PR #570, opened 2018-01-24
Select all pods, so pods in state Unknown are not deleted from the Cluster State.

schylek:
State Unknown is may happen due to transient error (https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase), so it shouldn't be deleted from the Cluster State.

### DIFF
--- a/vertical-pod-autoscaler/recommender/recommender.go
+++ b/vertical-pod-autoscaler/recommender/recommender.go
@@ -196,8 +196,7 @@ func newMetricsClient(config *rest.Config) cluster.MetricsClient {
 
 // Creates PodLister, listing only not terminated pods.
 func newPodLister(kubeClient kube_client.Interface) v1lister.PodLister {
-	selector := fields.ParseSelectorOrDie("status.phase!=" + string(apiv1.PodPending) + ",status.phase!=" + string(apiv1.PodUnknown))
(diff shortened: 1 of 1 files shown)


vertical-pod-autoscaler/recommender/recommender.go
```diff
@@ -196,8 +196,7 @@ func newMetricsClient(config *rest.Config) cluster.MetricsClient {
 
 // Creates PodLister, listing only not terminated pods.
 func newPodLister(kubeClient kube_client.Interface) v1lister.PodLister {
-	selector := fields.ParseSelectorOrDie("status.phase!=" + string(apiv1.PodPending) + ",status.phase!=" + string(apiv1.PodUnknown))
-	podListWatch := cache.NewListWatchFromClient(kubeClient.CoreV1().RESTClient(), "pods", apiv1.NamespaceAll, selector)
+	podListWatch := cache.NewListWatchFromClient(kubeClient.CoreV1().RESTClient(), "pods", apiv1.NamespaceAll, fields.Everything())
```
mwielgus: Why podPending selector is removed?

mwielgus: Ping

schylek:
I have accidentally removed phase!=Pending selector at first. However, I started to analyze if it could speedup delivery of a first recommendation to VPA object, so admission controller can see it earlier. Nevertheless it seems like a premature optimization, so I've restored this filter.

PR #574, opened 2018-01-24
Get CPU, memory and labels history from Prometheus

tkulczynski:
This also refactors out history provider from recommender.go (as you originally requested) and fixes issues with the recommender flags, client queries etc. I now tested it e2e in my cluster and it pushes some samples and labels to clusterState.

### DIFF
--- a/vertical-pod-autoscaler/recommender/Dockerfile
+++ b/vertical-pod-autoscaler/recommender/Dockerfile
@@ -17,4 +17,4 @@ MAINTAINER Karol Kraskiewicz "karol.kraskiewicz@gmail.com"
 
 ADD recommender recommender
 
-CMD ./recommender --v=4 --stderrthreshold=info
+CMD ./recommender --v=4 --stderrthreshold=info --prometheus-address=http://prometheus.monitoring.svc
--- a/vertical-pod-autoscaler/recommender/main.go
+++ b/vertical-pod-autoscaler/recommender/main.go
@@ -22,14 +22,16 @@ import (
 
 	"github.com/golang/glog"
 	kube_flag "k8s.io/apiserver/pkg/util/flag"
+	"k8s.io/autoscaler/vertical-pod-autoscaler/recommender/signals"
 	"k8s.io/client-go/rest"
 	kube_restclient "k8s.io/client-go/rest"
 )
 
 var (
-	namespace              = *flag.String("namespace", "default", `Namespace to manage`)
-	metricsFetcherInterval = *flag.Duration("recommender-interval", 1*time.Minute, `How often metrics should be fetched`)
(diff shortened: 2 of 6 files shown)


tkulczynski:
cc @mwielgus

vertical-pod-autoscaler/recommender/Dockerfile
```diff
@@ -17,4 +17,4 @@ MAINTAINER Karol Kraskiewicz "karol.kraskiewicz@gmail.com"
 
 ADD recommender recommender
 
-CMD ./recommender --v=4 --stderrthreshold=info
+CMD ./recommender --v=4 --stderrthreshold=info --prometheus-address=http://my-release-prometheus-server.default.svc
```
mwielgus: Can we pick some more reasonable prometheus server name and location? Like 
namespace: monitoring
service: prometheus
cc: @piosz

vertical-pod-autoscaler/recommender/main.go
```diff
@@ -22,22 +22,23 @@ import (
...
+	"k8s.io/autoscaler/vertical-pod-autoscaler/recommender/signals"
 	"k8s.io/client-go/rest"
 	kube_restclient "k8s.io/client-go/rest"
 )
 
 var (
-	namespace              = *flag.String("namespace", "default", `Namespace to manage`)
-	metricsFetcherInterval = *flag.Duration("recommender-interval", 1*time.Minute, `How often metrics should be fetched`)
-	prometheusAddress      = *flag.String("prometheus-address", "", `Where to reach for Prometheus metrics`)
+	namespace              = flag.String("namespace", "default", `Namespace to manage`)
```
mwielgus: namespace of what? VPA recommenders are expected to be global.

tkulczynski: Completely agree, but it's not the scope of this change. The flags were already here, I just noticed that with the star here it doesn't work (takes the default value before initializing the flag with value from the command line).

mwielgus: ok. Please add a todo to fix it.

tkulczynski: Done

vertical-pod-autoscaler/recommender/Dockerfile
piosz: SGTM though it's not a final commitment. Consider making this configurable via a flag.

tkulczynski: Done

mwielgus:
Please resolve conflicts.

tkulczynski:
Rebased.

PR #575, opened 2018-01-24
Tighten CF ASG permissions for prod systems

domhauton:
The tighter CF only allows modifications to the ASGs specified by the users. A much safer alternative to the previous.

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/README.md
+++ b/cluster-autoscaler/cloudprovider/aws/README.md
@@ -26,7 +26,7 @@ A minimum IAM policy would look like:
 }
 ```
 
-If you'd like to auto-discover node groups by specifing the `--node-group-auto-discover` flag, a `DescribeTags` permission is also required. We highly recommend you specify the ASGs explicitly in your IAM Roles for productions systems. The autoscaler will still be able to provide
+If you'd like to auto-discover node groups by specifing the `--node-group-auto-discover` flag, a `DescribeTags` permission is also required. We highly recommend you specify the ASGs explicitly in your IAM Roles for production systems. The autoscaler will only be able to modify ASGs you give it access to.
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


MaciekPytel:
@mumoshu Can you take a look?

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -26,29 +26,37 @@ A minimum IAM policy would look like:
...
     "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": [
+                "autoscaling:SetDesiredCapacity",
+                "autoscaling:TerminateInstanceInAutoScalingGroup"
+            ],
+            "Resource": [
+                "arn:aws:autoscaling:<REGION>:<ACCOUNTID>:autoScalingGroup:*:autoScalingGroupName/<ASG-1-NAME>",
+                "arn:aws:autoscaling:<REGION>:<ACCOUNTID>:autoScalingGroup:*:autoScalingGroupName/<ASG-2-NAME>"
```
mumoshu: I'd rather like to set `Condition` instead, so that we don't need to repeat ASG names in the IAM policy - my original intention to introduce auto-discovery was to reduce repetitions!

```yaml
                 "Condition": {
                    "Null": { "autoscaling:ResourceTag/kubernetes.io/cluster/{{.ClusterName}}": "false" }
                  }
```

Note: replace `{{.ClusterName}}` with your cluster name.

Actually, this is what kube-aws has been using since disasters of CA+Auto-Discovery terminating nodes unrelated to the k8s cluster.

domhauton:
@mumoshu Updated to use conditions as well. Tested and it works.

I left the resource name as wildcarded. We prefer it internally for safety.

domhauton:
@MaciekPytel @mumoshu Any luck?

szuecs:
@domhauton in general I would prefer that you add a separate section. I am not sure if I would run it like you suggest. I like it more simple and don't allow mixed plain EC2 and kubernetes workloads. 
Maybe a section about AWS account setup with mixed workloads (Kubernetes+EC2) makes more sense.

PR #77, opened 2017-05-17
cluster-autoscaler: Resource Slack Proposal

mumoshu:
This is a proposal for #56 
Any comment is appreciated.
@mwielgus @MaciekPytel Could you take a look into this?

PR #587, opened 2018-01-29
Burst capacity

integrii:
Here is a proposal for burstable headroom.  Please review.  The goal here is to have the autoscaler also factor in "burstable headroom" when deciding to change the scale of a cluster.

We really appreciate the project and use it in many production clusters globally.

### DIFF
--- a/cluster-autoscaler/proposals/burst_capacity.md
+++ b/cluster-autoscaler/proposals/burst_capacity.md
@@ -1,0 +1,5 @@
+https://github.com/kubernetes/autoscaler/issues/148
+
+https://github.com/kubernetes/autoscaler/pull/77
+
+https://github.com/kubernetes/autoscaler/pull/56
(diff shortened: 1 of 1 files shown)


MaciekPytel:
As you noticed this has been discussed in #77. It turned out that this is very hard. Some part of it was purely technical reasons: CA works by binpacking pods using scheduler logic. At no point does it calculate anything like total required resources, so there is nothing to add those extra resources to. We have a general idea for how to workaround this by injecting non-existing 'placeholder' pods into CA simulation. However, this is far from trivial to implement.

An even more serious problem (that ultimately stopped us from doing this) is how to specify the extra capacity? CPU and memory are just one of many things taken into account in scheduling. Having enough mem/CPU doesn't mean a new pod will be able to schedule - there are many other constraints (ex. hostports, affinity). Finally CA has no API - would this be specified as flags? How would they look like exactly?

===

As it stands I would consider this a feature request. The way we see proposals is as a technical document describing how all non-trivial issues will be solved (basically developer describing how he will implement a feature before implementing it).

Finally we think the issue can now be solved by user with https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/ (this is in alpha). Just create low-priority placeholder pods that will be evicted as soon as your real pods show up.

integrii:
Thanks for explaining the issue and reading my proposal so quickly.  I will take a deeper look into the codebase and get back to you...

mwielgus:
Closing the PR for now ...

PR #590, opened 2018-01-29
Use real recommender logic (based on usage).

kgrygiel:


kgrygiel:
cc @tkulczynski

vertical-pod-autoscaler/recommender/model/types.go
```diff
@@ -50,6 +50,11 @@ func CoresFromCPUAmount(cpuAmunt ResourceAmount) float64 {
 	return float64(cpuAmunt) / 1000.0
 }
 
+// QuantityFromCPUAmount converts CPU cores to a resource.Quantity.
```
tkulczynski: Cores? I tihnk millicores, no?

vertical-pod-autoscaler/examples/hamster-vpa.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: "poc.autoscaling.k8s.io/v1alpha1"
```
tkulczynski: I think hamster-deployment and hamster-vpa might be a single yaml file, unless you really want to use them separately which I doubt.

vertical-pod-autoscaler/recommender/model/types.go
kgrygiel: Done. Changed to "ResourceAmount". It shouldn't mention units at all.

vertical-pod-autoscaler/examples/hamster-vpa.yaml
kgrygiel: Done.

tkulczynski:
Thanks, LGTM now, handing over to @bskiba.

bskiba:
I missed this before, but are examples/ and samples/ fundamentally different or we can merge them?

kgrygiel:
I missed that too. I'm going to merge it. Thanks.

PR #591, opened 2018-01-30
Improvements to deploy scripts

tkulczynski:
Also unify default registries and tags in all vpa components.


tkulczynski:
cc @schylek @bskiba

PR #594, opened 2018-01-31
Write some READMEs for VPA

tkulczynski:
cc @kgrygiel

### DIFF
--- a/vertical-pod-autoscaler/README.md
+++ b/vertical-pod-autoscaler/README.md
@@ -1,0 +1,120 @@
+# Vertical Pod Autoscaler
+
+# Intro
+
+Vertical Pod Autoscaler (VPA) frees the users from necessity of setting
+up-to-date resource requests for their containers in pods.
+When configured, it will set the requests automatically based on usage and
+thus allow proper scheduling onto nodes so that appropriate resource amount is
+available for each pod.
+
+# For users
+
+### Installation
+
+**Prerequisites** (to be automatized):
+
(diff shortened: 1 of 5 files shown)


vertical-pod-autoscaler/README.md
```diff
@@ -0,0 +1,120 @@
...
+    updateMode: "Auto"
+```
+
+### Known limitations
+
+* The VPA admission controller is an admission webhook. The order of admission
+  controllers is defined by flag on APIserver.
+  VPA admission controller might have **relations to other admission controllers**,
+  e.g. running it after quota admission might cause making incorrect decisions.
+* Out-of-memory events / pod evictions are not taken into account for memory
```
mwielgus: not taken YET

vertical-pod-autoscaler/README.md
```diff
@@ -0,0 +1,120 @@
...
+  name: my-app-vpa
+spec:
+  selector:
+    matchLabels:
+      app: my-app
+  updatePolicy:
+    updateMode: "Auto"
+```
+
+### Known limitations
```
mwielgus: Limitations of the alpha versionion

vertical-pod-autoscaler/README.md
tkulczynski: Done

vertical-pod-autoscaler/README.md
tkulczynski: Done

PR #596, opened 2018-01-31
metrics and specs clients package refactoring

KarolKraskiewicz:
Proposal of a simple refactoring.
I'm not 100% sure about repeating package name in Interface name like: `spec.SpecClient`, however having two different `Client` interfaces imported in a single file also seemed a little confusing.

Let me know if it seems good to you or if you suggest a different approach.

CC: @kgrygiel @mwielgus

### DIFF
--- a/vertical-pod-autoscaler/recommender/input/history/history_provider.go
+++ b/vertical-pod-autoscaler/recommender/input/history/history_provider.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package signals
+package history
 
 import (
 	"fmt"
--- a/vertical-pod-autoscaler/recommender/input/history/history_provider_test.go
+++ b/vertical-pod-autoscaler/recommender/input/history/history_provider_test.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package signals
+package history
 
 import (
 	"fmt"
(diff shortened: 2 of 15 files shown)


tkulczynski:
Thanks for this, some refactoring is definitely needed!

I'm not sure about the naming though. There are currently four inputs:
1. VPA lister, used directly in recommender.go
2. Pod specs from spec_client.go
3. Real time metrics from metrics_client.go
4. Historical metrics from history_provider.go (that also used prometheus*.go underneath).

I would put 3 & 4 into signals/ directory (possibly in subdirs), but not necessarily 1 & 2. We can also put 2, 3 & 4 in one common directory like you're doing here, but let's not call it signals then (I'm thinking something like data_inputs summarizes what they all are better, but am open to better suggestions) and let's have three subdirs then and nothing in the top level dir.

What do you guys think?

KarolKraskiewicz:
Hi @tkulczynski, thanks for your feedback!

I like the idea of moving 2, 3 & 4 into separate dirs under signals and renaming it so something like input_data (I renamed it to "input" but I'm open for other suggestions ;) ). I was thinking spec_client fits neatly to the rest, mostly because in the future pods/containers specs will have the ability to be updated in the real-time. It is now all added to the PR. 

VPA lister is now so minimalistic, so I agree it sounds completely fair to keep it where it is for now.

tkulczynski:
Thanks for the changes! Looks good to me now.

BTW, please resolve the conflicts.

KarolKraskiewicz:
rebased.

tkulczynski:
LGTM, cc @mwielgus to approve.

PR #461, opened 2017-11-10
Consider GPU nodes unready until allocatable GPU is > 0

MaciekPytel:
Fix for https://github.com/kubernetes/kubernetes/issues/54959.

The idea is to internally override status of nodes which should have GPUs to unready until they actually have allocatable GPU. That way our normal logic for handling long starting up nodes kicks in and counts those nodes in simulations.

The downside is we have to use the same timeout for GPU node start-up + driver installation as for normal nodes (defaults to 15 minutes, which may not be enough - we may need to temporarily increase this for 1.8).

Also fixes the issue where GCP pricing model was using old GPU resource name and not taking GPU into account when calculating node cost (leading to very sad scale-up decisions in my test...).

PR #597, opened 2018-01-31
allow specifying arbitrary resource capacity via aws asg tags

peter-bitfusion:
We're starting to work with device plugins and want to be able to scale up node groups from 0 when an unschedulable pod requests custom resources provided by a device plugin

for example my pod requests
```
resources:
  requests:
    acme.inc/widget: 5
```

This changeset allows adding aws asgs tags to specify additional resources capacity, using a autoscaler prefix similar to how node labels are specified. example tags: 
```
Key: k8s.io/cluster-autoscaler/node-template/resource-capacity/acme.inc/widget
Value: 5
```
Is this an acceptable approach?

Making the PR against master. Please advise if master isn't the right branch.

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_manager.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_manager.go
@@ -406,6 +406,7 @@ func (m *AwsManager) buildNodeFromTemplate(asg *Asg, template *asgTemplate) (*ap
 	node.Status.Capacity[apiv1.ResourceCPU] = *resource.NewQuantity(template.InstanceType.VCPU, resource.DecimalSI)
 	node.Status.Capacity[apiv1.ResourceNvidiaGPU] = *resource.NewQuantity(template.InstanceType.GPU, resource.DecimalSI)
 	node.Status.Capacity[apiv1.ResourceMemory] = *resource.NewQuantity(template.InstanceType.MemoryMb*1024*1024, resource.DecimalSI)
+	node.Status.Capacity = cloudprovider.JoinResourceLists(node.Status.Capacity, extractCapacityFromAsg(template.Tags))
 
 	// TODO: use proper allocatable!!
 	node.Status.Allocatable = node.Status.Capacity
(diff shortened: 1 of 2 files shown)


mwielgus:
cc: @sethpollack @mumoshu

MaciekPytel:
I suspect this may run into the same issue as https://github.com/kubernetes/kubernetes/issues/54959.

The problem is when new node is added, CA knows it is 'upcoming' and it takes it knows some pods will go on that node once it boots up. So it won't trigger another scale-up for those pods. So far all is good. Enter device plugins. Unless something changed in the last few weeks the plugin triggers _after_ the node becomes ready and the extra resource required by pod will show up in node spec _even later_ (once the plugin is done). By that time CA will see that it was wrong and the pods for which it triggered scale-up can't actually go to a node it added for them (because it doesn't have the special resource yet and CA doesn't know it will be added later). So CA will go and create more nodes.

I added a hack for handling this specifically for GPUs on GKE (https://github.com/kubernetes/autoscaler/pull/461). However, this hack relies on CA being able to identify the nodes with GPUs by knowing some labels that GKE adds to those nodes. This is by no means a generic solution. We need to talk with owners of device plugins and figure out a proper fix before device plugins beta.

cc: @jiayingz @vishh

jiayingz:
I am not sure how aws CA works, but taking a quick look at the PR, it seems you directly updated the node status resource capacity to reflect the upcoming availability of the resource to be exported by device plugin. This  should allow the pod to be scheduled on the new node. However, there will be a time gap for your device plugin to establish communication with Kubelet so that Kubelet can make proper calls during container start. How long this time gap is will depend on how your device plugin is deployed on the node and what setup operations your device plugin needs to perform to make the "acme.inc/widget" resource consumable on the node. On GKE GPU case, this takes some non-neglectable time, and that is why @MaciekPytel needs to add some special handling to handle this time gap. Agree with @MaciekPytel that the current solution is not ideal and we may think about a general way to handle this resource exporting time gap.

peter-bitfusion:
@jiayingz @MaciekPytel I think I understand why extra nodes are scaled up.
It sounds like it would be helpful if nodes could be marked as NotReady until all their device plugins are Ready. However then you couldn't schedule pods that need X devices until Y devices were also available. It would be helpful if nodes exposed what device plugins were present and their statuses. Then CA could use that info to consider a node not ready yet for pods that require resources provided by device plugins that are present on that node but not ready. Perhaps CA could infer that information by looking at daemonsets that should be scheduled on a node? Would have to somehow know if a daemonset represents a device plugin though

MaciekPytel:
Yup, what we need is for device plugin to provide some information as to what extra resources will show up. We can't get that from daemonsets unless there is some standardized field or annotation or something we could look at. And even if there is how would we know how much of a given resource is on the node (ex. how many GPUs)?

There are many ways this information could be exposed (example: set capacity but not allocatable until resource is ready, have a CRD for each custom resource describing how to recognize node with this resource, worst-case some standardized annotations). However, until we come up with something like that I don't think we can fix it for generic use-case in CA.

jiayingz:
/cc @bsalamat @dchen1107 @thockin 

Agree we need some way for CA to know how it should config the new node to satisfy the "extended-resource" requests from the pending pod, and also be able to tell what resources will show up on the newly created node and how long it should wait. As @MaciekPytel mentioned, there could be different ways to expose this information. I cc-ed some folks who may be interested in this discussion. Right now, we probably don't have a good solution for this problem, and CA probably need to use some kind of timer to deal with this time gap.

PR #602, opened 2018-02-01
Remove old unregistered nodes before checking cluster healthiness

### LINKED ISSUE #547: Fix situation when scale up overshot 33% broken nodes while hitting qouta

bskiba:
Fixes #547

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -173,16 +173,6 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 				status.GetReadableString(), a.AutoscalingContext.LogRecorder)
 		}
 	}()
-	if !a.ClusterStateRegistry.IsClusterHealthy() {
-		glog.Warning("Cluster is not ready for autoscaling")
-		scaleDown.CleanUpUnneededNodes()
-		autoscalingContext.LogRecorder.Eventf(apiv1.EventTypeWarning, "ClusterUnhealthy", "Cluster is unhealthy")
-		return nil
-	}
-
-	metrics.UpdateDurationFromStart(metrics.UpdateState, runStart)
-	metrics.UpdateLastTime(metrics.Autoscaling, time.Now())
-
 	// Check if there are any nodes that failed to register in Kubernetes
 	// master.
 	unregisteredNodes := a.ClusterStateRegistry.GetUnregisteredNodes()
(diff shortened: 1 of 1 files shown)


MaciekPytel:
Maybe we should also change UpdateClusterStateMetrics? It seems it will still report that cluster is not ready for autoscaling for a while after each failed scale-up (IIRC it takes a few loops to delete unregistered nodes)?

MaciekPytel:
Can you describe how you tested that?

bskiba:
Tested following:

1. Start cluster with num_nodes=23, min=23, max=200, quota allows for 23 nodes. 
2. Spin up a workload that triggers scale up to 82 nodes
3. Wait 15 minutes during which we have 59 upcoming nodes. [ScaleUp:     InProgress (ready=23 cloudProviderTarget=82)]
4. After 15 minutes FailedScaleUp event is emitted, node group is backed off [ ScaleUp:     Backoff (ready=23 cloudProviderTarget=82)], unregistered nodes are removed, we have 0 upcoming nodes and no expansion options. Cluster is in a healthy state.

case: removing load leaves cluster in healthy state
5. Delete the workload
6. State changes to [ScaleUp:     Backoff (ready=23 cloudProviderTarget=23)]
7. After the scale up backoff period the state changes to [ ScaleUp:     NoActivity (ready=23 cloudProviderTarget=23)]

case: raising the quota leaves cluster in healthy state
5. Raise quota to allow for 82 nodes.
6. Wait for the backoff period to pass 
7. After the backoff autoscaler tries scaling up again. [ ScaleUp:     InProgress (ready=23 cloudProviderTarget=82)]
8. Scale up succeeds [ ScaleUp:     NoActivity (ready=82 cloudProviderTarget=82)]

Let me know if there's any other scenario you would like me to test.

bskiba:
> Maybe we should also change UpdateClusterStateMetrics? It seems it will still report that cluster is not ready for autoscaling for a while after each failed scale-up (IIRC it takes a few loops to delete unregistered nodes)?

I haven't observed this. Since the nodes are upcoming for >15minutes, the unregistered nodes are removed in the very next loop iteration AFAIU, but let me double check it.

MaciekPytel:
Hmm, I think they're 'removed' each iteration until they're actually deleted by underlying MIG. I guess I may remember it taking multiple iterations when I was testing failure scenarios with kubelet failing to register (ie. when there was an actual VM that needed to be deleted).

bskiba:
In that case I think it's reasonable to move UpdateClusterStateMetrics after the "removing unregistered nodes part" to make sure this doesn't happen.

MaciekPytel:
Nah, let's not do that. This way we don't update them at all. I guess cluster does actually get into unhealthy state, even if just for a single loop and it makes sense for metric to reflect that.

Sorry for causing confusion. Not enough coffeine in my system yet.

PR #612, opened 2018-02-03
Refactoring: loading ClusterState

KarolKraskiewicz:
Comming back with the idea of "ClusterStateFeeder", now in a different form. 
The goal of this refactoring is to:
1. encapsulate logic which loads data into clusterState
2. limit the number of Recommender dependencies

@kgrygiel for a while, I was playing around with the idea of making feeder stateless, but then Recommender would stay with a number of dependencies, which would be only passed as parameters to external functions.

@kgrygiel @mwielgus - please review!

(Please merge https://github.com/kubernetes/autoscaler/pull/596 before review, as this PR depends on it)

### DIFF
--- a/vertical-pod-autoscaler/recommender/input/cluster_feeder.go
+++ b/vertical-pod-autoscaler/recommender/input/cluster_feeder.go
@@ -1,0 +1,211 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package input
+
+import (
+	"time"
+
+	"github.com/golang/glog"
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
+	vpa_clientset "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/client/clientset/versioned"
+	vpa_api "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/client/clientset/versioned/typed/poc.autoscaling.k8s.io/v1alpha1"
+	vpa_lister "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/client/listers/poc.autoscaling.k8s.io/v1alpha1"
+	vpa_api_util "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/vpa"
+	"k8s.io/autoscaler/vertical-pod-autoscaler/recommender/input/history"
+	"k8s.io/autoscaler/vertical-pod-autoscaler/recommender/input/metrics"
+	"k8s.io/autoscaler/vertical-pod-autoscaler/recommender/input/spec"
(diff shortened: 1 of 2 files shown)


KarolKraskiewicz:
rebased to improve readability

vertical-pod-autoscaler/recommender/input/cluster_feeder.go
```diff
@@ -0,0 +1,191 @@
...
+// ClusterStateFeeder can update state of ClusterState object.
+type ClusterStateFeeder interface {
+
+	// LoadHistory loads historical pod spec and metrics into clusterState.
+	LoadHistory(clusterState *model.ClusterState)
+
+	// LoadVPAs updtes clusterState with current state of VPAs.
+	LoadVPAs(clusterState *model.ClusterState)
+
+	// LoadPods updates slusterState with current specification of Pods theirs Containers.
```
tkulczynski: Pods and their Containers.

vertical-pod-autoscaler/recommender/recommender.go
```diff
@@ -256,19 +125,26 @@ func createPodResourceRecommender() logic.PodResourceRecommender {
...
 		vpaClient:               vpa_clientset.NewForConfigOrDie(config).PocV1alpha1(),
-		vpaLister:               vpa_api_util.NewAllVpasLister(vpa_clientset.NewForConfigOrDie(config), make(chan struct{})),
 		podResourceRecommender:  createPodResourceRecommender(),
 	}
 	glog.V(3).Infof("New Recommender created %+v", recommender)
 
 	return recommender
 }
 
+func newClusterStateFeeder(config *rest.Config, historyProvider history.HistoryProvider) input.ClusterStateFeeder {
```
tkulczynski: There are two [nN]ewClusterStateFeeder methods. Can you please merge them? Also, how about moving the creation of the clients out from recommender? Ideally, recommender shouldn't have to pull in the input/history, input/metrics, input/spec packages, because it already has clusterFeeder that does everything, right?

vertical-pod-autoscaler/recommender/input/cluster_feeder.go
```diff
@@ -0,0 +1,191 @@
...
+	"k8s.io/autoscaler/vertical-pod-autoscaler/recommender/input/metrics"
+	"k8s.io/autoscaler/vertical-pod-autoscaler/recommender/input/spec"
+	"k8s.io/autoscaler/vertical-pod-autoscaler/recommender/model"
+)
+
+// ClusterStateFeeder can update state of ClusterState object.
+type ClusterStateFeeder interface {
+
+	// LoadHistory loads historical pod spec and metrics into clusterState.
+	LoadHistory(clusterState *model.ClusterState)
```
tkulczynski: clusterState is an argument to every method (and in general needed for the feeder to be useful at all), so why not pass it in constructor and make a field in feeder struct?

vertical-pod-autoscaler/recommender/recommender.go
KarolKraskiewicz: Thanks for catching the two methods.
I also moved all the clients/lister creation out of recommender, leaving it much skinnier :)

vertical-pod-autoscaler/recommender/input/cluster_feeder.go
KarolKraskiewicz: I was not in my original feeder but I kind of liked the idea of passing clusterState as a param, because it made it obvious the certain method would load data to a particular instance of a clusterState. The argument about passing the same object over and over again as a param was good though, so I moved clusterState back into the Feeder. See for yourself how you like it.

bskiba:
Looks like there's conflicts :(
Otherwise, waiting for lgtm from @tkulczynski

tkulczynski:
I like the change. After the merge/rebase, I'd really like to see the diff on the moved code is minimal. Other than that it looks good.

KarolKraskiewicz:
rebased!

tkulczynski:
LGTM

PR #614, opened 2018-02-05
Fix reference to the docker image in the example

ankon:
It seems 1.8 doesn't exist, but 1.8.0 does -- and there is a 1.8.1 already.

----
Note that on the _1.8_ branch pointed to by the README the example actually talks about _1.7_, so I'm somewhat assuming that master is more "up-to-date" than that branch.

### DIFF
--- a/addon-resizer/README.md
+++ b/addon-resizer/README.md
@@ -57,7 +57,7 @@ spec:
         kubernetes.io/cluster-service: "true"
     spec:
       containers:
-        - image: k8s.gcr.io/addon-resizer:1.8
+        - image: k8s.gcr.io/addon-resizer:1.8.1
           imagePullPolicy: Always
           name: pod-nanny
           resources:


bskiba:
@x13n can you take a look if this is what we want?

x13n:
1.8.1 is what we use in main k8s repo e.g. for GCE, but I don't think this will work after @kawych changes. Karol, can you help figuring out how the config should look like now?

kawych:
I think this will fail due to a bug that was noticed and fixed in PR #559 
Currently it requires a config map to be mounted, example configuration for Heapster: https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/cluster-monitoring/stackdriver/heapster-controller.yaml

/cc @MaciekPytel @mwielgus 
We probably want to release this bugfix as v1.8.2. I suggest to update the example only when the bugfix is released.

ankon:
Hmm. I think I'm missing something -- are you saying that 1.8.1 is broken, and one should use 1.8.0? In that case I think I should update this PR to refer to `1.8.0`. The point of this PR is to fix the completely broken `image` value, as it right now points to something non-existing.

> We probably want to release this bugfix as v1.8.2. I suggest to update the example only when the bugfix is released.
When you have a 1.8.2, then yes, please do update the example :)

x13n:
1.8.1 isn't entirely broken, it just requires a specific flag which is not provided in the example.

PR #628, opened 2018-02-07
replace deprecated maintainer instruction with label

andyxning:
As Dockerfile `MAINTAINER` instruction has been deprecated, we should use `LABEL maintainer=xxx` to replace original `MAINTAINER`. 

/cc @bskiba @mwielgus

### DIFF
--- a/addon-resizer/Dockerfile
+++ b/addon-resizer/Dockerfile
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 FROM BASEIMAGE
-MAINTAINER Quintin Lee "qlee@google.com"
 
 COPY pod_nanny /
 


andyxning:
Ping @Q-Lee @piosz @wojtek-t

addon-resizer/Dockerfile
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 FROM BASEIMAGE
-MAINTAINER Quintin Lee "qlee@google.com"
+LABEL maintainer="Quintin Lee <qlee@google.com>"
```
piosz: Please just remove it.

andyxning: IIUC, we should remove this completely? @piosz

PR #629, opened 2018-02-07
Add release notes for 1.1.0 and 1.1.1 to readme

### LINKED ISSUE #586: cluster-autoscaler release for kubernetes 1.9.x missing
https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler#releases does not list a kubernetes 1.9.x release.

bskiba:
fixes #586

### DIFF
--- a/cluster-autoscaler/README.md
+++ b/cluster-autoscaler/README.md
@@ -19,6 +19,7 @@ there is a big chance that it won't work as expected.
 
 | Kubernetes Version  | CA Version   |
 |--------|--------|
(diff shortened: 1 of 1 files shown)


MaciekPytel:
Adding 1.1.1, but not 1.0.4 (and not a single patch release for 0.6) seems inconsistent. Maybe we should remove patch releases from readme? It made sense around 0.5 when we only had a few releases, but it's starting to grow. Maybe we can just do 0.6, 1.0, 1.1, 1.2 and add a link to github release page if someone wants to check patch releases or something?

PR #633, opened 2018-02-07
Support for Conditions in VPA.

kgrygiel:


mwielgus:
!!! 'gofmt -s' needs to be run on the following files: 
./vertical-pod-autoscaler/pkg/utils/vpa/api.go

vertical-pod-autoscaler/pkg/utils/vpa/api.go
```diff
@@ -49,32 +50,34 @@ func patchVpa(vpaClient vpa_api.VerticalPodAutoscalerInterface, vpaName string,
...
-		Value: vpa_types.VerticalPodAutoscalerStatus{},
-	},
-	})
-}
+// UpdateVpaStatus updates the status field of the VPA API object.
+// It prevents race conditions by verifying that the lastUpdateTime of the
+// API object and its model representation are equal.
+func UpdateVpaStatus(vpaClient vpa_api.VerticalPodAutoscalerInterface, vpa *model.Vpa) (result *vpa_types.VerticalPodAutoscaler, err error) {
+	status := vpa_types.VerticalPodAutoscalerStatus{
+		LastUpdateTime: metav1.Now(),
```
tkulczynski: To be perfectly sure this solves races, you need to check that each patch strictly increases LastUpdateTime. While this should be true in sane worlds (but not e.g. if different nodes in the cluster have desynchronized clocks), I would add the if statement .

vertical-pod-autoscaler/pkg/utils/vpa/api.go
```diff
@@ -49,32 +50,34 @@ func patchVpa(vpaClient vpa_api.VerticalPodAutoscalerInterface, vpaName string,
...
+		status.Recommendation = *vpa.Recommendation
+	}
 
-// UpdateVpaRecommendation updates VPA's status/recommendation object and status/lastUpdateTime.
-func UpdateVpaRecommendation(vpaClient vpa_api.VerticalPodAutoscalerInterface, vpaName string, recommendation vpa_types.RecommendedPodResources) (result *vpa_types.VerticalPodAutoscaler, err error) {
-	patches := []patchRecord{
-		{
-			Op:    "add",
+	patches := make([]patchRecord, 0)
+	if !vpa.LastUpdateTime.IsZero() {
```
tkulczynski: Why not test in that case? There's a race for setting the first status then, no?

kgrygiel: Unfortunately I cannot set a test on a missing field in the patch method.
On the other hand we can make the assumption that the Updater (or other agents) don't do any operations on a VPA that doesn't have the Status field. In particular it has no recommendation at this point, so the Updater can't do anything useful. WDYT?

kgrygiel: Done (discussed offline, if the lastUpdateTime is missing I set a test on the whole status field).

vertical-pod-autoscaler/pkg/utils/vpa/api.go
kgrygiel: Done.

tkulczynski:
LGTM.

@bskiba: could you approve?

PR #636, opened 2018-02-08
Merge admission controller RBAC into main vpa-rbac file.

schylek:


vertical-pod-autoscaler/hack/vpa-process-yamls.sh
```diff
@@ -38,7 +38,7 @@ if [ $# -gt 2 ]; then
   exit 1
 fi
 
-YAMLS="api/vpa-crd.yaml deploy/vpa-rbac.yaml deploy/admission-controller-rbac.yaml deploy/updater-deployment.yaml deploy/recommender-deployment.yaml deploy/admission-controller-deployment.yaml"
```
bskiba: I think you meant to remove the admission-contraoller-rbac.yaml?

schylek: Good catch. Thanks!

PR #637, opened 2018-02-08
Scripts for handing E2E testing in test-infra

schylek:


vertical-pod-autoscaler/hack/deploy-for-e2e.sh
```diff
@@ -0,0 +1,43 @@
...
+
+COMPONENTS="recommender updater admission-controller"
+
+if [ $# -ne 0 ]; then
+  COMPONENTS=$*
+fi
+
+for i in ${COMPONENTS}; do
+  (
+    cd ${SCRIPT_ROOT}/${i} && make release
```
tkulczynski: Hint:
```
make --directory ${i} release
```
It's a nice shortcut that also goes back to your previous directory afterwards.

vertical-pod-autoscaler/hack/vpa-process-yamls.sh
```diff
@@ -14,9 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -o errexit
```
tkulczynski: The [no]errexit strategy is confusing to me, especially in nested scripts calling scripts. I would be glad to have it set always and instead use "|| true" on the commands that we actually expect to fail and are ok with that.

schylek: Redinements applied.

tkulczynski:
Thanks, LGTM now.

PR #639, opened 2018-02-09
Use Debian base to build Cluster Autoscaler image

aleksandra-malinowska:
Use Debian base to build Cluster Autoscaler image.

### DIFF
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -12,19 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM gcr.io/google-containers/ubuntu-slim:0.8
-MAINTAINER Marcin Wielgus "mwielgus@google.com"
+FROM golang:1.8.3
+LABEL maintainer="Marcin Wielgus <mwielgus@google.com>"
 
-RUN apt-get update && apt-get install --yes git wget make gcc libseccomp-dev \
-  && apt-get clean \
-  && rm -rf /var/lib/apt/lists/*
-RUN wget https://storage.googleapis.com/golang/go1.8.3.linux-amd64.tar.gz \
-   && tar -xvf go1.8.3.linux-amd64.tar.gz \
-   && rm go1.8.3.linux-amd64.tar.gz
-
-ENV GOROOT /go
 ENV GOPATH /gopath/
-ENV PATH $GOPATH/bin:$GOROOT/bin:$PATH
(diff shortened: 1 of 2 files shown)


builder/Dockerfile
```diff
@@ -12,11 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM gcr.io/google-containers/ubuntu-slim:0.8
-MAINTAINER Marcin Wielgus "mwielgus@google.com"
+FROM gcr.io/google-containers/debian-base-amd64:0.1
```
tallclair: Please use 0.3 (the latest)

aleksandra-malinowska: Done

cluster-autoscaler/Dockerfile
```diff
@@ -12,11 +12,11 @@
...
-FROM gcr.io/google-containers/ubuntu-slim:0.8
-MAINTAINER Marcin Wielgus "mwielgus@google.com"
+FROM gcr.io/google-containers/debian-base-amd64:0.3
+LABEL maintainer="Marcin Wielgus <mwielgus@google.com>"
 
-RUN apt-get update && apt-get install --yes ca-certificates \
-  && apt-get clean \
+ENV DEBIAN_FRONTEND noninteractive
+RUN clean-install ca-certificates \
   && rm -rf /var/lib/apt/lists/*
```
tallclair: clean-install handles this, no need to duplicate here.

builder/Dockerfile
```diff
@@ -12,11 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM gcr.io/google-containers/ubuntu-slim:0.8
-MAINTAINER Marcin Wielgus "mwielgus@google.com"
+FROM gcr.io/google-containers/debian-base-amd64:0.3
```
tallclair: The slim-ness of the build image probably doesn't matter as much - have you considered just using the official go images? (e.g. golang:1.8.3). This is how kubernetes releases [are built](https://github.com/kubernetes/kubernetes/blob/master/build/build-image/cross/Dockerfile).

aleksandra-malinowska: Done

cluster-autoscaler/Dockerfile
aleksandra-malinowska: removed

PR #640, opened 2018-02-09
Disable checking inter pod affinity predicate if not necessary

aleksandra-malinowska:
Add disabling inter pod affinity predicate if only preferred / node affinity used.

### DIFF
--- a/cluster-autoscaler/core/utils.go
+++ b/cluster-autoscaler/core/utils.go
@@ -412,23 +412,38 @@ func getPotentiallyUnneededNodes(context *AutoscalingContext, nodes []*apiv1.Nod
 	return result
 }
 
+func hasHardInterPodAffinity(affinity *apiv1.Affinity) bool {
+	if affinity == nil {
+		return false
+	}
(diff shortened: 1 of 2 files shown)


aleksandra-malinowska:
This has been tested manually for the following scenarios:
- deployment has pods using hard anti-affinity, verified that new nodes were added
- deployment has pods using soft anti-affinity, verified that predicate disabled log info is present, no new nodes are added as long as there's sufficient space on existing ones

cluster-autoscaler/core/utils.go
```diff
@@ -412,23 +412,38 @@ func getPotentiallyUnneededNodes(context *AutoscalingContext, nodes []*apiv1.Nod
 	return result
 }
 
+func hasInterPodAffinity(affinity *apiv1.Affinity) bool {
```
mwielgus: hasInterPod *HARD* Affinity

aleksandra-malinowska: Done

PR #676, opened 2018-02-21
Make requests smaller or equal to limits

tkulczynski:
This also extracts common code from updater & admission controller to a library.

### DIFF
--- a/vertical-pod-autoscaler/admission-controller/logic/recommendation_provider.go
+++ b/vertical-pod-autoscaler/admission-controller/logic/recommendation_provider.go
@@ -41,74 +41,18 @@ func NewRecommendationProvider(vpaLister vpa_lister.VerticalPodAutoscalerLister)
 	return &recommendationProvider{vpaLister: vpaLister}
 }
 
-// getRecomendedResources overwrites pod resources Request field with recommended values.
+// getRecomendedResources returns the recommended resources Request for each container in the given pod.
 func getRecomendedResources(pod *v1.Pod, podRecommendation vpa_types.RecommendedPodResources, policy vpa_types.PodResourcePolicy) []v1.ResourceList {
 	res := make([]v1.ResourceList, len(pod.Spec.Containers))
 	for i, container := range pod.Spec.Containers {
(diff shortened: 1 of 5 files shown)


vertical-pod-autoscaler/admission-controller/logic/recommendation_provider.go
```diff
@@ -41,76 +41,20 @@ func NewRecommendationProvider(vpaLister vpa_lister.VerticalPodAutoscalerLister)
 	return &recommendationProvider{vpaLister: vpaLister}
 }
 
-// getRecomendedResources overwrites pod resources Request field with recommended values.
+// getRecomendedResources returns the recommended resources Request for each container in the given pod.
```
kgrygiel: "in the same order they are specified in the pod.Spec"

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
```diff
@@ -0,0 +1,109 @@
...
+// GetCappedRecommendationForContainer returns a recommendation extracted for the given container, adjusted to obey policy and limits.
+func GetCappedRecommendationForContainer(
+	container apiv1.Container,
+	podRecommendation *vpa_types.RecommendedPodResources,
+	policy *vpa_types.PodResourcePolicy) (apiv1.ResourceList, error) {
+	containerRecommendation := getRecommendationForContainer(podRecommendation, container)
+	if containerRecommendation == nil {
+		return nil, fmt.Errorf("no recommendation available for container name %v", container.Name)
+	}
+	capRecommendationToContainerLimit(containerRecommendation, container)
```
kgrygiel: I think capping to the limit should be applied *after* the VPA policy. It normally shouldn't matter, but in the unlikely case of a conflict (currently: minAllowed > limit), capping to limit should take precedence. I guess the updater could set some error condition in such case (maybe a TODO for now?)

tkulczynski:
Applied all your comments, removed the glogs from vpa/capping.go completely. PTAL

tkulczynski:
@mwielgus could you please merge?

PR #681, opened 2018-02-24
Improve azure node's providerID handling

feiskyer:
This PR improves azure node's providerID handling by:

- Avoid converting cases for azure node's providerID 
- Use GET API to fetch node's ID for vmss virtual machines. This is because the virtual machines ID format in LIST is not consistent with GET results. We use GET here because azure cloud provider also uses GET API.


feiskyer:
@mwielgus Thanks

PR #684, opened 2018-02-26
Generated code for VPA Snapshot CRD object.

schylek:


schylek:
@kgrygiel PTAL pkg/apis/poc.autoscaling.k8s.io/v1alpha1/types.go

vertical-pod-autoscaler/pkg/apis/poc.autoscaling.k8s.io/v1alpha1/types.go
```diff
@@ -207,3 +207,64 @@ type RecommendedContainerResources struct {
...
+// VerticalPodAutoscalerSnapshotList is a list of VerticalPodAutoscalerSnapshot objects.
+type VerticalPodAutoscalerSnapshotList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata"`
+	Items           []VerticalPodAutoscalerSnapshot `json:"items"`
+}
+
+// VerticalPodAutoscalerSnapshotSpec is the specification of the snapshot object.
+type VerticalPodAutoscalerSnapshotSpec struct {
+	// Name of the VPA object that stored VerticalPodAutoscalerSnapshot object.
```
mwielgus: There should be content version information.

vertical-pod-autoscaler/pkg/apis/poc.autoscaling.k8s.io/v1alpha1/types.go
```diff
@@ -207,3 +207,64 @@ type RecommendedContainerResources struct {
...
+	VPAObjectName string `json:"vpaObjectName,omitempty" protobuf:"bytes,1,opt,name=vpaObjectName"`
+}
+
+// VerticalPodAutoscalerSnapshotStatus contains data of the snapshot.
+type VerticalPodAutoscalerSnapshotStatus struct {
+	// The time when the status was last refreshed.
+	LastUpdateTime metav1.Time `json:"lastUpdateTime,omitempty" protobuf:"bytes,1,opt,name=lastUpdateTime"`
+
+	// Schema of the snapshot that identifies format of the stored data.
+	Schema string `json:"schema,omitempty" protobuf:"bytes,2,opt,name=schema"`
```
mwielgus: What is store in schema?

vertical-pod-autoscaler/pkg/apis/poc.autoscaling.k8s.io/v1alpha1/types.go
schylek: Status/Schema field plays role of content version info. I was thinking whether it should be part of spec - my rationale to put it in status is that this is not configuration, but property of the stored data.

However, I have no strong feelings about that.

vertical-pod-autoscaler/pkg/apis/poc.autoscaling.k8s.io/v1alpha1/types.go
```diff
@@ -207,3 +207,64 @@ type RecommendedContainerResources struct {
...
+// HistogramSnapshot contains data needed to reconstruct the histogram.
+type HistogramSnapshot struct {
+	// Name of the resource that this HistogramSnapshot refers to.
+	Resource string `json:"resource,omitempty" protobuf:"bytes,1,opt,name=resource"`
+
+	// End of the window for which current histogram was computed.
+	WindowEnd metav1.Time `json:"windowEnd,omitempty" protobuf:"bytes,2,opt,name=windowEnd"`
+
+	// Map from bucket index to bucket weight.
+	BucketWeights map[int]float64 `json:"bucketWeights,omitempty" protobuf:"bytes,3,opt,name=bucketWeights"`
```
kgrygiel: For the sake of compactness I would recommend using float32 or even better uint32 + TotalWeight as the denominator.

vertical-pod-autoscaler/pkg/apis/poc.autoscaling.k8s.io/v1alpha1/types.go
```diff
@@ -207,3 +207,64 @@ type RecommendedContainerResources struct {
...
+	Histograms []HistogramSnapshot `json:"" protobuf:"bytes,3,rep,name=histograms"`
+}
+
+// HistogramSnapshot contains data needed to reconstruct the histogram.
+type HistogramSnapshot struct {
+	// Name of the resource that this HistogramSnapshot refers to.
+	Resource string `json:"resource,omitempty" protobuf:"bytes,1,opt,name=resource"`
+
+	// End of the window for which current histogram was computed.
+	WindowEnd metav1.Time `json:"windowEnd,omitempty" protobuf:"bytes,2,opt,name=windowEnd"`
```
kgrygiel: I don't think this field is needed to deserialize the histogram and I'm not sure how we would use it in this context.

You probably want to make the following assumptions:
- The CPU histogram aggregates all CPU samples until the timestamp of the checkpoint.
- The memory peaks histogram aggregates all peaks until the most recent aggregation interval that ended before the timestamp of the checkpoint. It is not obvious whether to include the peaks from the ongoing interval (it may lead to multiple peaks from a single period, esp. when you restart from the checkpoint many times during a single interval, so you may want to exclude the most recent peak).

vertical-pod-autoscaler/pkg/apis/poc.autoscaling.k8s.io/v1alpha1/types.go
```diff
@@ -207,3 +207,64 @@ type RecommendedContainerResources struct {
...
+	Schema string `json:"schema,omitempty" protobuf:"bytes,2,opt,name=schema"`
+
+	// Snapshots of histograms describing consumption of resources.
+	Histograms []HistogramSnapshot `json:"" protobuf:"bytes,3,rep,name=histograms"`
+}
+
+// HistogramSnapshot contains data needed to reconstruct the histogram.
+type HistogramSnapshot struct {
+	// Name of the resource that this HistogramSnapshot refers to.
+	Resource string `json:"resource,omitempty" protobuf:"bytes,1,opt,name=resource"`
```
kgrygiel: You should be able to remove this field if you follow the previous comment.

vertical-pod-autoscaler/pkg/apis/poc.autoscaling.k8s.io/v1alpha1/types.go
schylek: WindowEnd is meant to store DecayingHistogram.decayStart which is != timestamp of the checkpoint. If we don't receive samples for some time DecayingHistogram won't shift and if read from checkpoint it would have higher weights then those stored in memory before recommender's restart.

vertical-pod-autoscaler/pkg/apis/poc.autoscaling.k8s.io/v1alpha1/types.go
schylek: See other comment.

vertical-pod-autoscaler/pkg/apis/poc.autoscaling.k8s.io/v1alpha1/types.go
schylek: Done.

vertical-pod-autoscaler/pkg/apis/poc.autoscaling.k8s.io/v1alpha1/types.go
```diff
@@ -207,3 +207,64 @@ type RecommendedContainerResources struct {
...
+// VerticalPodAutoscalerSnapshotStatus contains data of the snapshot.
+type VerticalPodAutoscalerSnapshotStatus struct {
+	// The time when the status was last refreshed.
+	LastUpdateTime metav1.Time `json:"lastUpdateTime,omitempty" protobuf:"bytes,1,opt,name=lastUpdateTime"`
+
+	// Schema of the snapshot that identifies format of the stored data.
+	Schema string `json:"schema,omitempty" protobuf:"bytes,2,opt,name=schema"`
+
+	// Snapshots of histograms describing consumption of resources.
+	Histograms []HistogramSnapshot `json:"" protobuf:"bytes,3,rep,name=histograms"`
```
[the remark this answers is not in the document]
schylek: IMO having a list is more flexible since it enables adding new histograms without changing format of the snapshot.

vertical-pod-autoscaler/api/vpa-crd.yaml
```diff
@@ -50,4 +50,18 @@ spec:
                         type: object
                       maxAllowed:
                         type: object
-
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: verticalpodautoscalersnapshots.poc.autoscaling.k8s.io
```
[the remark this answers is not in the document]
schylek: Done.

vertical-pod-autoscaler/pkg/apis/poc.autoscaling.k8s.io/v1alpha1/types.go
schylek: Renamed to Version.

vertical-pod-autoscaler/pkg/apis/poc.autoscaling.k8s.io/v1alpha1/types.go
schylek: Version information is stored in status/version

schylek:
@kgrygiel PTAL, I have applied fixes discussed offline.

PR #685, opened 2018-02-27
Introduce recommendation confidence.

kgrygiel:
The confidence grows with the amount of aggregated utilization history.
The recommender multiplies MaxRecommendation by the factor of (1+1/confidence) and divides MinRecommendation by the same factor. This means that the [Min..Max] range is very wide when there is small amount of available history (either the VPA recommender has just started and didn't have access to the historical data or all pods matched by the VPA are new). The [Min..Max] range gets more narrow as the history accumulates.
As the effect, the Updater will be less likely to evict a pod when the confidence of the recommendation is low.

### DIFF
--- a/vertical-pod-autoscaler/recommender/logic/estimator_test.go
+++ b/vertical-pod-autoscaler/recommender/logic/estimator_test.go
@@ -26,7 +26,7 @@ import (
 )
 
 var (
-	anyTime = time.Time{}
+	anyTime = time.Unix(0, 0)
 )
 
 // Verifies that the PercentileEstimator returns requested percentiles of CPU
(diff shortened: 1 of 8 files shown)


vertical-pod-autoscaler/recommender/model/types.go
```diff
@@ -92,6 +100,22 @@ func ResourcesAsResourceList(resources Resources) apiv1.ResourceList {
 	return result
 }
 
+// RoundResourceAmount returns the given resource amount rounded down to the
+// whole multiple of another resource amount (unit).
+func RoundResourceAmount(amount ResourceAmount, unit ResourceAmount) ResourceAmount {
```
schylek: amount, unit ResourceAmount

vertical-pod-autoscaler/recommender/logic/estimator_test.go
```diff
@@ -48,8 +48,60 @@ func TestPercentileEstimator(t *testing.T) {
...
+func TestConfidenceMultiplierEstimatorNoHistory(t *testing.T) {
+	baseEstimator := NewConstEstimator(model.Resources{
+		model.ResourceCPU:    model.CPUAmountFromCores(3.14),
+		model.ResourceMemory: model.MemoryAmountFromBytes(3.14e9),
+	})
+	testedEstimator1 := &confidenceMultiplierEstimator{1.0, baseEstimator}
+	testedEstimator2 := &confidenceMultiplierEstimator{-1.0, baseEstimator}
+	s := newAggregateContainerState()
+	// Expect testedEstimator1 to return the maximum possible resource amount.
+	assert.Equal(t, 1e11, model.CoresFromCPUAmount(
```
schylek: CPU/milliCPU silent conversion can be hard to follow when you read that max is 1e11 while:
MaxResourceAmount = ResourceAmount(1e14)

schylek:
/lgtm

Note to self - new fields in AggregateContainerState have to be stored in Snapshot CRD

PR #688, opened 2018-03-02
Better GPU handling in price expander

MaciekPytel:
Use constant, very high unfitness for nodes with GPU to make them very unattractive for any pod that doesn't require GPU and to avoid situation where CA values CPU usage efficiency over GPU price optimization.

Also additionally delay scale-up if there are very new pods requesting GPU. We lose a bit of latency for pods requesting GPU, but we can pack them more densely that way.

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -37,6 +37,10 @@ import (
 const (
 	// How old the oldest unschedulable pod should be before starting scale up.
 	unschedulablePodTimeBuffer = 2 * time.Second
+	// How old the oldest unschedulable pod with GPU should be before starting scale up.
+	// The idea is that nodes with GPU are very expensive and we're ready to sacrifice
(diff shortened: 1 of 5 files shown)


MaciekPytel:
Tested using following scenario: 4 pods requesting 1 GPU and 5 cores, node groups gpu8 {2 GPU, 8 cores}, gpu16 {2 GPU, 16 cores}, a small node group without gpu for running system pods.

Without changes 4 nodes were added to gpu8. This is because node unfitness dominated pricing formula (1.6x difference in price, 2x difference in unfitness). Basically CA thinks 16 core nodes are too large for such small cluster and may hurt availability if too many pods are scheduled on such a large node.

With changes 2 nodes were added to gpu16 (after waiting for 30s for all pods to be created).

I think for nodes with GPU the new behavior is desired - those are dedicated nodes for pods that need GPU, not shared nodes for any pods. Therefore current heuristic approach doesn't make too much sense for them, while cost difference is very significant.

PR #690, opened 2018-03-05
Extend the Updater policy with new rules

kgrygiel:
1. Update priority is calculated based on the ratio between the total request and total recommended target for all containers, as opposed to sum of ratio for each container (previous policy favored many small containers).
2. Pods with a container that needs to grow get higher priority.
3. Pods with any request outside the [MinRecommended...MaxRecommended] range can be updated regardless of the relative diff between request and target.
4. Pods with all requests within the [MinRecommended...MaxRecommended] range can be updated if they live for at least 12h and the relative diff exceeds the threshold.

### DIFF
--- a/vertical-pod-autoscaler/pkg/utils/test/test_utils.go
+++ b/vertical-pod-autoscaler/pkg/utils/test/test_utils.go
@@ -19,6 +19,7 @@ package test
 import (
 	"fmt"
 	"log"
+	"time"
 
 	"github.com/stretchr/testify/mock"
 	apiv1 "k8s.io/api/core/v1"
@@ -30,9 +31,14 @@ import (
 	v1 "k8s.io/client-go/listers/core/v1"
 )
 
+var (
+	TimeLayout       = "2006-01-02 15:04:05"
+	testTimestamp, _ = time.Parse(TimeLayout, "2017-04-18 17:35:05")
+)
+
 // BuildTestPod creates a pod with specified resources.
 func BuildTestPod(name, containerName, cpu, mem string, creatorObjectMeta *metav1.ObjectMeta, creatorTypeMeta *metav1.TypeMeta) *apiv1.Pod {
-	startTime := metav1.Now()
+	startTime := metav1.Time{testTimestamp}
 	pod := &apiv1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: "default",
--- a/vertical-pod-autoscaler/pkg/utils/test/test_utils.go
+++ b/vertical-pod-autoscaler/pkg/utils/test/test_utils.go
@@ -32,6 +32,7 @@ import (
 
 // BuildTestPod creates a pod with specified resources.
 func BuildTestPod(name, containerName, cpu, mem string, creatorObjectMeta *metav1.ObjectMeta, creatorTypeMeta *metav1.TypeMeta) *apiv1.Pod {
+	startTime := metav1.Now()
 	pod := &apiv1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: "default",
@@ -40,6 +41,9 @@ func BuildTestPod(name, containerName, cpu, mem string, creatorObjectMeta *metav
 		},
 		Spec: apiv1.PodSpec{
 			Containers: []apiv1.Container{BuildTestContainer(containerName, cpu, mem)},
+		},
+		Status: apiv1.PodStatus{
+			StartTime: &startTime,
 		},
 	}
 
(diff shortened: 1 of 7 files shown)


schylek:
The Travis CI build failed.

vertical-pod-autoscaler/pkg/utils/test/test_utils.go
```diff
@@ -32,6 +32,7 @@ import (
 
 // BuildTestPod creates a pod with specified resources.
 func BuildTestPod(name, containerName, cpu, mem string, creatorObjectMeta *metav1.ObjectMeta, creatorTypeMeta *metav1.TypeMeta) *apiv1.Pod {
+	startTime := metav1.Now()
```
schylek: Please use a fixed date instead of Now()

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
```diff
@@ -27,47 +27,53 @@ import (
...
 	// containerPolicy can be nil (user does not have to configure it).
 	containerPolicy := getContainerPolicy(container.Name, policy)
-	applyVPAPolicy(containerRecommendation, containerPolicy)
-	// TODO: If limits and policy are conflicting, set some condition on the VPA.
-	capRecommendationToContainerLimit(containerRecommendation, container)
-	res := make(apiv1.ResourceList)
-	for resource, recommended := range containerRecommendation.Target {
-		res[resource] = recommended
+
+	cappedRecommendations := containerRecommendation.DeepCopy()
```
schylek: Why DeepCopy is needed here?

vertical-pod-autoscaler/updater/priority/update_priority_calculator.go
```diff
@@ -85,44 +101,62 @@ func (calc *UpdatePriorityCalculator) GetSortedPods() []*apiv1.Pod {
...
+				if recommended.MilliValue() > request.MilliValue() {
+					scaleUp = true
+				}
+				if (hasMin && request.MilliValue() < minRecommneded.MilliValue()) ||
+					(hasMax && request.MilliValue() > maxRecommneded.MilliValue()) {
+					outsideRecommendedRange = true
+				}
+			} else {
+				scaleUp = true
+				outsideRecommendedRange = true
```
schylek: AFAIU this code if we start new pod without request it will be evicted immediately just to assign request. Seems dangerous - I would prefer to let such pod live until we get better confidence in recommendation or it dies by itself (but giving us some usage samples before that)

vertical-pod-autoscaler/updater/priority/update_priority_calculator.go
```diff
@@ -132,6 +166,11 @@ func (list byPriority) Swap(i, j int) {
 	list[i], list[j] = list[j], list[i]
 }
 func (list byPriority) Less(i, j int) bool {
-	// reverse ordering, highest priority first
-	return list[i].priority > list[j].priority
+	// Reverse ordering, highest priority first.
+	// 1. If any container wants to grow, the pod takes precedence.
```
schylek: I find this disturbing - why those with shrinking resources shouldn't be first (e.g. to free resources for growing ones to use)?

vertical-pod-autoscaler/updater/priority/update_priority_calculator_test.go
```diff
@@ -146,19 +170,56 @@ func TestUsePolicy(t *testing.T) {
...
+		calculator.AddPod(pods[i], recommendation)
+	}
+	result := calculator.GetSortedPods()
+	assert.Exactly(t, []*apiv1.Pod{pods[1], pods[2]}, result, "Exactly POD2 and POD3 should be updated")
+}
 
-	calculator.AddPod(pod1, recommendation)
-	calculator.AddPod(pod2, recommendation)
+// Verify that a pod that lives for less than podLifetimeUpdateThreshold is
+// updated only if the request is outside the [MinRecommended...MaxRecommended]
```
schylek: Is there a unit test for a pod without request set?

vertical-pod-autoscaler/pkg/utils/test/test_utils.go
kgrygiel: Done.

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
kgrygiel: Well, previously it was also a "deep copy" of containerRecommendation.Target (just written explicitly, copying field by field).
As we now want to cap other fields (MinRecommended and MaxRecommended) as well, it would be very verbose to explicitly copy every field. Instead I'm just calling DeepCopy() which does it for me.

schylek:
Please fix golint check, one nit discussed offline. The rest /lgtm

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
kgrygiel: Oh, you're probably right that the copy wasn't necessary in the first place, because getRecommendationForContainer() already returns a copy. But I think I'd prefer not to depend on it, because it's fragile and easy to break without noticing.

PR #704, opened 2018-03-12
Add regional flag

aleksandra-malinowska:
Use flag to identify regional clusters iinstead of relying on multi-zone property.

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/gce_manager.go
+++ b/cluster-autoscaler/cloudprovider/gce/gce_manager.go
@@ -152,7 +152,7 @@ type gceManagerImpl struct {
 	mode                  GcpCloudProviderMode
 	templates             *templateBuilder
 	interrupt             chan struct{}
-	isRegional            bool
+	regional              bool
 	explicitlyConfigured  map[GceRef]bool
 	migAutoDiscoverySpecs []cloudprovider.MIGAutoDiscoveryConfig
 	resourceLimiter       *cloudprovider.ResourceLimiter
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -220,7 +218,7 @@ func CreateGceManager(configReader io.Reader, mode GcpCloudProviderMode, cluster
 		gceService:  gceService,
 		migCache:    make(map[GceRef]*Mig),
 		location:    location,
-		isRegional:  isRegional,
+		isRegional:  regional,
```
bskiba: nit: let's make this consistent and go for one of isRegional and regional in both places.

bskiba:
One small comment, otherwise lgtm

cluster-autoscaler/cloudprovider/gce/gce_manager.go
aleksandra-malinowska: Done. Went with `regional` following the AllElseBeingEqualLessCamelCaseIsBetter principle;)

negz:
Just curious - what was the reasoning behind this change? Is the multi-zone config knob being removed?

bskiba:
The two flags have slightly different meanings so they'll both stay: 
multi-zone - the worker nodes may be in multiple zones
regional - master nodes may be in multiple zones

So for example these two configs are valid and different from autoscaler point of view:
1) --multi-zone=true --regional=false (nodes across zones but master in just one zone)
2) --multi-zone=true --regional=true (nodes across zones and master replicated across zones)

PR #710, opened 2018-03-13
Extend cluster autoscaler to check AUTOSCALER_ENV_VARS in kube-env

mtaufen:
This is the second part of the fix in https://github.com/kubernetes/kubernetes/pull/61119

This provides a temporary way for the cluster autoscaler to get at
values that were removed from kube-env in https://github.com/kubernetes/kubernetes/pull/60020.
Ideally this information will eventually be available via e.g. the Cluster API,
because kube-env is an internal interface that carries no stability
guarantees.

/cc @roberthbailey @bskiba @mwielgus

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/templates.go
+++ b/cluster-autoscaler/cloudprovider/gce/templates.go
@@ -332,17 +332,31 @@ func parseKubeReserved(kubeReserved string) (apiv1.ResourceList, error) {
 }
 
 func extractLabelsFromKubeEnv(kubeEnv string) (map[string]string, error) {
-	labels, err := extractFromKubeEnv(kubeEnv, "NODE_LABELS")
+	// In v1.10+, labels are only exposed for the autoscaler via AUTOSCALER_ENV_VARS
+	// see kubernetes/kubernetes#61119. We try AUTOSCALER_ENV_VARS first, then
+	// fall back to the old way.
+	labels, err := extractAutoscalerVarFromKubeEnv(kubeEnv, "node_labels")
 	if err != nil {
-		return nil, err
+		glog.Errorf("node_labels not found via AUTOSCALER_ENV_VARS due to error, will try NODE_LABELS: %v", err)
+		labels, err = extractFromKubeEnv(kubeEnv, "NODE_LABELS")
+		if err != nil {
+			return nil, err
+		}
 	}
 	return parseKeyValueListToMap(labels)
 }
 
 func extractTaintsFromKubeEnv(kubeEnv string) ([]apiv1.Taint, error) {
-	taints, err := extractFromKubeEnv(kubeEnv, "NODE_TAINTS")
+	// In v1.10+, taints are only exposed for the autoscaler via AUTOSCALER_ENV_VARS
+	// see kubernetes/kubernetes#61119. We try AUTOSCALER_ENV_VARS first, then
+	// fall back to the old way.
+	taints, err := extractAutoscalerVarFromKubeEnv(kubeEnv, "node_taints")
 	if err != nil {
-		return nil, err
+		glog.Errorf("node_taints not found via AUTOSCALER_ENV_VARS due to error, will try NODE_TAINTS: %v", err)
+		taints, err = extractFromKubeEnv(kubeEnv, "NODE_TAINTS")
+		if err != nil {
+			return nil, err
+		}
 	}
 	taintMap, err := parseKeyValueListToMap(taints)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


bskiba:
Looks like wee need to handle possible empty values in parseKeyValueListToMap, sth along the lines of: https://github.com/bskiba/autoscaler/commit/a15c29df2af6d351ff47dd1b0c07a5d7597da3a0

Otherwise this looks good to me and I have successfully tested it on GCE together with https://github.com/kubernetes/kubernetes/pull/61119

mtaufen:
>Looks like wee need to handle possible empty values in parseKeyValueListToMap, sth along the lines of: bskiba/autoscaler@a15c29d

Do you have a test case that will trigger this?

>Otherwise this looks good to me and I have successfully tested it on GCE together with kubernetes/kubernetes#61119

Great!

bskiba:
I tested this on GCE with the new variables and when node-taints where empty (as is in the default case), this errored on me from here: https://github.com/mtaufen/autoscaler/blob/037467fad2c566d901f04a3f78024aff0a55aa54/cluster-autoscaler/cloudprovider/gce/templates.go#L452

I can write a unit-test for this case tomorrow.

bskiba:
Here's a unit test to make it break: https://github.com/bskiba/autoscaler/commit/faa2151808b85f17476918448127a15fd57bbe31 (tests for empty node taints). With bskiba/autoscaler@a15c29d this passes.

Could you incorporate the fix?

mtaufen:
Updated, PTAL :)

cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -332,17 +332,37 @@ func parseKubeReserved(kubeReserved string) (apiv1.ResourceList, error) {
...
+	// fall back to the old way.
+	labels, err := extractAutoscalerVarFromKubeEnv(kubeEnv, "node_labels")
 	if err != nil {
-		return nil, err
+		glog.Infof("node_labels not found via AUTOSCALER_ENV_VARS, will try NODE_LABELS")
+		labels, err := extractFromKubeEnv(kubeEnv, "NODE_LABELS")
+		if err != nil {
+			return nil, err
+		}
+		return parseKeyValueListToMap(labels)
```
roberthbailey: nit: you could remove this line and just exit the if block and let the line below return the map to reduce code duplication. up to you whether you think that's more confusing.

cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -332,17 +332,37 @@ func parseKubeReserved(kubeReserved string) (apiv1.ResourceList, error) {
...
+	// fall back to the old way.
+	taints, err := extractAutoscalerVarFromKubeEnv(kubeEnv, "node_taints")
 	if err != nil {
-		return nil, err
+		glog.Infof("node_taints not found via AUTOSCALER_ENV_VARS, will try NODE_TAINTS")
+		taints, err := extractFromKubeEnv(kubeEnv, "NODE_TAINTS")
+		if err != nil {
+			return nil, err
+		}
+		taintMap, err := parseKeyValueListToMap(taints)
```
roberthbailey: Same comment as above, but in this case there is more duplicated code. Once you get the taints extracted, you can leave the if block and parse them into a map in one place.

cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -352,19 +372,55 @@ func extractTaintsFromKubeEnv(kubeEnv string) ([]apiv1.Taint, error) {
...
+		for _, value := range kubeletArgs {
+			matches := resourcesRegexp.FindStringSubmatch(value)
+			if len(matches) > 1 {
+				return matches[1], nil
+			}
+		}
+		return "", fmt.Errorf("kube-reserved not in kubelet args in kube-env: %q", strings.Join(kubeletArgs, " "))
 	}
-	resourcesRegexp := regexp.MustCompile(`--kube-reserved=([^ ]+)`)
+	return kubeReserved[0], nil
```
roberthbailey: do we need a check here to verify that kubeReserved will have a zero-th entry and this won't cause a panic?

cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -352,19 +372,55 @@ func extractTaintsFromKubeEnv(kubeEnv string) ([]apiv1.Taint, error) {
...
-		matches := resourcesRegexp.FindStringSubmatch(value)
-		if len(matches) > 1 {
-			return matches[1], nil
+func extractAutoscalerVarFromKubeEnv(kubeEnv, name string) ([]string, error) {
+	const autoscalerVars = "AUTOSCALER_ENV_VARS"
+	autoscalerVals, err := extractFromKubeEnv(kubeEnv, autoscalerVars)
+	if err != nil {
+		return nil, err
+	}
+	result := []string{}
```
roberthbailey: nit: i personally prefer to write this as `var result []string` and then let the append calls do the allocation.

cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -352,19 +372,55 @@ func extractTaintsFromKubeEnv(kubeEnv string) ([]apiv1.Taint, error) {
...
+	for _, val := range autoscalerVals {
```
roberthbailey: we always expect a single val right? so this outer loop is expected to either execute once or not at all?

cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -352,19 +372,55 @@ func extractTaintsFromKubeEnv(kubeEnv string) ([]apiv1.Taint, error) {
...
+		for _, v := range strings.Split(val, ";") {
```
roberthbailey: Can labels or taints contain a semi-colon? If so, we can't use it as the separator char.

cluster-autoscaler/cloudprovider/gce/templates_test.go
```diff
@@ -297,26 +297,41 @@ func TestBuildAllocatableFromCapacity(t *testing.T) {
 }
 
 func TestExtractLabelsFromKubeEnv(t *testing.T) {
```
roberthbailey: are there test cases to check for expected errors during parsing (e.g. malformed vars, missing vars, etc)?

cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -352,19 +372,55 @@ func extractTaintsFromKubeEnv(kubeEnv string) ([]apiv1.Taint, error) {
 }
 
 func extractKubeReservedFromKubeEnv(kubeEnv string) (string, error) {
-	kubeletArgs, err := extractFromKubeEnv(kubeEnv, "KUBELET_TEST_ARGS")
+	// In v1.10+, kube-reserved is only exposed for the autoscaler via AUTOSCALER_ENV_VARS
+	// see kubernetes/kubernetes#61119. We try AUTOSCALER_ENV_VARS first, then
+	// fall back to the old way.
+	kubeReserved, err := extractAutoscalerVarFromKubeEnv(kubeEnv, "kube_reserved")
```
roberthbailey: In https://github.com/kubernetes/kubernetes/pull/61119 you have a comment that kube-reserved will need to be pulled from KUBELET_ARGS but this code only checks AUTOSCALER_ARGS with a fallback to KUBELET_TEST_ARGS so it won't find the parameter for 1.10 nodes.

bskiba:
Tested this once again after fixes from @mtaufen on GCE and GKE and still works good. Please fix comments from @roberthbailey and I'll be happy to merge.

cluster-autoscaler/cloudprovider/gce/templates.go
mtaufen: Users can just edit their MIG templates to include it in AUTOSCALER_ENV_VARS on GCE if they need it. We can probably just remove the comment. I'd recommend we have a release note in the next autoscaler version to that effect.

cluster-autoscaler/cloudprovider/gce/templates.go
mtaufen: I don't think so:
labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
>Labels are key/value pairs. Valid label keys have two segments: an optional prefix and name, separated by a slash (/). The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between. The prefix is optional. If specified, the prefix must be a DNS subdomain: a series of DNS labels separated by dots (.), not longer than 253 characters in total, followed by a slash (/). If the prefix is omitted, the label Key is presumed to be private to the user. Automated system components (e.g. kube-scheduler, kube-controller-manager, kube-apiserver, kubectl, or other third-party automation) which add labels to end-user objects must specify a prefix. The kubernetes.io/ prefix is reserved for Kubernetes core components.
>Valid label values must be 63 characters or less and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.

taints: https://github.com/kubernetes/kubernetes/blob/ca02c11887c65431af85f809606527be7f08062d/pkg/apis/core/validation/validation.go#L3931

Taints use the same validation as labels for keys and values. 

The effect part of the taint args to Kubelet (after the `:`) is split out from the value part here when Kubelet parses the flag: https://github.com/kubernetes/kubernetes/blob/ca02c11887c65431af85f809606527be7f08062d/pkg/util/taints/taints.go#L46

cluster-autoscaler/cloudprovider/gce/templates_test.go
mtaufen: I added some additional cases for expected errors.

cluster-autoscaler/cloudprovider/gce/templates.go
bskiba: Good idea, I can take care of adding to release notes.

bskiba:
Looks like we don't have github.com/stretchr/testify/require vendored

mtaufen:
/sigh
`require` would result in less garbage when a test fails due to an unexpected error, but I don't want to deal with Godep, so I'll just switch back to `assert`.

mtaufen:
kubernetes/kubernetes#61119 merged, so this is all that's left

PR #715, opened 2018-03-14
Limit logging of pods found to be schedulable

aleksandra-malinowska:
Cluster Autoscaler logs every pod it finds to be schedulable once per loop, which causes excessive logging if the number of such pods is large. This changes it to log up to 1000 pods per loop if running with `--v=10`, and up to 5 with lower verbosity levels.


cluster-autoscaler/core/utils.go
```diff
@@ -49,6 +49,9 @@ import (
 const (
 	// ReschedulerTaintKey is the name of the taint created by rescheduler.
 	ReschedulerTaintKey = "CriticalAddonsOnly"
+	// MaxSchedulableLog is the maximum number of schedulable pods for which
+	// we will log information where they could fit for verbosity < 10.
+	MaxSchedulableLog = 5
```
MaciekPytel: I would log full output at V5 or V6. It will still be off by default, but you can change it if you need logs for a specific pod, without getting ridiculous amount of predicate logs at V10.

Also I would set the threshold higher. Each pod logs at maximum 6 lines / minute, so 20 would still keep us around 100 lines / minute and it's more likely to give a better sample for debugging.

MaciekPytel:
Maybe we want to add a generic method for throttling logs and put it in utils instead of fixing it in this one place? I've seen other log lines get really spammy in the past (especially in performance tests) and it would be quite nice to have a generic way of dealing with them.

Also I remember how useful https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/simulator/cluster.go#L198 was for debugging exactly what's going on in small clusters. Maybe with throttling we could actually bring it back? 

WDYT @aleksandra-malinowska?

cluster-autoscaler/core/utils.go
mwielgus: 20 or more.

aleksandra-malinowska:
Makes sense. Added simple util for limiting logging and used it in both places.

cluster-autoscaler/simulator/cluster.go
```diff
@@ -195,7 +197,7 @@ func findPlaceFor(removedNode string, pods []*apiv1.Pod, nodes []*apiv1.Node, no
 				return false
 			}
 			err := predicateChecker.CheckPredicates(pod, predicateMeta, nodeInfo, ReturnVerboseError)
-			glog.V(5).Infof("Evaluation %s for %s/%s -> %v", nodename, pod.Namespace, pod.Name, err)
+			glogx.V(4).UpTo(&evaluateLogLimit).Infof("Evaluation %s for %s/%s -> %v", nodename, pod.Namespace, pod.Name, err)
```
mwielgus: It is easy to get lost where evaluateloglimit is declared, set and used.

cluster-autoscaler/utils/glogx/glogx.go
```diff
@@ -0,0 +1,60 @@
...
+	return *limit < 0
+}
+
+// V calls V from glog and wraps the result into glogx.Verbose.
+func V(n glog.Level) Verbose {
+	return Verbose(glog.V(n))
+}
+
+// Verbose is a wrapper for glog.Verbose that implements UpTo and Over.
+type Verbose glog.Verbose
```
mwielgus: This is too complicated. Can we simply create a struct that will allow to log at the given level only N times, with no extra int pointers etc.

aleksandra-malinowska: Do you mean that I should instead create an object with V() method, and use it instead of calling V() from package (something like Logger in Java?)

aleksandra-malinowska: refactored as discussed offline, PTAL

mwielgus:
Please add some tests.

PR #720, opened 2018-03-16
Update metadata.go, a clerical error

AdamDang:
line 48: a clerical error "This is variable name is not defined by any spec"
It should be "This variable name is not defined by any spec"


bskiba:
This is vendored code. Please update in the respective repo.

PR #723, opened 2018-03-16
VPA initial detection of OOMs caused by limit violation.

schylek:


vertical-pod-autoscaler/recommender/input/cluster_feeder.go
```diff
@@ -56,42 +57,46 @@ type ClusterStateFeeder interface {
...
-	podLister := newPodLister(kubeClient)
-	return spec.NewSpecClient(podLister)
-}
-
 func newMetricsClient(config *rest.Config) metrics.MetricsClient {
 	metricsGetter := resourceclient.NewForConfigOrDie(config)
 	return metrics.NewMetricsClient(metricsGetter)
 }
 
 // Creates PodLister, listing only not terminated pods.
```
bskiba: Can you update the comment?

PR #730, opened 2018-03-21
fix some typo

fqsghostcloud:
Here are some spelling mistakes that need to be corrected


feiskyer:
@fqsghostcloud we don't update vendor files this way. You should fix those typo in their original repo, e.g. https://github.com/kubernetes/kubernetes

fqsghostcloud:
thanks

PR #738, opened 2018-03-22
Disable VolumeScheduling predicate

aleksandra-malinowska:
Disable VolumeScheduling predicate (which breaks scale-up) until there's a workaround.


mwielgus:
Please open an issue to track this.

PR #748, opened 2018-03-26
Add Cluster Autoscaler 1.2 as recommended version with Kubernetes 1.10

aleksandra-malinowska:
Add Cluster Autoscaler 1.2 as recommended version with Kubernetes 1.10


MaciekPytel:
Do we also want to add release notes for recent CA releases? Or should we move all release notes to a separate file, as it's starting to get big? I vaguely recall already having this discussion, but I don't think we reached any conclusion.

aleksandra-malinowska:
Not sure about this. I just wanted to avoid the same situation as we had last time with missing current k8s/ca versions.

PR #749, opened 2018-03-26
Rely on cluster.yaml to restore cluster state

resouer:
Rely on cluster.yaml to restore cluster state

Do not modify config.yaml


resouer:
Sorry, wrong PR, closed!

PR #751, opened 2018-03-28
add dot

sunlintong:


bskiba:
Actually, we have comments with dots and without dots all over the codebase. I don't think this increases readability of the code.

PR #755, opened 2018-03-28
Update pricing.md

AdamDang:
"n1-standard-..." should be changed to "N1-standard-..."
In this doc, the machine type is defined as "N1-standard-...", but there are several "n1-standard-...".
It's better to use same format.


bskiba:
Actually, I think it's better to change it all to lower case, since AFAIK this is what actual machine type names look like.

AdamDang:
@bskiba Thanks. I've changed it all to lower case.

bskiba:
Cool, thanks

PR #761, opened 2018-03-29
Fix a typo

tkulczynski:
@bskiba: sorry, I made a typo in the previous one and it seems test code is not unit tested... ;)


bskiba:
Hm, the e2es are excluded from testing because of this commit: https://github.com/kubernetes/autoscaler/commit/1199a4ac7d702095b70b3ce3bdc84ad9e7f44367 but it doesn't seem to good that we can check in code that doesn't build.

PR #765, opened 2018-03-30
Fix parsing kube-env

### LINKED ISSUE #745: Properly fix parsing multi line values in kube-env
Multi line kube env values mess up our parsing. Temporary fix was introduced in https://github.com/kubernetes/autoscaler/pull/741. Let's do a proper one.

bskiba:
Use the yaml-parsing library to parse KUBE_ENV in gce instance templates to deal with possible multi line values.
Fixes #745


bskiba:
Actually, now I look at this, I can make extractFromKubeEnv return a single string instead of a list of strings. Let me fix that.

bskiba:
Ok, ready for review. Tested on GKE and GCE by triggering scale up from zero.

mwielgus:
Cool

PR #766, opened 2018-03-30
Print version number on startup.

kgrygiel:


vertical-pod-autoscaler/common/version.go
```diff
@@ -0,0 +1,20 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+// VerticalPodAutoscalerVersion is the version of VPA.
+const VerticalPodAutoscalerVersion = "0.0.1"
```
mwielgus: 0.1.0

kgrygiel: Done.

PR #769, opened 2018-03-30
Version update in vpa deployments

mwielgus:
cc: @kgrygiel @tkulczynski

### DIFF
--- a/vertical-pod-autoscaler/deploy/admission-controller-deployment.yaml
+++ b/vertical-pod-autoscaler/deploy/admission-controller-deployment.yaml
@@ -13,7 +13,7 @@ spec:
       serviceAccountName: vpa-admission-controller
       containers:
       - name: admission-controller
-        image: k8s.gcr.io/vpa-admission-controller:v0.0.1
+        image: gcr.io/google_containers/vpa-admission-controller:v0.1.0
         imagePullPolicy: Always
         volumeMounts:
           - name: tls-certs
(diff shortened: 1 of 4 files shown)


tkulczynski:
I'm afraid 0.0.1 is also hardcoded in Makefiles and in hack/vpa-process-yaml.sh...

mwielgus:
OK, i will update all places but we should eliminate them on the earliest convenience and have one file where registery and tag is defined.

vertical-pod-autoscaler/deploy/admission-controller-deployment.yaml
```diff
@@ -13,19 +13,19 @@ spec:
       serviceAccountName: vpa-admission-controller
       containers:
       - name: admission-controller
-        image: k8s.gcr.io/vpa-admission-controller:v0.0.1
+        image: gcr.io/google_containers/vpa-admission-controller:v0.1.0
```
bskiba: I think the current 'right way' is to use k8s.gcr.io?

mwielgus:
Actually, makefiles don't need to be updated.

mwielgus:
Updated.

vertical-pod-autoscaler/hack/vpa-process-yaml.sh
```diff
@@ -31,8 +31,8 @@ if [ $# -eq 0 ]; then
   exit 1
 fi
 
-DEFAULT_REGISTRY="k8s.gcr.io"
-DEFAULT_TAG="v0.0.1"
+DEFAULT_REGISTRY="gcr.io/google_containers/vpa-updater"
```
kgrygiel: Should be just "gcr.io/google_containers".

mwielgus: Done.

PR #774, opened 2018-04-03
2nd pass: k8s.gcr.io

thockin:
I will be sending out more messages before eventually disabling the legacy URL.  In the mean time, please be aware that gcr.io/google_containers is to be replaced with k8s.gcr.io in almost every case.


thockin:
Yes, I will probably add a pre-submit or something first

PR #783, opened 2018-04-06
AWS: Support selecting ASG by values

discordianfish:
So far aws asg discovery only works by matching on tag keys, not values.
But I have a tag "KubernetesCluster" with a value to identify the cluster, so this isn't working for me.

This PR (should) adds support for filtering the ASG by key *and* value, e.g:
```
--node-group-auto-discovery=asg:tag=KubernetesCluster=my-cluster
```

Not working since something appears to modify the flag value before reaching the cloudprovider stuff. Will pick this up next week but though some early feedback would be useful.

### DIFF
--- a/cluster-autoscaler/cloudprovider/node_group_discovery_options.go
+++ b/cluster-autoscaler/cloudprovider/node_group_discovery_options.go
@@ -169,9 +169,9 @@ func parseMIGAutoDiscoverySpec(spec string) (MIGAutoDiscoveryConfig, error) {
 
 // An ASGAutoDiscoveryConfig specifies how to autodiscover AWS ASGs.
 type ASGAutoDiscoveryConfig struct {
-	// TagKeys to match on.
+	// Tags to match on.
 	// Any ASG with all of the provided tag keys will be autoscaled.
-	TagKeys []string
+	Tags map[string]string
 }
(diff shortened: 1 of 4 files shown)


discordianfish:
Turned out my changes are working fine, I just didn't manage to rebuild the image properly last week.
So while some better tests would be nice, this is good to get merged from my POV.

cluster-autoscaler/cloudprovider/node_group_discovery_options.go
```diff
@@ -186,8 +186,8 @@ func parseASGAutoDiscoverySpec(spec string) (ASGAutoDiscoveryConfig, error) {
 		return cfg, fmt.Errorf("Unsupported discoverer specified: %s", discoverer)
 	}
 	param := tokens[1]
-	kv := strings.Split(param, "=")
-	if len(kv) != 2 {
+	kv := strings.SplitN(param, "=", 2)
+	if len(kv) < 2 {
```
mwielgus: why are you ok with len(kv) > 2 ?

discordianfish: You're right, this should stay `!= 2`. Fixed it!

PR #787, opened 2018-04-10
Fix absent image tag version in README

nak3:
Since addon-resizer does not have the tag `1.8`, this patch updates README to use `1.8.1`.

https://console.cloud.google.com/gcr/images/google-containers/GLOBAL/addon-resizer?gcrImageListsize=50


x13n:
/assign @kawych 
Karol, will these flags to addon resizer container work with 1.8.1?

kawych:
@x13n yes
/lgtm

PR #809, opened 2018-04-19
Factor the checkpoint specific code out of aggregations.

kgrygiel:


vertical-pod-autoscaler/recommender/output/checkpoint_writer.go
```diff
@@ -75,3 +76,26 @@ func (writer *checkpointWriter) StoreCheckpoints() {
 		}
 	}
 }
+
+func buildAggregateContainerStateMap(vpa *model.Vpa, now time.Time) map[string]*model.AggregateContainerState {
```
schylek: I'd add some info why this function is introduced (compared to original model.BuildAggregateContainerStateMap)

E.g. change name to buildCheckpointAggregateContainerStateMap

vertical-pod-autoscaler/recommender/output/checkpoint_writer.go
```diff
@@ -75,3 +76,26 @@ func (writer *checkpointWriter) StoreCheckpoints() {
...
+		for containerName, container := range pod.Containers {
+			if aggregateContainerState, exists := aggregateContainerStateMap[containerName]; exists {
+				subtractCurrentContainerPeak(aggregateContainerState, container, now)
+			}
+		}
+	}
+	return aggregateContainerStateMap
+}
+
+func subtractCurrentContainerPeak(a *model.AggregateContainerState, container *model.ContainerState, now time.Time) {
```
schylek: This method covers only memory. 
Maybe subtractCurrentContainer*Memory*Peak?

PR #815, opened 2018-04-26
Switch from per-pod aggregations to per-labelset aggregations.

kgrygiel:
* Usage samples are no longer aggregated per pod (which prevented from garbage collecting terminated pods and could lead to very large number of aggregations, when many pods with different names were created in the cluster).
* We introduce a new concept of the AggregateStateKey, which determines the set of containers
for which usage samples are aggregated. In the current implementation the AggregateStateKey
consists of { namespace, containerName, label set }, however it can be easily changed/extended in future, if necessary.
* Each container holds a pointer to the AggregateContainerState to which it contributes usage samples.
* Each VPA holds a set of all AggregateContainerStates which it matches.
* Pod no longer keeps links to the VPAs that match it. VPA no longer keeps links to Pods it matches.
* Checkpoint specific code is moved out from AggregateContainerState to Vpa.
* Memory peaks are stored in a histogram, similarly to CPU samples. Each container keeps its last peak.
  Sliding window is no longer used.
* Added more test cases.
* TODO: Garbage collect old Pods and AggregateContainerStates.
* TODO: Delete sliding window.

### DIFF
--- a/vertical-pod-autoscaler/recommender/model/aggregate_container_state.go
+++ b/vertical-pod-autoscaler/recommender/model/aggregate_container_state.go
@@ -18,6 +18,7 @@ package model
 
 import (
 	"fmt"
+	"math"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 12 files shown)


vertical-pod-autoscaler/recommender/model/aggregate_container_state.go
```diff
@@ -18,45 +18,66 @@ package model
...
+// ContainerNameToAggregateStateMap maps a container name to AggregateContainerState
+// that aggregates state of containers with that name.
+type ContainerNameToAggregateStateMap map[string]*AggregateContainerState
+
 const (
 	// SupportedCheckpointVersion is the tag of the supported version of serialized checkpoints.
+	// Version id should be incremented on every non incompatible change, i.e. if the new
+	// version of the recommender binary can't initialize from the old checkpoint format or the the
+	// previous version of the recommender binary can't initialize from the new checkpoint format.
 	SupportedCheckpointVersion = "v1"
```
mwielgus: Shouldn't we bump version with this change?

vertical-pod-autoscaler/recommender/model/aggregate_container_state.go
```diff
@@ -67,6 +88,50 @@ func NewAggregateContainerState() *AggregateContainerState {
...
+		a.AggregateMemoryPeaks.AddSample(BytesFromMemoryAmount(sample.Usage), 1.0, sample.MeasureStart)
+	default:
+		panic(fmt.Sprintf("AddSample doesn't support resource '%s'", sample.Resource))
+	}
+}
+
+// SubtractSample removes a single usage sample from an aggregation.
+// The subtracted sample should be equal to some sample that was aggregated with
+// AddSample() in the past.
+// Only memory samples can be subtracted at the moment. Support for CPU could be
```
mwielgus: Can you elaborate why cpu is not handled?

vertical-pod-autoscaler/recommender/model/aggregate_container_state.go
kgrygiel: We only checkpoint the VPA state, which should not be affected by this change.

vertical-pod-autoscaler/recommender/model/aggregate_container_state.go
kgrygiel: Subtracting CPU samples is not implemented because we don't currently have a use-case for it.
For memory we store daily peaks in the histogram - whenever the peak from today grows we had to remove the old one from the histogram and add a new one. For CPU we just add every sample to the histogram and never remove anything.

vertical-pod-autoscaler/recommender/model/vpa.go
```diff
@@ -70,53 +71,72 @@ type Vpa struct {
...
 
-// UpdatePodLink marks the Pod as controlled or not-controlled by the VPA
-// depending on whether the pod labels match the Vpa pod selector.
-// If multiple VPAs match the same Pod, only one of them will effectively
-// control the Pod.
-func (vpa *Vpa) UpdatePodLink(pod *PodState) bool {
-	_, previouslyMatched := pod.MatchingVpas[vpa.ID]
-	currentlyMatching := vpa.MatchesPod(pod)
+// UsesAggregation returns true iff an aggregation with the given key contributes to the VPA.
+func (vpa *Vpa) UsesAggregation(aggregationKey AggregateStateKey) bool {
```
schylek: Nit: s/UsesAggregation/HasAggregation

vertical-pod-autoscaler/recommender/model/vpa.go
```diff
@@ -70,53 +71,72 @@ type Vpa struct {
...
 	return vpa
 }
 
-// MatchesPod returns true iff a given pod is matched by the Vpa pod selector.
-func (vpa *Vpa) MatchesPod(pod *PodState) bool {
-	if vpa.ID.Namespace != pod.ID.Namespace {
-		return false
+// UseAggregationIfMatching checks if the given aggregation matches (contributes to) this VPA
+// and adds it to the set of VPA's aggregations if that is the case.
+func (vpa *Vpa) UseAggregationIfMatching(aggregationKey AggregateStateKey, aggregation *AggregateContainerState) {
```
schylek: Nit: In this convenience it should be named UseAggregationIfFirstAndMatching. Maybe TryUseAggregation? TryBindAggregation?

vertical-pod-autoscaler/recommender/model/cluster_test.go
```diff
@@ -104,45 +108,32 @@ func addTestPod(cluster *ClusterState) *PodState {
...
 	cluster := NewClusterState()
 	vpa := addTestVpa(cluster)
-	pod := addTestPod(cluster)
-	cluster.DeleteVpa(vpa.ID)
-	assert.Empty(t, pod.MatchingVpas)
-	assert.NotContains(t, cluster.Vpas, vpa.ID)
+	addTestPod(cluster)
+	container := addTestContainer(cluster)
+	aggregateStateKey := cluster.aggregateStateKeyForContainerID(testContainerID)
+	assert.True(t, container.aggregator == vpa.aggregateContainerStates[aggregateStateKey])
```
schylek: assert.Equal / assert.NotEqual won't work? xN

kgrygiel: This is slightly tricky, assert.Equal checks that values are equal, this asserts that the references are identical (https://godoc.org/github.com/stretchr/testify/assert#Assertions.Equal).

PR #818, opened 2018-04-26
"pod" and "Pod" are mixed except that in proper noun

AdamDang:
In this doc, "pod" and "Pod" are mixed. For the proper noun like "Vertical Pod Autoscaler", it's in capital. That's OK.
But in other situations, it's mixed in capital and lowercase. It's better to use same format.


kgrygiel:
/lgtm
Thanks for fixing.

PR #830, opened 2018-05-07
Add support for rescheduled pods with the same name in drain

aleksandra-malinowska:
Add support for rescheduled pods with the same name in drain.

### DIFF
--- a/cluster-autoscaler/core/scale_down.go
+++ b/cluster-autoscaler/core/scale_down.go
@@ -766,14 +766,15 @@ func drainNode(node *apiv1.Node, pods []*apiv1.Pod, client kube_client.Interface
 		allGone = true
 		for _, pod := range pods {
 			podreturned, err := client.CoreV1().Pods(pod.Namespace).Get(pod.Name, metav1.GetOptions{})
-			if err == nil {
+			if err == nil && (podreturned == nil || podreturned.Spec.NodeName == node.Name) {
 				glog.Errorf("Not deleted yet %v", podreturned)
 				allGone = false
 				break
 			}
-			if !kube_errors.IsNotFound(err) {
+			if err != nil && !kube_errors.IsNotFound(err) {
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/core/scale_down.go
```diff
@@ -763,12 +763,12 @@ func drainNode(node *apiv1.Node, pods []*apiv1.Pod, client kube_client.Interface
 		allGone = true
 		for _, pod := range pods {
 			podreturned, err := client.CoreV1().Pods(pod.Namespace).Get(pod.Name, metav1.GetOptions{})
-			if err == nil {
+			if err == nil && podreturned.Spec.NodeName == node.Name {
```
MaciekPytel: I will feel better if you add `podreturned != nil` check. It should probably never happen without some err, but it's still a pointer.

cluster-autoscaler/core/scale_down.go
```diff
@@ -763,12 +763,12 @@ func drainNode(node *apiv1.Node, pods []*apiv1.Pod, client kube_client.Interface
...
 				glog.Errorf("Not deleted yet %v", podreturned)
 				allGone = false
 				break
 			}
-			if !kube_errors.IsNotFound(err) {
+			if err != nil && !kube_errors.IsNotFound(err) {
 				glog.Errorf("Failed to check pod %s/%s: %v", pod.Namespace, pod.Name, err)
 				allGone = false
```
MaciekPytel: Since you're already working on this file - shouldn't there be a break here?

aleksandra-malinowska: Done

MaciekPytel:
/lgtm
Please merge once you're done testing. Also I think this is worth cherry-picking.

mwielgus:
Is the testing finished?

aleksandra-malinowska:
Sorry, found (and fixed) another issues with taints persisting through autoscaler restarts. 

Tested manually that StatefulSet pods are now correctly recognized as being moved to another node.

PR #831, opened 2018-05-07
Remove implementation details (CleanUp) from the interface.

kgolab:
The CleanUp method is instead called directly from the implementation,
when required.
Test updated in a quick way since the mock we're using does not support
AtLeast(1) - thus Times(2).


MaciekPytel:
Please squash your commits, otherwise lgtm.

kgolab:
Squashed.

PR #836, opened 2018-05-08
Use timestamp argument

kgolab:


MaciekPytel:
Do you know what the effect of changing this will actually be? I'm not particularly happy with merging random cleanup that may actually break things, without understanding it or knowing (how) it was tested.

We still don't have any CI running at master and if this introduces any problem the likely outcome is we will notice it right before release and do yet another session of panicked debugging + last minute updates.

kgolab:
@MaciekPytel , it might be a bit moot since it's merged already but as far as I understand the code clearing the data improperly should only influence efficiency of the simulation.

As for testing I haven't done anything apart of running the unit tests.

PR #848, opened 2018-05-11
Add caching to get machine type calls.

krzysztof-jastrzebski:


cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -172,7 +172,11 @@ func (gce *GceCloudProvider) NewNodeGroup(machineType string, labels map[string]
 		},
 		gceManager: gce.gceManager,
 	}
-	_, err := gce.gceManager.getTemplates().buildNodeFromAutoprovisioningSpec(mig)
+	cpu, mem, err := mig.gceManager.getCpuAndMemoryForMachineType(mig.spec.machineType, mig.GceRef.Zone)
```
MaciekPytel: Why are we now refering to gceManager via mig? I guess it doesn't really matter, but it's a strange thing to change.

krzysztof-jastrzebski: Fixed. It was copy and paste.

krzysztof-jastrzebski: done

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -392,17 +396,25 @@ func (mig *Mig) Autoprovisioned() bool {
...
+			return nil, err
+		}
+		node, err = mig.gceManager.getTemplates().buildNodeFromTemplate(mig, template, cpu, mem)
 		if err != nil {
 			return nil, err
 		}
 	} else if mig.Autoprovisioned() {
 		var err error
-		node, err = mig.gceManager.getTemplates().buildNodeFromAutoprovisioningSpec(mig)
+		cpu, mem, err := mig.gceManager.getCpuAndMemoryForMachineType(mig.spec.machineType, mig.GceRef.Zone)
```
MaciekPytel: nitpicking: this call could happen before if (it's almost identical in both branches)

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -1048,6 +1075,45 @@ func (m *gceManagerImpl) GetResourceLimiter() (*cloudprovider.ResourceLimiter, e
...
+	m.machinesCacheLastRefresh = time.Now()
+	glog.V(2).Infof("Cleared machine types cache, next clear after %v", m.machinesCacheLastRefresh.Add(machinesRefreshInterval))
+	return nil
+}
+
+func (m *gceManagerImpl) fetchMachinesCache() error {
+	if m.machinesCacheLastRefresh.Add(machinesRefreshInterval).After(time.Now()) {
+		return nil
+	}
+	cluster, err := m.gkeAlphaService.Projects.Zones.Clusters.Get(m.projectId, m.location, m.clusterName).Do()
```
MaciekPytel: I'm pretty sure this will segfault (nil ptr dereference) in ModeGKE. The reason is we only use gkeAlphaService in ModeGKENAP. In ModeGKE we use gkeService and / or gkeBetaService if on regional cluster. It's a horrible, horrible mess, but that's where we are.

Bottom line is you need to distinguish and use gkeAlphaService or gkeService depending on mode (I think you can still use gkeService in regional cluster - can you confirm @bskiba?)

MaciekPytel: Not sure if Locations are already available in v1? If no you may actually need 3 implementations :(

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -101,6 +103,12 @@ type migInformation struct {
 	basename string
 }
 
+type machineTypeKey struct {
+	projectId   string
```
MaciekPytel: Do we need projectId? We don't support running CA across projects anyway.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
bskiba: Locations is only available in beta now unfortunately. Generally, if we want to be sure that this works for regional clusters for the long run, we should use Location API instead of the Zone API so unfortunately gkeBetaService for regional :(

MaciekPytel:
@krzysztof-jastrzebski Given the mess that is all the different GKE modes I think this needs to be tested against all scenarios (I caught one problem, I may have missed other). By all scenarios I mean:
1. Cluster on GCE
1. Cluster on GKE using GKE mode
1. Regional cluster using GKE mode
1. NAP cluster

Sorry about that. You're paying now for our past mistakes :(

cluster-autoscaler/cloudprovider/gce/gce_manager.go
MaciekPytel: Right, so once again we need 3 implementations of the same :( Let's try to clean this mess for 1.12.

bskiba: Actually, it looks like the docs are out of date, https://container.googleapis.com/$discovery/rest?version=v1 shows that the location API has been moved to v1. This means we should be able to eliminate gkeBetaService if we move all calls to the location path of the v1 api \o/

MaciekPytel: Let's remove gkeBetaService in a separate PR though. This one is large enough and removing beta service will also be a big change. Good news is you will only need to have 2 versions in your new code (v1 service is also initialized in regional, so you can just use it in ModeGKE).

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
krzysztof-jastrzebski: It is not executed in else.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
krzysztof-jastrzebski: done

cluster-autoscaler/cloudprovider/gce/gce_manager.go
krzysztof-jastrzebski: done

Issue #814, opened 2018-04-25
DynamicAutoscaler lacks CleanUp call in Reconfigure


kgolab:
DynamicAutoscaler's Reconfigure method can create a [new Autoscaler](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/dynamic_autoscaler.go#L88) but it does not call CleanUp for it, breaking Autoscaler contract for the wrapped instance.

To be decided whether it's worth fixing, I've heard that DynamicAutoscaler is not actually used anymore.

kgolab:
Since there are some refactoring efforts planned to beautify CA code, I'd like to invoke Lazy Consensus here.
If there are no objections, in particular from @mumoshu , I'll remove DynamicAutoscaler next week (probably already May 14th).

mumoshu:
@kgolab Ah, sorry I was a bit late to party.
kube-aws still relies on this feature anyway. Any difficulty to keep it after the refactoring effort? Can I help somehow?

MaciekPytel:
@mumoshu I thought kube-aws moved from DynamicAutoscaler to PollingAutoscaler a long time ago?

kgolab:
/assign mumoshu

It's not a big deal to keep the DynamicAutoscaler, and of course I won't remove it, knowing it's still used.

The only tricky part is that I cannot easily test if it still works. But if you could help here, it would be great. Actually, pull request #831 should have fixed the problem mentioned here, could you please verify that using kube-aws?

mumoshu:
@MaciekPytel @kgolab Ah, sorry! You are correct. I was just confused - kube-aws has been migrated to PollingAutoscaler(`--node-group-auto-discovery`) long time ago.

So I have no worry about removing the dynamic config thing now.

kgolab:
Closing as due to #851 this ticket is no longer valid.

PR #851, opened 2018-05-14
Remove DynamicAutoscaler since it's unused

kgolab:
As discussed in #814 it's no longer needed.


MaciekPytel:
Shouldn't we also remove the flag, not just ignore it?

kgolab:
I've removed the configmap flag and its dependencies which were not used elsewhere.

PR #855, opened 2018-05-15
Add gpu related tests to scale_up_test

losipiuk:


MaciekPytel:
gofmt presubmit is failing

cluster-autoscaler/core/scale_down_test.go
```diff
@@ -1209,6 +1209,15 @@ func getStringFromChan(c chan string) string {
 	}
 }
 
+func getGroupSizeChangeFromChan(c chan groupSizeChange) *groupSizeChange {
```
MaciekPytel: As far as I can see you only use it in scale_up_test.go, why do you define it here?

cluster-autoscaler/core/scale_test_common.go
```diff
@@ -0,0 +1,54 @@
+/*
+Copyright 2016 The Kubernetes Authors.
```
MaciekPytel: 2018 :P

cluster-autoscaler/core/scale_test_common.go
```diff
@@ -0,0 +1,54 @@
...
+	sizeChange int
+}
+
+type scaleTestConfig struct {
+	nodes                  []nodeConfig
+	pods                   []podConfig
+	extraPods              []podConfig
+	expectedScaleUpOptions []groupSizeChange // we expect that all those options should be included in expansion options passed to expander strategy
+	scaleUpOptionToChoose  groupSizeChange   // this will be selected by assertingStrategy.BestOption
+	expectedFinalScaleUp   groupSizeChange   // we expect this to be delivered via scale-up event
```
MaciekPytel: If you want to distinguish between scaleUpOptionToChoose and expectedFinalScaleUp, can you make expectedFinalScaleUp a list? With `--balance-similar-node-groups` a single scale-up can actually be divided into size change in multiple NodeGroups

cluster-autoscaler/utils/test/test_utils.go
```diff
@@ -64,6 +65,19 @@ func BuildTestPod(name string, cpu int64, mem int64) *apiv1.Pod {
 	return pod
 }
 
+// RequestGpuForPod modifies pod's resource requests by adding a number of GPUs to them.
+func RequestGpuForPod(pod *apiv1.Pod, gpusCount int64) {
```
MaciekPytel: You're adding a taint automatically in AddGpusToNode. Shouldn't you add toleration here in similar way? How is it even working without toleration?

losipiuk: True. Actually we are not using the same predicates set in this UT as one which is used in production (`simulator.NewTestPredicateChecker` vs `simulator.NewPredicateChecker`).

This is poor but not much we can do about it know as we discussed offline.

losipiuk:
Note to reviewers (@MaciekPytel )

Needed to add following constants section to test_utils.go

```go
const (
	// cannot use constants from gpu module due to cyclic package import
	resourceNvidiaGPU = "nvidia.com/gpu"
	gpuLabel          = "cloud.google.com/gke-accelerator"
	defaultGPUType    = "nvidia-tesla-k80"
)
```

PR #856, opened 2018-05-15
Update go version to 1.10 and deprecate all <1.9

losipiuk:


hack/check-go-version.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# Copyright 2014 The Kubernetes Authors.
```
mwielgus: 2018

losipiuk: Done

PR #858, opened 2018-05-15
 Split Autoscaler.RunOnce into separate logical steps

kgolab:
Please have a look whether we can use the initial commit or do we really need the error-handling boilerplate to make the code look more like a "normal" go.

### DIFF
--- a/cluster-autoscaler/core/loop_state.go
+++ b/cluster-autoscaler/core/loop_state.go
@@ -1,0 +1,430 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package core
+
+import (
+	"fmt"
+	"time"
+
+	"k8s.io/autoscaler/cluster-autoscaler/clusterstate/utils"
+	"k8s.io/autoscaler/cluster-autoscaler/context"
+	"k8s.io/autoscaler/cluster-autoscaler/metrics"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/gpu"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/tpu"
+
+	apiv1 "k8s.io/api/core/v1"
+
+	"github.com/golang/glog"
+)
+
+// LoopStateHandler hides complexity of Autoscaler.RunOnce implementation,
+// splitting it into a number of phases.
+// Each phase method returns bool value to show whether processing can continue further.
+// LoopStateHandler instance is a very short-lived one, persisting state of a single RunOnce call.
+type LoopStateHandler interface {
+	// RefreshState obtains new Node-level data from the cloud provider
+	// and updates internal ClusterStateRegistry, until metrics.UpdateState.
+	// Returns true iff the processing can continue further.
+	RefreshState() bool
+	// CheckPreconditions verifies Node-level preconditions for continuing
+	// to processing Pod-level data.
+	// Returns true iff the processing can continue further.
+	CheckPreconditions() bool
+	// PreparePodData obtains Pod-level data for further processing.
+	// Returns true iff the processing can continue further.
+	PreparePodData() bool
+	// ScaleUp executes scale-up, if it's required.
+	// Returns true iff the processing can continue further.
+	ScaleUp() bool
+	// ScaleDown executes scale-down logic, both dry-run & actual scale-down, if it's required.
+	// Returns true iff the processing can continue further.
+	ScaleDown() bool
+
+	// ReturnError returns the result (error or nil) to return from Autoscaler.RunOnce.
+	ReturnError() errors.AutoscalerError
+}
+
+// loopState is the default implementation of LoopStateHandler
+type loopState struct {
+	currentTime        time.Time
+	startTime          time.Time
+	retError           errors.AutoscalerError
+	scaleDownForbidden bool
+
+	autoscaler *StaticAutoscaler
+	context    *context.AutoscalingContext
+
+	allNodes                                       []*apiv1.Node
+	readyNodes                                     []*apiv1.Node
+	allScheduled                                   []*apiv1.Pod
+	unschedulablePodsToHelp                        []*apiv1.Pod
+	unschedulableWaitingForLowerPriorityPreemption []*apiv1.Pod
+}
+
+// NewLoopStateHandler creates new instance of the default LoopStateHandler implementation
+func NewLoopStateHandler(autoscaler *StaticAutoscaler, currentTime time.Time) *loopState {
+	return &loopState{
+		currentTime: currentTime,
+		startTime:   time.Now(),
+		autoscaler:  autoscaler,
+		context:     autoscaler.AutoscalingContext,
+	}
+}
+
+// See LoopStateHandler interface
+func (ls *loopState) RefreshState() bool {
+	return ls.refreshCloudProvider() &&
+		ls.obtainNodeLists() &&
+		ls.updateClusterStateRegistry()
+}
+
+// See LoopStateHandler interface
+func (ls *loopState) CheckPreconditions() bool {
+	return ls.cleanUnregisteredNodes() &&
+		ls.checkClusterHealth() &&
+		ls.fixNodeGroupSizes()
+}
+
+// See LoopStateHandler interface
+func (ls *loopState) PreparePodData() bool {
+	allUnschedulablePods, err := ls.autoscaler.UnschedulablePodLister().List()
+	if err != nil {
+		glog.Errorf("Failed to list unscheduled pods: %v", err)
+		return ls.onError(errors.ApiCallError, err)
+	}
+	metrics.UpdateUnschedulablePodsCount(len(allUnschedulablePods))
+
+	ls.allScheduled, err = ls.autoscaler.ScheduledPodLister().List()
+	if err != nil {
+		glog.Errorf("Failed to list scheduled pods: %v", err)
+		return ls.onError(errors.ApiCallError, err)
+	}
+
+	allUnschedulablePods, ls.allScheduled, err = ls.autoscaler.podListProcessor.Process(
+		ls.context, allUnschedulablePods, ls.allScheduled, ls.allNodes)
+	if err != nil {
+		glog.Errorf("Failed to process pod list: %v", err)
+		return ls.onError(errors.InternalError, err)
+	}
+
+	ConfigurePredicateCheckerForLoop(allUnschedulablePods, ls.allScheduled, ls.autoscaler.PredicateChecker)
+
+	// We need to check whether pods marked as unschedulable are actually unschedulable.
+	// It's likely we added a new node and the scheduler just haven't managed to put the
+	// pod on in yet. In this situation we don't want to trigger another scale-up.
+	//
+	// It's also important to prevent uncontrollable cluster growth if CA's simulated
+	// scheduler differs in opinion with real scheduler. Example of such situation:
+	// - CA and Scheduler has slightly different configuration
+	// - Scheduler can't schedule a pod and marks it as unschedulable
+	// - CA added a node which should help the pod
+	// - Scheduler doesn't schedule the pod on the new node
+	//   because according to it logic it doesn't fit there
+	// - CA see the pod is still unschedulable, so it adds another node to help it
+	//
+	// With the check enabled the last point won't happen because CA will ignore a pod
+	// which is supposed to schedule on an existing node.
+	ls.scaleDownForbidden = false
+
+	unschedulablePodsWithoutTPUs := tpu.ClearTPURequests(allUnschedulablePods)
+
+	// Some unschedulable pods can be waiting for lower priority pods preemption so they have nominated node to run.
+	// Such pods don't require scale up but should be considered during scale down.
+	var unschedulablePods []*apiv1.Pod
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/core/loop_state.go
```diff
@@ -0,0 +1,448 @@
...
+	if len(ls.unschedulablePodsToHelp) == 0 {
+		glog.V(1).Info("No unschedulable pods")
+	} else if ls.autoscaler.MaxNodesTotal > 0 && len(ls.readyNodes) >= ls.autoscaler.MaxNodesTotal {
+		glog.V(1).Info("Max total nodes in cluster reached")
+	} else if allPodsAreNew(ls.unschedulablePodsToHelp, ls.currentTime) {
+		// The assumption here is that these pods have been created very recently and probably there
+		// is more pods to come. In theory we could check the newest pod time but then if pod were created
+		// slowly but at the pace of 1 every 2 seconds then no scale up would be triggered for long time.
+		// We also want to skip a real scale down (just like if the pods were handled).
+		ls.scaleDownForbidden = true
```
aleksandra-malinowska: If the purpose of this change is splitting the code by logic, I would expect that:
- ScaleUp() method only deals with scale-up, 
- ScaleDown() only deals with scale-down,
- whatever method calls ScaleUp and/or ScaleDown should decide when/if to call them.

Right now, both ScaleUp() and PreparePodData() explicitly decide that scale-down is forbidden. The logic is still all over the place, but now the names suggest it isn't (making it harder to find out what happens.)

cluster-autoscaler/core/loop_state.go
```diff
@@ -0,0 +1,448 @@
...
+		ls.autoscaler.lastScaleDownDeleteTime = ls.currentTime
+	}
+
+	return ls.onSuccess()
+}
+
+// single-use helper methods start here, extracted to limit complexity of any single method
+// all these methods return the values used by phase methods of LoopStateHandler interface
+
+func (ls *loopState) refreshCloudProvider() (ExecutionResult, errors.AutoscalerError) {
```
aleksandra-malinowska: Why does refreshCloudProvider return ExecutionResult enum? It can't break out of main loop without error

cluster-autoscaler/core/loop_state.go
```diff
@@ -0,0 +1,448 @@
...
+)
+
+// LoopStateHandler hides complexity of Autoscaler.RunOnce implementation,
+// splitting it into a number of phases.
+// Each phase method returns ExecutionResult to show if processing can continue further
+// and a result (error or nil) to return from Autoscaler.RunOnce if the processing should stop.
+// NOTE: Do not use `err != nil` idiom to decide if the processing can continue.
+// LoopStateHandler instance is a very short-lived one, persisting state of a single RunOnce call.
+type LoopStateHandler interface {
+	// RefreshState obtains new Node-level data from the cloud provider
```
aleksandra-malinowska: from the cloud provider and Kubernetes API server

cluster-autoscaler/core/loop_state.go
```diff
@@ -0,0 +1,448 @@
...
+	} else {
+		glog.V(4).Info("No schedulable pods")
+	}
+
+	return ls.onSuccess()
+}
+
+// See LoopStateHandler interface
+func (ls *loopState) ScaleUp() (ExecutionResult, errors.AutoscalerError) {
+	if len(ls.unschedulablePodsToHelp) == 0 {
```
aleksandra-malinowska: Instead of multiple else ifs and success path code contained in final else, can we have a series of ifs with early return? Like this:

if () {
 return
}
if () {
 return
}
// success path

cluster-autoscaler/core/loop_state.go
```diff
@@ -0,0 +1,448 @@
...
+	if typedErr != nil {
+		glog.Errorf("Failed to scale down: %v", typedErr)
+		ls.onTypedError(typedErr)
+	}
+
+	metrics.UpdateDurationFromStart(metrics.FindUnneeded, unneededStart)
+
+	if glog.V(4) {
+		for key, val := range ls.autoscaler.scaleDown.unneededNodes {
+			glog.V(4).Infof("%s is unneeded since %s duration %s", key, val.String(), ls.currentTime.Sub(val).String())
```
aleksandra-malinowska: V(4) is checked twice - can we either remove it here, or delete the enclosing if?

cluster-autoscaler/core/loop_state.go
```diff
@@ -0,0 +1,448 @@
...
+		// Some nodes were removed. Let's skip this iteration, the next one should be better.
+		if removedAny {
+			glog.V(0).Infof("Some unregistered nodes were removed, skipping iteration")
+			return ls.onEarlyExit()
+		}
+	}
+	return ls.onSuccess()
+}
+
+func (ls *loopState) checkClusterHealth() (ExecutionResult, errors.AutoscalerError) {
```
aleksandra-malinowska: It doesn't just "check" cluster health, but also act on this (cleaning up internal state and exiting early from the main loop)

cluster-autoscaler/core/loop_state.go
```diff
@@ -0,0 +1,448 @@
...
+		context:     autoscaler.AutoscalingContext,
+	}
+}
+
+// See LoopStateHandler interface
+func (ls *loopState) RefreshState() (ExecutionResult, errors.AutoscalerError) {
+	if result, err := ls.refreshCloudProvider(); result != CanContinue {
+		return result, err
+	}
+	if result, err := ls.obtainNodeLists(); result != CanContinue {
```
aleksandra-malinowska: This also doesn't look good - why should obtainNodeList decide whether to continue with the main loop? The logic for handling empty cluster could easily be moved outside of it.

aleksandra-malinowska:
In general, my main concern with this PR is that it restructures the code into not-so-logically-separate parts:
- methods that seem to contain only part of the logic (ScaleUp, ScaleDown) actually must be aware of and make decisions regarding other parts of it (forbidding scale down, "CanContinue" execution result etc.)
- loopState is like a new staticAutoscaler. It's not obvious if it should contain state local to the single iteration of the loop, or persisting between iterations. In effect, it has all of it and also configuration as its field.

cluster-autoscaler/core/loop_state.go
kgolab: It's just to have a consistent interface - I'd rather avoid exposing implementation detail (which might only temporarily hold) to the outside.

cluster-autoscaler/core/loop_state.go
```diff
@@ -0,0 +1,448 @@
...
+		currentTime: currentTime,
+		startTime:   time.Now(),
+		autoscaler:  autoscaler,
+		context:     autoscaler.AutoscalingContext,
+	}
+}
+
+// See LoopStateHandler interface
+func (ls *loopState) RefreshState() (ExecutionResult, errors.AutoscalerError) {
+	if result, err := ls.refreshCloudProvider(); result != CanContinue {
```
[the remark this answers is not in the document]
kgolab: Same response as for refreshCloudProvider - it's about consistency.

cluster-autoscaler/core/loop_state.go
kgolab: Well, not so easily since the obtainNodeList has two exits on errors and two exits on business conditions, intertwined.
As discussed with @MaciekPytel we'd have to make sure that gpu.FilterOutNodesWithUnreadyGpus is safe to be called on empty allNodes list (it looks like it is but it's not tested for this) before reordering the conditions or just sticking with the final one.

The `result != CanContinue` construct is for both error handling and business decisions (here and in general). The method could return separate statuses for errors and business decisions but actually that won't affect this particular line of code.

Of course we could argue that obtainXYZ method should not have side effects and I'd happily agree there. But currently the logic is following the pattern seen prior to the refactor, I tried to avoid introducing functional changes, even if they are tiny (there are few, though, where they seemed to be bug-fixes).

I'm open for further discussion there - the goal of the refactor was not to end up with ideal code but just to move slightly further away from the monstrous RunOnce we've had.

In this particular case I can only see changing the method name or splitting it further, the latter will remove most of the gains we have. If you see a better way forward I'm happy to follow as I don't expect you want only a name change.

cluster-autoscaler/core/loop_state.go
kgolab: Yes, we can, committed.

I've copied the code structure inside the new methods so it looks more similar to the old one - but we can improve on it, of course.

cluster-autoscaler/core/loop_state.go
kgolab: Thanks, I should have spotted this when moving the code (as I did move the if outside the loop).
Fixed.

cluster-autoscaler/core/loop_state.go
```diff
@@ -0,0 +1,448 @@
...
+		return result, err
+	}
+	if result, err := ls.obtainNodeLists(); result != CanContinue {
+		return result, err
+	}
+	return ls.updateClusterStateRegistry()
+}
+
+// See LoopStateHandler interface
+func (ls *loopState) CheckPreconditions() (ExecutionResult, errors.AutoscalerError) {
```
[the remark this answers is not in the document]
kgolab: Sorry, the name is a left-over from my initial split. Fixed it to express the intent slightly better.

cluster-autoscaler/core/loop_state.go
kgolab: Yes, it's one of the many places where we have side-effects in innocent-looking places.
I don't see (yet) a good way to fix this particular issue.

cluster-autoscaler/core/loop_state.go
kgolab: A very valid point. As is evident by the sheer size of ScaleUp and ScaleDown methods they were refactored the least and, as you observed, contain unwanted interactions from the code prior to refactoring.

I think I can improve the situation a bit here, avoiding `scaleDownForbidden` flag but I'm afraid some entanglement issues will still be left.
TODO(kgolab) - revisit this

MaciekPytel:
I'm not convinced this actually helps the problem of RunOnce. It just splits the whole loop into arbitrary chunks. Those chunks have names that are their "main theme", however, there is plenty of calls that have nothing to do with what particular function is meant to do (ex. ConfigurePredicateCheckerForLoop in PreparePodData, allPodsAreNew in ScaleUp). Different sub-functions still share state (aptly named loopState) and are full of side-effects impacting each other.

Basically, you still have to follow the whole thing to now what is actually going on. Except now you have more scrolling and additional code gluing it together. Last, but not least, RunOnce is an evil we already know. I know exactly what happens where and I suspect other CA maintainers are equally knowledgeable. I'm not sure it's worth giving up that knowledge without very clear benefit in terms of readability / maintainability.

I think to get that benefit you would have to go deeper, see which calls can be pushed inside functions we're already calling, which parts can be reordered safely, so that sub-functions actually only do a single thing, etc.

kgolab:
Regarding the general concerns:
* I'll split the logic a bit more but please do not expect that it will be a pure solution, as far as I understand the current CA logic it's not expressible as a nice set of functions; but we'll get closer, for sure,
* the "CanContinue" execution result is necessary for error handling; it could be a bool (as in the first commit of this PR), it can be an enum (as now), we can change it to something nicer if you have a better solution in mind; part of the problem here is that we _want_ CA to abort parts of ScaleDown based on results of ScaleUp,
* loopState is a short-lived object which exists only for a single execution of Autoscaler.RunOnce (see comment on the LoopStateHandler interface),
* the reason that loopState contains links to configuration is that I didn't want to pass these as arguments in all the calls, having a stateful object anyway; if you prefer I can move this to method arguments, I just tend to avoid typing.

kgolab:
As for RunOnce being the known evil - yes, I'm aware that this exercise is a risky one with regards to this knowledge.
I can split the code more / extract more functions - actually one of the reasons not to do it immediately was to have a slightly better code structure but still have resemblance to the original code, if only to keep that knowledge.

I can also restart from another direction, making smaller changes in RunOnce until it becomes more readable. My feeling is simply that each of these won't be nice looking either, even if maybe all together they will improve the code.

aleksandra-malinowska:
>I'll split the logic a bit more but please do not expect that it will be a pure solution, as far as I understand the current CA logic it's not expressible as a nice set of functions; but we'll get closer, for sure,
>
>the "CanContinue" execution result is necessary for error handling; it could be a bool (as in the first commit of this PR), it can be an enum (as now), we can change it to something nicer if you have a better solution in mind; part of the problem here is that we want CA to abort parts of ScaleDown based on results of ScaleUp,

> I can split the code more / extract more functions - actually one of the reasons not to do it immediately was to have a slightly better code structure but still have resemblance to the original code, if only to keep that knowledge.
>
> I can also restart from another direction, making smaller changes in RunOnce until it becomes more readable. My feeling is simply that each of these won't be nice looking either, even 
if maybe all together they will improve the code.

I respectfully disagree. "CanContinue" isn't necessary, it's an artifact of pushing existing logic in a new arbitrary structure that is (as you mentioned yourself) not well-suited for it. 

The options you list sound like "top-down" vs "bottom-up" approach. I'm not arguing for sticking to either one, quite the opposite. Improving smaller parts of the code will lead to better results if we keep a larger goal of restructuring it in mind. Similarly, imposing a new structure should simplify the logic, not complicate it.

What I'd like to achieve is a code structure that reflects the fundamental logic of the application. This makes the code way more readable ("readable" as in "understand how it works just by looking at it"). That logic (and the resulting structure) should be as simple as possible. I just don't think this PR brings us closer to either goal.

>loopState is a short-lived object which exists only for a single execution of Autoscaler.RunOnce (see comment on the LoopStateHandler interface),
the reason that loopState contains links to configuration is that I didn't want to pass these as arguments in all the calls, having a stateful object anyway; if you prefer I can move this to method arguments, I just tend to avoid typing.

My point is that loopState becomes the new staticAutoscaler, i.e. it's an object holding all state (mixed up), with all the main loop logic is in its methods. The only change (apart from the structure discussed above) is that we pushed all of this down to another object that will be recreated in every iteration.

>As for RunOnce being the known evil - yes, I'm aware that this exercise is a risky one with regards to this knowledge.

This doesn't worry me too much. To be honest, it probably makes me more confident that we can simplify this code rather than just shuffle it around. What worries me is adding structure on top of it that doesn't do what it says it does, making it harder to follow the code paths.

kgolab:
Thank you for all the comments.
As suggested previously I will try another approach to reorganise the code as I don't see a quick fix to salvage this PR.

PR #864, opened 2018-05-16
Revert "Cherry pick #813 and #852 to cluster-autoscaler-release-1.2"

aleksandra-malinowska:
Reverts kubernetes/autoscaler#861

cc @feiskyer This added a dependency that isn't in kubernetes/kubernetes: https://github.com/kubernetes/kubernetes/blob/release-1.10/Godeps/Godeps.json#L33 . Can you add it to kubernetes/kubernetes 1.10 release branch (to which we sync Godeps for Cluster Autoscaler 1.2) before we merge it back?

cc @MaciekPytel @mwielgus

PR #866, opened 2018-05-16
Add override for AKS API

feiskyer:
Alternative way for #864, add override for AKS API.

This PR should be cherry picked to 1.2 branch.

cc @MaciekPytel @mwielgus


cluster-autoscaler/fix_gopath.sh
```diff
@@ -14,14 +14,11 @@ for item in $to_fix; do
...
 done
 
-with_vendor=`find $GOPATH/src/ -type d | grep vendor | grep -v 'vendor/'`
-for item in $with_vendor; do
-    echo Removing vendor from $item
-    cd $item
-    cd ..
-    rm -rf $item
-    git commit -a -m no_vendor
-done
```
feiskyer: By the way, with this change, only vendor/Godeps in cluster-autoscaler are removed ().

aleksandra-malinowska: The purpose of this loop is to remove vendor from dependencies that are going to be saved to Cluster Autoscaler's vendor (e.g kubernetes/kubernetes' vendor/)

cluster-autoscaler/fix_gopath.sh
```diff
@@ -30,3 +27,10 @@ mkdir -p $GOPATH/src/google.golang.org/api/container/v1beta1
 cp $GOPATH/src/k8s.io/autoscaler/cluster-autoscaler/_override/google.golang.org/api/container/v1beta1/*  $GOPATH/src/google.golang.org/api/container/v1beta1
 cd $GOPATH/src/google.golang.org/api/
 git commit -a -m "Api override for NAP"
+
+echo Overridng AKS API
```
aleksandra-malinowska: Overriding

aleksandra-malinowska:
Before committing, please run fix_gopath.sh so that GKE and AKS APIs are in vendor (yes, they're going to be duplicated in _override)

cluster-autoscaler/fix_gopath.sh
feiskyer: OK, let me revert this back.

cluster-autoscaler/fix_gopath.sh
feiskyer: ACK

cluster-autoscaler/fix_gopath.sh
```diff
@@ -14,14 +14,11 @@ for item in $to_fix; do
...
 
-with_vendor=`find $GOPATH/src/ -type d | grep vendor | grep -v 'vendor/'`
-for item in $with_vendor; do
-    echo Removing vendor from $item
-    cd $item
-    cd ..
-    rm -rf $item
-    git commit -a -m no_vendor
-done
+echo Removing vendor from cluster-autoslcer
```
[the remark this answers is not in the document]
feiskyer: ACK

cluster-autoscaler/fix_gopath.sh
feiskyer: Wait, the original commands also removes vendors from vpa which is not expected. I think we should only removes vendors under CA

feiskyer:
@aleksandra-malinowska Addressed comments. PTAL

feiskyer:
And CI green now.

cluster-autoscaler/fix_gopath.sh
aleksandra-malinowska: It is intended to remove vendors from `$GOPATH/src/k8s.io/kubernetes` etc., not just Cluster Autoscaler.

If we want to exclude VPA (and addon resizer), we should explicitly check for those.

feiskyer: > It is intended to remove vendors from $GOPATH/src/k8s.io/kubernetes etc., not just Cluster Autoscaler.

But actually not, it removes vendors for vpa. And it is not required to remove vendors from $GOPATH/src/k8s.io/kubernetes., godeps won't include vendored packages.

mwielgus: Can we, for simplicity, keep it as is for a moment? We had mixed experience with godeps in the past. We would like to do the release today so right now we don't have enough capacity to validate that it will work without removals. 

We can get back to this discussion tomorrow :).

aleksandra-malinowska: >  And it is not required to remove vendors from $GOPATH/src/k8s.io/kubernetes., godeps won't include vendored packages.

I'll double-check this, but can you please revert it for now, so that we may proceed with the release? This hack works, and the cherry-pick with new version to 1.10 branch needs to be ready today. Let's continue this discussion after we made it in time for 1.10.3

feiskyer:
@aleksandra-malinowska thanks

PR #868, opened 2018-05-17
Move metrics update to proper place

kgolab:


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -180,6 +180,9 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 	}
 	UpdateClusterStateMetrics(a.ClusterStateRegistry)
 
+	metrics.UpdateDurationFromStart(metrics.UpdateState, runStart)
+	metrics.UpdateLastTime(metrics.Autoscaling, time.Now())
```
aleksandra-malinowska: If the cluster is unhealthy, or if we remove unregistered nodes, we'll exit early and never start the autoscaling step. I don't think we should update this metric then. (This is actually also the case for fixing node group size, so maybe we should move it even later, past that call?)

MaciekPytel:
Right, maybe we need one more label?

cluster-autoscaler/core/static_autoscaler.go
kgolab: @aleksandra-malinowska , I hope you mean only the second line here.

The doc says:
> autoscaling - current state of the cluster has been updated, started autoscaling logic.

So I guess it all boils down to the definition of what "autoscaling logic" actually means. I'm happy to accept a different interpretation here.

kgolab: Split the two:
* duration metric stays where it is as this fits description in metrics.md,
* lastTime metric is moved to after all Node-level operations and precondition checks.

PR #870, opened 2018-05-17
Set lastScaleDownFailTime properly

kgolab:
The ScaleDownError check was ureachable


aleksandra-malinowska:
@MaciekPytel do you recall what was the intent of that TODO? Should we add some (more concrete) reminder/note back?

PR #872, opened 2018-05-17
Move ClusterStateRegistry to StaticAutoscaler

kgolab:
AutoscalingContext is basically a configuration and few static helpers and API handles.
ClusterStateRegistry is a state and thus moved closer to other state-keeping objects.

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -19,6 +19,7 @@ package core
 import (
 	"time"
 
+	"k8s.io/autoscaler/cluster-autoscaler/clusterstate"
 	"k8s.io/autoscaler/cluster-autoscaler/clusterstate/utils"
 	"k8s.io/autoscaler/cluster-autoscaler/context"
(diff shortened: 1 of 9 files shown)


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -52,6 +53,8 @@ type StaticAutoscaler struct {
 	// AutoscalingContext consists of validated settings and options for this autoscaler
 	*context.AutoscalingContext
 	kube_util.ListerRegistry
+	// ClusterState for maintaining the state of cluster nodes.
+	ClusterStateRegistry    *clusterstate.ClusterStateRegistry
```
aleksandra-malinowska: Why does it need to be public?

kgolab: It doesn't, it just was, similar to the two above. Changed that.

mwielgus:
@kgolab  please resolve confilcts

kgolab:
I've rebased, resolved conflicts and squashed the commits since the middle state was no longer compiling.

PR #880, opened 2018-05-17
Allow passing taints to Node Group creation.

jbartosik:


MaciekPytel:
@jbartosik You need to sort out CLA before we can accept any PR from you. Also this doesn't pass go fmt presubmit.

MaciekPytel:
Other than the comment above looks good to me. Thanks :)

smarterclayton:
On master in cluster-autoscaler I see a build fail on kubemark with:

```
$ go build .
# k8s.io/autoscaler/cluster-autoscaler/cloudprovider/builder
cloudprovider/builder/cloud_provider_builder.go:208:2: cannot use provider (type *"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/kubemark".KubemarkCloudProvider) as type "k8s.io/autoscaler/cluster-autoscaler/cloudprovider".CloudProvider in return argument:
	*"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/kubemark".KubemarkCloudProvider does not implement "k8s.io/autoscaler/cluster-autoscaler/cloudprovider".CloudProvider (wrong type for NewNodeGroup method)
		have NewNodeGroup(string, map[string]string, map[string]string, map[string]resource.Quantity) ("k8s.io/autoscaler/cluster-autoscaler/cloudprovider".NodeGroup, error)
		want NewNodeGroup(string, map[string]string, map[string]string, []"k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/api/core/v1".Taint, map[string]resource.Quantity) ("k8s.io/autoscaler/cluster-autoscaler/cloudprovider".NodeGroup, error)
```

Was kubemark supposed to be removed, or was this just accidental?

PR #899, opened 2018-05-28
Delete autoprovisioned node pool after all nodes are deleted.

krzysztof-jastrzebski:


cluster-autoscaler/core/scale_down.go
```diff
@@ -839,11 +839,15 @@ func cleanUpNodeAutoprovisionedGroups(cloudProvider cloudprovider.CloudProvider,
 		if !nodeGroup.Autoprovisioned() {
 			continue
 		}
-		size, err := nodeGroup.TargetSize()
+		targetSize, err := nodeGroup.TargetSize()
 		if err != nil {
 			return err
 		}
-		if size == 0 {
+		nodes, err := nodeGroup.Nodes()
```
MaciekPytel: Since the reason for this change is performance, maybe we can avoid this additional API call by doing something along the lines of:
```
if targetSize > 0 {
  continue
}
```
before getting nodes?

PR #900, opened 2018-05-28
Move some GKE-specific logic outside core

MaciekPytel:
At some point we added a whole bunch of GKE-specific code to core. This is the first part of cleaning that mess up.

No change in actual logic. Added a new NodeGroupListProcessor interface to encapsulate the existing logic. Moved PodListProcessor and refactor how it's passed around to make it consistent and easy to add more 'processor' interfaces.

I thought about placing processors in context, but gave up on that due to problems with cyclic imports. Instead I put all processors into a single struct to avoid passing multiple processors around.


cluster-autoscaler/core/scale_up.go
```diff
@@ -22,18 +22,16 @@ import (
...
 	apiv1 "k8s.io/api/core/v1"
 	extensionsv1 "k8s.io/api/extensions/v1beta1"
-	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
 	"k8s.io/autoscaler/cluster-autoscaler/clusterstate"
 	"k8s.io/autoscaler/cluster-autoscaler/context"
 	"k8s.io/autoscaler/cluster-autoscaler/estimator"
 	"k8s.io/autoscaler/cluster-autoscaler/expander"
 	"k8s.io/autoscaler/cluster-autoscaler/metrics"
+	ca_processors "k8s.io/autoscaler/cluster-autoscaler/processors"
```
kgolab: Do we need a package alias here? It would be nice to get consistent naming between different files (autoscaler.go does not alias this import, others do).

cluster-autoscaler/core/scale_up.go
```diff
@@ -96,8 +94,10 @@ func ScaleUp(context *context.AutoscalingContext, clusterStateRegistry *clusters
 	podsRemainUnschedulable := make(map[*apiv1.Pod]bool)
 	expansionOptions := make([]expander.Option, 0)
 
-	if context.AutoscalingOptions.NodeAutoprovisioningEnabled {
-		nodeGroups, nodeInfos = addAutoprovisionedCandidates(context, nodeGroups, nodeInfos, unschedulablePods)
+	var errProc error
```
kgolab: Is there any reason to declare this variable separately?

cluster-autoscaler/core/scale_up_test.go
```diff
@@ -368,7 +370,9 @@ func simpleScaleUpTest(t *testing.T, config *scaleTestConfig) {
 		extraPods[i] = pod
 	}
 
-	result, err := ScaleUp(context, clusterState, extraPods, nodes, []*extensionsv1.DaemonSet{})
+	processors := ca_processors.TestProcessors()
```
kgolab: It would be nice if we didn't need this initialization by handling nil/empty values properly inside ScaleUp.

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -200,7 +201,7 @@ func TestStaticAutoscalerRunOnce(t *testing.T) {
 		lastScaleUpTime:       time.Now(),
 		lastScaleDownFailTime: time.Now(),
 		scaleDown:             sd,
-		podListProcessor:      pods.NewDefaultPodListProcessor(),
+		processors:            ca_processors.TestProcessors(),
```
kgolab: Similar to scale_up_test.go - it would be nice if we didn't have to initialize this in all the places where we don't care about exact values (3 out of 4).
Maybe the tests could use the NewStaticAutoscaler method, this might require more refactoring though.

cluster-autoscaler/processors/processors.go
```diff
@@ -0,0 +1,45 @@
...
+
+// DefaultProcessors returns default set of processors.
+func DefaultProcessors() *AutoscalingProcessors {
+	return &AutoscalingProcessors{
+		PodListProcessor:       pods.NewDefaultPodListProcessor(),
+		NodeGroupListProcessor: nodegroups.NewDefaultNodeGroupListProcessor()}
+}
+
+// TestProcessors returns a set of simple processors for use in tests.
+func TestProcessors() *AutoscalingProcessors {
```
kgolab: As commented previously it would be nice to get rid of this function and handle the defaults safely.

It is also slightly confusing that NoOpPodListProcessor is created both here and in DefaultProcessors() but in two different ways.

cluster-autoscaler/core/scale_up.go
MaciekPytel: I don't want to use := operator, because the code relies on overriding existing variables. If we decided that we need to move this call into an if statement for whatever reason, we would suddenly create temporary nodeGroups and nodeInfos variables. I've debugged exactly this kind of error too many times already (and it's the reason why I, personally, consider := operator the worst design choice of golang).

Still, I agree this variable declaration is kinda ugly. If you have a more pretty way of achieving the same, I'm happy to change it.

cluster-autoscaler/processors/processors.go
MaciekPytel: I expect the default implementations to be non-empty for most processors (in particular I still hope to move FilterOutSchedulable to PodListProcessor). And I expect more processors to be added very soon. So I think it makes sense to add a dedicated set of no-op processors for tests (we may change them to mocks if we ever need to).

cluster-autoscaler/core/scale_up_test.go
MaciekPytel: I added handling for nil values. I still like creating this explicitly more than passing nil though.

cluster-autoscaler/core/scale_up.go
MaciekPytel: I like processors as a variable / parameter name. It's very descriptive and better than any other option I came up with. I'll change it if you feel strongly though.

kgolab: I'm fine with ca_processors as long as we keep it consistent - so please update the autoscaler.go import (I think it's the only one not aliased) to have both consistency and nice variable names.

MaciekPytel: done

cluster-autoscaler/core/scale_up.go
MaciekPytel: Just wanted to point out that adding handling for nil values required moving inside if statement, as described in my comment above :) Operator := is pure evil.

PR #904, opened 2018-05-29
Autoscaler doesn't drain nodes that have terminal pods

smarterclayton:
Terminal pods (Succeeded or Failed phase) are drained by kubectl drain,
and autoscaler should also drain them.

I hit this while testing a cluster that created lots of

1. directly created pods with restartPolicy=Never
2. using a custom controller that creates restartPolicy=Never pods as "jobs"

In both cases, the autoscaler left those nodes up, even though all the pods on the node were terminated, with the "is not replicated" error. The autoscaler should not consider terminal pods as blockers for drain.

### DIFF
--- a/cluster-autoscaler/utils/drain/drain.go
+++ b/cluster-autoscaler/utils/drain/drain.go
@@ -77,6 +77,7 @@ func GetPodsForDeletionOnNodeDrain(
 		daemonsetPod := false
 		replicated := false
 		safeToEvict := hasSafeToEvictAnnotation(pod)
+		terminal := isPodTerminal(pod)
 
 		controllerRef := ControllerRef(pod)
 		refKind := ""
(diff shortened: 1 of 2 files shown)


feiskyer:
> t's not clear to me that the lack of a known controller kind on the owner reference should block scale down. CronJobs for instance aren't represented, and any custom controller that runs on the cluster would be incompatible with node autoscaler.

Seems there's no general way to check whether a controller exists or not, especially for custom controllers?

MaciekPytel:
If you use a custom controller you can add `cluster-autoscaler.kubernetes.io/safe-to-evict` annotation to those pods (IIRC the annotation was added specifically for this use-case).

aleksandra-malinowska:
Should we even list failed/succeeded pods? It doesn't seem we need them at all, in fact we already filter them out elsewhere, e.g. https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/utils/kubernetes/listers.go#L169

The only place we don't is here: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/simulator/nodes.go#L38 , which is why this fix was required. @MaciekPytel WDYT? Should we add it to our backlog?

aleksandra-malinowska:
>The call in simulator/nodes.go is only used in scale-up (to make template node), it shouldn't matter for scale-down at all.

Oops. Good point.

But it looks like all other paths use pods obtained via informers (which filter out Failed/Succeeded pods.) So why was this even happening?

PR #912, opened 2018-05-30
Add post scale up status processor.

bskiba:
After a scale up, process the state. This is currently implemented as and eventing processor, that emits appropriate events for pods in different states. The processor is wrapped in a chaning processor for easy extensibility.

### DIFF
--- a/cluster-autoscaler/core/scale_up.go
+++ b/cluster-autoscaler/core/scale_up.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/expander"
 	"k8s.io/autoscaler/cluster-autoscaler/metrics"
 	ca_processors "k8s.io/autoscaler/cluster-autoscaler/processors"
+	"k8s.io/autoscaler/cluster-autoscaler/processors/status"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/glogx"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/nodegroupset"
@@ -41,12 +42,12 @@ import (
 // false if it didn't and error if an error occurred. Assumes that all nodes in the cluster are
 // ready and in sync with instance groups.
 func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.AutoscalingProcessors, clusterStateRegistry *clusterstate.ClusterStateRegistry, unschedulablePods []*apiv1.Pod,
-	nodes []*apiv1.Node, daemonSets []*extensionsv1.DaemonSet) (bool, errors.AutoscalerError) {
+	nodes []*apiv1.Node, daemonSets []*extensionsv1.DaemonSet) (*status.ScaleUpStatus, errors.AutoscalerError) {
 	// From now on we only care about unschedulable pods that were marked after the newest
 	// node became available for the scheduler.
 	if len(unschedulablePods) == 0 {
 		glog.V(1).Info("No unschedulable pods")
-		return false, nil
+		return &status.ScaleUpStatus{ScaledUp: false}, nil
 	}
 
 	now := time.Now()
@@ -63,14 +64,14 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 	nodeInfos, err := GetNodeInfosForGroups(nodes, context.CloudProvider, context.ClientSet,
 		daemonSets, context.PredicateChecker)
 	if err != nil {
-		return false, err.AddPrefix("failed to build node infos for node groups: ")
+		return nil, err.AddPrefix("failed to build node infos for node groups: ")
 	}
 
 	nodeGroups := context.CloudProvider.NodeGroups()
 
 	resourceLimiter, errCP := context.CloudProvider.GetResourceLimiter()
 	if errCP != nil {
-		return false, errors.ToAutoscalerError(
+		return nil, errors.ToAutoscalerError(
 			errors.CloudProviderError,
 			errCP)
 	}
@@ -81,7 +82,7 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 	for nodeGroup, numberOfNodes := range clusterStateRegistry.GetUpcomingNodes() {
 		nodeTemplate, found := nodeInfos[nodeGroup]
 		if !found {
-			return false, errors.NewAutoscalerError(
+			return nil, errors.NewAutoscalerError(
 				errors.InternalError,
 				"failed to find template node for node group %s",
 				nodeGroup)
@@ -99,7 +100,7 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 		var errProc error
 		nodeGroups, nodeInfos, errProc = processors.NodeGroupListProcessor.Process(context, nodeGroups, nodeInfos, unschedulablePods)
 		if errProc != nil {
-			return false, errors.ToAutoscalerError(errors.InternalError, errProc)
+			return nil, errors.ToAutoscalerError(errors.InternalError, errProc)
 		}
 	}
 
@@ -180,13 +181,7 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 
 	if len(expansionOptions) == 0 {
 		glog.V(1).Info("No expansion options")
-		for pod, unschedulable := range podsRemainUnschedulable {
-			if unschedulable {
-				context.Recorder.Event(pod, apiv1.EventTypeNormal, "NotTriggerScaleUp",
-					"pod didn't trigger scale-up (it wouldn't fit if a new node is added)")
-			}
-		}
-		return false, nil
+		return &status.ScaleUpStatus{ScaledUp: false, PodsRemainUnschedulable: getRemainingPods(podsRemainUnschedulable)}, nil
 	}
 
 	// Pick some expansion option.
@@ -204,7 +199,7 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 			glog.V(1).Infof("Capping size to max cluster total size (%d)", context.MaxNodesTotal)
 			newNodes = context.MaxNodesTotal - len(nodes)
 			if newNodes < 1 {
-				return false, errors.NewAutoscalerError(
+				return nil, errors.NewAutoscalerError(
 					errors.TransientError,
 					"max node total count already reached")
 			}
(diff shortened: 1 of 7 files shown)


bskiba:
@kgolab

bskiba:
Fixed the failing test

cluster-autoscaler/processors/status/scale_up_post_processor.go
```diff
@@ -0,0 +1,61 @@
...
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/autoscaler/cluster-autoscaler/context"
+)
+
+// ScaleUpStatus is the status of a scale-up attempt. This includes information
+// on if scale-up happened, description of scale-up operation performed and
+// status of pods that took part in the scale-up evaluation.
+type ScaleUpStatus struct {
+	ScaledUp                bool
+	ScaleUpDescription      []string
```
MaciekPytel: Why not make this ScaleUpInfo? You end up converting them to string specifically so you can pass them to processor. You can do it in processor just as well and having structured information feels more useful than a bunch of strings (in a format designed for human reader, not parsing).

cluster-autoscaler/processors/status/scale_up_post_processor.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// NewDefaultScaleUpPostProcessor creates a default instance of ScaleUpPostProcessor.
+func NewDefaultScaleUpPostProcessor() ScaleUpPostProcessor {
+	return &ScaleUpChainPostProcessor{&EventingScaleUpPostProcessor{}}
+}
+
+// ScaleUpChainPostProcessor processes the state of the cluster after a scale-up
+// by evoking chained ScaleUpPostProcessors.
+type ScaleUpChainPostProcessor []ScaleUpPostProcessor
```
MaciekPytel: Not sure about this. We don't have anything similar for other processors and I think we should either add it to all processors or not at all. Personally I think it's premature to add it now. I'd rather go back and add it everywhere once we really need it somewhere. But I don't feel very strongly about it (I do feel strongly about being consistent between processors though).

cluster-autoscaler/processors/status/scale_up_post_processor.go
```diff
@@ -0,0 +1,61 @@
...
+
+// Process processes the state of the cluster after a scale-up by evoking chained
+// ScaleUpPostProcessors.
+func (cp *ScaleUpChainPostProcessor) Process(context *context.AutoscalingContext, state *ScaleUpStatus) (*context.AutoscalingContext, *ScaleUpStatus, error) {
+	for _, p := range *cp {
+		c, s, err := p.Process(context, state)
+		if err != nil {
+			return context, state, err
```
aleksandra-malinowska: Do we want to exit here? E.g. if we fail to publish an event, do we want to give up on reporting scale-up status via another channel?

cluster-autoscaler/processors/processors.go
```diff
@@ -28,18 +29,23 @@ type AutoscalingProcessors struct {
 	PodListProcessor pods.PodListProcessor
 	// NodeGroupListProcessor is used to process list of NodeGroups that can be used in scale-up.
 	NodeGroupListProcessor nodegroups.NodeGroupListProcessor
+	// ScaleUpStatusProcessor is used to process the state of the cluster after a scale-up.
+	ScaleUpStatusProcessor status.ScaleUpPostProcessor
```
aleksandra-malinowska: I know it may be a controversial question, but is this even a 'processor' in the same sense as the above two? 
Both PodListProcessor and NodeGroupListProcessor modify the list they accept, hopefully without too many side-effects.
The one implementation for ScaleUpStatusProcessor that we have now only produces side-effects (publishing events), without modifying scale-up status. Let's say we want to add another implementation, e.g. responsible for publishing metrics related to scale-up activity. The chain processor implementation has two choices:
a) call the second one on the data returned by the former and fail if the former failed (not good for monitoring),
b) call the second one on the data returned by the former, unless the former failed, then use the original input (not good for consistency),
b) call all of them on the same input (disregarding the returned list, which may be unintuitive.)

So the question is, do we want this processor to modify the status? If yes, whichever processor is called first becomes a single point of failure for monitoring use-case (which is currently the only one.) If no, we probably shouldn't introduce this dependency of scale-up status processors on each other.

MaciekPytel: Not sure why this comment is in this place in code (unless I misunderstand it completely). Assuming I do understand it - all of the above seem like a problem with chaining. I already commented I'd rather remove chaining until we need it. This is one more reason to do it. If we want chaining we should probably explicitly specify what changes and side effects are allowed and how we handle errors.

It's much easier to just say that if someone wants to implement a new processor it would be their job to call the default one and handle the result / errors. In this case the author of code will have full information on what every processor does and there will be no need for a generic solution.

aleksandra-malinowska: I suppose this comment should have been added where `ScaleUpPostProcessor` interface is defined, sorry.

Let's forget about chaining for a moment, then. We have an interface method that:
1) is supposed to return status (and context!),
2) in all places in code where it is called (excluding chaining, 3), the results of this call are disregarded (we don't even check or log error)

If we aren't considering chaining, we definitely shouldn't be returning those values. And if we were, my point is that we should probably design this interface with the intended use-case in mind.

MaciekPytel: Oh, I didn't notice the return values. I think we shouldn't be returning any of those. Most certainly not context. Status only if we think it's meaningful to mutate it (which I don't think it is).

cluster-autoscaler/core/scale_up.go
```diff
@@ -180,11 +181,9 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 
 	if len(expansionOptions) == 0 {
 		glog.V(1).Info("No expansion options")
-		for pod, unschedulable := range podsRemainUnschedulable {
-			if unschedulable {
-				context.Recorder.Event(pod, apiv1.EventTypeNormal, "NotTriggerScaleUp",
-					"pod didn't trigger scale-up (it wouldn't fit if a new node is added)")
-			}
+		if processors != nil && processors.ScaleUpStatusProcessor != nil {
```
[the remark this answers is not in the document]
bskiba: From the alternatives I like returning ScaleUpStatus from scale up the most, if there are no objections I'll go with that.

cluster-autoscaler/processors/status/eventing_scale_up_processor.go
```diff
@@ -0,0 +1,44 @@
...
+
+// EventingScaleUpPostProcessor processes the state of the cluster after
+// a scale-up by emitting relevant events for pods depending on their post
+// scale-up status.
+type EventingScaleUpPostProcessor struct{}
+
+// Process processes the state of the cluster after a scale-up by emitting
+// relevant events for pods depending on their post scale-up status.
+func (p *EventingScaleUpPostProcessor) Process(context *context.AutoscalingContext, state *ScaleUpStatus) (*context.AutoscalingContext, *ScaleUpStatus, error) {
+	if !state.ScaledUp {
```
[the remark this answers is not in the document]
bskiba: Done.

cluster-autoscaler/processors/status/scale_up_post_processor.go
bskiba: Done.

cluster-autoscaler/processors/status/scale_up_post_processor.go
bskiba: Removed.

cluster-autoscaler/processors/status/scale_up_post_processor.go
```diff
@@ -0,0 +1,61 @@
...
+type ScaleUpStatus struct {
+	ScaledUp                bool
+	ScaleUpDescription      []string
+	PodsTriggeredScaleUp    []*apiv1.Pod
+	PodsRemainUnschedulable []*apiv1.Pod
+	PodsAwaitEvaluation     []*apiv1.Pod
+}
+
+// ScaleUpPostProcessor processes the state of the cluster after a scale-up.
+type ScaleUpPostProcessor interface {
```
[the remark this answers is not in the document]
bskiba: Done

cluster-autoscaler/processors/processors.go
bskiba: Removed return values and the chaining processor.

cluster-autoscaler/processors/status/scale_up_post_processor.go
bskiba: Removed

cluster-autoscaler/core/scale_up.go
```diff
@@ -283,32 +278,49 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
...
 	}
-	for pod, unschedulable := range podsRemainUnschedulable {
-		if unschedulable {
-			context.Recorder.Event(pod, apiv1.EventTypeNormal, "NotTriggerScaleUp",
-				"pod didn't trigger scale-up (it wouldn't fit if a new node is added)")
+
+	return false, &status.ScaleUpStatus{PodsRemainUnschedulable: podSetToList(podsRemainUnschedulable)}, nil
+}
+
+func podSetToList(podSet map[*apiv1.Pod]bool) []*apiv1.Pod {
```
aleksandra-malinowska: This doesn't just convert a map to a list, but also filters out keys for which the value is false. How about "getRemainingPods" or sth like this?

cluster-autoscaler/core/scale_up.go
```diff
@@ -63,14 +64,14 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 	nodeInfos, err := GetNodeInfosForGroups(nodes, context.CloudProvider, context.ClientSet,
 		daemonSets, context.PredicateChecker)
 	if err != nil {
-		return false, err.AddPrefix("failed to build node infos for node groups: ")
+		return false, &status.ScaleUpStatus{}, err.AddPrefix("failed to build node infos for node groups: ")
```
aleksandra-malinowska: nit: I think we could just return a nil status on error

cluster-autoscaler/core/scale_up.go
```diff
@@ -41,12 +42,12 @@ import (
 // false if it didn't and error if an error occurred. Assumes that all nodes in the cluster are
 // ready and in sync with instance groups.
 func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.AutoscalingProcessors, clusterStateRegistry *clusterstate.ClusterStateRegistry, unschedulablePods []*apiv1.Pod,
-	nodes []*apiv1.Node, daemonSets []*extensionsv1.DaemonSet) (bool, errors.AutoscalerError) {
+	nodes []*apiv1.Node, daemonSets []*extensionsv1.DaemonSet) (bool, *status.ScaleUpStatus, errors.AutoscalerError) {
```
aleksandra-malinowska: We probably don't need the bool return value anymore now that we have ScaleUpStatus

cluster-autoscaler/core/scale_up.go
bskiba: That's why it's not podMap but podSet, if you treat the map as a set (and having a pod to bool map is an ok option for a set in go AFAIU). However, if it's not immediately clear, I guess I should change it.

cluster-autoscaler/core/scale_up.go
bskiba: Done

cluster-autoscaler/core/scale_up.go
bskiba: Done

bskiba:
If you think this looks good, let me know and I'll squash the commits.

bskiba:
Squashed

bskiba:
@aleksandra-malinowska @MaciekPytel good to go?

Issue #790, opened 2018-04-11
aws - autoscaler failing with Wrong id: expected format aws:///<zone>/<name>


sujithvs74:
I have deployed autoscaler in aws with auto discovery option and application fails with below error,

> I0411 03:07:37.172013       1 polling_autoscaler.go:111] Poll finished
I0411 03:07:37.172031       1 static_autoscaler.go:97] Starting main loop
I0411 03:07:37.353080       1 reflector.go:428] k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes/listers.go:174: Watch close - *v1.Pod total 0 items received
I0411 03:07:37.355165       1 round_trippers.go:436] GET https://10.96.0.1:443/api/v1/pods?fieldSelector=spec.nodeName%21%3D%2Cstatus.phase%21%3DFailed%2Cstatus.phase%21%3DSucceeded&resourceVersion=3644534&timeoutSeconds=422&watch=true 200 OK in 1 milliseconds
W0411 03:07:37.393124       1 clusterstate.go:514] Failed to get nodegroup for dev-k8s-node-asg-230-i-089e4d2f163533989: Wrong id: expected format aws:///<zone>/<name>, got
W0411 03:07:37.393145       1 clusterstate.go:514] Failed to get nodegroup for stg-k8s-w2-npe-master-3: Wrong id: expected format aws:///<zone>/<name>, got
W0411 03:07:37.393152       1 clusterstate.go:514] Failed to get nodegroup for dev-k8s-node-prm-1: Wrong id: expected format aws:///<zone>/<name>, got
W0411 03:07:37.393158       1 clusterstate.go:514] Failed to get nodegroup for dev-k8s-node-asg-230-i-0eb3341fce85be39c: Wrong id: expected format aws:///<zone>/<name>, got
W0411 03:07:37.393164       1 clusterstate.go:514] Failed to get nodegroup for dev-k8s-node-asg-230-i-091d1a037311d5daf: Wrong id: expected format aws:///<zone>/<name>, got
W0411 03:07:37.393169       1 clusterstate.go:514] Failed to get nodegroup for dev-k8s-node-asg-230-i-041dd54f2baaa4553: Wrong id: expected format aws:///<zone>/<name>, got
W0411 03:07:37.393188       1 clusterstate.go:560] Readiness for node group dev-k8s-node-asg-230 not found
[...]

PR #921, opened 2018-06-05
#790 - aws - autoscaler failing with Wrong id

sancyx:
#790 

I've experienced the same problem "autoscaler failing with Wrong id: expected format aws:///<zone>/<name>"  described in #790. This small patched resolved it, as it turned out that some node might not have providerId sometimes, returning an error in this case will result GetNodeInfosForGroups returning an empty map, even if other nodes have correct providerId.

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
@@ -21,6 +21,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/golang/glog"
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -80,6 +80,9 @@ func (aws *awsCloudProvider) NodeGroups() []cloudprovider.NodeGroup {
 
 // NodeGroupForNode returns the node group for the given node.
 func (aws *awsCloudProvider) NodeGroupForNode(node *apiv1.Node) (cloudprovider.NodeGroup, error) {
+	if len(node.Spec.ProviderID) == 0 {
+		return nil, nil
```
aleksandra-malinowska: There should be at least an error-level log message here. Otherwise, the issues with missing ProviderId of nodes will be much harder to debug.

aleksandra-malinowska:
Please sign CLA and verify that all units tests pass.

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
sancyx: I've added some logging, otherwise I agree this is just a quick workaround to solve the missing providerId problem. I think it would be useful warn about this problem but this should not affect other nodes in the node group if they have correct provdeirId's. Still reason of missing id's need to be discovered.

aleksandra-malinowska:
cc @mumoshu

mumoshu:
LGTM.

@aleksandra-malinowska In case the warning message is a important sign of the issue in the k8s, should we make the logger verbose enough to include warning logs by default? I remember that the default CA deployment wrote only fatal logs.

aleksandra-malinowska:
> In case the warning message is a important sign of the issue in the k8s, should we make the logger verbose enough to include warning logs by default?

Probably yes. Default for GCE is --v=4. I think this should be logged at an error level, but leaving it to you to decide as it's in cloud provider code.

PR #926, opened 2018-06-06
[WIP] Gpu cluster wide resource limits (w/fixups)

losipiuk:


cluster-autoscaler/core/scale_down.go
```diff
@@ -434,7 +434,8 @@ func (sd *ScaleDown) TryToScaleDown(allNodes []*apiv1.Node, pods []*apiv1.Pod, p
 
 			nodeCPU, nodeMemory, err := getNodeCoresAndMemory(node)
 			if err != nil {
-				glog.Warningf("Error getting node resources: %v", err)
+				glog.Errorf("Error getting node resources: %v", err)
+				continue
```
aleksandra-malinowska: I'm not at all sure if we should exit here. In the unlikely case there's a node object with corrupted resource data, we may actually prefer to remove it (assuming its pods can be moved and group is above min size), rather than keep it in the cluster forever. While Cluster Autoscaler isn't responsible for healing the cluster, it already does some things like this (e.g. remove unregistered nodes.)

losipiuk: I am not convinced. I agree with the fact that our responsibility is not healing cluster from nodes with broken definitions. 
Also it seems wrong to me to assume that resources are equal to 0 on some random error (which could be due to bug in our code for example).

cluster-autoscaler/cloudprovider/cloud_provider_test.go
```diff
@@ -0,0 +1,31 @@
+/*
+Copyright 2016 The Kubernetes Authors.
```
MaciekPytel: s/2016/2018

cluster-autoscaler/core/scale_down.go
aleksandra-malinowska: Looking at what scheduler does: https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/cache/node_info.go#L150
it seems similar (assuming 0 if there's no CPU in resource list.) So nothing will get scheduled on such node anyway (since minimal CPU requests for all containers are always set by scheduler, even if the user doesn't specify them.)

I think we can agree that deleting a corrupted node where nothing can be scheduled is good (regardless of whose responsibility it is.)

As for the bug in our code scenario, unfortunately it'll still be true that if there's a bug in how we calculate resources, we risk going outside the specified limits (e.g. if we return incorrect value and no error.)

cluster-autoscaler/core/scale_up.go
```diff
@@ -417,6 +477,27 @@ func applyMaxClusterCoresMemoryLimits(newNodes int, coresTotal, memoryTotal, max
 				"max memory already reached")
 		}
 	}
+
+	gpuType, gpuCount, err := gpu.GetNodeTargetGpus(nodeInfo.Node(), nodeGroup)
+	if err != nil {
+		// This is not very elegant, but it allows us to proceed even if we're
+		// unable to compute cpu/memory limits (not breaking current functionality)
```
aleksandra-malinowska: cpu/memory -> gpu

Also, isn't this the same case as with calculating total cluster GPUs (see "not safe to just let NAP go on GPU rampage" comment)? @MaciekPytel

cluster-autoscaler/utils/gpu/gpu.go
```diff
@@ -156,3 +162,59 @@ func GetGpuRequests(pods []*apiv1.Pod) map[string]GpuRequestInfo {
...
+func GetNodeTargetGpus(node *apiv1.Node, nodeGroup cloudprovider.NodeGroup) (gpuType string, gpuCount int64, err error) {
+	gpuLabel, found := node.Labels[GPULabel]
+	if !found {
+		return "", 0, nil
+	}
+	gpuAllocatable, found := node.Status.Allocatable[ResourceNvidiaGPU]
+	if found && gpuAllocatable.Value() > 0 {
+		return gpuLabel, gpuAllocatable.Value(), nil
+	}
+	// A node is supposed to have GPUs (based on label), but it's they're not available yet
```
aleksandra-malinowska: "but they're not"

cluster-autoscaler/utils/gpu/gpu.go
```diff
@@ -156,3 +162,59 @@ func GetGpuRequests(pods []*apiv1.Pod) map[string]GpuRequestInfo {
...
+	// secret.
+	template, err := nodeGroup.TemplateNodeInfo()
+	if err != nil {
+		glog.Errorf("Failed to build template for getting GPU estimation for node %v: %v", node.Name, err)
+		return "", 0, err
+	}
+	if gpuCapacity, found := template.Node().Status.Capacity[ResourceNvidiaGPU]; found {
+		return gpuLabel, gpuCapacity.Value(), nil
+	}
+	errStr := fmt.Sprintf("Node %v with gpu label is not supposed to have GPU", node.Name)
```
aleksandra-malinowska: "is supposed to have"

cluster-autoscaler/core/scale_down.go
```diff
@@ -107,11 +107,22 @@ type resourceName string
 type scaleDownResourcesLimits map[resourceName]int64
 type scaleDownResourcesDelta map[resourceName]int64
 
-func computeScaleDownResourcesLeftLimits(nodes []*apiv1.Node, resourceLimiter *cloudprovider.ResourceLimiter, timestamp time.Time) (scaleDownResourcesLimits, errors.AutoscalerError) {
+func computeScaleDownResourcesLeftLimits(nodes []*apiv1.Node, cp cloudprovider.CloudProvider, timestamp time.Time) (scaleDownResourcesLimits, errors.AutoscalerError) {
```
aleksandra-malinowska: Why pass CloudProvider object here, instead of just ResourceLimiter?

aleksandra-malinowska: Nevermind, didn't notice the calculateGpusTotal call.

cluster-autoscaler/core/scale_down.go
```diff
@@ -103,6 +103,93 @@ func (n *NodeDeleteStatus) SetDeleteInProgress(status bool) {
...
+
+	totals := make(map[resourceName]int64)
+	totals[cloudprovider.ResourceNameCores], totals[cloudprovider.ResourceNameMemory] = calculateCoresAndMemoryTotal(nodes, timestamp)
+
+	for gpuName, gpuCount := range calculateGpusTotal(nodes, cp, timestamp) {
+		totals[resourceName(gpuName)] = gpuCount
+	}
+
+	resultScaleDownLimits := make(scaleDownResourcesLimits)
+	for _, resource := range resourceLimiter.GetResources() {
```
aleksandra-malinowska: I know it's nitpicking, but we call a method here that creates a set of all resources for which either min or max limit is given and returns their names, only to retrieve only non-zero mins for them anyway. What about:

```for resource, min := range resourceLimiter.GetMinResources() {```

where `GetMinResources` would simply return the map with min resources from ResourceLimiter?

File: cluster-autoscaler/main.go [hunk 8]
```diff
@@ -437,3 +451,37 @@ func validateMinMaxFlag(min, max int64) error {
...
+		parsedFlag, err := parseSingleGpuLimit(flag)
+		if err != nil {
+			return nil, err
+		}
+		parsedFlags = append(parsedFlags, parsedFlag)
+	}
+	return parsedFlags, nil
+}
+
+func parseSingleGpuLimit(config string) (context.GpuLimits, error) {
```
MaciekPytel: This function is basically duplicated in gpu.go (ParseGpuLimits). And I'm not sure if I like having it here. On one hand it is consistent with how we handle CPU/mem limits. On the other I don't like having all this code in main.

cluster-autoscaler/context/autoscaling_context.go
```diff
@@ -51,6 +51,16 @@ type AutoscalingContext struct {
 	LogRecorder *utils.LogEventRecorder
 }
 
+// GpuLimits define lower and upper bound on GPU instances of given type in cluster
+type GpuLimits struct {
```
MaciekPytel: This is tied to my comment in main.go. Obviously structured information is more useful than a bunch of strings, but I'm not convinced it should be the responsibility of the code creating AutoscalingOptions, or code using them to actually parse this stuff. Of course right now we use both approaches, which is the worst option :( (parse before creating AutoscalingOptions for limits, parse when creating stuff using AutoscalingOptions for other stuff, for example autodiscovery config).

cluster-autoscaler/core/scale_down.go
```diff
@@ -103,6 +103,93 @@ func (n *NodeDeleteStatus) SetDeleteInProgress(status bool) {
...
+		total := totals[resourceName(resource)]
+		min := resourceLimiter.GetMin(resource)
+
+		// we put only actual limits into final map. No entry means no limit.
+		if min > 0 {
+			resultScaleDownLimits[resourceName(resource)] = total - min
```
MaciekPytel: This can definitely end up being negative. Is that ok? If so I'd appreciate at least a comment saying we're ok with that, because this sort of stuff scares me :)

cluster-autoscaler/core/scale_down.go
```diff
@@ -103,6 +103,93 @@ func (n *NodeDeleteStatus) SetDeleteInProgress(status bool) {
...
+		}
+	}
+	return resultScaleDownLimits, nil
+}
+
+func noLimitsOnResources() scaleDownResourcesLimits {
+	return nil
```
MaciekPytel: Shouldn't that be an empty map, not a nil?

cluster-autoscaler/core/scale_down.go
```diff
@@ -103,6 +103,93 @@ func (n *NodeDeleteStatus) SetDeleteInProgress(status bool) {
...
+}
+
+func copyScaleDownResourcesLimits(source scaleDownResourcesLimits) scaleDownResourcesLimits {
+	copy := scaleDownResourcesLimits{}
+	for k, v := range source {
+		copy[k] = v
+	}
+	return copy
+}
```
MaciekPytel: Missing newline between functions (I'm really surprised gofmt didn't pick that up...)

cluster-autoscaler/core/scale_down.go
```diff
@@ -103,6 +103,93 @@ func (n *NodeDeleteStatus) SetDeleteInProgress(status bool) {
...
+	}
+
+	return scaleDownResourcesDelta{
+		cloudprovider.ResourceNameCores:  nodeCPU,
+		cloudprovider.ResourceNameMemory: nodeMemory,
+		resourceName(gpuType):            gpuCount,
+	}, nil
+}
+
+func (limits *scaleDownResourcesLimits) checkDeltaWithinLimits(delta scaleDownResourcesDelta) *resourceName {
```
MaciekPytel: Why return a pointer? I thought in go strings are already immutable and passed without copy?

MaciekPytel: Also based on name I'd expect this function to return a bool...

cluster-autoscaler/core/scale_down.go
```diff
@@ -103,6 +103,93 @@ func (n *NodeDeleteStatus) SetDeleteInProgress(status bool) {
...
+	for resource, resourceDelta := range delta {
+		resourceLeft, found := (*limits)[resource]
+		if found {
+			if resourceDelta > resourceLeft {
+				return &resource
```
MaciekPytel: In scale-up we want to treat no GPU limit as 0 and never add a node with that GPU. It is less obvious for scale-down though. Can we delete a node with GPU if no limit at all was specified? I slightly lean towards "yes", but it should be a very conscious decision. WDYT @mwielgus @aleksandra-malinowska?

cluster-autoscaler/core/scale_down.go
```diff
@@ -103,6 +103,93 @@ func (n *NodeDeleteStatus) SetDeleteInProgress(status bool) {
...
+			}
+		}
+	}
+	return nil
+}
+
+func (limits *scaleDownResourcesLimits) tryDecrementLimitsByDelta(delta scaleDownResourcesDelta) *resourceName {
```
MaciekPytel: Once again this looks like it should return a bool or an error.

cluster-autoscaler/core/scale_down.go
```diff
@@ -382,9 +450,11 @@ func (sd *ScaleDown) TryToScaleDown(allNodes []*apiv1.Node, pods []*apiv1.Pod, p
 			errors.CloudProviderError,
 			errCP)
 	}
-	coresTotal, memoryTotal := calculateCoresAndMemoryTotal(nodesWithoutMaster, currentTime)
-	coresLeft := coresTotal - resourceLimiter.GetMin(cloudprovider.ResourceNameCores)
-	memoryLeft := memoryTotal - resourceLimiter.GetMin(cloudprovider.ResourceNameMemory)
+
+	scaleDownResourcesLeft, err := computeScaleDownResourcesLeftLimits(nodesWithoutMaster, resourceLimiter, currentTime)
+	if err != nil {
+		return ScaleDownError, err.AddPrefix("failed to obtain nodes resources: ")
```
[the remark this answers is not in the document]
losipiuk: Actually code new code and old code will behave exactly the same in terms of returning from `TryToScaleDown` in case of an error here.

Note the section above (which is removed in subsequent fixup commit)
```golang
 	resourceLimiter, errCP := sd.context.CloudProvider.GetResourceLimiter()
 	if errCP != nil {
 		return ScaleDownError, errors.ToAutoscalerError(
 			errors.CloudProviderError,
 			errCP)
 	}
```
The `computeScaleDownResourcesLeftLimits` will return an error only in exactly same situation.

But even without it I strongly feel we should not mask errors. I would rather see cluster-autoscaler not performing scale-down on some misbehaving cluser. And add specific patch (in cluster autoscaler or in other place on case-by-case basis), to mitigate that specific situation, rather than blindly ignore errors and proceed.

cluster-autoscaler/core/scale_down.go
losipiuk: The semantics is to return name of resource which fails the limit, or nil if everything went fine. 
So pointer is only to have special "nil" value.
I considered returning two values (bool,  string). But found it to verbose.

cluster-autoscaler/core/scale_down.go
```diff
@@ -376,15 +462,10 @@ func (sd *ScaleDown) TryToScaleDown(allNodes []*apiv1.Node, pods []*apiv1.Pod, p
...
 	readinessMap := make(map[string]bool)
 
-	resourceLimiter, errCP := sd.context.CloudProvider.GetResourceLimiter()
-	if errCP != nil {
-		return ScaleDownError, errors.ToAutoscalerError(
-			errors.CloudProviderError,
-			errCP)
+	scaleDownResourcesLeft, err := computeScaleDownResourcesLeftLimits(nodesWithoutMaster, sd.context.CloudProvider, currentTime)
+	if err != nil {
+		return ScaleDownError, err.AddPrefix("failed to obtain nodes resources: ")
```
MaciekPytel: Can you change error message so it's more obvious it's about resource limits?

cluster-autoscaler/core/scale_down.go
losipiuk: same thing :)
go....

cluster-autoscaler/core/scale_down.go
losipiuk: Then I would need to extend the interface with `GetMinResources`. Which is less general. Also I feel that it is better to have methods which do just one thing (separate method for getting resource names and separate for getting min/max value for a resource). Not a method which does two things together.

cluster-autoscaler/core/scale_down.go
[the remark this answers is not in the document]
aleksandra-malinowska: Good point, although I'd rather not exit on error from this function regardless. Can we retrieve resource limiter earlier, and pass it as a parameter?

Mitigation (by limiting impact of errors) is exactly what I mean here. This doesn't need to cause autoscaling downtime, so it probably shouldn't. If observability is a concern, perhaps we can emit an event or publish a metric that will alert us when this happens?

cluster-autoscaler/core/scale_down.go
losipiuk: True. Should do no harm. Yet I will change cap to 0 so it is more readable.

cluster-autoscaler/core/scale_down.go
aleksandra-malinowska: I don't think we always want to enforce specifying resource limits for GPUs in scale-up. Right now they're optional and it may surprise the users badly if autoscaling stops working for node groups with GPUs until they're set.

cluster-autoscaler/core/scale_down.go
losipiuk: I could somewhat agree (but still have very mixed feelings about it) if we were sure that error here means broken nodes  (e.g. nodes which are lacking resource definitions). But we are not sure if that is the case. 
E.g. when we compute number of gpu's used by node we call `nodeGroup.TemplateNodeInfo()`. Which is some arbitrary interface function, which could for example do an API call. And can temporarily return error. I strongly feel that we should not perform any scale_down action for considered node in such case.

IMHO we should count the errors here. Have metric with it so we can spot that something bad is happening. And if such thing happens react on that. Alternative to explicit metric could be a graph based on log grepping.

cluster-autoscaler/main.go
losipiuk: Yeah - actually I forgot to remove the function from `gpu.go` which is not used now. It seems to me that it is valid to have a function which parses user flags around the place which defines those flags.
Should it be main.go ... I am not sure. But as we define (for now) flags here it seems to me that it is valid place.

cluster-autoscaler/context/autoscaling_context.go
losipiuk: IDK. It seems good direction to me. And it is obvious that if we want to move into this direction there will be a phase (probably never ending) where some options are still passed not in a way we would like.
I would not call it worst option. We cannot fix everything. Is that a reason not to fix some things (those we are working on at the moment). 

Above assume that we agree that it is actually direction we want to move on :)

cluster-autoscaler/core/scale_down.go
losipiuk: > In scale-up we want to treat no GPU limit as 0 and never add a node with that GPU.

This is weird. 
I agree with what @aleksandra-malinowska wrote.

cluster-autoscaler/core/scale_down.go
losipiuk: same answer

cluster-autoscaler/core/scale_down.go
aleksandra-malinowska: > Then I would need to extend the interface with `GetMinResources`. 
I don't think we have a Go interface for it, it's just one struct type shared across cloud providers. And `GetResources` is only used here, so it's trading one method for another. 

>  Which is less general. 
Do we have a use case for this (supposedly) more general method?

> Also I feel that it is better to have methods which do just one thing (separate method for getting resource names and separate for getting min/max value for a resource). Not a method which does two things together.

GetMinResources (a simple getter with a one-line implementation):
- returns minResources field.

GetResources (a method that extracts just the names of all resources for which either min or max limit was defined):
- creates a map of structs to act like a set,
- adds resources from maxResources to it,
- adds resources from minResources to it,
- copies the keys from it to a slice.

It's kind of a stretch to say that the former does "two things" and more than the latter. In fact, we could even get rid of it entirely if the ResourceLimiter's fields were public.

cluster-autoscaler/utils/gpu/gpu.go
```diff
@@ -156,3 +162,59 @@ func GetGpuRequests(pods []*apiv1.Pod) map[string]GpuRequestInfo {
...
+	glog.Error(errStr)
+	return "", 0, errors.New(errStr)
```
aleksandra-malinowska: Note: as discussed offline, reconsider error here

cluster-autoscaler/core/scale_down.go
losipiuk: > I don't think we have a Go interface for it, it's just one struct type shared across cloud providers. And GetResources is only used here, so it's trading one method for another.

I did not mean "golang" maning of word "interface". Just test of methods defined on `ResourceLimiter`.

> Do we have a use case for this (supposedly) more general method?

It is just a good principle to have methods which have one responsibility. The `GetMinResources` method you are proposing contradicts that. And we are not winning anything at all IMO (just less readable code - at least to me).

> It's kind of a stretch to say that the former does "two things" and more than the latter.

The inner implementation is irrelevant. By "thing" I mean responsiblity. And given that:
 * `getResources` just lists resource names for which limits are defined (one thing)
 * `getMinResources` lists resource names for which min limit is defined AND lists values of those limits (two things).

losipiuk:
Closing. Superseded by followup PRs.

PR #930, opened 2018-06-07
Add config map to example Addon Resizer deployment

kawych:


addon-resizer/README.md
```diff
@@ -35,6 +35,19 @@ Usage of pod_nanny:
 The following yaml is an example deployment where the nanny watches and resizes itself.
 
 ```yaml
+# Config map for resource configuration.
+# Specify 'cpu', 'extra-cpu', 'memory' and 'extra-memory'
+# to overwrite resource requirements.
+apiVersion: v1
```
losipiuk: Was the example broken without this section?

kawych: No, but I want to include a deployment with config map to give an example how to use it.

losipiuk: Cool. Maybe it is worthy to note the intention in description above the yaml snippet?

PR #934, opened 2018-06-08
Cleanup how multi-string flags are handled in main()

losipiuk:


losipiuk:
Extracted out of https://github.com/kubernetes/autoscaler/pull/926

PR #945, opened 2018-06-11
Resource limits refactor in scale_down

losipiuk:
This is preparatory work before introducing gpu limits to the codebase

### DIFF
--- a/cluster-autoscaler/core/scale_down.go
+++ b/cluster-autoscaler/core/scale_down.go
@@ -27,7 +27,6 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
 	"k8s.io/autoscaler/cluster-autoscaler/clusterstate"
 	"k8s.io/autoscaler/cluster-autoscaler/clusterstate/utils"
-	"k8s.io/autoscaler/cluster-autoscaler/config"
 	"k8s.io/autoscaler/cluster-autoscaler/context"
 	"k8s.io/autoscaler/cluster-autoscaler/metrics"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator"
@@ -44,6 +43,7 @@ import (
 	kube_record "k8s.io/client-go/tools/record"
 
 	"github.com/golang/glog"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 // ScaleDownResult represents the state of scale down.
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/core/scale_down.go
```diff
@@ -103,6 +102,107 @@ func (n *NodeDeleteStatus) SetDeleteInProgress(status bool) {
...
+		}
+	}
+	return resultScaleDownLimits, nil
+}
+
+func calculateCoresAndMemoryTotal(nodes []*apiv1.Node, timestamp time.Time) (int64, int64) {
+	var coresTotal, memoryTotal int64
+	for _, node := range nodes {
+		deleteTime, _ := deletetaint.GetToBeDeletedTime(node)
+		if deleteTime != nil && (timestamp.Sub(*deleteTime) < MaxCloudProviderNodeDeletionTime || timestamp.Sub(*deleteTime) < MaxKubernetesEmptyNodeDeletionTime) {
```
MaciekPytel: side note: this is an ugly check that we probably need to do all over the place. Maybe we should add IsNodeBeingDeleted() function? Let's not do it in this PR though - it's large enough as is.

cluster-autoscaler/core/scale_down.go
```diff
@@ -382,9 +481,11 @@ func (sd *ScaleDown) TryToScaleDown(allNodes []*apiv1.Node, pods []*apiv1.Pod, p
 			errors.CloudProviderError,
 			errCP)
 	}
-	coresTotal, memoryTotal := calculateCoresAndMemoryTotal(nodesWithoutMaster, currentTime)
-	coresLeft := coresTotal - resourceLimiter.GetMin(cloudprovider.ResourceNameCores)
-	memoryLeft := memoryTotal - resourceLimiter.GetMin(cloudprovider.ResourceNameMemory)
+
+	scaleDownResourcesLeft, err := computeScaleDownResourcesLeftLimits(nodesWithoutMaster, resourceLimiter, currentTime)
+	if err != nil {
+		return ScaleDownError, err.AddPrefix("failed to obtain nodes resources: ")
```
MaciekPytel: I'd like the error message to be more descriptive. We probably get node resources all over the place and just looking at this error I wouldn't guess it means 'we failed to calculate resource limits for scale-down'.

MaciekPytel:
General comments:
1. Let's use k8s.io/apimachinery/pkg/util/sets for string sets (we already do it in cloudprovider). 
2. I prefer to print all exceeded resources, not just the one we happen to hit first in a loop.

cluster-autoscaler/core/scale_down.go
losipiuk: Actually this call should not return error.
We will only mark resources for which we could not compute limit due to error as `limitUnknown`.
Changed.

MaciekPytel:
Except for travis failing lgtm. Please squash before merge. 

@aleksandra-malinowska your last chance to disagree :)

cluster-autoscaler/core/scale_down.go
```diff
@@ -103,6 +102,112 @@ func (n *NodeDeleteStatus) SetDeleteInProgress(status bool) {
...
+
+// used as a value in scaleDownResourcesLimits if actual limit could not be obtained due to errors talking to cloud provider
+const limitUnknown = math.MinInt64
+
+func computeScaleDownResourcesLeftLimits(nodes []*apiv1.Node, resourceLimiter *cloudprovider.ResourceLimiter, timestamp time.Time) scaleDownResourcesLimits {
+	totals := make(map[resourceName]int64)
+	totals[cloudprovider.ResourceNameCores], totals[cloudprovider.ResourceNameMemory] = calculateCoresAndMemoryTotal(nodes, timestamp)
+	resultScaleDownLimits := make(scaleDownResourcesLimits)
+
+	for _, resource := range resourceLimiter.GetResources() {
```
aleksandra-malinowska: I still think it doesn't make sense to recompute the map we already have here (see discussion at https://github.com/kubernetes/autoscaler/pull/926/files#r193725222 .) It adds unnecessary ~30 lines of code (plus 1 file, if counting tests) and complicates rather than simplifies the logic. It's probably otherwise harmless, though.

cluster-autoscaler/core/scale_down.go
aleksandra-malinowska: +1

cluster-autoscaler/core/scale_down.go
losipiuk: Just for sake of not going back and forth with that I will stay with this shape. And as we finish scale_up let's take a look how it looks like and if look redundant then I will prepare a cleanup commit. Ok?

cluster-autoscaler/core/scale_down.go
```diff
@@ -102,28 +103,27 @@ func (n *NodeDeleteStatus) SetDeleteInProgress(status bool) {
...
 
 func computeScaleDownResourcesLeftLimits(nodes []*apiv1.Node, resourceLimiter *cloudprovider.ResourceLimiter, timestamp time.Time) scaleDownResourcesLimits {
-	totals := make(map[resourceName]int64)
+	totals := make(map[string]int64)
 	totals[cloudprovider.ResourceNameCores], totals[cloudprovider.ResourceNameMemory] = calculateCoresAndMemoryTotal(nodes, timestamp)
 	resultScaleDownLimits := make(scaleDownResourcesLimits)
 
 	for _, resource := range resourceLimiter.GetResources() {
-		total := totals[resourceName(resource)]
+		total := totals[string(resource)]
```
MaciekPytel: This converts string to string.

cluster-autoscaler/core/scale_down.go
aleksandra-malinowska: Let's leave it as it is if you prefer, I just disagree:) We can probably revisit it if we ever refactor ResourceLimiter

PR #947, opened 2018-06-11
Update Go version to 1.10.2

aleksandra-malinowska:
Update Go version to 1.10.2 (required by k8s.io/kubelet)


MaciekPytel:
Does VPA also use builder? If so does it work with go1.10?

aleksandra-malinowska:
Doesn't seem to use it. Only CA Makefile references it.

PR #950, opened 2018-06-11
Respect GPU limits in scale_down

losipiuk:


cluster-autoscaler/core/scale_down.go
```diff
@@ -44,6 +44,8 @@ import (
 
 	"github.com/golang/glog"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/gpu"
+	"k8s.io/kubernetes/bazel-kubernetes/pkg/kubelet/kubeletconfig/util/log"
```
MaciekPytel: We use glog literally everywhere. Please don't introduce different logging packages.

cluster-autoscaler/core/scale_down.go
```diff
@@ -146,6 +177,49 @@ func calculateCoresAndMemoryTotal(nodes []*apiv1.Node, timestamp time.Time) (int
...
+		if isNodeBeingDeleted(node, timestamp) {
+			// Nodes being deleted do not count towards total cluster resources
+			continue
+		}
+		nodeGroup, err := cp.NodeGroupForNode(node)
+		if err != nil {
+			return nil, fmt.Errorf("can not get node group for node %v when calculating cluster gpu usage (%s)", node.Name, err.Error())
+		}
+		if nodeGroup == nil || reflect.ValueOf(nodeGroup).IsNil() {
+			return nil, fmt.Errorf("group not found for node %v when calculating cluster gpu usage", node.Name)
```
MaciekPytel: I think we should log and continue here. NodeGroupForNode() returns nil for every non-autoscaled node (explicitly said here: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/cloud_provider.go#L44). Having non-autoscaled nodes is not an error condition and shouldn't prevent autoscaling from working.

cluster-autoscaler/core/scale_down.go
```diff
@@ -163,13 +237,21 @@ func copyScaleDownResourcesLimits(source scaleDownResourcesLimits) scaleDownReso
...
+	resultScaleDownDelta[cloudprovider.ResourceNameMemory] = nodeMemory
 
-	return scaleDownResourcesDelta{
-		cloudprovider.ResourceNameCores:  nodeCPU,
-		cloudprovider.ResourceNameMemory: nodeMemory,
-	}, nil
+	if containsGpuResources(resourcesWithLimits) {
+		gpuType, gpuCount, err := gpu.GetNodeTargetGpus(node, nodeGroup)
+		if err != nil {
+			return scaleDownResourcesDelta{}, errors.ToAutoscalerError(errors.ApiCallError, err).AddPrefix("Failed to get node %v gpu: %v", node.Name)
```
MaciekPytel: The only place that can return an error in GetNodeTargetGpus is a cloudprovider call, so this should be a CloudProviderError.

cluster-autoscaler/core/scale_down.go
```diff
@@ -163,13 +237,21 @@ func copyScaleDownResourcesLimits(source scaleDownResourcesLimits) scaleDownReso
 	return copy
 }
 
-func computeScaleDownResourcesDelta(node *apiv1.Node) (scaleDownResourcesDelta, errors.AutoscalerError) {
+func computeScaleDownResourcesDelta(node *apiv1.Node, nodeGroup cloudprovider.NodeGroup, resourcesWithLimits []string) (scaleDownResourcesDelta, errors.AutoscalerError) {
```
MaciekPytel: You only create resourcesWithLimits, so you can pass it here. I think passing either ResourceLimiter or a bool would be cleaner than this sort of half-calculated value.

MaciekPytel: I guess a bool would be more efficient, given that we call this in a loop. I suspect it doesn't matter too much though.

cluster-autoscaler/main.go
```diff
@@ -445,3 +452,37 @@ func validateMinMaxFlag(min, max int64) error {
 func minMaxFlagString(min, max int64) string {
 	return fmt.Sprintf("%v:%v", min, max)
 }
+
+func parseMultipleGpuLimits(flags MultiStringFlag) ([]context.GpuLimits, error) {
```
MaciekPytel: Ideally we should have some unittest for parsing. I can accept adding it later if you create an issue for it.

losipiuk: Good catch. I have - some. Just forgot to backport it from previous PR version.

cluster-autoscaler/core/scale_down.go
MaciekPytel: I thought about it some more. The problem is as follows: NodeGroup is needed to predict the number of GPUs on a node that is currently installing GPU driver. We don't have node NodeGroup for non autoscaled node and so we won't be able to guess how many GPUs it will have once drivers are installed. We can't fix this without some large refactors, that I disagree with both in principle and for practical reasons. We can still get the number from Capacity once the driver is installed.

The simplest thing to do is to change GetNodeTargetGpus to accept a nil NodeGroup. It would than return an error if it actually got to the point where it needs to call TemplateNodeInfo and NodeGroup is nil. In most cases we would be perfectly fine. During driver installation on non-autoscaled Node we would return an error resulting in setting limitUnknown and effectively stopping GPU autoscaling until the driver is installed. It's not perfect, but the scenario is unlikely and I agreed with @mwielgus that it's acceptable as a temporary limitation.

Also it's gpu.go, which is already the hack el dorado.

Anyway - WDYT? Do you see anything I missed?

Finally if you do end up implementing that idea, please return an InternalError if the hack is triggered (AutoscalerError is still an error and ToAutoscalerError does type assertion and only changes type if it's not an AutoscalerError).

cluster-autoscaler/utils/gpu/gpu.go
```diff
@@ -156,3 +158,36 @@ func GetGpuRequests(pods []*apiv1.Pod) map[string]GpuRequestInfo {
...
+		return gpuLabel, gpuAllocatable.Value(), nil
+	}
+
+	// A node is supposed to have GPUs (based on label), but they're not available yet
+	// (driver haven't installed yet?).
+	// Unfortunately we can't deduce how many GPUs it will actually have from labels (just
+	// that it will have some).
+	// Ready for some evil hacks? Well, you won't be disappointed - let's pretend we haven't
+	// seen the node and just use the template we use for scale from 0. It'll be our little
+	// secret.
```
MaciekPytel: I appreciate you leaving my original comment here :+1:

cluster-autoscaler/main.go
losipiuk: Hmm - actually I must have dreamt something. I see no tests. Will add some.

cluster-autoscaler/utils/gpu/gpu.go
losipiuk: I left whole method. Why would I remove such a beatiful comment :)

cluster-autoscaler/core/scale_down.go
losipiuk: > I guess a bool would be more efficient, given that we call this in a loop. I suspect it doesn't matter too much though

What do you refer to here?

losipiuk: > You only create resourcesWithLimits, so you can pass it here. I think passing either ResourceLimiter or a bool would be cleaner than this sort of half-calculated value

I used to have it that way. But I changed it to pass an array here because of usage of method in `getEmptyNodes`. I do not have `ResourceLimiter` there.

MaciekPytel: Right, I missed that. By bool I meant having passing something like `applyGpuLimits bool`. I think it's slightly more readable, but honestly it doesn't really matter. So I guess I'm fine with both.

cluster-autoscaler/core/scale_down.go
losipiuk: I changed `GetNodeTargetGpus` itself to return `AutoscalerError`. And also the two places which use the method - but after initial change the error type is not important any more.

cluster-autoscaler/core/scale_down.go
losipiuk: seems reasonable. Unless sth is screwed up which would imply that node will never get the gpu capacity (even though it contain gpu label) the situation should be self healing.

cluster-autoscaler/core/scale_down.go
losipiuk: >  So I guess I'm fine with both.

I will leave it then :)

losipiuk:
Addressed everything (I believe). Please take a look at fixups.

cluster-autoscaler/utils/gpu/gpu.go
```diff
@@ -156,3 +158,46 @@ func GetGpuRequests(pods []*apiv1.Pod) map[string]GpuRequestInfo {
...
+	template, err := nodeGroup.TemplateNodeInfo()
+	if err != nil {
+		glog.Errorf("Failed to build template for getting GPU estimation for node %v: %v", node.Name, err)
+		return "", 0, errors.ToAutoscalerError(errors.CloudProviderError, err)
+	}
+	if gpuCapacity, found := template.Node().Status.Capacity[ResourceNvidiaGPU]; found {
+		return gpuLabel, gpuCapacity.Value(), nil
+	}
+
+	// if template does not define gpus we assume node will not have any even if ith has gpu label
```
aleksandra-malinowska: nit: maybe we could log a warning here? Just so we know it's happening if we're ever debugging a cluster with such nodes.

cluster-autoscaler/core/scale_down.go
```diff
@@ -109,27 +110,57 @@ type scaleDownResourcesDelta map[string]int64
...
-			if total > min {
-				resultScaleDownLimits[resource] = total - min
-			} else {
-				resultScaleDownLimits[resource] = 0
+			switch {
+			case resource == cloudprovider.ResourceNameCores:
+				resultScaleDownLimits[resource] = computeLeft(totalCores, min)
+			case resource == cloudprovider.ResourceNameMemory:
+				resultScaleDownLimits[resource] = computeLeft(totalMem, min)
+			case cloudprovider.IsGpuResource(resource):
```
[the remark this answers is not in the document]
losipiuk: IDK if we will be able to do it  better later. But I would rather have this logic in single place. That is `cloudprovider` module which already defines names for other resource types (CPU, GPU).
If we later have better way to distinguish `GPU` resources from other types - only `IsGpuResource` method will need to be changed. Not `scale_down.go` and `scale_up.go`.

cluster-autoscaler/core/scale_down.go
```diff
@@ -146,6 +177,58 @@ func calculateCoresAndMemoryTotal(nodes []*apiv1.Node, timestamp time.Time) (int
...
+		if isNodeBeingDeleted(node, timestamp) {
+			// Nodes being deleted do not count towards total cluster resources
+			continue
+		}
+		nodeGroup, err := cp.NodeGroupForNode(node)
+		if err != nil {
+			return nil, errors.ToAutoscalerError(errors.CloudProviderError, err).AddPrefix("can not get node group for node %v when calculating cluster gpu usage", node.Name)
+		}
+		if reflect.ValueOf(nodeGroup).IsNil() {
+			// TODO[lukaszos] why do we need this? do we?
```
aleksandra-malinowska: Welcome to Golang: https://golang.org/doc/faq#nil_error

cluster-autoscaler/core/scale_down.go
```diff
@@ -146,6 +177,58 @@ func calculateCoresAndMemoryTotal(nodes []*apiv1.Node, timestamp time.Time) (int
...
+			nodeGroup = nil
+		}
+
+		var gpuType string
+		var gpuCount int64
+
+		cacheHit := false
+		if nodeGroup != nil {
+			cached, found := ngCache[nodeGroup.Id()]
```
aleksandra-malinowska: Either cacheHit or found is redunant, e.g.:
```cached, cacheHit := ngCache[nodeGroup.Id()]```

File: cluster-autoscaler/core/scale_down.go (hunk 8 above)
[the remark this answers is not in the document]
Comment (MaciekPytel) [reply cluster-autoscaler/core/scale_down.go]: +1 for encapsulating the hack in cloudprovider and not explicitly doing it here.

File: cluster-autoscaler/core/scale_down.go (hunk 10 above)
Comment (aleksandra-malinowska) [reply cluster-autoscaler/core/scale_down.go]: Nevermind, the scope doesn't work out.

File: cluster-autoscaler/core/scale_down.go (hunk 9 above)
Comment (MaciekPytel) [reply cluster-autoscaler/core/scale_down.go]: I understand why we need to use reflect to check for nil, what I don't understand is why do we need this if at all? What I'm seeing is:

```
if x == a {
  x = a
}
```
Am I missing something obvious here???

File: cluster-autoscaler/core/scale_down.go (hunk 10 above)
Comment (MaciekPytel) [reply cluster-autoscaler/core/scale_down.go]: I usually deal with that by declaring both variables and avoiding the pure, vile evil that is := with multiple assignement.
```
var cached, found bool
if stuff() {
  cached, found = whatever()
}
```
I'm fine with @losipiuk's solution too. Also - operator :=, scopes and multiple assignement is the single largest design flow of golang :(

cluster-autoscaler/core/scale_down.go
aleksandra-malinowska: The second nil is a nil interface, so it can be safely used later

losipiuk: Added comment

cluster-autoscaler/core/scale_down.go
losipiuk: Changed

MaciekPytel:
lgtm, please squash and I'll merge

losipiuk:
> lgtm, please squash and I'll merge
Done

PR #955, opened 2018-06-12
Restructure checking resource limits in scale_up.go

losipiuk:
Preparatory work for before introducing GPU limits

### DIFF
--- a/cluster-autoscaler/core/scale_up.go
+++ b/cluster-autoscaler/core/scale_up.go
@@ -18,10 +18,13 @@ package core
 
 import (
 	"bytes"
+	"fmt"
+	"math"
 	"time"
 
 	apiv1 "k8s.io/api/core/v1"
 	extensionsv1 "k8s.io/api/extensions/v1beta1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
 	"k8s.io/autoscaler/cluster-autoscaler/clusterstate"
 	"k8s.io/autoscaler/cluster-autoscaler/context"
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/core/scale_up.go
```diff
@@ -35,9 +35,129 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/utils/nodegroupset"
 	schedulercache "k8s.io/kubernetes/pkg/scheduler/cache"
 
+	"fmt"
```
MaciekPytel: The import order is stdlib, k8s, other.

cluster-autoscaler/core/scale_up.go
```diff
@@ -35,9 +35,129 @@ import (
...
+	resultScaleUpLimits := make(scaleUpResourcesLimits)
+	for _, resource := range resourceLimiter.GetResources() {
+		max := resourceLimiter.GetMax(resource)
+
+		// we put only actual limits into final map. No entry means no limit.
+		if max > 0 {
+			switch {
+			case resource == cloudprovider.ResourceNameCores:
+				if errCoresMem != nil {
+					resultScaleUpLimits[resource] = scaleUpLimitUnknown
```
MaciekPytel: Setting unknown instead of immediately failing whole scale-up with error only makes sense if we consider nodes without CPU? I know it's technically possible to have a Node object without CPU resource, but I'm not sure how practical it is to consider that. Also IIRC in scale-down you don't allow unknown CPU.

Same for memory.

cluster-autoscaler/core/scale_up.go
```diff
@@ -35,9 +35,129 @@ import (
...
+			continue
+		}
+		if currentSize > 0 {
+			nodeCPU, nodeMemory := getNodeInfoCoresAndMemory(nodeInfo)
+			coresTotal = coresTotal + int64(currentSize)*nodeCPU
+			memoryTotal = memoryTotal + int64(currentSize)*nodeMemory
+		}
+	}
+
+	if failed {
```
MaciekPytel: Is there any reason for setting `failed` to true and returning error here as opposed to immediately returning error?

cluster-autoscaler/core/scale_up.go
```diff
@@ -368,57 +487,36 @@ func executeScaleUp(context *context.AutoscalingContext, clusterStateRegistry *c
...
+		limit, limitFound := scaleUpResourcesLeft[resource]
+		if !limitFound {
+			continue
+		}
+		if limit == scaleUpLimitUnknown {
+			// should never happen - checked before
 			return 0, errors.NewAutoscalerError(
 				errors.TransientError,
-				"max cores already reached")
+				fmt.Sprintf("limit unknown for resource %s", resource))
```
MaciekPytel: This should be an InternalError.

cluster-autoscaler/core/scale_up.go
```diff
@@ -368,57 +487,36 @@ func executeScaleUp(context *context.AutoscalingContext, clusterStateRegistry *c
...
+		newNodes = int(limit / resourceDelta)
+		glog.V(1).Infof("Capping scale-up size due to limit for resource %s", resource)
 		if newNodes < 1 {
-			// This should never happen, as we already check that
-			// at least one node will fit when considering nodegroup
+			// should never happen - checked before
 			return 0, errors.NewAutoscalerError(
 				errors.TransientError,
-				"max memory already reached")
+				fmt.Sprintf("max limit for resource %s reached", resource))
```
MaciekPytel: I'd make that an InternalError as well.

cluster-autoscaler/core/scale_up_test.go
```diff
@@ -809,3 +809,57 @@ func TestScaleUpAutoprovisionedNodeGroup(t *testing.T) {
...
+		{
+			limits:            scaleUpResourcesLimits{"a": 10},
+			delta:             scaleUpResourcesDelta{"a": 10},
+			exceededResources: []string{},
+		},
+		{
+			limits:            scaleUpResourcesLimits{"a": 10},
+			delta:             scaleUpResourcesDelta{"a": 10},
+			exceededResources: []string{},
+		},
```
MaciekPytel: Duplicate test case?

cluster-autoscaler/core/scale_up.go
losipiuk: More logging I guess.

cluster-autoscaler/core/scale_up.go
```diff
@@ -196,7 +201,10 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 			errCP)
 	}
 
-	scaleUpResourcesLeft := computeScaleUpResourcesLeftLimits(nodeGroups, nodeInfos, resourceLimiter)
+	scaleUpResourcesLeft, errLimits := computeScaleUpResourcesLeftLimits(nodeGroups, nodeInfos, resourceLimiter)
+	if errLimits != nil {
+		return nil, errors.ToAutoscalerError(errors.InternalError, errLimits).AddPrefix("Could not compute total resources: ")
```
MaciekPytel: Shouldn't that be a cloudprovider error?

MaciekPytel:
lgtm, please squash

MaciekPytel:
/lgtm

It looks like travis is failing due to upstream problem in golint installation: https://github.com/golang/go/issues/25872. This is not related to this PR and verify passes locally, so I'm merging this.

PR #959, opened 2018-06-13
Typo fix: automaticaly -> automatically

mirake:


bskiba:
Do we use/need this script anyway? @MaciekPytel @aleksandra-malinowska

aleksandra-malinowska:
I don't think we do. @mwielgus should we remove it?

mwielgus:
Yes we can remove.

aleksandra-malinowska:
Removed in #963

PR #960, opened 2018-06-13
Move backoff mechanism to utils.

krzysztof-jastrzebski:


cluster-autoscaler/utils/backoff/backoff.go
```diff
@@ -0,0 +1,82 @@
...
+
+// Backoff handles backing off executions.
+type Backoff struct {
+	maxBackoffDuration     time.Duration
+	initialBackoffDuration time.Duration
+	backoffResetTimeout    time.Duration
+	backoffInfo            map[string]backoffInfo
+}
+
+func NewBackoff(maxBackoffDuration time.Duration, initialBackoffDuration time.Duration, backoffResetTimeout time.Duration) *Backoff {
```
aleksandra-malinowska: golint requires comment

cluster-autoscaler/utils/backoff/backoff_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+func TestMaxBackoff(t *testing.T) {
+	backoff := NewBackoff(3*time.Minute, 1*time.Minute, 3*time.Hour)
+	startTime := time.Now()
+	backoff.Backoff("key1", startTime)
+	assert.True(t, backoff.IsBackedOff("key1", startTime))
+	assert.False(t, backoff.IsBackedOff("key1", startTime.Add(1*time.Minute)))
+	backoff.Backoff("key1", startTime.Add(1*time.Minute))
+	assert.True(t, backoff.IsBackedOff("key1", startTime.Add(1*time.Minute)))
+	assert.False(t, backoff.IsBackedOff("key1", startTime.Add(3*time.Minute)))
```
aleksandra-malinowska: nit: 2*time.Minute

aleksandra-malinowska: Feel free to ignore, it just seems that this assertion would pass even if max backoff was ignored. Next one would catch it though:)

cluster-autoscaler/utils/backoff/backoff_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBackoffTwoKeys(t *testing.T) {
+	backoff := NewBackoff(time.Hour, 10*time.Minute, 3*time.Hour)
```
aleksandra-malinowska: nit: this (and every other test case in this file) effectively tests only max backoff, since it's always larger than init. Changing it here to max > init would improve coverage of that scenario, e.g.:
```NewBackoff(time.Hour, 10*time.Minute, 3*time.Hour)```

aleksandra-malinowska: We could also use a test case of incremental backoff scenario, sth like:
- backoff for init duration
- check that after this duration backoff is lifted
- backoff again
- check that this time backoff duration is 2x

Doesn't have to be in this PR, just a general note.

aleksandra-malinowska:
A couple nits regarding unit tests, otherwise it looks good.

MaciekPytel:
Also looks good to me, after you address @aleksandra-malinowska comments.

cluster-autoscaler/utils/backoff/backoff.go
krzysztof-jastrzebski: done

PR #961, opened 2018-06-13
Scale down works. Mocked rancher api calls.

jepp2078:


jepp2078:
Ignore.
PR'ed by mistake.

Working on adding rancher as a cloud-provider.

PR #974, opened 2018-06-15
Add GPU-related scaled_up & scaled_down metrics

kgolab:


MaciekPytel:
Unittests are failing

cluster-autoscaler/utils/gpu/gpu.go
```diff
@@ -68,6 +68,21 @@ func FilterOutNodesWithUnreadyGpus(allNodes, readyNodes []*apiv1.Node) ([]*apiv1
 	return newAllNodes, newReadyNodes
 }
 
+// GetGpuType returns name of the GPU used on the node or empty string if there's no GPU
+// if the GPU type is unknown, "generic" is returned
+// NOTE: current implementation is GKE/GCE-specific
+func GetGpuType(node *apiv1.Node) string {
```
MaciekPytel: Why add new method if we already have GetNodeTargetGpus? Especially if you already use output of GetNodeTargetGpus in one place and the output of this function in another place?

kgolab: TBH I just haven't noticed it.

But the logic behind both is a little bit different - here I just need GPU type and not count, I'd also prefer not to touch Cloud Provider since it's not needed for the name itself.

Additionally, if I understand correctly, GetNodeTargetGpus returns non-zero value only in GKE whereas I'd like the metrics to work everywhere, even if they are not as detailed.

kgolab:
Rebased & fixed the import which broke compilation.

cluster-autoscaler/utils/gpu/gpu.go
MaciekPytel: The problem is this won't work reliably. GPU capacity / allocatable is only set on node after the GPU driver is installed, so it's not a reliable way of telling if the underlying VM has a GPU.

kgolab: As discussed - I've switched the logic to use label first but still support "generic" GPU if the node has GPU capacity set.

cluster-autoscaler/utils/gpu/gpu.go
```diff
@@ -68,6 +68,24 @@ func FilterOutNodesWithUnreadyGpus(allNodes, readyNodes []*apiv1.Node) ([]*apiv1
 	return newAllNodes, newReadyNodes
 }
 
+// GetGpuTypeForMetrics returns name of the GPU used on the node or empty string if there's no GPU
+// if the GPU type is unknown, "generic" is returned
+// NOTE: current implementation is GKE/GCE-specific
+func GetGpuTypeForMetrics(node *apiv1.Node) string {
+	// we use the GKE label if there is one
+	gpuType, found := node.Labels[GPULabel]
```
aleksandra-malinowska: IIUC the issue here is that just because a node has this label doesn't mean it'll have GPU. We treat such nodes as not having a GPU when making scale-up/scale-down decisions. Our metrics should hopefully be consistent with it (or alternatively, report it as a possible misconfiguration.)

aleksandra-malinowska: I agree scale-up and scale-down should be consistent as much as possible. What would be the inconsistency if we added checking template here?

kgolab: Please have a look at PR #1017 which addresses this discussion.

PR #983, opened 2018-06-18
 Add checkpoint versioning

kgrygiel:
1. Fill the VerticalPodAutoscalerCheckpointStatus.Version field.
2. Do not load checkpoint with incompatible version.


bskiba:
I can merge, but I'm kinda unhappy we don't have any tests for cluster_feeder at all :(

PR #985, opened 2018-06-19
Improve histogram exponential buckets.

kgrygiel:
Currently the bucket boundaries are calculated as: FirstBucketSize * ExpFactor^n.
For example for FirstBucketSize=100M and ExpFactor=1.05, the bucket boundaries will be:
0:  [0...100M)
1:  [100M...105M)
2:  [105M...110.25M)
etc..
Natural and more reasonable boundaries would instead be such that the second bucket is 5% larger than the first bucket, so:
0:  [0...100M)
1:  [100M...205M)
2:  [205M...315.25M)
etc.

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/util/histogram_options.go
+++ b/vertical-pod-autoscaler/pkg/recommender/util/histogram_options.go
@@ -52,8 +52,10 @@ func NewLinearHistogramOptions(
 
 // NewExponentialHistogramOptions returns HistogramOptions describing a
 // histogram with exponentially growing bucket boundaries. The first bucket
-// covers the range [0..firstBucketSize). Consecutive buckets are of the form
-// [x(n)..x(n) * ratio) for n = 1 .. numBuckets - 1.
+// covers the range [0..firstBucketSize). Bucket with index n has size equal to firstBucketSize * ratio^n.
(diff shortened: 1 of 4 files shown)


vertical-pod-autoscaler/pkg/recommender/util/histogram_options.go
```diff
@@ -109,25 +111,30 @@ func (o *exponentialHistogramOptions) NumBuckets() int {
...
 
+// Returns the index of the bucket for given value. This is the inverse function to
+// GetBucketStart(), which yields the following formula for the bucket index:
+//    bucket(value) = floor(log(value/firstBucketSize*(ratio-1)+1) / log(ratio))
 func (o *exponentialHistogramOptions) FindBucket(value float64) int {
 	if value < o.firstBucketSize {
 		return 0
 	}
-	bucket := int(math.Log(value/o.firstBucketSize)/math.Log(o.ratio)) + 1
+	bucket := int(math.Log(value/o.firstBucketSize*(o.ratio-1)+1) / math.Log(o.ratio))
```
tkulczynski: These formulas got pretty complex. I would recommend splitting to a few statements with named subparts or even providing a separate short function to calculate the bucket number.

PR #999, opened 2018-06-20
update the aws ec2 instance types

mtougeron:
Ran `make generate` to update the available ec2 instance types in aws.


mtougeron:
thanks!

PR #1000, opened 2018-06-21
Move VPA default tag to 0.2.0

bskiba:


bskiba:
@mwielgus @schylek @kgrygiel

mwielgus:
Please also update the yaml files.

bskiba:
Done

vertical-pod-autoscaler/deploy/admission-controller-deployment.yaml
```diff
@@ -13,7 +13,7 @@ spec:
       serviceAccountName: vpa-admission-controller
       containers:
       - name: admission-controller
-        image: k8s.gcr.io/vpa-admission-controller:v0.1.0
+        image: k8s.gcr.io/vpa-admission-controller:0.2.0
```
mwielgus: Why are we removing v from the TAG?

bskiba: This seems to be the convention for other images in the k8s.gcr.io/ repo

PR #1007, opened 2018-06-22
VPA split min resources across containers in a pod

schylek:


vertical-pod-autoscaler/pkg/recommender/logic/estimator.go
```diff
@@ -126,18 +135,24 @@ func (e *confidenceMultiplier) GetResourceEstimation(s *model.AggregateContainer
...
 		newResources[resource] = originalResources[resource] + margin
 	}
 	return newResources
 }
+
+func (e *minResourcesEstimator) GetResourceEstimation(s *model.AggregateContainerState) model.Resources {
+	originalResources := e.baseEstimator.GetResourceEstimation(s)
+	newResources := make(model.Resources)
+	for resource, resourceAmount := range originalResources {
+		if resourceAmount < e.minResources[resource] {
```
mwielgus: why e.maxResources is not included?

schylek: This is minimum resources recommended to a pod. Recommendation has no max boundary (not to confuse with min/maxAllowed that is part of resourcePolicy).

PR #1012, opened 2018-06-22
Fix construction of the pollPeriod duration

vantuvt:
The construction of the pollPeriod duration was done in a manner that
did not use the unit of the poll-period flag correctly.

### DIFF
--- a/addon-resizer/nanny/main/pod_nanny.go
+++ b/addon-resizer/nanny/main/pod_nanny.go
@@ -73,7 +73,7 @@ func main() {
 	checkPercentageFlagBounds("recommendation-offset", *recommendationOffset)
 	checkPercentageFlagBounds("acceptance-offset", *acceptanceOffset)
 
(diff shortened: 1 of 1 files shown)


addon-resizer/nanny/main/pod_nanny.go
```diff
@@ -73,7 +73,7 @@ func main() {
 	checkPercentageFlagBounds("recommendation-offset", *recommendationOffset)
 	checkPercentageFlagBounds("acceptance-offset", *acceptanceOffset)
 
-	pollPeriod := time.Millisecond * time.Duration(*pollPeriodMillis)
+	pollPeriod := time.Duration(int64(*pollPeriodMillis) * int64(time.Millisecond))
```
mwielgus: Why these casts are needed?

vantuvt: Without the casts the compiler throws the following error:

invalid operation: *pollPeriodMillis * time.Millisecond (mismatched types int and time.Duration)

PR #1017, opened 2018-06-26
Make GetGpuTypeForMetrics more robust

kgolab:
As discussed I've made GetGpuTypeForMetrics more in line with GetNodeTargetGpus - but looking at the amount of plumbing needed I'm no longer fully convinced it's really worth the effort.

Updated metrics.md for a good measure.

### DIFF
--- a/cluster-autoscaler/utils/gpu/gpu.go
+++ b/cluster-autoscaler/utils/gpu/gpu.go
@@ -43,6 +43,8 @@ const (
 	MetricsUnknownGPU = "not-listed"
 	// MetricsErrorGPU - for when there was an error obtaining GPU type
 	MetricsErrorGPU = "error"
+	// MetricsInternalErrorGPU - for when there was a CA-internal error obtaining GPU type
+	MetricsInternalErrorGPU = "own-error"
 	// MetricsNoGPU - for when there is no GPU at all
 	MetricsNoGPU = ""
 )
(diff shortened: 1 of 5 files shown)


kgolab:
Could you please have a look?

aleksandra-malinowska:
Going over this again, I changed my mind a bit. Let me know if this makes sense:

For nodes with labels:
1. If the node has capacity, I don't believe we need validation of GPU name.
2. If the node doesn't have capacity, then if template has capacity (only possible in scale-down), it means GPU didn't appear in expected time (error on our side).
3. If the node doesn't have capacity, and template doesn't have capacity, it means there's a problem with configuration.

(2) would be a useful error to track. However, if we don't care about distinguishing between (2) and (3) at this time, we can simplify a lot:
- label + capacity - publish metric with label value (assumed GPU name)
- label + no capacity - publish metric with error
- no label + capacity - publish metric with unknown GPU
- no label + no capacity - don't publish anything

cluster-autoscaler/utils/gpu/gpu.go
```diff
@@ -71,19 +93,51 @@ func FilterOutNodesWithUnreadyGpus(allNodes, readyNodes []*apiv1.Node) ([]*apiv1
...
+	capacity, capacityFound := node.Status.Capacity[ResourceNvidiaGPU]
+
+	// GKE-specific label & capacity are present - consistent state
+	if labelFound && capacityFound {
+		return validateGpuType(gpuType)
+	}
+	// GKE-specific label present but no capacity (yet?) - check the node template
+	if labelFound {
+		if nodeGroup == nil && template == nil {
+			return MetricsErrorGPU
```
aleksandra-malinowska: As discussed offline, let's use another constant to distinguish the cases

kgolab: Done.

PR #1018, opened 2018-06-27
VPA add flag switch for setting memory limit in admission-controller

schylek:


vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider_test.go
```diff
@@ -65,6 +65,9 @@ func TestUpdateResourceRequests(t *testing.T) {
 	vpaWithEmptyRecommendation := vpaBuilder.Get()
 	vpaWithEmptyRecommendation.Status.Recommendation = &vpa_types.RecommendedPodResources{}
 
+	checkLimit := true
+	setMemoryLimit = &checkLimit
```
mwielgus: Do we have a tests that checks what happens if flag is false?

schylek: Test case added.

Issue #789, opened 2018-04-10
AWS: updated instance type doesn't update cluster-autoscaler instance template


julianvmodesto:
In AWS, I updated my instance type for my autoscaling group's launch configuration and doubled the cpu and memory.

However, this didn't update the cluster-autoscaler instance template and I see that my Pods still won't schedule even though they'll fit on the new instance type:

```
  Normal   NotTriggerScaleUp  6s                 cluster-autoscaler  pod didn't trigger scale-up (it wouldn't fit if a new node is added)
```

Version:
```
          image: gcr.io/google_containers/cluster-autoscaler:v1.0.4
```

aleksandra-malinowska:
Were there any instances in that group that weren't yet updated? Cluster Autoscaler uses an existing node from a group as "template", and falls back to reading group's template only if there are no existing nodes in it.

julianvmodesto:
Ah yes, all but one of my nodes haven't been updated. My intent wasn't to rolling-update them, but only update the launch configuration so only new nodes are created with more resources.

Would it be feasible to change this default behavior and reverse it? What would the implications / side effects be?

From a user's intent, it's closer to what I want e.g. as a user, I expected cluster-autoscaler to scale up since if it did, the new template would be used, not the old one.

aermakov-zalando:
I agree with @julianvmodesto. The group template (launch configuration in AWS) is a much better representation on what the newly created node would look like, and this is what matters when scaling up. It also simplifies the mental model of how the autoscaler behaves and gets rid of the nasty edge cases when node configuration is changed. For example, our cluster upgrade process sets a `lifecycle-status=decommission-pending` label on all nodes that would be removed during the upgrade and new nodes come up with `lifecycle-status=ready`. However, one of the users decided to add a node selector matching only `lifecycle-status=ready` nodes (which is something you would expect to work) and during the next cluster upgrade the pods were stuck in Pending without triggering scale up.

tecnobrat:
I was trying to find out if there is a way to do something where when and ASG it updated to a new launch template that the cluster autoscaler noticed that the LT has changed, and has new instance types on it, or simply just a new AMI.  Then autoscaler is responsible for decommissioning and terminating instances one by one in a rolling fashion.

The idea is that autoscaler is responsible for my nodes, and also ensuring that my nodes are running based on the latest LT.

aweeks:
[...]

PR #1021, opened 2018-06-28
Scale up: ignore existing nodes

aermakov-zalando:
Add a new flag, `--scale-up-cloud-provider-template`, that forces the autoscaler to always construct the template node from the cloud provider data (e.g. AWS launch configuration) instead of a random existing node. This fixes both the trivial issues where instance type change is not taken into account until the user gets lucky with node ordering and more complicated ones where pods using node affinities sometimes don't trigger scale up (see comments in #789).

I think that this should be the default behaviour, because the current one is non-intuitive and hard to debug. However, since the users might rely on existing logic and use node selectors without the corresponding tags on the ASGs, the flag defaults to `false`.

One more change I've made selects a random AZ when constructing a template node for a multi-AZ ASG in the AWS cloud provider. Previously, the autoscaler always used the first AZ, but since this was masked by the default logic of using a random existing node it *sort of* worked when existing nodes got added or removed. Using a random AZ allows the autoscaler to progress even in these non-ideal setups.

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_manager.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_manager.go
@@ -214,7 +214,7 @@ func (m *AwsManager) getAsgTemplate(asg *asg) (*asgTemplate, error) {
 		return nil, fmt.Errorf("Unable to get first AvailabilityZone for %s", asg.Name)
 	}
 
-	az := asg.AvailabilityZones[0]
+	az := asg.AvailabilityZones[rand.Intn(len(asg.AvailabilityZones))]
 	region := az[0 : len(az)-1]
 
 	if len(asg.AvailabilityZones) > 1 {
--- a/cluster-autoscaler/context/autoscaling_context.go
+++ b/cluster-autoscaler/context/autoscaling_context.go
@@ -146,6 +146,9 @@ type AutoscalingOptions struct {
 	ExpendablePodsPriorityCutoff int
 	// Regional tells whether the cluster is regional.
 	Regional bool
+	// ScaleUpTemplateFromCloudProvider tells if template node should be built from the up-to-date provider configuration (e.g. ASG launch configuration)
+	// instead of a random existing node.
+	ScaleUpTemplateFromCloudProvider bool
 }
 
 // NewResourceLimiterFromAutoscalingOptions creates new instance of cloudprovider.ResourceLimiter
--- a/cluster-autoscaler/core/scale_up.go
+++ b/cluster-autoscaler/core/scale_up.go
@@ -250,7 +250,13 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 		podsRemainUnschedulable[pod] = true
 	}
 	glogx.V(1).Over(loggingQuota).Infof("%v other pods are also unschedulable", -loggingQuota.Left())
-	nodeInfos, err := GetNodeInfosForGroups(nodes, context.CloudProvider, context.ClientSet,
+
+	existingNodes := nodes
+	if context.ScaleUpTemplateFromCloudProvider {
+		existingNodes = []*apiv1.Node{}
+	}
+
+	nodeInfos, err := GetNodeInfosForGroups(existingNodes, context.CloudProvider, context.ClientSet,
 		daemonSets, context.PredicateChecker)
 	if err != nil {
 		return nil, err.AddPrefix("failed to build node infos for node groups: ")
--- a/cluster-autoscaler/core/scale_up_test.go
+++ b/cluster-autoscaler/core/scale_up_test.go
@@ -887,6 +887,118 @@ func TestScaleUpAutoprovisionedNodeGroup(t *testing.T) {
 	assert.Equal(t, "autoprovisioned-T1-1", getStringFromChan(expandedGroups))
 }
 
+func TestScaleUpAutoprovisionFavorExistingNodes(t *testing.T) {
+	expandedGroups := make(chan string, 10)
+
+	p1 := BuildTestPod("p1", 5000, 0)
+
+	fakeClient := &fake.Clientset{}
+
+	existingNode := BuildTestNode("n1", 4000, 1000000)
+	SetNodeReadyState(existingNode, true, time.Time{})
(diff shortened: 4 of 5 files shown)


aermakov-zalando:
Rebased on master. Any updates?

MaciekPytel:
Sorry for late reply, I was on vacation and completely missed your reply. 

First of all - there are a lot of implementations of cloudprovider (3 + 2 in progress in main repo and I know about at least 3 different providers that implemented their cloudprovider in fork). We've been GA for a year now and it's effectively our API. I'm ok-ish with extending the interface with new optional methods, but I'm against any major changes (especially backward incompatible ones) without an extremely good reason. Also if we were to do such changes we probably should go through a deprecation process, which makes me even less happy about it.

Fundamentally, a NodeGroup is a set of identical, interchangeable nodes. It's the central assumption in CA and a we depend on that both in core code and various cloudprovider-specific places. Just create a separate NodeGroup with a different instance type - CA works quite well with multiple NodeGroups. It's the same advice we give people who want multi-zone clusters: use a separate ASG in each zone instead of multi-zonal ASGs.

I'm more sympathetic to the labels and taints use-case. I can see a lot of reasons for temporarily tainting a node and CA should play better with it. The simplest solution would be to just say that any label or taint prefixed with `custom.cluster-autoscaler.kubernetes.io/` (or something like that) would be ignored when creating template node. Would that help? If not we can think about adding some optional method to cloudprovider that filters out labels and taints, but that's likely to bring some complexity (in my experience an extra API call per NodeGroup per loop is likely to be a scalability problem and caching makes things complex).

aermakov-zalando:
I still think that what the current CA behaviour is a surprising and undocumented hack that's very painful when you actually run into it. Software should be as simple as possible and not try to magically guess for the user which is what CA does now (badly!).

I see your point about not wanting to change the provider interface. What about reconciling the difference in the CA code itself? You could still keep the provider interface the same as before, where CA asks it to provide _a_ template node, and then merge it with the existing nodes:
 * labels/taints from the existing will be overwritten by the template node's labels/taints (if configured)
 * `allocatable` can be computed in a smarter way (e.g. get it from the latest existing node if the instance type is the same as the template's one, or just recompute it using the existing bad heuristics)

This should make it possible to implement better logic while keeping the provider interface intact, even though it doesn't cover more complicated future improvements (e.g. node pools with different instance types, which is something that'd be possible with AWS Spot Fleets, for example). For this you'd actually need to implement scale-up-from-zero calculation properly.

MaciekPytel:
I think the problem once again boils to the definition of NodeGroup. A NodeGroup is not an ASG, it is a scalable set of identical nodes. Nothing in the core logic expects it to be anything more and I don't think we want to change that - the interface is meant to be a common denominator between clouds and there are significant differences between ASG, MIG, VMSS, etc.

I'd argue that once we assume the nodes in NodeGroup are identical using a copy of an existing node as a template is the simplest solution. I can explain it in a single sentence to someone who doesn't know CA, which is not the case for combining allocatable of existing node with taints and labels from TemplateNodeInfo with some sort of fallback if one or the other is unavailable. 

If we want to support cloud provider specific features like changing instance type in ASG perhaps the right way to do this would be to have cloudprovider implementation that handles it behind the scenes, representing machines with different instance type as a separate NodeGroup. There is no requirement for NodeGroup to map 1-1 to ASG, as long as we manage to implement things like TargetSize() in a consistent way. 

---

Taints are a special case, because they can represent both node config (ex. tainting node with expensive resources to limit what pods can schedule on this node) and state (some temporary condition of the node). If it's the latter we should discard the taint. The problem is how to distinguish those - I dislike both the flag and the prefix approach proposed in https://github.com/kubernetes/autoscaler/issues/1350, but I don't have any better idea. We don't have any CRD or dynamic config that we could use to pass those and we're not likely to add it in predictable future.

aermakov-zalando:
> I'd argue that once we assume the nodes in NodeGroup are identical using a copy of an existing node as a template is the simplest solution.

Yes, but it's also a bad solution, unless you expect the cluster operators to never change taints or labels. And the way it's completely automagical is even worse, because it will work most of the time so people won't even read the documentation and just rely on it, until they change a label or add a taint and then their cluster suddenly stops scaling for a couple of pods. Once that happens it's also impossible to fix it without either getting rid of affinities/selectors, babysitting the cluster to manually scale it up or bringing up nodes until you get lucky and the provider gives you a node with a name that'll sort before the "broken" one.

You now also have two code paths, one for `scale-up-from-0` and another one for `magically-copy-from-existing-node`. One of them might be exercised a lot less often, so the configuration getting out of sync could be noticed at the worst possible moment, triggering an incident. If CA used just one set of truth instead of trying to be smart, the error would've been noticed immediately.

I would also argue that the assumption that all the nodes in the node pool are identical is also dubious at best. Yes, it works for GKE because you have tooling that allows you to treat node pools as immutable and migrate from one to another. Is every user of CA supposed to implement the same tooling now just to achieve something as simple as changing the instance type? Or are the users supposed to manually babysit the nodes, draining them one by one after creating the fiftieth copy of the same ASG even though AWS easily allows changing an instance type of an existing ASG? The only real problem, as you've said, is the capacity estimation of new nodes, but there's no real requirement that all nodes spawned from an ASG look exactly the same as existing ones, they just have to look exactly the same as each other so the scale-up calculations can work correctly.

losipiuk:
As we discussed during sig-meeting, it seems like an option to be able to influence how the function of predicting next node shape for give node group looks like, may came handy.
On the other hand it is not obvious that everyone needs the same logic.

Given that we suggest implementing adding new `*Processor` interface which will be called in places when we need to determine the node shape for node pool.
Then we can have default implementation of processor which basically replicates current logic of CA.
And we can have another implementation which allows handling Zalando case (preferably in as generic way as possible - so it is not limited to Zalando only). 
The processor implementation as well as maybe some parametrization of the processor itself would be done using CA program flags.

As an example how we use `Processors` to have pluggable logic you may want to look at `NodeGroupSetProcessor` which was introduced by https://github.com/kubernetes/autoscaler/commit/6f5e6aab6fe3129e01d5edce9f0cf0308e7e94d1

@MaciekPytel, do you want to add something?

mwielgus:
No activity for almost 2 months, closing. Please reopen once needed.

PR #1025, opened 2018-07-02
Handle nil recommendation in admission controller.

bskiba:


bskiba:
@schylek

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
```diff
@@ -134,10 +134,14 @@ func (p *recommendationProvider) GetContainersResourcesForPod(pod *v1.Pod) ([]Co
 		glog.V(2).Infof("no matching VPA found for pod %s", pod.Name)
 		return nil, "", nil
 	}
-	recommendedPodResources, err := p.recommendationProcessor.Apply(vpaConfig.Status.Recommendation, vpaConfig.Spec.ResourcePolicy, pod)
-	if err != nil {
-		glog.V(2).Infof("cannot process recommendation for pod %s", pod.Name)
-		return nil, "", err
+	recommendedPodResources := &vpa_types.RecommendedPodResources{}
+	var err error
```
schylek: Why standalone var definition? ":=" 2 lines below should work AFAIU

bskiba: This is weirdness of the := operator. If I do it like that, the recommendedPodResources gets redefined inside the if and when compiling go will tell me that recommendedPodResources in line 140 is defined but not used :(

schylek: Ok, makes sense. However "var err error" could be probably defined inside "if" to reduce the distance between definition and use.

bskiba: Done

PR #1031, opened 2018-07-03
Fix the parsing of resource values

vantuvt:
When parsing a resource value such as "0.5m" the unit information was
being extracted as "00u" instead of "u".

### DIFF
--- a/addon-resizer/nanny/estimator.go
+++ b/addon-resizer/nanny/estimator.go
@@ -19,6 +19,7 @@ package nanny
 import (
 	"fmt"
 
+	log "github.com/golang/glog"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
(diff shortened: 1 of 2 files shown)


x13n:
Please add a test for that.

vantuvt:
@x13n - Refactored the code and added unit tests. PTAL. Thanks.

addon-resizer/nanny/estimator.go
```diff
@@ -18,6 +18,7 @@ package nanny
 
 import (
 	"fmt"
+	"log"
```
x13n: nanny_lib.go uses glog, can you use it here too, to stay consistent?

addon-resizer/nanny/estimator_test.go
```diff
@@ -239,3 +239,23 @@ func TestEstimateResources(t *testing.T) {
...
 	}
 }
+
+func TestComputeResourceOverheadValueString(t *testing.T) {
+	testCases := []struct {
+		numNodes         uint64
+		resource         Resource
+		expectedOverhead string
+	}{
+		{1, Resource{Base: resource.MustParse("1"), ExtraPerNode: resource.MustParse("1"), Name: "cpu"}, "1.000000"},
```
x13n: This function computes per node overhead, so Base values are not needed in test cases.

addon-resizer/nanny/estimator.go
vantuvt: Done.

addon-resizer/nanny/estimator_test.go
vantuvt: Done.

wojtek-t:
Please also squash commits.

vantuvt:
@wojtek-t - Squashed.

PR #1043, opened 2018-07-05
Refactor and extend GPU metrics error types

aleksandra-malinowska:
Refactor and extend GPU metrics error types to better track missing GPUs.

### DIFF
--- a/cluster-autoscaler/utils/gpu/gpu.go
+++ b/cluster-autoscaler/utils/gpu/gpu.go
@@ -21,7 +21,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
-	schedulercache "k8s.io/kubernetes/pkg/scheduler/cache"
 
 	"github.com/golang/glog"
 )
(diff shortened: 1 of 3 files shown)


aleksandra-malinowska:
cc @kgolab  - a couple of last-minute changes, let me know if I'm breaking something :)

cluster-autoscaler/utils/gpu/gpu.go
```diff
@@ -95,45 +96,44 @@ func FilterOutNodesWithUnreadyGpus(allNodes, readyNodes []*apiv1.Node) ([]*apiv1
...
 	gpuType, labelFound := node.Labels[GPULabel]
 	capacity, capacityFound := node.Status.Capacity[ResourceNvidiaGPU]
 
+	if !labelFound {
+		// no label, fallback to generic solution
+		if capacityFound && !capacity.IsZero() {
+			return MetricsGenericGPU
+		}
+
+		// no signs of GPu
```
kgolab: /nit

MaciekPytel:
At this point (multiple different error states), I'd define a separate 'state' label, rather than push all the different errors as fake 'gpu types'. That being said this is not doing anything the previous version wasn't doing already and it's not worth the effort to refactor this whole thing again.

/lgtm

kgolab:
LGTM, too.
As discussed off-line I understand that we don't want to merge "missing GPU" with label value to have both error & name of the problematic GPU.
Maybe introducing separate state would indeed be good but let's keep it simple for now.

PR #1047, opened 2018-07-05
Add support for custom min cluster sizes

vantuvt:
Add a command line flag allowing the use of an ExponentialEstimator with
a custom minimum cluster size. The default minimum cluster size is 16 to
retain compatibility with existing behaviour. Also tweak some unit test
error logging to make it easier to diagnose failures.

### DIFF
--- a/addon-resizer/nanny/estimator.go
+++ b/addon-resizer/nanny/estimator.go
@@ -43,19 +43,28 @@ func (e LinearEstimator) scaleWithNodes(numNodes uint64) *corev1.ResourceRequire
 	return calculateResources(numNodes, e.Resources)
 }
 
-// ExponentialEstimator estimates the amount of resources in the way that
-// prevents from frequent updates but may end up with larger resource usage
-// than actually needed (though no more than ScaleFactor).
+// ExponentialEstimator estimates resource requirements in a way that prevents
+// frequent updates but may end up with larger estimates than actually needed.
 type ExponentialEstimator struct {
-	Resources   []Resource
+	// The collection of resources to provide estimates for.
+	Resources []Resource
+	// The minimum cluster size for which the estimator will provide
+	// resource estimates. Must be greater than 1.
+	MinClusterSize uint64
+	// The multiplier used to compute the next cluster size to provide
+	// estimates for. For example, suppose the cluster has 9 nodes,
+	// MinClusterSize is 5, and ScaleFactor is 1.5. Then the estimator will
+	// provide estimates for FLOOR(5 * 1.5 * 1.5) = 11 nodes since that is
+	// the smallest cluster size larger than 9 nodes.
 	ScaleFactor float64
 }
 
 func (e ExponentialEstimator) scaleWithNodes(numNodes uint64) *corev1.ResourceRequirements {
-	n := uint64(16)
+	n := e.MinClusterSize
 	for n < numNodes {
 		n = uint64(float64(n)*e.ScaleFactor + eps)
 	}
+
 	return calculateResources(n, e.Resources)
 }
 
--- a/addon-resizer/nanny/main/pod_nanny.go
+++ b/addon-resizer/nanny/main/pod_nanny.go
@@ -55,8 +55,9 @@ var (
 	podName       = flag.String("pod", os.Getenv("MY_POD_NAME"), "The name of the pod to watch. This defaults to the nanny's own pod.")
 	containerName = flag.String("container", "pod-nanny", "The name of the container to watch. This defaults to the nanny itself.")
 	// Flags to control runtime behavior.
-	pollPeriod = time.Millisecond * time.Duration(*flag.Int("poll-period", 10000, "The time, in milliseconds, to poll the dependent container."))
-	estimator  = flag.String("estimator", "linear", "The estimator to use. Currently supported: linear, exponential")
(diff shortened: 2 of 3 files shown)


addon-resizer/nanny/main/pod_nanny.go
```diff
@@ -55,8 +55,9 @@ var (
 	podName       = flag.String("pod", os.Getenv("MY_POD_NAME"), "The name of the pod to watch. This defaults to the nanny's own pod.")
 	containerName = flag.String("container", "pod-nanny", "The name of the container to watch. This defaults to the nanny itself.")
 	// Flags to control runtime behavior.
-	pollPeriod = time.Millisecond * time.Duration(*flag.Int("poll-period", 10000, "The time, in milliseconds, to poll the dependent container."))
-	estimator  = flag.String("estimator", "linear", "The estimator to use. Currently supported: linear, exponential")
+	pollPeriod     = time.Millisecond * time.Duration(*flag.Int("poll-period", 10000, "The time, in milliseconds, to poll the dependent container."))
+	estimator      = flag.String("estimator", "linear", "The estimator to use. Currently supported: linear, exponential")
+	minClusterSize = flag.Uint64("minClusterSize", 16, "Specifies the smallest number of nodes resources will be scaled to.")
```
x13n: Please add a note that this is for exponential estimator only.

addon-resizer/nanny/estimator.go
```diff
@@ -43,19 +43,28 @@ func (e LinearEstimator) scaleWithNodes(numNodes uint64) *corev1.ResourceRequire
...
+	// estimates for. For example, suppose the cluster has 9 nodes,
+	// MinClusterSize is 5, and ScaleFactor is 1.5. Then the estimator will
+	// provide estimates for FLOOR(5 * 1.5 * 1.5) = 11 nodes since that is
+	// the smallest cluster size larger than 9 nodes.
 	ScaleFactor float64
 }
 
 func (e ExponentialEstimator) scaleWithNodes(numNodes uint64) *corev1.ResourceRequirements {
-	n := uint64(16)
+	n := e.MinClusterSize
```
x13n: With this change, if anyone passes 0 or 1 as a flag value, this will go into infinite loop. Probably not what one might expect.

addon-resizer/nanny/main/pod_nanny.go
vantuvt: Done.

vantuvt:
Looking at the cluster growth rates, I think it is reasonable to restrict the value of the minClusterSize to to >= 4. That eliminates the degenerate cases (0, 1) and eliminates scenarios where the cluster is scaled up or down by 1 node. Thoughts?

x13n:
It doesn't eliminate scenarios where the cluster is scaled by 1 node, it merely enables scenarios where resize by 1 doesn't always restart Heapster/Metrics Server. If someone sets this to 4, then cluster size change from 4 to 5 and back will still trigger 2 restarts.

vantuvt:
IIUC, the situation you're describing is one where a cluster starts out at 4 nodes and is then manually increased to 5 nodes. This triggers the resizer to use the next bucket size, which would be 6 so a restart is needed. Then changing it back from 5 to 4 triggers another restart because the previous bucket size of 4 is sufficient. Assuming my understanding is correct, wouldn't this happen even with the existing implementation? Suppose your cluster starts at 16 nodes and you add 1 node and then remove it. Wouldn't the resizer go between the buckets of 16 and 24 nodes also triggering two restarts?

x13n:
That is correct. This problem wasn't solved in addon resizer 1.8, but it wasn't a huge concern because bucket of (1, 16) was good enough for a large number of clusters, and the risk of frequent restarts in bigger clusters was smaller, because the bucket size is growing exponentially. Making first bucket smaller will increase the impact of this problem.

vantuvt:
I feel we can reasonably ignore the corner case where a cluster is sitting right on a bucket boundary and document that as a limitation. I agree that setting a smaller value for the minClusterSize flag can result in more frequent occurrences. However, the default still remains set at 16. So for most (?) users nothing will change. Users who want to experiment with smaller values OTOH will need to be more aware of the potential issues that may arise. Although setting a fixed lower bound is a reasonable thing to do.

x13n:
Yes, the default doesn't change, but I guess later you'd like to change the value of that flag, which is when the problem will manifest itself. Now may be a good moment to think how this can be mitigated.

vantuvt:
Agreed that more thought around mitigation is required. Beyond enforcing a lower bound to handle the degenerate cases, I believe that discussion is orthogonal to the mechanism used to support smaller bucket sizes being introduced in this PR. Specifically, different scenarios will likely require different mitigations.

x13n:
Yeah, the mitigation needs to be in place before the flag value changes, not when the flag is introduced. I won't block this PR on that. Please handle 0&1 cases though.

addon-resizer/nanny/estimator.go
vantuvt: Done.

x13n:
Looks good. Please squash commits before merging.

vantuvt:
@x13n - Thanks for the review! Squashed.

vantuvt:
@x13n - A quick follow-up here. I could be mistaken, but I don't believe I have write privileges to the repository. So I'm going to need someone to help me with the merge. Can you please let me know who can help me with that? Thanks.

vantuvt:
@MaciekPytel - Friendly ping. Can you please help me get this PR merged into the addon-resizer-release-1.8 branch? Thanks.

PR #1048, opened 2018-07-05
fix spelling and indentation

pauloancheta:
Small fix for spelling and indentation


pauloancheta:
Finished cncf-cla yesterday. Also confirmed that my email is the same as my git commit

Issue #510, opened 2017-12-22
Add annotation to stop the autoscaler from moving a specific pod


vainu-arto:
There is an annotation to stop specific nodes from being removed, but not one for marking pods as "do not touch". The "cluster-autoscaler.kubernetes.io/safe-to-evict": "true" annotation only seems to work one way (you can use it to mark a pod as safe to move, not to mark a pod as unsafe to move).

MaciekPytel:
You can achieve this with PodDisruptionBudget. Just set maxUnavailable to 0 and CA will never touch those pods.

vainu-arto:
That works, but affects a lot of other things as well (such as draining nodes with this pod on them). I'd prefer having a specific method for affecting just the cluster autoscaler.

DerrickMartinez:
+1 Having this would be awesome. Having an annotation not to touch it would be nice

MaciekPytel:
I still think using PDB is the right way to do it, but I'm fine accepting a PR if someone feels like contributing one. Feel free to ping me on k8s slack if you need any help.

vainu-arto:
How about for example long running Job pods that should not be interrupted? I can't seem to find documentation about how PDBs interact with jobs, do you have any hints?

vainu-arto:
No, based on tests using a PDB to protect a Job does not appear to be supported at least on Kubernetes 1.9. The controllermanager logs "found no controllers for pod "test"" and "Failed to calculate the number of expected pods: found no controllers for pod "test"".

Having this annotation available to mark long running jobs as not interruptible by the autoscaler seems like the simple solution to this issue, but others exist as well. One option would be to make the cluster autoscaler use a NoExecute taint on the node instead of manually deleting each pod. In that case it would be possible to simply add a toleration to pods that should not be removed.

Yet another alternative would be to simply add the NoSchedule taint as is currently done and then wait for the currently running jobs to complete. Both of these seem rather inefficient, though.

MaciekPytel:
PDB not working with Jobs is indeed a very good reason for this annotation. Thanks for your PR, merged.

PR #1054, opened 2018-07-11
Add annotation to mark pods as unsafe to evict

vainu-arto:
Allow using the cluster-autoscaler.kubernetes.io/safe-to-evict annotation also to mark the pod as not safe to evict, this fixes issue #510.


vainu-arto:
@MaciekPytel Thanks!

What is the release process like, is there a chance of including this in a 1.1-series release at some point?

MaciekPytel:
The easiest way would be if you opened cherry-pick PRs against release branches. I'm not sure when next 1.1-series release will happen, but if your PR is already on branch it will be included.

prat0318:
@MaciekPytel It has been cherry-picked to 1.1 on July 12 but the last release of 1.1 is still 1.1.3 which was on July 4. Any future timelines on when 1.1.4 would be cut?

sdonataccio-handy:
Hello, I'm also curious about a 1.1.4 release that includes this change.

PR #1071, opened 2018-07-19
WIP: upgrade AKS API to 2018-03-31

### LINKED ISSUE #1051: [Azure AKS] Scaling fails with "Changing property 'networkProfile.networkPlugin' is not allowed when using Advanced Networking
v1.2.2 fails with AKS when configured with Advanced Networking:

```
E0708 09:51:50.493997       1 azure_container_service_pool.go:262] containerservice.ManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="PropertyChangeNotAllowed" Message="Changing property 'networkProfile.networkPlugin' is not allowed."
W0708 09:51:50.494266       1 clusterstate.go:249] Disabling scale-up for node group nodepool1 until 2018-07-08 09:56:50.494249582 +0000 UTC
E0708 09:51:50.506118       1 static_autoscaler.go:304] Failed to scale up: failed to increase node group size: containerservice.ManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="PropertyChangeNotAllowed" Message="Changing property 'networkProfile.networkPlugin' is not allowed."
```

I rembember this was issue with old AZ CLI version as well so I assume there is change in API call needed for latest AKS clusters.

feiskyer:
This PR upgrades AKS API to new 2018-03-31 (which is required for clusters with custom VNET) together with `go-autorest` v10.12.0.

It also updates AKS/ACS agent pools to use new versioned APIs.

Fix #1051


feiskyer:
Still WIP (validating together with aks team).

/sig azure
/cc @khenidak @kkmsft

cluster-autoscaler/Godeps/Godeps.json
```diff
@@ -71,33 +71,33 @@
 		},
 		{
 			"ImportPath": "github.com/Azure/go-autorest/autorest",
-			"Comment": "v10.6.2",
-			"Rev": "1ff28809256a84bb6966640ff3d0371af82ccba4"
+			"Comment": "v10.12.0",
```
aleksandra-malinowska: This version doesn't match the one in kubernetes/kubernetes and the package isn't included in _override: https://github.com/kubernetes/kubernetes/blob/master/Godeps/Godeps.json#L75

feiskyer: hmm, yep. Kubernetes doesn't upgrade to this version yet. And the fix should also be cherry-picked to old versions. Let me add it in _override instead.

feiskyer:
Moved the new changes to _override. @aleksandra-malinowska  PTAL

cluster-autoscaler/cloudprovider/azure/azure_container_service_pool.go
```diff
@@ -96,7 +104,7 @@ func (agentPool *ContainerServiceAgentPool) GetPool(agentProfiles *[]containerse
 		for _, value := range *agentProfiles {
```
kkmsft: Do we still need the check: "agentPool.serviceType == vmTypeACS" since we are already in the ACS path here ?

cluster-autoscaler/cloudprovider/azure/azure_container_service_pool.go
```diff
@@ -78,12 +72,26 @@ func NewContainerServiceAgentPool(spec *dynamic.NodeGroupSpec, am *AzureManager)
...
 }
 
-//GetPool is an internal function which figures out agentPoolProfile from the list based on the pool name provided in the --node parameter passed
+//GetAKSAgentPool is an internal function which figures out ManagedClusterAgentPoolProfile from the list based on the pool name provided in the --node parameter passed
+//to the autoscaler main
+func (agentPool *ContainerServiceAgentPool) GetAKSAgentPool(agentProfiles *[]containerservice.ManagedClusterAgentPoolProfile) (ret *containerservice.ManagedClusterAgentPoolProfile) {
+	for _, value := range *agentProfiles {
+		profileName := *value.Name
+		glog.V(5).Infof("AKS AgentPool profile name: %s", profileName)
+		if profileName == (agentPool.azureRef.Name) {
```
kkmsft: Is it possible to move this to a common method since this is similar in case of AKS and ACS  ?

feiskyer: Though they look same, the agentProfile is different struct for ACS and AKS. If we want to use the same logic, then we need a conversion between them, which is much more complicated.

cluster-autoscaler/cloudprovider/azure/azure_container_service_pool.go
feiskyer: hmm, yep, this is not required here

cluster-autoscaler/cloudprovider/azure/azure_container_service_pool.go
kkmsft: Ah.. yes.. I remember now ...:-)

cluster-autoscaler/cloudprovider/azure/azure_container_service_pool.go
feiskyer: Fixed now

feiskyer:
@kkmsft @aleksandra-malinowska PTAL

cluster-autoscaler/Godeps/Godeps.json
```diff
@@ -23,43 +23,43 @@
...
 		},
 		{
 			"ImportPath": "github.com/Azure/azure-sdk-for-go/services/containerregistry/mgmt/2017-10-01/containerregistry",
-			"Comment": "v14.6.0-1-g925867aa",
-			"Rev": "925867aa958c240707f9f937687cefc67f6cd86d"
+			"Comment": "v14.6.0-1-g50002f7f",
+			"Rev": "50002f7f8b030be047e67bbbd5108c5f5beed4ca"
 		},
 		{
 			"ImportPath": "github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2017-09-30/containerservice",
```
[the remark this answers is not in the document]
feiskyer: Let me have a check. Maybe I should change code first and then run godep save.

feiskyer:
Reworked vendor updates. Now there are many commit id changes in Godeps.json, while no real changes in vendors except azure packages. @aleksandra-malinowska Is this expected?

feiskyer:
ping @aleksandra-malinowska @MaciekPytel

PR #1072, opened 2018-07-19
Set RecommendationProvided if recommendation not empty.

bskiba:


bskiba:
@MaciekPytel @aleksandra-malinowska can one of you take a look? This is a minimal change.

vertical-pod-autoscaler/pkg/recommender/initialization/recommender.go
```diff
@@ -82,7 +82,10 @@ func (r *recommender) updateVPAs() {
 
 		}
 		vpa.Recommendation = &vpa_types.RecommendedPodResources{containerResources}
-		vpa.Conditions.Set(vpa_types.RecommendationProvided, true, "", "")
+		// Set RecommendationProvided if recommendation not empty.
+		if vpa.Recommendation.ContainerRecommendations != nil && len(vpa.Recommendation.ContainerRecommendations) > 0 {
```
aleksandra-malinowska: nit: nil check is unnecessary, len of nil slice will be 0:

```
The zero value of a slice is nil. The len and cap functions will both return 0 for a nil slice.
```

https://blog.golang.org/go-slices-usage-and-internals

bskiba: Done

Issue #863, opened 2018-05-16
Azure AKS Kubernetes error "runtime error: invalid memory address or nil pointer dereference"


dhirenrouterabbit:
Kuberentes version: 1.8.7
Cluster-autoscaler image version: v1.0.3

While autoscaling, cluster-autoscaler generates the following error

```
$ kubectl -n kube-system logs cluster-autoscaler-7458cd4887-gzlvh
.......
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x60 pc=0x2805da6]
goroutine 83 [running]:
k8s.io/autoscaler/cluster-autoscaler/core.(*StaticAutoscaler).RunOnce(0xc420b75b90, 0xed28dac2e, 0xe331d0368, 0x518f0a0, 0x0, 0x0)
	/gopath/src/k8s.io/autoscaler/cluster-autoscaler/core/static_autoscaler.go:99 +0x256
main.run(0xc420717bd0)
	/gopath/src/k8s.io/autoscaler/cluster-autoscaler/main.go:262 +0x46f
main.main.func2(0xc420010720)
	/gopath/src/k8s.io/autoscaler/cluster-autoscaler/main.go:344 +0x2a
created by k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/client-go/tools/leaderelection.(*LeaderElector).Run
	/gopath/src/k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/client-go/tools/leaderelection/leaderelection.go:145 +0x97
bash-4.3# kubectl -n kube-system logs cluster-autoscaler-7458cd4887-gzlvh
```

feiskyer:
Hi, @dhirenrouterabbit AKS support is not included in any autoscaler releases yet.  #861 is aiming to include it in v1.2.2.

feiskyer:
However, build from master works now, e.g.

```
cd cluster-autoscaler
REGISTRY=<your-registry> make dev-release
```

feiskyer:
> For which Kuberentes version builds from master works?

1.10+. It should also work for v1.9.X, but it's not recommended because autoscaler vendors kube-scheduler while there may be some conflicts with new versioned kube-scheduler.

feiskyer:
[...]

PR #1076, opened 2018-07-23
Fix potential panic

feiskyer:
err may be nil some cases, so err.Error() would panic CA.

Fix https://github.com/kubernetes/autoscaler/issues/863#issuecomment-406857408

### DIFF
--- a/cluster-autoscaler/utils/errors/errors.go
+++ b/cluster-autoscaler/utils/errors/errors.go
@@ -71,7 +71,7 @@ func ToAutoscalerError(defaultType AutoscalerErrorType, err error) AutoscalerErr
 	if e, ok := err.(AutoscalerError); ok {
 		return e
 	}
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/utils/errors/errors.go
```diff
@@ -71,7 +71,7 @@ func ToAutoscalerError(defaultType AutoscalerErrorType, err error) AutoscalerErr
 	if e, ok := err.(AutoscalerError); ok {
```
bskiba: Shouldn't we return nil from ToAutoscalerError if err is nil? It seems weird that we construct an AutoscalerError when there is actually no error.

feiskyer: It's still an error message, and even err is nill, the error type (defaultType) isn't. So we should still return an error here.

PR #1078, opened 2018-07-23
Fix error on node info not found for group

aleksandra-malinowska:
Instead of using unrelated (and always nil) error variable, create a new error from scratch. Fixes potential cases for #1076


feiskyer:
lgtm

PR #1079, opened 2018-07-23
Simplify building cloud provider

aleksandra-malinowska:
Pass the entire options object to build cloud provider.
Remove no longer necessary builder pattern for cloud provider (it's never re-constructed after initialization.)

### DIFF
--- a/cluster-autoscaler/cloudprovider/builder/cloud_provider_builder.go
+++ b/cluster-autoscaler/cloudprovider/builder/cloud_provider_builder.go
@@ -25,7 +25,7 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/azure"
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/gce"
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/kubemark"
-	"k8s.io/autoscaler/cluster-autoscaler/config/static"
+	"k8s.io/autoscaler/cluster-autoscaler/config"
 	"k8s.io/client-go/informers"
 	kubeclient "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -48,7 +48,7 @@ var AvailableCloudProviders = []string{
 const DefaultCloudProvider = gce.ProviderNameGCE
 
 // NewCloudProvider builds a cloud provider from provided parameters.
-func NewCloudProvider(opts static.AutoscalingOptions, do cloudprovider.NodeGroupDiscoveryOptions, rl *cloudprovider.ResourceLimiter) cloudprovider.CloudProvider {
+func NewCloudProvider(opts config.AutoscalingOptions, do cloudprovider.NodeGroupDiscoveryOptions, rl *cloudprovider.ResourceLimiter) cloudprovider.CloudProvider {
 	glog.V(1).Infof("Building %s cloud provider.", opts.CloudProviderName)
 	switch opts.CloudProviderName {
 	case gce.ProviderNameGCE:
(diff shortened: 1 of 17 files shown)


cluster-autoscaler/cloudprovider/builder/cloud_provider_builder.go
```diff
@@ -47,70 +47,49 @@ var AvailableCloudProviders = []string{
...
-
-// Build a cloud provider from static settings contained in the builder and dynamic settings passed via args
-func (b CloudProviderBuilder) Build(discoveryOpts cloudprovider.NodeGroupDiscoveryOptions, resourceLimiter *cloudprovider.ResourceLimiter) cloudprovider.CloudProvider {
-	glog.V(1).Infof("Building %s cloud provider.", b.cloudProviderFlag)
-	switch b.cloudProviderFlag {
+// NewCloudProvider builds a cloud provider from provided parameters.
+func NewCloudProvider(opts static.AutoscalingOptions, do cloudprovider.NodeGroupDiscoveryOptions, rl *cloudprovider.ResourceLimiter) cloudprovider.CloudProvider {
+	glog.V(1).Infof("Building %s cloud provider.", opts.CloudProviderName)
+	switch opts.CloudProviderName {
 	case gce.ProviderNameGCE:
```
MaciekPytel: I'd rather keep the log above. It's 1 line / CA instance and it can be useful to quickly figure out cloudprovider from logs.

cluster-autoscaler/config/static/autoscaling_options.go
```diff
@@ -0,0 +1,118 @@
+/*
```
MaciekPytel: 'static' part of the path feels sort of redundant. There isn't that much stuff in config/, not sure if it's worth creating package for this single file.

'config/dynamic' exists for historical reasons only, all it contains this days is a function for parsing a string, nothing I would call "dynamic config"

cluster-autoscaler/config/static/autoscaling_options.go
```diff
@@ -0,0 +1,118 @@
...
+	// Type of the GPU (e.g. nvidia-tesla-k80)
+	GpuType string
+	// Lower bound on number of GPUs of given type in cluster
+	Min int64
+	// Upper bound on number of GPUs of given type in cluster
+	Max int64
+}
+
+// AutoscalingOptions contain various options to customize how autoscaling works
+type AutoscalingOptions struct {
```
MaciekPytel: Github diff sucks, but I assume this is just copy-paste from context/autoscaling_context.go?

cluster-autoscaler/main_test.go
```diff
@@ -17,24 +17,25 @@ limitations under the License.
 package main
 
 import (
-	"github.com/stretchr/testify/assert"
-	"k8s.io/autoscaler/cluster-autoscaler/context"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"k8s.io/autoscaler/cluster-autoscaler/config/static"
```
MaciekPytel: If you're already re-organizing imports with the convention we use it should be:
```
"testing" 

"k8s.io/autoscaler/cluster-autoscaler/config/static"

```
Admittedly it looks kinda strange with just one import in each group :)

MaciekPytel:
Few nits, overall looks good. Thanks for killing dead code.

cluster-autoscaler/cloudprovider/builder/cloud_provider_builder.go
aleksandra-malinowska: Done

cluster-autoscaler/config/static/autoscaling_options.go
aleksandra-malinowska: Yes

cluster-autoscaler/main_test.go
aleksandra-malinowska: Done

cluster-autoscaler/config/static/autoscaling_options.go
aleksandra-malinowska: Done

PR #1083, opened 2018-07-23
Do not garbage collect newly created ContainerStates.

bskiba:
Do not garbage collect newly created ContainerStates (LastSampleStart.IsZero() == true). 

Also fix typo.


bskiba:
@schylek

PR #1087, opened 2018-07-25
Move creating cloud provider out of context

aleksandra-malinowska:
Move creating cloud provider out of context.
Move creating helper structs with dynamic config into cloud provider constructor.

### DIFF
--- a/cluster-autoscaler/cloudprovider/builder/cloud_provider_builder.go
+++ b/cluster-autoscaler/cloudprovider/builder/cloud_provider_builder.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/gce"
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/kubemark"
 	"k8s.io/autoscaler/cluster-autoscaler/config"
+	"k8s.io/autoscaler/cluster-autoscaler/context"
 	"k8s.io/client-go/informers"
 	kubeclient "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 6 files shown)


cluster-autoscaler/cloudprovider/builder/cloud_provider_builder.go
```diff
@@ -48,8 +49,15 @@ var AvailableCloudProviders = []string{
 const DefaultCloudProvider = gce.ProviderNameGCE
 
 // NewCloudProvider builds a cloud provider from provided parameters.
-func NewCloudProvider(opts config.AutoscalingOptions, do cloudprovider.NodeGroupDiscoveryOptions, rl *cloudprovider.ResourceLimiter) cloudprovider.CloudProvider {
+func NewCloudProvider(opts config.AutoscalingOptions) cloudprovider.CloudProvider {
```
MaciekPytel: I wonder if it wouldn't be preferable to keep explicitly passing ResourceLimiter? I know we construct it from options anyway, but it doesn't feel like building it should be the responsibility of NewCloudProvider. It would make it easier if we ever need to customize RL and also marginally cleaner to mock. WDYT?

aleksandra-malinowska: Not feeling particularly strongly about it, but to me initializing ResourceLimiter here seems cleaner, since it's owned by CloudProvider object anyway. As for customizing and/or mocking, it isn't currently hidden behind any interface, so there's no way to substitute a custom struct/mock in a transparent way anyway.

MaciekPytel: Fair point about mocking, I guess the difference in our opinions is because I'm still not thinking about RL as a part of CP, more like something separate we put in CP as a hack. Since we're not going to refactor RL again I probably should just get used to it.
I'm fine with leaving it as is.

PR #1088, opened 2018-07-25
Fix node memory prediction in GCP cloudprovider

MaciekPytel:


MaciekPytel:
Currently we're overestimating node memory capacity and allocatable by ~4.9%, so in a rare scenario we can scale-up a node that can't fit the pending pod. We'll need to cherry-pick this to all branches I think.

mwielgus:
seriously?

MaciekPytel:
Added one more commit to account for kubelet evictionHard config (which goes into allocatable calculation on top of kube-reserved). @bskiba @aleksandra-malinowska can you review?

@mwielgus Does your comment refer to the error itself or my fix? I hope it's the former :) Going through all those MB / MiB conversions is a rather silly looking problem (and, as I just learned, it's very painful to implement).

cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -87,6 +92,9 @@ func (t *templateBuilder) buildAllocatableFromKubeEnv(capacity apiv1.ResourceLis
 	if err != nil {
 		return nil, err
 	}
+	if quantity, found := reserved[apiv1.ResourceMemory]; found {
+		reserved[apiv1.ResourceMemory] = *resource.NewQuantity(quantity.Value()+kubeletEvictionHardMemory, resource.BinarySI)
```
bskiba: quantity.Add() 
https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go#L531

PR #1090, opened 2018-07-26
Use proxy for container state aggregator.

bskiba:
When a ContainerAggregatedState of a live container is garbage collected due to lack of samples, is will be recreated once the samples start appearing again. This prevents us from not providing recommendations for containers that are live but didn't get samples for 8 days.

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
+++ b/vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
@@ -104,6 +104,46 @@ func TestClusterGCAggregateContainerStateLeavesValid(t *testing.T) {
 	assert.NotEmpty(t, vpa.aggregateContainerStates)
 }
 
+func TestAddSampleAfterAggregateContainerStateGCed(t *testing.T) {
+	// Create a pod with a single container.
+	cluster := NewClusterState()
+	vpa := addTestVpa(cluster)
+	pod := addTestPod(cluster)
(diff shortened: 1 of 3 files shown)


bskiba:
@schylek

schylek:
The Travis CI build failed

bskiba:
Fixed now.

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
```diff
@@ -166,7 +208,7 @@ func TestAddVpaThenAddPod(t *testing.T) {
 	addTestPod(cluster)
 	container := addTestContainer(cluster)
 	aggregateStateKey := cluster.aggregateStateKeyForContainerID(testContainerID)
-	assert.True(t, container.aggregator == vpa.aggregateContainerStates[aggregateStateKey])
+	assert.True(t, container.aggregator.Get() == vpa.aggregateContainerStates[aggregateStateKey])
```
schylek: IMO this Get metod should be deleted:
- it is not obvious what it returns and why it should be part of the interface
- it is used only in tests
I recommend testing with Contains/NotContains(t, vpa.aggregateContainerStates, aggregateStateKey)

bskiba: You're right, I was just being lazy. Removed.

PR #1092, opened 2018-07-26
Group all Kubernetes API clients

aleksandra-malinowska:
Move all clients used for communicating with Kubernetes API server to a separate struct within context. This doesn't affect current usage as the struct is embedded, the only benefit is cleaner structs and logically divided responsibility. Before, most clients lived in AutoscalingContext, except for ListerRegistry, which was embedded directly in StaticAutoscaler.

Delta except for tests is 32 lines. The rest is reducing test setup verbosity.


feiskyer:
Much cleaner now. LGTM

MaciekPytel:
Agreed, it makes the test setup significantly cleaner. Thanks!
/lgtm

PR #1101, opened 2018-07-31
Initialization package renamed to routines.

schylek:


bskiba:
/lgtm
Not too sure about "routines", but it's definitely closer to the truth than initialization :) Cluster autoscaler has "core", but I guess it is a similar level of vagueness.

PR #1102, opened 2018-07-31
Move initializing defaults out of main

aleksandra-malinowska:
Move initializing defaults out of main. Extract building autoscaler from run(). Extract constructing auxiliary kube clients to a separate function.

### DIFF
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -37,7 +37,6 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/estimator"
 	"k8s.io/autoscaler/cluster-autoscaler/expander"
 	"k8s.io/autoscaler/cluster-autoscaler/metrics"
-	"k8s.io/autoscaler/cluster-autoscaler/simulator"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
 	kube_util "k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/main.go
```diff
@@ -250,38 +249,39 @@ func registerSignalHandlers(autoscaler core.Autoscaler) {
...
+	// This metric should be published only once.
+	metrics.UpdateNapEnabled(autoscalingOptions.NodeAutoprovisioningEnabled)
+
+	// Create autoscaler.
+	return core.NewAutoscaler(opts)
+}
+
+func run(healthCheck *metrics.HealthCheck) {
+	// Register metrics.
+	metrics.RegisterAll()
```
MaciekPytel: nit: I'm all for commenting stuff, but this may be getting too far :)

cluster-autoscaler/main.go
```diff
@@ -250,38 +249,39 @@ func registerSignalHandlers(autoscaler core.Autoscaler) {
...
 
-	autoscaler, err := core.NewAutoscaler(opts, cloudProvider)
+	// Create autoscaler.
+	autoscaler, err := buildAutoscaler()
 	if err != nil {
 		glog.Fatalf("Failed to create autoscaler: %v", err)
 	}
+
+	// Register signal handlers for graceful shutdown.
```
MaciekPytel: This comment I like, because there is an additional information on why we need signal handlers.

MaciekPytel:
This seems much more logical than what we're doing now, nice work. 
/lgtm

Issue #1103, opened 2018-07-31
VPA replaces annotations of pods, in stead of updating. This breaks PSP.


wmuizelaar:
Hi,

Testing with the vertical-pod-autoscaler, we figured out that the annotations on a pod are completely replaced by the admission controller, in stead of merged/added.

Used version of the containers:
```
k8s.gcr.io/vpa-admission-controller:0.2.0
k8s.gcr.io/vpa-recommender:0.2.0
k8s.gcr.io/vpa-updater:0.2.0
```

What I see is that containers that are run from a deployment, and I set a custom annotation in my specfile, these will be overwritten by the vpa-admission-controller:

Before:
```
kubectl describe pod nginx-deployment-5d78d59c86-dxg7h
Name:           nginx-deployment-5d78d59c86-dxg7h
Namespace:      default
Node:           gke-bolcom-sbx-4wu-internal-4e3396ab-b2lj/10.16.4.18
Start Time:     Tue, 31 Jul 2018 13:37:25 +0200
Labels:         app=nginx
                pod-template-hash=1834815742
Annotations:    bol.io/test=test
                kubernetes.io/limit-ranger=LimitRanger plugin set: cpu request for container nginx
```

After:
```
kubectl describe pod nginx-deployment-5d78d59c86-lr2wv
Name:           nginx-deployment-5d78d59c86-lr2wv
Namespace:      default
Node:           <none>
Labels:         app=nginx
                pod-template-hash=1834815742
Annotations:    vpaUpdates=Pod resources updated by nginx-vpa: container 0: cpu request, memory request, memory limit
```

When using Pod Security Policies in a cluster, this will also set annotations, and since those are overwritten by the admission controller, those pods won't get scheduled again, since there cannot be found a valid podsecuritypolicy.

bskiba:
Hm, can't seem to reproduce this... 
I've added annotations to the hamster example for the repo, like this: https://gist.github.com/bskiba/1aad99ecb52e75f744b3f88c21e1ea3f

And here is what I see:
```
kubectl describe pod
Name:           hamster-7cb9f4785d-ww8gh
Namespace:      default
Start Time:     Tue, 31 Jul 2018 20:24:13 +0200
Labels:         app=hamster
                pod-template-hash=3765903418
Annotations:    vpa-test=test
                vpaUpdates=Pod resources updated by hamster-vpa: container 0: cpu request, memory request, memory limit
```
[...]

PR #1104, opened 2018-07-31
VPA: don't overwrite existing annotations when patching pod

maartenvandenbogaard:
Fixes: https://github.com/kubernetes/autoscaler/issues/1103

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
@@ -121,17 +121,17 @@ func (s *AdmissionServer) getPatchesForPodResourceRequest(raw []byte, namespace 
 		updatesAnnotation = append(updatesAnnotation, fmt.Sprintf("container %d: ", i)+strings.Join(annotations, ", "))
 	}
 	if len(updatesAnnotation) > 0 {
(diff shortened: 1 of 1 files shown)


vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
```diff
@@ -121,11 +121,18 @@ func (s *AdmissionServer) getPatchesForPodResourceRequest(raw []byte, namespace
...
-			Op:   "add",
-			Path: "/metadata/annotations",
-			Value: map[string]string{
-				"vpaUpdates": fmt.Sprintf("Pod resources updated by %s: ", vpaName) + strings.Join(updatesAnnotation, "; ")}})
+		if pod.Annotations == nil {
+			patches = append(patches, patchRecord{
+				Op:   "add",
+				Path: "/metadata/annotations",
+				Value: map[string]string{
+					"vpaUpdates": fmt.Sprintf("Pod resources updated by %s: ", vpaName) + strings.Join(updatesAnnotation, "; ")}})
```
bskiba: Can you construct the annotation value string in one place? can be a variable or function.

bskiba:
Just one small comment.

PR #1107, opened 2018-08-02
Don't use typed errors during initialization

aleksandra-malinowska:
We don't check or report errors at this stage, just fail.

### DIFF
--- a/cluster-autoscaler/core/autoscaler.go
+++ b/cluster-autoscaler/core/autoscaler.go
@@ -52,10 +52,10 @@ type Autoscaler interface {
 }
 
 // NewAutoscaler creates an autoscaler of an appropriate type according to the parameters
-func NewAutoscaler(opts AutoscalerOptions) (Autoscaler, errors.AutoscalerError) {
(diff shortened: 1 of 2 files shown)


MaciekPytel:
I'm not sure about this one. Is there any downside to using AutoscalerErrors?

aleksandra-malinowska:
Just aesthetics/cleanup. Using autoscaler internal error type here doesn't seem right since it really means "can't even create the autoscaler". Also, we never publish a metric for it at this stage, unlike for typed errors returned during autoscaler operations (side question: do we even have any sort of "CA is crashlooping signal?")

Apart from this, it really doesn't matter much, I guess.

MaciekPytel:
I'd lean towards 'use AutoscalerErrors everywhere we can without running into too much problems with external dependencies'. 

To answer your question: if CA is crashlooping at this stage it doesn't have any way to expose it I think (we don't have metrics endpoint if the process is crashing immediately). It should be up to cluster monitoring to notice crashlooping system pod and react.

PR #1108, opened 2018-08-02
Capping to limit or min/max allowed leaves annotation

schylek:


vertical-pod-autoscaler/pkg/utils/vpa/capping.go
```diff
@@ -30,88 +30,127 @@ func NewCappingRecommendationProcessor() RecommendationProcessor {
 	return &cappingRecommendationProcessor{}
 }
 
+type cappingAction = string
+
+var (
+	cappedToMinAllowed cappingAction = "capped to min allowed"
```
bskiba: capped to min allowed from VPA resourcePolicy? or VPA containerPolicy
min allowed seems a bit too generic

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
```diff
@@ -30,88 +30,127 @@ func NewCappingRecommendationProcessor() RecommendationProcessor {
...
+	cappedToMaxAllowed cappingAction = "capped to max allowed"
+	cappedToLimit      cappingAction = "capped to limit"
```
bskiba: capped to request limit? Not very sure though, limit might be specific enough

bskiba:
Do you have an example of what the annotation looks like? I wonder if we're not putting too much info in one place.

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
bskiba: I was wondering if I see minAllowed I'll know immediately that this is the minAllowed from the VPA object. But I might be paranoid.

schylek:
Annotations:    vpaUpdates=Pod resources updated by hamster-vpa:
container 0: cpu capped to max allowed, memory capped to min allowed,
cpu request, memory request

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
schylek: Changed to "capped to container limit".

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
schylek: ok, changed minAllowed/maxAllowed

PR #1109, opened 2018-08-02
Clean up estimators

aleksandra-malinowska:
Hide estimators behind interface. Check flag only once, during initialization. Deprecate BasicEstimator.

### DIFF
--- a/cluster-autoscaler/context/autoscaling_context.go
+++ b/cluster-autoscaler/context/autoscaling_context.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
 	"k8s.io/autoscaler/cluster-autoscaler/clusterstate/utils"
 	"k8s.io/autoscaler/cluster-autoscaler/config"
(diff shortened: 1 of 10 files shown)


cluster-autoscaler/context/autoscaling_context.go
```diff
@@ -42,6 +43,8 @@ type AutoscalingContext struct {
 	PredicateChecker *simulator.PredicateChecker
 	// ExpanderStrategy is the strategy used to choose which node group to expand when scaling up
 	ExpanderStrategy expander.Strategy
+	// EstimatorBuilder is the builder function for node count estimator to be used.
+	EstimatorBuilder estimator.EstimatorBuilder
```
aleksandra-malinowska: The reason this is a builder and not the estimator is because BasicEstimator has internal state. Re-using the same object rather than constructing a new one in each iteration (which is how it currently works) would go beyond refactoring IMO.

losipiuk:
@aleksandra-malinowska is this one still relevant or should we close it?

PR #1110, opened 2018-08-03
Extract autoscaling GCE client

aleksandra-malinowska:
Extract GCE calls and hide them behind a new interface. This will allow for easier mocking (and getting rid of jsons from gke_manager_test.go) in the near future.

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
+++ b/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
@@ -43,7 +43,7 @@ type AutoscalingGceClient interface {
 	FetchMigBasename(GceRef) (string, error)
 	FetchMigInstances(GceRef) ([]GceRef, error)
 	FetchMigTemplate(GceRef) (*gce.InstanceTemplate, error)
-	FetchMigs(zone string, filter *regexp.Regexp) ([]string, error)
+	FetchMigsWithName(zone string, filter *regexp.Regexp) ([]string, error)
 	FetchZones(region string) ([]string, error)
 
 	// modifying resources
@@ -177,14 +177,14 @@ func (client *autoscalingGceClientV1) FetchMigTemplate(migRef GceRef) (*gce.Inst
 	return client.gceService.InstanceTemplates.Get(migRef.Project, templateName).Do()
 }
 
-func (client *autoscalingGceClientV1) FetchMigs(zone string, name *regexp.Regexp) ([]string, error) {
+func (client *autoscalingGceClientV1) FetchMigsWithName(zone string, name *regexp.Regexp) ([]string, error) {
 	filter := fmt.Sprintf("name eq %s", name)
 	links := make([]string, 0)
 	req := client.gceService.InstanceGroups.List(client.projectId, zone).Filter(filter)
 	if err := req.Pages(context.TODO(), func(page *gce.InstanceGroupList) error {
 		for _, ig := range page.Items {
 			links = append(links, ig.SelfLink)
-			glog.V(3).Infof("autodiscovered managed instance group %s using regexp %s", ig.Name, name)
+			glog.V(3).Infof("found managed instance group %s matching regexp %s", ig.Name, name)
 		}
 		return nil
 	}); err != nil {
--- a/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
+++ b/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
@@ -30,8 +30,8 @@ import (
 )
 
 const (
-	operationWaitTimeout  = 5 * time.Second
-	operationPollInterval = 100 * time.Millisecond
+	defaultOperationWaitTimeout  = 5 * time.Second
+	defaultOperationPollInterval = 100 * time.Millisecond
 )
 
 // AutoscalingGceClient is used for communicating with GCE API.
@@ -55,6 +55,10 @@ type autoscalingGceClientV1 struct {
 	gceService *gce.Service
 
 	projectId string
+
+	// These can be overridden, e.g. for testing.
+	operationWaitTimeout  time.Duration
+	operationPollInterval time.Duration
 }
 
 // NewAutoscalingGceClientV1 creates a new client for communicating with GCE v1 API.
@@ -65,8 +69,10 @@ func NewAutoscalingGceClientV1(client *http.Client, projectId string) (*autoscal
 	}
 
 	return &autoscalingGceClientV1{
-		projectId:  projectId,
-		gceService: gceService,
+		projectId:             projectId,
+		gceService:            gceService,
+		operationWaitTimeout:  defaultOperationWaitTimeout,
+		operationPollInterval: defaultOperationPollInterval,
 	}, nil
 }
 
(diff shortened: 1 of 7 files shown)


cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	templateUrl, err := url.Parse(igm.InstanceTemplate)
+	if err != nil {
+		return nil, err
+	}
+	_, templateName := path.Split(templateUrl.EscapedPath())
+	return client.gceService.InstanceTemplates.Get(migRef.Project, templateName).Do()
+}
+
+func (client *autoscalingGceClientV1) FetchMigs(zone string, name *regexp.Regexp) ([]string, error) {
```
MaciekPytel: Can we rename it to make it more clear what the method is doing? It was not at all obvious to me this lists MIGs with name matching regexp.

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -0,0 +1,194 @@
...
+	filter := fmt.Sprintf("name eq %s", name)
+	links := make([]string, 0)
+	req := client.gceService.InstanceGroups.List(client.projectId, zone).Filter(filter)
+	if err := req.Pages(context.TODO(), func(page *gce.InstanceGroupList) error {
+		for _, ig := range page.Items {
+			links = append(links, ig.SelfLink)
+			glog.V(3).Infof("autodiscovered managed instance group %s using regexp %s", ig.Name, name)
```
MaciekPytel: This is only used for autodiscovery, right? Anyway, if this is just a wrapper on GCE client it shouldn't really know about autodiscovery. Can we just change the log to "found managed instance group %s matching regexp %s"?

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client_test.go
```diff
@@ -0,0 +1,52 @@
...
+func TestWaitForOp(t *testing.T) {
+	server := test_util.NewHttpServerMock()
+	defer server.Close()
+	g := newTestAutoscalingGceClient(t, "project1", server.URL)
+	server.On("handle", "/project1/zones/us-central1-b/operations/operation-1505728466148-d16f5197").Return(operationRunningResponse).Times(3)
+	server.On("handle", "/project1/zones/us-central1-b/operations/operation-1505728466148-d16f5197").Return(operationDoneResponse).Once()
+
+	operation := &gce_api.Operation{Name: "operation-1505728466148-d16f5197"}
+
+	err := g.waitForOp(operation, projectId, zoneB)
```
MaciekPytel: I think unittests tests doing sleeps inside are a big anti-pattern. If UT takes more than a few seconds to run people just stop running them. I know in this case it's just 300ms, but I'd rather avoid it on principle (also because we may have more tests calling waitForOp internally in future).

Can we just make operationPollInterval a variable on autoscalingGceClientV1 (defaulting to current 100ms) and override it in tests?

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -1205,37 +1157,12 @@ func (m *gceManagerImpl) findMigsInRegion(region string, name *regexp.Regexp) ([
...
-	if err != nil {
-		return nil, err
-	}
-	templateUrl, err := url.Parse(igm.InstanceTemplate)
-	if err != nil {
-		return nil, err
-	}
-	_, templateName := path.Split(templateUrl.EscapedPath())
-	instanceTemplate, err := m.gceService.InstanceTemplates.Get(mig.Project, templateName).Do()
+	template, err := m.GceService.FetchMigTemplate(mig.GceRef)
```
MaciekPytel: Do we need this method anymore? All it does is proxy to FetchMigTemplate, which is available everywhere where we can call this. Maybe we should just update the places calling it and scrap this method? Same for getZones.

cluster-autoscaler/cloudprovider/gce/gce_url.go
```diff
@@ -35,20 +35,41 @@ func ParseMigUrl(url string) (project string, zone string, name string, err erro
 	return parseGceUrl(url, "instanceGroups")
 }
 
+// ParseIgmUrl expects url in format:
+// https://content.googleapis.com/compute/v1/projects/<project-id>/zones/<zone>/instanceGroupManagers/<name>
+func ParseIgmUrl(url string) (project string, zone string, name string, err error) {
```
MaciekPytel: Do we call this anywhere?

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client_test.go
MaciekPytel: I just realized this is just copy-pasted. Still, I think applying my comment would be a nice improvement.

MaciekPytel:
I left a few minor comments, but overall it looks like a very nice refactor and a good step in the direction of cleaning up the cloudprovider mess :+1:.

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -0,0 +1,194 @@
...
+	gce "google.golang.org/api/compute/v1"
+)
+
+const (
+	operationWaitTimeout  = 5 * time.Second
+	operationPollInterval = 100 * time.Millisecond
+)
+
+// AutoscalingGceClient is used for communicating with GCE API.
+type AutoscalingGceClient interface {
```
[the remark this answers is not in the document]
aleksandra-malinowska: Yes to everything except changing anything for GKE client. Next PR will extract it in a similar way (with two implementations, v1 and v1beta1). 

In particular, I want to add fake implementation for easier testing (no need to set up HTTP server returning fake jsons just to test GCE manager). But we may also need to use v1beta1 for suspend/resume at first.

cluster-autoscaler/cloudprovider/gce/gce_url.go
aleksandra-malinowska: Now we do (3 times in `gce_manager.go`). Thanks for catching this!

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
aleksandra-malinowska: Done

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
aleksandra-malinowska: Done. Not sure if this is clear now, let me know if you have a better name in mind

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client_test.go
aleksandra-malinowska: Done, and added another tests case (timeout). I hope to move from GceManager tests that require waiting for operation each time to ones using fake client, but for now changing this brought down time needed for running unit tests by about 0.5s:

Before:
$ go clean -testcache && go test ./...
ok  	k8s.io/autoscaler/cluster-autoscaler/cloudprovider/gce	1.260s

After:
$ go clean -testcache && go test ./...
ok  	k8s.io/autoscaler/cluster-autoscaler/cloudprovider/gce	0.671s

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -55,6 +55,10 @@ type autoscalingGceClientV1 struct {
 	gceService *gce.Service
 
 	projectId string
+
+	// These can be overridden, e.g. for testing.
```
MaciekPytel: They can't really be overridden for anything else given they're private fields in a struct :P

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client_test.go
```diff
@@ -41,6 +42,10 @@ func TestWaitForOp(t *testing.T) {
 	server := test_util.NewHttpServerMock()
 	defer server.Close()
 	g := newTestAutoscalingGceClient(t, "project1", server.URL)
+
+	g.operationPollInterval = 5 * time.Millisecond
```
MaciekPytel: Is there any reason for this not to be 0?

aleksandra-malinowska: Done

PR #1113, opened 2018-08-06
Update container-gen.go

wangxy518:
update some urls.


MaciekPytel:
This is imported code (our _override is similar in nature to vendor, we use it to hack around some diamond dependency problems). You should report this to the original repo, so they fix it upstream. Any fix here will be lost as soon as we need to update to a newer version.

PR #1114, opened 2018-08-06
Add AutoscalingStatusProcessor

towca:


MaciekPytel:
You need to figure out CLA before we can accept any contribution. Also let's leave this one as is, but for future reference - it looks like this contains 3 separate and unrelated changes (adding processor, changing gcp provider and adding method to clusterstate). It really should be 3 separate PRs.

towca:
@kgolab @krzysztof-jastrzebski please review

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -200,6 +200,12 @@ func (gce *GceCloudProvider) Refresh() error {
 	return gce.gceManager.Refresh()
 }
 
+// GetProjectInfo returns the project id, location and cluster name.
+func (gce *GceCloudProvider) GetProjectInfo() (projectId, location, clusterName string) {
```
krzysztof-jastrzebski: How are you going to use this function as it is not part of CloudProvider interface?

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -106,6 +106,15 @@ func (a *StaticAutoscaler) cleanUpIfRequired() {
 
 // RunOnce iterates over node groups and scales them up/down if necessary
 func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError {
+	if a.processors != nil && a.processors.AutoscalingStatusProcessor != nil {
```
krzysztof-jastrzebski: Can processors and a.processors.AutoscalingStatusProcessor be nil? I other executions we don't check that.

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
krzysztof-jastrzebski: This is rather ClusterInfo.

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -200,6 +200,12 @@ func (gce *GceCloudProvider) Refresh() error {
...
+	manager := gce.gceManager
```
krzysztof-jastrzebski: You don't need to create local variable.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -893,3 +893,13 @@ func getNotRegisteredNodes(allNodes []*apiv1.Node, cloudProvider cloudprovider.C
 	}
 	return notRegistered, nil
 }
+
+// GetClusterSize calculates and returns cluster's current size and target size. The current size is the
+// actual number of nodes registered in Kubernetes, the target size is the number of nodes the CA wants.
+func (csr *ClusterStateRegistry) GetClusterSize() (currentSize, targetSize int) {
+	for _, accRange := range csr.acceptableRanges {
```
krzysztof-jastrzebski: Is it thread safe?

cluster-autoscaler/processors/processors.go
```diff
@@ -51,7 +55,42 @@ func TestProcessors() *AutoscalingProcessors {
...
-		NodeGroupManager:       nodegroups.NewDefaultNodeGroupManager(),
+		ScaleUpStatusProcessor:     &status.EventingScaleUpStatusProcessor{},
+		AutoscalingStatusProcessor: &status.NoOpAutoscalingStatusProcessor{},
+		NodeGroupManager:           nodegroups.NewDefaultNodeGroupManager(),
+	}
+}
+
+// CleanUp cleans up the processors' internal structures.
+func (ap *AutoscalingProcessors) CleanUp() {
+	if ap.PodListProcessor != nil {
```
krzysztof-jastrzebski: Do we need to check if it is nil? Same below.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -344,8 +353,12 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 	return nil
 }
 
-// ExitCleanUp removes status configmap.
+// ExitCleanUp performs all necessary clean-ups when the autoscaler's exiting.
 func (a *StaticAutoscaler) ExitCleanUp() {
+	if a.processors != nil {
```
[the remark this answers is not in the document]
towca: I think it's currently checked everywhere `a.processors` is used.

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
MaciekPytel: This should not be added to cloudprovider interface. If you need such method in your extension you will need to use type assertion.

MaciekPytel: +1

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
MaciekPytel: I don't think it hurts to create it and have a shorter return line. I'm fine either way.

towca:
@krzysztof-jastrzebski @MaciekPytel I fixed the issues you mentioned. I'm not familiar with github CR workflow, should i amend and force push my commit as in Gitlab or gerrit for example, or push new commits with the fixes? Because your comments can't be accessed in the code now.

cluster-autoscaler/clusterstate/clusterstate.go
MaciekPytel: Let's add locking here.

BTW you shouldn't really trust in thread safety of clusterstate. With exception of RegisterScaleDown nothing is ever called from a different goroutine and nothing is really tested for thread safety. Actually, we should probably revisit this at some point. I'll open an issue to do so.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -106,6 +106,13 @@ func (a *StaticAutoscaler) cleanUpIfRequired() {
 
 // RunOnce iterates over node groups and scales them up/down if necessary
 func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError {
+	defer func() {
+		err := a.processors.AutoscalingStatusProcessor.Process(a.AutoscalingContext, a.clusterStateRegistry, time.Now())
+		if err != nil {
+			glog.Warningf("AutoscalingStatusProcessor error: %v.", err)
```
MaciekPytel: Let's make it Errorf, seems appropriate for an error.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -106,6 +106,13 @@ func (a *StaticAutoscaler) cleanUpIfRequired() {
 
 // RunOnce iterates over node groups and scales them up/down if necessary
 func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError {
+	defer func() {
```
MaciekPytel: This uses CSR, shouldn't we move it after updateClusterState, similar to when we defer writing status configmap? Otherwise you can't really assume anything you get from CSR is true.

cluster-autoscaler/processors/nodegroups/autoprovisioning_nodegroup_manager.go
```diff
@@ -95,3 +95,8 @@ func (p *AutoprovisioningNodeGroupManager) RemoveUnneededNodeGroups(context *con
 	}
 	return nil
 }
+
+// CleanUp cleans up the processor's internal structures.
+func (p *AutoprovisioningNodeGroupManager) CleanUp() error {
```
MaciekPytel: Is there even a point in this (and other CleanUp methods) returning any error? It's not like we can do anything about errors at this stage. And you don't even log them, so it's very easy for someone to implement something that returns an error which won't even be visible in logs.

towca: There can be pending tasks on a processor that are flushed in a CleanUp, I think it can be meaningful to be notified about dropping any of them. Also, external clients executing tasks asynchronously may not even notify about any errors until they are closed, which would happen in CleanUp.

MaciekPytel: I fully agree that it may be useful to log some failure in CleanUp. The question is why not leave it to CleanUp implementation to log whatever it thinks is worth logging? Not allowing error makes it more explicit that it's up to the author of specific processor implementation to log if necessary and gives them more control over exactly what is logged, at what level, etc.

towca: Fair enough :)

towca: @MaciekPytel fixed, everything should be ok now

MaciekPytel:
/lgtm
I'll merge it after travis passes.

PR #1115, opened 2018-08-06
Add metrics to VPA

kgolab:
Added some basic metrics to all 3 VPA components.

@bskiba , please have a look.

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/poc.autoscaling.k8s.io/v1alpha1"
+	metrics_admission "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/metrics/admission"
 	vpa_api_util "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/vpa"
 )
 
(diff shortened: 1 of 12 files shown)


vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
```diff
@@ -135,51 +136,66 @@ func getPatchesForVPADefaults(raw []byte) ([]patchRecord, error) {
...
+	if resource == metrics_admission.Pod {
+		metrics_admission.OnAdmittedPod(status == metrics_admission.Applied)
+	}
+
+	return &response, status, resource
 }
 
 // Serve is a handler function of AdmissionServer
 func (s *AdmissionServer) Serve(w http.ResponseWriter, r *http.Request) {
+	t := metrics_admission.NewAdmissionLatency()
```
bskiba: nit: t seems a bit not descriptive

vertical-pod-autoscaler/pkg/recommender/routines/recommender.go
```diff
@@ -81,11 +83,16 @@ func (r *recommender) updateVPAs() {
 			})
 
 		}
+		had := vpa.HasRecommendation()
 		vpa.Recommendation = &vpa_types.RecommendedPodResources{containerResources}
 		// Set RecommendationProvided if recommendation not empty.
 		if len(vpa.Recommendation.ContainerRecommendations) > 0 {
 			vpa.Conditions.Set(vpa_types.RecommendationProvided, true, "", "")
+			if !had {
+				metrics_recommender.ObserveRecommendationLatency(vpa.Created)
```
bskiba: I am not super sure about this metric. If there is no recommendation ever provided for a vpa we will not know from this metric, right? What we really need is information - how long has this vpa gone without a recommendation (and does it already have it or is it still waiting). 
Since admission controller only sees the vpa again once it is modified, maybe this should be moved into the recommender, as it looks at all vpas every loop? 
Open to discussion, I'd like this metric to give some meaningful signal on how well the system is performing.

vertical-pod-autoscaler/pkg/utils/metrics/admission/admission.go
```diff
@@ -0,0 +1,104 @@
...
+	Error AdmissionStatus = "error"
+	// Skipped denotes an Admission Control execution w/o applying a recommendation
+	Skipped AdmissionStatus = "skipped"
+	// Applied denotes an Admission Control execution when a recommendation was applied
+	Applied AdmissionStatus = "applied"
+)
+
+const (
+	// Unknown means that the resource was not yet determined
+	Unknown AdmissionResource = "n/a"
```
bskiba: I'd go for "unknown". And I don't agree with the comment, it seems from it that the resource will be determined in the future, but this is rather - we were not able to determine the resource (e.g. unmarshalling error)

bskiba:
Some comments. On the whole looks good. 
Seems there's a merge conflict.

vertical-pod-autoscaler/pkg/recommender/routines/recommender.go
kgolab: This metric is there to only measure time it took to provide the first recommendation.
As for tracing the number of VPAs w/o recommendation one could use vpa_objects_count.

vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
kgolab: Fixed.

vertical-pod-autoscaler/pkg/utils/metrics/admission/admission.go
kgolab: Thanks, fixed.

kgolab:
Resolved the conflict in imports and fixed the nits. Please have another look.

bskiba:
Looks good. Would you mind squashing commits?

kgolab:
Squashed.

PR #1119, opened 2018-08-07
Update external_docs.go

wangxy518:
update a url


MaciekPytel:
Files under vendor/ directory are just how go manages external dependencies. It's just a copy from external repository and it's not meant to be modified here. You should open a PR against the upstream project (go-openapi).

PR #1120, opened 2018-08-07
Update CHANGELOG.md

wangxy518:
"https://github.com/magiconair/properties/tags/v1.7.0"   is an unvalid url.
So I changed it .


MaciekPytel:
Files under vendor/ directory are just how go manages external dependencies. It's just a copy from external repository and it's not meant to be modified here. You should open a PR against the upstream project (magiconair).

PR #1123, opened 2018-08-07
Remove faulty validation for VPA

bskiba:
The CRD creation doesn't work at the moment. Validation fails for enums and lists. This should fix e2e tests for now. 
Will add validation to admission webhook in a followup pr.

### DIFF
--- a/vertical-pod-autoscaler/deploy/vpa-crd.yaml
+++ b/vertical-pod-autoscaler/deploy/vpa-crd.yaml
@@ -26,31 +26,10 @@ spec:
               properties:
                 updateMode:
                   type: string
-                  enum:
-                    - "Off"
-                    - "Initial"
-                    - "Recreate"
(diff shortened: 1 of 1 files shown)


bskiba:
@schylek

vertical-pod-autoscaler/deploy/vpa-crd.yaml
```diff
@@ -26,31 +26,11 @@ spec:
...
-                  enum:
-                    - "Off"
-                    - "Initial"
-                    - "Recreate"
-                    - "Auto"
             resourcePolicy:
               properties:
                 containerPolicies:
                   type: array
                   items:
```
schylek: After removing lines below items will be empty. Should be removed too?

bskiba: Done

PR #1129, opened 2018-08-08
Extract cache from GCE manager

aleksandra-malinowska:
This extracts cache management part of GCE manager in order to make it:
1) easier to understand and verify it is used correctly,
2) reusable for separating GKE cloud provider.

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/cache.go
+++ b/cluster-autoscaler/cloudprovider/gce/cache.go
@@ -21,33 +21,12 @@ import (
 	"reflect"
 	"strings"
 	"sync"
-	"time"
 
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
 
 	"github.com/golang/glog"
 	gce "google.golang.org/api/compute/v1"
 )
-
-// GceCache is used for caching cluster resources state.
-type GceCache struct {
-	// Cache content.
-	migs            []*MigInformation
-	migCache        map[GceRef]Mig
-	resourceLimiter *cloudprovider.ResourceLimiter
-	machinesCache   map[MachineTypeKey]*gce.MachineType
-	// Freshness timestamps.
-	machinesCacheLastRefresh time.Time
-	// Locks. Rules of locking:
-	// - migsMutex protects only migs.
-	// - cacheMutex protects migCache, resourceLimiter and machinesCache.
-	// - methods that take cacheMutex may call methods that take migsMutex.
-	// - methods that take migsMutex may not call methods that take cacheMutex.
-	cacheMutex sync.Mutex
-	migsMutex  sync.Mutex
-	// Service used to refresh cache.
-	GceService AutoscalingGceClient
-}
 
 // MigInformation is a wrapper for Mig.
 type MigInformation struct {
@@ -61,13 +40,49 @@ type MachineTypeKey struct {
 	MachineType string
 }
 
+// GceCache is used for caching cluster resources state.
+//
+// It is needed to:
+// - keep track of autoscaled MIGs in the cluster,
+// - keep track of instances and which MIG they belong to,
+// - limit repetitive GCE API calls.
+//
+// Cached resources:
+// 1) MIG configuration,
+// 2) instance->MIG mapping,
+// 3) resource limits (self-imposed quotas),
+// 4) machine types.
+//
+// How it works:
+// - migs (1), resource limits (3) and machine types (4) are only stored in this cache,
+// not updated by it.
+// - instancesCache (2) is based on registered migs (1). For each mig, its instances
+// are fetched from GCE API using gceService.
+// - instancesCache (2) is NOT updated automatically when migs field (1) is updated. Calling
+// RegenerateInstancesCache is required to sync it with registered migs.
+type GceCache struct {
+	// Cache content.
+	migs            []*MigInformation
+	instancesCache  map[GceRef]Mig
+	resourceLimiter *cloudprovider.ResourceLimiter
+	machinesCache   map[MachineTypeKey]*gce.MachineType
+	// Locks. Rules of locking:
+	// - migsMutex protects only migs.
+	// - cacheMutex protects instancesCache, resourceLimiter and machinesCache.
+	// - if both locks are needed, cacheMutex must be obtained before migsMutex.
+	cacheMutex sync.Mutex
+	migsMutex  sync.Mutex
+	// Service used to refresh cache.
+	GceService AutoscalingGceClient
+}
+
 // NewGceCache creates empty GceCache.
 func NewGceCache(gceService AutoscalingGceClient) GceCache {
 	return GceCache{
-		migs:          []*MigInformation{},
-		migCache:      map[GceRef]Mig{},
-		machinesCache: map[MachineTypeKey]*gce.MachineType{},
-		GceService:    gceService,
+		migs:           []*MigInformation{},
+		instancesCache: map[GceRef]Mig{},
+		machinesCache:  map[MachineTypeKey]*gce.MachineType{},
+		GceService:     gceService,
 	}
 }
 
@@ -90,6 +105,7 @@ func (gc *GceCache) RegisterMig(mig Mig) bool {
 	}
 
 	glog.V(1).Infof("Registering %s", mig.GceRef().String())
+	// TODO(aleksandra-malinowska): fetch and set MIG basename here.
 	gc.migs = append(gc.migs, &MigInformation{
 		Config: mig,
 	})
@@ -143,7 +159,7 @@ func (gc *GceCache) updateMigBasename(ref GceRef, basename string) {
 
 // Methods locking on cacheMutex.
 
-// GetMigForInstance returns MigConfig of the given Instance from cache.
+// GetMigForInstance returns Mig to which the given instance belongs.
 // Attempts to regenerate cache if there is a Mig with matching prefix in migs list.
 // TODO(aleksandra-malinowska): reconsider failing when there's a Mig with
 // matching prefix, but instance doesn't belong to it.
@@ -151,7 +167,7 @@ func (gc *GceCache) GetMigForInstance(instance *GceRef) (Mig, error) {
 	gc.cacheMutex.Lock()
 	defer gc.cacheMutex.Unlock()
 
-	if mig, found := gc.migCache[*instance]; found {
+	if mig, found := gc.instancesCache[*instance]; found {
 		return mig, nil
 	}
 
@@ -162,7 +178,7 @@ func (gc *GceCache) GetMigForInstance(instance *GceRef) (Mig, error) {
 			if err := gc.regenerateCache(); err != nil {
 				return nil, fmt.Errorf("Error while looking for MIG for instance %+v, error: %v", *instance, err)
 			}
-			if mig, found := gc.migCache[*instance]; found {
+			if mig, found := gc.instancesCache[*instance]; found {
 				return mig, nil
 			}
 			return nil, fmt.Errorf("Instance %+v does not belong to any configured MIG", *instance)
@@ -172,8 +188,8 @@ func (gc *GceCache) GetMigForInstance(instance *GceRef) (Mig, error) {
 	return nil, nil
 }
 
-// RegenerateCacheWithLock triggers cache regeneration under lock.
-func (gc *GceCache) RegenerateCacheWithLock() error {
+// RegenerateInstancesCache triggers instances cache regeneration under lock.
+func (gc *GceCache) RegenerateInstancesCache() error {
 	gc.cacheMutex.Lock()
 	defer gc.cacheMutex.Unlock()
 
@@ -182,7 +198,7 @@ func (gc *GceCache) RegenerateCacheWithLock() error {
 
 // internal method - should only be called after locking on cacheMutex.
 func (gc *GceCache) regenerateCache() error {
-	newMigCache := make(map[GceRef]Mig)
+	newInstancesCache := make(map[GceRef]Mig)
 
 	for _, migInfo := range gc.GetMigs() {
 		mig := migInfo.Config
@@ -200,11 +216,11 @@ func (gc *GceCache) regenerateCache() error {
 			return err
 		}
 		for _, ref := range instances {
-			newMigCache[ref] = mig
+			newInstancesCache[ref] = mig
 		}
 	}
 
-	gc.migCache = newMigCache
+	gc.instancesCache = newInstancesCache
 	return nil
 }
 
@@ -240,19 +256,10 @@ func (gc *GceCache) AddMachineToCache(machineType string, zone string, machine *
 	gc.machinesCache[MachineTypeKey{zone, machineType}] = machine
 }
 
-// SetMachinesCache sets the machines cache under lock and returns next recommended refresh time.
-func (gc *GceCache) SetMachinesCache(machinesCache map[MachineTypeKey]*gce.MachineType) time.Time {
+// SetMachinesCache sets the machines cache under lock.
+func (gc *GceCache) SetMachinesCache(machinesCache map[MachineTypeKey]*gce.MachineType) {
 	gc.cacheMutex.Lock()
 	defer gc.cacheMutex.Unlock()
 
 	gc.machinesCache = machinesCache
-	gc.machinesCacheLastRefresh = time.Now()
-	return gc.machinesCacheLastRefresh.Add(machinesRefreshInterval)
-}
-
-// Methods that need no lock.
-
-// MachinesCacheFresh return true if the last update to machines cache was within refesh interval.
-func (gc *GceCache) MachinesCacheFresh() bool {
-	return gc.machinesCacheLastRefresh.Add(machinesRefreshInterval).After(time.Now())
 }
--- a/cluster-autoscaler/cloudprovider/gce/cache.go
+++ b/cluster-autoscaler/cloudprovider/gce/cache.go
@@ -1,0 +1,258 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package gce
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+	"sync"
+	"time"
+
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
+
+	"github.com/golang/glog"
+	gce "google.golang.org/api/compute/v1"
+)
+
+// GceCache is used for caching cluster resources state.
+type GceCache struct {
+	// Cache content.
+	migs            []*MigInformation
+	migCache        map[GceRef]Mig
+	resourceLimiter *cloudprovider.ResourceLimiter
+	machinesCache   map[MachineTypeKey]*gce.MachineType
+	// Freshness timestamps.
+	machinesCacheLastRefresh time.Time
+	// Locks. Rules of locking:
+	// - migsMutex protects only migs.
+	// - cacheMutex protects migCache, resourceLimiter and machinesCache.
+	// - methods that take cacheMutex may call methods that take migsMutex.
+	// - methods that take migsMutex may not call methods that take cacheMutex.
+	cacheMutex sync.Mutex
+	migsMutex  sync.Mutex
+	// Service used to refresh cache.
+	GceService AutoscalingGceClient
+}
+
+// MigInformation is a wrapper for Mig.
+type MigInformation struct {
+	Config   Mig
+	Basename string
+}
+
+// MachineTypeKey is used to identify MachineType.
+type MachineTypeKey struct {
+	Zone        string
+	MachineType string
+}
+
+// NewGceCache creates empty GceCache.
+func NewGceCache(gceService AutoscalingGceClient) GceCache {
+	return GceCache{
+		migs:          []*MigInformation{},
+		migCache:      map[GceRef]Mig{},
+		machinesCache: map[MachineTypeKey]*gce.MachineType{},
+		GceService:    gceService,
+	}
+}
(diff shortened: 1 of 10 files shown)


MaciekPytel:
For future reference: it would be easier to review if making MIG an interface and refactoring cache were separate PRs.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -661,15 +661,43 @@ func (m *gceManagerImpl) fetchExplicitMigs(specs []string) error {
...
+	s := &dynamic.NodeGroupSpec{
+		Name:               link,
+		MinSize:            cfg.MinSize,
+		MaxSize:            cfg.MaxSize,
+		SupportScaleToZero: scaleToZeroSupported,
+	}
+	return m.buildMigFromSpec(s)
+}
+
+func (m *gceManagerImpl) buildMigFromSpec(s *dynamic.NodeGroupSpec) (Mig, error) {
```
MaciekPytel: Extracting this is a drive-by cleanup, that has nothing to do with this refactor. I'm always somewhat worried about those, especially in cases like that, where we're changing code specific to feature we don't test at all (autodiscovery). I think that just including this means we should probably include autodiscovery in your post-refactor testing.

Regardless of the above this should at least be a separate PR. Reviewing just this change would be quite simple, but it adds up, resulting in this huge PR (which it turn makes me delay review until I have enough time to go through the whole thing, rather than just keep reviewing as you send PRs).

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -661,15 +661,43 @@ func (m *gceManagerImpl) fetchExplicitMigs(specs []string) error {
...
+	if err := s.Validate(); err != nil {
```
MaciekPytel: This only happened in buildMigFromAutoCfg before, but not for explicitly configured migs. It sounds like a reasonable thing to do, but I have no idea what it really does. Are you sure this is ok?

cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -193,26 +193,26 @@ func (t *templateBuilder) buildNodeFromTemplate(mig *Mig, template *gce.Instance
 	return &node, nil
 }
 
-func (t *templateBuilder) buildNodeFromAutoprovisioningSpec(mig *Mig, cpu int64, mem int64) (*apiv1.Node, error) {
-	if mig.spec == nil {
-		return nil, fmt.Errorf("no spec in mig %s", mig.Name)
+func (t *templateBuilder) buildNodeFromAutoprovisioningSpec(mig Mig, cpu int64, mem int64) (*apiv1.Node, error) {
```
MaciekPytel: It's no longer called 'autoprovisioningSpec'.

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+// MigInformation is a wrapper for Mig.
+type MigInformation struct {
+	Config   Mig
+	Basename string
+}
+
+// MachineTypeKey is used to identify MachineType.
+type MachineTypeKey struct {
```
MaciekPytel: Can you move the struct definitions up, before you use them?

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -0,0 +1,258 @@
...
+	migCache        map[GceRef]Mig
+	resourceLimiter *cloudprovider.ResourceLimiter
+	machinesCache   map[MachineTypeKey]*gce.MachineType
+	// Freshness timestamps.
+	machinesCacheLastRefresh time.Time
+	// Locks. Rules of locking:
+	// - migsMutex protects only migs.
+	// - cacheMutex protects migCache, resourceLimiter and machinesCache.
+	// - methods that take cacheMutex may call methods that take migsMutex.
+	// - methods that take migsMutex may not call methods that take cacheMutex.
```
MaciekPytel: What's the reasoning behind last 2 rules? Do you mean "if you ever need both cacheMutex and migMutex always grab cacheMutex first to avoid deadlocks"?

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -0,0 +1,258 @@
...
+// GetMigs returns a copy of migs list.
+func (gc *GceCache) GetMigs() []*MigInformation {
+	gc.migsMutex.Lock()
+	defer gc.migsMutex.Unlock()
+
+	migs := make([]*MigInformation, 0, len(gc.migs))
+	for _, mig := range gc.migs {
+		migs = append(migs, &MigInformation{
+			Basename: mig.Basename,
+			Config:   mig.Config,
```
MaciekPytel: It seems like the semantic here is changing rather significantly. Previously MigInformation had a Mig struct, but now it has Mig interface implemented by *gceMig. Are we implicitly moving from deep to shallow copy here?

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -0,0 +1,258 @@
...
+	for _, mig := range gc.migs {
+		if mig.Config.GceRef() == ref {
+			mig.Basename = basename
+		}
+	}
+}
+
+// Methods locking on cacheMutex.
+
+// GetMigForInstance returns MigConfig of the given Instance from cache.
```
MaciekPytel: I realize this is a copy-paste, but this comment is really bad. MigConfig makes it look like a name, but we don't have anything named that. Also what is mig config of instance?

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -0,0 +1,258 @@
...
+			}
+			return nil, fmt.Errorf("Instance %+v does not belong to any configured MIG", *instance)
+		}
+	}
+	// Instance doesn't belong to any configured mig.
+	return nil, nil
+}
+
+// RegenerateCacheWithLock triggers cache regeneration under lock.
+func (gc *GceCache) RegenerateCacheWithLock() error {
```
MaciekPytel: I don't like this method name. It's the only way to regenerate cache that is externally available and the caller shouldn't need to know about internal locking. It should be called RegenerateCache or even better RegenerateInstanceCache.

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -0,0 +1,258 @@
...
+			newMigCache[ref] = mig
+		}
+	}
+
+	gc.migCache = newMigCache
+	return nil
+}
+
+// SetResourceLimiter sets resource limiter.
+func (gc *GceCache) SetResourceLimiter(resourceLimiter *cloudprovider.ResourceLimiter) {
```
MaciekPytel: This has really nothing to do with this cache, except we happen to have some mutex in cache that we could use to protect it.

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -0,0 +1,258 @@
...
+	gc.machinesCache = machinesCache
+	gc.machinesCacheLastRefresh = time.Now()
+	return gc.machinesCacheLastRefresh.Add(machinesRefreshInterval)
+}
+
+// Methods that need no lock.
+
+// MachinesCacheFresh return true if the last update to machines cache was within refesh interval.
+func (gc *GceCache) MachinesCacheFresh() bool {
+	return gc.machinesCacheLastRefresh.Add(machinesRefreshInterval).After(time.Now())
```
MaciekPytel: Does this really need no lock? Is resolving a field in struct and calling a method on it atomic in golang? We didn't protect the last refresh timestamp before, because we only ever accessed it from one goroutine anyway. If we're placing it behind a thread-safe interface we may need to reevaluate that.

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -0,0 +1,258 @@
...
+
+// GceCache is used for caching cluster resources state.
+type GceCache struct {
+	// Cache content.
+	migs            []*MigInformation
+	migCache        map[GceRef]Mig
+	resourceLimiter *cloudprovider.ResourceLimiter
+	machinesCache   map[MachineTypeKey]*gce.MachineType
+	// Freshness timestamps.
+	machinesCacheLastRefresh time.Time
```
MaciekPytel: Why move machinesCacheLastRefresh here and related logic here, but keep lastRefresh in GceManager? It seems very inconsistent.

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -0,0 +1,258 @@
...
+// AddMachineToCache adds machine to cache under lock.
+func (gc *GceCache) AddMachineToCache(machineType string, zone string, machine *gce.MachineType) {
+	gc.cacheMutex.Lock()
+	defer gc.cacheMutex.Unlock()
+
+	gc.machinesCache[MachineTypeKey{zone, machineType}] = machine
+}
+
+// SetMachinesCache sets the machines cache under lock and returns next recommended refresh time.
+func (gc *GceCache) SetMachinesCache(machinesCache map[MachineTypeKey]*gce.MachineType) time.Time {
```
MaciekPytel: Why would refreshing cache return a timestamp? It feels really surprising and not particularly useful if the cache is keeping track of when it should be updated by implementing MachinesCacheFresh. Since we only use it to log, we could log inside this method and not care about it (assuming we don't want to move machinesCacheLastRefresh back to GceManager in the first place).

cluster-autoscaler/cloudprovider/gce/gce_manager.go
aleksandra-malinowska: Fair point. Although more PRs also mean more review cycles :)

cluster-autoscaler/cloudprovider/gce/gce_manager.go
aleksandra-malinowska: Validate() is a basic sanity check: node group name must not be empty, maxSize >= minSize, minSize > 0 (or >= 0 if scale to zero isn't supported.)  It's true we never checked any of this for explicitly configured migs before, which is all the more reason I'd like to add it.

https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/config/dynamic/node_group_spec.go#L68

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -232,40 +235,63 @@ func GceRefFromProviderId(id string) (*GceRef, error) {
...
 	machineType    string
 	labels         map[string]string
 	taints         []apiv1.Taint
 	extraResources map[string]resource.Quantity
 }
 
 // Mig implements NodeGroup interface.
-type Mig struct {
-	GceRef
+type Mig interface {
```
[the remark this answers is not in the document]
aleksandra-malinowska: Reusing cache requires it to operate on an interface type. These methods may be removed from the interface if they're no longer needed.

cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -227,29 +227,27 @@ func (t *templateBuilder) buildNodeFromAutoprovisioningSpec(mig *Mig, cpu int64,
...
-	if mig.spec.labels != nil {
-		for k, v := range mig.spec.labels {
-			if existingValue, found := labels[k]; found {
-				if v != existingValue {
-					return map[string]string{}, fmt.Errorf("conflict in labels requested: %s=%s  present: %s=%s",
-						k, v, k, existingValue)
-				}
-			} else {
-				labels[k] = v
+	for k, v := range mig.Spec().labels {
```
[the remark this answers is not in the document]
aleksandra-malinowska: This check would have never saved us anyway, given that earlier in this function there was:

```labels, err := buildGenericLabels(mig.GceRef, mig.spec.machineType, nodeName)```

I. e., we would have failed with nil pointer there before ever reaching the if in question.

I can add checking if mig.Spec() != nil at the beginning of every function that uses it,  if that makes it better.

File: cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go [hunk 15]
```diff
@@ -232,40 +235,63 @@ func GceRefFromProviderId(id string) (*GceRef, error) {
 	}, nil
 }
 
-// Information about what machines in an autoprovisioned MIG would look like.
-type autoprovisioningSpec struct {
+// Information about what machines in a MIG look like.
+type MigSpec struct {
```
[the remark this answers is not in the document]
aleksandra-malinowska: I thought we wanted to be able to access information about a node group's characteristics between restarts/even if it wasn't autoprovisioned?

Also, nothing in this struct is specific to this use case. "Autoprovisioning" spec describes how we use it now, not what it is. But I can change it back if you're sure this is better.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -84,22 +82,8 @@ func init() {
...
-
-type machineTypeKey struct {
-	zone        string
-	machineType string
-}
-
 // GceManager handles gce communication and data caching.
 type GceManager interface {
-	// RegisterMig registers mig in GceManager. Returns true if the node group didn't exist before.
-	RegisterMig(mig Mig) bool
```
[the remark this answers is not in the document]
aleksandra-malinowska: GceManager is only used by gce_cloud_provider.go (and it only exists for testing GceCloudProvider with a mock, BTW). However, RegisterMig() was always called only by the gceManagerImpl itself. Therefore, it's perfectly useless as part of this interface, whether exported or not.

I originally intended to embed gceManagerImpl (after exporting it) in gkeManager, in order to reuse the logic as much as possible. For that, almost all of its methods and fields would have to be exported. However, given that most of the shared logic is being extracted to clients and cache, we can probably avoid embedding and create independent implementation with only a limited amount of duplicated code. Then this method could be private. WDYT?

cluster-autoscaler/cloudprovider/gce/cache.go
aleksandra-malinowska: We used this same lock to protect resource limits before. Also, I'd argue that if the intended behavior is to fetch limits from the API, but we don't want to do this too often, so we keep them and refresh only now and then, it technically makes it a cached value.

cluster-autoscaler/cloudprovider/gce/templates.go
aleksandra-malinowska: Done

cluster-autoscaler/cloudprovider/gce/cache.go
aleksandra-malinowska: Yes. Rephrased since it wasn't clear

cluster-autoscaler/cloudprovider/gce/cache.go
aleksandra-malinowska: Done

cluster-autoscaler/cloudprovider/gce/cache.go
aleksandra-malinowska: Done. Also renamed migCache to instancesCache for consistency.

cluster-autoscaler/cloudprovider/gce/cache.go
aleksandra-malinowska: Moved to gceManagerImpl as suggested in other comments

cluster-autoscaler/cloudprovider/gce/cache.go
aleksandra-malinowska: Good point, moved back to gceManagerImpl.

cluster-autoscaler/cloudprovider/gce/cache.go
aleksandra-malinowska: Moved to gceManagerImpl as suggested in other comments

cluster-autoscaler/cloudprovider/gce/cache.go
aleksandra-malinowska: I can if you prefer, although I personally find it more readable to start with top-level/most relevant struct and define pieces as necessary (similar to proto definitions.) Looking at standard library, it seems to be mixing type declarations before and after use, so I guess it's not something Go is opinionated about.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
MaciekPytel: It seems we actually do check it already: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/config/dynamic/node_group_spec.go#L60. We call SpecFromString immediately before buildMigFromSpec, so with your change we end up calling Validate() twice for explicitly configured MIGs. I guess it doesn't matter all that much in that case.

cluster-autoscaler/cloudprovider/gce/templates.go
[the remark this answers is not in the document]
MaciekPytel: It does. Though, depending on the number of such functions I may prefer it in a separate PR. Since this PR doesn't make it any worse, I guess it's not blocking it anyway.

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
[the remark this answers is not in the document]
MaciekPytel: I'd love to have it available for non-autoprovisioned node groups. The only question is how to populate it, though I guess we could re-use some of TemplateNodeInfo logic for at least some of this stuff. But we don't really need it outside of GKE right now, so we'll most likely never end up implementing it.

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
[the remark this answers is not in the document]
MaciekPytel: Do we need an extended implementation for GKE? Other than NodePoolName(), which is very clearly GKE specific.

cluster-autoscaler/cloudprovider/gce/cache.go
MaciekPytel: Right, I thought it was a field, because I was going commit-by-commit through the review and not looking at the whole thing (because of how large it is :P). I think it *should* be a deep copy, but that would require a more careful consideration. Let's leave it as is for this PR and think if we need a follow up making it a deepcopy (note - it's probably only relevant with autoprovisioning, so it may not be as battle-tested as we may think it is).

cluster-autoscaler/cloudprovider/gce/gce_manager.go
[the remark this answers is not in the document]
MaciekPytel: By that logic having anything public on GceManager is pointless (since we never use it outside of package and, hopefully, never will). My intuition would be that anything we call from different files (gce_cloud_provider.go in particular) should be public and anything else private, but apparently that's not true and it's just completely random :(

Still, we don't call it from outside manager, so if we don't need this in GKE I'm ok with it being public.

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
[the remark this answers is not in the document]
aleksandra-malinowska: I suppose if we're happy with keeping node pool name, spec, exist and autoprovisioned fields in GCE implementation, we might not need it after all.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -120,15 +120,16 @@ type gceManagerImpl struct {
...
+	location                 string
+	projectId                string
+	clusterName              string
+	mode                     GcpCloudProviderMode
+	templates                *templateBuilder
+	interrupt                chan struct{}
+	regional                 bool
+	explicitlyConfigured     map[GceRef]bool
+	migAutoDiscoverySpecs    []cloudprovider.MIGAutoDiscoveryConfig
+	machinesCacheLastRefresh time.Time
```
MaciekPytel: nit: I would prefer if this was defined next to lastRefresh since the two are very similar and connected to cache.

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -0,0 +1,245 @@
...
+
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
+
+	"github.com/golang/glog"
+	gce "google.golang.org/api/compute/v1"
+)
+
+// GceCache is used for caching cluster resources state.
+type GceCache struct {
+	// Cache content.
```
MaciekPytel: I know it wasn't there before, but I'd like some general comment about what we're caching, why we need it and how all this stuff works. Hopefully it would make it easier for people to approach cloudprovider code in future.

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -0,0 +1,245 @@
...
+				glog.V(4).Infof("Updated Mig %s", mig.GceRef().String())
+				return true
+			}
+			return false
+		}
+	}
+
+	glog.V(1).Infof("Registering %s", mig.GceRef().String())
+	gc.migs = append(gc.migs, &MigInformation{
+		Config: mig,
```
MaciekPytel: Can you add TODO: fetch and set Basename here? Currently we rely on someone outside of cache calling RegenerateInstancesCache before GetMigForInstance (if they called Mig with empty basename matches the prefix of every node bad things will happen). At least that's the impression I get from the code.

MaciekPytel:
/lgtm
Please squash fixup commits and let's merge this.

cluster-autoscaler/cloudprovider/gce/cache.go
aleksandra-malinowska: Done

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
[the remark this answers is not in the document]
aleksandra-malinowska: Done

cluster-autoscaler/cloudprovider/gce/cache.go
aleksandra-malinowska: Done

cluster-autoscaler/cloudprovider/gce/gce_manager.go
aleksandra-malinowska: Done

cluster-autoscaler/cloudprovider/gce/cache.go
aleksandra-malinowska: Done

PR #648, opened 2018-02-13
AWS: Fix nvidia gpu resource name, re-enable ScaleToZero

discordianfish:
These are two somewhat separate changes, so let me know if I should split this PR:

1) Rename gpu resource name, the same as #407 but for AWS
2) Set scaleToZero = true. It looks like this was changed (by accident?) in #462

I've tested the changes manually and scaling to and from 0 works on my cluster with the nvidia k8s device plugin.

PR #1130, opened 2018-08-08
cherry pick: "Fix nvidia gpu resource name on AWS" to 1.1

jeruane:
cherry pick of:
https://github.com/kubernetes/autoscaler/pull/648


rpunetha:
When will this merge be performed? I'm having issue with version 1.1.1 of ASG.

MaciekPytel:
/lgtm

A word of caution - regardless of this fix CA only works correctly with GPU in GCP/GKE. For other providers it will create too many nodes with GPU when scaling-up due to https://github.com/kubernetes/kubernetes/issues/54959. We still don't have a generic fix for this issue, due to reasons explained in my comment https://github.com/kubernetes/autoscaler/pull/597#issuecomment-362231521.

PR #1131, opened 2018-08-08
cherry pick: "Fix nvidia gpu resource name on AWS" to 1.0

jeruane:
cherry pick of:
https://github.com/kubernetes/autoscaler/pull/648


MaciekPytel:
There will probably be no more CA 1.0.X releases. CA versions are normally released together with Kubernetes version, but Kubernetes community only supports last 3 versions. CA 1.0.X was bundled with Kubernetes 1.8, which has reached end-of-life. 

I would recommend moving to a newer CA version. We only test each version with the corresponding k8s version, but using newer CA with older Kubernetes should generally work fine unless you use some really complex scheduling.

PR #1142, opened 2018-08-13
Create GKE cloud provider

aleksandra-malinowska:
For now, GKE cloud provider is just copied GCE cloud provider. Next PRs will prune redundant functionality (GCE-only stuff from GKE and vice versa.)

Also contains the last (hopefully) of necessary refactors - mostly exporting stuff so that it can be used from another package.


cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -76,6 +76,24 @@ func NewAutoscalingGceClientV1(client *http.Client, projectId string) (*autoscal
 	}, nil
 }
 
+// NewCustomAutoscalingGceClientV1 creates a new client using custom server url and timeouts
+// for communicating with GCE v1 API.
+func NewCustomAutoscalingGceClientV1(client *http.Client, projectId, serverUrl string,
```
aleksandra-malinowska: This is just for use in unit tests of GKE manager. Once we get rid of mocking API via http server, it'll be replaced with fake or mock.

cluster-autoscaler/cloudprovider/gce/autoscaling_gke_client.go
```diff
@@ -24,7 +24,7 @@ import (
 
 // This flag is outside main as it's only useful for test/development.
 var (
-	gkeAPIEndpoint = flag.String("gke-api-endpoint", "", "GKE API endpoint address. This flag is used by developers only. Users shouldn't change this flag.")
+	GkeAPIEndpoint = flag.String("gke-api-endpoint", "", "GKE API endpoint address. This flag is used by developers only. Users shouldn't change this flag.")
```
aleksandra-malinowska: Will be private again once it's removed from GCE. For now, it has to be available in both packages since flag can't be redeclared.

PR #1147, opened 2018-08-14
Cluster Autoscaler 1.2.3-beta.1

aleksandra-malinowska:
Update Cluster Autoscaler version to 1.2.3-beta.1

### DIFF
--- a/cluster-autoscaler/version.go
+++ b/cluster-autoscaler/version.go
@@ -17,4 +17,4 @@ limitations under the License.
 package main
 
 // ClusterAutoscalerVersion contains version of CA.
-const ClusterAutoscalerVersion = "1.2.2"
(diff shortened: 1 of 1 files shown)


MaciekPytel:
/lgtm
Although I'm not sure how are we planning to test this beta. In the past we usually only did a manual test of changes included in patch release. After changing godeps we really should run all e2e tests. That being said as 1.10 patch release manager I don't feel too happy about cherry-picking beta version just to run e2e on it.

Same applies to https://github.com/kubernetes/autoscaler/pull/1146.

aleksandra-malinowska:
Good point about testing - we've never committed a beta version to a release branch before. I wonder if we could override the image just for e2e tests (at least on GCE)?

PR #1151, opened 2018-08-15
Improve getMatchingVPA logging to be more clear (with pod name) and using logging level

phsiao:
Currently in getMatchingVPA, it can generate a lot of lines of "Let's choose from x configs" without any context if you set v=1.  This PR sets the logging level to be identical to surrounding logging level, as well as add pod name to the log line for clarity.


phsiao:
That is even better, should have done.  Updated.

PR #1153, opened 2018-08-16
Add support for customizing qps and burst for recommender via --kubeapi-qps and --kube-api-burst

phsiao:
For a cluster with a large number of vpa objects, the default qps is too low (5 I believe) for it to finish processing quickly and you end up getting a lot of noise related to throttling.

This PR adds basic support for specifying and customizing qps and burst, and will be applied the setting to both checkpoint writer and vpcClient.

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/main.go
+++ b/vertical-pod-autoscaler/pkg/recommender/main.go
@@ -37,13 +37,15 @@ var (
 	prometheusAddress      = flag.String("prometheus-address", "", `Where to reach for Prometheus metrics`)
 	storage                = flag.String("storage", "", `Specifies storage mode. Supported values: prometheus, checkpoint (default)`)
 	address                = flag.String("address", ":8942", "The address to expose Prometheus metrics.")
+	kubeApiQps             = flag.Float64("kube-api-qps", 5.0, `QPS limit when making requests to Kubernetes apiserver`)
+	kubeApiBurst           = flag.Float64("kube-api-burst", 10.0, `QPS burst limit when making requests to Kubernetes apiserver`)
 )
 
 func main() {
 	kube_flag.InitFlags()
 	glog.V(1).Infof("Vertical Pod Autoscaler %s Recommender", common.VerticalPodAutoscalerVersion)
 
-	config := createKubeConfig()
+	config := createKubeConfig(float32(*kubeApiQps), int(*kubeApiBurst))
 
 	metrics.Initialize(*address)
 	metrics_recommender.Register()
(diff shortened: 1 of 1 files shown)


bskiba:
I wonder if this is the right way to address this. Maybe the problem is we make to many calls to the API server? Can you share a bit about which calls are throttled and how big is the setup when you run into this?

phsiao:
The calls being throttled include

1. loading from checkpoints during startup, https://github.com/kubernetes/autoscaler/blob/9feb612e2200e04c2ac8f7dd7cc8e7adbe300add/vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go#L190-L192

2. updating vpa object periodically, https://github.com/kubernetes/autoscaler/blob/9feb612e2200e04c2ac8f7dd7cc8e7adbe300add/vertical-pod-autoscaler/pkg/recommender/routines/recommender.go#L105-L106

The second one is much more visible due to its frequent nature.

I am testing this on a cluster with ~100 nodes and there are 242 vpa objects.

With the throttling it would take more than a couple of minutes for UpdateVPAs() to complete, and the recommender pauses for about a minute and repeat the process again.

Do you think we can avoid updating vpa periodically?

bskiba:
From the checkpoints side, I believe that the problem is with "MaintainCheckpoints" step which includes storing checkpoints for all VPAs (adding informer for checkpoints won't help here.

So I experimented a bit with this, and I think regardless fo the possible fixes that would allow us to do less object updates, for big clusters the default throttling will be too low. I think we should merge this, just have one minor comment on the pr.

vertical-pod-autoscaler/pkg/recommender/main.go
```diff
@@ -37,13 +37,17 @@ var (
...
+	kubeApiQps             = flag.Float64("kube-api-qps", 5.0, `QPS limit when making requests to Kubernetes apiserver`)
+	kubeApiBurst           = flag.Float64("kube-api-burst", 10.0, `QPS burst limit when making requests to Kubernetes apiserver`)
 )
 
 func main() {
 	kube_flag.InitFlags()
 	glog.V(1).Infof("Vertical Pod Autoscaler %s Recommender", common.VerticalPodAutoscalerVersion)
 
 	config := createKubeConfig()
+	config.QPS = float32(*kubeApiQps)
```
bskiba: Can you do the assignment inside of createKubeConfig function and pass kubeApiQps and kubeApiBurst through parameters?

phsiao: Yep, I will take a look today.

phsiao:
@bskiba revise as suggested.

PR #1154, opened 2018-08-17
Mention --maxGracefulTerminationSec in FAQ.md

devkid:
The FAQ was wrong in a way that it said that pods are given at most 10 minutes to gracefully terminate, but the maximum time is actually configurable.


MaciekPytel:
Hi @devkid,
We can't accept your contribution before you sign CLA. Otherwise looks good to me.

devkid:
Done

PR #1156, opened 2018-08-17
Cherry pick of #1155: Update Cluster Autoscaler base image

aleksandra-malinowska:
Cherry pick of #1155: Update Cluster Autoscaler base image to debian-base-amd:0.3.2

### DIFF
--- a/cluster-autoscaler/Dockerfile
+++ b/cluster-autoscaler/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM k8s.gcr.io/debian-base-amd64:0.3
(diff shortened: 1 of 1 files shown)


MaciekPytel:
/lgtm
Should we also cherry-pick to 1.1 and 1.2?

aleksandra-malinowska:
> Should we also cherry-pick to 1.1 and 1.2?

I guess it depends for how long are we going to maintain those. Since there's a pending release for 1.2, I'd say cherry-pick it there. Not sure about 1.1, WDYT?

MaciekPytel:
Kubernetes policy is to maintain last 3 versions, so I guess we support 1.1 until we release 1.4.

PR #1174, opened 2018-08-21
Update owners files for autoscaling

### LINKED ISSUE #1111: Create an OWNERS file
ref: https://github.com/kubernetes/community/issues/1721

This repo is listed in sigs.yaml as belonging to sig-autoscaling, but it lacks an OWNERS file.  Can you please add one with the appropriate list of approvers and reviewers? eg:
- [robust example: approvers, reviewers, aliases](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/OWNERS)
- [simple example: only approvers, no aliases](https://github.com/kubernetes-incubator/external-storage/blob/master/OWNERS)

For more information see the [OWNERS docs](https://go.k8s.io/owners)

/sig autoscaling

/assign @mwielgus @DirectXMan12 
sig-autoscaling chairs

/assign @aleksandra-malinowska @bskiba @MaciekPytel 
as folks who've merged PR's here recently

bskiba:
fixes #1111

### DIFF
--- a/cluster-autoscaler/OWNERS
+++ b/cluster-autoscaler/OWNERS
@@ -1,9 +1,9 @@
 approvers:
-- fgrzadkowski
-- mwielgus
 - piosz
+- aleksandram
+- lukaszos
 reviewers:
 - feiskyer
-- fgrzadkowski
-- mwielgus
 - piosz
+- aleksandram
+- lukaszos
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,8 @@
+approvers:
+- mwielgus
+- maciekpytel
+- bskiba
+reviewers:
+- mwielgus
+- maciekpytel
+- bskiba
--- a/addon-resizer/OWNERS
+++ b/addon-resizer/OWNERS
@@ -1,8 +1,10 @@
 approvers:
-- Q-Lee
 - piosz
 - wojtek-t
+- kawych
+- x13n
 reviewers:
-- Q-Lee
 - piosz
 - wojtek-t
+- kawych
+- x13n
--- a/vertical-pod-autoscaler/OWNERS
+++ b/vertical-pod-autoscaler/OWNERS
@@ -1,0 +1,7 @@
+approvers:
+- schylek
+- kgolab
+reviewers:
+- schylek
+- kgolab
+


bskiba:
@mwielgus @MaciekPytel

cluster-autoscaler/OWNERS
```diff
@@ -1,9 +1,11 @@
 approvers:
-- fgrzadkowski
 - mwielgus
```
MaciekPytel: mwielgus is already a top level owner, so probably redundant here (or I want to be included as well :) )

aleksandra-malinowska:
Do we want to add an OWNERS file for builder/ as well? It's specific to Cluster Autoscaler and changes there are necessary e.g. when adding extra build dependency (that won't be included in the release image anyway). However, with the new setup it will require approval by top-level OWNERS. Given that changes such as adding dependencies to be included in the release image only require cluster-autoscaler/OWNERS approval, this doesn't seem right.

bskiba:
@aleksandra-malinowska shouldn't the builder be used for all autoscaling images (I have some wip that adds build-in-docker to vpa)? Then this would imply that top-level OWNERS is right for it.

aleksandra-malinowska:
If this is going to be the case, then sure, it should probably require top-level approval, to avoid one project's dependencies' update from breaking another one. Although I don't think it's the best idea to share it, given that each project may require different dependencies. In particular, sharing it would mean we'd always have to sync upgrading to use newer version of Go for all projects (which would in some cases mean synchronized godeps update for all of them.)

MaciekPytel:
Maybe we should add addon-resizer owners too while we're doing it? I guess that would be @x13n and @kawych?

aleksandra-malinowska:
addon-resizer/OWNERS already exists, but +1 to updating it (AFAIU Q-Lee is no longer working on this, and current maintainers are missing)

cluster-autoscaler/OWNERS
bskiba: Done

bskiba:
Added addon-resizer changes.

bskiba:
@kawych @x13n FYI

PR #1179, opened 2018-08-22
Handle channel closes from pod event watcher

bskiba:
Issue: After ~30 minutes pod eviction event watch channel is closed. We are reading from a closed channel in an infinite loop that burns cpu. Causes recommender to start eating way too much cpu ~30 mins after start.
Fix: Watch for channel closure and restart the watch.


bskiba:
@jbartosik Can you review?

jbartosik:
Looks good to me.

PR #1180, opened 2018-08-22
Add batch jobs and cronjobs to Azure cluster-autoscaler

romanlutz:
Jobs and CronJobs were both missing so far which makes the cluster-autoscaler complain that it doesn't have access to "batch".


feiskyer:
@romanlutz  Thanks for the fix. Could you also add this to other yamls under same directory?

romanlutz:
@feiskyer I'm trying to get the cla/linuxfoundation check to go green, but it's not changing. Any idea? The hints they've given under "Details" don't work / are outdated.

feiskyer:
And after that,  comment with a single line `/check-cla` should make it green.

romanlutz:
It still refuses to go green... Not sure what else to do. I definitely signed the agreement and the email is set correctly. Maybe I need to recreate the PR since the first commit is with a different "user" apparently.

PR #1185, opened 2018-08-23
Report reason why pod didn't trigger scale-up

aleksandra-malinowska:
Report reason why pod didn't trigger scale-up in NotTriggeredScaleUp event. Format is similar to FailedScheduling, i.e. list of "\<failure reason\> (\<number of affected node groups\>)".

This can be further improved by:
- detailing reason why node group was skipped (e.g. max size reached, not healthy),
- communicating why NAP couldn't create a node group,
- passing raw predicate errors plus the above information to scale up status processor, so it can decide the message format.

### DIFF
--- a/cluster-autoscaler/core/scale_up.go
+++ b/cluster-autoscaler/core/scale_up.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"fmt"
 	"math"
+	"strings"
 	"time"
 
 	apiv1 "k8s.io/api/core/v1"
@@ -33,6 +34,7 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/metrics"
 	ca_processors "k8s.io/autoscaler/cluster-autoscaler/processors"
 	"k8s.io/autoscaler/cluster-autoscaler/processors/status"
+	"k8s.io/autoscaler/cluster-autoscaler/simulator"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/glogx"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/gpu"
@@ -243,11 +245,11 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 
 	loggingQuota := glogx.PodsLoggingQuota()
 
-	podsRemainUnschedulable := make(map[*apiv1.Pod]bool)
+	podsRemainUnschedulable := make(map[*apiv1.Pod][]*simulator.PredicateError)
 
 	for _, pod := range unschedulablePods {
 		glogx.V(1).UpTo(loggingQuota).Infof("Pod %s/%s is unschedulable", pod.Namespace, pod.Name)
-		podsRemainUnschedulable[pod] = true
+		podsRemainUnschedulable[pod] = []*simulator.PredicateError{}
 	}
 	glogx.V(1).Over(loggingQuota).Infof("%v other pods are also unschedulable", -loggingQuota.Left())
 	nodeInfos, err := GetNodeInfosForGroups(nodes, context.CloudProvider, context.ClientSet,
@@ -301,7 +303,11 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 		}
 	}
 
+	skippedNodeGroups := 0
 	for _, nodeGroup := range nodeGroups {
+		// All node groups are potentially skippable until proven otherwise.
+		skippedNodeGroups++
+
 		// Autoprovisioned node groups without nodes are created later so skip check for them.
 		if nodeGroup.Exist() && !clusterStateRegistry.IsNodeGroupSafeToScaleUp(nodeGroup.Id(), now) {
 			glog.Warningf("Node group %s is not ready for scaleup", nodeGroup.Id())
@@ -314,7 +320,6 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 			continue
 		}
 		if currentTargetSize >= nodeGroup.MaxSize() {
-			// skip this node group.
 			glog.V(4).Infof("Skipping node group %s - max size reached", nodeGroup.Id())
 			continue
 		}
@@ -336,14 +341,32 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 			continue
 		}
 
+		// Beyond this point, scaling the node group is considered an option.
+		skippedNodeGroups--
+
 		option := expander.Option{
 			NodeGroup: nodeGroup,
 			Pods:      make([]*apiv1.Pod, 0),
 		}
 
-		option.Pods = FilterSchedulablePodsForNode(context, unschedulablePods, nodeGroup.Id(), nodeInfo)
+		schedulableOnNode := CheckPodsSchedulableOnNode(context, unschedulablePods, nodeGroup.Id(), nodeInfo)
+		for pod, err := range schedulableOnNode {
+			if err != nil {
+				// Aggregate errors across existing node groups.
+				// TODO(aleksandra-malinowska): figure out how to communicate
+				// reasons NAP can't create a node-pool, if it's enabled.
+				if nodeGroup.Exist() {
+					errs, found := podsRemainUnschedulable[pod]
+					if found {
+						podsRemainUnschedulable[pod] = append(errs, err)
+					}
+				}
+			} else {
+				option.Pods = append(option.Pods, pod)
+			}
+		}
 		for _, pod := range option.Pods {
-			podsRemainUnschedulable[pod] = false
+			delete(podsRemainUnschedulable, pod)
 		}
 		passingPods := make([]*apiv1.Pod, len(option.Pods))
 		copy(passingPods, option.Pods)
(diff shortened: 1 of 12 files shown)


MaciekPytel:
Can you paste an example event in github comment? It would be useful to see how those look like and maybe discuss the format, etc.

cluster-autoscaler/core/scale_up.go
```diff
@@ -475,36 +498,54 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
...
-			result = append(result, pod)
+func getRemainingPods(schedulingErrors map[*apiv1.Pod][]*simulator.PredicateError, skipped int) map[*apiv1.Pod]string {
+	remaining := map[*apiv1.Pod]string{}
+	for pod, errs := range schedulingErrors {
+		messages := []string{}
+		aggregated := map[string]int{}
+		for _, err := range errs {
+			// Aggregate by reason, since predicate name isn't as helpful.
+			// If two different predicates fail with the same reason, so be it.
+			for _, reason := range err.Reasons() {
```
MaciekPytel: Scheduler aggregates by predicate name. I agree reason is probably more useful, but I wonder how well it aggregates? For example will we end with separate entries looking something like "insufficient cpu: required 2500m, available 2000m", "insufficient cpu: required 2500m, available 1000m?

I'd like to see some example events for most common scenarios (insufficient cpu/mem/gpu, taint, nodeSelector/nodeAffinity).

cluster-autoscaler/core/utils.go
```diff
@@ -67,9 +67,10 @@ const (
 // UID as a key in initial lookup and only run full comparison on a set of
 // podSchedulableInfos created for pods owned by this controller.
 type podSchedulableInfo struct {
-	spec        apiv1.PodSpec
-	labels      map[string]string
-	schedulable bool
+	spec            apiv1.PodSpec
+	labels          map[string]string
+	schedulable     bool
```
MaciekPytel: Do we need a bool field if we have error? I assume `schedulable == (schedulingError == nil)` is always going to be true?

MaciekPytel: Ok, I see we don't always use it. Still I sort of don't like the resulting semantics What would be the meaning of scheduled==true and schedulingError!=nil? Maybe just have a singleton throwaway PredicateError and use this if we don't care about actual error?

cluster-autoscaler/core/utils.go
```diff
@@ -121,20 +122,24 @@ func FilterOutSchedulable(unschedulableCandidates []*apiv1.Pod, nodes []*apiv1.N
...
+				glogx.V(4).UpTo(loggingQuota).Infof("Pod %s marked as unschedulable can be scheduled on %s. Ignoring in scale up.", pod.Name, nodeName)
 			}
-			continue
 		}
-		if nodeName, err := predicateChecker.FitsAny(pod, nodeNameToNodeInfo); err == nil {
-			glogx.V(4).UpTo(loggingQuota).Infof("Pod %s marked as unschedulable can be scheduled on %s. Ignoring in scale up.", pod.Name, nodeName)
-			podSchedulable.set(pod, true)
-		} else {
+
+		if !schedulable {
```
MaciekPytel: Why refactor this whole method just because of minor change of signature of cache? Personally I find old version much easier to follow - check for cached result, run predicates if its missing.

aleksandra-malinowska: I thought this is what the new version does more clearly ("check for cached result, run predicates if missing"), i.e. follows the general pattern of:
- try to get data from cache,
- if missing, get it the hard way and put it in cache,
- do common logic on the data we wanted in the first place (whether it came from cache or not.)

MaciekPytel: I guess it's personal preference. I see your point, but the common logic is a single line so duplication is minimal. If it was more I would extract it to function instead of refactoring like you described. The old way felt like a decision tree cache/no cache, schedulable/non schedulable. Only one variable on each level to think about. Here the conditions are more complex and I have to think if each possible combination is handled, which was trivial in old version.

cluster-autoscaler/core/utils.go
```diff
@@ -173,34 +178,38 @@ func FilterOutExpendablePods(pods []*apiv1.Pod, expendablePodsPriorityCutoff int
...
+		// Check if pod isn't repeated before overwriting result for it.
+		if _, repeated := schedulingErrors[pod]; repeated {
+			// This shouldn't really happen.
+			glog.Warningf("Pod %v appears multiple time on pods list, will only count it once in scale-up simulation", pod)
+		}
+
+		schedulingErrors[pod] = err
+
+		if !schedulable {
+			if found {
```
MaciekPytel: Similar to previous method. I'd expect found to only show up at the beginning of the method. As soon as we need to do anything custom for cached result I find old flow much easier to follow.

cluster-autoscaler/core/scale_up.go
aleksandra-malinowska: I think basic resources are all checked by PodFitsResources predicate, and scheduler doesn't aggregate by this name.

Example event running current version of this PR for insufficient CPU:

```3m          23m          110       long-5786d47bf5-jr5ct.154d90743d10c944   Pod                      Normal    NotTriggerScaleUp   cluster-autoscaler      pod didn't trigger scale-up (it wouldn't fit if a new node is added): Insufficient cpu (1)```

It's accompanied by scheduler event for this pod:

```3m          23m          420       long-5786d47bf5-jr5ct.154d9071973d7470   Pod                      Warning   FailedScheduling    default-scheduler       0/3 nodes are available: 3 Insufficient cpu.```

Given that the scheduler's message format is now "<number of affected nodes> <failure reason>", we should probably use this as well. I'll change this.

File: cluster-autoscaler/core/utils.go (hunk 2 above)
Comment (aleksandra-malinowska) [reply cluster-autoscaler/core/utils.go]: I thought having two fields is the lesser evil (i.e. we won't have to use artificial predicate error), but I can change this if you think it's better.

File: cluster-autoscaler/core/scale_up.go [hunk 5]
```diff
@@ -475,36 +498,54 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
...
+				aggregated[reason]++
+			}
+		}
+		for msg, count := range aggregated {
+			messages = append(messages, fmt.Sprintf("%s (%d)", msg, count))
 		}
+		if skipped > 0 {
+			messages = append(messages, fmt.Sprintf("can't increase node group size (%d)", skipped))
+		}
+		remaining[pod] = strings.Join(messages, ", ")
```
MaciekPytel: As far as I can see we only do the whole string formatting so we can use it in EventingScaleUpStatusProcessor. Maybe we should move this to a utility function and call it from inside the processor? In general a data structure may be more useful for some custom processor than a string.

aleksandra-malinowska: Agreed, this is one of the points for potential improvement mentioned in the description. 

I can work on it now, but it'll make this PR even bigger (or hacky - e.g. I could add an artificial PredicateError for skipped groups, instead of yet-another-custom-struct).

cluster-autoscaler/core/utils.go
aleksandra-malinowska: Changed back to checking schedulable in each branch, PTAL

cluster-autoscaler/core/utils.go
aleksandra-malinowska: Removed the field as suggested, PTAL

cluster-autoscaler/core/scale_up.go
aleksandra-malinowska: Fixed format to match scheduler's

aleksandra-malinowska: There's still a lot of room for improvement, e.g. we should probably make it clear that we're talking about node groups rejected for scale-up. Ideally there'd be a per-node-group reason, but then the message size would grow unreasonably long in some cases.

MaciekPytel: Thanks for the example outputs. They look good for me. We can always improve formatting later if people find them hard to use, but I think using the same format as scheduler provides a consistent experience.

cluster-autoscaler/core/utils.go
```diff
@@ -112,34 +112,41 @@ func (podMap podSchedulableMap) set(pod *apiv1.Pod, schedulable bool) {
...
-			continue
 		}
-		if nodeName, err := predicateChecker.FitsAny(pod, nodeNameToNodeInfo); err == nil {
-			glogx.V(4).UpTo(loggingQuota).Infof("Pod %s marked as unschedulable can be scheduled on %s. Ignoring in scale up.", pod.Name, nodeName)
-			podSchedulable.set(pod, true)
-		} else {
-			unschedulablePods = append(unschedulablePods, pod)
-			podSchedulable.set(pod, false)
+		// Not found in cache, have to run the predicates.
+		if !found {
```
MaciekPytel: nit: that looks strange to me:
```
if found {
  stuff()
}
if !found {
  otherStuff()
}
```
I liked old continue at the end of first branch. `else` would be pretty normal as well (personally I prefer continue to keep code "flat", but whatever).

cluster-autoscaler/core/utils.go
```diff
@@ -112,34 +112,41 @@ func (podMap podSchedulableMap) set(pod *apiv1.Pod, schedulable bool) {
...
+			nodeName, err := predicateChecker.FitsAny(pod, nodeNameToNodeInfo)
+			// err returned from FitsAny isn't a PredicateError.
+			// Hello, ugly hack. I wish you weren't here.
+			var predicateError *simulator.PredicateError
+			if err != nil {
+				predicateError = simulator.NewPredicateError("FitsAny", err, nil)
+			}
+			podSchedulable.set(pod, predicateError)
+			if err != nil {
```
MaciekPytel: nit: I would merge this with identical if 3 lines above and just push podSchedulable.set to the end of the function.

cluster-autoscaler/core/utils.go
aleksandra-malinowska: Done.

cluster-autoscaler/core/utils.go
aleksandra-malinowska: Done.

MaciekPytel:
/lgtm
Sorry it took that long. Can you squash some of the commits, or is it too painful (I'll re-apply lgtm if you do).

aleksandra-malinowska:
Nothing painful about squashing commits ;) Done.

PR #1188, opened 2018-08-24
Pass NoScaleUpInfo to ScaleUpStatus processor

aleksandra-malinowska:
Follow-up to #1185.

### DIFF
--- a/cluster-autoscaler/core/scale_up.go
+++ b/cluster-autoscaler/core/scale_up.go
@@ -20,7 +20,6 @@ import (
 	"bytes"
 	"fmt"
 	"math"
-	"strings"
 	"time"
 
 	apiv1 "k8s.io/api/core/v1"
@@ -34,7 +33,6 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/metrics"
 	ca_processors "k8s.io/autoscaler/cluster-autoscaler/processors"
 	"k8s.io/autoscaler/cluster-autoscaler/processors/status"
-	"k8s.io/autoscaler/cluster-autoscaler/simulator"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/glogx"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/gpu"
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/core/scale_up.go
```diff
@@ -229,6 +227,19 @@ func getNodeInfoCoresAndMemory(nodeInfo *schedulercache.NodeInfo) (int64, int64)
...
+type skippedReasons struct {
+	message []string
+}
+
+func (sr *skippedReasons) Reasons() []string {
+	return sr.message
+}
+
+var (
+	notReadyReason        = &skippedReasons{[]string{"not ready for scale-up"}}
```
MaciekPytel: We know why the group can be not ready for scale-up, but I feel this doesn't really clarify things for users. Maybe we can come up with a better message? Possibly a few different messages, given that this one is used for a few different things?

cluster-autoscaler/processors/status/eventing_scale_up_processor.go
```diff
@@ -46,3 +47,22 @@ func (p *EventingScaleUpStatusProcessor) Process(context *context.AutoscalingCon
 // CleanUp cleans up the processor's internal structures.
 func (p *EventingScaleUpStatusProcessor) CleanUp() {
 }
+
+func reasonsMessage(noScaleUpInfo NoScaleUpInfo) string {
```
MaciekPytel: This looks useful for other implementations of processor. Can you make it public?

MaciekPytel:
Looks pretty cool. cc: @bskiba WDYT?

cluster-autoscaler/core/scale_up.go
MaciekPytel: Agreed we need some catch-all message for "stuff didn't work, sorry" (aka 500). But the 2 most likely reasons to get this would be:
1. group is unhealthy
2. group is in backoff
We can distinguish those and they sound at least somewhat meaningful to user. WDYT?

aleksandra-malinowska: Added new message for the back-off case, PTAL. This requires extra check (duplicated in IsNodeGroupSafeToScaleUp) to separate it from unhealthy-because-of-sizes case.

MaciekPytel:
/lgtm
@bskiba Are you also ok with this?

PR #1189, opened 2018-08-26
Add batch jobs and cronjobs to Azure cluster-autoscaler

romanlutz:
Jobs and CronJobs were both missing so far which makes the cluster-autoscaler complain that it doesn't have access to "batch".


romanlutz:
This is identical to #1180 but there my CLA check didn't go green. That works here, so I've closed the other one!
@feiskyer

feiskyer:
@romanlutz Thanks.

/lgtm

PR #1191, opened 2018-08-27
[examples] Update resource version

koooge:
Hi. 
Use v1 version in examples. apps/v1 may be just alias of exensions/v1beta1 though.


aleksandra-malinowska:
Thanks for making this PR! Waiting for someone with AWS setup to test it works as expected and LGTM.

PR #1192, opened 2018-08-27
Update README

aleksandra-malinowska:
Add 1.3 / K8s 1.11 to recommended release combinations. Add link to release notes for recent releases. Rephrase a couple other sentences.

### DIFF
--- a/cluster-autoscaler/README.md
+++ b/cluster-autoscaler/README.md
@@ -2,10 +2,10 @@
 
 # Introduction
 
-Cluster Autoscaler is a tool that automatically adjusts the size of the Kubernetes cluster when:
-* there are pods that failed to run in the cluster due to insufficient resources.
-* some nodes in the cluster are so underutilized, for an extended period of time,
-that they can be deleted and their pods will be easily placed on some other, existing nodes.
+Cluster Autoscaler is a tool that automatically adjusts the size of the Kubernetes cluster when one of the following conditions is true:
+* there are pods that failed to run in the cluster due to insufficient
+  resources,
+* there are nodes in the cluster that have been underutilized for an extended period of time and their pods can be placed on other existing nodes.
 
 # FAQ/Documentation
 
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/README.md
```diff
@@ -27,10 +26,13 @@ there is a big chance that it won't work as expected.
 | 1.5.X  | 0.4.X  |
 | 1.4.X  | 0.3.X  |
 
-<sup>*</sup>Cluster Autoscaler 0.5.X is the official version shipped with k8s 1.6. We've done some basic tests using k8s 1.6 / CA 0.6 and we're not aware of any problems with this setup. However, CA internally simulates k8s scheduler and using different versions of scheduler code can lead to subtle issues.
+<sup>*</sup>Cluster Autoscaler 0.5.X is the official version shipped with k8s 1.6. We've done some basic tests using k8s 1.6 / CA 0.6 and we're not aware of any problems with this setup. However, Cluster Autoscaler internally simulates Kubernetes' scheduler and using different versions of scheduler code can lead to subtle issues.
```
MaciekPytel: Recently we've had a few issues created for bugs in versions we no longer support (ex. 0.6). Basic predicates (PodsFitsResouces, PodToleratesNodeTaints, etc) haven't really changed in a long time. In most cases running a newer CA on older cluster is the best solution available to user. Maybe we should explicitly say what versions are not supported and that it is _probably_ ok to use newer CA if they're on old k8s version and run into a bug in CA.

cluster-autoscaler/README.md
```diff
@@ -27,10 +26,13 @@ there is a big chance that it won't work as expected.
...
 
 # Notable changes
 
+For CA 1.1.2 and later, please check [release
```
MaciekPytel: Do we have releases on github for older versions? If so maybe we can completely drop this section. It's not too useful anyway and it's quickly getting outdated. I want to keep this short to avoid tl;dr effect of our FAQ. If we drop this maybe we could add some stuff that users often get confused about (I'd like some buildup on my "things that CA do / doesn't do" slides from deep dive on last kubecon, but I'm open to discussion :)). WDYT?

aleksandra-malinowska: There are some. I didn't read all of them, but they don't seem as extensive as the ones here. I suppose it's a question of how long we want to keep the history here.

cluster-autoscaler/README.md
aleksandra-malinowska: Do we know which versions will break for sure? 

Rephrased the earlier part to reflect that sometimes using a newer version makes sense, PTAL.

PR #1193, opened 2018-08-27
Fix GetClusterSize to return actual size in line with the rest of CSR

towca:
It returned the number of registered nodes, but should return the number
of provisioned nodes instead.


towca:

PR #1197, opened 2018-08-29
Fix bad syntax for struct tag value in prometheus_util.go

hello2mao:
This PR fixes bad syntax for struct tag value in prometheus_util.go.


bskiba:
Thanks for the PR!
Does the current way break for you or is this just a change to follow some recommended way to do this?

hello2mao:
@bskiba According to  [Go conventions](https://golang.org/pkg/reflect/#StructTag),  tag strings are a concatenation of optionally space-separated key:"value" pairs. Each key is a non-empty string consisting of non-control characters other than space (U+0020 ' '), quote (U+0022 '"'), and colon (U+003A ':'). Each value is quoted using U+0022 '"' characters and Go string literal syntax.

bskiba:
Thanks for explaining!
/lgtm

PR #1198, opened 2018-08-30
Clean up eviction admission no VPAs in the cluster 

bskiba:
We need to make sure, that even if there are no VPAs, Eviction Admission will correctly initialize its state.


bskiba:
@kgolab

kgolab:
This breaks tests (TestRunOnceNotingToProcess to be precise).

But maybe it suggests that we should add some CleanUp() method instead of executing LoopInit() here?

bskiba:
Added CleanUp

kgolab:
LGTM now, thanks.

PR #1199, opened 2018-08-30
Check for ready condition not true

aleksandra-malinowska:
Check for ready condition not true instead of just false.

/cc @losipiuk

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -468,9 +468,9 @@ func (csr *ClusterStateRegistry) updateReadinessStats(currentTime time.Time) {
 		current.Registered++
 		if deletetaint.HasToBeDeletedTaint(node) {
 			current.Deleted++
-		} else if isNodeNotStarted(node) && node.CreationTimestamp.Time.Add(MaxNodeStartupTime).Before(currentTime) {
+		} else if stillStarting := isNodeStillStarting(node); stillStarting && node.CreationTimestamp.Time.Add(MaxNodeStartupTime).Before(currentTime) {
 			current.LongNotStarted++
-		} else if isNodeNotStarted(node) {
+		} else if stillStarting {
 			current.NotStarted++
 		} else if ready {
 			current.Ready++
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -773,7 +773,7 @@ func buildScaleDownStatusClusterwide(candidates map[string][]string, lastProbed
 func isNodeNotStarted(node *apiv1.Node) bool {
```
losipiuk: nit: If you are changing this part of code consider removing Not from name of this function.
Maybe call it "isNodeStillStarting" or "isNodeStarted" negating the semantics.

cluster-autoscaler/clusterstate/clusterstate_test.go
```diff
@@ -722,3 +722,35 @@ func TestGetClusterSize(t *testing.T) {
 	assert.Equal(t, 3, currentSize)
 	assert.Equal(t, 30, targetSize)
 }
+
+func TestIsNodeNotStarted(t *testing.T) {
+	testCases := []struct {
+		desc         string
+		condition    apiv1.NodeConditionType
+		status       apiv1.ConditionStatus
+		recentResult bool
```
losipiuk: name this "expectedResult"

cluster-autoscaler/clusterstate/clusterstate.go
aleksandra-malinowska: Done

cluster-autoscaler/clusterstate/clusterstate_test.go
```diff
@@ -722,3 +722,35 @@ func TestGetClusterSize(t *testing.T) {
 	assert.Equal(t, 3, currentSize)
 	assert.Equal(t, 30, targetSize)
 }
+
+func TestIsNodeNotStarted(t *testing.T) {
```
losipiuk: rename the the test method to match renamed name :)

losipiuk:
LGTM

cluster-autoscaler/clusterstate/clusterstate_test.go
aleksandra-malinowska: Done, thanks for catching this

Issue #1181, opened 2018-08-23
VPA object should have condition to signalize no matching pods 


bskiba:
When the label selector on VPA doesn't match any pods in the cluster, the recomender should set a condition on VPA object to indicate this as it will cause us not to provide recommendations.

phsiao:
@bskiba Happy to take this one.  Would something like

```
  status:
    conditions:
    - lastTransitionTime: 2018-08-30T23:14:13Z
      status: "False"
      reason: "NoMatchingPod"
      type: RecommendationProvided
```

be what you are looking for?   Or a separate condition?

bskiba:
I think this looks very reasonable. 👍

phsiao:
Right now, the status does not contain a conditions map like below if there is no matching pod

```
  status:
    recommendation: {}
```

it is easy to fix that by also setting things when `len(vpa.Recommendation.ContainerRecommendations) == 0`

to become

```
  status:
    conditions:
    - lastTransitionTime: 2018-08-31T19:34:44Z
      reason: NoMatchingPod
      status: "False"
      type: RecommendationProvided
    recommendation: {}
```

and would become normal again when pods are added to match

```
  status:
    conditions:
    - lastTransitionTime: 2018-08-31T19:40:01Z
      status: "True"
      type: RecommendationProvided
    recommendation:
      containerRecommendations:
      - containerName: hamster
        lowerBound:
          cpu: 190m
          memory: "104857600"
        target:
          cpu: 564m
          memory: "104857600"
        upperBound:
          cpu: 406644m
          memory: 4145750k
```

I still need to test for a couple of other cases and see how to address them, specifically,

1. when there are matching pods, but the pods are in crashloop/error/error pulling image so no stats to report, recommender would still report there are matching pods but no recommendations
2. when the pods are deleted, because of the internal model and vpa, it will continue to recommend stale data.

phsiao:
I did run into a weird case that I have not had a chance to look into while working on this.

When I created a VPA first without matching pods, then add pods later, recommender does not make recommendations for a long time as if the pods don't exist.  Restarting recommender fixes that.

bskiba:
Hm, I'm testing this right now from the current master branch, and I don't see this. I first created VPA pointing at a non-existing deployment. No recommendation was generated. Then created the deployment. After ~2 minutes got the recommendation.

phsiao:
[...]

PR #1201, opened 2018-08-31
WIP: VPA object status reflects no matching pods

phsiao:
Working on solution for #1181


bskiba:
Generally, I'm not sold on this approach. This does not signalize a config error: "the selector provided matches no pods", but rather "There is no recommendation for any containers" which can be caused by multiple situations, as you mention yourself (crashlooping containers, metric collection system fails to provides samples).
I would like to have a clear and actionable message for the user: "there is probably some misconfiguration, you have a VPA that is not controlling any pods at all, are you sure this is what you want?"

phsiao:
Agreed.  Will take a look and see how to get more diagnosing information.

mwielgus:
No activity for 1 month - closing.
Please reopen the PR once you have the updated version.

PR #1203, opened 2018-09-03
update resource version

mikeweiwei:
update resource version


feiskyer:
@mikeweiwei Thanks for the updating.

PR #1205, opened 2018-09-03
Add caching node locations in GKE cloud provider

aleksandra-malinowska:
Refresh node locations when refreshing other GKE resources. Add a method for getting it from cloud provider. Use zone from system labels when constructing new node group in GKE cloud provider.

### DIFF
--- a/cluster-autoscaler/cloudprovider/gke/gke_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/gke/gke_cloud_provider.go
@@ -204,6 +204,11 @@ func (gke *GkeCloudProvider) GetClusterInfo() (projectId, location, clusterName 
 	return gke.gkeManager.GetProjectId(), gke.gkeManager.GetLocation(), gke.gkeManager.GetClusterName()
 }
 
+// GetNodeLocations returns the list of zones in which the cluster has nodes.
+func (gke *GkeCloudProvider) GetNodeLocations() []string {
+	return gke.gkeManager.GetNodeLocations()
+}
+
 // MigSpec contains information about what machines in a MIG look like.
 type MigSpec struct {
 	MachineType    string
--- a/cluster-autoscaler/cloudprovider/gke/gke_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/gke/gke_cloud_provider.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/gce"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/gpu"
+	kubeletapis "k8s.io/kubernetes/pkg/kubelet/apis"
 	schedulercache "k8s.io/kubernetes/pkg/scheduler/cache"
 )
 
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/cloudprovider/gke/gke_manager.go
```diff
@@ -106,11 +107,37 @@ type GkeManager interface {
...
+
+	cache.nodeLocations = locations
+}
+
+func (cache *gkeConfigurationCache) getNodeLocations() []string {
+	cache.Lock()
+	defer cache.Unlock()
+
+	locations := make([]string, len(cache.nodeLocations))
+	for i, location := range cache.nodeLocations {
```
losipiuk: use `copy(locations, cache.nodeLocations`) instead a loop

cluster-autoscaler/cloudprovider/gke/gke_manager.go
```diff
@@ -106,11 +107,37 @@ type GkeManager interface {
...
+type gkeConfigurationCache struct {
+	sync.Mutex
+	nodeLocations []string
+}
+
+func (cache *gkeConfigurationCache) setNodeLocations(locations []string) {
+	cache.Lock()
+	defer cache.Unlock()
+
+	cache.nodeLocations = locations
```
losipiuk: I would suggest copying slice here. That way we ensure that internal array is not modified even if caller by mistake changes array which he passed as parameter to `setNodeLocations`

losipiuk:
`Add caching node locations in GKE cloud provider` lgtm module minor comments :)

cluster-autoscaler/cloudprovider/gke/gke_manager.go
aleksandra-malinowska: Done

cluster-autoscaler/cloudprovider/gke/gke_manager.go
aleksandra-malinowska: Done

cluster-autoscaler/cloudprovider/gke/gke_manager.go
```diff
@@ -467,6 +508,10 @@ func (m *gkeManagerImpl) forceRefresh() error {
 			glog.Errorf("Failed to fetch node pools: %v", err)
 			return err
 		}
+		if err := m.refreshLocations(); err != nil {
```
MaciekPytel: Unless I'm missing something with this implementation we do 3 identical API calls in a row when calling forceRefresh() (a separate GET cluster call for nodepools, locations and limits). Can we somehow avoid that? Maybe cache cluster in GKEService or something? I'm fine accepting this and improving in follow up PR if you prefer it that way.

aleksandra-malinowska: Agreed, repetitive API calls are definitely something we can optimize. I'd prefer to do it in a separate PR since it's unrelated/bigger change. In the meantime, moved this call to NAP case.

cluster-autoscaler/cloudprovider/gke/gke_cloud_provider.go
```diff
@@ -121,10 +122,10 @@ func (gke *GkeCloudProvider) GetAvailableMachineTypes() ([]string, error) {
 func (gke *GkeCloudProvider) NewNodeGroup(machineType string, labels map[string]string, systemLabels map[string]string,
 	taints []apiv1.Taint, extraResources map[string]resource.Quantity) (cloudprovider.NodeGroup, error) {
 	nodePoolName := fmt.Sprintf("%s-%s-%d", nodeAutoprovisioningPrefix, machineType, time.Now().Unix())
-	// TODO(aleksandra-malinowska): GkeManager's location will be a region
-	// for regional clusters. We should support regional clusters by looking at
-	// node locations instead.
-	zone := gke.gkeManager.GetLocation()
+	zone, found := systemLabels[kubeletapis.LabelZoneFailureDomain]
```
MaciekPytel: Huh, I just realised this will break all our tests until we start passing zone label. Maybe let's default it for now if it's missing?

aleksandra-malinowska: As discussed offline, this isn't a blocker

PR #1212, opened 2018-09-04
Reduce GKE API calls

aleksandra-malinowska:
Replace 2 GET cluster and 1 LIST node pools call with a single GET cluster on forceRefresh().

This also changes LIST node pools call to GET cluster call after creating/deleting node pool. I don't think it should have much performance impact - the object is larger, but it counts as a full refresh, so it reduces the total number of calls. @MaciekPytel WDYT?

### DIFF
--- a/cluster-autoscaler/cloudprovider/gke/autoscaling_gke_client_v1beta1.go
+++ b/cluster-autoscaler/cloudprovider/gke/autoscaling_gke_client_v1beta1.go
@@ -44,7 +44,6 @@ type autoscalingGkeClientV1beta1 struct {
 	gkeBetaService *gke_api_beta.Service
 
 	clusterPath   string
-	nodePoolsPath string
 	nodePoolPath  string
 	operationPath string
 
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/cloudprovider/gke/autoscaling_gke_client_v1beta1.go
```diff
@@ -93,22 +91,16 @@ func (m *autoscalingGkeClientV1beta1) FetchNodePools() ([]NodePool, error) {
...
-func (m *autoscalingGkeClientV1beta1) FetchResourceLimits() (*cloudprovider.ResourceLimiter, error) {
-	cluster, err := m.gkeBetaService.Projects.Locations.Clusters.Get(m.clusterPath).Do()
-	if err != nil {
-		return nil, err
-	}
+func buildResourceLimiter(cluster *gke_api_beta.Cluster) *cloudprovider.ResourceLimiter {
 	if cluster.Autoscaling == nil {
-		glog.Warningf("FetchResourceLimits called without autoscaling limits set")
-		return nil, nil
+		return nil
```
MaciekPytel: I don't really expect to get here, so maybe we should keep the warning?

aleksandra-malinowska: Done

PR #1214, opened 2018-09-05
Add a method for inserting custom test node group to test cloud provider

aleksandra-malinowska:
Add a method for inserting real node group to test cloud provider.


MaciekPytel:
I really dislike the implication in PR description. We really shouldn't use real node group implementations in unittests. If it's just so you can inject a more complex mock than I'm fine with this.

aleksandra-malinowska:
Fair enough. Refactored to allow building test node group, which can then be used as base for a more complex testing struct.

PR #1216, opened 2018-09-05
Update godeps

losipiuk:


losipiuk:
@feiskyer  please take a look at change when I update the azure compute client version. Should the azure cloud provider still work.

aleksandra-malinowska:
Please remove changes to addon resizer and vertical pod autoscaler vendor/

losipiuk:
> Please remove changes to addon resizer and vertical pod autoscaler vendor/

Done - I was not aware `fix_godeps.sh` would remove vendor from other subprojects of c-a (makes sense if you give it a thought)  and I squashed updated godeps commit with `no_vendor` commits. My bad.

PR #1219, opened 2018-09-06
Add client-go metrics to VPA

kgolab:
Import k/k/pkg/client/metrics/prometheus to register rest_client_request_* metrics.

Commits are separated to split the real change from godeps update.


kgolab:
@schylek , could please review?

aleksandra-malinowska:
Merging based on LGTM from @schylek.

PR #1221, opened 2018-09-06
update ca version

mikeweiwei:
update ca version


aleksandra-malinowska:
1.3.2-beta.2 already exists and was built from another branch, with very different commit history over the past two months. We shouldn't use this version on master branch.

PR #1222, opened 2018-09-06
Add client-go metrics to CA

kgolab:
Import k/k/pkg/client/metrics/prometheus to register rest_client_request_* metrics.

Commits are separated to split the real change from godeps update.


MaciekPytel:
@aleksandra-malinowska can you take this one?

Issue #734, opened 2018-03-21
Cluster-autoscaler support for OpenStack cloud provider


dklyle:
There are numerous potential ways to support cluster autoscaling in OpenStack. Services like Heat and Senlin are not guaranteed to be support in an OpenStack deployment. So base support using only core OpenStack services should be targeted. The eventual path could be to optionally support using other services, but that would be outside the scope of this issue, other than driving the implementation to be abstract around ties to underlying services.

I found #230, but it has been closed as stale.

varuntalus:
@dklyle 

We are also struggling with cluster auto-scalar. Our requirement is to prov\deprov nodes (VMs) automatically based on the usage of the resources on each of the node.

We are planning to have a "VM Farm" having n number of VMs which can be utilized by any k8s cluster (in our on-prim env) looking for a node. This VM farm would be maintained\managed by OpenStack.

Currently we do not find any implementation\solution where K8s CA is able to communicate properly with OpenStack and is able to do node prov\deprov automatically.

Could you please advise by when you would be able to release the solution for the same?

Thanks,
Varun T

dklyle:
@varuntalus This is very much a work in progress. I'm working on code in spurts. I will update here when I have enough progress to be useful. But I also wanted to solicit input of use cases and any discussion of desired criteria to help shape the implementation.

timstoop:
We are very interested in getting this to work. @dklyle would you be able to provide an update for this? Anything we can do to help you move forward with this?

ricolin:
@zetaab appears I have to fix gophercloud as well (already some fix merged,but needs more), will also implement in kops once we get full test here.

ricolin:
@timstoop Since this has been pending for long time, I will take over if that's fine with all. Any help or comments are welcome

ricolin:
Current patch still missing Test, and some implementation, I need to fix some issue in gophercloud (OpenStack Golang SDK) first before I can push all implement and tests up.

Rajat-0:
@ricolin  I am also interested in implementing this( As discussed on slack), let me know how I can help.

chaosaffe:
[...]

PR #1226, opened 2018-09-06
[WIP]Support openstack cluster autoscaler

ricolin:
WIP for #734

### DIFF
--- a/cluster-autoscaler/Godeps/Godeps.json
+++ b/cluster-autoscaler/Godeps/Godeps.json
@@ -1001,123 +1001,131 @@
 		},
 		{
 			"ImportPath": "github.com/gophercloud/gophercloud",
-			"Rev": "781450b3c4fcb4f5182bcc5133adb4b2e4a09d1d"
+			"Rev": "13af1c2a38cfbc60a7a13062338838301035d91d"
 		},
 		{
 			"ImportPath": "github.com/gophercloud/gophercloud/openstack",
-			"Rev": "781450b3c4fcb4f5182bcc5133adb4b2e4a09d1d"
+			"Rev": "13af1c2a38cfbc60a7a13062338838301035d91d"
 		},
 		{
 			"ImportPath": "github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/volumeactions",
-			"Rev": "781450b3c4fcb4f5182bcc5133adb4b2e4a09d1d"
+			"Rev": "13af1c2a38cfbc60a7a13062338838301035d91d"
 		},
 		{
 			"ImportPath": "github.com/gophercloud/gophercloud/openstack/blockstorage/v1/volumes",
-			"Rev": "781450b3c4fcb4f5182bcc5133adb4b2e4a09d1d"
+			"Rev": "13af1c2a38cfbc60a7a13062338838301035d91d"
 		},
 		{
 			"ImportPath": "github.com/gophercloud/gophercloud/openstack/blockstorage/v2/volumes",
-			"Rev": "781450b3c4fcb4f5182bcc5133adb4b2e4a09d1d"
+			"Rev": "13af1c2a38cfbc60a7a13062338838301035d91d"
 		},
 		{
 			"ImportPath": "github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes",
-			"Rev": "781450b3c4fcb4f5182bcc5133adb4b2e4a09d1d"
+			"Rev": "13af1c2a38cfbc60a7a13062338838301035d91d"
 		},
 		{
 			"ImportPath": "github.com/gophercloud/gophercloud/openstack/common/extensions",
-			"Rev": "781450b3c4fcb4f5182bcc5133adb4b2e4a09d1d"
+			"Rev": "13af1c2a38cfbc60a7a13062338838301035d91d"
 		},
 		{
 			"ImportPath": "github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/attachinterfaces",
-			"Rev": "781450b3c4fcb4f5182bcc5133adb4b2e4a09d1d"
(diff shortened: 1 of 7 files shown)


zetaab:
Wow, great job :)

aleksandra-malinowska:
This looks promising :) Great job so far! A note on unit tests, seeing as this is based on GCE cloud provider: I don't recommend following the same legacy approach used there (mock HTTP server sending JSONs = long, unreadable test scenarios, hard to verify for correctness.) It'd be much easier to use a fake client, and we're planning to refactor them this way soon-ish.

xmik:
Thank you @ricolin for implementing this! Does your implementation need OpenStack Heat? I am not sure after skimming through this PR.

MrDienns:
Lovely to see that someone is working on this! We'd love to get started on autoscaling our OpenStack cloud, so I can't wait to see this PR merged. Keep up the great work @ricolin!

HashFail:
@ricolin are you still working on this? I can take over working on this if you're not able to work on it right now.

ricolin:
@HashFail I'm still working on this, but I would like to have more help together:)
I have left some information we got now in [1]. feel free to take a look. 
[1] https://github.com/kubernetes/autoscaler/issues/734

tghartland:
I have been working on an openstack cloud provider as well. The basic functionality is done, scaling up with magnum and scaling down by interacting directly with heat to remove specific nodes.
It's a bit of a mess at the moment, but I'm working on a WIP pull request [here](https://github.com/cernops/autoscaler/pull/3) which I want to upstream when it's ready, perhaps by combining with this PR.

I'm allowing only a single nodegroup for now until nodegroups are implemented in magnum, and I'm told there will be a way to delete specific nodes through magnum instead of having to use heat sometime in the future.

I'm also strictly enforcing that the autoscaler can only scale the cluster if the cluster has finished the previous update operation. To speed things up a bit I've batched deletions together where possible, since the autoscaler calls DeleteNodes only ever with a single node even when there are multiple that need removing at the same time.

I don't yet have any caching for the nodegroup, but it looks like that has already been covered in this PR.

scott-chaney:
@tghartland Awesome to see this moving along! I have a working-condition similar approach. I am waiting on my organization's contribution approval before I can release anything but I would love to help contribute through discussion until then.

My approach is based on clusters that Magnum creates, but does all of the scaling through Heat due to Magnum's current inability to scale out specific nodes, as you mentioned.  I am curious as to why your implementation would prefer to use the Magnum API?

> I'm also strictly enforcing that the autoscaler can only scale the cluster if the cluster has finished the previous update operation. To speed things up a bit I've batched deletions together where possible, since the autoscaler calls DeleteNodes only ever with a single node even when there are multiple that need removing at the same time.

For me, solving this was a little awkward due to the declarative nature of Heat vs the expected imperative behavior of the cloud provider from the autoscaler's perspective.  I ended up caching any update parameters to allow for subsequent updates to make changes relative to the previous update request.  This prevents having to do any batching on the cloud provider side, and allows for subsequent updates to just overwrite the previous one (maybe this is similar to how you do the batching?).

Something that remains to be solved in my implementation is the mapping of Heat Stack update parameters to the autoscaler code.  The coupling across projects doesn't feel great and has motivated us to look into driving the scaling operations through Senlin (i.e. Magnum would deploy a Senlin cluster through Heat).  Until then, my plan is to just provide a piece of configuration to map the node group scaling parameters in Heat to the autoscaler calls.  I would love to hear some ideas on how others are looking at solving this.

scott-chaney:
@tghartland 

>  found that when scaling the heat stack directly, the node_count reported by magnum wasn't updated to reflect the new value.

This is a fair point, we currently just let magnum's numbers get out of sync, which isn't great, but my thinking was that it would be easier to implement a "watch" on the stack in magnum, rather than implement the update path.  Based on some of the discussion around this, it sounds like there may be merit to have both approaches as well as a Senlin approach and an approach that uses the more core services directly (nova, neutron, cinder, etc) to allow for Openstack environments that do not run Heat to take advantage of the autoscaler.

> As for your last paragraph, I don't quite understand what you're trying to say. Do you mean that the autoscaler being designed for aws/gke makes it difficult to adapt to openstack?

I was more getting at this point in the 2nd to last paragraph but the final paragraph is related. The node group naming conventions are specific to magnum's templates resources (minions_to_remove, number_of_minions).  My concern is that those strings are not guaranteed for a given Heat stack, or even a given Magnum template so a mapping of the update parameters to node group in the cluster-autoscaler allows for this sort of flexibility.  For example, as multiple node groups become supported, there will be a pattern for each node groups' update parameters (number_of_minons_m4.xlarge or something) and it would be good to support dynamic naming patterns to avoid code changes in the openstack cloud provider autoscaler code if possible.  I made the comment about Senlin because Senlin allows for action queuing as well as a node deletion method that only requires the node ID and does not need to know anything about the heat stack/magnum template parameters, avoiding the coupling that we have with the current Heat::ResourceGroups.  I hope that makes sense.

tghartland:
@scott-chaney 
> For example, as multiple node groups become supported, there will be a pattern for each node groups' update parameters (number_of_minons_m4.xlarge or something)

From what I understand of nodegroups,  different VM flavours will be separated into different groups so this shouldn't be a problem. That's what the [nodegroup design spec](https://github.com/openstack/magnum-specs/blob/master/specs/stein/magnum-nodegroups.rst) seems to be going for anyway, so the names needed to interact with each nodegroup should be identical.
I don't know how this would work if you did want to interact directly with heat instead of the nodegoups themselves, so maybe in that situation you would need some way to figure out which stack values belong to which nodegroup.

mwielgus:
No activity for almost 2 months, still WIP, closing. Please reopen once there is a chance to get out of WIP state.

ricolin:
oops https://github.com/kubernetes/autoscaler/pull/1691

PR #1231, opened 2018-09-07
VPA histogram percentile should return end of the bucket instead of the middle point

schylek:


vertical-pod-autoscaler/pkg/recommender/util/histogram.go
```diff
@@ -164,9 +164,8 @@ func (h *histogram) Percentile(percentile float64) float64 {
 	}
 	bucketStart := h.options.GetBucketStart(bucket)
```
mwielgus: this variable is not needed.

mwielgus:
1 comment, otherwise lgtm

vertical-pod-autoscaler/pkg/recommender/util/histogram.go
schylek: Done.

kgolab:
LGTM

PR #1233, opened 2018-09-07
Add a scale down status processor, refactor so that there's more scal…

towca:
…e down info available to it

### DIFF
--- a/cluster-autoscaler/core/scale_down.go
+++ b/cluster-autoscaler/core/scale_down.go
@@ -43,23 +43,8 @@ import (
 
 	"github.com/golang/glog"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/autoscaler/cluster-autoscaler/processors/status"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/gpu"
-)
-
-// ScaleDownResult represents the state of scale down.
-type ScaleDownResult int
-
-const (
-	// ScaleDownError - scale down finished with error.
-	ScaleDownError ScaleDownResult = iota
-	// ScaleDownNoUnneeded - no unneeded nodes and no errors.
-	ScaleDownNoUnneeded
-	// ScaleDownNoNodeDeleted - unneeded nodes present but not available for deletion.
-	ScaleDownNoNodeDeleted
-	// ScaleDownNodeDeleted - a node was deleted.
-	ScaleDownNodeDeleted
-	// ScaleDownNodeDeleteStarted - a node deletion process was started.
-	ScaleDownNodeDeleteStarted
 )
 
 const (
@@ -85,6 +70,10 @@ const (
 type NodeDeleteStatus struct {
 	sync.Mutex
 	deleteInProgress bool
+	// A map of node delete results by node name. It contains nil if the delete was successful and an error otherwise.
+	// It's being constantly drained into ScaleDownStatus objects in order to notify the ScaleDownStatusProcessor that
+	// the node drain has ended or that an error occurred during the deletion process.
+	nodeDeleteResults map[string]error
 }
 
 // IsDeleteInProgress returns true if a node is being deleted.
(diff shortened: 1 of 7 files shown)


towca:

cluster-autoscaler/core/scale_down.go
```diff
@@ -697,6 +720,7 @@ func (sd *ScaleDown) TryToScaleDown(allNodes []*apiv1.Node, pods []*apiv1.Pod, p
 		// Finishing the delete process once this goroutine is over.
 		defer sd.nodeDeleteStatus.SetDeleteInProgress(false)
 		err := sd.deleteNode(toRemove.Node, toRemove.PodsToReschedule)
+		sd.nodeDeleteStatus.AddNodeDeleteError(toRemove.Node.Name, err)
```
aleksandra-malinowska: Do we still want to do this if err == nil?

aleksandra-malinowska: Can we name it sth like result instead, then?

Also, is there any reason not to defer this call? That way, waiting for lock won't stop logging error and updating metrics.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -334,17 +335,23 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
 				glog.Errorf("Failed to scale down: %v", err)
 				a.lastScaleDownFailTime = currentTime
 				return typedErr
 			}
-			if result == ScaleDownNodeDeleted {
+			if scaleDownStatus.Result == status.ScaleDownNodeDeleted {
 				a.lastScaleDownDeleteTime = currentTime
 			}
+			if scaleDownStatus.Result == status.ScaleDownNodeDeleted || scaleDownStatus.Result == status.ScaleDownNodeDeleteStarted {
+				a.clusterStateRegistry.Recalculate()
```
aleksandra-malinowska: Do we also need to recalculate when we just started draining the node? At this point, we'll be evicting pods, but target size shouldn't yet be affected (unless the drain already completed, which is unlikely)

cluster-autoscaler/core/scale_down.go
```diff
@@ -554,9 +559,23 @@ func (sd *ScaleDown) chooseCandidates(nodes []*apiv1.Node) ([]*apiv1.Node, []*ap
...
+		})
+	}
+	return result
+}
+
+// TryToScaleDown tries to scale down the cluster. It returns a result inside a ScaleDownStatus indicating if any node was
 // removed and error if such occurred.
-func (sd *ScaleDown) TryToScaleDown(allNodes []*apiv1.Node, pods []*apiv1.Pod, pdbs []*policyv1.PodDisruptionBudget, currentTime time.Time) (ScaleDownResult, errors.AutoscalerError) {
+func (sd *ScaleDown) TryToScaleDown(allNodes []*apiv1.Node, pods []*apiv1.Pod, pdbs []*policyv1.PodDisruptionBudget, currentTime time.Time) (*status.ScaleDownStatus, errors.AutoscalerError) {
+	scaleDownStatus := &status.ScaleDownStatus{NodeDeleteResults: sd.nodeDeleteStatus.DrainNodeDeleteResults()}
```
aleksandra-malinowska: I sort of agree with the idea of draining node delete results only at the beginning of scale-down (i.e., when we know we'll be processing a scale down status), but I was wondering what will happen if:
- we start draining a node,
- a schedulable pod appears (i.e., schedulable in simulation, but not schedulable according to scheduler - it happens from time to time), or there's just a large churn of pods for a long while,
- node drain completes, but we don't enter TryToScaleDown because of the above condition for ~hours, so we don't process the results for this time. I assume it's up to ScaleDownStatusProcessor to know the deletion results it was expecting are stale, but it seems a pity it can't access them even though they've been ready this whole time.

Or:
- we start draining a node,
- it completes,
- next run of TryToScaleDown returns an error, and we exit from RunOnce without running ScaleDownStatus through ScaleDownStatusProcessor. However, the deletion results are lost without ever being processed.

cluster-autoscaler/core/static_autoscaler.go
towca: That's true, I'll change it

cluster-autoscaler/core/scale_down.go
towca: I've changed the processor invocation in RunOnce() to deal with the second scenario, but the first one looks out of scope of this PR.

cluster-autoscaler/core/static_autoscaler.go
towca: Done

cluster-autoscaler/processors/status/scale_down_status_processor.go
```diff
@@ -0,0 +1,77 @@
...
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
+	"k8s.io/autoscaler/cluster-autoscaler/context"
+	"k8s.io/autoscaler/cluster-autoscaler/simulator"
+)
+
+// ScaleDownStatus represents the state of scale down.
+type ScaleDownStatus struct {
+	Result            ScaleDownResult
+	ScaledDownNodes   []*ScaleDownNode
+	NodeDeleteResults map[string]error
```
aleksandra-malinowska: Given that nil is a meaningful value for success and a default value for non-existent key at the same time, I wonder if this is what we want to have here long-term. @MaciekPytel WDYT?

MaciekPytel: Discussed with @towca, I think it's reasonable to explicitly check for whether key is in map or not.

PR #1234, opened 2018-09-09
add AWS t3 instance types

delgod:


delgod:
> It may take a couple minutes for the CLA signature to be fully registered; after that, please reply here with a new comment and we'll verify. 
Hi @k8s-ci-robot, I have signed CLA a few minutes ago, please verify.

hikhvar:
Any plans to backport this patch to 1.2.x? Since EKS is currently on kubernetes 1.10, we can not use the cluster autoscaler together with t3.* instances. Doing so results in nil pointer panics.

delgod:
@aleksandra-malinowska, give me know If it is needed to prepare PR to another branch

PR #1235, opened 2018-09-09
fix 3 spell

sunlintong:
maybe I'm too boring this weekend :)


vertical-pod-autoscaler/README.md
```diff
@@ -87,7 +87,7 @@ There are three modes in which *VPAs* operate:
   the `"Auto"` mode.
 * `"Recreate"`: VPA assigns resource requests on pod creation as well as updates
   them on existing pods by evicting them when the requested resources differ significantly
-  from the new recommendation (respecting the Pod Distruption Budget, if defined).
+  from the new recommendation (respecting the Pod Distribution Budget, if defined).
```
kgolab: This should be fixed to Disruption.

sunlintong: That's ok, I've fix.

PR #1237, opened 2018-09-10
Remove unused test code

aleksandra-malinowska:
Remove no longer used JSON responses.


losipiuk:
Sth is wrong with Travis. Seems intermittent.

PR #1241, opened 2018-09-12
Regenerate godeps for VPA to drop unnecessary dependencies.

kgolab:
Thanks to updates in upstream grpc some tools (lint, misspell, staticcheck) are no longer required as transitive dependencies.

@bskiba , please have a look.


bskiba:
Awesome, thank you!
/lgtm

PR #1244, opened 2018-09-13
Call CheckPodsSchedulableOnNode in scale_up.go via caching layer

losipiuk:


cluster-autoscaler/core/scale_up.go
```diff
@@ -367,27 +370,36 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 			Pods:      make([]*apiv1.Pod, 0),
 		}
 
-		schedulableOnNode := CheckPodsSchedulableOnNode(context, unschedulablePods, nodeGroup.Id(), nodeInfo)
-		for pod, err := range schedulableOnNode {
-			if err != nil {
+		// add list of pods which pass predicates to option
+		podsPassing, err := getPodsPassingPredicates(nodeGroup.Id())
+		if err != nil {
+			glog.V(4).Infof("Skipping node group %s; cannot compute pods passing predicates", nodeGroup.Id())
```
aleksandra-malinowska: skippedNodeGroups[nodeGroup.Id()] = ...reason

cluster-autoscaler/core/scale_up.go
```diff
@@ -367,27 +370,36 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
...
+
+		// mark that there is a scheduling option for pods which can be scheduled to node from currently analyzed node group
+		for _, pod := range podsPassing {
+			delete(podsRemainUnschedulable, pod)
+		}
+
+		// update information why we cannot schedule pods for which we did not find a working extension option so far
+		podsNotPassing, err := getPodsNotPassingPredicates(nodeGroup.Id())
+		if err != nil {
+			glog.V(4).Infof("Skipping node group %s; cannot compute pods not passing predicates", nodeGroup.Id())
```
aleksandra-malinowska: skippedNodeGroups[nodeGroup.Id()] = ...reason

cluster-autoscaler/core/scale_up.go
```diff
@@ -367,27 +370,36 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
...
+		}
+
+		// mark that there is a scheduling option for pods which can be scheduled to node from currently analyzed node group
+		for _, pod := range podsPassing {
+			delete(podsRemainUnschedulable, pod)
+		}
+
+		// update information why we cannot schedule pods for which we did not find a working extension option so far
+		podsNotPassing, err := getPodsNotPassingPredicates(nodeGroup.Id())
+		if err != nil {
```
aleksandra-malinowska: Since we exit on this error without saving expansion option for this node group, we should do it before deleting pod from "podsRemainUnschedulable".

cluster-autoscaler/core/scale_up.go
```diff
@@ -524,6 +535,80 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
...
+	nodeInfos map[string]*schedulercache.NodeInfo) podsPredicatePassingCheckFunctions {
+
+	podsPassingPredicatesCache := make(map[string][]*apiv1.Pod)
+	podsNotPassingPredicatesCache := make(map[string]map[*apiv1.Pod]status.Reasons)
+	errorsCache := make(map[string]error)
+
+	computeCaches := func(nodeGroupId string) {
+		nodeInfo, found := nodeInfos[nodeGroupId]
+		if !found {
+			errorsCache[nodeGroupId] = errors.NewAutoscalerError(errors.InternalError, "NodeInfo not found for node group %v", nodeGroupId)
```
aleksandra-malinowska: Shouldn't we return early on this error?

cluster-autoscaler/core/scale_up.go
```diff
@@ -524,6 +535,80 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
...
+			_, errorFound := errorsCache[nodeGroupId]
+
+			if !passingFound && !errorFound {
+				computeCaches(nodeGroupId)
+			}
+			pods, found := podsPassingPredicatesCache[nodeGroupId]
+			if found {
+				return pods, nil
+			}
+			err, found := errorsCache[nodeGroupId]
```
aleksandra-malinowska: Just in case (see the early exit comment above), I'd prefer it if we checked errorsCache first. This way, if there's an error we'll always find it, even if we accidentally set other entries for the node group. Especially given that we may e.g. want to modify cache entries for autoprovisioned node groups in the future.

cluster-autoscaler/core/scale_up.go
```diff
@@ -444,11 +456,10 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 			if err != nil {
 				return nil, err
 			}
+
 			// Node group id may change when we create node group and we need to update
 			// our data structures.
 			if oldId != bestOption.NodeGroup.Id() {
-				podsPassingPredicates[bestOption.NodeGroup.Id()] = podsPassingPredicates[oldId]
-				delete(podsPassingPredicates, oldId)
```
aleksandra-malinowska: IIUC, this means that for an autoprovisioned node group created with different id, we'll:
- run all predicates again (call computeCache(newId)),
- use the same nodeInfo object as in simulation.

Is this WAI? I'm OK with either assuming the result of simulation is fine, or with running predicates again on a node info built from actual instance template, but right now it seems we'll be running the same simulation twice for selected node group.

cluster-autoscaler/core/scale_up.go
losipiuk: Makes sense. Updated.

losipiuk:
Oh - new comments on the way. :) looking at those.

cluster-autoscaler/core/scale_up.go
losipiuk: Running same simulation was my intention. IIUC the previous code was just moving `podsPassingPredicates[oldId]` to `podsPassingPredicates[newId]`. I wanted to avoid explicit cache modifications from external code, at cost of one extra computation.

aleksandra-malinowska: Ack, let's leave it for now then.

aleksandra-malinowska: True, but we assume this in so many places that fixing it would probably require rethinking this whole file. Accessor won't help with e.g. cache invalidation - ideally when we replace or remove a nodeInfo for a node group, we should clean all results (errors, pods passing, pods not passing etc.) for this group.

losipiuk: > True, but we assume this in so many places that fixing it would probably require rethinking this whole file.

Well - we mostly pass it around as a parameter. Which IMO is fine. The difference with caching is that it is stored within closure returned from `getPodsPredicatePassingCheckFunctions`. This is somewhat different and less readable IMO.

> Accessor won't help with e.g. cache invalidation - ideally when we replace or remove a nodeInfo for a node group, we should clean all results (errors, pods passing, pods not passing etc.) for this group.

This cache is rebuilt every iteration. I do not see why we would like to invalidate any entries in it really.

aleksandra-malinowska: > This is somewhat different and less readable IMO.

Agreed, but I don't think an accessor (= same logic, but yet another layer of abstraction to follow) would help much with it. I was thinking about encapsulating this in a "simulation results cache" object, and letting it own nodeInfos, but that's probably impractical given that they're used extensively by other part of the code.

> This cache is rebuilt every iteration. I do not see why we would like to invalidate any entries in it really.

If we create a node group, we may want to (eventually) run the simulation on the node built from actual instance template. But it's just a possibility at this point, yeah. Right now, we're only adding a new nodeInfo object (or rather, moving an existing one under a new key), and re-running simulations on it to update remaining cache objects. How about instead add another function, e.g. "addNodeGroupAlias(oldId, newId)", which will basically copy all the results associated with oldId to be associated with newId?

Issue #923, opened 2018-06-06
Cluster Autoscaler erroneously triggered during PVC binding


gtie:
Scaling up is currently triggered by any uschedulable pod. A pod can be unschedulable for just a few seconds, however, while it's waiting for a volume to be bound. Those few seconds are enough for Cluster Autoscaler to kick off the creation of a new instance.

Here is how the sequence looks in the event stream:
```
$ k get events --field-selector involvedObject.uid=45b82c29-6970-11e8-a9c2-02145a4f4030 -o=custom-columns=CREATED:.metadata.creationTimestamp,SOURCE:.source.component,REASON:.reason,MSG:.message
CREATED                SOURCE                    REASON                   MSG
2018-06-06T09:59:16Z   default-scheduler         FailedScheduling         pod has unbound PersistentVolumeClaims (repeated 92 times)
2018-06-06T09:59:22Z   cluster-autoscaler        TriggeredScaleUp         pod triggered scale-up: [{1.default.nodes.ASG-XXXXXX 37->38 (max: 39)}]
2018-06-06T09:59:23Z   default-scheduler         Scheduled                Successfully assigned os-strg-osq6574-2-68d64976f6-4s5j2 to ip-10-117-5-157.XXXXXX
2018-06-06T09:59:23Z   kubelet                   SuccessfulMountVolume    MountVolume.SetUp succeeded for volume "logs-volume" 
2018-06-06T09:59:23Z   kubelet                   SuccessfulMountVolume    MountVolume.SetUp succeeded for volume "default-token-tm7c7" 
2018-06-06T09:59:23Z   attachdetach-controller   FailedAttachVolume       AttachVolume.Attach failed for volume "pvc-45afef4a-6970-11e8-96b5-0a38a7e18608" : "Error attaching EBS volume \"vol-0a2eddf3918053d85\"" to instance "i-0c6399097e297ff80" since volume is in "creating" state
2018-06-06T09:59:27Z   attachdetach-controller   SuccessfulAttachVolume   AttachVolume.Attach succeeded for volume "pvc-45afef4a-6970-11e8-96b5-0a38a7e18608" 
2018-06-06T09:59:32Z   kubelet                   SuccessfulMountVolume    MountVolume.SetUp succeeded for volume "pvc-45afef4a-6970-11e8-96b5-0a38a7e18608" 
2018-06-06T09:59:35Z   kubelet                   Pulling                  pulling image "XXXXXX"
[...]

PR #1245, opened 2018-09-14
Add configurable delay for pod age before considering for scale-up

hercynium:
- This is intended to address the issue described in https://github.com/kubernetes/autoscaler/issues/923
  - the delay is configurable via a CLI option
  - in production (on AWS) we set this to a value of 2m
  - the delay could possibly be set as low as 30s and still be effective depending on your workload and environment
  - the default of 0 for the CLI option results in no change to the CA's behavior from defaults.

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -358,11 +358,10 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 	return nil
 }
 
-// don't consider pods newer than newPodScaleUpBuffer seconds old as unschedulable
+// don't consider pods newer than newPodScaleUpDelay seconds old as unschedulable
 func (a *StaticAutoscaler) filterOutYoungPods(allUnschedulablePods []*apiv1.Pod, currentTime time.Time) []*apiv1.Pod {
-	// only consider unschedulable pods older than X
 	var oldUnschedulablePods []*apiv1.Pod
-	newPodScaleUpBuffer := a.AutoscalingOptions.NewPodScaleUpBuffer
+	newPodScaleUpBuffer := a.AutoscalingOptions.NewPodScaleUpDelay
 	for _, pod := range allUnschedulablePods {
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/main.go
```diff
@@ -148,6 +148,7 @@ var (
 	unremovableNodeRecheckTimeout = flag.Duration("unremovable-node-recheck-timeout", 5*time.Minute, "The timeout before we check again a node that couldn't be removed before")
 	expendablePodsPriorityCutoff  = flag.Int("expendable-pods-priority-cutoff", -10, "Pods with priority below cutoff will be expendable. They can be killed without any consideration during scale down and they don't cause scale up. Pods with null priority (PodPriority disabled) are non expendable.")
 	regional                      = flag.Bool("regional", false, "Cluster is regional.")
+	newPodScaleUpBuffer           = flag.Duration("new-pod-scale-up-buffer", 0*time.Second, "Pods less than this many seconds old will not be considered for scale-up.")
```
aleksandra-malinowska: Can we rename it to sth like newPodScaleUpDelay?

Also, "Pods less than this old" is sufficient in description, flag.Duration already handles units.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -355,6 +358,23 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 	return nil
 }
 
+// don't consider pods newer than newPodScaleUpBuffer seconds old as unschedulable
+func (a *StaticAutoscaler) filterOutYoungPods(allUnschedulablePods []*apiv1.Pod, currentTime time.Time) []*apiv1.Pod {
+	// only consider unschedulable pods older than X
```
aleksandra-malinowska: nit: this comment is probably redundant, given there's one above this method

hercynium:
Make the requested changes, plus added some info about this option to the FAQ

aleksandra-malinowska:
Thanks!

/lgtm

PR #1249, opened 2018-09-17
Update GCB config for Cluster Autoscaler

aleksandra-malinowska:
@rvkubiak - follow-up to #665, PTAL if it makes sense


mwielgus:
No-one needed it for a quarter, closing.

PR #1251, opened 2018-09-17
Added r5 instance type

nburn42:
https://aws.amazon.com/ec2/instance-types/r5/


nburn42:
Looks like someone already made the changes when I wasn't looking.

PR #1254, opened 2018-09-18
NodeGroupManager.CreateNodeGroup can return extra created node groups.

losipiuk:


cluster-autoscaler/processors/nodegroups/nodegroup_manager.go
```diff
@@ -35,9 +35,20 @@ type NodeGroupManager interface {
 type NoOpNodeGroupManager struct {
 }
 
+// CreateNodeGroupResult result captures result of successful NodeGroupManager.CreateNodeGroup call.
+type CreateNodeGroupResult struct {
+	// Main created node group, matching the requested node group passed to CreateNodeGroup call
+	MainCreatedNodeGroup cloudprovider.NodeGroup
```
aleksandra-malinowska: Any reason we need to split this (i.e., why a list of node groups won't work)?

losipiuk: Sure it would work. It just looked cleaner to me. Especially given fact that it is quite common to return just one NodeGroup.

aleksandra-malinowska: I tend to disagree re: cleanliness, but that's a nit. One way it changes the behavior is that we run the simulation again on the original nodeInfo (autoprovisioning path) for the main group, and on node built from template (scale from 0 path) for the extra groups. It's inconsistent but I suppose it makes sense as a heuristic - if we already created a group, it's better to add a node even if we couldn't scale it from 0 than to not add one possibly just because of faulty template.

PR #1256, opened 2018-09-18
Refactor the scale-(up|down) status processors so that they have more…

towca:
… info available

Replace the simple boolean ScaledUp property of ScaleUpStatus with a more
comprehensive ScaleUpResult. Add more possible values to ScaleDownResult.
Refactor the processors execution so that they are always executed every
iteration, even if RunOnce exits earlier.

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -132,12 +132,26 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 	}
 	metrics.UpdateDurationFromStart(metrics.UpdateState, stateUpdateStart)
 
+	scaleUpStatus := &status.ScaleUpStatus{Result: status.ScaleUpNotTried}
+	scaleUpStatusProcessorAlreadyCalled := false
+	scaleDownStatus := &status.ScaleDownStatus{Result: status.ScaleDownNotTried}
+	scaleDownStatusProcessorAlreadyCalled := false
+
 	defer func() {
 		// Update status information when the loop is done (regardless of reason)
 		if autoscalingContext.WriteStatusConfigMap {
 			status := a.clusterStateRegistry.GetStatus(currentTime)
 			utils.WriteStatusConfigMap(autoscalingContext.ClientSet, autoscalingContext.ConfigNamespace,
 				status.GetReadableString(), a.AutoscalingContext.LogRecorder)
+		}
+
+		// This deferred processor execution allows the processors to handle a situation when a scale-(up|down)
+		// wasn't even attempted because e.g. the iteration exited earlier.
+		if !scaleUpStatusProcessorAlreadyCalled && a.processors != nil && a.processors.ScaleUpStatusProcessor != nil {
+			a.processors.ScaleUpStatusProcessor.Process(a.AutoscalingContext, scaleUpStatus)
+		}
+		if !scaleDownStatusProcessorAlreadyCalled && a.processors != nil && a.processors.ScaleDownStatusProcessor != nil {
+			a.processors.ScaleDownStatusProcessor.Process(a.AutoscalingContext, scaleDownStatus)
 		}
 
 		err := a.processors.AutoscalingStatusProcessor.Process(a.AutoscalingContext, a.clusterStateRegistry, currentTime)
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/processors/status/scale_down_status_processor.go
```diff
@@ -52,6 +52,23 @@ const (
...
+	// ScaleDownDeleteInProgress - the scale down wasn't attempted, because a node delete was still in progress.
+	ScaleDownDeleteInProgress
+	// ScaleDownScaleUpTooRecent - the scale down wasn't attempted, because a scale-up happened too recently.
+	ScaleDownScaleUpTooRecent
+	// ScaleDownFailureTooRecent - the scale down wasn't attempted, because a scale-down failure happened too recently.
+	ScaleDownFailureTooRecent
+	// ScaleDownNodeDeleteTooRecent - the scale down wasn't attempted, because a scale-down node delete happened too recently.
+	ScaleDownNodeDeleteTooRecent
+	// ScaleDownForbidden - the scale down wasn't attempted, because it was forbidden by previous logic.
+	ScaleDownForbidden
```
aleksandra-malinowska: This is leaking a huge part of internal implementation details to the API. Can we replace it with something more generic?

towca: Done

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -251,47 +265,53 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 	unschedulablePodsToHelp = a.filterOutYoungPods(unschedulablePodsToHelp, currentTime)
 
 	if len(unschedulablePodsToHelp) == 0 {
+		scaleUpStatus.Result = status.ScaleUpNoUnschedulablePods
 		glog.V(1).Info("No unschedulable pods")
 	} else if a.MaxNodesTotal > 0 && len(readyNodes) >= a.MaxNodesTotal {
+		scaleUpStatus.Result = status.ScaleUpLimitExceeded
```
aleksandra-malinowska: This is inconsistent with what happens when we hit limits for each node group's size. Can we use "no options" result here as well?

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -345,9 +370,10 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 				a.lastScaleDownDeleteTime = currentTime
 				a.clusterStateRegistry.Recalculate()
 			}
-			if a.processors != nil && a.processors.ScaleDownStatusProcessor != nil {
```
aleksandra-malinowska: Can we keep this check for now?

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -345,9 +370,10 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
-				a.processors.ScaleDownStatusProcessor.Process(autoscalingContext, scaleDownStatus)
-			}
+
+			a.processors.ScaleDownStatusProcessor.Process(autoscalingContext, scaleDownStatus)
```
aleksandra-malinowska: Is there any reason we need to process result from here, instead of relying on deferred call?

cluster-autoscaler/processors/status/scale_up_status_processor.go
```diff
@@ -40,6 +40,27 @@ type NoScaleUpInfo struct {
...
+
+const (
+	// ScaleUpSuccessful - a scale-up successfully occurred.
+	ScaleUpSuccessful ScaleUpResult = iota
+	// ScaleUpError - an unexpected error occurred during the scale-up attempt.
+	ScaleUpError
+	// ScaleUpNoOptionsAvailable - there were no node groups that could be considered for the scale-up.
+	ScaleUpNoOptionsAvailable
+	// ScaleUpNoUnschedulablePods - there was no need for a scale-up because there were no unschedulable pods.
+	ScaleUpNoUnschedulablePods
```
aleksandra-malinowska: The "no unschedulable pods" part may not always be true (e.g. if we remove a pod from list during preprocessing). ScaleUpNotNeeded?

cluster-autoscaler/core/static_autoscaler.go
towca: Consistency with the scale-up processor, i.e. if we added some logic between the scale down and the deferred call, this logic should be able to depend on the fact that if a scale down happens, its result is processed immediately after (as is the case with a scale up). Unless we don't need that guarantee for the scale-up processor, then they can both be executed in the deferred call.

cluster-autoscaler/core/static_autoscaler.go
towca: done

cluster-autoscaler/core/static_autoscaler.go
towca: done

cluster-autoscaler/processors/status/scale_up_status_processor.go
towca: done

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -132,6 +132,11 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 	}
 	metrics.UpdateDurationFromStart(metrics.UpdateState, stateUpdateStart)
 
+	scaleUpStatus := &status.ScaleUpStatus{Result: status.ScaleUpNotTried}
+	scaleUpStatusHandled := false
```
losipiuk: nit: maybe `scaleUpStatusAlreadyHandled` or more explicitly `scaleUpStatusProcessorAlreadyCalled`.

PR #1260, opened 2018-09-19
Add simple health-check to all VPA components.

kgolab:
For updater & recommender we stop returning "healthy" if more than five loops passed without an update (process stuck somewhere).
For admission controller we always return "healthy".

@bskiba please decide whether you want to have this configurable or report "unhealthy" in some other situations.

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/main.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"flag"
 	"net/http"
+	"time"
 
 	"github.com/golang/glog"
 	kube_flag "k8s.io/apiserver/pkg/util/flag"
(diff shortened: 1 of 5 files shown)


vertical-pod-autoscaler/pkg/admission-controller/main.go
```diff
@@ -30,6 +30,7 @@ import (
 	metrics_admission "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/metrics/admission"
 	vpa_api_util "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/vpa"
 	"k8s.io/client-go/rest"
+	"time"
```
bskiba: nit: organize imports

bskiba:
/lgtm 
one nit only

Issue #1228, opened 2018-09-06
Failure to create AWS manager: limited to 50 ASG names


downeast:
Hello, we are running the cluster autoscaler on a cluster with many different scaling groups, and hit the following error on startup:

```
E0906 20:27:17.672470       1 aws_manager.go:122] Failed to regenerate ASG cache: ValidationError: The number of group names that may be passed in is limited to 50
	status code: 400, request id: 3f77f9d5-b213-11e8-9faf-6919af635e71
F0906 20:27:17.672503       1 cloud_provider_builder.go:137] Failed to create AWS Manager: ValidationError: The number of group names that may be passed in is limited to 50
	status code: 400, request id: 3f77f9d5-b213-11e8-9faf-6919af635e71
```

Seems like we need some batching logic here?  This is version 1.3.1.

johanneswuerbach:
@downeast sorry for the late reply, I recently rewrote the ASG autodiscovery logic, could you share whether you are using auto-discovery or explicitly defined asg names?

As far as I know all API calls should use paginate or support > 50 names.

johanneswuerbach:
Scratch that, this method currently doesn't handle if number of ASG names is larger then the set max records https://github.com/kubernetes/autoscaler/blob/b8022c728cf3fbf17e8570f734618f739f3cc816/cluster-autoscaler/cloudprovider/aws/auto_scaling.go#L62-L65

PR #1261, opened 2018-09-20
Fetch ASG names in chunks of 50

johanneswuerbach:
Fixes https://github.com/kubernetes/autoscaler/issues/1228

While https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_DescribeAutoScalingGroups.html reports a `maximum length of 1600`, it seems that only max. 50 names are allowed as input https://github.com/cristim/autospotting/issues/199#issuecomment-355764282


losipiuk:
Thanks!

PR #1264, opened 2018-09-21
Add kube-state-metrics which uses addon-resizer

MIBc:
Can we add kube-state-metrics which also use addon-resizer to improve its performance:
https://github.com/kubernetes/kube-state-metrics/blob/master/kubernetes/kube-state-metrics-deployment.yaml#L34


warmchang:
lgtm

PR #1265, opened 2018-09-21
Check for ready condition not true

losipiuk:


losipiuk:
Cherry pick from master

PR #1269, opened 2018-09-23
Adds a FAQ entry with all commandline parameters

### LINKED ISSUE #918: [cluster-autoscaler]: parameter documentation
I can't find any documentation of all parameters - forcing me to download binary and running --help. It would be awesome to have a complete list of all parameters and their use.

WebSpider:
This PR adds an entry in FAQ.md with all current commandline parameters, description and default value
Fixes #918

### DIFF
--- a/cluster-autoscaler/FAQ.md
+++ b/cluster-autoscaler/FAQ.md
@@ -41,6 +41,7 @@ this document:
   * [How fast is HPA when combined with CA?](#how-fast-is-hpa-when-combined-with-ca)
   * [Where can I find the designs of the upcoming features?](#where-can-i-find-the-designs-of-the-upcoming-features)
   * [What are Expanders?](#what-are-expanders)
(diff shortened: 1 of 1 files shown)


WebSpider:
Fixed the CLA signing

xmik:
There are more command line parameters:
   * `--nodes` and `--node-group-auto-discovery` configured [here](https://github.com/kubernetes/autoscaler/blob/1.12.0/cluster-autoscaler/main.go#L124). Earlier they were configured in the [`main`](https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-1.2.2/cluster-autoscaler/main.go#L287) function, so that is probably why you missed them.
   * leader-election flags [here](https://github.com/kubernetes/kubernetes/blob/v1.13.0-alpha.0/pkg/client/leaderelectionconfig/config.go#L31). 

I don't know about any other parameters.

PR #1274, opened 2018-09-25
Cherry-pick: Fix error on node info not found for group

losipiuk:


losipiuk:
Fixes observed panic (but not the root cause) for https://github.com/kubernetes/autoscaler/issues/1270

aleksandra-malinowska:
IIRC we've recently been refactoring NewAutoscalerError to ToAutoscalerError in order to avoid re-packaging of errors. Are we sure we want to go back here?

aleksandra-malinowska:
Can we cherry-pick #1076 instead?

aleksandra-malinowska:
It would prevent the panic, but error message would be less useful. Sorry, I misread the original issue.

PR #1279, opened 2018-09-27
Fix info on Min and Max Allowed

bskiba:
Min/Max Allowed does not affect recommendation but actuation.


bskiba:
So the problem is not with this PR but with the way this is actually implemented (because this is how it works at the moment). 
I think the intent was to keep the recommendation unmodified - actually recommend how much the workload needs so that if it's actually over the max, the user can see it. Then only enforce this during actuation. Of course this can be done in another way. Let's discuss offline.

PR #1283, opened 2018-09-28
Updater uses informers to fetch pod controllers.

bskiba:
This includes SatefulSets, Replicasets and ReplicationContollers.

### DIFF
--- a/vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go
+++ b/vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go
@@ -18,14 +18,23 @@ package eviction
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/golang/glog"
+	appsv1 "k8s.io/api/apps/v1"
 	apiv1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	metrics_updater "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/metrics/updater"
+	appsinformer "k8s.io/client-go/informers/apps/v1"
+	coreinformer "k8s.io/client-go/informers/core/v1"
 	kube_client "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
+)
+
+const (
+	resyncPeriod time.Duration = 1 * time.Minute
 )
 
 // PodsEvictionRestriction controls pods evictions. It ensures that we will not evict too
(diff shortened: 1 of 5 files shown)


vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go
```diff
@@ -220,35 +263,54 @@ func getPodID(pod *apiv1.Pod) string {
...
-func getReplicaCount(creator podReplicaCreator, client kube_client.Interface) (int, error) {
+func (f *podsEvictionRestrictionFactoryImpl) getReplicaCount(creator podReplicaCreator) (int, error) {
 	switch creator.Kind {
-	case "ReplicationController":
-		rc, err := client.CoreV1().ReplicationControllers(creator.Namespace).Get(creator.Name, metav1.GetOptions{})
-
-		if err != nil || rc == nil {
+	case replicationController:
+		rcObj, exists, err := f.rcInformer.GetStore().GetByKey(creator.Namespace + "/" + creator.Name)
+		if err != nil || rcObj == nil {
```
kgolab: Shouldn't the == nil check be done together with !exists.

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go
```diff
@@ -220,35 +263,54 @@ func getPodID(pod *apiv1.Pod) string {
...
 		}
 		return int(*rc.Spec.Replicas), nil
 
-	case "ReplicaSet":
-		rs, err := client.ExtensionsV1beta1().ReplicaSets(creator.Namespace).Get(creator.Name, metav1.GetOptions{})
-
-		if err != nil || rs == nil {
+	case replicaSet:
+		rsObj, exists, err := f.rsInformer.GetStore().GetByKey(creator.Namespace + "/" + creator.Name)
+		if err != nil || rsObj == nil {
```
kgolab: Same as for line 270.

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go
```diff
@@ -113,9 +134,9 @@ func (e *podsEvictionRestrictionImpl) Evict(podToEvict *apiv1.Pod, eventRecorder
 			Name:      podToEvict.Name,
 		},
 	}
-	err := e.client.CoreV1().Pods(podToEvict.Namespace).Evict(eviction)
+	err := e.client.CoreV1().Pods(eviction.Namespace).Evict(eviction)
```
kgolab: This seems unrelated (and does not change the code behaviour).

bskiba: Done.

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go
bskiba: I think this check is redundant anyway. Removed.

bskiba:
Also fixed failing test

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go
```diff
@@ -61,14 +70,26 @@ type PodsEvictionRestrictionFactory interface {
 
 type podsEvictionRestrictionFactoryImpl struct {
 	client                    kube_client.Interface
+	rcInformer                cache.SharedIndexInformer
```
kgolab: Could you please either extend the names here or at least comment them? Looking below it's easy to understand what they mean but without this context it's not clear so let's have some context directly here.

bskiba: Done

PR #1286, opened 2018-09-28
AWS: Support selecting ASG by values - 1.2 Branch

### LINKED ISSUE #1280: node-group-auto-discovery no longer accepts multiple tags
We're running a Kubernetes 1.10.8 cluster in AWS (provisioned via KOPS) with the following flags:
```
--namespace=k8s-autoscaler
--v=4
--stderrthreshold=info
--cloud-provider=aws
--skip-nodes-with-local-storage=false
--expander=least-waste
--node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler=enabled,KubernetesCluster=k8s.ourcluster.domain
```

This worked fine with Autoscaler 1.1.3, however when we upgrade to 1.2.3 it fails with the following error:
```
I0927 17:20:35.816328       1 cloud_provider_builder.go:72] Building aws cloud provider.
I0927 17:20:35.916130       1 request.go:481] Throttling request took 399.921532ms, request: GET:https://100.64.0.1:443/api/v1/services?limit=500&resourceVersion=0
F0927 17:20:35.916206       1 cloud_provider_builder.go:137] Failed to create AWS Manager: invalid key=value pair [tag k8s.io/cluster-autoscaler enabled, KubernetesCluster k8s.ourcluster.domain]
```

Looks like the behavior was changed by this commit:
53fc344ecabbad73c0063d065bc2223d43318033

cablespaghetti:
Fixes #1280

I've pushed a built docker container here under the v1.2-1280fix tag: https://hub.docker.com/r/cablespaghetti/cluster-autoscaler/tags/

@ReillyProcentive can you verify that this fixes your problem?

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/auto_scaling.go
+++ b/cluster-autoscaler/cloudprovider/aws/auto_scaling.go
@@ -92,17 +92,23 @@ func (m *autoScalingWrapper) getAutoscalingGroupsByNames(names []string) ([]*aut
 	return asgs, nil
 }
 
-func (m *autoScalingWrapper) getAutoscalingGroupsByTags(keys []string) ([]*autoscaling.Group, error) {
+func (m *autoScalingWrapper) getAutoscalingGroupsByTags(kvs map[string]string) ([]*autoscaling.Group, error) {
 	// DescribeTags does an OR query when multiple filters on different tags are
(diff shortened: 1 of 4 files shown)


ReillyProcentive:
I'm already running with a docker image built locally with this patch included and it works fine.

PR #1294, opened 2018-10-02
Recalculate clusterStateRegistry after adding multiple node groups

losipiuk:


MaciekPytel:
Wouldn't the same issue affect balancing main created group with pre-existing similar nodegroups? Maybe we should just always recalculate after creating nodegroup? I'm not sure if any performance gain of only doing it when multiple node groups were created is worth spilling the logic all over the place anyway.

losipiuk:
>Wouldn't the same issue affect balancing main created group with pre-existing similar nodegroups? Maybe we should just always recalculate after creating nodegroup? I'm not sure if any performance gain of only doing it when multiple node groups were created is worth spilling the logic all over the place anyway.

Definitely makes sense.
Updated.

MaciekPytel:
/lgtm
/approve

/hold 
For any manual testing required. Feel free to cancel hold once tested.

losipiuk:
I tested code before last change. I will risk it still work :) Adding e2e tests anyway.

PR #1303, opened 2018-10-09
Keep small-weight samples for at least MemoryAggregationWindowLength

kgolab:
Together with PR #1302 this fixes a problem with "zero recommendation" appearing in some new clusters.

NOTE: This PR changes how long samples are kept, in particular when restoring a checkpoint.

/assign @bskiba 
/cc @schylek


bskiba:
lgtm from me, @schylek, can you take a look this makes sense?

kgolab:
Changed to const, as requested.

PR #1312, opened 2018-10-11
VPA drain context.Done() channel in checkpoint writer.

schylek:


vertical-pod-autoscaler/pkg/recommender/checkpoint/checkpoint_writer.go
```diff
@@ -79,6 +79,11 @@ func (writer *checkpointWriter) StoreCheckpoints(ctx context.Context, now time.T
 	vpas := getVpasToCheckpoint(writer.cluster.Vpas)
 	for _, vpa := range vpas {
 
+		select {
+		case <-ctx.Done():
```
bskiba: Is seems very weird that we do nothing in either case. What happens if there is a message in ctx.Done? Will ctx.Err() != nil below?

schylek: Agree this is weird, but this is required if we want to have a min number of checkpoints written (i.e. ignore context for a little while). When Done is closed ctx.Err() will repeatedly return error  (https://golang.org/pkg/context/#Context)

bskiba: OKay, then please just leave a short comment why we're doing it this way.

PR #1319, opened 2018-10-15
Refactor cloud provider builder to allows build for a single provider

aleksandra-malinowska:
This PR:
- moves code for creating cloud providers inside the respective modules,
- adds support for creating {AWS,AZURE,GCE,Kubemark}-only builds.

Binary size depending on build:
- all (default): 166MB
- AWS: 65MB
- Azure: 56MB
- GCE: 75MB
- Kubemark: 164MB


cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -303,3 +307,27 @@ func (mig *gceMig) TemplateNodeInfo() (*schedulercache.NodeInfo, error) {
 	nodeInfo.SetNode(node)
 	return nodeInfo, nil
 }
+
+// BuildGCE builds GCE cloud provider, manager etc.
+func BuildGCE(opts config.AutoscalingOptions, do cloudprovider.NodeGroupDiscoveryOptions, rl *cloudprovider.ResourceLimiter) cloudprovider.CloudProvider {
```
losipiuk: nit: Following go convention this method should be called BuildCloudProvider. And the the call would look gce.BuildCloudProvider.

With that said I am not fan of the convention :)

cluster-autoscaler/cloudprovider/builder/builder_all.go
```diff
@@ -0,0 +1,58 @@
+// +build !gce,!aws,!azure,!kubemark
```
losipiuk: I would prefer having explicit tag for that `all_cp` so we do not have to update this list when new cloud provider is added.
Given that adding `_cp` suffix to cloud-provider specific tags seems reasonable to me.

cluster-autoscaler/Makefile
```diff
@@ -6,29 +6,38 @@ ENVVAR=
 GOOS?=linux
 REGISTRY?=staging-k8s.gcr.io
 BASEIMAGE?=k8s.gcr.io/debian-base-amd64:0.3.2
+ifdef BUILD_TAGS
+  TAGS_FLAG=--tags ${BUILD_TAGS}
+  PROVIDER=-${BUILD_TAGS}
+  FOR_PROVIDER=" for ${BUILD_TAGS}"
```
losipiuk: This one is weird? Can we somehow use `$BUILD_TAGS` in `echo` calls below.
That said I admit that I have more or less 0 knowledge on Makefile - so maybe it is actually hard.

aleksandra-malinowska: I'm not at all familiar with Makefile best practices either, just wanted do avoid ifdefs everywhere (passing empty --tags is invalid, I'd expect so is image name ending in dash (cluster-autoscaler-:<tag>) etc.) If there's a better way, I'd be glad to learn :)

PR #1320, opened 2018-10-16
E2e test for no PDB violations.

bskiba:
@kgolab


vertical-pod-autoscaler/e2e/updater.go
```diff
@@ -78,18 +74,96 @@ var _ = UpdaterE2eDescribe("Updater", func() {
...
 
 	})
+
+	ginkgo.It("observes pod disruption budget", func() {
+
+		ginkgo.By("Setting up a hamster deployment")
+		c := f.ClientSet
+		ns := f.Namespace.Name
+
+		cpuQuantity := ParseQuantityOrDie("10m")
```
kgolab: Could you please extract this (up to podList creation) to a helper method? This code is very similar to the other test.

vertical-pod-autoscaler/e2e/updater.go
```diff
@@ -78,18 +74,96 @@ var _ = UpdaterE2eDescribe("Updater", func() {
...
+				MaxUnavailable: &maxUnavailable,
+				Selector: &metav1.LabelSelector{
+					MatchLabels: map[string]string{"app": "hamster"},
+				},
+			},
+		}
+		_, err = c.PolicyV1beta1().PodDisruptionBudgets(ns).Create(pdb)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+		ginkgo.By("Setting up a VPA CRD")
```
kgolab: This section could be extracted to a helper method too.

vertical-pod-autoscaler/e2e/updater.go
```diff
@@ -78,18 +74,96 @@ var _ = UpdaterE2eDescribe("Updater", func() {
...
+		d, err := c.AppsV1().Deployments(ns).Create(d)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		err = framework.WaitForDeploymentComplete(c, d)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+		podList, err := framework.GetPodsForDeployment(c, d)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		podSet := makePodSet(podList)
+
+		ginkgo.By("Setting up prohibitive PDB for hamster deployment")
```
kgolab: If you extract this too (even though it's used only once), the test will become easier to read - there is no need to skim the code to next ginkgo.By then.

vertical-pod-autoscaler/e2e/updater.go
```diff
@@ -78,18 +74,96 @@ var _ = UpdaterE2eDescribe("Updater", func() {
...
+		newPdb := pdb.DeepCopy()
+		newPdb.Spec.MaxUnavailable = &maxUnavailable
+		newPdb.ObjectMeta.Name = "hamster-pdb-2"
+		// Creating new PDB and removing old one, since PDBs are immutable at the moment
+		_, err = c.PolicyV1beta1().PodDisruptionBudgets(ns).Create(newPdb)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		err = c.PolicyV1beta1().PodDisruptionBudgets(ns).Delete(pdb.Name, &metav1.DeleteOptions{})
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+		ginkgo.By(fmt.Sprintf("Waiting for pods to be restarted, sleep for %s", VpaEvictionTimeout.String()))
```
kgolab: Is it possible to verify here that the Pod evictions were done in parallel? I think this is one of the goals of the test (using PDB > 1) so it would be nice to have this verified.

vertical-pod-autoscaler/e2e/common.go
```diff
@@ -39,6 +39,9 @@ const (
 	actuationSuite               = "actuation"
 	pollInterval                 = 10 * time.Second
 	pollTimeout                  = 15 * time.Minute
+	// VpaEvictionTimeout is a timeout for VPA to restart a pod if there are no
+	// mechanisms blocking it (for example PDB).
+	VpaEvictionTimeout = 3 * time.Minute
```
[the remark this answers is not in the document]
bskiba: Will be needed in later changes.

vertical-pod-autoscaler/e2e/updater.go
kgolab: OK. Maybe a one-line comment on the relation between this 7 and 3 minutes in time-out would be nice - we can see it now but it might be not so obvious in a month.

vertical-pod-autoscaler/e2e/updater.go
```diff
@@ -78,20 +63,110 @@ var _ = UpdaterE2eDescribe("Updater", func() {
...
+		gomega.Expect(restartedCount >= permissiveMaxUnavailable).To(gomega.BeTrue())
 	})
 })
 
+func setupHamsterDeployment(f *framework.Framework, cpu, memory string, replicas int32) (*appsv1.Deployment, *apiv1.PodList) {
+	cpuQuantity := ParseQuantityOrDie(cpu)
+	memoryQuantity := ParseQuantityOrDie(memory)
+
+	d := NewHamsterDeploymentWithResources(f, cpuQuantity, memoryQuantity)
+	if replicas > 0 {
```
kgolab: This looks like something that could be improved - either by allowing to pass nil explicitly (I understand it's not too nice for the callers) or at least specifying this behaviour in function comment.

kgolab:
Thanks for the updates. Found two non-blocking nits so LGTM already.

vertical-pod-autoscaler/e2e/updater.go
bskiba: Good point. Done

vertical-pod-autoscaler/e2e/updater.go
bskiba: Went for passing nil, it's a bit awkward for the caller not not too bad.

PR #1325, opened 2018-10-18
Remove duplicate words

xichengliudui:
Remove duplicate words


aleksandra-malinowska:
This change is modifying Cluster Autoscaler's dependencies, please make it to upstream instead.

PR #1326, opened 2018-10-18
Ignore lo/fluentd-ds-ready when checking node similarity

losipiuk:


cluster-autoscaler/utils/nodegroupset/compare_nodegroups.go
```diff
@@ -89,19 +89,20 @@ func IsNodeInfoSimilar(n1, n2 *schedulercache.NodeInfo) bool {
...
 		for label, value := range node.Node().ObjectMeta.Labels {
-			if label == kubeletapis.LabelHostname {
-				continue
-			}
-			if label == kubeletapis.LabelZoneFailureDomain {
-				continue
-			}
-			if label == kubeletapis.LabelZoneRegion {
-				continue
+			ignore, _ := ignoredLabels[label]
```
aleksandra-malinowska: nit: if disregarding second returned value, you can just as well write `if !ignoredLabels[label] {`

PR #1327, opened 2018-10-18
Start using beta VPA API

bskiba:
The permissions and scripts are updated so that both apis can be used.
This way both ./hack/vpa-up.sh and ./hack/run-e2e.sh will work.


bskiba:
@kgolab @schylek

PR #1331, opened 2018-10-19
Allow updating Increase for ScaleUpRequest in ClusterStateRegistry

losipiuk:


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -154,21 +154,39 @@ func NewClusterStateRegistry(cloudProvider cloudprovider.CloudProvider, config C
...
 
-// RegisterScaleUp registers scale up.
-func (csr *ClusterStateRegistry) RegisterScaleUp(request *ScaleUpRequest) {
+// RegisterOrUpdateScaleUp registers scaleup for give node group or changes requested node increase
+// count.
+// If increaseDelta is positive then number of new nodes requested is increased; Time and expectedAddTime
+// are reset.
+// If increaseDelete is negative the number of new nodes requested is decreased; Time and expectedAddTime are
+// left intact.
+func (csr *ClusterStateRegistry) RegisterOrUpdateScaleUp(nodeGroupName string, increaseDelta int) {
```
aleksandra-malinowska: Why rename this method? It seems we still only use it to register a scale-up, just like before - how we store the data internally (updating a struct or always appending a new one) is an implementation detail that caller shouldn't have to know.

Secondly, it seems like this change is preparing ground for "registering" a scale-up with negative delta (scale-down). Can you explain how this will work before we commit to this change?

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -154,21 +154,39 @@ func NewClusterStateRegistry(cloudProvider cloudprovider.CloudProvider, config C
...
 	defer csr.Unlock()
 
-	oldScaleUpRequest, found := csr.scaleUpRequests[request.NodeGroupName]
+	scaleUpRequest, found := csr.scaleUpRequests[nodeGroupName]
 	if !found {
-		csr.scaleUpRequests[request.NodeGroupName] = request
-		return
+		scaleUpRequest = &ScaleUpRequest{
+			NodeGroupName:   nodeGroupName,
+			Time:            time.Now(),
```
aleksandra-malinowska: Hard-coding time.Now() here seems like a step backwards - we've been reducing it across the codebase, both to make unit testing easier and the reported behavior more consistent.

Note that the previous change was implementation-only and didn't require updating any unit tests - they continued to work and helped us verify that the refactoring indeed didn't change the behavior in covered scenarios. In other words, they weren't "change-detection" tests. With this change, you had to add manipulating internal state of ClusterStateRegistry (substituting timestamps) for them to continue to work. This means that:
1. if we were to revert the other change, or made another refactoring of internal implementation, it would break the tests this time, because they now make assumptions about internal state and manipulate it,
2. old tests no longer verify the logic of aggregating scale-up requests and using a correct timeout - we replace these values. If you set ExpectedAddTime to time.Now() in this method, backoff tests wouldn't fail. So a new test had to be added to cover this, making even more assumptions about internal state (which makes subsequent code changes harder, see (1)). 

(Yes, a lot of our unit tests make assertions on internal state, I'm aware of this;P)

cluster-autoscaler/clusterstate/clusterstate.go
losipiuk: The other usecase for this method will be to to decrease the "Increase" in case we notice that part of the scale-up is not possible due to stockout/quota exceeded errors.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -154,21 +154,39 @@ func NewClusterStateRegistry(cloudProvider cloudprovider.CloudProvider, config C
...
+			ExpectedAddTime: time.Now().Add(csr.config.MaxNodeProvisionTime),
```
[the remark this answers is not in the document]
losipiuk: I am fine with passing `ExpectedAddDuration` as parameter here. Though I believe we should not expose internals of ScaleUpRequest (i.e. the `Time` field to caller of this method). It is implementation detail.

cluster-autoscaler/clusterstate/clusterstate.go
losipiuk: > Hard-coding time.Now() here seems like a step backwards - we've been reducing it across the codebase, both to make unit testing easier and the reported behavior more consistent.

The problem here is using static `time.Now()` method. We could fix that if we passed some `Watch` object on construction of `ClusterStateRegistry`. Standard implementation would represent wall time. And in tests we could put sth where we can manipulate the time programtically. Not sure if golang have sth like this.

> With this change, you had to add manipulating internal state of ClusterStateRegistry

I agree that I do that. Though I disagree that these tests did not do that already. Just the internal state was exposed to the outer caller via `ScaleUpRequest` struct.

aleksandra-malinowska: > I faintly recall seeing some flags and configuration options (i.e. things normally not exposed to GKE users) getting moved into ConfigMaps. I don't know the details, but perhaps this approach would some day work for tuning the autoscaler in GKE (although I'm sure that further complicates the GKE support story).

Technically, we could use an internal timeNow (or sth like this) function, assign time.Now() to it in prod and custom fakeTime() in tests. This complicates test setup, though.

> Just the internal state was exposed to the outer caller via ScaleUpRequest struct.

To me, it seems we passed input parameter and then never checked how CSR stored it, only that it made correct assessments based on it. 

Would it be OK to pass startTime and expectedAddDuration (or expectedAddTime, doesn't matter) as parameters directly? Or if ScaleUpRequest (exported) was part of this method's interface, but internally another type was used for storing scale-up state?

cluster-autoscaler/clusterstate/clusterstate.go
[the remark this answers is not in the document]
aleksandra-malinowska: Perhaps the problem is that the same exported struct, ScaleUpRequest, is used as part of this methods interface and to keep track of the state internally? I checked and it seems we only use it outside CSR when registering scale-up, and if this is removed (either in favor of the list of parameters, or another struct), it can be made private.

cluster-autoscaler/clusterstate/clusterstate.go
aleksandra-malinowska: Thanks. Given the discussion below about passing time and duration parameters, I wonder if it's really a single method? That is, passing startTime and expectedAddTime/Duration makes sense when first registering a scale-up, but not really when updating it on encountering a stockout/quota error. Perhaps CSR can expose two methods, RegisterScaleUp and sth else (UpdateScaleUpRequest? RegisterScaleUpError? DontExpectNoMoreScaleUpFromHere?), which may share part of the logic internally (or not). Then we'd also have more flexibility with handling such conditions, e.g. backing off a node group on stockout, but not on quota error (this is just an example/idea, not sure if we'd want such logic).

losipiuk: > Thanks. Given the discussion below about passing time and duration parameters, I wonder if it's really a single method? That is, passing startTime and expectedAddTime/Duration makes sense when first registering a scale-up, but not really when updating it on encountering a stockout/quota error. Perhaps CSR can expose two methods, RegisterScaleUp and sth else (UpdateScaleUpRequest? RegisterScaleUpError? DontExpectNoMoreScaleUpFromHere?), which may share part of the logic internally (or not). Then we'd also have more flexibility with handling such conditions, e.g. backing off a node group on stockout, but not on quota error (this is just an example/idea, not sure if we'd want such logic).

We can possibly have two methods. One to increase and one to decrease. Though I would like to avoid that if keeping simpler interface if feasible. 
To me `startTime` should not be part of interface no matter if we pass it via struct or parameter. I production code we always pass `time.Now()` there, and we exploit the fact it exists only in tests. 
I do not feel it is better than introspecting internal CSR's data structures. Actually I feel it is worse as it puts test code above clarity of interface in prod code.

Anyway. Let me build a bit on top of this and with some commits on top it could be easier to come to common understanding how it should look like.

aleksandra-malinowska: > Actually I feel it is worse as it puts test code above clarity of interface in prod code.

Fair point. I suppose in other places it's more justified by consistency - we process the state from start of the loop, and it makes sense to use that timestamp when reporting node group health etc., but not here.

WDYT about substituting time.Now() in tests?

losipiuk:
Closing this one for sake of https://github.com/kubernetes/autoscaler/pull/1332.
The other one has more context.

PR #1335, opened 2018-10-22
VPA reducing memory pressure for inspecting pod updates

schylek:


bskiba:
Something is broken with the checks, I'll try closing and reopening.

PR #1337, opened 2018-10-22
Cluster Autoscaler 1.2.4

aleksandra-malinowska:
Update Cluster Autoscaler version to 1.2.4

### DIFF
--- a/cluster-autoscaler/version.go
+++ b/cluster-autoscaler/version.go
@@ -17,4 +17,4 @@ limitations under the License.
 package main
 
 // ClusterAutoscalerVersion contains version of CA.
-const ClusterAutoscalerVersion = "1.2.3"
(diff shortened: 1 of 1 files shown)


aleksandra-malinowska:
It isn't, see #1346

feiskyer:
Since this is still not ready yet, could we wait a while for a critical bug fix for Azure?

MaciekPytel:
Note that Kubernetes 1.10 (and by extension CA 1.2) will go end of life in the next few days. Ideally CA 1.2.4 should be released in time to make it into the final 1.10 release. So we don't have much time for fixes.

spronin-aurea:
Can we move forward here to release with [#1478](https://github.com/kubernetes/autoscaler/pull/1478) please ?

PR #1340, opened 2018-10-23
NodeGroup.Nodes() return Instance struct instead instance name

losipiuk:
This is preparatory work for handling resource related
(stockout/quota-exceeded) error conditions in CA.


losipiuk:
cc: @feiskyer, @mumoshu, @sethpollack, @ringtail

MaciekPytel:
Please fix unittests. Otherwise lgtm.

PR #1341, opened 2018-10-23
Move nodegroup balancing to processor, add GKE-specific implementation

MaciekPytel:
This allows customizing the balancing logic for different use-cases. In particular this PR implements GKE specific version (only enabled if provider is gke) that considers node groups with the same gke-nodepool as similar.


MaciekPytel:
@losipiuk Refactored based on our conversation, PTAL.

cluster-autoscaler/processors/nodegroupset/balancing_processor.go
```diff
@@ -17,30 +17,54 @@ limitations under the License.
...
+		if ngId == nodeGroupId {
+			continue
+		}
+		ngNodeInfo, found := nodeInfosForGroups[ngId]
+		if !found {
+			glog.Warningf("Failed to find nodeInfo for group %v", ngId)
+			continue
+		}
+		comparator := b.Comparator
+		if comparator == nil {
```
losipiuk: I would rather not have this logic here and treat Comparator as obligatory paremetrization.

cluster-autoscaler/processors/nodegroupset/balancing_processor.go
```diff
@@ -17,30 +17,54 @@ limitations under the License.
...
+		ngId := ng.Id()
+		if ngId == nodeGroupId {
+			continue
+		}
+		ngNodeInfo, found := nodeInfosForGroups[ngId]
+		if !found {
+			glog.Warningf("Failed to find nodeInfo for group %v", ngId)
+			continue
+		}
+		comparator := b.Comparator
```
losipiuk: why copy to variable?

cluster-autoscaler/processors/nodegroupset/compare_nodegroups_test.go
```diff
@@ -29,22 +29,22 @@ import (
...
 	ni2.SetNode(n2)
-	assert.Equal(t, shouldEqual, IsNodeInfoSimilar(ni1, ni2))
+	assert.Equal(t, shouldEqual, comparator(ni1, ni2))
 }
 
 func TestIdenticalNodesSimilar(t *testing.T) {
 	n1 := BuildTestNode("node1", 1000, 2000)
 	n2 := BuildTestNode("node2", 1000, 2000)
-	checkNodesSimilar(t, n1, n2, true)
+	checkNodesSimilar(t, n1, n2, IsNodeInfoSimilar, true)
```
losipiuk: what is the reason for passing same comparator for every test?
Drop argument?

losipiuk: Ok I see the other test.

cluster-autoscaler/processors/nodegroupset/gke_compare_nodegroups.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// IsGkeNodeInfoSimilar compares if two nodes should be considered part of the
+// same NodeGroupSet. This is true if they either belong to the same GKE nodepool
+// or match usual conditions checked by IsNodeInfoSimilar.
+func IsGkeNodeInfoSimilar(n1, n2 *schedulercache.NodeInfo) bool {
+	if nodesFromSameGkeNodePool(n1, n2) {
+		return true
+	}
+	return IsNodeInfoSimilar(n1, n2)
```
losipiuk: Or maybe return false? WDYT? Too much a change of semantics?

cluster-autoscaler/processors/nodegroupset/balancing_processor.go
MaciekPytel: I think it's a standard coding pattern for applying defaults :p

losipiuk:
@MaciekPytel unhold if you do are not willing to do any changes. Everything is optional from my side.

PR #1354, opened 2018-10-25
Recommender capps recommendation according to policy.

bskiba:


bskiba:
@schylek @kgolab Will one of you have capacity to review?

kgolab:
I'll take a look.

vertical-pod-autoscaler/pkg/recommender/routines/recommender.go
```diff
@@ -100,15 +100,21 @@ func (r *recommender) UpdateVPAs() {
...
 			containerResources = append(containerResources, vpa_types.RecommendedContainerResources{
-				ContainerName: containerName,
-				Target:        model.ResourcesAsResourceList(res.Target),
-				LowerBound:    model.ResourcesAsResourceList(res.LowerBound),
-				UpperBound:    model.ResourcesAsResourceList(res.UpperBound),
+				ContainerName:  containerName,
+				Target:         model.ResourcesAsResourceList(res.Target),
+				LowerBound:     model.ResourcesAsResourceList(res.LowerBound),
+				UpperBound:     model.ResourcesAsResourceList(res.UpperBound),
+				UncappedTarget: model.ResourcesAsResourceList(res.Target),
```
kgolab: This is confusing even though correct.

Could you please add some explanation what's going on here as a comment?
Or maybe extract full creation of containerResources [up to line 116 or thereabouts] to a function where you could simply comment on its internals w/o overloading the code here.

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
```diff
@@ -85,14 +85,15 @@ func (c *cappingRecommendationProcessor) Apply(
 
 // getCappedRecommendationForContainer returns a recommendation for the given container, adjusted to obey policy and limits.
 func getCappedRecommendationForContainer(
-	container apiv1.Container,
+	containerName string,
```
kgolab: Why pass both the name & container if the name is trivially accessible?

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
```diff
@@ -153,6 +156,30 @@ func applyVPAPolicy(recommendation apiv1.ResourceList, policy *vpa_types.Contain
...
+		return nil, nil
+	}
+	if policy == nil {
+		return podRecommendation, nil
+	}
+
+	updatedRecommendations := []vpa_types.RecommendedContainerResources{}
+	for _, containerRecommendation := range podRecommendation.ContainerRecommendations {
+		containerName := containerRecommendation.ContainerName
+		updatedContainerResources, _, err := getCappedRecommendationForContainer(
```
kgolab: Can you please double-check if reusing the same function while ignoring the computed annotations is performant?

I recall a story that somewhere in CA creation of error messages which were later ignored proved to be the biggest performance issue around simulations. There it has been solved by adding an argument like "silent/verbose".

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
```diff
@@ -153,6 +156,30 @@ func applyVPAPolicy(recommendation apiv1.ResourceList, policy *vpa_types.Contain
...
+			containerName, nil, &containerRecommendation, policy, false /* capToLimit */)
+
+		if err != nil {
```
kgolab: This currently cannot happen which might be treated as an argument to have a specialised function instead of resuing getCapped[...]

vertical-pod-autoscaler/pkg/utils/vpa/capping_test.go
```diff
@@ -184,3 +184,57 @@ func TestApply(t *testing.T) {
...
+				ContainerName: "ctr-name",
+				Target: apiv1.ResourceList{
+					apiv1.ResourceCPU:    *resource.NewScaledQuantity(10, 1),
+					apiv1.ResourceMemory: *resource.NewScaledQuantity(5000, 1),
+				},
+				UncappedTarget: apiv1.ResourceList{
+					apiv1.ResourceCPU:    *resource.NewScaledQuantity(10, 1),
+					apiv1.ResourceMemory: *resource.NewScaledQuantity(5000, 1),
+				},
+				LowerBound: apiv1.ResourceList{
```
kgolab: UpperBound is missing - on purpose?

vertical-pod-autoscaler/pkg/utils/vpa/capping_test.go
```diff
@@ -184,3 +184,57 @@ func TestApply(t *testing.T) {
...
+		apiv1.ResourceMemory: *resource.NewScaledQuantity(4500, 1),
+	}, res.ContainerRecommendations[0].Target)
+
+	assert.Equal(t, apiv1.ResourceList{
+		apiv1.ResourceCPU:    *resource.NewScaledQuantity(10, 1),
+		apiv1.ResourceMemory: *resource.NewScaledQuantity(5000, 1),
+	}, res.ContainerRecommendations[0].UncappedTarget)
+
+	assert.Equal(t, apiv1.ResourceList{
+		apiv1.ResourceCPU:    *resource.NewScaledQuantity(45, 1),
```
kgolab: Do we really expect LowerBound > Target ?

vertical-pod-autoscaler/pkg/utils/vpa/capping_test.go
```diff
@@ -184,3 +184,57 @@ func TestApply(t *testing.T) {
 		assert.Equal(t, testCase.ExpectedError, err)
 	}
 }
+
+func TestApplyVpa(t *testing.T) {
```
kgolab: The behaviour of this test is difficult to understand.
Could you please add an initial comment about what's supposed to happen (CPU limited up, mem limited down) - it will help future readers.

vertical-pod-autoscaler/pkg/recommender/routines/recommender.go
```diff
@@ -100,15 +100,21 @@ func (r *recommender) UpdateVPAs() {
...
 			})
-
 		}
 		had := vpa.HasRecommendation()
-		vpa.Recommendation = &vpa_types.RecommendedPodResources{containerResources}
+		recommendation := &vpa_types.RecommendedPodResources{containerResources}
+		cappedRecommendation, err := vpa_utils.ApplyVPAPolicy(recommendation, observedVpa.Spec.ResourcePolicy)
+		if err != nil {
+			glog.Errorf("Failed to apply policy for VPA %v/%v: %v", observedVpa.Namespace, observedVpa.Name, err)
+			cappedRecommendation = cappedRecommendation
```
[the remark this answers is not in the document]
bskiba: Fixed in the new version.

vertical-pod-autoscaler/pkg/recommender/routines/recommender.go
bskiba: I have extracted to a new method. Reads way better, thanks!

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
bskiba: Removed this change.

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
bskiba: Went with not reusing.

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
bskiba: Went for not reusing.

vertical-pod-autoscaler/pkg/utils/vpa/capping_test.go
bskiba: I have change the test a bit so that it should be more obvious. Let me know if you still want the comment.

vertical-pod-autoscaler/pkg/utils/vpa/capping_test.go
bskiba: Fixed.

bskiba:
Reworked a bit. I think this looks better now.

kgolab:
Thanks, it indeed looks better. LGTM

PR #1357, opened 2018-10-26
Modify execution_latency_seconds buckets

bskiba:


bskiba:
@kgolab

kgolab:
LGTM but I also think that if we care more about total execution time maybe a few point between 120 and 300 (150 & 240 perhaps?) could be useful.

bskiba:
Adde. Also added some more buckets for recommendation latency.

Issue #1182, opened 2018-08-23
Cap recommendation upperBound to some sane value.


bskiba:
When we start providing recommendation for a VPA object, initially the upperBound is ridiculously high. This is confusing to the users and provides no information. We should cap this value to sth sane.

Some ideas are to cap upperBound to target recommendation +X cores or target recommendation + XX% of cores (analogous for memory).

bskiba:
Need to choose a good capping value so that we will make sure this will not cause updater to start killing pods prematurely.

Rajat-0:
I would like to work on this

davidquarles:
I'm building/running current master and targeting a prometheus statefulset with no container memory limit and no `resourcePolicy` specified.  The pods were getting oom-killed for awhile before I deployed the vpa, if that's relevant.  Once stabilized, utilization was relatively low:
```
NAME           CPU(cores)   MEMORY(bytes)
prometheus-0   93m          1595Mi
prometheus-1   88m          1198Mi
```
but the recommender ended up with _very_ (unschedulably) high values for the memory request:
```
    - containerName: prometheus
      lowerBound:
        cpu: 92m
        memory: "2042845301"
      target:
        cpu: 109m
        memory: "1174275820239"
      uncappedTarget:
        cpu: 109m
        memory: "1174275820239"
      upperBound:
        cpu: 296m
        memory: "2762029041970"
```
I redeployed the vpa with a cap of 10Gi so that it would schedule and ingest some usage data, then tried removing the cap.  The recommendation target then returns to 1093Gi.  Does this sound correct, and is there any mitigation strategy outside of hard-capping, if so?  I'm not sure if i've done something wrong, or if it simply needs to be capped to function correctly today.

kgolab:
Hi David,

is this scenario easily reproducible?

We'd be very interested in more data, in particular what was the CPU & memory utilization from the very beginning and what were the initial VPA recommendations.
If possible could you please share logs & vpacheckpoint objects?

At the first glance this looks like a wrong recommendation but I'd love to see more so we can trace the issue.

davidquarles:
[...]

PR #1363, opened 2018-10-30
Use real-usage sample to estimate memory usage after OOM

kgolab:
Use real-usage sample (exclude previous OOMs) to estimate memory usage after OOM.

This should fix the problem reported by @davidquarles under #1182 - but *not* the initial issue reported by @bskiba there.

Note for a reviewer: I've unexported memoryPeak & oomPeak to decrease chance of invalid usage.


bskiba:
Please make golint happy, otherwise

/lgtm

kgolab:
Sorry for that, refactoring didn't fix the comment automatically :(

PR #1372, opened 2018-11-02
Add e2e tests for updater with different controllers.

bskiba:


bskiba:
@kgolab

vertical-pod-autoscaler/e2e/updater.go
```diff
@@ -82,18 +92,76 @@ var _ = UpdaterE2eDescribe("Updater", func() {
...
+}
+
+type getCurrentPodListFunc func() (*apiv1.PodList, error)
+
+func setupHamsterController(f *framework.Framework, controllerKind, cpu, memory string, replicas *int32) (getCurrentPodListFunc, *apiv1.PodList) {
+	switch controllerKind {
+	case "Deployment":
+		d, podList := setupHamsterDeployment(f, cpu, memory, replicas)
+		getCurrentPodList := func() (*apiv1.PodList, error) {
+			return framework.GetPodsForDeployment(f.ClientSet, d)
```
kgolab: I wonder if we could use getHamsterPods() everywhere.
Even if it's not the most efficient nor specific function, it would make this case contain only one-liners and getCurrentPodList would be constant.

vertical-pod-autoscaler/e2e/updater.go
```diff
@@ -113,6 +181,106 @@ func setupHamsterDeployment(f *framework.Framework, cpu, memory string, replicas
...
+	err = framework.WaitForAllJobPodsRunning(f.ClientSet, f.Namespace.Name, job.Name, parallelism)
+	podList, err := framework.GetJobPods(f.ClientSet, f.Namespace.Name, job.Name)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	return job, podList
+}
+
+func setupHamsterRS(f *framework.Framework, cpu, memory string, replicas *int32) (*appsv1.ReplicaSet, *apiv1.PodList) {
+	rs := framework.NewReplicaSet("hamster-rs", f.Namespace.Name, 3,
+		hamsterLabels, "", "")
+	if replicas != nil {
```
kgolab: Nit: I think it would be cleaner to use a construct similar to other functions so:
```
desiredReplicas := 3
if replicas != nill { desiredReplicas = *replicas }
...NewReplicaSet(..., desiredReplicas)

```

vertical-pod-autoscaler/e2e/updater.go
```diff
@@ -113,6 +181,106 @@ func setupHamsterDeployment(f *framework.Framework, cpu, memory string, replicas
...
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	podList, err := getHamsterPods(f)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	return rs, podList
+}
+
+func setupHamsterStateful(f *framework.Framework, cpu, memory string, replicas *int32) (*appsv1.StatefulSet,
+	*framework.StatefulSetTester, *apiv1.PodList) {
+	stateful := framework.NewStatefulSet("hamster-stateful", f.Namespace.Name,
+		"hamster-service", 3, nil, nil, hamsterLabels)
```
kgolab: Same comment as with setupHamsterRS.

And, btw, could you please extract 3 to a const? We have it in all setupHamsterXYZ methods and it always means "default number of replicas".
Even better maybe we could require this number to be always set and just populate it from testRestartsPods or setupHamsterController. This would simplify the awkward call in "observes pod disruption budget" test too.

vertical-pod-autoscaler/e2e/updater.go
```diff
@@ -21,36 +21,47 @@ import (
...
 	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta1"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
+	testutils "k8s.io/kubernetes/test/utils"
 
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/gomega"
 )
 
+var hamsterLabels = map[string]string{"app": "hamster"}
```
kgolab: It's nice that you've extracted it but it still needs to be in line with NewHamsterDeployment.
Or could we add this in common.go and use everywhere?

vertical-pod-autoscaler/e2e/updater.go
```diff
@@ -152,48 +320,91 @@ func setupPDB(f *framework.Framework, name string, maxUnavailable int) *policyv1
...
+	return getPodSetChanges(c, currentPodSet, initialPodSet) > 0
 }
 
-func getPodSetChanges(c clientset.Interface, deployment *appsv1.Deployment, initialPodSet map[string]bool) (int, error) {
-	currentPodList, err := framework.GetPodsForDeployment(c, deployment)
-	if err != nil {
-		return 0, err
-	}
-	currentPodSet := makePodSet(currentPodList)
+func getPodSetChanges(c clientset.Interface, currentPodSet podSet, initialPodSet podSet) int {
```
kgolab: Based by the function usage it might not be needed but based on the name itself I'd also expect new pods that are in currentPodSet to count towards diff (e.g. there is a new name in currentPodSet and no such name in initialPodSet).

Since that would change the expected results maybe we want a different name for the function?
Sadly the only one I see is getChangesFromInitialSet but maybe you can come up with something better.

vertical-pod-autoscaler/e2e/updater.go
```diff
@@ -152,48 +320,91 @@ func setupPDB(f *framework.Framework, name string, maxUnavailable int) *policyv1
...
+	return diffs
+}
+
+func createReplicaSetWithRetries(c clientset.Interface, namespace string, obj *appsv1.ReplicaSet) error {
+	if obj == nil {
+		return fmt.Errorf("Object provided to create is empty")
+	}
+	createFunc := func() (bool, error) {
+		_, err := c.AppsV1().ReplicaSets(namespace).Create(obj)
+		if err == nil || apierrs.IsAlreadyExists(err) {
```
kgolab: I can see that this is a general way to create test resources but aren't we afraid that an existing ReplicaSet could signify an error (conflicting definitions could or could not influence our test)?

kgolab:
Thanks for these tests.
I've left a few comments about possible improvements, happy to do them myself in a separate PR if you don't have time.

vertical-pod-autoscaler/e2e/updater.go
bskiba: There should be no conflicting definitions. Even if cleanup from a previous test fails, every test sets up a new namespace.

vertical-pod-autoscaler/e2e/updater.go
bskiba: Done.

vertical-pod-autoscaler/e2e/updater.go
bskiba: Done

vertical-pod-autoscaler/e2e/updater.go
bskiba: Done

vertical-pod-autoscaler/e2e/updater.go
```diff
@@ -113,6 +181,106 @@ func setupHamsterDeployment(f *framework.Framework, cpu, memory string, replicas
...
+		Name:  "hamster",
+		Image: "k8s.gcr.io/ubuntu-slim:0.1",
+		Resources: apiv1.ResourceRequirements{
+			Requests: apiv1.ResourceList{
+				apiv1.ResourceCPU:    cpuQuantity,
+				apiv1.ResourceMemory: memoryQuantity,
+			},
+		},
+		Command: []string{"/bin/sh"},
+		Args:    []string{"-c", "while true; do timeout 0.5s yes >/dev/null; sleep 0.5s; done"},
```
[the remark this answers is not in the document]
bskiba: Done

vertical-pod-autoscaler/e2e/updater.go
bskiba: Done

vertical-pod-autoscaler/e2e/updater.go
bskiba: Changed naming WDYT?

bskiba:
Addressed almost all comments. PTAL and let me know if this looks good. If yes, I'll squash and merge.

vertical-pod-autoscaler/e2e/updater.go
kgolab: Thanks, getRestartedPodsCount sounds good.

kgolab:
Thanks, LGTM now.

PR #1376, opened 2018-11-05
Build VPA releases in docker

bskiba:
@MaciekPytel

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/Makefile
+++ b/vertical-pod-autoscaler/pkg/admission-controller/Makefile
@@ -1,33 +1,59 @@
 all: build
 
-TAG?=v0.0.1
-REGISTRY?=k8s.gcr.io
+TAG?=dev
+REGISTRY?=staging-k8s.gcr.io
 FLAGS=
 ENVVAR=
 GOOS?=linux
+COMPONENT=admission-controller
+FULL_COMPONENT=vpa-${COMPONENT}
 
 deps:
 	go get github.com/tools/godep
 
 build: clean deps
 	$(ENVVAR) GOOS=$(GOOS) godep go build ./...
-	$(ENVVAR) GOOS=$(GOOS) godep go build -o admission-controller
+	$(ENVVAR) GOOS=$(GOOS) godep go build -o ${COMPONENT}
+
+build-binary: clean deps
+	$(ENVVAR) GOOS=$(GOOS) godep go build -o ${COMPONENT}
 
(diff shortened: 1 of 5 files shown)


vertical-pod-autoscaler/pkg/recommender/Makefile
```diff
@@ -1,33 +1,59 @@
...
-	docker build --pull -t ${REGISTRY}/vpa-recommender:${TAG} .
-	docker push ${REGISTRY}/vpa-recommender:${TAG}
+ifndef TAG
+	ERR = $(error TAG is undefined)
+	$(ERR)
+endif
+	docker push ${REGISTRY}/${FULL_COMPONENT}:${TAG}
+
+docker-builder:
+	docker build -t autoscaling-builder ../../../builder
```
aleksandra-malinowska: Using the same builder for CA and VPA will mean we always have to do things like upgrade Go version simultaneously etc. Do we want to do this?

bskiba:
@aleksandra-malinowska @MaciekPytel does this look resonable now?

vertical-pod-autoscaler/pkg/admission-controller/Makefile
```diff
@@ -1,33 +1,59 @@
...
+	ERR = $(error TAG is undefined)
+	$(ERR)
+endif
+	docker push ${REGISTRY}/${FULL_COMPONENT}:${TAG}
+
+docker-builder:
+	docker build -t vpa-autoscaling-builder ../../builder
+
+build-in-docker: clean docker-builder
+	docker run -v `pwd`/../..:/gopath/src/k8s.io/autoscaler/vertical-pod-autoscaler vpa-autoscaling-builder:latest bash -c 'cd /gopath/src/k8s.io/autoscaler/vertical-pod-autoscaler && make build-binary -C pkg/admission-controller'
```
aleksandra-malinowska: Is it intentional to not parametrize `admission-controller` here?

aleksandra-malinowska: Looking at the rest of the Makefiles, I guess my question is more if there's any difference between them except this? And why parametrize them using COMPONENT if they're forked anyway?

bskiba: No specific reason I guess I can inline everywhere if that's more readable

aleksandra-malinowska: Ack, as you prefer

PR #1378, opened 2018-11-05
AWS: Improved balancing

johanneswuerbach:
Various ClusterAutoscaler components like the BalancingProcessor act based on the current `TargetSize` of a `NodeGroup`. https://github.com/kubernetes/autoscaler/blob/f341d8a68a6aad996964cda5db67e2f254b71c8d/cluster-autoscaler/processors/nodegroupset/balancing_processor.go#L89

As the current AWS implementation only updates `TargetSize` when refreshing the ASG list every minutes (less when being rate-limited), the cached `TargetSize` might quickly mismatch the actual value.

While each ASG change could trigger a cache refresh, I don't think this is optimal as it introduces again further requests.

Instead I decided to pre-calculate the `TargetSize` change for AWS on AutoScalingGroup changes and Node deletion to make better balancing decisions.

Any wrongly pre-calculated `TargetSize` will be auto-corrected on the next successful ASG list refresh.

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
+++ b/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
@@ -19,6 +19,7 @@ package aws
 import (
 	"fmt"
 	"reflect"
+	"strings"
 	"sync"
 
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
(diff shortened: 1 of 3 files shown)


johanneswuerbach:
Friendly ping to @bskiba or @aleksandra-malinowska, a review would be appreciated. I'm running this change in one of our production since ~7 days and the balancing looks much better.

losipiuk:
I am not that much familiar with AWS code. Looks fine aftern skimming though.

/lgtm
/approve

losipiuk:
Thanks @johanneswuerbach

PR #1386, opened 2018-11-06
Update RBAC example to include replicasets in the apps apigroup

zegl:
This is needed as of Kubernetes 1.12


losipiuk:
Looks fine though it seems a good cleanup would be useful for all those example files.

/lgtm
/approve

PR #1391, opened 2018-11-08
Migrate pod_nanny to fetch only nodes metadata.

mborsz:
This pod requires only number of nodes so we can fetch only nodes
metadata to reduce resource consumption and latency.
Tests on 5000-node cluster show that this change reduces 'GET /api/v1/nodes' latency from ~5s to ~100ms.


mborsz:
I ran verify locally and had to do small fixes.

The problem with travis seems to be unrelated to the change, according to google it's rather related to the fact we are still using unsupported go 1.8: https://github.com/apache/incubator-openwhisk-cli/issues/376

wojtek-t:
That means, that to be able to merge this PR, we should probably first upgrade the Go version. That should hopefully be pretty simple.

mborsz:
Created #1392, once it's merged I will rebase this PR.

MaciekPytel:
Ping me once @wojtek-t lgtms this and I'll merge it.

mborsz:
Travis is happy now,

wojtek-t PTAL?

PR #1392, opened 2018-11-08
Migrate to go1.10

mborsz:
Go 1.8 is no longer supported and travis is not able to run presubmit with it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ language: go
 
 matrix:
   include:
-    - go: 1.8
+    - go: "1.10"
 
 install:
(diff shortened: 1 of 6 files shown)


mborsz:
Context: #1391

MaciekPytel:
Merging manually since we never cherry-picked owner files on this branch. @mborsz Are you planning a lot of development here? If so we should probably fix the problem with owners files.

mborsz:
Thanks a lot Maciek, I'm planing to do only a small fix here: https://github.com/kubernetes/autoscaler/pull/1391

PR #1399, opened 2018-11-13
Clarify down-scale capabilities

nikopen:
Wasn't clear by initially reading the readme.

I'd also like to add specifics on the time intervals (configurable or not? etc), please let me know how that works.


bskiba:
Regarding the time intervals, there is no specific time after which autoscaler kicks in. If your app is severely over-requesting resources, the downscale should happen quite fast.

nikopen:
@bskiba should I open another docs PR specifying that VPA acts as fast as the 'severity' of the discrepancy between usage-over-time and the pinned requests?

also, are there perhaps plans to make the timing configurable?

monotek:
Any update on this?

PR #1400, opened 2018-11-13
Inject Backoff instance to ClusterStateRegistry on creation

losipiuk:


MaciekPytel:
Looks good, but I wonder - if we're allowing a custom Backoff shouldn't we put it into context as well? Seems like a useful thing to have available in node-related processors.

MaciekPytel:
After discussing offline @losipiuk convinced me it can be easily passed on initialization as needed. We're not very consistent on what we pass through context and what we don't, but this is a separate issue that deserves a larger discussion and a separate PR to clean up. 

/lgtm
/approve

PR #1407, opened 2018-11-16
add flag to ignore daemonsets when calculating resource utilization of a node

awprice:
Adds the flag `--ignore-daemon-set-pods-in-utilization` (defaults to false) and when enabled, factors DaemonSet pods out when calculating the resource utilization of a node. This is useful when there are only system or metrics DaemonSets on the node that are not user workloads.

Not sure how to test this as there don't seem to be any other existing tests that test the functionality of the flags.

### DIFF
--- a/cluster-autoscaler/simulator/cluster_test.go
+++ b/cluster-autoscaler/simulator/cluster_test.go
@@ -38,14 +38,42 @@ func TestUtilization(t *testing.T) {
 	node := BuildTestNode("node1", 2000, 2000000)
 	SetNodeReadyState(node, true, time.Time{})
 
-	utilInfo, err := CalculateUtilization(node, nodeInfo)
+	utilInfo, err := CalculateUtilization(node, nodeInfo, false, false)
 	assert.NoError(t, err)
 	assert.InEpsilon(t, 2.0/10, utilInfo.Utilization, 0.01)
 
 	node2 := BuildTestNode("node1", 2000, -1)
 
-	_, err = CalculateUtilization(node2, nodeInfo)
+	_, err = CalculateUtilization(node2, nodeInfo, false, false)
 	assert.Error(t, err)
+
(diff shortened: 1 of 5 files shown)


awprice:
@frittentheke Yea I agree with you that they shouldn't be considered, but I went with default false so the default behaviour of the cluster-autoscaler doesn't change when this version is used. 

Happy to change it to true if the maintainers see it as okay.

awprice:
@losipiuk Thanks for the comments.

I've updated the PR with the changes in the review.

awprice:
@losipiuk Is it possible for this to reviewed soon? We would love to be able to get this feature into the 1.13.0 release of cluster-autoscaler. Thanks!

cluster-autoscaler/simulator/cluster_test.go
```diff
@@ -38,13 +38,13 @@ func TestUtilization(t *testing.T) {
...
 
-	utilInfo, err := CalculateUtilization(node, nodeInfo)
+	utilInfo, err := CalculateUtilization(node, nodeInfo, false)
 	assert.NoError(t, err)
 	assert.InEpsilon(t, 2.0/10, utilInfo.Utilization, 0.01)
 
 	node2 := BuildTestNode("node1", 2000, -1)
 
-	_, err = CalculateUtilization(node2, nodeInfo)
+	_, err = CalculateUtilization(node2, nodeInfo, false)
```
aleksandra-malinowska: Can you add a test case for ignoring DaemonSet pod?

awprice: 👍 Tests added for ignoring DaemonSet pods

awprice:
@aleksandra-malinowska Thanks for your feedback. I've updated the PR with the changes you have requested. There is also now a flag (and unit tests) for ignoring mirror pods as well.

Both flags are now in the "ignore-<type>-utilization" format and default to false to match the current behaviour of cluster-autoscaler where everything is included.

PR #1413, opened 2018-11-18
pass kubeconfig to cloudprovider builder

multi-io:
We have an internal cloud provider here (which we're planning to publish in a separate PR later) that uses the K8s cluster-api for creating nodes and thus needs to access to the cluster itself (its CloudConfig is just the cluster kubeconfig). As a preparation for that, this PR changes AutoscalerOptions and the cloud provider builder so we pass the kubeconfig down to the `buildCloudProvider` in case any actual cloud provider needs it.


losipiuk:
Hi @multi-io!

I am not a big fan of preparatory changes with no use in the codebase. I would rather see the change bundled with the PR which adds cloudprovider you mentioned. I suggest the change lives on your fork until then.

losipiuk:
Closing for now. Please send out new PR where usages are bundled with changes.

Issue #1411, opened 2018-11-16
Crash when Autoscaler starts and AutoscalinGroups have 0 instances


gtie:
We are using Cluster Autoscaler (CA) to right-size several AutoScalingGroups in AWS (each featuring different AZ/taint combination). When the leader encouters such an autoscaling group, it crashes with SIGSEGV (and then, next leader does the same, and so on).

Here is the traceback:
```
I1116 18:30:35.877019       1 request.go:485] Throttling request took 90.934181ms, request: GET:https://100.64.0.1:443/api/v1/namespaces/kube-system/configmaps/cluster-autoscaler-status
I1116 18:30:36.077059       1 request.go:485] Throttling request took 196.99294ms, request: PUT:https://100.64.0.1:443/api/v1/namespaces/kube-system/configmaps/cluster-autoscaler-status
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x134867f]

goroutine 119 [running]:
k8s.io/autoscaler/cluster-autoscaler/cloudprovider/aws.(*AwsManager).buildNodeFromTemplate(0xc421932bd0, 0xc42359f6c0, 0xc422d9a7c0, 0x0, 0x0, 0xc4206fe750)
        /gopath/src/k8s.io/autoscaler/cluster-autoscaler/cloudprovider/aws/aws_manager.go:243 +0x3df
k8s.io/autoscaler/cluster-autoscaler/cloudprovider/aws.(*AwsNodeGroup).TemplateNodeInfo(0xc4208737e0, 0xc421784e70, 0xc423623c00, 0x40)
        /gopath/src/k8s.io/autoscaler/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go:303 +0x68
k8s.io/autoscaler/cluster-autoscaler/core.GetNodeInfosForGroups(0xc421b52b00, 0xa, 0x10, 0x3b71440, 0xc4235f9c10, 0x3b932e0, 0xc420372000, 0xc422578940, 0x5, 0x8, ...)
        /gopath/src/k8s.io/autoscaler/cluster-autoscaler/core/utils.go:258 +0x296
[...]

PR #1422, opened 2018-11-19
Update AWS EC2 instance type catalog

seh:
Run _make generate_ to pull and translate a fresh catalog of [AWS's available EC2 instance types](https://aws.amazon.com/ec2/instance-types/) used by the cluster autoscaler.

This addresses #1411 but isn't a complete solution, as [we should also trap these instance type lookup failures](https://github.com/kubernetes/autoscaler/issues/1411#issuecomment-439999208).

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/ec2_instance_types.go
+++ b/cluster-autoscaler/cloudprovider/aws/ec2_instance_types.go
@@ -255,6 +255,12 @@ var InstanceTypes = map[string]*instanceType{
 		MemoryMb:     124928,
 		GPU:          0,
 	},
+	"f1.4xlarge": {
+		InstanceType: "f1.4xlarge",
+		VCPU:         16,
+		MemoryMb:     249856,
+		GPU:          0,
+	},
 	"g2": {
 		InstanceType: "g2",
 		VCPU:         32,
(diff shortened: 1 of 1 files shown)


aleksandra-malinowska:
/lgtm
/approve

> This addresses #1411 but isn't a complete solution, as we should also trap these instance type lookup failures.

If you have a solution in mind (e.g. go with whatever resources other nodes in this group have if there's no entry, ignore this group, get list of available instance types from API, or anything that makes sense for this use-case), it would be great. It seems rather inconvenient that autoscaler code on all supported branches has to have all instance types hard-coded, and breaks whenever a new one is used.

seh:
> If you have a solution in mind (e.g. go with whatever resources other nodes in this group have if there's no entry, ignore this group, get list of available instance types from API, or anything that makes sense for this use-case), it would be great.

So far the approach I've taken is to return an indicative error, as it wasn't clear to me how to continue there without a rather drastic change to the code—as I imagine fetching a refreshed list of instance types would involve. Having said that, though, I don't think AWS would let anyone create a Launch Template using an unknown instance type, so if the autoscaler runs into a type it's never heard of, it's most likely that it's the autoscaler that's out of date, and not that it's receiving invalid input from AWS.

I'll post a patch for your consideration shortly.

seh:
> I'll post a patch for your consideration shortly.

Please see #1425.

PR #1423, opened 2018-11-20
Run feature gates based logic to fix consistency of CA and scheduler

losipiuk:


cluster-autoscaler/simulator/predicates.go
```diff
@@ -56,6 +56,15 @@ type PredicateChecker struct {
 // There are no const arrays in Go, this is meant to be used as a const.
 var priorityPredicates = []string{"PodFitsResources", "GeneralPredicates", "PodToleratesNodeTaints"}
 
+func init() {
+	// This results in filtering out some predicate functions registered by defaults.init() method.
+	// In scheduler this method is run from app.runCommand().
+	// We also need to call it in CA to have simulation behaviour consistent with scheduler.
+	// Note: the logic of method is conditional and depends on feature gates enabled. To have same
```
aleksandra-malinowska: s/on feature gates/on default feature gates

AFAIU we aren't using exactly the same set of feature gates as they may have been changed by flag in kube-controller-manager. Can we perhaps add a link to where the defaults come from? For future maintainers' sake...

PR #1432, opened 2018-11-22
Update examples and deployments for new version.

bskiba:


bskiba:
@kgolab

bskiba:
VPA will no longer use it. But it will be needed for proper cleanup when switching from alpha to beta so bringing it back. Thanks/

vertical-pod-autoscaler/hack/vpa-process-yamls.sh
```diff
@@ -39,7 +39,7 @@ if [ $# -gt 2 ]; then
 fi
 
 ACTION=$1
-COMPONENTS="vpa-crd vpa-beta-crd vpa-rbac updater-deployment recommender-deployment admission-controller-deployment"
+COMPONENTS="vpa-beta-crd vpa-rbac updater-deployment recommender-deployment admission-controller-deployment"
```
kgolab: I think we need to decide whether VPA beta should take care about switching off VPA alpha.

If no - vpa-crd.yaml can be removed after all (unless we want to keep it for manual clean-up - but then this needs to be documented somewhere).
If yes - we need to add vpa-crd to COMPONENTS if ACTION==delete.

kgolab:
We also have an example in README.md - are you planning to update these separately?

vertical-pod-autoscaler/hack/vpa-process-yamls.sh
bskiba: I plan on adding a switch-me-to-beta script or instruction. Added vpa-crd on delete, thanks.

bskiba:
Yup, I plan to send a separate PR with changes to docs.

kgolab:
OK, then LGTM.

PR #1438, opened 2018-11-22
Ignore starting nodes in node group health

aleksandra-malinowska:
When a user (or test) manually creates new nodes, or if Cluster Autoscaler restarts and loses track of scale-ups it requested, it may consider a node group unhealthy because of upcoming nodes. This is inconsistent with how cluster state is computed and with how scale-up expansion options are calculated, and most likely not what we want to do.

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -323,7 +323,7 @@ func (csr *ClusterStateRegistry) IsNodeGroupHealthy(nodeGroupName string) bool {
 	unjustifiedUnready := 0
 	// Too few nodes, something is missing. Below the expected node count.
 	if readiness.Ready < acceptable.MinNodes {
-		unjustifiedUnready += acceptable.MinNodes - readiness.Ready
+		unjustifiedUnready += acceptable.MinNodes - readiness.Ready - readiness.NotStarted
 	}
 	// TODO: verify against max nodes as well.
 


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -323,7 +323,7 @@ func (csr *ClusterStateRegistry) IsNodeGroupHealthy(nodeGroupName string) bool {
 	unjustifiedUnready := 0
 	// Too few nodes, something is missing. Below the expected node count.
 	if readiness.Ready < acceptable.MinNodes {
-		unjustifiedUnready += acceptable.MinNodes - readiness.Ready
+		unjustifiedUnready += acceptable.MinNodes - readiness.Ready - readiness.NotStarted
```
losipiuk: It seems like with this change in common case (when CA was not restarted, and was originator of the change of size of MIG) we will account for upcoming nodes twice.
Once in updateAcceptableRanges when MinNodes is computed. And then here.

aleksandra-malinowska: Right. It's probably less harmful this way (at most, we'll consider the node group healthy a little while longer while waiting for scale up), but how about using max from {scaleUpRequests, notStarted} when computing minNodes?

https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/clusterstate/clusterstate.go#L482

losipiuk: It would not change anything in common case, as scaleUpRequest would only disappear if all the requested nodes are up, right?

mwielgus:
PR Close warning. 

@aleksandra-malinowska  @MaciekPytel  Please decide whether you want this change or not. With no decision I will close the PR tomorrow (you will still be able to reopen it once you make up your mind).

PR #1446, opened 2018-11-26
Update cluster-autoscaler godeps on master

losipiuk:


aleksandra-malinowska:
kubernetes.sync

losipiuk:
https://github.com/kubernetes/autoscaler/pull/1446/files#diff-1bb6f9325d0fa20db0ad737ee0ff9cfa

feiskyer:
@losipiuk @aleksandra-malinowska Any reasons for rolling the kubernetes version back? Just noticed the master branch is still using vendor from kubernetes v1.12.

losipiuk:
> Any reasons for rolling the kubernetes version back? Just noticed the master branch is still using vendor from kubernetes v1.12.

I am not sure now but I do not think it was intentional. We should update vendor on master against the newest k8s release there is right now. 
Probably `v1.15.0-alpha.0` is the way to go right now.

feiskyer:
OK, let me do that.

feiskyer:
Just find an issue because of old-versioned packages.

PR #1447, opened 2018-11-26
Add OWNERS to /builder

aleksandra-malinowska:
Builder is used only for Cluster Autoscaler. It'll be useful if more maintainers can e.g. update Go version there. Long-term, we should probably move it to /cluster-autoscaler though.

### DIFF
--- a/builder/OWNERS
+++ b/builder/OWNERS
@@ -1,0 +1,10 @@
+approvers:
+- aleksandra-malinowska
+- losipiuk
+- maciekpytel
+- mwielgus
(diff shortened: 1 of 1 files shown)


MaciekPytel:
/approve
/lgtm
@bskiba Are you ok with that? I think it doesn't impact VPA in any way.
/hold

bskiba:
/lgtm 
We have the separate builder now. You might consider moving into cluster-autoscaler and noting that the builder is CA specific though.

PR #1464, opened 2018-11-29
Better quota-exceeded/stockout handling

losipiuk:


cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -274,16 +326,47 @@ func (mig *gceMig) Debug() string {
...
+	for _, instanceInfo := range instanceInfos {
+		instance := cloudprovider.Instance{
+			Id: instanceInfo.GceRef.ToProviderId(),
+		}
+		instance.Status = &cloudprovider.InstanceStatus{}
+
+		switch instanceInfo.Status {
+
+		case RUNNING:
+			instance.Status.State = cloudprovider.STATE_RUNNING
```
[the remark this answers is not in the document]
losipiuk: I wanted to have separate layer of GCE related types returned by `GceManager` and the types used in `cloudprovider` interface.

I agree they are very similar but it feels cleaner to me.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -398,6 +403,93 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
+		if !found {
+			continue
+		}
+
+		outOfResourcesInstancesByErrorCode := make(map[string][]cloudprovider.Instance)
+		for _, instance := range instances {
+			if instance.Status != nil && instance.Status.State == cloudprovider.StateCreating && instance.Status.ErrorInfo != nil {
+				errorInfo := instance.Status.ErrorInfo
+				if errorInfo.ErrorClass == cloudprovider.ErrorOutOfResources {
+					errorInstances, found := outOfResourcesInstancesByErrorCode[errorInfo.ErrorCode]
```
aleksandra-malinowska: why not just:

`errorInstancesByCode[errorInfo.ErrorCode] = append(errorInstancesByCode[errorInfo.ErrorCode], errorInstance)`

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -274,6 +299,15 @@ func (csr *ClusterStateRegistry) Recalculate() {
 	csr.updateAcceptableRanges(targetSizes)
 }
 
+// Cleanup cleans up internal structures to avoid memory leaks. Should be called periodically.
+func (csr *ClusterStateRegistry) Cleanup(currentTime time.Time) {
+	for nodeName, requestTime := range csr.nodeDeleteRequested {
+		if requestTime.Add(MaxNodeStartupTime).Before(currentTime) {
```
aleksandra-malinowska: Why use MaxNodeStartupTime here? It has nothing to do with node startup, right?

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -398,6 +403,93 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
+
+	// Issue a delete
+	err := nodeGroup.DeleteNodes(nodes)
+	if err != nil {
+		klog.Warningf("Error while trying to delete nodes from %v: %v", nodeGroup.Id(), err)
+		return false
+	}
+
+	// Remember the fact that we deleted the nodes
+	// Question: Maybe for safety we should mark it in the loop above.
```
aleksandra-malinowska: That's an interesting problem. I see what you're doing here - only on successful delete call to cloud provider we'll update failed scale ups metric and back-off, which will hopefully mean we do it just once... unless we restart. But if the instance is stuck and can't be deleted, it means we'll never back-off the node group at all. On the other hand, we'll keep publishing events about it every loop.

A side-effect is also that if the instance delete fails, we'll keep retrying every loop, which may not be the best idea if there's a problem with the underlying service which may be limited to a specific zone.

I'm not sure if it's worth the added complexity to attempt to solve this 100% correctly, but perhaps we can consider the priorities to make the right trade-offs?

The way I see, from most to least important:
- don't get stuck trying to delete the same instance every loop if it doesn't work,
- back-off the group ASAP,
- don't update the metric twice if possible (loss of information is somehow more acceptable, since it'll be the case on restarts anyway - but let's not do false positives),
- everything else (events, metric loss).

I wonder if something like this may work:
- as soon as errors are recognized, report instance ids to cluster state and save,
- on first seeing a given instance id, back-off the node group,
- on first seeing a given instance id, update the metric but only if there's a scale up in progress,
- delete instances in a way that won't block us (i.e. if it fails don't retry in every iteration, scale up other groups in the meantime).

Still figuring out how to fit updating scale up information in it. My initial thought was that we should also do it ASAP (when updating metric), but then we'd have to triple-check how it affects everything else.

aleksandra-malinowska: OK, so it seems we ignore that error and don't fail the loop, so we'll keep scaling the rest of the cluster no matter what (which is good).

I still think we should back-off even if we don't manage to delete the node for whatever reason (e.g. because the group is at or below minimum), and publish metric if there was an active scale-up attempt.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -398,6 +403,93 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 	return nil
 }
 
+func (a *StaticAutoscaler) handleQuoteAndStockoutErrors(currentTime time.Time) {
+	cloudProviderNodeInstances := a.clusterStateRegistry.GetCloudProviderNodeInstances()
+
+	nodeGroups := a.CloudProvider.NodeGroups()
+	for _, nodeGroup := range nodeGroups {
```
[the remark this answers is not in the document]
aleksandra-malinowska: Fair point, nevermind then.

cluster-autoscaler/core/static_autoscaler.go
aleksandra-malinowska: I think the logic with saving last-seen-instances in CSR can work for this. And I really think we should only do failed scale up stuff (metrics, events, etc.) IF there's a scale up in progress in that group, and update that scale up soon-ish.

That way, if a user resizes group, but we don't know about it, we'll not start reporting failed scale up. And on restart, we'll also not do that (because we lost scale up state).

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -166,22 +167,67 @@ func (client *autoscalingGceClientV1) DeleteInstances(migRef GceRef, instances [
...
+				} else if isQuotaExceededErrorCoce(instanceError.Code) {
+					info.ErrorClass = QUOTA_EXCEEDED
+				} else if info.ErrorClass == NONE {
+					info.ErrorClass = OTHER
+				}
+				if instanceError.Message != "" {
+					if info.ErrorMessage != "" {
+						info.ErrorMessage += "; "
+					}
+					info.ErrorMessage += instanceError.Message
```
MaciekPytel: It's probably premature optimization, but it hurts me to see string being build this way. 

strings.Builder?

aleksandra-malinowska: Out of curiosity, is gathering a list and calling strings.Join significantly terribly inefficient? We seem to use that a lot

MaciekPytel: Nah, it's ok. I just didn't notice your comment as it was already marked as resolved. I consider both ways equally good, I just thought of Builder first.

Python developer in me feels a bit ashamed about that :s

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -156,6 +161,53 @@ type gceMig struct {
 	maxSize    int
 }
 
+// InstanceStatus represents status of instance. It collapses more verbose list of statuses from
+// GCE to just three values (running, being created, being deleted)
+type InstanceStatus uint64
```
MaciekPytel: Is there any reason not to use cloudprovider.InstanceState? They seem identical.

MaciekPytel: Honestly all the data structures here look like a duplicate of stuff you already added at cloudprovider level. The only difference I see is distinguishing between quota and stockout and I'm not sure how valuable that is.

cluster-autoscaler/cloudprovider/gce/gce_manager_test.go
```diff
@@ -346,10 +346,10 @@ func TestDeleteInstances(t *testing.T) {
...
-	server.On("handle", "/project1/zones/us-central1-b/instanceGroupManagers/gke-cluster-1-extra-pool-323233232").Return(getInstanceGroupManager(zoneB)).Once()
-	server.On("handle", "/project1/zones/us-central1-b/instanceGroupManagers/gke-cluster-1-extra-pool-323233232/listManagedInstances").Return(getManagedInstancesResponse2(zoneB)).Once()
+	server.On("handle", "/project1/zones/us-central1-b/instanceGroupManagers/gke-cluster-1-default-pool").Return(buildDefaultInstanceGroupManagerResponse(zoneB)).Once()
+	server.On("handle", "/project1/zones/us-central1-b/instanceGroupManagers/gke-cluster-1-default-pool/listManagedInstances").Return(buildFourRunningInstancesOnDefaultMigManagedInstancesResponse(zoneB)).Once()
+	server.On("handle", "/project1/zones/us-central1-b/instanceGroupManagers/gke-cluster-1-extra-pool-323233232").Return(buildDefaultInstanceGroupManagerResponse(zoneB)).Once()
+	server.On("handle", "/project1/zones/us-central1-b/instanceGroupManagers/gke-cluster-1-extra-pool-323233232/listManagedInstances").Return(buildOneRunningInstanceOnExtraPoolMigManagedInstancesResponse(zoneB)).Once()
```
MaciekPytel: Have I ever mentioned how much I hate everything about this file? Just looking at this is painful. 

I don't mean you're making anything wrong or breaking anything. This stuff is just so ugly.

aleksandra-malinowska: I was going to refactor these tests as after GCE & GKE cloud provider remake this is no longer necessary, but it turned out we don't have time for it :(

cluster-autoscaler/cloudprovider/gke/gke_manager.go
```diff
@@ -449,7 +449,8 @@ func (m *gkeManagerImpl) GetMigNodes(mig gce.Mig) ([]string, error) {
 		return []string{}, err
 	}
 	result := make([]string, 0)
-	for _, ref := range instances {
+	for _, instance := range instances {
+		ref := instance.GceRef
```
MaciekPytel: You added utility function for building this string. Why not use it here?

cluster-autoscaler/core/static_autoscaler.go
losipiuk: Updated

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -151,24 +156,43 @@ func NewClusterStateRegistry(cloudProvider cloudprovider.CloudProvider, config C
 		nodeGroupBackoffInfo:    backoff,
 		lastStatus:              emptyStatus,
 		logRecorder:             logRecorder,
+		nodeDeleteRequested:     make(map[string]time.Time),
```
MaciekPytel: Shouldn't you initialize cloudProviderNodeInstances here?

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -154,21 +156,39 @@ func NewClusterStateRegistry(cloudProvider cloudprovider.CloudProvider, config C
 	}
 }
 
-// RegisterScaleUp registers scale up.
-func (csr *ClusterStateRegistry) RegisterScaleUp(request *ScaleUpRequest) {
+// RegisterOrUpdateScaleUp registers scale-up for give node group or changes requested node increase
+// count.
+// If increaseDelta is positive then number of new nodes requested is increased; Time and expectedAddTime
+// are reset.
+// If increaseDelete is negative the number of new nodes requested is decreased; Time and expectedAddTime are
```
MaciekPytel: s/Delete/Delta

More importantly I find the possibility of having negative `increaseDelta` very non-intuitive. Can we just name it delta or something?

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -154,21 +156,39 @@ func NewClusterStateRegistry(cloudProvider cloudprovider.CloudProvider, config C
...
+		}
+		csr.scaleUpRequests[nodeGroup.Id()] = scaleUpRequest
 	}
 
 	// update the old request
-	oldScaleUpRequest.Time = request.Time
-	oldScaleUpRequest.ExpectedAddTime = request.ExpectedAddTime
-	oldScaleUpRequest.Increase += request.Increase
+	if scaleUpRequest.Increase+increaseDelta <= 0 {
+		// increase <= 0 means that there is no scale-up intent really
```
MaciekPytel: Do we expect to get negative numbers here? Should we just silently ignore it if we do?

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -154,21 +156,39 @@ func NewClusterStateRegistry(cloudProvider cloudprovider.CloudProvider, config C
...
+	scaleUpRequest, found := csr.scaleUpRequests[nodeGroup.Id()]
 	if !found {
-		csr.scaleUpRequests[request.NodeGroup.Id()] = request
-		return
+		scaleUpRequest = &ScaleUpRequest{
+			NodeGroup:       nodeGroup,
+			Time:            currentTime,
+			ExpectedAddTime: currentTime.Add(csr.config.MaxNodeProvisionTime),
+		}
+		csr.scaleUpRequests[nodeGroup.Id()] = scaleUpRequest
```
MaciekPytel: This looks very much like it should return here.

MaciekPytel: Or maybe not. But it's a bit tricky.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -154,21 +156,39 @@ func NewClusterStateRegistry(cloudProvider cloudprovider.CloudProvider, config C
...
 
-	oldScaleUpRequest, found := csr.scaleUpRequests[request.NodeGroup.Id()]
+	scaleUpRequest, found := csr.scaleUpRequests[nodeGroup.Id()]
 	if !found {
-		csr.scaleUpRequests[request.NodeGroup.Id()] = request
-		return
+		scaleUpRequest = &ScaleUpRequest{
+			NodeGroup:       nodeGroup,
+			Time:            currentTime,
+			ExpectedAddTime: currentTime.Add(csr.config.MaxNodeProvisionTime),
```
MaciekPytel: You don't actually need to set anything other than NodeGroup.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -398,6 +403,110 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
+func (a *StaticAutoscaler) handleQuotaAndStockoutErrorsForNodeGroup(
+	nodeGroup cloudprovider.NodeGroup,
+	currentInstances []cloudprovider.Instance,
+	previousInstances []cloudprovider.Instance,
+	currentTime time.Time) {
+
+	_, currentErrorCodeToInstance := a.buildInstanceToOutOfResourcesErrorCodeMappings(currentInstances)
+	previousInstanceToErrorCode, _ := a.buildInstanceToOutOfResourcesErrorCodeMappings(previousInstances)
+
+	// If node group is scaling up and there are new node-create requests which cannot be satisfied because of
```
MaciekPytel: Why only do this for out-of-resources errors? Do we expect other errors to be recoverable?

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -398,6 +403,110 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
+}
+
+func (a *StaticAutoscaler) handleQuotaAndStockoutErrorsForNodeGroup(
+	nodeGroup cloudprovider.NodeGroup,
+	currentInstances []cloudprovider.Instance,
+	previousInstances []cloudprovider.Instance,
+	currentTime time.Time) {
+
+	_, currentErrorCodeToInstance := a.buildInstanceToOutOfResourcesErrorCodeMappings(currentInstances)
+	previousInstanceToErrorCode, _ := a.buildInstanceToOutOfResourcesErrorCodeMappings(previousInstances)
```
MaciekPytel: You never actually use the mapping. All you care about is the key set.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -398,6 +403,110 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
+				unseenInstanceIds = append(unseenInstanceIds, instanceId)
+			}
+		}
+
+		klog.Infof("Failed adding %v nodes (%v unseen previously) to group %v due to %v", len(instanceIds), len(unseenInstanceIds), nodeGroup.Id(), errorCode)
+		if len(unseenInstanceIds) > 0 && a.clusterStateRegistry.IsNodeGroupScalingUp(nodeGroup.Id()) {
+			a.LogRecorder.Eventf(
+				apiv1.EventTypeWarning,
+				"ScaleUpFailed",
+				"Failed adding %v nodes to group %v due to %v",
```
MaciekPytel: Cloudprovider gives us ErrorMessage, it would be nice to expose it to user (otherwise what's the point of having it?).

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -398,6 +403,110 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 	return nil
 }
 
+func (a *StaticAutoscaler) handleQuotaAndStockoutErrors(currentTime time.Time) {
+	currentInstancesPerNodeGroup := a.clusterStateRegistry.GetCloudProviderNodeInstances()
+	previousInstancesPerNodeGroup := a.clusterStateRegistry.GetPreviousCloudProviderNodeInstances()
```
MaciekPytel: I don't feel super strongly about this, but exposing this from clusterstate feels kinda ugly. We could move the whole set difference logic to cloudprovider and only expose something like NodesThatFailedToCreate() (new, old).
To me it seems more consistent with what we already do for things like UnregisteredNodes - build data structure describing state in clusterstate, act on it in core/. WDYT?

aleksandra-malinowska: Did you mean move it to clusterstate?

MaciekPytel:
General comment - cloudprovider returns OutOfResourcesError and you call any method handling it blahQuotaAndStockoutBleh. Those are 2 out of resources scenarios, but I think it would be more consistent to have say handleOutOfResourcesErrors than handleQuotaAndStockoutErrors. Same for all other methods.

cluster-autoscaler/core/static_autoscaler.go
MaciekPytel: I did mean clusterstate, sorry :(

cluster-autoscaler/clusterstate/clusterstate.go
losipiuk: I think it can if someone manually increase MIG size and then the nodes cannot be provisioned because of stockout/quota/whatever.

cluster-autoscaler/core/static_autoscaler.go
losipiuk: Yeah noticed that. I could return:
* map EC->Instance
* map Instance->bool instead

Or have two methods to compute those independenetly.
That will be a bit more code though.
Let me know if you really want to change it.

cluster-autoscaler/core/static_autoscaler.go
losipiuk: Added log as we discussed.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -398,6 +403,110 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
+
+func (a *StaticAutoscaler) deleteNodesByIds(nodeGroup cloudprovider.NodeGroup, instanceIds []string) {
+	if len(instanceIds) == 0 {
+		return
+	}
+
+	nodes := make([]*apiv1.Node, 0, len(instanceIds))
+	for _, nodeName := range instanceIds {
+		// create dummy node name
+		node := &apiv1.Node{
```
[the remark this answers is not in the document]
losipiuk: Would skip it for now. There is no obvious place to put it. And I do not want to do too much refactoring in this PR. I will add TODO and adress it as followup.

cluster-autoscaler/core/static_autoscaler.go
losipiuk: Let's talk

losipiuk:
> General comment - cloudprovider returns OutOfResourcesError and you call any method handling it blahQuotaAndStockoutBleh. Those are 2 out of resources scenarios, but I think it would be more consistent to have say handleOutOfResourcesErrors than handleQuotaAndStockoutErrors. Same for all other methods.

Done

losipiuk:
@MaciekPytel Addressed some comments.
Intentionally skipped the rest. :)

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -154,21 +157,44 @@ func NewClusterStateRegistry(cloudProvider cloudprovider.CloudProvider, config C
...
+}
 
-	oldScaleUpRequest, found := csr.scaleUpRequests[request.NodeGroup.Id()]
+func (csr *ClusterStateRegistry) registerOrUpdateScaleUpNoLock(nodeGroup cloudprovider.NodeGroup, delta int, currentTime time.Time) {
+	scaleUpRequest, found := csr.scaleUpRequests[nodeGroup.Id()]
 	if !found {
-		csr.scaleUpRequests[request.NodeGroup.Id()] = request
+		scaleUpRequest = &ScaleUpRequest{
+			NodeGroup:       nodeGroup,
+			Increase:        delta,
```
MaciekPytel: You can actually register a scale-up with negative delta here. I'm fairly certain this would lead to something bad happening down the line (messed up acceptableRanges calculation? negative scale-up timing out triggering backoff?).

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -874,35 +908,45 @@ func (csr *ClusterStateRegistry) GetUpcomingNodes() map[string]int {
 	return result
 }
 
+// getCloudProviderNodeInstances reterns map keyed on node group id where value is list of node instances
```
MaciekPytel: s/reterns/returns

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -917,3 +961,117 @@ func (csr *ClusterStateRegistry) GetClusterSize() (currentSize, targetSize int)
...
+		for errorMessage := range uniqueErrorMessages {
+			uniqueErrorMessagesForErrorCode[errorCode] = append(uniqueErrorMessagesForErrorCode[errorCode], errorMessage)
+		}
+	}
+
+	return
+}
+
+// GetInstancesWithCreateErrorsToBeDeleted returns instances being created which reported create error and should be deleted. Returned map is keyed on nodegroup id.
+func (csr *ClusterStateRegistry) GetInstancesWithCreateErrorsToBeDeleted() map[string][]cloudprovider.Instance {
```
MaciekPytel: I would skip "ToBeDeleted" part. CSR doesn't really care why you need those.

cluster-autoscaler/clusterstate/clusterstate_test.go
```diff
@@ -725,6 +702,45 @@ func TestGetClusterSize(t *testing.T) {
...
+	assert.Equal(t, clusterstate.scaleUpRequests["ng1"].Increase, 80)
+	assert.Equal(t, clusterstate.scaleUpRequests["ng1"].Time, now)
+	assert.Equal(t, clusterstate.scaleUpRequests["ng1"].ExpectedAddTime, now.Add(10*time.Second))
+
+	// update times on positive delta
+	clusterstate.RegisterOrUpdateScaleUp(provider.GetNodeGroup("ng1"), 30, later)
+	assert.Equal(t, clusterstate.scaleUpRequests["ng1"].Increase, 110)
+	assert.Equal(t, clusterstate.scaleUpRequests["ng1"].Time, later)
+	assert.Equal(t, clusterstate.scaleUpRequests["ng1"].ExpectedAddTime, later.Add(10*time.Second))
+
```
MaciekPytel: Test case for creating a new scale-up with negative delta ("ng2", -100).

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -398,6 +403,60 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
 }
 
+func (a *StaticAutoscaler) deleteCreatedNodesWithErrors() {
+	// We always schedule deleting of incoming errornous nodes
+	// TODO[lukaszos] Consider adding logic to not retry delete every loop iteration
+	instancesWithCreateErrorsByNodeGroupId := a.clusterStateRegistry.GetInstancesWithCreateErrorsToBeDeleted()
+	nodeGroups := a.nodeGroupsById()
+	for nodeGroupId, instancesToBeDeleted := range instancesWithCreateErrorsByNodeGroupId {
+		var err error
+		klog.Infof("Deleting %v from %v node group because of create errors", len(instancesToBeDeleted), nodeGroupId)
```
MaciekPytel: General comment - we should always use V(X) for Infof.

cluster-autoscaler/clusterstate/clusterstate.go
MaciekPytel: Also it would be more consistent to create fake Node objects and return those. We already do that for unregistered nodes.

losipiuk: Renamed to `GetCreatedInstancesWithOutOfResourcesErrors`

PR #1473, opened 2018-12-03
cluster autoscaler: Support scaling AWS groups from 0 when using MixedInstancesPolicy

errm:
AWS recently added the MixedInstancesPolicy as an alternative to LaunchTemplate
or LaunchConfigruration when specifying the config for the instances in an autoscaling group.

https://aws.amazon.com/blogs/aws/new-ec2-auto-scaling-groups-with-multiple-instance-types-purchase-options/

The MixedInstancesPolicy holds a reference to a LaunchTemplate... plus an number of overides
that can overide the instance type specified in the LaunchTemplate.  Thus when using this
type an autoscaling group can schedule multiple instance types.

The MixedInstancesPolicy also allows some percentage of instances in the autoscaling group to be
launched as spot instances, this is kindof othoganal to this PR, but the main reason I wan't to be
able to configure ASGs in this way + use cluster autoscaler.

MixedInstancesPolicy breaks Cluster Autoscaler a bit since there is no longer an LaunchTemplate / Configuration
to discover the instance type from, when the ASG was scaled to 0.

This change makes it so that the cluster autoscaler will work again ...

I have made a few (hopefully reasonable) assumptions:

Firstly where there are instance types specified in multiple overidees, I presume that they will all
have an equivelent amount of resources.

e.g.If I creating an ASGs that has m5.large, m4.large and m3.large overides .... and AWS is choosing
the instance type with the best spot price ... but these three instance types have (roughly) the same
level of resources... so the autoscaler can still work properly... even though we don't know exactly
which instance type will be chosen. The code change just take the instance type from the first override.


losipiuk:
Thanks @errm,

You assumed that all the instance types within ASG will have same amount of resources. And just take first instance shape as template.
I think that we actually need to check  if above assumption holds true, before using asg with mixed instance types as node group in CA.
CA has strong requirement that all the nodes created by the node group have same shape and if that does not hold true it will misbehave.
IMO without validation it is too easy to misconfigure CA.

It seems that with added validation it should be safe to put that in. @MaciekPytel, @aleksandra-malinowska WDYT?

Nit: you cannot directly update vendor dir in PR. But let's talk about it when we come to common understanding of how the actual logic should look like.

cc: @gjtempleton

losipiuk:
Linking the doc which dicusses the problem of multiple instances in single instance group:
https://docs.google.com/document/d/1m-2lQCOwxMrlv1rCz1JqUyBrZAjLOZAELswjuljJVjg/edit

errm:
OK it seems reasonable to have a sanity check here ... and I think we already have all the info we need to compare the shape of the instances in an ASG...

 What kind of behaviour do we want if the instances are mismatching:
* a warning
* fail hard with an error of some kind...
* Ignore ASGs with mismatching instance types ...

Do we want to enforce that instances should be exactly the same shape... or do we need some fuzziness...

e.g. c5.large has 4 GiB Memory, whilst c4.large has only 3.75 GiB ... I think that in practice this should be close enough ... 

I think a rigid check here would prevent useful configurations....

errm:
Having thought about this a little more I think that the correct strategy here is to just pick the "smallest" instance instance type ... that way scale up should provision at least enough capacity...

Being able to have quite a diverse set of instance types makes a lot of sense when using spot instances where the actual price is not entirely proportional to the instance size, the ASG chooses to use the cheapest type in this case... so I don't think a weird over provision will be that harmful ...

Clearly there are a bunch of stuff that isn't really going to work properly since cluster autoscaler doesn't have any price awareness for AWS yet anyway...

I am going to:

* Add some code to pick the smallest instance type from the MixedInstancesPolicy
* Add a warning when the instance types are more than 20% different in size...

@losipiuk WDYT ?

On the other hand this does feel a bit like banging a square peg into a round hole ... I am wondering how feasible it would be for the autoscaler to just directly launch ec2 instances from launch templates, without the overhead of needing the ASG there anyway? Clearly that would be way to big of a change to make without some good evidence in production... but I might experiment with the idea ...

aleksandra-malinowska:
> Having thought about this a little more I think that the correct strategy here is to just pick the "smallest" instance instance type ... that way scale up should provision at least enough capacity...

That could work, if you're OK with overprovisioning.

> I am wondering how feasible it would be for the autoscaler to just directly launch ec2 instances from launch templates, without the overhead of needing the ASG there anyway? Clearly that would be way to big of a change to make without some good evidence in production... but I might experiment with the idea ...

The question is, what would be the overhead for the user to configure autoscaler using templates etc. compared to setting up ASGs?

errm:
> The question is, what would be the overhead for the user to configure autoscaler using templates etc. compared to setting up ASGs?

If you setup an ASG you need a template anyway ... so no harder I presume...

losipiuk:
> > Having thought about this a little more I think that the correct strategy here is to just pick the "smallest" instance instance type ... that way scale up should provision at least enough capacity...
> 
> That could work, if you're OK with overprovisioning.

I think it would not work, with just AWS cloud provider changes. The template node info (which would now returned the template of the "smallest" node) is only used in scale-up from 0 scenario.
If there are already nodes from a node group in cluster, they are used as a proxy of what next node coming from same node group would look like.
Currently cluster autoscaler just picks a random (first) one, and assumes all the other look the same. If we wanted the "smallest" logic to be used in that case, we would need to also implement it in the scale-up algorithm. Specifically in `GetNodeInfosForGroups` function. 

I guess we could add some extension points in the `GetNodeInfosForGroups`, so we are not breaking default behaviour, and allow changing it via system flag (`--node-info-estimation-strategy`?
@MaciekPytel, @aleksandra-malinowska, WDYT?

> What kind of behaviour do we want if the instances are mismatching:
> 
> * a warning
> * fail hard with an error of some kind...
> * Ignore ASGs with mismatching instance types ...

I would go with "ignore". It would be also great to tell the user that the node group is ignored. It should probably be exposed via `cluster-autoscaler-status`. I guess it would require some refactorings in code to pass all the needed information there.

errm:
Having read through the documentation a bit more I think it is feasible to build quite an accurate simulation of the instance type that will be launched next by an ASG. https://aws.amazon.com/blogs/aws/new-ec2-auto-scaling-groups-with-multiple-instance-types-purchase-options/

* We can tell if the next instance to be launched will be on demand be checking the on demand base and percentage fields ... and comparing these numbers with the current composition of the group...
* If the next instance to be launched is indeed on demand, we can be fairly certain that the instance type will be the first one ...
* If we think that the next instance to be launched will be a spot instance then we need to compute the `n` lowest priced instance types (using the spot pricing API) and choose the one that currently has the fewest instances ... 

All of this seems reasonably feasible... but clearly will need a bit of refactoring ...

MaciekPytel:
Note that CA can request multiple nodes in a single request (it calculates how many nodes would be needed to accomodate all pending pods). So predicting how the next one will look like doesn't really help (unless all nodes added in a single request will be the same?), you're probably better off using the smallest one.

mwielgus:
cc: @Jeffwan

mwielgus:
The idea of mixed machines in a single node group seems to require significant amount of refactoring in core as it is against the very idea of a node group. 

As have been no activity on this PR for 1.5 month and the PR itself is controversial, I'm closing it for now. Feel free to reopen once you are ready for more discussion and fixing concerns.

errm:
Thanks @mwielgus ... I am happy to continue to working on this but no consensus around the best way forward seems to have formed.

If there is some indication that the strategy to pick the smallest instance type and accept over-provisioning would be acceptable then I can update the PR... but there didn't really seem to be any agreement that this was even desirable ... 

I personally I am going to start work on an autoscaler that doesn't use AWS autoscaling groups at all... I think this is a significant enough departure, that it doesn't really belong anywhere near this repo though 😂 at least for now!

aleksandra-malinowska:
Using autoscaling groups is just a convenience - you can probably write an alternative cloud provider module which manages individual instances and groups them by type for core algorithm to use. Unless you want to depart completely from the concept of node groups and treat each scale up as "from 0" in the old model (i.e. without using existing node as blueprint, but only looking at templates), you may end up doing it anyway.

errm:
Yes I was thinking about something along those lines...

The only wrinkle I can see is when dealing with spot priced node groups there might be the need for some rescheduling + rebalancing activity when the price on existing instances increases and another (cheeper) type is available .... And as far as I can see the core algorithm doesn't really have an way to deal with this...

guarly:
I'm being hit with Amazon being Over Capacity in the majority of my spot ASGs for 24h straight, which is causing a lot of churn in the nodes (maybe they are alive for 10 minutes and then taken away).

It was suggested https://github.com/aws/containers-roadmap/issues/144 to create a mixed autoscaling group of similar machines which is what I've implemented, to discover that CA is not detecting the instance type, and thus not scaling up the mixed ASG. (BTW, I use kops for the ASG handling)

It would be great if the CA looked at all the instances in a mixed ASG and find the lowest specs. It would solve the problem for me. Maybe it could be stated in the documentation the problem that mixing heterogeneous types could do (maybe link to this pull conversation), but it would be of much help for me.

You could also peek this solution from Zalando: https://github.com/zalando-incubator/autoscaler/pull/1/commits

Jeffwan:
@guarly Do you have at least 1 instance in your mixed autoscaling group? how do you use this feature?  Are you using in this way? https://github.com/kubernetes/autoscaler/pull/1886 and https://github.com/kubernetes/autoscaler/pull/1983

Issue #608, opened 2018-02-02
Cluster Autoscaler AWS cloudprovider tests are failing


bskiba:
I believe this is a flake (example: https://travis-ci.org/kubernetes/autoscaler/jobs/336570454)
k8s.io/autoscaler/cluster-autoscaler/cloudprovider/aws tests are sometimes failing. Logs from the failed test below

```
E0202 14:19:07.636800   11264 runtime.go:66] Observed a panic: "\n\nmock: Unexpected Method Call\n-----------------------------\n\nDescribeAutoScalingGroupsPages(*autoscaling.DescribeAutoScalingGroupsInput,func(*autoscaling.DescribeAutoScalingGroupsOutput, bool) bool)\n\t\t0: {\n  AutoScalingGroupNames: [\"coolasg\"],\n  MaxRecords: 100\n}\n\t\t1: (func(*autoscaling.DescribeAutoScalingGroupsOutput, bool) bool)(0x12e7a50)\n\nThe closest call I have is: \n\nDescribeAutoScalingGroupsPages(*autoscaling.DescribeAutoScalingGroupsInput,mock.AnythingOfTypeArgument)\n\t\t0: {\n  AutoScalingGroupNames: [\"coolasg\"],\n  MaxRecords: 100\n}\n\t\t1: \"func(*autoscaling.DescribeAutoScalingGroupsOutput, bool) bool\"\n\n\n" (
mock: Unexpected Method Call
-----------------------------
DescribeAutoScalingGroupsPages(*autoscaling.DescribeAutoScalingGroupsInput,func(*autoscaling.DescribeAutoScalingGroupsOutput, bool) bool)
		0: {
  AutoScalingGroupNames: ["coolasg"],
  MaxRecords: 100
}
		1: (func(*autoscaling.DescribeAutoScalingGroupsOutput, bool) bool)(0x12e7a50)
The closest call I have is: 
DescribeAutoScalingGroupsPages(*autoscaling.DescribeAutoScalingGroupsInput,mock.AnythingOfTypeArgument)
		0: {
  AutoScalingGroupNames: ["coolasg"],
  MaxRecords: 100
}
		1: "func(*autoscaling.DescribeAutoScalingGroupsOutput, bool) bool"
)
/home/travis/gopath/src/k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:72
[...]

PR #1478, opened 2018-12-04
ec2_instance_types CA cherry picks to 1.2

jeruane:


losipiuk:
There is Travis failure. Can you look at it?

jeruane:
looks to be related to:
https://github.com/kubernetes/autoscaler/issues/608
looks like this test is disabled on master:
https://github.com/kubernetes/autoscaler/blob/991873c2375575ef7cd5cdb6ac6009a41122e92e/cluster-autoscaler/cloudprovider/aws/aws_manager_test.go#L306

I can cherry-pick the commit to disable that test into this branch if you'd like:
https://github.com/kubernetes/autoscaler/commit/e93e25dc718f7f932b23eff381982eac15b978ac

losipiuk:
I guess /retest does not work here.

@jeruane Thanks for checking.
It would be great if you added cherry-picking the e93e25d as part of this PR.

PR #1481, opened 2018-12-05
cloud provider for cluster-api

multi-io:
This is an initial implementation of a new cloud provider (named "clusterapi") that uses cluster-api (https://github.com/kubernetes-sigs/cluster-api/) for creating and managing nodes via the K8s API itself rather than any particular cloud API. We've tested this in clusters with the Kubermatic machine-controller (https://github.com/kubermatic/machine-controller). The way it works it that cluster-api defines CRDs for "Machines" -- which represent machines in the backing cloud (which become nodes after successful initialization; see below), and MachineSets and MachineDeployments, which are to machines what ReplicaSets and Deployments are to pods: A MachineSet or MachineDeployment declares a replica count, and a corresponding controller (part of cluster-api) will create and (in the case of MachineDeployments) rolling-upgrade (if needed) a corresponding set of machine objects. Machine-controller, which has its own cloud provider layer, will then create and provision Kubernetes machines in a cloud environment, which will boot up, register with the API server and eventually show up as nodes.

What this autoscaler cloud provider does is expose all the MachineDeployments as node groups, and simply manipulate the replica count when asked to scale a node group.

The cloud provider works nicely in our setup (we're running on an OpenStack cloud). It's not feature-complete yet though; there are a couple of caveats:

- we've vendored a slightly older version of cluster-api because the current master contains a breaking change (https://github.com/kubernetes-sigs/cluster-api/pull/548) that machine-controller's version of cluster-api doesn't yet have

- downscaling doesn't work currently -- https://github.com/kubernetes-sigs/cluster-api/pull/558 is needed for this and wasn't merged until very recently -- we'd implement this next, wanted to get some feedback on the PR first

- ClusterapiNodeGroup.TemplateNodeInfo(), which is used to estimate new nodes' resources when scaling up empty MachineDeployments, is entirely Openstack-specific right now (because that's what we use) and won't work anywhere else. It extracts the information from the MachineDeployment's providerConfig, which contains the cloud-specific configuration to use for new machines in the deployment. We're not sure how to implement this for other clouds except manually.

- we might want to support scaling of MachineSets, not just MachineDeployments

The PR contains two required changes outside the new cloud provider itself, currently provided as separate commits in the PR to make  the review easier:

- since this cloud provider's cloud-config is just the cluster's kubeconfig, we had to change `AutoscalerOptions` and the cloud provider builder so we pass the kubeconfig down to `buildCloudProvider` (https://github.com/kubernetes/autoscaler/commit/fdb83cde659f011c1d6f1dd84ef6e3050899a402). This change was previously provided as a separate PR (https://github.com/kubernetes/autoscaler/pull/1413).

- cluster-api currently still uses glog, while autoscaler uses klog. In the current upstream, this causes a runtime panic ("redefined: log_dir") at https://github.com/kubernetes/autoscaler/blob/d53928a11d422256ee452a8aa5e397b9e4823216/cluster-autoscaler/main.go#L324 because `glog.init()` adds a log_dir flag to the global `flag.CommandLine`, and then `klog.InitFlags(nil)` does the same, hence the error. https://github.com/kubernetes/autoscaler/commit/f4f7986862f953e9aa0d4497189dc5175186fec6 fixes this by registering klog's flags into a local flagset, then merging that and `flag.CommandLine` into `pflag.CommandLine`, which is then parsed. The fix will be needed not just for cluster-api but also with any other dependency that uses glog, so this should be of general interest.


aleksandra-malinowska:
Thanks for contributing this!

> downscaling doesn't work currently -- kubernetes-sigs/cluster-api#558 is needed for this and wasn't merged until very recently -- we'd implement this next, wanted to get some feedback on the PR first

That PR most likely won't be sufficient (we want to *only* delete marked machine, and do nothing if there's none).

> ClusterapiNodeGroup.TemplateNodeInfo(), which is used to estimate new nodes' resources when scaling up empty MachineDeployments, is entirely Openstack-specific right now (because that's what we use) and won't work anywhere else. It extracts the information from the MachineDeployment's providerConfig, which contains the cloud-specific configuration to use for new machines in the deployment. We're not sure how to implement this for other clouds except manually.

Agreed it probably won't be possible until and unless something like template node becomes part of Cluster API as well. However, if it's the only Openstack-specific part, can you perhaps parameterize the cloud provider with this logic? Then we can add support for starting autoscaler with ClusterAPI+Openstack-template or just ClusterAPI.

Regarding the dependencies added, we have a rather hacky and sad setup which requires updating them using kubernetes/kubernetes /staging and /vendor, and putting the rest in kubernetes/autoscaler's /_override. Unless you have an idea how to solve this long-standing issue, the instructions are here: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-update-ca-dependencies-particularly-k8siokubernetes

multi-io:
I forgot to mention one other thing: Unit test coverage of the cloud provider isn't great right now. We have an additional test that works against cluster-api's fake (mock) clientset. That clientset doesn't compile against the currently vendored `k8s.io/client-go/testing` due to API changes in the latter. So the test isn't included in the PR. I guess we'll have to wait until cluster-api regenerates the fake clientset with the newer API, would add the test then.

frobware:
Hi @multi-io, @aleksandra-malinowska  - I just wanted to highlight that I have been working on a clusterapi-based cloudprovider implementation for the autoscaler here: https://github.com/openshift/kubernetes-autoscaler. 

We are doing these as carry patches (on master) so that we can easily identify them for upstream integration:

```console
$ git log | grep UPSTREAM | grep cluster-api | uniq
    UPSTREAM: <carry>: cluster-api switch to using informers for cluster state
    UPSTREAM: <carry>: cluster-api switch to annotations
    UPSTREAM: <carry>: initial cluster-api cloudprovider implementation
    UPSTREAM: <carry>: initial cluster-api provider implementation
    UPSTREAM: <carry>: vendor sigs.k8s.io/cluster-api
```

In a similar vein, there are a few small things I am aware of [that are not working 100%](https://github.com/openshift/kubernetes-autoscaler/pull/14) but the missing pieces in my mind right now are:
- unit tests
- e2e tests
- the ability to scale from zero

I have been testing this primarily on AWS using the [AWS actuator](https://github.com/openshift/cluster-api-provider-aws) that we have also been developing.

/cc @hardikdr @derekwaynecarr @ingvagabund @enxebre

frobware:
> It would probably be easier if you could sync and figure out if it makes sense to converge these efforts at some point.

Yes, totally agree.

I struggled initially to get the dependencies to work having vendored cluster-api but things have moved on and that seems easier now.

multi-io:
> Thanks for contributing this!
> 
> > downscaling doesn't work currently -- [kubernetes-sigs/cluster-api#558](https://github.com/kubernetes-sigs/cluster-api/pull/558) is needed for this and wasn't merged until very recently -- we'd implement this next, wanted to get some feedback on the PR first
> 
> That PR most likely won't be sufficient (we want to _only_ delete marked machine, and do nothing if there's none).

Yeah, I talked to some folks at the cluster-api SIG yesterday, they said the same thing.

> > ClusterapiNodeGroup.TemplateNodeInfo(), which is used to estimate new nodes' resources when scaling up empty MachineDeployments, is entirely Openstack-specific right now (because that's what we use) and won't work anywhere else. It extracts the information from the MachineDeployment's providerConfig, which contains the cloud-specific configuration to use for new machines in the deployment. We're not sure how to implement this for other clouds except manually.
> 
> Agreed it probably won't be possible until and unless something like template node becomes part of Cluster API as well. However, if it's the only Openstack-specific part, can you perhaps parameterize the cloud provider with this logic? Then we can add support for starting autoscaler with ClusterAPI+Openstack-template or just ClusterAPI.

I can do that, the question is how long-term such a solution would be. It feels weird to have a sort of mini "cloud provider" layer inside the clusterapi cloud provider, right? The cluster-api people said that there are plans to extend their "MachineClass" CRD so it declares some attributes of the machine templates like cores and RAM in a provider-neutral way, which would allow us to implement `TemplateNodeInfo()` just once and have it work everywhere. I guess this might happen sometime in the next year, not sure.

> 
> Regarding the dependencies added, we have a rather hacky and sad setup which requires updating them using kubernetes/kubernetes /staging and /vendor, and putting the rest in kubernetes/autoscaler's /_override. Unless you have an idea how to solve this long-standing issue, the instructions are here: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-update-ca-dependencies-particularly-k8siokubernetes

Yep. I read and used that document. I essentially recreated Godeps/vendor using the process laid out in that FAQ, and then added the cluster-api dependency by cloning cluster-api into the GOPATH and then running `godep save ./...` in cluster-autoscaler and resolving any missing library errors by cloning those libraries too, taking care to clone the version referenced in cluster-api's `Gopkg.lock`. I repeated this until the `godep save ./...` succeeded.

MaciekPytel:
Regarding TemplateNodeInfo(): it's optional to implement, you can have a cloudprovider that doesn't implement it and it will work (except for scale-to-0). If you have an implementation that only works in openstack CA will still try to use it on AWS cluster using cluster API cloudprovider. I don't see that ending well. 

I think the right way to do it would be to default to not-implemented TemplateNodeInfo() and use openstack implementation if an appropriate flag is set. 

Of course once Cluster API provides a generic way to implement it we should just use that. I'll admit I'm rather pessimistic with regards to when this will be available though. It is a hard problem.

multi-io:
> Regarding TemplateNodeInfo(): it's optional to implement, you can have a cloudprovider that doesn't implement it and it will work (except for scale-to-0). If you have an implementation that only works in openstack CA will still try to use it on AWS cluster using cluster API cloudprovider. I don't see that ending well.
> 
> I think the right way to do it would be to default to not-implemented TemplateNodeInfo() and use openstack implementation if an appropriate flag is set.

Oh, I do check the MachineDeployment's cloudProvider field to see if this is an OpenStack MachineDeployment, and return an error (not ErrNotImplemented though) if it isn't. I'll change this to ErrNotImplemented.

MaciekPytel:
Yeah, in that case I don't think returning ErrNotImplemented solves the issue. We don't actually handle that error gracefully for TemplateNodeInfo(). More importantly when we call TemplateNodeInfo() it's too late anyway - we need to know NodeGroup won't be able to scale back from 0, so we don't scale it down to 0 nodes).

The way we handle it is by setting SupportScaleToZero when creating cloudprovider and rejecting config with min set to 0 if cloudprovider doesn't support it. Everywhere else we just rely on the fact that NodeGroup without working TemplateNodeInfo must have min >0 nodes.

multi-io:
> Hi @multi-io, @aleksandra-malinowska - I just wanted to highlight that I have been working on a clusterapi-based cloudprovider implementation for the autoscaler here: https://github.com/openshift/kubernetes-autoscaler.

Hey, that looks quite nice. The informer-based implementation is interesting. I think you're based on a slightly older autoscaler version with no cloudprovider.Machines yet. You support MachineSets, not MachineDeployments, right? It's the other way round with our implementation (since MachineDeployments are more feature-rich, which suits our application).

> 
> We are doing these as carry patches (on master) so that we can easily identify them for upstream integration:
> 
> ```
> $ git log | grep UPSTREAM | grep cluster-api | uniq
>     UPSTREAM: <carry>: cluster-api switch to using informers for cluster state
>     UPSTREAM: <carry>: cluster-api switch to annotations
>     UPSTREAM: <carry>: initial cluster-api cloudprovider implementation
>     UPSTREAM: <carry>: initial cluster-api provider implementation
>     UPSTREAM: <carry>: vendor sigs.k8s.io/cluster-api
> ```

I'm not sure how to proceed here. Were you planning to offer these to upstream as PRs?

> 
> In a similar vein, there are a few small things I am aware of [that are not working 100%](https://github.com/openshift/kubernetes-autoscaler/pull/14) but the missing pieces in my mind right now are:
> 
> * unit tests
> * e2e tests
> * the ability to scale from zero

As mentioned earlier, I have a unit test that's not currently part of the PR, and it uses cluster-api's mock clientsets (which @alvaroaleman repaired and re-enabled ~3 weeks ago) to unit-tests the cluster-api facade (named `MachineManager` in our PR, `machineController` in your code AFAICS). As I explained, the mock clientsets don't compile against the k8s version that the current autoscaler upstream master has vendored; we might either wait for cluster-api to update or generate our own one as part of the PR.

frobware:
> Hey, that looks quite nice. The informer-based implementation is interesting. I think you're based on a slightly older autoscaler version with no cloudprovider.Machines yet. You support MachineSets, not MachineDeployments, right? It's the other way round with our implementation (since MachineDeployments are more feature-rich, which suits our application).

Yes, I'm currently using version 1.12.

Haven't looked at machine deployments right now but it is something I plan to look at.

frobware:
> As I explained, the mock clientset doesn't compile against the k8s version that the current autoscaler upstream master has vendored; we might either wait for cluster-api to update their clientset or generate our own one as part of the PR.

I ran into this a lot when we first started out in our implementation. Will look at the mock facade -  thanks for the link.

multi-io:
> I forgot to mention one other thing: Unit test coverage of the cloud provider isn't great right now. We have an additional test that works against cluster-api's fake (mock) clientset. That clientset doesn't compile against the currently vendored `k8s.io/client-go/testing` due to API changes in the latter. So the test isn't included in the PR. I guess we'll have to wait until cluster-api regenerates the fake clientset with the newer API, would add the test then.

Update 2019/02/14: Vendored the latest cluster-api (ae564acc7c608c90d51684e854d3ca8fa36b1e68), which has updated their fake clientset to be compatible with the `k8s.io/client-go/testing` vendored by autoscaler, so all our unit tests are now included in the PR and succeed. We're running CA built from this PR branch in our production setup now, seems to work nicely. We're currently working on AWS support, so a corresponding `ClusterapiNodeGroup.TemplateNodeInfo()` implementation is forthcoming in the coming weeks probably.

Btw. I'm not sure what happened with the cla/linuxfoundation check -- can someone re-trigger this maybe?

PR #1484, opened 2018-12-07
update log output detailed warning info

gaozhenhai:
Signed-off-by: Zhenhai Gao <gaozh1988@live.com>

### DIFF
--- a/cluster-autoscaler/core/utils.go
+++ b/cluster-autoscaler/core/utils.go
@@ -407,7 +407,7 @@ func removeOldUnregisteredNodes(unregisteredNodes []clusterstate.UnregisteredNod
 			}
 			size, err := nodeGroup.TargetSize()
 			if err != nil {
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/core/utils.go
```diff
@@ -407,7 +407,7 @@ func removeOldUnregisteredNodes(unregisteredNodes []clusterstate.UnregisteredNod
 			}
 			size, err := nodeGroup.TargetSize()
 			if err != nil {
-				klog.Warningf("Failed to get node group size, err: %v", err)
+				klog.Warningf("Failed to get node group size for %s: %v", unregisteredNode.Node.Name, err)
```
losipiuk: If already changing this message then add node group id to it also. So it is sth like
"Failed to get node group size; unregisteredNode=%v; nodeGroup=%v; err=%v"

Issue #1449, opened 2018-11-26
AWS cloud provider tests take 120 seconds


losipiuk:
The problem started to occur after merging 
https://github.com/kubernetes/autoscaler/commit/a36f8007afa65fcb2ff7ae291ee815be5512e01c
Each of 6 test method tries to get region from AWS api (see aws_manager.go `getRegion` method). This call times out after 20seconds.

@shatil Is there a chance that you could fix that?

shatil:
@losipiuk, certainly!

Inside of AWS this returns instantly, and outside, it will always timeout, so I'm thinking to add a 1 second timeout around `svc.Region()`.

losipiuk:
I was thinking more that you could mock the http call in tests so it returns immediately. 
Depending on build environment and sleeping (even 1s) in unit tests are both bad practices.

aleksandra-malinowska:
> I was thinking more that you could mock the http call in tests so it returns immediately.

This. Or set the env var in tests. Or read the env var earlier, e.g. when constructing AWS manager, and pass it as parameter. Etc.

shatil:
@losipiuk @aleksandra-malinowska I like the environment variable idea. I had already mocked out another environment variable lookup when it occurred to me:

If the tests (indirectly) relying on `getRegion` supply an `AWS_REGION` env variable, there is no timeout. Is this approach acceptable? Or will this throw something off?

I think tests using `createAWSManagerInternal` would need it, and that's about it.

```
$ git grep createAWSManagerInternal
aws_manager.go:func createAWSManagerInternal(
aws_manager.go: return createAWSManagerInternal(configReader, discoveryOpts, nil, nil)
aws_manager_test.go:    m, err := createAWSManagerInternal(nil, do, &autoScalingWrapper{s}, nil)
aws_manager_test.go:    m, err := createAWSManagerInternal(nil, cloudprovider.NodeGroupDiscoveryOptions{}, nil, &ec2Wrapper{s})
aws_manager_test.go:                    m, err := createAWSManagerInternal(nil, cloudprovider.NodeGroupDiscoveryOptions{}, nil, &ec2Wrapper{s})
aws_manager_test.go:    m, err := createAWSManagerInternal(nil, do, &autoScalingWrapper{s})
```

I could set and reset it for that test alone, unless it's okay to set it for _all_ tests. Is there some central location where environment variables for tests are configured?

aleksandra-malinowska:
[...]

PR #1490, opened 2018-12-07
Set AWS_REGION so tests don't wait for SDK timeout #1449

shatil:
This patch sets and then resets the `AWS_REGION` environment variable in
tests that call `createAWSManagerInternal`, which calls `getRegion`.

`getRegion` introduced `AWS_REGION` lookup at runtime, but tests that
don't (a) run inside AWS or (b) set `AWS_REGION` environment variable
encountered AWS SDK's EC2 Metadata lookup timeout. Tests that indirectly
invoked `getRegion`, e.g., via `createAWSManagerInternal`, saw at least
a 5-second timeout per invocation, ballooning test times.

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_manager.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_manager.go
@@ -80,6 +80,13 @@ func getRegion(cfg ...*aws.Config) string {
 }
 
 // createAwsManagerInternal allows for a customer autoScalingWrapper to be passed in by tests
+//
+// #1449 If running tests outside of AWS without AWS_REGION among environment
+// variables, avoid a 5+ second EC2 Metadata lookup timeout in getRegion by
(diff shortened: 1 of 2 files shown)


losipiuk:
I am not a big fan of the fact that tests commicate with internet at all.
@aleksandra-malinowska please take a look if that is what you intended this to look like.

shatil:
@losipiuk fwiw, EC2 Metadata endpoint's default address is link-local, and not an Internet address. During the previous round of tests, I set up an `httptest` server to handle the requests that were intended to hit the endpoint.

During tests it makes sense to mock out AWS endpoints that are hit, maybe in a test main somewhere? Or have the `AWS_REGION` environment variable specified when running tests. I have that in my local environment, for example, in `~/.aws/config`, which is probably why I never was hit by the timeout before.

PR #1485, opened 2018-12-07
Fix aws flaking Unit Tests

Jeffwan:
Looks like  this Unit Test blocks several tasks. I notice someone comment this test and I think we can use `MatchedBy` to customize comparison logic.

This will resolve #1346 and other dependent issues for 1.2

~~I run hack/verify-gofmt.sh and found several files are not formatted and then run `gofmt -s -w ${filename}` Not sure if that's needed. if not, I will remove second commit~~

PR #1495, opened 2018-12-08
Cherry-pick of #1485 to master : Fix aws flaking Unit Tests

Jeffwan:


losipiuk:
/approve
/lgtm

Thanks. Commit name should not reference 1.13 but that is just a nit.

PR #1499, opened 2018-12-11
Add cache for resource IDs of vmss instances

feiskyer:
This PR adds caches for resource IDs of vmss instances. Those IDs won't change after creation, so it's safe for CA to keep them in memory.

Without this, cluster-autoscaler would be Throttled by Azure API: 

```json
"details": [
    {
      "code": "TooManyRequests",
      "message": "{\"operationGroup\":\"GetVMScaleSetVM30Min\",\"startTime\":\"2018-11-29T20:49:14.3565951+00:00\",\"endTime\":\"2018-11-29T21:04:14.3565951+00:00\",\"allowedRequestCount\":2500,\"measuredRequestCount\":2596}",
      "target": "GetVMScaleSetVM30Min"
    }
  ],
  "innererror": {
    "internalErrorCode": "TooManyRequestsReceived"
  },
  "code": "OperationNotAllowed",
  "message": "The server rejected the request because too many requests have been received for this subscription."
}
```

cc @kkmsft

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
@@ -22,17 +22,17 @@ import (
 	"sync"
 	"time"
 
-	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
-	"k8s.io/klog"
-
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
 	"k8s.io/autoscaler/cluster-autoscaler/config/dynamic"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/gpu"
+	"k8s.io/klog"
 	kubeletapis "k8s.io/kubernetes/pkg/kubelet/apis"
 	schedulercache "k8s.io/kubernetes/pkg/scheduler/cache"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
 )
 
 // ScaleSet implements NodeGroup interface.
@@ -46,6 +46,8 @@ type ScaleSet struct {
 	mutex       sync.Mutex
 	lastRefresh time.Time
 	curSize     int64
+	// virtualMachines holds a list of vmss instances (instanceID -> resourceID).
+	virtualMachines map[string]string
 }
 
 // NewScaleSet creates a new NewScaleSet.
@@ -54,10 +56,11 @@ func NewScaleSet(spec *dynamic.NodeGroupSpec, az *AzureManager) (*ScaleSet, erro
 		azureRef: azureRef{
 			Name: spec.Name,
 		},
-		minSize: spec.MinSize,
-		maxSize: spec.MaxSize,
-		manager: az,
-		curSize: -1,
+		minSize:         spec.MinSize,
+		maxSize:         spec.MaxSize,
+		manager:         az,
+		curSize:         -1,
+		virtualMachines: make(map[string]string),
 	}
 
 	return scaleSet, nil
@@ -188,8 +191,7 @@ func (scaleSet *ScaleSet) IncreaseSize(delta int) error {
 // Note that the list results is not used directly because their resource ID format
 // is not consistent with Get results.
 // TODO(feiskyer): use list results directly after the issue fixed in Azure VMSS API.
-func (scaleSet *ScaleSet) GetScaleSetVms() ([]compute.VirtualMachineScaleSetVM, error) {
-	instanceIDs := make([]string, 0)
+func (scaleSet *ScaleSet) GetScaleSetVms() ([]string, error) {
 	ctx, cancel := getContextWithCancel()
 	defer cancel()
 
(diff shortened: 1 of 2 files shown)


losipiuk:
@feiskyer you never delete from the map. Is that omission or not done on purpose. It guess it will not grow big so it should not be a problem.

/lgtm
/approve
/hold

Please unhold if you do not want to make any changes.

feiskyer:
> you never delete from the map. Is that omission or not done on purpose? I guess it will not grow big so it should not be a problem.

Yep, this is intentional. It won't grow too big and can be used as another cache next time.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -188,8 +191,7 @@ func (scaleSet *ScaleSet) IncreaseSize(delta int) error {
 // Note that the list results is not used directly because their resource ID format
 // is not consistent with Get results.
 // TODO(feiskyer): use list results directly after the issue fixed in Azure VMSS API.
```
kkmsft: Can you please add a GitHub issue link here ? Is it the case that If this issue is fixed then, we don't have to do get but use what comes from the list ?

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -223,7 +232,14 @@ func (scaleSet *ScaleSet) GetScaleSetVms() ([]compute.VirtualMachineScaleSetVM,
 			}
 		}
 
-		allVMs = append(allVMs, vm)
+		// The resource ID is empty string, which indicates the instance may be in deleting state.
+		if len(*vm.ID) == 0 {
```
kkmsft: Is there a bug for this issue, that is the resource ID empty  causing issues with the vmss?

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -447,17 +463,17 @@ func (scaleSet *ScaleSet) TemplateNodeInfo() (*schedulercache.NodeInfo, error) {
...
 		return nil, err
 	}
 
 	instances := make([]cloudprovider.Instance, 0, len(vms))
 	for i := range vms {
-		if len(*vms[i].ID) == 0 {
-			continue
-		}
-		name := "azure://" + *vms[i].ID
+		name := "azure://" + vms[i]
```
kkmsft: okay, so we are moving the check which was here to one more layer below - right ?

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -200,15 +202,22 @@ func (scaleSet *ScaleSet) GetScaleSetVms() ([]compute.VirtualMachineScaleSetVM,
...
 	for _, vm := range result {
 		instanceIDs = append(instanceIDs, *vm.InstanceID)
 	}
 
-	allVMs := make([]compute.VirtualMachineScaleSetVM, 0)
+	allVMs := make([]string, 0)
 	for _, instanceID := range instanceIDs {
+		// Get from cache first.
+		if v, ok := scaleSet.virtualMachines[instanceID]; ok {
+			allVMs = append(allVMs, v)
```
kkmsft: Could we please add the logic here to ensure that the mapping is deleted when the instance is no longer present in the list ?

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -46,6 +46,8 @@ type ScaleSet struct {
 	mutex       sync.Mutex
 	lastRefresh time.Time
 	curSize     int64
+	// virtualMachines holds a list of vmss instances (instanceID -> resourceID).
+	virtualMachines map[string]string
```
kkmsft: Isn't this generic map enough that we need to move this to azure manager and use this in non-vmss cases ?

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
kkmsft: Perhaps we can have the map and utility functions move there and only implement it for vmss for now and take up the rest later onwards ?

cluster-autoscaler/cloudprovider/azure/azure_scale_set_test.go
```diff
@@ -158,3 +159,29 @@ func TestDebug(t *testing.T) {
...
+		},
+	}
+	group, err := provider.NodeGroupForNode(node)
+	assert.NoError(t, err)
+	assert.NotNil(t, group, "Group should not be nil")
+	assert.Equal(t, group.Id(), "test-asg")
+	assert.Equal(t, group.MinSize(), 1)
+	assert.Equal(t, group.MaxSize(), 5)
+
+	instances, err := group.Nodes()
```
kkmsft: Shouldn't we do this twice to make sure that the cache is working or check whether the cache got populated ?

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
feiskyer: I'm asking them to fix the issue, no Github issue yet.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
feiskyer: This is expected when the instance is in deleting state, and we won't want an empty string.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
feiskyer: we don't need to remove them, because in next scale up, if the instanceID is same, then resourceID would also be same.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
feiskyer: No, this is only for vmss.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
feiskyer: right

cluster-autoscaler/cloudprovider/azure/azure_scale_set_test.go
feiskyer: Good catch. Added validation for it.

feiskyer:
@kkmsft Addressed comments. PTAL

PR #1506, opened 2018-12-12
Remove namespace from hamster example.

thealmightygrant:
Very minor change, but in order to ensure that the vpa ends up in the same namespace as the deployment, the default namespace needs to be removed here. This fixes any issues with the current context have a default namespace that is not named "default".


thealmightygrant:
I signed it

PR #1521, opened 2018-12-21
Doc: scheduler.alpha.kubernetes.io/critical-pod annotations are deprecated. Use priorityClassName instead.

aerostitch:
Hi guys,

Thanks for this awesome work!
I noticed in the documentation the use of the `scheduler.alpha.kubernetes.io/critical-pod` annotation that has been deprecated as of version 1.13 and will be removed in 1.14.

So this PR changes the doc to use `priorityClassName: system-cluster-critical`.

Let me know  if I need to change some wording.

Thanks for your help,
Joseph

### DIFF
--- a/cluster-autoscaler/README.md
+++ b/cluster-autoscaler/README.md
@@ -122,8 +122,8 @@ default deployment strategy on GCP.
 It is possible to run a customized deployment of Cluster Autoscaler on worker nodes, but extra care needs
 to be taken to ensure that Cluster Autoscaler remains up and running. Users can put it into kube-system
 namespace (Cluster Autoscaler doesn't scale down node with non-mirrored kube-system pods running
-on them) and add `scheduler.alpha.kubernetes.io/critical-pod` annotation (so that the rescheduler,
-if enabled, will kill other pods to make space for it to run).
+on them) and set a `priorityClassName: system-cluster-critical` property on your pod spec
+(so that the rescheduler, if enabled, will kill other pods to make space for it to run).
 
 Supported cloud providers:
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/README.md
```diff
@@ -122,8 +122,8 @@ default deployment strategy on GCP.
 It is possible to run a customized deployment of Cluster Autoscaler on worker nodes, but extra care needs
 to be taken to ensure that Cluster Autoscaler remains up and running. Users can put it into kube-system
 namespace (Cluster Autoscaler doesn't scale down node with non-mirrored kube-system pods running
-on them) and add `scheduler.alpha.kubernetes.io/critical-pod` annotation (so that the rescheduler,
-if enabled, will kill other pods to make space for it to run).
+on them) and set a `priorityClassName: system-cluster-critical` property on your pod spec
+(so that the rescheduler, if enabled, will kill other pods to make space for it to run).
```
MaciekPytel: Can we remove the part about rescheduler as well? Rescheduler was the thing using the critical-pod annotation that has been deperecated, it has nothing to do with priority classes.

aerostitch: Good catch thanks!
I changed it to `to prevent your pod from being evicted`.

MaciekPytel:
/approve
Thanks for fixing this. 

@ringtail Since most of the changes are in alicloud documentation do you want to also have a look before I merge this?

ringtail:
@MaciekPytel Thanks,I'll check it later.

ringtail:
@MaciekPytel Thanks! It works well as expected.

PR #1526, opened 2018-12-21
Use k8s.io/klog instead github.com/golang/glog

Rajat-0:


bskiba:
@Rajat-0 thanks for taking the time to do this! Could you run hack/update-gofmt.sh to fix gofmt issues? Then I think this should be good to merge.

PR #1531, opened 2018-12-24
refactor(*): fix some golint warning

lsytj0413:
fix some golint warning.

- add some export function/type comment
- replace some increment with ++
- omit some will be inferred type name

### DIFF
--- a/cluster-autoscaler/cloudprovider/kubemark/kubemark_other.go
+++ b/cluster-autoscaler/cloudprovider/kubemark/kubemark_other.go
@@ -34,30 +34,40 @@ const (
 	ProviderName = "kubemark"
 )
 
+// KubemarkCloudProvider implements CloudProvider interface.
 type KubemarkCloudProvider struct{}
 
+// BuildKubemarkCloudProvider builds a CloudProvider for kubemark. Builds
+// node groups from passed in specs.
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/cloudprovider/kubemark/kubemark_other.go
```diff
@@ -34,30 +34,40 @@ const (
 	ProviderName = "kubemark"
 )
 
+// KubemarkCloudProvider implements CloudProvider interface.
 type KubemarkCloudProvider struct{}
 
+// BuildKubemarkCloudProvider builds a CloudProvider for kubemark. Builds
```
bskiba: Not sure these are the right comments, provided that most of the methods return an ErrNotImplemented.

lsytj0413: kubemark only works on linux environment, so this function in this file is Dummy implementation and returns ErrNotImplemented.

bskiba: It doesn't build outside of linux. That's why I think instead of copy-pasting the comments it's better to phrase all the comments sth like this:
//BuildKubemarkCloudProvider is not implemented for non-linux environments.

But now I see that there already are methods here with comments in the same way you added them, so I guess LGTM

PR #1534, opened 2018-12-25
More flexible VPA certificate configuration

### LINKED ISSUE #1530: increase flexibility on reading the certificates
I am trying to install the admission controller using a different method of providing certificates.
If I understand correctly the code right now expects certificates in the following locations:
```
/etc/tls-certs/caKey.pem
/etc/tls-certs/caCert.pem
/etc/tls-certs/serverKey.pem
/etc/tls-certs/serverCert.pem
```
It is possible to change the directory where the certs are, but not the name of the files.
Can you add a way to specify the name and location of each file independently?

Also something must be escaping me, but why is the caKey there? That is unusual and doesn't look right. Just to be sure, who is going to call this endpoint and how is it going to validate the certificate?

petr-k:
This resolves #1530 in a way that's flexible but also maintains backward compatibility.


vertical-pod-autoscaler/pkg/admission-controller/main.go
```diff
@@ -34,8 +34,14 @@ import (
 )
 
 var (
-	certsDir = flag.String("certs-dir", "/etc/tls-certs", `Where the TLS cert files are stored.`)
-	address  = flag.String("address", ":8944", "The address to expose Prometheus metrics.")
+	certsConfiguration = &certsConfig{
+		certsDir:       flag.String("certs-dir", "/etc/tls-certs", `Where the TLS cert files are stored.`),
+		caCertFile:     flag.String("ca-cert-file", "caCert.pem", "Path to CA PEM file. If relative, it is joined with whatever given to certs-dir."),
```
bskiba: I'm kinda not convinced by the way this is set up. You can pass a certs-dir and then make one of the file paths absolute and have this mixed setup, where for some files you ignore certs-dir and for others not. Or you can pass all the files as absolute and a non-empty certs-dir and certs-dir is completely ignored.

What is the intended use case? Are the files in different dirs?

petr-k: The thinking behind this is that, yes, those files can be in different dirs. I'm not sure if that makes sense or not.

There's multiple options to tackle this:
1) Ditch `certs-dir` and require all cert file paths to be absolute. Breaks compatibility (is that na issue?).
2) Keep `certs-dir` and treat all cert file names as relative to `certs-dir`. Maintains compatibility, but a bit inflexible as all those files needs to be in the same directory.
3) What I went for here. Most flexible, but admittedly a bit too complex.

What do you think would be the best option? I went for 3 just because it covers just about any scenario.

bskiba: I think we can go for 1 here for simplification. The previous setup was so inflexible I doubt anyone was actually overriding this instead of using default.

petr-k: Cool, changed. I also modified flag file names to better match what's already being used by many other Kubernetes components.

bskiba:
Thanks!
/lgtm
/approve

PR #1539, opened 2018-12-28
Add listers for more controllers

MaciekPytel:
Currently we call the API many times per loop to get various controllers. Turns out it impacts CA performance quite a bit in large clusters (and it puts an extra load on API server). 

This is preparatory work for migrating those API calls to use watch cache instead.

I decided to use upstream listers directly, rather than wrap them in our own interface. Partly because we will actually need to replace Get() calls which the upstream lister already provides (and our interfaces don't), partly because all we did in our interface was forward the call to upstream interface anyway and using upstream interface feels more consistent with k8s ecosystem.
I've migrated existing daemonSetLister to use upstream interface as well to be consistent with other controllers (and migrated it to use apps/v1.DaemonSet instead of old extensions/v1beta1). I've left other listers as they were, because they actually have custom logic and we won't need Get calls for those.

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler_test.go
+++ b/cluster-autoscaler/core/static_autoscaler_test.go
@@ -168,7 +168,8 @@ func TestStaticAutoscalerRunOnce(t *testing.T) {
 	}
 	context := NewScaleTestAutoscalingContext(options, &fake.Clientset{}, provider)
 	listerRegistry := kube_util.NewListerRegistry(allNodeListerMock, readyNodeListerMock, scheduledPodMock,
-		unschedulablePodMock, podDisruptionBudgetListerMock, daemonSetListerMock)
+		unschedulablePodMock, podDisruptionBudgetListerMock, daemonSetListerMock,
+		nil, nil, nil, nil)
 	context.ListerRegistry = listerRegistry
 
 	clusterStateConfig := clusterstate.ClusterStateRegistryConfig{
@@ -347,7 +348,8 @@ func TestStaticAutoscalerRunOnceWithAutoprovisionedEnabled(t *testing.T) {
 	}
 	context := NewScaleTestAutoscalingContext(options, &fake.Clientset{}, provider)
 	listerRegistry := kube_util.NewListerRegistry(allNodeListerMock, readyNodeListerMock, scheduledPodMock,
-		unschedulablePodMock, podDisruptionBudgetListerMock, daemonSetListerMock)
+		unschedulablePodMock, podDisruptionBudgetListerMock, daemonSetListerMock,
+		nil, nil, nil, nil)
 	context.ListerRegistry = listerRegistry
 
 	clusterStateConfig := clusterstate.ClusterStateRegistryConfig{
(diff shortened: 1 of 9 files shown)


cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -168,7 +168,8 @@ func TestStaticAutoscalerRunOnce(t *testing.T) {
 	}
 	context := NewScaleTestAutoscalingContext(options, &fake.Clientset{}, provider)
 	listerRegistry := kube_util.NewListerRegistry(allNodeListerMock, readyNodeListerMock, scheduledPodMock,
-		unschedulablePodMock, podDisruptionBudgetListerMock, daemonSetListerMock)
+		unschedulablePodMock, podDisruptionBudgetListerMock, daemonSetListerMock,
```
losipiuk: Either all parameters in single line or each in separate one.

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -40,28 +44,42 @@ type ListerRegistry interface {
...
+	podDisruptionBudgetLister   PodDisruptionBudgetLister
+	daemonSetLister             DaemonSetLister
+	replicationControllerLister v1lister.ReplicationControllerLister
+	jobLister                   v1batchlister.JobLister
+	replicaSetLister            v1appslister.ReplicaSetLister
+	statefulSetLister           v1appslister.StatefulSetLister
 }
 
 // NewListerRegistry returns a registry providing various listers to list pods or nodes matching conditions
 func NewListerRegistry(allNode NodeLister, readyNode NodeLister, scheduledPod PodLister,
```
losipiuk: each parameter in separate line.

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -107,6 +130,26 @@ func (r listerRegistryImpl) DaemonSetLister() DaemonSetLister {
 	return r.daemonSetLister
 }
 
+// ReplicationControllerLister returns the replicationControllerLister registered to this registry
```
losipiuk: registered in?

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -72,9 +74,24 @@ type daemonSetListerMock struct {
 	mock.Mock
 }
 
-func (l *daemonSetListerMock) List() ([]*extensionsv1.DaemonSet, error) {
+func (l *daemonSetListerMock) List(selector labels.Selector) ([]*appsv1.DaemonSet, error) {
```
losipiuk: Just a general comment. We should not be writing mocks by hand. Mockery saves a lot of work. And generated mocks are actually more versatile. Could you log ticket to move to mockery?

cluster-autoscaler/core/utils.go
```diff
@@ -301,7 +301,7 @@ func GetNodeInfosForGroups(nodes []*apiv1.Node, cloudProvider cloudprovider.Clou
 }
 
 // GetNodeInfoFromTemplate returns NodeInfo object built base on TemplateNodeInfo returned by NodeGroup.TemplateNodeInfo().
-func GetNodeInfoFromTemplate(nodeGroup cloudprovider.NodeGroup, daemonsets []*extensionsv1.DaemonSet, predicateChecker *simulator.PredicateChecker) (*schedulercache.NodeInfo, errors.AutoscalerError) {
```
losipiuk: Did you grep all codebase for `extensionsv1.DaemonSet`?

MaciekPytel:
/hold
For more testing

cluster-autoscaler/core/static_autoscaler_test.go
MaciekPytel: We do that all over CA code. And in case of a dozen parameters that are just variables (like here) I don't like either of those - both 300 char lines and 20 LOC to pass trivial parameters feel like a bad solution.

MaciekPytel: Dunno, I've never seen any nice solution to this in golang.

PR #1541, opened 2018-12-29
Add multi stage Dockerfile

sylr:
Allows to build & package image inside docker.

### DIFF
--- a/cluster-autoscaler/Dockerfile.multi-stage
+++ b/cluster-autoscaler/Dockerfile.multi-stage
@@ -1,0 +1,22 @@
+ARG BASEIMAGE=k8s.gcr.io/debian-base-amd64:0.4.0
+
+FROM golang:1.11-stretch AS builder
+
+ADD . $GOPATH/src/k8s.io/autoscaler/cluster-autoscaler
+WORKDIR $GOPATH/src/k8s.io/autoscaler/cluster-autoscaler
+
+RUN apt-get update && apt-get upgrade -y && apt-get install -y build-essential linux-headers-amd64
+RUN make build
+
+# -----------------------------------------------------
+
+FROM $BASEIMAGE
+LABEL maintainer="Marcin Wielgus <mwielgus@google.com>"
+
+ENV DEBIAN_FRONTEND noninteractive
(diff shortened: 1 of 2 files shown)


Jeffwan:
What's the minimum version of docker it needed?

sylr:
> Multi-stage builds are a new feature requiring Docker 17.05 or higher on the daemon and client. Multistage builds are useful to anyone who has struggled to optimize Dockerfiles while keeping them easy to read and maintain.

https://docs.docker.com/develop/develop-images/multistage-build/

Jeffwan:
Cool. Looks like all latest version will support it except for old docker.io distribution.

losipiuk:
@sylr This is cool thanks. I am just thinking is it really beneficial to have this file if it is not tied to any build flow described via Makefile.
Presumably makefile would still be standard way of building release image of CA.

sylr:
@losipiuk I've added a makefile target which uses this new Dockerfile

losipiuk:
> Łukasz Osipiuk I've added a makefile target which uses this new Dockerfile

Yeah I saw. Thanks.
I am wandering if we still want original make-image target or the new one should just override the old one. It seems there is no reason for having two. @MaciekPytel WDYT?

sylr:
@losipiuk for development purposes it could make sense to keep the current target as it cause the multi-stage building takes a bit of time due to the installation/building of all dependencies.

losipiuk:
> Łukasz Osipiuk for development purposes it could make sense to keep the current target as it cause the multi-stage building takes a bit of time due to the installation/building of all dependencies.

Why would `make-image-multi-stage` take more time that `build-in-docker` and then `make-image`?
I am not saying about getting rid of `build-in-docker` (which is useful if we only want to do the compilation) but not sure if `make-image` is needed. 
If we want to build image for development now we need to compile and then build image. Which should take more-or-less same amount of time as running `make-image-multi-stage`. Right?

sylr:
@losipiuk indeed.

mwielgus:
No activity for 1 month, closing. Please reopen if you really need it.

PR #1543, opened 2018-12-31
Add functions for testing new listers

MaciekPytel:


cluster-autoscaler/utils/kubernetes/testlisters.go
```diff
@@ -0,0 +1,89 @@
...
+	v1appslister "k8s.io/client-go/listers/apps/v1"
+	v1batchlister "k8s.io/client-go/listers/batch/v1"
+	v1lister "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+)
+
+// NewTestDaemonSetLister returns a lister that returns provided DaemonSets
+func NewTestDaemonSetLister(dss []*appsv1.DaemonSet) (v1appslister.DaemonSetLister, error) {
+	store := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+	for _, ds := range dss {
```
losipiuk: seems like all but last line can be extracted to private helper method. Right?

MaciekPytel: Types are different every time. You'd have to convert list of objects to list of interface{}, which would result in similar amount of boilerplate.

PR #1544, opened 2018-12-31
Use listers in scale-down

MaciekPytel:
I'll migrate scale-up in follow up PR

### DIFF
--- a/cluster-autoscaler/utils/drain/drain_test.go
+++ b/cluster-autoscaler/utils/drain/drain_test.go
@@ -119,7 +119,6 @@ func TestDrain(t *testing.T) {
 		},
 	}
 
-
 	statefulset := appsv1.StatefulSet{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "ss",
@@ -130,8 +129,8 @@ func TestDrain(t *testing.T) {
 
 	ssPod := &apiv1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        "bar",
-			Namespace:   "default",
+			Name:            "bar",
+			Namespace:       "default",
 			OwnerReferences: GenerateOwnerReferences(statefulset.Name, "StatefulSet", "apps/v1", ""),
 		},
 	}
--- a/cluster-autoscaler/utils/drain/drain_test.go
+++ b/cluster-autoscaler/utils/drain/drain_test.go
@@ -17,21 +17,21 @@ limitations under the License.
 package drain
 
 import (
-	"fmt"
 	"testing"
 	"time"
 
-	//appsv1beta1 "k8s.io/api/apps/v1beta1"
+	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	apiv1 "k8s.io/api/core/v1"
-	extensions "k8s.io/api/extensions/v1beta1"
 	policyv1 "k8s.io/api/policy/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
+	kube_util "k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
 	. "k8s.io/autoscaler/cluster-autoscaler/utils/test"
-	"k8s.io/client-go/kubernetes/fake"
-	core "k8s.io/client-go/testing"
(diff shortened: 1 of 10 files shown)


cluster-autoscaler/utils/drain/drain_test.go
```diff
@@ -476,75 +476,74 @@ func TestDrain(t *testing.T) {
...
-		// register("statefulsets", &statefulset, statefulset.ObjectMeta)
-
+		var rsLister v1appslister.ReplicaSetLister
 		if len(test.replicaSets) > 0 {
-			register("replicasets", &test.replicaSets[0], test.replicaSets[0].ObjectMeta)
+			rsLister, err = kube_util.NewTestReplicaSetLister(test.replicaSets)
+			assert.NoError(t, err)
 		}
+
+		dsLister, err := kube_util.NewTestDaemonSetLister([]*appsv1.DaemonSet{&ds})
```
losipiuk: Nit: consider changing interface of NewTest*Listers so they take argument list of variable length instead of array.

cluster-autoscaler/utils/drain/drain_test.go
```diff
@@ -476,75 +476,74 @@ func TestDrain(t *testing.T) {
...
-			register("replicationcontrollers", &test.rcs[0], test.rcs[0].ObjectMeta)
+			rcLister, err = kube_util.NewTestReplicationControllerLister(test.rcs)
+			assert.NoError(t, err)
 		}
-		register("daemonsets", &ds, ds.ObjectMeta)
-		register("jobs", &job, job.ObjectMeta)
-		// register("statefulsets", &statefulset, statefulset.ObjectMeta)
-
+		var rsLister v1appslister.ReplicaSetLister
 		if len(test.replicaSets) > 0 {
```
losipiuk: can we get rid of this checks? Are they needed?

cluster-autoscaler/utils/drain/drain_test.go
```diff
@@ -476,75 +476,74 @@ func TestDrain(t *testing.T) {
...
-			fakeClient.Fake.AddReactor("get", resource, func(action core.Action) (bool, runtime.Object, error) {
-				getAction := action.(core.GetAction)
-				if getAction.GetName() == meta.GetName() && getAction.GetNamespace() == meta.GetNamespace() {
-					return true, obj, nil
-				}
-				return false, nil, fmt.Errorf("Not found")
-			})
-		}
+		var err error
+		var rcLister v1lister.ReplicationControllerLister
```
losipiuk: I would prefer `replicationControllerLister` as variable name. Abbreviations are bad!!!!

cluster-autoscaler/utils/drain/drain_test.go
MaciekPytel: I just left this part as it was, but it's a bit of a mess. RC and RS are passed as test param, but other controllers are just always set. I'd rather avoid cleaning up all those issues from tests as it will bloat this PR. I can open an issue to clean it up later, though https://github.com/kubernetes/autoscaler/issues/215 illustrates the problem with that (I only re-enabled the test it mentions now as I was making changes in this area).

cluster-autoscaler/utils/drain/drain_test.go
MaciekPytel: Why? I bet go naming convention likes them (it seems to hate long variable names).

cluster-autoscaler/utils/drain/drain_test.go
losipiuk: I am not opting for fixing everything. But if we can clean something up, as part adding feature, I am for it. I know you prefer the PRs to address only one thing but the issue is that with that approach the cleanup PRs never happen. On the other hand if can clean just a bit, as part of feature-PR, we are a bit more in the better world.

losipiuk:
/lgtm
/hold

Decide what you want to do with nits

MaciekPytel:
/hold cancel
I'll try to clean up in a separate PR for both this and scale-up.

cluster-autoscaler/utils/drain/drain_test.go
losipiuk: This does not mean thay are good.
As long as longer variable name does not make lines very long they there is no point in using abbreviations.
They are just harder to read and makes understanding the code harder. Unless you are reading letter by letter of course....

PR #1545, opened 2018-12-31
Add cloudprovider.GetNodeResources method

losipiuk:


losipiuk:
Closing for now. Will try to facilitate already existent `TemplateNodeInfo` method for constructing `NodeResources`. For node groups with nodes we will use one of those as a representant.

PR #1549, opened 2019-01-02
Use lister in GetNodeInfosForGroups

MaciekPytel:
This significantly improves performance on very large clusters (3x faster in my tests).

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler_test.go
+++ b/cluster-autoscaler/core/static_autoscaler_test.go
@@ -227,7 +227,7 @@ func TestStaticAutoscalerRunOnce(t *testing.T) {
 	// Scale up.
 	readyNodeListerMock.On("List").Return([]*apiv1.Node{n1}, nil).Once()
 	allNodeListerMock.On("List").Return([]*apiv1.Node{n1}, nil).Once()
-	scheduledPodMock.On("List").Return([]*apiv1.Pod{p1}, nil).Once()
+	scheduledPodMock.On("List").Return([]*apiv1.Pod{p1}, nil).Times(2) // 1 to get pods + 1 per nodegroup when building nodeInfo map
 	unschedulablePodMock.On("List").Return([]*apiv1.Pod{p2}, nil).Once()
 	daemonSetListerMock.On("List", labels.Everything()).Return([]*appsv1.DaemonSet{}, nil).Once()
 	onScaleUpMock.On("ScaleUp", "ng1", 1).Return(nil).Once()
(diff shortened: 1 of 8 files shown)


cluster-autoscaler/simulator/nodes.go
```diff
@@ -21,28 +21,25 @@ import (
...
 // GetRequiredPodsForNode returns a list of pods that would appear on the node if the
 // node was just created (like daemonset and manifest-run pods). It reuses kubectl
 // drain command to get the list.
-func GetRequiredPodsForNode(nodename string, client kube_client.Interface) ([]*apiv1.Pod, errors.AutoscalerError) {
-
-	// TODO: we should change this to use informer
-	podListResult, err := client.CoreV1().Pods(apiv1.NamespaceAll).List(
-		metav1.ListOptions{FieldSelector: fields.SelectorFromSet(fields.Set{"spec.nodeName": nodename}).String()})
+func GetRequiredPodsForNode(nodename string, listers kube_util.ListerRegistry) ([]*apiv1.Pod, errors.AutoscalerError) {
+	pods, err := listers.ScheduledPodLister().List()
```
losipiuk: Why not extend ScheduledPodLister expose method with selector and use it here instead doing filtering in for loop?

cluster-autoscaler/simulator/nodes_test.go
```diff
@@ -21,21 +21,23 @@ import (
...
-	"k8s.io/client-go/kubernetes/fake"
-	core "k8s.io/client-go/testing"
+	kube_util "k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
 	"k8s.io/kubernetes/pkg/kubelet/types"
 
 	"github.com/stretchr/testify/assert"
 )
 
 func TestRequiredPodsForNode(t *testing.T) {
+	nodeName := "node1"
```
losipiuk: You could add a pod from different node here too.

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -524,7 +524,7 @@ func TestStaticAutoscalerRunOnceWithALongUnregisteredNode(t *testing.T) {
 	// Scale up.
 	readyNodeListerMock.On("List").Return([]*apiv1.Node{n1}, nil).Times(2) // due to initialized=false
 	allNodeListerMock.On("List").Return([]*apiv1.Node{n1}, nil).Once()
-	scheduledPodMock.On("List").Return([]*apiv1.Pod{p1}, nil).Once()
+	scheduledPodMock.On("List").Return([]*apiv1.Pod{p1}, nil).Times(2) // 1 to get pods + 1 per nodegroup when building nodeInfo map
```
losipiuk: I was looking if there maybe is some `AnyNumberOfTimes`, yet it does not seem so.

cluster-autoscaler/simulator/nodes.go
MaciekPytel: Selector wouldn't allow you to filter by NodeName. We'd have to write our own indexing function, which would be unnecessarily complex and a premature optimization.

PR #1563, opened 2019-01-08
Extend backoff interface with NodeInfo and error information

losipiuk:


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -260,10 +261,10 @@ func (csr *ClusterStateRegistry) backoffNodeGroup(nodeGroup cloudprovider.NodeGr
...
 	csr.Lock()
 	defer csr.Unlock()
-	csr.registerFailedScaleUpNoLock(nodeGroup, reason, currentTime)
+	csr.registerFailedScaleUpNoLock(nodeGroup, reason, cloudprovider.OtherErrorClass, "", currentTime)
 }
 
-func (csr *ClusterStateRegistry) registerFailedScaleUpNoLock(nodeGroup cloudprovider.NodeGroup, reason metrics.FailedScaleUpReason, currentTime time.Time) {
+func (csr *ClusterStateRegistry) registerFailedScaleUpNoLock(nodeGroup cloudprovider.NodeGroup, reason metrics.FailedScaleUpReason, errorClass cloudprovider.InstanceErrorClass, errorCode string, currentTime time.Time) {
 	metrics.RegisterFailedScaleUp(reason)
 	csr.backoffNodeGroup(nodeGroup, currentTime)
```
MaciekPytel: What's the point of all this if you're not passing errorClass and errorCode you get?

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -221,7 +221,7 @@ func (csr *ClusterStateRegistry) updateScaleRequests(currentTime time.Time) {
 			// scale-out finished successfully
 			// remove it and reset node group backoff
 			delete(csr.scaleUpRequests, nodeGroupName)
-			csr.backoff.RemoveBackoff(scaleUpRequest.NodeGroup)
+			csr.backoff.RemoveBackoff(scaleUpRequest.NodeGroup, csr.nodeInfosForGroups[scaleUpRequest.NodeGroup.Id()])
```
MaciekPytel: This is just sanity check - you know you may end up passing nil nodeInfo here? I think it's ok and just needs to be handled in backoff implementation. 

Also it got me thinking - what will happend if nodegroup is removed / re-configured to be no longer autoscaled during scale-up? Can this function handle scaleUp request existing for a no-longer existing NodeGroup?

cluster-autoscaler/clusterstate/clusterstate.go
losipiuk: Yeah - that is not how it should look like ...

cluster-autoscaler/clusterstate/clusterstate.go
losipiuk: > This is just sanity check - you know you may end up passing nil nodeInfo here?

Yeah - I know that. The backoff needs to handle calls with nil NodeInfo gracefully.

> Also it got me thinking - what will happend if nodegroup is removed / re-configured to be no longer autoscaled during scale-up? Can this function handle scaleUp request existing for a no-longer existing NodeGroup?

Good point. Seems like potential bug. I will file an issue.

losipiuk: #1565

PR #1564, opened 2019-01-08
GCE CA: use public selfLink URL format.

elithrar:
- gcloud commands & APIs return "www.googleapis.com", not
"content.googleapis.com" as the selfLink. This documentation & error message
change makes this more obvious.
- Improve checks & tests around valid expectedResource names.

---

Example output that end-users see:

```
➜  gcloud compute instance-groups describe gke-your-first-cluster-1-pool-1-4f8729cc-grp --zone=us-central1-a
creationTimestamp: '2019-01-08T10:05:47.710-08:00'
description: "This instance group is controlled by Instance Group Manager 'gke-your-first-cluster-1-pool-1-4f8729cc-grp'.\
  \ To modify instances in this group, use the Instance Group Manager API: https://cloud.google.com/compute/docs/reference/latest/instanceGroupManagers"
fingerprint: 42WmSpB8rSM=
id: '5710522466650642580'
kind: compute#instanceGroup
name: gke-your-first-cluster-1-pool-1-4f8729cc-grp
network: https://www.googleapis.com/compute/v1/projects/silverlock-sandbox/global/networks/default
selfLink: https://www.googleapis.com/compute/v1/projects/silverlock-sandbox/zones/us-central1-a/instanceGroups/gke-your-first-cluster-1-pool-1-4f8729cc-grp
size: 1
subnetwork: https://www.googleapis.com/compute/v1/projects/silverlock-sandbox/regions/us-central1/subnetworks/default
zone: https://www.googleapis.com/compute/v1/projects/silverlock-sandbox/zones/us-central1-a
```

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/gce_url.go
+++ b/cluster-autoscaler/cloudprovider/gce/gce_url.go
@@ -17,6 +17,7 @@ limitations under the License.
 package gce
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 )
@@ -24,31 +25,31 @@ import (
 const (
 	gceUrlSchema        = "https"
 	gceDomainSuffix     = "googleapis.com/compute/v1/projects/"
-	gcePrefix           = gceUrlSchema + "://content." + gceDomainSuffix
+	gcePrefix           = gceUrlSchema + "://www." + gceDomainSuffix
 	instanceUrlTemplate = gcePrefix + "%s/zones/%s/instances/%s"
 	migUrlTemplate      = gcePrefix + "%s/zones/%s/instanceGroups/%s"
 )
 
 // ParseMigUrl expects url in format:
-// https://content.googleapis.com/compute/v1/projects/<project-id>/zones/<zone>/instanceGroups/<name>
+// https://www.googleapis.com/compute/v1/projects/<project-id>/zones/<zone>/instanceGroups/<name>
 func ParseMigUrl(url string) (project string, zone string, name string, err error) {
 	return parseGceUrl(url, "instanceGroups")
 }
 
 // ParseIgmUrl expects url in format:
-// https://content.googleapis.com/compute/v1/projects/<project-id>/zones/<zone>/instanceGroupManagers/<name>
+// https://www.googleapis.com/compute/v1/projects/<project-id>/zones/<zone>/instanceGroupManagers/<name>
 func ParseIgmUrl(url string) (project string, zone string, name string, err error) {
 	return parseGceUrl(url, "instanceGroupManagers")
 }
 
 // ParseInstanceUrl expects url in format:
-// https://content.googleapis.com/compute/v1/projects/<project-id>/zones/<zone>/instances/<name>
+// https://www.googleapis.com/compute/v1/projects/<project-id>/zones/<zone>/instances/<name>
 func ParseInstanceUrl(url string) (project string, zone string, name string, err error) {
 	return parseGceUrl(url, "instances")
 }
 
 // ParseInstanceUrlRef expects url in format:
-// https://content.googleapis.com/compute/v1/projects/<project-id>/zones/<zone>/instances/<name>
+// https://www.googleapis.com/compute/v1/projects/<project-id>/zones/<zone>/instances/<name>
 // and returns a GceRef struct for it.
 func ParseInstanceUrlRef(url string) (GceRef, error) {
 	project, zone, name, err := parseGceUrl(url, "instances")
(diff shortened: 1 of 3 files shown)


elithrar:
Re-pushed with corp email address.

cluster-autoscaler/cloudprovider/gce/gce_url_test.go
```diff
@@ -23,21 +23,54 @@ import (
...
+	assert.Equal(t, "mwielgus-proj", proj)
+	assert.Equal(t, "us-central1-b", zone)
+	assert.Equal(t, "kubernetes-minion-group", name)
+
+	// Invalid TLD
+	_, _, _, err = parseGceUrl("https://www.googleapis.net/compute/v1/projects/mwielgus-proj/zones/us-central1-b/instanceGroups/kubernetes-minion-group", "instanceGroups")
+	assert.NotNil(t, err)
+
+	// Empty 'expectedResource'
+	_, _, _, err = parseGceUrl("googleapis.net/compute/v1/projects/mwielgus-proj/zones/us-central1-b/instanceGroups/kubernetes-minion-group", "")
```
aleksandra-malinowska: This one also has invalid TLD?

cluster-autoscaler/cloudprovider/gce/gce_url_test.go
```diff
@@ -23,21 +23,54 @@ import (
...
+	assert.NotNil(t, err)
+
+	// Invalid domain & URL
+	_, _, _, err = parseGceUrl("www.onet.pl", "instanceGroups")
 	assert.NotNil(t, err)
 
-	proj, zone, name, err = parseGceUrl("https://content.googleapis.com/compute/vabc/projects/mwielgus-proj/zones/us-central1-b/instanceGroups/kubernetes-minion-group", "instanceGroups")
+	// Validate the URL suffix
```
aleksandra-malinowska: s/suffix/API version?

cluster-autoscaler/cloudprovider/gce/gce_url.go
```diff
@@ -72,8 +73,14 @@ func GenerateMigUrl(ref GceRef) string {
 	return fmt.Sprintf(migUrlTemplate, ref.Project, ref.Zone, ref.Name)
 }
 
+// parseGceUrl parses a GCE Compute API URL into the project ID, zone name and resource name.
 func parseGceUrl(url, expectedResource string) (project string, zone string, name string, err error) {
-	errMsg := fmt.Errorf("Wrong url: expected format https://content.googleapis.com/compute/v1/projects/<project-id>/zones/<zone>/%s/<name>, got %s", expectedResource, url)
+	errMsg := fmt.Errorf("Wrong URL: expected format https://www.googleapis.com/compute/v1/projects/<project-id>/zones/<zone>/%s/<name>, got %s", expectedResource, url)
+
+	if expectedResource == "" {
```
aleksandra-malinowska: This check is IMO redundant, but if we're going to keep it, let's rephrase the message to sth more like wrong resource error message. Logging what we got and what we expected is usually a lot more useful than a generic error message (which is completely different from anything else we have in the code, btw, this is not user-facing API). 

Also, it seems errors.New() and fmt.Errorf() are basically identical, can we stick to using just one everywhere?

cluster-autoscaler/cloudprovider/gce/gce_url_test.go
aleksandra-malinowska: Also, this only tests for extra error check against empty resource. Can you add a test case for unexpected/wrong resource?

cluster-autoscaler/cloudprovider/gce/gce_url.go
elithrar: Noted RE: error message intent.

- I explicitly did not use `fmt.Errorf` as I wasn't using a format string, and that'd be a linter error.
- I'll add an explicit check for `instances` and `instanceGroups`
- Likely, I'll take this opportunity to improve the URL parsing via `net/url.Parse` so that we can correctly validate scheme vs. host vs. path components.

cluster-autoscaler/cloudprovider/gce/gce_url_test.go
elithrar: Good catch, will fix.

cluster-autoscaler/cloudprovider/gce/gce_url_test.go
elithrar: Will fix, should be two tests.

mwielgus:
The request for fixing tests has not been handled for almost 1 month, closing. Please reopen once you have a chance to work on the tests.

Issue #1135, opened 2018-08-10
Could GPU support move to cloud provider?


ringtail:
autoscaler has GKE GPU support and I want to add alicloud GPU to autoscaler. But I found the code is outside the scope of cloud provider, Is there any plan to move those code to cloud provider?

MaciekPytel:
That's a very reasonable request, but I'd like to understand some details to figure out how this could be done.

Basically, most of gpu.go is a one big hack for the problem discussed in https://github.com/kubernetes/autoscaler/pull/597#issuecomment-362231521. Ideally, once the problem is fixed upstream, this whole file should be removed and there should be no special handling for GPU required on cloudprovider side.
That being said the issue upstream has been there for almost a year now and we don't know when it'll finally get fixed. The question is how are you planning to solve it for alicloud? You need some way to recognize a node will have GPU, before the resource shows up in Node requests. In GKE we use the fact that every node with GPU always has a `cloud.google.com/gke-accelerator` label. Is there some similar label in alicloud that you're planning to use? Or is it something completely different?

ringtail:
Yes,We do have the similar label `aliyun.accelerator/nvidia_name`. We need to check wether a node is a gpu node and what is the gpu type of this node. We can abstract two interface function in cloud provider with nodeInfo as param. one is to judge weather a node has gpu and the other is to get the gpu type of this node.

aleksandra-malinowska:
Technically, wouldn't TemplateNodeInfo work for this? I.e., core autoscaler algorithm could get the template node for each node group at the beginning of the iteration. Then, check each node for GPUs and consider it not ready (for all cloud providers) if the template has GPUs, but the node doesn't. Node template has to have correct GPU information for scaling from 0 to work already; the only thing to do would be to optimize (cache) getting template nodes, so it doesn't require N=number of node groups extra API calls each loop.

MaciekPytel:
IIRC for the base GPU implementation all you need is FilterOutNodesWithUnreadyGpus. As a short term fix you could probably just change that function to look at 2 different GPU labels instead of one. 

Moving the implementation to cloudprovider makes sense, but it may take some time before we have time to do it. Also all this gpu code is only a temporary hack, so it would only be a temporary addition until the upstream issue is fixed.

GetNodeTargetGpus is only used for resource limits (`--gpu-total`) which is experimental, not very well tested and not used anywhere in prod, so I'm not sure how much you care about that.

ringtail:
[...]

PR #1584, opened 2019-01-16
Move GPULabel and GPUTypes to cloud provider

Jeffwan:
This PR is to address #1135 

1. Add GPULabel() and GetAvailableGPUTypes() in cloudprodiver interface 
2. Caller will pass GPULabel to methods in gpu.go which original relies on GPULabel
3. Pass cloud.CloudProvider to methods in scale_up.go and scale_down.go that don't use GPULabel directly. 

Please don't merge this PR until I confirm with all cloud provider owners and check if they're ok about labels and supported GPU types.

### DIFF
--- a/cluster-autoscaler/expander/price/price.go
+++ b/cluster-autoscaler/expander/price/price.go
@@ -40,7 +40,7 @@ import (
 // **********
 
 type priceBased struct {
-	pricingModel          cloudprovider.PricingModel
+	cloudProvider         cloudprovider.CloudProvider
 	preferredNodeProvider PreferredNodeProvider
 	nodeUnfitness         NodeUnfitness
 }
@@ -75,12 +75,12 @@ var (
 )
 
 // NewStrategy returns an expansion strategy that picks nodes based on price and preferred node type.
-func NewStrategy(pricingModel cloudprovider.PricingModel,
+func NewStrategy(cloudProvider cloudprovider.CloudProvider,
 	preferredNodeProvider PreferredNodeProvider,
 	nodeUnfitness NodeUnfitness,
 ) expander.Strategy {
 	return &priceBased{
-		pricingModel:          pricingModel,
+		cloudProvider:         cloudProvider,
 		preferredNodeProvider: preferredNodeProvider,
 		nodeUnfitness:         nodeUnfitness,
 	}
@@ -98,7 +98,13 @@ func (p *priceBased) BestOption(expansionOptions []expander.Option, nodeInfos ma
 		klog.Errorf("Failed to get preferred node, switching to default: %v", err)
 		preferredNode = defaultPreferredNode
 	}
-	stabilizationPrice, err := p.pricingModel.PodPrice(priceStabilizationPod, now, then)
+
+	pricingModel, err := p.cloudProvider.Pricing()
+	if err != nil {
+		klog.Errorf("Failed to get pricing model from cloud provider: %v", err)
+	}
+
+	stabilizationPrice, err := pricingModel.PodPrice(priceStabilizationPod, now, then)
 	if err != nil {
 		klog.Errorf("Failed to get price for stabilization pod: %v", err)
 		// continuing without stabilization.
@@ -111,7 +117,7 @@ nextoption:
 			klog.Warningf("No node info for %s", option.NodeGroup.Id())
 			continue
 		}
-		nodePrice, err := p.pricingModel.NodePrice(nodeInfo.Node(), now, then)
+		nodePrice, err := pricingModel.NodePrice(nodeInfo.Node(), now, then)
 		if err != nil {
 			klog.Warningf("Failed to calculate node price for %s: %v", option.NodeGroup.Id(), err)
 			continue
@@ -119,7 +125,7 @@ nextoption:
 		totalNodePrice := nodePrice * float64(option.NodeCount)
 		totalPodPrice := 0.0
 		for _, pod := range option.Pods {
-			podPrice, err := p.pricingModel.PodPrice(pod, now, then)
+			podPrice, err := pricingModel.PodPrice(pod, now, then)
 			if err != nil {
 				klog.Warningf("Failed to calculate pod price for %s/%s: %v", pod.Namespace, pod.Name, err)
 				continue nextoption
@@ -141,7 +147,7 @@ nextoption:
 
 		// Set constant, very high unfitness to make them unattractive for pods that doesn't need GPU and
 		// avoid optimizing them for CPU utilization.
-		if gpu.NodeHasGpu(nodeInfo.Node()) {
+		if gpu.NodeHasGpu(p.cloudProvider.GPULabel(), nodeInfo.Node()) {
 			klog.V(4).Infof("Price expander overriding unfitness for node group with GPU %s", option.NodeGroup.Id())
 			supressedUnfitness = gpuUnfitnessOverride
 		}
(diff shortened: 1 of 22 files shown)


ringtail:
@Jeffwan in Alibaba Cloud the label is `aliyun.accelerator/nvidia_name`

ringtail:
@Jeffwan Thanks for shimming code for Alibaba Cloud. Could we reach an agreement with the interface and develop the code by ourself. Because some logic may be different between providers.

hello2mao:
@Jeffwan no special GPU label for Baidu Cloud, thanks.

cluster-autoscaler/expander/price/price.go
```diff
@@ -39,6 +39,7 @@ import (
 // **********
 
 type priceBased struct {
+	cloudProvider         cloudprovider.CloudProvider
```
aleksandra-malinowska: We probably don't need both cloudProvider and pricingModel (which can be obtained from cloudProvider) here.

Disregard this comment if you choose to follow the suggestion with GpuInfo struct, of course.

Jeffwan: True. I will clean it up.

Jeffwan:
> Where error can be `cloudprovider.ErrNotImplemented` for cloudproviders not supporting GPUs
Agree. The only concern I have is `gpu.go` is an utility and no matter CloudProvider supports it or not, scaling logic go through helper functions now. And even cloud provider doesn't support it, I know some of the users label node and use it in a hack way. 
After we use error to indicate the GPU support, process will be changed. Using default value is a compromise way. If we want to use struct or interface to capture all GPU infos, err would be necessary. 

> Also one more nit. As we are changing this part of code we can also change availableGPUTypes
to be map[string]bool. The lookup value can then be used directly and we do not have to unpack the second tuple value to check if key is found in the map.

The reason not using bool is in the future CA potential wants to know more Info for specific GPU type, like GPU memory to calculate utilization. (if sharing GPU is supported later).  This might be a good case to make GPUInfo more extensible.

Jeffwan:
@ringtail If we‘d like to step further, I think we can remove gpu.go and move everything in cloud providers or withhold some interface in cloud provider like GPUInfo() and reference it in main logic (scale_up and scale_down). Could you give some examples or use cases?

MaciekPytel:
The current approach with returning label or simple gpu info is perfectly fine, though I agree with @losipiuk there should be a more explicit way to say 'not implemented' than returning gke label.

Maybe the way to proceed would be to allow ErrNotImplemented to make it explicit it's not supported by a given cloudprovider and let cloudprovider owners do their own implementation. To keep compatibility for users who add gke label manually we can just handle cloudprovider returning ErrNotImplemented by defaulting to gke label (it's ugly, but it keeps current behavior while still being explicit about what is officially supported by given cloudprovider and what is not).

---

I'm against moving utils/gpu logic into cloudprovider though. cloudprovider is meant to be a client for communicating with underlying platform. FilterOutNodesWithUnreadyGpus is a hack, but it's also a complex part of CA logic based on knowledge of internal implementation (it's abusing how clusterstate is implemented to hack around the driver installation problem). Moving this kind of logic to cloudprovider means we have to start treating a lot of internal CA logic as a quasi-API, as we do with cloudprovider today.

This kind of complex, provider-specific logic is why we have Processors. If we think we really need custom GPU-hacks I'd suggest moving StaticAutoscaler.obtainNodeLists() into a new processor and implementing custom logic there. (If someone were to actually do this -  I'm really interested to learn how you're planning to deal with the problem - it took us a while to come up with the current hack).

Jeffwan:
Thanks. I will make the change to support ErrNotImplemented. For moving GPU utils/gpu logic into cloudprovider, unless we get reasonable requirements, it will not be considered in this PR's scope

mwielgus:
@Jeffwan @MaciekPytel @aleksandra-malinowska  - Where are we with this PR? Who is waiting for who :)

MaciekPytel:
Sorry for leaving this hanging for so long. I don't think all outstanding comments have been applied yet - in particular we should decide if we want to keep gke-accelarator label working on cloudproviders that have their own label to avoid breaking people who are already using it as a workaround? Current version will work fine for people on EKS, but it may break people running self-hosted clusters on AWS. 
Alternatively we can leave this as is and put it into 'Action required' part of Kubernetes release notes for the release it lands in (1.14 if we merge this in time).

Jeffwan:
> To keep compatibility for users who add gke label manually we can just handle cloudprovider returning ErrNotImplemented by defaulting to gke label (it's ugly, but it keeps current behavior while still being explicit about what is officially supported by given cloudprovider and what is not).

@MaciekPytel I sent you a message on slack to ask for this. I am not sure how to `returning ErrNotImplemented by defaulting to gke label` ? You mean for cloud providers doesn't have this label, we return something like this? 

func (aws *awsCloudProvider) Name() (string, errors.AutoscalerError) {
    return "cloud.google.com/gke-accelerator", cloudprovider.ErrNotImplemented
}

feiskyer:
There's no label for GPU type on Azure yet, does this change still supports the default device-plugin based GPU scaling?

Jeffwan:
> There's no label for GPU type on Azure yet, does this change still supports the default device-plugin based GPU scaling?

Yes. It won't change default behavior, just for cloud provider which has its own label, otherwise, default label is still being used.

ringtail:
> > There's no label for GPU type on Azure yet, does this change still supports the default device-plugin based GPU scaling?
> 
> Yes. It won't change default behavior, just for cloud provider which has its own label, otherwise, default label is still being used.

So,What's the status of this PR. Could we follow your interface to implement?

MaciekPytel:
> I am not sure how to returning ErrNotImplemented by defaulting to gke label ? You mean for cloud providers doesn't have this label, we return something like this? The logic calling these two functions assume they always return value back, how should we use ErrNotImplemented there?

If we wanted to keep backward compatibility we'd need to respect gke label even if cloudprovider defines a different one (ie. AWS would respect both EKS and GKE labels). There are already people running on AWS who depend on GKE label (ex. https://github.com/kubernetes/autoscaler/issues/1659). I'd imagine that would be implemented in the code calling the cloudprovider method (ex. in NodeHasGpu, etc).
That being said I'm not sure how many people actually use it - maybe we can just handle it with "action required" release note?

@ringtail Is the current interface (GPULabel() + GetAvailableGPUTypes()) sufficient for you?

Jeffwan:
> That being said I'm not sure how many people actually use it - maybe we can just handle it with "action required" release note?

Given "action required" release note is reasonable. This will affect some users using CA master branch.

ringtail:
> GetAvailableGPUTypes

Yes, we use custom label and it seems to work for me.

mwielgus:
@MaciekPytel @Jeffwan @ringtail - Where we are with this PR. Who is waiting for what/who?

ringtail:
> @MaciekPytel @Jeffwan @ringtail - Where we are with this PR. Who is waiting for what/who?

@mwielgus I think I am waiting this pr to merge so that I can implement for my provider. I don't know what this pr is waiting for.

Jeffwan:
rebase the upstream change.

MaciekPytel:
/approve
/lgtm
Sorry to keep it waiting so long.

Jeffwan:
No problem. I rebase upstream changes and need another /lgtm. 
@MaciekPytel  @mwielgus

PR #1585, opened 2019-01-16
VPA - protection against malformed pod status in OomObserver

schylek:


schylek:
For issue #1585

schylek:
fixes #1585

PR #1587, opened 2019-01-16
Tainting unneeded nodes as PreferNoSchedule

jkaniuk:


cluster-autoscaler/utils/deletetaint/delete_test.go
```diff
@@ -45,7 +45,40 @@ func TestMarkNodes(t *testing.T) {
...
 
+func TestSoftMarkNodes(t *testing.T) {
+	node := BuildTestNode("node", 1000, 1000)
+	fakeClient := buildFakeClient(t, node)
+	err := MarkDeletionCandidate(node, fakeClient)
+	assert.NoError(t, err)
+
+	updatedNode, err := fakeClient.Core().Nodes().Get("node", metav1.GetOptions{})
+	assert.NoError(t, err)
+	assert.False(t, HasToBeDeletedTaint(updatedNode))
```
aleksandra-malinowska: nit: what is this scenario intended to test? Looks like it may make sense to assert soft taint is on the node here?

cluster-autoscaler/utils/deletetaint/delete_test.go
```diff
@@ -60,20 +93,57 @@ func TestCheckNodes(t *testing.T) {
 	assert.True(t, time.Now().Sub(*val) < 10*time.Second)
 }
 
+func TestSoftQueryNodes(t *testing.T) {
```
aleksandra-malinowska: nit: how is this scenario intended to be different from TestSoftMarkNodes, other than making a couple more assertions?

aleksandra-malinowska:
A few nits, otherwise looks good. Just as a sanity check, can you describe how you tested it?

cluster-autoscaler/core/scale_down.go
```diff
@@ -572,6 +572,46 @@ func (sd *ScaleDown) mapNodesToStatusScaleDownNodes(nodes []*apiv1.Node, nodeGro
...
+			// Do not consider nodes that are scheduled to be deleted
+			continue
+		}
+		alreadyTainted := deletetaint.HasDeletionCandidateTaint(node)
+		_, unneeded := sd.unneededNodes[node.Name]
+
+		if unneeded != alreadyTainted {
+			apiCallBudget--
+			if apiCallBudget < 0 {
+				continue
```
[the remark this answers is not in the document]
jkaniuk: a) better observability, b) on idle we scan through all of that, so optimizations during only burst down scale seems premature

cluster-autoscaler/utils/deletetaint/delete_test.go
jkaniuk: First scenario tests Mark and Has (so baseline scenario), second tests Get. I could remove Has call to make it more obvious, but Has is a general assertion strategy in all of those tests.

cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -32,6 +32,8 @@ type GpuLimits struct {
 
 // AutoscalingOptions contain various options to customize how autoscaling works
 type AutoscalingOptions struct {
+	// MaxBulkSoftTaint is a number of nodes that can be tainted PreferNoSchedule at the same time.
```
losipiuk: The wording does not make it obvious if this is about:
- how many nodes can have a taint at any single time
- or how many nodes can be modified with a taint in one go.

Please change it so it is more understandable.

cluster-autoscaler/utils/deletetaint/delete_test.go
jkaniuk: Also I have included additional basic scenario. TestMark tests with Mark, TestCheck tests with addTaintToSpec. Other tests uses addTaintToSpec as a building block so I felt that it would be beneficial to test it separately

cluster-autoscaler/config/autoscaling_options.go
aleksandra-malinowska: > how many nodes can have a taint at any single time

I don't think the current version limits it at all. Perhaps it should?

cluster-autoscaler/core/scale_down.go
```diff
@@ -572,6 +572,46 @@ func (sd *ScaleDown) mapNodesToStatusScaleDownNodes(nodes []*apiv1.Node, nodeGro
...
+	// Check if some new nodes should be tainted
+	for _, node := range allNodes {
+		if deletetaint.HasToBeDeletedTaint(node) {
+			// Do not consider nodes that are scheduled to be deleted
+			continue
+		}
+		alreadyTainted := deletetaint.HasDeletionCandidateTaint(node)
+		_, unneeded := sd.unneededNodes[node.Name]
+
+		if unneeded != alreadyTainted {
```
losipiuk: Maybe it is just me but I would prefer to have this if split into two:

```
if unneeded && !alreadyTainted  ...
if !unneeded && alreadyTainted ...
```

cluster-autoscaler/core/scale_down.go
```diff
@@ -572,6 +572,46 @@ func (sd *ScaleDown) mapNodesToStatusScaleDownNodes(nodes []*apiv1.Node, nodeGro
...
+			apiCallBudget--
+			if apiCallBudget < 0 {
+				continue
+			}
+			if unneeded {
+				// Node unneeded but not tainted
```
losipiuk: If you follow my suggestion above you can drop this comment and one below :)

cluster-autoscaler/core/scale_down.go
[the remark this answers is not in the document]
aleksandra-malinowska: Can you elaborate on (a)? I'm not sure how it contributes to observability, so I'd like to learn about it :)

As for (b), it's true we don't need the optimization, it just seems like the code would be simpler.

cluster-autoscaler/utils/deletetaint/delete_test.go
jkaniuk: Bug spotted, I have fixed it

cluster-autoscaler/core/scale_down.go
jkaniuk: But there is a common block at the top of that if ...

cluster-autoscaler/core/scale_down.go
```diff
@@ -572,6 +572,46 @@ func (sd *ScaleDown) mapNodesToStatusScaleDownNodes(nodes []*apiv1.Node, nodeGro
...
+				cleaned, err := deletetaint.CleanDeletionCandidate(node, sd.context.ClientSet)
+				if err != nil || !cleaned {
+					errors = append(errors, err)
+					klog.Warningf("Soft taint on %s removal error %v", node.Name, err)
+				}
+			}
+		}
+	}
+	if apiCallBudget < 0 {
+		klog.V(4).Infof("Skipped soft taints on %v nodes - API call limit exceeded", -apiCallBudget)
```
losipiuk: "Skipped adding/removing soft taints ...." ?

cluster-autoscaler/core/scale_down.go
jkaniuk: I had that previously, I have refactored into current form to account for apiCallBudget.

cluster-autoscaler/core/scale_down.go
jkaniuk: ok

cluster-autoscaler/config/autoscaling_options.go
jkaniuk: done

cluster-autoscaler/core/scale_down.go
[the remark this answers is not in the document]
losipiuk: a) I guess it is about log message below

cluster-autoscaler/core/scale_down.go
losipiuk: Yeah - still I would rather have it repeated.

cluster-autoscaler/core/scale_down.go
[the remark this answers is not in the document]
aleksandra-malinowska: Good point, thanks

jkaniuk:
> A few nits, otherwise looks good. Just as a sanity check, can you describe how you tested it?

GKE v1.12.3-gke.1 with replaced cluster-autoscaling.
Passed tests:
scaling pods down - checking if on unneeded node, taint is added
scaling pods down and up - checking if taint is removed if node is needed again
scaling pods down and waiting for deletion - checking if taints are not added once again when node is in process of being decommisioned
checking if with bulk limit = 1 nodes are tainted and untainted one by one on each pass
checking if with default bulk limit = 30, all nodes are tainted and untainted one by one on each pass (I have not tested with more than 5 to add/remove)
restarting cluster-autoscaling when some nodes are tainted but not needed - checking if they will be untainted
restarting cluster-autoscaling when some nodes are unneeded - checking if they will be tainted

cluster-autoscaler/config/autoscaling_options.go
jkaniuk: @aleksandra-malinowska - we could add it as a second parameter, but how to calculate such limit?

cluster-autoscaler/core/scale_down.go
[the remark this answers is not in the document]
jkaniuk: klog.V(4).Infof("Skipped adding/removing soft taints on %v nodes - API call limit exceeded", -apiCallBudget)

cluster-autoscaler/config/autoscaling_options.go
aleksandra-malinowska: It could be a percentage of nodes, and we could keep track of how many we've tainted. But then we'd probably want to taint them in some order (first the empty ones, then candidates), so let's keep it as it is for now and discuss later.

aleksandra-malinowska:
/lgtm

Leaving it to @losipiuk to approve. Thanks!

cluster-autoscaler/core/scale_down_test.go
```diff
@@ -1339,3 +1351,142 @@ func TestCheckScaleDownDeltaWithinLimits(t *testing.T) {
...
+	fakeClient := &fake.Clientset{}
+
+	job := batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "job",
+			Namespace: "default",
+			SelfLink:  "/apivs/batch/v1/namespaces/default/jobs/job",
+		},
+	}
+	n1 := BuildTestNode("n1", 2000, 2000)
```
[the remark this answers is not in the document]
jkaniuk: done

cluster-autoscaler/core/scale_down.go
jkaniuk: done

cluster-autoscaler/config/autoscaling_options.go
MaciekPytel: This thread reflects all my thoughts:
1. When I saw the field I was 100% sure it was about max number of tainted nodes, not rate limiting. I was very surprised when I read the code.
1. As @aleksandra-malinowska wrote I'm not convinced we should be marking every candidate like that. We know each candidate can be deleted, but that in no way implies all of them can be deleted. In a small cluster majority of nodes can be candidates and tainting most nodes defeats the purpose of tainting them in the first place. Tainting just some candidates requires some persistent ordering of the candidates of course.
@aleksandra-malinowska's ideas sound good to me - percentage of all nodes as a max threshold. Not perfect but consistent order can be achieved by just sorting candidates - already tainted go first, empty second (or should it be the other way round?), followed by the rest. WDYT?

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -379,6 +379,8 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 		} else {
 			klog.V(4).Infof("Starting scale down")
 
+			scaleDown.SoftTaintUnneededNodes(allNodes)
```
MaciekPytel: Can you add a metric for how long this takes (ie. metrics.UpdateDurationFromStart)? This potentially involves multiple API calls, so it's worth keeping track of the performance impact.

MaciekPytel: Also - any specific reason this is called here? It's not obvious to me why IsDeleteInProgress() should stop us from soft tainting other nodes.

cluster-autoscaler/config/autoscaling_options.go
losipiuk: >Aleksandra Malinowska's ideas sound good to me - percentage of all nodes as a max threshold. 

Percentage of all nodes seems like a good starting point. Yet I would love have some metric which would tell us in the future if the percentage (or other heuristic) we picked is a good one or should be changed. This could be one of CA KPIs.

losipiuk:
/lgtm
/approve
/hold

I think this is good to go in current shape and I suggest you add metrics suggested by @MaciekPytel and me in followup commit.
Also the possiblity to specify max percentage of nodes which can be soft tainted can be done as a followup as we reach agreement how it should look like.
Feel free to unhold as you address the "Also - any specific reason this is called here?" comment.

cluster-autoscaler/config/autoscaling_options.go
aleksandra-malinowska: What about sth like tracking when nodes stop being unneeded according to us? Then we'd know what's the likelihood that given node will be removed in the end, and how taints affect this.

cluster-autoscaler/core/static_autoscaler.go
jkaniuk: I have added scaleDown:softTaintUnneeded metric with combined time of that loop.

It turns out that with default limit of 30 requests, it takes up to 10s.
Why: on scaling down, API server burts requests limit could be drained for pod removal request, after that API server limits to 10 qps, which translates to 5 soft taints (each take GET and PUT).
For each subsequent requests, around 200ms has to pass, so one tainting per 400ms.
I have changed default limit to 10 requests per autoscaler loop to keep up with bulk deletion. 

I plan to add another PR changing tainting retry loop from GET + PUT, to try PUT, then GET on failure. That will reduce number of requests.

jkaniuk: As for the place to put SoftTaintUnneededNodes:
- it should not slow down scaleUp (it is only soft tainting)
- there is flag calculateUnneededOnly marked "In dry run only utilization is updated" - I have put softTaint to run when it is false
- it does not run when scaleDownForbidden or ScaleDownInCooldown, as then we do not have full understanding of situation
- it does not run when IsDeleteInProgress to not conflict with given node removed from unneeded nodes during node deletion (in other words - to be sure that taint is not removed in edge cases)

MaciekPytel: Sounds good to me. 

Another thing to consider in follow up PR - if it can take long, maybe we should also check elapsed time as loop condition?

MaciekPytel:
/hold
Actually, just one more thing - can I disable that with feature with flags? I guess I sort of can by setting the flag to 0? But it will still go through some extra logic - are there any side effects? If setting to 0 is the intended way to disable can that be included in flag description?

cluster-autoscaler/core/static_autoscaler.go
jkaniuk: We could. Do you have any specific timeout in mind? Like 10s? Adding additional flag seems an overkill.

MaciekPytel: Agreed we can go with const for a timeout. I'd say 10s is too much - ideally the whole loop should take no more than 10s (not true on large clusters, but one can always dream). Something like 3s sound more reasonable to me. If we're waiting 10 minutes to delete a node we can taint it one or two loops later.

jkaniuk: I have added time limiting as well

cluster-autoscaler/core/scale_down.go
```diff
@@ -76,6 +78,9 @@ type NodeDeleteStatus struct {
 	nodeDeleteResults map[string]error
 }
 
+// Proxy for unit tests
+var getCurrentTime func() time.Time = time.Now
```
losipiuk: maybe just `now`?

cluster-autoscaler/core/scale_down.go
```diff
@@ -64,6 +64,8 @@ const (
 	// PodEvictionHeadroom is the extra time we wait to catch situations when the pod is ignoring SIGTERM and
 	// is killed with SIGKILL after MaxGracefulTerminationTime
 	PodEvictionHeadroom = 30 * time.Second
+	// MaxSoftTaintTime limits single run of SoftTaintUnneededNodes
+	MaxSoftTaintTime = 3 * time.Second
```
losipiuk: The comment is fine, but variable name is misleading. I would name it `SoftTaintUnnededNodesRuntimeLimit`

cluster-autoscaler/core/scale_down.go
```diff
@@ -572,6 +575,50 @@ func (sd *ScaleDown) mapNodesToStatusScaleDownNodes(nodes []*apiv1.Node, nodeGro
...
+			continue
+		}
+		alreadyTainted := deletetaint.HasDeletionCandidateTaint(node)
+		_, unneeded := sd.unneededNodes[node.Name]
+
+		// Check if expected taints match existing taints
+		if unneeded != alreadyTainted {
+			if apiCallBudget <= 0 || now().Sub(startTime) >= timeBudget {
+				skippedNodes++
+				continue
```
MaciekPytel: Is there any reason not to return here? I'm not convinced logging number of skipped nodes is really worth it.

jkaniuk: Please look at the discussion with Ola above

jkaniuk: but basically: on idle we scan through all of that, so optimizations during only burst down scale seems premature

MaciekPytel: Arguably that's when we most care :)
Seriously though, I know performance impact is negligible. I meant it makes the code more complex just so we can log what seems like a not particularly important number. But I don't feel strongly about it - if you disagree let's leave it as is.

Issue #1367, opened 2018-10-31
GPU node should only calculate gpu utilization


ringtail:
GPU instance is quite different from other instance. It's much more expansive and not sensitive to cpu or memory. If this instance has GPU,Should we have different utilization formula for it. 

```
I1031 20:38:32.303631   31733 scale_down.go:407] Node cn-shenzhen.i-wz9ex1vq9wpu4sxvyy8o - utilization 0.600000
I1031 20:38:32.303670   31733 scale_down.go:411] Node cn-shenzhen.i-wz9ex1vq9wpu4sxvyy8o is not suitable for removal - utilization too big (0.600000)
I1031 20:38:32.303694   31733 scale_down.go:407] Node cn-shenzhen.i-wz91n8ywxxvcmqa2jxxh - utilization 0.600000
I1031 20:38:32.303711   31733 scale_down.go:411] Node cn-shenzhen.i-wz91n8ywxxvcmqa2jxxh is not suitable for removal - utilization too big (0.600000)
```
Here is a 2C8G1GPU instance and  nvidia-device-plugin will request at least 1 core so this instance can't scale down forever even no one use this GPU.

MaciekPytel:
+1 sounds like a good improvement. If you're willing to implement I'm happy to review, otherwise I can add it to my backlog, but it will probably take a while before I get to it.

Jeffwan:
@MaciekPytel  I can help look at this issue and submit a patch.

pierreozoux:
I'm interested in this features as well.
From what I read here, there is no request/limit on the Nvidia pod 
https://github.com/NVIDIA/k8s-device-plugin/blob/v1.11/nvidia-device-plugin.yml
If the add-on doesnt request 1 core, and as it is a deamon set, I think CA should work as expected? Am I right?
If not, I guess the CA for scaledown, should ignore add on and deamon sets.

What do you think about this idea? Did you workaround it for you?

Jeffwan:
@pierreozoux  Right. Daemonset is not a blocker for CA to remove nodes.  The problem here is GPU Node speciality or GPU utilization is not considered when CA make scale down decision.

seh:
@Jeffwan, that's a different complaint than what @ringtail mentioned originally. That problem was that the GPUs were idle, but there was (ignorable) CPU being used by the device plugin.

[...]

PR #1589, opened 2019-01-17
Consider GPU utilization in scaling down

Jeffwan:
Address #1367 

1. Consider GPU resource in utilization calculation. 
2. Add separate flag for GPU Utilization Threshold. 
3. Add UTs and also cover the case that annotate GPU pods to prevent eviction. (Training use case)

### DIFF
--- a/cluster-autoscaler/config/autoscaling_options.go
+++ b/cluster-autoscaler/config/autoscaling_options.go
@@ -34,9 +34,12 @@ type GpuLimits struct {
 type AutoscalingOptions struct {
 	// MaxEmptyBulkDelete is a number of empty nodes that can be removed at the same time.
 	MaxEmptyBulkDelete int
-	// ScaleDownUtilizationThreshold sets threshold for nodes to be considered for scale down.
+	// ScaleDownUtilizationThreshold sets threshold for nodes to be considered for scale down if cpu or memory utilization is over threshold.
 	// Well-utilized nodes are not touched.
 	ScaleDownUtilizationThreshold float64
+	// ScaleDownGpuUtilizationThreshold sets threshold for gpu nodes to be considered for scale down if gpu utilization is over threshold.
+	// Well-utilized nodes are not touched.
+	ScaleDownGpuUtilizationThreshold float64
 	// ScaleDownUnneededTime sets the duration CA expects a node to be unneeded/eligible for removal
 	// before scaling down the node.
 	ScaleDownUnneededTime time.Duration
--- a/cluster-autoscaler/core/scale_down.go
+++ b/cluster-autoscaler/core/scale_down.go
@@ -402,16 +402,16 @@ func (sd *ScaleDown) UpdateUnneededNodes(
 			klog.Errorf("Node info for %s not found", node.Name)
 			continue
 		}
-		utilInfo, err := simulator.CalculateUtilization(node, nodeInfo, sd.context.IgnoreDaemonSetsUtilization, sd.context.IgnoreMirrorPodsUtilization)
 
+		utilInfo, err := simulator.CalculateUtilization(node, nodeInfo, sd.context.IgnoreDaemonSetsUtilization, sd.context.IgnoreMirrorPodsUtilization, sd.context.CloudProvider.GPULabel())
 		if err != nil {
 			klog.Warningf("Failed to calculate utilization for %s: %v", node.Name, err)
 		}
-		klog.V(4).Infof("Node %s - utilization %f", node.Name, utilInfo.Utilization)
+		klog.V(4).Infof("Node %s - %s utilization %f", node.Name, utilInfo.ResourceName, utilInfo.Utilization)
 		utilizationMap[node.Name] = utilInfo
 
-		if utilInfo.Utilization >= sd.context.ScaleDownUtilizationThreshold {
-			klog.V(4).Infof("Node %s is not suitable for removal - utilization too big (%f)", node.Name, utilInfo.Utilization)
+		if !sd.isNodeBelowUtilzationThreshold(node, utilInfo) {
+			klog.V(4).Infof("Node %s is not suitable for removal - %s utilization too big (%f)", node.Name, utilInfo.ResourceName, utilInfo.Utilization)
 			continue
 		}
 		currentlyUnneededNodes = append(currentlyUnneededNodes, node)
@@ -504,6 +504,20 @@ func (sd *ScaleDown) UpdateUnneededNodes(
 	sd.clusterStateRegistry.UpdateScaleDownCandidates(sd.unneededNodesList, timestamp)
 	metrics.UpdateUnneededNodesCount(len(sd.unneededNodesList))
 	return nil
+}
+
+// isNodeBelowUtilzationThreshold determintes if a given node utilization is blow threshold.
+func (sd *ScaleDown) isNodeBelowUtilzationThreshold(node *apiv1.Node, utilInfo simulator.UtilizationInfo) bool {
+	if gpu.NodeHasGpu(sd.context.CloudProvider.GPULabel(), node) {
+		if utilInfo.Utilization >= sd.context.ScaleDownGpuUtilizationThreshold {
+			return false
+		}
+	} else {
+		if utilInfo.Utilization >= sd.context.ScaleDownUtilizationThreshold {
+			return false
+		}
+	}
+	return true
 }
 
 // updateUnremovableNodes updates unremovableNodes map according to current
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -93,7 +93,10 @@ var (
 	scaleDownUnreadyTime = flag.Duration("scale-down-unready-time", 20*time.Minute,
 		"How long an unready node should be unneeded before it is eligible for scale down")
 	scaleDownUtilizationThreshold = flag.Float64("scale-down-utilization-threshold", 0.5,
-		"Node utilization level, defined as sum of requested resources divided by capacity, below which a node can be considered for scale down")
+		"Sum of cpu or memory of all pods running on the node divided by node's corresponding allocatable resource, below which a node can be considered for scale down")
+	scaleDownGpuUtilizationThreshold = flag.Float64("scale-down-gpu-utilization-threshold", 0.5,
+		"Sum of gpu requests of all pods running on the node divided by node's allocatable resource, below which a node can be considered for scale down."+
+			"Utilization calculation only cares about gpu resource for accelerator node. cpu and memory utilization will be ignored.")
 	scaleDownNonEmptyCandidatesCount = flag.Int("scale-down-non-empty-candidates-count", 30,
 		"Maximum number of non empty nodes considered in one iteration as candidates for scale down with drain."+
 			"Lower value means better CA responsiveness but possible slower scale down latency."+
@@ -210,6 +213,7 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 		ScaleDownUnneededTime:               *scaleDownUnneededTime,
 		ScaleDownUnreadyTime:                *scaleDownUnreadyTime,
 		ScaleDownUtilizationThreshold:       *scaleDownUtilizationThreshold,
+		ScaleDownGpuUtilizationThreshold:    *scaleDownGpuUtilizationThreshold,
 		ScaleDownNonEmptyCandidatesCount:    *scaleDownNonEmptyCandidatesCount,
 		ScaleDownCandidatesPoolRatio:        *scaleDownCandidatesPoolRatio,
 		ScaleDownCandidatesPoolMinCount:     *scaleDownCandidatesPoolMinCount,
--- a/cluster-autoscaler/simulator/cluster.go
+++ b/cluster-autoscaler/simulator/cluster.go
@@ -19,13 +19,13 @@ package simulator
 import (
 	"flag"
 	"fmt"
-	"math"
 	"math/rand"
 	"time"
 
 	"k8s.io/autoscaler/cluster-autoscaler/utils/drain"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/glogx"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/gpu"
 	kube_util "k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
 	scheduler_util "k8s.io/autoscaler/cluster-autoscaler/utils/scheduler"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/tpu"
@@ -62,7 +62,10 @@ type NodeToBeRemoved struct {
 type UtilizationInfo struct {
 	CpuUtil float64
 	MemUtil float64
-	// Max(CpuUtil, MemUtil).
+	GpuUtil float64
+	// Resource name of highest utilization resource
+	ResourceName apiv1.ResourceName
+	// Max(CpuUtil, MemUtil) or GpuUtils
 	Utilization float64
 }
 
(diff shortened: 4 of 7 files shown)


cluster-autoscaler/.vscode/launch.json
```diff
@@ -0,0 +1,31 @@
+{
```
MaciekPytel: Can you remove this file?

cluster-autoscaler/simulator/cluster.go
```diff
@@ -188,6 +201,8 @@ func calculateUtilizationOfResource(node *apiv1.Node, nodeInfo *schedulercache.N
 			}
 		}
 	}
+	fmt.Println(float64(podsRequest.MilliValue()))
+	fmt.Println(float64(nodeAllocatable.MilliValue()))
```
MaciekPytel: Please remove prints.

cluster-autoscaler/.vscode/launch.json
Jeffwan: Yeah.. My bad. Carelessly add editor file.

Jeffwan:
@MaciekPytel  Clean up code and remove useless files. Please have a look

cluster-autoscaler/core/scale_down.go
```diff
@@ -403,14 +403,27 @@ func (sd *ScaleDown) UpdateUnneededNodes(
 
 		if err != nil {
 			klog.Warningf("Failed to calculate utilization for %s: %v", node.Name, err)
+
+			// Skip node that GPU resource is not ready to avoid scaling down new created node.
+			if gpu.NodeHasGpu(node) {
```
MaciekPytel: I understand what problem you're solving with this, but this feels like a very hacky way to do this. 
We consider node with GPU not ready as NotStarted in clusterstate. The correct thing to do would be to have proper handling for not started nodes (don't scale them down, delete them after timeout). 
Slightly less correct thing would be an explicit if checking for node with unready GPU and skipping it in there.

Jeffwan: Oh. I should leave a comment here. Originally, I think it doesn't need extra handling here. For node with GPU not ready case, utilization will be 0 and it will be added to candidate collection. Ideally, GPU is ready before `ScaleDownUnneededTime`. Next, CA will reconsider this node based on GPU usage. Do you think that's reasonable?  Delete line 408-410

Jeffwan: But this is still an explicit way..

cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -37,6 +37,9 @@ type AutoscalingOptions struct {
 	// ScaleDownUtilizationThreshold sets threshold for nodes to be considered for scale down.
 	// Well-utilized nodes are not touched.
 	ScaleDownUtilizationThreshold float64
+	// ScaleDownGPUUtilizationThreshold sets threshold for GPU nodes to be considered for scale down.
+	// Well-utilized nodes are not touched.
+	ScaleDownGPUUtilizationThreshold float64
```
losipiuk: Please make it `ScaleDownGpuUtilizationThreshold`.
Also expand comment to more explicitly say what the threshold means. (Yeah, I know comment for `ScaleDownUtilizationThresholds` is not meaningful neither; feel free to fix that one too :) ).
Specifically it is not obvious from variable name (nor comment) what resources does the utilization refer to (are those CPU, memory, or maybe GPU)

cluster-autoscaler/core/scale_down.go
```diff
@@ -403,14 +403,27 @@ func (sd *ScaleDown) UpdateUnneededNodes(
...
+				continue
+			}
 		}
 		klog.V(4).Infof("Node %s - utilization %f", node.Name, utilInfo.Utilization)
 		utilizationMap[node.Name] = utilInfo
 
-		if utilInfo.Utilization >= sd.context.ScaleDownUtilizationThreshold {
-			klog.V(4).Infof("Node %s is not suitable for removal - utilization too big (%f)", node.Name, utilInfo.Utilization)
-			continue
+		if gpu.NodeHasGpu(node) {
```
losipiuk: Please extract a function (`isNodeUnderutilized`/`isNodeBelowUtilzationThreshold`)

cluster-autoscaler/main.go
```diff
@@ -95,6 +95,8 @@ var (
 		"How long an unready node should be unneeded before it is eligible for scale down")
 	scaleDownUtilizationThreshold = flag.Float64("scale-down-utilization-threshold", 0.5,
 		"Node utilization level, defined as sum of requested resources divided by capacity, below which a node can be considered for scale down")
+	scaleDownGPUUtilizationThreshold = flag.Float64("scale-down-gpu-utilization-threshold", 0.50,
```
losipiuk: Nit: `s/0.50/0.5/` :)

losipiuk: Also here we could extend help strings so they are more meaningful.

cluster-autoscaler/config/autoscaling_options.go
Jeffwan: Great! Yeah. I will make it clear.

cluster-autoscaler/core/scale_down.go
Jeffwan: @MaciekPytel. I am trying to add logic to explicitly handle unready GPU Nodes. A new collection is needed to track those nodes. 
1. The `delete` you mean is delete kubernetes node and delete them in cloudprovider? Like `removeOldUnregisteredNodes` . The difference here is unready GPU nodes may have workloads running after timeout period unless user taints and tolerance or node affinity. If there's workload, better to scale them down.
2.  Does CA already have a timeout to use here? I searched configs and only find `unremovableNodeRecheckTimeout`. It probably need another option, like 5 or 10?

cluster-autoscaler/simulator/cluster.go
```diff
@@ -62,7 +64,8 @@ type NodeToBeRemoved struct {
 type UtilizationInfo struct {
 	CpuUtil float64
 	MemUtil float64
-	// Max(CpuUtil, MemUtil).
+	GPUUtil float64
```
christopherhein: to follow the `CpuUtil` convention this should probably be `GpuUtil`

cluster-autoscaler/core/scale_down.go
MaciekPytel: The current implementation will never remove a node with unready GPU. There should be some timeout for device plugin after which the node will be deleted (to avoid keeping defunct costly nodes around). Have you actually tested what will happen with those nodes? I suspect they will be deleted after `--scale-down-unready-time` (which is not perfect, but at least possible to change with a flag)?

Jeffwan: @MaciekPytel No, the case device plugin failed to report has not been covered in my test. This is a good edge case. In current implementation, if we just consider gpu utilization, calculator will return error here and node will be added into unneededList in scale down phase1. 

Since node status is still `ready`, it will be treated as ready node in phase 2. 

I think another option is
1. Add `FilterOutNodesWithUnreadyGpus` at the beginning of phase1, this will make sure CA won't consider unReady GPU nodes, this won't cover the case we want to delete nodes after 

2. Add hack logic here
https://github.com/kubernetes/autoscaler/blob/bd84757b7eedee6e122194ef7a1f41af27e45783/cluster-autoscaler/core/scale_down.go#L656

mark unready gpu node `unready` in phase 2, force the case go through 
https://github.com/kubernetes/autoscaler/blob/bd84757b7eedee6e122194ef7a1f41af27e45783/cluster-autoscaler/core/scale_down.go#L664-L666

cluster-autoscaler/simulator/cluster.go
Jeffwan: Resolved.

cluster-autoscaler/core/scale_down.go
Jeffwan: Hi @MaciekPytel  I double verified behaviors in tests. The problem I think is no matter node is ready or unready, CA calculates utilization for the node and add it into `unneededNodeList`.  

In this case, we assume node is unready, 
1. If device plugin is working fine, after a while (normally before ScaleDownUnneededTime), the node becomes ready. If utilization rate is low, it will be added into `unneededNodeList` 

2. If device plugin isn't working well, node will be unready and it's added into `unneededNodeList`.  CA will wait ScaleDownUnreadyTime to delete node. 


I think if we want to explicitly manage unready nodes, a reasonable solution is to skip unready node and only add those nodes to unneededNostList after `--scale-down-unready-time` in Phase I. Doesn't need to calculate utilization for them.  Need a map similar like `unneededNodes` to track those nodes.

MaciekPytel:
Sorry for delay. I'm travelling, I'll review tomorrow or Wednesday.

mwielgus:
@MaciekPytel  Ping, please review.

cluster-autoscaler/core/scale_down.go
```diff
@@ -227,6 +227,36 @@ func calculateScaleDownGpusTotal(nodes []*apiv1.Node, cp cloudprovider.CloudProv
 	return result, nil
 }
 
+// skipCalculatingUnreadyNode checks if a node should be skipped calucating utilization.
+func (sd *ScaleDown) skipCalculatingUnreadyNode(node *apiv1.Node, timestamp time.Time) bool {
```
losipiuk: I did not do a full review.
1. Just comment on this method. Name is very confusing and does not match description
2. The fact that predicate is modifying a map in ScaleDown is highly unexpected - I would expect this method to not have any side effects

Jeffwan: Sure. I will clean them up this weekend. I was stuck in some urgent work recently. Also waiting for feedbacks from @MaciekPytel

cluster-autoscaler/main.go
```diff
@@ -94,7 +94,10 @@ var (
 	scaleDownUnreadyTime = flag.Duration("scale-down-unready-time", 20*time.Minute,
 		"How long an unready node should be unneeded before it is eligible for scale down")
 	scaleDownUtilizationThreshold = flag.Float64("scale-down-utilization-threshold", 0.5,
-		"Node utilization level, defined as sum of requested resources divided by capacity, below which a node can be considered for scale down")
+		"Sum of cpu or memory of all pods running on the node divided by node's corresponding allocatable resource, below which a node can be considered for scale down")
+	scaleDownGpuUtilizationThreshold = flag.Float64("scale-down-gpu-utilization-threshold", 0.5,
```
MaciekPytel: I think the only values that make sense here are 0 (don't restart pods using GPU) or 1 (try to keep maximum GPU utilization). I don't think there is much use-case for 0.5, so I'd prefer a different default. Feels like most people don't want GPU workloads restarted, so I'd go for 0.0.

cluster-autoscaler/core/scale_down.go
MaciekPytel: If it's working well, why do we need to introduce the new map? Seems like all it's doing is save us the utilization check, which is basically free. The actual behavior would stay the same with your change, we've just added extra map and more code to handle it.

cluster-autoscaler/simulator/cluster.go
```diff
@@ -168,6 +181,7 @@ func calculateUtilizationOfResource(node *apiv1.Node, nodeInfo *schedulercache.N
 	nodeAllocatable, found := node.Status.Allocatable[resourceName]
 	if !found {
 		return 0, fmt.Errorf("failed to get %v from %s", resourceName, node.Name)
+		return 0, fmt.Errorf("Failed to get %v from %s, resource is not ready yet", resourceName, node.Name)
```
MaciekPytel: Seems like you added a new line, instead of modifying the existing one.

cluster-autoscaler/main.go
Jeffwan: Last time we talked about use case, we think this would be used for inference workloads, training jobs can not be interrupted and they will be labeled. `cluster-autoscaler.kubernetes.io/safe-to-evict: false`

Inference pods can be restarted and behavior should be close to general service pod. 0 here means no gpu pods on this node which is not good enough. If one of 8 gpus is taken by a pod, we also want to move it to other gpu nodes.

cluster-autoscaler/simulator/cluster.go
Jeffwan: Removed.

cluster-autoscaler/core/scale_down.go
Jeffwan: @MaciekPytel  I think last time, you mentioned it's better to `explicitly handle unready GPU Nodes`. Maybe I am doing the wrong way? 

The original way was a little bit tricky. Every nodes get into resource calculation but new coming GPU node will fail calculation and return error. Then we check if it's gpu case, we ignore it until node get allocatable gpu.  Do you think this is fine? I can change back to this way. It's working well but not that elegant.

The latest change brings new map and behavior won't change, just explicitly filter unready nodes before going to utilization calculation phase. But it has to control time because it could never come back and then we want to remove it.

cluster-autoscaler/core/scale_down.go
Jeffwan: @losipiuk Thanks. I rename it and make it more understandable. 
UnreadyNode map is to identify those nodes with unready resources explicitly. The original version doesn't have this map and we use a tricky way to skip unready nodes. This part can be removed. Both way work

Jeffwan:
@MaciekPytel  No problem. Yeah, that's true, from result perspective, there's no change. If we consider behavior, it's slightly different. What do you think? Do you think first way make sense to you? I can change back or if you have better ideas? 

1. Original revision: calculate utilization including unready nodes, then we skip unready GPU nodes from failures
2. Recent revision: explicitly skip unready nodes before utilization calculation.

cluster-autoscaler/main.go
MaciekPytel: I would also be fine with 1.0 (always try to pack GPU as tightly as possible). It feels that with GPU you always want to either avoid restarts or minimize number used and the sort of in-between approach we use for cpu/memory (0.5) makes less sense. But I'm ok to leave this default if you think this actually makes sense.

cluster-autoscaler/core/scale_down.go
MaciekPytel: I think I misunderstood what you were trying to do with that error handling - I thought you wanted to change what happens if driver installation falls and now I think it's just about handling missing Allocatable in calculateUtilizationOfResource. Is that correct?

cluster-autoscaler/main.go
mwielgus: I would assume that PDBs are not created with ML jobs on regular basis. At least I wouldn't do this. 1.0 is extremely aggressive towards pods that finish a bit later than other pods. They will be constantly interrupted, with big chance of starting from the very beginning. 
I would go for something small - like 0.25 or, maybe, in extreme 0. 
1.0 is IMHO a wrong choice.

cluster-autoscaler/core/scale_down.go
Jeffwan: Yes. Just as you said. I have not seen any situation driver installation falls yet so I didn't consider this case at that time.

mwielgus:
@MaciekPytel are you good with the PR or do you request some other changes?

cluster-autoscaler/core/scale_down.go
MaciekPytel: Right, so I indeed misunderstood. I'm thinking maybe the best way to handle this would be to just handle it in calculateUtilizationOfResource (make it return 0 if GPU utilization is missing)? The current solution feels overly complicated and I don't like the fact that previous one would ignore any error, not just this specific one.

Jeffwan: I see. Let me have a refactor and try to simply this change within small scope. Once I am done, I will come to you for another review.

Jeffwan:
@mwielgus Maciek left comments there and we determine to simplify process. I will submit revision later.

Jeffwan:
Done with change. We just return 0 if GPU utilization is missing. This will guarantee we can still delete node with unready resources. 

@MaciekPytel  @mwielgus  @losipiuk  Please have a review.

Jeffwan:
Manually run some tests and looks good to me.

GPU is not ready 
```
0404 01:04:17.785940   78940 static_autoscaler.go:365] Calculating unneeded nodes
I0404 01:04:17.785968   78940 utils.go:552] Skipping ip-192-168-79-122.us-west-2.compute.internal - node group min size reached
I0404 01:04:17.786047   78940 cluster.go:162] node ip-192-168-30-187.us-west-2.compute.internal has unready GPU
I0404 01:04:17.786060   78940 scale_down.go:410] Node ip-192-168-30-187.us-west-2.compute.internal - nvidia.com/gpu utilization 0.000000
I0404 01:04:17.786075   78940 cluster.go:162] node ip-192-168-13-48.us-west-2.compute.internal has unready GPU
I0404 01:04:17.786084   78940 scale_down.go:410] Node ip-192-168-13-48.us-west-2.compute.internal - nvidia.com/gpu utilization 0.000000
I0404 01:04:17.786161   78940 static_autoscaler.go:381] ip-192-168-30-187.us-west-2.compute.internal is unneeded since 2019-04-04 01:04:06.62706 -0700 PDT m=+124.298764166 duration 11.037588431s
I0404 01:04:17.786180   78940 static_autoscaler.go:381] ip-192-168-13-48.us-west-2.compute.internal is unneeded since 2019-04-04 01:04:06.62706 -0700 PDT m=+124.298764166 duration 11.037588431s

```

Normal case
```
I0404 01:05:08.849883   78940 scale_down.go:410] Node ip-192-168-30-187.us-west-2.compute.internal - nvidia.com/gpu utilization 0.250000
I0404 01:05:08.849912   78940 scale_down.go:410] Node ip-192-168-13-48.us-west-2.compute.internal - nvidia.com/gpu utilization 1.000000
I0404 01:05:08.849929   78940 scale_down.go:414] Node ip-192-168-13-48.us-west-2.compute.internal is not suitable for removal - nvidia.com/gpu utilization too big (1.000000)
I0404 01:05:08.850042   78940 cluster.go:93] Fast evaluation: ip-192-168-30-187.us-west-2.compute.internal for removal
I0404 01:05:08.850087   78940 cluster.go:248] Evaluation ip-192-168-13-48.us-west-2.compute.internal for default/v100-from-0-758f89485c-wtlxr -> PodFitsResources predicate mismatch, reason: Insufficient nvidia.com/gpu
I0404 01:05:08.850142   78940 cluster.go:248] Evaluation ip-192-168-79-122.us-west-2.compute.internal for default/v100-from-0-758f89485c-wtlxr -> PodFitsResources predicate mismatch, reason: Insufficient nvidia.com/gpu
I0404 01:05:08.850169   78940 cluster.go:129] Fast evaluation: node ip-192-168-30-187.us-west-2.compute.internal is not suitable for removal: failed to
```

scaling down case
```
I0404 01:06:20.804389   78940 scale_down.go:410] Node ip-192-168-13-48.us-west-2.compute.internal - nvidia.com/gpu utilization 0.250000
I0404 01:06:20.804577   78940 cluster.go:93] Fast evaluation: ip-192-168-13-48.us-west-2.compute.internal for removal
I0404 01:06:20.804695   78940 cluster.go:251] Pod default/v100-from-0-758f89485c-5jc9r can be moved to ip-192-168-30-187.us-west-2.compute.internal
I0404 01:06:20.804799   78940 cluster.go:124] Fast evaluation: node ip-192-168-13-48.us-west-2.compute.internal may be removed
I0404 01:06:20.804891   78940 static_autoscaler.go:381] ip-192-168-13-48.us-west-2.compute.internal is unneeded since 2019-04-04 01:06:09.207485 -0700 PDT m=+246.877340965 duration 11.594556145s

```

cluster-autoscaler/main.go
Jeffwan: 0.25 inclusive means 1/4 gpu will be considered over threshold which will not trigger scale down. Most of large GPU machines comes with 4 or 8 gpus, I think it's reasonable to try to scale it down if there're 1 or 3 gpus used. 0.5 will work for both cases.

MaciekPytel:
/lgtm
/approve
Thanks for handling that and once again sorry for review latency.

Jeffwan:
No problem. Thanks for your help on this task! @MaciekPytel

Issue #1591, opened 2019-01-17
Binary in docker image of cluster-autoscaler:v1.3.5 does not match source code


gjasny:
I tried to use cluster-autoscaler version 1.3.5 with the Docker image at `gcr.io/google-containers/cluster-autoscaler:v1.3.5` on AWS. I got strange error messages saying it tried to call `DescribeLaunchConfigurations` and the `LaunchConfigurations` parameter was an empty list (or list of an empty string). This is weird because the auto scaling groups it complained about use LaunchTemplates, not LaunchConfigurations.

I used gdb to break at the error and saw that the `type aws` did only contain the `LaunchConfiguration` field, but none of the `LaunchTemplate*` fields that are present in this structure in the source code at tag `cluster-autoscaler-1.3.5`. That means the binary in the v1.3.5 Docker image was not built from the code at tag 1.3.5. Is it possible that an old version of the cluster-autoscaler binary was used to build the v.1.3.5 Docker image by mistake?

I manually built the binary and container from tag `cluster-autoscaler-1.3.5` and this works.

aleksandra-malinowska:
That sounds worrying. We'll take a look, thanks for reporting this. 

cc @losipiuk

aleksandra-malinowska:
Looks like the `v1.3.5` tag is on the master branch, not on `cluster-autoscaler-release-1.3` :/ This is certainly not as intended. It should point to [commit updating version to 1.3.5](https://github.com/kubernetes/autoscaler/commit/0b7a14651e9f1e4be4c3464b4ca4fb30b0542efe)).

I'll double check, but if the tree at that point contains the code you'd expect from the image, my suspicion is that the release was build from correct source, but a mistake was made when publishing release notes and creating tag on github (this step is done manually).

MaciekPytel:
Sorry, my bad. I accidentally put tag on wrong branch.

devkid:
Colleague of OP here. @aleksandra-malinowska @MaciekPytel Do you mean that the Docker image was correct and the tag in git was wrong or vice versa? If the Docker image is correct, does that mean that 1.3.5 does not support Auto Scaling Groups with Launch Templates? According to the `README.md`, we need to use 1.3.x for Kubernetes 1.11, would that mean we cannot use ASGs with Launch Templates?

MaciekPytel:
[...]

PR #1594, opened 2019-01-18
Cherry pick to backport support for AWS EC2 Launch Templates onto 1.3 release branch to support Launch Templates with Kubernetes 1.11

devkid:
This feature is included in cluster-autoscaler starting from 1.12.


devkid:
As discussed in #1591.

@sethpollack, @Jeffwan

aleksandra-malinowska:
Cherry-picking godeps update effectively replaces all dependencies from Kubernetes 1.11 with those from Kubernetes 1.12. Since Cluster Autoscaler 1.3 is intended to work with Kubernetes 1.11, this won't do. 

If this change requires newer AWS SDK, first confirm that this is in kubernetes/kubernetes dependencies on `release-1.11` branch. If it is, [update godeps using this branch](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-update-ca-dependencies-particularly-k8siokubernetes).

If it isn't, the alternatives are:
1) cherry-pick newer version to kubernetes/kubernetes `release-1.11` first, or
2) put newer version of SDK in Cluster Autoscaler's `/override` directory and update `fix_gopath.sh` to pick it up,
then update godeps as described above.

devkid:
I did not cherry pick the Godeps dependendices, I just updated the aws sdk dependencies using `godep update github.com/aws/aws-sdk-go/...` because the old aws sdk did not know about LaunchTemplates.

Thanks for pointing that out. I will probably go with option 2 next week.

devkid:
I started looking into this and encountered the following problems:
* when I add the whole `aws-sdk-go` package under `cluster-autoscaler/_override`, I increase the size of the whole repo by roughly 90 MiB; this seems undesirable, considering that the whole repo is only roughly 110 MiB and the other folders in `_override` are significantly smaller
* when I only add the `aws-sdk-go/service/ec2` folder, it doesn't compile because there were probably changes in that package that require a complete new `aws-sdk-go` package

Am I missing something here? Or is there no other way than to get the newer aws-sdk-go version into `k8s.io/kubernetes` first?

aleksandra-malinowska:
> Am I missing something here? Or is there no other way than to get the newer aws-sdk-go version into k8s.io/kubernetes first?

Not really. Dependency management is definitely a sad part of this project, sorry about this :( TL;DR is that we require scheduler code, which means vendoring all of core Kubernetes, which keeps biting us in strange ways on ~every godeps update.

mwielgus:
PR not active. Dependencies https://github.com/kubernetes/kubernetes/pull/73349 are also stuck. 

Closing. Please reopen once it will be possible to merge it.

Jeffwan:
I will leave on eye once Micah from my team get SDK support in kubernetes upstream. Then we can easily cherry pick this change,

PR #1598, opened 2019-01-21
Tainting nodes - update first, refresh on conflict

jkaniuk:
Depend on https://github.com/kubernetes/autoscaler/pull/1587


jkaniuk:
/assign @losipiuk
/assign @MaciekPytel 

Please review only last commit but do not merge as it depend on other PR
/hold

PR #1600, opened 2019-01-22
Cherry pick: More flexible VPA certificate configuration #1534

bskiba:


bskiba:
@schylek

schylek:
@jbartosik

vertical-pod-autoscaler/pkg/admission-controller/main.go
```diff
@@ -34,8 +34,13 @@ import (
 )
 
 var (
-	certsDir = flag.String("certs-dir", "/etc/tls-certs", `Where the TLS cert files are stored.`)
```
jbartosik: Just to check - do we normally just remove flags? Or do we mark them as deprecated but keep them around for a while before removing?

bskiba: Not normally, but I figured this one is unlikely to be used by anyone, and definitely not used by default.

PR #1610, opened 2019-01-24
Use k8s.io/klog instead github.com/golang/glog

Rajat-0:


mwielgus:
This PR touches both cluster autoscaler (in a way that is completely not related to klog) and VPA. Please submit these PRs separately. Closing for now.

PR #1621, opened 2019-01-29
add prometheus as history provider to FAQ

wyb1:
Added how to set prometheus as the history provider to the FAQ. 
@bskiba


bskiba:
This is perfect, thank you!
/lgtm
/approve

Issue #1561, opened 2019-01-07
vertical-pod-autoscaler: make the admission controller self registration optional


raffaelespazzoli:
currently the `vpa-admission-controller` self register itself as an admission controller to the master API.
As a consequence the vpa-admission-controller service account must have enough permissions to do so.
Security aware organization may have a problem with this set-up as the `vpa-admission-controller` main duty is not to register new admission controller it is just to set correct resources for pod being created. 
I suggest to make it optional to have the `vpa-admission-controller` self-registering  and allow a path where the vpa-admission-controller is registered as a part of the installation process. 
This would allow for better separation of duties and least privilege, making the entire deployment more secure.

bskiba:
That sounds like a very reasonable request and shouldn't be hard to set up.

Rajat-0:
I would like to work on this

bskiba:
Great! Feel free to send PRs my way.

bskiba:
Fixed by #2949

PR #1625, opened 2019-01-30
Allow vpa-admission-controler to installl on specific path

Rajat-0:
#1561

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/config.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/config.go
@@ -57,7 +57,7 @@ func configTLS(clientset *kubernetes.Clientset, serverCert, serverKey []byte) *t
 
 // register this webhook admission controller with the kube-apiserver
 // by creating MutatingWebhookConfiguration.
-func selfRegistration(clientset *kubernetes.Clientset, caCert []byte, namespace *string) {
+func selfRegistration(clientset *kubernetes.Clientset, caCert []byte, namespace *string, url string, registerByURL bool) {
 	time.Sleep(10 * time.Second)
 	client := clientset.AdmissionregistrationV1beta1().MutatingWebhookConfigurations()
 	_, err := client.Get(webhookConfigName, metav1.GetOptions{})
@@ -66,6 +66,14 @@ func selfRegistration(clientset *kubernetes.Clientset, caCert []byte, namespace 
 			glog.Fatal(err2)
 		}
 	}
+	RegisterClientConfig := v1beta1.WebhookClientConfig{}
+	if !registerByURL {
+		RegisterClientConfig.Service.Name = *namespace
+		RegisterClientConfig.Service.Name = "vpa-webhook"
+	} else {
+		RegisterClientConfig.URL = &url
+	}
+	RegisterClientConfig.CABundle = caCert
 	webhookConfig := &v1beta1.MutatingWebhookConfiguration{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: webhookConfigName,
@@ -90,13 +98,7 @@ func selfRegistration(clientset *kubernetes.Clientset, caCert []byte, namespace 
 							Resources:   []string{"verticalpodautoscalers"},
 						},
 					}},
-				ClientConfig: v1beta1.WebhookClientConfig{
-					Service: &v1beta1.ServiceReference{
-						Namespace: *namespace,
-						Name:      "vpa-webhook",
-					},
-					CABundle: caCert,
-				},
+				ClientConfig: RegisterClientConfig,
 			},
 		},
 	}
--- a/vertical-pod-autoscaler/pkg/admission-controller/main.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/main.go
@@ -48,9 +48,12 @@ var (
 		tlsPrivateKey: flag.String("tls-private-key", "/etc/tls-certs/serverKey.pem", "Path to server certificate key PEM file."),
 	}
 
-	port      = flag.Int("port", 8000, "The port to listen on.")
-	address   = flag.String("address", ":8944", "The address to expose Prometheus metrics.")
-	namespace = os.Getenv("NAMESPACE")
+	port           = flag.Int("port", 8000, "The port to listen on.")
+	address        = flag.String("address", ":8944", "The address to expose Prometheus metrics.")
+	namespace      = os.Getenv("NAMESPACE")
+	webhookAddress = flag.String("webhook-address", "", "Address under which webhook is registered. Used when registerByURL is set to true.")
+	webhookPort    = flag.String("webhook-port", "", "Server Port for Webhook")
+	registerByURL  = flag.Bool("register-by-url", false, "If set to true, admission webhook will be registered by URL (webhookAddress:webhookPort) instead of by service name")
 )
 
 func main() {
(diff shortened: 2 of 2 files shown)


bskiba:
Can you reference the original issue that this is addressing?

vertical-pod-autoscaler/pkg/admission-controller/main.go
```diff
@@ -43,6 +43,7 @@ var (
 
 	address   = flag.String("address", ":8944", "The address to expose Prometheus metrics.")
 	namespace = os.Getenv("NAMESPACE")
+        path      = flag.String("path","https://127.0.01","path for admission controller")
```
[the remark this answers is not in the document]
Rajat-0: I was using default value for testing only, I'll split in into two flags

Rajat-0:
@bskiba I have sperated it into two flags

bskiba:
From https://github.com/kubernetes/kubernetes/blob/92e0c231fa0fff744a1ce26955cf46c693dfa5e5/pkg/apis/admissionregistration/types.go#L239:6

> Exactly one of `url` or `service` must be specified.

it seems that you can't register both by Service and URL. Does this work for you?

Rajat-0:
Can we specify the url where it is currently installing the vpa-admission controller as the initial/default url?

bskiba:
Not sure I understand the question. The current way of registering the webhook is by specifying a Kubernetes Service to call. Registration by URL is an alternative, it shouldn't be used at the same time as Service is specified. That's why I am curious if this works for you in the current setup - seems it shouldn't work if you specify a not-empty URL.

We could introduce recognizing registration mode to avoid this clash - register as Service only if the url is not empty or have a --registerByURL flag

Rajat-0:
Having a --registerByURL flag sounds good.

Rajat-0:
@bskiba I have updated the PR  PTAL.

vertical-pod-autoscaler/pkg/admission-controller/main.go
```diff
@@ -75,6 +79,8 @@ func main() {
 		Addr:      ":8000",
 		TLSConfig: configTLS(clientset, certs.serverCert, certs.serverKey),
 	}
-	go selfRegistration(clientset, certs.caCert, &namespace)
+	Path := webhookConfigName + ":"
+	url := fmt.Sprintf(Path + *serverPort)
```
bskiba: These two lines can be shortened to url := fmt.Sprintf("%s:%v", webhookAddress, serverPort)

vertical-pod-autoscaler/pkg/admission-controller/main.go
```diff
@@ -41,8 +42,11 @@ var (
 		tlsPrivateKey: flag.String("tls-private-key", "/etc/tls-certs/serverKey.pem", "Path to server certificate key PEM file."),
 	}
 
-	address   = flag.String("address", ":8944", "The address to expose Prometheus metrics.")
-	namespace = os.Getenv("NAMESPACE")
+	address         = flag.String("address", ":8944", "The address to expose Prometheus metrics.")
+	namespace       = os.Getenv("NAMESPACE")
+	webhookHostName = flag.String("webhookHostName", "", "path for admission controller.Blank for default")
```
bskiba: I'd rather name the two flags webhookAddress, webhookPort.

bskiba: Can you change the description to "Address under which webhook is registered. Used when registerByURL is set to true."

vertical-pod-autoscaler/pkg/admission-controller/main.go
```diff
@@ -41,8 +42,11 @@ var (
...
+	serverPort      = flag.String("Server Port", "", "Server Port for Webhook")
+	registerByURL   = flag.Bool("registerByURL", false, "True, if admission webhook is to be register on specific URL")
```
bskiba: Can you change the description to "If set to true, admission webhook will be registered by URL (:) instead of by service name."

vertical-pod-autoscaler/pkg/admission-controller/config.go
```diff
@@ -82,7 +79,7 @@ func configTLS(clientset *kubernetes.Clientset, serverCert, serverKey []byte) *t
 
 // register this webhook admission controller with the kube-apiserver
 // by creating MutatingWebhookConfiguration.
-func selfRegistration(clientset *kubernetes.Clientset, caCert []byte, namespace *string) {
+func selfRegistration(clientset *kubernetes.Clientset, caCert []byte, namespace *string, url *string, registerByURL *bool) {
```
bskiba: url string, registerByURL bool (please remove pointers)

vertical-pod-autoscaler/pkg/admission-controller/config.go
```diff
@@ -91,6 +88,15 @@ func selfRegistration(clientset *kubernetes.Clientset, caCert []byte, namespace
 			glog.Fatal(err2)
 		}
 	}
+	RegisterClientConfig := v1beta1.WebhookClientConfig{}
+	if *registerByURL == false {
```
bskiba: if !registerByURL

vertical-pod-autoscaler/pkg/admission-controller/config.go
```diff
@@ -91,6 +88,15 @@ func selfRegistration(clientset *kubernetes.Clientset, caCert []byte, namespace
...
+		RegisterClientConfig.Service.Name = *namespace
+		RegisterClientConfig.Service.Name = "vpa-webhook"
+		RegisterClientConfig.CABundle = caCert
+	} else {
+		RegisterClientConfig.URL = url
+		RegisterClientConfig.CABundle = caCert
```
bskiba: nit: RegisterClientConfig.CABundle = caCert can be done once outside of the if.

vertical-pod-autoscaler/pkg/admission-controller/config.go
```diff
@@ -1,12 +1,9 @@
 /*
 Copyright 2018 The Kubernetes Authors.
-
```
bskiba: Please don't modify the Copyright comment.

Rajat-0: that was by mistake

Rajat-0:
@bskiba PTAL

vertical-pod-autoscaler/pkg/admission-controller/main.go
```diff
@@ -41,8 +42,11 @@ var (
 		tlsPrivateKey: flag.String("tls-private-key", "/etc/tls-certs/serverKey.pem", "Path to server certificate key PEM file."),
 	}
 
-	address   = flag.String("address", ":8944", "The address to expose Prometheus metrics.")
-	namespace = os.Getenv("NAMESPACE")
+	address        = flag.String("address", ":8944", "The address to expose Prometheus metrics.")
+	namespace      = os.Getenv("NAMESPACE")
+	webhookAddress = flag.String("webhookHostName", "", "Address under which webhook is registered. Used when registerByURL is set to true.")
```
bskiba: Please name the flag webhook-address

vertical-pod-autoscaler/pkg/admission-controller/main.go
```diff
@@ -41,8 +42,11 @@ var (
...
+	webhookPort    = flag.String("Server Port", "", "Server Port for Webhook")
```
bskiba: Please name the flag webhook-port

vertical-pod-autoscaler/pkg/admission-controller/main.go
```diff
@@ -41,8 +42,11 @@ var (
...
+	registerByURL  = flag.Bool("registerByURL", false, "If set to true, admission webhook will be registered by URL (webhookAddress:webhookPort) instead of by service name")
```
bskiba: please name the flag register-by-url

bskiba:
Probably last minor comments, please also resolve conflict.

Rajat-0:
@bskiba PTAL

PR #1639, opened 2019-01-31
Default VPA version 0.3.1

bskiba:


bskiba:
@jbartosik

bskiba:
@Rajat-0 The change is on branch vpa-release-0.3, I can make it in master also, but it shouldn't make any difference.

Rajat-0:
@bskiba thanks for clarification.

Issue #1565, opened 2019-01-09
Do not store NodeGroup object in ScaleUpRequest/ScaleDownRequest


losipiuk:
We recently changed the `ScaleUp/DownRequest` objects to hold reference to `NodeGroup`.
This is not a good idea as we do not guarantee that `NodeGroup` objects created by cloud provider are valid after alogorithm loop completes. We should get back to storing `NodeGroup` id instead and gracefully handle situation when node group for given id is no longer available.

losipiuk:
/remove-lifecycle rotten

*From: *fejta-bot <notifications@github.com>
*Date: *Thu, May 9, 2019, 20:41
*To: *kubernetes/autoscaler
*Cc: *Łukasz Osipiuk, Author

Stale issues rot after 30d of inactivity.

PR #1642, opened 2019-02-01
[WIP] Storing store NodeGroupName object instead of NodeGroup.id

Rajat-0:


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -67,6 +67,8 @@ type ScaleUpRequest struct {
 	ExpectedAddTime time.Time
 	// How much the node group is increased.
 	Increase int
+	// NodeGroupName is the node group to be scaled up.
+	NodeGroupName string
```
MaciekPytel: Is there any reason for storing this if we already have the NodeGroup? The assumption is NodeGroup.Id() should not make any costly operations like API calls (or cache them internally) - we call it all over CA code.

mwielgus:
What is the purpose of this change? So far ID worked for everyone...

losipiuk:
This PR is about fixing #1565. So we are not storing `NodeGroup` instances in `ScaleUp/DownRequest` objects over multiple loop iterations.

@Rajat-0 as there is not much happening with this PR I will close it for now. Please reopen where it is in shape ready to be reviewed.
Thanks for work so far.

PR #1643, opened 2019-02-01
Capacity prediction based on physical memory

jkaniuk:


cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -478,7 +478,7 @@ func (m *gceManagerImpl) getCpuAndMemoryForMachineType(machineType string, zone
 		}
 		m.cache.AddMachineToCache(machineType, zone, machine)
 	}
-	return machine.GuestCpus, machine.MemoryMb * bytesPerMB, nil
+	return machine.GuestCpus, machine.MemoryMb * 1024 * 1024, nil
```
aleksandra-malinowska: Why not use units.MiB in this file as well, since it's already being added?

cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -59,13 +65,14 @@ func (t *GceTemplateBuilder) getAcceleratorCount(accelerators []*gce.Accelerator
...
-// BuildCapacity builds a list of resource capacities for a node.
-func (t *GceTemplateBuilder) BuildCapacity(machineType string, accelerators []*gce.AcceleratorConfig, zone string, cpu int64, mem int64) (apiv1.ResourceList, error) {
+// BuildCapacity builds a list of resource capacities given list of hardware.
+func (t *GceTemplateBuilder) BuildCapacity(cpu int64, mem int64, accelerators []*gce.AcceleratorConfig) (apiv1.ResourceList, error) {
 	capacity := apiv1.ResourceList{}
 	// TODO: get a real value.
 	capacity[apiv1.ResourcePods] = *resource.NewQuantity(110, resource.DecimalSI)
 	capacity[apiv1.ResourceCPU] = *resource.NewQuantity(cpu, resource.DecimalSI)
-	capacity[apiv1.ResourceMemory] = *resource.NewQuantity(mem, resource.DecimalSI)
+	memTotal := mem - CalculateCapacityFromPhysicalMemory(mem)
```
aleksandra-malinowska: IIUC, this is calculating _reservation_ and assigning it to memTotal. Is it intended?

cluster-autoscaler/cloudprovider/gce/gce_manager.go
jkaniuk: done

cluster-autoscaler/cloudprovider/gce/templates.go
jkaniuk: it was not, I have fixed that and added additional unit test

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -492,6 +493,6 @@ func parseCustomMachineType(machineType string) (cpu, mem int64, err error) {
 		return 0, 0, fmt.Errorf("failed to parse all params in %s", machineType)
 	}
 	// Mb to bytes
-	mem = mem * bytesPerMB
+	mem = mem * 1024 * 1024
```
aleksandra-malinowska: units.MiB

jkaniuk: done here and in few other places

cluster-autoscaler/core/utils.go
```diff
@@ -293,6 +301,16 @@ func getNodeInfosForGroups(nodes []*apiv1.Node, cloudProvider cloudprovider.Clou
 			continue
 		}
 
+		// No good template, check cache of previously running nodes.
+		if exemplarNodes != nil {
+			if node, found := exemplarNodes[id]; found {
+				added, _ := processNode(node)
```
aleksandra-malinowska: I think it may work (i.e. that the logic is correct), but it seems to make a lot of assumptions about internal calls:
1) that processNode will not fail for a non-existing node just because it's gone - cache miss in cloud provider when retrieving NodeGroupForNode seems like the most obvious point of failure,
2) that BuildNodeInfo will work the same whether it exists or not - seems to be true for now, but it's not really part of any contract,
2) that seenGroups[id] will always be updated within processNode - not incorrect, just spreading logic.

Is there any reason we can't save sanitized node info and return it here when needed?

jkaniuk: Rewritten to cache nodeInfo

aleksandra-malinowska:
/lgtm

How did you test it?

losipiuk:
Skimmed. Seems fine.
/lgtm
/approve

cluster-autoscaler/core/utils_test.go
```diff
@@ -379,70 +379,190 @@ func TestFilterSchedulablePodsForNode(t *testing.T) {
...
+	cachedInfo, found = nodeInfoCache["ng2"]
+	assert.True(t, found)
+	assertEqualNodeCapacities(t, ready2, cachedInfo.Node())
+	cachedInfo, found = nodeInfoCache["ng3"]
+	assert.False(t, found)
+	cachedInfo, found = nodeInfoCache["ng4"]
+	assert.False(t, found)
+
+	// Invalidate part of cache in two different ways
+	provider1.DeleteNodeGroup("ng1")
```
aleksandra-malinowska: Doesn't the below call (nodeGroup.Delete()) also call this internally?

jkaniuk:
Regarding tests:
- spinning up cluster with empty autoscaling node-pool
- adding pod with capacity 1Mi bigger than computed according to new allocatable calculations minus 220Mi (DeamonSets)
- confirming that will not spin up a node
- removing pod
- adding pod with capacity computed according to new allocatable calculation minus 220Mi (DeamonSets)
- confirming that will spin up a node
- pod will not fit as real allocatable is slightly less (1-2MB)
- waiting for node to be taken down as unneeded
- confirming that no new node is added to that nodepool because of that pod (so cache is working)
- deleting node pool
- confirming that cluster-autoscaler handles node pool removal
- adding new node pool
- confirming that new node once again is spin up for given pod and new node pool

Separately, capacity and allocatable calculations were checked against many machine-types and GKE versions 1.11 and 1.12

cluster-autoscaler/core/utils_test.go
jkaniuk: It does, with small differences. Those two different way do delete a group test more those mocks than anything else. I think that it is still beneficial to leave it.

aleksandra-malinowska: Not sure I'm following - Delete() doesn't call DeleteNodeGroup() if and only if callback failed; we aren't checking any error in the below call, do we expect it to fail?

PR #1644, opened 2019-02-01
Cluster Autoscaler 1.3.6

aleksandra-malinowska:
Update Cluster Autoscaler version to 1.3.6.


miry:
@aleksandra-malinowska It seems there is missing Tag for github https://github.com/kubernetes/autoscaler/tree/cluster-autoscaler-1.3.6 or was not pushed.

aleksandra-malinowska:
Fixed, thanks!

PR #1651, opened 2019-02-05
Fix readme to reflect addon-resizer version 2.1.

carlosedp:
Updated readme.md to reflect latest parameters from addon-resizer 2.1. Also added references to the images pushed into the registry.

### DIFF
--- a/addon-resizer/README.md
+++ b/addon-resizer/README.md
@@ -5,7 +5,11 @@ vertically scales the dependent container up and down. Currently the only
 option is to scale it linearly based on the number of nodes, and it only works
 for a singleton.
 
-Currently recommended version is 1.8, on addon-resizer-release-1.8 branch.
+The latest version and Docker images are 2.1. The images are pushed into:
+
+* gcr.io/google-containers/addon-resizer-amd64:2.1
+* gcr.io/google-containers/addon-resizer-arm64:2.1
+* gcr.io/google-containers/addon-resizer-arm:2.1
 
 ## Nanny program and arguments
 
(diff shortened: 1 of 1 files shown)


bskiba:
Thanks a lot for this contribution, very much appreciated!

My main concern is that the recommended addon-resizer version is still 1.8
There are issues with the 2.1 version that make it not suitable for all use-cases. Can we keep the info about 1.8 in the main readme? 

I'll add specific comments directly on the diff.

addon-resizer/README.md
```diff
@@ -5,7 +5,11 @@ vertically scales the dependent container up and down. Currently the only
 option is to scale it linearly based on the number of nodes, and it only works
 for a singleton.
 
-Currently recommended version is 1.8, on addon-resizer-release-1.8 branch.
```
bskiba: Can we keep this line, ideally bolded, prepend with a "NOTE: " and then link to the 1.8 readme:
See [1.8 documentation for details](https://github.com/kubernetes/autoscaler/blob/addon-resizer-release-1.8/addon-resizer/README.md)

(The readme in that branch is a bit outdated but I'll try to fix that right away)

bskiba:
Actually, scratch all my comments, the 1.8 branch has such an old version of the readme that I need to reconsider. I'll need to sync with the folks who used to maintain it. I'll get back to you later this week.
/hold

bskiba:
Didn't have time for sync, so let's do it the other way round. Please apply my comment about keeping the line that says 1.8 is still recommended. Then we can merge and I will update the 1.8 readme later.

PR #1628, opened 2019-01-30
change file name to ca-bundle.crt

VinayVanama:

PR #1654, opened 2019-02-06
ssl-certs path mismatch

VinayVanama:


aleksandra-malinowska:
Thanks for contributing. Please sign CLA, we can't accept any PRs without it.

mwielgus:
This is the exactly same PR as #1628. The previous one was closed based on AWS maintainer comment (@Jeffwan).  I guess the situation has not changed dramatically over the last 3 days so I'm closing this one as well.

Jeffwan:
@mwielgus Yes. Right now, there're still lots of kops users. EKS users needs to update cert path. I think  instead of updating example now, I will update documentation to make it stand out. 

In the future, we probably want to update all example if most of users migrate to EKS.

PR #1656, opened 2019-02-08
Adding ability to override allocatable resources via ASG tags. 

### LINKED ISSUE #1650: Cluster Autoscaler unaware of ephemeral storage of ASGs without nodes
When an ASG has 0 nodes, the cluster autoscaler does not scale up the nodes for pods which have `ephemeral-storage` requests with the error message: `reason: Insufficient ephemeral-storage`

When a node of the same ASG already exist, the scaling works fine. 

This seems to be similar to the issue with taints and labels of nodes, which can be set via tags on the ASG. I could not find a similar functionality for indicating allocatable resources of a node.

In my eyes this is clearly missing functionality / bug. What is the best way to go about it? 
Maybe also do it via tags on the ASG or can we autodetect it via describing the vanilla ASG.
Is there something I overlooked? I can't be the first one to run into this issue!?

Image version: k8s.gcr.io/cluster-autoscaler:v1.13.1
Flags used:
            - ./cluster-autoscaler
            - --v=4
            - --stderrthreshold=info
            - --cloud-provider=aws
            - --expander=least-waste
            - --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,k8s.io/cluster-autoscaler/${CLUSTER}

Link to the same question in #sig-autoscaling
https://kubernetes.slack.com/archives/C09R1LV8S/p1549291479482900

nielsole:
Fixes #1650

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_manager.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_manager.go
@@ -247,6 +247,11 @@ func (m *AwsManager) buildNodeFromTemplate(asg *asg, template *asgTemplate) (*ap
 	node.Status.Capacity[gpu.ResourceNvidiaGPU] = *resource.NewQuantity(template.InstanceType.GPU, resource.DecimalSI)
 	node.Status.Capacity[apiv1.ResourceMemory] = *resource.NewQuantity(template.InstanceType.MemoryMb*1024*1024, resource.DecimalSI)
 
+	resourcesFromTags := extractAllocatableResourcesFromAsg(template.Tags)
+	if val, ok := resourcesFromTags["ephemeral-storage"]; ok {
+		node.Status.Capacity[apiv1.ResourceEphemeralStorage] = *val
+	}
+
 	// TODO: use proper allocatable!!
 	node.Status.Allocatable = node.Status.Capacity
 
@@ -286,6 +291,28 @@ func extractLabelsFromAsg(tags []*autoscaling.TagDescription) map[string]string 
 			label := splits[1]
 			if label != "" {
 				result[label] = v
+			}
+		}
+	}
+
+	return result
+}
+
+func extractAllocatableResourcesFromAsg(tags []*autoscaling.TagDescription) map[string]*resource.Quantity {
+	result := make(map[string]*resource.Quantity)
+
+	for _, tag := range tags {
+		k := *tag.Key
+		v := *tag.Value
+		splits := strings.Split(k, "k8s.io/cluster-autoscaler/node-template/resources/")
+		if len(splits) > 1 {
+			label := splits[1]
+			if label != "" {
+				quantity, err := resource.ParseQuantity(v)
+				if err != nil {
+					continue
+				}
+				result[label] = &quantity
 			}
 		}
 	}
--- a/cluster-autoscaler/cloudprovider/aws/aws_manager_test.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_manager_test.go
@@ -33,6 +33,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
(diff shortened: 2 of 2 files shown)


Jeffwan:
Thanks for the contribution. I will have a look and come back to you by EOD

cluster-autoscaler/cloudprovider/aws/aws_manager.go
```diff
@@ -247,6 +247,17 @@ func (m *AwsManager) buildNodeFromTemplate(asg *asg, template *asgTemplate) (*ap
 	node.Status.Capacity[gpu.ResourceNvidiaGPU] = *resource.NewQuantity(template.InstanceType.GPU, resource.DecimalSI)
 	node.Status.Capacity[apiv1.ResourceMemory] = *resource.NewQuantity(template.InstanceType.MemoryMb*1024*1024, resource.DecimalSI)
 
+	resourcesFromTags := extractResourceAllocatableFromAsg(template.Tags)
+	if val, ok := resourcesFromTags["cpu"]; ok {
```
Jeffwan: Do you need to override cpu and memory resource? I thought you only want `ephemeral-storage`  in this case?

cluster-autoscaler/cloudprovider/aws/aws_manager.go
```diff
@@ -293,6 +304,28 @@ func extractLabelsFromAsg(tags []*autoscaling.TagDescription) map[string]string
 	return result
 }
 
+func extractResourceAllocatableFromAsg(tags []*autoscaling.TagDescription) map[string]*resource.Quantity {
```
Jeffwan: we already have one extractLabelsFromAsg method, is it possible to reuse method?

cluster-autoscaler/cloudprovider/aws/aws_manager.go
nielsole: I don't think we can predict the value of `kubelet`s `--kube-reserved=` so I thought this would be a nice addition for those, who need to set the nodes to the exact value that is gonna be allocatable.
Should I remove it?

cluster-autoscaler/cloudprovider/aws/aws_manager.go
nielsole: I didn't do it, because `extractTaintsFromAsg` and `extractLabelsFromAsg` were already different functions. So I sticked with what was already there and made `extractResourceAllocatableFromAsg` separate.

I could go for a function with the signature like 

```
extractLabelsAndResourcesFromAsg(tags []*autoscaling.TagDescription) (map[string]string,  map[string]*resource.Quantity)
```
Although the `And` in a function name is already a red flag to me.

Another option would be to extract the parsing of the ASG labels, but that would only save us like 3 lines in the main function body (because we would still need to loop), so I don't think it is really worth it.

Which would you prefer?

cluster-autoscaler/cloudprovider/aws/aws_manager.go
Jeffwan: It's true and kubelet can not predict the value, actually neither of the user, since there's kube-reserved and system-reserved value.  Do you meet any problems using node capacity now?

cluster-autoscaler/cloudprovider/aws/aws_manager.go
Jeffwan: Make sense. Let's keep current signature. I will find opportunity to refactor all into one method later.

cluster-autoscaler/cloudprovider/aws/aws_manager.go
nielsole: no, we are quite new to the autoscaling game. We don't even set `--kube-reserved` currently.
It felt like something that could become an issue down the road, when we increase our cluster utilization. Shall I remove CPU and memory it then and just open another merge request should we ever face an issue with that?

Jeffwan: Sounds good. As I know I don't see anyone experience this issue. this is just used to get a similar node from cloud provider. If you do see some issues, please submit another one. Let's concentrate on `ephemeral-storage` issue in this PR.

Jeffwan:
@nielsole Please also squash into one commit.

nielsole:
ok, done. Thanks for being so responsive! :)

PR #1657, opened 2019-02-08
Vpa v1beta2 api

bskiba:
v1beta2 api changes the way VPA references groups of pods to scale. We are switching from label selectors to a mechanism used by other autoscaling component - Horizontal Pod Autoscaler. 

VPA objects will reference a resource via CrossVersionObjectReference. This object will need to implement the scale subresource, but only to the extent that it is possible to fetch the resource's label selector to identify pods grouped by this resource.

The most common use case will be referencing a controller such as Deployment or StatefulSet.

This change also adds new conditions for VPA to signal deprecated or unsupported configurations.


bskiba:
It's marked WIP

bskiba:
@schylek @kgolab

bskiba:
Ready for review @mwielgus

schylek:
LGTM
@kgolab 
1. "1. it should be enough if VPA knows how to list Pods of a referenced object" - I believe that for admission controller this would be not enough. Admission controller has to be able to tell if the pod belongs to VPA during creation.
2. It seems that new machinery to generate code was used and this resulted with extra dependencies.

vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2/types.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// ControllerRef points to the controller managing the set of pods for the
+	// autoscaler to control. The pod set is retrieved from the controller's
+	// ScaleStatus or in case of some well known controllers (Daemonset, Job
+	// or CronJob) from the controller's spec.
+	// Note that VerticalPodAutoscaler does not require full implementation
+	// of scale subresource - it will not use it to modify the replica count.
+	// The only thing retrieved is a label selector matching pods grouped by
+	// the target resource.
+	ControllerRef *autoscaling.CrossVersionObjectReference `json:"controllerRef" protobuf:"bytes,1,name=controllerRef"`
```
mwielgus: This is not obvious what should be put there by looking at the field name foe Deployments. For pods under  deployment the controller is ReplicaSet, not Deployment. Please update the description or pod name to handle this ambiguity.

mwielgus:
Closing in favour of #1668

PR #483, opened 2017-11-22
Added aws c5 instance types.

miry:
C5 instances: https://www.ec2instances.info/?filter=c5

Added new instances types to aws provider, otherwise we have exception everytime:

```
I1122 07:03:30.997937       1 scale_up.go:54] Pod bi/presto.worker-59788c9f65-565tx is unschedulable
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x145189d]
goroutine 76 [running]:
k8s.io/autoscaler/cluster-autoscaler/cloudprovider/aws.(*AwsManager).buildNodeFromTemplate(0xc420b1c040, 0xc421bcc000, 0xc421e0c900, 0xc421e0c900, 0x0, 0x0)
	/gopath/src/k8s.io/autoscaler/cluster-autoscaler/cloudprovider/aws/aws_manager.go:252 +0x48d
k8s.io/autoscaler/cluster-autoscaler/cloudprovider/aws.(*Asg).TemplateNodeInfo(0xc421bcc000, 0xc421dd3b00, 0xc421a2f380, 0x20)
	/gopath/src/k8s.io/autoscaler/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go:355 +0x78
k8s.io/autoscaler/cluster-autoscaler/core.GetNodeInfosForGroups(0xc421201080, 0x27, 0x2a, 0x50ba120, 0xc421bc1f80, 0x50c44e0, 0xc420ab5c70, 0xc42116ba80, 0x2, 0x2, ...)
	/gopath/src/k8s.io/autoscaler/cluster-autoscaler/core/utils.go:211 +0x65a
k8s.io/autoscaler/cluster-autoscaler/core.ScaleUp(0xc420b9b520, 0xc42116e360, 0x4, 0x4, 0xc421201080, 0x27, 0x2a, 0xc42116ba80, 0x2, 0x2, ...)
	/gopath/src/k8s.io/autoscaler/cluster-autoscaler/core/scale_up.go:57 +0x28e
k8s.io/autoscaler/cluster-autoscaler/core.(*StaticAutoscaler).RunOnce(0xc4218d4230, 0xed1a7183c, 0xe37618c72, 0x518d0a0, 0x0, 0x0)
	/gopath/src/k8s.io/autoscaler/cluster-autoscaler/core/static_autoscaler.go:246 +0x1378
[...]

PR #1662, opened 2019-02-10
Don't panic on unknown aws instance type

miry:
Cluster Autoscaler crashes if there are some AWS ASG that use instances
types and was not updated in the build.
    
Update tests for case when there is unknown instance type and return
error if there is one.

Sample of excepttion:  #483

Covers changes for 1.3 branch only

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider_test.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package aws
 
 import (
+	"fmt"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


Rajat-0:
lgtm

aleksandra-malinowska:
/approve

Any reason this is not a cherry-pick of #1425? I don't have a preference for either solution, just for consistency (it also makes subsequent cherry-picks easier).

Jeffwan:
> /approve
> 
> Any reason this is not a cherry-pick of #1425? I don't have a preference for either solution, just for consistency (it also makes subsequent cherry-picks easier).

Yes. Just notice that, should be a cherry-pick pf existing patch.

PR #1663, opened 2019-02-10
Updated aws instance types actual for 2019.02.10

miry:
Sync instance type for aws via `make generate`

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/ec2_instance_types.go
+++ b/cluster-autoscaler/cloudprovider/aws/ec2_instance_types.go
@@ -27,6 +27,42 @@ type instanceType struct {
 
 // InstanceTypes is a map of ec2 resources
 var InstanceTypes = map[string]*instanceType{
+	"a1": {
+		InstanceType: "a1",
+		VCPU:         16,
+		MemoryMb:     0,
+		GPU:          0,
+	},
+	"a1.2xlarge": {
+		InstanceType: "a1.2xlarge",
+		VCPU:         8,
+		MemoryMb:     16384,
+		GPU:          0,
+	},
+	"a1.4xlarge": {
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/cloudprovider/aws/ec2_instance_types.go
```diff
@@ -705,6 +783,18 @@ var InstanceTypes = map[string]*instanceType{
 		MemoryMb:     249856,
 		GPU:          4,
 	},
+	"p3dn": {
```
Jeffwan: em.. It looks weird. Not sure why here's a separate p3dn instance type here. 
The other thing is `p3dn.24xlarge` is not fully tested on kubernetes. (it supposed to be release together with latest CNI, and this instance type also need whitelist)

Let me double confirm and then come back to you.

Jeffwan: Looks lik p3dn belongs to dedicated host type. Not an issue.

mwielgus:
@Jeffwan  - does it mean LGTM?

cluster-autoscaler/cloudprovider/aws/ec2_instance_types.go
miry: @Jeffwan should I do something more? Should I check makefile to ignore p3dn?

Jeffwan: Just confirmed that p3dn.24xlarge test is done. Right now, AFAIK no one use dedicated host, I think we can still use API results directly. Nothing needs to be changed.

Jeffwan:
> @Jeffwan - does it mean LGTM?

Yes. Confirmed and LGTM

Issue #1540, opened 2018-12-28
Add the ability to better pass additional arguments to containers


raffaelespazzoli:
I suggest using `ENTRYPOINT` instead of `CMD` to set the default command of the containers that are used by VPA. This allows to use the `args` field in Kubernetes pods to customize the arguments to be passed to the containers.

PR #1664, opened 2019-02-10
Add the ability to better pass additional arguments to containers

Rajat-0:
#1540

### DIFF
--- a/vertical-pod-autoscaler/pkg/updater/Dockerfile
+++ b/vertical-pod-autoscaler/pkg/updater/Dockerfile
@@ -18,4 +18,4 @@ MAINTAINER Marcin Wielgus "mwielgus@google.com"
 
 ADD updater updater
 
-CMD ./updater --v=4 --stderrthreshold=info
+ENTRYPOINT ./updater --v=4 --stderrthreshold=info
--- a/vertical-pod-autoscaler/pkg/admission-controller/Dockerfile
+++ b/vertical-pod-autoscaler/pkg/admission-controller/Dockerfile
@@ -17,4 +17,4 @@ MAINTAINER Tomasz Kulczynski "tkulczynski@google.com"
 
 ADD admission-controller admission-controller
 
-CMD ./admission-controller --v=4 --stderrthreshold=info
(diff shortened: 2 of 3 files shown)


Rajat-0:
@mwielgus I was trying to implement https://github.com/kubernetes/autoscaler/issues/1540

mwielgus:
OK, I noticed the bug in the description, but I'm not sure what is your plan in modifying the build scripts.

Rajat-0:
@mwielgus So if I have to customize the arguments to be passed to the containers, then I have to change in Its pod's DockerFile( Recommender, Admission Controller, Updater)

mwielgus:
You are changing the wrong Docker file. For example you should change this one:

https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/admission-controller/Dockerfile

Instead of the one that is used to compile stuff.

vertical-pod-autoscaler/pkg/updater/Dockerfile
```diff
@@ -18,4 +18,4 @@ MAINTAINER Marcin Wielgus "mwielgus@google.com"
 
 ADD updater updater
 
-CMD ./updater --v=4 --stderrthreshold=info
+ENTRYPOINT ./updateri --v=4 --stderrthreshold=info
```
mwielgus: typo

mfilotto:
Maybe we could move default parameters to CMD as a best practice
https://medium.freecodecamp.org/docker-entrypoint-cmd-dockerfile-best-practices-abc591c30e21

And there is an extra i at the end of updater

vertical-pod-autoscaler/pkg/updater/Dockerfile
Rajat-0: Resolved

mfilotto:
Any plan to build a new vpa images version to unlock this feature ?
Looking forward to be able to set arguments (set replicas to 1 in my use case)

Leluke:
Hi how can I take advantage of this option? I wanted my VPA to restart pods even when there is only one replica of each.

PR #1668, opened 2019-02-11
Vpa v1beta2 api 

schylek:
This PR is a replacement of #1657 to address minor comments.

Original note:

v1beta2 api changes the way VPA references groups of pods to scale. We are switching from label selectors to a mechanism used by other autoscaling component - Horizontal Pod Autoscaler.

VPA objects will reference a resource via CrossVersionObjectReference. This object will need to implement the scale subresource, but only to the extent that it is possible to fetch the resource's label selector to identify pods grouped by this resource.

The most common use case will be referencing a controller such as Deployment or StatefulSet.

This change also adds new conditions for VPA to signal deprecated or unsupported configurations.


schylek:
This PR is a replacement of #1657
@mwielgus @kgolab

PR #1671, opened 2019-02-12
Cherry-pick of #1288: Read AWS Region from EC2 Metadata

Jeffwan:
cherry-pick of #1288

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_manager.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_manager.go
@@ -22,14 +22,17 @@ import (
 	"fmt"
 	"io"
 	"math/rand"
+	"os"
 	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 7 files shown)


Jeffwan:
go v1.11 fails for files that pass on v1.10. Use container to do verification instead..

Fix cloudprovider/builder/cloud_provider_builder.go:150: Info call has possible formatting directive %v
Not sure how does it pass previous tests..

Jeffwan:
Recently know a few users use example yaml files in master branch with v1.3.6 CA. This change was a breaking change but not in v1.3.x yet. Even though user should use examples in right branch, I think it's better to cherry pick to v1.3 branch to reduce change they encounter this problem.

aleksandra-malinowska:
Can you please cherry-pick #1490 as well, so that tests don't start taking too long on this branch?

/lgtm
/approve

PR #1680, opened 2019-02-13
Use k8s.io/klog instead of glog

Rajat-0:


mwielgus:
Doesn't compile.

Rajat-0:
@mwielgus this is due to some dependency issue.

Rajat-0:
klog package is not in Vendor directory that's why its failing . How is vendor directory updated?
@bskiba

PR #1550, opened 2019-01-02
Pass nodeGroup->NodeInfo map to ClusterStateRegistry

losipiuk:

PR #1682, opened 2019-02-13
Cherry-pick of #1643: Capacity prediction based on physical memory

jkaniuk:
Cherry-pick of https://github.com/kubernetes/autoscaler/pull/1643
with dependency on https://github.com/kubernetes/autoscaler/pull/1550

/cc aleksandra-malinowska
/cc MaciekPytel
/cc losipiuk

/hold

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/gce_manager.go
+++ b/cluster-autoscaler/cloudprovider/gce/gce_manager.go
@@ -26,6 +26,7 @@ import (
 
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
 	"k8s.io/autoscaler/cluster-autoscaler/config/dynamic"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/units"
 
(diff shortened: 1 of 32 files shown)


jkaniuk:
Tested by:

* spinning up cluster with empty autoscaling node-pool
* adding pod with capacity 1Mi bigger than computed according to new allocatable calculations minus 220Mi (DeamonSets)
* confirming that will not spin up a node
* removing pod
* adding pod with capacity computed according to new allocatable calculation minus 220Mi (DeamonSets)
* confirming that will spin up a node
* pod will not fit as real allocatable is slightly less (1-2MB)
* waiting for node to be taken down as unneeded
* confirming that no new node is added to that nodepool because of that pod (so cache is working)
* deleting node pool
* confirming that cluster-autoscaler handles node pool removal
* adding new node pool with the same name
* confirming that new node once again is spin up for given pod and new node pool (so cache is cleared)

Separately, capacity and allocatable calculations were checked against many machine-types and GKE versions 1.11 and 1.12

PR #1690, opened 2019-02-14
Add OpenStack cloud provider

tghartland:
This pull request adds a cloud provider for OpenStack. This requires a more up to date version of gophercloud. The specific newer commit of gophercloud is chosen because the commit following it [changes a function signature](https://github.com/gophercloud/gophercloud/commit/7bd760e8eb1d7c94108db464cc992df1620278d1#diff-18ac4c04e89cf251c4c250ccfa91b7f8R84), and this leads to [problems with k8s.io/kubernetes itself](https://gist.github.com/tghartland/e564a2e57b09e7008a82b4ad18b5c4da).

As of right now this autoscaler only supports OpenStack clusters using Magnum and Heat, but this is done through an interface OpenstackManager. I will be adding another implementation of this interface to support Magnum clusters with proper nodegroups when that is available, but this could also be extended to support any other kind of OpenStack cluster. There is only the Magnum/Heat manager now but it is already set up to support more via changing an environment variable.

All nodegroups managed by the OpenstackCloudProvider interact with openstack through a single instance of the manager.

Since Magnum does not yet have nodegroups, only a single nodegroup is currently allowed and this represents the entire cluster.

PR #1691, opened 2019-02-14
Support openstack cluster autoscaler

ricolin:
#734

This is for update status and plan to co-work with https://github.com/kubernetes/autoscaler/pull/1690/commits/a37e287e1b5727987896e5df63f256c1de77195b


jim-bach:
slight request, is there any way to reduce the vendor folder like the other CAS PR to better view/comment on this PR?

mwielgus:
Looks like #1690 is getting much more traction and interest from the reviewers and Openstack folks. Closing this one for now. Should #1690 fail for some reason, we can get back to this one.

PR #1692, opened 2019-02-14
VPA: Remove TLS client cert lookup logic

forkbomber:
Actual verification was turned off from the initial commit and is still [not enabled](../blob/6b1a97c16001a98607424568faf7239b5c364d11/vertical-pod-autoscaler/pkg/admission-controller/config.go#L79), so the whole construct does not add value.

Also depending on `requestheader-client-ca-file` by itself renders VPA incompatible with apiserver installations which does not use such certificate with VPA Admission Controller crash-looping with:
`vpa-admission-controller-X-Y:admission-controller F0214 18:15:09.448204       9 config.go:60] cannot find the ca.crt in the configmap, configMap.Data is map[string]string{"client-ca-file":"...skipped..."}`

In addition to the above it worth noting that `requestheader-client-ca-file` appears to be just a wrong place to look for _possible_ apiserver client certificate (I don't think there is any).

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/config.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/config.go
@@ -18,8 +18,6 @@ package main
 
 import (
 	"crypto/tls"
-	"crypto/x509"
-	"fmt"
 	"time"
 
 	"k8s.io/api/admissionregistration/v1beta1"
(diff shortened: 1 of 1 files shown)


mwielgus:
cc: @schylek

mwielgus:
Doesn't compile. Please fix.

forkbomber:
@mwielgus, thanks, cleaned up unused imports.

mwielgus:
@tkulczynski - I believe you are the one who implemented this. Could you please take a look at the PR? Is this code needed?

schylek:
cc: @jbartosik

jbartosik:
I'll take a look later today.

jbartosik:
/lgtm
looks like dead code to me

bskiba:
Thanks @joachim! 
/approve

PR #1693, opened 2019-02-15
VPA - Implement obtaining selector from well known controllers

schylek:
(partial contribution from kgolab)


schylek:
cc: @kgolab

PR #1696, opened 2019-02-15
VPA - Switch to v1beta2 api

schylek:


mwielgus:
How was it tested?

schylek:
I did run all e2e tests (which still target v1beta1) locally + some manual tests of v1beta2. e2e tests against v1beta2 will follow.

PR #1711, opened 2019-02-22
Fix vmss scaling issues when node's providerIDs are in different cases

feiskyer:
Fix VMSS scaling issues when node's providerIDs are in different cases.

For VMSS node's providerIDs, they may be in different cases depending user's PUT requests to the resources. This would cause autoscaler to delete all newly provisioned nodes. And more seriously, CA would scale up/down again and again.

This PR adds a workaround for this issue, so that different cased providerIDs are processed as same node.

Cross refer https://github.com/kubernetes/kubernetes/issues/71994.

/assign @aleksandra-malinowska @losipiuk 

cc @andyzhangx @ritazh

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -277,11 +277,11 @@ func (csr *ClusterStateRegistry) UpdateNodes(nodes []*apiv1.Node, nodeInfosForGr
 		return err
 	}
 
-	cloudProviderNodeInstances, err := getCloudProviderNodeInstances(csr.cloudProvider)
+	cloudProviderNodeInstances, extended, err := getCloudProviderNodeInstances(csr.cloudProvider)
 	if err != nil {
 		return err
 	}
-	notRegistered := getNotRegisteredNodes(nodes, cloudProviderNodeInstances, currentTime)
+	notRegistered := getNotRegisteredNodes(nodes, cloudProviderNodeInstances, currentTime, extended)
 
 	csr.Lock()
 	defer csr.Unlock()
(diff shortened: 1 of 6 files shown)


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -935,12 +935,12 @@ func getCloudProviderNodeInstances(cloudProvider cloudprovider.CloudProvider) (m
 func getNotRegisteredNodes(allNodes []*apiv1.Node, cloudProviderNodeInstances map[string][]cloudprovider.Instance, time time.Time) []UnregisteredNode {
 	registered := sets.NewString()
 	for _, node := range allNodes {
-		registered.Insert(node.Spec.ProviderID)
+		registered.Insert(strings.ToLower(node.Spec.ProviderID))
```
mwielgus: This doesn't seem right. First of all, it is (probably) cloud provider specific. Second, if we agree that providerIds are not case sensitive, then we should update them at the very beginning (nodeLister?), so that we have no chance of forgetting to add `toLower` in some other place in the code.

feiskyer: Right, it may be cloud provider-specific, but I think it should work for all providers.

I added it here not in nodeLister because the ID in both cloudProviderNodeInstances and allNodes are only needed to be lower here. In other places (e.g. logs), we may still want it's logged with the original value.

mwielgus: The thing is that if, in 6 months, someone else will be doing something new with ProviderIDs, it is more than guaranteed that they will forget about using toLower when comparing them. And you will have another crash/bug on Azure.

feiskyer: Oh, good catch, thanks. Then let me move the logic to nodeLister.

feiskyer: @mwielgus After checking more codes, convert cases in nodeLister doesn't work because of two reasons:

* The nodes listed from nodeLister is also used in other places, e.g. update node taints
* The providerID of Node object is readonly and hence Node update API would fail if the providerID is different from its original value.

Seems it's still the best place of casting cases here while keeping the providerIDs are same with their original cases outside.

@mwielgus @losipiuk Any more ideas on this?

losipiuk: IMO the more proper way would  be extending `CloudProvider` (or maybe `NodeGroup`) interface with `GetIntsanceId(Node)` method which would return instance id compatible with one returned be `NodeGroup.Nodes()`.
Then as a followup the `DeleteNodes` method could be changed to take just list of Ids as parameter.

cc: @MaciekPytel

feiskyer: @mwielgus @MaciekPytel @losipiuk Added anther interface `NodeGroupExtended`. So now the conversion is only needed when this interface is implemented. PTAL

By the way, I'm trying to make the changes small enough, so that we could cherry pick it to stable branches.

losipiuk: This is very, very hacky. I do not like idea of having two interfaces (aside the fact the the name of both interface and the method is very not descriptive), as well as idea of passing `extended bool` flag around.
Why don't you follow what I suggested above?

feiskyer:
Would open another PR for this.

cluster-autoscaler/clusterstate/clusterstate.go
feiskyer: Thanks.

PR #1712, opened 2019-02-22
VPA switches to 0.4.0

bskiba:
Change default version
Update docs


bskiba:
@kgolab @schylek

Rajat-0:
lgtm

vertical-pod-autoscaler/README.md
```diff
@@ -12,49 +12,113 @@ It can both down-scale pods that are over-requesting resources, and also up-scal
 
 Autoscaling is configured with a
 [Custom Resource Definition object](https://kubernetes.io/docs/concepts/api-extension/custom-resources/)
-called [VerticalPodAutoscaler](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta1/types.go).
+called [VerticalPodAutoscaler](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2/types.go).
 It allows to specify which pods should be under vertically autoscaled as well as if/how the
```
kgolab: I guess "under" is not required here.

vertical-pod-autoscaler/examples/hamster.yaml
```diff
@@ -3,14 +3,15 @@
 # 0.5s and sleeping 0.5s).
 # It also creates a corresponding Vertical Pod Autoscaler that adjusts the requests.
 # Note that the update mode is left unset, so it defaults to "Auto" mode.
-apiVersion: "autoscaling.k8s.io/v1beta1"
+apiVersion: "autoscaling.k8s.io/v1beta2"
 kind: VerticalPodAutoscaler
 metadata:
-  name: hamster-vpa
+  name: hamster-vpa-beta2
```
kgolab: I think it would be slightly better if we kept the naming here as was and add -deprecated suffix to all objects in the other yaml.
This way a) it would be harder to miss deprecation b) there would be no conflicts between the two examples.
WDYT?

vertical-pod-autoscaler/README.md
```diff
@@ -12,49 +12,113 @@ It can both down-scale pods that are over-requesting resources, and also up-scal
...
 resource recommendations are applied.
 
 To enable vertical pod autoscaling on your cluster please follow the installation
 procedure described below.
 
-
 # Installation
 
-### Notice on switching from alpha to beta (<0.3.0 to >=0.3.0)
+The current default version is Verical Pod Autoscaler 0.4.0
```
Rajat-0: nit:Vertical

vertical-pod-autoscaler/README.md
```diff
@@ -12,49 +12,113 @@ It can both down-scale pods that are over-requesting resources, and also up-scal
...
+In 0.4.0 we introduced a new version of the API - `autoscaling.k8s.io/v1beta2`.
+Full API is accessible [here](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2/types.go).
 
-If you want to migrate your objects between versions, you can use the [object conversion script](./hack/convert-alpha-objects.sh).
-It will save your VPA objects into temporary yaml files with the new apiVersion.
-After running the script:
-1. disable alpha VPA via vpa-down.sh script
-1. enable beta VPA via vpa-up.sh script
-1. re-create VPA objects following the instructions from the script.
+The change introduced is in the way you express what pods should be scaled by a
```
kgolab: nit: s/what pods/which pods/

vertical-pod-autoscaler/README.md
```diff
@@ -12,49 +12,113 @@ It can both down-scale pods that are over-requesting resources, and also up-scal
...
+    matchLabels:
+      app: hamster
+```
+
+**[RECOMMENDED]** In `v1beta2` pods to scale by VPA are specified by a 
+target reference. This target will usually be a Deployment, as configured in the
+example below.
+
+```yaml
+apiVersion: "autoscaling.k8s.io/v1beta1"
```
kgolab: v1beta2 here

vertical-pod-autoscaler/README.md
```diff
@@ -12,49 +12,113 @@ It can both down-scale pods that are over-requesting resources, and also up-scal
...
+  targetRef:
+    apiVersion: "extensions/v1beta1"
+    kind:       Deployment
+    name:       hamster
 ```
-./hack/convert-alpha-objects.sh
+
+The target object can be a well known controller (Deployment, ReplicaSet, DaemonSet, StatefulSet etc.)
+or any object that implements the scale subresource. VPA uses ScaleStatus to
+retrieve the pod set controlled by this objetc.
```
kgolab: typo: objetc -> object

vertical-pod-autoscaler/README.md
```diff
@@ -12,49 +12,113 @@ It can both down-scale pods that are over-requesting resources, and also up-scal
...
 procedure described below.
 
-
 # Installation
 
-### Notice on switching from alpha to beta (<0.3.0 to >=0.3.0)
+The current default version is Verical Pod Autoscaler 0.4.0
+
+**NOTE:** version 0.4 requires at least Kubernetes 1.11 to work (needs certain
+Custom Resource Definition capabilities). With older Kubernetes versions we
```
schylek: In version 0.5 when we drop v1beta1 we won't need needs certain Custom Resource Definition capabilities and 1.11 won't be required.

Proposal:
**NOTE:** version 0.4 requires at least Kubernetes 1.11 to work (needs certain Custom Resource Definition capabilities to support multiple API versions). When deprecated v1beta1 API is removed requirement of Kubernetes in version at least 1.11 may be obsolete.

vertical-pod-autoscaler/examples/hamster.yaml
schylek: +1

vertical-pod-autoscaler/README.md
```diff
@@ -12,49 +12,113 @@ It can both down-scale pods that are over-requesting resources, and also up-scal
...
 
-Between versions 0.2.x and 0.3.x there is an alpha to beta switch which includes
-a change of VPA apiVersion from `poc.autoscaling.k8s.io/v1alpha1` to `autoscaling.k8s.io/v1beta1`.
-The safest way to switch is to use `vpa-down.sh` script to tear down
-the old installation of VPA first. This will delete your old VPA objects that
-have been defined with `poc.autoscaling.k8s.io/v1alpha1` apiVersion.
-Then use `vpa-up.sh` to bring up the new version of VPA and create you VPA
-objects from the scratch, passing apiVersion `autoscaling.k8s.io/v1beta1`
-(See [example](./examples/hamster.yaml)).
+In 0.4.0 we introduced a new version of the API - `autoscaling.k8s.io/v1beta2`.
```
schylek: I'd add a word why the change is introduced:
- avoid misuse of selectors (e.g. too wide or too narrow label selector)
- API aligned with HPA

vertical-pod-autoscaler/README.md
```diff
@@ -130,7 +193,7 @@ kubectl create -f examples/hamster.yaml
 
 The above command creates a deployment with 2 pods, each running a single container
 that requests 100 millicores and tries to utilize slightly above 500 millicores.
-The command also creates a VPA config with selector that matches the pods in the deployment.
+The command also creates a VPA config point at the deployment.
```
kgolab: nit: point -> pointing

vertical-pod-autoscaler/README.md
```diff
@@ -151,9 +214,10 @@ kind: VerticalPodAutoscaler
 metadata:
   name: my-app-vpa
 spec:
-  selector:
-    matchLabels:
-      app: my-app
+  targetRef:
```
kgolab: Please change the apiVersion above.

vertical-pod-autoscaler/README.md
bskiba: I'm not sure about that. There is better validation coming to CRDs in 1.12 and in general a lot of capabilities being added that we will benefit greatly from. The current statement doesn't comment on what will happen onwards, so I would rather leave it as is.

vertical-pod-autoscaler/README.md
bskiba: Added below.

vertical-pod-autoscaler/README.md
bskiba: Done

vertical-pod-autoscaler/README.md
bskiba: Done

vertical-pod-autoscaler/examples/hamster.yaml
bskiba: Good point. Done

vertical-pod-autoscaler/README.md
bskiba: Done

vertical-pod-autoscaler/README.md
bskiba: Done

vertical-pod-autoscaler/README.md
bskiba: Done

vertical-pod-autoscaler/README.md
bskiba: Done

vertical-pod-autoscaler/README.md
bskiba: Done

vertical-pod-autoscaler/README.md
schylek: OK. Fair.

PR #1713, opened 2019-02-22
Rename vpa-target-reader ClusterRoleBinding name

forkbomber:
This looks like unintended effect of renaming VPA ClusterRole `scale-reader` to `vpa-target-reader` in 0bc25d5.

Repetitive `vpa-vpa-` while being technically correct and in accordance with naming convention in a surrounding blocks just looks wrong.


bskiba:
I believe this was intended, it's general prefix vpa- followed by vpa-target. @schylek, can you confirm?

schylek:
I am fine with this change.

bskiba:
Thanks!

/lgtm
/approve

PR #1719, opened 2019-02-25
AliCloud - Add asg configuration's tag to node

lookstar:
This code allows labels of node to be set from cluster-autoscaler based on tags in asg of Alicloud. 
The tag format follows AWS's convention -  k8s.io/cluster-autoscaler/node-template/label/


losipiuk:
Tests would be nice to have. Although, as this is cloud provider code we are not very strict about that. 
Feel free to "/hold cancel" if you feel they are not needed.

mwielgus:
I guess someone from Alibaba cloud should review this. UnLGTMing.

mwielgus:
Closing per @ringtail comments on #1723.

PR #1721, opened 2019-02-25
Only soft taint nodes if there's no scale down to do

aleksandra-malinowska:
Soft tainting nodes when there are nodes we can delete immediately uses up quota, increasing the chance of delete failure. Moving it to trigger only as "best effort" - i.e. when there's nothing else to do and no errors.

cc @jkaniuk

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -394,10 +394,6 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 		} else {
 			klog.V(4).Infof("Starting scale down")
 
-			if a.AutoscalingContext.AutoscalingOptions.MaxBulkSoftTaintCount != 0 {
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -412,6 +408,13 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 				a.clusterStateRegistry.Recalculate()
 			}
 
+			if (scaleDownStatus.Result == status.ScaleDownNoNodeDeleted ||
+				scaleDownStatus.Result == status.ScaleDownInCooldown ||
```
jkaniuk: We are already in branch where ScaleDownInCooldown and ScaleDownInProgress did not happen.
If you decide that it should be moved outside of the branch, make sure that it would not remove soft taints after node deletion starts.

aleksandra-malinowska: You're completely right. Removing these conditions from the check to keep the change minimal.

PR #1723, opened 2019-02-25
Add labels support for Alibaba Cloud Provider

ringtail:
link to the pr (https://github.com/kubernetes/autoscaler/pull/1719) . @lookstar Hi, If you use autoscaler in Alibaba Cloud Kubernetes,it has supported labels right now. And I am sorry to forget merge these code from internal branch. The difference is we will attach every ess label as node label instead of the value under a specific key. because the other node labels are also plain tag.


mwielgus:
Should we close #1719?

ringtail:
@mwielgus Yes, We have done some offline contacts already. Thanks!

ringtail:
I am sorry to close this pr by accident.

PR #1724, opened 2019-02-25
Use patch request to taint nodes

aleksandra-malinowska:
Tainting by update frequently fails due to conflicts unrelated to taints.

### DIFF
--- a/cluster-autoscaler/core/scale_down_test.go
+++ b/cluster-autoscaler/core/scale_down_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package core
 
 import (
+	"encoding/json"
 	"fmt"
 	"sort"
 	"testing"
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/core/scale_down_test.go
```diff
@@ -514,6 +515,34 @@ func TestDeleteNode(t *testing.T) {
...
 					return true, obj, nil
 				})
+			fakeClient.Fake.AddReactor("patch", "nodes",
+				func(action core.Action) (bool, runtime.Object, error) {
+					patch := action.(core.PatchAction)
+					patchMap := make(map[string]interface{})
+					if err := json.Unmarshal(patch.GetPatch(), &patchMap); err != nil {
+						return false, nil, err
+					}
+					// Ugly but works for retrieving taints from patch.
```
losipiuk: Isn't there some utility (maybe in pkg/strategicpatch module) to apply a patch. It would be much more cleaner.

aleksandra-malinowska: Not that I could find any, but if you do, I'm happy to update the PR. Note this is only in test code (I wouldn't use this approach in production path).

losipiuk: `StrategicMergePatchUsingLookupPatchMeta`?

losipiuk: Also I am not a big fan of having different code quality standards for tests code and prod code. 
We spend a lot of time changing test code and we do not want to put a lot of tech-debt there as it basically slows us down.

aleksandra-malinowska:
Patching can accidentally overwrite unrelated taint changes due to race conditions. We've discussed this with @MaciekPytel and decided to try to improve chances of update succeeding by better retry/error handling logic.

PR #1353, opened 2018-10-25
Remove vertical-pod-autoscaler dependencies on k/k

justinsb:
* SelfLink was unused
* Build Schemas as needed
* Repoint imports to k8s.io/api

PR #1725, opened 2019-02-25
Godeps update to pull in klog

bskiba:
Prerequisite for #1626 

@Rajat-0


bskiba:
@kgolab

Rajat-0:
@bskiba thanks for updating deps.

kgolab:
I was a little worried about removed files, in particular testapi but it got removed via #1353 , this also removed dependency on apis/imagepolicy.
I'm not sure why parts of apis/admission are gone but they seem unused so LGTM.

PR #1735, opened 2019-02-27
Added target size cache to gceCache for GKEMig

vivekbagade:
Added migTargetSizeCache in GceCache to cache TargetSize calls for
each GKEMig for each loop of the CA. The cache is invalidated during
refresh that occurs during the start of each loop and during
CreateNodeGroup and DeleteNodeGroup.

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/cache.go
+++ b/cluster-autoscaler/cloudprovider/gce/cache.go
@@ -62,13 +62,14 @@ type MachineTypeKey struct {
 // RegenerateInstancesCache is required to sync it with registered migs.
 type GceCache struct {
 	// Cache content.
-	migs            []*MigInformation
-	instancesCache  map[GceRef]Mig
-	resourceLimiter *cloudprovider.ResourceLimiter
-	machinesCache   map[MachineTypeKey]*gce.MachineType
+	migs               []*MigInformation
+	instancesCache     map[GceRef]Mig
+	resourceLimiter    *cloudprovider.ResourceLimiter
+	machinesCache      map[MachineTypeKey]*gce.MachineType
+	migTargetSizeCache map[GceRef]int64
 	// Locks. Rules of locking:
 	// - migsMutex protects only migs.
-	// - cacheMutex protects instancesCache, resourceLimiter and machinesCache.
+	// - cacheMutex protects instancesCache, resourceLimiter, machinesCache and migTargetSizeCache.
 	// - if both locks are needed, cacheMutex must be obtained before migsMutex.
 	cacheMutex sync.Mutex
 	migsMutex  sync.Mutex
@@ -79,10 +80,11 @@ type GceCache struct {
 // NewGceCache creates empty GceCache.
 func NewGceCache(gceService AutoscalingGceClient) GceCache {
 	return GceCache{
-		migs:           []*MigInformation{},
-		instancesCache: map[GceRef]Mig{},
-		machinesCache:  map[MachineTypeKey]*gce.MachineType{},
-		GceService:     gceService,
+		migs:               []*MigInformation{},
+		instancesCache:     map[GceRef]Mig{},
+		machinesCache:      map[MachineTypeKey]*gce.MachineType{},
+		GceService:         gceService,
+		migTargetSizeCache: map[GceRef]int64{},
 	}
 }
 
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/cloudprovider/gke/gke_manager.go
```diff
@@ -498,7 +504,9 @@ func (m *gkeManagerImpl) refreshClusterResources() error {
 	if err != nil {
 		return err
 	}
-	m.refreshNodePools(cluster.NodePools)
+	if err = m.refreshNodePools(cluster.NodePools); err != nil {
```
aleksandra-malinowska: This makes me wonder how we want to handle errors here. I don't think a single broken node group should cause us to abort all operations, but then we probably shouldn't break from the loop in refreshNodePools either. @MaciekPytel WDYT?

aleksandra-malinowska:
Please also invalidate cache when node group is resized (i.e. on DeleteInstances() and SetMigSize()).

cluster-autoscaler/cloudprovider/gke/gke_manager.go
vivekbagade: +1 on that thought. We probably shouldn't react too fiercely on one error. Ideally we should have a simple error budget for the refreshNodePools call. Eg. Greater than 50% throw errors

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -244,6 +246,46 @@ func (gc *GceCache) GetResourceLimiter() (*cloudprovider.ResourceLimiter, error)
...
 }
 
+// GetMigTargetSize returns the cached targetSize for a GceRef
+func (gc *GceCache) GetMigTargetSize(ref GceRef) (int64, bool) {
+	gc.cacheMutex.Lock()
+	defer gc.cacheMutex.Unlock()
+
+	size, found := gc.migTargetSizeCache[ref]
+	if found {
+		klog.V(1).Infof("target size cache hit for %s", ref)
```
aleksandra-malinowska: increase the log level above 4

cluster-autoscaler/cloudprovider/gke/gke_manager.go
aleksandra-malinowska: I'm not confident with this change as-is. Since this is a drive-by fix, can we revert it and address the issue in a separate PR?

cluster-autoscaler/cloudprovider/gke/gke_manager.go
```diff
@@ -274,6 +274,8 @@ func (m *gkeManagerImpl) refreshNodePools(nodePools []NodePool) error {
 	existingMigs := map[gce.GceRef]struct{}{}
 	changed := false
 
+	m.cache.InvalidateTargetSizeCache()
```
MaciekPytel: This is not the right place to invalidate. It _must_ be invalidated every loop and refreshNodePools is only called once in a while (https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/gke/gke_manager.go#L476).

MaciekPytel:
Can you add the same logic to gce_manager.go? It's not really specific to GKE, it should work just as well for gce provider.

cluster-autoscaler/cloudprovider/gce/gce_manager_test.go
```diff
@@ -412,6 +413,44 @@ func TestGetMigSize(t *testing.T) {
...
+		},
+		gceManager: g,
+		minSize:    0,
+		maxSize:    1000,
+	}
+	size, err := g.GetMigSize(mig)
+	assert.NoError(t, err)
+	assert.Equal(t, int64(3), size)
+
+	err = g.SetMigSize(mig, 3)
```
MaciekPytel: Can you use a different value? This would pass if SetMigSize was a no-op.

MaciekPytel:
one comment to unittest, otherwise lgtm

cluster-autoscaler/cloudprovider/gce/gce_manager_test.go
vivekbagade: Sure, changed it. Got lazy there and didn't add a response. Though, I don't think that it'll be a no-op.

PR #1738, opened 2019-02-27
Cluster Autoscaler:  add GetInstanceID() for cloudprovider interface

feiskyer:
This PR adds a new `GetInstanceID()` for cloud provider interface, which is required for fixing the VMSS cases different issues.

For VMSS node's providerIDs, they may be in different cases depending user's PUT requests to the resources. This would cause autoscaler to delete all newly provisioned nodes. And more seriously, CA would scale up/down again and again.

This PR adds a workaround for this issue, so that they are handled in lower cases in CA.

/assign @mwielgus  @losipiuk

cc @andyzhangx @ritazh

### DIFF
--- a/cluster-autoscaler/cloudprovider/cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/cloud_provider.go
@@ -56,6 +56,9 @@ type CloudProvider interface {
 	// GetResourceLimiter returns struct containing limits (max, min) for resources (cores, memory etc.).
 	GetResourceLimiter() (*ResourceLimiter, error)
 
+	// GetInstanceID gets the instance ID for the specified node.
+	GetInstanceID(node *apiv1.Node) string
+
 	// Cleanup cleans up open resources before the cloud provider is destroyed, i.e. go routines etc.
 	Cleanup() error
 
--- a/cluster-autoscaler/cloudprovider/kubemark/kubemark_other.go
+++ b/cluster-autoscaler/cloudprovider/kubemark/kubemark_other.go
@@ -85,6 +85,11 @@ func (kubemark *KubemarkCloudProvider) Refresh() error {
 	return cloudprovider.ErrNotImplemented
 }
 
+// GetInstanceID gets the instance ID for the specified node.
+func (kubemark *KubemarkCloudProvider) GetInstanceID(node *apiv1.Node) string {
+	return ""
+}
+
 // Cleanup cleans up all resources before the cloud provider is removed
 func (kubemark *KubemarkCloudProvider) Cleanup() error {
 	return cloudprovider.ErrNotImplemented
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -281,7 +281,7 @@ func (csr *ClusterStateRegistry) UpdateNodes(nodes []*apiv1.Node, nodeInfosForGr
 	if err != nil {
 		return err
 	}
-	notRegistered := getNotRegisteredNodes(nodes, cloudProviderNodeInstances, currentTime)
+	notRegistered := getNotRegisteredNodes(csr.cloudProvider, nodes, cloudProviderNodeInstances, currentTime)
 
 	csr.Lock()
 	defer csr.Unlock()
@@ -932,10 +932,10 @@ func getCloudProviderNodeInstances(cloudProvider cloudprovider.CloudProvider) (m
 }
 
 // Calculates which of the existing cloud provider nodes are not registered in Kubernetes.
-func getNotRegisteredNodes(allNodes []*apiv1.Node, cloudProviderNodeInstances map[string][]cloudprovider.Instance, time time.Time) []UnregisteredNode {
+func getNotRegisteredNodes(cloudProvider cloudprovider.CloudProvider, allNodes []*apiv1.Node, cloudProviderNodeInstances map[string][]cloudprovider.Instance, time time.Time) []UnregisteredNode {
 	registered := sets.NewString()
 	for _, node := range allNodes {
-		registered.Insert(node.Spec.ProviderID)
+		registered.Insert(cloudProvider.GetInstanceID(node))
 	}
 	notRegistered := make([]UnregisteredNode, 0)
 	for _, instances := range cloudProviderNodeInstances {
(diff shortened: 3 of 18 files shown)


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -932,10 +932,10 @@ func getCloudProviderNodeInstances(cloudProvider cloudprovider.CloudProvider) (m
 }
 
 // Calculates which of the existing cloud provider nodes are not registered in Kubernetes.
-func getNotRegisteredNodes(allNodes []*apiv1.Node, cloudProviderNodeInstances map[string][]cloudprovider.Instance, time time.Time) []UnregisteredNode {
+func getNotRegisteredNodes(cloudProvider cloudprovider.CloudProvider, allNodes []*apiv1.Node, cloudProviderNodeInstances map[string][]cloudprovider.Instance, time time.Time) []UnregisteredNode {
 	registered := sets.NewString()
 	for _, node := range allNodes {
-		registered.Insert(node.Spec.ProviderID)
+		registered.Insert(cloudProvider.GetInstanceID(node))
```
losipiuk: Technically the commit with implementation should precede the usage.

feiskyer: Let me split it into another commit

feiskyer:
> Did you grep the code for usages of `ProviderId`?

Yep, the providerIDs got from cloud provider are only used in clusterstate.

cluster-autoscaler/cloudprovider/cloud_provider.go
```diff
@@ -56,6 +56,9 @@ type CloudProvider interface {
 	// GetResourceLimiter returns struct containing limits (max, min) for resources (cores, memory etc.).
 	GetResourceLimiter() (*ResourceLimiter, error)
 
+	// GetInstanceID gets the instance ID for the specified node.
+	GetInstanceID(node *apiv1.Node) string
```
mwielgus: Are we 100% percent sure that this is the right approach? Honestly, I have doubts that a (hopefully) temporary bugfix/workaround for minor issues in NodeController/Azure node registration process should involve an api change for all cloud providers (including ones that are not in this repo). Do we see any other applications of this method other "toLower" in Azure?

cluster-autoscaler/clusterstate/clusterstate.go
mwielgus: This is comparable to the previous code. Now, instead of `if (azure) providerId=tolower(providerId)` we have `cloudProvider.GetInstanceID(node)`

cluster-autoscaler/cloudprovider/cloud_provider.go
feiskyer: I don't expect to change all the cloud providers, but since the providerID is handled in clusterstate and it is case sensitive, the changes are still required. (there are also other ways, e.g. only this for Azure, but that's a little hacky).

In other applications, we are also keeping the resourceID of Azure VM, but when the resourceID is used for comparison, it is converted to lower or upper cases.

cluster-autoscaler/cloudprovider/kubemark/kubemark_other.go
```diff
@@ -85,6 +85,11 @@ func (kubemark *KubemarkCloudProvider) Refresh() error {
 	return cloudprovider.ErrNotImplemented
 }
 
+// GetInstanceID gets the instance ID for the specified node.
+func (kubemark *KubemarkCloudProvider) GetInstanceID(node *apiv1.Node) string {
+	return ""
```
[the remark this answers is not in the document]
feiskyer: keep same with other interfaces, e.g. `Name()`

cluster-autoscaler/clusterstate/clusterstate.go
feiskyer: yep, it's cleaner that previous one.

mwielgus:
@MaciekPytel @losipiuk @aleksandra-malinowska What do you think? I understand the problem and the need to solve it but I have mixed feelings about the approach - actually every approach would be ugly to some extent. Which is the least ugly to you?

losipiuk:
> @MaciekPytel Łukasz Osipiuk Aleksandra Malinowska What do you think? I understand the problem and the need to solve it but I have mixed feelings about the approach - actually every approach would be ugly to some extent. Which is the least ugly to you?

I like this PR more than explicit `if(azure)` call in core logic. It is mostly due to aesthetics, yet still. (I agree the two solutions are similar).
I am not concerned very much about API changes. As long as CPs live in this repository we can rollback the change easily if it is no longer needed.
As for CPs living elsewhere - we are not giving any guarantees about interface stability - and I think we should not give such.

With that said I will not force on this approach if over-voted.

feiskyer:
@mwielgus @losipiuk Thanks for the review. If we don't have other better solutions, could we get this way in first?

aleksandra-malinowska:
+1 from me. As was mentioned offline, it may also open the way for implementations that don't rely on ProviderId, which was brought up multiple times by users running into this not-so-obvious requirement.

mwielgus:
I still don't like it too much but we can merge it if neither @MaciekPytel or @losipiuk have strong negative feelings about it.

cluster-autoscaler/cloudprovider/cloud_provider.go
MaciekPytel: I'd like to understand if there is a plan to fix it in Azure (ie. make ProviderID match the name of VM)? Basically is this a temporary workaround? Are we planning to remove it later on? If so I'd like a comment stating that.

MaciekPytel:
/lgtm
/hold
Feel free to cancel hold after replying to my comment.

cluster-autoscaler/cloudprovider/cloud_provider.go
feiskyer: @MaciekPytel This is actually for fixing the issues with old releases, as providerID couldn't be changed after node initialized. 

For new releases, we should fix the providerID inconsistent issues in the cloud provider, probably cast to lower cases instead of using VM name.

mwielgus: Are you talking about master or node versions? How long will we support these old releases? What should w do in, let's say half a year from now - remove this method?

feiskyer: I mean cherry pick to old stable releases (for each kubernetes release). Those workarounds should be there and never get deleted. But for master branch, we may delete it after fixing the issues in cloud provider (e.g. in v1.14).

MaciekPytel: Ok, makes sense.

mwielgus:
Closing and reopening to reset stalled Travis.

PR #1740, opened 2019-03-01
Regenerate the ec2 instance types using latest metadata

stefansedich:
Originally I found that the g3s.xlarge instance type was not reporting a GPU so after much back and forth I finally got AWS to add it to the metadata endpoint.

The changes were generated by running `gen.go`, it looks to have fixed up the g3s.xlarge instance but also added a bunch of new instance types, if I ran it wrong let me know.

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/ec2_instance_types.go
+++ b/cluster-autoscaler/cloudprovider/aws/ec2_instance_types.go
@@ -27,6 +27,42 @@ type instanceType struct {
 
 // InstanceTypes is a map of ec2 resources
 var InstanceTypes = map[string]*instanceType{
+	"a1": {
+		InstanceType: "a1",
+		VCPU:         16,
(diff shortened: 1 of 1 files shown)


aleksandra-malinowska:
@Jeffwan is there any chance this list could be fetched from some API? It's great that the community keeps it up-to-date, but it really doesn't seem like good user experience.

Jeffwan:
@aleksandra-malinowska 
Yes. Actually there's a way to generate from API, the file is a little bit large ~50MBs as I remembered and if we move to that way initialization may takes some time.  What do you think?

stefansedich:
@Jeffwan I wonder if we can provide a snapshot config but then update in the background once it starts running, this would at least provide a baseline for quick initialization but then getting updates do not rely waiting for a release.

aleksandra-malinowska:
I'm fine with either way.

PR #1745, opened 2019-03-03
Allow custom AWS region overrides #1707

### LINKED ISSUE #1707: Allow custom AWS region overrides
Unique regional endpoints exist in disconnected AWS environments. Currently the aws-go-sdk that is utilized heavily in k8s hard codes these regions in json [1] and doesn't appear to expose an easy method to override that would flow down to a repo like the k8s/autoscaler. A similar issue [2] with a merged PR  [3] was filed against the k8s repo.

The previous solution was to read a file that specified the custom endpoints and associated services. This is something that would benefit those same environments tremendously and all for the use of the auto-scaler. I'm happy to take a crack at this patch, but it's taking me a while to find the right place in this repo to inject the reading of such a file or env var.

Because the hostnames and certs for these custom regions are private, they will never see the light of day in the aws-go-sdk models json file defining all available services. Being able to override these hostnames dynamically would be extremely helpful.

1. [AWS endpoints models](https://github.com/aws/aws-sdk-go/blob/182cda27d0921b14139ff6d352c09e0cb20e4578/models/endpoints/endpoints.json)
2. [RFE Allow to override default AWS endpoint](https://github.com/kubernetes/kubernetes/issues/70588)
3. [Add AWS Custom Endpoint capability #70588](https://github.com/kubernetes/kubernetes/pull/72245)

jottofar:
Replicated changes from kubernetes "Add AWS Custom Endpoint capability #70588" into cluster-autoscaler:
- Modified aws_manager similar to kubernetes aws.
- Modified kubernetes aws exposing type/methods used by cluster-autoscaler aws_manager.

Closes #1707

Fixes #1707

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_manager.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_manager.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
 	"github.com/aws/aws-sdk-go/service/ec2"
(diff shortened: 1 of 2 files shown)


bskiba:
Seems there is an issue with the cloudprovider/aws/aws_manager_test

cluster-autoscaler/vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/aws/aws.go
```diff
@@ -595,7 +595,7 @@ type CloudConfig struct {
 	}
 }
 
-func (cfg *CloudConfig) validateOverrides() error {
+func (cfg *CloudConfig) ValidateOverrides() error {
```
ingvagabund: Were all the methods capitalized by syncing the vendored file or by hand?

jottofar: Yes, that is how I did it to get the change in here quickly.

ingvagabund: Anything under vendor dir is untouchable. Since anytime autoscaler dependencies are updated, your changes are lost. The original idea was to have every component to implement its own mechanism of overriding the custom endpoints. Once we update all relevant component, we may identify which code is generic enough to be shared among components. Right now, it means to copy paste both `validateOverrides` and `getResolver` methods from the vendor dir into `cluster-autoscaler/cloudprovider/aws/aws_manager.go`.

jottofar: Thanks for the input. I replicated the necessary changes from vendor into cluster-autoscaler/cloudprovider/aws/aws_manager.go: CloudConfig struct and validateOverrides and getResolver methods.
Should autoscaler have it's own AWS config file with overrides and not rely on the one necessary for the Openshift install? I assume so to adhere to the original idea of every component implementing it's own mechanism.

cuppett: The intention (as I understand it) is to replicate the same code in here. Share the config struct, but just have another implementation of it here. That's how these are set up for the time being and have behaved historically.

jottofar: Ok, so I'll change it to reuse the vendor provided CloudConfig struct. I thought the intent may have been to completely remove any dependency on the vendor code.

ingvagabund: > Ok, so I'll change it to reuse the vendor provided CloudConfig struct. I thought the intent may have been to completely remove any dependency on the vendor code.

AWS endpoint customization was merged only into kubernetes master branch. New feature are not usually backported into older releases unless they are proven very useful and important. Given (and please prove me wrong) the autoscaler deps are synced only with the stable versions of kubernetes, it would take some time to fetch the master changes into the vendor directory anyway. So yes, there was no intent to have the AWS endpoint customization to be consumed from the vendor directory.

ingvagabund:
@aleksandra-malinowska PTAL

mwielgus:
@ingvagabund This PR requires LGTM from someone from AWS. 

@Jeffwan Ping :).

Jeffwan:
@mwielgus Miss this PR, I will have a look and leave comments by EOD.

cuppett:
@Jeffwan Does somebody else have to "approve" this also?

mwielgus:
@cuppett No, we merge (approve) as soon as we get LGTM from AWS.

PR #1746, opened 2019-03-04
Improve AWS "gotchas" list and discoverability of cloud-provider READMEs.

### LINKED ISSUE #1744: AWS AZ Rebalancing Unexpectedly Terminates Nodes
Not strictly speaking an autoscaler issue, but I will raise a PR to at least mention this in the docs if nothing else.

We couldn't work out why some of our nodes were getting _surprise_ terminated without being drained first. Our apps need to get a nice SIGTERM and shutdown nicely or things get a bit wobbly.

I've now traced it to a feature of AWS Autoscaling Groups, which is designed to keep an even number of instances in each AZ. It will automatically spin up a new instance in the AZ with less instances and terminate an instance in the AZ with too many. Which is of course not very helpful in our use case. https://docs.aws.amazon.com/autoscaling/ec2/userguide/auto-scaling-benefits.html#arch-AutoScalingMultiAZ

Here's the docs of how to disable or "Suspend" this feature for your Autoscaling group, which I imagine anyone using this autoscaler on AWS will probably want to do: https://docs.aws.amazon.com/autoscaling/ec2/userguide/as-suspend-resume-processes.html

cablespaghetti:
Better document the caveats of running a multi-az autoscaling group with the cluster autoscaler on AWS. Also help people find the cloud provider specific documentation rather than just the FAQ.

Fixes #1744.


bskiba:
Thank you for this change!

I wonder now, this is about multi-zone ASGs, right? Wasn't our answer that Cluster Autoscaler shouldn't be used at all with ASGs spanning multiple zones, as it expects to have homogeneous node-groups to scale and you should always do what is described in third bullet - use an ASG per zone?
@aleksandra-malinowska @MaciekPytel

cablespaghetti:
I should probably change my setup to use multiple ASGs then and update the PR.

cablespaghetti:
Hopefully this better encourages recommended practice whilst still explaining what's going on to people with the same issue as me.

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -142,8 +142,8 @@ If you'd like to scale node groups from 0, an `autoscaling:DescribeLaunchConfigu
 ```
 
 ## Common Notes and Gotchas:
-- The `/etc/ssl/certs/ca-certificates.crt` should exist by default on your ec2 instance. If you use Amazon Linux 2 (EKS woker node AMI by default), use `/etc/ssl/certs/ca-bundle.crt` instead.
-- Cluster autoscaler is not zone aware (for now), so if you wish to span multiple availability zones in your autoscaling groups beware that cluster autoscaler will not evenly distribute them. For more information, see https://github.com/kubernetes/contrib/pull/1552#discussion_r75532949.
+- The `/etc/ssl/certs/ca-certificates.crt` should exist by default on your ec2 instance. If you use Amazon Linux 2 (EKS worker node AMI by default), use `/etc/kubernetes/pki/ca.crt` instead for the volume hostPath of your cluster-autoscaler manifest.
```
bskiba: nit: "instead of the volume hostPath of your cluster-autoscaler manifest"

bskiba:
This looks great, thank you. One nit.

cablespaghetti:
Hopefully that makes a little more sense.

PR #1768, opened 2019-03-07
Fixing minor error handling bug in static autoscaler

uruddarraju:
The previous code was checking an invalid error condition that no longer is valid in the statement's scope.


uruddarraju:
I signed it

lxlxok:
@uruddarraju  LGTM 😄

PR #1777, opened 2019-03-11
Fix bulk removal issues of AKS scaling

### LINKED ISSUE #1597: Azure AKS cluster autoscaler doesn't support bulk removal of empty nodes
CloudProvider: Azure AKS
Kubernetes version: 1.11.5
Cluster Autoscaler version: 1.3.5 and current master
ACS-Engine version: v0.26.3-aks

In our setup we run an Azure AKS cluster with multiple nodes that each run a single pod requiring all resources on the node. The pods are part of a statefulset that we scale quite aggressively from 2 to 20 and then back to 2 replicas at regular intervals.

When scaling down we observe unexpected behaviours in the cluster autoscaler, where initially nodes are removed quickly but then mysteriously are added back a few minutes later.

I believe this is caused by `scheduleDeleteEmptyNodes` bulk removing nodes in parallel. And `DeleteNodesInternal` in `azure_container_service_pool.go` decreasing the AKS cluster size with only one instead of actual number of nodes in the bulk set, since the function is called multiple times in parallel with a list of only a single VM each time.

The effect is that scaling down from 20 to 2 nodes takes about 1.5 to 2 hours for us. A workaround is setting `--max-empty-bulk-delete=1` (default: 10) since this decreases the scale down to 30 minutes, but running in parallel should be much faster.

feiskyer:
Fix bulk removal issues of AKS scaling:

> When scaling down we observe unexpected behaviors in the cluster autoscaler, where initially nodes are removed quickly but then mysteriously are added back a few minutes later.

The issue happens because after the virtual machines are removed, CA would also tries to correct the node count by invoking the AKS PUT API. Unfortunately, for bulk removal, multiple virtual machines are deleted in different Go routines, and hence their calculated new counts are wrong.

This PR fixed this issue by adding a lock for ContainerServiceAgentPool. It also adds a cache for getting current node count, so as to reduce the API calls.

Fix #1597

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_container_service_pool.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_container_service_pool.go
@@ -19,6 +19,8 @@ package azure
 import (
 	"fmt"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2018-03-31/containerservice"
 	"k8s.io/klog"
@@ -41,6 +43,10 @@ type ContainerServiceAgentPool struct {
 	clusterName       string
 	resourceGroup     string
 	nodeResourceGroup string
+
+	curSize     int
+	lastRefresh time.Time
+	mutex       sync.Mutex
 }
 
 //NewContainerServiceAgentPool constructs ContainerServiceAgentPool from the --node param
@@ -53,6 +59,7 @@ func NewContainerServiceAgentPool(spec *dynamic.NodeGroupSpec, am *AzureManager)
 		minSize: spec.MinSize,
 		maxSize: spec.MaxSize,
 		manager: am,
+		curSize: -1,
 	}
 
 	asg.util = &AzUtil{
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/cloudprovider/azure/azure_container_service_pool.go
```diff
@@ -307,24 +314,55 @@ func (agentPool *ContainerServiceAgentPool) MinSize() int {
 //TargetSize gathers the target node count set for the cluster by
 //querying the underlying service.
 func (agentPool *ContainerServiceAgentPool) TargetSize() (int, error) {
-	return agentPool.GetNodeCount()
+	agentPool.mutex.Lock()
+	defer agentPool.mutex.Unlock()
+
+	if agentPool.lastRefresh.Add(15 * time.Second).After(time.Now()) {
+		return agentPool.curSize, nil
```
andyzhangx: is it possible that the returned target size is not accurate since it's cached? what behavior would happen if it's a wrong num?

feiskyer: yep, if user invokes the scaling operation on other places, then the count may be wrong. In that case, autoscaler would correct it back based on resource usage.

Issue #1133, opened 2018-08-09
Node pool scale up timeout


aermakov-zalando:
The autoscaler has a timeout for non-ready nodes which forces it to kill those nodes and potentially select a different node pool in the next iteration. However, in the situation where the node pool cannot scale up at all it'll happily wait forever, keeping pods in Pending state without trying to compensate.

For example, setting multiple AWS Spot node pools with different instance types, or setting up a Spot pool and an On Demand pool doesn't really work. We'd expect CA to scale up one of the ASGs, detect a few minutes later that there's still no nodes coming up (because the corresponding Spot pool doesn't have capacity) and fall back to another pool. What actually happens is that CA will scale up the node pool by increasing desired capacity and then not do anything at all other than printing `Upcoming 1 nodes`/`Failed to find readiness information for ...`.

MaciekPytel:
CA should timeout scale-up (IIRC after 15 minutes) and put the node group that failed to scale-up in 'backoff' state. At that point it should try to scale-up again, ignoring this node group. You can see this by looking for one of the following:
* `Scale-up timed out for node group <name> after <time>` in logs
* `ScaleUpTimedOut` event on cluster-autoscaler-status configmap (in kube-system ns)
* Node group showing scale-up status as "Backoff" in status configmap
* `failed_scale_ups_total` metric increasing

If you can reproduce it and doesn't see any of the above it's a bug. In that case can you provide some more details (especially CA version you're using).

aermakov-zalando:
We're running version 1.2.2, but I can try with 1.3.1.

MaciekPytel:
It shouldn't make any difference, this was added earlier than 1.2 (I don't remember exactly, but probably 1.1 timeframe?).

So you're saying it's stuck on `Upcoming 1 node` for more than 15 minutes? Can you provide a log of initial scale-up, a loop immediately after and another loop after 15+ minutes?

MaciekPytel:
As @aleksandra-malinowska pointed out to me the timeout is effectively reset if there is another scale-up on the same node group (ie. CA only notices it if the last of multiple overlapping scale-ups times out). So that may be another thing to look for in the logs.

aermakov-zalando:
The logs are in [this gist](https://gist.github.com/aermakov-zalando/c9465f420ba13b1814b1474531629792). Please ignore all ASGs and pods in `1b`, the nodes there were being destroyed by the Spot termination.

[...]

PR #1779, opened 2019-03-11
Ignoring upcoming nodes from unhealthy nodegroups.

viggeh:
This is a fix for https://github.com/kubernetes/autoscaler/issues/1133

I set the base branch as the 1.13 release branch. Please let me know if that is not the way to go.

If a nodegroup is in an Unhealthy state but has upcoming nodes, it is still considered to be ready to serve a pod that cannot be scheduled.
This means that other nodegroups that might be able to host the pod will not be expanded.

This happens for instance on AWS when you have an ASG with spot instances and the your price is either too high or the market does not have any spot instances available. The autoscaling group will be marked as `Unhealthy` and yet have upcoming nodes that will (maybe) never come. Other autoscaling groups will not get the chance to be expanded.

### DIFF
--- a/cluster-autoscaler/core/scale_up.go
+++ b/cluster-autoscaler/core/scale_up.go
@@ -285,6 +285,10 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 
 	upcomingNodes := make([]*schedulercache.NodeInfo, 0)
 	for nodeGroup, numberOfNodes := range clusterStateRegistry.GetUpcomingNodes() {
+		if !clusterStateRegistry.IsNodeGroupHealthy(nodeGroup) {
+			klog.Warningf("Node group %s has %d upcoming nodes but is unhealthy. Ignoring those nodes.", nodeGroup, numberOfNodes)
+			continue
+		}
 		nodeTemplate, found := nodeInfos[nodeGroup]
 		if !found {
 			return &status.ScaleUpStatus{Result: status.ScaleUpError}, errors.NewAutoscalerError(


cluster-autoscaler/core/scale_up.go
```diff
@@ -285,6 +285,10 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 
 	upcomingNodes := make([]*schedulercache.NodeInfo, 0)
 	for nodeGroup, numberOfNodes := range clusterStateRegistry.GetUpcomingNodes() {
+		if !clusterStateRegistry.IsNodeGroupHealthy(nodeGroup) {
+			klog.Warningf("Node group %s has %d upcoming nodes but is unhealthy. Ignoring those nodes.", nodeGroup, numberOfNodes)
```
mwielgus: Even if 80% of the group is broken, but some nodes are just being started, we should not ignore them. They have a chance of being healthy and should be included in the upcoming nodes for further evaluation.

viggeh: `upcomingNodes` is only being used to evaluate if a nodeGroup should be expanded. The assumption being that all the upcoming nodes are indeed upcoming, healthy nodes.

The problem is that in some cases, like with spot instances in AWS, the nodes will never show up. So the worst case scenario is that a pod will never be scheduled.
If upcoming nodes from a broken nodegroup do end up showing up healthy, the worst case scenario is that we have over-provisioned the cluster and it will need to scale down in the next possible ScaleDown event.

MaciekPytel: The fact that nodegroup is unhealthy doesn't imply that the upcoming nodes won't turn up ok. They may or may not and your solution is trading one problem for the other. 

There is already a mechanism in CA for handling your case - if we know the nodes will never show up, NodeGroup.Nodes() should return Instances specifying InstanceState=InstanceCreating and appropriate ErrorInfo. If  CA gets such Instances from cloudprovider it will delete those VMs immediately, backoff nodegroup and trigger a scale-up in a different nodegroup. I think the right way to fix your issue would be to improve implementation of NodeGroup.Nodes() in AWS cloudprovider, so that it's actually returns Instances with appropriate error.

mwielgus: @viggeh Please fix the problem in the way @MaciekPytel described and submit a new PR. I'm closing this PR as the right solution will be significantly different.

viggeh: Thank you. I'll take a look at that approach.

PR #1785, opened 2019-03-12
Forbid creating VPA objects without targetRef

kgolab:
Follow-up to #1783 .

It's a bit of a poor man's solution since we don't have the version nor v1beta1 fields at hand so I've used (lack of) TargetRef for validation.
Moreover, the validation can be by-passed by updating the object but I guess at that point the user has been already warned.

An attempt to create v1beta1 object (or simply v1beta2 w/o TargetRef) ends up with:
```
Error from server (InternalError): error when creating "b1.yaml": Internal error occurred: admission webhook "vpa.k8s.io" denied the request: TargetRef is required
```

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
@@ -147,7 +147,7 @@ var (
 	}
 )
 
-func validateVPA(vpa *vpa_types.VerticalPodAutoscaler) error {
+func validateVPA(vpa *vpa_types.VerticalPodAutoscaler, isCreate bool) error {
 	if vpa.Spec.UpdatePolicy != nil {
 		mode := vpa.Spec.UpdatePolicy.UpdateMode
 		if mode == nil {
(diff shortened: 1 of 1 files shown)


vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
```diff
@@ -178,16 +178,22 @@ func validateVPA(vpa *vpa_types.VerticalPodAutoscaler) error {
 		}
 	}
 
+	if isCreate {
```
bskiba: You can fold the two ifs into one.

vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
```diff
@@ -147,7 +147,7 @@ var (
 	}
 )
 
-func validateVPA(vpa *vpa_types.VerticalPodAutoscaler) error {
+func validateVPA(vpa *vpa_types.VerticalPodAutoscaler, isCreate bool) error {
```
bskiba: nit: I am inclined to pass ar.Request.Operation instead of boolean but you can ignore my inclination :)

bskiba:
Nits only

vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
```diff
@@ -178,16 +178,22 @@ func validateVPA(vpa *vpa_types.VerticalPodAutoscaler) error {
 		}
 	}
 
+	if isCreate {
+		if vpa.Spec.TargetRef == nil {
+			return fmt.Errorf("TargetRef is required")
```
bskiba: Actually one more thing, can you make this more descriptive? Sth like, you are trying to use an unsupported version of the api, please migrate to v1beta2 and use targetRef to specify pods to autoscale

vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
kgolab: Sorry, I started with a more complicated code and haven't cleaned this. Fixed.

vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
kgolab: I'll choose "ignore" this time.

vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
kgolab: Reworded.
Since we're not sure of the API version used here, it makes the message a bit long but maybe it's for the best.

kgolab:
@bskiba , please take another look.

bskiba:
Thanks!
/lgtm
/approve

PR #1791, opened 2019-03-13
Notes about multiple AWS AZ  and PV

ismailyenigul:


Jeffwan:
This is a known issue. Thanks for addressing it in the documentation. 
Once I get some time, I will check if there's a tradeoff to make CA to respect pod's PV but this will against multiple zone setting.

ismailyenigul:
> Once I get some time, I will check if there's a tradeoff to make CA to respect pod's PV but this will against multiple zone setting.

Actually, the problem is using ASG with  Multi-AZ.  ASG is not aware of EBS limitation. It just launches a new node instance if needed more resource. Also, ASG may terminate the single node in the AZ where we created PV :) then our pod will be unusable until ASG creates another node in this AZ again.

Jeffwan:
> > Once I get some time, I will check if there's a tradeoff to make CA to respect pod's PV but this will against multiple zone setting.
> 
> Actually, the problem is using ASG with Multi-AZ. ASG is not aware of EBS limitation. It just launches a new node instance if needed more resource. Also, ASG may terminate the single node in the AZ where we created PV :) then our pod will be unusable until ASG creates another node in this AZ again.

Yes. That's by design and something we can not control on ASG. The thing we can control is inside CA but CA also needs to respect ASG's behavior.  Anyway, once I get some idea, I will talk with you on this.

ismailyenigul:
Here is the reply from AWS Support team.

I think EFS is not  a right solution for databases.
Options 2 requires creating at least one node per AZ.

```Solution 1
-----

Using ElasticFileSystem (EFS)

AWS EFS is the scalable, shared and fully managed network file system. You can use EFS across Availability Zones. Postgres supports network file systems as mentioned here [1]. There is also a kubernetes incubator project for provisioning EFS storage as persistent volumes in kubernetes. You can refer to it here [2]. Here, I am not sure if RabbitMQ supports EFS. So, solution 2 described below might help in this case.

Solution 2
-----

Assigning Pods to Nodes

Within this approach you can schedule the pods on nodes (instances) in particular Availability Zone. To follow this approach, you should first put node labels on the instances. For example, AZ=<az>. Labels can be put on the instance while bootstrapping.  Following is the way the arguments are passed to EKS worker node bootstrap script.

--kubelet-extra-args --node-labels=AZ=$(curl http://169.254.169.254/latest/meta-data/placement/availability-zone)

After the labels are assigned to the worker nodes, Pods can be assigned to the nodes using the nodeSelector 

You can follow this approach, to schedule nodes on the instances with particular AZ labels.

by default Amazon EC2 Auto Scaling attempts to distribute instances evenly between the Availability Zones that are enabled for your Auto Scaling group. So, if the min instance count within the AutoScaling Group is kept equal to number of Availability Zones within the region, then there will be at least 1 instance within each AZ. And, as node Selector configuration will assign the pods on the instance within particular Availability Zone.
```

Issue #1790, opened 2019-03-13
VPA for CronJob


estahn:
Would be great to have the VPA update CronJob resources.

bskiba:
Seems reasonable, we need to find a consistent way to fetch a labelSelector from a CronJob. (CronJob does not support scale subresource).
@schylek @kgolab FYI

bskiba:
This should not be very difficult to support. The parts to fix are:
- give VPA recommender permissions to read CronJobs
- add code to return JobSpec.LabelSelector for CronJobs in this switch: https://github.com/kubernetes/autoscaler/blob/e79fbeee44f6474cd80029ded60ef092d6a4c602/vertical-pod-autoscaler/pkg/target/fetcher.go#L153

Marking this as help-wanted if anyone wants to pick it up. Will be happy to review.

estahn:
@bskiba Thanks for your input. I will make it my Friday project and report back.

schylek:
One small addition: CronJob has to be added to the list of observed controllers as it is already happening for Job. https://github.com/kubernetes/autoscaler/blob/e79fbeee44f6474cd80029ded60ef092d6a4c602/vertical-pod-autoscaler/pkg/target/fetcher.go#L86

PR #1797, opened 2019-03-15
[WIP]VPA for CronJob

Rajat-0:
#1790

### DIFF
--- a/vertical-pod-autoscaler/pkg/target/fetcher.go
+++ b/vertical-pod-autoscaler/pkg/target/fetcher.go
@@ -23,6 +23,7 @@ import (
 	"github.com/golang/glog"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
+	v1batch "k8s.io/api/batch/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -61,6 +62,7 @@ const (
 	statefulSet           wellKnownController = "StatefulSet"
 	replicationController wellKnownController = "ReplicationController"
 	job                   wellKnownController = "Job"
+	cronJob               wellKnownController = "CronJob"
 )
 
 // NewVpaTargetSelectorFetcher returns new instance of VpaTargetSelectorFetcher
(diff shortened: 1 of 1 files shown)


vertical-pod-autoscaler/pkg/target/fetcher.go
```diff
@@ -187,6 +190,12 @@ func getLabelSelector(informer cache.SharedIndexInformer, kind, namespace, name
 			return nil, fmt.Errorf("Failed to parse %s %s/%s", kind, namespace, name)
 		}
 		return metav1.LabelSelectorAsSelector(metav1.SetAsLabelSelector(apiObj.Spec.Selector))
+	case (*v1batch.CronJob):
+		apiObj, ok := obj.(*v1batch.CronJob)
+		if !ok {
+			return nil, fmt.Errorf("Failed to parse %s %s/%s", kind, namespace, name)
+		}
+		return metav1.LabelSelectorAsSelector(metav1.SetAsLabelSelector(apiObj.Spec.JobTemplate.Spec.Selector))
```
mwielgus: This will not work in all (most?) cases. Here is a sample CronJob i created with kubectl run command. It doesn't have Selector. Selector for (Cron)Jobs is optional.
```
$  kubectl run --schedule="* * * * *" bb --image=busybox --restart=Never --command -- date
$ kubectl get cronjob -o yaml
apiVersion: v1
items:
- apiVersion: batch/v1beta1
  kind: CronJob
  metadata:
    creationTimestamp: 2019-03-15T12:59:13Z
    labels:
      run: bb
    name: bb
    namespace: default
    resourceVersion: "1162"
    selfLink: /apis/batch/v1beta1/namespaces/default/cronjobs/bb
    uid: 22104d0a-4722-11e9-a206-0242ac110028
  spec:
    concurrencyPolicy: Allow
    failedJobsHistoryLimit: 1
    jobTemplate:
      metadata:
        creationTimestamp: null
      spec:
        template:
          metadata:
            creationTimestamp: null
            labels:
              run: bb
          spec:
            containers:
            - command:
              - date
              image: busybox
              imagePullPolicy: Always
              name: bb
              resources: {}
              terminationMessagePath: /dev/termination-log
              terminationMessagePolicy: File
            dnsPolicy: ClusterFirst
            restartPolicy: Never
            schedulerName: default-scheduler
            securityContext: {}
            terminationGracePeriodSeconds: 30
    schedule: '* * * * *'
    successfulJobsHistoryLimit: 3
    suspend: false
  status:
    lastScheduleTime: 2019-03-15T13:06:00Z
[...]
```

Rajat-0: Thanks for the review I'll update it accordingly.

mwielgus:
Inactivity warning. Please proceed with the PR or it will be closed.

Rajat-0:
@mwielgus I am looking for a way via which CronJob without selector can be used.

mwielgus:
@Rajat-0 That is totally OK, and your contribution is more than welcome. 
Nothing bad will happen if you, once you find a proper solution, open another PR, especially if your approach will change significantly. I suspect that not much of the current content will remain there, so starting a new PR would probably be even advisable :).

PR #1801, opened 2019-03-18
New expander: priority expander

piontec:
## The PR

This commit introduces code, tests and docs for a new expander called 'priority'. The motivation is placed in [this proposal](proposals/priority_expander.md). Docs are updated as well in [FAQ](FAQ.md#what-are-expanders).

## Testing procedure
Full autoamted kubernetes e2e tests were not done. The testing procedure was as follows:
1. Unit tests
  `go test ./expander/priority/...`

1. Manual e2e tests
    - I started a test cluster and connected the new cluster-autoscaler binary to it
    - checked for selection of the correct node group
    - check for handling of ConfigMap reloads

### DIFF
--- a/cluster-autoscaler/FAQ.md
+++ b/cluster-autoscaler/FAQ.md
@@ -590,7 +590,7 @@ new nodes will be added.
 Expanders can be selected by passing the name to the `--expander` flag, i.e.
 `./cluster-autoscaler --expander=random`.
 
-Currently Cluster Autoscaler has 4 expanders:
+Currently Cluster Autoscaler has 5 expanders:
 
 * `random` - this is the default expander, and should be used when you don't have a particular
 need for the node groups to scale differently.
@@ -607,12 +607,15 @@ after scale-up. This is useful when you have different classes of nodes, for exa
 would match the cluster size. This expander is described in more details
 [HERE](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/proposals/pricing.md). Currently it works only for GCE and GKE (patches welcome.)
 
+* `priority` - selects the node group that has the highest priority assigned by the user. It's configuration is described in more details [here](expander/priority/readme.md)
+
 ### Does CA respect node affinity when selecting node groups to scale up?
+
 CA respects `nodeSelector` and `requiredDuringSchedulingIgnoredDuringExecution` in nodeAffinity given that you have labelled your node groups accordingly. If there is a pod that cannot be scheduled with either `nodeSelector` or `requiredDuringSchedulingIgnoredDuringExecution` specified, CA will only consider node groups that satisfy those requirements for expansion.
 
 However, CA does not consider "soft" constraints like `preferredDuringSchedulingIgnoredDuringExecution` when selecting node groups. That means that if CA has two or more node groups available for expansion, it will not use soft constraints to pick one node group over another.
 
-************
+****************
 
 ### What are the parameters to CA?
 
--- a/cluster-autoscaler/cloudprovider/alicloud/examples/cluster-autoscaler-standard.yaml
+++ b/cluster-autoscaler/cloudprovider/alicloud/examples/cluster-autoscaler-standard.yaml
@@ -63,8 +63,8 @@ rules:
   verbs: ["create"]
 - apiGroups: [""]
   resources: ["configmaps"]
-  resourceNames: ["cluster-autoscaler-status"]
-  verbs: ["delete","get","update"]
+  resourceNames: ["cluster-autoscaler-status", "cluster-autoscaler-priority-expander"]
+  verbs: ["delete","get","update","watch"]
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1beta1
--- a/cluster-autoscaler/expander/priority/configmap_handler.go
+++ b/cluster-autoscaler/expander/priority/configmap_handler.go
@@ -1,0 +1,73 @@
+/*
+Copyright 2016 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package priority
+
+import (
+	"errors"
+	"fmt"
+
+	kube_errors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/watch"
+	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/klog"
+	api "k8s.io/kubernetes/pkg/apis/core"
+)
+
+const (
+	// PriorityConfigMapName defines a name of the ConfigMap used to store priority expander configuration
+	PriorityConfigMapName = "cluster-autoscaler-priority-expander"
+	// ConfigMapKey defines the key used in the ConfigMap to configure priorities
+	ConfigMapKey = "priorities"
+)
+
+// InitPriorityConfigMap initializes ConfigMap with priority expander configurations. It checks if the map exists
+// and has the correct top level key. If it doesn't, it returns error or Exits. If the value is found,
+// the current value is fetched and a Watcher is started to watch for changes. It returns the current value of
+// the config map, the channel with value updates and an error.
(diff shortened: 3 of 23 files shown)


cluster-autoscaler/expander/priority/priority.go
```diff
@@ -0,0 +1,162 @@
...
+			}
+
+			prioString, found := cm.Data[ConfigMapKey]
+			if !found {
+				klog.Warningf("Wrong configmap for priority expander, doesn't contain %s key. Ignoring update.", ConfigMapKey)
+				res.badConfigUpdates++
+				continue
+			}
+			if err := res.parsePrioritiesYAMLString(prioString); err != nil {
+				klog.Warningf("Wrong configuration for priority expander: %v. Ignoring update.", err)
```
mwielgus: I would also send an event, update status config map, etc. Be extremely noisy in case of misconfiguriation.

cluster-autoscaler/FAQ.md
```diff
@@ -606,6 +606,26 @@ after scale-up. This is useful when you have different classes of nodes, for exa
 would match the cluster size. This expander is described in more details
 [HERE](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/proposals/pricing.md). Currently it works only for GCE and GKE (patches welcome.)
 
+* `priority` - selects the node group that has the highest priority assigned by the user. The priority configuration is based on the values stored in a ConfigMap. This ConfigMap has to be created before cluster autoscaler with priority expander can be started. The ConfigMap must be named `cluster-autoscaler-priority-expander` and it must be placed in the same namespace as cluster autoscaler pod. The format of the [ConfigMap](expander/priority/priority-expander-configmap.yaml) is as follows:
```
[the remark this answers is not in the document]
piontec: Do you want it only in expander dir and linked here?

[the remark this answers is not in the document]
mwielgus: Just a brief description and a link, please.

piontec:
@mwielgus Changes are in place. I added events only, as the status-map seemed not to be a good fit to me: it keeps CAS-wide info only, no matter what expander you use. And if I'm not wrong, no other expander keeps status there. Still, if you think it would be beneficial, I can work on it.

mwielgus:
```
!!! 'golint' problems: 
/home/travis/gopath/src/k8s.io/autoscaler/cluster-autoscaler/expander/priority/priority.go:181:6: exported type EventRecorder should have comment or be unexported
```

mwielgus:
You have too many (21) commits in this PR. Please squash them to no more than 3 commits.

mwielgus:
"merge with master" commit has misleading/ambiguous name. Please rename or squash with "priority expander".

piontec:
rebased on master

cluster-autoscaler/cloudprovider/alicloud/examples/cluster-autoscaler-standard.yaml
```diff
@@ -63,8 +63,8 @@ rules:
   verbs: ["create"]
 - apiGroups: [""]
   resources: ["configmaps"]
-  resourceNames: ["cluster-autoscaler-status"]
-  verbs: ["delete","get","update"]
+  resourceNames: ["cluster-autoscaler-status", "cluster-autoscaler-priority-expander"]
```
MaciekPytel: Once this merges we should also update https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/rbac/cluster-autoscaler/cluster-autoscaler-rbac.yaml#L54.

piontec:
@mwielgus Thanks for the review! One question though: what is the best approach to backport it, we would need it back in 1.11 (1.3) kubernetes release.

cluster-autoscaler/expander/priority/configmap_handler.go
```diff
@@ -0,0 +1,73 @@
...
+			"with priority expander.", namespace, PriorityConfigMapName)
+	} else {
+		errMsg = fmt.Sprintf("Failed to retrieve priority expander configmap %s/%s: %v", namespace, PriorityConfigMapName,
+			getStatusError)
+	}
+	if errMsg != "" {
+		return "", nil, errors.New(errMsg)
+	}
+
+	watcher, err := maps.Watch(metav1.ListOptions{
```
MaciekPytel: Listers already handle watch cache and all necessary error handling and we use those for all other objects. I don't think we should re-implement this functionality. In particular if you want to use watch directly you need to handle channel closing - in both HPA and VPA we've already run into issues caused by watch channel abruptly closing and we needed to implement a retry logic for this case. I don't see such retry logic in this PR. 
That being said - unless you have a very strong reason I strongly recommend using a lister or an informer, rather than directly using watch API.

cluster-autoscaler/expander/priority/priority.go
```diff
@@ -0,0 +1,189 @@
...
+	for _, re := range nameRegexpList {
+		if re.FindStringIndex(id) != nil {
+			return true
+		}
+	}
+	return false
+}
+
+// EventRecorder is an interface to abstract kubernetes event recording.
+type EventRecorder interface {
```
MaciekPytel: Why are you redefining this here?

cluster-autoscaler/expander/priority/priority.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	if err := res.parsePrioritiesYAMLString(initialPriorities); err != nil {
+		return nil, errors.ToAutoscalerError(errors.InternalError, err)
+	}
+	go func() {
+		// TODO: how to terminate on process shutdown?
+		for event := range priorityChangesChan {
+			cm, ok := event.Object.(*apiv1.ConfigMap)
+			if !ok {
+				klog.Exit("Unexpected object type received on the configmap update channel in priority expander")
```
MaciekPytel: You should return an appropriate error (API or Internal), not crash the application.

cluster-autoscaler/expander/priority/priority.go
```diff
@@ -0,0 +1,189 @@
...
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/klog"
+	schedulernodeinfo "k8s.io/kubernetes/pkg/scheduler/nodeinfo"
+)
+
+type priority struct {
+	fallbackStrategy expander.Strategy
+	changesChan      <-chan watch.Event
+	priorities       map[int][]*regexp.Regexp
+	padlock          sync.RWMutex
```
MaciekPytel: Expander is called once per loop, at most 1 call / 10s (more likely 1 per 15-20s), after a ton of complex calculations done all over CA code. Do we really need async config update? Feels like we could just get configmap from watch cache every time and synchronously check if the config changed - the performance impact should be negligible and the complexity of this expander will greatly reduce.

cluster-autoscaler/expander/priority/configmap_handler.go
piontec: I didn't know about any problems using watch API - when is the channel closed? Can you point me to some docs?

cluster-autoscaler/expander/priority/priority.go
piontec: For testability - check unit tests. Have I missed exactly the same interface being already defined elsewhere?

cluster-autoscaler/expander/priority/priority.go
piontec: OK.

cluster-autoscaler/expander/priority/priority.go
piontec: Hmm, I agree, if there's no plan to make the updates/loops more frequent in the future, probably we can drop the watch and go with fetching the map every time. I'm just not sure about caching - can you point me to some docs? How is stuff cached, for how long and what is your default caching toolset for CAS?

MaciekPytel: By 'watch cache' I mean using listers or informers. Those are a higher level abstraction for Kubernetes client provided by client-go. They provide a local cache that is backed by a watch on apiserver and updated on any change. So you get fresh data, except you don't have to care about creating goroutines doing the watch and error handling - it's all provided.
The general recommendation is to always use lister or informer over raw client, unless there is a very good reason not to.

cluster-autoscaler/expander/priority/configmap_handler.go
MaciekPytel: I don't know if there is any documentation on the details. IIRC @jbartosik once run into similar issue, so he may provide more details. The general recommendation is to always use higher-level abstractions provided by client-go library, rather than raw client.

cluster-autoscaler/expander/priority/priority.go
MaciekPytel: After taking a closer look at unittests - they're very inconsistent with existing unittests. We have utilities we use for making fake nodegroups (https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/test/test_cloud_provider.go) and there is an existing utility for faking event recording (used for example here: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/processors/status/eventing_scale_up_processor_test.go#L87).

Please use existing tooling for your unittests.

piontec:
I created this PR to update RBAC YAML in kubernetes: https://github.com/kubernetes/kubernetes/pull/75814

Issue #1812, opened 2019-03-22
Admission controller should set limit in case of LimitRange with max ratio


safanaj:
In case VPA is setting requests to lower values and there is deployed a LimitRange with MaxLimitRequestRatio the pod could be not scheduled because it is not respecting the limts/requests ratio for some resource.

Admission controller should be aware of limit ranges and if needed explicitly decrease limits to respect the max ratio.

bskiba:
Thank you for filing this. I agree this would be a good thing for VPA to check. 
One thing I would modify is I think with the current implementation, where VPA doesn't change limits at all, we would want to modify the request to not violate MaxLimitRequestRatio instead of the limit.

@kgolab This is relevant to discussion on VPA controlling limits, Kubernetes offers a way to express the maximum desired request/limit ratio.

safanaj:
I think that in VPA objects the user can set minAllowed to keep the ratio respected, but in scenario where some pods are really wasting requests VPA should be able to take care of limits (optionally).

I think a scenatio where k8s is offered as PaaS and who manage the cluster have no control over the workloads definition deployed, when using LimitRanger with ratio,default request and limits we can find out that that "static" approach is not fitting well to get high utilization.

Optionally allow VPA to set limits would be helpful to avoid users to implement another webhook that take care of this situation.

kgolab:
Looking at LimitRange spec I think we need one more change in VPA: LimitRange min & max should be treated similar to VPA minAllowed & maxAllowed - the question is whether it should be recommender or admission controller which takes care of it. Personally I'd go for the latter so we keep recommender unconstrained by configuration outside of VPA objects.

OTOH, being used to Unix and its "small utils + pipeline" methodology, I also like an approach of having a separate webhook which takes care of enforcing LimitRange *after* VPA makes its (current) changes. If we went this route it could take care of MaxLimitRequestRatio too.

bskiba:
LimitRange is also an admission webhook, so it could be configured to run after the VPA one, but it is currently only a validating webhook - it doesn't change the spec to match constraints, just rejects it if incorrect.

kgolab:
What I meant (and what I think @safanaj wrote) was that there would have to be a new mutating webhook for applying LimitRange before it gets enforced. And I actually like the idea of having it instead of adding functionality to VPA's admission controller (not necessarily prefer it).

bskiba:
/close
fixed by @jbartosik

PR #1813, opened 2019-03-22
Ac take care of limit range

safanaj:
https://github.com/kubernetes/autoscaler/issues/1812

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/logic/limitrange_checker.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/logic/limitrange_checker.go
@@ -1,0 +1,249 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package logic
+
+import (
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/informers"
+
+	v1_listers "k8s.io/client-go/listers/core/v1"
+)
+
+// LimitsHints provides hinted limits that respect limit range max ratio
+type LimitsHints interface {
+	IsNil() bool
+	RequestsExceedsRatio(indexOfContainer int, resourceName v1.ResourceName) bool
+	HintedLimit(indexOfContainer int, resourceName v1.ResourceName) resource.Quantity
+}
+
+// LimitRangeHints implements LimitsHints interface
+type LimitRangeHints struct {
+	requestsExceedsRatio  []map[v1.ResourceName]bool
+	limitsRespectingRatio []v1.ResourceList
+}
+
+var _ LimitsHints = &LimitRangeHints{}
+
+// LimitsChecker checks for LimitRange and if container needs limits to be set
+type LimitsChecker interface {
+	NeedsLimits(*v1.Pod, []ContainerResources) LimitsHints
+}
+
+// IsNil return true if there is no hints to set limits
+func (lrh *LimitRangeHints) IsNil() bool {
+	return lrh == (*LimitRangeHints)(nil)
+}
+
+// RequestsExceedsRatio return true if limits have to be set to respect limit range max ratio
+func (lrh *LimitRangeHints) RequestsExceedsRatio(indexOfContainer int, resourceName v1.ResourceName) bool {
+	if !lrh.IsNil() {
+		yes, ok := lrh.requestsExceedsRatio[indexOfContainer][resourceName]
+		return ok && yes
+	}
+	return false
+}
+
+// HintedLimit return the limit Quantity that respect the limit range max ration
+func (lrh *LimitRangeHints) HintedLimit(indexOfContainer int, resourceName v1.ResourceName) resource.Quantity {
+	if !lrh.IsNil() {
+		limit, ok := lrh.limitsRespectingRatio[indexOfContainer][resourceName]
+		if ok {
+			return limit
+		}
+		return resource.Quantity{}
+	}
+	return resource.Quantity{}
+}
+
+type neverNeedsLimitsChecker struct{}
+
+var _ LimitsChecker = &neverNeedsLimitsChecker{}
+
+func (lc *neverNeedsLimitsChecker) NeedsLimits(pod *v1.Pod, containersResources []ContainerResources) LimitsHints {
+	return LimitsHints((*LimitRangeHints)(nil))
+}
+
+type limitsChecker struct {
+	limitrangeLister v1_listers.LimitRangeLister
+}
+
+var _ LimitsChecker = &limitsChecker{}
+
+// NewLimitsChecker creates a LimitsChecker
+func NewLimitsChecker(f informers.SharedInformerFactory) LimitsChecker {
+	if f != nil {
+		limitrangeLister := f.Core().V1().LimitRanges().Lister()
+		stopCh := make(chan struct{})
+		f.Start(stopCh)
+		for _, ok := range f.WaitForCacheSync(stopCh) {
+			if !ok {
+				if ok := f.Core().V1().LimitRanges().Informer().HasSynced(); !ok {
+					return &neverNeedsLimitsChecker{}
+				}
+			}
+		}
+		return &limitsChecker{limitrangeLister}
+	}
+	return &neverNeedsLimitsChecker{}
+}
+
+type interestingData struct {
+	MaxLimitRequestRatio v1.ResourceList
+	Default              v1.ResourceList
+}
+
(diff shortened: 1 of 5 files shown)


bskiba:
Thank you for this PR. There is good work here. We

vertical-pod-autoscaler/pkg/admission-controller/logic/limitrange_checker.go
```diff
@@ -0,0 +1,251 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
bskiba: 2019

vertical-pod-autoscaler/pkg/admission-controller/logic/limitrange_checker.go
```diff
@@ -0,0 +1,251 @@
...
+				}
+			}
+		}
+		return &limitsChecker{limitrangeLister}
+	}
+	return &neverNeedsLimitsChecker{}
+}
+
+type interestingData struct {
+	// Min                  v1.ResourceList
```
bskiba: Should this be part of this PR?

vertical-pod-autoscaler/pkg/admission-controller/logic/limitrange_checker.go
```diff
@@ -0,0 +1,251 @@
...
+	return LimitsHints((*LimitRangeHints)(nil))
+}
+
+type limitsChecker struct {
+	limitrangeLister v1_listers.LimitRangeLister
+}
+
+var _ LimitsChecker = &limitsChecker{}
+
+// NewLimitsChecker create a LimitsChecker
```
bskiba: nit: creates

vertical-pod-autoscaler/pkg/admission-controller/logic/limitrange_checker.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	return resource.Quantity{}
+}
+
+type neverNeedsLimitsChecker struct{}
+
+var _ LimitsChecker = &neverNeedsLimitsChecker{}
+
+func (lc *neverNeedsLimitsChecker) NeedsLimits(pod *v1.Pod, containersResources []ContainerResources) LimitsHints {
+	return LimitsHints((*LimitRangeHints)(nil))
```
bskiba: It might be simpler if we returned an object with two empty lists instead of nil.

vertical-pod-autoscaler/pkg/recommender/model/vpa.go
```diff
@@ -121,7 +121,7 @@ func (vpa *Vpa) UseAggregationIfMatching(aggregationKey AggregateStateKey, aggre
 	}
 }
 
-// UsesAggregation returns true iff an aggregation with the given key contributes to the VPA.
+// UsesAggregation returns true if an aggregation with the given key contributes to the VPA.
```
bskiba: This is not a typo - iff means if and only if

vertical-pod-autoscaler/pkg/admission-controller/logic/limitrange_checker.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+type limitsChecker struct {
+	limitrangeLister v1_listers.LimitRangeLister
+}
+
+var _ LimitsChecker = &limitsChecker{}
+
+// NewLimitsChecker create a LimitsChecker
+func NewLimitsChecker(i interface{}) LimitsChecker {
```
bskiba: Can we have a SharedInformerFactory passed here?

vertical-pod-autoscaler/pkg/admission-controller/main.go
```diff
@@ -75,11 +78,14 @@ func main() {
 	vpaLister := vpa_api_util.NewAllVpasLister(vpaClient, make(chan struct{}))
 	kubeClient := kube_client.NewForConfigOrDie(config)
 	factory := informers.NewSharedInformerFactory(kubeClient, defaultResyncPeriod)
+	if *allowToAdjustLimits {
+		factoryForLimitsChecker = factory
```
bskiba: I'd rather explicitly pass the no-op limitChecker when *allowToAdjustLimits = false

vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
```diff
@@ -104,6 +116,19 @@ func (s *AdmissionServer) getPatchesForPodResourceRequest(raw []byte, namespace
 				Path:  fmt.Sprintf("/spec/containers/%d/resources/requests/%s", i, resource),
 				Value: request.String()})
 			annotations = append(annotations, fmt.Sprintf("%s request", resource))
+
+			// Set limit if needed
+			if limitsHints.RequestsExceedsRatio(i, resource) {
+				// we need just to take care of max ratio
```
bskiba: Why?

vertical-pod-autoscaler/pkg/admission-controller/logic/limitrange_checker.go
```diff
@@ -0,0 +1,251 @@
...
+// LimitsHints provide hinted limits that respect limit range max ratio
+type LimitsHints interface {
+	IsNil() bool
+	RequestsExceedsRatio(indexOfContainer int, resourceName v1.ResourceName) bool
+	HintedLimit(indexOfContainer int, resourceName v1.ResourceName) resource.Quantity
+}
+
+// LimitRangeHints implements LimitsHints interface
+type LimitRangeHints struct {
+	requestsExceedsRatio  []map[v1.ResourceName]bool
```
bskiba: since v1.ResourceList is already a map, can we store proposedLimits map[container_name]proposed_limit_changes
Then:

requestExceedsRatio(container_name, resourceName) bool {
  _, found := proposedLimits[container_name][resourceName]
  return found
}

vertical-pod-autoscaler/pkg/admission-controller/logic/limitrange_checker.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+	}
+}
+
+func (lc *limitsChecker) getLimitRangeItem(pod *v1.Pod) (ret *v1.LimitRangeItem) {
+	ret = nil
+	limitranges, err := lc.limitrangeLister.
+		LimitRanges(pod.GetNamespace()).
+		List(labels.Everything())
+	if err == nil {
```
bskiba: I'd rather fail fast:
if err != nil {
  return nil
}

vertical-pod-autoscaler/pkg/admission-controller/logic/limitrange_checker.go
```diff
@@ -0,0 +1,251 @@
...
+			for _, lri := range lr.Spec.Limits {
+				if lri.Type != v1.LimitTypeContainer && lri.Type != v1.LimitTypePod {
+					continue
+				}
+				if lri.MaxLimitRequestRatio == nil &&
+					lri.Default == nil {
+					continue
+				}
+				foundInterstingData = true
+				id.parse(&lri)
```
bskiba: Is it possible to have multiple conflicting MaxLimitRequestRatio per namespace?

vertical-pod-autoscaler/pkg/admission-controller/logic/limitrange_checker.go
bskiba: Also, it would be good to log the error.

vertical-pod-autoscaler/pkg/admission-controller/logic/limitrange_checker_test.go
```diff
@@ -0,0 +1,236 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
bskiba: nit: 2019

bskiba:
Thank you very much for this PR, it is needed functionality for VPA.

I left some initial comments. My main concern is why do we only care about MaxRatio? I assume we can also easily violate the MinRatio constraint.

safanaj:
Thanks for the comments, I will take these as soon as I have time

safanaj:
Hi @bskiba I am sorry for the delay but was a very busy month for me.
I re-pushed  the branch for the PR rebased on master. I addressed several of you comments. For other I am not sure.

> My main concern is why do we only care about MaxRatio? I assume we can also easily violate the MinRatio constraint.

The min ratio is 1:1 and this would make limits equals to requests would make a pod in Guaranteed (QoS). In [LimitRange API](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#limitrangeitem-v1-core) there is no MinRatio property.

> Is it possible to have multiple items that have non nil MaxLimitRequestRatio per namespace?

Yes it is possible and in this case [all ratio have to be respected](https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/admission/limitranger/admission.go#L106) so I get the minimum one. For the Default, to consider the limit quantity in case on the container there is not an explicit limit, is not so clear (and I am not sure if my code is correct) because looks like that the LimitRanger set the default limit from [the first LimitRange it gets](https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/admission/limitranger/admission.go#L241) the problem here is that the query using the client-go could return a different order compared to what [LimitRanger get on the server-side](https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/admission/limitranger/admission.go#L149) . I didn't verify this, btw I think that is quite uncommon to have multiple LimitRange for the same resources in a namespace.

bskiba:
Thanks for the changes and the explanation. It was my oversight that I din't check properly that there is no corresponding minLimit, apologies. 
I need some more time to review.

bskiba:
@safanaj I believe the other PRs (based on this one) fix the original issue. I'm closing this, let me know if there is something still to be adressed.
/close

PR #1815, opened 2019-03-24
Adding link to Sig-Autoscaling for better reference.

Rajat-0:


MaciekPytel:
You're right, that's a better link. Thanks!

/lgtm
/approve

BTW @mwielgus - maybe we should update the README in community repo? It says the meetings are biweekly at 2pm UTC. I think it would be more true to say they're weekly at 4pm CET?

PR #1816, opened 2019-03-24
Adding steps to register VPA-Admission-Controller on specific path.

Rajat-0:


vertical-pod-autoscaler/pkg/admission-controller/README.md
```diff
@@ -28,8 +28,10 @@ up the changes: ```sudo systemctl restart kubelet.service```
 1. Create the pod:
    `kubectl create -f ../deploy/admission-controller-deployment.yaml`.
    The first thing this will do is it will register itself with the apiserver as
-   an Webhook Admission Controller and start changing resource requirements
-   for pods on their creation & updates.
+   Webhook Admission Controller and start changing resource requirements
+   for pods on their creation & updates, you can specify a path for it to registered
```
bskiba: Can you break the sentence after creation & updates?  So that it will be "creation & updates. You can specify..."

bskiba: s/registered/register on/

Rajat-0:
@bskiba  Done

bskiba:
Thank you! 
/lgtm
/approve

Issue #1501, opened 2018-12-11
vertical-pod-autoscaler - the way vpa recommender uses prometheus history cpu data


patrickshan:
when using prometheus as history data provider, how does VPA recommender calculate CPU recoomendation ? 

It seems currently it uses `container_cpu_usage_seconds_total` which is a counter but it doesn't seem to do any rate calculation based on it ? is this correct ? while for memory it uses a gauge metric `container_memory_usage_bytes` which makes sense

patrickshan:
thanks @bskiba . I'd love to contribute back but I don't have a working setup with Prometheus at the moment. will do that when I have one.

yashbhutwala:
I think this is fixed via #1824 and then fixed further by #3009 🔨

PR #1824, opened 2019-03-25
VPA recommender: use rate() to query CPU historical usage from promet…

safanaj:
…heus

This is a proposed fix for #1501


bskiba:
Thanks! 
/lgtm
/approve

PR #1825, opened 2019-03-25
Use debian-base-amd64:v1.0.0

losipiuk:


losipiuk:
/hold

Seems like the 1.0.0 is still in staging gcr only

PR #1843, opened 2019-03-29
Priority expander relies now on listers instead of direct watch api

piontec:
This is a rework of this PR: https://github.com/kubernetes/autoscaler/pull/1801
The changes are as follows:
* it uses lister now
* unittests rely on already existing tooling
* app is never crashed, just error is logged

Please review @MaciekPytel


cluster-autoscaler/expander/factory/expander_factory.go
```diff
@@ -48,12 +55,30 @@ func ExpanderStrategyFromString(expanderFlag string, cloudProvider cloudprovider
...
+		stopChannel := make(chan struct{})
+		lister := getPriorityExpandersConfigMapLister(kubeClient, stopChannel, configNamespace)
+		return priority.NewStrategy(lister.ConfigMaps(configNamespace), autoscalingKubeClients.Recorder)
 	}
 	return nil, errors.NewAutoscalerError(errors.InternalError, "Expander %s not supported", expanderFlag)
 }
+
+func getPriorityExpandersConfigMapLister(kubeClient kube_client.Interface, stopChannel <-chan struct{},
+	namespace string) v1lister.ConfigMapLister {
+	// FIXME: how to be sure the reflector completed at least one run and switch to the line below?
```
piontec: @MaciekPytel I have a problem here: how to make sure the reflector completed at least 1 sync loop before I access the lister? I want to validate the configmap before starting.
If this can't be done easily, I probably have to give up on startup time validation and assume reflector will be ready when first call for `Best` comes. But even that smells like race condition.

cluster-autoscaler/expander/factory/expander_factory.go
```diff
@@ -48,12 +55,30 @@ func ExpanderStrategyFromString(expanderFlag string, cloudProvider cloudprovider
 			price.NewSimplePreferredNodeProvider(autoscalingKubeClients.AllNodeLister()),
 			price.SimpleNodeUnfitness), nil
 	case expander.PriorityBasedExpanderName:
-		maps := kubeClient.CoreV1().ConfigMaps(configNamespace)
-		initialPriorities, priorityChangesChan, err := priority.InitPriorityConfigMap(maps, configNamespace)
-		if err != nil {
-			return nil, errors.ToAutoscalerError(errors.InternalError, err)
-		}
-		return priority.NewStrategy(initialPriorities, priorityChangesChan, autoscalingKubeClients.LogRecorder)
+		// TODO: how to get proper termination info? It seems other listers do the same here
```
piontec: Other listers seems to be doing the same - please make sure it's ok to leave the termination channel like that

piontec:
And one important question here: is there any comprehensive doc arund here about how to use the `client-go` lib properly? Either I can't find it or it's just a bunch of blog entries of kind "do this and that". The API is pretty big, I was running around different packages like crazy trying to figure out what's what and how to use it.

MaciekPytel:
> is there any comprehensive doc arund here about how to use the client-go lib properly?

I'm not aware of any such doc. I'm not an expert on client-go though, it may be worth asking directly on that repo or in kubernetes slack sig-apimachinery channel. As far as I know Kubernetes documentation is mostly limited to how to use it. Documentation for developing is minimal and the entry point for contributing is quite high.

cluster-autoscaler/expander/factory/expander_factory.go
MaciekPytel: CA loop generally takes a few seconds before it gets to expander and other listers must have already synced for a scale-up to have any chance of happening. Since you have a fallback strategy and the race is quite unlikely I'm ok to merge this without logic to explicitly make sure lister has initialized. 
If you think you really need to make sure: I think the LastSyncResourceVersion() you do below is the best way unfortunately (at least the best way I know of). I don't like creating lister in here though. This file is really not the place for all those client-go internals. If you really want to sync, I'd rather you created your own configmap lister interface in utils/kubernetes/listers.go (like we do for pod listers) and expose a HasSynced() method.

piontec:
OK, I didn't want to leave the code there, just make sure first how to solve this. In this case I'm in favor of not running configmap validation at startup and leave that until the first `BestOption()` call happens - then any potential problems will go to logs and CM's events. Let me know if you're OK with that and I can quickly change it.

piontec:
ping @MaciekPytel

MaciekPytel:
Sorry for delay. I'm ok with what you proposed in https://github.com/kubernetes/autoscaler/pull/1843#issuecomment-479765720.

piontec:
@MaciekPytel Done, on-init config check is removed. Please review again.

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -360,3 +360,14 @@ func NewStatefulSetLister(kubeClient client.Interface, stopchannel <-chan struct
 	go reflector.Run(stopchannel)
 	return lister
 }
+
+// NewConfigMapListerForNamespace builds a configmap lister for the passed namespace (including all).
+func NewConfigMapListerForNamespace(kubeClient client.Interface, stopchannel <-chan struct{},
+	namespace string) v1lister.ConfigMapLister {
+	listWatcher := cache.NewListWatchFromClient(kubeClient.AppsV1().RESTClient(), "configmaps", namespace, fields.Everything())
```
MaciekPytel: Configmap is in core/v1, not apps/v1. Does that work for you? Seems like it shouldn't.

piontec: Yeah, my bad, I ran tests actually before removing the test code here. I was sure it was exactly the same as in Listers - it wasn't. Fixing, and I'll re-run tests once more. Thanks for catching!

piontec: @MaciekPytel please check now

PR #1847, opened 2019-04-01
Only refresh instances cache for one mig on cache miss in GetMigForInstance

losipiuk:


cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -165,71 +160,110 @@ func (gc *GceCache) updateMigBasename(ref GceRef, basename string) {
...
 	gc.cacheMutex.Lock()
 	defer gc.cacheMutex.Unlock()
 
-	if mig, found := gc.instancesCache[*instance]; found {
-		return mig, nil
+	if migRef, found := gc.instanceRefToMigRef[instanceRef]; found {
+		mig, found := gc.migs[migRef]
+		if !found {
+			// It is fine that mig is not found in gc.migs map even if we have entry in gc.instanceRefToMigRef
+			// The mig could have bin removed rom gc.migs if we (for example) disabled autoscaling for that or the node pool
```
aleksandra-malinowska: s/bin/been
s/that/it

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -165,71 +160,110 @@ func (gc *GceCache) updateMigBasename(ref GceRef, basename string) {
...
-			newInstancesCache[*gceRef] = mig
-		}
 	}
-
-	gc.instancesCache = newInstancesCache
 	return nil
 }
 
+// InvalidateInstanceCache invalidates whole InstanceRef->MigRef cache
+func (gc *GceCache) InvalidateInstanceCache() {
```
aleksandra-malinowska: Do we use it anywhere?

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -165,71 +160,110 @@ func (gc *GceCache) updateMigBasename(ref GceRef, basename string) {
...
 // matching prefix, but instance doesn't belong to it.
-func (gc *GceCache) GetMigForInstance(instance *GceRef) (Mig, error) {
+func (gc *GceCache) GetMigForInstance(instanceRef GceRef) (Mig, error) {
 	gc.cacheMutex.Lock()
 	defer gc.cacheMutex.Unlock()
 
-	if mig, found := gc.instancesCache[*instance]; found {
-		return mig, nil
+	if migRef, found := gc.instanceRefToMigRef[instanceRef]; found {
+		mig, found := gc.migs[migRef]
```
aleksandra-malinowska: Previously list of migs was obtained here by `for _, mig := range gc.GetMigs()  {`. Can we still use GetMigs() here, to copy under migs lock?

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -165,71 +160,110 @@ func (gc *GceCache) updateMigBasename(ref GceRef, basename string) {
...
+	gc.cacheMutex.Lock()
+	defer gc.cacheMutex.Unlock()
+
+	return gc.regenerateInstancesCacheNoLock()
+}
 
-		instances, err := gc.GceService.FetchMigInstances(mig.GceRef())
+func (gc *GceCache) regenerateInstancesCacheNoLock() error {
+	gc.invalidateInstanceCacheNoLock()
+	for migRef := range gc.migs {
```
aleksandra-malinowska: Can we use GetMigs here, to copy under lock?

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -165,71 +160,110 @@ func (gc *GceCache) updateMigBasename(ref GceRef, basename string) {
...
+				return nil, fmt.Errorf("error while looking for MIG for instance %+v, error: %v", instanceRef, err)
 			}
-			if mig, found := gc.instancesCache[*instance]; found {
-				return mig, nil
+
+			migRef, found := gc.instanceRefToMigRef[instanceRef]
+			if !found {
+				return nil, fmt.Errorf("instance %+v belongs to unknown mig %+v", instanceRef, migRef)
+			}
+			mig, found := gc.migs[migRef]
```
aleksandra-malinowska: Can we add some method that will retrieve mig under migsMutex lock?

cluster-autoscaler/cloudprovider/gce/cache.go
losipiuk: Only the `NoLock` version. I will remove the public one.

cluster-autoscaler/cloudprovider/gce/cache.go
losipiuk: Yeah - sure. I did not notice we have two mutexes here.
Side question: is that really needed? Was it driven by actual perfromance problemes - or it was done so "just in case"

losipiuk: gc.getMig

cluster-autoscaler/cloudprovider/gce/cache.go
losipiuk: Done - seem comments above

cluster-autoscaler/cloudprovider/gce/cache.go
losipiuk: I added getMigRefs methods and use it here.

losipiuk:
@aleksandra-malinowska @MaciekPytel comments addressed.

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -165,71 +172,110 @@ func (gc *GceCache) updateMigBasename(ref GceRef, basename string) {
...
 	return nil, nil
 }
 
-// RegenerateInstancesCache triggers instances cache regeneration under lock.
-func (gc *GceCache) RegenerateInstancesCache() error {
+func (gc *GceCache) getMig(migRef GceRef) (*MigInformation, bool) {
+	gc.migsMutex.Lock()
+	defer gc.migsMutex.Unlock()
+	mig, found := gc.migs[migRef]
+	return mig, found
```
aleksandra-malinowska: nit: can it return a copy of MigInformation (similarly to GetMigs())?

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -165,71 +172,110 @@ func (gc *GceCache) updateMigBasename(ref GceRef, basename string) {
...
+func (gc *GceCache) RegenerateInstancesCache() error {
+	gc.cacheMutex.Lock()
+	defer gc.cacheMutex.Unlock()
+
+	return gc.regenerateInstancesCacheNoLock()
+}
 
-		instances, err := gc.GceService.FetchMigInstances(mig.GceRef())
+func (gc *GceCache) regenerateInstancesCacheNoLock() error {
+	gc.invalidateInstanceCacheNoLock()
```
aleksandra-malinowska: nit: this is the only use of this method, right? If so, can we inline it?

cluster-autoscaler/cloudprovider/gce/cache.go.rej
```diff
@@ -0,0 +1,86 @@
+diff a/cluster-autoscaler/cloudprovider/gce/cache.go b/cluster-autoscaler/cloudprovider/gce/cache.go	(rejected hunks)
```
aleksandra-malinowska: accidental git add?

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -165,71 +172,110 @@ func (gc *GceCache) updateMigBasename(ref GceRef, basename string) {
...
+	for _, migRef := range gc.getMigRefs() {
+		err := gc.regenerateInstanceCacheForMigNoLock(migRef)
 		if err != nil {
```
aleksandra-malinowska: nit: if err := ...; err != nil { return err }

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -165,71 +172,110 @@ func (gc *GceCache) updateMigBasename(ref GceRef, basename string) {
...
+	}
+	return nil
+}
+
+// RegenerateInstancesCache triggers instances cache regeneration under lock.
+func (gc *GceCache) RegenerateInstancesCache() error {
+	gc.cacheMutex.Lock()
+	defer gc.cacheMutex.Unlock()
+
+	return gc.regenerateInstancesCacheNoLock()
```
aleksandra-malinowska: nit:  this is the only use of this method, why not inline? Do we plan to call it from another place where we lock on cacheMutex?

cluster-autoscaler/cloudprovider/gce/cache.go
losipiuk: done

cluster-autoscaler/cloudprovider/gce/cache.go.rej
losipiuk: Yeah.

cluster-autoscaler/cloudprovider/gce/cache.go
losipiuk: Less readable to me

cluster-autoscaler/cloudprovider/gce/cache.go
losipiuk: less readable to me

cluster-autoscaler/cloudprovider/gce/cache.go
losipiuk: Makes sense.

cluster-autoscaler/cloudprovider/gce/cache.go
losipiuk: I just followed common pattern. I guess boilerplate of acquiring lock is not that big so it makes method less readable if actual logic is inlined to method obtaining lock.

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -165,68 +172,99 @@ func (gc *GceCache) updateMigBasename(ref GceRef, basename string) {
...
 	defer gc.cacheMutex.Unlock()
 
-	if mig, found := gc.instancesCache[*instance]; found {
-		return mig, nil
+	if migRef, found := gc.instanceRefToMigRef[instanceRef]; found {
+		mig, found := gc.getMig(migRef)
+		if !found {
+			// It is fine that mig is not found in gc.migs map even if we have entry in gc.instanceRefToMigRef
+			// The mig could have been removed from gc.migs if we (for example) disabled autoscaling for it or the node pool
+			// was just deleted. Yet we are not removing entries from gc.instanceRefToMigRef in such case due to performance reasons.
```
MaciekPytel: What performance reasons are those? Feels like you can easily clean up the map in UnregisterMig. Sure, you'd need to iterate over the whole map, but that should still be negligible.

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -165,68 +172,99 @@ func (gc *GceCache) updateMigBasename(ref GceRef, basename string) {
...
+			strings.HasPrefix(instanceRef.Name, mig.Basename) {
+			if err := gc.regenerateInstanceCacheForMigNoLock(mig.Config.GceRef()); err != nil {
+				return nil, fmt.Errorf("error while looking for MIG for instance %+v, error: %v", instanceRef, err)
 			}
-			if mig, found := gc.instancesCache[*instance]; found {
-				return mig, nil
+
+			migRef, found := gc.instanceRefToMigRef[instanceRef]
+			if !found {
+				return nil, fmt.Errorf("instance %+v belongs to unknown mig %+v", instanceRef, migRef)
```
MaciekPytel: Seems like migRef will be nil in here, no?

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -165,68 +172,99 @@ func (gc *GceCache) updateMigBasename(ref GceRef, basename string) {
...
-// internal method - should only be called after locking on cacheMutex.
-func (gc *GceCache) regenerateCache() error {
-	newInstancesCache := make(map[GceRef]Mig)
+func (gc *GceCache) regenerateInstanceCacheForMigNoLock(migRef GceRef) error {
+	klog.V(4).Infof("Regenerating MIG information for %s", migRef.String())
+
+	// Here we are on purpose not deleting old entries pointing to migRef from gc.instanceRefToMig.
+	// They are no harm as same GceRef should not be used for new instance and instance should not move from one Mig to
+	// the other.
+	// And garbage-collection will be enforces as whole cache is invalidated hourly.
```
MaciekPytel: This is correct, but it also feels like premature optimization. You're about to do an API call, and the size of the map is limited, so the cost of iteration is probably negligible.
Feel free to ignore this comment if you disagree, as I said you're right and it's just a stylistic difference.

MaciekPytel:
/hold
/approve
/lgtm

Some minor comments - feel free to ignore, except for the nil pointer in log one.

PR #1851, opened 2019-04-02
Add steps to use AWS Credentials for on premise case

### LINKED ISSUE #1798: on premise scaling to aws
How do i provide credentials to the aws cloudprovider so i can run from on premise bare metal and scale out to aws when it there are not enough resources on the node? I believe the aws cloudprovider is grabbing creds to create the new aws nodes.

Jeffwan:
Resolve #1798 

User runs cluster autoscaler in on premise cluster and cluster autoscaler uses AWS cloud provider to scale up/down.

Rather than using IAM policy for autoscaler to get access to AWS services, they should use AWS credentials in Env to get AWS SDK authenticated and authorized.


tikajon:
Good

PR #1855, opened 2019-04-03
Update MIG target size cache via LIST api call

losipiuk:


losipiuk:
Please review just second commit. The first one is just squashed #1847 PR. This one is based on that one because same files are touched. And otherwise more extra work related to conflict resolution would have to be done.

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -47,6 +47,7 @@ type AutoscalingGceClient interface {
 	// reading resources
 	FetchMachineType(zone, machineType string) (*gce.MachineType, error)
 	FetchMachineTypes(zone string) ([]*gce.MachineType, error)
+	ListMigs(zone string) ([]*gce.InstanceGroupManager, error)
```
aleksandra-malinowska: 1) I don't like returning *gce.InstanceGroupManager here. The reason this object (autoscalingGceClientV1) exists is to hide as much as possible of stuff from GCE API, so that the rest of the code doesn't need to know (or care) what kind, and how many requests are made.
I realize this is already not the case for MachineType and InstanceTemplate (because processing templates is messy), but here we only need target sizes. 
2) Following the pattern from autodiscovery (FetchMigsWithName) and getting only MIGs we're interested in, I'd prefer if this method accepted []GceRef of configured MIGs and filtered them inside. 
3) Lastly, please follow the 'Fetch' naming convention in this file. It may not be perfect but at least it's consistent;)

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -293,15 +295,79 @@ func (gc *GceCache) GetResourceLimiter() (*cloudprovider.ResourceLimiter, error)
...
+	size, found := gc.migTargetSizeCache[migRef]
+	if !found {
+		return 0, fmt.Errorf("Could not get target size for mig %v", migRef.String())
+	}
+
+	// we are good
+	return size, nil
+}
+
+func (gc *GceCache) fillInMigTargetSizeCache() error {
```
aleksandra-malinowska: migsMutex

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -293,15 +295,79 @@ func (gc *GceCache) GetResourceLimiter() (*cloudprovider.ResourceLimiter, error)
...
+		}
 	}
-	return size, found
+	gc.migTargetSizeCache = newMigTargetSizeCache
+	return nil
+}
+
+func (gc *GceCache) listAllZonesForMigs() map[string]bool {
+	zones := map[string]bool{}
+	for migRef := range gc.migs {
```
aleksandra-malinowska: GetMigs()

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -206,22 +206,19 @@ func (m *gceManagerImpl) registerMig(mig Mig) bool {
...
-	if migSize, found := m.cache.GetMigTargetSize(mig.GceRef()); found {
-		return migSize, nil
-	}
-	targetSize, err := m.GceService.FetchMigTargetSize(mig.GceRef())
-	if err != nil {
-		return -1, err
-	}
-	m.cache.SetMigTargetSize(mig.GceRef(), targetSize)
-	return targetSize, nil
+	return m.cache.GetMigTargetSize(mig.GceRef())
```
aleksandra-malinowska: I don't really like moving this logic inside the cache itself. Sth called cache.GetMigTargetSize doesn't imply in any way that the cache will self-refresh, or that the negative result isn't just a cache miss (but that the actual check was performed). I'd rather keep the cache simple, and use it as cache.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -206,22 +206,19 @@ func (m *gceManagerImpl) registerMig(mig Mig) bool {
...
 // SetMigSize sets MIG size.
 func (m *gceManagerImpl) SetMigSize(mig Mig, size int64) error {
 	klog.V(0).Infof("Setting mig size %s to %d", mig.Id(), size)
 	m.cache.InvalidateTargetSizeCacheForMig(mig.GceRef())
-	return m.GceService.ResizeMig(mig.GceRef(), size)
+	err := m.GceService.ResizeMig(mig.GceRef(), size)
+	if err != nil {
+		return err
+	}
+	m.cache.SetMigTargetSize(mig.GceRef(), size)
```
aleksandra-malinowska: Do we need this, considering we invalidate this cache just before resizing?

cluster-autoscaler/cloudprovider/gce/gce_manager.go.rej
```diff
@@ -0,0 +1,31 @@
+diff a/cluster-autoscaler/cloudprovider/gce/gce_manager.go b/cluster-autoscaler/cloudprovider/gce/gce_manager.go	(rejected hunks)
+@@ -206,22 +206,19 @@ func (m *gceManagerImpl) registerMig(mig Mig) bool {
```
aleksandra-malinowska: git add?

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
losipiuk: > 1. I don't like returning *gce.InstanceGroupManager here. The reason this object (autoscalingGceClientV1) exists is to hide as much as possible of stuff from GCE API, so that the rest of the code doesn't need to know (or care) what kind, and how many requests are made.
> I realize this is already not the case for MachineType and InstanceTemplate (because processing templates is messy), but here we only need target sizes.

I will pass on this one. As you noticed we are not doing it for other entitiies anyway. And it does not seem that creating a wrapper struct around targetSize will do us any good right now.

> 2. Following the pattern from autodiscovery (FetchMigsWithName) and getting only MIGs we're interested in, I'd prefer if this method accepted []GceRef of configured MIGs and filtered them inside.

I do not see any benefit of that (given you cannot filter out migs on the server side) . And such interface would be more cumbersome to use.  I will skip it for now.

> 3. Lastly, please follow the 'Fetch' naming convention in this file. It may not be perfect but at least it's consistent;)

Done

cluster-autoscaler/cloudprovider/gce/cache.go
losipiuk: N/A

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -293,15 +295,79 @@ func (gc *GceCache) GetResourceLimiter() (*cloudprovider.ResourceLimiter, error)
...
+		zones[migRef.Zone] = true
+	}
+	return zones
+}
+
+func (gc *GceCache) refreshMigTargetSizeCacheForMig(migRef GceRef) error {
```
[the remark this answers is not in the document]
losipiuk: N/A

cluster-autoscaler/cloudprovider/gce/gce_manager.go
losipiuk: Moved out of cache

cluster-autoscaler/cloudprovider/gce/gce_manager.go
losipiuk: I guess it to handle the case when `ResizeMig` succeeded (changed the mig size on the server side) but returned error due to some network issues later on.
We do not want to have stale data in cache in that situation.

cluster-autoscaler/cloudprovider/gce/mig_target_sizes_provider.go
```diff
@@ -0,0 +1,119 @@
...
+
+		for _, zoneMig := range zoneMigs {
+			zoneMigRef := GceRef{
+				c.projectId,
+				zone,
+				zoneMig.Name,
+			}
+
+			if registeredMigRefs[zoneMigRef] {
+				newMigTargetSizeCache[zoneMigRef] = zoneMig.TargetSize
```
MaciekPytel: This feels a bit artificial, you could just SetMigTargetSize here. Tho only difference is if listing fails for one zone we would still update zones we listed previously, which seems reasonable to me.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
MaciekPytel: But in that case you would still fetch target size of that mig next time you need it and not return stale data (since you invalidated it before ResizeMig call)?

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -116,6 +117,22 @@ func (client *autoscalingGceClientV1) FetchMachineTypes(zone string) ([]*gce.Mac
 	return machines.Items, nil
 }
 
+func (client *autoscalingGceClientV1) ListMigs(zone string) ([]*gce.InstanceGroupManager, error) {
```
MaciekPytel: Name mismatch.

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
losipiuk: > I do not see any benefit of that (given you cannot filter out migs on the server side) 

I take that back. Actually you can filter out using "ORs" and multiple condition on mig name. I will ask MIGs team what is the expected performance and limitations of that, and possibly make the change so this method takes a list of GCERefs.

cluster-autoscaler/cloudprovider/gce/mig_target_sizes_provider.go
losipiuk: Yeah - yet I still want to return the map from `fillInMigTargetSizeCache` method. So I can check what is the target size of mig without going to cache again (which can be invalidated at this point).
So I would keep it here.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
losipiuk: With invalidate call yes. Without we would return stale value from cache.

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
losipiuk: > I take that back. Actually you can filter out using "ORs" and multiple condition on mig name. I will ask MIGs team what is the expected performance and limitations of that, and possibly make the change so this method takes a list of GCERefs.

I am doing some more testing related to filtering - but based on conversation it seems it would not help us much (in terms of performance) to pass list of refs as arg. Let's merge this one and iterate with more data.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
MaciekPytel: After discussing with @losipiuk - yes it would be safe to skip this, but it's also safe to leave it and it potentially saves us a completely unnecessary API call (note we'll invalidate all target sizes for next loop anyway).

Ps. I think @losipiuk misunderstood original question :)

cluster-autoscaler/cloudprovider/gce/mig_target_sizes_provider.go
MaciekPytel: Ok, I actually missed the fact you're returning the map. In that case it makes sense.

PR #1857, opened 2019-04-04
Document behavior of min_nodes and max_nodes

### LINKED ISSUE #1849: Behavior wrt min/max limits is not publicly documented
[Cluster autoscaler FAQ](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md) does not describe how CA handles min/max limits.

People often expect min/max to be enforced by CA, and that changing the minimum number of nodes alone will trigger a scale-up ([example](https://b.corp.google.com/issues/129746085)). The actual behavior is that CA does not enforce these limits, so if someone manually resizes the cluster beyond those limits, CA will not intervene.

Since this is a common cause of misunderstanding, we need some public documentation on the expected behavior.

Shnatsel:
Initial documentation for the handling of min_nodes and max_nodes parameters. These are currently not publicly documented and have caused some confusion.

Fixes #1849


bskiba:
Thanks!
/lgtm
/approve

Issue #1858, opened 2019-04-04
addon-resizer modifies different attributes than request/limit only


piontec:
We're running addon-resizer with https://github.com/kubernetes/kube-state-metrics, which is the default deployment. Unfortunately, addon-resizer changes other properties than limits/requests on the managed pod:
```
   name: kube-state-metrics
   namespace: kube-system
 spec:
-  progressDeadlineSeconds: 2147483647
+  progressDeadlineSeconds: 600
   replicas: 1
   revisionHistoryLimit: 10
   selector:
@@ -189,6 +189,7 @@
         terminationMessagePath: /dev/termination-log
         terminationMessagePolicy: File
       dnsPolicy: ClusterFirst
+      priorityClassName: system-cluster-critical
       restartPolicy: Always
       schedulerName: default-scheduler
       securityContext: {}
```

SInce the code seems to not update them anywhere, it must be a bug in underlying libs.

piontec:
ping @bskiba , can we close this?

bskiba:
Still haven't released it :disappointed: 
Will try to wrap it up this week

bskiba:
released. Image is: gcr.io/google-containers/addon-resizer:2.3
/close

Thanks @piontec!

Issue #1799, opened 2019-03-17
[addon-resizer] deployment update drops existing nodeAffinity fields


jstangroome:
Seems to be a repeat of issue #188.

Using addon-resizer Docker image: `gcr.io/google-containers/addon-resizer-amd64:2.1`

Original deployment yaml contains:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  # ... truncated ...
spec:
  template:
      # ... truncated ...
      tolerations:
      - key: the-taint
        effect: NoSchedule
      affinity:
        nodeAffinity:
          preferredDuringSchedulingIgnoredDuringExecution:
          - weight: 1
            preference:
              matchExpressions:
              - key: the-taint
                operator: Exists
```

And after addon-resizer logs "Resources are not within the expected limits, updating the deployment", the `apiVersion` has regressed to `extensions/v1beta1` and both the `tolerations` and `affinity` members have been removed.

bskiba:
The fix for this has been released. Image is: gcr.io/google-containers/addon-resizer:2.3
/close

Thanks @piontec!

Issue #1457, opened 2018-11-28
addon resizer changes `optional: true` of volume mounts


williamhuangh:
original yaml
```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: prometheus-deployment
  namespace: monitoring
  annotations:
    k8sdeploy.lyft.net/skip-rollout-monitoring: true
spec:
  replicas: 1
  strategy:
    rollingUpdate:
      maxSurge: 0
      maxUnavailable: 1
    type: RollingUpdate
  selector:
    matchLabels:
      app: prometheus
  template:
    metadata:
      name: prometheus
      labels:
        app: prometheus
        version: ${sha}
    spec:
      serviceAccount: prometheus
      securityContext:
        runAsUser: 0 # This is required to use the EBS volume which is owned by root
      containers:
      - name: addon-resizer
        image: k8s.gcr.io/addon-resizer:1.7
        resources:
          limits:
            cpu: 100m
            memory: 30Mi
          requests:
            cpu: 100m
            memory: 30Mi
        env:
          - name: MY_POD_NAME
            valueFrom:
              fieldRef:
                fieldPath: metadata.name
          - name: MY_POD_NAMESPACE
            valueFrom:
              fieldRef:
                fieldPath: metadata.namespace
        command:
          - /pod_nanny
          - --container=prometheus
          - --cpu=3
          - --extra-cpu=10m
          - --memory=2Gi
          - --extra-memory=100Mi
          - --threshold=5
          - --deployment=prometheus        
      - name: prometheus
        image: quay.io/prometheus/prometheus:v2.3.2
        args:
            - '--config.file=/etc/prometheus/prometheus.yaml'
            - --storage.tsdb.path=/prometheus
            - --web.console.libraries=/etc/prometheus/console_libraries
            - --web.console.templates=/etc/prometheus/consoles
            - --web.enable-lifecycle            
        ports:
        - name: web
          containerPort: 9090
        readinessProbe:
          httpGet:
            path: /-/ready
            port: 9090
          initialDelaySeconds: 30
          timeoutSeconds: 30
        livenessProbe:
          httpGet:
            path: /-/healthy
            port: 9090
          initialDelaySeconds: 30
          timeoutSeconds: 30
        resources:
          limits:
            cpu: 8
            memory: 12000Mi
          requests:
            cpu: 3
            memory: 2000Mi
        env:
        - name: STORAGE_RETENTION
          valueFrom:
            configMapKeyRef:
[...]

PR #1859, opened 2019-04-04
rework addon-resizer to use client-go 

piontec:
The addon-resizer code doesn't modify attributes other than limits/requests, so it seems the bug is in libraries used. This is causing problems like #1858 and probably also #1799 and #1457 .  To solve that, I upgraded libraries to more recent ones. 

Changes include:
* drop `godep` in favor of `go mod` - I was becoming sick trying to do that using `godep`
* code is reworked to use `client-go` (version v8.0.0) and api libs (from 1.11 k8s release)
* go version is updated to 1.12.1
* `Makefile` builds docker image OK

Testing:
* unit tests are passing
* I manually verified that `addon-scaler` works OK running against an already provisioned cluster with `kube-state-metrics` pod.


piontec:
Assuming the switch to `go mod` is OK, we can drop `vendor/` in the repo and fetch dependencies on build time. Let me know if that's ok.

bskiba:
In general, huge, huge thank you for taking the time and the effort!

Seeing core Kubernetes moving to go mod (https://groups.google.com/forum/#!topic/kubernetes-dev/ECejidBCELI) I think it's the right thing to move addon resizer too.

I have one request, can you split the go mod change from the client-go one? It will make it way easier for me to remove the code that actually changed.

piontec:
Crap, I squashed my commits - I can't easily separate them now. But if we remove `vendor/` and rely entirely on the new `go mod`, the changes should be much easier to review - the whole switch will be just in removing `Godeps/` and adding `go.mod`. What do you think?

bskiba:
There are several problems:
- I need to make sure the dependencies haven't changed - we are not pulling in problematic licenses
- I believe we need to have a copy of the license for all vendored dependencies

I can try to review this as a whole but I'll need time. I also need to consult someone around the license thing, I'll get back to you next week.
It's easy to omit things with such a huge change. Example: you have two different go versions specified, one in Makefile (1.12) and another in go.mod (1.11). Is that intentional?

piontec:
Well, dependencies have changed, that's for sure. `client-go` wasn't even used before and it is now. The same with `apimachinery`. The problem with this project was that dependencies were so old and the project so small, that I actually removed all dependencies, started with empty `go.mod` and then I was rewriting the code to use new libraries, adding them to `go.mod` on the way.

As for version mismatch - my mistake, I forgot to switch everything to 1.12. Will fix in a minute.

piontec:
OK, I'll try to rework using `godeps`. Do you have some hints about how to include `client-go` with it the easiest way?

piontec:
@bskiba Done, it's using `godep` now. Can you also have a look at https://github.com/kubernetes/autoscaler/pull/1843, it seems to be stuck a little...

addon-resizer/nanny/kubernetes_client.go
```diff
@@ -20,40 +20,59 @@ import (
...
+		deploymentClient: kubeClient.AppsV1().Deployments(namespace),
 	}
-	return uint64(len(k.nodeStore.List())), nil
+	return result
 }
 
-func (k *kubernetesClient) ContainerResources() (*apiv1.ResourceRequirements, error) {
-	pod, err := k.clientset.CoreClient.Pods(k.namespace).Get(k.pod)
+func (k *kubernetesClient) Stop() {
+	for i := 0; i < 3; i++ {
```
bskiba: The i < 3 seems really error prone - if someone adds another lister, no way they will know thay have to increase it. Can you add a stopChan per lister instead?

addon-resizer/nanny/kubernetes_client.go
```diff
@@ -20,40 +20,59 @@ import (
...
 
-	api "k8s.io/kubernetes/pkg/api"
-	apiv1 "k8s.io/kubernetes/pkg/api/v1"
-	cache "k8s.io/kubernetes/pkg/client/cache"
-	client "k8s.io/kubernetes/pkg/client/clientset_generated/release_1_3"
-	runtime "k8s.io/kubernetes/pkg/runtime"
-	wait "k8s.io/kubernetes/pkg/util/wait"
-	watch "k8s.io/kubernetes/pkg/watch"
+	apps "k8s.io/api/apps/v1"
+	api "k8s.io/api/core/v1"
```
bskiba: nit: s/api/core/

addon-resizer/main.go
```diff
@@ -0,0 +1,175 @@
...
+func buildOutOfClusterConfig() (*rest.Config, error) {
+	kubeconfigPath := os.Getenv("KUBECONFIG")
+	if kubeconfigPath == "" {
+		kubeconfigPath = os.Getenv("HOME") + "/.kube/config"
+	}
+	return clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+}
+
+// GetClientOutOfCluster returns a k8s clientset to the request from outside of cluster
+func GetClientOutOfCluster() kubernetes.Interface {
```
bskiba: nit: GetClientOutOfClusterOrDie() (here and below)

piontec: GetClientOutOfCluster() has a fallback to GetClient(). I'm changing the latter one to "OrDie()"

piontec:
@bskiba please check now

piontec:
Hi @bskiba , do you have any time frame for releasing this?

diclophis:
Also where can we find newer tagged images of this addon-resizer docker image?

bskiba:
@piontec I'll try to release in the next two weeks.
@diclophis All images should be in k8s.gcr.io/addon-resizer, the newest is 2.1

wojtek-t:
@bskiba - note that we're still using 1.8.* releases, as there were some issues with 2.* releases
I can't remember the details, but it's something to keep in mind.

bskiba:
@wojtek-t Thanks! I wasn't planning on changing the default, and I think that there are folks successfully using the 2.X releases so it still makes sense to release. But it's good to note that maybe this fix should go to the 1.8 release as well. @piontec are you interested in porting to 1.8 or is the master branch satisfactory?

PR #1872, opened 2019-04-09
Separate mig-basename caching in GceCache

losipiuk:


cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -381,12 +376,124 @@ func (m *gceManagerImpl) fetchAutoMigs() error {
...
+
+func (gc *GceCache) removeInstancesForMig(migRef GceRef) {
+	for instanceRef, instanceMigRef := range gc.instanceRefToMigRef {
+		if migRef == instanceMigRef {
+			delete(gc.instanceRefToMigRef, instanceRef)
+		}
+	}
+}
+
+func (gc *GceCache) getMig(migRef GceRef) (mig Mig, found bool) {
```
aleksandra-malinowska: why do we need this method?

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -381,12 +376,124 @@ func (m *gceManagerImpl) fetchAutoMigs() error {
...
+
+			instanceRefs, err := m.regenerateInstanceCacheForMigReturnInstances(migRef)
+			if err != nil {
+				return nil, fmt.Errorf("error while looking for MIG for instance %+v, error: %v", instanceRef, err)
+			}
+
+			// see if instance found among instances obatained from GCE
+			found := false
+			for _, migInstanceRef := range instanceRefs {
+				if migInstanceRef == instanceRef {
```
aleksandra-malinowska: why not look it up in cache again?

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -381,12 +376,124 @@ func (m *gceManagerImpl) fetchAutoMigs() error {
...
+}
+
+// RegenerateInstanceCacheForMig triggers instances cache regeneration for single MIG
+func (m *gceManagerImpl) RegenerateInstanceCacheForMig(migRef GceRef) error {
+	_, err := m.regenerateInstanceCacheForMigReturnInstances(migRef)
+	return err
+}
+
+// RegenerateInstanceCacheForMig triggers instances cache regeneration for single MIG
+func (m *gceManagerImpl) regenerateInstanceCacheForMigReturnInstances(migRef GceRef) (instanceRefs []GceRef, err error) {
```
aleksandra-malinowska: nit: please don't name it like this:(

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -381,12 +376,124 @@ func (m *gceManagerImpl) fetchAutoMigs() error {
...
+	}
+}
+
+func (gc *GceCache) getMig(migRef GceRef) (mig Mig, found bool) {
+	mig, found = gc.migs[migRef]
+	return
+}
+
+// RegenerateInstanceCacheForMig triggers instances cache regeneration for single MIG
+func (m *gceManagerImpl) RegenerateInstanceCacheForMig(migRef GceRef) error {
```
aleksandra-malinowska: do we need this method? it just calls one other method

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -381,12 +376,124 @@ func (m *gceManagerImpl) fetchAutoMigs() error {
...
 }
 
+// GetMigForInstance gets mig for given instance ref
+// Attempts to regenerate cache if there is a Mig with matching prefix in migs list.
+// TODO(aleksandra-malinowska): reconsider failing when there's a Mig with
+// matching prefix, but instance doesn't belong to it.
+func (m *gceManagerImpl) GetMigForInstance(instanceRef GceRef) (Mig, error) {
+	migRef, found := m.cache.GetMigRefForInstance(instanceRef)
+	if found {
+		mig, found := m.cache.getMig(migRef)
```
aleksandra-malinowska: GetMig

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -381,12 +376,124 @@ func (m *gceManagerImpl) fetchAutoMigs() error {
...
 	}
 
 	return nil
 }
 
+// GetMigForInstance gets mig for given instance ref
+// Attempts to regenerate cache if there is a Mig with matching prefix in migs list.
+// TODO(aleksandra-malinowska): reconsider failing when there's a Mig with
+// matching prefix, but instance doesn't belong to it.
+func (m *gceManagerImpl) GetMigForInstance(instanceRef GceRef) (Mig, error) {
```
aleksandra-malinowska: This was previously done under lock, are we sure we don't need it here?

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -61,217 +53,127 @@ type MachineTypeKey struct {
...
 
-// Methods locking on cacheMutex.
-
-// GetMigForInstance returns Mig to which the given instance belongs.
-// Attempts to regenerate cache if there is a Mig with matching prefix in migs list.
-// TODO(aleksandra-malinowska): reconsider failing when there's a Mig with
-// matching prefix, but instance doesn't belong to it.
-func (gc *GceCache) GetMigForInstance(instanceRef GceRef) (Mig, error) {
+// GetMigRefForInstance returns Mig to which the given instance belongs.
+func (gc *GceCache) GetMigRefForInstance(instanceRef GceRef) (migRef GceRef, found bool) {
```
aleksandra-malinowska: Can't we keep 'dumb' GetMigForInstance (which will only succeed if both MIG and instance are cached)? On cache miss, user can trigger regeneration, then try again.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -372,19 +369,131 @@ func (m *gceManagerImpl) fetchAutoMigs() error {
...
+				if migInstanceRef == instanceRef {
+					found = true
+					break
+				}
+			}
+			if !found {
+				return nil, fmt.Errorf("instance %+v belongs to unknown mig", instanceRef)
+			}
+
+			mig, found := m.cache.GetMig(migRef)
```
aleksandra-malinowska: I may be missing something, but this happens in body of the loop starting with:

`for _, mig := range m.cache.GetMigs() `

Isn't `mig` already set to the correct value?

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -61,217 +53,127 @@ type MachineTypeKey struct {
...
 func NewGceCache(gceService AutoscalingGceClient) GceCache {
 	return GceCache{
-		migs:                map[GceRef]*MigInformation{},
+		migs:                map[GceRef]Mig{},
 		instanceRefToMigRef: map[GceRef]GceRef{},
 		machinesCache:       map[MachineTypeKey]*gce.MachineType{},
-		GceService:          gceService,
 		migTargetSizeCache:  map[GceRef]int64{},
+		migBaseNameCache:    map[GceRef]string{},
+		GceService:          gceService,
```
aleksandra-malinowska: we no longer need gceService here, right?

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -339,3 +328,32 @@ func (gc *GceCache) SetMachinesCache(machinesCache map[MachineTypeKey]*gce.Machi
...
+// GetMigBasename get basename for given mig from cache.
+func (gc *GceCache) GetMigBasename(migRef GceRef) (basename string, found bool) {
+	gc.cacheMutex.Lock()
+	defer gc.cacheMutex.Unlock()
+	basename, found = gc.migBaseNameCache[migRef]
+	return
+}
+
+// InvalidateMigBasename invalidates basename entry for given mig.
+func (gc *GceCache) InvalidateMigBasename(migRef GceRef) {
```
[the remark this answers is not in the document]
losipiuk: Renamed

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -85,7 +85,7 @@ func (gce *GceCloudProvider) NodeGroups() []cloudprovider.NodeGroup {
 	migs := gce.gceManager.GetMigs()
 	result := make([]cloudprovider.NodeGroup, 0, len(migs))
 	for _, mig := range migs {
-		result = append(result, mig.Config)
+		result = append(result, mig)
```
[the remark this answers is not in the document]
losipiuk: Apparently yes. Cannot convert `[]Mig` to `[]NodeGroup`. Not an expert on gloang type system :)

cluster-autoscaler/cloudprovider/gce/gce_manager.go
losipiuk: I do not see a reason. gce_manager is not thread safe. And the cache itself is guarded by lock.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
losipiuk: It could have been invalidated already.
I have some idea to add some "transactions" to cache interface. But it is more work and it is low prio. So I would skip it for now.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
losipiuk: Yeah - it does not make sense now.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
losipiuk: We dont

cluster-autoscaler/cloudprovider/gce/gce_manager.go
losipiuk: Dropping and renaming the other one.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
losipiuk: :)

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -206,9 +201,9 @@ func (gc *GceCache) removeInstancesForMig(migRef GceRef) {
 	}
 }
 
-func (gc *GceCache) getMig(migRef GceRef) (MigInformation, bool) {
-	mig, found := gc.migs[migRef]
-	return *mig, found
+func (gc *GceCache) getMigNoLock(migRef GceRef) (mig Mig, found bool) {
```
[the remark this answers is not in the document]
losipiuk: It is gone

cluster-autoscaler/cloudprovider/gce/cache.go
aleksandra-malinowska: Not sure I'm following. Isn't the list of MIGs something we fetch from the API, repackage and cache on our side (in case of autodiscovery setup, and no-longer-here GKE cloud provider)?

cluster-autoscaler/cloudprovider/gce/gce_manager.go
aleksandra-malinowska: I believe it's possible to have ongoing regenerateInstancesCache (from goroutine running it every hour, calling regenerateInstanceCacheForMig for every MIG) and regenerateInstanceCacheForMig (from this method) calls. In this case, we need to consider the interactions.

cluster-autoscaler/cloudprovider/gce/cache.go
losipiuk: Mig. As per 'managed instance group` in GCE nomenclature yet. But not `Mig` object as defined in `gce` package. 
The former should be cached. And actually we do that by caching basename, targetsize, instance list - all of those are parts of managed instance group.
Yet `Mig` as a node-group which is subject to autoscaling by GCE cloud provider should IMO reside in gce_manager - and just use the cached bits from cache. Not be part of cache itself. 
Is my intention more clear?

cluster-autoscaler/cloudprovider/gce/gce_manager.go
losipiuk: Yeah - still I do not believe that it should be a problem. It could have if we depended on updating instances cache at some given time spot (like beginning of the loop). But we did not do that previously - and we do not do that now. 
The worst that can happen if extra `regenerateInstancesCache` is called that we will operate on more "recent" data.

cluster-autoscaler/cloudprovider/gce/cache.go
aleksandra-malinowska: I mean caching list of MigRefs for which autoscaling is enabled, together with min/max/autoprovisioned.

losipiuk: The `min/max/autoprovisioned` is not gce thing. Also there is no notion of "invalidating" of list of Migs. It still does not seem natural to hold `Migs` in GceCache.
(Though as I said I am not chaning it right now so we have some more time to think about it :) )

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -179,6 +179,8 @@ func CreateGceManager(configReader io.Reader, discoveryOpts cloudprovider.NodeGr
 		if err := manager.cache.RegenerateInstancesCache(); err != nil {
 			klog.Errorf("Error while regenerating Mig cache: %v", err)
 		}
+		klog.Info("Invalidating MIG basename cache")
+		manager.cache.InvalidateAllMigBasenames()
```
aleksandra-malinowska: Do we need this call? I realize it was previously filled in regenerateInstancesCache, but:
- does basename ever change? 
- if it does/we care, does it matter that there's a race condition (another goroutine may access cache after regenerating instances, but before invalidating this)?

losipiuk: > Do we need this call? I realize it was previously filled in regenerateInstancesCache, but:

Not strictly. It should not change. This is just a "just-in-case-our-cache-got-rotten" kinda call.
I can drop it if you prefer.
 
> * does basename ever change?

I do not believe so.

> * if it does/we care, does it matter that there's a race condition (another goroutine may access cache after regenerating instances, but before invalidating this)?

losipiuk: I will drop this.

PR #1873, opened 2019-04-09
Fixes for CA Magnum cloud provider

### LINKED ISSUE #1870: Magnum cloud provider includes master node in node group
The assumption for the single node group in the magnum autoscaler was that it covers the entire cluster. But since the master node is also registered in kubernetes the cloud provider was including it as a member of the node group as well.

This is the cause of the issue discussed in #1690 where having a single cordoned minion node in the cluster does not allow a scale up. The autoscaler sees that the cordoned minion should not be used in simulations and uses the master node as the template for a healthy node in the node group instead. The pending pods could not be scheduled if another master node was added, so no scale up is triggered.

By excluding the master from the node group
```go
 func (os *magnumCloudProvider) NodeGroupForNode(node *apiv1.Node) (cloudprovider.NodeGroup, error) {
        // TODO: wait for magnum nodegroup support
+       if _, found := node.ObjectMeta.Labels["node-role.kubernetes.io/master"]; found {
+               return nil, nil
+       }
        return &(os.nodeGroups[0]), nil
 }
```
the autoscaler falls back to using a cached version of the minion node from before it was cordoned as the template, and a scale up is possible.

I can make the PR for this tomorrow but I need to do some more testing first.
This needs to be fixed before the magnum cloud provider is included in a patch release of 1.14, as it is necessary for @openstacker for cluster auto healing.

tghartland:
This PR has 3 separate commits

* One which renames things that were overlooked when the cloud provider name was changed from openstack to magnum.
* One which excludes the master node from being considered as a part of the nodegroup (fixes #1870)
* One which converts machine UUIDs from a format without dashes which kubernetes stores, to a format with dashes which openstack uses. Needed for looking up minion stack indices when deleting them.

Regarding the second commit, is there a utility function somewhere in the kubernetes codebase for checking if a role exists on a node, or is this the best way to do it?
```go
if _, found := node.ObjectMeta.Labels["node-role.kubernetes.io/master"]; found {
```

### DIFF
--- a/cluster-autoscaler/cloudprovider/magnum/magnum_manager_heat_test.go
+++ b/cluster-autoscaler/cloudprovider/magnum/magnum_manager_heat_test.go
@@ -296,10 +296,10 @@ func createManagerGetClusterSuccess() *magnumManagerHeat {
 
 	sc := createTestServiceClient()
 
-	osm := createTestMagnumManagerHeat(sc)
-	osm.clusterName = clusterUUID
+	manager := createTestMagnumManagerHeat(sc)
+	manager.clusterName = clusterUUID
 
-	return osm
+	return manager
 }
 
 func createManagerGetClusterFail() *magnumManagerHeat {
@@ -312,19 +312,19 @@ func createManagerGetClusterFail() *magnumManagerHeat {
 
 	sc := createTestServiceClient()
 
-	osm := createTestMagnumManagerHeat(sc)
-	osm.clusterName = badClusterUUID
+	manager := createTestMagnumManagerHeat(sc)
+	manager.clusterName = badClusterUUID
 
-	return osm
+	return manager
 }
 
 func TestNodeGroupSizeSuccess(t *testing.T) {
 	th.SetupHTTP()
 	defer th.TeardownHTTP()
 
-	osm := createManagerGetClusterSuccess()
+	manager := createManagerGetClusterSuccess()
 
-	nodeCount, err := osm.nodeGroupSize("default")
+	nodeCount, err := manager.nodeGroupSize("default")
 	assert.NoError(t, err)
 	assert.Equal(t, clusterNodeCount, nodeCount)
 }
@@ -333,9 +333,9 @@ func TestNodeGroupSizeFail(t *testing.T) {
 	th.SetupHTTP()
 	defer th.TeardownHTTP()
 
-	osm := createManagerGetClusterFail()
+	manager := createManagerGetClusterFail()
 
-	_, err := osm.nodeGroupSize("default")
+	_, err := manager.nodeGroupSize("default")
 	assert.Error(t, err)
 	assert.Equal(t, "could not get cluster: Resource not found", err.Error())
 }
@@ -344,9 +344,9 @@ func TestGetClusterStatusSuccess(t *testing.T) {
 	th.SetupHTTP()
 	defer th.TeardownHTTP()
 
-	osm := createManagerGetClusterSuccess()
+	manager := createManagerGetClusterSuccess()
 
-	gotStatus, err := osm.getClusterStatus()
+	gotStatus, err := manager.getClusterStatus()
 	assert.NoError(t, err)
 	assert.Equal(t, clusterStatus, gotStatus)
 }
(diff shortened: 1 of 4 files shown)


mwielgus:
cc: @strigazi @openstacker @scott-chaney

strigazi:
renaming from os to mcp and so on is trivial, looks fine.

Checking the master label works as well, in the logs it ignores the master node. kube-flannel-autoscale-5xr5tj5dgujs-master-0 - no node group config

For scaling down before the third patch, I reproduced and saw that it couldn't resolve the uuid without dashes to an index. After this patch it works fine.

LGTM, thanks

tghartland:
@scott-chaney in #1690 you said that deleting using the MachineID was working for you, but for @openstacker it isn't due to the different UUID formats ([logs](http://paste.openstack.org/show/749051/), [stack](http://paste.openstack.org/show/749052/)).

Maybe you have a different set up either for the stack or for the MachineIDs in kubernetes, are they both in the same format for you? And would this patch break scaling down for whatever setup you have?

scott-chaney:
> @scott-chaney in #1690 you said that deleting using the MachineID was working for you, but for @openstacker it isn't due to the different UUID formats ([logs](http://paste.openstack.org/show/749051/), [stack](http://paste.openstack.org/show/749052/)).
> 
> Maybe you have a different set up either for the stack or for the MachineIDs in kubernetes, are they both in the same format for you? And would this patch break scaling down for whatever setup you have?

We have continued to use the resource name (same as the instance index) for scaling.  I tested the MachineID scaling manually but we never adopted it in our version.  Looking at it now, the refs_map output set do not match the providerIDs in Kubernetes.  Scaling does work using the values in the refs_map though.  I must have assumed incorrectly that the map contained the instanceIDs in my initial test.  Sorry if this lent to the confusion.

Since we haven't adopted the upstream version yet, there are no immediate concerns from our end, though this may come back up when we start working to converge with upstream.

cluster-autoscaler/cloudprovider/magnum/magnum_manager_heat_test.go
```diff
@@ -260,8 +260,8 @@ var stackGetKubeMinionsStackResponse = fmt.Sprintf(`
         "outputs":[
             {
                 "output_value":{
-                    "0":"ffbc651da661462d94352e01f3020688",
```
openstacker: That's the thing I still don't understand well. Where did you got this sample UUID when you worked on this initially? And why the instance ID in MachineID and ProviderID are using different format?

openstacker: And based on the changed you're doing in this test, I can't see how the case is covered. Did I miss something?

tghartland: My clusters are still using IPs as references in their stacks so I had to modify this example response and I was generating IDs with python `uuid.uuid4().hex` when I should have been using `str(uuid.uuid4())` to get the right format for openstack.

The delete tests use the example stack response and try to delete one of the minions.
https://github.com/kubernetes/autoscaler/blob/d887ba5e873e7fadfae1c7bae949cfab6d97e8db/cluster-autoscaler/cloudprovider/magnum/magnum_manager_heat_test.go#L155-L157
Now that I'm thinking about it, that ProviderID should have dashes in as well, but it isn't used anyway.

As for why the MachineID and ProviderID are different, it seems that the kubelet checks the file `/etc/machine-id` [by default](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/)
```
--machine-id-file string | Comma-separated  list of files to check for machine-id. Use the first one that exists.  (default "/etc/machine-id,/var/lib/dbus/machine-id")
```

```
[fedora@scaler-53-w74zzswdismh-minion-0 ~]$ cat /etc/machine-id
9f7da7a47b514e5197c37607828e850f
```
but I think the ProviderID would be set by the cloud provider by querying the cloud and so would use the openstack format.
```
Name                              MachineID                          ProviderID
scaler-53-w74zzswdismh-master-0   8faa9ed686c44e9ea30eb8fb7ca30b21   openstack:///8faa9ed6-86c4-4e9e-a30e-b8fb7ca30b21
scaler-53-w74zzswdismh-minion-0   9f7da7a47b514e5197c37607828e850f   openstack:///9f7da7a4-7b51-4e51-97c3-7607828e850f
```

strigazi:
looks good to me

cluster-autoscaler/cloudprovider/magnum/magnum_manager_heat_test.go
openstacker: That's a very impressive clarification. Thank you so much, @tghartland

PR #1877, opened 2019-04-10
Migrate filter out schedulabe to PodListProcessor

losipiuk:


cluster-autoscaler/core/filter_out_schedulable.go
```diff
@@ -0,0 +1,68 @@
...
+	// - Scheduler doesn't schedule the pod on the new node
+	//   because according to it logic it doesn't fit there
+	// - CA see the pod is still unschedulable, so it adds another node to help it
+	//
+	// With the check enabled the last point won't happen because CA will ignore a pod
+	// which is supposed to schedule on an existing node.
+
+	// Some unschedulable pods can be waiting for lower priority pods preemption so they have nominated node to run.
+	// Such pods don't require scale up but should be considered during scale down.
+	unschedulablePods, unschedulableWaitingForLowerPriorityPreemption := filterOutExpendableAndSplit(unschedulablePods, context.ExpendablePodsPriorityCutoff)
```
MaciekPytel: This was not created in this PR, but reading it now I think it's actually wrong. Why do we need to keep track of `unschedulableWaitingForLowerPriorityPreemption`? My intuition is we want to just treat them as if they were scheduled on nominated node. This is reflected in code - literally all we ever do with this list is append those pods to allScheduled (both in scale down and filterOutSchedulableBy*). 
Except we only don't append them to allScheduled when calling TryToScaleDown. This make some sense - if they're not really scheduled we shouldn't drain them. But the same could be said about any other pods artificially scheduled by PLPs. So it feels like a more correct thing to do would be:
- save original list of scheduled pods (and pass that to TryToScaleDown). 
- make filterOutExpendableAndSplit a separate PLP (just add unschedulableWaitingForLowerPriorityPreemption to scheduledPods).
WDYT?

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -253,62 +276,20 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
 		klog.Errorf("Failed to list scheduled pods: %v", err)
 		return errors.ToAutoscalerError(errors.ApiCallError, err)
 	}
 
-	allUnschedulablePods, allScheduled, err = a.processors.PodListProcessor.Process(a.AutoscalingContext, allUnschedulablePods, allScheduled, allNodes)
-	if err != nil {
-		klog.Errorf("Failed to process pod list: %v", err)
-		return errors.ToAutoscalerError(errors.InternalError, err)
-	}
+	ConfigurePredicateCheckerForLoop(allUnschedulablePods, allScheduledPods, a.PredicateChecker)
```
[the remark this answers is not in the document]
losipiuk: Possibly we can reconfigure it after PLP.
It needs to be configured based on original pods before PLPs because PLPs can (and do) use predicate checker

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -253,62 +276,20 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
-	metrics.UpdateDurationFromStart(metrics.FilterOutSchedulable, filterOutSchedulableStart)
-
-	if len(unschedulablePodsToHelp) != len(unschedulablePods) {
-		klog.V(2).Info("Schedulable pods present")
-		scaleDownForbidden = true
-	} else {
-		klog.V(4).Info("No schedulable pods")
-	}
+	// todo: this is also computed in filterOutSchedulablePodListProcessor; avoid that.
+	_, unschedulableWaitingForLowerPriorityPreemption := filterOutExpendableAndSplit(unschedulablePodsWithoutTPUs, a.ExpendablePodsPriorityCutoff)
```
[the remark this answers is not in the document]
losipiuk: Not sure what you meant here. But hopefully it is addressed by update for this PR.

cluster-autoscaler/core/static_autoscaler.go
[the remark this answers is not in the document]
MaciekPytel: You're right. Let's leave it the way it is in this PR and we can consider adding additional call in a follow up PR if we find it's needed.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -246,69 +269,36 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
-	//   because according to it logic it doesn't fit there
-	// - CA see the pod is still unschedulable, so it adds another node to help it
-	//
-	// With the check enabled the last point won't happen because CA will ignore a pod
-	// which is supposed to schedule on an existing node.
-	scaleDownForbidden := false
+	ConfigurePredicateCheckerForLoop(unschedulablePods, scheduledPods, a.PredicateChecker)
 
-	unschedulablePodsWithoutTPUs := tpu.ClearTPURequests(allUnschedulablePods)
+	scaleDownForbidden := false
```
MaciekPytel: I think this is no longer needed with your changes.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -246,69 +269,36 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
+	unschedulablePods = tpu.ClearTPURequests(unschedulablePods)
 
+	// todo: move split and append below to separate PodListProcessor
```
MaciekPytel: Are you going to do this in follow-up PR?

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -367,7 +357,7 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 		scaleDown.CleanUp(currentTime)
 		potentiallyUnneeded := getPotentiallyUnneededNodes(autoscalingContext, allNodes)
 
-		typedErr := scaleDown.UpdateUnneededNodes(allNodes, potentiallyUnneeded, append(allScheduled, unschedulableWaitingForLowerPriorityPreemption...), currentTime, pdbs)
+		typedErr := scaleDown.UpdateUnneededNodes(allNodes, potentiallyUnneeded, originalScheduledPods, currentTime, pdbs)
```
MaciekPytel: Shouldn't you pass scheduledPods here, not originalScheduledPods? That's how it worked before and it seems correct to me -> it makes sense for nominated pod to block scale-down of a node I think.

cluster-autoscaler/main.go
```diff
@@ -278,7 +278,10 @@ func buildAutoscaler() (core.Autoscaler, error) {
 	autoscalingOptions := createAutoscalingOptions()
 	kubeClient := createKubeClient(getKubeConfig())
 	eventsKubeClient := createKubeClient(getKubeConfig())
+
 	processors := ca_processors.DefaultProcessors()
+	processors.PodListProcessor = core.NewFilterOutSchedulablePodListProcessor()
```
MaciekPytel: This should really be part of DefaultProcessors. I guess it's tricky due to cyclic dependencies though, right? If so it's not worth refactoring everything just to put it there.

cluster-autoscaler/core/static_autoscaler.go
losipiuk: Yes.

cluster-autoscaler/core/static_autoscaler.go
losipiuk: Changed to use processorCallbacks with static_autoscaler.go

cluster-autoscaler/core/static_autoscaler.go
losipiuk: hmm on the other hand there are pods to be preempted on that node already. Seems like using `scheduledPods` should be a bit safer approach. Yet both approaches seem kinda broken.

cluster-autoscaler/main.go
losipiuk: I will send ahother PR with just moving things around (and no logic changes) when we decide this one is good to go

MaciekPytel:
Seems like you have a lot of unrelated comments in your PR. Rebase gone wrong?

PR #1878, opened 2019-04-11
Fix garbage collection for empty aggregate container states.

bskiba:
Fixes the issue that we never removed aggregate container states for pods that lived too short too get any samples (common with short-lived cronjobs).
Adds garbage collection of old empty containers. Also removes aggregate container states that are empty and inactive - there are no live pods that can contribute metric samples.

/assign @jbartosik 

FYI @kgolab @schylek

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/model/cluster.go
+++ b/vertical-pod-autoscaler/pkg/recommender/model/cluster.go
@@ -292,13 +292,24 @@ func (cluster *ClusterState) findOrCreateAggregateContainerState(containerID Con
 }
 
 // GarbageCollectAggregateCollectionStates removes obsolete AggregateCollectionStates from the ClusterState.
+// AggregateCollectionState is obsolete in following situations:
+// 1) It has no samples and there are no more active pods that can contribute,
+// 2) The last sample is too old to give meaningful recommendation (>8 days),
+// 3) There are no samples and the aggregate state was created >8 days ago.
 func (cluster *ClusterState) GarbageCollectAggregateCollectionStates(now time.Time) {
 	klog.V(1).Info("Garbage collection of AggregateCollectionStates triggered")
 	keysToDelete := make([]AggregateStateKey, 0)
+	activeKeys := cluster.getActiveAggregateStateKeys()
 	for key, aggregateContainerState := range cluster.aggregateStateMap {
+		isKeyActive := activeKeys[key]
+		if !isKeyActive && aggregateContainerState.isEmpty() {
+			keysToDelete = append(keysToDelete, key)
+			klog.V(1).Infof("Removing empty and inactive AggregateCollectionState for %+v", key)
+			continue
+		}
 		if aggregateContainerState.isExpired(now) {
 			keysToDelete = append(keysToDelete, key)
-			klog.V(1).Infof("Removing AggregateCollectionStates for %+v", key)
+			klog.V(1).Infof("Removing expired AggregateCollectionState for %+v", key)
 		}
 	}
 	for _, key := range keysToDelete {
@@ -307,6 +318,20 @@ func (cluster *ClusterState) GarbageCollectAggregateCollectionStates(now time.Ti
 			vpa.DeleteAggregation(key)
 		}
 	}
+}
+
+func (cluster *ClusterState) getActiveAggregateStateKeys() map[AggregateStateKey]bool {
+	activeKeys := map[AggregateStateKey]bool{}
+	for _, pod := range cluster.Pods {
+		// Pods that will not run anymore are considered inactive.
+		if pod.Phase == apiv1.PodSucceeded || pod.Phase == apiv1.PodFailed {
+			continue
+		}
+		for container := range pod.Containers {
+			activeKeys[cluster.MakeAggregateStateKey(pod, container)] = true
+		}
+	}
+	return activeKeys
 }
 
 // Implementation of the AggregateStateKey interface. It can be used as a map key.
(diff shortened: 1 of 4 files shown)


vertical-pod-autoscaler/pkg/recommender/model/aggregations_config.go
```diff
@@ -23,8 +23,6 @@ import (
 )
 
 var (
-	// MemoryAggregationWindowLength is the length of the memory usage history
-	// aggregated by VPA, which is 8 days.
```
jbartosik: Why remove the comment?

vertical-pod-autoscaler/pkg/recommender/model/cluster.go
```diff
@@ -292,13 +292,23 @@ func (cluster *ClusterState) findOrCreateAggregateContainerState(containerID Con
...
 // GarbageCollectAggregateCollectionStates removes obsolete AggregateCollectionStates from the ClusterState.
+// AggregateCollectionState is obsolete in two situations:
+// 1) It has no samples and there are no more active pods that can contribute,
+// 2) The last sample is too old to give meaningful recommendation.
 func (cluster *ClusterState) GarbageCollectAggregateCollectionStates(now time.Time) {
 	klog.V(1).Info("Garbage collection of AggregateCollectionStates triggered")
 	keysToDelete := make([]AggregateStateKey, 0)
+	activeKeys := cluster.getActiveAggregateStateKeys()
 	for key, aggregateContainerState := range cluster.aggregateStateMap {
+		isKeyActive := activeKeys[key]
```
jbartosik: What do you think about making activeKeys map[AggregateStateKey]struct{} and checking here if the key is present (instead of checking if it maps to true)?

vertical-pod-autoscaler/pkg/recommender/model/cluster.go
```diff
@@ -292,13 +292,23 @@ func (cluster *ClusterState) findOrCreateAggregateContainerState(containerID Con
 }
 
 // GarbageCollectAggregateCollectionStates removes obsolete AggregateCollectionStates from the ClusterState.
+// AggregateCollectionState is obsolete in two situations:
+// 1) It has no samples and there are no more active pods that can contribute,
+// 2) The last sample is too old to give meaningful recommendation.
```
jbartosik: // 3) There are no samples and the container aggregator was created >8 days ago.

?

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
```diff
@@ -82,6 +82,59 @@ func TestClusterGCAggregateContainerStateDeletesOld(t *testing.T) {
...
+	addTestPod(cluster)
+
+	assert.NoError(t, cluster.AddOrUpdateContainer(testContainerID, testRequest))
+	// No usage samples added.
+
+	assert.NotEmpty(t, cluster.aggregateStateMap)
+	assert.NotEmpty(t, vpa.aggregateContainerStates)
+
+	assert.Len(t, cluster.aggregateStateMap, 1)
+	var creationTime time.Time
```
jbartosik: cluster.aggregateStateMap[0].CreationTime = ...
instead of the loop? The test asserts the length is 1 right above this.

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
```diff
@@ -82,6 +82,59 @@ func TestClusterGCAggregateContainerStateDeletesOld(t *testing.T) {
...
+	for _, aggregateState := range cluster.aggregateStateMap {
+		creationTime = aggregateState.CreationTime
+	}
+
+	// AggegateContainerState are valid for 8 days since creation
+	cluster.GarbageCollectAggregateCollectionStates(creationTime.Add(9 * 24 * time.Hour))
```
jbartosik: Please do a clean up with smaller timestamp, assert the aggregation isn't cleaned up just because it lacks active pods, then do clean up with creationTime + 9 days to check that it's cleaned up because it's old.

vertical-pod-autoscaler/pkg/recommender/model/cluster.go
bskiba: It's basically the same (both are correct ways of representing a set in go) and I really hate writing activeKeys[cluster.MakeAggregateStateKey(pod, container)] = struct{}{}

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
bskiba: The map keys are not ints but aggregateStateKeys. I could construct the key, but the fastest way is to iterate.

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
bskiba: Done

vertical-pod-autoscaler/pkg/recommender/model/aggregations_config.go
bskiba: Unintentional, added back.

vertical-pod-autoscaler/pkg/recommender/model/cluster.go
bskiba: Done

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
jbartosik: Ack.

PR #1882, opened 2019-04-11
Rebase addon-resizer to distroless

yuwenma:
This is part of the effort described in KEP kubernetes/enhancements#900
This mitigates the affect of GVE issues.


yuwenma:
> Do we have support for all the removed architectures in distroless?

Great question. Just double checked with Tim Allstair and we should be good.
See [this](https://github.com/GoogleContainerTools/distroless/issues/346) for detail

bskiba:
@yuwenma  Sorry to be late to the party, but addon-resizer still uses glog, won't this be an issue?

PR #1885, opened 2019-04-12
Update go to 1.12.1

losipiuk:
Change-Id: Ibfb0aeeb6e70ca0aba44b192f67470aec31cef04


tghartland:
Should 1.12 also be added to the go version check?
https://github.com/kubernetes/autoscaler/blob/22dba813bee8d82c0e653b1dc36fdd6ddcc46d2d/hack/check-go-version.sh#L21

PR #1889, opened 2019-04-12
Disable failing priority expander tests

losipiuk:


jkaniuk:
cc: @piontec

MaciekPytel:
/lgtm
/approve

I suspect https://github.com/kubernetes/autoscaler/pull/1843 will fix those test.

jkaniuk:
> I suspect #1843 will fix those test.

I've checked and it does fix it indeed

PR #1895, opened 2019-04-12
Keep track of VPAs with no recommendation.

bskiba:


bskiba:
@kgolab

vertical-pod-autoscaler/pkg/recommender/routines/recommender.go
```diff
@@ -105,8 +107,19 @@ func (r *recommender) UpdateVPAs() {
...
 				metrics_recommender.ObserveRecommendationLatency(vpa.Created)
 			}
+			delete(r.clusterState.EmptyVPAs, vpa.ID)
 		} else {
 			vpa.Conditions.Set(vpa_types.RecommendationProvided, false, "", "")
+			firstObserved, ok := r.clusterState.EmptyVPAs[vpa.ID]
+			if !ok {
+				r.clusterState.EmptyVPAs[vpa.ID] = time.Now()
+			} else {
+				if firstObserved.Add(RecommendationMissingMaxDuration).Before(time.Now()) {
```
kgolab: I'd change the clusterState.EmptyVPAs meaning slightly so keep track the time from the last warning (and reset it inside this if) so we'd emit a warning for a problematic VPA twice an hour, not every minute. WDYT?

bskiba: Done.

bskiba:
I've reworked this change a little, added unit test (and rightly so as they caught some bugs...). PTAL

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
```diff
@@ -386,3 +388,78 @@ func TestEmptySelector(t *testing.T) {
 	assert.Contains(t, vpa.aggregateContainerStates, cluster.aggregateStateKeyForContainerID(containerID1))
 	assert.Contains(t, vpa.aggregateContainerStates, cluster.aggregateStateKeyForContainerID(containerID2))
 }
+
+func TestRecordRecommendation(t *testing.T) {
+	testTimestamp := time.Unix(1555400, 0)
```
kgolab: Any reason why this particular timestamp?

bskiba: Nah, removed. I didn't notice there is already a testTimestamp available

kgolab:
Thanks, LGTM.

PR #1902, opened 2019-04-16
Shelllcheck fixes for shell scripts in ./hack directory

achinthagunasekara:
Shelllcheck fixes for shell scripts in ./hack directory


hack/for-go-proj.sh
```diff
@@ -49,6 +49,7 @@ for godep_file in ${godep_projects}; do
     cd "${project}"
     case "${CMD}" in
       "test")
+        #shellcheck disable=SC2046
```
MaciekPytel: Can we avoid adding all those comments everywhere?

hack/for-go-proj.sh
```diff
@@ -18,7 +18,7 @@ set -o errexit
 set -o pipefail
 set -o nounset
 
-CONTRIB_ROOT="$(dirname ${BASH_SOURCE})/.."
+CONTRIB_ROOT="$(dirname "${BASH_SOURCE[0]}")/.."
```
kgolab: I'm not sure that the extra quotation marks here help any, the whole variable is already surrounded by quotes.
Similar issue is visible in kube-env.sh.

hack/install-verify-tools.sh
```diff
@@ -18,7 +18,8 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-KUBE_ROOT=$(dirname "${BASH_SOURCE}")/..
+KUBE_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+"${KUBE_ROOT}"/hack/check-go-version.sh
```
kgolab: I think it would be good not to mix functional changes with formatting/code fixes.

MaciekPytel:
Have you actually tested all those scripts after the changes? I'm suspicious about just applying linter changes to shell scripts, it's just so easy for those to have unexpected behavior changes (especially when adding quotes to bash commands, which this PR seems to do quite liberally).

achinthagunasekara:
I'll break this into smaller PRs and do more testing :)

PR #1911, opened 2019-04-17
Magnum cloud provider for 1.14.1

tghartland:
Auto scaling as a default option is now merged in magnum https://review.openstack.org/#/c/631378/.
Since the first fixes in #1873 we have not found any other issues with the cloud provider, so it should be ready for a patch release now.
@losipiuk

### DIFF
--- a/cluster-autoscaler/Godeps/Godeps.json
+++ b/cluster-autoscaler/Godeps/Godeps.json
@@ -1221,6 +1221,10 @@
 			"Rev": "c818fa66e4c88b30db28038fe3f18f2f4a0db9a8"
 		},
 		{
+			"ImportPath": "github.com/gophercloud/gophercloud/openstack/containerinfra/v1/clusters",
+			"Rev": "c818fa66e4c88b30db28038fe3f18f2f4a0db9a8"
+		},
+		{
 			"ImportPath": "github.com/gophercloud/gophercloud/openstack/identity/v2/tenants",
 			"Rev": "c818fa66e4c88b30db28038fe3f18f2f4a0db9a8"
 		},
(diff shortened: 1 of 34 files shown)


cluster-autoscaler/Godeps/Godeps.json
```diff
@@ -1220,6 +1220,10 @@
 			"ImportPath": "github.com/gophercloud/gophercloud/openstack/compute/v2/servers",
 			"Rev": "c818fa66e4c88b30db28038fe3f18f2f4a0db9a8"
 		},
+		{
+			"ImportPath": "github.com/gophercloud/gophercloud/openstack/containerinfra/v1/clusters",
```
losipiuk: It seems that you are adding the new CA dependencies which are not dependencies of k8s.
This is not compatible with how we manage dependencies for CA.

The whole process is described here: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-update-ca-dependencies-particularly-k8siokubernetes.
tl;dr; on updating dependencies we are deleting `cluster-autoscaler/Godeps/` and `cluster-autoscaler/vendor/` directories and rebuild those based on `GOPATH`  which we get after running `godeps restore` in k8s directory. (*)

The extra dependencies you add will not be available after revendoring. 

We strongly prefer that cloud providers only use dependencies already available in k8s. And all other dependencies are included in `cluster-autoscaler/cloudprovider/XXXXX` source directory.

Alternative (much less preferred) is to put extra dependencies in `cluster-autoscaler/_override` directory and modifying `cluster-autoscaler/fix_gopath.sh` to include those as part of revendoring.

(*) We will need to change the process somehow as k8s moved to use of go modules - but I do not expect that we will change it significantly - I expect the `Godeps` and `vendor` dirs will still be recreated.

PS. Sorry for late spotting that in cherry-pick and not in original PR.

tghartland:
Will wait for 1.15 instead, to avoid adding a feature in a patch release.

PR #1912, opened 2019-04-17
[cherry-pick] Add logging for errors on adding samples.

bskiba:


vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -211,10 +211,12 @@ func (feeder *clusterStateFeeder) InitFromHistoryProvider(historyProvider histor
 				ContainerName: containerName}
 			glog.V(4).Infof("Adding %d samples for container %v", len(sampleList), containerID)
 			for _, sample := range sampleList {
-				feeder.clusterState.AddSample(
+				if err := feeder.clusterState.AddSample(
```
jbartosik: This is a cherry pick so please leave as-is but I think that in case of multi line function invocations 

err := functionWithLongArgList(...)
if err != nil {
  // stuff
}

 is easier to read.

PR #1915, opened 2019-04-17
Set appropriate conditions on VPAs with no active pods matched.

bskiba:


bskiba:
Fixes #1181

bskiba:
@kgolab

vertical-pod-autoscaler/pkg/recommender/input/spec/spec_client.go
```diff
@@ -76,6 +76,12 @@ func (client *specClient) GetPodSpecs() ([]*BasicPodSpec, error) {
 	}
 	return podSpecs, nil
 }
+
+// IsActive returns true if Pod is in active phase.
+func (pod *BasicPodSpec) IsActive() bool {
+	return pod.Phase != v1.PodSucceeded && pod.Phase != v1.PodFailed
```
kgolab: What exactly do you want to be covered by "active" phase? Is it a state where we still expect usage samples from the Pod's contaners?
Looking at https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase I'd expect at least Completed to be excluded here too.

vertical-pod-autoscaler/pkg/recommender/model/cluster.go
```diff
@@ -90,14 +94,20 @@ type PodState struct {
 	Phase apiv1.PodPhase
 }
 
+// IsActive returns true if Pod is in active phase.
```
kgolab: Looks like a duplicate from spec_client.go - do we need both?

vertical-pod-autoscaler/pkg/recommender/model/cluster.go
```diff
@@ -292,15 +302,17 @@ func (cluster *ClusterState) aggregateStateKeyForContainerID(containerID Contain
...
 	aggregateStateKey := cluster.aggregateStateKeyForContainerID(containerID)
 	aggregateContainerState, aggregateStateExists := cluster.aggregateStateMap[aggregateStateKey]
 	if !aggregateStateExists {
 		aggregateContainerState = NewAggregateContainerState()
 		cluster.aggregateStateMap[aggregateStateKey] = aggregateContainerState
 		// Link the new aggregation to the existing VPAs.
 		for _, vpa := range cluster.Vpas {
-			vpa.UseAggregationIfMatching(aggregateStateKey, aggregateContainerState)
+			if vpa.UseAggregationIfMatching(aggregateStateKey, aggregateContainerState) && isPodActive {
+				cluster.VpasWithMatchingPods[vpa.ID] = true
```
kgolab: Why are we checking isPodActive here?
This way we'll mark [Cron]Job-related VPA as not matching pods quite often, won't we?

vertical-pod-autoscaler/pkg/recommender/model/cluster.go
```diff
@@ -369,6 +382,22 @@ func (cluster *ClusterState) RecordRecommendation(vpa *Vpa, now time.Time) error
 	return nil
 }
 
+// GetActiveMatchingPods returns a list of currently active pods that match the
```
kgolab: This is a potentially heavy function. I can see it being used now only for logging an erroneous situation but I'd like to make sure it won't become accidentally used elsewhere.
Maybe a comment that it should be used sparingly would be good?

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -363,13 +363,14 @@ func (feeder *clusterStateFeeder) LoadPods() {
 			feeder.clusterState.DeletePod(key)
 		}
 	}
+	feeder.clusterState.VpasWithMatchingPods = make(map[model.VpaID]bool, len(feeder.clusterState.Vpas))
```
kgolab: I don't like this, it looks like we're exposing internals of feeder.clusterState here. OTOH we do it already with other parts of the state so maybe it's not really actionable.

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
```diff
@@ -462,3 +463,139 @@ func TestRecordRecommendation(t *testing.T) {
...
+			name:        "No matching pods",
+			vpaSelector: testSelectorStr,
+			pods: []podDesc{
+				{
+					id:     testPodID,
+					labels: emptyLabels,
+					phase:  apiv1.PodRunning,
+				}, {
+					id:     PodID{Namespace: "different-than-vpa", PodName: "pod-1"},
+					labels: emptyLabels,
```
kgolab: It should testLabels here, right?

vertical-pod-autoscaler/pkg/recommender/model/vpa.go
```diff
@@ -164,3 +170,42 @@ func (vpa *Vpa) matchesAggregation(aggregationKey AggregateStateKey) bool {
...
+
+// UpdateConditions updates the conditions of VPA objects based on it's state.
+// PodsMatched is passed to indicate if there are currently active pods in the
+// cluster matching this VPA.
+func (vpa *Vpa) UpdateConditions(podsMatched bool) {
+	reason := ""
+	msg := ""
+	if !podsMatched {
+		reason = "NoPodsMatched"
+		msg = "No live pods match this VPA object"
```
kgolab: As noted elsewhere - I wonder why we limit ourselves to live Pods. Can we discuss it off-line?

vertical-pod-autoscaler/pkg/recommender/model/vpa.go
```diff
@@ -164,3 +170,42 @@ func (vpa *Vpa) matchesAggregation(aggregationKey AggregateStateKey) bool {
 	}
 	return vpa.PodSelector != nil && vpa.PodSelector.Matches(aggregationKey.Labels())
 }
+
+// AsStatus returns this objects equivalent of VPA Status.
+func (vpa *Vpa) AsStatus() *vpa_types.VerticalPodAutoscalerStatus {
```
kgolab: Do I understand correctly that this function requires UpdateConditions being called first? If so - please comment as such.
The same applies to HasMatchedPods.

vertical-pod-autoscaler/pkg/recommender/model/vpa_test.go
```diff
@@ -39,3 +44,89 @@ func TestMergeAggregateContainerState(t *testing.T) {
...
+			expectedConditions: []vpa_types.VerticalPodAutoscalerCondition{
+				{
+					Type:    vpa_types.RecommendationProvided,
+					Status:  core.ConditionTrue,
+					Reason:  "",
+					Message: "",
+				}, {
+					Type:    vpa_types.NoPodsMatched,
+					Status:  core.ConditionTrue,
+					Reason:  "",
```
kgolab: Why not "NoPodsMatched" here?

vertical-pod-autoscaler/pkg/recommender/model/vpa.go
```diff
@@ -164,3 +170,42 @@ func (vpa *Vpa) matchesAggregation(aggregationKey AggregateStateKey) bool {
...
+// UpdateConditions updates the conditions of VPA objects based on it's state.
+// PodsMatched is passed to indicate if there are currently active pods in the
+// cluster matching this VPA.
+func (vpa *Vpa) UpdateConditions(podsMatched bool) {
+	reason := ""
+	msg := ""
+	if !podsMatched {
+		reason = "NoPodsMatched"
+		msg = "No live pods match this VPA object"
+		vpa.Conditions.Set(vpa_types.NoPodsMatched, true, "", msg)
```
kgolab: true, reason, msg instead of true, "", msg ?

vertical-pod-autoscaler/pkg/recommender/input/spec/spec_client.go
bskiba: I'd expect it too if it existed :) It seems the docs are not in sync with the api: https://github.com/kubernetes/api/blob/7b2abe14437878fa82cec898c0cd89bd09ad5250/core/v1/types.go#L2380

vertical-pod-autoscaler/pkg/recommender/model/cluster.go
bskiba: They work on different types. (BasicPodSpec and PodStatus) and  I can try and merge into one if it's worth it.

vertical-pod-autoscaler/pkg/recommender/model/cluster.go
bskiba: Yes, this is something I'm not entirely sure about. For short running pods this doesn't work too well. On the other hand, for deployments, if we count terminated pods, after deployment scales down to 0 it will take us 8 days to start saying that there are no matching pods. I'm open to discussion on this.

vertical-pod-autoscaler/pkg/recommender/routines/recommender.go
```diff
@@ -102,22 +102,27 @@ func (r *recommender) UpdateVPAs() {
 		vpa.Recommendation = getCappedRecommendation(vpa.ID, resources, observedVpa.Spec.ResourcePolicy)
 		// Set RecommendationProvided if recommendation not empty.
```
kgolab: This comment became obsolete.

vertical-pod-autoscaler/pkg/recommender/routines/recommender.go
```diff
@@ -102,22 +102,27 @@ func (r *recommender) UpdateVPAs() {
...
 		if len(vpa.Recommendation.ContainerRecommendations) > 0 {
-			vpa.Conditions.Set(vpa_types.RecommendationProvided, true, "", "")
 			if !had {
```
kgolab: Could you please merge these two lines into a single if?

vertical-pod-autoscaler/pkg/utils/metrics/recommender/recommender.go
```diff
@@ -48,7 +48,7 @@ var (
 			Namespace: metricsNamespace,
 			Name:      "vpa_objects_count",
 			Help:      "Number of VPA objects present in the cluster.",
-		}, []string{"update_mode", "has_recommendation", "api"},
+		}, []string{"update_mode", "has_recommendation", "api", "pods_matched"},
```
kgolab: nit: matches_pods is (for me) more suggestive of a bool value.

vertical-pod-autoscaler/pkg/recommender/input/spec/spec_client.go
bskiba: Removed anyway as I moved away from checking if matched pods are active per offline chat.

vertical-pod-autoscaler/pkg/recommender/model/cluster.go
bskiba: Removed.

vertical-pod-autoscaler/pkg/recommender/model/cluster.go
bskiba: Agreed to not take into account if pod is active. Check removed.

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
bskiba: Done.

vertical-pod-autoscaler/pkg/recommender/model/vpa.go
bskiba: Done.

vertical-pod-autoscaler/pkg/recommender/model/vpa.go
bskiba: Done.

vertical-pod-autoscaler/pkg/recommender/model/vpa_test.go
bskiba: Done

vertical-pod-autoscaler/pkg/recommender/model/vpa.go
bskiba: Done

vertical-pod-autoscaler/pkg/recommender/routines/recommender.go
bskiba: Done.

vertical-pod-autoscaler/pkg/recommender/routines/recommender.go
bskiba: Done

vertical-pod-autoscaler/pkg/utils/metrics/recommender/recommender.go
bskiba: Done.

vertical-pod-autoscaler/pkg/recommender/model/cluster.go
bskiba: Done.

kgolab:
Thanks, LGTM.

vertical-pod-autoscaler/pkg/recommender/model/vpa.go
```diff
@@ -164,3 +170,45 @@ func (vpa *Vpa) matchesAggregation(aggregationKey AggregateStateKey) bool {
...
+	if !podsMatched {
+		reason = "NoPodsMatched"
+		msg = "No live pods match this VPA object"
+		vpa.Conditions.Set(vpa_types.NoPodsMatched, true, reason, msg)
+	}
+	if vpa.HasRecommendation() {
+		vpa.Conditions.Set(vpa_types.RecommendationProvided, true, "", "")
+	} else {
+		vpa.Conditions.Set(vpa_types.RecommendationProvided, false, reason, msg)
+	}
```
hjkatz: Should recommendations continue to be provided on the status if the vpa doesn't match any Pods?

I'm seeing a case where a vpa has no matching pods and still has recommendations, and these recommendations are meaningless (because no pods match to help make the calculations reasonable). So when graphing or viewing data about these vpas we're seeing upperBounds of 30,000mCores as a recommendation for a workload that isn't running at all.

What are your thoughts on adjusting this logic to always remove recommendations after a grace period of time (perhaps 8 days? or the same timeframe as the calculations?) for vpas that have status `NoPodsMatched`?

hjkatz: Perhaps here (https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/recommender/model/vpa.go#L145) could be the place to remove ContainerRecommendations for containers that no longer exist or no longer match if the `NoPodsMatch` condition is set?

Let me know your thoughts and I can open a PR or Issue to further discussion :)

bskiba: Ufortunately the NoPodsMatched condition does not work well atm: see #2465, so we can't rely on it.

As good of an idea as it sounds for some class workloads, there are some that would be harmed by removing recommendations that are "old". A good example would be rarely running CronJobs. There is a benefit of having a bootstrap recommendation you can use for new pods even if it is an "old" one.

However, please feel free to open an issue to discuss, perhaps there is a way we can make this work.

hjkatz: I was just reading #2465 yesterday as I'm experiencing the same issues in our clusters too. I was reading through the code a bit to try and understand the behaviour of the NoPodsMatched condition, but I haven't gotten far.

As for old recommendations during the NoPodsMatched condition, I don't think CronJobs would benefit from being recommended to have 30000mCores when NoPodsMatched. The behaviour we're seeing is that when NoPodsMatched is set (and there _really_ aren't any Pods) then the recommender is updating the recommendations to be at the extremes with lowerbound <= 10m and upperbound >= 30000m for CPU. 

I agree that having old _correct_ recommendations for rarely running CronJobs would be beneficial as when the CronJob does eventually run the Pod would be admitted with the last good recommendation. However, that's not what I'm seeing at this time. Currently the CronJob's Pod would be admitted with very poor and extreme recommendations.

bskiba: The lowerBound and upperBound are not used to set the actual resource, only target matters. It's a bit unfortunate that they are together in the recommendation object. UpperBound and lowerBound signal the confidence of recommendation and are used only by updater to decide if it should evict running pods. So it does not mean that we recommend 30000 mCPU

hjkatz: Oh interesting. I was unable to find much documentation on what the different values mean. We're using https://github.com/FairwindsOps/goldilocks this tool to provide many VPAs for all workloads and then build dashboards off of this information. Goldilocks seems to believe that lowerbound == requests and upperbound == limits, is this not the case?

What should we use from the VPA recommendation as requests and which for limits?

bskiba: Yeah, we need to document this better, it's true.

classical VPA sets request=target and limit in such a way that the original ratio of resource to request is kept constant.

So if you have in a deployment
request=100mCPU
limit=mCPU

VPA will set request=target and limit=2*target

Setting request=lowerBound seems risky, I'd keep request=target. limit=upperBound might work, with the condition that you need to wait until upperBound is reasonably close to target.

bskiba: As a side note, does the Goldilocks tool provide a way to automatically apply the requests or just display the recommended resources?

hjkatz: Goldilocks only displays recommendations. We have an internal fork that does more than that and I'm actively working on getting their CLA signed so I can submit patches upstream. So soon™️ 😄

PR #1920, opened 2019-04-18
priority expander based on client-go for cluster autoscaler is missing permissions

piontec:
@MaciekPytel 
Related to: https://github.com/kubernetes/autoscaler/pull/1843
Priority expander needs now wider permissions, so I updated deployment examples.


piontec:
Hi @MaciekPytel , do you have any time frame for releasing this? What about backports to previous versions?

mgalgs:
@piontec 

> What about backports to previous versions?

I have a 1.14 backport in progress at #2435

PR #1921, opened 2019-04-18
Unset NoPodsMatched if pods match

bskiba:


vertical-pod-autoscaler/pkg/recommender/model/vpa_test.go
```diff
@@ -51,6 +51,7 @@ func TestUpdateConditions(t *testing.T) {
 		podsMatched        bool
 		hasRecommendation  bool
 		expectedConditions []vpa_types.VerticalPodAutoscalerCondition
+		expectedAbsent     []vpa_types.VerticalPodAutoscalerConditionType
```
kgolab: I think a real test for this would be to call UpdateConditions twice, otherwise the default behaviour (not touching the condition) will pass this check.
Feel free to keep it for a later PR.

kgolab:
Thanks, LGTM.
One possible test improvement noted.

PR #1924, opened 2019-04-19
VPA - add monitoring of VPA object with unsupported config condition.

schylek:


vertical-pod-autoscaler/pkg/utils/metrics/recommender/recommender.go
```diff
@@ -134,11 +135,13 @@ func (oc *ObjectCounter) Add(vpa *model.Vpa) {
...
 	key := objectCounterKey{
-		mode:        mode,
-		has:         vpa.HasRecommendation(),
-		apiVersion:  api,
-		matchesPods: vpa.HasMatchedPods(),
+		mode:              mode,
+		has:               vpa.HasRecommendation(),
+		apiVersion:        api,
+		matchesPods:       vpa.HasMatchedPods(),
+		unsupportedConfig: vpa.Conditions.ConditionActive(vpa_types.ConfigUnsupported),
```
bskiba: I think you need to modify also the NewObjectCounter function.

schylek: Done.

PR #1925, opened 2019-04-19
[cherry-pick] VPA - addmonitoring of VPA object with unsupported config condition.

bskiba:
@schylek @kgolab


kgolab:
Thanks, LGTM.

PR #1935, opened 2019-04-20
Add cache to getInstanceTypeByLCName for avoid aws rate limit

yagonobre:
**What this PR does / why we need it**:
Add cache to `getInstanceTypeByLCName`.

**Special notes for your reviewer**:
If you have a launch configuration that can't bootstrap a valid Kubernetes node (in our case, a bug on user-data), the cluster-autoscaler try to execute a lot of `DescribeLaunchConfigurations` making the Auto Scaling Group and Launch Configurations aws console unresponsive due  to rate limit. 

Launch configuration is immutable, so add a cache by LC name seems safe.


Jeffwan:
#1841 This PR also helps to reduce API Calls. With cache, users should not see any rate limit issues.

losipiuk:
/approve
Thanks.

PR #1942, opened 2019-04-22
CA FAQs - Clarify error situations support

gjtempleton:
Clarification to text raised off the back of discussion in SIG Autoscaling Slack: https://kubernetes.slack.com/archives/C09R1LV8S/p1555945612041900


MaciekPytel:
/lgtm
/approve

Sad to see that the whole mechanism doesn't work, but thanks for documenting this.

gjtempleton:
@MaciekPytel ditto. Going to give the code a look to see whether I can come up with something to understand the spot bids as an entity now I understand the code better, so hopeful I can come back to this at some point and remove it...

PR #1947, opened 2019-04-23
Provide ScaleDownStatusProcessor with more info about scale-down results

towca:


cluster-autoscaler/core/scale_down.go
```diff
@@ -70,10 +70,10 @@ const (
 type NodeDeleteStatus struct {
 	sync.Mutex
 	deleteInProgress bool
-	// A map of node delete results by node name. It contains nil if the delete was successful and an error otherwise.
-	// It's being constantly drained into ScaleDownStatus objects in order to notify the ScaleDownStatusProcessor that
-	// the node drain has ended or that an error occurred during the deletion process.
-	nodeDeleteResults map[string]error
+	// A map of node delete results by node name. It's being constantly drained into ScaleDownStatus
```
losipiuk: the `drain` verb is overloaded in this description (and in logic). Could we rename the `DrainNodeDeleteResults` to `GetAndClearNodeDeleteResults` and avoid using drain for other thing than draining the pods from node?

losipiuk:
Overly looks fine. But review from someone with more context could be helpful.

cluster-autoscaler/core/scale_down.go
towca: Good idea, done!

cluster-autoscaler/core/scale_down.go
```diff
@@ -1031,25 +1040,36 @@ func drainNode(node *apiv1.Node, pods []*apiv1.Pod, client kube_client.Interface
...
 			}
 		case <-time.After(retryUntil.Sub(time.Now()) + 5*time.Second):
+			// All successfully evicted pods were deleted from podsNotYetEvicted, so after a timeout it contains pods that either failed to evict or haven't evicted in time.
+			failingPods := make([]*apiv1.Pod, 0, len(podsNotYetEvicted))
+			for _, pod := range podsNotYetEvicted {
+				failingPods = append(failingPods, pod)
+			}
 			return errors.NewAutoscalerError(
-				errors.ApiCallError, "Failed to drain node %s/%s: timeout when waiting for creating evictions", node.Namespace, node.Name)
+				errors.ApiCallError, "Failed to drain node %s/%s: timeout when waiting for creating evictions", node.Namespace, node.Name), status.NodeDeleteErrorFailedToEvictPods, failingPods
```
aleksandra-malinowska: Looks like we lose information about eviction errors collected so far in this case. It may be OK, just wondering if it's a conscious decision (other than "we never logged them before").

cluster-autoscaler/core/scale_down.go
```diff
@@ -1005,23 +1010,27 @@ func evictPod(podToEvict *apiv1.Pod, client kube_client.Interface, recorder kube
...
-	return fmt.Errorf("failed to evict pod %s/%s within allowed timeout (last error: %v)", podToEvict.Namespace, podToEvict.Name, lastError)
+	return podEvictionResult{pod: podToEvict, err: fmt.Errorf("failed to evict pod %s/%s within allowed timeout (last error: %v)", podToEvict.Namespace, podToEvict.Name, lastError)}
 }
 
 // Performs drain logic on the node. Marks the node as unschedulable and later removes all pods, giving
 // them up to MaxGracefulTerminationTime to finish.
 func drainNode(node *apiv1.Node, pods []*apiv1.Pod, client kube_client.Interface, recorder kube_record.EventRecorder,
-	maxGracefulTerminationSec int, maxPodEvictionTime time.Duration, waitBetweenRetries time.Duration) errors.AutoscalerError {
+	maxGracefulTerminationSec int, maxPodEvictionTime time.Duration, waitBetweenRetries time.Duration,
+	podEvictionHeadroom time.Duration) (err errors.AutoscalerError, resultType status.NodeDeleteResultType, failingPods []*apiv1.Pod) {
```
aleksandra-malinowska: Would it make sense to return map of pod names to errors here? Disconnecting the two here may be OK for aggregation purposes, but it also blocks processing them further.

cluster-autoscaler/core/scale_down.go
```diff
@@ -1067,11 +1087,20 @@ func drainNode(node *apiv1.Node, pods []*apiv1.Pod, client kube_client.Interface
 		if allGone {
 			klog.V(1).Infof("All pods removed from %s", node.Name)
 			// Let the deferred function know there is no need for cleanup
-			return nil
+			return nil, -1, nil
```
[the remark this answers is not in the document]
towca: Changed to sth meaningful

cluster-autoscaler/processors/status/scale_down_status_processor.go
```diff
@@ -61,6 +61,34 @@ const (
...
+const (
+	// NodeDeleteOk - the node was deleted successfully.
+	NodeDeleteOk NodeDeleteResultType = iota
+
+	// NodeDeleteErrorFailedToMarkToBeDeleted - node deletion failed because the node couldn't be marked to be deleted.
+	NodeDeleteErrorFailedToMarkToBeDeleted
+	// NodeDeleteErrorFailedToEvictPods - node deletion failed because some of the pods couldn't be evicted from the node.
+	NodeDeleteErrorFailedToEvictPods
+	// NodeDeleteErrorPodsFailedToDisappear - node deletion failed because some of the pods didn't disappear from the node in time after eviction.
+	NodeDeleteErrorPodsFailedToDisappear
```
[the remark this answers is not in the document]
towca: Added error information

cluster-autoscaler/core/scale_down.go
```diff
@@ -963,23 +963,28 @@ func (sd *ScaleDown) deleteNode(node *apiv1.Node, pods []*apiv1.Pod) errors.Auto
 	sd.context.Recorder.Eventf(node, apiv1.EventTypeNormal, "ScaleDown", "marked the node as toBeDeleted/unschedulable")
 
 	// attempt drain
-	if err := drainNode(node, pods, sd.context.ClientSet, sd.context.Recorder, sd.context.MaxGracefulTerminationSec, MaxPodEvictionTime, EvictionRetryTime); err != nil {
-		return err
+	err, drainResultType, failingPods := drainNode(node, pods, sd.context.ClientSet, sd.context.Recorder, sd.context.MaxGracefulTerminationSec, MaxPodEvictionTime, EvictionRetryTime, PodEvictionHeadroom)
```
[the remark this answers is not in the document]
towca: Yep of course, thanks for that - returns NodeDeleteResult now

cluster-autoscaler/core/scale_down.go
towca: Not anymore!

cluster-autoscaler/core/scale_down.go
towca: Yes, that's a good idea! Although there can still be pods that caused the node failure but don't have any specific error associated (e.g. timed out while waiting for their eviction/disappearance), so I'd leave both fields actually.

cluster-autoscaler/processors/status/scale_down_status_processor.go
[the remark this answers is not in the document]
aleksandra-malinowska: I meant do we even need 2 types - it feels like it could be covered by one (like "NodeDeleteErrorFailedToEvictPods" or "NodeDeleteErrorFailedDrain" or sth) ;)

cluster-autoscaler/processors/status/scale_down_status_processor.go
```diff
@@ -61,6 +61,36 @@ const (
...
+	NodeDeleteErrorFailedToDelete
+)
+
+// NodeDeleteResult contains information about the result of a node deletion.
+type NodeDeleteResult struct {
+	// Err contains nil if the delete was successful and an error otherwise.
+	Err error
+	// ResultType contains the type of the result of a node deletion.
+	ResultType NodeDeleteResultType
+	// FailingPods contains the pods that prevented the node from being deleted, if there were any.
```
aleksandra-malinowska: Just an idea: you already create podEvictionResult type, why not move it here and replace both 'FailingPods' and 'FailingPodErrors' with:

PodEvictionResults map[string]PodEvictionResult

For pods that still exist, attach a new error to result. Then the processors can deal with this to aggregate etc.

towca: Done, something similar at least. Looks much better!

cluster-autoscaler/processors/status/scale_down_status_processor.go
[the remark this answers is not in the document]
towca: Agreed, only one left!

cluster-autoscaler/processors/status/scale_down_status_processor.go
```diff
@@ -72,6 +98,18 @@ func NewDefaultScaleDownStatusProcessor() ScaleDownStatusProcessor {
 	return &NoOpScaleDownStatusProcessor{}
 }
 
+// PodEvictionResult contains the result of an eviction of a pod.
+type PodEvictionResult struct {
+	Pod      *apiv1.Pod
+	TimedOut bool
```
aleksandra-malinowska: nit: timeout could also be represented as an error

cluster-autoscaler/core/scale_down.go
```diff
@@ -1067,11 +1073,22 @@ func drainNode(node *apiv1.Node, pods []*apiv1.Pod, client kube_client.Interface
...
-		errors.TransientError, "Failed to drain node %s/%s: pods remaining after timeout", node.Namespace, node.Name)
+
+	for _, pod := range pods {
+		podReturned, err := client.CoreV1().Pods(pod.Namespace).Get(pod.Name, metav1.GetOptions{})
+		if err == nil && (podReturned == nil || podReturned.Spec.NodeName == node.Name) {
+			evictionResults[pod.Name] = status.PodEvictionResult{Pod: pod, TimedOut: true, Err: nil}
+		} else if err != nil && !kube_errors.IsNotFound(err) {
+			evictionResults[pod.Name] = status.PodEvictionResult{Pod: pod, TimedOut: true, Err: err}
+		} else {
+			evictionResults[pod.Name] = status.PodEvictionResult{Pod: pod, TimedOut: false, Err: nil}
```
aleksandra-malinowska: nit: I'd prefer to gather errors where the original checks are made, instead of duplicating the code and introducing the new case which basically means "we succeeded after all, but it's too late".

PR #1948, opened 2019-04-23
Add local copy of gophercloud in magnum cloud provider

### LINKED ISSUE #1945: Remove extra non-k8s dependencies of Magnum cloud provider
See the comment: https://github.com/kubernetes/autoscaler/pull/1911/files#r277652727

tghartland:
The magnum cloud provider is using some extra gophercloud components compared to what kubernetes/kubernetes vendors. The extra components should not have been vendored in this repository because it is not supported by the vendoring process.
Rather than adding an override for the vendoring, I have copied gophercloud into the cloud provider and reverted the old changes to the vendor directory. The gophercloud version was not changed.
Fixes #1945


tghartland:
The included code is failing all the `hack/verify-*` checks. It will need to be ignored somehow.

losipiuk:
As files are moved out of vendor directory the script checking license header is now complaining.
@MaciekPytel It seems we should exclude requirement for header for those files as they are imported from other project. I guess a license file in directory which contains the gophercloud files should be enough.

losipiuk:
@tghartland there currently is no nice way to specify that boilerplate checking should be skipped for a file. We just edit the `hack/boilerplate/boilerplate.py` extending the `skipped_dirs` list.
As the LICENSE file is already in `cluster-autoscaler/cloudprovider/magnum/gophercloud` it seems that extending the list and a comment in `boilerplate.py` should be enough.
Let's wait util tomorrow for opinion of @MaciekPytel, though.

MaciekPytel:
I agree with @losipiuk. Let's fix the presubmit and merge this.

/approve

PR #1950, opened 2019-04-24
Break getPatchesForPodResourceRequest into smaller functions

jbartosik:
Prepare for setting limits

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
@@ -26,6 +26,7 @@ import (
 
 	"k8s.io/api/admission/v1beta1"
 	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2"
 	metrics_admission "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/metrics/admission"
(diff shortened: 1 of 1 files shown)


vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
```diff
@@ -125,6 +96,53 @@ func (s *AdmissionServer) getPatchesForPodResourceRequest(raw []byte, namespace
...
+
+func getAddResourcesSubPatch(i int, kind string) patchRecord {
+	return patchRecord{
+		Op:    "add",
+		Path:  fmt.Sprintf("/spec/containers/%d/resources/%s", i, kind),
+		Value: v1.ResourceList{},
+	}
+}
+
+func getAddResourceParamPatch(i int, kind, resource v1.ResourceName, quantity resource.Quantity) patchRecord {
```
bskiba: nit: Shouldn't kind be explicitly a string?

vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
```diff
@@ -125,6 +96,53 @@ func (s *AdmissionServer) getPatchesForPodResourceRequest(raw []byte, namespace
 	return patches, nil
 }
 
+func getAddResourcesPatches(i int) patchRecord {
```
bskiba: getAddResourcesPatches is initializes resource requirements to an empty object, so maybe put "empty" explicitly in the function name. Same for getAddResourcesSubpatch. Then the comments in getContainerPatch would become unneeded.

vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
bskiba: getAddResourceRequirementValuePatch?

vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
jbartosik: Done.

jbartosik: Done.

Issue #1515, opened 2018-12-18
Incorrect handling of condition based taints if TaintNodesByCondition is used


losipiuk:
According to @aermakov-zalando in https://github.com/kubernetes/kubernetes/issues/72129
CA misbehaves if TainNodesByCondition is in use:

> The autoscaler itself doesn't treat these conditions in a special way, from what I understand by looking at the source code. This means that if the tolerations are not there, it'll spin up a new node, determine that the pods > can't run on it because of taints and happily create a new node instead of waiting for the node to be ready.

That needs to be verified and fixed if needed.

aermakov-zalando:
We're migrating to 1.12, and this actually doesn't happen in practice. It looks like the actual state of unready nodes is ignored and a template node is substituted instead in https://github.com/kubernetes/autoscaler/blob/7f6344ef2a48fa64c9c0ae5a257f642eabdc0de5/cluster-autoscaler/core/scale_up.go#L288, so it works correctly.

sanyu:
When using taints as node conditions, wouldn't the node still be **_Ready_** (at least for things like disk pressure, memory pressure...)?
Adding these taints doesn't doesn't change the status of the node and the user is allowed to ignore some of the Node's problems by adding the appropriate tolerations.

I added a random taint to a node in an ASG, this node was used as a template by the autoscaler so no Pods where able to run in this group (they didn't have the appropriate toleration), but the autoscaler decided not to spin up a new node in this group as it assumes every node will be identical to this special tainted node.

We haven't enabled TaintNodesByCondition yet, but based on this, I don't think this will work as expected, if the cluster-autoscaler happens to use a node that has been tainted as a template (the node is still **Ready** in some cases), will it decide that the entire group has this taint and therefore spinning up a new node is useless?

SleepyBrett:
related: https://github.com/kubernetes/autoscaler/issues/1350

aermakov-zalando:
@sanyu Good point! We run a fork that handles the templates more predictably, which is why we haven't noticed.

vikaschoudhary16:
is not #1963 a complete fix for this issue?

aermakov-zalando:
Yeah, #1963 looks like it fix the issue. This needs verification, however, and since we took a different approach in our internal fork, someone would need to test it.

mvz27:
Have same issue in AWS cluster autoscaler 1.14.5.
Should this issue be fixed ??

losipiuk:
[...]

PR #1963, opened 2019-04-24
add --ignore-taint flag and ignore taints added by TaintNodesByCondition

### LINKED ISSUE #1350: [Cluster Autoscaler] --IgnoreTaint=sometaint proposal.
For our clusters we use Immutable nodes and therefore certain changes require us to roll over our node pools. We currently use the following system.

1) We apply the new node config to the cloud provider, assuring that new nodes that appear will have the new configuration.

2) We taint all the existing nodes in the pool with `recycle=true:NoSchedule` assuring that any pods that get drained don't just end up on another old node leading to constant leapfrogging.

3) We have a pod that looks for the recycle taint, when it finds one it drains the node and calls the cloud provider to terminate it.

The problem we are finding is that the cluster autoscale in this scenario will not actually scale up these node pools. My assumption is that since it sees all the nodes in the pool as tainted with `recycle=true:NoSchedule` it has decided that any node that comes from scaling that pool up will result in the taint. This makes a certain amount of sense.

I'm proposing a flag that allows us to pass taints that should be ignored by the cluster autoscaler for purposes of scaleup.

Our current workaround is to manually go into the cloud provider and scale the group up by 1 so that it will have one non-tainted "new configuration" to disrupt the behavior.

cjbradfield:
Fixes #1350, #1515

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -70,6 +70,7 @@ type StaticAutoscaler struct {
 	initialized             bool
 	// Caches nodeInfo computed for previously seen nodes
 	nodeInfoCache map[string]*schedulernodeinfo.NodeInfo
+	ignoredTaints taintKeySet
 }
 
 type staticAutoscalerProcessorCallbacks struct {
@@ -109,6 +110,13 @@ func NewStaticAutoscaler(
 		OkTotalUnreadyCount:       opts.OkTotalUnreadyCount,
 		MaxNodeProvisionTime:      opts.MaxNodeProvisionTime,
 	}
+
+	ignoredTaints := make(taintKeySet)
+	for _, taintKey := range opts.IgnoredTaints {
+		klog.V(4).Infof("Ignoring taint %s on all NodeGroups", taintKey)
+		ignoredTaints[taintKey] = true
+	}
+
 	clusterStateRegistry := clusterstate.NewClusterStateRegistry(autoscalingContext.CloudProvider, clusterStateConfig, autoscalingContext.LogRecorder, backoff)
 
 	scaleDown := NewScaleDown(autoscalingContext, clusterStateRegistry)
@@ -124,6 +132,7 @@ func NewStaticAutoscaler(
 		processorCallbacks:      processorCallbacks,
 		clusterStateRegistry:    clusterStateRegistry,
 		nodeInfoCache:           make(map[string]*schedulernodeinfo.NodeInfo),
+		ignoredTaints:           ignoredTaints,
 	}
 }
 
@@ -183,7 +192,7 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 	}
 
 	nodeInfosForGroups, autoscalerError := getNodeInfosForGroups(
-		readyNodes, a.nodeInfoCache, autoscalingContext.CloudProvider, autoscalingContext.ListerRegistry, daemonsets, autoscalingContext.PredicateChecker)
+		readyNodes, a.nodeInfoCache, autoscalingContext.CloudProvider, autoscalingContext.ListerRegistry, daemonsets, autoscalingContext.PredicateChecker, a.ignoredTaints)
 	if autoscalerError != nil {
 		return autoscalerError.AddPrefix("failed to build node infos for node groups: ")
 	}
@@ -320,7 +329,7 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 		scaleUpStart := time.Now()
 		metrics.UpdateLastTime(metrics.ScaleUp, scaleUpStart)
 
-		scaleUpStatus, typedErr = ScaleUp(autoscalingContext, a.processors, a.clusterStateRegistry, unschedulablePodsToHelp, readyNodes, daemonsets, nodeInfosForGroups)
+		scaleUpStatus, typedErr = ScaleUp(autoscalingContext, a.processors, a.clusterStateRegistry, unschedulablePodsToHelp, readyNodes, daemonsets, nodeInfosForGroups, a.ignoredTaints)
 
 		metrics.UpdateDurationFromStart(metrics.ScaleUp, scaleUpStart)
 
--- a/cluster-autoscaler/core/utils.go
+++ b/cluster-autoscaler/core/utils.go
@@ -38,6 +38,7 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/utils/glogx"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/gpu"
 	kube_util "k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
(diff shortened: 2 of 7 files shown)


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -109,6 +110,13 @@ func NewStaticAutoscaler(
 		OkTotalUnreadyCount:       opts.OkTotalUnreadyCount,
 		MaxNodeProvisionTime:      opts.MaxNodeProvisionTime,
 	}
+
+	ignoredTaints := make(taintKeySet)
+	for _, taintKey := range opts.IgnoredTaints {
+		klog.V(4).Info("Ignoring taint on all NodeGroups: ", taintKey)
```
losipiuk: nit: "Ignoring taint %s on all NodeGroups"

cluster-autoscaler/core/utils.go
```diff
@@ -372,13 +389,29 @@ func sanitizeTemplateNode(node *apiv1.Node, nodeGroup string) (*apiv1.Node, erro
...
+			continue
 		case deletetaint.ToBeDeletedTaint:
 			klog.V(4).Infof("Removing autoscaler taint when creating template from node %s", node.Name)
+			continue
 		case deletetaint.DeletionCandidateTaint:
 			klog.V(4).Infof("Removing autoscaler soft taint when creating template from node %s", node.Name)
+			continue
 		default:
-			newTaints = append(newTaints, taint)
+			break
```
losipiuk: Just drop the default section

cluster-autoscaler/core/utils.go
```diff
@@ -48,6 +49,20 @@ const (
 	ReschedulerTaintKey = "CriticalAddonsOnly"
 )
 
+var (
+	nodeConditionTaints = taintKeySet{
```
losipiuk: It would be great to automatically extend the set if new taits are defined by scheduler. 
Can we list them somehow - instead of enumerating each separately?

cluster-autoscaler/core/utils.go
```diff
@@ -48,6 +49,20 @@ const (
 	ReschedulerTaintKey = "CriticalAddonsOnly"
 )
 
+var (
```
losipiuk: We should consider adding a feature flag to disable/enable dropping standard taints.
I personally do not see a scenario where we would like to NOT drop them. But maybe I am missing something.
@MaciekPytel WDYT?

cluster-autoscaler/core/utils.go
```diff
@@ -291,7 +308,7 @@ func getNodeInfoFromTemplate(nodeGroup cloudprovider.NodeGroup, daemonsets []*ap
 	pods = append(pods, baseNodeInfo.Pods()...)
 	fullNodeInfo := schedulernodeinfo.NewNodeInfo(pods...)
 	fullNodeInfo.SetNode(baseNodeInfo.Node())
-	sanitizedNodeInfo, typedErr := sanitizeNodeInfo(fullNodeInfo, id)
+	sanitizedNodeInfo, typedErr := sanitizeNodeInfo(fullNodeInfo, id, nil)
```
MaciekPytel: Feels like we should also pass ignoredTaints here, no? You really shouldn't ignore a taint that is on your template, but if for some reason you do it would be inconsistent between scale-from-0 and normal scale-up.
I guess a possible use-case would be if your cloudprovider implementation automatically assumes some taint will be there, but in your cluster it won't, you may want to filter it out?

cluster-autoscaler/core/utils.go
MaciekPytel: I don't see any scenario for keeping those taints either. You should never use those specific taint keys for anything else, whether you use taint by condition or not. I don't think we need a flag to disable that logic (also - we don't have flag that would prevent CA from ignoring rescheduler taint, so there is precedent).

cluster-autoscaler/core/static_autoscaler.go
cjbradfield: OK, will change

cluster-autoscaler/core/utils.go
cjbradfield: Agreed, I didn't like writing this. Unfortunately the scheduler api (where these are defined) only defines these a string constants. Unless they/we add code in that repo, the only solution I could see would be to ignore the entire key prefix.

cluster-autoscaler/core/utils.go
cjbradfield: Yeah I thought about this too and arrived at the same rationale for not providing a flag.

cluster-autoscaler/core/utils.go
cjbradfield: My understanding is this call is used to sanitize the node info you get from the Node Group template (e.g. the one created from the taint tags on the ASG) for the scale-from-zero case. Presumably, you wouldn't have taints coming from that template you want to ignore.

That said, I don't mind adding it for consistency as it shouldn't hurt anything either.

cjbradfield:
Just a note, I don't mind back porting these change either. In fact, the clusters I am running are a mix of 1.12 and 1.11 so I have a vested interest in seeing these changes in those branches.

Thanks for the quick turnaround on the code review!

cluster-autoscaler/core/utils.go
cjbradfield: those prefixes for your reference would be:
node.kubernetes.io/.*
node.cloudprovider.kubernetes.io/.*

cluster-autoscaler/core/utils.go
cjbradfield: OK, will change

cluster-autoscaler/core/utils.go
cjbradfield: I didn't read your comment fully. I see that you already considered what I said. I'll make the change.

cjbradfield:
@losipiuk @MaciekPytel Please re-review when you get a chance.

cluster-autoscaler/core/utils.go
losipiuk: Seems to vast. Let's leave it.

cjbradfield:
@losipiuk Thanks! I've squashed the commits and I believe is ready for final review.

losipiuk:
/lgtm
/approve
Thanks!

nhoughto:
love to see it backported 🙏

losipiuk:
> love to see it backported

By principle we should not cherry-pick that. As this is not really a bugfix. But given that it is small and useful change we can make an exception. Could you prepare cherry-picks to cluster-autoscaler-release-{1.14,1.13} branches?

iMartyn:
Please can we get this backported, we have nodes that are tainted on a 1.12 cluster and the autoscaler is not able to scale up because it thinks we have capacity.

FWIW I think the flag is misleading and CAS should be ignoring taints by default (like pods do in kubernetes) and invert the logic and name of this flag, but that's less urgent than getting the functionality backported in my view.

nhoughto:
FWIW instead of waiting for backport we just rolled the dice on new version of autoscaler with older version of kube and haven’t had any issues with it. So you could try that?

PR #1967, opened 2019-04-26
Modify the info passed to ScaleDownStatusProcessor when empty nodes a…

towca:
…re deleted

Previously, if any of the nodes fails to delete, the processor gets
a ScaleDownError status. After this commit, it will get the list of
nodes that were successfully deleted.

### DIFF
--- a/cluster-autoscaler/core/scale_down.go
+++ b/cluster-autoscaler/core/scale_down.go
@@ -730,16 +730,22 @@ func (sd *ScaleDown) TryToScaleDown(allNodes []*apiv1.Node, pods []*apiv1.Pod, p
 	emptyNodes := getEmptyNodes(candidates, pods, sd.context.MaxEmptyBulkDelete, scaleDownResourcesLeft, sd.context.CloudProvider)
 	if len(emptyNodes) > 0 {
 		nodeDeletionStart := time.Now()
-		confirmation := make(chan errors.AutoscalerError, len(emptyNodes))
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/core/scale_down.go
```diff
@@ -911,32 +922,34 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, cp cloud
 					metrics.RegisterScaleDown(1, gpu.GetGpuTypeForMetrics(cp.GPULabel(), cp.GetAvailableGPUTypes(), nodeToDelete, nodeGroup), metrics.Unready)
 				}
 			}
-			confirmation <- deleteErr
+			confirmation <- emptyNodeDeletionConfirmation{node: node, err: deleteErr}
```
aleksandra-malinowska: Let's drop the 'empty' part from emptyNodeDeleteConfirmation (both the type and variables). It's sent for both empty and non-empty nodes.

towca: In my understanding, this method and this channel is only used for empty nodes, but no problem!

aleksandra-malinowska: Right, nevermind:/ Shorter name is still nice, but that's just a nit in this case.

towca: Done already :)

Issue #1936, opened 2019-04-21
Support for AWS r5ad. instances


casidiablo:
They are just like `r5a.` instances, but use NVMe SSD instead of EBS.

Right now the autoscaler is crashing because of lack of support for them:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x1406f8f]

goroutine 49 [running]:
k8s.io/autoscaler/cluster-autoscaler/cloudprovider/aws.(*AwsManager).buildNodeFromTemplate(0xc420adeb00, 0xc420f2b440, 0xc42341bb40, 0x0, 0x0, 0xc420f7d110)
	/gopath/src/k8s.io/autoscaler/cluster-autoscaler/cloudprovider/aws/aws_manager.go:241 +0x3df
k8s.io/autoscaler/cluster-autoscaler/cloudprovider/aws.(*AwsNodeGroup).TemplateNodeInfo(0xc4212cdcf0, 0xc420db5020, 0xc421b21e30, 0x61)
	/gopath/src/k8s.io/autoscaler/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go:308 +0x68
k8s.io/autoscaler/cluster-autoscaler/core.GetNodeInfosForGroups(0xc4221b7600, 0x40, 0x40, 0xc420db5020, 0x4137d60, 0xc420cdfb90, 0x415a8a0, 0xc4201bd000, 0xc4234c22c0, 0x8, ...)
	/gopath/src/k8s.io/autoscaler/cluster-autoscaler/core/utils.go:286 +0x443
k8s.io/autoscaler/cluster-autoscaler/core.(*StaticAutoscaler).RunOnce(0xc420f2b950, 0xbf27363d55b28006, 0x2766d29a2, 0x64687c0, 0x0, 0x0)
	/gopath/src/k8s.io/autoscaler/cluster-autoscaler/core/static_autoscaler.go:138 +0x36a
main.run(0xc42043c0a0)
	/gopath/src/k8s.io/autoscaler/cluster-autoscaler/main.go:302 +0x1dd
main.main.func2(0x4118ce0, 0xc4207c4300)
	/gopath/src/k8s.io/autoscaler/cluster-autoscaler/main.go:381 +0x2a
created by k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/client-go/tools/leaderelection.(*LeaderElector).Run
[...]

PR #1970, opened 2019-04-29
Add m5ad, r5ad and t3a instance families

Jeffwan:
Add m5ad, r5ad and t3a instance families.  Address #1936 

Will take some time later to find a way to dynamically load latest instance in runtime. This is not elegant that every time we need to add new instance types


sc250024:
When is this getting released? Currently none of the latest CA releases support the new `t3a` instances types.

Jeffwan:
@sc250024 This is included in 1.15 branch. Do you need t3a instance in older branches? I can cherry-pick back to other branches as well

sc250024:
@Jeffwan Thanks for your reply, my comment above sounds a bit bitchy, sorry for that 😅

Anyway, if you could cherry pick the changes into the 1.13 and 1.14 branches, that would be very much desired. Case in point, the latest version of EKS on AWS only supports up to 1.13 at the latest, with 1.14 coming soon.

Thank you sir!

Jeffwan:
Sure. I will cherry-pick the change back to 1.12-1.14. BTW, this won't release immediately, release team has it's schedule.

Issue #1976, opened 2019-04-30
Ability to pause the autoscaler based on a tag in the AutoScaling Group


achinthagunasekara:
It would be really nice to have a feature to pause node autoscaler based on a tag in the autoscaling group (AWS) and have the ability to pass in the tag when node autoscaler starts.

E.G `--pause-tags=pause-node-autoscaler`

Then we should be able to add the tag `pause-node-autoscaler` to the ASG and have autoscaler pause any autoscaling activities.

MaciekPytel:
Why not just just temporarily disable autoscaling for that ASG? How is this different?

achinthagunasekara:
@MaciekPytel
Issue with that is, we'd have to update your Kube deployment for Autoscaler every time we roll the ASG. With this approach, it's all handled by adding/removing a tag to the ASG.

aleksandra-malinowska:
Isn't it already possible when using autodiscovery? From [FAQ](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/aws#auto-discovery-setup):

> It is recommended to use a second tag like k8s.io/cluster-autoscaler/<YOUR CLUSTER NAME> when k8s.io/cluster-autoscaler/enabled is used across many clusters to prevent ASGs from different clusters recognized as the node groups

This sounds like you should be able to enable/disable autoscaler just by changing the tags, without autoscaler restart?

MaciekPytel:
There are 2 parallel discussions on this (here and on https://github.com/kubernetes/autoscaler/pull/1977). Let's continue on https://github.com/kubernetes/autoscaler/pull/1977 to avoid duplication.

Jeffwan:
We have two PRs to help reduce API calls. #1935 #1841  Please check if you still have throttling issue or not using  autodiscovery

PR #1977, opened 2019-04-30
Ability to pause the autoscaler based on a tag in the AutoScaling Group

achinthagunasekara:
# Description

I created this issue earlier https://github.com/kubernetes/autoscaler/issues/1976 and this is my attempt at solving it for AWS cloud provider.

This adds a feature to pause node autoscaler based on a tag in the autoscaling group (AWS) and have the ability to pass in the tag when node autoscaler starts.

E.G --pause-tags=pause-node-autoscaler

In this example, then we should be able to add the tag `pause-node-autoscaler` to the ASG and have autoscaler pause any autoscaling activities.

## Backgroud

The reason I believe we need this feature is, when we cycle instances in an ASG (for deploying a new AMI for an example), Cluster Autoscaler interferes with that operation and terminates/launch instances during the update process. So we'd like to pause Cluster Autoscaler by adding a tag to the ASG and remove this tag when the update is complete.

### DIFF
--- a/cluster-autoscaler/cloudprovider/alicloud/alicloud_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/alicloud/alicloud_cloud_provider.go
@@ -230,3 +230,8 @@ func BuildAlicloud(opts config.AutoscalingOptions, do cloudprovider.NodeGroupDis
 	}
 	return cloudProvider
 }
+
+func (ali *aliCloudProvider) Paused() bool {
+	// Support to be added later
+	return false
+}
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
@@ -52,13 +52,15 @@ var (
 type awsCloudProvider struct {
 	awsManager      *AwsManager
 	resourceLimiter *cloudprovider.ResourceLimiter
+	pauseTag        string
 }
 
 // BuildAwsCloudProvider builds CloudProvider implementation for AWS.
-func BuildAwsCloudProvider(awsManager *AwsManager, resourceLimiter *cloudprovider.ResourceLimiter) (cloudprovider.CloudProvider, error) {
+func BuildAwsCloudProvider(awsManager *AwsManager, resourceLimiter *cloudprovider.ResourceLimiter, pauseTag string) (cloudprovider.CloudProvider, error) {
 	aws := &awsCloudProvider{
 		awsManager:      awsManager,
 		resourceLimiter: resourceLimiter,
+		pauseTag:        pauseTag,
 	}
 	return aws, nil
 }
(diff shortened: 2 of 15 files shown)


bskiba:
Thank you for the contribution. 
Please fix failing test and squash commits into 1 or 2 for easier review.

/assign @MaciekPytel

MaciekPytel:
As I asked on the issue you created - what's the benefit of this pausing mechanism over just disabling autoscaling for a given NodeGroup?

achinthagunasekara:
@MaciekPytel 
Issue with that is, we'd have to update your Kube deployment for Autoscaler every time we roll the ASG. With this approach, it's all handled by adding/removing a tag to the ASG.

MaciekPytel:
Why not use [autodiscovery][1]? Seems like using it you could disable autoscaling for a nodegroup using tags. 

[1]: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md#auto-discovery-setup

achinthagunasekara:
We also looked at it :) The was it was loading up the AWS API casing other API calls to throttle.

MaciekPytel:
Sorry, I didn't get what was the problem with autodiscovery. Can you explain in more detail? Is it querying the API too much? It seems like there is no conceptual difference between what you're trying to achieve and disabling autoscaling. I'd much rather we fix whatever problem you have with existing mechanism than add a new overlapping one. Especially if it involves extending cloudprovider interface, which is something we try to avoid as much as possible and only do once we figure there is no other option.

achinthagunasekara:
No worries :)

The issue we have is that we have a large number of Autoscaling groups in one AWS account. Also we have several EKS clusters running inside this account (each running Cluster Autoscaler).
With auto discovery it seems to query All ASGs inside the region every time and filter them based on the given tags. Several clusters doing this seems to load up the AWS API.

With the proposed solution, we no longer need to look through all ASGs and filter them to find relevant ASGs, since ASG names are passed into the Cluster Autoscaler. We only need to check if the provided `pauseTag` is present in the given ASG.

achinthagunasekara:
@MaciekPytel any feedback/suggestions on this? :)

achinthagunasekara:
@MaciekPytel any feedback/suggestions on this? :)

Jeffwan:
@achinthagunasekara  I agree with @MaciekPytel point, actually we have two PRs to help reduce API calls. https://github.com/kubernetes/autoscaler/pull/1935 https://github.com/kubernetes/autoscaler/pull/1841

Can you check if these two changes can resolve the throttling issue? If so, that might be easier

vikaschoudhary16:
Can priority expander be an alternate solution, https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/expander/priority/readme.md#priority-based-expander-for-cluster-autoscaler ?

On the run, configmap can be changed to remove the ASG that we dont want to be enabled. Is this correct?

achinthagunasekara:
I think auto discovery seems to be the best bet at this stage.
`priority expander` won't work for our use case, because that means my scripts that's cycling the ASGs needs knowledge of Kube environments to be able to touch config maps. I'd like to keep this script as low level as possible so all it needs to know is that it's just cycling a ASG and modify some ASG tags.

PR #1978, opened 2019-04-30
Receive v1beta2 VPAs

jbartosik:


bskiba:
The master branch does not support v1beta1 already, so I think we can just switch instead of supporting both.

jbartosik:
/hold
It needs also to subscribe for updates on v1beta2 VPAs.

jbartosik:
> The master branch does not support v1beta1 already, so I think we can just switch instead of supporting both.

Done.

Issue #1795, opened 2019-03-14
AWS Ondemand not scaled up if Spot requests remain "Open"


sc250024:
Greetings,

I'm running cluster-autoscaler v1.3.8 in a Kubernetes v1.11.8 cluster in AWS created using Kops. I'm not sure if this falls under feature request, a bug, or a misconfiguration on my part, so apologies in advance if it's a misconfiguration. Basically, for our pre-production cluster, I want to run spot instances as much as possible. I have the following node groups (and thus, ASGs), and they look like the following if the cluster is running normally:

```text
ASG Name                   Instances  Desired  Min  Max
nodes-ondemand-eu-west-1a  0          0        0    10
nodes-ondemand-eu-west-1b  0          0        0    10
nodes-ondemand-eu-west-1c  0          0        0    10
nodes-spot-eu-west-1a      2          2        2    10
nodes-spot-eu-west-1b      2          2        2    10
nodes-spot-eu-west-1c      2          2        2    10
```

I also have [k8s-spot-rescheduler](https://github.com/pusher/k8s-spot-rescheduler) running to take pods from any `ondemand` nodes that are provisioned, and move them to spot instances so that the `ondemand` nodes can be removed. However, lately, there always seems to be spot requests that are open, but not yet fulfilled:

<img width="1130" alt="Screen Shot 2019-03-14 at 14 55 19" src="https://user-images.githubusercontent.com/8792243/54362601-7f22ac80-4669-11e9-963a-61d8d019c97c.png">

This is normal behavior in AWS, but the problem is, the cluster autoscaler *does not* use the `ondemand` node groups at all. The following log output is what I'll see during this situation:

```
cluster-autoscaler aws_manager.go:148] Refreshed ASG list, next refresh after 2019-03-14 13:59:04.235594224 +0000 UTC m=+241976.799746087
cluster-autoscaler clusterstate.go:542] Readiness for node group nodes-spot-eu-west-1a.my-k8s-cluster.com not found
cluster-autoscaler utils.go:541] No pod using affinity / antiaffinity found in cluster, disabling affinity predicate for this loop
cluster-autoscaler static_autoscaler.go:274] No unschedulable pods
cluster-autoscaler utils.go:498] Skipping ip-10-YY-XXX-ZZZ.eu-west-1.compute.internal - node group min size reached
cluster-autoscaler utils.go:498] Skipping ip-10-YY-XXX-ZZZ.eu-west-1.compute.internal - node group min size reached
[...]

PR #1980, opened 2019-05-02
Stop waiting for upcoming nodes from unhealthy node groups

mvisonneau:
This is an attempt to solve a long going issue, preventing CA from failing over healthy node groups when a scale-up occurred onto a non-healthy one. This is particularly painful whilst working with spot based ASGs which are very prone to these kinds of disruptions.

References in #1795 and #1133

if people want to easily try it out, I published a docker release: `docker.io/mvisonneau/cluster-autoscaler:1.14.2-fix_scale_up_unavail_node_groups`

it seems to work perfectly fine for my use case :

```
Events:
  Type     Reason            Age                    From                Message
  ----     ------            ----                   ----                -------
  Normal   TriggeredScaleUp  52m (x7 over 54m)      cluster-autoscaler  pod triggered scale-up: [{sb1-in-k8s-worker-spot-r5.2xlarge-b 0->1 (max: 20)}]
  Warning  FailedScheduling  4m10s (x195 over 60m)  default-scheduler   0/9 nodes are available: 6 Insufficient cpu, 9 Insufficient memory.
  Normal   TriggeredScaleUp  52s                    cluster-autoscaler  pod triggered scale-up: [{sb1-in-k8s-worker-spot-m5.4xlarge-b 0->1 (max: 20)}]
```

### DIFF
--- a/cluster-autoscaler/core/scale_up.go
+++ b/cluster-autoscaler/core/scale_up.go
@@ -288,16 +288,20 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 	}
 
 	upcomingNodes := make([]*schedulernodeinfo.NodeInfo, 0)
-	for nodeGroup, numberOfNodes := range clusterStateRegistry.GetUpcomingNodes() {
-		nodeTemplate, found := nodeInfos[nodeGroup]
+	for nodeGroupID, numberOfNodes := range clusterStateRegistry.GetUpcomingNodes() {
+		nodeTemplate, found := nodeInfos[nodeGroupID]
 		if !found {
 			return &status.ScaleUpStatus{Result: status.ScaleUpError}, errors.NewAutoscalerError(
 				errors.InternalError,
 				"failed to find template node for node group %s",
-				nodeGroup)
+				nodeGroupID)
 		}
-		for i := 0; i < numberOfNodes; i++ {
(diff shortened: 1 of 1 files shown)


MaciekPytel:
I don't think this solves the issue. A single failed scale-up won't make a NodeGroup unhealthy, it will likely take multiple retries before it gets there. And just because a NodeGroup is unhealthy doesn't mean a node that is already being created will fail. 

There is already plenty of logic to deal with this scenario. I think the problem here is that AWS cloudprovider implementation doesn't use it. I've put some details in https://github.com/kubernetes/autoscaler/issues/1996#issuecomment-490422582.

mvisonneau:
Indeed @MaciekPytel, I did eventually end up with this issue :( I'll try to have a closer look following your comments.

Jeffwan:
> There is already plenty of logic to deal with this scenario. I think the problem here is that AWS cloudprovider implementation doesn't use it. I've put some details in [#1996 (comment)](https://github.com/kubernetes/autoscaler/issues/1996#issuecomment-490422582).

Hi @MaciekPytel Anywhere I can check for the logic you mentioned? I'd love to add to AWS cloudprovider side.

MaciekPytel:
The implementation is already in progress in https://github.com/kubernetes/autoscaler/pull/2008 - let's continue discussion there.

PR #1983, opened 2019-05-03
Added documentation about AWS ASG MixedInstancesPolicy

drewhemm:
Ref https://github.com/kubernetes/autoscaler/pull/1886

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/README.md
+++ b/cluster-autoscaler/cloudprovider/aws/README.md
@@ -175,6 +175,42 @@ If you'd like to scale node groups from 0, an `autoscaling:DescribeLaunchConfigu
 }
 ```
 
+## Using AutoScalingGroup MixedInstancesPolicy
+
+It is possible to use Cluster Autoscaler with a [mixed instances policy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-autoscaling-autoscalinggroup-mixedinstancespolicy.html), to enable diversification across on-demand and spot instances, of multiple instance types in a single ASG. When using spot instances, this increases the likelihood of successfully launching a spot instance to add the desired capacity to the cluster versus a single instance type, which may be in short supply.
+
+Note that the instance types should have the same amount of RAM and number of CPU cores, since this is fundamental to CA's scaling calculations. Using mismatched instances types can produce unintended results.
+
+Additionally, there are other factors which affect scaling, such as node labels. If you are currently using `nodeSelector` with the [beta.kubernetes.io/instance-type](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#interlude-built-in-node-labels) label, you will need to apply a common propagating label to the ASG and use that instead, since the instance-type label can no longer be relied upon. One may also use auto-generated tags such as `aws:cloudformation:stack-name` for this purpose. [Node affinity and anti-affinity](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity) are not affected in the same way, since these selectors natively accept multiple values; one must add all the configured instances types to the list of values, for example:
+
+```yaml
+spec:
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+        - matchExpressions:
+          - key: beta.kubernetes.io/instance-type
+            operator: In
+            values:
+            - r5.2xlarge
+            - r5d.2xlarge
+            - i3.2xlarge
+            - r5a.2xlarge
+            - r5ad.2xlarge
+```
+
+### Example usage:
+
+* Create a [Launch Template](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-autoscaling-autoscalinggroup-launchtemplate.html) (LT) with an instance type, for example, r5.2xlarge. Consider this the 'base' instance type. Do not define any spot purchase options here.
+* Create an ASG with a MixedInstancesPolicy that refers to the newly-created LT.
+* Set LaunchTemplateOverrides to include the 'base' instance type r5.2xlarge and suitable alternatives, e.g. r5d.2xlarge, i3.2xlarge, r5a.2xlarge and r5ad.2xlarge. Differing processor types and speeds should be evaluated depending on your use-case(s).
+* Set [InstancesDistribution](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_InstancesDistribution.html) according to your needs.
+* See [Allocation Strategies](https://docs.aws.amazon.com/autoscaling/ec2/userguide/asg-purchase-options.htlm#asg-allocation-strategies) for information about the ASG fulfils capacity from the specified instance types.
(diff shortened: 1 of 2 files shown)


drewhemm:
Any thoughts on whether a CloudFormation json snippet should be added?

```json
{
    "Resources": {
        "LaunchTemplate": {
            "Type": "AWS::EC2::LaunchTemplate",
            "Properties": {
                "LaunchTemplateName": "memory-opt-2xlarge",
                "LaunchTemplateData": {
                    "InstanceType": "r5.2xlarge"
                }
            }
        },
        "ASGA": {
            "Type": "AWS::AutoScaling::AutoScalingGroup",
            "Properties": {
                "MinSize": 1,
                "MaxSize": 10,
                "MixedInstancesPolicy": {
                    "InstancesDistribution": {
                        "OnDemandBaseCapacity": 0,
                        "OnDemandPercentageAboveBaseCapacity": 0
                    },
                    "LaunchTemplate": {
                        "LaunchTemplateSpecification": {
                            "LaunchTemplateId": {
                                "Ref": "LaunchTemplate"
                            },
                            "Version": {
                                "Fn::GetAtt": [
                                    "LaunchTemplate",
                                    "LatestVersionNumber"
                                ]
                            }
                        },
                        "Overrides": [
                            {
                                "InstanceType": "r5.2xlarge"
                            },
                            {
                                "InstanceType": "i3.2xlarge"
                            },
                            {
                                "InstanceType": "r5a.2xlarge"
                            }
                        ]
                    }
                },
                "VPCZoneIdentifier": [
                    "subnet-###############"
                ],
            }
        },
        "ASGB": {},
        "ASGC": {}
    }
}
```

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -175,6 +175,21 @@ If you'd like to scale node groups from 0, an `autoscaling:DescribeLaunchConfigu
 }
 ```
 
+## AWS::AutoScaling::AutoScalingGroup MixedInstancesPolicy
```
Jeffwan: Could we use a different title? I think a template resource reference may not clear to users.

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -175,6 +175,21 @@ If you'd like to scale node groups from 0, an `autoscaling:DescribeLaunchConfigu
...
+
+It is possible to use Cluster Autoscaler with a [mixed instances policy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-autoscaling-autoscalinggroup-mixedinstancespolicy.html), to enable diversification across on-demand and spot instances, of multiple instance types in a single ASG. When using spot instances, this increases the likelihood of successfully launching a spot instance to add the desired capacity to the cluster versus a single instance type, which may be in short supply.
+
```
Jeffwan: Let's move  line 191-192 here to remind user has to use instance with same cpu, memory or GPU. otherwise, users have to take risk.

Jeffwan:
> Any thoughts on whether a CloudFormation json snippet should be added?
> 
> ```json
> {
>     "Resources": {
>         "LaunchTemplate": {
>             "Type": "AWS::EC2::LaunchTemplate",
>             "Properties": {
>                 "LaunchTemplateName": "memory-opt-2xlarge",
>                 "LaunchTemplateData": {
>                     "InstanceType": "r5.2xlarge"
>                 }
>             }
>         },
>         "ASGA": {
>             "Type": "AWS::AutoScaling::AutoScalingGroup",
>             "Properties": {
>                 "MinSize": 1,
>                 "MaxSize": 10,
>                 "MixedInstancesPolicy": {
>                     "InstancesDistribution": {
>                         "OnDemandBaseCapacity": 0,
>                         "OnDemandPercentageAboveBaseCapacity": 0
>                     },
>                     "LaunchTemplate": {
>                         "LaunchTemplateSpecification": {
>                             "LaunchTemplateId": {
>                                 "Ref": "LaunchTemplate"
>                             },
>                             "Version": {
>                                 "Fn::GetAtt": [
>                                     "LaunchTemplate",
>                                     "LatestVersionNumber"
>                                 ]
>                             }
>                         },
>                         "Overrides": [
>                             {
>                                 "InstanceType": "r5.2xlarge"
>                             },
>                             {
>                                 "InstanceType": "i3.2xlarge"
>                             },
>                             {
>                                 "InstanceType": "r5a.2xlarge"
>                             }
>                         ]
>                     }
>                 },
>                 "VPCZoneIdentifier": [
>                     "subnet-###############"
>                 ],
>             }
>         },
>         "ASGB": {},
>         "ASGC": {}
>     }
> }
> ```

That would be super helpful. What about create a separate doc? We can add Spot, MixedInstancePolicy content you submitted and CloudFormation example there and reference it in README.md

drewhemm:
> What about create a separate doc? We can add Spot, MixedInstancePolicy content you submitted and CloudFormation example there and reference it in README.md

Another markdown file with a different filename in the same directory?

Jeffwan:
> > What about create a separate doc? We can add Spot, MixedInstancePolicy content you submitted and CloudFormation example there and reference it in README.md
> 
> Another markdown file with a different filename in the same directory?

Yes. should be fine. If we have more docs in the future, we can have a `docs/` dir later.

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -175,6 +175,21 @@ If you'd like to scale node groups from 0, an `autoscaling:DescribeLaunchConfigu
 }
 ```
 
+## Using AutoScalingGroup MixedInstancesPolicy
+
+It is possible to use Cluster Autoscaler with a [mixed instances policy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-autoscaling-autoscalinggroup-mixedinstancespolicy.html), to enable diversification across on-demand and spot instances, of multiple instance types in a single ASG. When using spot instances, this increases the likelihood of successfully launching a spot instance to add the desired capacity to the cluster versus a single instance type, which may be in short supply.
+
+Note that the instance types should have the same amount of RAM and number of CPU cores, since these are the criteria used by CA when scaling.
```
MaciekPytel: This are not the only criteria, just the most commonly used ones. Fundamentally, CA works by assuming every node in ASG is identical in every way. Any difference between the nodes is potentially unsafe. 
If you use nodes with the same cpu/memory/gpu and don't use complex scheduling features (ex. nodeAffinity on system labels) you're _probably_ safe. 
No one is going to guarantee you're ever 100% safe when using mixed ASGs though - it all depends on the detail of how your cluster is configured. For every setup with mixed ASGs you can come up with a pod spec that would lead to autoscaler doing something bad.

Jeffwan: Thanks for the details. @drewhemm Could you help update to doc?

Jeffwan: @MaciekPytel Seems we have lots of use cases now and any discussion going on for this topic? I'd like to know anything we can do to make logic not violate core assumption of CA?

drewhemm: I believe the only impacted functionality relates to `nodeSelector`, (which will eventually be deprecated) and node affinity and anti-affinity:

https://kubernetes.io/docs/concepts/configuration/assign-pod-node/

I have updated the readme to reflect this.

PR #1986, opened 2019-05-06
Move CA version string to own package, for use by cloud providers

tghartland:
Moving the version string into its own package lets it be imported from other packages, which can be useful in the cloud providers.

The second commit in this PR uses the version string to set the user-agent for gophercloud in the magnum cloud provider, to provide more information in request logs. This is a goal of the OpenStack K8s SIG: https://etherpad.openstack.org/p/k8s-sig-ptg-train
```
- Enhancement User agent header patch - gopher cloud
    - append to the user agent the particular project you're working on
    - this is useful for logging and tracing and telemetry
```
The user-agent then looks like
```
cluster/20000bea-83fd-4e21-b6cd-1cd0ea07c157 cluster-autoscaler/1.14.0-beta.2 gophercloud/2.0.0
```
instead of just `gophercloud/2.0.0`.


strigazi:
LGTM

cc @jim-bach

jim-bach:
LGTM =D

PR #1995, opened 2019-05-07
Make "Full VPA" E2E tests stricter

kgolab:
Check that initial recommendation is provided for both CPU & memory and only then await for an updated recommendation on resource increase.

Also fixes expected value for CPU since the minimal recommendation is nowadays 25mCPU.
Uses large range for minimal memory recommendation to be future-proof in case we lower the default from current 250 MiB.


kgolab:
/assign @bskiba 
Please have a look.

PR #1997, opened 2019-05-07
When limit is set but request is not treat them as matching

jbartosik:
@kgolab pointed out that when container has limit and no request it's treated as if it had request equal to limit.

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
@@ -18,6 +18,7 @@ package logic
 
 import (
 	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/labels"
 
 	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2"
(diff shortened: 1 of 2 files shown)


vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
```diff
@@ -60,6 +61,19 @@ func NewRecommendationProvider(vpaLister vpa_lister.VerticalPodAutoscalerLister,
 	}
 }
 
+func limitMatchesRequest(limit, request *resource.Quantity) bool {
+	// Limit not set, it cannot match the request.
+	if limit == nil || limit.Value() == 0 {
+		return false
+	}
+	// Limit set but request not set - K8s interprets will treat the pod as if it had them equal.
```
kgolab: nit: grammar seems broken in "K8s interprets..."

jbartosik: Fixed.

kgolab:
LGTM, thanks.

bskiba:
So the pod with empty request and limit set is treated as guaranteed?

jbartosik:
> So the pod with empty request and limit set is treated as guaranteed?

Yes: https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/#what-if-you-specify-a-container-s-limit-but-not-its-request

bskiba:
Actually, the documentation you point to states that the pod will be modified to have request set, so AFAIU we should not actually in the admission controller see the case when request = 0 and limit > 0

jbartosik:
> Actually, the documentation you point to states that the pod will be modified to have request set, so AFAIU we should not actually in the admission controller see the case when request = 0 and limit > 0

In a 1.12 cluster I created a pod specifying limit but no request. Admission controller received a pod with limit = request.

PR #2000, opened 2019-05-08
Monitor diffs between usage and recommendation.

bskiba:
Initial stab at monitoring diffs between recommendation and actual usage. Putting this up for discussion.

This has some drawbacks:
- pollutes the ContainerState interface
- When no recommendation is provided, we don't notice (I will work on fixing this one later)

However thanks to this approach we record diff on every usage sample we get (including OOMs).

@kgolab @schylek let me know what you think.

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/model/vpa.go
+++ b/vertical-pod-autoscaler/pkg/recommender/model/vpa.go
@@ -127,9 +127,23 @@ func (vpa *Vpa) UseAggregationIfMatching(aggregationKey AggregateStateKey, aggre
 	}
 	if vpa.matchesAggregation(aggregationKey) {
 		vpa.aggregateContainerStates[aggregationKey] = aggregation
+		aggregation.IsUnderVPA = true
 		return true
 	}
 	return false
+}
+
+// UpdateRecommendation updates the recommended resources in the VPA and its
+// aggregations with the given recommendation.
+func (vpa *Vpa) UpdateRecommendation(recommendation *vpa_types.RecommendedPodResources) {
+	vpa.Recommendation = recommendation
+	for _, containerRecommendation := range recommendation.ContainerRecommendations {
+		for container, state := range vpa.aggregateContainerStates {
+			if container.ContainerName() == containerRecommendation.ContainerName {
+				state.LastRecommendation = containerRecommendation.UncappedTarget
+			}
+		}
+	}
 }
 
 // UsesAggregation returns true iff an aggregation with the given key contributes to the VPA.
@@ -140,6 +154,8 @@ func (vpa *Vpa) UsesAggregation(aggregationKey AggregateStateKey) bool {
 
 // DeleteAggregation deletes aggregation used by this container
 func (vpa *Vpa) DeleteAggregation(aggregationKey AggregateStateKey) {
+	state := vpa.aggregateContainerStates[aggregationKey]
+	state.MarkNotAutoscaled()
 	delete(vpa.aggregateContainerStates, aggregationKey)
 }
 
--- a/vertical-pod-autoscaler/pkg/recommender/model/vpa_test.go
+++ b/vertical-pod-autoscaler/pkg/recommender/model/vpa_test.go
@@ -20,7 +20,8 @@ import (
 	"testing"
 	"time"
 
-	core "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	labels "k8s.io/apimachinery/pkg/labels"
 	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2"
 	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/test"
@@ -60,7 +61,7 @@ func TestUpdateConditions(t *testing.T) {
 			expectedConditions: []vpa_types.VerticalPodAutoscalerCondition{
 				{
 					Type:    vpa_types.RecommendationProvided,
-					Status:  core.ConditionTrue,
+					Status:  corev1.ConditionTrue,
 					Reason:  "",
 					Message: "",
 				},
@@ -73,12 +74,12 @@ func TestUpdateConditions(t *testing.T) {
 			expectedConditions: []vpa_types.VerticalPodAutoscalerCondition{
 				{
 					Type:    vpa_types.RecommendationProvided,
-					Status:  core.ConditionTrue,
+					Status:  corev1.ConditionTrue,
 					Reason:  "",
 					Message: "",
 				}, {
 					Type:    vpa_types.NoPodsMatched,
-					Status:  core.ConditionTrue,
+					Status:  corev1.ConditionTrue,
 					Reason:  "NoPodsMatched",
 					Message: "No pods match this VPA object",
 				},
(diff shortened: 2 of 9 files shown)


schylek:
While you are on the topic - how about adding metrics between original request and applied request in admission-controller? This will help assess impact of VPA on the cluster.

kgolab:
Regarding the drawbacks:
* if you're worried about polluting ContainerState, you can move the logic to ClusterState.AddSample and .RecordOOM - I think all the required data should be readily available
* when exporting value on no recommendation - are you planning to add a label to distinguish these cases from "normal" diff?

I have a few questions about the code but not sure if it's worth asking them already since it's WIP.

bskiba:
> Regarding the drawbacks:
> 
> * if you're worried about polluting ContainerState, you can move the logic to ClusterState.AddSample and .RecordOOM - I think all the required data should be available there though maybe we'd need a map of Vpa objects indexed by PodID too
I know, but this has a drawback of keeping that map up to date. Currently we just move aggregateContainer states between VPA objects which is quite convenient.
> * when exporting value on no recommendation - are you planning to add a label to distinguish these cases from "normal" diff?
Yeah, most probably. Though I am wondering what value to export. Should it be +Inf? or maybe usage/request since if recommendation is missing it's the request that counts (though the request can be 0 as well). 
> 
> I have a few questions about the code but not sure if it's worth asking them already since it's WIP.
I'm fixing this up and will send for review today.

bskiba:
> While you are on the topic - how about adding metrics between original request and applied request in admission-controller? This will help assess impact of VPA on the cluster.

Great idea, will do (maybe on a followup PR)

bskiba:
PTAL

vertical-pod-autoscaler/pkg/recommender/model/vpa.go
```diff
@@ -127,11 +127,25 @@ func (vpa *Vpa) UseAggregationIfMatching(aggregationKey AggregateStateKey, aggre
 	}
 	if vpa.matchesAggregation(aggregationKey) {
 		vpa.aggregateContainerStates[aggregationKey] = aggregation
+		aggregation.IsUnderVPA = true
 		return true
 	}
 	return false
```
kgolab: I think we should set IsUnderVPA = false here in case someone deleted VPA object.

vertical-pod-autoscaler/pkg/recommender/model/vpa.go
```diff
@@ -140,6 +154,9 @@ func (vpa *Vpa) UsesAggregation(aggregationKey AggregateStateKey) bool {
 
 // DeleteAggregation deletes aggregation used by this container
 func (vpa *Vpa) DeleteAggregation(aggregationKey AggregateStateKey) {
+	state := vpa.aggregateContainerStates[aggregationKey]
```
kgolab: Why do you clear these two fields if delete is executed just afterwards?

vertical-pod-autoscaler/pkg/recommender/model/vpa_test.go
```diff
@@ -142,3 +142,166 @@ func TestUpdateConditions(t *testing.T) {
...
 }
+
+func TestUpdateRecommendation(t *testing.T) {
+	cases := []struct {
+		name           string
+		containers     []string
+		recommendation *vpa_types.RecommendedPodResources
+		expectedLast   map[string]corev1.ResourceList
+	}{
+		{
```
kgolab: A test I'm missing here is a recommendation update - calling UpdateRecommendation with data similar to 2nd TC on the output of 1st TC; we should check that the test-container has updated recommendation and second-container keeps the old one.

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
```diff
@@ -0,0 +1,69 @@
...
+	metricsNamespace = metrics.TopMetricsNamespace + "quality"
+)
+
+var (
+	usageRecommendationRelativeDiff = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: metricsNamespace,
+			Name:      "usage_recommendation_relative_diffs",
+			Help:      "Diffs between usage and recommendation, normalized by recommendation value",
+			Buckets: []float64{-1., -.75, -.5, -.25, -.1, -.05, -.005, -.001,
```
kgolab: Please add -.01 for symmetry.
I also wonder about adding +-0.002 & +-0.02, similar to .25 or even more detailed buckets; WDYT?

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
```diff
@@ -0,0 +1,69 @@
...
+
+const (
+	metricsNamespace = metrics.TopMetricsNamespace + "quality"
+)
+
+var (
+	usageRecommendationRelativeDiff = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: metricsNamespace,
+			Name:      "usage_recommendation_relative_diffs",
```
kgolab: Are you planning to also add absolute diffs (usage - recommendation) and recommendation values in some upcoming PR? It would give more information about the observed diff values.

kgolab:
Thanks, this looks quite good already. Few comments to be addressed.

vertical-pod-autoscaler/pkg/recommender/model/vpa.go
bskiba: Because if the aggregation stops being autoscaled I don't want to keep exporting the diff - the recommendation is bound to get stale, so I need to clear recommendation. I also don't want to mark as recommendation missing, so I clear IsUnderVPA.

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
bskiba: Yes, I have exactly this and it is almost ready, will send today :)

vertical-pod-autoscaler/pkg/recommender/model/vpa_test.go
bskiba: Good point, added.

vertical-pod-autoscaler/pkg/recommender/model/vpa.go
bskiba: Actually I added this to Delete VPA in cluster.go

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
bskiba: Added .01
As for more detailed buckets in the < 0 part, I'm not sure it is that interesting, I'm more invested in the usage > recommendation part, but I can add more.

kgolab: I would keep the buckets symmetric (at least as far as they can be kept that way).
What I meant by added 0.002 and 0.02 boundaries was to get more precise information, jump from 1 to 5 seems big.

bskiba: Got it now. Done.

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
```diff
@@ -0,0 +1,69 @@
...
+			Help:      "Diffs between usage and recommendation, normalized by recommendation value",
+			Buckets: []float64{-1., -.75, -.5, -.25, -.1, -.05, -0.025, -.01, -.005, -0.0025 - .001,
```
kgolab: -0.0025 - .001, - something wrong happened here
And we lack 0, I think.

bskiba: Indeed. Fixed. Thanks for the attentive review :)

kgolab:
Thanks, LGTM.

PR #2005, opened 2019-05-09
Preserve limit / request ratio, don't cap recommendations to limit

jbartosik:
The test was failing because it recommendation was higher than the limit so it got capped. With changes I'm making to how VPA handles limits capping to limit doesn't make much sense.

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
@@ -17,9 +17,12 @@ limitations under the License.
 package logic
 
 import (
+	"fmt"
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/labels"
+	"math"
+	"math/big"
 
 	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2"
 	vpa_lister "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/client/listers/autoscaling.k8s.io/v1beta2"
@@ -61,21 +64,38 @@ func NewRecommendationProvider(vpaLister vpa_lister.VerticalPodAutoscalerLister,
 	}
 }
 
-func limitMatchesRequest(limit, request *resource.Quantity) bool {
-	// Limit not set, it cannot match the request.
-	if limit == nil || limit.Value() == 0 {
-		return false
+func getProportionalLimit(originalLimit, originalRequest, recommendedRequest *resource.Quantity) (limit *resource.Quantity, capped bool) {
+	// originalLimit not set, don't set limit.
+	if originalLimit == nil || originalLimit.Value() == 0 {
+		return nil, false
 	}
-	// Limit set but request not set - K8s will treat the pod as if they were equal.
-	if request == nil || request.Value() == 0 {
-		return true
+	// originalLimit set but originalRequest not set - K8s will treat the pod as if they were equal,
+	// recommend limit equal to request
+	if originalRequest == nil || originalRequest.Value() == 0 {
+		result := *recommendedRequest
+		return &result, false
 	}
-	// Limit and request are set. Check if they are equal.
-	return request.Value() == limit.Value()
+	// originalLimit and originalRequest are set. If they are equal recommend limit equal to request.
+	if originalRequest.MilliValue() == originalLimit.MilliValue() {
+		result := *recommendedRequest
+		return &result, false
+	}
+
+	// Input and output milli values should fit in int64 but intermediate values might be bigger.
+	originalMilliRequest := big.NewInt(originalRequest.MilliValue())
+	originalMilliLimit := big.NewInt(originalLimit.MilliValue())
+	recommendedMilliRequest := big.NewInt(recommendedRequest.MilliValue())
+	var recommendedMilliLimit big.Int
+	recommendedMilliLimit.Mul(recommendedMilliRequest, originalMilliLimit)
+	recommendedMilliLimit.Div(&recommendedMilliLimit, originalMilliRequest)
+	if recommendedMilliLimit.IsInt64() {
+		return resource.NewMilliQuantity(recommendedMilliLimit.Int64(), recommendedRequest.Format), false
+	}
+	return resource.NewMilliQuantity(math.MaxInt64, recommendedRequest.Format), true
 }
 
-// getContainersResources returns the recommended resources for each container in the given pod in the same order they are specified in the pod.Spec.
-func getContainersResources(pod *v1.Pod, podRecommendation vpa_types.RecommendedPodResources) []ContainerResources {
+// GetContainersResources returns the recommended resources for each container in the given pod in the same order they are specified in the pod.Spec.
+func GetContainersResources(pod *v1.Pod, podRecommendation vpa_types.RecommendedPodResources, annotations vpa_api_util.ContainerToAnnotationsMap) []ContainerResources {
 	resources := make([]ContainerResources, len(pod.Spec.Containers))
 	for i, container := range pod.Spec.Containers {
 		resources[i] = newContainerResources()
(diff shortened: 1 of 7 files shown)


jbartosik:
I'm going to check that this doesn't break anything else, marking as WIP until then.

bskiba:
Couple issues:
- This is no longer "fix and e2e test" PR - it removes a significant chunk of VPA logic
- This PR removes the logic  for capping recommendation to limit, but the case when request < recommendation is yet unhandled. So what will happen is:
1. Pod has request < limit
2. Pod grows it's usage
3. Pods recommendation grows, everything is ok until it is lower that limit
4. Recommendation grows over limit
4. Recommendations are not capped to limit and we create pods with request > limit and they are rejected
I don't think this is the behavior we want.

jbartosik:
Fixed, I think.

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
```diff
@@ -61,17 +64,36 @@ func NewRecommendationProvider(vpaLister vpa_lister.VerticalPodAutoscalerLister,
...
+
+	// Input and output milli values should fit in int64 but intermediate values might be bigger.
+	originalMilliRequest := big.NewInt(originalRequest.MilliValue())
+	originalMilliLimit := big.NewInt(originalLimit.MilliValue())
+	recommendedMilliRequest := big.NewInt(recommendedRequest.MilliValue())
+	var recommendedMilliLimit big.Int
+	recommendedMilliLimit.Mul(recommendedMilliRequest, originalMilliLimit)
+	recommendedMilliLimit.Div(&recommendedMilliLimit, originalMilliRequest)
+	if recommendedMilliLimit.IsInt64() {
+		fmt.Printf("recommendedMilliLimit: %d (%d * %d / %d)\n", recommendedMilliLimit.Int64(), recommendedMilliRequest.Int64(), originalMilliLimit.Int64(), originalMilliRequest.Int64())
```
bskiba: Leftover debug?

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
```diff
@@ -61,17 +64,36 @@ func NewRecommendationProvider(vpaLister vpa_lister.VerticalPodAutoscalerLister,
...
-		return true
+	// originalLimit and originalRequest are set. If they are equal recommend limit equal to request.
+	if originalRequest.MilliValue() == originalLimit.MilliValue() {
+		return recommendedRequest
 	}
-	// Limit and request are set. Check if they are equal.
-	return request.Value() == limit.Value()
+
+	// Input and output milli values should fit in int64 but intermediate values might be bigger.
+	originalMilliRequest := big.NewInt(originalRequest.MilliValue())
```
bskiba: Wouldn't it be safer (without risking values > maxint) to compute actual ratio and then scale by ratio?

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
```diff
@@ -61,17 +64,36 @@ func NewRecommendationProvider(vpaLister vpa_lister.VerticalPodAutoscalerLister,
 	}
 }
 
-func limitMatchesRequest(limit, request *resource.Quantity) bool {
-	// Limit not set, it cannot match the request.
-	if limit == nil || limit.Value() == 0 {
-		return false
+func limitMatchesRequest(originalLimit, originalRequest, recommendedRequest *resource.Quantity) *resource.Quantity {
```
bskiba: Function name does not match the contents any more.

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
bskiba: Does it make sense to take MilliValue for memory? I think the Value() is in bytes

bskiba: I seem to recall that the k8s resource types have some nifty Scale() function (I may be wrong though)

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
jbartosik: Removed.

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
jbartosik: > Wouldn't it be safer (without risking values > maxint) to compute actual ratio and then scale by ratio?

It wouldn't be safer - user can still set request to 1 millibyte and limit to 1GB (for limit/request ratio of about 2^40). If VPA recommends limit of 1 GB limit recommendation will be about 2^80 milli bytes (2^30 bytes request * 2^10 (converting to milli units) * 2^40 ratio), well out of int64 range, regardless how we calculate the value.

But I think it doesn't really matter - we can't really give more than 2^64 bytes to anything anyways.

jbartosik: > I seem to recall that the k8s resource types have some nifty Scale() function (I may be wrong though)

[I can't find it](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/vendor/k8s.io/apimachinery/pkg/api/resource/quantity.go).

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
jbartosik: I don't think it hurts to use millibytes for memory. It limits max value this code can work with to ~2^54 bytes, orders of  magnitude more than we might need to use those years.

In the meantime it simplifies code somewhat.

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
```diff
@@ -61,17 +63,35 @@ func NewRecommendationProvider(vpaLister vpa_lister.VerticalPodAutoscalerLister,
...
+	originalMilliLimit := big.NewInt(originalLimit.MilliValue())
+	recommendedMilliRequest := big.NewInt(recommendedRequest.MilliValue())
+	var recommendedMilliLimit big.Int
+	recommendedMilliLimit.Mul(recommendedMilliRequest, originalMilliLimit)
+	recommendedMilliLimit.Div(&recommendedMilliLimit, originalMilliRequest)
+	if recommendedMilliLimit.IsInt64() {
+		return resource.NewMilliQuantity(recommendedMilliLimit.Int64(), recommendedRequest.Format)
+	}
+	klog.Warningf(
+		"Trying to keep proportion between original millilimit of %s and original millirequest of %s with recommended millirequest of %s the recommended millilimit of %s doesn't fit in int64, capping to MaxInt64",
```
bskiba: nit: the milli... makes this really long and convoluted. How about:
Failed to keep limit to request proportion of %s to %s with recommended request of %s milliUnits; doesn't fit in int64. Capping limit to MaxInt64

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
```diff
@@ -61,17 +63,35 @@ func NewRecommendationProvider(vpaLister vpa_lister.VerticalPodAutoscalerLister,
...
-		return false
+func getProportionalLimit(originalLimit, originalRequest, recommendedRequest *resource.Quantity) *resource.Quantity {
+	// originalLimit not set, don't set limit.
+	if originalLimit == nil || originalLimit.Value() == 0 {
+		return nil
+	}
+	// originalLimit set but originalRequest not set - K8s will treat the pod as if they were equal,
+	// recommend limit equal to request
+	if originalRequest == nil || originalRequest.Value() == 0 {
+		return recommendedRequest
```
bskiba: Maybe return a copy?

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
bskiba: How about we calculate request / limit and if that is < epsilon we decide that this is an error, and actually not move the limit at all (instead of setting to infinity)?

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
jbartosik: Setting to infinity is our best guess at users intent here. I added an annotation as you suggested.

bskiba:
/lgtm
/approve

Please leave a mental note to add documentation before we release. Thanks!

Issue #1263, opened 2018-09-20
Reaction on "failed to launch a new EC2 instance" event


Dregathar:
Hello,
Our application uses different types of instances as k8s nodes: c5 and i3. Pods running on c5 instances (let's name them 'core pods') use EBS volumes, so these pods are bound to a particular AZ. Pods running on i3 instances (let's name them 'compute pods') don't have EBS volumes and can run in any AZ. However, the application performs better if both core pods and compute pods run in the same AZ. The problem with i3 instances is that they might be lacking in some AZ's at some points in time (I personally saw 2 cases when i3.2xlarge instances were not available in a given AZ, and there was a recommendation in AWS console to create instances in other AZ's). Ideally we want to run compute pods in the same AZ as core pods, but in case when i3 instances are not available in the AZ, it is fine to run instances in another AZ.
To actually configure this scenario and to test how autoscaler would behave, I've created 2 node groups:

```
nodes-1a
  machineType: i3.xlarge
  maxSize: 10
  minSize: 0
  nodeLabels:
    instance_macro_group: computenodes
    kops.k8s.io/instancegroup: nodes-1a
	
nodes-1b
  machineType: i3.2xlarge
  maxSize: 10
  minSize: 0
  nodeLabels:
    instance_macro_group: computenodes
    kops.k8s.io/instancegroup: nodes-1b
```

Then I've created a deployment with the template spec containing the following:
  
```
    spec:
      affinity:
        nodeAffinity:
          preferredDuringSchedulingIgnoredDuringExecution:
          - preference:
              matchExpressions:
              - key: kops.k8s.io/instancegroup
                operator: In
                values:
                - nodes-1a
            weight: 1
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: instance_macro_group
                operator: In
                values:
                - computenodes
```
				
and container memory has the following configuration:
```
        resources:
          requests:
            memory: 25Gi
```

nodes-1a instance group has maxSize = 10, however, the AWS account has a limit on number of running i3.xlarge instances = 6 (I'm not able to reproduce lacking of instances of a particular type, so, using instance limits fits the purpose of the test)

[...]

PR #2008, opened 2019-05-10
fix: correctly handle lack of capacity of AWS spot ASGs

### LINKED ISSUE #1996: cluster autoscaler failed to scale up when AWS couldn't start a new instance
We're using recent `cluster-autoscaler`, built from the `master` branch. It's used with AWS EKS and k8s 1.12.7. We're using it to run a cluster based on AWS spot instances.

Recently we had a situation, where CAS failed to scale up the cluster. During scale-up, 2 ASGs were selected to increase the capacity, but AWS had no spot capacity in them, so no new instances were really started (they were stuck in a state like `requested instances = 5, running instances = 0`). We're using `--max-node-provision-time=10m` option, but new nodes in other ASGs (which had capacity) were not started in this time. The CAS was stuck in this state for over an hour.

We found the following in CAS logs:
```
...
I0501 12:50:26.228744 1 scale_up.go:264] Pod blah-r-ebbf1d5c18a2f81565-7489bfcc6c-7zxp2 is unschedulable
...
I0501 12:50:26.228818 1 scale_up.go:303] Upcoming 3 nodes
...
Pod cg-service-r-1885076a5de3ca3fee-55f44f94f4-tcvcr can't be scheduled on central-eks-ondemand-workers-asg-r4.4xlarge, predicate failed: PodFitsResources predicate mismatch, reason: Insufficient ephemeral-storage
...
I0501 12:50:26.228846 1 scale_up.go:341] Skipping node group central-eks-canary-asg-r5.12xlarge - max size reached
...
I0501 12:50:26.265802 1 scale_up.go:411] No need for any nodes in central-eks-workers-asg-m4.16xlarge
I0501 12:50:26.359310 1 scale_up.go:411] No need for any nodes in central-eks-workers-asg-m5.12xlarge
...
W0501 12:50:26.359343 1 scale_up.go:325] Node group central-eks-workers-asg-m5a.12xlarge is not ready for scaleup - unhealthy
```

As you can see, we had some groups that couldn't be used (unhealthy were the once where AWS had no spot capacity), but some of them were perfectly fine ("the ones with "No need for any nodes...").

This situation is hard to reproduce, but can someone please help me review the code?

If an instance is not coming up for more than `MaxNodeProvisonTime`, it is added to the `LongUnregistered` counter [here](https://github.com/kubernetes/autoscaler/blob/e05ceb51e1217914bb46c6600c889084fcadc077/cluster-autoscaler/clusterstate/clusterstate.go#L562).
Then, when calculating upcoming nodes, `LongUnregistered`, is subtracted, as can be seen in [clusterstate.go](https://github.com/kubernetes/autoscaler/blob/e05ceb51e1217914bb46c6600c889084fcadc077/cluster-autoscaler/clusterstate/clusterstate.go#L910), which is used in [scale_up.go](https://github.com/kubernetes/autoscaler/blob/e05ceb51e1217914bb46c6600c889084fcadc077/cluster-autoscaler/core/scale_up.go#L291).
In our case `newNodes := ar.CurrentTarget - (readiness.Ready + readiness.Unready + readiness.LongNotStarted + readiness.LongUnregistered)` should correctly evaluate to 0, which means no new nodes are coming. 
Yet, our logs show "3 upcoming nodes".

Is it possible, that we had `CurrentTarget = X, Ready = Unready = LongNotStarted = LongUnregistered = 0`, which set the value to `0`, but later it was updated to `CurrentTarget = X, Ready = Unready = 0, LongNotStarted = LongUnregistered = X`, the result was negative and the check prevent the counter from updating?

piontec:
This fixes a situation, where AWS cloud provider is used and at least one of the managed ASGs is using spot instances. Instances requested from such ASG can never come up due to too low bid or lack of capacity in spot pool. RIght now, this triggers a situation, where CAS considers such new nodes as "coming up", because `comig_up_nodes = requested - real`. As there's no instance identity to track, timeout mechanism for startup time won't kick in.

This patch introduces artificial placeholder IDs for instances requested from any ASG. If such instance won't be available before the timeout kicks in, the size of the ASG is decreased and the ASG group is marked as "backoff".

Fixes #1996 and most probably also #1133 and #1795

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
+++ b/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
@@ -19,6 +19,7 @@ package aws
 import (
 	"fmt"
 	"reflect"
+	"regexp"
 	"strings"
 	"sync"
 
@@ -30,7 +31,10 @@ import (
 	"k8s.io/klog"
 )
 
-const scaleToZeroSupported = true
+const (
+	scaleToZeroSupported          = true
+	placeholderInstanceNamePrefix = "i-placeholder-"
+)
 
 type asgCache struct {
 	registeredAsgs []*asg
@@ -195,6 +199,10 @@ func (m *asgCache) SetAsgSize(asg *asg, size int) error {
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
 
+	return m.setAsgSizeNoLock(asg, size)
+}
+
+func (m *asgCache) setAsgSizeNoLock(asg *asg, size int) error {
 	params := &autoscaling.SetDesiredCapacityInput{
 		AutoScalingGroupName: aws.String(asg.Name),
 		DesiredCapacity:      aws.Int64(int64(size)),
@@ -210,6 +218,10 @@ func (m *asgCache) SetAsgSize(asg *asg, size int) error {
 	asg.curSize = size
 
 	return nil
+}
+
+func (m *asgCache) decreaseAsgSizeByOneNoLock(asg *asg) error {
+	return m.setAsgSizeNoLock(asg, asg.curSize-1)
 }
 
 // DeleteInstances deletes the given instances. All instances must be controlled by the same ASG.
@@ -239,22 +251,34 @@ func (m *asgCache) DeleteInstances(instances []*AwsInstanceRef) error {
 	}
 
 	for _, instance := range instances {
-		params := &autoscaling.TerminateInstanceInAutoScalingGroupInput{
-			InstanceId:                     aws.String(instance.Name),
-			ShouldDecrementDesiredCapacity: aws.Bool(true),
-		}
-		resp, err := m.service.TerminateInstanceInAutoScalingGroup(params)
-		if err != nil {
-			return err
+		// check if the instance is a placeholder - a requested instance that was never created by the node group
+		// if it is, just decrease the size of the node group, as there's no specific instance we can remove
+		if m.isPlaceholderInstance(instance) {
+			klog.V(4).Infof("instance %s is detected as a placeholder, decreasing ASG requested size instead "+
+				"of deleting instance", instance.Name)
+			m.decreaseAsgSizeByOneNoLock(commonAsg)
+		} else {
+			params := &autoscaling.TerminateInstanceInAutoScalingGroupInput{
+				InstanceId:                     aws.String(instance.Name),
+				ShouldDecrementDesiredCapacity: aws.Bool(true),
+			}
+			resp, err := m.service.TerminateInstanceInAutoScalingGroup(params)
+			if err != nil {
+				return err
+			}
+			klog.V(4).Infof(*resp.Activity.Description)
 		}
 
 		// Proactively decrement the size so autoscaler makes better decisions
 		commonAsg.curSize--
-
-		klog.V(4).Infof(*resp.Activity.Description)
 	}
-
 	return nil
+}
+
+// isPlaceholderInstance checks if the given instance is only a placeholder
+func (m *asgCache) isPlaceholderInstance(instance *AwsInstanceRef) bool {
+	matched, _ := regexp.MatchString(fmt.Sprintf("^%s\\d+$", placeholderInstanceNamePrefix), instance.Name)
+	return matched
 }
 
 // Fetch automatically discovered ASGs. These ASGs should be unregistered if
@@ -323,6 +347,11 @@ func (m *asgCache) regenerate() error {
 		return err
 	}
 
+	// If currently any ASG has more Desired than running Instances, introduce placeholders
+	// for the instances to come up. This is required to track Desired instances that
+	// will never come up, like with Spot Request that can't be fulfilled
+	groups = m.createPlaceholdersForDesiredNonStartedInstances(groups)
+
 	// Register or update ASGs
 	exists := make(map[AwsRef]bool)
 	for _, group := range groups {
@@ -353,6 +382,27 @@ func (m *asgCache) regenerate() error {
 	m.asgToInstances = newAsgToInstancesCache
 	m.instanceToAsg = newInstanceToAsgCache
 	return nil
+}
+
+func (m *asgCache) createPlaceholdersForDesiredNonStartedInstances(groups []*autoscaling.Group) []*autoscaling.Group {
+	for _, g := range groups {
+		desired := *g.DesiredCapacity
+		real := int64(len(g.Instances))
+		if desired <= real {
+			continue
+		}
+
+		for i := real; i < desired; i++ {
+			id := fmt.Sprintf("%s%d", placeholderInstanceNamePrefix, i)
+			klog.V(4).Infof("Instance group %s has only %d instances created while requested count is %d. "+
+				"Creating placeholder instance with ID %s.", *g.AutoScalingGroupName, real, desired, id)
+			g.Instances = append(g.Instances, &autoscaling.Instance{
+				InstanceId:       &id,
+				AvailabilityZone: g.AvailabilityZones[0],
+			})
+		}
+	}
+	return groups
 }
 
 func (m *asgCache) buildAsgFromAWS(g *autoscaling.Group) (*asg, error) {
--- a/cluster-autoscaler/cloudprovider/aws/aws_manager.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_manager.go
@@ -277,7 +277,10 @@ func (m *AwsManager) SetAsgSize(asg *asg, size int) error {
 
 // DeleteInstances deletes the given instances. All instances must be controlled by the same ASG.
 func (m *AwsManager) DeleteInstances(instances []*AwsInstanceRef) error {
-	return m.asgCache.DeleteInstances(instances)
+	if err := m.asgCache.DeleteInstances(instances); err != nil {
+		return err
+	}
+	return m.forceRefresh()
 }
 
 // GetAsgNodes returns Asg nodes.
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -274,6 +274,14 @@ func (csr *ClusterStateRegistry) updateScaleRequests(currentTime time.Time) {
 	csr.scaleDownRequests = newScaleDownRequests
 }
 
+// BackoffNodeGroup is used to force the specified nodeGroup to go into backoff mode, which
+// means it won't be used for scaling out temporarily
+func (csr *ClusterStateRegistry) BackoffNodeGroup(nodeGroup cloudprovider.NodeGroup, currentTime time.Time) {
+	csr.Lock()
+	defer csr.Unlock()
+	csr.backoffNodeGroup(nodeGroup, cloudprovider.OtherErrorClass, "cloudProviderError", currentTime)
+}
+
 // To be executed under a lock.
 func (csr *ClusterStateRegistry) backoffNodeGroup(nodeGroup cloudprovider.NodeGroup, errorClass cloudprovider.InstanceErrorClass, errorCode string, currentTime time.Time) {
 	nodeGroupInfo := csr.nodeInfosForGroups[nodeGroup.Id()]
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -254,7 +254,9 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 	unregisteredNodes := a.clusterStateRegistry.GetUnregisteredNodes()
 	if len(unregisteredNodes) > 0 {
 		klog.V(1).Infof("%d unregistered nodes present", len(unregisteredNodes))
-		removedAny, err := removeOldUnregisteredNodes(unregisteredNodes, autoscalingContext, currentTime, autoscalingContext.LogRecorder)
+		removedAny, err := removeOldUnregisteredNodes(unregisteredNodes, autoscalingContext, a.clusterStateRegistry,
+			currentTime, autoscalingContext.LogRecorder)
+		// There was a problem with removing unregistered nodes. Retry in the next loop.
 		if err != nil {
 			klog.Warningf("Failed to remove unregistered nodes: %v", err)
 		}
--- a/cluster-autoscaler/core/utils.go
+++ b/cluster-autoscaler/core/utils.go
@@ -484,7 +484,7 @@ func sanitizeTemplateNode(node *apiv1.Node, nodeGroup string, ignoredTaints tain
 
 // Removes unregistered nodes if needed. Returns true if anything was removed and error if such occurred.
 func removeOldUnregisteredNodes(unregisteredNodes []clusterstate.UnregisteredNode, context *context.AutoscalingContext,
-	currentTime time.Time, logRecorder *utils.LogEventRecorder) (bool, error) {
+	clusterStateRegistry *clusterstate.ClusterStateRegistry, currentTime time.Time, logRecorder *utils.LogEventRecorder) (bool, error) {
 	removedAny := false
 	for _, unregisteredNode := range unregisteredNodes {
 		if unregisteredNode.UnregisteredSince.Add(context.MaxNodeProvisionTime).Before(currentTime) {
@@ -514,6 +514,7 @@ func removeOldUnregisteredNodes(unregisteredNodes []clusterstate.UnregisteredNod
 					"Failed to remove node %s: %v", unregisteredNode.Node.Name, err)
 				return removedAny, err
 			}
+
 			logRecorder.Eventf(apiv1.EventTypeNormal, "DeleteUnregistered",
 				"Removed unregistered node %v", unregisteredNode.Node.Name)
 			removedAny = true
--- a/cluster-autoscaler/core/utils_test.go
+++ b/cluster-autoscaler/core/utils_test.go
@@ -451,12 +451,12 @@ func TestRemoveOldUnregisteredNodes(t *testing.T) {
 	assert.Equal(t, 1, len(unregisteredNodes))
 
 	// Nothing should be removed. The unregistered node is not old enough.
-	removed, err := removeOldUnregisteredNodes(unregisteredNodes, context, now.Add(-50*time.Minute), fakeLogRecorder)
+	removed, err := removeOldUnregisteredNodes(unregisteredNodes, context, clusterState, now.Add(-50*time.Minute), fakeLogRecorder)
 	assert.NoError(t, err)
 	assert.False(t, removed)
 
 	// ng1_2 should be removed.
-	removed, err = removeOldUnregisteredNodes(unregisteredNodes, context, now, fakeLogRecorder)
+	removed, err = removeOldUnregisteredNodes(unregisteredNodes, context, clusterState, now, fakeLogRecorder)
 	assert.NoError(t, err)
 	assert.True(t, removed)
 	deletedNode := getStringFromChan(deletedNodes)
--- a/cluster-autoscaler/cloudprovider/aws/aws_manager_test.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_manager_test.go
@@ -207,9 +207,16 @@ func TestFetchExplicitAsgs(t *testing.T) {
 		mock.AnythingOfType("func(*autoscaling.DescribeAutoScalingGroupsOutput, bool) bool"),
 	).Run(func(args mock.Arguments) {
 		fn := args.Get(1).(func(*autoscaling.DescribeAutoScalingGroupsOutput, bool) bool)
+		zone := "test-1a"
 		fn(&autoscaling.DescribeAutoScalingGroupsOutput{
 			AutoScalingGroups: []*autoscaling.Group{
-				{AutoScalingGroupName: aws.String(groupname)},
+				{
+					AvailabilityZones:    []*string{&zone},
+					AutoScalingGroupName: aws.String(groupname),
+					MinSize:              aws.Int64(int64(min)),
+					MaxSize:              aws.Int64(int64(max)),
+					DesiredCapacity:      aws.Int64(int64(min)),
+				},
 			}}, false)
 	}).Return(nil)
 
@@ -381,11 +388,14 @@ func TestFetchAutoAsgs(t *testing.T) {
 		mock.AnythingOfType("func(*autoscaling.DescribeAutoScalingGroupsOutput, bool) bool"),
 	).Run(func(args mock.Arguments) {
 		fn := args.Get(1).(func(*autoscaling.DescribeAutoScalingGroupsOutput, bool) bool)
+		zone := "test-1a"
 		fn(&autoscaling.DescribeAutoScalingGroupsOutput{
 			AutoScalingGroups: []*autoscaling.Group{{
+				AvailabilityZones:    []*string{&zone},
 				AutoScalingGroupName: aws.String(groupname),
 				MinSize:              aws.Int64(int64(min)),
 				MaxSize:              aws.Int64(int64(max)),
+				DesiredCapacity:      aws.Int64(int64(min)),
 			}}}, false)
 	}).Return(nil).Twice()
 
--- a/cluster-autoscaler/clusterstate/utils/node_instances_cache.go
+++ b/cluster-autoscaler/clusterstate/utils/node_instances_cache.go
@@ -102,8 +102,9 @@ func (cache *CloudProviderNodeInstancesCache) GetCloudProviderNodeInstances() (m
 			wg.Add(1)
 			go func() {
 				defer wg.Done()
-				_, err := cache.fetchCloudProviderNodeInstancesForNodeGroup(nodeGroup)
-				klog.Errorf("Failed to fetch cloud provider node instances for %v, error %v", nodeGroup.Id(), err)
+				if _, err := cache.fetchCloudProviderNodeInstancesForNodeGroup(nodeGroup); err != nil {
+					klog.Errorf("Failed to fetch cloud provider node instances for %v, error %v", nodeGroup.Id(), err)
+				}
 			}()
 		}
 	}


cluster-autoscaler/core/utils.go
```diff
@@ -411,10 +411,20 @@ func removeOldUnregisteredNodes(unregisteredNodes []clusterstate.UnregisteredNod
...
-					"Failed to remove node %s: %v", unregisteredNode.Node.Name, err)
-				return removedAny, err
+				_, wasPlaceholder := err.(*cloudprovider.PlaceholderDeleteError)
+				// this means only a placeholder instance was deleted - it is an instance, that was requested,
+				// but was not create before StartUpTimeout. It means something's wrong with this specific
+				// node group and we temporarily suspend requesting new instances from it by registering
+				// a failed scale up
+				if wasPlaceholder {
+					klog.Warningf("Timeout trying to scale node group %s, enabling backoff for the group", nodeGroup.Id())
+					clusterStateRegistry.RegisterFailedScaleUp(nodeGroup, metrics.Timeout, time.Now())
```
MaciekPytel: The failed scale-up logic should happen regardless, triggered from https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/clusterstate/clusterstate.go#L237. At minimum this will mess up metrics (you'll double count failed scale-ups), not sure if it breaks anything else.

MaciekPytel:
I think adding the placeholders alone is enough, all the changes outside AWS cloudprovider are redundant (and mess up metrics). The timeout/backoff mechanism has worked for years in GCP, all the core parts should be there already (it may not be obvious from code because in GCP placeholder instances are created by MIG, so they're already returned by API calls to GCP).

piontec:
Hmm, so if tracking scaleUpRequest is enough, why it didn't work in the first place? After scaleUpRequest times out, `backoffNodeGroup` should solve the problem...?

MaciekPytel:
`backoffNodeGroup` temporarily prevents CA from trying another scale-up on the same NodeGroup, but doesn't make it ignore already upcoming nodes. My guess (no way to test...) would be that NodeGroup was backedOff, but it didn't matter as you were already stuck with forever-upcoming nodes preventing any future scale-up.

piontec:
OK, that might be it. Let me remove the code in `utils.go` and retest - I completely missed that Request status update logic

aleksandra-malinowska:
> Hmm, so if tracking scaleUpRequest is enough, why it didn't work in the first place? After scaleUpRequest times out, backoffNodeGroup should solve the problem...?

I may be missing some context here, but I think I recall there was a problem with ASG immediately changing its target size back to actual size when the instance fails to start (see #1263). So scaleUpRequest never times out.

piontec:
> `backoffNodeGroup` temporarily prevents CA from trying another scale-up on the same NodeGroup, but doesn't make it ignore already upcoming nodes. My guess (no way to test...) would be that NodeGroup was backedOff, but it didn't matter as you were already stuck with forever-upcoming nodes preventing any future scale-up.

OK, I think I know better now what is going on. Indeed, after scaling up a node group that won't increase its size the `Long*` counters are not updated, so on every check CAS computes the upcoming node as `Desired - Running`, which gives a positive number and keeps other scaling groups from scaling up.

Still, my first proposal is (almost) good. The problem is that at some point the node group needs to be scaled down by the `Increase` amount of the `ScaleUpRequest`. In my patch, I did that 1-by-1: every time a node startup time timeout was triggered, I was decreasing the size of node group and returning error indicating a backoff is necessary. 

The above approach is needed, as we can't rely on `ScaleUpRequest` to timeout. Imagine a situation, where CAS crashes and node group state is already "Wanted = 6, Running = 4". Now, CAS starts up and sees 2 nodes coming up in the node group, but there's no `ScaleUpRequest` object. So, the only thing we can rely on is node startup timeout, after which we have to immediately trigger `backoff`, as otherwise the same node group will be used to scale up in the next iteration. So, the only problem in my original patch is double logging of failed metrics.

Am I right? What do you think?

mvisonneau:
Thanks a lot @piontec! I managed to test it and it seems to be working like a charm 👌

cluster-autoscaler/core/utils.go
```diff
@@ -411,10 +411,20 @@ func removeOldUnregisteredNodes(unregisteredNodes []clusterstate.UnregisteredNod
 			}
 			err = nodeGroup.DeleteNodes([]*apiv1.Node{unregisteredNode.Node})
 			if err != nil {
-				klog.Warningf("Failed to remove node %s: %v", unregisteredNode.Node.Name, err)
-				logRecorder.Eventf(apiv1.EventTypeWarning, "DeleteUnregisteredFailed",
-					"Failed to remove node %s: %v", unregisteredNode.Node.Name, err)
-				return removedAny, err
+				_, wasPlaceholder := err.(*cloudprovider.PlaceholderDeleteError)
```
losipiuk: I do not believe this is right approach to determine the nature of deleted instance based on error returned from `DeleteNodes()` (especially given the fact that `DeleteNodes` takes multiple nodes as an argument).

Instead we should based the behavior on what we got from `NodeGroup.Nodes()`.
Actually it seems you already have all the needed fields in the interface. The `NodeGroup.Nodes()` returns `InstanceStatus` for each node. And there is `InstanceState` field which can currently be one of `Instance{Running,Creating,Deleting}`. It seems to me that you want to backoff nodegroup if deleted node is in `InstanceCreating` state. If you think you would need to extend the `InstanceStatus` in some way (e.g. introduce more error classes or something - we can discuss it).

piontec: I don't think the above is right. We want to have feedback about the whole `DeleteNodes()`, we don't need this to be instance-by-instance. The thing is: if you requested 4 new instances, 4 came up successfully, but there's no spot capacity for the 5th, the ASG should be "backed-off" exactly the same as if no instance was started.
Of course, we can set `InstanceState` for each instance, then after the call iterate instances again to check if any is not coming up, but I'd like to introduce a new state for that, like `CreationTimeout`

losipiuk: > I don't think the above is right. We want to have feedback about the whole DeleteNodes(), we don't need this to be instance-by-instance. The thing is: if you requested 4 new instances, 4 came up successfully, but there's no spot capacity for the 5th, the ASG should be "backed-off" exactly the same as if no instance was started.

I agree that we want to backoff in such case.

> Of course, we can set InstanceState for each instance, then after the call iterate instances again to check if any is not coming up, but I'd like to introduce a new state for that, like CreationTimeout

`InstanceState` should still be `InstanceCreating`. Maybe the wording is not best but `InstanceCreating` means both:
* instance which is being created and we still do not know what will be outcome of that process
* instance which was created and creation result in a failure

For the latter case the `ErrorInfo` in the `InstanceStatus` is set to non-nil value. And IMO you should use this field to differentiate between successful and failed creation. It seems that you can still use the `OtherErrorClass` when you set `ErrorInfo`.

piontec: OK, I think I can have a look at that today. BTW, is someone of you going to KubeCon Barcelona next week?

losipiuk: THanks. I will be in BCN and bunch of other people from autoscaling team too.

piontec: There's a problem with this approach. It's not easy to update `Node.InstanceState` within aws provider, as it only keeps track of instances refs (a single string). This is common in the whole provider. Only when `Nodes()` is called, the array of strings (instance ids) is converted to `[]Nodes`, so there's no easy way to keep additional state of nodes without heavily changing interfaces and returned types of aws provider.

losipiuk: I understand. Yet I do not believe that changing signature a `DeleteNodes` is the way to go. I do not like the idea of making point changes to general interface, so it is somewhat less of an effort to support it for one specific implementation.

If exposing InstanceState for all the instances in AWS is problematic, maybe you can fairly easily set those to the placeholder nodes only?

piontec:
I did just that, please check now, although it's still not pretty

piontec:
ping @losipiuk @MaciekPytel , what do you think now?

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
```diff
@@ -30,7 +31,12 @@ import (
 	"k8s.io/klog"
 )
 
-const scaleToZeroSupported = true
+const (
+	scaleToZeroSupported          = true
+	placeholderInstanceNamePrefix = "i-placeholder-"
+	// TimeoutedPlaceholderName is used to mark placeholder instances that did not come up with timeout
+	TimeoutedPlaceholderName = "i-timeouted-placeholder"
```
seh: s/timeouted/timedout/

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
```diff
@@ -239,21 +253,38 @@ func (m *asgCache) DeleteInstances(instances []*AwsInstanceRef) error {
...
-		params := &autoscaling.TerminateInstanceInAutoScalingGroupInput{
-			InstanceId:                     aws.String(instance.Name),
-			ShouldDecrementDesiredCapacity: aws.Bool(true),
-		}
-		resp, err := m.service.TerminateInstanceInAutoScalingGroup(params)
-		if err != nil {
-			return err
+		// check if the instance is a placeholder - a requested instance that was never created by the node group
+		// if it is, just decrease the size of the node group, as there's no specific instance we can remove
+		matched, err := regexp.MatchString(fmt.Sprintf("^%s\\d+$", placeholderInstanceNamePrefix), instance.Name)
```
seh: We could use `regexp.Compile` to compile this regular expression just once, given that the prefix string is fixed. The resulting `regexp.Regex` value is safe for repeated and concurrent use.

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
```diff
@@ -239,21 +253,38 @@ func (m *asgCache) DeleteInstances(instances []*AwsInstanceRef) error {
...
+		if err == nil && matched {
+			klog.V(4).Infof("instance %s is detected as a placeholder, decreasing ASG requested size instead "+
+				"of deleting instance", instance.Name)
+			m.decreaseAsgSizeByOneNoLock(commonAsg)
+			// mark this instance using its name as a timeouted placeholder
```
seh: s/timeouted/timed out/

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
```diff
@@ -355,6 +391,27 @@ func (m *asgCache) regenerate() error {
...
+		desired := *g.DesiredCapacity
+		real := int64(len(g.Instances))
+		if desired <= real {
+			continue
+		}
+
+		for i := real; i < desired; i++ {
+			id := fmt.Sprintf("%s%d", placeholderInstanceNamePrefix, i)
+			klog.V(4).Infof("Instance group %s has only %d instances created while requested count is %d."+
+				"Creating placeholder instance with ID %s", *g.AutoScalingGroupName, real, desired, id)
```
seh: Either add a leading space before "Creating" or a trailing space after the period on the preceding line. As written, the two concatenated sentences will wind up with no intervening space character.

Also, punctuate the second sentence like the first one.

cluster-autoscaler/core/utils.go
```diff
@@ -416,6 +416,20 @@ func removeOldUnregisteredNodes(unregisteredNodes []clusterstate.UnregisteredNod
...
 				return removedAny, err
 			}
+
+			failedPlaceholder, err := anyPlaceholderInstanceStartupFailed(nodeGroup)
+			if err != nil {
+				return removedAny, err
+			}
+			if failedPlaceholder {
+				// this means only a placeholder instance was deleted - it is an instance, that was requested,
+				// but was not create before StartUpTimeout. It means something's wrong with this specific
```
seh: s/create/created/

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
```diff
@@ -239,21 +253,38 @@ func (m *asgCache) DeleteInstances(instances []*AwsInstanceRef) error {
...
+		// if it is, just decrease the size of the node group, as there's no specific instance we can remove
+		matched, err := regexp.MatchString(fmt.Sprintf("^%s\\d+$", placeholderInstanceNamePrefix), instance.Name)
+		if err == nil && matched {
+			klog.V(4).Infof("instance %s is detected as a placeholder, decreasing ASG requested size instead "+
+				"of deleting instance", instance.Name)
+			m.decreaseAsgSizeByOneNoLock(commonAsg)
+			// mark this instance using its name as a timeouted placeholder
+			asg := m.instanceToAsg[*instance]
+			delete(m.instanceToAsg, *instance)
+			instance.Name = TimeoutedPlaceholderName
```
losipiuk: I seems very much not right. 

IMO the instance should be marked as "timed-out" in some routine which refreshes node groups state. I guess in the same place when you are creating placeholders you should also monitor if some of the place holders are "old enough" to treat them as timed out. Then you can mark those as such and when instances are listed set `OutOfResources` error for them.

Then the DeleteInstances will be called by the core logic for timed-out instances. And here you will only remove the place holder, and decrease the asg size.

piontec: I get your point, but the problem is again in how AWS cloud provider works. Inside, it keeps instance IDs only and these IDs need to stay constant. It also doesn't track any timeouts - that's on the core's side, specifically in [clusterstate.go](https://github.com/kubernetes/autoscaler/blob/2543e3e5ea02867da4b40fca88976a9c575380ad/cluster-autoscaler/clusterstate/clusterstate.go#L288). The logic that tracks instances and handles registration timeouts is there. 

We either have to reuse/link this logic over here: [auto_scaling_groups.go](https://github.com/kubernetes/autoscaler/blob/2543e3e5ea02867da4b40fca88976a9c575380ad/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go#L306) or assume that if a `DeleteInstances` call comes for a placeholder, it's always because of timeout related to non-existing ScaleUp request. Basically, this whole logic is to guard against just a single use case: when CAS is started when already `required > running` for an ASG and the instances never come up.

I think there are the following possible cases for placeholder instances:
1) ScaleUp request times out. In that case, we don't have to modify Nodes state at all, as the time out handler will scale down the related ASG, mark it as "backed off" and decrease its size, which will make placeholders to be gone on the next refresh (that starts every main loop).
2) As I wrote above, on CAS start "required > running" already; in that case, `DeleteNodes` will be invoked by the Core logic (after detecting, that the node is listed by the cloud provider, but not registered in k8s) and we assume, that node is deleted because of time out. One possible modification is to explicitly pass "isTimeout" flag here.

What do you think?

seh:
I've been testing this patch in a cluster running Kubernetes version 1.13.4. I just saw the container crash like this:
```
I0531 14:40:04.609595       1 static_autoscaler.go:161] Starting main loop
[signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0x2164a15]

goroutine 99 [running]:
k8s.io/autoscaler/cluster-autoscaler/clusterstate.(*ClusterStateRegistry).updateReadinessStats(0xc0009dd7c0, 0xbf346d1124555c8c, 0x2c0acc8908, 0x4b7e140)
        /gopath/src/k8s.io/autoscaler/cluster-autoscaler/clusterstate/clusterstate.go:569 +0x9a5
k8s.io/autoscaler/cluster-autoscaler/clusterstate.(*ClusterStateRegistry).UpdateNodes(0xc0009dd7c0, 0xc00132d6c0, 0x5, 0x8, 0xc000aa2270, 0xbf346d1124555c8c, 0x2c0acc8908, 0x4b7e140, 0x0, 0x0)
        /gopath/src/k8s.io/autoscaler/cluster-autoscaler/clusterstate/clusterstate.go:303 +0x234
k8s.io/autoscaler/cluster-autoscaler/core.(*StaticAutoscaler).updateClusterState(0xc000a3c280, 0xc00132d6c0, 0x5, 0x8, 0xc000aa2270, 0xbf346d1124555c8c, 0x2c0acc8908, 0x4b7e140, 0xc00132d700, 0x8)
        /gopath/src/k8s.io/autoscaler/cluster-autoscaler/core/static_autoscaler.go:541 +0x94
k8s.io/autoscaler/cluster-autoscaler/core.(*StaticAutoscaler).RunOnce(0xc000a3c280, 0xbf346d1124555c8c, 0x2c0acc8908, 0x4b7e140, 0x0, 0x0)
        /gopath/src/k8s.io/autoscaler/cluster-autoscaler/core/static_autoscaler.go:191 +0x5a8
main.run(0xc0002db360)
        /gopath/src/k8s.io/autoscaler/cluster-autoscaler/main.go:322 +0x201
main.main.func2(0x2ef6780, 0xc00040cac0)
        /gopath/src/k8s.io/autoscaler/cluster-autoscaler/main.go:394 +0x2a
created by k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/client-go/tools/leaderelection.(*LeaderElector).Run
        /gopath/src/k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/client-go/tools/leaderelection/leaderelection.go:182 +0xec
```
I don't know yet whether that's related to these changes.

seh:
That failure above seems to happen within about two seconds after deleting a bunch of placeholder instances and reducing the desired count on the corresponding ASG. I can reproduce it easily.

seh:
If the line numbers are to be believed, it looks like the call to `(*awsCloudProvider).NodeGroupForNode` in `(*ClusterStateRegistry).updateReadinessStats`—at line 564 in file _cluster-autoscaler/clusterstate/clusterstate.go_—returns both a nil error and a nil `cloudprovider.NodeGroup`, as would happen if `(*AwsManager).GetAsgForInstance` returned nil.

Trying to call `cloudprovider.NodeGroup.ID` on such a nil value at line 569 would provoke the segmentation violation shown above.

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
losipiuk: > I get your point, but the problem is again in how AWS cloud provider works. Inside, it keeps instance IDs only and these IDs need to stay constant. It also doesn't track any timeouts - that's on the core's side, specifically in [clusterstate.go](https://github.com/kubernetes/autoscaler/blob/2543e3e5ea02867da4b40fca88976a9c575380ad/cluster-autoscaler/clusterstate/clusterstate.go#L288). The logic that tracks instances and handles registration timeouts is there.
> 
> We either have to reuse/link this logic over here: [auto_scaling_groups.go](https://github.com/kubernetes/autoscaler/blob/2543e3e5ea02867da4b40fca88976a9c575380ad/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go#L306) or assume that if a `DeleteInstances` call comes for a placeholder, it's always because of timeout related to non-existing ScaleUp request. Basically, this whole logic is to guard against just a single use case: when CAS is started when already `required > running` for an ASG and the instances never come up.

My understanding was that you wanted speed up the reaction time of CA when node insntances coming from spot-backed ASG are not showing up.
If this is just about fixing bug that the 15-minutes timeout mechanism built-in into CA does not trigger up situation is simpler (but also benefits are smaller).

> 1. ScaleUp request times out. In that case, we don't have to modify Nodes state at all, as the time out handler will scale down the related ASG, mark it as "backed off" and decrease its size, which will make placeholders to be gone on the next refresh (that starts every main loop).

Yes. Should work that way.

> 2. As I wrote above, on CAS start "required > running" already; in that case, `DeleteNodes` will be invoked by the Core logic (after detecting, that the node is listed by the cloud provider, but not registered in k8s) and we assume, that node is deleted because of time out. One possible modification is to explicitly pass "isTimeout" flag here.

If we just added placeholders and not modify code in any other way we would get to stable state after DeleteNodes are called. Just ASG would not be backed-off this time. Is that a problem? Do we care about (possibly rare) case when CA is restarted while scale-up is in progress? The state will get stable eventually. I do not believe we want to optimize for restart scenario. CA was never designed to be restarted very often. It survives restarts but it does not necessarily behaves optimally in such situations.

seh: Is that 15 minute timeout the one that's configured by the `--max-node-provision-time` command-line flag? There are several flags with default values of 15 minutes, and there's a const `MaxNodeStartupTime` with the same value. It doesn't look like that one can be changed without rebuilding the program.

15 minutes is too long for us to wait to react to an ASG's failure to acquire a machine, so we'd like to be able to dial this down.

seh:
I rebuilt the program using the latest f43101a5f418658c4aee331710d0275919e8966e commit, but it fails in the same way [reported above](https://github.com/kubernetes/autoscaler/pull/2008#issuecomment-497733797).

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
piontec: @losipiuk OK, I'll try to split this patch in 2: first provide a general fix, that has the worst case of recovering in `2 * NodeStartupTime`, then create another one, that can work in just `1 * NodeStartupTime`

@seh It's `--max-node-provision-time`. And yeah... and the general recommendation is "do not change that". Still, in my tests, I dialed it down to 2 minutes and it seemed to work. YMMV.

seh:
I added a diagnostic statement to the `(asgCache).findForInstance` method, as follows:
```go
func (m *asgCache) findForInstance(instance AwsInstanceRef) *asg {
	if asg, found := m.instanceToAsg[instance]; found {
		return asg
	}

	klog.Errorf("asgCache.findForInstance: no instance known for reference %q", instance.ProviderID)
	return nil
}
```
I also added log messages to the other functions along the suspect call path.

Right before the crash above, we see the following log messages:
```
I0603 15:18:51.773124       1 static_autoscaler.go:161] Starting main loop
E0603 15:18:51.773492       1 auto_scaling_groups.go:187] asgCache.findForInstance: no instance known for reference "aws:///us-east-1b/i-0ad71731e7ba8ba16"
E0603 15:18:51.773510       1 aws_cloud_provider.go:114] (*awsCloudProvider).NodeGroupForNode returning nil, nil for node "ip-10-105-51-26.ec2.internal"
E0603 15:18:51.773519       1 auto_scaling_groups.go:187] asgCache.findForInstance: no instance known for reference "aws:///us-east-1c/i-098187e4abb713cf1"
E0603 15:18:51.773523       1 aws_cloud_provider.go:114] (*awsCloudProvider).NodeGroupForNode returning nil, nil for node "ip-10-105-52-194.ec2.internal"
E0603 15:18:51.773703       1 auto_scaling_groups.go:187] asgCache.findForInstance: no instance known for reference "aws:///us-east-1a/i-01a4952e92aa98610"
E0603 15:18:51.773713       1 aws_cloud_provider.go:114] (*awsCloudProvider).NodeGroupForNode returning nil, nil for node "ip-10-105-49-120.ec2.internal"
E0603 15:18:51.775825       1 auto_scaling_groups.go:187] asgCache.findForInstance: no instance known for reference "aws:///us-east-1b/i-0ad71731e7ba8ba16"
E0603 15:18:51.775841       1 aws_cloud_provider.go:114] (*awsCloudProvider).NodeGroupForNode returning nil, nil for node "ip-10-105-51-26.ec2.internal"
E0603 15:18:51.775848       1 auto_scaling_groups.go:187] asgCache.findForInstance: no instance known for reference "aws:///us-east-1a/i-01a4952e92aa98610"
E0603 15:18:51.775851       1 aws_cloud_provider.go:114] (*awsCloudProvider).NodeGroupForNode returning nil, nil for node "ip-10-105-49-120.ec2.internal"
E0603 15:18:51.775864       1 auto_scaling_groups.go:187] asgCache.findForInstance: no instance known for reference "aws:///us-east-1c/i-098187e4abb713cf1"
E0603 15:18:51.775872       1 aws_cloud_provider.go:114] (*awsCloudProvider).NodeGroupForNode returning nil, nil for node "ip-10-105-52-194.ec2.internal"
E0603 15:18:51.775892       1 auto_scaling_groups.go:187] asgCache.findForInstance: no instance known for reference "aws:///us-east-1b/i-placeholder-5"
E0603 15:18:51.775899       1 aws_cloud_provider.go:114] (*awsCloudProvider).NodeGroupForNode returning nil, nil for node "aws:///us-east-1b/i-placeholder-5"
E0603 15:18:51.775906       1 clusterstate.go:570] No nodegroup retured for aws:///us-east-1b/i-placeholder-5
[signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0x2164b56]
```

piontec:
OK, this now only provides identity for failed ASG instances. Worst-case recovery time is 2*NodeStartupTime

seh:
@piontec, I'm trying to keep up. Thank you for iterating on this.

Testing your latest commit , it appears to work—reducing the size of the "stuck" ASGs—without crashing. It seems to take longer than the `--max-node-provision-time` command-line flag value I set (at two minutes), but it does work.

I do see log messages now like this, which I'd never noticed before:
```
I0603 15:46:15.058938       1 utils.go:393] Removing unregistered node aws:///us-east-1a/i-placeholder-0
W0603 15:46:15.058943       1 utils.go:409] Failed to remove node aws:///us-east-1a/i-placeholder-0: node group min size reached, skipping unregistered node removal
W0603 15:46:15.058953       1 static_autoscaler.go:250] Cluster is not ready for autoscaling
I0603 15:46:15.059089       1 event.go:209] Event(v1.ObjectReference{Kind:"ConfigMap", Namespace:"kube-system", Name:"cluster-autoscaler-status", UID:"fb5984ff-8615-11e9-aa89-0ecbc831ee82", APIVersion:"v1", ResourceVersion:"41809315", FieldPath:""}): type: 'Warning' reason: 'ClusterUnhealthy' Cluster is unhealthy
```
What does "not ready for autoscaling" mean? Why would the autoscaler believe that the cluster is unhealthy?

seh:
I think I spoke too soon. About five minutes later, the autoscaler tried to use those "stuck" ASGs again. I don't see a command-line flag available to tune how long these ASGs will stay "out of commission." In our clusters, the right value is usually on the order of days, not minutes.

piontec:
> @piontec, I'm trying to keep up. Thank you for iterating on this.
> 
> I0603 15:46:15.058938       1 utils.go:393] Removing unregistered node aws:///us-east-1a/i-placeholder-0
> W0603 15:46:15.058943       1 utils.go:409] Failed to remove node aws:///us-east-1a/i-placeholder-0: node group min size reached, skipping unregistered node removal
> What does "not ready for autoscaling" mean? Why would the autoscaler believe that the cluster is unhealthy?

Hmm, the strange part is where it thinks your ASG is already at min capacity. Are you forcing the test scenario by setting "min instances" to some value >0, but with a price that won't result in actual startup? If yes, try to set "min instances = 0", "max instances = N" and trigger the scenario with setting "desired" in [o, N] range.

piontec:
> Perhaps that was [`clusterstate.InitialNodeGroupBackoffDuration`](https://github.com/kubernetes/autoscaler/blob/62dcc2e81875bf5118ea591a02c8b26a711eae28/cluster-autoscaler/clusterstate/clusterstate.go#L53-L54) that we experienced, given the similarity in duration.

Yes, it probably was. Still, if the next try with the same ASG fails, the next backoff duration will be longer. Maybe we need to make "initial backoff duration" a config param?

seh:
> Hmm, the strange part is where it thinks your ASG is already at min capacity. Are you forcing the test scenario by setting "min instances" to some value >0, but with a price that won't result in actual startup?

No, these ASGs have a minimum size of zero. Their spot bid is indeed artificially set too low to force this scenario, where the autoscaler will try to use them, but then find that no instances arrive to fulfill the desired capacity.

seh:
> Still, if the next try with the same ASG fails, the next backoff duration will be longer.

I did figure that out after posting, seeing that there's a exponential backoff in use, with [the maximum capped at 30 minutes](https://github.com/kubernetes/autoscaler/blob/62dcc2e81875bf5118ea591a02c8b26a711eae28/cluster-autoscaler/clusterstate/clusterstate.go#L50-L51).

> Maybe we need to make "initial backoff duration" a config param?

I'd appreciate being able to tune both the initial duration and the maximum duration. For our common scenarios, I'd start it at around 20 minutes and let it creep up to a few hours.

seh:
Testing with the latest commit b7c879b3cd63495750c56066793ce533e5b05efb, it appears to work well managing the ASG sizes, though it still notes failure to remove these placeholders:
```
I0604 15:33:27.341742       1 static_autoscaler.go:161] Starting main loop
W0604 15:33:27.344164       1 clusterstate.go:601] Readiness for node group kubernetes-000-003-worker-ac-1-in-us-east-1a not found
W0604 15:33:27.344178       1 clusterstate.go:601] Readiness for node group kubernetes-000-003-worker-ac-1-in-us-east-1b not found
I0604 15:33:27.344203       1 static_autoscaler.go:229] 20 unregistered nodes present
I0604 15:33:27.344210       1 utils.go:393] Removing unregistered node aws:///us-east-1a/i-placeholder-1
W0604 15:33:27.344218       1 utils.go:409] Failed to remove node aws:///us-east-1a/i-placeholder-1: node group min size reached, skipping unregistered node removal
I0604 15:33:27.344222       1 utils.go:393] Removing unregistered node aws:///us-east-1c/i-placeholder-3
W0604 15:33:27.344227       1 utils.go:409] Failed to remove node aws:///us-east-1c/i-placeholder-3: node group min size reached, skipping unregistered node removal
I0604 15:33:27.344233       1 utils.go:393] Removing unregistered node aws:///us-east-1c/i-placeholder-4
W0604 15:33:27.344238       1 utils.go:409] Failed to remove node aws:///us-east-1c/i-placeholder-4: node group min size reached, skipping unregistered node removal
I0604 15:33:27.344241       1 utils.go:393] Removing unregistered node aws:///us-east-1a/i-placeholder-0
W0604 15:33:27.344246       1 utils.go:409] Failed to remove node aws:///us-east-1a/i-placeholder-0: node group min size reached, skipping unregistered node removal
I0604 15:33:27.344249       1 utils.go:393] Removing unregistered node aws:///us-east-1a/i-placeholder-6
W0604 15:33:27.344257       1 utils.go:409] Failed to remove node aws:///us-east-1a/i-placeholder-6: node group min size reached, skipping unregistered node removal
I0604 15:33:27.344261       1 utils.go:393] Removing unregistered node aws:///us-east-1c/i-placeholder-2
W0604 15:33:27.344269       1 utils.go:409] Failed to remove node aws:///us-east-1c/i-placeholder-2: node group min size reached, skipping unregistered node removal
I0604 15:33:27.344276       1 utils.go:393] Removing unregistered node aws:///us-east-1c/i-placeholder-5
W0604 15:33:27.344283       1 utils.go:409] Failed to remove node aws:///us-east-1c/i-placeholder-5: node group min size reached, skipping unregistered node removal
I0604 15:33:27.344304       1 utils.go:393] Removing unregistered node aws:///us-east-1a/i-placeholder-2
W0604 15:33:27.344314       1 utils.go:409] Failed to remove node aws:///us-east-1a/i-placeholder-2: node group min size reached, skipping unregistered node removal
I0604 15:33:27.344319       1 utils.go:393] Removing unregistered node aws:///us-east-1a/i-placeholder-5
W0604 15:33:27.344327       1 utils.go:409] Failed to remove node aws:///us-east-1a/i-placeholder-5: node group min size reached, skipping unregistered node removal
I0604 15:33:27.344331       1 utils.go:393] Removing unregistered node aws:///us-east-1c/i-placeholder-0
W0604 15:33:27.344339       1 utils.go:409] Failed to remove node aws:///us-east-1c/i-placeholder-0: node group min size reached, skipping unregistered node removal
I0604 15:33:27.344346       1 utils.go:393] Removing unregistered node aws:///us-east-1a/i-placeholder-3
W0604 15:33:27.344353       1 utils.go:409] Failed to remove node aws:///us-east-1a/i-placeholder-3: node group min size reached, skipping unregistered node removal
I0604 15:33:27.344360       1 utils.go:393] Removing unregistered node aws:///us-east-1a/i-placeholder-4
W0604 15:33:27.344369       1 utils.go:409] Failed to remove node aws:///us-east-1a/i-placeholder-4: node group min size reached, skipping unregistered node removal
I0604 15:33:27.344373       1 utils.go:393] Removing unregistered node aws:///us-east-1c/i-placeholder-1
W0604 15:33:27.344381       1 utils.go:409] Failed to remove node aws:///us-east-1c/i-placeholder-1: node group min size reached, skipping unregistered node removal
W0604 15:33:27.344398       1 static_autoscaler.go:250] Cluster is not ready for autoscaling
```
Is that indicative of an accounting problem?

seh:
@losipiuk, do you have an opinion on [my question posed here](https://github.com/kubernetes/autoscaler/pull/2008#issuecomment-498724588) concerning these warning messages?

I'm wondering what the next step is here. My testing shows that this patch improves the CA's behavior, getting itself out of the "stuck" condition that's plaguing our use of it today. How can I help bring this to completion?

piontec:
I'll try to reproduce and understand your problem, I hope I can do it tomorrow. Can you drop here your CAS configuration and ASG setup you're testing with?

seh:
> Can you drop here your CAS configuration and ASG setup you're testing with?

Here are the command-line flags we've used in our testing:
- `--v=4`
- `--stderrthreshold=info`
- `--cloud-provider=aws`
- `--skip-nodes-with-local-storage=false`
- `--expander=least-waste`
- `--node-group-auto-discovery=asg:tag=kubernetes.io/cluster-autoscaler/enabled,kubernetes.io/cluster/OUR_ID`
- `--max-node-provision-time=2m`
- `--scale-down-delay-after-add=3m`
- `--scale-down-unneeded-time=5m`
- `--scale-down-unready-time=7m`
- `--scale-down-non-empty-candidates-count=50`

In this test environment, we have nine ASGs, for two different instance types—one of them repeated with different Spot bids—across three availability zones. That is, three "group configurations," repeated so that each related ASG is confined to a single AZ, per the cluster autoscaler prescription.

Is that enough detail on the ASGs?

piontec:
> Is that enough detail on the ASGs?
Can you please check earlier logs, if these instances were successfully "removed"? The ASG should be scaled down and you should have an information about it. I think that in this case, this log just means that the ASG was scaled back down to 0 - should be harmless, yet the log message should be avoided

piontec:
@seh I replicated your problem and warning message and I think it should work now. Can you please check how it works for you now?

piontec:
@losipiuk @MaciekPytel can you have a look over here, please?

seh:
@piontec, so far my testing with commit f7465d5419b141de529dbc75893178ddfc889cb3 is going well. The logs are chatty, but none of the messages look to ominous like they did before.

seh:
After further testing, I'm satisfied with what I've seen of the change in behavior this patch brings.

I still think it would be better if [I could adjust the backoff parameters with command-line flags](https://github.com/kubernetes/autoscaler/pull/2008#issuecomment-498699026), but that's a separate debate and a separate change.

MaciekPytel:
@losipiuk Can you take this one? You're already familiar with it and I won't be able to look at it next week (travelling).

Jeffwan:
I will also put some time review and help test. Will come back by end of week.

seh:
@Jeffwan and @losipiuk, do you have any feedback to share?

Jeffwan:
@seh I did check codes and it looks good to me. But I did not get a chance to test it manually. Have you guys tried it? If it's working well for the case. I am ok to merge it.

seh:
We have indeed been running this version in our clusters for a couple of weeks, and have found that it mostly alleviates the problem with “stuck” ASGs.

The chattiness of the log messages about placeholder instances still makes me anxious, wondering if everything is working as it should, but the net observable behavior is what we asked for.

The backoff delay parameters could be exposed as configuration, but I understand that doing so would then require validating that they all work sensibly together, and we’d have to document several more command-line flags. We don’t need to take that on here. This patch is about enabling behavior that was already supposed to work this way, as opposed to introducing new features.

I hope @piontec agrees with my summary.

cluster-autoscaler/core/utils.go
```diff
@@ -484,7 +484,7 @@ func sanitizeTemplateNode(node *apiv1.Node, nodeGroup string, ignoredTaints tain
 
 // Removes unregistered nodes if needed. Returns true if anything was removed and error if such occurred.
 func removeOldUnregisteredNodes(unregisteredNodes []clusterstate.UnregisteredNode, context *context.AutoscalingContext,
-	currentTime time.Time, logRecorder *utils.LogEventRecorder) (bool, error) {
+	clusterStateRegistry *clusterstate.ClusterStateRegistry, currentTime time.Time, logRecorder *utils.LogEventRecorder) (bool, error) {
```
losipiuk: Please revert.

cluster-autoscaler/core/utils.go
```diff
@@ -514,6 +514,7 @@ func removeOldUnregisteredNodes(unregisteredNodes []clusterstate.UnregisteredNod
 					"Failed to remove node %s: %v", unregisteredNode.Node.Name, err)
 				return removedAny, err
 			}
+
```
losipiuk: plaese revert

cluster-autoscaler/core/utils_test.go
```diff
@@ -451,12 +451,12 @@ func TestRemoveOldUnregisteredNodes(t *testing.T) {
 	assert.Equal(t, 1, len(unregisteredNodes))
 
 	// Nothing should be removed. The unregistered node is not old enough.
-	removed, err := removeOldUnregisteredNodes(unregisteredNodes, context, now.Add(-50*time.Minute), fakeLogRecorder)
+	removed, err := removeOldUnregisteredNodes(unregisteredNodes, context, clusterState, now.Add(-50*time.Minute), fakeLogRecorder)
```
losipiuk: please revert

cluster-autoscaler/core/utils_test.go
```diff
@@ -451,12 +451,12 @@ func TestRemoveOldUnregisteredNodes(t *testing.T) {
...
 	assert.NoError(t, err)
 	assert.False(t, removed)
 
 	// ng1_2 should be removed.
-	removed, err = removeOldUnregisteredNodes(unregisteredNodes, context, now, fakeLogRecorder)
+	removed, err = removeOldUnregisteredNodes(unregisteredNodes, context, clusterState, now, fakeLogRecorder)
```
losipiuk: please revert

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -274,6 +274,14 @@ func (csr *ClusterStateRegistry) updateScaleRequests(currentTime time.Time) {
 	csr.scaleDownRequests = newScaleDownRequests
 }
 
+// BackoffNodeGroup is used to force the specified nodeGroup to go into backoff mode, which
+// means it won't be used for scaling out temporarily
+func (csr *ClusterStateRegistry) BackoffNodeGroup(nodeGroup cloudprovider.NodeGroup, currentTime time.Time) {
```
losipiuk: please revert

cluster-autoscaler/cloudprovider/aws/aws_manager.go
```diff
@@ -277,7 +277,10 @@ func (m *AwsManager) SetAsgSize(asg *asg, size int) error {
 
 // DeleteInstances deletes the given instances. All instances must be controlled by the same ASG.
 func (m *AwsManager) DeleteInstances(instances []*AwsInstanceRef) error {
-	return m.asgCache.DeleteInstances(instances)
+	if err := m.asgCache.DeleteInstances(instances); err != nil {
+		return err
+	}
+	return m.forceRefresh()
```
gjtempleton: It feels like we should either have a log message here stating that a refresh is being forced or update the message in `forceRefresh` as currently it could be confusing for users with the log message
in `forceRefresh` stating: `"Refreshed ASG list, next refresh after %v", m.lastRefresh.Add(refreshInterval))`

gjtempleton:
Awesome work, really glad to see someone finally getting around to resolving this.

One brief comment in addition to @losipiuk 's.

seh:
@piontec, do you anticipate being able to address @losipiuk's and @gjtempleton's suggestions soon?

seh:
I'm willing to take over this patch, and can handle the requests to revert a few lines. I'm not sure what's involved with writing end-to-end tests for this change in behavior.

What would the procedure be? Submit a fresh PR that starts from where this branch leaves off?

MaciekPytel:
Regarding e2e tests: I'm not exactly sure what you mean by that. Currently CA uses unittests and e2e tests in real cluster - the latter only exist for GCP. Obviously there is no way to cover this without porting them to AWS. I have no idea how much work that would involve (I expect a lot), but I don't think this PR should be gated on it.

Jeffwan:
> Regarding e2e tests: I'm not exactly sure what you mean by that. Currently CA uses unittests and e2e tests in real cluster - the latter only exist for GCP. Obviously there is no way to cover this without porting them to AWS. I have no idea how much work that would involve (I expect a lot), but I don't think this PR should be gated on it.

Agree. It's more like some efforts we can put on sig-testing separately. 

>I'm willing to take over this patch, and can handle the requests to revert a few lines. I'm not sure >what's involved with writing end-to-end tests for this change in behavior.
>What would the procedure be? Submit a fresh PR that starts from where this branch leaves off?

Hi @seh Is there a way we can get in touch with @piontec, not sure what happend to `missing Github user`

seh:
@Jeffwan, I see that Lukasz does publish an email address in his GitHub profile, but I expect that he's receiving updates for each message posted here. I'm willing to write to him once, just in case I'm wrong about those notifications. I don't want to bother him if he's had to move on to other projects.

Stay tuned.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -254,7 +254,9 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 	unregisteredNodes := a.clusterStateRegistry.GetUnregisteredNodes()
 	if len(unregisteredNodes) > 0 {
 		klog.V(1).Infof("%d unregistered nodes present", len(unregisteredNodes))
-		removedAny, err := removeOldUnregisteredNodes(unregisteredNodes, autoscalingContext, currentTime, autoscalingContext.LogRecorder)
+		removedAny, err := removeOldUnregisteredNodes(unregisteredNodes, autoscalingContext, a.clusterStateRegistry,
```
gjtempleton: Just twigged that this will need reverting as well whilst playing around with backporting this into a 1.3 version.

piontec:
Hey! I no longer work at aurea/devfactory and I lost write access there. I need to switch this work to my private github account. Please continue the discussion here: https://github.com/kubernetes/autoscaler/pull/2235. I did the requested cleanup there already.

Jeffwan:
> Hey! I no longer work at aurea/devfactory and I lost write access there. I need to switch this work to my private github account. Please continue the discussion here: #2235. I did the requested cleanup there already.

Good to now. Thanks!

JacobHenner:
To those who might be trying to track down issues with speedy transitions from spot to OnDemand, see https://github.com/kubernetes/autoscaler/issues/3241.

PR #2013, opened 2019-05-13
[cherry-pick] Add ASG MixedInstancesPolicy Support

danigrmartinez:
Hello! I would like this feature to be included in the 1.14 release.
This is a cherrypick of the two PR. Code and documentation from @drewhemm 
https://github.com/kubernetes/autoscaler/pull/1886
https://github.com/kubernetes/autoscaler/pull/1983

@drewhemm I am taking the liberty to propose your change. If you prefer to do it your self, please let me know and I will close this PR. Thanks!

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/MixedInstancePolicy.md
+++ b/cluster-autoscaler/cloudprovider/aws/MixedInstancePolicy.md
@@ -1,0 +1,69 @@
+The following is an excerpt from a CloudFormation template showing how a MixedInstancesPolicy can be used with ClusterAutoscaler:
+
+```json
+{
+    "Resources": {
+        "LaunchTemplate": {
+            "Type": "AWS::EC2::LaunchTemplate",
+            "Properties": {
+                "LaunchTemplateName": "memory-opt-2xlarge",
+                "LaunchTemplateData": {
+                    "InstanceType": "r5.2xlarge"
+                }
+            }
+        },
+        "ASGA": {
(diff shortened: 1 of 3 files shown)


drewhemm:
I am indeed happy for this to go in :)

losipiuk:
Normally we would not allow cherry pick as this one (the rule of thumb is only bug fixes are cherry-pickable). Yet this is change specific to AWS cloud provider only, so I do not feel strongly against it.
@Jeffwan what do you think? If you are fine with that let's merge this one. 

Also WDYT about older releases. Should we also cherry-pick to 1.13, 1.12 if we decide on going with 1.14?

Jeffwan:
@losipiuk Some CA customers on AWS would like to use MixedInstancePolicy for spot instances. As we know the Spot support for another PR has been closed. I think merge this back to 1.12-1.14 would give users a choice. The part missing is LaunchTemplate doesn't consider MixedInstancePolicy case. It's actually a patch and should be safe enough to cherry-pick to 1.12-1.14. 

@danigrmartinez Can you help cherry-pick this change to 1.13 and 1.12 as well?

danigrmartinez:
@Jeffwan @losipiuk Thank you!

@Jeffwan, sure I cal help with that. I didn't test with 1.13 or 1.12, But should not be much different. here the PRs:
https://github.com/kubernetes/autoscaler/pull/2019
https://github.com/kubernetes/autoscaler/pull/2018

PR #2022, opened 2019-05-15
Add tghartland as an approver for magnum cloud provider

losipiuk:
cc: @tghartland 
cc: @MaciekPytel 
cc: @mwielgus


tghartland:
Thanks 👍

PR #2025, opened 2019-05-15
Sync up go-autorest vendoring on v1.13.2

### LINKED ISSUE #1811: Long running operations may fail because of Azure go-autorest bugs
According to Azure/go-autorest#357, the timeout set in context doesn't take effect and 15min is always used for long running operations.

We should upgrade the version of go-autorest >= 11.3.2 or a patch version 11.1.2.

Depending on https://github.com/kubernetes/kubernetes/issues/75579.

/kind bug

mboersma:
Fixes #1811 

...or maybe not? I see that [go-autorest v1.13.2](https://github.com/Azure/go-autorest/releases/tag/v11.3.2) fixes the issue, but it was already vendored in, albeit inconsistently. So perhaps #1811 is already fixed in practice and this mainly just cleans up `Godeps.json`.


feiskyer:
It's still not resolved yet. Kubernetes has fixed the issue by upgrading go-autorest to v11.1.2.

Since autoscaler vendors Kubernetes, we should fix this by upgrading Kubernetes vendors (go-autorest would be upgraded automatically). 

@mboersma could you upgrade Kubernetes vendor version to v1.15.0-beta.0 ?

mboersma:
Since Kubernetes has moved to go modules, vendoring in its dependencies here with `godep` is not simple. I've tried a few approaches but may have to do a lot of it with copy-and-paste.

I'm going to close this PR and open a new one soon with the approach @feiskyer recommended.

feiskyer:
@mboersma thanks

PR #2028, opened 2019-05-16
VPA: Record more quality metrics

bskiba:
- Absolute diff between usage and recommendation
- Recommendation as observed on receiving a usage sample


bskiba:
The Travis failure seems unrelated to the change
@kgolab can you review?

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
```diff
@@ -30,6 +33,12 @@ const (
 	metricsNamespace = metrics.TopMetricsNamespace + "quality"
 )
 
+var (
+	// Both buckets are of length 17, 35 with prepended negative buckets
```
kgolab: I don't understand the comment, why are there two numbers?

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
```diff
@@ -30,6 +33,12 @@ const (
...
+	cpuBuckets    = getExponentialBuckets(0.01, 1000., 2.)
+	memoryBuckets = getExponentialBuckets(1e4, 1e12, 3.)
```
kgolab: I think we can safely start from 1e7 (that's only 10 MB, after all) and keep 2.0 growth factor.

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
```diff
@@ -44,15 +53,51 @@ var (
...
 			Name:      "usage_sample_missing_recommendation_count",
-			Help:      "Count of usage samples when a recommendation should be present but is missing.",
+			Help:      "Count of usage samples when a recommendation should be present but is missing",
 		}, []string{"resource", "is_oom"},
 	)
+	cpuUsageRecommendationDiff = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Namespace: metricsNamespace,
+			Name:      "cpu_usage_recommendation_diffs_cores",
+			Help:      "Absolute diffs between usage and recommendation for cpu",
```
kgolab: nit: s/cpu/CPU/ - to make it consistent with cpu_recommendation_cores

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
```diff
@@ -44,15 +53,51 @@ var (
...
+			Buckets:   cpuBuckets,
+		},
+	)
+	memoryRecommendations = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: metricsNamespace,
+			Name:      "mem_recommendations_bytes",
+			Help:      "Memory recommendation values as observed on recorded usage sample",
+			Buckets:   memoryBuckets,
+		}, []string{"is_oom"},
```
kgolab: Sorry for bringing it up only now and not in the previous PR - but do you think it would make sense to add "VPA mode" as a label to all these new metrics?

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
```diff
@@ -67,3 +112,59 @@ func ObserveUsageRecommendationRelativeDiff(usage, recommendation float64, isOOM
...
+		cpuUsageRecommendationDiff.Observe((usage - recommendation))
+	case corev1.ResourceMemory:
+		memoryUsageRecommendationDiff.WithLabelValues(strconv.FormatBool(isOOM)).Observe(usage - recommendation)
+	default:
+		klog.Warningf("Unknown resource: %v", resource)
+	}
+}
+
+// ObserveRecommendation records the value of recommendation.
+func ObserveRecommendation(recommendation float64, isOOM bool, resource corev1.ResourceName) {
```
kgolab: Have you considered not exporting this function (and 3 previous ones)?
I understand the two unchanged functions were used directly in the previous PR but now we have the ObserveQualityMetrics* as public API.

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
```diff
@@ -67,3 +112,59 @@ func ObserveUsageRecommendationRelativeDiff(usage, recommendation float64, isOOM
 func ObserveMissingRecommendation(isOOM bool, resource corev1.ResourceName) {
 	usageMissingRecommendationCounter.WithLabelValues(string(resource), strconv.FormatBool(isOOM)).Inc()
 }
+
+// ObserveUsageRecommendationDiff records absolute diff between usage and
+// recommendation.
+func ObserveUsageRecommendationDiff(usage, recommendation float64, isOOM bool, resource corev1.ResourceName) {
+	switch resource {
+	case corev1.ResourceCPU:
+		cpuUsageRecommendationDiff.Observe((usage - recommendation))
```
kgolab: nit: why double parentheses?

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
```diff
@@ -67,3 +112,59 @@ func ObserveUsageRecommendationRelativeDiff(usage, recommendation float64, isOOM
...
+func ObserveQualityMetrics(usage, recommendation float64, isOOM bool, resource corev1.ResourceName) {
+	ObserveRecommendation(recommendation, isOOM, resource)
+	ObserveUsageRecommendationDiff(usage, recommendation, isOOM, resource)
+	ObserveUsageRecommendationRelativeDiff(usage, recommendation, isOOM, resource)
+}
+
+// ObserveQualityMetricsRecommendationMissing records all quality metrics that we can derive from usage when recommendation is missing.
+func ObserveQualityMetricsRecommendationMissing(usage float64, isOOM bool, resource corev1.ResourceName) {
+	ObserveMissingRecommendation(isOOM, resource)
+	ObserveUsageRecommendationDiff(usage, 0, isOOM, resource)
```
kgolab: I think it would be useful to label these entries with "recommendation missing".

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
```diff
@@ -67,3 +112,59 @@ func ObserveUsageRecommendationRelativeDiff(usage, recommendation float64, isOOM
...
+		if bucket < 0 {
+			panic(fmt.Sprintf("Positive buckets expected. Got: %v", bucket))
+		}
+		result[bucketCount-i-1] = -bucket
+		result[bucketCount+i+1] = bucket
+	}
+	return result
+}
+
+func getExponentialBuckets(min, max, step float64) []float64 {
```
kgolab: Upon reading this I'm not sure if the function together with comments at var definition bring value.
Maybe it would more straightforward to use prometheus function directly and only comment on what the exponential factor is in each case.
WDYT?

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+func TestPrependBuckets(t *testing.T) {
+	assert.Equal(t, []float64{0.}, prependNegativeBuckets([]float64{}))
+	assert.Equal(t, []float64{-1., 0., 1}, prependNegativeBuckets([]float64{1.}))
+	assert.Equal(t, []float64{-2., -1., -0.5, -0.1, 0., 0.1, 0.5, 1., 2.}, prependNegativeBuckets([]float64{0.1, 0.5, 1., 2.}))
+	assert.Panics(t, func() { prependNegativeBuckets([]float64{-1.}) })
+}
+
+func TestGetExponentialBuckets(t *testing.T) {
+	assert.Equal(t, []float64{1., 2., 4.}, getExponentialBuckets(1., 5., 2.))
```
kgolab: The fact that 5.0 value seems ignored (unused) feels wrong.

kgolab:
Thanks for this work! Few remarks left.

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
bskiba: Removed altogether.

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
bskiba: Done.

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
bskiba: Done.

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
bskiba: No problem. Added.

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
bskiba: Done.

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
bskiba: Removed.

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
bskiba: Good point, done,

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
bskiba: Done.

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality_test.go
bskiba: Removed funtion.

bskiba:
Actually, changed this a little bit more - now diffs will be reported as positive values always, with a label telling which of usage, recommendation is higher. This actually allows to take a sum of all diffs when recommendation > usage out of the histogram.

bskiba:
Updated to split into two histograms each.

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
```diff
@@ -57,20 +57,36 @@ var (
...
+			Buckets:   cpuBuckets,
+		}, []string{"update_mode", "recommendation_missing"},
+	)
+	memoryRecommendationOverUsageDiff = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: metricsNamespace,
+			Name:      "mem_recommendation_over_usage_diffs_bytes",
+			Help:      "Absolute diffs between usage and recommendation for memory when recommendation > usage",
+			Buckets:   memoryBuckets,
+		}, []string{"update_mode", "recommendation_missing", "is_oom"},
```
kgolab: I don't think we can have "recommendation_missing" set to true for this particular metric (and the one for CPU) but I'd still like to keep it to have the labels in line with "LowerOrEquals" metric. WDYT?

kgolab:
Thanks! LGTM, I have one small doubt but that's it.

vertical-pod-autoscaler/pkg/utils/metrics/quality/quality.go
bskiba: I agree

PR #2047, opened 2019-05-20
Use fakeNodeLister instead of mocks.

krzysztof-jastrzebski:


cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -45,17 +46,25 @@ import (
 	"k8s.io/klog"
 )
 
-type nodeListerMock struct {
-	mock.Mock
+type fakeNodeLister struct {
```
losipiuk: Please renamte to `TestNodeLister` and move to `testlister.go` making it consistent with other test listers.
Please also add the `NewTestNodeLister(nodes...) TestNodeLister` method. And if you neet mutability `TestNodeLister.SetNodes()`

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -667,22 +676,22 @@ func TestStaticAutoscalerRunOncePodsWithFilterOutSchedulablePodsUsesPackingFalse
 	}
 
 	// Scale up
-	readyNodeListerMock.On("List").Return([]*apiv1.Node{n1, n2, n3}, nil).Times(2) // due to initialized=false
-	allNodeListerMock.On("List").Return([]*apiv1.Node{n1, n2, n3}, nil).Once()
+	readyNodeLister.SetNodes([]*apiv1.Node{n1, n2, n3})
+	allNodeLister.SetNodes([]*apiv1.Node{n1, n2, n3})
 	scheduledPodMock.On("List").Return([]*apiv1.Pod{p1, p2, p3}, nil).Times(3) // 1 to get pods + 1 per nodegroup when building nodeInfo map
```
losipiuk: Maybe migrate other listers mock to test listers too?

krzysztof-jastrzebski: It would be quite big change.

cluster-autoscaler/core/static_autoscaler_test.go
krzysztof-jastrzebski: Changed to be consistent but I think we use 'fake' prefix for fake implementations.

krzysztof-jastrzebski:
Comments resolved.

PR #2049, opened 2019-05-20
Limit range limit support

jbartosik:
This is built on #1813. I follow it with a change addressing comments from review of the original PR. On top of that I make changes to logic of handling limit range.

VPA now preserves limit/request ratio configured by user. VPA will now preserve the ratio if limit comes from default in limit range. It will also preserve the ratio if it must cap limit to keep it below max defined in the limit range.

This uses limit range that appears first in the list and applies to containers. It should be able to handle multiple limit ranges. It should also handle limit ranges on pods.


jbartosik:
@safanaj I'd like support for limit range to VPA. I would appreciate if you find time to check if changes I'm applying on top of your PR make sense.

jbartosik:
Issue: #1812

jbartosik:
Marking as WIP - I think this is ready for review but I didn't perform any E2E tests yet.

jbartosik:
@kgolab  PTAL if you have time.

vertical-pod-autoscaler/pkg/admission-controller/main.go
```diff
@@ -48,12 +48,15 @@ var (
...
-	registerByURL  = flag.Bool("register-by-url", false, "If set to true, admission webhook will be registered by URL (webhookAddress:webhookPort) instead of by service name")
+	port                = flag.Int("port", 8000, "The port to listen on.")
+	address             = flag.String("address", ":8944", "The address to expose Prometheus metrics.")
+	namespace           = os.Getenv("NAMESPACE")
+	webhookAddress      = flag.String("webhook-address", "", "Address under which webhook is registered. Used when registerByURL is set to true.")
+	webhookPort         = flag.String("webhook-port", "", "Server Port for Webhook")
+	registerByURL       = flag.Bool("register-by-url", false, "If set to true, admission webhook will be registered by URL (webhookAddress:webhookPort) instead of by service name")
+	allowToAdjustLimits = flag.Bool("allow-to-adjust-limits", false, "If set to true, admission webhook will set limits per container too if needed")
+
+	factoryForLimitsChecker interface{}
```
safanaj: this is not used anymore

bskiba:
@jbartosik There seems to still be a lot of logic for checking the request/limit ratio. Since your changes for VPA to keep the request to limit ratio constant, I think we no longer need it. The only thing we need to observe is the appropriate maxRatio.

safanaj:
@jbartosik Thanks, I think you addressed what I missed in my PR review cycle and does make sense for me.

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator.go
```diff
@@ -0,0 +1,88 @@
...
+func (lc *limitsChecker) GetContainerLimitRangeItem(namespace string) (*v1.LimitRangeItem, error) {
+	limitRanges, err := lc.limitRangeLister.LimitRanges(namespace).List(labels.Everything())
+	if err != nil {
+		fmt.Printf("1")
+		return nil, fmt.Errorf("error loading limit ranges: %s", err)
+	}
+
+	for _, lr := range limitRanges {
+		for _, lri := range lr.Spec.Limits {
+			if lri.Type == v1.LimitTypeContainer || lri.Type == v1.LimitTypePod && (lri.MaxLimitRequestRatio == nil || lri.Default == nil) {
```
safanaj: Hi @jbartosik , I don't get this. In case a LimitRange has MaxRatio and Default both not nil we don't choose that Limit?
Also, in server code I understand that [any limit range have to be respected](https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/admission/limitranger/admission.go#L106) so the more restrictive (minumum max ratio) have to be respected. In this code I understand that we are not checking for the more restrictive limits max ratio.

vertical-pod-autoscaler/pkg/admission-controller/main.go
jbartosik: Removed.

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator.go
jbartosik: My mistake. We care about limit range if it has at least one of Max / Default set.

safanaj: but what about the MaxRatio? the purpose of my initial PR was exactly just to shrink limits in case a max ratio was present in LimitRange definition to prevent that after the vpa-updater kill a pod the pod couldn't be recreated anymore because the vpa-admission-controller is setting the requests too low and the LimitRanger admission controller prevent the pod to be created.
I am sorry but I am a bit lost on this rework. this PR like it is now doesn't address anymore the issue https://github.com/kubernetes/autoscaler/issues/1812

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator.go
```diff
@@ -0,0 +1,85 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package logic
+
+import (
+	"fmt"
```
bskiba: nit: split imports after "fmt"

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator.go
```diff
@@ -0,0 +1,85 @@
...
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/informers"
+
+	v1_listers "k8s.io/client-go/listers/core/v1"
```
bskiba: nit: just listers?

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+// LimitsRangeCalculator checks for LimitRange and if container needs limits to be set
+type LimitsRangeCalculator interface {
```
bskiba: This is actually a LimitRangeItemLister, with a filter applied. So maybe LimitRangeFetcher would be a more appropriate name.

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator.go
```diff
@@ -0,0 +1,85 @@
...
+	limitRanges, err := lc.limitRangeLister.LimitRanges(namespace).List(labels.Everything())
+	if err != nil {
+		return nil, fmt.Errorf("error loading limit ranges: %s", err)
+	}
+
+	for _, lr := range limitRanges {
+		for _, lri := range lr.Spec.Limits {
+			if lri.Type == v1.LimitTypeContainer && (lri.Max != nil || lri.Default != nil) {
+				// TODO: handle multiple limit ranges matching a pod.
+				return &v1.LimitRangeItem{
```
bskiba: Why do we make a copy?

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator_test.go
```diff
@@ -0,0 +1,124 @@
...
+*/
+
+package logic
+
+import (
+	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/informers"
+
+	//"fmt"
```
bskiba: nit: please organise import and remove comments

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator_test.go
```diff
@@ -0,0 +1,124 @@
...
+			assert.Nil(t, limitRange)
+		}
+	}
+}
+
+func TestUpdateResourceLimits(t *testing.T) {
+
+	testCases := []struct {
+		name         string
+		pod          *apiv1.Pod
```
bskiba: tc.pod seems to be never used

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator.go
```diff
@@ -0,0 +1,85 @@
...
+func NewLimitsRangeCalculator(f informers.SharedInformerFactory) (*limitsChecker, error) {
+	if f == nil {
+		return nil, fmt.Errorf("NewLimitsRangeCalculator requires a SharedInformerFactory but got nil")
+	}
+	limitRangeLister := f.Core().V1().LimitRanges().Lister()
+	stopCh := make(chan struct{})
+	f.Start(stopCh)
+	for _, ok := range f.WaitForCacheSync(stopCh) {
+		if !ok {
+			if f.Core().V1().LimitRanges().Informer().HasSynced() {
```
bskiba: if HasSynced() we return an error? I think you're missing a negation here.

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator_test.go
```diff
@@ -0,0 +1,124 @@
...
+		limitRanges  []runtime.Object
+		expectErr    error
+		expectLimits *apiv1.LimitRangeItem
```
bskiba: nit: expectedLimits

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	if assert.NoError(t, err) {
+		limitRange, err := lc.GetContainerLimitRangeItem(getPod().Namespace)
+		if assert.NoError(t, err) {
+			assert.Nil(t, limitRange)
+		}
+	}
+}
+
+func TestUpdateResourceLimits(t *testing.T) {
```
bskiba: nit: TestGetContainerLimitRangeItem

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
```diff
@@ -82,53 +103,75 @@ func getProportionalLimit(originalLimit, originalRequest, recommendedRequest *re
...
-	originalMilliLimit := big.NewInt(originalLimit.MilliValue())
-	recommendedMilliRequest := big.NewInt(recommendedRequest.MilliValue())
-	var recommendedMilliLimit big.Int
-	recommendedMilliLimit.Mul(recommendedMilliRequest, originalMilliLimit)
-	recommendedMilliLimit.Div(&recommendedMilliLimit, originalMilliRequest)
-	if recommendedMilliLimit.IsInt64() {
-		return resource.NewMilliQuantity(recommendedMilliLimit.Int64(), recommendedRequest.Format), false
-	}
-	return resource.NewMilliQuantity(math.MaxInt64, recommendedRequest.Format), true
+	return scaleQuantityProportionally( /*scaledQuantity=*/ originalLimit /*scaleBase=*/, originalRequest /*scaleResult=*/, recommendedRequest)
```
bskiba: The commas seem to be in weird places.

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
```diff
@@ -82,53 +103,75 @@ func getProportionalLimit(originalLimit, originalRequest, recommendedRequest *re
...
+		return recommendedRequest, recommendedLimit
+	}
+	if recommendedLimit.Cmp(*maxLimit) <= 0 {
+		return recommendedRequest, recommendedLimit
+	}
+	scaledRequest, _ := scaleQuantityProportionally(recommendedRequest, recommendedLimit, maxLimit)
+	return scaledRequest, maxLimit
+}
+
+func proportionallyCapLimitsToMax(recommendedRequests v1.ResourceList, cpuLimit, memLimit, maxCpuLimit, maxMemLimit *resource.Quantity) ContainerResources {
```
bskiba: I'd call it proportionallyCapResourcesToMaxLimit, since it also caps request (I was suprised)

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
```diff
@@ -82,53 +103,75 @@ func getProportionalLimit(originalLimit, originalRequest, recommendedRequest *re
...
+	scaledCpuRequest, scaledCpuLimit := proportionallyCapLimitToMax(recommendedRequests.Cpu(), cpuLimit, maxCpuLimit)
+	scaledMemRequest, scaledMemLimit := proportionallyCapLimitToMax(recommendedRequests.Memory(), memLimit, maxMemLimit)
+	result := newContainerResources()
+	if scaledCpuRequest != nil {
```
bskiba: When will those be nil?

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
bskiba: This logic actually changes the recommended request, but this is not included in the updater - updater and admission controller will disagree about the request - we can get into an infinite updating loop.

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator.go
bskiba: @safanaj since your PR we have changed the logic of VPA to keep the ratio between request and limit constant. This means that VPA will never break things - since we keep the original request to limit ratio that was specified for the deployment, we rely on the user to specify it in a way that respects the MaxRatio.

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"fmt"
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/informers"
+
+	v1_listers "k8s.io/client-go/listers/core/v1"
+)
+
+// LimitsRangeCalculator checks for LimitRange and if container needs limits to be set
```
kgolab: It's probably nitpicking but I don't see why this interface is called Calculator, Finder would be more appropriate.
I also don't get the second part of the comment.

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator.go
kgolab: Shouldn't this be ! HasSynced() ?

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator_test.go
```diff
@@ -0,0 +1,124 @@
...
+		pod          *apiv1.Pod
+		limitRanges  []runtime.Object
+		expectErr    error
+		expectLimits *apiv1.LimitRangeItem
+	}{
+		{
+			name: "no matching limit ranges",
+			pod:  getPod(),
+			limitRanges: []runtime.Object{
+				test.LimitRange().WithName("different-namespace").WithNamespace("different").WithType(apiv1.LimitTypePod).WithMax(test.Resources("2", "2")).Get(),
```
kgolab: nit: since it's different namespace already maybe it's not worth changing type too?

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (lc *limitsChecker) GetContainerLimitRangeItem(namespace string) (*v1.LimitRangeItem, error) {
+	limitRanges, err := lc.limitRangeLister.LimitRanges(namespace).List(labels.Everything())
+	if err != nil {
+		return nil, fmt.Errorf("error loading limit ranges: %s", err)
+	}
+
+	for _, lr := range limitRanges {
+		for _, lri := range lr.Spec.Limits {
+			if lri.Type == v1.LimitTypeContainer && (lri.Max != nil || lri.Default != nil) {
```
kgolab: What happens with LimitTypePod limits?

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator_test.go
```diff
@@ -0,0 +1,124 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package logic
+
+import (
+	"github.com/stretchr/testify/assert"
```
kgolab: Please remove unneeded imports & sort.

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+func getPod() *apiv1.Pod {
+	labels := map[string]string{"app": "testingApp"}
+	return test.Pod().WithName("test_uninitialized").AddContainer(test.BuildTestContainer("container1", "", "")).WithLabels(labels).Get()
+}
+
+func TestNewNoopLimitsChecker(t *testing.T) {
+	nlc := NewNoopLimitsCalculator()
+	limitRange, err := nlc.GetContainerLimitRangeItem(getPod().Namespace)
+	if assert.NoError(t, err) {
```
kgolab: Do we need this if? Elsewhere we're simply using:
```
assert.NoError
assert.MoreChecks
```

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
```diff
@@ -31,104 +29,63 @@ import (
...
 			klog.V(2).Infof("no matching recommendation found for container %s", container.Name)
 			continue
 		}
-		resources[i].Requests = recommendation.Target
-
-		cpuLimit, capped := getProportionalLimit(container.Resources.Limits.Cpu(), container.Resources.Requests.Cpu(), resources[i].Requests.Cpu())
-		if cpuLimit != nil {
-			resources[i].Limits[v1.ResourceCPU] = *cpuLimit
-		}
+		cpuLimit, capped := vpa_api_util.GetProportionalLimit(container.Resources.Limits.Cpu(), container.Resources.Requests.Cpu(), recommendation.Target.Cpu(), defaultCpu)
```
bskiba: Can we have GetProportionalLimit to return annotation instead of bool in case of capping? I think it would make this code clearer.

vertical-pod-autoscaler/pkg/admission-controller/logic/limit_range_calculator.go
bskiba: Ok, then since it's LimitRange not LimitsRange, lets call it LimitRangeCalculator

vertical-pod-autoscaler/pkg/admission-controller/main.go
```diff
@@ -48,12 +48,13 @@ var (
...
-	webhookAddress = flag.String("webhook-address", "", "Address under which webhook is registered. Used when registerByURL is set to true.")
-	webhookPort    = flag.String("webhook-port", "", "Server Port for Webhook")
-	registerByURL  = flag.Bool("register-by-url", false, "If set to true, admission webhook will be registered by URL (webhookAddress:webhookPort) instead of by service name")
+	port                = flag.Int("port", 8000, "The port to listen on.")
+	address             = flag.String("address", ":8944", "The address to expose Prometheus metrics.")
+	namespace           = os.Getenv("NAMESPACE")
+	webhookAddress      = flag.String("webhook-address", "", "Address under which webhook is registered. Used when registerByURL is set to true.")
+	webhookPort         = flag.String("webhook-port", "", "Server Port for Webhook")
+	registerByURL       = flag.Bool("register-by-url", false, "If set to true, admission webhook will be registered by URL (webhookAddress:webhookPort) instead of by service name")
+	allowToAdjustLimits = flag.Bool("allow-to-adjust-limits", false, "If set to true, admission webhook will set limits per container too if needed")
```
bskiba: Actually this flag does not seem necessary any more. We are already adjusting limits even when this flag is set to false (see your previous PRs) so this is not accurate. I think we can remove it and also get rid of NoopLimitsCalculator - if we already manipulate limits we might as well always observe LimitRange.

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
```diff
@@ -183,6 +140,11 @@ func (p *recommendationProvider) GetContainersResourcesForPod(pod *v1.Pod) ([]Co
 			return nil, annotations, vpaConfig.Name, err
 		}
 	}
-	containerResources := GetContainersResources(pod, *recommendedPodResources, annotations)
+	podLimitRange, err := p.limitsRangeCalculator.GetContainerLimitRangeItem(pod.Namespace)
+	// TODO: Support limit range on pod level.
```
bskiba: This might be useful https://github.com/kubernetes/kubernetes/blob/016e9d5c06089774c6286fd825302cbae661a446/plugin/pkg/admission/limitranger/admission.go#L557

vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go
```diff
@@ -0,0 +1,104 @@
...
+		return &result, false
+	}
+	// originalLimit and originalRequest are set. If they are equal recommend limit equal to request.
+	if originalRequest.MilliValue() == originalLimit.MilliValue() {
+		result := *recommendedRequest
+		return &result, false
+	}
+
+	// Input and output milli values should fit in int64 but intermediate values might be bigger.
+	return scaleQuantityProportionally( /*scaledQuantity=*/ originalLimit /*scaleBase=*/, originalRequest /*scaleResult=*/, recommendedRequest)
```
bskiba: nit: The commas are still weird

vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go
```diff
@@ -0,0 +1,104 @@
...
+		result := *recommendedRequest
+		return &result, false
+	}
+	// originalLimit and originalRequest are set. If they are equal recommend limit equal to request.
+	if originalRequest.MilliValue() == originalLimit.MilliValue() {
+		result := *recommendedRequest
+		return &result, false
+	}
+
+	// Input and output milli values should fit in int64 but intermediate values might be bigger.
```
bskiba: This comment seems mispalaced.

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/admission-controller/main.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go
jbartosik: Removed.

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
jbartosik: Thanks :)

vertical-pod-autoscaler/pkg/admission-controller/logic/recommendation_provider.go
```diff
@@ -31,104 +29,55 @@ import (
...
-			annotations[container.Name] = append(
-				annotations[container.Name],
-				fmt.Sprintf(
-					"Failed to keep CPU limit to request proportion of %d to %d with recommended request of %d milliCPU; doesn't fit in int64. Capping limit to MaxInt64",
-					container.Resources.Limits.Cpu().MilliValue(), container.Resources.Requests.Cpu().MilliValue(), resources[i].Requests.Cpu().MilliValue()))
-		}
-		memLimit, capped := getProportionalLimit(container.Resources.Limits.Memory(), container.Resources.Requests.Memory(), resources[i].Requests.Memory())
-		if memLimit != nil {
-			resources[i].Limits[v1.ResourceMemory] = *memLimit
+		cpuLimit, annotation := vpa_api_util.GetProportionalLimit(container.Resources.Limits.Cpu(), container.Resources.Requests.Cpu(), recommendation.Target.Cpu(), defaultCpu)
```
bskiba: I was hoping to have a list of strings returned and lose the if, but I guess this is fine as well.

vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go
```diff
@@ -0,0 +1,110 @@
...
+		return recommendedRequest, recommendedLimit
+	}
+	if recommendedLimit.Cmp(*maxLimit) <= 0 {
+		return recommendedRequest, recommendedLimit
+	}
+	scaledRequest, _ := scaleQuantityProportionally(recommendedRequest, recommendedLimit, maxLimit)
+	return scaledRequest, maxLimit
+}
+
+// ProportionallyCapResourcesToMaxLimit caps CPU and memory limit to maximu and scales requests to maintain limit/request ratio.
```
bskiba: nit: s/maximu/maximum/

bskiba:
Hm, wait, I just noticed. We also care about the **min** limit, don't we? atm we can downscale below min and make pod invalid.

bskiba:
i.e. I expect the "raises request according to min limit set in LimitRange" test to pass ahen we are done. See test here: https://github.com/bskiba/autoscaler/commit/da76008702cfd1ae3aeebae41b6308a73fc7590b

vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go
jbartosik: Done

jbartosik:
@bskiba PTAL

vertical-pod-autoscaler/pkg/utils/vpa/capping_test.go
```diff
@@ -215,3 +215,69 @@ func TestApplyVpa(t *testing.T) {
...
+			},
+		},
+	}
+	expectedRecommendation := vpa_types.RecommendedPodResources{
+		ContainerRecommendations: []vpa_types.RecommendedContainerResources{
+			{
+				ContainerName: "container",
+				Target: apiv1.ResourceList{
+					apiv1.ResourceCPU:    resource.MustParse("1000m"),
+					apiv1.ResourceMemory: resource.MustParse("1000000000000m"),
```
bskiba: (I'm not sure if my comment was lost) Why millivalue here? It's not too readable

vertical-pod-autoscaler/pkg/utils/vpa/recommendation_processor.go
```diff
@@ -30,7 +30,7 @@ type RecommendationProcessor interface {
 	// VPA policy and possibly other internal RecommendationProcessor context.
 	// Must return a non-nil pointer to RecommendedPodResources or error.
 	Apply(podRecommendation *vpa_types.RecommendedPodResources,
-		policy *vpa_types.PodResourcePolicy,
-		conditions []vpa_types.VerticalPodAutoscalerCondition,
+		_ *vpa_types.PodResourcePolicy,
```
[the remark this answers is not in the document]
jbartosik: Was in hurry, made a mistake.

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
```diff
@@ -84,6 +91,36 @@ func (c *cappingRecommendationProcessor) Apply(
 	return &vpa_types.RecommendedPodResources{ContainerRecommendations: updatedRecommendations}, containerToAnnotationsMap, nil
 }
 
+func (c *cappingRecommendationProcessor) scaleInProportionWithLimit(podRecommendation *vpa_types.RecommendedPodResources, pod *apiv1.Pod) ([]vpa_types.RecommendedContainerResources, error) {
```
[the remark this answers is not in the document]
jbartosik: Done.

vertical-pod-autoscaler/pkg/utils/vpa/capping_test.go
jbartosik: I want to use assert.equal to keep test easy to read. The way assert.equal works representation of quantities must be the same, not just their actual values.

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
```diff
@@ -62,7 +65,11 @@ func (c *cappingRecommendationProcessor) Apply(
 	}
 	updatedRecommendations := []vpa_types.RecommendedContainerResources{}
 	containerToAnnotationsMap := ContainerToAnnotationsMap{}
-	for _, containerRecommendation := range podRecommendation.ContainerRecommendations {
+	limitAdjustedRecommendation, err := c.capProportionallyToMaxLimit(podRecommendation, pod, containerToAnnotationsMap)
```
bskiba: This doesn't work. We need to apply capping also to upper and lower bound. What we need is to integrate capping to max/min limitrange into getCappedRecommendationForContainer function. 
Just as we have ApplyVpaPolicy, we need to have ApplyMaxRange here:
https://github.com/kubernetes/autoscaler/blob/5094dc06c0bdfc50a543e4ae719c4427dcf029d3/vertical-pod-autoscaler/pkg/utils/vpa/capping.go#L146

Apologies for not spotting it earlier.

jbartosik: Changed to adjust not only target but also bounds. I think this will work correctly (if limit range, limit/request ratio and policy allow assigning a recommendation).

bskiba:
/lgtm
/approve

please remove WIP from the title

PR #2056, opened 2019-05-22
Move static scheduler initialization out of module init()

losipiuk:
Having call to algorithmprovider.ApplyFeatureGates() in module init
function was problematic as the mathod calls out to klog.Infof().
The call to klog.Infof() was then happening before klog related
application flags were parsed which resulted in klog misbehavior.
See https://github.com/kubernetes/klog/issues/67.


MaciekPytel:
/lgtm
/approve

Nice job finding this.

PR #2059, opened 2019-05-26
Dockerfile changes and readme update

saiyam1814:
In order to test out the new features that are getting added each day in the  autoscaler it becomes very common for developers to build the code from master branch , currently there is no such way to build from the Dockerfile as the current Dockerfile expects the code to be prebuilt from the master. 
So I have added the build step using go Docker image and some configs for the cluster autoscaler to be build the code and get the image baked from the master(with no additional step).

Also I have updated the readme file for building the cluster autoscaler from the Dockerfile (including one type from the previous Readme).

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ git clone https://github.com/$YOUR_GITHUB_USERNAME/autoscaler.git
 # Replace "GO_Versiom" below with the version of go you want to use(eg. go1.11)
 wget https://dl.google.com/go/$GO_Version.linux-amd64.tar.gz
 # Extract the tar file
-tar -xvf go1.11.linux-amd64.tar.gz
+tar -xvf $GO_Version.linux-amd64.tar.gz
 # Move go to /usr/local
 mv go /usr/local
 # Set GOROOT,GOPATH,PATH and add to profile
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ cd $GOPATH/src/k8s.io/autoscaler/cluster-autoscaler
 # start the build
 GOOS=linux make build-binary
 #Copy  run.sh and Run the Dockerfile from $GOPATH/src/k8s.io/autoscaler directory
-cp $GOPATH/src/k8s.io/autoscaler/cluster-autoscaler ..
+cp $GOPATH/src/k8s.io/autoscaler/cluster-autoscaler/run.sh ..
 docker build -t autoscaler -f cluster-autoscaler/Dockerfile .
 ```
 
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ of our weekly meetings.  See [the Kubernetes Community Repo](https://github.com/
 
 Fork the repository in the cloud:
 1. Visit https://github.com/kubernetes/autoscaler
-1. Click Fork button (top right) to establish a cloud-based fork.
+2. Click Fork button (top right) to establish a cloud-based fork.
 
 The code must be checked out as a subdirectory of `k8s.io`, and not `github.com`.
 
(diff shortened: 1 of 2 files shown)


saiyam1814:
CLA verified

README.md
```diff
@@ -38,6 +38,16 @@ git clone https://github.com/$YOUR_GITHUB_USERNAME/autoscaler.git
 cd autoscaler
 ```
 
+## Building the Cluster autoscaler
+
+```shell
+# Replace "$YOUR_GITHUB_USERNAME" below with your github username
+git clone https://github.com/$YOUR_GITHUB_USERNAME/autoscaler.git
+cd autoscaler
+docker build -t autoscaler cluster-autoscaler/Dockerfile .
```
Jeffwan: Please check Makefile for the commands to build binary and container image.

I would suggest to use existing commands to rewrite this section. Dockerfile doesn't have to changed.

saiyam1814: Sure , I will probably change the Readme file and keep the Dockerfile same so that those  who are using it already do not have to change .

saiyam1814: @Jeffwan I have updated only the readme now, can you check ? LMK if any additions or deletions required

README.md
```diff
@@ -38,6 +38,34 @@ git clone https://github.com/$YOUR_GITHUB_USERNAME/autoscaler.git
...
 
+## Building the Cluster autoscaler from master
+
+```shell
+# Replace "$YOUR_GITHUB_USERNAME" below with your github username
+git clone https://github.com/$YOUR_GITHUB_USERNAME/autoscaler.git
+# Replace "GO_Versiom" below with the version of go you want to use(eg. go1.11)
+wget https://dl.google.com/go/$GO_Version.linux-amd64.tar.gz
+# Extract the tar file
+tar -xvf go1.11.linux-amd64.tar.gz
```
Jeffwan: Not sure if that's redundant to have golang setups here.. I think CA doc should concentrate on build CA images?

Jeffwan: You use GO_Version env but here it's fixed version. This can be removed.

saiyam1814: @Jeffwan Updated the GoLang version  and what I feel is there might be people , sys admins you actually might not know GO , so if little generic steps are provided for setting up go and building the binary from the master to test out the latest features from the branch , it would be helpful IMO

saiyam1814:
@Jeffwan @mwielgus Since we cannot change the Dockerfile as it is directly used by GKE at Google , can we have a new Dockerfile something like Docker_complete in the repo this will reduce the readme file contents and the go commands will then become part of new dockerfile. Then we can have minimum number of Readme steps . Let me know your thoughts on this .

krzysztof-jastrzebski:
There is an easier way to build cluster autoscaler:
1. Download your sources.
2. Go to cluster autoscaler directory.
3. Export following variables
export REGISTRY=<your docker registry>
export TAG=<your tag>
4 Exacute:
make release
You will be asked if you want to push to registry during building release.
Does it solve your problem?
Could you put this into FAQ instead of Readme please?

saiyam1814:
@krzysztof-jastrzebski actually the whole point is having a separate Dockerfile apart from the existing one  so that user do not have to do anything and can just build the dockerfile and have a new build of cluster autoscaler available.

losipiuk:
> Krzysztof Jastrzębski actually the whole point is having a separate Dockerfile apart from the existing one so that user do not have to do anything and can just build the dockerfile and have a new build of cluster autoscaler available.

To support we could make Makefile understand `DOCKERFILE` env variable.
Then you could do
`REGISTRY=... TAG=... DOCKERFILE=.... make release`

saiyam1814:
@losipiuk actually I want to do that but not worked on Makefile before.

saiyam1814:
closing after discussion with @losipiuk

PR #2062, opened 2019-05-27
Create stores and reflectors with a simpler way

mrkm4ntr:
Use `cache.NewNamespaceKeyedIndexerAndReflector` to create stores and reflectors.
The `Resync` does nothing without informers, so I set resyncPeriod 0.
https://github.com/kubernetes/autoscaler/blob/5c743c1ffefbfba36326cf2e4bfd67c4443e41a1/cluster-autoscaler/vendor/k8s.io/client-go/tools/cache/thread_safe_store.go#L300-L303

### DIFF
--- a/cluster-autoscaler/utils/kubernetes/listers.go
+++ b/cluster-autoscaler/utils/kubernetes/listers.go
@@ -185,10 +185,9 @@ func NewUnschedulablePodInNamespaceLister(kubeClient client.Interface, namespace
 	selector := fields.ParseSelectorOrDie("spec.nodeName==" + "" + ",status.phase!=" +
 		string(apiv1.PodSucceeded) + ",status.phase!=" + string(apiv1.PodFailed))
 	podListWatch := cache.NewListWatchFromClient(kubeClient.CoreV1().RESTClient(), "pods", namespace, selector)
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -185,10 +185,9 @@ func NewUnschedulablePodInNamespaceLister(kubeClient client.Interface, namespace
 	selector := fields.ParseSelectorOrDie("spec.nodeName==" + "" + ",status.phase!=" +
 		string(apiv1.PodSucceeded) + ",status.phase!=" + string(apiv1.PodFailed))
 	podListWatch := cache.NewListWatchFromClient(kubeClient.CoreV1().RESTClient(), "pods", namespace, selector)
-	store := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+	store, reflector := cache.NewNamespaceKeyedIndexerAndReflector(podListWatch, &apiv1.Pod{}, 0)
```
losipiuk: Here we are changing resyncPeriod from one hour to 0. Not sure if that matters but please leave old value just in case.

mrkm4ntr: Resync does nothing actually, but it's OK to leave old value. I will fix it.

losipiuk:
/lgtm
/a
/lgtm
/approve
Thanks!

PR #2064, opened 2019-05-27
Rebase Cluster-Autoscaler on top of distroless/static

losipiuk:


losipiuk:
This is part of the effort to rebase all k8s components images on top of distroless. Having minimalistic base image makes lower the probability of having exploitable security holes in CA image.

cc: @feiskyer @Jeffwan @ringtail @tghartland @hello2mao

tghartland:
This seems to be the relevant issue -> https://github.com/kubernetes/kubernetes/issues/70249

No problems for me with building and running on magnum.

There's a nice reduction in image size as well:
```
REPOSITORY                  TAG               IMAGE ID          CREATED            SIZE
cluster-autoscaler          distroless        81a658ea648e      7 minutes ago      104MB
cluster-autoscaler          debian            3d379eedb826      10 minutes ago     151MB
```
and even more so when built with only the tag for a single cloud provider, down to half the size:
```
REPOSITORY                  TAG               IMAGE ID          CREATED            SIZE
cluster-autoscaler-magnum   distroless        3d084222133a      4 minutes ago      51.3MB
cluster-autoscaler-magnum   debian            8d4eac7d7360      5 minutes ago      98.5MB
```

/lgtm

cluster-autoscaler/Dockerfile
```diff
@@ -11,15 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-ARG BASEIMAGE=k8s.gcr.io/debian-base-amd64:v1.0.0
+ARG BASEIMAGE=gcr.io/distroless/static:latest
```
Jeffwan: Wondering what's the size of this distroless image?

MaciekPytel:
/lgtm

That being said we're only going to use it for 1.15+ and we need to bump dependencies for that anyway. So we probably don't need the cherry-picking into vendor. Let's wait with merging this until we bump dependencies.

/hold

PR #2066, opened 2019-05-28
Delete zero values from deletionsInProgress map in NodeDeletionTracker.

krzysztof-jastrzebski:
Delete zero values from deletionsInProgress map in NodeDeletionTracker.


losipiuk:
I do not believe commit message is very informative. 
lgtm otherwise.

krzysztof-jastrzebski:
changed

PR #2067, opened 2019-05-28
Add caching for mig instance templates

losipiuk:


cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -305,6 +307,46 @@ func (gc *GceCache) InvalidateAllMigTargetSizes() {
 	gc.migTargetSizeCache = map[GceRef]int64{}
 }
 
+// GetMigInstanceTemplate returns the cached gce.InstanceTemplate for a mig GceRef
+func (gc *GceCache) GetMigInstanceTemplate(ref GceRef) (*gce.InstanceTemplate, bool) {
```
krzysztof-jastrzebski: Do you need to return found? It is hard to tell what it means without looking at implementation.
If you add nil to cache SetMigInstanceTemplate(some_ref, nil) then GetMigInstanceTemplate(some_ref) will return (nil, true). I think it is better to return nil in this case.

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -305,6 +307,46 @@ func (gc *GceCache) InvalidateAllMigTargetSizes() {
...
+	gc.cacheMutex.Lock()
+	defer gc.cacheMutex.Unlock()
+
+	instanceTemplate, found := gc.instanceTemplatesCache[ref]
+	if found {
+		klog.V(5).Infof("instance template cache hit for %s", ref)
```
krzysztof-jastrzebski: Shouldn't logs start from capital letters? This is not an error. Same below.

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -305,6 +307,46 @@ func (gc *GceCache) InvalidateAllMigTargetSizes() {
...
+
+// SetMigInstanceTemplate sets gce.InstanceTemplate for a mig GceRef
+func (gc *GceCache) SetMigInstanceTemplate(ref GceRef, instanceTemplate *gce.InstanceTemplate) {
+	gc.cacheMutex.Lock()
+	defer gc.cacheMutex.Unlock()
+
+	gc.instanceTemplatesCache[ref] = instanceTemplate
+}
+
+// InvalidateMigInstanceTemplate clears the instance template cache
```
krzysztof-jastrzebski: Add "for a mig GceRef"

cluster-autoscaler/cloudprovider/gce/mig_instance_templates_provider.go
```diff
@@ -0,0 +1,74 @@
...
+		gceClient: gceClient,
+	}
+}
+
+// GetMigInstanceTemplate returns instance template for MIG with given ref
+func (p *CachingMigInstanceTemplatesProvider) GetMigInstanceTemplate(migRef GceRef) (*gce.InstanceTemplate, error) {
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+
+	if !p.lastRefresh.Add(migInstanceCacheRefreshInterval).After(time.Now()) {
```
krzysztof-jastrzebski: Wouldn't it better to fill this cache asynchronously?

cluster-autoscaler/cloudprovider/gce/cache.go
losipiuk: I would leave it. It is consistent with how go generally deals with getting values from map. That there is separate bool value returned to determine if value was found or not.

cluster-autoscaler/cloudprovider/gce/mig_instance_templates_provider.go
losipiuk: I think it would actually. It is kinda bigger refactor though. The CP would need to start to be active component doing stuff in the background. Also it would be nice to do the change holistically for (at least some) other caches too. I would skip it before GA.

cluster-autoscaler/cloudprovider/gce/mig_instance_templates_provider.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"sync"
+	"time"
+
+	gce "google.golang.org/api/compute/v1"
+)
+
+const (
+	migInstanceCacheRefreshInterval = 60 * time.Minute
```
MaciekPytel: 1. Having those timeouts scattered all over the place makes them hard to find and quickly figure out what is cached and for how long. Maybe we should just put them all in one place, to make it easier? Something like cloudprovider/gce/timeouts.go? WDYT?
2. I'm trying to figure out if those can ever change? Upgrading to a newer k8s version is the only thing that comes to my mind. Are there other scenarios? Even if there aren't 1 hour seems like a long timeout. Maybe we should make it 10 minutes or so?
3. On a somewhat related question (not part of this PR, just something to consider for future) - we're going to get one much slower loop every time we drop cache. Maybe we should consider including some random jitter to timeout, so that we invalidate instances for different groups at different times and spread the pain between loops? Same goes for other caching we do in cloudprovider.

cluster-autoscaler/cloudprovider/gce/mig_instance_templates_provider.go
MaciekPytel: Agree with @losipiuk.

cluster-autoscaler/cloudprovider/gce/mig_instance_templates_provider.go
losipiuk: (3) is the problem I also though about and I do not like the behavior. The anychronous approach would let as avoid this problem. 
Having `lastRefresh` per each entry in the map would help too - but is kinda ugly and not sure who should hold it - cache itself or the provider. I will give it a though and maybe send out a followup.

losipiuk: > Having those timeouts scattered all over the place makes them hard to find and quickly figure out what is cached and for how long. Maybe we should just put them all in one place, to make it easier? Something like cloudprovider/gce/timeouts.go? WDYT

Actually other timeouts are outside of GCE cloud provider. And are part of core logic - so there is no natural place group them. I will leave the const here and only start grouping if we end up having more of those.

losipiuk: (2) "szewskim targiem" changing to 30s

PR #2078, opened 2019-05-30
Cache cloud provider node instances in cluster state.

krzysztof-jastrzebski:
Cache cloud provider node instances in cluster state.

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -134,6 +134,8 @@ type ClusterStateRegistry struct {
 	logRecorder                        *utils.LogEventRecorder
 	cloudProviderNodeInstances         map[string][]cloudprovider.Instance
 	previousCloudProviderNodeInstances map[string][]cloudprovider.Instance
+	cloudProviderNodeInstancesCache    *utils.CloudProviderNodeInstancesCache
+	interrupt                          chan struct{}
 }
 
 // NewClusterStateRegistry creates new ClusterStateRegistry.
@@ -142,21 +144,34 @@ func NewClusterStateRegistry(cloudProvider cloudprovider.CloudProvider, config C
 		ClusterwideConditions: make([]api.ClusterAutoscalerCondition, 0),
 		NodeGroupStatuses:     make([]api.NodeGroupStatus, 0),
 	}
+
 	return &ClusterStateRegistry{
-		scaleUpRequests:         make(map[string]*ScaleUpRequest),
-		scaleDownRequests:       make([]*ScaleDownRequest, 0),
-		nodes:                   make([]*apiv1.Node, 0),
-		cloudProvider:           cloudProvider,
-		config:                  config,
-		perNodeGroupReadiness:   make(map[string]Readiness),
-		acceptableRanges:        make(map[string]AcceptableRange),
-		incorrectNodeGroupSizes: make(map[string]IncorrectNodeGroupSize),
-		unregisteredNodes:       make(map[string]UnregisteredNode),
-		candidatesForScaleDown:  make(map[string][]string),
-		backoff:                 backoff,
-		lastStatus:              emptyStatus,
-		logRecorder:             logRecorder,
+		scaleUpRequests:                 make(map[string]*ScaleUpRequest),
+		scaleDownRequests:               make([]*ScaleDownRequest, 0),
+		nodes:                           make([]*apiv1.Node, 0),
+		cloudProvider:                   cloudProvider,
+		config:                          config,
+		perNodeGroupReadiness:           make(map[string]Readiness),
+		acceptableRanges:                make(map[string]AcceptableRange),
+		incorrectNodeGroupSizes:         make(map[string]IncorrectNodeGroupSize),
+		unregisteredNodes:               make(map[string]UnregisteredNode),
+		candidatesForScaleDown:          make(map[string][]string),
+		backoff:                         backoff,
+		lastStatus:                      emptyStatus,
+		logRecorder:                     logRecorder,
+		cloudProviderNodeInstancesCache: utils.NewCloudProviderNodeInstancesCache(cloudProvider),
+		interrupt:                       make(chan struct{}),
 	}
+}
+
+// Start starts components running in background.
+func (csr *ClusterStateRegistry) Start() {
+	csr.cloudProviderNodeInstancesCache.Start(csr.interrupt)
+}
+
+// Stop stops components running in background.
+func (csr *ClusterStateRegistry) Stop() {
+	close(csr.interrupt)
 }
 
 // RegisterOrUpdateScaleUp registers scale-up for give node group or changes requested node increase
@@ -277,7 +292,7 @@ func (csr *ClusterStateRegistry) UpdateNodes(nodes []*apiv1.Node, nodeInfosForGr
 		return err
 	}
 
-	cloudProviderNodeInstances, err := getCloudProviderNodeInstances(csr.cloudProvider)
+	cloudProviderNodeInstances, err := csr.getCloudProviderNodeInstances()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -55,6 +56,11 @@ const (
 
 	// NodeGroupBackoffResetTimeout is the time after last failed scale-up when the backoff duration is reset.
 	NodeGroupBackoffResetTimeout = 3 * time.Hour
+
+	// CloudProviderNodeInstancesCacheRefreshInterval is the interval between cache refreshes.
```
losipiuk: between nodegroup instances cache refreshes.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -55,6 +56,11 @@ const (
...
+	CloudProviderNodeInstancesCacheRefreshInterval = 2 * time.Minute
```
losipiuk: CloudProviderNodeInstancesCacheRefreshInterval

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -55,6 +56,11 @@ const (
...
+	// CloudProviderNodeInstancesCacheEntryFreshnessInterval is the duration when cache entry is fresh.
+	CloudProviderNodeInstancesCacheEntryFreshnessInterval = 3 * time.Minute
```
losipiuk: CloudProviderNodeInstancesCacheEntryFreshnessThreshold?

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -1090,3 +1115,114 @@ func fakeNode(instance cloudprovider.Instance) *apiv1.Node {
...
+	}
+	return results, nil
+}
+
+// GetCloudProviderNodeInstancesForNodeGroup returns cloud provider node instances for the given node group.
+func (cache *CloudProviderNodeInstancesCache) GetCloudProviderNodeInstancesForNodeGroup(nodeGroup cloudprovider.NodeGroup) ([]cloudprovider.Instance, error) {
+	cacheEntry, found := cache.getCacheEntryLocked(nodeGroup)
+	if found {
+		if cacheEntry.isStale() {
+			klog.Warningf("Entry cloudProviderNodeInstances is stale, refresh time is %v", cacheEntry.refreshTime)
```
losipiuk: If we are doing refresh every two minutes and freshness interval is 3 minutes we will be seeing a lot of those.  If we dozens/hundreds of node groups we will not manage to refresh all within 1 minut (3-1)

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -1090,3 +1115,114 @@ func fakeNode(instance cloudprovider.Instance) *apiv1.Node {
...
+	}
+	cache.updateCacheEntryLocked(nodeGroup, &cloudProviderNodeInstancesCacheEntry{nodeGroupInstances, time.Now()})
+	return nodeGroupInstances, nil
+
+}
+
+// InvalidateCacheEntry removes entry for the given node group from cache.
+func (cache *CloudProviderNodeInstancesCache) InvalidateCacheEntry(nodeGroup cloudprovider.NodeGroup) {
+	klog.V(5).Infof("Invalidate entry in cloud provider node instances cache %v", nodeGroup.Id())
+	cache.removeCacheEntryLocked(nodeGroup)
```
losipiuk: I am thinking that maybe instead removing entry here we can just make it stale.
Then we can change the refresh algorithm in such way that we are running it more often (say every 10 seconds) and it refreshes only stale entries?

Then if we have many scale-ups in progress we still will not wait for all to be refreshed every loop. But we will prioritize that those are refreshed more often.

I am thinking about the case when we have multiple GPU-enabled nodegroups for which scale-ups take a lot of time (as GPU nodes are slow to come up). Then we can end up with each loop still taking lots of time. 

But maybe this is premature optimization. WDYT?

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -1090,3 +1115,114 @@ func fakeNode(instance cloudprovider.Instance) *apiv1.Node {
...
+}
+
+// Refresh refreshes cache.
+func (cache *CloudProviderNodeInstancesCache) Refresh() {
+	klog.Infof("Start refreshing cloud provider node instances cache")
+	defer klog.Infof("Refresh cloud provider node instances cache finished")
```
losipiuk: nit: You can log how much it took

cluster-autoscaler/clusterstate/clusterstate_test.go
```diff
@@ -781,3 +782,68 @@ func TestIsNodeStillStarting(t *testing.T) {
 func newBackoff() backoff.Backoff {
 	return backoff.NewIdBasedExponentialBackoff(InitialNodeGroupBackoffDuration, MaxNodeGroupBackoffDuration, NodeGroupBackoffResetTimeout)
 }
+
+func Test(t *testing.T) {
```
losipiuk: Please name it somewhat more specific ;)

cluster-autoscaler/main.go
```diff
@@ -314,6 +314,9 @@ func run(healthCheck *metrics.HealthCheck) {
 	// Start updating health check endpoint.
 	healthCheck.StartMonitoring()
 
+	// Start components running in background.
+	autoscaler.Start()
```
losipiuk: Either handle error here or drop it from signature

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -1090,3 +1115,114 @@ func fakeNode(instance cloudprovider.Instance) *apiv1.Node {
 		},
 	}
 }
+
```
losipiuk: please move cache to separate file.

cluster-autoscaler/clusterstate/clusterstate.go
krzysztof-jastrzebski: I see following problem with such solution: We need to refresh all instances every x minutes. If there is a lot of instance to be refreshed during one refresh then we might end up waiting longer than 10 seconds for refreshing scaled up node groups. Other node groups can starve scaled up node groups for some time.

cluster-autoscaler/clusterstate/clusterstate.go
krzysztof-jastrzebski: done

cluster-autoscaler/clusterstate/clusterstate.go
krzysztof-jastrzebski: What needs to be changed

cluster-autoscaler/clusterstate/clusterstate.go
krzysztof-jastrzebski: done

cluster-autoscaler/clusterstate/clusterstate.go
krzysztof-jastrzebski: changed

cluster-autoscaler/clusterstate/clusterstate.go
krzysztof-jastrzebski: done

cluster-autoscaler/clusterstate/clusterstate_test.go
krzysztof-jastrzebski: done

cluster-autoscaler/clusterstate/clusterstate.go
krzysztof-jastrzebski: done

krzysztof-jastrzebski:
Comments resolved.

cluster-autoscaler/clusterstate/clusterstate.go
losipiuk: I was thinking usign Threshold instead of interval. But ignore this. Not important.

losipiuk: Nothing here. I edited it a few times. And got to same version as original one :)

cluster-autoscaler/main.go
krzysztof-jastrzebski: done

losipiuk:
/lgtm
/approve
Thanks!

PR #2079, opened 2019-05-30
Account for min Limit from LimitRange.

bskiba:
Refactoring to make the code consistent.
Removed applying to upper and lower bound - I was mistaken that
it was needed >.<.

Note that since the recommendation is already capped
to max/min limit, there is no additional need for capping
in the GetContainerResources method.

### DIFF
--- a/vertical-pod-autoscaler/pkg/utils/vpa/capping.go
+++ b/vertical-pod-autoscaler/pkg/utils/vpa/capping.go
@@ -18,6 +18,7 @@ package api
 
 import (
 	"fmt"
+
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2"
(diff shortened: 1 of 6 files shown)


vertical-pod-autoscaler/pkg/utils/vpa/capping.go
```diff
@@ -34,8 +35,10 @@ func NewCappingRecommendationProcessor(limitsRangeCalculator limitrange.LimitRan
 type cappingAction string
 
 var (
-	cappedToMinAllowed cappingAction = "capped to minAllowed"
-	cappedToMaxAllowed cappingAction = "capped to maxAllowed"
+	cappedToMinAllowed             cappingAction = "capped to minAllowed"
+	cappedToMaxAllowed             cappingAction = "capped to maxAllowed"
+	cappedProportionallyToMaxLimit cappingAction = "capped to fit Max in LimitRange"
+	cappedProportionallyToMinLimit cappingAction = "capped to fit Min in LimitRange"
```
jbartosik: ...in container LimitRange" ?

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
```diff
@@ -278,3 +248,55 @@ func getContainer(containerName string, pod *apiv1.Pod) *apiv1.Container {
 	}
 	return nil
 }
+
+// applyLimitRange updates recommendation if recommended resources are outside of limits defined in VPA resources policy
+func applyLimitRange(recommendation apiv1.ResourceList, container apiv1.Container, limitRange *apiv1.LimitRangeItem) []string {
```
jbartosik: applyPodLimitRange?

bskiba: I'm confused, should it be pod LimitRange or container LimitRange?

jbartosik: Container limit range, sorry for the confusion.

bskiba: Done

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
bskiba: Done

PR #2080, opened 2019-05-31
Pod limit range

jbartosik:


jbartosik:
@bskiba I'll sync & ping you after I finish

jbartosik:
@bskiba rebased, PTAL

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
```diff
@@ -300,3 +304,71 @@ func getBoundaryRecommendation(recommendation apiv1.ResourceList, container apiv
...
+
+func applyPodLimitRange(resources []vpa_types.RecommendedContainerResources,
+	pod *apiv1.Pod, limitRange apiv1.LimitRangeItem, resourceName apiv1.ResourceName,
+	get func(vpa_types.RecommendedContainerResources) *apiv1.ResourceList) []vpa_types.RecommendedContainerResources {
+	minLimit := limitRange.Min[resourceName]
+	maxLimit := limitRange.Max[resourceName]
+	defaultLimit := limitRange.Default[resourceName]
+
+	var sumLimit resource.Quantity
+	for i, item := range pod.Spec.Containers {
```
bskiba: s/item/container/

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
```diff
@@ -300,3 +304,71 @@ func getBoundaryRecommendation(recommendation apiv1.ResourceList, container apiv
...
+func (c *cappingRecommendationProcessor) capProportionallyToPodLimitRange(
+	containerRecommendations []vpa_types.RecommendedContainerResources, pod *apiv1.Pod) ([]vpa_types.RecommendedContainerResources, error) {
+	podLimitRange, err := c.limitsRangeCalculator.GetPodLimitRangeItem(pod.Namespace)
+	if err != nil {
+		return nil, fmt.Errorf("error obtaining limit range: %s", err)
+	}
+	if podLimitRange == nil {
+		return containerRecommendations, nil
+	}
+	getLower := func(r vpa_types.RecommendedContainerResources) *apiv1.ResourceList {
```
bskiba: I've mentioned in the PR before that it is actually not needed to compute lower and upper as well, I was mistaken.

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/utils/vpa/capping_test.go
```diff
@@ -276,11 +281,218 @@ func TestApplyCapsToLimitRange(t *testing.T) {
...
+								},
+								Limits: apiv1.ResourceList{
+									apiv1.ResourceCPU: resource.MustParse("1"),
+								},
+							},
+						},
+					},
+				},
+			},
+			limitRange: apiv1.LimitRangeItem{
```
bskiba: is Pod type the default?

vertical-pod-autoscaler/pkg/utils/vpa/capping_test.go
```diff
@@ -276,11 +281,218 @@ func TestApplyCapsToLimitRange(t *testing.T) {
...
+	tests := []struct {
+		name         string
+		resources    []vpa_types.RecommendedContainerResources
+		pod          apiv1.Pod
+		limitRange   apiv1.LimitRangeItem
+		resourceName apiv1.ResourceName
+		expect       []vpa_types.RecommendedContainerResources
+	}{
+		{
+			name: "cap target cpu to max",
```
bskiba: Maybe also both min and max

bskiba:
I was thinking how pod limit and container limit will work together. Currently, if we trim due to podLimit, we give each of the containers limits proportionally to the current recommended resources, right? But then it might turn out that we have to trim to the containerLimit anyway and we end up with some unused podLimit and another container will have unnecessarily trimmed resources. But I guess it is sort of an edgecase anyway.

vertical-pod-autoscaler/pkg/utils/vpa/capping_test.go
jbartosik: No, it doesn't matter for this test but I see how this could be confusing. Done.

vertical-pod-autoscaler/pkg/utils/vpa/capping_test.go
jbartosik: Test is now redundant. I'll change limit ranges to have both min and max.

bskiba:
Actually those comments are not so important
/lgtm
/approve

PR #2081, opened 2019-05-31
Retain information about scale-up failures in CSR

towca:
This will provide the AutoscalingStatusProcessor with information
about failed scale-ups.

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -177,6 +177,7 @@ func (a *StaticAutoscaler) cleanUpIfRequired() {
 func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError {
 	a.cleanUpIfRequired()
 	a.processorCallbacks.reset()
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -240,6 +240,9 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 		if err != nil {
 			klog.Errorf("AutoscalingStatusProcessor error: %v.", err)
 		}
+
+		// Clear the scale-up failures info each loop so they don't accumulate.
+		a.clusterStateRegistry.ClearScaleUpFailures()
```
losipiuk: It seems too explicit.

I suggest you add `clusterStateRegistry.OnLoopStart()`. Then make `ClearScaleUpFailures` private and call it from `OnLoopStart`. That way we will be able to add more "per-loop-cleanup" routines into `OnLoopStart` without overgrowing `clusterStateRegistry` interface

towca: Agreed, done!

PR #2082, opened 2019-05-31
Optimize building node infos by using map with pods for nodes.

krzysztof-jastrzebski:
Optimize building node infos by using map with pods for nodes.

### DIFF
--- a/cluster-autoscaler/simulator/nodes.go
+++ b/cluster-autoscaler/simulator/nodes.go
@@ -23,25 +23,14 @@ import (
 	policyv1 "k8s.io/api/policy/v1beta1"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/drain"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/simulator/nodes.go
```diff
@@ -23,25 +23,14 @@ import (
...
-		return []*apiv1.Pod{}, errors.ToAutoscalerError(errors.ApiCallError, err)
-	}
-	allPods := make([]*apiv1.Pod, 0)
-	for _, p := range pods {
-		if p.Spec.NodeName == nodename {
-			allPods = append(allPods, p)
-		}
-	}
-
+func GetRequiredPodsForNode(nodename string, podsForNodes map[string][]*apiv1.Pod) ([]*apiv1.Pod, errors.AutoscalerError) {
```
losipiuk: btw. Does this have to be public.

krzysztof-jastrzebski: Done.

PR #2094, opened 2019-06-05
fix: ignore agentpool label when looking for similar node groups with Azure provider

### LINKED ISSUE #2044: Azure CA doesn't scale multiple agent pools in parallel
Scale-up plans in Azure never include multiple agent pools, even when it is needed to make room for new pods. It looks like, for some reason, the implementation for Azure never includes more than one agent pool in the same scale-up plan. The consecuence is that it serializes the operations, scaling only one agent pool at a time. Can you confirm the issue or whether it is the expected behavior?

At least cluster-autoscaler 1.3.9 for Azure (K8S 1.11.8) is affected. The autoscaler works (all the agent pools end up being scaled) but the issue makes it much slower than other implementations (e.g. AWS is able to scale multiple node pools in parallel, as part of the same same scale-up plan). With N agent pools in the cluster that need to be scaled and M minutes to scale each one, you need to wait N x M minutes instead of just M.

CecileRobertMichon:
Related to #2044 

Currently, the cluster-autoscaler never puts two different agent pools in the scale-up plan when used with Azure clusters deployed with aks-engine or AKS. Each node has an "agentpool" label that identifies its node pool. This PR adds the "agentpool" label to the ignored labels of `IsNodeInfoSimilar` in order to allow two similar node groups with different "agentpool" labels to be detected. 

Also added documentation on using `--balance-similar-node-groups` in the Azure README.

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/README.md
+++ b/cluster-autoscaler/cloudprovider/azure/README.md
@@ -68,6 +68,12 @@ or to autoscale multiple VM scale sets:
         - --nodes=1:10:k8s-nodepool-2-vmss
 ```
 
+To allow scaling similar node pools simultaneously, or when using separate node groups per zone and to keep nodes balanced across zones, use the `--balance-similar-node-groups` flag. Add it to the `command` section to enable it:
+
+```yaml
+        - --balance-similar-node-groups=true
+```
+
 Save the updated deployment manifest, then deploy cluster-autoscaler by running:
 
 ```sh
--- a/cluster-autoscaler/cloudprovider/azure/README.md
+++ b/cluster-autoscaler/cloudprovider/azure/README.md
@@ -74,6 +74,8 @@ To allow scaling similar node pools simultaneously, or when using separate node 
         - --balance-similar-node-groups=true
 ```
 
+See the [FAQ](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#im-running-cluster-with-nodes-in-multiple-zones-for-ha-purposes-is-that-supported-by-cluster-autoscaler) for more details.
+
 Save the updated deployment manifest, then deploy cluster-autoscaler by running:
 
 ```sh
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -30,6 +30,7 @@ import (
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/azure"
 	cloudBuilder "k8s.io/autoscaler/cluster-autoscaler/cloudprovider/builder"
 	"k8s.io/autoscaler/cluster-autoscaler/config"
 	"k8s.io/autoscaler/cluster-autoscaler/core"
@@ -37,6 +38,7 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/expander"
 	"k8s.io/autoscaler/cluster-autoscaler/metrics"
 	ca_processors "k8s.io/autoscaler/cluster-autoscaler/processors"
+	"k8s.io/autoscaler/cluster-autoscaler/processors/nodegroupset"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
 	kube_util "k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/units"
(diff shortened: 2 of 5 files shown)


cluster-autoscaler/cloudprovider/azure/README.md
```diff
@@ -68,6 +68,12 @@ or to autoscale multiple VM scale sets:
         - --nodes=1:10:k8s-nodepool-2-vmss
 ```
 
+To allow scaling similar node pools simultaneously, or when using separate node groups per zone and to keep nodes balanced across zones, use the `--balance-similar-node-groups` flag. Add it to the `command` section to enable it:
+
+```yaml
+        - --balance-similar-node-groups=true
```
CecileRobertMichon: @feiskyer do you know if this flag defaults to true?

MaciekPytel:
The whole point of making node group comparison a 'processor' (ie. hiding it behind an interface) was that such cloudprovider specific changes can be easily isolated. Please check this commit, which shows code addressing the exact same problem for GKE: https://github.com/kubernetes/autoscaler/commit/34a4262ad884b4b6fdaf20a0321b58015f18d8a2#diff-83aae18ad286ee688abea380ae132efa (the commit removes the relevant code, since GKE is now supported from fork and there was no reason to keep GKE specific logic in main repo, but the code still illustrates our preferred pattern).

cluster-autoscaler/cloudprovider/azure/README.md
feiskyer: the default value is false.

feiskyer:
@CecileRobertMichon Could you update the PR according to  34a4262#diff-83aae18ad286ee688abea380ae132efa?

cluster-autoscaler/main.go
```diff
@@ -285,6 +287,10 @@ func buildAutoscaler() (core.Autoscaler, error) {
 
 	processors := ca_processors.DefaultProcessors()
 	processors.PodListProcessor = core.NewFilterOutSchedulablePodListProcessor()
+	if autoscalingOptions.CloudProviderName == azure.ProviderName {
+		processors.NodeGroupSetProcessor = &nodegroupset.BalancingNodeGroupSetProcessor{
+			Comparator: nodegroupset.IsAzureNodeInfoSimilar}
```
CecileRobertMichon: this overrides the node comparator for Azure

cluster-autoscaler/processors/nodegroupset/compare_nodegroups.go
```diff
@@ -52,12 +52,45 @@ func compareResourceMapsWithTolerance(resources map[apiv1.ResourceName][]resourc
 	return true
 }
 
+func compareLabels(nodes []*schedulernodeinfo.NodeInfo, ignoredLabels map[string]bool) bool {
```
CecileRobertMichon: refactored this into a function

cluster-autoscaler/processors/nodegroupset/compare_nodegroups.go
```diff
@@ -52,12 +52,41 @@ func compareResourceMapsWithTolerance(resources map[apiv1.ResourceName][]resourc
...
+	}
+	return true
+}
+
 // IsNodeInfoSimilar returns true if two NodeInfos are similar enough to consider
 // that the NodeGroups they come from are part of the same NodeGroupSet. The criteria are
 // somewhat arbitrary, but generally we check if resources provided by both nodes
 // are similar enough to likely be the same type of machine and if the set of labels
 // is the same (except for a pre-defined set of labels like hostname or zone).
 func IsNodeInfoSimilar(n1, n2 *schedulernodeinfo.NodeInfo) bool {
```
CecileRobertMichon: this is a thin wrapper around the existing function that provides the default set of ignored labels for all cloud providers.

CecileRobertMichon:
@MaciekPytel thanks for the suggestion. I've updated the PR, let me know what you think. I wasn't able to just reuse the implementation for GKE (although I did reuse the same pattern for overriding the node comparator) because the requirement here is slightly different: the GKE implementation allows two nodes to be "similar" if they have the _same_ node pool label (regardless of their other characteristics). What we're trying to do here for Azure is to allow two node groups to be "similar" if even if their node pool label is _ different_, given that everything else matches. The reason for doing this is to allow for agent pools that have the same configuration to be considered as similar even if their "agentpool" label differs (since they are in different pools). @feiskyer thoughts?

cluster-autoscaler/processors/nodegroupset/compare_nodegroups.go
```diff
@@ -52,12 +52,41 @@ func compareResourceMapsWithTolerance(resources map[apiv1.ResourceName][]resourc
...
+	ignoredLabels := map[string]bool{
```
feiskyer: those labels may be changed in the future. Could we make a variable for it and share also in IsAzureNodeInfoSimilar()?

CecileRobertMichon: done

feiskyer:
@MaciekPytel @losipiuk Could you help to take a look at this PR again?

CecileRobertMichon:
@MaciekPytel @losipiuk please take a look.

losipiuk:
Makes sense but we cannot explicitly call Azure CP from main as we have build tags which allow building without Azure code at all. Could you think of way to extend CP interface to hide such CP specific customization behind it? Sth like CP.setupProcessors(originalProcessors). This would allow to replace/wrap/extend processors set as needed by CP.

feiskyer:
#2171 is working on https://github.com/kubernetes/autoscaler/pull/2094#issuecomment-502979184, so close this and prefer #2171.

/close

PR #2102, opened 2019-06-06
 Updating vendor against git@github.com:kubernetes/kubernetes.git:master (ef7808fec57284582c7bf3fc834570e5769df83e

losipiuk:


losipiuk:
Update was done using script from https://github.com/kubernetes/autoscaler/pull/2103.

@feiskyer Please note I removed the _override directory. Everything compiles and unit tests pass but please see if some Azure specific dependencies do not need to be added to go.mod-extra

feiskyer:
> @feiskyer Please note I removed the _override directory. Everything compiles and unit tests pass but please see if some Azure specific dependencies do not need to be added to go.mod-extra

It's ok since the override parts are added automatically now in the vendor directory.

PR #2110, opened 2019-06-10
Do not fail loop iteration if unregistered nodes cannot be removed

losipiuk:


krzysztof-jastrzebski:
Could you write in description why we are changing this behavior?

losipiuk:
> Could you write in description why we are changing this behavior?

Updated commit message

PR #2112, opened 2019-06-10
Remove legacy selector fetcher

bskiba:
In preparation for introducing v1 VPA API.

@schylek would you mind taking a look?


vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -471,21 +464,11 @@ func (feeder *clusterStateFeeder) validateTargetRef(vpa *vpa_types.VerticalPodAu
 }
 
 func (feeder *clusterStateFeeder) getSelector(vpa *vpa_types.VerticalPodAutoscaler) (labels.Selector, []condition) {
-	legacySelector, fetchLegacyErr := feeder.legacySelectorFetcher.Fetch(vpa)
-	if fetchLegacyErr != nil {
-		klog.Errorf("Error while fetching legacy selector. Reason: %+v", fetchLegacyErr)
-	}
 	selector, fetchErr := feeder.selectorFetcher.Fetch(vpa)
 	if fetchErr != nil {
```
schylek: Seems like this could be moved to if under L485 for simplicity..

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -113,17 +111,16 @@ func NewClusterStateFeeder(config *rest.Config, clusterState *model.ClusterState
...
 	controllerFetcher := controllerfetcher.NewControllerFetcher(config, kubeClient, factory)
 	return ClusterStateFeederFactory{
-		PodLister:             podLister,
-		OOMObserver:           oomObserver,
-		KubeClient:            kubeClient,
-		MetricsClient:         newMetricsClient(config),
-		VpaCheckpointClient:   vpa_clientset.NewForConfigOrDie(config).AutoscalingV1beta2(),
-		VpaLister:             vpa_api_util.NewAllVpasLister(vpa_clientset.NewForConfigOrDie(config), make(chan struct{})),
-		ClusterState:          clusterState,
-		LegacySelectorFetcher: target.NewBeta1TargetSelectorFetcher(config),
```
schylek: I can see that LegacySelectorFetcher is deleted, so it means that v1beta1 objects will be read as v1beta2 and marked as UnsupportedConfig since they don't have target ref. Is it expected? Don't we want to have additional information for the user that they use v1beta1?

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
bskiba: Done

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
bskiba: Added more info to the error. Thanks@

PR #2120, opened 2019-06-12
VPA uses v1 API version

bskiba:
Also switches e2e tests to deploy the v1 crd.


vertical-pod-autoscaler/pkg/target/v1beta1_fetcher.go
```diff
@@ -0,0 +1,87 @@
+/*
```
jbartosik: This file looks unused - why is it part of the PR?

bskiba: Ah, thanks! Rebase confusion. Removed.

bskiba:
Gah, I broke something while fixing the other thing, one more round.

jbartosik:
Looks like some tests still use objects from wrong API version. Otherwise LGTM

PR #2116, opened 2019-06-11
Cherry limits to 04

jbartosik:
Cherry pick:

PRs preparing VPA for limit and limit range support:
#1879
#1880
#1950
#1990
PRs implementing support for limits:
#1993
#1997
#1998
#2005

This is analogous to #2105 but this one had conflicts (mostly import changes, one in unit tests)

PR #2106, opened 2019-06-07
Cherry pick limit support

jbartosik:
Follow up to https://github.com/kubernetes/autoscaler/pull/2105 which had commits that cherry picked automatically. Some commits in this PR are not exact cherry pick and needed changes to

* vertical-pod-autoscaler/deploy/vpa-rbac.yaml
* vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
* vertical-pod-autoscaler/pkg/admission-controller/logic/server_test.go
* vertical-pod-autoscaler/pkg/admission-controller/main.go

Because master has VPA preprocessor but vpa-release-0.5 doesn't (and some changes to RBAC). Each commit notes in its message any files I modified manually.

Cherry pick:
* https://github.com/kubernetes/autoscaler/pull/2049
* https://github.com/kubernetes/autoscaler/pull/2068
* https://github.com/kubernetes/autoscaler/pull/2072
* https://github.com/kubernetes/autoscaler/pull/2073
* https://github.com/kubernetes/autoscaler/pull/2076
* https://github.com/kubernetes/autoscaler/pull/2077
* https://github.com/kubernetes/autoscaler/pull/2079
* https://github.com/kubernetes/autoscaler/pull/2080
* https://github.com/kubernetes/autoscaler/pull/2091
* https://github.com/kubernetes/autoscaler/pull/2092
* https://github.com/kubernetes/autoscaler/pull/2093
* https://github.com/kubernetes/autoscaler/pull/2095
* https://github.com/kubernetes/autoscaler/pull/2099
* https://github.com/kubernetes/autoscaler/pull/2100

PR #2105, opened 2019-06-07
Cherry limit 05

jbartosik:
Cherry pick:
* PRs preparing VPA for limit and limit range support:
  * https://github.com/kubernetes/autoscaler/pull/1879
  * https://github.com/kubernetes/autoscaler/pull/1880
  * https://github.com/kubernetes/autoscaler/pull/1950
  * https://github.com/kubernetes/autoscaler/pull/1990
* PRs implementing support for limits:
  * https://github.com/kubernetes/autoscaler/pull/1993
  * https://github.com/kubernetes/autoscaler/pull/1997
  * https://github.com/kubernetes/autoscaler/pull/1998
  * https://github.com/kubernetes/autoscaler/pull/2005

PR #2123, opened 2019-06-13
Update VPA README with description of limit control

jbartosik:


bskiba:
I propose to create the 0.6.0 release and do an update of the README together with updating the deploy scripts - currently they default is still 0.5.0

jbartosik:
> I propose to create the 0.6.0 release and do an update of the README together with updating the deploy scripts - currently they default is still 0.5.0

I cherry picked limit changes to `vpa-release-0.4` (#2116, #2117) and `vpa-release-0.5` (#2105, #2106) so I think we need those changes to README on `master`, `vpa-release-0.4`, `vpa-release-0.5`, and `vpa-release-0.6`.

I thought that I would cherry pick README change to release branches when building new patch releases.

jbartosik:
@bskiba please take a look.

vertical-pod-autoscaler/README.md
```diff
@@ -214,6 +218,18 @@ kubectl delete clusterrolebinding myname-cluster-admin-binding
 
 # Known limitations
 
+## Limits control
```
bskiba: I don't think this section belongs in Known limitations.

vertical-pod-autoscaler/README.md
```diff
@@ -214,6 +218,18 @@ kubectl delete clusterrolebinding myname-cluster-admin-binding
...
+
+When setting limits VPA will conform to
+[resource policies](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2/types.go#L82).
+It will maintain limit to request ratio specified for all containers.
+
+VPA will also attempt to fit within
+[limit ranges](https://kubernetes.io/docs/concepts/policy/limit-range/).
+However if maintaining limit/request ratio, following resource policies, and
```
bskiba: This sentence is very difficult to parse. I think we should write that we try to cap to max/min limit, but that VPA policy (minRecommended/maxRecommended) take precedence. It may clarify things a bit if we give concrete examples maybe?

bskiba: Concrete examples = concrete request,limit,ratio, limitrange values?

vertical-pod-autoscaler/README.md
jbartosik: Moved.

vertical-pod-autoscaler/README.md
jbartosik: Reworded and added examples

bskiba:
Thanks!
/lgtm
/approve

PR #2127, opened 2019-06-18
Move pod packing in upcoming nodes to RunOnce from Estimator

vivekbagade:
For performance reasons, move to pod bin packing on upcoming nodes to RunOnce


cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -135,6 +135,10 @@ type AutoscalingOptions struct {
 	// Setting it to false employs a more lenient filtering approach that does not try to pack the pods on the nodes.
 	// Pods with nominatedNodeName set are always filtered out.
 	FilterOutSchedulablePodsUsesPacking bool
+	// FilterOutSchedulablePodsInUpcomingNodes if set, will prevent CA from considering pods that can
```
losipiuk: Seems like `AgainsUpcomingNodes` would be better than `InUpcomingNodes`

cluster-autoscaler/core/filter_out_schedulable.go
```diff
@@ -63,11 +64,17 @@ func (filterOutSchedulablePodListProcessor) Process(
...
+
+	if context.FilterOutSchedulablePodsInUpcomingNodes {
+		unschedulablePodsToHelp = filterOutSchedulableByPacking(unschedulablePods, upcomingNodes, allScheduledPods,
+			context.PredicateChecker, context.ExpendablePodsPriorityCutoff, false)
+	}
+
 	if context.FilterOutSchedulablePodsUsesPacking {
-		unschedulablePodsToHelp = filterOutSchedulableByPacking(unschedulablePods, readyNodes, allScheduledPods,
-			context.PredicateChecker, context.ExpendablePodsPriorityCutoff)
+		unschedulablePodsToHelp = filterOutSchedulableByPacking(unschedulablePodsToHelp, readyNodes, allScheduledPods,
```
losipiuk: It can be unset if `FilterOutSchedulablePodsInUpcomingNodes` is false.
The less error prone approach would be to:
* create `tmpUnschedulablePods`
* start with assigning `unschedulablePods` to it.
* then always use the `tmpUnschedulablePods` as both imput and output when `filterOut*` functions are called.
* finally assign `tmpUnschedulablePods` to `unschedulablePodsToHelp`

cluster-autoscaler/core/filter_out_schedulable.go
```diff
@@ -104,12 +112,18 @@ func filterOutSchedulableByPacking(unschedulableCandidates []*apiv1.Pod, nodes [
 		if err != nil {
 			unschedulablePods = append(unschedulablePods, pod)
 		} else {
-			glogx.V(4).UpTo(loggingQuota).Infof("Pod %s marked as unschedulable can be scheduled on %s. Ignoring in scale up.", pod.Name, nodeName)
+			nodeType := "existing"
```
losipiuk: I would prefer
```go
var nodeType string
if nodeExists {
  nodeType = "existing"
else 
  nodeType = "upcoming"
}
```

cluster-autoscaler/estimator/basic_estimator.go
```diff
@@ -98,51 +98,27 @@ func (basicEstimator *BasicNodeEstimator) GetDebug() string {
 }
 
 // Estimate estimates the number needed of nodes of the given shape.
-func (basicEstimator *BasicNodeEstimator) Estimate(pods []*apiv1.Pod, nodeInfo *schedulernodeinfo.NodeInfo, upcomingNodes []*schedulernodeinfo.NodeInfo) int {
```
losipiuk: what if `FilterOutSchedulablePodsInUpcomingNodes` is `false` then we will not handle upcoming nodes at all. 
That is broken, right?

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -588,3 +590,18 @@ func allPodsAreNew(pods []*apiv1.Pod, currentTime time.Time) bool {
 	found, oldest := getOldestCreateTimeWithGpu(pods)
 	return found && oldest.Add(unschedulablePodWithGpuTimeBuffer).After(currentTime)
 }
+
+func getUpcomingNodeInfos(registry *clusterstate.ClusterStateRegistry, nodeInfos map[string]*schedulernodeinfo.NodeInfo) []*apiv1.Node {
```
losipiuk: The tests using mocks are unreadable. Do we have test covering this method anyway? Maybe writing a unit test to just cover this one would be beneficial?

vivekbagade: Just discovered that the tests with mocks are broken as they are stateful. There is a cache used in scale up. I hacked it by passing a distinct node group id. We need to fix it properly soon.

To answer your question, there are tests that cover this method already.

cluster-autoscaler/estimator/basic_estimator.go
vivekbagade: Solved

cluster-autoscaler/core/filter_out_schedulable.go
vivekbagade: Don't think this makes much of difference for readability, changed it anyways

cluster-autoscaler/core/filter_out_schedulable.go
vivekbagade: Solved a little differently

cluster-autoscaler/config/autoscaling_options.go
vivekbagade: Solved.

cluster-autoscaler/core/static_autoscaler.go
vivekbagade: The hack doesn't work in all cases, so, I have changed the test case a little. Will discuss with you tomorrow.

cluster-autoscaler/estimator/estimator.go
```diff
@@ -26,18 +26,14 @@ import (
 )
 
 const (
-	//BasicEstimatorName is the name of basic estimator.
-	BasicEstimatorName = "basic"
 	// BinpackingEstimatorName is the name of binpacking estimator.
 	BinpackingEstimatorName = "binpacking"
+	// BinpackingOldEstimatorName is the name of the older binpacking estimator.
+	OldBinpackingEstimatorName = "oldbinpacking"
```
losipiuk: shouldn't this one be deprecated to and marked as to be removed soon?

cluster-autoscaler/estimator/estimator.go
```diff
@@ -55,11 +51,10 @@ func NewEstimatorBuilder(name string) (EstimatorBuilder, error) {
...
 	// Deprecated.
-	// TODO(aleksandra-malinowska): remove in 1.5.
-	case BasicEstimatorName:
-		klog.Warning(basicEstimatorDeprecationMessage)
-		return func(_ *simulator.PredicateChecker) Estimator {
-			return NewBasicNodeEstimator()
+	case OldBinpackingEstimatorName:
+		klog.Warning(oldBinPackingEstimatorDeprecationMessage)
+		return func(predicateChecker *simulator.PredicateChecker) Estimator {
+			return NewOldBinpackingNodeEstimator(predicateChecker)
```
losipiuk: NewOld ... decide man ;)

cluster-autoscaler/estimator/old_binpacking_estimator.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"sort"
+
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/autoscaler/cluster-autoscaler/simulator"
+	schedulerUtils "k8s.io/autoscaler/cluster-autoscaler/utils/scheduler"
+	schedulernodeinfo "k8s.io/kubernetes/pkg/scheduler/nodeinfo"
+)
+
+const oldBinPackingEstimatorDeprecationMessage = "old binpacking estimator is deprecated. It will be removed in Cluster Autoscaler 1.15."
```
losipiuk: I did not check yet I assume this is just a copy of previoius verison of binpacking_estimator.go

cluster-autoscaler/estimator/estimator.go
vivekbagade: I knew you were going to catch that :D

cluster-autoscaler/estimator/old_binpacking_estimator.go
vivekbagade: Yes. The Estimate func

cluster-autoscaler/estimator/estimator.go
vivekbagade: Its in the desc of OldBinpackingEstimator. Just like it was in the case of BasicEstimator

vivekbagade:
> looks fine. Validation does not pass.

Fixed fmt

PR #2129, opened 2019-06-19
update images, they reference the wrong version

lawrencedudley:


Jeffwan:
@lawrencedudley Thanks for your first contribution!

PR #2131, opened 2019-06-19
update images, they reference the wrong version

lawrencedudley:


Jeffwan:
For 1.15. Could we update master directly and cherry-pick back to 1.15?

Jeffwan:
I fixed all rest of the branches. Let's merge it one and get 1.15 fixed.

PR #2133, opened 2019-06-21
Delaying events in the case of scaleup and scaleup error

vivekbagade:


cluster-autoscaler/processors/status/eventing_scale_up_processor.go
```diff
@@ -32,9 +32,12 @@ type EventingScaleUpStatusProcessor struct{}
 // Process processes the state of the cluster after a scale-up by emitting
 // relevant events for pods depending on their post scale-up status.
 func (p *EventingScaleUpStatusProcessor) Process(context *context.AutoscalingContext, status *ScaleUpStatus) {
-	for _, noScaleUpInfo := range status.PodsRemainUnschedulable {
-		context.Recorder.Event(noScaleUpInfo.Pod, apiv1.EventTypeNormal, "NotTriggerScaleUp",
-			fmt.Sprintf("pod didn't trigger scale-up (it wouldn't fit if a new node is added): %s", ReasonsMessage(noScaleUpInfo)))
+	if status.Result != ScaleUpSuccessful && status.Result != ScaleUpError {
```
losipiuk: Maybe log that you are suppressing event emission stating what was the reason for suppression

PR #2138, opened 2019-06-24
Validation to verify if node.Spec.ProviderID is nil - for GCE

emailksravi:


losipiuk:
/lgtm
/approve
Thanks!

PR #2142, opened 2019-06-26
Fixes/Improvements to update-vendor.sh

losipiuk:


cluster-autoscaler/hack/update-vendor.sh
```diff
@@ -17,16 +18,19 @@ fi
...
-OPTS=`getopt -o f::r::d::v:: --long k8sfork::,k8srev::,workdir:: -n $SCRIPT_NAME -- "$@"`
+OPTS=`getopt -o f::r::d::v::b:: --long k8sfork::,k8srev::,workdir::,batch:: -n $SCRIPT_NAME -- "$@"`
 if [ $? != 0 ] ; then echo "Failed parsing options." >&2 ; exit 1 ; fi
 eval set -- "$OPTS"
 while true; do
   case "$1" in
     -f | --k8sfork ) K8S_FORK="$2"; shift; shift ;;
     -r | --k8srev ) K8S_REV="$2"; shift; shift ;;
     -d | --workdir ) WORK_DIR="$2"; shift; shift ;;
+    -b | --batch ) BATCH_MODE="true"; shift; shift ;;
```
jkaniuk: I have a feeling that this is an overkill

losipiuk: I believe it does not add very much of compexity. And error messages without this switch would be very cumbersome if this script is called as part of other script.

PR #2146, opened 2019-06-28
Adding config for event filtering

vivekbagade:
CA Scalability improvement by throttling number of events
fired in the case where there are a lot of unschedulable
pods.

### DIFF
--- a/cluster-autoscaler/utils/kubernetes/factory.go
+++ b/cluster-autoscaler/utils/kubernetes/factory.go
@@ -27,12 +27,31 @@ import (
 	"k8s.io/klog"
 )
 
+const (
+	// Rate of refill for the event spam filter in client go
+	// 1 per event key per 5 minutes.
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/utils/kubernetes/factory.go
```diff
@@ -27,12 +27,31 @@ import (
 	"k8s.io/klog"
 )
 
+const (
+	// Rate of refill for the event spam filter in client go
+	// 1 per event key per 5 minutes.
+	defaultQPS = 1. / 300.
```
losipiuk: Why are those called `default*` if they cannot be overriden with something non-default?

vivekbagade: Default just means preselected. For eg. there were defaults in client-go that couldn't be overridden. 

I'll change it if you feel strongly about this.

PR #2147, opened 2019-06-28
Fix logging error

silenceper:
Fix logging error avoid panic
nodeGroup may be nil


silenceper:
I signed it

losipiuk:
/lgtm
/approve
Thanks!

PR #2149, opened 2019-06-30
nodeGroup judy IsNil to avoid crashed

yan234280533:
nodeGroup judy IsNil to avoid crashed

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -585,7 +585,7 @@ func (csr *ClusterStateRegistry) updateReadinessStats(currentTime time.Time) {
 			klog.Warningf("Failed to get nodegroup for %s: %v", unregistered.Node.Name, errNg)
 			continue
 		}
-		if nodeGroup == nil {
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -585,7 +585,7 @@ func (csr *ClusterStateRegistry) updateReadinessStats(currentTime time.Time) {
 			klog.Warningf("Failed to get nodegroup for %s: %v", unregistered.Node.Name, errNg)
 			continue
 		}
-		if nodeGroup == nil {
+		if nodeGroup == nil || reflect.ValueOf(nodeGroup).IsNil() {
```
losipiuk: Did you actually encounter this? I think the cloudprovider which returned nil wrapped in an interface should be fixed. Not this place.

yan234280533: I  actually encounter this. And i found other place judied the value is nil except here.

mwielgus:
Let's merge this to avoid buggy cloud providers to crash CA. This is lesser evil :/.

PR #2151, opened 2019-07-01
Fix Azure client requests stuck issues on http.StatusTooManyRequests

feiskyer:
In go-autorest SDK https://github.com/Azure/go-autorest/blob/master/autorest/sender.go#L242,
if ARM returns http.StatusTooManyRequests, the sender doesn't increase the retry attempt count,
hence the Azure clients will keep retrying forever until it gets a status code other than 429.

```go
// don't count a 429 against the number of attempts
// so that we continue to retry until it succeeds
if resp == nil || resp.StatusCode != http.StatusTooManyRequests {
    attempt++
}
```

So this PR explicitly remove `http.StatusTooManyRequests` from `autorest.StatusCodesForRetry`. To reduce the API throttling issues, it also adds caches for vmss instances.

Fix #2124.

/cc @andyzhangx @nilo19

PR #2152, opened 2019-07-01
Allow scaling multiple Azure vmss synchronously

### LINKED ISSUE #1984: bulk scale-up in azure creates only one node per iteration sometimes
I think that cluster-autoscaler (CA) 1.3.x in Azure has problems dealing with affinity rules.

I use the following deployment to deploy a "pause" pod with two rules:
- affinity: They must use a node in the agentpool named "genmlow"
- podAntiAffinity: pods must not be deployed in the same node

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: pause
  labels:
    app: pause
spec:
  replicas: 1
  selector:
    matchLabels:
      app: pause
  template:
    metadata:
      labels:
        app: pause
    spec:
      affinity:
        nodeAffinity:
            nodeSelectorTerms:
              - matchExpressions:
                  - key: poolName
                    operator: In
                    values:
                      - genmlow
        podAntiAffinity:
            - labelSelector:
                matchExpressions:
                  - key: app
                    operator: In
                    values:
                      - pause
              topologyKey: kubernetes.io/hostname
      containers:
      - image: "karlherler/pause:1.0"
        name: pause
```

The agentpool "genmlow" uses Standard_DS2_v2 machines (8GB) in a virtual machine scale set.

When I scale the number of replicas to 10 (`kubectl scale deployment pause --replicas=10`), I see that the cluster autoscaler (version 1.3.9, k8s 1.11.8) creates only one node per iteration, as if it was ignoring the affinity rules. See cluster-autoscaler logs, where nodes go from 0->1->2->...->N.

```
I0503 14:03:19.299146       1 azure_manager.go:261] Refreshed ASG list, next refresh after 2019-05-03 14:04:19.2991386 +0000 UTC m=+948.211672501
I0503 14:03:19.993383       1 scale_up.go:249] Pod default/pause-66cf84dcdb-2khzb is unschedulable
I0503 14:03:19.993412       1 scale_up.go:249] Pod default/pause-66cf84dcdb-l7587 is unschedulable
I0503 14:03:19.993418       1 scale_up.go:249] Pod default/pause-66cf84dcdb-t5mb8 is unschedulable
I0503 14:03:19.993422       1 scale_up.go:249] Pod default/pause-66cf84dcdb-xp2kn is unschedulable
I0503 14:03:19.993426       1 scale_up.go:249] Pod default/pause-66cf84dcdb-rpskf is unschedulable
I0503 14:03:19.993429       1 scale_up.go:249] Pod default/pause-66cf84dcdb-kkxc5 is unschedulable
I0503 14:03:19.993433       1 scale_up.go:249] Pod default/pause-66cf84dcdb-lbprj is unschedulable
I0503 14:03:19.993437       1 scale_up.go:249] Pod default/pause-66cf84dcdb-lmwmf is unschedulable
I0503 14:03:19.993441       1 scale_up.go:249] Pod default/pause-66cf84dcdb-c8njm is unschedulable
I0503 14:03:19.993446       1 scale_up.go:249] Pod default/pause-66cf84dcdb-gg6xh is unschedulable
...
I0503 14:03:20.071931       1 utils.go:187] Pod pause-66cf84dcdb-kkxc5 can't be scheduled on k8s-genl-24772259-vmss. Used cached predicate check results
I0503 14:03:20.072229       1 utils.go:187] Pod pause-66cf84dcdb-lbprj can't be scheduled on k8s-genl-24772259-vmss. Used cached predicate check results
I0503 14:03:20.072529       1 utils.go:187] Pod pause-66cf84dcdb-lmwmf can't be scheduled on k8s-genl-24772259-vmss. Used cached predicate check results
I0503 14:03:20.073242       1 utils.go:187] Pod pause-66cf84dcdb-c8njm can't be scheduled on k8s-genl-24772259-vmss. Used cached predicate check results
...
I0503 14:03:20.076758       1 scale_up.go:378] Best option to resize: k8s-genmlow-24772259-vmss
I0503 14:03:20.076770       1 scale_up.go:382] Estimated 1 nodes needed in k8s-genmlow-24772259-vmss
I0503 14:03:20.076783       1 scale_up.go:461] Final scale-up plan: [{k8s-genmlow-24772259-vmss 0->1 (max: 1000)}]
I0503 14:03:20.076796       1 scale_up.go:531] Scale-up: setting group k8s-genmlow-24772259-vmss size to 1
...
I0503 14:06:13.334377       1 scale_up.go:378] Best option to resize: k8s-genmlow-24772259-vmss
I0503 14:06:13.334411       1 scale_up.go:382] Estimated 1 nodes needed in k8s-genmlow-24772259-vmss
I0503 14:06:13.334470       1 scale_up.go:461] Final scale-up plan: [{k8s-genmlow-24772259-vmss 1->2 (max: 1000)}]
I0503 14:06:13.334503       1 scale_up.go:531] Scale-up: setting group k8s-genmlow-24772259-vmss size to 2
...
I0503 14:09:02.059191       1 scale_up.go:378] Best option to resize: k8s-genmlow-24772259-vmss
I0503 14:09:02.059243       1 scale_up.go:382] Estimated 1 nodes needed in k8s-genmlow-24772259-vmss
I0503 14:09:02.059310       1 scale_up.go:461] Final scale-up plan: [{k8s-genmlow-24772259-vmss 2->3 (max: 1000)}]
I0503 14:09:02.059350       1 scale_up.go:531] Scale-up: setting group k8s-genmlow-24772259-vmss size to 3
...
I0503 14:11:50.214206       1 scale_up.go:378] Best option to resize: k8s-genmlow-24772259-vmss
I0503 14:11:50.214228       1 scale_up.go:382] Estimated 1 nodes needed in k8s-genmlow-24772259-vmss
I0503 14:11:50.214245       1 scale_up.go:461] Final scale-up plan: [{k8s-genmlow-24772259-vmss 3->4 (max: 1000)}]
I0503 14:11:50.214262       1 scale_up.go:531] Scale-up: setting group k8s-genmlow-24772259-vmss size to 4
...
...
```

However, it only behaves this way when the pod has no requests. If I add the following requests:

```yaml
  resources:
    requests:
      memory: 5Gi
```

Everything works as expected and the cluster autoscaler works as expected, creating the 10 virtual machines in a single batch (1->10). I guess it is because this time the autoscaler knows that it can not fit two pods in a single node (5Gi + 5Gi > 8GB), even if it still ignoring the affinity rules.

```
I0503 14:31:36.574678       1 scale_up.go:378] Best option to resize: k8s-genmlow-24772259-vmss
I0503 14:31:36.574722       1 scale_up.go:382] Estimated 10 nodes needed in k8s-genmlow-24772259-vmss
I0503 14:31:36.574752       1 scale_up.go:461] Final scale-up plan: [{k8s-genmlow-24772259-vmss 0->10 (max: 1000)}]
I0503 14:31:36.574786       1 scale_up.go:531] Scale-up: setting group k8s-genmlow-24772259-vmss size to 10
```

It looks like a bug to me. Using the same setup in AWS (cluster autoscaler 1.2.x instead of 1.3.x is the only difference) works fine, and the CA creates the 10 virtual machines no matter whether you specify the container memory requests or not.

### LINKED ISSUE #2044: Azure CA doesn't scale multiple agent pools in parallel

feiskyer:
This PR allows scaling multiple Azure vmss synchronously by delaying the vmss capacity updates in different goroutines.

To make it work, the upcoming nodes from `getUpcomingNodeInfos()` are also changed to different names, so that multiple nodes won't be merged as one node in `filterOutSchedulableByPacking()`.

Partially fix #2044 (similar node issues are tracked at #2094)
Fix #1984

/cc @andyzhangx @nilo19

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
@@ -19,6 +19,7 @@ package azure
 import (
 	"fmt"
 	"math/rand"
+	"net/http"
 	"strings"
 	"sync"
 	"time"
@@ -35,6 +36,10 @@ import (
 	schedulernodeinfo "k8s.io/kubernetes/pkg/scheduler/nodeinfo"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
+)
+
+var (
+	vmssSizeRefreshPeriod = 15 * time.Second
 )
 
 // ScaleSet implements NodeGroup interface.
@@ -114,7 +119,7 @@ func (scaleSet *ScaleSet) getCurSize() (int64, error) {
 	scaleSet.mutex.Lock()
 	defer scaleSet.mutex.Unlock()
 
-	if scaleSet.lastRefresh.Add(15 * time.Second).After(time.Now()) {
+	if scaleSet.lastRefresh.Add(vmssSizeRefreshPeriod).After(time.Now()) {
 		return scaleSet.curSize, nil
 	}
 
@@ -135,34 +140,56 @@ func (scaleSet *ScaleSet) GetScaleSetSize() (int64, error) {
 	return scaleSet.getCurSize()
 }
 
-// SetScaleSetSize sets ScaleSet size.
-func (scaleSet *ScaleSet) SetScaleSetSize(size int64) error {
-	scaleSet.mutex.Lock()
-	defer scaleSet.mutex.Unlock()
+// updateVMSSCapacity invokes virtualMachineScaleSetsClient to update the capacity for VMSS.
+func (scaleSet *ScaleSet) updateVMSSCapacity(size int64) {
+	var op compute.VirtualMachineScaleSet
+	var resp *http.Response
+	var isSuccess bool
+	var err error
+
+	defer func() {
+		if err != nil {
+			klog.Errorf("Failed to update the capacity for vmss %s with error %v, invalidate the cache so as to get the real size from API", scaleSet.Name, err)
+			// Invalidate the VMSS size cache in order to fetch the size from the API.
+			scaleSet.mutex.Lock()
+			defer scaleSet.mutex.Unlock()
+			scaleSet.lastRefresh = time.Now().Add(-1 * vmssSizeRefreshPeriod)
+		}
+	}()
 
 	resourceGroup := scaleSet.manager.config.ResourceGroup
-	op, err := scaleSet.getVMSSInfo()
+	op, err = scaleSet.getVMSSInfo()
 	if err != nil {
-		return err
+		klog.Errorf("Failed to get information for VMSS (%q): %v", scaleSet.Name, err)
+		return
 	}
 
 	op.Sku.Capacity = &size
 	op.Identity = nil
 	op.VirtualMachineScaleSetProperties.ProvisioningState = nil
-	updateCtx, updateCancel := getContextWithCancel()
-	defer updateCancel()
+	ctx, cancel := getContextWithCancel()
+	defer cancel()
 	klog.V(3).Infof("Waiting for virtualMachineScaleSetsClient.CreateOrUpdate(%s)", scaleSet.Name)
-	resp, err := scaleSet.manager.azClient.virtualMachineScaleSetsClient.CreateOrUpdate(updateCtx, resourceGroup, scaleSet.Name, op)
-	isSuccess, realError := isSuccessHTTPResponse(resp, err)
+	resp, err = scaleSet.manager.azClient.virtualMachineScaleSetsClient.CreateOrUpdate(ctx, resourceGroup, scaleSet.Name, op)
+	isSuccess, err = isSuccessHTTPResponse(resp, err)
 	if isSuccess {
 		klog.V(3).Infof("virtualMachineScaleSetsClient.CreateOrUpdate(%s) success", scaleSet.Name)
-		scaleSet.curSize = size
-		scaleSet.lastRefresh = time.Now()
-		return nil
+		return
 	}
 
-	klog.Errorf("virtualMachineScaleSetsClient.CreateOrUpdate for scale set %q failed: %v", scaleSet.Name, realError)
-	return realError
+	klog.Errorf("virtualMachineScaleSetsClient.CreateOrUpdate for scale set %q failed: %v", scaleSet.Name, err)
+	return
+}
+
+// SetScaleSetSize sets ScaleSet size.
+func (scaleSet *ScaleSet) SetScaleSetSize(size int64) {
+	scaleSet.mutex.Lock()
+	defer scaleSet.mutex.Unlock()
+
+	// Proactively set the VMSS size so autoscaler makes better decisions.
+	scaleSet.curSize = size
+	scaleSet.lastRefresh = time.Now()
+	go scaleSet.updateVMSSCapacity(size)
 }
 
 // TargetSize returns the current TARGET size of the node group. It is possible that the
@@ -187,7 +214,8 @@ func (scaleSet *ScaleSet) IncreaseSize(delta int) error {
 		return fmt.Errorf("size increase too large - desired:%d max:%d", int(size)+delta, scaleSet.MaxSize())
 	}
 
-	return scaleSet.SetScaleSetSize(size + int64(delta))
+	scaleSet.SetScaleSetSize(size + int64(delta))
+	return nil
 }
 
 // GetScaleSetVms returns list of nodes for the given scale set.
@@ -249,7 +277,8 @@ func (scaleSet *ScaleSet) DecreaseTargetSize(delta int) error {
 			size, delta, len(nodes))
 	}
 
-	return scaleSet.SetScaleSetSize(size + int64(delta))
+	scaleSet.SetScaleSetSize(size + int64(delta))
+	return nil
 }
 
 // Belongs returns true if the given node belongs to the NodeGroup.
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -22,6 +22,7 @@ import (
 
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
 	"k8s.io/autoscaler/cluster-autoscaler/clusterstate"
 	"k8s.io/autoscaler/cluster-autoscaler/clusterstate/utils"
@@ -591,6 +592,13 @@ func allPodsAreNew(pods []*apiv1.Pod, currentTime time.Time) bool {
 	return found && oldest.Add(unschedulablePodWithGpuTimeBuffer).After(currentTime)
 }
 
+func buildNodeForNodeTemplate(nodeTemplate *schedulernodeinfo.NodeInfo, index int) *apiv1.Node {
+	node := nodeTemplate.Node().DeepCopy()
+	node.Name = fmt.Sprintf("%s-%d", node.Name, index)
+	node.UID = uuid.NewUUID()
+	return node
+}
+
 func getUpcomingNodeInfos(registry *clusterstate.ClusterStateRegistry, nodeInfos map[string]*schedulernodeinfo.NodeInfo) []*apiv1.Node {
 	upcomingNodes := make([]*apiv1.Node, 0)
 	for nodeGroup, numberOfNodes := range registry.GetUpcomingNodes() {
@@ -600,7 +608,8 @@ func getUpcomingNodeInfos(registry *clusterstate.ClusterStateRegistry, nodeInfos
 			continue
 		}
 		for i := 0; i < numberOfNodes; i++ {
-			upcomingNodes = append(upcomingNodes, nodeTemplate.Node())
+			// Ensure new nodes having different names because nodeName would used as a map key.
+			upcomingNodes = append(upcomingNodes, buildNodeForNodeTemplate(nodeTemplate, i))
 		}
 	}
 	return upcomingNodes


cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -135,34 +136,56 @@ func (scaleSet *ScaleSet) GetScaleSetSize() (int64, error) {
...
 
-	klog.Errorf("virtualMachineScaleSetsClient.CreateOrUpdate for scale set %q failed: %v", scaleSet.Name, realError)
-	return realError
+	klog.Errorf("virtualMachineScaleSetsClient.CreateOrUpdate for scale set %q failed: %v", scaleSet.Name, err)
+	return
+}
+
+// SetScaleSetSize sets ScaleSet size.
+func (scaleSet *ScaleSet) SetScaleSetSize(size int64) error {
+	scaleSet.mutex.Lock()
```
andyzhangx: there is also `scaleSet.mutex.Lock()` inside `updateVMSSCapacity` func, will there be potential issue?

feiskyer: it's ok, updateVMSSCapacity() is running as goroutine.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -135,34 +136,56 @@ func (scaleSet *ScaleSet) GetScaleSetSize() (int64, error) {
...
-		scaleSet.curSize = size
-		scaleSet.lastRefresh = time.Now()
-		return nil
+		return
 	}
 
-	klog.Errorf("virtualMachineScaleSetsClient.CreateOrUpdate for scale set %q failed: %v", scaleSet.Name, realError)
-	return realError
+	klog.Errorf("virtualMachineScaleSetsClient.CreateOrUpdate for scale set %q failed: %v", scaleSet.Name, err)
+	return
```
CecileRobertMichon: why is this not returning an error anymore? If the `CreateOrUpdate` fails we will print the error but not return it, that doesn't seem right.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -135,34 +136,56 @@ func (scaleSet *ScaleSet) GetScaleSetSize() (int64, error) {
...
+// SetScaleSetSize sets ScaleSet size.
+func (scaleSet *ScaleSet) SetScaleSetSize(size int64) error {
+	scaleSet.mutex.Lock()
+	defer scaleSet.mutex.Unlock()
+
+	// Proactively set the VMSS size so autoscaler makes better decisions.
+	scaleSet.curSize = size
+	scaleSet.lastRefresh = time.Now()
+	go scaleSet.updateVMSSCapacity(size)
+	return nil
```
CecileRobertMichon: this function never returns an error, it shouldn't have an error return type

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -600,7 +601,10 @@ func getUpcomingNodeInfos(registry *clusterstate.ClusterStateRegistry, nodeInfos
 			continue
 		}
 		for i := 0; i < numberOfNodes; i++ {
-			upcomingNodes = append(upcomingNodes, nodeTemplate.Node())
+			// Ensure new nodes having different names because nodeName would used as a map key.
+			node := nodeTemplate.Node().DeepCopy()
+			node.Name = fmt.Sprintf("%s-%d", node.Name, rand.Int63())
```
CecileRobertMichon: will this affect non-Azure nodes as well?

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -135,34 +136,56 @@ func (scaleSet *ScaleSet) GetScaleSetSize() (int64, error) {
...
+	var isSuccess bool
+	var err error
+
+	defer func() {
+		if err != nil {
+			klog.Errorf("Failed to update the capacity for vmss %s with error %v, invalidate the cache so as to get the real size from API", scaleSet.Name, err)
+			// Invalidate the vmss size cache, so that it would be got by API.
+			scaleSet.mutex.Lock()
+			defer scaleSet.mutex.Unlock()
+			scaleSet.lastRefresh = time.Now().Add(-1 * 15 * time.Second)
```
CecileRobertMichon: can you explain why we are setting last refresh to -15 seconds from Now()? Also, why not use Sub()? https://golang.org/pkg/time/#Time.Sub

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -135,34 +136,56 @@ func (scaleSet *ScaleSet) GetScaleSetSize() (int64, error) {
...
+func (scaleSet *ScaleSet) updateVMSSCapacity(size int64) {
+	var op compute.VirtualMachineScaleSet
+	var resp *http.Response
+	var isSuccess bool
+	var err error
+
+	defer func() {
+		if err != nil {
+			klog.Errorf("Failed to update the capacity for vmss %s with error %v, invalidate the cache so as to get the real size from API", scaleSet.Name, err)
+			// Invalidate the vmss size cache, so that it would be got by API.
```
CecileRobertMichon: I don't understand the second part of this comment "so that it would be got by API."

Would "Invalidate the VMSS size cache in order to fetch the size from the API." make more sense?

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
CecileRobertMichon: nevermind, this is a goroutine

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -135,34 +136,56 @@ func (scaleSet *ScaleSet) GetScaleSetSize() (int64, error) {
...
+			scaleSet.mutex.Lock()
+			defer scaleSet.mutex.Unlock()
+			scaleSet.lastRefresh = time.Now().Add(-1 * 15 * time.Second)
+		}
+	}()
 
 	resourceGroup := scaleSet.manager.config.ResourceGroup
-	op, err := scaleSet.getVMSSInfo()
+	op, err = scaleSet.getVMSSInfo()
 	if err != nil {
```
CecileRobertMichon: should we print an error log here?

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
feiskyer: Yep, this is a goroutine

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
feiskyer: good catch. Would update to never return err

cluster-autoscaler/core/static_autoscaler.go
feiskyer: Yep, it would affect non-Azure nodes, but I think it should be fixed for all cloud providers.

@losipiuk @mwielgus Could you help to take a look at this?

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
feiskyer: 15 second is from L122, I renamed it to vmssSizeRefreshPeriod in PR #2151, but forgot to change here. Let me change it also to `vmssSizeRefreshPeriod` for clear.

Sub is a different use case, it accepts a param with `time.Time`, not `time.Duration`.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
feiskyer: Yep, thanks for improvements.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
feiskyer: Good catch, would add a log here.

cluster-autoscaler/core/static_autoscaler.go
losipiuk: It feels like serious bug. Thanks for spotting and fixing that.
One note here. Could you please use counter (increased atomically using AddUint64) instead random value. 

cc: @vivekbagade. Vivek could you please scan the code and see if there are no more places where we use node name as dictionary key.

vivekbagade: We use node name as a dictionary key in a few places. Need to check if they could be causing any issues. Even if they are not, we probably should change this to avoid future issues.

vivekbagade: This is because of CreateNodeNameToInfoMap func that could potentially mask a few nodes.

feiskyer: Thanks for the suggestion, would update the PR. 

> cc: @vivekbagade. Vivek could you please scan the code and see if there are no more places where we use node name as dictionary key.
> We could also add a sanity check in filterOutSchedulableUsingPacking to detect situation when it gets list of nodes with repeating names.
> Also probably we should use UID (not name) as map key.

That looks good. We can do the check and optimization after the code scan.

feiskyer: >  Could you please use counter (increased atomically using AddUint64) instead random value.

Actually, the index would be ok. The node name here only used for this single filterOutSchedulableUsingPacking() step.

feiskyer: @losipiuk  Updated, PTAL

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -591,6 +592,13 @@ func allPodsAreNew(pods []*apiv1.Pod, currentTime time.Time) bool {
 	return found && oldest.Add(unschedulablePodWithGpuTimeBuffer).After(currentTime)
 }
 
+func buildNodeForNodeTemplate(nodeTemplate *schedulernodeinfo.NodeInfo, index int) *apiv1.Node {
```
losipiuk: Looks good. 
@vivekbagade is doing some testing on this (thanks!).
I will LGTM when we are done with that.

vivekbagade:
@losipiuk My testing is done. Works as expected. LGTM from my side.

losipiuk:
Thanks!

feiskyer:
@losipiuk  Thanks

PR #2158, opened 2019-07-03
Use Metrics Server instead of Heapster in the FAQ

rata:
Heapster is deprecated and retired now:

https://github.com/kubernetes-retired/heapster/

The Metrics Server should be used now, so let's mention that on the FAQ
to avoid confusions.


rata:
Thank you very much! :)

PR #2160, opened 2019-07-03
Add resource limit type to NotTriggerScaleUp event

aleksandra-malinowska:
Resource limit message was indistinguishable from resource limits message. This change adds the type of resource on which limit would be exceeded.

Had to refactor scale up tests a bit in order to add a unit test without duplicating the entire test setup.

### DIFF
--- a/cluster-autoscaler/core/scale_up.go
+++ b/cluster-autoscaler/core/scale_up.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"fmt"
 	"math"
+	"strings"
 	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/core/scale_up.go
```diff
@@ -238,10 +239,14 @@ func (sr *skippedReasons) Reasons() []string {
 
 var (
 	backoffReason         = &skippedReasons{[]string{"in backoff after failed scale-up"}}
-	maxLimitReachedReason = &skippedReasons{[]string{"max limit reached"}}
+	maxLimitReachedReason = &skippedReasons{[]string{"max size reached"}}
```
MaciekPytel: maybe "max node group size reached"? I don't think it's obvious.

cluster-autoscaler/core/scale_up.go
```diff
@@ -238,10 +239,14 @@ func (sr *skippedReasons) Reasons() []string {
...
 	notReadyReason        = &skippedReasons{[]string{"not ready for scale-up"}}
 )
 
+func maxResourceLimitReached(resources []string) *skippedReasons {
+	return &skippedReasons{[]string{fmt.Sprintf("max %s limit reached", strings.Join(resources, ", "))}}
```
MaciekPytel: "cluster %s limit reached"?

cluster-autoscaler/core/scale_up.go
aleksandra-malinowska: Done

cluster-autoscaler/core/scale_up.go
aleksandra-malinowska: Done

MaciekPytel:
/lgtm
/approve
/hold
@aleksandra-malinowska feel free to cancel hold once you're confident the new messages are more understandable than current ones.

Issue #2162, opened 2019-07-04
[VPA] can't set arguments only on recommender container


cristifalcas:
It will be helpful to have the possibility to add arguments to the container.

Currently, in the recommender deployment you can't add just the arguments, as they will be ignored. You have to also add the command overwrite, that it's not so intuitive to find out (./recommender).

By adding a CMD line, this can be solved as in https://stackoverflow.com/a/34245657/451348

Affected file:

https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/recommender/Dockerfile

Denton24646:
I can open a PR for this 👍

cristifalcas:
thank you

PR #2166, opened 2019-07-07
tweak: add CMD options to recommender Dockerfile Entrypoint

Denton24646:
PR addresses a recent issue brought up in #2162 

This change will still keep the default behavior of the recommender container the same - if the user does not specify anything the entrypoint and default commands will be executed as before. If the user does specify additional commands to the recommender container when running it those arguments will override the existing defaults.


bskiba:
Thank you! 
If you have time for another contribution, updater and admission controller would also benefit from this kind of change.

/lgtm
/approve

fixes  #2162

Denton24646:
@bskiba sure, no problem

PR #2173, opened 2019-07-08
Fix race conditions when checking the VMSS size

feiskyer:
Fix race conditions when checking the VMSS size.


losipiuk:
/lgtm
/approve
Thanks!

PR #2175, opened 2019-07-09
tweak: add CMD to updater and admission-controller Dockerfile Entrypoint

Denton24646:
PR updates VPA updater and admission-controller Dockerfiles to use CMD
Conversation tracked in previously merged PR (which updated the Dockerfile for the recommender). 
https://github.com/kubernetes/autoscaler/pull/2166#issuecomment-509225247

This change will still keep the default behavior of the updater and admission-controller containers the same - if the user does not specify anything the entrypoint and default commands will be executed as before. If the user does specify additional commands to the container when running it those arguments will override the existing defaults.

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/Dockerfile
+++ b/vertical-pod-autoscaler/pkg/admission-controller/Dockerfile
@@ -17,4 +17,5 @@ MAINTAINER Tomasz Kulczynski "tkulczynski@google.com"
 
 ADD admission-controller admission-controller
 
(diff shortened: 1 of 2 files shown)


bskiba:
Can you fix this along the lines of #2176? Otherwise the container won't start .

Denton24646:
> Can you fix this along the lines of #2176? Otherwise the container won't start .

sure, no problem

Denton24646:
ready @bskiba

PR #2186, opened 2019-07-11
Use ioutil for certificate file reading.

jackkleeman:
Currently we read certificate files into a buffer of 5000 bytes. This sets a max size for our certificates; above this max, we will get invalid certs, most likely, leading to runtime bad certificate errors.

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/certs.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/certs.go
@@ -17,7 +17,7 @@ limitations under the License.
 package main
 
 import (
-	"os"
+	"io/ioutil"
 
 	"k8s.io/klog"
 )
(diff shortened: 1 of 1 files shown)


vertical-pod-autoscaler/pkg/admission-controller/certs.go
```diff
@@ -31,19 +31,14 @@ type certsConfig struct {
 }
 
 func readFile(filePath string) []byte {
-	file, err := os.Open(filePath)
+	res, err := ioutil.ReadFile(filePath)
 	if err != nil {
 		klog.Error(err)
```
bskiba: Would you mind adding some more context to this error log? sth like ("Error reading certificate file: %v", err)

vertical-pod-autoscaler/pkg/admission-controller/certs.go
```diff
@@ -31,19 +31,14 @@ type certsConfig struct {
...
-	res := make([]byte, 5000)
-	count, err := file.Read(res)
-	if err != nil {
-		klog.Error(err)
-		return nil
-	}
-	klog.Infof("Successfully read %d bytes from %v", count, filePath)
-	return res[:count]
+
+	klog.Infof("Successfully read %d bytes from %v", len(res), filePath)
```
bskiba: klog.V(3)

jackkleeman:
Done @bskiba :)

bskiba:
Thanks!
/lgtm
/approve

PR #2187, opened 2019-07-14
add test case for controller_fetcher and cluster_feeder

Moriadry:
to raise code coverage of controller_fetcher and cluster_feeder.

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder_test.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder_test.go
@@ -120,6 +120,14 @@ func TestLoadPods(t *testing.T) {
 			expectedConfigDeprecated:  nil,
 		},
 		{
+			name:                      "no targetRef",
+			selector:                  parseLabelSelector("app = test"),
+			fetchSelectorError:        nil,
+			expectedSelector:          labels.Nothing(),
+			expectedConfigUnsupported: nil,
+			expectedConfigDeprecated:  nil,
+		},
+		{
 			name:               "can't decide if top-level-ref",
 			selector:           nil,
 			fetchSelectorError: nil,
--- a/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
@@ -24,6 +24,8 @@ import (
 	"github.com/stretchr/testify/assert"
 
 	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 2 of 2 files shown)


Moriadry:
@bskiba  can you help me check this? This is my very start of k8s, thx!

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder_test.go
```diff
@@ -119,6 +119,14 @@ func TestLoadPods(t *testing.T) {
 			expectedConfigUnsupported: nil,
 			expectedConfigDeprecated:  nil,
 		},
+		{
+			name:                      "no targetRef",
+			selector:                  parseLabelSelector("app = test"),
+			fetchSelectorError:        nil,
+			expectedSelector:          labels.Nothing(),
+			expectedConfigUnsupported: nil,
```
bskiba: I think it would be better if this was actually configUnsupported. Also VPA withouth targetRef is rejected by the admission controller so it shouldn't be possible to create an object like this.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
```diff
@@ -117,6 +117,37 @@ func TestControllerFetcher(t *testing.T) {
...
+				Name: "test-ss", Kind: "StatefulSet", Namespace: "test-namesapce"}},
+			objects: []runtime.Object{&appsv1.Deployment{
+				TypeMeta: metav1.TypeMeta{
+					Kind: "Deployment",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-deployment",
+					Namespace: "test-namesapce",
+				},
+			}, &appsv1.StatefulSet{
```
bskiba: I guess this is OK for test purposes, but it is kind of weird to see stateful set controlled by a deployment. Maybe it would make more sense to make it standalone, e.g. it has no parent?

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder_test.go
Moriadry: yes. But this will hit this line, and it's return condition is  default value. https://github.com/kubernetes/autoscaler/blob/de322e2034df4413381586d50d68620f5c222ed1/vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go#L442

I believe this case raise code coverage of cluster_feeder, and it really works.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
Moriadry: I wrote just as ReplicaSet case did. Okay, I will try to figure it out :)

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder_test.go
bskiba: Agreed, though I meant maybe we should actually change the code :) It doesn't have to be part of this PR though.

Moriadry: Yeah, maybe I can help with this afterwards~

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
```diff
@@ -117,6 +119,70 @@ func TestControllerFetcher(t *testing.T) {
...
+				TypeMeta: metav1.TypeMeta{
+					Kind: "StatefulSet",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-statefulset",
+					Namespace: "test-namesapce",
+				},
+			}},
+			expectedKey: &ControllerKeyWithAPIVersion{ControllerKey: ControllerKey{
+				Name: "test-statefulset", Kind: "StatefulSet", Namespace: "test-namesapce"}}, // Deployment has no parrent
```
bskiba: s/Deployment/Statefulset/
s/parrent/parent

Can you also update the other comments?

Moriadry: okay

bskiba:
Looks good! Can you please squash to one commit and we're good to submit.

Moriadry:
@bskiba done.

PR #2191, opened 2019-07-16
xrange() was removed in Python 3 in favor of range()

cclauss:
This PR ensures equivalent functionality in both Python 2 and Python 3.


losipiuk:
/lgtm
/approve
Thanks!

PR #2193, opened 2019-07-17
Update fetcher.go

PualrDwade:
Use reflection to reduce duplication of code


bskiba:
Hi @PualrDwade, you need to sign the CLA before we can accept your pull reqests.

PR #2195, opened 2019-07-17
Add support for cronjobs in VPA

### LINKED ISSUE #1790: VPA for CronJob

bskiba:
Fixes #1790

### DIFF
--- a/vertical-pod-autoscaler/deploy/vpa-rbac.yaml
+++ b/vertical-pod-autoscaler/deploy/vpa-rbac.yaml
@@ -178,6 +178,7 @@ rules:
       - batch
     resources:
       - jobs
+      - cronjobs
     verbs:
       - get
       - list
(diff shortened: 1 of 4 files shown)


bskiba:
@schylek

schylek:
In general /lgtm. Do you think Cron Jobs are worth adding e2e test?

bskiba:
I was thinking about adding e2e. Currently we only have almost all tests with VPA on Deployment, the only ones where we exercise different controllers is updater (Evicts pods in ...). I came to a conclusion that it makes little sense to evict pods from a CronJob.
However it would be nice to have a "VPA provides recommendations" scenario for all the known controllers.

bskiba:
Yes, once this releases, if you set VPA on CronJob and updateMode: Initial, each new launched Job will get request based on recommendations calculated from previous jobs usage.

Note, this will not work well with short lived jobs (<1minute).

johanneswuerbach:
@bskiba is there any timeline when this is going to be released?

PR #2206, opened 2019-07-19
Update CA version to 1.12.7

losipiuk:


aleksandra-malinowska:
Kubernetes 1.12 is EOL, is there any emergency fix/other reason we need another release?

losipiuk:
> Kubernetes 1.12 is EOL, is there any emergency fix/other reason we need another release

There are still people maintaining their own clusters which may find this update useful. 
The fact that 1.12 is EOL implies we will not update the cluster-autoscaler.manifest file in k8s on 1.12 release branch.

Issue #496, opened 2017-12-06
Support whitelisting labels to ignore when balancing similar node groups


negz:
Running the Cluster Autoscaler with `--balance-similar-node-groups` attempts to keep the size of identical node groups balanced when scaling up.

https://github.com/kubernetes/autoscaler/blob/d43029c/cluster-autoscaler/utils/nodegroupset/compare_nodegroups.go#L57

Per `IsNodeInfoSimilar` node groups are considered similar when a sample node from each node group has approximately the same resource capacity and labels. A fixed set of labels are ignored during this comparison - specifically hostname, failure domain, and region.

At Planet Labs we label nodes with their node group's unique identifier, which we call their pool. This label looks like `kubernetes.planet.com/pool_id=worker-7kd9`. Our node groups are typically identical and thus we would like to balance them, but they are considered dissimilar due to the existence of the `pool_id` label.

We'd like to be able to configure a set of labels for `IsNodeInfoSimilar` to ignore in addition to the hostname, failure domain, and region. This would involve passing the CA a flag like `--ignore-labels-when-comparing-node-groups=kubernetes.planet.com/pool_id,someotherlabel`. This flag would be used to set `IgnoreLabelsWhenComparingNodeGroups` in the autoscaling context, and that setting would be passed down to `IsNodeInfoSimilar` via `FindSimilarNodeGroups`.

Does this sound like a reasonable approach? If so we're happy to send a PR to implement it.

MaciekPytel:
I'm curious what is your use-case for having those labels? I'm generally not opposed to adding something like that, just not sure how useful it would be. `--balance-similar-node-groups` only works if your nodes are similar and the same subset of pending pods can go to each "similar" node pool. So it won't necessarily work well if you'll try to use this custom label for things like nodeSelector or nodeAffinity.

negz:
To be honest it's mostly historical and worth thinking through. We don't use the labels for node selection or affinity - I'm fairly sure they only exist so that humans can easily tie a node back to the pool it belongs to.

negz:
Upon reflection we decided the use case for the label isn't worth adding this functionality to the CA - we'll just remove the label instead. Thanks for the perspective!

PR #2207, opened 2019-07-22
add kops instance group label to ignore list for similar node group identification.

jhohertz:
For kops-built clusters, an instance label added by that tooling will prevent the balance-similar-nodegroups feature from working as intended.

This patch simply adds the kops instance group label to the ignored list.

It looks like this should apply cleanly as far back as the release 1.13 branch, and I have a PR for the 1.12 branch ready to make as well since it's still in common use. (And in fact the most recent version released for kops)

Signed-off-by: Joe Hohertz <joe@viafoura.com>

### DIFF
--- a/cluster-autoscaler/processors/nodegroupset/compare_nodegroups.go
+++ b/cluster-autoscaler/processors/nodegroupset/compare_nodegroups.go
@@ -41,7 +41,8 @@ var BasicIgnoredLabels = map[string]bool{
 	apiv1.LabelZoneFailureDomain:          true,
 	apiv1.LabelZoneRegion:                 true,
 	"beta.kubernetes.io/fluentd-ds-ready": true, // this is internal label used for determining if fluentd should be installed as deamon set. Used for migration 1.8 to 1.9.
-	"kops.k8s.io/instancegroup":           true, // this is a label used by kops to identify instance group names. it's value is variable, defeating check of similar node groups
+	"kops.k8s.io/instancegroup":           true, // this is a label used by kops to identify "instance group" names. it's value is variable, defeating check of similar node groups
+	"alpha.eksctl.io/nodegroup-name":      true, // this is a label used by eksctl to identify "node group" names, similar in spirit to the kops label above
 }
 
 // NodeInfoComparator is a function that tells if two nodes are from NodeGroups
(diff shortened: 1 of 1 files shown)


jhohertz:
rebased

MaciekPytel:
To clarify - is the value of this label always unique for each nodegroup? ie. if I have 3 ASGs in different zones will they always get a different value for this flag? Or can you have nodegroups in different zones with the same instancegroup id?
If it's the former I'm happy to lgtm this. 
If it's the latter you may actually want to balance only if id is identical - this is the case for GKE nodepools (multiple MIGs can belong to a single nodepool) and (I think) Azure agentpool.

jhohertz:
Correct. Each "instance group" in kops terminology, will have it's own unique "kops.k8s.io/instancegroup" labelling. (And each "instance group" relates to an autoscaling group). Here's a collection from a live cluster:

```
$ kubectl describe nodes | grep kops.k8s.io/instancegroup  | sort | uniq
                    kops.k8s.io/instancegroup=master-us-east-1a
                    kops.k8s.io/instancegroup=master-us-east-1b
                    kops.k8s.io/instancegroup=master-us-east-1d
                    kops.k8s.io/instancegroup=master-us-east-1e
                    kops.k8s.io/instancegroup=master-us-east-1f
                    kops.k8s.io/instancegroup=nodes-us-east-1a
                    kops.k8s.io/instancegroup=nodes-us-east-1b
                    kops.k8s.io/instancegroup=nodes-us-east-1d
                    kops.k8s.io/instancegroup=nodes-us-east-1e
                    kops.k8s.io/instancegroup=nodes-us-east-1f
```

chrisvanderpennen:
Hey, I found this PR tracking down why balance-similar-node-groups wasn't working with an eksctl-created cluster.  eksctl adds the `alpha.eksctl.io/nodegroup-name=xxx` label to each node in an autoscaling group, and each availability zone needs a dedicated autoscaling group so cluster-autoscaler can guarantee nodes will be able to access EBS volumes, which can only be mounted by VMs in the same availability zone.

May I propose exposing configuration to allow the label whitelist to be extended as requested in #496 instead of maintaining a list of tool-specific labels in the core of cluster-autoscaler?  Doing so would allow users to resolve this issue when using kops, eksctl, or any other future tool that introduces its own nodegroup annotations.

jhohertz:
That's one possibility, but would it hurt if we just added both kops and eksctl tags to the ignored list?

Would be happy to update the PR with the additional tag if the maintainers are cool with that.

MaciekPytel:
I don't mind adding both, if they're both always unique for nodegroup.

jhohertz:
Added the eksctl label variant to the PR

losipiuk:
@MaciekPytel are we good with this one?

selslack:
@losipiuk @aleksandra-malinowska @MaciekPytel What needs to be done to have this PR moving?

jhohertz:
Let me know if there is anything I can do / clarify for this PR.

losipiuk:
/lgtm
/approve
Thanks!

Sorry for the delay

PR #2212, opened 2019-07-24
Don't consider to move pods to empty nodes when scaling down

devkid:
We had noticed the following happening in our clusters:
* some node A becomes empty after previously running a workload that blocked some host ports
* some other node B that is running the same type of workload (blocking the same host ports) is now considered unneeded because the resource utilization is low and the pod could be evicted onto the now empty node
* apparently node A is only marked as unneeded _slightly_ after node B is marked as unneeded
* the 10 minute timer elapses first for node B and the node B is selected for scale down, the pod is evicted and another one is being started as a replacement on node A; afterwards node B is terminated and node A is now longer unneeded

From my point of view this behavior doesn't really make sense as it would have been better to just wait for the already empty node to be scaled-down instead of moving pods around to scale down another node instead.

This pull request changes this behavior to not consider empty nodes for targets of possible pod evictions. As far as I could tell, nodes are only considered as empty in the code if they are actually removable (e.g. minimum size in the node group is not yet reached), therefor empty nodes should _always_ be scaled down before scaling down any other node.

### DIFF
--- a/cluster-autoscaler/core/scale_down.go
+++ b/cluster-autoscaler/core/scale_down.go
@@ -489,7 +489,7 @@ func (sd *ScaleDown) UpdateUnneededNodes(
 	// Look for nodes to remove in the current candidates
 	nodesToRemove, unremovable, newHints, simulatorErr := simulator.FindNodesToRemove(
 		currentCandidates, destinationNodes, nonExpendablePods, nil, sd.context.PredicateChecker,
-		len(currentCandidates), true, sd.podLocationHints, sd.usageTracker, timestamp, pdbs, emptyNodesList)
+		len(currentCandidates), true, sd.podLocationHints, sd.usageTracker, timestamp, pdbs, emptyNodes)
 	if simulatorErr != nil {
 		return sd.markSimulationError(simulatorErr, timestamp)
 	}
@@ -512,7 +512,7 @@ func (sd *ScaleDown) UpdateUnneededNodes(
 		additionalNodesToRemove, additionalUnremovable, additionalNewHints, simulatorErr :=
 			simulator.FindNodesToRemove(currentNonCandidates[:additionalCandidatesPoolSize], destinationNodes, nonExpendablePods, nil,
 				sd.context.PredicateChecker, additionalCandidatesCount, true,
-				sd.podLocationHints, sd.usageTracker, timestamp, pdbs, emptyNodesList)
+				sd.podLocationHints, sd.usageTracker, timestamp, pdbs, emptyNodes)
 		if simulatorErr != nil {
 			return sd.markSimulationError(simulatorErr, timestamp)
 		}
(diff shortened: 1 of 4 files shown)


losipiuk:
Thanks - this looks reasonable to me. @devkid can you resolve the conflicts.

aleksandra-malinowska:
> As far as I could tell, nodes are only considered as empty in the code if they are actually removable (e.g. minimum size in the node group is not yet reached), therefor empty nodes should always be scaled down before scaling down any other node.

True, but it's more of an implementation detail right now. Since after this change it would be a dependency, can you please add a test case for such scenario? E.g. 1 empty node in group at min size + 1 drainable node in a group above its min size. It will help avoid regressions if the code is refactored.

devkid:
> True, but it's more of an implementation detail right now. Since after this change it would be a dependency, can you please add a test case for such scenario? E.g. 1 empty node in group at min size + 1 drainable node in a group above its min size. It will help avoid regressions if the code is refactored.
Good catch, will do.

devkid:
Conflicts resolved and test case added. @aleksandra-malinowska Can you check if this is what you wanted to have?

cluster-autoscaler/core/scale_down.go
```diff
@@ -518,7 +518,7 @@ func (sd *ScaleDown) UpdateUnneededNodes(
 		additionalNodesToRemove, additionalUnremovable, additionalNewHints, simulatorErr :=
 			simulator.FindNodesToRemove(currentNonCandidates[:additionalCandidatesPoolSize], destinationNodes, nonExpendablePods, nil,
 				sd.context.PredicateChecker, additionalCandidatesCount, true,
-				sd.podLocationHints, sd.usageTracker, timestamp, pdbs)
+				sd.podLocationHints, sd.usageTracker, timestamp, pdbs, emptyNodesList)
```
aleksandra-malinowska: Why pass `emptyNodesList` and not `emptyNodes`? Map would be more convenient to use (no need to loop over the list searching for each node)

devkid: The reason was that `simulator.FindNodesToRemove` is also invoked in `TryToScaleDown` where there is only `emptyNodes`, which is a `[]*apiv1.Node`, not a map. Should I maybe introduce another helper function to generate this map out of the list and also invoke it in `TryToScaleDown` to get the map?

aleksandra-malinowska: Hey, sorry, missed the emails from github :/ I'd say yes, let's put it into map. It's probably irrelevant performance-wise for all but the largest clusters, but it seems cleaner.

devkid: Changed & rebased the branch.

devkid:
Can we get this merged somehow?

cluster-autoscaler/core/scale_down_test.go
```diff
@@ -176,6 +176,63 @@ func TestFindUnneededNodes(t *testing.T) {
...
+
+	// node group with empty node, but min size reached
+	provider.AddNodeGroup("ng1", 1, 10, 1)
+	n1 := BuildTestNode("n1", 1000, 10)
+	SetNodeReadyState(n1, true, time.Time{})
+	provider.AddNode("ng1", n1)
+
+	// node group with one underutilized node and node min size not yet reached
+	provider.AddNodeGroup("ng2", 1, 10, 2)
+	n2 := BuildTestNode("n1", 1000, 10)
```
aleksandra-malinowska: Shouldn't this node be called "n2"? Right now there are 2 nodes named "n1" and p1 pod is assigned to a non-existent "n2".

aleksandra-malinowska: (Test no longer passes when this is changed, BTW)

PR #2217, opened 2019-07-26
cluster-autoscaler FAQ: clarify what "deleting nodes" means in this context

matthiasr:
I was trying to debug why, in our (custom) clusters, nodes continue to exist (with status NotReady) when the cluster-autoscaler _deletes_ them.

"delete node" has a specific meaning in the context of the Kubernetes
API: deleting the node object.

In the end, I found that it is not actually expected to do that: kube-controller-manager (or cloud-controller-manager) is. The repeated use of "delete(ing/s) node" in the FAQ obscured this.

I am proposing to

* change the FAQ to say "terminate"
  * it may make sense to change the flags later as well?
* add an FAQ entry that makes this relationship clear


MaciekPytel:
/lgtm
/approve

Thanks for contributing this!

PR #2219, opened 2019-07-28
Fix bug in balancing processor. 

krzysztof-jastrzebski:
Cluster Autoscaler was stopping scaling up when there was a multizonal node pool pool with number of nodes exceeding limit for one zone. For example if we have multizonal node pool with 2 zones with max set to 3. Node pools have sizes of 2 and 4 then Cluster Autoscaler gets stuck as it tries to scale only this node pool and balancing processor prevents it as total size is already 6.

### DIFF
--- a/cluster-autoscaler/processors/nodegroupset/balancing_processor.go
+++ b/cluster-autoscaler/processors/nodegroupset/balancing_processor.go
@@ -103,7 +103,9 @@ func (b *BalancingNodeGroupSetProcessor) BalanceScaleUpBetweenGroups(context *co
 			NewSize:     currentSize,
 			MaxSize:     maxSize}
 		scaleUpInfos = append(scaleUpInfos, info)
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/processors/nodegroupset/balancing_processor.go
```diff
@@ -103,7 +103,9 @@ func (b *BalancingNodeGroupSetProcessor) BalanceScaleUpBetweenGroups(context *co
 			NewSize:     currentSize,
 			MaxSize:     maxSize}
 		scaleUpInfos = append(scaleUpInfos, info)
-		totalCapacity += maxSize - currentSize
+		if maxSize-currentSize > 0 {
```
aleksandra-malinowska: nit: maybe we can also change the check above (in line 96) to "currentSize >= maxSize"?

PR #2222, opened 2019-07-29
when only have on as group, not to backoff

yan234280533:
when only have on as group, not to backoff

### DIFF
--- a/cluster-autoscaler/core/scale_up.go
+++ b/cluster-autoscaler/core/scale_up.go
@@ -329,11 +329,17 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 			if !clusterStateRegistry.IsNodeGroupHealthy(nodeGroup.Id()) {
 				klog.Warningf("Node group %s is not ready for scaleup - unhealthy", nodeGroup.Id())
 				skippedNodeGroups[nodeGroup.Id()] = notReadyReason
+				continue
 			} else {
-				klog.Warningf("Node group %s is not ready for scaleup - backoff", nodeGroup.Id())
-				skippedNodeGroups[nodeGroup.Id()] = backoffReason
+				if len(nodeGroups) > 1 {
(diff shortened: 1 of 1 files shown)


MaciekPytel:
I disagree with this change. The current behavior is intentional, with backoff serving 2 purposes:
1. Make CA pick a different nodegroup if the first one can't be scaled-up.
2. Avoid pounding the cloudprovider API / using up API call quota when no scale-up is possible at all (for example because of hitting cloudprovider side quota). 

This PR removes point 2. 
Please re-open if you disagree.

yan234280533:
@MaciekPytel  I re-open the PR, for those reasons:
1. Make CA pick a different nodegroup if the first one can't be scaled-up.
You mistake me, this PR not to  make a difference of the first nodegroup and other  nodegroups. It only work when the CA only have one nodegroup. It make the nodegroup to as soon as possible to try to scale-up.

2.Avoid pounding the cloudprovider API 
In generally, we set CA loop 10s, so if we call the scale-up api in every loop , it only 6QPS/min, it absolutely not to reach the API quota. 
In fact, the frequency call cloudprovider API when not backoff one nodegroup, it not greater than there it have an cluster which scalp-up and scale-down very frequently， every loop it have an scale-up or scale-down action.

MaciekPytel:
> It make the nodegroup to as soon as possible to try to scale-up.

As I said the fact that it does not re-try immediately is by design. The assumption is that if a scale-up failed it is because of some cloudprovider-side issue and retrying immediately will run into the same issue.
Can you describe exactly what is your use-case is?

yan234280533:
@MaciekPytel  In my case, our user configuration only one node asg. They hope to quickly retry when the asg scale-up failed.Sometimes, when specific instance type sold out, we will throw some machines for this user, the user hope to scale-up when we throw machines as quickly as possible.

mwielgus:
Closing due to inactivity.

PR #2224, opened 2019-07-30
Fix v1beta1 VPA limit range e2e tests

jbartosik:
I forgot to update them when fixing v1beta2 tests


jbartosik:
/assign @bskiba 

This mostly copy-pasted from v1beta2 tests (only modification is for using label selector instead of target ref).

PR #2227, opened 2019-08-01
cloudprovider: add DigitalOcean

### LINKED ISSUE #254: Implement DigitalOcean cloud provider
I'm not exactly sure how to implement this, but I think the easiest way would be creating droplet from a snapshot already configured to join the existing cluster.

fatih:
This adds a new cluster autoscaler for [DigitalOcean Kubernetes ](https://www.digitalocean.com/products/kubernetes/) offering. 

* Because there is no native NodeGroup offering in DigitalOcean this autoscaler can be used only within a Managed DigitalOcean Kubernetes cluster. 
* Customers will be able to enable/disable it for existing clusters. The functionality is not added yet and will be implemented separately. 
* As suggested, I didn't vendored the client (`cloudprovider/digitalocean/godo`) and added it as a sub package. 

cc @timoreimann @snormore

Closes #254


fatih:
I've splitted my changes into two commits:

1. The actual implementation of the cloud provider: https://github.com/kubernetes/autoscaler/pull/2227/commits/cd2cb0fed3015b36bbe85ae515e141557de124cd

2. Adding the `godo` package (DigitalOcean API client): https://github.com/kubernetes/autoscaler/pull/2227/commits/776bba5e1059da5c5b7a3937e0f09836b7bf515e

cluster-autoscaler/cloudprovider/cloud_provider.go
```diff
@@ -35,6 +35,8 @@ const (
 	AwsProviderName = "aws"
 	// BaiducloudProviderName gets the provider name of baiducloud
 	BaiducloudProviderName = "baiducloud"
+	// DigitalOceanProviderName gets the provider name of baiducloud
```
neshmi: Typo here, baiducloud should be digitalocean

tghartland:
Your client package will need to be ignored by the verify tools. See https://github.com/kubernetes/autoscaler/pull/1948/commits/c2f2e9cd635a09ee616a13160bff6cb0d8e8a743.

cluster-autoscaler/cloudprovider/cloud_provider.go
fatih: Thanks, fixed it 👍

fatih:
> Your client package will need to be ignored by the verify tools. See c2f2e9c.

Thanks! I've added it now with https://github.com/kubernetes/autoscaler/pull/2227/commits/40b28f0133ced4d8bbcee5340af53b9166095cf0 Albeit, I think all this could be avoided if I could use the `vendor/` folder :/

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_node_group.go
```diff
@@ -0,0 +1,269 @@
...
+// It is assumed that cloud provider will not delete the existing nodes when there
+// is an option to just decrease the target. Implementation required.
+func (n *NodeGroup) DecreaseTargetSize(delta int) error {
+	if delta >= 0 {
+		return fmt.Errorf("delta must be negative, have: %d", delta)
+	}
+
+	targetSize := n.nodePool.Count + delta
+	if targetSize <= n.MinSize() {
+		return fmt.Errorf("size decrease is too small. current: %d desired: %d max: %d",
```
andrewsykim: s/max/min

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_node_group.go
```diff
@@ -0,0 +1,269 @@
...
+// kube-proxy). Implementation optional.
+func (n *NodeGroup) TemplateNodeInfo() (*schedulernodeinfo.NodeInfo, error) {
+	return nil, cloudprovider.ErrNotImplemented
+}
+
+// Exist checks if the node group really exists on the cloud provider side.
+// Allows to tell the theoretical node group from the real one. Implementation
+// required.
+func (n *NodeGroup) Exist() bool {
+	return n.nodePool != nil
```
andrewsykim: Wondering if this should actually call the provider cause a node pool could be deleted between calls to `Refresh`?

andrewsykim: Looks like the other providers don't do this so maybe not

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_cloud_provider.go
```diff
@@ -0,0 +1,193 @@
...
+		// that as nodeID.
+		nodeID = node.Spec.ProviderID
+	}
+
+	klog.V(5).Infof("checking nodegroup for node ID: %q", nodeID)
+
+	// NOTE(arslan): the number of node groups per cluster is usually very
+	// small. So even though this looks like quadratic runtime, it's OK to
+	// proceed with this.
+	klog.V(5).Infof("iterating over %d node groups", len(d.manager.nodeGroups))
```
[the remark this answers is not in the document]
fatih: Done

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_cloud_provider_test.go
```diff
@@ -0,0 +1,232 @@
...
+*/
+
+package digitalocean
+
+import (
+	"bytes"
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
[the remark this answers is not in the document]
fatih: Done

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_manager.go
```diff
@@ -0,0 +1,231 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"strconv"
+	"strings"
+
+	"golang.org/x/oauth2"
```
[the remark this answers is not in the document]
fatih: Done

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_node_group.go
fatih: Done

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_node_group.go
fatih: @andrewsykim that was the case for me when I wrote this initially. However upon using this and checking the code, seems like `Refresh()` is called every `--scan-interval` and then all the methods in this file are called based on when they are need or not.  The cluster-autoscaler basically reconciles every `--scan-interval` (default 10 seconds) therefore we should be fairly stable for race conditions. 

The other issue is, using `Refresh()` in these methods triggers the rate limit on the public API of DigitalOcean, because the cluster-autoscaler aggressively calls these during each reconcile period. Hence the most efficient and performant  way (based on my observation) is to use an internal cache that is populated (therefore invalidated) each reconcile period. Which is defined by the `--scan-interval` period.

andrewsykim: That seems reasonable to me

fatih:
I've added a way of dynamically configuring the node groups via tags associated with DO node pools. The changes are here: https://github.com/kubernetes/autoscaler/pull/2227/commits/44bfcce1d53187b8a671cf7dbfb434664bf91466

fatih:
Closing this in favor of: https://github.com/kubernetes/autoscaler/pull/2245. This PR was based on my personal repository. The new one is based on `digitalocean/autoscaler` which is better for me.

Issue #2139, opened 2019-06-25
Cluster Autoscaler AWS example uses wrong host path for SSL certs


edmorley:
Hi!

The latest version of Kubernetes available on EKS is 1.13. The Cluster Autoscaler docs recommend using a version of the autoscaler that matches that of the Kubernetes version in use, so EKS users will be using the examples from this branch:

https://github.com/kubernetes/autoscaler/tree/cluster-autoscaler-1.13.5/cluster-autoscaler/cloudprovider/aws/examples

However in those examples, the SSL certs path for the host is set to:

```yaml
      volumes:
        - name: ssl-certs
          hostPath:
            path: "/etc/ssl/certs/ca-certificates.crt"
```

...when it should be `/etc/ssl/certs/ca-bundle.crt`.

This results in:

```
Error: failed to start container "cluster-autoscaler":
Error response from daemon: OCI runtime create failed: container_linux.go:348: starting container process caused "process_linux.go:402:container init caused \"rootfs_linux.go:58: mounting \\\"/etc/ssl/certs/ca-certificates.crt\\\" to rootfs \\\"/var/lib/docker/overlay2/da2f555b196e50cd870a45f7c1eb8011351fa95cbb148e33ad7dd4a0fc4f3689/merged\\\" at \\\"/var/lib/docker/overlay2/da2f555b196e50cd870a45f7c1eb8011351fa95cbb148e33ad7dd4a0fc4f3689/merged/etc/ssl/certs/ca-certificates.crt\\\" caused \\\"not a directory\\\"\"": unknown:
Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type
```

This appears to be fixed on `master`, however I imagine many people will instead be using the branch versions.

xrl:
This can be fixed by overriding the `sslCertHostPath` value on the release.

Mine looks like:

```
autoDiscovery:
  clusterName: MyCluster
rbac:
  create: true
sslCertHostPath: /etc/ssl/certs/ca-bundle.crt
podAnnotations:
  iam.amazonaws.com/role: k8s-autoscaler
```

Jeffwan:
Please check common notes and Gotchas here.
https://github.com/kubernetes/autoscaler/tree/cluster-autoscaler-1.13.5/cluster-autoscaler/cloudprovider/aws#common-notes-and-gotchas

I plan to update default value to EKS friendly since more users move from kops to EKS now..

[...]

PR #2230, opened 2019-08-02
CA 1.13: cherry pick of #1970, #1864 and #1740

Jeffwan:
#1970 Add m5ad, r5ad and t3a instance families
#1864 Upgrade CA version in aws example and fix autodiscover example
#1740 Regenerate the ec2 instance types using latest metadata


MaciekPytel:
Why do we need to cherry-pick documentation PR? Do we expect many people to go for documentation to release branch rather than master?
I don't mind approving, but I'd like to understand the rationale.

/approve
/lgtm

Jeffwan:
> Why do we need to cherry-pick documentation PR? Do we expect many people to go for documentation to release branch rather than master?
> I don't mind approving, but I'd like to understand the rationale.
> 
> /approve
> /lgtm

I see some issues like https://github.com/kubernetes/autoscaler/issues/2139 frequently and looks like some of the users go to the version branch to check examples and readme for guidance.

MaciekPytel:
Fair enough. Thanks for pointing it out.

PR #2235, opened 2019-08-03
correctly handle lack of capacity of AWS spot ASGs

piontec:
This is a direct continuation of the https://github.com/kubernetes/autoscaler/pull/2008 PR. I had to switch the source repo.

Please read the PR linked above to get the full context and history. From the last state in that PR I did the changes mentioned by @losipiuk and @gjtempleton.

Please also note, that I no longer have easy access to a spot based EKS cluster on AWS. @seh would you be so kind as to test it?

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
+++ b/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
@@ -19,6 +19,7 @@ package aws
 import (
 	"fmt"
 	"reflect"
+	"regexp"
 	"strings"
 	"sync"
 
@@ -30,7 +31,10 @@ import (
 	"k8s.io/klog"
 )
 
-const scaleToZeroSupported = true
+const (
+	scaleToZeroSupported          = true
+	placeholderInstanceNamePrefix = "i-placeholder"
+)
 
 type asgCache struct {
 	registeredAsgs []*asg
@@ -195,6 +199,10 @@ func (m *asgCache) SetAsgSize(asg *asg, size int) error {
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
 
+	return m.setAsgSizeNoLock(asg, size)
+}
+
+func (m *asgCache) setAsgSizeNoLock(asg *asg, size int) error {
 	params := &autoscaling.SetDesiredCapacityInput{
 		AutoScalingGroupName: aws.String(asg.Name),
 		DesiredCapacity:      aws.Int64(int64(size)),
@@ -210,6 +218,10 @@ func (m *asgCache) SetAsgSize(asg *asg, size int) error {
 	asg.curSize = size
 
 	return nil
+}
+
+func (m *asgCache) decreaseAsgSizeByOneNoLock(asg *asg) error {
+	return m.setAsgSizeNoLock(asg, asg.curSize-1)
 }
 
 // DeleteInstances deletes the given instances. All instances must be controlled by the same ASG.
@@ -239,22 +251,34 @@ func (m *asgCache) DeleteInstances(instances []*AwsInstanceRef) error {
 	}
 
 	for _, instance := range instances {
-		params := &autoscaling.TerminateInstanceInAutoScalingGroupInput{
-			InstanceId:                     aws.String(instance.Name),
-			ShouldDecrementDesiredCapacity: aws.Bool(true),
-		}
-		resp, err := m.service.TerminateInstanceInAutoScalingGroup(params)
-		if err != nil {
-			return err
+		// check if the instance is a placeholder - a requested instance that was never created by the node group
+		// if it is, just decrease the size of the node group, as there's no specific instance we can remove
+		if m.isPlaceholderInstance(instance) {
+			klog.V(4).Infof("instance %s is detected as a placeholder, decreasing ASG requested size instead "+
+				"of deleting instance", instance.Name)
+			m.decreaseAsgSizeByOneNoLock(commonAsg)
+		} else {
+			params := &autoscaling.TerminateInstanceInAutoScalingGroupInput{
+				InstanceId:                     aws.String(instance.Name),
+				ShouldDecrementDesiredCapacity: aws.Bool(true),
+			}
+			resp, err := m.service.TerminateInstanceInAutoScalingGroup(params)
+			if err != nil {
+				return err
+			}
+			klog.V(4).Infof(*resp.Activity.Description)
 		}
 
 		// Proactively decrement the size so autoscaler makes better decisions
 		commonAsg.curSize--
-
-		klog.V(4).Infof(*resp.Activity.Description)
 	}
-
 	return nil
+}
+
+// isPlaceholderInstance checks if the given instance is only a placeholder
+func (m *asgCache) isPlaceholderInstance(instance *AwsInstanceRef) bool {
+	matched, _ := regexp.MatchString(fmt.Sprintf("^%s.*\\d+$", placeholderInstanceNamePrefix), instance.Name)
+	return matched
 }
 
 // Fetch automatically discovered ASGs. These ASGs should be unregistered if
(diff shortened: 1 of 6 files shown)


piontec:
BTW, why the bot thinks I don't have CLA signed? I did that long time ago....

Jeffwan:
> BTW, why the bot thinks I don't have CLA signed? I did that long time ago....

How did you move your commits? Probably your old commits linked to your old email address? Please have a check.

piontec:
> > BTW, why the bot thinks I don't have CLA signed? I did that long time ago....
> 
> How did you move your commits? Probably your old commits linked to your old email address? Please have a check.

Yeah, my bad, still wrong author was in the commits.

xrl:
I have experienced this style of failure in the past. I have a 1/2 ondemand, 1/2 spot EKS cluster under the cluster autoscaler. Let me know if I can help test in any way.

piontec:
@losipiuk @gjtempleton can we merge it now, before it needs rebasing again?

gjtempleton:
Unfortunately I don't have merge permissions on the repo so you'll need someone from the OWNERS file to do it. 

/assign @losipiuk

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
```diff
@@ -355,6 +384,27 @@ func (m *asgCache) regenerate() error {
...
+func (m *asgCache) createPlaceholdersForDesiredNonStartedInstances(groups []*autoscaling.Group) []*autoscaling.Group {
+	for _, g := range groups {
+		desired := *g.DesiredCapacity
+		real := int64(len(g.Instances))
+		if desired <= real {
+			continue
+		}
+
+		for i := real; i < desired; i++ {
+			id := fmt.Sprintf("%s%d", placeholderInstanceNamePrefix, i)
```
MaciekPytel: If you have multiple ASGs with placeholders you'll duplicate ids. I'm not sure if it will create problems or not, but it's probably worth testing.

MaciekPytel:
@losipiuk is on vacation.

/lgtm
/approve

Hold for a single comment I have. Feel free to cancel hold if you've tested it and it works fine.
/hold

cablespaghetti:
Thanks for this PR. I think it solves the risk I have using spot instances for production clusters. 

I've raised a feature request with my company's AWS Technical Account Manager to see if they can implement some kind of "failover" to on demand instance if there is no spot capacity available for an autoscaling group. The mixed instances policy goes most of the way to supporting this but not quite. https://aws.amazon.com/blogs/aws/new-ec2-auto-scaling-groups-with-multiple-instance-types-purchase-options/

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
gjtempleton: Good point, **I'm not 100% sure it'll hold true for this PR as we're running a patched version of CA v1.3.8** but we've had this occur over the past couple of days. 

Behaviour we've seen: The CA only picks up that there's one unregistered node called `i-placeholder-0` at a time despite 2 ASGs having placeholder nodes with the name `i-placeholder-0` so only attempts to decrement the size of one of them, but not the other, until the first group is resolved.

@piontec how about adding the `AutoScalingGroupName` to the end of the placeholder instance name to distinguish these from each other?

Jeffwan:
> Thanks for this PR. I think it solves the risk I have using spot instances for production clusters.
> 
> I've raised a feature request with my company's AWS Technical Account Manager to see if they can implement some kind of "failover" to on demand instance if there is no spot capacity available for an autoscaling group. The mixed instances policy goes most of the way to supporting this but not quite. https://aws.amazon.com/blogs/aws/new-ec2-auto-scaling-groups-with-multiple-instance-types-purchase-options/

I confirmed with Internal AWS ASG team who worked on this feature. `failover` will not be supported at least in short term. If spot has resource constraints, ondemand will have similar problem. But the idea for CA user I think is more on the retry on demand and if could be other AZs.

Seems [spotinst](https://spotinst.com/) have this supported and not sure if there's a reasonable solution in CA. Let's create an separate issue and discuss there?

MaciekPytel:
> But the idea for CA user I think is more on the retry on demand and if could be other AZs.

Retry should happen with current CA error handling, assuming all of those are separate nodegroups. Either because:
1. You have one ASG/config rather than use MixedInstancesPolicy.
2. AWS cloudprovider could represent ASG with MixedInstancesPolicy / multi-AZ as multiple NodeGroups. 

Further improvement would be to return OutOfResouces errors in case of missing spot capacity.

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
jaypipes: Yes, @MaciekPytel has an excellent point. Seems like this is an easy fix (as Guy suggests, simply add the autoscaling group name to the placeholder instance name template).

piontec: Yes, great catch, fixing

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
```diff
@@ -239,24 +251,36 @@ func (m *asgCache) DeleteInstances(instances []*AwsInstanceRef) error {
...
-
-		klog.V(4).Infof(*resp.Activity.Description)
 	}
-
 	return nil
 }
 
+// isPlaceholderInstance checks if the given instance is only a placeholder
+func (m *asgCache) isPlaceholderInstance(instance *AwsInstanceRef) bool {
+	matched, _ := regexp.MatchString(fmt.Sprintf("^%s.*\\d+$", placeholderInstanceNamePrefix), instance.Name)
```
jaypipes: A future improvement might be to compile this regex into the module scope and re-use the compiled regex here. Either that, or just use `strings.HasPrefix()` and ignore the need to verify a numeric component to the name.

seh: This idea [came up here](https://github.com/kubernetes/autoscaler/pull/2008#discussion_r288575645) on 29 May 2019.

jaypipes: ah, indeed it did! :)

Jeffwan:
@piontec 
Looks like there're test case failed. Could you help fix that? 
```
--- FAIL: TestDeleteNodes (0.00s)
    aws_cloud_provider_test.go:426: 
        	Error Trace:	aws_cloud_provider_test.go:426
        	Error:      	Not equal: 
        	            	expected: 1
        	            	actual  : 2
        	Test:       	TestDeleteNodes
        	Messages:   	Expected number of calls (1) does not match the actual number of calls (2).
    aws_cloud_provider_test.go:430: 
        	Error Trace:	aws_cloud_provider_test.go:430
        	Error:      	Not equal: 
        	            	expected: 1
        	            	actual  : 2
        	Test:       	TestDeleteNodes
FAIL
FAIL	k8s.io/autoscaler/cluster-autoscaler/cloudprovider/aws	0.162s

```

piontec:
@Jeffwan Tests are fixed. We need 'lgtm' again.

MaciekPytel:
/lgtm
/hold cancel

Seems like all interested parties have already lgtm-ed. Merging this PR. 
Thanks for the fix @piontec and sorry it got stuck in review for so long.

MaciekPytel:
Given a large interest in addressing underlying issue we should probably cherry-pick this fix. I'll leave the final call on this to AWS folks though.
@jaypipes @Jeffwan

Jeffwan:
@MaciekPytel I will file cherry-picks today.

viggeh:
Thanks for the great work all! Did this ever get cherry picked to previous versions?

gjtempleton:
It doesn't look like it has been yet.

gjtempleton:
@Jeffwan hope I'm not treading on your toes, but I've started raising the cherry-picks of this to 1.12-1.14 as it's of use to us.

Jeffwan:
@gjtempleton Thanks. I was busy with oncall this week and didn't file PR.

BradLugo:
I don't think this found its way into release-1.15

gjtempleton:
@BradLugo Well caught, I'll raise a PR for cherry-picking into 1.15 now.

PR #2241, opened 2019-08-06
Fix linker flags in CA Makefile

### LINKED ISSUE #2239: CA Makefile not setting linker flags properly
Running `make build-binary` produces a 99M binary, and uses an environment variable `LD_FLAGS=-s`.
```
$ make build-binary 
rm -f cluster-autoscaler
CGO_ENABLED=0 LD_FLAGS=-s GOOS=linux go build -o cluster-autoscaler 

$ ls -lh cluster-autoscaler 
-rwxrwxr-x 1 thomas thomas 99M Aug  5 14:04 cluster-autoscaler
```

But, setting the linker flags by passing the flag `-ldflags=-s` directly to `go build` instead produces a smaller binary, 73M.
```
$ CGO_ENABLED=0 GOOS=linux go build -ldflags="-s" -o cluster-autoscaler .

$ ls -lh cluster-autoscaler 
-rwxrwxr-x 1 thomas thomas 73M Aug  5 14:07 cluster-autoscaler
```

I can't find any documentation that suggests that Go does anything with an LD_FLAGS environment variable. There is one called CGO_LDFLAGS but I also see this building a 99M binary.

The linker flags should be changed to be similar to `TAGS_FLAG`.

tghartland:
Fixes #2239 

Linker flags are moved to their own variable `LDFLAGS`, and `LDFLAGS_FLAG` is constructed to pass to `go build`.

Default is `-s` as intended with the original change. Also works with multiple flags:

```
$ LDFLAGS="-s -w" make build-binary 
rm -f cluster-autoscaler
CGO_ENABLED=0 GOOS=linux go build -o cluster-autoscaler --ldflags "-s -w"
```

```
$ LDFLAGS="-s -w" make build-in-docker
.....
.....
docker run -v `pwd`:/gopath/src/k8s.io/autoscaler/cluster-autoscaler/ autoscaling-builder:latest bash -c 'cd /gopath/src/k8s.io/autoscaler/cluster-autoscaler && BUILD_TAGS= LDFLAGS="-s -w" make build-binary'
rm -f cluster-autoscaler
CGO_ENABLED=0 GOOS=linux go build -o cluster-autoscaler --ldflags "-s -w"
```

### DIFF
--- a/cluster-autoscaler/Makefile
+++ b/cluster-autoscaler/Makefile
@@ -2,7 +2,8 @@ all: build
 
 TAG?=dev
 FLAGS=
-ENVVAR=CGO_ENABLED=0 LD_FLAGS=-s
+LDFLAGS?=-s
+ENVVAR=CGO_ENABLED=0
 GOOS?=linux
 REGISTRY?=staging-k8s.gcr.io
 ifdef BUILD_TAGS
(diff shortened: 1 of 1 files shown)


MaciekPytel:
I think @losipiuk knows most about our build process these days, so I'll leave this one to him. He's on vacation this week though.

losipiuk:
/lgtm
/approve
Thanks!

Issue #2225, opened 2019-07-30
Incorrect backoff duration in case of overlapping scale-ups


MaciekPytel:
`duration` variable is not properly updated in https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/utils/backoff/exponential_backoff.go#L69 if the NodeGroup is already in backoff. As a result, if there are multiple concurrent scale-ups the duration of backoff will keep resetting to `initialBackoffDuration`.

In default configuration `--scale-down-delay-after-add` is 10m and `initialBackoffDuration` is 5m. This means that CA will not do any scale-down within 10m of scale-up and it will retry failed scale-up every 5 minutes. In other words a single NodeGroup that consistently fails to scale-up will block any scale-down in cluster.

Denton24646:
I'd like to take a shot at this bug - if I understand correctly, we need to check for backoff already inside `Backoff()` via `IsBackedOff()` and in the case where the NodeGroup is already in backoff, increment it appropriately so it ends up higher than `--scale-down-delay-after-add` at some point to not block as you described. It's nice because it's fairly testable.

MaciekPytel:
I meant to fix it myself later this week, but I'm happy to review your fix if you beat me to it. I think the fix is literally 1 LOC in Backoff() + unittest.

PR #2244, opened 2019-08-07
fix: increase exponential backoff duration regardless of backoff status

Denton24646:
PR addresses issue raised in https://github.com/kubernetes/autoscaler/issues/2225.

Backoff duration was not being exponentially increased for cases where the node group was already in a backoff state, instead setting the duration to the initial default duration. This could cause the cluster autoscaler to be in a locked state, where the scale-down-after-scale-up cooldown would always be greater than the duration between scale up attempts. So as scale-up attempts successively failed within the window of cooldown, no scale down could take place. 

The solution is to not set `duration=initialDuration` if `found=true`, instead setting it to the existing`backoffInfo.duration`, which allows for the duration to increase beyond initial.

### DIFF
--- a/cluster-autoscaler/utils/backoff/exponential_backoff.go
+++ b/cluster-autoscaler/utils/backoff/exponential_backoff.go
@@ -70,10 +70,12 @@ func (b *exponentialBackoff) Backoff(nodeGroup cloudprovider.NodeGroup, nodeInfo
 	duration := b.initialBackoffDuration
 	key := b.nodeGroupKey(nodeGroup)
 	if backoffInfo, found := b.backoffInfo[key]; found {
-		// Multiple concurrent scale-ups failing shouldn't cause backoff
-		// duration to increase, so we only increase it if we're not in
-		// backoff right now.
+		// Multiple concurrent scale-ups failing shouldn't cause
+		// backoff duration to increase exponentially
+		duration = backoffInfo.duration
 		if backoffInfo.backoffUntil.Before(currentTime) {
+			// NodeGroup is not currently in backoff, but was recently
+			// Increase backoff duration exponentially
 			duration = 2 * backoffInfo.duration
 			if duration > b.maxBackoffDuration {
 				duration = b.maxBackoffDuration
--- a/cluster-autoscaler/utils/backoff/exponential_backoff_test.go
+++ b/cluster-autoscaler/utils/backoff/exponential_backoff_test.go
@@ -81,3 +81,23 @@ func TestResetStaleBackoffData(t *testing.T) {
 	backoff.RemoveStaleBackoffData(startTime.Add(5 * time.Hour))
 	assert.Equal(t, 0, len(backoff.(*exponentialBackoff).backoffInfo))
 }
+
+func TestIncreaseExistingBackoff(t *testing.T) {
+	backoff := NewIdBasedExponentialBackoff(1*time.Second, 10*time.Minute, 3*time.Hour)
+	startTime := time.Now()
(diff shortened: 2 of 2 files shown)


cluster-autoscaler/utils/backoff/exponential_backoff.go
```diff
@@ -70,14 +70,12 @@ func (b *exponentialBackoff) Backoff(nodeGroup cloudprovider.NodeGroup, nodeInfo
...
-		// backoff right now.
-		if backoffInfo.backoffUntil.Before(currentTime) {
-			duration = 2 * backoffInfo.duration
-			if duration > b.maxBackoffDuration {
-				duration = b.maxBackoffDuration
-			}
+		// Multiple concurrent scale-ups failing should cause backoff
+		// duration to increase, to not block cluster scale-down
+		// Increase backoff duration if we are in backoff already or not
+		duration = 2 * backoffInfo.duration
```
MaciekPytel: I'd prefer to keep the if statement and just set duration correctly in else or something like that. 
A single failed scale-up can easily trigger calling Backoff() more than once (if you add multiple instances in one go it's likely for errors to crop up over multiple CA loop). With your change you're likely to just immediately jump to maximum backoff duration, rather than increase it exponentially as you retry.

Denton24646: Ok that makes sense. I'll refactor it to keep the existing if statement and add logic to the else condition instead. Thanks for the additional context.

Denton24646: @MaciekPytel I have updated it to something that more closely resembles what you're looking for but I'm having some questions. I know we want to increase the duration in `Backoff()` if the NG is already in backoff, but you're saying we do not want to do it on every reentrant call to `Backoff()` because then it goes up to the max too fast (like in my initial commit). Seems like we need some additional state in `backoffInfo` to inform how many retries we have, but you mention the solution should be only one LOC. Maybe you could provide the condition you're looking for?

MaciekPytel: Right, so this requires more context than I initially thought. When NG goes off backoff, the backoff entry stays around for a few more hours (NodeGroupBackoffResetTimeout const). My idea when first implementing backoff was not to increase duration of the backoff if the NG is already in backoff (hence the if statement). The duration would be increased only if NG is not in backoff now, but has been in backoff recently (ie. the backoff entry is still there).

The rationale is that CA will not try another scale-up while the NG is already in backoff, so any new failure can likely be attributed to the original scale-up. Once NG goes off backoff and another scale-up fails the backoff duration will increase.

Not sure if this is the best way to do this, but your current approach doesn't fix the original issue - it is possible to set `duration = initialBackoffDuration` even if `found = true`.
I think the fix may be just as simple as adding `duration = backoffInfo.duration` in https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/utils/backoff/exponential_backoff.go#L75.

Denton24646: Ok thanks for the context, this all makes more sense now. I will implement the approach and update the test.

Denton24646: @MaciekPytel updated

MaciekPytel:
Can you squash your commits? Otherwise lgtm.

Denton24646:
> Can you squash your commits? Otherwise lgtm.

Squashed @MaciekPytel -- thanks

MaciekPytel:
/lgtm
/approve

Thanks for fixing this!

PR #2245, opened 2019-08-08
cloudprovider: add DigitalOcean

### LINKED ISSUE #254: Implement DigitalOcean cloud provider

fatih:
This adds a new cluster autoscaler for [DigitalOcean Kubernetes ](https://www.digitalocean.com/products/kubernetes/) offering. 

* Because there is no native NodeGroup offering in DigitalOcean this autoscaler can be used only within a Managed DigitalOcean Kubernetes cluster. 
* Customers will be able to enable/disable it by adding tags to their **node pools**. Some of the valid tags are:

```
k8s-cluster-autoscaler-enabled:true
k8s-cluster-autoscaler-min:3
k8s-cluster-autoscaler-max:10
```

* As suggested, I didn't vendored the client (`cloudprovider/digitalocean/godo`) and added it as a sub package. 

For reviewers:

I've tried to add each individual commit in a packaged way. The first commit (https://github.com/kubernetes/autoscaler/commit/776bba5e1059da5c5b7a3937e0f09836b7bf515e) adds our DigitealOcean API package called `godo`. The second and following commits are all related to the actual DigitalOcean cloud provider. 

cc @timoreimann @snormore

cc @andrewsykim @MaciekPytel 

Closes #254


fatih:
I've closed the previous PR as it was based on my personal repository. This new PR is the same, but the changes are based on https://github.com/digitalocean/autoscaler

cluster-autoscaler/cloudprovider/digitalocean/README.md
```diff
@@ -0,0 +1,58 @@
...
+* To set the minimum number of nodes to use `k8s-cluster-autoscaler-min`
+* To set the maximum number of nodes to use `k8s-cluster-autoscaler-max`
+
+
+If you don't set the minimum and maximum tags, node pools will have the
+following default limits:
+
+```
+minimum number of nodes: 1
+maximum number of nodes: 200
```
MaciekPytel: That is significantly different UX from other providers - I think all (?) existing providers require explicitly setting at least max limit.

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_cloud_provider.go
```diff
@@ -0,0 +1,193 @@
...
+
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
+	"k8s.io/autoscaler/cluster-autoscaler/config"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
+	"k8s.io/klog"
+)
+
+var _ cloudprovider.CloudProvider = (*digitaloceanCloudProvider)(nil)
```
MaciekPytel: What is the purpose of this line?

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_cloud_provider.go
```diff
@@ -0,0 +1,193 @@
...
+	// proceed with this.
+	for _, group := range d.manager.nodeGroups {
+		klog.V(5).Infof("iterating over node group %q", group.Id())
+		nodes, err := group.Nodes()
+		if err != nil {
+			return nil, err
+		}
+
+		for _, node := range nodes {
+			klog.V(5).Infof("checking node have: %q want: %q", node.Id, nodeID)
```
MaciekPytel: This will produce incredible amount of logs in a decently sized cluster. There are some V(5) logs that are sometimes useful for debugging, but this make V(5) unusable - can you make it V(6) or more?

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_cloud_provider.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+// BuildDigitalOcean builds the DigitalOcean cloud provider.
+func BuildDigitalOcean(
+	opts config.AutoscalingOptions,
+	do cloudprovider.NodeGroupDiscoveryOptions,
+	rl *cloudprovider.ResourceLimiter,
+) cloudprovider.CloudProvider {
+	var configFile io.ReadCloser
+	if opts.CloudConfig != "" {
```
MaciekPytel: Shouldn't that error/fatal if CloudConfig == ""?

cluster-autoscaler/cloudprovider/digitalocean/README.md
```diff
@@ -0,0 +1,58 @@
+# Cluster Autoscaler for DigitalOcean
+
+The cluster autoscaler for DigitalOcean scales worker nodes within any
+specified DigitalOcean Kubernetes cluster's node pool. This is part of the DOKS
+offering which can be enabled/disable dynamically for an existing cluster.
+
+# Configuration
+
+The `cluster-autoscaler` dynamically runs based on tags associated with node
+pools. These are the current valid tags:
```
MaciekPytel: This is completely different from how you run CA on different providers. Any particular reason for this design?

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_manager.go
```diff
@@ -0,0 +1,248 @@
...
+func (m *Manager) Refresh() error {
+	ctx := context.Background()
+	nodePools, _, err := m.client.ListNodePools(ctx, m.clusterID, nil)
+	if err != nil {
+		return err
+	}
+
+	var group []*NodeGroup
+	for _, np := range nodePools {
+		nodePool, resp, err := m.client.GetNodePool(ctx, m.clusterID, np.ID)
```
MaciekPytel: Mostly asking out of curiosity - why do you need to do this? Seems like ListNodePools already return *godo.KubernetesNodePool for each node pool, why fetch the same thing again? Is the struct coming from list not completely populated?

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_manager.go
```diff
@@ -0,0 +1,248 @@
...
+		switch key {
+		case "enabled":
+			if value == "true" {
+				spec.enabled = true
+			}
+		case "min":
+			min, err := strconv.Atoi(value)
+			if err != nil {
+				return nil, fmt.Errorf("invalid minimum nodes: %q", value)
+			}
```
MaciekPytel: Explicitly return an error for min=0? Currently you just silently override it to 1. Many other providers support scale-to-0 so it's possible for someone to assume digitalocean also supports it and just set it.

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_node_group.go
```diff
@@ -0,0 +1,269 @@
...
+	if err != nil {
+		return err
+	}
+
+	if updatedNodePool.Count != targetSize {
+		return fmt.Errorf("couldn't increase size to %d (delta: %d). Current size is: %d",
+			targetSize, delta, updatedNodePool.Count)
+	}
+
+	return nil
```
MaciekPytel: You seem to cache TargetSize, in which case you should update it after changing it.

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_node_group.go
```diff
@@ -0,0 +1,269 @@
...
+			nodeID = node.Spec.ProviderID
+		}
+
+		_, err := n.client.DeleteNode(ctx, n.clusterID, n.id, nodeID, nil)
+		if err != nil {
+			return fmt.Errorf("deleting node failed for cluster: %q node pool: %q node: %q: %s",
+				n.clusterID, n.id, nodeID, err)
+		}
+	}
+
```
MaciekPytel: Update cached TargetSize?

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_node_group.go
```diff
@@ -0,0 +1,269 @@
...
+	updatedNodePool, _, err := n.client.UpdateNodePool(ctx, n.clusterID, n.id, req)
+	if err != nil {
+		return err
+	}
+
+	if updatedNodePool.Count != targetSize {
+		return fmt.Errorf("couldn't increase size to %d (delta: %d). Current size is: %d",
+			targetSize, delta, updatedNodePool.Count)
+	}
+
```
MaciekPytel: Update cached TargetSize?

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_node_group.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+// Nodes returns a list of all nodes that belong to this node group.  It is
+// required that Instance objects returned by this method have Id field set.
+// Other fields are optional.
+func (n *NodeGroup) Nodes() ([]cloudprovider.Instance, error) {
+	if n.nodePool == nil {
+		return nil, errors.New("node pool instance is not created")
+	}
+	return toInstances(n.nodePool.Nodes), nil
```
MaciekPytel: To get proper error handling you should return 'placeholder' instances for any instance that doesn't exist yet (n.nodePool.Count - len(n.nodePool.Nodes). Unless these is always 0?

It's fine to add it in a separate PR later. This PR is a recent example of this: https://github.com/kubernetes/autoscaler/pull/2235 and could be used as a reference.

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_cloud_provider.go
andrewsykim: This verifies at compile time that `digitaloceanCloudProvider` implements the interface `cloudprovider.CloudProvider`.

andrewsykim:
> Left a few comments, but overall looks quite good. Seems like @andrewsykim is also reviewing this? Happy to approve once he lgtms.

Left most of my review comments in the previous PR https://github.com/kubernetes/autoscaler/pull/2227.  Overall PR LGTM as well. 

> Also - feel free to add digitalocean/OWNERS file with yourself and / or whoever is interested in maintaining this from DO side. Core developers have only so much review capacity and as the number of cloudproviders grow we don't want to block internal fixes / improvements on our reviews.

@fatih feel free to add myself here as well if you need someone in the OWNERs file (note you need to be an kubernetes org member for this to work).

cluster-autoscaler/cloudprovider/digitalocean/README.md
fatih: This is the default limit for a given node group right now. It's out upper limit. I can change it to 50 or something else if that makes more sense to you? In the UI (cloud.digitalocean.com) we're going to expose a UI that will have sensible pre-default values.

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_cloud_provider.go
fatih: Let's us catch issues if someone modifies the interface. So you'll see immediately that the digitalocean cloud provider doesn't satisfy the interface.

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_cloud_provider.go
fatih: Made it V(6).

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_cloud_provider.go
fatih: This resembles the same layout as the other cloud providers. It doesn't have to error, because in the next step , `newManager()` will throw an error because of an invalid configuration.

cluster-autoscaler/cloudprovider/digitalocean/README.md
fatih: This let us dynamically enable/disable the cluster-autoscaler and make it by default as `opt-out`. So to opt-in you need to add those tags. Because our UI is not ready yet, this will allow  some customers to enable it by adding the tags themself. Also this doesn't require us fiddle with CA flag arguments. It's easier to maintain and ship cluster-autoscaler and requires zero to minimum interaction once it's deployed. Is there something you believe is wrong in this approach?

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_manager.go
fatih: Good catch Maciek! This came up in our internal meeting as well. I think this was initially added because I wasn't using `ListNodePools` in the first place, it was a cached list of nodePools. I'll change it 👍

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_manager.go
fatih: We don't support `0` yet :/ Hence I return an error here. This is one of our short comings on our end we're working on fixing it.

cluster-autoscaler/cloudprovider/digitalocean/README.md
```diff
@@ -0,0 +1,58 @@
...
+# Development
+
+Make sure you're inside the root path of the [autoscaler
+repository](https://github.com/kubernetes/autoscaler)
+
+1.) Build the `cluster-autoscaler` binary:
+
+
+```
+make build-binary
```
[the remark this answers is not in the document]
fatih: I've just tested it and seems like our API depends on https://github.com/google/go-querystring. Can I vendor this or should I also copy this as a subfolder and rewrite the import paths of our API?

fatih:
@andrewsykim sounds good thanks! I've added you to `reviewers` section and our @timoreimann as I think he is a Kubernetes Org Member.

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_cloud_provider.go
MaciekPytel: Makes sense. Thanks for clarification.

cluster-autoscaler/cloudprovider/digitalocean/README.md
MaciekPytel: I don't have any specific recommendation for the value. I was referring to the fact that there is a default at all, rather than an explicit requirement to specify a value.

cluster-autoscaler/cloudprovider/digitalocean/README.md
MaciekPytel: Oh, I didn't realize you're planning to host CA pod (or at least start it automatically). I don't think there is anything inherently wrong with your approach, it's just inconsistent with other providers.
Given that autodiscovery already works differently for each provider I don't think it's a blocker for merging this either. Just a suboptimal experience for someone moving between clouds / running multicloud / etc.

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_manager.go
MaciekPytel: My comment is that you don't actually return any error. You just treat 0 as 'not set' and silently default to 1.

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_manager.go
```diff
@@ -0,0 +1,235 @@
...
+			spec.min = min
+		case "max":
+			max, err := strconv.Atoi(value)
+			if err != nil {
+				return nil, fmt.Errorf("invalid minimum nodes: %q", value)
```
MaciekPytel: s/minimum/maximum

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_manager.go
MaciekPytel: Interestingly it seems like you can also set max to a negative value as long as you don't set min?

cluster-autoscaler/cloudprovider/digitalocean/README.md
fatih: I see what you mean. This is just a way of limiting the hard maximum we allow. The user experience will be different and our API will validate requests.

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_manager.go
fatih: Yeah you're right. Added this test case and it fails:

```
		{
			name: "bad tags - max is set to negative, no min",
			tags: []string{
				"k8s-cluster-autoscaler-enabled:true",
				"k8s-cluster-autoscaler-max:-5",
			},
			wantErr: true,
		},
```

Making sure to fix it.

MaciekPytel:
Please squash your commits once you're ready to merge, so that there are just 2: adding client and implementation.

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_manager.go
fatih: Fixed

fatih:
> Please squash your commits once you're ready to merge, so that there are just 2: adding client and implementation.

@MaciekPytel just did it. The PR now only contains two commits, including recent fixes around zero minimum and negative maxes. 

There is only one conversation that is not resolved though for us: https://github.com/kubernetes/autoscaler/pull/2245#discussion_r312374847 What should I do here?

cluster-autoscaler/cloudprovider/digitalocean/README.md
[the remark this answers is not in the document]
MaciekPytel: @mwielgus what do you think about this one?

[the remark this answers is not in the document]
mwielgus: So far we have been very strict on adding any cloud-provider-specific dependencies to global CA vendors. To keep our integrity and not create precedent(even for small Google library) we would kindly ask you to contain everything what your cloud provider needs in your directory.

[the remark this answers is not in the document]
fatih: @mwielgus gotcha. I vendored everything locally into the `digitalocean` folder. I re-ran `make build-in-docker` again and now it works without any issue. I've force pushed the changes as I've rebased and squashed everything.

fatih:
@MaciekPytel everything is settled down now. I've fixed all your comments (I think I didn't missed anything). I also squashed and splitted all my changes into two commits. The first commit contains the client (including the dependencies) inside the `digitalocean` folder, and the second commit contains the actual cloudprovider implementation. PTAL

micahhausler:
I'm new to this project, but why is the whole godo source copied in a subdirectory instead of just add godo as a vendored dependency? This seems to break `go mod` in awful ways

fatih:
@micahhausler I also wanted to use `go mod`, but the requirement was that new cloud providers are not allowed to vendor any code. Hence we had to copy all dependencies into subfolders.

micahhausler:
Hmm that seems strange, where is that documented?

fatih:
I'm no longer involved with this project, please ask #sig-autoscaling in Slack or @MaciekPytel.

Issue #2246, opened 2019-08-08
Unable to get instance type from launch config or launch template - v1.13.6


mmingorance-dh:
We have started to use launch-templates-mixed in our Kubernetes cluster 1.13 on EKS and apparently, when we upgraded to cluster-autoscaler 1.13.6 we started fo face this issue with the autoscaling groups using launch-templates-mixed
*Unable to build proper template node for $AUTOSCALING_GROUP_NAME: Unable to get instance type from launch config or launch template*

When rolling back to cluster-autoscaler 1.13.1 this problems does not appear at all.

Here are some logs from 1.13.6
[logs-ca-1-13-6.txt](https://github.com/kubernetes/autoscaler/files/3481798/logs-ca-1-13-6.txt)

MaciekPytel:
/cc @Jeffwan @jaypipes
This seems very specific to AWS provider.

jaypipes:
https://github.com/kubernetes/autoscaler/commit/4bd79c83531d449e9ed1451cd030a9f34cc3e470 introduced the tracking of launch template or config. The `AwsManager.buildInstanceType` function hasn't changed substantially since July last year (well before 1.13.1).

This error will be returned if the ASG has no launch configuration name or launch template name+version.

What does your ASG look like?

Also, see in your logs this warning:
```
W0808 12:28:18.713213       1 aws_manager.go:194] Found multiple availability zones for ASG "v1-13-testing-eu-tools-blue20190807074452737800000026"; using eu-west-1b
```

Please note that multi-AZ ASGs are not supported. Please see the second bullet point here:

https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/aws#common-notes-and-gotchas

mmingorance-dh:
@jaypipes Thanks for replying.
multi-AZ ASGs can lead to nodes being terminated by AWS instead of cluster-autoscaler but it doesn't cause any problem with cluster-autoscaler terminating and launching new instances. We are running our ASGs in multiple AZs since very long time ago.

My ASG looks like this:
```
{
    "AutoScalingGroups": [
        {
            "AutoScalingGroupARN": "$ARN",
            "ServiceLinkedRoleARN": "$RoleARN",
            "TargetGroupARNs": [],
            "SuspendedProcesses": [
                {
                    "ProcessName": "AZRebalance",
                    "SuspensionReason": "User suspended at 2019-08-07T07:45:04Z"
                }
            ],
[...]

PR #2248, opened 2019-08-12
Add MixedInstancesPolicy struct to better handle instance type

Jeffwan:
Fix issue that user can not build instance type using MixedInstanceType. 

https://github.com/kubernetes/autoscaler/issues/2246

@jaypipes

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/README.md
+++ b/cluster-autoscaler/cloudprovider/aws/README.md
@@ -179,6 +179,8 @@ If you'd like to scale node groups from 0, an `autoscaling:DescribeLaunchConfigu
 
 ## Using AutoScalingGroup MixedInstancesPolicy
 
+> Note: The minimum version of cluster autoscaler to support MixedInstancePolicy is v1.14.x.
+
 It is possible to use Cluster Autoscaler with a [mixed instances policy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-autoscaling-autoscalinggroup-mixedinstancespolicy.html), to enable diversification across on-demand and spot instances, of multiple instance types in a single ASG. When using spot instances, this increases the likelihood of successfully launching a spot instance to add the desired capacity to the cluster versus a single instance type, which may be in short supply.
 
 Note that the instance types should have the same amount of RAM and number of CPU cores, since this is fundamental to CA's scaling calculations. Using mismatched instances types can produce unintended results.
--- a/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
+++ b/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
@@ -44,6 +44,16 @@ type asgCache struct {
 	explicitlyConfigured  map[AwsRef]bool
 }
 
+type launchTemplate struct {
+	name    string
+	version string
+}
+
+type mixedInstancesPolicy struct {
+	launchTemplate         *launchTemplate
+	instanceTypesOverrides []string
+}
+
 type asg struct {
 	AwsRef
 
@@ -52,9 +62,9 @@ type asg struct {
 	curSize int
 
 	AvailabilityZones       []string
-	LaunchTemplateName      string
-	LaunchTemplateVersion   string
 	LaunchConfigurationName string
+	LaunchTemplate          *launchTemplate
+	MixedInstancesPolicy    *mixedInstancesPolicy
 	Tags                    []*autoscaling.TagDescription
 }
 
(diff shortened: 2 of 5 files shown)


jsharpe:
I've been looking at this issue; I've not tried this patch but based on the code in master I'm seeing the LaunchTemplate from the MixedInstances configuration returning an empty string for the template version. Setting this to "$Default" when the string is empty appears to fix the issue of it not being able to construct node information when scaling to 0.

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -179,6 +179,8 @@ If you'd like to scale node groups from 0, an `autoscaling:DescribeLaunchConfigu
 
 ## Using AutoScalingGroup MixedInstancesPolicy
 
+> Note: The minimum supported version of MixedInstancePolicy is CA v1.14.x.
```
jaypipes: This should read "The minimum version of cluster autoscaler to support MixedInstancePolicy is CA v1.14.x"

The way it currently reads makes it sound like there is a minimum version of AutoScalingGroup.MixedInstancesPolicy, not cluster autoscaler.

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
```diff
@@ -377,23 +385,36 @@ func (m *asgCache) buildAsgFromAWS(g *autoscaling.Group) (*asg, error) {
...
 	return asg, nil
 }
 
-func (m *asgCache) buildLaunchTemplateParams(g *autoscaling.Group) (string, string) {
-	if g.LaunchTemplate != nil {
-		return aws.StringValue(g.LaunchTemplate.LaunchTemplateName), aws.StringValue(g.LaunchTemplate.Version)
-	} else if g.MixedInstancesPolicy != nil && g.MixedInstancesPolicy.LaunchTemplate != nil {
-		return aws.StringValue(g.MixedInstancesPolicy.LaunchTemplate.LaunchTemplateSpecification.LaunchTemplateName),
-			aws.StringValue(g.MixedInstancesPolicy.LaunchTemplate.LaunchTemplateSpecification.Version)
+func (m *asgCache) buildLaunchTemplateParams(ltSpec *autoscaling.LaunchTemplateSpecification) *launchTemplate {
```
jaypipes: recommend just calling this `buildLaunchTemplate()` or `launchTemplateFromSpec()`

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
```diff
@@ -44,6 +44,16 @@ type asgCache struct {
...
 }
 
+type launchTemplate struct {
+	launchTemplateName    string
+	launchTemplateVersion string
+}
+
+type mixedInstancesPolicy struct {
+	launchTemplate *launchTemplate
+	instanceTypes  []string
```
jaypipes: recommend calling this field `overrides` or `instanceTypeOverrides` to better describe what they represent, which is the `MixedInstancePolicy.Overrides` field in the ASG.

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
```diff
@@ -44,6 +44,16 @@ type asgCache struct {
 	explicitlyConfigured  map[AwsRef]bool
 }
 
+type launchTemplate struct {
+	launchTemplateName    string
```
jaypipes: recommend just calling these field "name" and "version". We already know they are for the launch template since they are fields on the launchTemplate :)

Jeffwan:
@jaypipes Thanks for the feedbacks. Resolve all the comments and please have another look

Jeffwan:
> > Fix issue that user can not build instance type using MixedInstanceType.
> > #2246
> 
> Jiaxin, I'd appreciate a little bit more descriptive commit message/PR comment than above. Something that describes exactly what was being fixed and why.
> 
> Perhaps something like this?
> 
> "Ensures that when MixedInstancePolicy is used in an AWS AutoScalingGroup, that
> the buildInstanceType() AWS manager method returns an instance type after looking
> at the MixedInstancePolicy.LaunchTemplateSpecification. The buildInstanceType()
> method is called in numerous places including on cluster scale up actions.
> 
> Also adds documentation highlighting the minimum version of cluster autoscaler
> supporting MixedInstancePolicy is 1.14."

Great! I add the comment to git commit.

losipiuk:
/lgtm
/approve
Thanks!

coreypobrien:
@losipiuk @Jeffwan Will this need to be cherry picked back to 1.15 and 1.14?

Jeffwan:
@coreypobrien This is a bug fix and should be a cherry-pick candidate to 1.14 and 1.15. 
Can you help cherry-pick the change and I can approve it.

Issue #2240, opened 2019-08-05
Load EC2 instance list dynamically when cluster autoscaler starts up


Jeffwan:
/feature /improvement

Right now, aws instance templates `ec2_instance_types.go` are generate by `ec2_instance_types/gen.go`. Everytime Ec2 rollout new instances, this file needs to be regenerated and CA needs to roll out new images.

This can be improved by generating ec2 instance types in runtime. Either at the time CA starts or the first time to get the instance list. With this, we don't need to frequently update this file and cherry-pick to elder versions.

PR #2249, opened 2019-08-12
Generate EC2 Instance Type list in runtime

Jeffwan:
https://github.com/kubernetes/autoscaler/issues/2240

We have some discussion here. https://github.com/kubernetes/autoscaler/pull/2231

Consider the case not all k8s clusters has network to aws pricing API, I still keep the static instance type list. If network is not accessible, it will fall back to static dict. 

dynamic loading will reuse code snippet from gen.go

I am trying reach out to folks who worked on this, not sure if we have any internal IP like EC2 metadata service. If so, we can use that and assume it can always get updated list in the same network. 

@jaypipes

### DIFF
--- a/cluster-autoscaler/FAQ.md
+++ b/cluster-autoscaler/FAQ.md
@@ -674,6 +674,7 @@ The following startup parameters are supported for cluster autoscaler:
 | `leader-elect-renew-deadline` | The interval between attempts by the acting master to renew a leadership slot before it stops leading.<br>This must be less than or equal to the lease duration.<br>This is only applicable if leader election is enabled | 10 seconds
 | `leader-elect-retry-period` | The duration the clients should wait between attempting acquisition and renewal of a leadership.<br>This is only applicable if leader election is enabled | 2 seconds
 | `leader-elect-resource-lock` | The type of resource object that is used for locking during leader election.<br>Supported options are `endpoints` (default) and `configmaps` | "endpoints"
+| `aws-use-static-instance-list` | Should CA fetch instance types in runtime or use a static list. AWS only | false
 
 # Troubleshooting:
 
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
@@ -49,13 +49,16 @@ var (
 type awsCloudProvider struct {
 	awsManager      *AwsManager
 	resourceLimiter *cloudprovider.ResourceLimiter
+	// InstanceTypes is a map of ec2 resources
+	instanceTypes map[string]*InstanceType
 }
 
 // BuildAwsCloudProvider builds CloudProvider implementation for AWS.
-func BuildAwsCloudProvider(awsManager *AwsManager, resourceLimiter *cloudprovider.ResourceLimiter) (cloudprovider.CloudProvider, error) {
+func BuildAwsCloudProvider(awsManager *AwsManager, instanceTypes map[string]*InstanceType, resourceLimiter *cloudprovider.ResourceLimiter) (cloudprovider.CloudProvider, error) {
 	aws := &awsCloudProvider{
 		awsManager:      awsManager,
 		resourceLimiter: resourceLimiter,
+		instanceTypes:   instanceTypes,
 	}
 	return aws, nil
 }
@@ -343,12 +346,37 @@ func BuildAWS(opts config.AutoscalingOptions, do cloudprovider.NodeGroupDiscover
 		defer config.Close()
 	}
 
+	// Generate EC2 list
+	var instanceTypes map[string]*InstanceType
+	var lastUpdateTime string
+	if opts.AWSUseStaticInstanceList {
+		instanceTypes, lastUpdateTime = GetStaticEC2InstanceTypes()
+		klog.Warningf("Use static EC2 Instance Types and list could be outdated. Last update time: %s", lastUpdateTime)
+	} else {
+		region, err := GetCurrentAwsRegion()
+		if err != nil {
+			klog.Fatalf("Failed to get AWS Region: %v", err)
+		}
+
+		instanceTypes, err = GenerateEC2InstanceTypes(region)
+		if err != nil {
+			klog.Fatalf("Failed to generate AWS EC2 Instance Types: %v", err)
+		}
+
+		keys := make([]string, 0, len(instanceTypes))
+		for key := range instanceTypes {
+			keys = append(keys, key)
+		}
+
+		klog.Infof("Successfully load %d EC2 Instance Types %s", len(keys), keys)
+	}
+
 	manager, err := CreateAwsManager(config, do)
 	if err != nil {
 		klog.Fatalf("Failed to create AWS Manager: %v", err)
 	}
 
-	provider, err := BuildAwsCloudProvider(manager, rl)
+	provider, err := BuildAwsCloudProvider(manager, instanceTypes, rl)
 	if err != nil {
 		klog.Fatalf("Failed to create AWS cloud provider: %v", err)
 	}
--- a/cluster-autoscaler/cloudprovider/aws/aws_util.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_util.go
@@ -1,0 +1,173 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package aws
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"io/ioutil"
+	"k8s.io/klog"
+	"net/http"
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+var (
+	ec2MetaDataServiceUrl        = "http://169.254.169.254/latest/dynamic/instance-identity/document"
+	ec2PricingServiceUrlTemplate = "https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonEC2/current/%s/index.json"
+	staticListLastUpdateTime     = "2019-10-14"
+)
+
+type response struct {
+	Products map[string]product `json:"products"`
+}
+
+type product struct {
+	Attributes productAttributes `json:"attributes"`
+}
+
+type productAttributes struct {
+	InstanceType string `json:"instanceType"`
+	VCPU         string `json:"vcpu"`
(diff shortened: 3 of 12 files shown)


Jeffwan:
> Tests are failing
> 
> ```
> FAIL	k8s.io/autoscaler/cluster-autoscaler/cloudprovider/aws	0.114s
> ```

I will update the test case.

Jeffwan:
@mwielgus Tests have been fixed. Please have a review.

cluster-autoscaler/cloudprovider/aws/aws_util_test.go
```diff
@@ -0,0 +1,53 @@
...
+func TestParseMemory(t *testing.T) {
+	memoryInGiB := 3.75
+	expectedResultInMiB := int64(memoryInGiB * 1024)
+	result := parseMemory("3.75 GiB")
+	assert.Equal(t, expectedResultInMiB, result)
+
+	result = parseMemory("3.75 Gib")
+	assert.Equal(t, expectedResultInMiB, result)
+
+	result = parseMemory("3.75 Gib")
```
jaypipes: AFAICT, this is the same test as on line 36.

jaypipes: Also, for a future improvement, consider using table-based testing for things like this. Something like this would work:

```go
tests := []struct{
    input string
    expect float
}{
    {
        input: "3.75 GiB",
        expect: 3.75,
    },
    {
        input: "3.75 Gib",
        expect: 3.75,
    },
    {
        input: "3.75GiB",
        expect: 3.75,
    },
    {
        input: "3.75",
        expect: 3.75,
    },
}

for _, test := range tests {
    got := parseMemory(test.input)
    assert.Equal(t, test.expect, got)
}
```

Jeffwan: Address feedbacks in latest commit. Thanks!

cluster-autoscaler/cloudprovider/aws/aws_util.go
```diff
@@ -0,0 +1,131 @@
...
+// GenerateEC2InstanceTypes returns a map of ec2 resources
+func GenerateEC2InstanceTypes() (map[string]*InstanceType, error) {
+	instanceTypes := make(map[string]*InstanceType)
+
+	resolver := endpoints.DefaultResolver()
+	partitions := resolver.(endpoints.EnumPartitions).Partitions()
+
+	for _, p := range partitions {
+		for _, r := range p.Regions() {
+			url := "https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonEC2/current/" + r.ID() + "/index.json"
```
jaypipes: just another little thing to improve in a followup, you might want to put the pricing API URL into a `const` at the top of this file.

cluster-autoscaler/cloudprovider/aws/aws_util_test.go
```diff
@@ -0,0 +1,75 @@
...
+			input:  strconv.FormatInt(8, 10),
+			expect: int64(8),
+		},
+	}
+
+	for _, test := range tests {
+		got := parseCPU(test.input)
+		assert.Equal(t, test.expect, got)
+	}
+}
```
jaypipes: :+1: rock on. :)

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -348,7 +351,15 @@ func BuildAWS(opts config.AutoscalingOptions, do cloudprovider.NodeGroupDiscover
 		klog.Fatalf("Failed to create AWS Manager: %v", err)
 	}
 
-	provider, err := BuildAwsCloudProvider(manager, rl)
+	// Generate EC2 list
```
losipiuk: I do not like the idea to fallback to static list of instances if runtime detection does not work. With this we can end up with  unpredicatable behavior of AWS cp if there is some random network/API outage while Cluster-Autoscaler is starting up.
I would suggest adding an program flag to specify if static or dynamic list should be used. And if dynamic is used and API endpoint cannot be queried, return an error from BuildAWS method.

losipiuk: @Jeffwan Are you still working on this one?

Jeffwan:
Sorry for late response. I will clean it up and submit revision

Jeffwan:
> ++ lgtm. Would still prefer a const for the pricing URL, but that's something to do in a follow up.

Move to const in latest change.

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
Jeffwan: Yes. I update the PR and expose an option. Please have a check

Jeffwan:
/hold cancel 

In the latest change, we expose option to users whose network is restricted to call EC2 pricing service.

cluster-autoscaler/cloudprovider/aws/aws_util.go
```diff
@@ -0,0 +1,172 @@
...
+type productAttributes struct {
+	InstanceType string `json:"instanceType"`
+	VCPU         string `json:"vcpu"`
+	Memory       string `json:"memory"`
+	GPU          string `json:"gpu"`
+}
+
+// GenerateEC2InstanceTypes returns a map of ec2 resources
+func GenerateEC2InstanceTypes(region string) (map[string]*InstanceType, error) {
+	instanceTypes := make(map[string]*InstanceType)
```
seh: No size hint?

Jeffwan: @seh  Not sure why I can not find your comments in this page.. I remember you left some comments here. Did you say it's better to have a initial capacity? The list is growing every quarter.

seh: Force pushes to files with comments hide them. I'm not sure why sometimes GitHub will retain them, but hide them in an "obsolete" block, or something to that effect.

In any case, I suggest you size the map to something like 200. Starting with the default size is going to induce reallocation and rehashing, given that we know there are more instance types than the default map size.

Jeffwan: I agree that go internally handles allocations and rehashing may hit performance. Not sure how golang implements it, probably exponential? rehashing for < 300 items may not be a problem. This method is only used during CA startup or the time we manually generate static list. Performance should not be a concern here.

The other reason I don't want to have a number here. If we set something like 200, it will be resized for some regions. Developers maybe also confused by the magic number there, this number is hard to tell from code itself.

seh: That's a fine reason. I just wanted to make sure that this is a choice, and not just a missed opportunity.

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -343,12 +346,36 @@ func BuildAWS(opts config.AutoscalingOptions, do cloudprovider.NodeGroupDiscover
 		defer config.Close()
 	}
 
+	// Generate EC2 list
+	var instanceTypes map[string]*InstanceType
+	if opts.StaticInstanceList {
+		klog.Warning("Use static EC2 Instance Types, list could be outdated")
+		instanceTypes = GetStaticEC2InstanceTypes()
```
losipiuk: Possibly this method can also return the date when the list was last updated. Then you could add it to warning message to give clear singal how up-to-date data is.

cluster-autoscaler/cloudprovider/aws/aws_util_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+func TestGetCurrentAwsRegion(t *testing.T) {
+	region := "us-west-2"
+	os.Unsetenv("AWS_REGION")
```
losipiuk: To be more correct please change to:
```go
if oldRegion, found := os.LookupEnv("AWS_REGION"); found {
  os.Unsetenv("AWS_REGION")
  defer os.Setenv("AWS_REGION", oldRegion)
}
```

cluster-autoscaler/cloudprovider/aws/aws_util_test.go
```diff
@@ -0,0 +1,105 @@
...
+	result, err := GetCurrentAwsRegion()
+
+	assert.Nil(t, err)
+	assert.NotNil(t, result)
+	assert.Equal(t, region, result)
+}
+
+func TestGetCurrentAwsRegionWithRegionEnv(t *testing.T) {
+	region := "us-west-2"
+	os.Setenv("AWS_REGION", region)
```
losipiuk: To be more correct please change to:
```go
if oldRegion, found := os.LookupEnv("AWS_REGION"); found {
  defer os.Setenv("AWS_REGION", oldRegion)
} else {
  defer os.Unsetenv("AWS_REGION")
}
os.Setenv("AWS_REGION", region)
```

Possibly not needed if we have just above two places. Yet you may consider writing a short wrapper with signature:
```go
func withEnv(variable string, variableValue string, f func())
```

cluster-autoscaler/FAQ.md
```diff
@@ -674,6 +674,7 @@ The following startup parameters are supported for cluster autoscaler:
 | `leader-elect-renew-deadline` | The interval between attempts by the acting master to renew a leadership slot before it stops leading.<br>This must be less than or equal to the lease duration.<br>This is only applicable if leader election is enabled | 10 seconds
 | `leader-elect-retry-period` | The duration the clients should wait between attempting acquisition and renewal of a leadership.<br>This is only applicable if leader election is enabled | 2 seconds
 | `leader-elect-resource-lock` | The type of resource object that is used for locking during leader election.<br>Supported options are `endpoints` (default) and `configmaps` | "endpoints"
+| `static-instance-list` | Should CA fetch instance types in runtime or use a static list. AWS only | false
```
losipiuk: please rename to "aws-use-static-instance-list". Also rename variable to pass the value accordingly.

cluster-autoscaler/cloudprovider/aws/aws_util_test.go
Jeffwan: Great! This is precise. I make the changes. I didn't see any other place test env here. I will do refactor if I see more tests with similar setups.

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
Jeffwan: This is great suggestion.
I create a const and add constructions in the doc that everytime the list is updated, people need to update date correspondingly.

Jeffwan: W1014 17:25:22.612552   75153 aws_cloud_provider.go:354] Use static EC2 Instance Types and list could be outdated. Last update time: 2019-10-14
I1014 17:25:22.612701   75153 reflector.go:120] Starting reflector *v1.PersistentVolumeClaim (0s) from k8s.io/client-go/informers/factory.go:134

losipiuk:
/lgtm
/approve
Thanks!

cablespaghetti:
Any chance of this making it over to the 1.14-1.16 branches? I've got a couple of clusters I want to build on t5a instances which aren't in the static list.

jaypipes:
@Jeffwan can you look into porting to the older branches please?

Jeffwan:
@jaypipes I mark this as a backport candidate in next release.

PR #2255, opened 2019-08-12
Provide more info to ScaleUpStatusProcessor

towca:
Add info about considered and created nodegroups to
ScaleUpStatusProcessor

### DIFF
--- a/cluster-autoscaler/processors/status/scale_up_status_processor.go
+++ b/cluster-autoscaler/processors/status/scale_up_status_processor.go
@@ -18,7 +18,10 @@ package status
 
 import (
 	apiv1 "k8s.io/api/core/v1"
+
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/processors/status/scale_up_status_processor.go
```diff
@@ -31,6 +34,8 @@ type ScaleUpStatus struct {
 	PodsTriggeredScaleUp    []*apiv1.Pod
 	PodsRemainUnschedulable []NoScaleUpInfo
 	PodsAwaitEvaluation     []*apiv1.Pod
+	CreateNodeGroupResults  []nodegroups.CreateNodeGroupResult
+	ConsideredNodeGroups    map[string]cloudprovider.NodeGroup
```
aleksandra-malinowska: nit: why not []cloudprovider.NodeGroup? Seems like it would be simpler

towca: No preference here, changed

PR #2259, opened 2019-08-13
Provide ScaleUpStatusProcessor with info about all rejected node groups

towca:
Previously, it had info only about the ones that actually exist.

The changes to the eventing processor are done to keep its previous
behavior the same.


cluster-autoscaler/core/scale_up.go
```diff
@@ -400,7 +400,7 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 
 		for pod, err := range podsNotPassing {
 			_, found := podsRemainUnschedulable[pod]
-			if found && nodeGroup.Exist() {
+			if found {
```
MaciekPytel: The comment below no longer makes sense with your change.

towca: Right, deleted!

PR #1613, opened 2019-01-25
Some configurable parameters

safanaj:
To use VPA on AWS EKS I needed to change the port for the admission controller webhook, so I would propose to have a flag for it.

For the recommender to be able to get historical metrics from prometheus I need to tune the metrics and labels to query prometheus properly, so I would add a bunch of flags for that.

This PR is just to share the ideas.

Issue #1551, opened 2019-01-03
vertiacal-pod-autoscaler: how to configure for integration with prometheus?


piontec:
I'm running VPA 0.3.0 on k8s 1.10.11. I'm trying to find some docs that explain how VPA fetches metrics' history from prometheus. Which metrics and labels it expects to be present in prometheus? Currently, no matter if I give it a correct or incorrect prometheus URL in `--prometheus-address`, nothing is logged about prometheus queries or connections, even with `--v=8`.

bskiba:
I think you also need --storage=prometheus
Otherwise this defaults to reading from checkpoints on startup.
https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/recommender/main.go#L37

piontec:
Thanks for the tip! Does it mean that VPA uses either CRD checkpoints or it stores checkpoints in prometheus (and reads them back from there), but not both at the same time? Does it read any custom metrics from prom or uses it only as a storage for data it got previously from metrics-server?

bskiba:
VPA uses checkpoints or history from Prometheus to initialise on startup. With checkpoints, it writes them periodically to etcd. With Prometheus it doesn't store any metrics, but assumes Proemetheus is already set up to gather pod metrics from the cluster. It reads `container_cpu_usage_seconds_total` and `container_memory_usage_bytes`.
You can take a look at [this file](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go) to see how it is done in more detail.

bskiba:
Any luck? If you managed to set this up successfully, it would be extremely helpful if you wrote a short note on how you set it up that we could put in the docs.

wyb1:
I setup the VPA to use prometheus. I set `--storage=prometheus` and `--prometheus-address=http://prometheus.default.svc.cluster.local:9090` Looking at the logs it looks like the vpa-recommender is using prometheus.
@bskiba is there any reason you are using `container_memory_usage_bytes` over `container_memory_working_set_bytes`? I found that `container_memory_working_set_bytes` delivers more accurate results regarding memory usage.

bskiba:
@wyb1 This is good to know, would you maybe have time to add a short note on how to set this up in the [FAQ](https://github.com/kubernetes/autoscaler/vertical-pod-autoscaler/faq.md)? It would be very helpful to other users!

[...]

PR #2266, opened 2019-08-19
VPA Recommender: Use WSS from Prometheus provider

realdimas:
Currently, VPA Recommender uses `container_memory_usage_bytes` cAdvisor metric from Prometheus history provider.
This metric includes page cache into container memory usage total.
As a result, cache-heavy workloads keep getting larger and larger memory recommendation.

This change takes memory held for page cache out of VPA's calculation removing the primary driver behind memory over-requesting.

A similar suggestion was mentioned twice ([[1]](https://github.com/kubernetes/autoscaler/issues/1551#issuecomment-456386360), [[2]](https://github.com/kubernetes/autoscaler/pull/1613#issuecomment-458918134)), but no PRs / issues were raised to discuss it.

This change harmonizes the metric which VPA Recommender Prometheus history provider uses with that of [k8s-prometheus-adapter](https://github.com/DirectXMan12/k8s-prometheus-adapter/pull/106/commits/c5801455ecdbecec363bb2d87927172690a7d983#diff-a24e018d491475644c4bb928d27db3c9R103).

Illustration:
thanos-store-wss-vs-total

On the screenshot above you see an example process which would use significant memory of memory for the page cache.

Since VPA keep taking into account page cache into memory requests, a container would be able to use more and more memory for it without experiencing memory pressure.

This creates a condition for over-requesting if not runaway(-ish) memory adjustments by VPA.

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
@@ -194,7 +194,7 @@ func (p *prometheusHistoryProvider) GetClusterHistory() (map[model.PodID]*PodHis
 	if err != nil {
 		return nil, fmt.Errorf("cannot get usage history: %v", err)
 	}
-	err = p.readResourceHistory(res, fmt.Sprintf("container_memory_usage_bytes{%s}[%s]", podSelector, p.config.HistoryLength), model.ResourceMemory)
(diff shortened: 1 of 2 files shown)


bskiba:
Sounds good. Thanks for providing the reference to prometheus metrics adapter!
/lgtm
/approve

PR #2274, opened 2019-08-22
reworked dependencies / go.mod

pdbogen:
howdy!

I found it very frustrating that I couldn't just `go build` this, because of things like hardcoded paths in go.mod.

(It also isn't clear that we _need_ to vendor anything... and `godo` was being vendored/pinned in a fairly hacky way...)

This PR does the following:
1. remove all vendoring, including weird vendoring of godo
2. clean up `/tmp/abc` paths from go.mod
3. add a script `hack/pin-dependency.sh`, heavily modified from https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-521493597 to manage the go mod replacements. as committed, they're pinned to 1.15.3.

Totally understand if the PR is not desirable; but if so, I'd love if you could provide instructions for how potential contributors are expected to build and run this code.

After these changes, tests (`go test ./...`) pass, and building within a clean `golang:1` docker container work perfectly.


pdbogen:
Sorry for the churn on those last few commits. travis-ci is passing now; I needed to update some dependencies and put `GO111MODULE` into `on` mode when running tests for `go.mod`, since within travis the code is unpacked inside GOPATH.

losipiuk:
Thanks @pdbogen!

There is another PR which addresses the fact that `go.mod` is unresolvable (#2286). See my comment https://github.com/kubernetes/autoscaler/pull/2286#issuecomment-527242023 why I believe `update-vendor.sh` is needed (or at least beneficial). 
It feels like resolvability of `go.mod` can be achieved with minor tweaks to `update-vendor.sh`, without sacrificing possibility to frequently sync dependencies of `k8s.io/autoscaling/cluster-autoscaler` and `k8s.io/kubernetes`.

Also I see that your PR you uses separate versions for all the modules which are based on `k8s.io/kubernetes/staging`. That approach works only for point releases of `k8s`. It is nice to have to be able to revendor to unreleased version of k8s (some random commit in k8s.io/kubernetes). Therefore I prefer the solution from #2286 when we use `replace` directive with subdirectories of `k8s.io/kubernetes/staging/`, all pinned to single version.

PR #1910, opened 2019-04-17
Validate that targetRef if set is a top level controller

jbartosik:
ControllerFetcher and its tests written by @schylek.

PR #2275, opened 2019-08-22
VPA: find topmost well-known or scalable controller

### LINKED ISSUE #2159: vertical-pod-autoscaler recommender stopped working for some targets
vpa-recommender stopped working for some targetRefs between vertical-pod-autoscaler-0.5.0 and vertical-pod-autoscaler-0.5.1.

Specifically, none of the [prometheus-operator](https://github.com/coreos/prometheus-operator) created [StatefulSets](https://github.com/coreos/prometheus-operator/blob/409ec7dc5dd6b88d33e67ba5d19ff1d6169e56b0/pkg/prometheus/statefulset.go#L162-L179) get any recommendation.

TargetRef:
```
targetRef:
  apiVersion: apps/v1
  kind: StatefulSet
  name: prometheus-kube-system
```

gets following  vpa-recommender v0.5.1 messages on the `.status.conditions`:
```
- lastTransitionTime: "2019-07-03T11:30:29Z"
  message: 'Error checking if target is a top level controller: Unhandled targetRef
    monitoring.coreos.com/v1 / Prometheus / kube-system, last error the server could
    not find the requested resource (get prometheuses kube-system)'
  status: "True"
  type: ConfigUnsupported
- lastTransitionTime: "2019-07-03T11:30:35Z"
  message: No pods match this VPA object
  reason: NoPodsMatched
  status: "False"
  type: RecommendationProvided
```

(it worked as expected in v0.5.0).

As an experiment, I tried specifying a targetRef against top-level controller's CRD
```
targetRef:
  apiVersion: monitoring.coreos.com/v1
  kind: Prometheus
  name: kube-system
```

which made vpa-recommender produce this error:
```
I0703 12:03:42.380087       1 request.go:485] Throttling request took 197.788691ms, request: GET:https://100.64.0.1:443/apis/monitoring.coreos.com/v1/namespaces/monitoring/prometheuses/kube-system/scale
E0703 12:03:42.380999       1 cluster_feeder.go:463] Cannot get target selector from VPA's targetRef. Reason: Unhandled targetRef monitoring.coreos.com/v1 / Prometheus / kube-system, last error the server could not find the requested resource (get prometheuses kube-system)
```

As far as I can tell `targetRef` validation logic added in #1910 does not play well with 3rd-party controllers managing 1st-party controllers objects via `ownerReference`.

realdimas:
Not all CRDs implement scale sub resource.
Previous logic lead to VPA failing with top-level resource which did not implement scale subresource.
This change ensures that we are walking up the controllers' chain to the last valid (i. e. well-known or scalable) controller.

Resolves #2159.

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
@@ -24,6 +24,7 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -69,10 +70,10 @@ type ControllerKeyWithAPIVersion struct {
 	ApiVersion string
 }
 
-// ControllerFetcher is responsible for finding the top level controller
+// ControllerFetcher is responsible for finding the topmost well-known or scalable controller
 type ControllerFetcher interface {
-	// FindTopLevel returns top level controller. Error is returned if top level controller cannot be found.
-	FindTopLevel(controller *ControllerKeyWithAPIVersion) (*ControllerKeyWithAPIVersion, error)
+	// FindTopMostWellKnownOrScalable returns topmost well-known or scalable controller. Error is returned if controller cannot be found.
+	FindTopMostWellKnownOrScalable(controller *ControllerKeyWithAPIVersion) (*ControllerKeyWithAPIVersion, error)
 }
 
 type controllerFetcher struct {
@@ -126,7 +127,7 @@ func NewControllerFetcher(config *rest.Config, kubeClient kube_client.Interface,
 
 func getOwnerController(owners []metav1.OwnerReference, namespace string) *ControllerKeyWithAPIVersion {
 	for _, owner := range owners {
-		if owner.Controller != nil && *owner.Controller == true {
+		if owner.Controller != nil && *owner.Controller {
 			return &ControllerKeyWithAPIVersion{
 				ControllerKey: ControllerKey{
 					Namespace: namespace,
@@ -207,23 +208,70 @@ func (f *controllerFetcher) getParentOfController(controllerKey ControllerKeyWit
 		return getParentOfWellKnownController(informer, controllerKey)
 	}
 
-	// TODO: cache response
-	groupVersion, err := schema.ParseGroupVersion(controllerKey.ApiVersion)
+	groupKind, err := controllerKey.groupKind()
 	if err != nil {
 		return nil, err
 	}
-	groupKind := schema.GroupKind{
-		Group: groupVersion.Group,
-		Kind:  controllerKey.Kind,
-	}
 
 	owner, err := f.getOwnerForScaleResource(groupKind, controllerKey.Namespace, controllerKey.Name)
+	if apierrors.IsNotFound(err) {
+		return nil, nil
+	}
 	if err != nil {
 		return nil, fmt.Errorf("Unhandled targetRef %s / %s / %s, last error %v",
 			controllerKey.ApiVersion, controllerKey.Kind, controllerKey.Name, err)
 	}
 
 	return owner, nil
+}
+
+func (c *ControllerKeyWithAPIVersion) groupKind() (schema.GroupKind, error) {
+	// TODO: cache response
+	groupVersion, err := schema.ParseGroupVersion(c.ApiVersion)
+	if err != nil {
+		return schema.GroupKind{}, err
+	}
+
+	groupKind := schema.GroupKind{
+		Group: groupVersion.Group,
+		Kind:  c.ControllerKey.Kind,
+	}
+
+	return groupKind, nil
+}
+
+func (f *controllerFetcher) isWellKnown(key *ControllerKeyWithAPIVersion) bool {
+	kind := wellKnownController(key.ControllerKey.Kind)
+	_, exists := f.informersMap[kind]
+	return exists
+}
+
+func (f *controllerFetcher) isWellKnownOrScalable(key *ControllerKeyWithAPIVersion) bool {
+	if f.isWellKnown(key) {
+		return true
+	}
+
+	//if not well known check if it supports scaling
+	groupKind, err := key.groupKind()
+	if err != nil {
+		klog.Errorf("Could not find groupKind for %s/%s: %v", key.Namespace, key.Name, err)
+		return false
+	}
+
+	mappings, err := f.mapper.RESTMappings(groupKind)
+	if err != nil {
+		klog.Errorf("Could not find mappings for %s: %v", groupKind, err)
+		return false
+	}
+
+	for _, mapping := range mappings {
+		groupResource := mapping.Resource.GroupResource()
+		scale, err := f.scaleNamespacer.Scales(key.Namespace).Get(groupResource, key.Name)
+		if err == nil && scale != nil {
+			return true
+		}
+	}
+	return false
 }
 
 func (f *controllerFetcher) getOwnerForScaleResource(groupKind schema.GroupKind, namespace, name string) (*ControllerKeyWithAPIVersion, error) {
@@ -246,10 +294,18 @@ func (f *controllerFetcher) getOwnerForScaleResource(groupKind schema.GroupKind,
 	return nil, lastError
 }
 
-func (f *controllerFetcher) FindTopLevel(key *ControllerKeyWithAPIVersion) (*ControllerKeyWithAPIVersion, error) {
+func (f *controllerFetcher) FindTopMostWellKnownOrScalable(key *ControllerKeyWithAPIVersion) (*ControllerKeyWithAPIVersion, error) {
 	if key == nil {
 		return nil, nil
 	}
+
+	var topMostWellKnownOrScalable *ControllerKeyWithAPIVersion
+
+	wellKnownOrScalable := f.isWellKnownOrScalable(key)
+	if wellKnownOrScalable {
+		topMostWellKnownOrScalable = key
+	}
+
 	visited := make(map[ControllerKeyWithAPIVersion]bool)
 	visited[*key] = true
 	for {
@@ -257,14 +313,22 @@ func (f *controllerFetcher) FindTopLevel(key *ControllerKeyWithAPIVersion) (*Con
 		if err != nil {
 			return nil, err
 		}
+
 		if owner == nil {
-			return key, nil
+			return topMostWellKnownOrScalable, nil
 		}
+
+		wellKnownOrScalable = f.isWellKnownOrScalable(owner)
+		if wellKnownOrScalable {
+			topMostWellKnownOrScalable = owner
+		}
+
 		_, alreadyVisited := visited[*owner]
 		if alreadyVisited {
 			return nil, fmt.Errorf("Cycle detected in ownership chain")
 		}
 		visited[*key] = true
+
 		key = owner
 	}
 }
@@ -272,7 +336,7 @@ func (f *controllerFetcher) FindTopLevel(key *ControllerKeyWithAPIVersion) (*Con
 type identityControllerFetcher struct {
 }
 
-func (f *identityControllerFetcher) FindTopLevel(controller *ControllerKeyWithAPIVersion) (*ControllerKeyWithAPIVersion, error) {
+func (f *identityControllerFetcher) FindTopMostWellKnownOrScalable(controller *ControllerKeyWithAPIVersion) (*ControllerKeyWithAPIVersion, error) {
 	return controller, nil
 }
 
@@ -280,7 +344,7 @@ type constControllerFetcher struct {
 	ControllerKeyWithAPIVersion *ControllerKeyWithAPIVersion
 }
 
-func (f *constControllerFetcher) FindTopLevel(controller *ControllerKeyWithAPIVersion) (*ControllerKeyWithAPIVersion, error) {
+func (f *constControllerFetcher) FindTopMostWellKnownOrScalable(controller *ControllerKeyWithAPIVersion) (*ControllerKeyWithAPIVersion, error) {
 	return f.ControllerKeyWithAPIVersion, nil
 }
 
@@ -289,7 +353,7 @@ type mockControllerFetcher struct {
 	result   *ControllerKeyWithAPIVersion
 }
 
-func (f *mockControllerFetcher) FindTopLevel(controller *ControllerKeyWithAPIVersion) (*ControllerKeyWithAPIVersion, error) {
+func (f *mockControllerFetcher) FindTopMostWellKnownOrScalable(controller *ControllerKeyWithAPIVersion) (*ControllerKeyWithAPIVersion, error) {
 	if controller == nil && f.expected == nil {
 		return f.result, nil
 	}
--- a/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
@@ -24,11 +24,17 @@ import (
 	"github.com/stretchr/testify/assert"
 
 	appsv1 "k8s.io/api/apps/v1"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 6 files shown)


realdimas:
`s/scaleable/scalable/`

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -207,25 +208,87 @@ func (f *controllerFetcher) getParentOfController(controllerKey ControllerKeyWit
...
+
+	groupKind := schema.GroupKind{
+		Group: groupVersion.Group,
+		Kind:  c.ControllerKey.Kind,
+	}
+
+	return groupKind, nil
+}
+
+func (f *controllerFetcher) isWellKnown(key *ControllerKeyWithAPIVersion) (bool, error) {
```
jbartosik: This never returns an error. Please return only bool.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -207,25 +208,87 @@ func (f *controllerFetcher) getParentOfController(controllerKey ControllerKeyWit
...
+	}
+	var lastError error
+	for _, mapping := range mappings {
+		groupResource := mapping.Resource.GroupResource()
+		_, err := f.scaleNamespacer.Scales(key.Namespace).Get(groupResource, key.Name)
+		if err == nil {
+			return true, nil
+		}
+		if !apierrors.IsNotFound(err) {
+			// Suppress errors due to missing subresource
```
jbartosik: Why do we want to suppress those errors?

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -207,25 +208,87 @@ func (f *controllerFetcher) getParentOfController(controllerKey ControllerKeyWit
...
+	groupKind, err := key.groupKind()
+	if err != nil {
+		return false, err
+	}
+
+	mappings, err := f.mapper.RESTMappings(groupKind)
+	if err != nil {
+		return false, err
+	}
+	var lastError error
```
jbartosik: Could we instead accumulate those errors and build new error with message like "errors checking if resurce is scalable. [list or resources with errors]"

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
realdimas: Fixed.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
realdimas: Expanded the comment to better explain the reason.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
realdimas: Not sure, wouldn't the resource be all the same here? 🤷‍♂
We are iterating over (possibly) multiple RESTMappings here, its too theoretical for me to make ajudgmentt call.

realdimas:
> @jbartosik: I'm not sure what I think about changing wording from "top-level" to "top most"
I'm not attached to "topmost", but "top-level controller" would no longer be correct given the new flow [proposed by @bskiba](../issues/2159#issuecomment-508217621).

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -207,25 +208,85 @@ func (f *controllerFetcher) getParentOfController(controllerKey ControllerKeyWit
...
+			// method should return false and conceal the IsNotFound error instead
+			lastError = err
+		}
+	}
+
+	return false, lastError
+}
+
+func (f *controllerFetcher) isWellKnownOrScalable(key *ControllerKeyWithAPIVersion) (bool, error) {
+	wellKnown := f.isWellKnown(key)
```
bskiba: if f.isWellKnown(key) {
 return true, nil
}

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -207,25 +208,85 @@ func (f *controllerFetcher) getParentOfController(controllerKey ControllerKeyWit
...
-	groupKind := schema.GroupKind{
-		Group: groupVersion.Group,
-		Kind:  controllerKey.Kind,
-	}
 
 	owner, err := f.getOwnerForScaleResource(groupKind, controllerKey.Namespace, controllerKey.Name)
-	if err != nil {
+	if apierrors.IsNotFound(err) {
+		return nil, nil
+	} else if err != nil {
```
bskiba: You can drop the else here.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
bskiba: I think it's fine to leave lastError

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -207,25 +208,85 @@ func (f *controllerFetcher) getParentOfController(controllerKey ControllerKeyWit
...
+	}
+
+	mappings, err := f.mapper.RESTMappings(groupKind)
+	if err != nil {
+		return false, err
+	}
+	var lastError error
+	for _, mapping := range mappings {
+		groupResource := mapping.Resource.GroupResource()
+		_, err := f.scaleNamespacer.Scales(key.Namespace).Get(groupResource, key.Name)
```
bskiba: This code is very similar to getOwnerForScaleResource, and we are performing a Get call to the API Server here if I'm reading this right and that is expensive. We should make sure that for each controllerKey we only perform this once while traversing the owner ref chain (Currenlty it may be twice, to check the owner and then check isScalable).

Joseph-Irving: @bskiba, so we've changed the code to only call this once, I've done this by setting a `scalable` boolean field on the controllers so we know if they implement the scale sub-resource and can re-use the information. Let me know if you think this is better.

realdimas:
> @bskiba: I'm missing some test cases to exercise the new code

While #1910 did add [nice controllers testing code](https://github.com/kubernetes/autoscaler/pull/1910/files#diff-915e734f7c0f030eff6fb340147b1942) for apps/v1, no provisions were made for scale subresource and/or CRDs.
Any tips & tricks on extending support for dynamic go-client? We started with `unstructured.Unstructured` for CRDs but then tests panic on https://github.com/jbartosik/autoscaler/blob/d1c90cb8848a3d9774007d565956d48b4cceb7b8/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go#L48-L51

bskiba:
Sorry for the slow replies, I really appreciate the good work here.

Minimal test case I see is Deployment (or any other controller) controlled by another object, probably one that we are not authorized to query. (sort of what happened with the Prometheus operator)
So what if you do sth like 
fun maybeAddObject(controller *controllerFetcher, obj runtime.Object) {
	kind := wellKnownController(obj.GetObjectKind().GroupVersionKind().Kind)
	store, ok := controller.informersMap[kind].GetStore(); ok {
             store.Add(obj)
        }
}

For scale subresource I don't have good tips OTOH, though let me take a look at HorizontalPodAutoscaler tests to see if there is material there.

bskiba:
https://github.com/kubernetes/kubernetes/blob/c7c89f8c6140e0639dc8ba2d8dcdd6fdbd5409ae/pkg/controller/podautoscaler/horizontal_test.go#L373
has setup for a fake scale client that could be useful here.

Joseph-Irving:
We've been trying to write some tests for this but it's proving to be quite tricky, maybe we're missing something obvious. But when you have a controller that isn't well known you need to fake the `controllerFetcher`'s `scaleNamespacer` and `mapper` as it attempts to query the api to get info about the resource in https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go#L229. 

So this needs us to fake a restmapper and some api groups/versions, do you know if there are any example of doing this? 

I've been trying with just creating some structs
```go
versioned := map[string][]metav1.APIResource{
		"Foo": []metav1.APIResource{metav1.APIResource{Kind: "Foo"}},
	}
	fake := []*restmapper.APIGroupResources{
		&restmapper.APIGroupResources{
			Group: metav1.APIGroup{
				Name:     "Foo",
				Versions: []metav1.GroupVersionForDiscovery{metav1.GroupVersionForDiscovery{GroupVersion: "Foo", Version: "Foo"}},
			},
			VersionedResources: versioned,
		},
	}
	mapper := restmapper.NewDiscoveryRESTMapper(fake)
```
but it's not really working I get errors like
```
Expected nil, but got: &errors.errorString{s:"Unhandled targetRef Foo/Foo / Foo / bah, last error unable to get client for .Foo: unable to get full preferred group-version-resource for .Foo: a resource must be present, got: Foo/, Resource="}
```
I would greatly appreciate any nudges in the right direction as I'm not convinced I'm going about this in the right way

bskiba:
Does it help if you specify Name, Group and Version as well in the APIResource?
Like this:

```
versioned := map[string][]metav1.APIResource{
		"Foo": []metav1.APIResource{metav1.APIResource{Name: "bah", Kind: "Foo", Version:"Foo", Group: "Foo"}}, // IIUC what resource you are trying to add
	}
```

If not, can you upload the WIP test that you have so I can play around with it and see if I can get it to work?

Joseph-Irving:
Thanks that suggestion got me a bit further, I get a panic in `k8s.io/client-go/scale/client.go:132` now. It appears the baseClient that I create using the fake clientset is equal to nil and I'm not entirely sure why. 
```go
clientSet := fake.NewSimpleClientset()
baseClient := clientSet.CoreV1().RESTClient() // THIS EQUALS NIL
resolver := scale.NewDiscoveryScaleKindResolver(clientSet.Discovery())
scaleNamespacer := scale.New(baseClient, mapper, dynamic.LegacyAPIPathResolverFunc, resolver)
f.scaleNamespacer = scaleNamespacer
```
I've pushed what I've done so far so you can have a look

bskiba:
I'm just getting back to this. Appreciate your patience!

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
```diff
@@ -38,6 +42,26 @@ var trueVar = true
...
+		},
+	}
+	mapper := restmapper.NewDiscoveryRESTMapper(fakeMapper)
+	f.mapper = mapper
+
+	clientSet := fake.NewSimpleClientset()
+	baseClient := clientSet.CoreV1().RESTClient()
+	resolver := scale.NewDiscoveryScaleKindResolver(clientSet.Discovery())
+	scaleNamespacer := scale.New(baseClient, mapper, dynamic.LegacyAPIPathResolverFunc, resolver)
+	f.scaleNamespacer = scaleNamespacer
```
bskiba: So it looks like RESTClient() of the fake ClientSet is nil (guess this is part of it being fake :) )

How about trying this instead:

```
scaleNamespacer := &scalefake.FakeScaleClient{}
f.scaleNamespacer = scaleNamespacer
```

This needs and import of 
scalefake "k8s.io/client-go/scale/fake" 
which is not in vendor, but if you can just copy it into vendor for now, I'll resolve the dependencies for you when we get this to work.

Joseph-Irving:
Hey @bskiba, sorry I've been away on holiday. 
I'm back now, and following your advice I managed to get the test to work. 
So the test has a Deployment with a made up owner resource (to simulate being owned by an arbitrary CRD), then fake owner returns 'not found' when it attempts to find a scale subresource on it, so the Deployment is returned as the Controller to act on. Let me know what you think!
(I cped the fake client to vendor for now)

Joseph-Irving:
@bskiba sorry to pester, do you think you might have some time to take a look at this?

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
```diff
@@ -239,17 +272,42 @@ func TestControllerFetcher(t *testing.T) {
...
+							APIVersion: "Foo/Foo",
+							Controller: &trueVar,
+							Kind:       "Foo",
+							Name:       "bah",
+						},
+					},
+				},
+			}},
+			expectedKey: &ControllerKeyWithAPIVersion{ControllerKey: ControllerKey{
+				Name: "test-deployment", Kind: "Deployment", Namespace: "test-namesapce"}}, // No known parent or a parent that can scale so should return itself
```
bskiba: I understand this case as "Parent does not support scale subresource so should return itself"

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
```diff
@@ -239,17 +272,42 @@ func TestControllerFetcher(t *testing.T) {
...
+			key: &ControllerKeyWithAPIVersion{ControllerKey: ControllerKey{
+				Name: "test-deployment", Kind: "Deployment", Namespace: "test-namesapce"}},
+			objects: []runtime.Object{&appsv1.Deployment{
+				TypeMeta: metav1.TypeMeta{
+					Kind: "Deployment",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-deployment",
+					Namespace: "test-namesapce",
+					// UnkownParent
```
bskiba: The parent seems to be known but not scalable?

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
```diff
@@ -239,17 +272,42 @@ func TestControllerFetcher(t *testing.T) {
 			expectedKey:   nil,
 			expectedError: fmt.Errorf("Cycle detected in ownership chain"),
 		},
+		{
```
bskiba: It should be possible to add a similar case where the parent actually supports scale subresource, WDYT?

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
```diff
@@ -27,8 +27,14 @@ import (
...
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/restmapper"
+	scalefake "k8s.io/client-go/scale/fake"
+	core "k8s.io/client-go/testing"
+
```
bskiba: nit: Remove blank line

bskiba:
Couple more comments (many apologies for taking so long).

In general this looks close to being ready.
We moved VPA to modules so running make build and go mod vendor should sort out dependencies for you.
Could you then squash the commits to have one commit with the logic and tests and one with vendor/modules changes?

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
Joseph-Irving: Maybe "not well-known" is better wording, I meant it's not in the `informersMap` map of `wellKnownController`. E.g not a Deployment, ReplicaSet, StatefulSet etc.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
Joseph-Irving: yeah I shall try that, I wanted to see if you thought this seemed sane before I attempted that :)

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
bskiba: Oh yeah, that makes perfect sense :+1:

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
bskiba: Thanks! Looks good

Joseph-Irving:
Ok, I've added a test for an owner that does support the scale subresource, it exposed a bit of a logic flaw in the code, previously if the resource was wellKnown (e.g a Deployment) it wouldn't actually check if the resource's owner could scale and therefore would always return the original resource. So I reworked it so that it will always check the owner can scale or is well known, regardless of what kind of resource it owns.

I've done the go mod changes, and squashed commits

Joseph-Irving:
@bskiba, gentle ping to look at this if you have a chance, would be much appreciated!

bskiba:
Apologies for the delay again. I'll give this a look today.

bskiba:
Thanks a lot for your effort on this and bearing with my review latency!
/lgtm
/approve

PR #2277, opened 2019-08-25
metrics: expose if its can scale down

JulienBalestra:
Expose through a prometheus gauge if the cluster autoscaler can scale down.

In very large clusters, it become hard to get a scale down window as the cluster autoscaler often scales-up.
The metrics allows to pull out some data about how often it can do this task. 

This is going to help to tune any flags like `--scale-down-delay-*`


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -437,6 +437,7 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 		// In dry run only utilization is updated
 		calculateUnneededOnly := scaleDownInCooldown || scaleDown.nodeDeletionTracker.IsNonEmptyNodeDeleteInProgress()
 
+		metrics.UpdateScalingDownStatus(!a.processorCallbacks.disableScaleDownForLoop)
```
losipiuk: It seems what you want to capture is the `!scaleDownInCooldown`

cluster-autoscaler/metrics/metrics.go
```diff
@@ -207,6 +207,14 @@ var (
 		},
 	)
 
+	scaleDownEnabled = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: caNamespace,
+			Name:      "scale_down_enabled",
```
losipiuk: It seems `scale_down_in_cooldown` is better name for the metric.

cluster-autoscaler/metrics/metrics.go
```diff
@@ -360,3 +369,12 @@ func RegisterNodeGroupCreation() {
 func RegisterNodeGroupDeletion() {
 	nodeGroupDeletionCount.Add(1.0)
 }
+
+// UpdateScalingDownStatus registers if the cluster autoscaler can scale down
+func UpdateScalingDownStatus(canScaleDown bool) {
```
losipiuk: name parameter `scaleDownInCooldown`

cluster-autoscaler/metrics/metrics.go
```diff
@@ -207,6 +207,14 @@ var (
 		},
 	)
 
+	scaleDownEnabled = prometheus.NewGauge(
```
losipiuk: name variable `scaleDownInCooldown`

losipiuk:
@MaciekPytel: you have better intuition when it comes to metrics. Does this one make sense to  you?

cluster-autoscaler/core/static_autoscaler.go
JulienBalestra: @losipiuk what do you think if I set this metric to `false` in these 2 codeblocks: [here](https://github.com/kubernetes/autoscaler/blob/8d9010e11eb01f0d8cb2f58ec53239240802eef7/cluster-autoscaler/core/static_autoscaler.go#L446) and [here](https://github.com/kubernetes/autoscaler/blob/8d9010e11eb01f0d8cb2f58ec53239240802eef7/cluster-autoscaler/core/static_autoscaler.go#L448)

And I set it to `true` [here](https://github.com/kubernetes/autoscaler/blob/8d9010e11eb01f0d8cb2f58ec53239240802eef7/cluster-autoscaler/core/static_autoscaler.go#L450)

JulienBalestra: @losipiuk I've just updated the PR to propose this change 👍

cluster-autoscaler/metrics/metrics.go
MaciekPytel: Agreed, the current one makes it look like it just exposes the value of --scale-down-enabled flag or equivalent config.

MaciekPytel:
I think this metric makes a lot of sense. I do agree with @losipiuk comments above, but once they're addressed this looks good to me.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -443,10 +443,13 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 			a.processorCallbacks.disableScaleDownForLoop, scaleDown.nodeDeletionTracker.IsNonEmptyNodeDeleteInProgress())
 
 		if scaleDownInCooldown {
+			metrics.UpdateScalingDownStatus(false)
 			scaleDownStatus.Result = status.ScaleDownInCooldown
 		} else if scaleDown.nodeDeletionTracker.IsNonEmptyNodeDeleteInProgress() {
+			metrics.UpdateScalingDownStatus(false)
```
losipiuk: I guess this is fine though no 100% convinced. Maybe we should have non-binary metric (`scale_down_status`)
- 0: scale_down_possbile
- 1: scale_down_in_cooldown
- 2: scale_down_non_empty_node_in_progress

???

losipiuk: @MaciekPytel pointed out that having different states coded in value of the metric does not play well (at all) with aggregations on metric. So lets keep you original idea.

JulienBalestra: @losipiuk awesome, I just made the requested changes, let me know what do you think 👍

losipiuk:
/lgtm
/approve
Thanks!

PR #2281, opened 2019-08-27
cluster-autoscaler: blocked if an instance is detached from MIG

JulienBalestra:
When an instance is detached from a managed instance group in GCP, the cluster-autoscaler return fast [here](https://github.com/kubernetes/autoscaler/blob/5a5845737328f03b4c1353988aded36fa17cc2ea/cluster-autoscaler/core/static_autoscaler.go#L219) and increase the metric `cluster_autoscaler_errors_total{type:cloudprovidererror}`.

I had some trouble to troubleshoot what caused this issue: I had to patch it to add new logs otherwise it was not possible to understand why the cluster autoscaler was stuck.

Terminating the instance from GCP fixed the issue and unstuck the cluster autoscaler.

As we heavily rely on this component, the current cluster autoscaler behaviour could cause us some issues because it prevents the entire cluster from scaling up/down.

The instance detach use case could happen again for troubleshoot/forensic purpose.

This PR has 2 things:
* add more logs around the error management like it is in other places
* continue on error when processing a node: like when the node is nil.

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/cache.go
+++ b/cluster-autoscaler/cloudprovider/gce/cache.go
@@ -58,13 +58,14 @@ type GceCache struct {
 	cacheMutex sync.Mutex
 
 	// Cache content.
-	migs                   map[GceRef]Mig
-	instanceRefToMigRef    map[GceRef]GceRef
-	resourceLimiter        *cloudprovider.ResourceLimiter
-	machinesCache          map[MachineTypeKey]*gce.MachineType
-	migTargetSizeCache     map[GceRef]int64
-	migBaseNameCache       map[GceRef]string
-	instanceTemplatesCache map[GceRef]*gce.InstanceTemplate
+	migs                     map[GceRef]Mig
+	instanceRefToMigRef      map[GceRef]GceRef
+	instancesFromUnknownMigs map[GceRef]struct{}
+	resourceLimiter          *cloudprovider.ResourceLimiter
+	machinesCache            map[MachineTypeKey]*gce.MachineType
+	migTargetSizeCache       map[GceRef]int64
+	migBaseNameCache         map[GceRef]string
+	instanceTemplatesCache   map[GceRef]*gce.InstanceTemplate
 
 	// Service used to refresh cache.
 	GceService AutoscalingGceClient
@@ -73,13 +74,14 @@ type GceCache struct {
 // NewGceCache creates empty GceCache.
 func NewGceCache(gceService AutoscalingGceClient) *GceCache {
 	return &GceCache{
-		migs:                   map[GceRef]Mig{},
-		instanceRefToMigRef:    map[GceRef]GceRef{},
-		machinesCache:          map[MachineTypeKey]*gce.MachineType{},
-		migTargetSizeCache:     map[GceRef]int64{},
-		migBaseNameCache:       map[GceRef]string{},
-		instanceTemplatesCache: map[GceRef]*gce.InstanceTemplate{},
-		GceService:             gceService,
+		migs:                     map[GceRef]Mig{},
+		instanceRefToMigRef:      map[GceRef]GceRef{},
+		instancesFromUnknownMigs: map[GceRef]struct{}{},
+		machinesCache:            map[MachineTypeKey]*gce.MachineType{},
+		migTargetSizeCache:       map[GceRef]int64{},
+		migBaseNameCache:         map[GceRef]string{},
+		instanceTemplatesCache:   map[GceRef]*gce.InstanceTemplate{},
+		GceService:               gceService,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -216,11 +217,13 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 	nodeInfosForGroups, autoscalerError := getNodeInfosForGroups(
 		readyNodes, a.nodeInfoCache, autoscalingContext.CloudProvider, autoscalingContext.ListerRegistry, daemonsets, autoscalingContext.PredicateChecker, a.ignoredTaints)
 	if autoscalerError != nil {
+		klog.Errorf("Failed to get node infos for groups: %v", autoscalerError)
 		return autoscalerError.AddPrefix("failed to build node infos for node groups: ")
```
JulienBalestra: note: here we add a prefix but this error is never displayed by the caller in the main.

cluster-autoscaler/core/utils.go
```diff
@@ -254,7 +254,9 @@ func getNodeInfosForGroups(nodes []*apiv1.Node, nodeInfoCache map[string]*schedu
 	processNode := func(node *apiv1.Node) (bool, string, errors.AutoscalerError) {
 		nodeGroup, err := cloudProvider.NodeGroupForNode(node)
 		if err != nil {
-			return false, "", errors.ToAutoscalerError(errors.CloudProviderError, err)
+			klog.Warningf("Failed to get nodegroup for node %s: %v", node.Name, err)
+			metrics.RegisterError(errors.ToAutoscalerError(errors.CloudProviderError, err))
```
losipiuk: I think the fix should go to GCP cloud provider. Where we have more control over which situation we should treat as an error and which not. I do not believe we want to treat any error from `NodeGroupForNode` as "not-found".

t seems to me that to implement the changed behavior we only need to make local changes to GceCache.GetMigForInstance, so we do not return an error if we spot the problematic instance. We should also add negative cache for such broken instances so we do not end up with sending to much API call listing MIGs instances.

JulienBalestra: @losipiuk I've done the changes about the initial issue.
How would you like to do this negative cache ?
Do we return a custom error and remove the instance from the cache at this moment ?

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -182,7 +182,8 @@ func (gc *GceCache) GetMigForInstance(instanceRef GceRef) (Mig, error) {
 
 			migRef, found := gc.instanceRefToMigRef[instanceRef]
 			if !found {
-				return nil, fmt.Errorf("instance %+v belongs to unknown mig", instanceRef)
+				klog.Warningf("instance %+v belongs to unknown mig", instanceRef)
+				return nil, nil
```
losipiuk: Please also add the `gc.instancesFromUnknownMigs map[GceRef]bool{}` and put the entry the map here.
At the top of the function in `else` branch of loop from line `154` add checking of this new map. If entry is found return `(nil, nil)` from function immediately.

Also add code for cleaning up the new map together with `gc.instanceRefToMigRef`.

Without the new `gc.instancesFromUnknownMigs` map we will be listing instances of mig, which appears to hold the problematic instance, every loop. That will harm the performance.

cluster-autoscaler/core/utils.go
losipiuk: Thanks. I replied what I meant by the change in `cache.go`

losipiuk:
@JulienBalestra are you planning to address the comments on this one. I heard from other parties which would find this fix useful to them :)

cluster-autoscaler/cloudprovider/gce/cache.go
JulienBalestra: @losipiuk I've done the requests changes 👍 
I've done a `map[GceRef]struct{}` which is lighter than a bool one as we just store it as index/set. Let me know if everything's fine

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -195,10 +201,11 @@ func (gc *GceCache) GetMigForInstance(instanceRef GceRef) (Mig, error) {
 	return nil, nil
 }
 
-func (gc *GceCache) removeInstancesForMig(migRef GceRef) {
+func (gc *GceCache) removeInstancesForMigs(migRef GceRef) {
 	for instanceRef, instanceMigRef := range gc.instanceRefToMigRef {
 		if migRef == instanceMigRef {
 			delete(gc.instanceRefToMigRef, instanceRef)
+			delete(gc.instancesFromUnknownMigs, instanceRef)
```
losipiuk: It should not be needed. No harm though.

losipiuk: I think we should clear the Unknown map in RegenerateInstancesCache.
Without it we have (probably harmless leak) as removeInstancesForMigs will not be called for 
Migs which no longer exist

losipiuk:
@JulienBalestra one more thought on this one. If you had some time would you like to send PR which updates the `cluster-autoscaler/cloudprovider/gce/gce_cloud_provider_test.go` to test the added logic?

JulienBalestra:
@losipiuk Yes will do !

losipiuk:
> Łukasz Osipiuk Yes will do !

THanks!

losipiuk:
> > Łukasz Osipiuk Yes will do !
> 
> THanks!

Is there any progress on that one? :)

PR #2286, opened 2019-08-30
[cluster-autoscaler] Fixed go module builds

micahhausler:
* Moved digitalocean/godo to a vendor module instead of just copied code (this broke `go mod` commands)
* Bumped AWS SDK to v1.23.12
* Moved kubernetes modules to point to k8s.io/kubernetes/staging instead of `/tmp/abc/kubernetes`


losipiuk:
It will not work :/. We do not use go modules for building. It is just used for building vendor directory using somewhat hacky update-vendor.sh script. It is done in a very similar fashion (unless something changed) in k8s/kubernetes. Any changes you made to go.mod and vendor will be overridden on next run of update-vendor.sh.

micahhausler:
@losipiuk this _does_ work. 

I used the Kubernetes commit ID from #2102 ([ef7808fec57284582c7bf3fc834570e5769df83e](https://github.com/kubernetes/kubernetes/tree/ef7808fec57284582c7bf3fc834570e5769df83e)) to pin the Kubernetes module _instead_ of using a local `/tmp/abc` directory (see changes to go.mod). You'll see that there are no changes to the `k8s.io` packages in the vendor folder.

I also updated the rewrite for the staged k8s repos to point to the `/staging` path within Kubernetes
```diff
@@ -266,35 +266,35 @@ replace (
-	k8s.io/api => /tmp/abc/kubernetes/staging/src/k8s.io/api
+       k8s.io/api => k8s.io/kubernetes/staging/src/k8s.io/api v1.16.0-alpha.0.0.20190606024034-ef7808fec572
```
Look at K8s commit [ef7808fec57284582c7bf3fc834570e5769df83e](https://github.com/kubernetes/kubernetes/tree/ef7808fec57284582c7bf3fc834570e5769df83e), you'll see there is a `go.mod` in the repository

You can easily test that this PR builds with go modules, and without:

```bash
# Clone the repo (if not already present)
cd $GOPATH/src/k8s.io/
git clone https://github.com/kubernetes/autoscaler.git

# Checkout this PR
git fetch origin pull/2286/head:go-module-pr
git checkout go-module-pr

# You'll see that the only files changed are
# * AWS SDK
# * digital ocean cloud provider file moves
# * go.mod, go.sum, /vendor/modules.txt
git diff HEAD~1 --stat

# Builds using go modules (may take ~15min to download modules + build)
docker run -it --rm \
  -v $GOPATH/src/k8s.io/autoscaler:/go/src/k8s.io/autoscaler \
  -w /go/src/k8s.io/autoscaler/cluster-autoscaler \
  -e GO111MODULE=on \
  golang:1.12 go build -v

# Builds without go modules (and even without network!) (took me ~5min to build)
docker run -it --rm \
  --network none \
  -v $GOPATH/src/k8s.io/autoscaler:/go/src/k8s.io/autoscaler \
  -w /go/src/k8s.io/autoscaler/cluster-autoscaler \
  golang:1.12 go build -v


```

With this change, you can do away with the hacky `update-vendor.sh`.

losipiuk:
@micahhausler thanks for detailed explanation how to test the PR :)

I think `update-vendor.sh` is needed. We want to have dependencies in `cluster-autoscaler` as near to what current `k8s.io/kubernetes` uses as possible. `update-vendor.sh` allows us to do syncups easy and often.

Yet it make perfect sense to put reasonable pseudo-versions for stuff from `k8s.io/kubernetes/staging` in generated `go.mod` (i did not think of using versioned subdirectory of `k8s.io/kubernetes/subdirectory` in replace section of generated `go.mod`. That is great idea which will make `go.mod` actually resolvable). Could you add logic for that to `update-vendor.sh`? Or possibly I can do it some time later this week if you tell me how to easily determine correct pseudo-version pointing to go module at given git repository.

As for the AWS related changes - I think you should put requirement for newer version of SDK in `go-mod.extra` and rerun `update-vendor.sh`. As target for update you can use last commit used (via `-r` parameter). Or tip of the master.

micahhausler:
> I think update-vendor.sh is needed. We want to have dependencies in cluster-autoscaler as near to what current k8s.io/kubernetes uses as possible. update-vendor.sh allows us to do syncups easy and often.

Makes sense, however this is non-intuitive and not mentioned anywhere in the root CONTRIBUTING.md or in cluster-autoscaler README. Could that be more clearly documented? And can you clarify what you want out of "as near to what current k8s.io/kubernetes uses"? Are there specific exceptions?

> Yet it make perfect sense to put reasonable pseudo-versions for stuff from k8s.io/kubernetes/staging in generated go.mod (i did not think of using versioned subdirectory of k8s.io/kubernetes/subdirectory in replace section of generated go.mod. That is great idea which will make go.mod actually resolvable). Could you add logic for that to update-vendor.sh? Or possibly I can do it some time later this week if you tell me how to easily determine correct pseudo-version pointing to go module at given git repository.

I'll take a crack at this. To be clear, is this the logic you want?
* Any k8s.io/kubernetes, k8s.io/* (or k8s.io/kubernetes/staging) dependency in cluster-autoscaler should all be pinned to the same version
* Any dependency that cluster-autoscaler and k8s.io/kubernetes both have should be in sync with the what the above pinned version of k8s.io/kubernetes requires
* Any dependency that cluster-autoscaler has that is _not_ in k8s.io/kubernetes can be independently upgraded at its own pace

> As for the AWS related changes - I think you should put requirement for newer version of SDK in go-mod.extra and rerun update-vendor.sh. As target for update you can use last commit used (via -r parameter). Or tip of the master.

Why is this the case? `go-mod.extra` doesn't work with any of the go toolchain and adds extra fragility.

micahhausler:
Take a look at the `hack/update-k8s-dependencies.sh` in this PR. It performs the above logic on the k8s/k8s dependencies.

losipiuk:
> > I think update-vendor.sh is needed. We want to have dependencies in cluster-autoscaler as near to what current k8s.io/kubernetes uses as possible. update-vendor.sh allows us to do syncups easy and often.
> 
> Makes sense, however this is non-intuitive and not mentioned anywhere in the root CONTRIBUTING.md or in cluster-autoscaler README. Could that be more clearly documented? 

It should be more clear. I will add note to the documentation.

> And can you clarify what you want out of "as near to what current k8s.io/kubernetes uses"? Are there specific exceptions?

CA is importing significant part of core k8s code  (i.e. scheduler code). For correct operation it is crucial that the imported code behaves the same way as actual scheduler in the cluster. Therefore to minimize the chance of misbehavior we would like to use the same set of dependencies (the same versions) in CA as in k8s. The easy way to achieve that was to take go.mod out of k8s and use it in CA. And only make changes when those are really needed (mostly adding extra dependencies needed by cloud providers). 


> > Yet it make perfect sense to put reasonable pseudo-versions for stuff from k8s.io/kubernetes/staging in generated go.mod (i did not think of using versioned subdirectory of k8s.io/kubernetes/subdirectory in replace section of generated go.mod. That is great idea which will make go.mod actually resolvable). Could you add logic for that to update-vendor.sh? Or possibly I can do it some time later this week if you tell me how to easily determine correct pseudo-version pointing to go module at given git repository.
> 
> I'll take a crack at this. To be clear, is this the logic you want?
> 
> * Any k8s.io/kubernetes, k8s.io/* (or k8s.io/kubernetes/staging) dependency in cluster-autoscaler should all be pinned to the same version

Yes.

> * Any dependency that cluster-autoscaler and k8s.io/kubernetes both have should be in sync with the what the above pinned version of k8s.io/kubernetes requires

Unless newer version is required by CA. `update-vendor.sh` allows currently to explicitly increase version of module used in `k8s.io/kubernetes` via `go.mod-extra`.

> * Any dependency that cluster-autoscaler has that is _not_ in k8s.io/kubernetes can be independently upgraded at its own pace

Yes

> 
> > As for the AWS related changes - I think you should put requirement for newer version of SDK in go-mod.extra and rerun update-vendor.sh. As target for update you can use last commit used (via -r parameter). Or tip of the master.
> 
> Why is this the case? `go-mod.extra` doesn't work with any of the go toolchain and adds extra fragility.

Yeah - it is not supported directly. Yet I see a benefit of having it separate from automatically generated `go.mod` file. With `go.mod-extra` we can easily determine what is the delta of dependencies, specifically needed by CA on top of those coming from `k8s.io/kubernetes`. With just having `go.mod` it is harder to see that. Diffing against `go.mod` from `k8s.io/kubernetes` is needed, which given size of it is not practical. 
Also the fact that `go.mod-extra` is smaller than generated `go.mod` means we have less stuff to care about, as with current apporach we treat `go.mod` file as volatile and regenerate it from scratch on each `update-vendor.sh` call.

micahhausler:
>> * Any dependency that cluster-autoscaler and k8s.io/kubernetes both have should be in sync with the what the above pinned version of k8s.io/kubernetes requires

> Unless newer version is required by CA. update-vendor.sh allows currently to explicitly increase version of module used in k8s.io/kubernetes via go.mod-extra.

The current CA/K8s diffs I found were:

* github.com/aws/aws-sdk-go@v1.23.12 != github.com/aws/aws-sdk-go@v1.16.26 (my change)
* github.com/stretchr/testify@v1.3.0 != github.com/stretchr/testify@v1.2.2

> Yeah - it is not supported directly. Yet I see a benefit of having it separate from automatically generated go.mod file. With go.mod-extra we can easily determine what is the delta of dependencies, specifically needed by CA on top of those coming from k8s.io/kubernetes. With just having go.mod it is harder to see that. Diffing against go.mod from k8s.io/kubernetes is needed, which given size of it is not practical.
Also the fact that go.mod-extra is smaller than generated go.mod means we have less stuff to care about, as with current apporach we treat go.mod file as volatile and regenerate it from scratch on each update-vendor.sh call.

This seems incongruent with the above statement and list. What do you think about putting a list of tolerated dependency differences in the `update-vendor.sh` script? This way, the normal go module toolchain will work, dependencies will be easy to update. Any modules with different versions that you want to tolerate will still be tracked in version control, but not in `go.mod`

micahhausler:
I updated the FAQ with desired vendor behavior and added a new script `hack/update-mods.sh` with the signature:
```
update-mods.sh - Updates or detects differences in go.mod, go.sum, /vendor

    -f --k8sfork    The Kubernetes fork to use. Defaults to kubernetes/kubernetes
    -r --k8srev     Kuberntes revision to use. Defaults to ef7808fec57284582c7bf3fc834570e5769df83e
    -d --workdir    Working dir. Defaults to a temporary dir
    -n --no-update  Set flag to only view differences in go.mod, /vendor
    -h --help       Print this help output
```
It performs the following logic:

* Forces k8s.io/kubernetes and k8s.io/* (or k8s.io/kubernetes/staging) dependencies to be pinned to the same version specified in `-r/--k8srev`
* Fetches the `go.mod` file from k8s.io/kubernetes (or `-f/--k8sfork`) at `-r/--k8srev` for comparison
* Forces all dependencies in cluster-autoscaler that are shared by k8s.io/kubernetes to be in sync with the what the above pinned version of k8s.io/kubernetes requires from `k8s.io/kubernetes/go.mod`, with a list of exceptions in `PERMITTED_PACKAGE_DIFFS`
    * Calls out any drift in dependencies 
* Calls `go mod vendor` and `go mod tidy`

K8s updates are now as trivial as `hack/update-mods.sh -r v1.16.0` and updates to any 3rd party dependencies can use `go get my.module@v1.0.0`. This all works with the above `docker build` test commands.

At this point, we could probably remove the `update-vendor.sh` and the `go.mod-extra`

cluster-autoscaler/hack/update-mods.sh
```diff
@@ -0,0 +1,206 @@
...
+  exit 1
+fi
+
+WORK_DIR="${WORK_DIR:-$(mktemp -d /tmp/ca-update-vendor.XXXX)}"
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        -h | --help ) PRINT_HELP="true"; break;;
+        -f | --k8sfork ) K8S_FORK="$2"; shift; shift ;;
+        -r | --k8srev ) K8S_REV="$2"; shift; shift ;;
```
losipiuk: Please allow tagname/branchname here. 
Resolve with git rev-parse.

If possible also add an option to automatically commit the changes (that will allow to have consistent commit message for go.mod updates).

cluster-autoscaler/hack/update-mods.sh
```diff
@@ -0,0 +1,206 @@
...
+        go mod edit -replace=${mod}=k8s.io/kubernetes/staging/src/${mod}@${K8S_PSEUDO_VERSION}
+    done
+fi
+
+################################################################################
+### Get k8s.io/kubernetes/go.mod@$K8S_REV ######################################
+################################################################################
+cd $WORK_DIR
+K8S_GO_MOD_JSON="k8s-go-mod.json"
+curl -O $WORK_DIR/go.mod -s "https://raw.githubusercontent.com/${K8S_FORK}/${K8S_REV}/go.mod"
```
losipiuk: This should be `-o`. With `-O` the `go.mod` file is not saved at all. Did this work for you?

cluster-autoscaler/hack/update-mods.sh
```diff
@@ -0,0 +1,206 @@
...
+# Get Cluster Autoscaler required packages
+cd $AUTOSCALE_DIR
+CA_GO_MOD_JSON="ca-go-mod.json"
+go mod edit -json > $WORK_DIR/$CA_GO_MOD_JSON
+CA_REQS=$(cat $WORK_DIR/$CA_GO_MOD_JSON | jq -r .Require[].Path)
+
+# Record Cluster Autoscaler differences from k8s/k8s
+CA_DIFFS=()
+CA_ONLY_MODS=()
+echo "Cluster Autoscale & Kubernetes Dependencies"
```
losipiuk: Autoscaler

cluster-autoscaler/hack/update-mods.sh
```diff
@@ -0,0 +1,206 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 The Kubernetes Authors.
+#
```
losipiuk: General comment on calling go mod.

Please redirect stderr to `/dev/null`. It clutters the output.
Or better than using `/dev/null`  create some debug log and pass stderr there.
You can take logging bits from update-vendor.sh

cluster-autoscaler/hack/update-mods.sh
```diff
@@ -0,0 +1,206 @@
...
+
+# Packages that are allowed to drift from k8s/k8s
+PERMITTED_PACKAGE_DIFFS=(
+    github.com/aws/aws-sdk-go
+    github.com/stretchr/testify
+)
+
+# Kubernetes staging repos: will need to be periodically updated
+# or update this script to check what is in k8s.io/kubernetes/stagin/src each time
+K8S_STAGING_MODULES=(
```
losipiuk: You can parse those out of k8s `go.mod` file.

cluster-autoscaler/hack/update-mods.sh
```diff
@@ -0,0 +1,206 @@
...
+
+err_report() {
+    echo "Exited with error on line $1"
+}
+trap 'err_report $LINENO' ERR
+
+SCRIPT_NAME=$(basename "$0")
+K8S_FORK="kubernetes/kubernetes"
+# Update here when we need a new version of Kubernetes
+K8S_REV="ef7808fec57284582c7bf3fc834570e5769df83e"
```
losipiuk: Can we update that automatically? Specifically in the auto-commit mode mentioned below. 
It seems quite fragile to modify a script which is being executed though. So maybe it should be move to some other file (`update-mods.k8s_rev`?)

cluster-autoscaler/hack/update-mods.sh
```diff
@@ -0,0 +1,206 @@
...
+            go mod edit -require=${req}@${K8S_DEPENDENCY_VERSION}
+            go mod edit -replace=${req}=${req}@${K8S_DEPENDENCY_VERSION}
+        fi
+    fi
+    if [ -z "$K8S_DEPENDENCY_VERSION" ]; then
+        CA_ONLY_MODS+=("${req}@${CA_VERSION}")
+    fi
+done
+
+# TODO Exit if CA_DIFFS is not empty?
```
losipiuk: What would happen if go.mod file contains a dependency `A` in version `V1` as a transitive dependency of some k8s code and on call to `update-mods.sh` the version of `A` required by k8s is bumped to `V2`.
It feels like we will have `A` in `CA_DIFFS` then. 

That is normal and expected situation. We should not warn user about that. 

The disturbing situation is when we are using `Replace` for `A` in `go.mod`, to use version which is different than one used in `k8s`. Or more specifically older than one used in `k8s`. IMO we should error out only in such case.

cluster-autoscaler/hack/update-mods.sh
```diff
@@ -0,0 +1,206 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
```
losipiuk: We should check if required go version in CA `go.mod` is the same as one in `k8s`.

cluster-autoscaler/hack/update-mods.sh
micahhausler: > What would happen if go.mod file contains a dependency A in version V1 as a transitive dependency of some k8s code and on call to update-mods.sh the version of A required by k8s is bumped to V2.
It feels like we will have A in CA_DIFFS then.

It will, but just below with `UPDATE_GO_MOD="true"`, we'll always set the `require` and the `replace` to whatever k8s uses

```bash
go mod edit -require=${req}@${K8S_DEPENDENCY_VERSION} 2&1       
go mod edit -replace=${req}=${req}@${K8S_DEPENDENCY_VERSION} 2&1
```

> That is normal and expected situation. We should not warn user about that.

I changed this to only report any detected drift from k8s/k8s when go.mod is not edited

> The disturbing situation is when we are using Replace for A in go.mod, to use version which is different than one used in k8s. Or more specifically older than one used in k8s. IMO we should error out only in such case.

This shouldn't happen with the above 

```bash
go mod edit -require=${req}@${K8S_DEPENDENCY_VERSION} 2&1       
go mod edit -replace=${req}=${req}@${K8S_DEPENDENCY_VERSION} 2&1
```

cluster-autoscaler/hack/update-mods.sh
micahhausler: Done

cluster-autoscaler/hack/update-mods.sh
```diff
@@ -0,0 +1,206 @@
...
+
+################################################################################
+### Flag and directory validation ##############################################
+################################################################################
+
+if [[ $(basename $(pwd)) != "cluster-autoscaler" ]];then
+  echo "The script must be run in cluster-autoscaler directory"
+  exit 1
+fi
+AUTOSCALE_DIR=$(pwd)
```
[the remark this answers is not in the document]
micahhausler: Done

cluster-autoscaler/hack/update-mods.sh
```diff
@@ -0,0 +1,206 @@
...
+        * ) break ;;
+    esac
+done
+
+if [[ "${PRINT_HELP}" == "true" ]]; then
+    cat << EOF >> /dev/stderr
+$SCRIPT_NAME - Updates or detects differences in go.mod, go.sum, /vendor
+
+    -f --k8sfork    The Kubernetes fork to use. Defaults to $K8S_FORK
+    -r --k8srev     Kuberntes revision to use. Defaults to ${K8S_REV}
```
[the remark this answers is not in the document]
micahhausler: Derp

cluster-autoscaler/hack/update-mods.sh
```diff
@@ -0,0 +1,206 @@
...
+    -d --workdir    Working dir. Defaults to a temporary dir
+    -n --no-update  Set flag to only view differences in go.mod, /vendor
+    -h --help       Print this help output
+EOF
+    exit 0
+fi
+
+export GO111MODULE=on
+
```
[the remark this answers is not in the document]
micahhausler: Done

cluster-autoscaler/hack/update-mods.sh
micahhausler: Since this is just referencing a GitHub raw URL any tag, branch, or commit ref should work fine here

cluster-autoscaler/hack/update-mods.sh
micahhausler: Derp, that was the last thing I added. I'll fix that

cluster-autoscaler/hack/update-mods.sh
micahhausler: Done

cluster-autoscaler/hack/update-mods.sh
micahhausler: Done

cluster-autoscaler/hack/update-mods.sh
micahhausler: Done

cluster-autoscaler/hack/update-mods.sh
micahhausler: The script updates a file (`hack/k8s_rev`) at the end and adds it to the commit

micahhausler:
Addressed all comments, and split commits.

cluster-autoscaler/hack/update-mods.sh
```diff
@@ -0,0 +1,206 @@
+#!/usr/bin/env bash
+
```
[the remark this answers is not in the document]
micahhausler: I re-ran and didn't get any checksum mismatch. Can you reproduce in a fresh container?

```
docker run -it --rm \
  -v $GOPATH/src/k8s.io/autoscaler:/go/src/k8s.io/autoscaler \
  -w /go/src/k8s.io/autoscaler/cluster-autoscaler \
  -e GO111MODULE=on \
  golang:1.12 /bin/bash
```

micahhausler:
There seems to be an issue with go mod vendor and go mod tidy on commits that have a tag, but not on untagged commits. We'll hold this PR until we can get that sorted out

losipiuk:
> There seems to be an issue with go mod vendor and go mod tidy on commits that have a tag, but not on untagged commits. We'll hold this PR until we can get that sorted out.

@micahhausler are you planning to push the issue towards go modules developers?

cluster-autoscaler/hack/update-mods.sh
jaypipes: Ref: https://github.com/golang/go/issues/35878

losipiuk:
Closing this one. @jaypipes volunteered to pick up the work on this one and will send followup PR.

PR #2304, opened 2019-09-05
Stop disabling Cluster Autoscaler when there is no ready nodes.

### LINKED ISSUE #2218: why can not  cluster-autoscaler work from empty cluster?
In cluster-autoscaler  we found this, it limit the cluster-autoscaler in empty node cluster, It why ? Can we remove the limit ?

The code is :
if a.actOnEmptyCluster(allNodes, readyNodes, currentTime) {
    return nil
}

krzysztof-jastrzebski:
Fixes #2218

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -362,7 +362,7 @@ func (csr *ClusterStateRegistry) IsClusterHealthy() bool {
 	csr.Lock()
 	defer csr.Unlock()
 
-	totalUnready := csr.totalReadiness.Unready + csr.totalReadiness.LongNotStarted + csr.totalReadiness.LongUnregistered
+	totalUnready := csr.totalReadiness.Unready
 
 	if totalUnready > csr.config.OkTotalUnreadyCount &&
 		float64(totalUnready) > csr.config.MaxTotalUnreadyPercentage/100.0*float64(len(csr.nodes)) {
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -130,6 +130,7 @@ var (
 	maxGracefulTerminationFlag = flag.Int("max-graceful-termination-sec", 10*60, "Maximum number of seconds CA waits for pod termination when trying to scale down a node.")
 	maxTotalUnreadyPercentage  = flag.Float64("max-total-unready-percentage", 45, "Maximum percentage of unready nodes in the cluster.  After this is exceeded, CA halts operations")
 	okTotalUnreadyCount        = flag.Int("ok-total-unready-count", 3, "Number of allowed unready nodes, irrespective of max-total-unready-percentage")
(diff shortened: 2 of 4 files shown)


cluster-autoscaler/main.go
```diff
@@ -130,6 +130,7 @@ var (
 	maxGracefulTerminationFlag = flag.Int("max-graceful-termination-sec", 10*60, "Maximum number of seconds CA waits for pod termination when trying to scale down a node.")
 	maxTotalUnreadyPercentage  = flag.Float64("max-total-unready-percentage", 45, "Maximum percentage of unready nodes in the cluster.  After this is exceeded, CA halts operations")
 	okTotalUnreadyCount        = flag.Int("ok-total-unready-count", 3, "Number of allowed unready nodes, irrespective of max-total-unready-percentage")
+	noScaleUpFromZero          = flag.Bool("no-scale-up-from-zero", false, "Shouldn't CA scale up when there 0 ready nodes.")
```
aleksandra-malinowska: nit: can we change this to a positive ("scale-up-from-zero")?

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -362,7 +362,7 @@ func (csr *ClusterStateRegistry) IsClusterHealthy() bool {
 	csr.Lock()
 	defer csr.Unlock()
 
-	totalUnready := csr.totalReadiness.Unready + csr.totalReadiness.LongNotStarted + csr.totalReadiness.LongUnregistered
```
aleksandra-malinowska: What is the reasoning for removing LongNotStarted and LongUnregistered from the condition?

krzysztof-jastrzebski: LongNotStarted and LongUnregistered are nodes that failed to start. In my opinion such nodes should be removed and other mechanism will backoff node pool, Cluster Autoscaler should recover from situation when there is a lot of such nodes. Deleting such nodes is not a problem as we are sure that nothing is running on these nodes. Cluster should be considered as unhealthy only when large part of nodes that were ready becomes unready because there are some pods running on these nodes and we don't want to break it even more.

cluster-autoscaler/main.go
krzysztof-jastrzebski: I used negative because I want unit tests to test it as default without changing tests' code.
If you think this is big problem then I can change it.

aleksandra-malinowska: Using positive with default value 'true' would have the same effect, right?

cluster-autoscaler/clusterstate/clusterstate.go
aleksandra-malinowska: Ack, didn't realize node group health remains unchanged.

cluster-autoscaler/main.go
krzysztof-jastrzebski: Changed. I will change tests to use default values in next PR

PR #2305, opened 2019-09-05
Revendor against tip of k8s.io/kubernetes to prepare for 1.16 release of CA

losipiuk:
Note: This pr uses old version of revendoring. The process will change slightly with #2286 merged


micahhausler:
@losipiuk this is going to cause quite a merge conflict with #2286, what do you think about holding off until that is merged?

losipiuk:
> Łukasz Osipiuk this is going to cause quite a merge conflict with #2286, what do you think about holding off until that is merged?

I would prefer to have it merged first. We need revendoring for sake of building release for k8s 1.16 which is in freeze already. IMO introducing new process just before cutting new release adds unnecessary risk. #2286 should be easy to rebase just bu rerunning script you prepared there. Right?

micahhausler:
The rebase won't be simple, because my PR also moves the digitalocean cloudprovider out into /vendor, and this change modifies that and other dependencies.

The risk should be really minimal, as re-running my change against the previous k8s commit (relative to this PR) resulted in no changes

micahhausler:
There seems to be an issue with `go mod vendor` and `go mod tidy`  on commits that have a tag, but not on untagged commits. We'll skip using my PR for now until we resolve that go mod issue

losipiuk:
Updated this PR to include bumping `github.com/aws/aws-sdk-go` to `v1.23.12` as discussed with @micahhausler.
@Jeffwan PTAL.

PR #2308, opened 2019-09-06
compare_nodegroups: Tolerate small differences in memory capacity

frobware:
The current comparator expects memory capacity values to be identical.
However across AWS, Azure and GCP I quite often see very small
differences in capacity, typically 8-16Ki. When this occurs the
nodegroups are considered not equal when balancing is in effect which
is unfortunate because, in reality, they are identical.

This change will now tolerate a 128Ki difference before memory
capacity values are considered unequal.

### DIFF
--- a/cluster-autoscaler/processors/nodegroupset/compare_nodegroups.go
+++ b/cluster-autoscaler/processors/nodegroupset/compare_nodegroups.go
@@ -31,6 +31,9 @@ const (
 	// MaxFreeDifferenceRatio describes how free resources (allocatable - daemon and system pods)
 	// can differ between groups in the same NodeGroupSet
 	MaxFreeDifferenceRatio = 0.05
+	// MaxMemoryDifferenceInKiloBytes describes how much memory
+	// capacity can differ but still be considered equal.
+	MaxMemoryDifferenceInKiloBytes = 128000
 )
 
 // BasicIgnoredLabels define a set of basic labels that should be ignored when comparing the similarity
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/processors/nodegroupset/compare_nodegroups.go
```diff
@@ -31,6 +31,9 @@ const (
 	// MaxFreeDifferenceRatio describes how free resources (allocatable - daemon and system pods)
 	// can differ between groups in the same NodeGroupSet
 	MaxFreeDifferenceRatio = 0.05
+	// MaxMemoryDifferenceInKiloBytes describes how much memory
+	// capacity can differ but still be considered equal.
+	MaxMemoryDifferenceInKiloBytes = 2 << 7 * 1000 // 128Ki
```
losipiuk: Please write it as 128000

cluster-autoscaler/processors/nodegroupset/compare_nodegroups.go
```diff
@@ -110,14 +113,29 @@ func IsCloudProviderNodeInfoSimilar(n1, n2 *schedulernodeinfo.NodeInfo, ignoredL
...
+	for kind, qtyList := range capacity {
+		if len(qtyList) != 2 {
 			return false
 		}
+		switch kind {
+		case apiv1.ResourceMemory:
+			// For memory capacity we allow a small tolerance
+			larger := math.Max(float64(qtyList[0].Value()), float64(qtyList[1].Value()))
+			smaller := math.Min(float64(qtyList[0].Value()), float64(qtyList[1].Value()))
+			if larger-smaller > MaxMemoryDifferenceInKiloBytes {
```
losipiuk:
```
memoryDifference := Math.Abs(qtyList[0].Value()) - float64(qtyList[1].Value())
if memoryDifference >
```

frobware:
> Why 128k? What are the actuall differences observed?

In some limited testing 8 or 16Ki. I was trying to avoid the case where we may see slightly larger values without going overboard with the value at this stage. Happy to readjust - not so happy to have to keep changing/rebuilding and distributing incremental fixes (in OpenShift at least).

cluster-autoscaler/processors/nodegroupset/compare_nodegroups.go
frobware: https://github.com/kubernetes/autoscaler/compare/a743c4a9743bcacc886d8bb61773472fb9e7f7d0..56d99b72b3384b20b33437571e605eb644cf5172

cluster-autoscaler/processors/nodegroupset/compare_nodegroups.go
frobware: Done in https://github.com/kubernetes/autoscaler/compare/56d99b72b3384b20b33437571e605eb644cf5172..e8b3c2a111eb9b3b4fe15b4d4081175267ff6d76

losipiuk:
/lgtm
/approve
Thanks!

frobware:
FYI @hardikdr

hardikdr:
Thanks @frobware for the quick PR.

PR #2323, opened 2019-09-10
Updated aws-sdk-go to 1.23.18

alexpekurovsky:
Update the AWS SDK to support Web Identity providers for IAM credentials

More infos here:

- https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
- https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-minimum-sdk.html
- https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/


alexpekurovsky:
Could it also be backported? EKS currently is 1.14, so even releasing it to 1.16/1.17 won't help with EKS

devkid:
I would also love to have this backported. The alternative would be to wait for Kubernetes 1.16 on EKS or everyone needs to build their own image with this fix included to get the advantage of IAM roles for serviceaccounts.

losipiuk:
The contents of this PR are unexpected. Why the only file changed in vendor is `cluster-autoscaler/vendor/github.com/aws/aws-sdk-go/aws/version.go`. It seems that something else should be changed. Otherwise what is the point of upgrading?

Btw. Did you run the `update-vendor.sh` script? After changing go.mod-extra it should be rerun. To limit amount of changes in `go.mod` and `vendor` we can rerun it agains the same commit which was used for previous updating. I.e.

`update-vendor.sh -r708a7f76dba706ed1e7dc4b4c3a53f63b0b7544a`

alexpekurovsky:
The point of upgrading of AWS SDK is to get support of new way of getting IAM Roles credentials (without reading them from metadata). I just did search&replace of aws-sdk-go version. If something is needed additionally - please process.

Identical PR for external-dns: https://github.com/kubernetes-incubator/external-dns/pull/1182

micahhausler:
@alexpekurovsky The external-dns change was only to `go.mod` and `go.sum` because that project doesn't vendor its dependencies. Autoscaler vendors its dependences, so there needed to be a vendor update in addition to a go.mod bump

#2305 includes an updated version of the AWS SDK which will support this

alexpekurovsky:
@micahhausler, I see. However #2305 upgrades aws-sdk-go to 1.23.12, but for WebIdentity at least 1.23.13 is required

alexpekurovsky:
@losipiuk, PR updated using `update-vendor.sh -r708a7f76dba706ed1e7dc4b4c3a53f63b0b7544a`

losipiuk:
> Łukasz Osipiuk Can this PR be cherry-picked to 1.13+ releases? I've checked 1.14 release is based on aws 1.16.26, so it shouldn't support Web Identity according to information from Micah Hausler

I am not big fan of the idea. This is not a bugfix. And also it cannot be cherry picked directly as all the release branches have very different set of libraries in vendor folder (each matching different k8s minor version). Revendoring 1.14 and and below is painful and risky process (see https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-release-1.14/cluster-autoscaler/FAQ.md#how-can-i-update-ca-dependencies-particularly-k8siokubernetes) and doing that for patch release feels to me as bad idea.

asthasr:
Is it not possible to revendor only the AWS SDK for 1.14? After all, updating the AWS SDK to support this feature is pointless if EKS does not support the version in question. It may be months before EKS catches up to the k8s version, and it strikes me as very bad if SDKs for a cloud provider cannot be updated at a cadence that matches the cloud provider instead of mainline Kubernetes.

whereisaaron:
@losipiuk I think that maintaining the auto-scaler 1.14 branch is a necessary consequence of not supporting backward compatibility in the cluster autoscaler. If each cluster autoscaler release supported the current _and_ previous couple k8s releases, then I see your point about not back-porting features. But since releases are 1:1, old cluster autoscaler releases need to be maintained and supported until that release of k8s is EOL. For 1.14 on AWS/EKS, EOL is still 9 months away. And 1.16 won't even be available to deploy on AWS for another ~5 months.

https://aws.amazon.com/blogs/compute/updates-to-amazon-eks-version-lifecycle/
https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html

losipiuk:
> Łukasz Osipiuk I think that maintaining the auto-scaler 1.14 branch is a necessary consequence of not supporting backward compatibility in the cluster autoscaler. If each cluster autoscaler release supported the current _and_ previous couple k8s releases, then I see your point about not back-porting features. But since releases are 1:1, old cluster autoscaler releases need to be maintained and supported until that release of k8s is EOL. For 1.14 on AWS/EKS, EOL is still 9 months away. And 1.16 won't even be available to deploy on AWS for another ~5 months.
> 
> https://aws.amazon.com/blogs/compute/updates-to-amazon-eks-version-lifecycle/
> https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html

We mostly follow the OSS k8s release cycle (which unless something changed) provide patch fixes for 3 most recent releases. We try our best to support a wider set of releases, as we understand that service providers are lagging behind OSS. Yet all that happens within constraints coming from how much resources we can put to maintenance.

PR #2326, opened 2019-09-11
Add rate limiter to the Vertical Pod autoscaler updater component

### LINKED ISSUE #2178: vertical-pod-autoscaler: rate limit pod evictions
We've been using the VPA for a couple of weeks, and it's generally working really well 👍unfortunately, on a couple of occasions, it's working a bit _too_ effectively, and creating a ton of pod evictions in a short space of time. We run a fairly large cluster, and on a couple of occasions it's evicted about 2-3k pods in the space of 15-20 minutes.

This large pod churn puts a huge strain on the apiservers and service discovery mechanisms - we're scaling these up, but we also think there should be a way of restricting how many pods the `vpa-updater` can kill in a period of time.

We could try doing this with the `--kube-api-qps` flag for this, but this stays generally low (4-5 RPS), so not ideal. I was thinking about introducing a new CLI flag on the updater, something like `--max-pod-evictions-per-sec`, which would explicitly rate limit the amount of pod evictions the updater could do.

I'm very happy to do the work myself on this; I just thought I'd start a conversation on how this might work best 🙂

guillaumebreton:
This patch adds a rate limiter to the vertical pod autoscaler updater.
It adds two flags
- eviction-rate-limit to control the number of pods that can be evicted
every seconds.
- eviction-rate-limit-burst to control the burst of that can be evicted
immediately.

fixes #2178

### DIFF
--- a/vertical-pod-autoscaler/pkg/updater/logic/updater.go
+++ b/vertical-pod-autoscaler/pkg/updater/logic/updater.go
@@ -17,8 +17,11 @@ limitations under the License.
 package logic
 
 import (
+	"context"
 	"fmt"
 	"time"
+
+	"golang.org/x/time/rate"
 
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/fields"
@@ -44,7 +47,7 @@ import (
 // Updater performs updates on pods if recommended by Vertical Pod Autoscaler
 type Updater interface {
 	// RunOnce represents single iteration in the main-loop of Updater
-	RunOnce()
+	RunOnce(context.Context)
 }
 
 type updater struct {
@@ -54,11 +57,13 @@ type updater struct {
 	evictionFactory         eviction.PodsEvictionRestrictionFactory
 	recommendationProcessor vpa_api_util.RecommendationProcessor
 	evictionAdmission       priority.PodEvictionAdmission
+	evictionRateLimiter     *rate.Limiter
 	selectorFetcher         target.VpaTargetSelectorFetcher
 }
 
 // NewUpdater creates Updater with given configuration
-func NewUpdater(kubeClient kube_client.Interface, vpaClient *vpa_clientset.Clientset, minReplicasForEvicition int, evictionToleranceFraction float64, recommendationProcessor vpa_api_util.RecommendationProcessor, evictionAdmission priority.PodEvictionAdmission, selectorFetcher target.VpaTargetSelectorFetcher) (Updater, error) {
+func NewUpdater(kubeClient kube_client.Interface, vpaClient *vpa_clientset.Clientset, minReplicasForEvicition int, evictionRateLimit float64, evictionRateBurst int, evictionToleranceFraction float64, recommendationProcessor vpa_api_util.RecommendationProcessor, evictionAdmission priority.PodEvictionAdmission, selectorFetcher target.VpaTargetSelectorFetcher) (Updater, error) {
+	evictionRateLimiter := getRateLimiter(evictionRateLimit, evictionRateBurst)
 	factory, err := eviction.NewPodsEvictionRestrictionFactory(kubeClient, minReplicasForEvicition, evictionToleranceFraction)
 	if err != nil {
 		return nil, fmt.Errorf("Failed to create eviction restriction factory: %v", err)
@@ -69,13 +74,14 @@ func NewUpdater(kubeClient kube_client.Interface, vpaClient *vpa_clientset.Clien
 		eventRecorder:           newEventRecorder(kubeClient),
 		evictionFactory:         factory,
 		recommendationProcessor: recommendationProcessor,
+		evictionRateLimiter:     evictionRateLimiter,
 		evictionAdmission:       evictionAdmission,
 		selectorFetcher:         selectorFetcher,
 	}, nil
 }
 
 // RunOnce represents single iteration in the main-loop of Updater
-func (u *updater) RunOnce() {
+func (u *updater) RunOnce(ctx context.Context) {
 	timer := metrics_updater.NewExecutionTimer()
 
 	vpaList, err := u.vpaLister.List(labels.Everything())
@@ -144,6 +150,11 @@ func (u *updater) RunOnce() {
 			if !evictionLimiter.CanEvict(pod) {
 				continue
 			}
+			err := u.evictionRateLimiter.Wait(ctx)
+			if err != nil {
+				klog.Warningf("evicting pod %v failed: %v", pod.Name, err)
+				return
+			}
 			klog.V(2).Infof("evicting pod %v", pod.Name)
 			evictErr := evictionLimiter.Evict(pod, u.eventRecorder)
 			if evictErr != nil {
@@ -153,6 +164,19 @@ func (u *updater) RunOnce() {
 	}
 	timer.ObserveStep("EvictPods")
 	timer.ObserveTotal()
+}
+
+func getRateLimiter(evictionRateLimit float64, evictionRateLimitBurst int) *rate.Limiter {
+	var evictionRateLimiter *rate.Limiter
+	if evictionRateLimit <= 0 {
+		// As a special case if the rate is set to rate.Inf, the burst rate is ignored
+		// see https://github.com/golang/time/blob/master/rate/rate.go#L37
+		evictionRateLimiter = rate.NewLimiter(rate.Inf, 0)
+		klog.V(1).Info("Rate limit disabled")
+	} else {
+		evictionRateLimiter = rate.NewLimiter(rate.Limit(evictionRateLimit), evictionRateLimitBurst)
+	}
+	return evictionRateLimiter
 }
 
 // getPodsUpdateOrder returns list of pods that should be updated ordered by update priority
--- a/vertical-pod-autoscaler/pkg/updater/logic/updater_test.go
+++ b/vertical-pod-autoscaler/pkg/updater/logic/updater_test.go
@@ -17,10 +17,14 @@ limitations under the License.
 package logic
 
 import (
+	"context"
 	"strconv"
 	"testing"
 
+	"golang.org/x/time/rate"
+
 	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
 	apiv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -87,12 +91,13 @@ func TestRunOnce(t *testing.T) {
 		vpaLister:               vpaLister,
 		podLister:               podLister,
 		evictionFactory:         factory,
+		evictionRateLimiter:     rate.NewLimiter(rate.Inf, 0),
 		recommendationProcessor: &test.FakeRecommendationProcessor{},
 		selectorFetcher:         mockSelectorFetcher,
 	}
 
 	mockSelectorFetcher.EXPECT().Fetch(gomock.Eq(vpaObj)).Return(selector, nil)
-	updater.RunOnce()
+	updater.RunOnce(context.Background())
 	eviction.AssertNumberOfCalls(t, "Evict", 5)
 }
 
@@ -134,11 +139,12 @@ func TestVPAOff(t *testing.T) {
 		vpaLister:               vpaLister,
 		podLister:               podLister,
 		evictionFactory:         factory,
+		evictionRateLimiter:     rate.NewLimiter(rate.Inf, 0),
 		recommendationProcessor: &test.FakeRecommendationProcessor{},
 		selectorFetcher:         target_mock.NewMockVpaTargetSelectorFetcher(ctrl),
 	}
 
-	updater.RunOnce()
+	updater.RunOnce(context.Background())
 	eviction.AssertNumberOfCalls(t, "Evict", 0)
 }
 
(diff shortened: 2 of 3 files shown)


bskiba:
@guillaumebreton thanks! This looks ok, I have some minor comments but first I wanted to clarify that this is what we want. 
@milesbxf does global setting like this solve your issues is there need for more fine grained control (for example ability to set this per-deployment)?

vertical-pod-autoscaler/pkg/updater/logic/updater.go
```diff
@@ -155,6 +165,17 @@ func (u *updater) RunOnce() {
 	timer.ObserveTotal()
 }
 
+func getRateLimiter(evictionRateLimit float64, evictionRateLimitBurst int) *rate.Limiter {
+	var evictionRateLimiter *rate.Limiter
+	if evictionRateLimit == -1 || evictionRateLimit == 0 {
+		evictionRateLimiter = rate.NewLimiter(rate.Inf, evictionRateLimitBurst)
+		klog.Warning("Rate limit disabled")
+	} else {
+		evictionRateLimiter = rate.NewLimiter(rate.Every(time.Duration(1.0/evictionRateLimit*float64(time.Second))), evictionRateLimitBurst)
```
bandesz: If I understand the documentation correctly the first parameter is the rate per second, so we could just use `rate.NewLimiter(rate.Limit(evictionRateLimit))`

bandesz: See https://github.com/golang/time/blob/master/rate/rate.go#L25

vertical-pod-autoscaler/pkg/updater/logic/updater.go
```diff
@@ -144,6 +149,11 @@ func (u *updater) RunOnce() {
 			if !evictionLimiter.CanEvict(pod) {
 				continue
 			}
+			err := u.evictionRateLimiter.Wait(ctx)
```
bandesz: `Wait` always asks for one token, so it doesn't make much sense to set the burst rate to anything else than 1. If I'm correct then we should remove the `eviction-rate-limit-burst` parameter.

bandesz: Without knowing the autoscaler well I'm a little bit worried if `Wait` will block here for long time (e.g. minutes). The state of the cluster might change significantly during that time.

bandesz: An idea for solving the Wait()/blocking problem:
1. Let's collect all pods to be evicted in a list
2. Randomize the list (to give all pods a fair chance to be evicted in a RunOnce run)
3. Go through the list and evict a pod if `u.evictionRateLimiter.Allow()` returns true.
4. If `u.evictionRateLimiter.Allow()` returns false then return from the RunOnce function. Any remaining pods to be evicted will be retried in the next RunOnce run

guillaumebreton: Thanks for you input 🙂 

I'm not really sure why the Wait is an issue here :
- A context with timeout it passed to the Wait function so the Wait will be cancelled if the delay is larger than the tick interval.
- As far as I understand it. The state of the cluster is updated by the recommender, and the admission controller applies it. The updater is only responsible for eviction here, and when the pod restarts, the last value of the VPA are applied by the admission controller.

In your proposed solution, we would hit the rate limiter quickly and exit. So we would "waste" n seconds of the ticker loop doing nothing. Running on fresher data is an excellent idea, but the only option I see would be to refresh the pod list between every Wait, which doesn't seem ideal to me.

milesbxf:
> @guillaumebreton thanks! This looks ok, I have some minor comments but first I wanted to clarify that this is what we want.
> @milesbxf does global setting like this solve your issues is there need for more fine grained control (for example ability to set this per-deployment)?

Yes, it does! 🙌 FYI @guillaumebreton and I are on the same team, so we've spoken about this offline 🙂

bskiba:
@milesbxf Thanks! In that case I'll get back to this early next week, appreciate your patience I am a bit short on time atm :)

guillaumebreton:
@bskiba No worries 👍 Thank you for reviewing it.

vertical-pod-autoscaler/pkg/updater/logic/updater.go
bskiba: +1 to this comment

vertical-pod-autoscaler/pkg/updater/logic/updater.go
```diff
@@ -155,6 +165,17 @@ func (u *updater) RunOnce() {
 	timer.ObserveTotal()
 }
 
+func getRateLimiter(evictionRateLimit float64, evictionRateLimitBurst int) *rate.Limiter {
+	var evictionRateLimiter *rate.Limiter
+	if evictionRateLimit == -1 || evictionRateLimit == 0 {
+		evictionRateLimiter = rate.NewLimiter(rate.Inf, evictionRateLimitBurst)
+		klog.Warning("Rate limit disabled")
```
bskiba: I think this is Info level log, it's the default currently to have rate limit disabled.

vertical-pod-autoscaler/pkg/updater/logic/updater.go
```diff
@@ -155,6 +165,17 @@ func (u *updater) RunOnce() {
 	timer.ObserveTotal()
 }
 
+func getRateLimiter(evictionRateLimit float64, evictionRateLimitBurst int) *rate.Limiter {
+	var evictionRateLimiter *rate.Limiter
+	if evictionRateLimit == -1 || evictionRateLimit == 0 {
+		evictionRateLimiter = rate.NewLimiter(rate.Inf, evictionRateLimitBurst)
```
bskiba: I haven't dived into the rate package docs, is this equivalent to having no rateLimiter at all? i.e. will the Wait() return immedately?

vertical-pod-autoscaler/pkg/updater/logic/updater.go
```diff
@@ -54,11 +56,13 @@ type updater struct {
...
 	recommendationProcessor vpa_api_util.RecommendationProcessor
 	evictionAdmission       priority.PodEvictionAdmission
+	evictionRateLimiter     *rate.Limiter
 	selectorFetcher         target.VpaTargetSelectorFetcher
 }
 
 // NewUpdater creates Updater with given configuration
-func NewUpdater(kubeClient kube_client.Interface, vpaClient *vpa_clientset.Clientset, minReplicasForEvicition int, evictionToleranceFraction float64, recommendationProcessor vpa_api_util.RecommendationProcessor, evictionAdmission priority.PodEvictionAdmission, selectorFetcher target.VpaTargetSelectorFetcher) (Updater, error) {
+func NewUpdater(kubeClient kube_client.Interface, vpaClient *vpa_clientset.Clientset, minReplicasForEvicition int, evictionRateLimit float64, evictionRateLimitBurst int, evictionToleranceFraction float64, recommendationProcessor vpa_api_util.RecommendationProcessor, evictionAdmission priority.PodEvictionAdmission, selectorFetcher target.VpaTargetSelectorFetcher) (Updater, error) {
+	evictionRateLimiter := getRateLimiter(evictionRateLimit, evictionRateLimitBurst)
```
bskiba: I think it can be simplified to evicitonRateBurst

vertical-pod-autoscaler/pkg/updater/main.go
```diff
@@ -45,6 +46,12 @@ var (
 	evictionToleranceFraction = flag.Float64("eviction-tolerance", 0.5,
 		`Fraction of replica count that can be evicted for update, if more than one pod can be evicted.`)
 
+	evictionRateLimit = flag.Float64("eviction-rate-limit", -1, `
+		Number of pods that can be evicted per seconds.`)
```
bskiba: Please add info that 0 and -1 turn rate limiting off. Also, I think the ` should be on the next line.

vertical-pod-autoscaler/pkg/updater/main.go
```diff
@@ -45,6 +46,12 @@ var (
...
+
+	evictionRateLimitBurst = flag.Int("eviction-rate-limit-burst", 1,
```
bskiba: Same comment on name as above.

vertical-pod-autoscaler/pkg/updater/logic/updater.go
bskiba: Also, is the Burst completely ignored.

vertical-pod-autoscaler/pkg/updater/logic/updater_test.go
```diff
@@ -153,9 +160,27 @@ func TestRunOnceNotingToProcess(t *testing.T) {
...
+
+func TestGetRateLimiter(t *testing.T) {
+	cases := []struct {
+		rateLimit       float64
+		rateLimitBurst  int
+		expectedLimiter *rate.Limiter
+	}{
+		{0.0, 1, rate.NewLimiter(rate.Inf, 1)},
+		{-1.0, 2, rate.NewLimiter(rate.Inf, 2)},
+		{10.0, 3, rate.NewLimiter(rate.Every(time.Duration(1.0/10*float64(time.Second))), 3)},
```
bskiba: From @bandesz's comment above, looks like this can be simplified.

vertical-pod-autoscaler/pkg/updater/logic/updater.go
bskiba: The default tick for the updater is 1 minute, so as @guillaumebreton writes, the Wait will be canceled after that time at most (unless someone uses a custom tick). So we can basically look at info that is stale up to 1 minute, which is also true if updater has a lot of work to do. I'm inclined to agree this solution is acceptable. 

As to randomizing the list of pods to evict - the updater has a priority mechanism to select which pods to evict first based on multiple factors (including how far away they are from their recommended resources). Shuffling the list wouldn't play well with that.

bskiba:
Thanks for your patience and apologies for the delay.

I've added some comments. My main concern is to make sure that the current behavior stays the same (i.e. if we are not using the rate limiter). Other comments are mostly nit's

vertical-pod-autoscaler/pkg/updater/logic/updater.go
guillaumebreton: Indeed, it makes more sense 👍 I fixed it.

vertical-pod-autoscaler/pkg/updater/logic/updater.go
guillaumebreton: Indeed, if we set the rate limite to `rate.Inf` the burst rate is completely ignored, and Wait will return immediately. 
-> Documentation about burst being ignored if rate equals to Rate.inf https://github.com/golang/time/blob/master/rate/rate.go#L37
-> Wait() returning immediately : https://github.com/golang/time/blob/master/rate/rate.go#L307

vertical-pod-autoscaler/pkg/updater/logic/updater.go
guillaumebreton: 👍 I missed that thank you @bandesz

vertical-pod-autoscaler/pkg/updater/main.go
guillaumebreton: 👍 Done

vertical-pod-autoscaler/pkg/updater/main.go
```diff
@@ -76,13 +84,15 @@ func main() {
...
+	updater, err := updater.NewUpdater(kubeClient, vpaClient, *minReplicas, *evictionRateLimit, *evictionRateBurst, *evictionToleranceFraction, vpa_api_util.NewCappingRecommendationProcessor(limitRangeCalculator), nil, targetSelectorFetcher)
 	if err != nil {
 		klog.Fatalf("Failed to create updater: %v", err)
 	}
 	ticker := time.Tick(*updaterInterval)
 	for range ticker {
-		updater.RunOnce()
+		ctx, cancel := context.WithTimeout(context.Background(), *updaterInterval)
+		updater.RunOnce(ctx)
+		cancel()
```
bskiba: I think it might be useful to run defer cancel() before updater.RunOnce() just to be on the safe side if RunOnce panics.

vertical-pod-autoscaler/pkg/updater/main.go
```diff
@@ -45,6 +46,13 @@ var (
 	evictionToleranceFraction = flag.Float64("eviction-tolerance", 0.5,
 		`Fraction of replica count that can be evicted for update, if more than one pod can be evicted.`)
 
+	evictionRateLimit = flag.Float64("eviction-rate-limit", -1,
+		`Number of pods that can be evicted per seconds. A rate limit set to 0 or -1 will disabled
```
bskiba: nit: s/disabled/disable/

vertical-pod-autoscaler/pkg/updater/main.go
```diff
@@ -45,6 +46,13 @@ var (
...
+		the rate limiter.`)
+
+	evictionRateBurst = flag.Int("eviction-rate-limit-burst", 1,
+		`Burst of pods that can be evicted.`)
```
bskiba: nit: Does this fit one line now?

vertical-pod-autoscaler/pkg/updater/logic/updater.go
```diff
@@ -155,6 +165,19 @@ func (u *updater) RunOnce() {
 	timer.ObserveTotal()
 }
 
+func getRateLimiter(evictionRateLimit float64, evictionRateLimitBurst int) *rate.Limiter {
+	var evictionRateLimiter *rate.Limiter
+	if evictionRateLimit == -1 || evictionRateLimit == 0 {
+		// As a special case if the rate is set to rate.Inf, the burst rate is ignored
+		// see https://github.com/golang/time/blob/master/rate/rate.go#L37
+		evictionRateLimiter = rate.NewLimiter(rate.Inf, 0)
+		klog.Info("Rate limit disabled")
```
bskiba: klog.V(1).Info("Eviction rate limit disabled")

vertical-pod-autoscaler/pkg/updater/logic/updater.go
```diff
@@ -155,6 +165,19 @@ func (u *updater) RunOnce() {
 	timer.ObserveTotal()
 }
 
+func getRateLimiter(evictionRateLimit float64, evictionRateLimitBurst int) *rate.Limiter {
+	var evictionRateLimiter *rate.Limiter
+	if evictionRateLimit == -1 || evictionRateLimit == 0 {
```
bskiba: What happens if evictionRateLimit is < 0?

vertical-pod-autoscaler/pkg/updater/logic/updater.go
```diff
@@ -17,9 +17,11 @@ limitations under the License.
 package logic
 
 import (
+	"context"
 	"fmt"
 	"time"
 
+	"golang.org/x/time/rate"
```
bskiba: nit: Can you add one blank line after this import?

vertical-pod-autoscaler/pkg/updater/logic/updater.go
guillaumebreton: I changed the condition 👍

guillaumebreton:
@bskiba  What would be the next step with this PR ? Should I ask @jbartosik and @mwielgus for a review ?

bskiba:
I'll review tomorrow, sorry for the wait!

guillaumebreton:
@bskiba No worries :) Github was showing you already approved it so I wasn't sure

guillaumebreton:
@bskiba No worries :) Github was showing you already approved it so I wasn't sure

guillaumebreton:
@bskiba Thank you 👍 I missed it and I fixed it.

bskiba:
Thanks!
/lgtm
/approve

PR #2339, opened 2019-09-16
Adding DOCKERFILE variable in Makefile

saiyam1814:
Signed-off-by: Saiyam Pathak <saiyam911@gmail.com>

This is wrt the pull request: https://github.com/kubernetes/autoscaler/pull/2059

Creating a Dockerfile variable in Makefile
If you want the builder Docekrfile or the Cluster autoscaler Docker file to be a custome one like a multi-stage build Dockerfile or some tests added Dockerfile then you would be able to pass onto as variables while doinf a make. Since you cannot change the Existing Dockerfile as its getting used in Googl'es production workload so the makefile changes would help here.

### DIFF
--- a/cluster-autoscaler/Makefile
+++ b/cluster-autoscaler/Makefile
@@ -1,6 +1,6 @@
 all: build
 
-DOCKERFILE?=../builder
+BUILDER_DOCKERFILE_DIR?=../builder
 TAG?=dev
 FLAGS=
 LDFLAGS?=-s
@@ -59,7 +59,7 @@ format:
     test -z "$$(find . -path ./vendor -prune -type f -o -name '*.go' -exec gofmt -s -w {} + | tee /dev/stderr)"
 
 docker-builder:
-	docker build -t autoscaling-builder ${DOCKERFILE}
+	docker build -t autoscaling-builder ${BUILDER_DOCKERFILE_DIR}
 
 build-in-docker: clean docker-builder
 	docker run -v `pwd`:/gopath/src/k8s.io/autoscaler/cluster-autoscaler/ autoscaling-builder:latest bash -c 'cd /gopath/src/k8s.io/autoscaler/cluster-autoscaler && BUILD_TAGS=${BUILD_TAGS} LDFLAGS="${LDFLAGS}" make build-binary'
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/Makefile
```diff
@@ -1,5 +1,6 @@
 all: build
 
+DOCKERFILE?=../builder
```
losipiuk: Please rename to BUILDER_DOCKERFILE_DIR.

saiyam1814: renamed

saiyam1814:
Yes @losipiuk Ideally what I wanted from the previous issue is a separate Dockerfile but its used officially but Google in their prod workloads so things might get impacted. So As I understood your suggestion was that what is we can make the Makefile understand the Custom Dockerfile. So this change would allow users to have their custome Dockerfile for the builder.

losipiuk:
> Yes @losipiuk Ideally what I wanted from the previous issue is a separate Dockerfile but its used officially but Google in their prod workloads so things might get impacted. So As I understood your suggestion was that what is we can make the Makefile understand the Custom Dockerfile. So this change would allow users to have their custome Dockerfile for the builder.

The issue here is that there are **two** dockerfiles involved in the CA build process. 
There is builder docker file. This is for building image which is then used for running build process (this is not dockerfile used as a source of image which contains CA binary). The image build with this docker file contains make, go compiler etc. Generally stuff needed to _build_ CA. Not CA binary itself.
It is referenced in the makefile here:
https://github.com/kubernetes/autoscaler/blob/7fee6108b439b2e6ef1409b80d3a731c2af6d87c/cluster-autoscaler/Makefile#L60
And this is dockerfile your PR allows to override.

There is second docker file which is source for actual CA image. It is referenced here in the Makefile:
https://github.com/kubernetes/autoscaler/blob/7fee6108b439b2e6ef1409b80d3a731c2af6d87c/cluster-autoscaler/Makefile#L40
(as well as 2 lines below).
The file is here:
https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/Dockerfile

Didn't you want to override the second one?

saiyam1814:
ohh I see, YEs should have been a variable like 
CA_DOCKERFILE_DIR
Also shuold I remove the other variable BUILDER_DOCKERFILE_DIR or keep that as well ? 
@losipiuk

saiyam1814:
Closing after discussion with @losipiuk

PR #2344, opened 2019-09-16
Explicitly set SideEffects=None for mutating webhook

### LINKED ISSUE #2278: Consider adding sideEffects to MutatingWebhookConfiguration
Since 1.13, k8s api has supported server side dry run, making it so a user can test the validity of an update, including the execution of webhooks (https://kubernetes.io/docs/reference/using-api/api-concepts/#make-a-dry-run-request). 

However, if webhooks are not explicit about whether they are "safe" to run in dry run mode, the dry run fails, which is what is currently happening with the VPA webhook:
```
LMASSA-MAC:~ luke$ cat pod.yaml 
apiVersion: v1
kind: Pod
metadata:
  name: pod-name
  namespace: sandbox
spec:
  containers:
  - name: foo
    image: centos:7
LMASSA-MAC:~ luke$ kubectl apply -f pod.yaml --server-dry-run
Error from server (InternalError): error when creating "pod.yaml": Internal error occurred: admission webhook "vpa.k8s.io" does not support dry run
```

`sideEffects` is currently set to `Unknown` (the default value) for `vpa-webhook-config`
```
LMASSA-MAC:~ luke$ kubectl get mutatingwebhookconfiguration  vpa-webhook-config  -o yaml | grep sideEffect
  sideEffects: Unknown
```

If (as I suspect) there are no side-effects to the VPA, I propose `sideEffects` be set to `None` here: https://github.com/kubernetes/autoscaler/blob/249aeab201bc97a8307f2f87d7e74ed16bfd15f7/vertical-pod-autoscaler/pkg/admission-controller/config.go#L82

If there are side-effects, then the webhook should respect the `dryrun` flag on the `AdmissionReview`, then set `sideEffects` to `NoneOnDryRun`.

lukemassa:
This closes #2278.

I believe that there are no SideEffects of the `vpa.k8s.io` controller, so I am setting that explicitly, instead of taking the default of `None`.

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/config.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/config.go
@@ -74,6 +74,7 @@ func selfRegistration(clientset *kubernetes.Clientset, caCert []byte, namespace 
 	} else {
 		RegisterClientConfig.URL = &url
 	}
+	sideEffects := v1beta1.SideEffectClassNone
 	RegisterClientConfig.CABundle = caCert
 	webhookConfig := &v1beta1.MutatingWebhookConfiguration{
 		ObjectMeta: metav1.ObjectMeta{
@@ -101,6 +102,7 @@ func selfRegistration(clientset *kubernetes.Clientset, caCert []byte, namespace 
 					},
 				},
 				ClientConfig: RegisterClientConfig,
+				SideEffects:  &sideEffects,
 			},
 		},
 	}


lukemassa:
Testing my CLA...

jbartosik:
Please fix tests

lukemassa:
Ah, my apologies, marking as WIP as I investigate.

lukemassa:
Sorry again, checks are passing now.

PR #2351, opened 2019-09-17
Vertical Pod Autoscaler version 0.7.0

bskiba:


bskiba:
Because there is a new version of the API in this one

PR #2356, opened 2019-09-18
Update golang sdk in 1.13 branch to use iam service account

rchernobelskiy:
More info: https://github.com/kubernetes/autoscaler/pull/2323
Thanks for looking into it, really appreciate the great work you guys do.


losipiuk:
@rchernobelskiy I have mixed feelings about that. It is very massive change for a bugfix. Also if we now followed to standard process of updating vendor as described in FAQ (which we probably will not do on 1.13 branch) we would overwrite your changes. Another thing is why the delta is that huge? Do we need all the SDK?

@MaciekPytel PTAL - I would appreciate second opinion here.

rchernobelskiy:
@losipiuk Thanks for looking into it. I figured you'd want the whole repo for the checksum to match. I can remove all the `_test` files, of which there are about 100. That would decrease the size of the PR significantly. Would that be of interest?

rchernobelskiy:
Fwiw, I'm building and running this locally for now and it works well.

rchernobelskiy:
Friendly reminder for people using this: `AWS_SDK_LOAD_CONFIG` var must be set to `"true"`

Jeffwan:
AWS_SDK_LOAD_CONFIG has been supported by using latest Session api. #2428

losipiuk:
/lgtm
/approve
Thanks!

PR #2367, opened 2019-09-19
Increase default GCE operation wait timeout to 20s

aleksandra-malinowska:
Increase default GCE operation wait timeout to 20s.


losipiuk:
Please fix tests

PR #2379, opened 2019-09-24
Bugfix - Allow arbitrary placeholder AWS instance names

### LINKED ISSUE #2378: [BUG] Cluster-Autoscaler - Fails to delete AWS placeholder instances
## What's not working as expected?

Due to my suggestion in #2235 of modifying the construction of placeholder IDs generated when ASGs have a difference between their desired and in-service count placeholders can now fail to be deleted when spot bids are unfulfilled depending on the naming format of the users ASGs.

I.e. we use the ASG naming format `<instance-type>.<cluster-identifier>.foo.bar.local` internally, this contains `.` and as such doesn't match the existing regex to check for valid node IDs.

This means that placeholder instances can fail to match the regex performed by the AWS cloudprovider when attempting to delete the placeholder node: 
https://github.com/kubernetes/autoscaler/blob/649ffcce3249d5db9027cc726097302a5496cff2/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go#L159

## What's the proposed fix?

Allow the `validAwsRefIdRegex` check to also pass if the node instead contains `i-placeholder` to allow for users to name their ASGs anything.

I'll raise the PR for this this evening.

gjtempleton:
Resolves #2378

Pulls in the `placeholderInstanceNamePrefix` to allow for arbitrary naming of ASGs

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
@@ -156,7 +156,7 @@ type AwsInstanceRef struct {
 	Name       string
 }
 
-var validAwsRefIdRegex = regexp.MustCompile(`^aws\:\/\/\/[-0-9a-z]*\/[-0-9a-z]*$`)
+var validAwsRefIdRegex = regexp.MustCompile(fmt.Sprintf(`^aws\:\/\/\/[-0-9a-z]*\/[-0-9a-z]*$|aws\:\/\/\/[-0-9a-z]*\/%s.*$`, placeholderInstanceNamePrefix))
 
 // AwsRefFromProviderId creates InstanceConfig object from provider id which
 // must be in format: aws:///zone/name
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -156,7 +156,7 @@ type AwsInstanceRef struct {
 	Name       string
 }
 
-var validAwsRefIdRegex = regexp.MustCompile(`^aws\:\/\/\/[-0-9a-z]*\/[-0-9a-z]*$`)
+var validAwsRefIdRegex = regexp.MustCompile(fmt.Sprintf(`^aws\:\/\/\/[-0-9a-z]*\/[-0-9a-z]*$|aws\:\/\/\/[-0-9a-z]*\/%s.*$`, placeholderInstanceNamePrefix))
```
Jeffwan: em.. Seems we miss ref validation for this name. Thanks for the fix!

gjtempleton:
Thanks for the super quick review! 

I'll raise cherry-picks back to 1.13 to fix this in other releases.

Jeffwan:
@gjtempleton Let's create a list to track items need to be cherry-picked to previous versions and we can do once before very release

gjtempleton:
Good idea. Thanks again for the quick reviews on this.

PR #2385, opened 2019-09-25
ensure valid AWS LaunchTemplate version

### LINKED ISSUE #1728: Continue operation if parsing some ASG fails
One of our ASGs uses mixed instance types and has 0 instances.

Expected Behaviour:
Autoscaler behaves like described in: https://github.com/kubernetes/autoscaler/issues/1647.
The autoscaler fails to parse the mixed ASG with
`Unable to build proper template node for {{ASG_NAME}}: Unable to get instance type from launch config or launch template`
 and continues with the existing others.

Actual Behaviour:
The autoscaler fails to parse the mixed ASG with
`Unable to build proper template node for {{ASG_NAME}}: Unable to get instance type from launch config or launch template` and stops processing ASGs with the following message:
`static_autoscaler.go:312] Failed to scale up: failed to build node infos for node groups: Unable to get instance type from launch config or launch template`

I just had to remove the ASG and everything was fine, but a single invalid ASG shouldn't incapacitate the Autoscaler IMO.

jaypipes:
The LaunchTemplateSpecification.Version is a pointer to
string. When the pointer is nil, EC2 AutoScaling API considers the value
to be "$Default", however aws.StringValue(ltSpec.Version) will return an
empty string (which is not considered the same as "$Default" or a nil
string pointer. So, in order to not pass an empty string as the version
for the launch template when we communicate with the EC2 AutoScaling API
using the information in the launchTemplate, we store the string
"$Default" when the ltSpec.Version is a nil pointer.

Fixes #1728

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
+++ b/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
@@ -461,9 +461,32 @@ func (m *asgCache) buildAsgFromAWS(g *autoscaling.Group) (*asg, error) {
 }
 
 func (m *asgCache) buildLaunchTemplateFromSpec(ltSpec *autoscaling.LaunchTemplateSpecification) *launchTemplate {
+	// NOTE(jaypipes): The LaunchTemplateSpecification.Version is a pointer to
+	// string. When the pointer is nil, EC2 AutoScaling API considers the value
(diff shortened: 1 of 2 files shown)


Jeffwan:
Great! Thanks for fixing this case.

Jeffwan:
This should be a candidate to cherry-pick to >= 1.14 versions. We can do it later before CA release next version

PR #2395, opened 2019-09-27
Allow for overriding VPA container default arguments

bskiba:


bskiba:
This ports needed changes to the 0.6 branch

PR #2407, opened 2019-10-01
Extract separate package for core/utils.go and split into multiple files.

losipiuk:
This PR extracts uitls.go from core package to separate utils package. The motivation for that is the size of core package. Is humongous which hinders maintenance. Expect subsequent refactors in the area.


cluster-autoscaler/core/utils/taint_key_set.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	schedulerapi "k8s.io/kubernetes/pkg/scheduler/api"
+)
+
+// TaintKeySet is a set of taint key
+type TaintKeySet map[string]bool
+
+var (
+	// NodeConditionTaints lists taint keys used as node conditions
+	NodeConditionTaints = TaintKeySet{
```
aleksandra-malinowska: nit: does it need to be exported? It's never used outside the package

aleksandra-malinowska:
Two thoughts (non-blocking):
1. Not sure what is the reason for moving some used-in-one-place-and-very-specific-functions into utils (e.g. filterOutExpendableAndSplit) while moving some such functions from utils (e.g. fixNodeGroupSize), i.e. what are the criteria for where each of them should live. But as there wasn't much order before, I don't mind either way.
2. This PR is very large, with one commit changing 1.7k lines of code total. It's one thing to sanity-check where the functions move, but it's very hard to verify that they weren't modified, and that the net difference (~100 loc) is only due to moving code around (imports, boilerplate etc.) In the future, can we have smaller PRs?

/lgtm
/approve

PR #2427, opened 2019-10-06
doks: use providerID for node.ID

snormore:
## Problem

The `cluster-autoscaler` uses `node.Spec.ProviderID` to determine [which nodes are unregistered](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/clusterstate/clusterstate.go#L963-L980), so that they can be [eventually deleted](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/static_autoscaler.go#L259-L274) in an attempt to replace by registered instances. The `node.Id` that we assign in cluster-autoscaler is the DOKS node UUID, so the nodes always look unregistered to the autoscaler, because they're using different IDs. This results in nodes flapping in and out of not ready as they're [deleted via k8s](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/static_autoscaler.go#L259-L274) by the cluster-autoscaler.

## Solution

Derive the cluster-autoscaler `node.Id` from the node droplet ID and in the providerID form that it expects (i.e. `digitalocean://12345678`), instead of the DOKS node UUID.

This also upgrades the vendored `godo` package.

cc @timoreimann @andrewsykim


cluster-autoscaler/cloudprovider/digitalocean/digitalocean_node_group.go
```diff
@@ -233,9 +234,9 @@ func (n *NodeGroup) Autoprovisioned() bool {
 // toInstances converts a slice of *godo.KubernetesNode to
 // cloudprovider.Instance
 func toInstances(nodes []*godo.KubernetesNode) []cloudprovider.Instance {
-	instances := make([]cloudprovider.Instance, len(nodes))
-	for i, nd := range nodes {
-		instances[i] = toInstance(nd)
+	instances := []cloudprovider.Instance{}
```
timoreimann: FWIW, you could use

```go
instances := make([]cloudprovider.Instance, 0, len(nodes))
```

to allocate the slice at the right size while also being able to use `append`.

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_cloud_provider.go
```diff
@@ -70,13 +71,8 @@ func (d *digitaloceanCloudProvider) NodeGroups() []cloudprovider.NodeGroup {
...
 func (d *digitaloceanCloudProvider) NodeGroupForNode(node *apiv1.Node) (cloudprovider.NodeGroup, error) {
-	nodeID, ok := node.Labels[nodeIDLabel]
-	if !ok {
-		// CA creates fake node objects to represent upcoming VMs that haven't
-		// registered as nodes yet. They have node.Spec.ProviderID set. Use
-		// that as nodeID.
-		nodeID = node.Spec.ProviderID
-	}
+	providerID := node.Spec.ProviderID
+	nodeID := strings.TrimPrefix(providerID, doProviderIDPrefix)
```
timoreimann: Since this PR adds a `toProviderID` function already, it might be nice to also add a `toDropletID` (or `toNodeID`?) function for consistency and encapsulation reasons.

PR #2358, opened 2019-09-18
Update golang sdk in 1.14 branch to use iam service account

rchernobelskiy:

PR #2428, opened 2019-10-07
AWS – use `session.NewSession` instead of `session.New` to

devkid:
`session.New` is deprecated and requires the `AWS_SDK_LOAD_CONFIG`
environment variable to be set in order to automatically call
`AssumeRoleWithWebIdentity` when `AWS_WEB_IDENTITY_TOKEN_FILE` is set
(which is not documented and most likely unintended).


devkid:
Original bug report: https://github.com/aws/aws-sdk-go/issues/2828
Issue with cluster-autoscaler: https://github.com/kubernetes/autoscaler/pull/2358#issuecomment-537599252

Jeffwan:
@devkid Thanks! Did you see any error when you get session?

Jeffwan:
Thanks for the change. This would be an easy way no matter `session.New` will be fixed or not. `NewSession` assumes IRSA role which is expected. 

/lgtm
/approve

PR #2432, opened 2019-10-08
Simplify building equivalence groups in scale up

aleksandra-malinowska:
Minor refactor to separate pod equivalence computation from running predicates.

Possible TODO for improving messages in events: pass equivalence groups to scale up status processor, rather than build a full NoScaleUpInfo per pod.

### DIFF
--- a/cluster-autoscaler/core/scale_up.go
+++ b/cluster-autoscaler/core/scale_up.go
@@ -19,10 +19,11 @@ package core
 import (
 	"bytes"
 	"fmt"
-	"k8s.io/autoscaler/cluster-autoscaler/core/utils"
 	"math"
 	"strings"
 	"time"
+
+	"k8s.io/autoscaler/cluster-autoscaler/core/utils"
 
 	appsv1 "k8s.io/api/apps/v1"
 	apiv1 "k8s.io/api/core/v1"
@@ -265,11 +266,8 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 
 	loggingQuota := glogx.PodsLoggingQuota()
 
-	podsRemainUnschedulable := make(map[*apiv1.Pod]map[string]status.Reasons)
-
 	for _, pod := range unschedulablePods {
 		glogx.V(1).UpTo(loggingQuota).Infof("Pod %s/%s is unschedulable", pod.Namespace, pod.Name)
-		podsRemainUnschedulable[pod] = make(map[string]status.Reasons)
 	}
 	glogx.V(1).Over(loggingQuota).Infof("%v other pods are also unschedulable", -loggingQuota.Left())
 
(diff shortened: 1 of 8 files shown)


cluster-autoscaler/core/scale_up.go
```diff
@@ -374,42 +369,27 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
...
+			samplePod := eg.pods[0]
+			if err := context.PredicateChecker.CheckPredicates(samplePod, nil, nodeInfo); err == nil {
+				// add pods to option
+				option.Pods = append(option.Pods, eg.pods...)
+				// mark pod group as (theoretically) schedulable
+				eg.schedulable = true
+			} else {
+				klog.V(2).Infof("Pod %s can't be scheduled on %s, predicate failed: %v", samplePod.Name, nodeGroup.Id(), err.VerboseError())
+				if podCount := len(eg.pods); podCount > 0 {
+					klog.V(2).Infof("%d other pods similar to %s can't be scheduled on %s", podCount, samplePod.Name, nodeGroup.Id())
```
losipiuk: podCount - 1

cluster-autoscaler/core/scale_up.go
```diff
@@ -374,42 +369,27 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
...
-			_, found := podsRemainUnschedulable[pod]
-			if found {
-				podsRemainUnschedulable[pod][nodeGroup.Id()] = err
+		for _, eg := range podEquivalenceGroups {
+			samplePod := eg.pods[0]
+			if err := context.PredicateChecker.CheckPredicates(samplePod, nil, nodeInfo); err == nil {
+				// add pods to option
+				option.Pods = append(option.Pods, eg.pods...)
+				// mark pod group as (theoretically) schedulable
+				eg.schedulable = true
```
losipiuk: This will not mutate the struct which is in the `podEquivalenceGroups` slice but the copy used for iteration.

Also I would prefer that `equivalenceGroup` to just be struct with groupId and slice of pods. And that we keep track of whether given group is schedulable or not and failed predicates in separate data struct. Pod equivalence regarding scheduling properties and information whether given equivalence group is schedulable on given node shape is a separate concern.

losipiuk: If I am not mistaken - and there is a bug I described - can we add test which would catch that?

aleksandra-malinowska: Good catch! Added more assertions to existing tests (that would catch it) and also a new test for scenario where some pods trigger scale up and some don't.

cluster-autoscaler/core/scale_up.go
aleksandra-malinowska: Fixed

losipiuk:
/lgtm
/approve
Thanks!

PR #2435, opened 2019-10-09
Priority expander 1.14 backport

mgalgs:
This is a backport of the `priority` expander to the 1.14 release branch.

Relevant PRs:
- #1801
- #1889
- #1920

---

## Test results

I was able to run this build on my cluster with other expanders, and there doesn't appear to be any regressions.  ~Unfortunately, I couldn't actually get the priority expander to work.  I'm not sure if the priority expander itself is busted (I know it's relatively new) or if my backport just sucks... @piontec any ideas on troubleshooting?  I do see a bunch of unschedulable pods but no scale-up event.~

Edit: it's working now!  Turns out I had my `cluster-autoscaler-priority-expander` ConfigMap in the wrong namespace :man_facepalming:


mgalgs:
> It may take a couple minutes for the CLA signature to be fully registered; after that, please reply here with a new comment and we'll verify.

Should be good to go (I guess I committed using a different email address than usual)

losipiuk:
This is very much not a bugfix and IMO should not be subject to cherry-pick to older release. 1.15 if feature is needed I suggest migrating to newer k8s version.

mgalgs:
@losipiuk EKS doesn't support k8s 1.15 yet, so my hands are kind of tied...  Didn't realize that release branches were bugfix-only...  Upgrading k8s versions isn't always simple...

Jeffwan:
@mgalgs Hi, Mitchel, do you get a chance to update git username and email to yours. I think original author change the email and account and that's the reason CLA check complain missing Github user.  Łukasz  will help backport this change. Trying to see if you can help on that. If I don't hear back from you, I will file a PR with correct Github user instead as we want to release today or tomorrow.

mgalgs:
> If I don't hear back from you, I will file a PR with correct Github user instead as we want to release today or tomorrow.

Your PR LGTM :+1: Will go ahead and close this guy. Thanks!

PR #2436, opened 2019-10-10
Skip acceptable range check before it has data

### LINKED ISSUE #2250: Failed to find acceptable ranges for node groups
Master branch. Need to investigate this issue.

```
I0812 01:08:34.619043   86316 static_autoscaler.go:191] Starting main loop
W0812 01:08:34.619861   86316 clusterstate.go:437] Failed to find acceptable ranges for eksctl-ca-testing-nodegroup-ondemand-NodeGroup-1C3FP32CDSL2M
W0812 01:08:34.621255   86316 clusterstate.go:437] Failed to find acceptable ranges for eksctl-ca-testing-nodegroup-spot-NodeGroup-1IKFMEITZKT7T
I0812 01:08:34.621565   86316 utils.go:619] No pod using affinity / antiaffinity found in cluster, disabling affinity predicate for this loop
I0812 01:08:34.621737   86316 filter_out_schedulable.go:65] Filtering out schedulables
I0812 01:08:34.622118   86316 filter_out_schedulable.go:130] 0 other pods marked as unschedulable can be scheduled.
I0812 01:08:34.622246   86316 filter_out_schedulable.go:130] 0 other pods marked as unschedulable can be scheduled.

```

Jeffwan:
Resolve #2250 

https://github.com/kubernetes/autoscaler/blob/a258103f8ec9c24a333338468fa51262df0ef24a/cluster-autoscaler/clusterstate/clusterstate.go#L307-L326

`csr.getCloudProviderNodeInstances()` is called before `updateAcceptableRanges`  and at that time, acceptable range doesn't have any data, current error message is misleading. 

Please have a check and let me know if it makes sense.

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -432,6 +432,11 @@ func (csr *ClusterStateRegistry) IsNodeGroupSafeToScaleUp(nodeGroup cloudprovide
 }
 
 func (csr *ClusterStateRegistry) getProvisionedAndTargetSizesForNodeGroup(nodeGroupName string) (provisioned, target int, ok bool) {
+	if len(csr.acceptableRanges) == 0 {
+		klog.Warningf("AcceptableRanges have not been populated yet. Skip checking")
+		return 0, 0, false
+	}
+
 	acceptable, found := csr.acceptableRanges[nodeGroupName]
 	if !found {
 		klog.Warningf("Failed to find acceptable ranges for %v", nodeGroupName)


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -432,6 +432,11 @@ func (csr *ClusterStateRegistry) IsNodeGroupSafeToScaleUp(nodeGroup cloudprovide
 }
 
 func (csr *ClusterStateRegistry) getProvisionedAndTargetSizesForNodeGroup(nodeGroupName string) (provisioned, target int, ok bool) {
+	if len(csr.acceptableRanges) == 0 {
```
losipiuk: Withe the change you will get N lines in logs saying `AcceptableRanges have not been populated yet. Skip checking` instead of N lines saying `Failed to find acceptable ranges for %v"` for different groups. 

It is hard for me to see a benefit here.

Jeffwan: Right, I agree it will repeat for N node groups. I think people just get confused if they see "Failed to find acceptable ranges for %v" in the logs and it indicates something is wrong. If we know we may call an uninitialized collection, I think it's better to add checks.

siwyd: I can vouch on being confused by seeing those messages and looking them up... :)

losipiuk: Fine. Let's merge it the. I personally do not see that it helps. But it does not break anything neither.

losipiuk:
/lgtm
/approve
Thanks!

PR #2439, opened 2019-10-10
track the scale down duration metric correctly for non-empty node case

### LINKED ISSUE #2425: nodeDeletionDuration metric is broken for non empty nodes
It seems that the metric calculation for the non-empty nodes case is broken. 

https://github.com/kubernetes/autoscaler/blob/3f0a5fa3c21df7f14846e742a91d3137e3791e7e/cluster-autoscaler/core/scale_down.go#L843-L846

I'm expecting that this metric should be updated to reflect the time after the go routine completes.

This seems to be tracked correctly for the empty nodes case.

marwanad:
It seemed that the node deletion duration has been broken for the non-empty node case since the node drainage/deletion was refactored into a goroutine.
 
Fixes: #2425

### DIFF
--- a/cluster-autoscaler/core/scale_down.go
+++ b/cluster-autoscaler/core/scale_down.go
@@ -692,7 +692,8 @@ func (sd *ScaleDown) TryToScaleDown(allNodes []*apiv1.Node, pods []*apiv1.Pod, p
 	scaleDownStatus := &status.ScaleDownStatus{NodeDeleteResults: sd.nodeDeletionTracker.GetAndClearNodeDeleteResults()}
 	nodeDeletionDuration := time.Duration(0)
 	findNodesToRemoveDuration := time.Duration(0)
-	defer updateScaleDownMetrics(time.Now(), &findNodesToRemoveDuration, &nodeDeletionDuration)
+	scaleDownFnStartTime := time.Now()
+	defer updateScaleDownMetrics(scaleDownFnStartTime, &findNodesToRemoveDuration, &nodeDeletionDuration)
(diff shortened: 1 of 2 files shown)


marwanad:
cc: @feiskyer

cluster-autoscaler/core/scale_down.go
```diff
@@ -692,7 +692,10 @@ func (sd *ScaleDown) TryToScaleDown(allNodes []*apiv1.Node, pods []*apiv1.Pod, p
 	scaleDownStatus := &status.ScaleDownStatus{NodeDeleteResults: sd.nodeDeletionTracker.GetAndClearNodeDeleteResults()}
 	nodeDeletionDuration := time.Duration(0)
 	findNodesToRemoveDuration := time.Duration(0)
-	defer updateScaleDownMetrics(time.Now(), &findNodesToRemoveDuration, &nodeDeletionDuration)
+	scaleDownFnStartTime := time.Now()
+	metrics.UpdateLastTime(metrics.ScaleDown, scaleDownFnStartTime)
+	defer updateScaleDownMetrics(scaleDownFnStartTime, &findNodesToRemoveDuration, &nodeDeletionDuration)
```
losipiuk: It seems that now we can be calling `updateScaleDownMetrics` twice per call to `TryToScaleDown`.

marwanad: good catch, that's a problem. 

Apart from updating the metrics at every return path, is there anything we can go to have an accurate `metrics.ScaleDownNodeDeletion`?

PR #2454, opened 2019-10-15
fix the webhook-address and webhook-port reference for vpa-admission-…

calixwu:
Just found this little bug when I try to deploy the vpa-admission-controller outside the cluster... Change ``webhook-address`` and ``webhook-port`` to  ``webhook-address`` and ``*webhook-port`` will work.


bskiba:
Thanks!
Looks like your commit is missing a GitHub user. Would you mind fixing it?
Otherwise the CLA check will fail and we will be unable to merge this.

PR #2464, opened 2019-10-17
Cherry pick #2428 to 1.13

Jeffwan:
#2428 AWS – use session.NewSession instead of session.New to


Jeffwan:
@losipiuk Thanks for the waiting. Last piece to 1.13.

PR #2448, opened 2019-10-13
cluster-autoscaler: fix some typos in the code

hectorj2f:
I found some typos while surfing across the code to understand what it does each component.

PR #2470, opened 2019-10-19
fix misspell failures

FayerZhang:


Jeffwan:
BTW, seems #2448 fix some of the typos. if everything is covered, let's work on the one open first.

PR #2474, opened 2019-10-22
Improve performance of the Azure cloud provider

DavidLangworthy:
Previously, the cloud provider called Get to obtain the status for each node pool.  When running with a large number of node pools, e.g. 10+, these will cause Azure to throttle, not only the autoscaler, but all services operating under the same identity.

This commit replaces the Get call for each node pool with a List call which obtains the status for all the node pools at once.  The frequency of these calls is set with the vmssSizeRefreshPeriod constant.

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_cloud_provider_test.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_cloud_provider_test.go
@@ -29,6 +29,8 @@ import (
 )
 
 func newTestAzureManager(t *testing.T) *AzureManager {
+	vmssName := "test-asg"
+	var vmssCapacity int64 = 3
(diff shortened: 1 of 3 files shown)


DavidLangworthy:
@feiskyer I have updated the expected values and am passing all tests but one.  This is TestDeleteNodes.  It overwrites the FakeStore.  How should I deal with this?  Also, are the expected values changes alright?

mboersma:
@feiskyer tests are passing now, PTAL when time permits.

feiskyer:
@DavidLangworthy would you like to cherry pick to stable branches?

DavidLangworthy:
@feiskyer Yes, I would like this in 1.16 and 1.15.  Should I just reissue this PR against those branches?

PR #2475, opened 2019-10-23
cluster-autoscaler/gce: create a test for an instance with unknown mig

JulienBalestra:
This PR follows #2281 

@losipiuk I created the test in `gce_manager_test.go` instead of `gce_cloud_provider_test.go` because there is no coverage of this file because of the mocks.

Would you prefer a dedicated test for this ?


losipiuk:
> Would you prefer a dedicated test for this ?

That is fine. Thanks a lot!

losipiuk:
/lgtm
/approve
Thanks!

PR #2479, opened 2019-10-24
VPA: handle quick OOMs in all containers

aermakov-zalando:
For some reason, the quick OOM handling logic in VPA is currently restricted to pods with just one container. This makes VPA unable to handle crashes in pods that use sidecars, so the user has to delete the pod manually to get the proper recommendations. This PR changes the logic to handle quick OOM crashes in all containers.

### DIFF
--- a/vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go
+++ b/vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go
@@ -86,13 +86,14 @@ func (calc *UpdatePriorityCalculator) AddPod(pod *apiv1.Pod, recommendation *vpa
 	updatePriority := calc.getUpdatePriority(pod, processedRecommendation)
 
 	quickOOM := false
-	if len(pod.Status.ContainerStatuses) == 1 {
-		terminationState := pod.Status.ContainerStatuses[0].LastTerminationState
+	for _, cs := range pod.Status.ContainerStatuses {
+		terminationState := cs.LastTerminationState
 		if terminationState.Terminated != nil &&
 			terminationState.Terminated.Reason == "OOMKilled" &&
 			terminationState.Terminated.FinishedAt.Time.Sub(terminationState.Terminated.StartedAt.Time) < *evictAfterOOMThreshold {
 			quickOOM = true
-			klog.V(2).Infof("quick OOM detected in pod %v", pod.Name)
+			klog.V(2).Infof("quick OOM detected in pod %v, container %s", pod.Name, cs.Name)
+			break
 		}
 	}
 
--- a/vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator_test.go
+++ b/vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator_test.go
@@ -31,7 +31,8 @@ import (
 )
 
 const (
-	containerName = "container1"
+	containerName       = "container1"
+	secondContainerName = "container2"
 )
 
 func TestSortPriority(t *testing.T) {
@@ -303,17 +304,24 @@ func TestDontUpdatePodWithOOMAfterLongRun(t *testing.T) {
 	assert.Exactly(t, []*apiv1.Pod{}, result, "Pod shouldn't be updated")
 }
 
-func TestDontUpdatePodWithOOMOnlyOnOneContainer(t *testing.T) {
+func TestUpdatePodWithOOMAnyContainer(t *testing.T) {
 	calculator := NewUpdatePriorityCalculator(
 		nil, nil, &UpdateConfig{MinChangePriority: 0.5}, &test.FakeRecommendationProcessor{})
 
-	pod := test.Pod().WithName("POD1").AddContainer(test.BuildTestContainer(containerName, "4", "")).Get()
+	pod := test.Pod().WithName("POD1").
+		AddContainer(test.BuildTestContainer(containerName, "4", "")).
+		AddContainer(test.BuildTestContainer(secondContainerName, "4", "")).
+		Get()
 
 	// Pretend that the test pod started 11 hours ago.
 	timestampNow := pod.Status.StartTime.Time.Add(time.Hour * 11)
 
 	pod.Status.ContainerStatuses = []apiv1.ContainerStatus{
 		{
+			Name: containerName,
+		},
+		{
+			Name: secondContainerName,
 			LastTerminationState: apiv1.ContainerState{
 				Terminated: &apiv1.ContainerStateTerminated{
 					Reason:     "OOMKilled",
@@ -322,18 +330,25 @@ func TestDontUpdatePodWithOOMOnlyOnOneContainer(t *testing.T) {
 				},
 			},
 		},
-		{},
 	}
 
 	// Pod is within the recommended range.
-	recommendation := test.Recommendation().WithContainer(containerName).
-		WithTarget("5", "").
-		WithLowerBound("1", "").
-		WithUpperBound("6", "").Get()
+	recommendation := &vpa_types.RecommendedPodResources{
+		ContainerRecommendations: []vpa_types.RecommendedContainerResources{
+			test.Recommendation().WithContainer(containerName).
+				WithTarget("5", "").
+				WithLowerBound("1", "").
+				WithUpperBound("6", "").GetContainerResources(),
+			test.Recommendation().WithContainer(secondContainerName).
+				WithTarget("5", "").
+				WithLowerBound("1", "").
+				WithUpperBound("6", "").GetContainerResources(),
+		},
+	}
 
 	calculator.AddPod(pod, recommendation, timestampNow)
 	result := calculator.GetSortedPods(NewDefaultPodEvictionAdmission())
-	assert.Exactly(t, []*apiv1.Pod{}, result, "Pod shouldn't be updated")
+	assert.Exactly(t, []*apiv1.Pod{pod}, result, "Pod should be updated")
 }
 
 func TestNoPods(t *testing.T) {


vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go
```diff
@@ -86,13 +86,14 @@ func (calc *UpdatePriorityCalculator) AddPod(pod *apiv1.Pod, recommendation *vpa
 	updatePriority := calc.getUpdatePriority(pod, processedRecommendation)
 
 	quickOOM := false
-	if len(pod.Status.ContainerStatuses) == 1 {
-		terminationState := pod.Status.ContainerStatuses[0].LastTerminationState
+	for _, cs := range pod.Status.ContainerStatuses {
```
bskiba: There is a possibility to switch off autoscaling for a given container, we probably don't want to react to ooms in those.

bskiba:
@schylek I think you were working on the quick oom feature, do you remember anything blocking us for doing this for all autoscaled containers if pods have multiple?

schylek:
We cannot say which container is the main application. It is possible that
*some* users wants the whole pod to be restarted, but other users may want
the main app to be up despite crashing sidecar. The point is that there is
no safe choice in that case.

aermakov-zalando:
@schylek Would you at least accept a PR if I change it to only handle containers for which autoscaling isn't disabled?

arjunrn:
@schylek so how do we proceed? do we add a configuration field to the VPA which decides the behavior. I would argue that when even the sidecar is crashing the main container will be affected. So in most of the cases it does make sense to evict. But I see your point and we would like to contribute a fix.

schylek:
Thank you offering help. Some sidecars are safe to crash and others are
required and there is no way to specify that.

What is your exact setup? I understand that the main app is autoscaled and
what about the sidecar? Is it a well known sidecar or something built
in-house?

I would start with a fix to check if the list *autoscaled* containers has
size of 1 as a guard for current quickOOM logic. Then VPA could react to
crashes of the main app and just ignore sidecars (which would have constant
resources). Would it help your case?

For having both the main app and the sidecar autoscaled probably we would
need a configuration knob, but I can imagine different combinations: list
of sidecar containers, name of the main app container, quick OOM policy
(current behavior, react to OOMs in all containers). I don't have a settled
option what would be the best choice here.

mikkeloscar:
@schylek I don't understand why the default is to assume that people want *some* "sidecars" which should not force the pod to be recreated in case of OOMKill. Presumably if you are using the VPA you are fine with your pod being recreated. If you are using Kubernetes in the first place you should be ok with that, so saying that by default it should be opt-in to have containers considered for OOMKIll handling seems like the wrong optimization to me. It should be opt-out rather than opt-in.

We only have pods where all the containers work together to provide the full functionality, I can't think of a case where you would be fine with one container crashing all the time because you don't want to kill another container in the pod to fix both of them.

Right now it's impossible to use the VPA for a pod with more than one container in the case of OOMKill handling.

The case where it breaks is simple. You have a "main" container and one "sidecar". The "sidecar" works fine, but your "main" container is OOMKilled and the VPA doesn't do anything to recover your "main" container, which it would do if you didn't have a "sidecar".

schylek:
Sample of a sidecar that should crash without disturbing the main app is
some log/metrics exporter. I can imagine users being upset for "main" being
evicted when some non-critical sidecar crashes. My point is that both
scenarios - sidecars crucial for functionality and sidecars that are "best
effort" are probable. In the current implementation the quick path suffers
when there are >1 containers, but eventually the recommendation should be
applied. I agree that it will be slower, but should be functional - am I
missing any problem that makes impossible to use VPA for such workload?

I understand your case. The root of the problem is that today VPA cannot
tell which container is "main" and "sidecar" and what is the class of the
sidecar. I think that fix for checking if len(the list of autoscaled
containers) == 1 is easy to implement right now, since it reduces the
problem to autoscaling a single application. For multiple autoscaled
containers it is more complicated. IMO changing the default behavior is
risky (some users may rely on current behavior) and introducing extra knobs
require some more thought. I would also consider a list of well known
sidecars that are crucial for functionality (e.g. service mesh proxies).

mikkeloscar:
I don't see what's the point of a "best-effort" container. If you have something that exports metrics/logs why would you be ok with this thing not running? I don't believe this is an actual production use case.
Further more, if your sidecar doesn't run your pod will not be ready so if your main container is supposed to get traffic from a service then it won't work anyway. The whole semantic of a Pod in Kubernetes is that it can go away so we shouldn't try to optimize for keeping a single pod (or a single container within a pod) running at all times. If a pod is to be evicted because "some" container is OOMKilled then this is not a problem in Kubernetes. The VPA respects PodDisruptionBudgets and Kubernetes will send the SIGTERM to the running containers so they can shut down safely.

Like @bskiba pointed out, it makes sense to not do the `OOMKilled` check for containers which are switched off for autoscaling. This is what I would consider opt-out rather than opt-in.

aermakov-zalando:
@schylek My use case (which prompted me to do this PR) was an application that had a sidecar responsible for authentication which was setup in a way where the main container was the second one. Since the quick OOM handling is the only way to fix a pod that has a memory spike immediately after startup, one of the administrators had to get paged for something that should've been handled automatically.

I still don't understand your point in general, though. If I, as a user, have any number of containers in a pod for which I _want_ to use autoscaling, and I've specified that VPA can safely evict my pod to change whatever it needs, why would I want the autoscaler to let my pod in a broken state? What purpose does it even serve? How do you think the recommendations would get applied "eventually"? Just by waiting for weeks until the node is terminated so the pod is recreated?

schylek:
I am aware of scenarios when crashes of a "best effort" sidecar due to
memory leak were causing pages. So setups like that are IMO valid and yes
your setup is valid too. VPA should support both cases and there is no way
to configure it today. The general guideline was that if VPA is not sure of
some decisions it should act slower to not make things worse. Regarding
slower actuation of recommendations - the recommendation has to grow beyond
current recommendation range to be applied by updater, it shouldn't wait
for some external events like node being terminated.

So to summarize:
1) We all agree that in case of multiple containers only autoscaled ones
should be considered for quick oom path, exactly what @bskiba suggested.
2) For multiple autoscaled containers you propose to change current
behavior from "let's not actuate changes if we are unaware of the user
intention" to "let's treat all autoscaled containers required for pod"
3) It seems that VPA should support both cases for "best effort" and
"required" autoscaled containers but there is no way for VPA to be aware
which type of setup the user has.

1 can be implemented right now. IMO implementing 2 without addressing 3 may
just break someone else without giving them the option to opt-out. Maybe
there are not many users like that, but there is no way to say.

Let's have @mwielgus (sig autoscaling lead) involved in this discussion.

mikkeloscar:
@schylek thanks for the summary.

I still don't understand how point 3 is different from what @bskiba suggests

> 3) It seems that VPA should support both cases for "best effort" and
"required" autoscaled containers but there is no way for VPA to be aware
which type of setup the user has.

If you can express that a "best-effort" sidecar should not be autoscaled then there would be a way for the VPA to be aware of what setup the user has expressed.

You're ofc. right that if we change this now, then the default behavior will change and users who relied on VPA not doing anything for pods with multiple containers may see it try to fix a pod by recreating it.
I would still argue that this is ok to break that "expectation" because the behavior is not intuitive as it is now.

You mentioned this situation

> I am aware of scenarios when crashes of a "best effort" sidecar due to
memory leak were causing pages.

I would argue that the root cause here is not that a "best effort" sidecar caused a page. But that the sidecar was leaking memory to begin with. :)

schylek:
> If you can express that a "best-effort" sidecar should not be
autoscaled then there would be a way for the VPA to be aware of what setup
the user has expressed.

There is a way to express that a container shouldn't be autoscaled by
specifying ContainerResourcePolicy.Mode
https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2/types.go#L134
and
AFAIU this what was suggested by @bskiba. So what I propose is the
following:
- for quick oom fix current logic of "1 container" to "1 autoscaled
container", where non-autoscaled containers are filtered out using
ContainerResourcePolicy.Mode
- then users can disable autoscaling for sidecars (to have quickOOM enabled
on the main app) and the logic for having multiple autoscaled containers is
not changed.
This is what I meant in point 1.

Regarding point 3 - it about a way to configure how VPA should react to OOM
events on a container if there multiple autoscaled containers. IMO this
requires more thought to specify how to express that.

aermakov-zalando:
@schylek Changing the logic to "only 1 autoscaling container" doesn't cover any of the applications we run with sidecars, IIRC, so this would make the PR useless for our use case. I still don't understand why a separate knob is needed for distinguishing between the cases where VPA is allowed to evict a pod when a container's memory usage needs to be adjusted normally and when it needs to be adjusted due to an OOM kill. From the user's perspective these two cases aren't different, and I'd argue that scaling the container when needed is something that the users would expect by default without setting yet another weird configuration item.

mikkeloscar:
I also don't understand the point of treating the OOMKilled handling and general autoscaling as two separate things that need different configuration.
For me as a user it's the same thing. Either a container should be considered for autoscaling (which includes OOMKill handling) or it should not be autoscaled at all.

schylek:
Still my point of view is that this is a change of current behavior and
should be accepted by @bskiba or @mwielgus.

szuecs:
@bskiba @mwielgus can we have an update on the decision, please?

schylek:
I had a chat with @bskiba IRL and she will accept the change. Just make sure only to observe autoscaled containers.

bskiba:
Yup, sorry to be late to the party here, thanks to everyone for the good discussion :+1:, I support adding quick oom logic for all autoscaled containers in the pod.

bskiba:
#2739 adresses this (quick ooms in all autoscaled containers)
/close

PR #2480, opened 2019-10-24
[aws] use simple string prefix placeholder match

jaypipes:
Addresses a [comment on PR2235](https://github.com/kubernetes/autoscaler/pull/2235/files#r313536379)
to remove the unnecessary use of a regex with a simpler string prefix
matching for placeholder names.


Jeffwan:
/cc @seh @gjtempleton

As you either changed or left comments on placeholder regex, please help review.

Issue #2453, opened 2019-10-15
Support `custom Daemonset` Pods


clamoriniere:
The Daemonset pod detection is currently done in the code by looking at the `OwnerRef.Kind` that reference ether a Daemonset, ReplicaSet, StatefulSet.

Internally at Datadog, we implemented thanks to CRD a custom Daemonset that aims to fix several issues that we have with the current Daemonset implementation. 
Like the official Daemonset controller, our CustomDaemonset controller is creating and assigning on each Node a Pod, but in our case the OwnerRef.Kind is not `Daemonset` but `CustomDaemonset`, which has the consequence that the `Autoscaler` considers those pods as `normal` pods, and tries to control them for the cluster scaling and also the bin packing.

To allow the support of our use case, we are open to participating in the project. We see two possible implementations:

1) Add new configuration flag to provide several ApiGroup/Kind to treat as Daemonsets
2) Introduce a new Annotation on Pod to allow the Autoscaler to consider the Pod as a Daemonset pod. `cluster-autoscaler.kubernetes.io/is-daemonset-pod: "true"` (annotation name TBD)

### 1) Add new configuration flag to provide several `Daemonset` ApiGroup/Kind
* Advantages:
   - Users don’t need to update their PodTemplate with a new annotation.
* Cons:
   - Need to know all ApiGroup/Kind that needs to be considered as Daemonset.
   - Requires to change the configuration of the Autoscaler deployment, which is not always possible in managed Kubernetes clusters like GKE.

### 2) Introduce a new Annotation on Pod to make the Autoscaler consider a Pod as a Daemonset pod.
* Advantages:
   - Better genericity compares to solution 1.
   - No Autoscaler configuration change needed. That allows better support for managed Kubernetes clusters.
* Cons:
   - The user needs to add this annotation on each PodTemplate/Pod that is used as Daemonset.

Solution 2 is our preferred solution because it opens the support of more use cases. Also, it seems aligned with other use cases logic already implemented in the Autoscaler like:
* Annotation for detecting Static Pods
* Annotation for ignoring pods in the eviction logic: `cluster-autoscaler.kubernetes.io/safe-to-evict: "true" `

Currently with fix it with the following workaround: https://github.com/kubernetes/autoscaler/commit/25754c89d4d6cdbc84a1e5bf9f5e37b1e97a39ef

Feel free to comment on this issue, if you want more information about our use case, if you see corner cases or you think of another possible solution.

losipiuk:
[...]

PR #2483, opened 2019-10-25
[cluster-autoscaler] Allow “custom” DaemonSet pods

clamoriniere:
Fixes: #2453

* Create `utils/pod` package: pod kind detection functions.
* Update DaemonSet Pod detection logic: introduce the annotation
  `cluster-autoscaler.kubernetes.io/daemonset-pod` to declare a Pod as a
  DaemonSet Pod.
* Update `simulator` package to use the new `utils/pod` package functions.
* Cleanup `getRequiredPodsForNode()` function.

Signed-off-by: cedric lamoriniere <cedric.lamoriniere@datadoghq.com>

### DIFF
--- a/cluster-autoscaler/simulator/nodes.go
+++ b/cluster-autoscaler/simulator/nodes.go
@@ -17,13 +17,11 @@ limitations under the License.
 package simulator
 
 import (
-	"time"
-
 	apiv1 "k8s.io/api/core/v1"
-	policyv1 "k8s.io/api/policy/v1beta1"
-	"k8s.io/autoscaler/cluster-autoscaler/utils/drain"
-	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
 	schedulernodeinfo "k8s.io/kubernetes/pkg/scheduler/nodeinfo"
+
+	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
+	pod_util "k8s.io/autoscaler/cluster-autoscaler/utils/pod"
 )
 
 // getRequiredPodsForNode returns a list of pods that would appear on the node if the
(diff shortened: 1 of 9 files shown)


clamoriniere:
Hi @losipiuk @seh could you have a look to this PR? it follows the discussion that we add in  #2453
cheers,

cluster-autoscaler/simulator/nodes.go
```diff
@@ -75,3 +45,20 @@ func BuildNodeInfoForNode(node *apiv1.Node, podsForNodes map[string][]*apiv1.Pod
...
+func filterRequiredPodsForNode(allPods []*apiv1.Pod) []*apiv1.Pod {
+	var selectedPods []*apiv1.Pod
+
+	for id, pod := range allPods {
+		// Ignore pod in deletion phase
+		if pod.DeletionTimestamp != nil {
+			continue
+		}
+
+		if pod_util.IsMirrorPod(pod) || pod_util.IsDaemonSetPod(pod) || pod_util.IsStaticPod(pod) {
```
losipiuk: IIUC the mirror pod is manifestation of static pod registered in API server. Static pod is observed only by kubelet so we should not get a static pod ever here. Right?

https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/

cluster-autoscaler/utils/pod/pod.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// IsDaemonSetPod returns true if the Pod should be considered as Pod managed by a DaemonSet
+func IsDaemonSetPod(pod *apiv1.Pod) bool {
+	controllerRef := metav1.GetControllerOf(pod)
+	if controllerRef != nil && controllerRef.Kind == "DaemonSet" {
+		return true
+	}
+
+	if pod.Annotations == nil {
```
losipiuk: I think you do not need this. Looking up in "nil" map should work fine.

cluster-autoscaler/simulator/nodes.go
```diff
@@ -17,50 +17,20 @@ limitations under the License.
...
 
 // getRequiredPodsForNode returns a list of pods that would appear on the node if the
 // node was just created (like daemonset and manifest-run pods). It reuses kubectl
 // drain command to get the list.
 func getRequiredPodsForNode(nodename string, podsForNodes map[string][]*apiv1.Pod) ([]*apiv1.Pod, errors.AutoscalerError) {
 	allPods := podsForNodes[nodename]
-	podsToRemoveList, err := drain.GetPodsForDeletionOnNodeDrain(
-		allPods,
-		[]*policyv1.PodDisruptionBudget{}, // PDBs are irrelevant when considering new node.
-		true,                              // Force all removals.
```
losipiuk: It seems we can drop the `deleteAll` parameter from `GetPodsForDeletionOnNodeDrain`

cluster-autoscaler/simulator/nodes.go
clamoriniere: absolutely

cluster-autoscaler/simulator/nodes.go
clamoriniere: right, it is always used with `false`

clamoriniere:
@losipiuk Thanks for the fast review.

Commits are now squashed.

losipiuk:
/lgtm
/approve
Thanks!

Issue #2447, opened 2019-10-11
VerticalPodAutoscaler not compatible with k8s 1.16.x libraries


tariq1890:
We, in kube-state-metrics use the VPA and K8s client libaries as dependencies. We seem to be blocked on updating to K8s 1.16.x libraries as the VPA dependencies fail to compile when importing the latest K8s deps.

As seen on this thread: https://github.com/kubernetes/kube-state-metrics/pull/945#issuecomment-540711969

PR #2485, opened 2019-10-25
[WIP] VPA: bring k8s dependencies to v1.16.1

wozniakjan:
The dependencies in VPA have been out of date for some time. This PR is trying to migrate `go dep` to `go mod` and upgrading k8s dependencies to v1.16.1.

This is branch is being used in https://github.com/kubernetes/kube-state-metrics/pull/949 as an interim solution to test that PR but before merging, it would be ideal to point the dependencies back to this repo.

closes: https://github.com/kubernetes/autoscaler/issues/2447


bskiba:
:hand: Looks like we worked on this simultaneously :) I'm working on getting #2486 merged. Will it will cover your needs as well?

PR #2490, opened 2019-10-27
Add tests for Packet

d-mo:
Test cases include scaling up a node pool and then scaling it back down. They run against a mock API by default, or against the actual Packet API if a valid PACKET_AUTH_TOKEN is set as an environmental variable.


losipiuk:
/lgtm
/approve
Thanks!

PR #2493, opened 2019-10-29
Add support for passing custom ignored labels

adammw:
Add support for passing custom ignored labels to the node group set comparator. This way end-users can add labels that they would like the autoscaler to ignore when considering two node groups the same for balancing between as a command line option, for example if they have their own readiness labels or other markings that should not affect balancing behaviour.

### DIFF
--- a/cluster-autoscaler/config/autoscaling_options.go
+++ b/cluster-autoscaler/config/autoscaling_options.go
@@ -135,6 +135,9 @@ type AutoscalingOptions struct {
 	MaxBulkSoftTaintTime time.Duration
 	// IgnoredTaints is a list of taints to ignore when considering a node template for scheduling.
 	IgnoredTaints []string
+	// BalancingExtraIgnoredLabels is a list of labels to additionally ignore when comparing if two node groups are similar.
+	// Labels in BasicIgnoredLabels and the cloud provider-specific ignored labels are always ignored.
+	BalancingExtraIgnoredLabels []string
 	// AWSUseStaticInstanceList tells if AWS cloud provider use static instance type list or dynamically fetch from remote APIs.
 	AWSUseStaticInstanceList bool
 	// Path to kube configuration if available
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -293,10 +293,10 @@ func buildAutoscaler() (core.Autoscaler, error) {
 	processors.PodListProcessor = core.NewFilterOutSchedulablePodListProcessor()
 	if autoscalingOptions.CloudProviderName == cloudprovider.AzureProviderName {
 		processors.NodeGroupSetProcessor = &nodegroupset.BalancingNodeGroupSetProcessor{
-			Comparator: nodegroupset.IsAzureNodeInfoSimilar}
+			Comparator: nodegroupset.CreateAzureNodeInfoComparator()}
 	} else if autoscalingOptions.CloudProviderName == cloudprovider.AwsProviderName {
 		processors.NodeGroupSetProcessor = &nodegroupset.BalancingNodeGroupSetProcessor{
-			Comparator: nodegroupset.IsAwsNodeInfoSimilar}
+			Comparator: nodegroupset.CreateAwsNodeInfoComparator()}
 	}
 
 	opts := core.AutoscalerOptions{
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -168,9 +168,10 @@ var (
 	regional                      = flag.Bool("regional", false, "Cluster is regional.")
 	newPodScaleUpDelay            = flag.Duration("new-pod-scale-up-delay", 0*time.Second, "Pods less than this old will not be considered for scale-up.")
 
-	ignoreTaintsFlag         = multiStringFlag("ignore-taint", "Specifies a taint to ignore in node templates when considering to scale a node group")
-	awsUseStaticInstanceList = flag.Bool("aws-use-static-instance-list", false, "Should CA fetch instance types in runtime or use a static list. AWS only")
-	enableProfiling          = flag.Bool("profiling", false, "Is debug/pprof endpoint enabled")
+	ignoreTaintsFlag          = multiStringFlag("ignore-taint", "Specifies a taint to ignore in node templates when considering to scale a node group")
+	balancingIgnoreLabelsFlag = multiStringFlag("balancing-ignore-label", "Specifies a label to ignore in addition to the basic and cloud-provider set of labels when comparing if two node groups are similar")
+	awsUseStaticInstanceList  = flag.Bool("aws-use-static-instance-list", false, "Should CA fetch instance types in runtime or use a static list. AWS only")
+	enableProfiling           = flag.Bool("profiling", false, "Is debug/pprof endpoint enabled")
 )
 
 func createAutoscalingOptions() config.AutoscalingOptions {
@@ -235,6 +236,7 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 		Regional:                         *regional,
 		NewPodScaleUpDelay:               *newPodScaleUpDelay,
 		IgnoredTaints:                    *ignoreTaintsFlag,
+		BalancingExtraIgnoredLabels:      *balancingIgnoreLabelsFlag,
 		KubeConfigPath:                   *kubeConfigFile,
 		NodeDeletionDelayTimeout:         *nodeDeletionDelayTimeout,
 		AWSUseStaticInstanceList:         *awsUseStaticInstanceList,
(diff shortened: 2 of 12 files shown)


adammw:
Hi @aleksandra-malinowska please review

cluster-autoscaler/processors/nodegroupset/balancing_processor.go
```diff
@@ -46,6 +46,18 @@ func (b *BalancingNodeGroupSetProcessor) FindSimilarNodeGroups(context *context.
 			"failed to find template node for node group %s",
 			nodeGroupId)
 	}
+
+	ignoredLabels := BasicIgnoredLabels
```
losipiuk: Intiialize this to empty map and always copy context.IgnoredLabels  and BasicIgnoredLabels.
We do not want to modify global variable as we do now by setting cp-specific values in it to true.

cluster-autoscaler/main.go
```diff
@@ -171,6 +171,7 @@ var (
 			"Pods with nominatedNodeName set are always filtered out.")
 
 	ignoreTaintsFlag         = multiStringFlag("ignore-taint", "Specifies a taint to ignore in node templates when considering to scale a node group")
+	ignoreLabelsFlag         = multiStringFlag("ignore-labels", "Specifies a label to ignore when comparing if two node groups are similar")
```
losipiuk: State in help that this this list extends a set of basic and cloud provider specific labels which are always ignored.

cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -139,6 +139,8 @@ type AutoscalingOptions struct {
 	FilterOutSchedulablePodsUsesPacking bool
 	// IgnoredTaints is a list of taints to ignore when considering a node template for scheduling.
 	IgnoredTaints []string
+	// IgnoredLabels is a list of labels to ignore when comparing if two node groups are similar.
```
losipiuk: State in comment that this is just an extension. That basic and cp-specific labels are always ignored.

losipiuk:
@adammw are you planning to still work on this one?

adammw:
updated @losipiuk

cluster-autoscaler/processors/nodegroupset/aws_nodegroups.go
```diff
@@ -21,16 +21,16 @@ import (
 )
 
 // IsAwsNodeInfoSimilar adds AWS specific node labels to the list of ignored labels.
-func IsAwsNodeInfoSimilar(n1, n2 *schedulernodeinfo.NodeInfo) bool {
+func IsAwsNodeInfoSimilar(n1, n2 *schedulernodeinfo.NodeInfo, ignoredLabels map[string]bool) bool {
```
losipiuk: please do not modify input parameter.
Copy the map and extend with standard awsIngornedLabels before passing to `IsNodeInfoSimilar`

cluster-autoscaler/processors/nodegroupset/azure_nodegroups.go
```diff
@@ -32,14 +32,10 @@ func nodesFromSameAzureNodePool(n1, n2 *schedulernodeinfo.NodeInfo) bool {
 // IsAzureNodeInfoSimilar compares if two nodes should be considered part of the
 // same NodeGroupSet. This is true if they either belong to the same Azure agentpool
 // or match usual conditions checked by IsAzureNodeInfoSimilar, even if they have different agentpool labels.
-func IsAzureNodeInfoSimilar(n1, n2 *schedulernodeinfo.NodeInfo) bool {
+func IsAzureNodeInfoSimilar(n1, n2 *schedulernodeinfo.NodeInfo, ignoredLabels map[string]bool) bool {
```
losipiuk: same here.
Do not modify input parameters.

cluster-autoscaler/processors/nodegroupset/aws_nodegroups.go
```diff
@@ -21,16 +21,16 @@ import (
 )
 
 // IsAwsNodeInfoSimilar adds AWS specific node labels to the list of ignored labels.
```
losipiuk: As you are changing this part of code please also update the comment. It does not capture hat function actually does - that it actually compares nodes for similarity.

cluster-autoscaler/processors/nodegroupset/aws_nodegroups.go
losipiuk: Actually I have another idea. It does not seem best idea to pass `ignoredLabels` to each call of comparator.
The passed labels will be the same ever time. And each call we do the same merging of maps

I suggest following apporach:
 * replace static Is*NodeInfoSimilar functions with builders.
   * The builder would be parametrized with `extraIgnoredLabels` and would merge the `Basic`, `extra`, and `cloud-provider specific` ignored labels. 
   * The generated comparator would just take two node `*NodeInfo` parameters (no ignoredLabels map)
   * E.g For AWS we create `CreateAwsNodeInfoSimilarFunction() func(n1, n2 *NodeInfo) bool`. Similarly for Azure. The returned function would call out to standard `IsNodeInfoSimilar(n1, n2 *NodeInfo, ignoredLabels map[string]bool ignoredLabels)`.
   * If neither AWS nor Azure is used, then comparator would be created with `CreateStandardNodeInfoSimilarFunction` which would just merge `Basic` and `extra` ignored node labels and also return function which calls out to `IsNodeInfoSimilar(n1, n2, ignoredLabels)`

I think that approach would be cleaner. WDYT?

adammw: Yep. Currently rewriting this PR to use this approach.

adammw:
@losipiuk updated this PR to match your suggestion of having builder functions for the comparator method that creates the full ignore label map rather than passing the ignore labels in each time.

For posterity - previous code before the force push: 29fed41

hectorj2f:
@adammw The last build failed 😕

adammw:
Yep, although I can't work out why - it seems unrelated to my changes as it's a failure in vertical pod autoscaler when I only touched cluster autoscaler files.

bskiba:
@adammw I restarted the build for now (we'll see if this is a flake)

@jbartosik could you take a look at this failure? I think this is one of the tests you added (please correct me if I'm wrong). [Logs](https://travis-ci.org/kubernetes/autoscaler/jobs/653596328?utm_medium=notification&utm_source=github_status)

adammw:
@hectorj2f @bskiba build is passing now...

bskiba:
@adammw Yeah, sorry about the trouble, I filed #2886 for the flaky test. 
I think @losipiuk is the reviewer here?

uthark:
@losipiuk @hectorj2f @bskiba Any updates on it? Thank you.

cluster-autoscaler/processors/nodegroupset/balancing_processor.go
```diff
@@ -58,7 +58,7 @@ func (b *BalancingNodeGroupSetProcessor) FindSimilarNodeGroups(context *context.
 		}
 		comparator := b.Comparator
 		if comparator == nil {
-			comparator = IsNodeInfoSimilar
+			panic("BalancingNodeGroupSetProcessor comparator not set")
```
losipiuk: please replace for klog.Fatal(). It logs and then panics.

cluster-autoscaler/processors/nodegroupset/aws_nodegroups.go
```diff
@@ -20,8 +20,10 @@ import (
 	schedulernodeinfo "k8s.io/kubernetes/pkg/scheduler/nodeinfo"
 )
 
-// IsAwsNodeInfoSimilar adds AWS specific node labels to the list of ignored labels.
-func IsAwsNodeInfoSimilar(n1, n2 *schedulernodeinfo.NodeInfo) bool {
+// CreateAwsNodeInfoComparator returns a comparator that checks if two nodes should be considered
+// part of the same NodeGroupSet. This is true if they match usual conditions checked by IsCloudProviderNodeInfoSimilar,
+// even if they have different AWS-specific labels.
+func CreateAwsNodeInfoComparator(ignoredLabels []string) NodeInfoComparator {
```
losipiuk: Please rename parameter to "extraIgnoredLablels" to point that this are not the only one ignored (there as still `Basic` ones).

Same for other comparator implementations.

cluster-autoscaler/main.go
```diff
@@ -169,6 +169,7 @@ var (
 	newPodScaleUpDelay            = flag.Duration("new-pod-scale-up-delay", 0*time.Second, "Pods less than this old will not be considered for scale-up.")
 
 	ignoreTaintsFlag         = multiStringFlag("ignore-taint", "Specifies a taint to ignore in node templates when considering to scale a node group")
+	ignoreLabelsFlag         = multiStringFlag("ignore-labels", "Specifies a label to ignore in addition to the basic and cloud-provider set of labels when comparing if two node groups are similar")
```
losipiuk: name the flag `ignore-label`

cluster-autoscaler/main.go
```diff
@@ -235,6 +236,7 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 		Regional:                         *regional,
 		NewPodScaleUpDelay:               *newPodScaleUpDelay,
 		IgnoredTaints:                    *ignoreTaintsFlag,
+		IgnoredLabels:                    *ignoreLabelsFlag,
```
losipiuk: Change `IgnoredLabels` to `BalancingExtraIgnoredLabels`

adammw:
@losipiuk what is this repo's policy on backporting features like this? 

we're running k8s 1.16 (soon to be 1.17), would be nice to get them into one of the release branches rather than just `master`

SmilingNavern:
Hello,

Could please someone help me. What is release version for this feature?

PR #2496, opened 2019-10-30
Clarify VPA logging

jbartosik:


vertical-pod-autoscaler/e2e/v1/admission_controller.go
```diff
@@ -77,11 +77,13 @@ var _ = AdmissionControllerE2eDescribe("Admission-controller", func() {
...
 
-		// Originally Pods had 100m CPU, 100Mi of memory, but admission controller
-		// should change it to recommended 250m CPU and 200Mi of memory.
-		for _, pod := range podList.Items {
-			gomega.Expect(pod.Spec.Containers[0].Resources.Requests[apiv1.ResourceCPU]).To(gomega.Equal(ParseQuantityOrDie("250m")))
-			gomega.Expect(pod.Spec.Containers[0].Resources.Requests[apiv1.ResourceMemory]).To(gomega.Equal(ParseQuantityOrDie("200Mi")))
+		ginkgo.By("Verifying hamster deployment")
+		for i, pod := range podList.Items {
+			podInfo := fmt.Sprintf("pod at index %d", i)
+			cpuDescription := fmt.Sprintf("%s: originally Pods had 100m CPU, admission controller should change it to recommended 250m CPU", podInfo)
```
bskiba: Please make the values into variables, so that we don't have to modify multiple places.

vertical-pod-autoscaler/e2e/v1/admission_controller.go
```diff
@@ -77,11 +77,13 @@ var _ = AdmissionControllerE2eDescribe("Admission-controller", func() {
...
 		podList := startDeploymentPods(f, d)
 
-		// Originally Pods had 100m CPU, 100Mi of memory, but admission controller
-		// should change it to recommended 250m CPU and 200Mi of memory.
-		for _, pod := range podList.Items {
-			gomega.Expect(pod.Spec.Containers[0].Resources.Requests[apiv1.ResourceCPU]).To(gomega.Equal(ParseQuantityOrDie("250m")))
-			gomega.Expect(pod.Spec.Containers[0].Resources.Requests[apiv1.ResourceMemory]).To(gomega.Equal(ParseQuantityOrDie("200Mi")))
+		ginkgo.By("Verifying hamster deployment")
+		for i, pod := range podList.Items {
+			podInfo := fmt.Sprintf("pod at index %d", i)
```
bskiba: Maybe just pod.Name?

vertical-pod-autoscaler/e2e/v1/admission_controller.go
jbartosik: Done

bskiba:
Expected <sth> to equal <sth> seems quite clear to me (although it's the other way round than I remembered). I don't want to add too much there since the code gets cluttered IMO.

jbartosik:
> Expected to equal seems quite clear to me (although it's the other way round than I remembered). I don't want to add too much there since the code gets cluttered IMO.

Order isn't the only information missing from the log. It's also not saying what those values are (limit or request). I think we should log more details about the failure, looking it up in test code is painful.

I added another commit, reducing amount of cluter a bit while trying to preserve most of information

jbartosik:
/hold
Rebased to resolve merge conflicts, putting on hold while I rerun tests.

jbartosik:
/hold cancel

Tests seem to work after rebase

bskiba:
Thanks a lot, this looks really good
/lgtm
/apprive

PR #2509, opened 2019-11-04
Cherry pick #1963 to 1.13: add --ignore-taint flag and ignore taints added by TaintNodesByCondition

marwanad:
Add an `--ignore-taint` flag and ignore taints added by TaintNodesByCondition


losipiuk:
/lgtm
/approve
Thanks!

Not sure if we will still cut another 1.13 release though.

Issue #2449, opened 2019-10-13
Action required: Move to scheduler.New to instantiate a scheduler


ahg-g:
Scheduler "factory" package will be deprecated in k8s 1.17. To instantiate a scheduler in the [simulator](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/simulator/predicates.go#L129-L160), please use [scheduler.New()](https://github.com/kubernetes/kubernetes/blob/f0e1d394f13560cd6094722ec0b3969b05830fd0/pkg/scheduler/scheduler.go#L250) instead.

Please let me know if you hav any questions.

MaciekPytel:
I'm not sure if this works for us - scheduler.New() seems to do a whole lot more than just initialize informers. We explicitly don't use [scheduler cache](https://github.com/kubernetes/kubernetes/blob/f0e1d394f13560cd6094722ec0b3969b05830fd0/pkg/scheduler/scheduler.go#L263) and I'm pretty sure we don't want any [event handlers](https://github.com/kubernetes/kubernetes/blob/f0e1d394f13560cd6094722ec0b3969b05830fd0/pkg/scheduler/https://github.com/kubernetes/kubernetes/blob/f0e1d394f13560cd6094722ec0b3969b05830fd0/pkg/scheduler/https://github.com/kubernetes/kubernetes/blob/f0e1d394f13560cd6094722ec0b3969b05830fd0/pkg/scheduler/eventhandlers.go#L359) registered on informers.

The point is we need predicates to be stateless. They can't use any precomputations (unless we explicitly provide it), because we're possibly invalidating them between simulations (moving pods around, creating in-memory node objects, etc). None of those changes are reflected in informers, so the scheduler state won't be updated accordingly.
Can we have a version of this function that we can use to only pass informers and skip all the other initialization of scheduler?

ahg-g:
> I'm not sure if this works for us - scheduler.New() seems to do a whole lot more than just initialize informers. We explicitly don't use [scheduler cache](https://github.com/kubernetes/kubernetes/blob/f0e1d394f13560cd6094722ec0b3969b05830fd0/pkg/scheduler/scheduler.go#L263) 

[...]

PR #2515, opened 2019-11-08
Revendor CA against tip of k8s.io/kubernetes

losipiuk:
Preparatory for 1.17 release

/assign @MaciekPytel 

cc: @ahg-g (`Migrate scheduler creation to scheduler.New()` commit)

Fixes: #2449


ahg-g:
/cc @alculquicondor 

See new scheduler invocation code: https://github.com/kubernetes/autoscaler/pull/2515/files#diff-aa62a4a46f434572b43666b621f733efR116

cluster-autoscaler/simulator/predicates.go
```diff
@@ -261,7 +228,16 @@ func (p *PredicateChecker) GetPredicateMetadata(pod *apiv1.Pod, nodeInfos map[st
 	if !p.enableAffinityPredicate {
 		return nil
 	}
-	return p.predicateMetadataProducer(pod, nodeInfos)
+	nodeInfoList := make([]*schedulernodeinfo.NodeInfo, 0, len(nodeInfos))
```
alculquicondor: how is `nodeInfos` created? Maybe you should try to use `NewSnapshot` or provide your own implementation of `listers.SharedLister`
This usage of snapshot seems to be relying too much on internals.

losipiuk: Multiple versions of `nodeInfos` are created as part of scale-down simulation here: https://github.com/kubernetes/autoscaler/blob/dc64d0aab25202144cf655c8526d5d6c2d7e5f97/cluster-autoscaler/simulator/cluster.go#L221.

It does not seem feasible to migrate the code to `NewSnapshot`. If possible at all major refactor of the method would be needed. 

Do you think I should create our own implementation of `SharedLister`? It would be a copy of what you have. Yet I agree it will improve the separation of CA and scheduler.

cluster-autoscaler/simulator/predicates.go
```diff
@@ -111,53 +112,24 @@ func NewPredicateChecker(kubeClient kube_client.Interface, stop <-chan struct{})
...
-		DisablePreemption:              false,
-		PercentageOfNodesToScore:       schedulerapi.DefaultPercentageOfNodesToScore,
-		BindTimeoutSeconds:             scheduler.BindTimeoutSeconds,
-	})
-	// Create the config from a named algorithm provider.
-	config, err := configurator.CreateFromProvider(algorithmProvider)
-	if err != nil {
-		return nil, fmt.Errorf("couldn't create scheduler using provider %q: %v", algorithmProvider, err)
+	defaultProviderName := schedulerconfig.SchedulerDefaultProviderName
+	algorithmSource := schedulerconfig.SchedulerAlgorithmSource{
```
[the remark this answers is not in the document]
ahg-g: sgtm, can you send a PR?

cluster-autoscaler/simulator/predicates.go
ahg-g: middle ground: if CA wants to maintain the ```NodeInfoMap```, then we can have a ```NewSnapshotFromNodeInfoMap``` func to create the snapshot. 

Note that we do a little more than just create the ```NodeInfoList```, but also a ```HavePodsWithAffinityNodeInfoList```:  https://github.com/kubernetes/kubernetes/blob/bcb171b375df0dd9326bf5d3a23428beea1aeb89/pkg/scheduler/nodeinfo/snapshot/snapshot.go#L52

ahg-g: Once https://github.com/kubernetes/kubernetes/pull/85139 is merged, you should be able to use NewSnapshot.

cluster-autoscaler/simulator/predicates.go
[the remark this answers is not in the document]
alculquicondor: Created #85140

cluster-autoscaler/simulator/predicates.go
losipiuk: Thanks! Updated the PR.

ahg-g: wait for this one as well https://github.com/kubernetes/kubernetes/pull/85151 , should allow you to not specify algorithm source explicitly

losipiuk:
Updated

PR #2520, opened 2019-11-12
remove temporary nodes logic

vivekbagade:


MaciekPytel:
Shouldn't you also remove https://github.com/kubernetes/autoscaler/blob/8d667b5a5aa4623a2377077e43809bb53c178c77/cluster-autoscaler/core/scale_down.go#L873?

vivekbagade:
Done

Issue #2507, opened 2019-11-01
VPA: go modules reference local files


wozniakjan:
https://github.com/kubernetes/autoscaler/pull/2486 has added support for go modules but the modules file references locally checked out repositories that may not be present on other machines

https://github.com/kubernetes/autoscaler/blob/086999e99674db5b92240356846bd6dcfab30e2f/vertical-pod-autoscaler/go.mod#L297-L325

The recommended hack scripts to update dependencies also seem to be missing from the repository
https://github.com/kubernetes/autoscaler/blob/086999e99674db5b92240356846bd6dcfab30e2f/vertical-pod-autoscaler/go.mod#L1-L3

cc: @bskiba

bskiba:
#2486 Does not unfortunately fully add support for go modules. TL;DR: we use go mod only to construct the vendor directory, based on kubernetes/kubernetes dependencies. 

Long version:
VPA (and Cluster Autoscaler as well) has an issue where we depend on both kubernetes/kubernetes repo and a bunch of stuff that is synced from kubernetes/staging repo to other repos. This makes managing dependencies really hard.

What we do is we copy the go.mod file from kubernetes/kubernetes repo and then fix all the dependencies to what is in the staging directory. Then we use go mod to construct the vendor directory. So the go.mod file is not fully valid. See https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/update_deps.sh

I'd like to both document the process better and work on making this setup less ugly, but it will take some time. If you have ideas, let me know. 
The scripts mentioned in the auto-generated files will not work with this setup unfortunately, but I don't know if I can make go mod skip generation of that part.

wozniakjan:
wondering if we can reuse any ideas from https://github.com/kubernetes/kubernetes/issues/79384 or in the worst case create a staging dir here as well to move the absolute paths `/usr/local/google/home/bskiba` from go.mod to relative so others can collaborate and update the dependencies?

I quite like and prefer the script mentioned in https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-521493597 which based on depending kubernetes/kubernetes version figures out the sha1 of those staged repos. That should transform all `replace xyz => /absolute/path/xyz` directives to `replace xyz => xyz v0.0.0-date-sha` which sounds to me like more portable and reproducible environment. What do you think? 

[...]

PR #2530, opened 2019-11-13
[WIP] VPA: update go modules

wozniakjan:
Effort to simplify go modules tracking removing the requirement for replace directives with filesystem absolute paths.

If possible, downstream projects would prefer something along the lines of https://github.com/kubernetes/autoscaler/pull/2531 to remove dependency on `k8s.io/kubernetes`

Closes: https://github.com/kubernetes/autoscaler/issues/2507

TODO:
- [ ] add script to handle modules automatically


wozniakjan:
closing in favor of https://github.com/kubernetes/autoscaler/pull/2531

PR #2544, opened 2019-11-19
vpa-recommender: Use Prometheus query_range endpoint for history

### LINKED ISSUE #2508: VPA recommender prometheus history provider - cannot get timeseries for cpu: invalid response status: success or type: vector
Prometheus history provider isn't working on VPA 0.6.0+ because the prometheus client expects a matrix for the cpu data but it's a vector as the cpu query now uses rate() as introduced in #1824.

`E1104 04:23:55.334996       1 cluster_feeder.go:213] Cannot get cluster history: cannot get usage history: cannot get timeseries for cpu: invalid response status: success or type: vector
`

milesbxf:
Fixes #2508 .

To populate historical data, we'd like to retrieve a set of data points over time from Prometheus. The canonical Prometheus method for doing this is the /api/v1/query_range endpoint, which means "evaluate the given query at the given interval/step between the start and the end time".

This change updates the Prometheus history provider to use this endpoint instead. We keep the --history-length flag, but use it to determine the start and end time of the query. We're also adding a new --history-resolution flag to control the step parameter.


milesbxf:
I have a new PR coming that refactors the history provider to use the Prometheus client, so closing in favour of that

PR #2545, opened 2019-11-19
Read subscription ID from instance metadata

fiunchinho:
The same way that we don't have to explicitly pass credentials as arguments when using managed identity extension, we could allow people to avoid passing the `subscriptionID` if they don't want to, and fetch it from the instance metadata.

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_manager.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_manager.go
@@ -26,9 +26,11 @@ import (
 
 	"github.com/Azure/go-autorest/autorest/azure"
 	"gopkg.in/gcfg.v1"
+	"k8s.io/klog"
+	azure2 "k8s.io/legacy-cloud-providers/azure"
+
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
 	"k8s.io/autoscaler/cluster-autoscaler/config/dynamic"
-	"k8s.io/klog"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/cloudprovider/azure/azure_manager.go
```diff
@@ -26,9 +26,11 @@ import (
 
 	"github.com/Azure/go-autorest/autorest/azure"
 	"gopkg.in/gcfg.v1"
+	"k8s.io/klog"
+	azure2 "k8s.io/legacy-cloud-providers/azure"
```
feiskyer: looks good to me in general. Could we rename azure2 to providerazure?

feiskyer: Also, please group k8s.io packages togather

fiunchinho: Fixed. I also added a note to explain the new behavior. Please take a look.

fiunchinho:
@feiskyer the build has failed, but I can't build the project locally. It seems like there is a problem with an outdated dependency? Can you help me with this?

fiunchinho:
I believe I fixed it.

cluster-autoscaler/cloudprovider/azure/azure_manager.go
```diff
@@ -534,3 +541,22 @@ func parseLabelAutoDiscoverySpec(spec string) (labelAutoDiscoveryConfig, error)
...
+// getSubscriptionId reads the Subscription ID from the instance metadata.
+func getSubscriptionIdFromInstanceMetadata() (string, error) {
+	subscriptionID, present := os.LookupEnv("ARM_SUBSCRIPTION_ID")
+	if !present {
+		metadataService, err := providerazure.NewInstanceMetadataService(metadataURL)
+		if err != nil {
+			return "", err
+		}
+
+		metadata, err := metadataService.GetMetadata(0)
```
feiskyer: why `(0)` is passed in here?

feiskyer: I see, a new param is added now

PR #2546, opened 2019-11-19
CSR: fix a bug in GetClusterSize

towca:
Currently, GetClusterSize reports the target number for all autoscaled
node groups, but the actual number for _all_ node groups, even those
that are not autoscaled. This commit fixes that behavior so that both
target and actual size reported are from autoscaled node groups only.

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -979,16 +979,18 @@ func getNotRegisteredNodes(allNodes []*apiv1.Node, cloudProviderNodeInstances ma
 	return notRegistered
 }
 
-// GetClusterSize calculates and returns cluster's current size and target size. The current size is the
-// actual number of nodes provisioned in Kubernetes, the target size is the number of nodes the CA wants.
-func (csr *ClusterStateRegistry) GetClusterSize() (currentSize, targetSize int) {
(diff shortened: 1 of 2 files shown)


losipiuk:
Looks reasonable but please confirm with someone else if we are not breaking some undocumented corner case with this change.

losipiuk:
cc: @MaciekPytel @aleksandra-malinowska

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -988,7 +988,9 @@ func (csr *ClusterStateRegistry) GetClusterSize() (currentSize, targetSize int)
 	for _, accRange := range csr.acceptableRanges {
 		targetSize += accRange.CurrentTarget
 	}
-	currentSize = csr.totalReadiness.Registered - csr.totalReadiness.NotStarted - csr.totalReadiness.LongNotStarted
```
aleksandra-malinowska: Let's update the comment above to clarify this method only returns current and target number of nodes in autoscaled node groups.

While I don't particularly like long names, it also feels like this method's name could better reflect what it's doing - e.g. GetAutoscaledNodesCount?

towca: Done.

aleksandra-malinowska:
/lgtm
/approve

Thanks!

PR #2550, opened 2019-11-20
CA: Update AWS instance type list

gjtempleton:
Mostly important due to existing Memory figures for some C5d instances being incorrect.

Despite the ability for the CA to dynamically generate this list added by #2240 users can still use this list and the memory figures for some of the C5d family currently are incorrect. These new generated figures match the published figures: https://aws.amazon.com/ec2/instance-types/c5/

This also pulls in the figures for the new super high memory instances: https://aws.amazon.com/blogs/aws/ec2-high-memory-update-new-18-tb-and-24-tb-instances/

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/ec2_instance_types.go
+++ b/cluster-autoscaler/cloudprovider/aws/ec2_instance_types.go
@@ -18,7 +18,7 @@ limitations under the License.
 
 package aws
 
-// InstanceType is sepc of EC2 instance
+// InstanceType is spec of EC2 instance
 type InstanceType struct {
 	InstanceType string
 	VCPU         int64
@@ -223,7 +223,7 @@ var InstanceTypes = map[string]*InstanceType{
 	"c5d.12xlarge": {
 		InstanceType: "c5d.12xlarge",
 		VCPU:         48,
-		MemoryMb:     131072,
+		MemoryMb:     98304,
 		GPU:          0,
 	},
 	"c5d.18xlarge": {
@@ -235,7 +235,7 @@ var InstanceTypes = map[string]*InstanceType{
 	"c5d.24xlarge": {
 		InstanceType: "c5d.24xlarge",
 		VCPU:         96,
-		MemoryMb:     262144,
+		MemoryMb:     196608,
 		GPU:          0,
 	},
 	"c5d.2xlarge": {
@@ -265,7 +265,7 @@ var InstanceTypes = map[string]*InstanceType{
 	"c5d.metal": {
 		InstanceType: "c5d.metal",
 		VCPU:         96,
-		MemoryMb:     262144,
+		MemoryMb:     196608,
 		GPU:          0,
 	},
 	"c5d.xlarge": {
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/cloudprovider/aws/ec2_instance_types.go
```diff
@@ -223,7 +223,7 @@ var InstanceTypes = map[string]*InstanceType{
 	"c5d.12xlarge": {
 		InstanceType: "c5d.12xlarge",
 		VCPU:         48,
-		MemoryMb:     131072,
```
Jeffwan: Em..Interesting. I do use the generator to populate the specs. Do you change manually or use the `gen.go`?

cluster-autoscaler/cloudprovider/aws/ec2_instance_types.go
```diff
@@ -1702,6 +1714,30 @@ var InstanceTypes = map[string]*InstanceType{
 		MemoryMb:     12582912,
 		GPU:          0,
 	},
+	"u-18tb1": {
+		InstanceType: "u-18tb1",
+		VCPU:         448,
+		MemoryMb:     0,
```
Jeffwan: Any reason it get 0 MemoryMb?

cluster-autoscaler/cloudprovider/aws/ec2_instance_types.go
gjtempleton: Yeah, I was confused as well when I noticed it, even weirder I changed it by running `make generate` again, so I can only assume the API was returning different results when you last generated it.

cluster-autoscaler/cloudprovider/aws/ec2_instance_types.go
gjtempleton: Ah, I failed to notice that! I think I'll need to dig into this and figure out why that's happening.

cluster-autoscaler/cloudprovider/aws/ec2_instance_types.go
Jeffwan: Yeah. The new values you give match with values in aws docs. Great!

cluster-autoscaler/cloudprovider/aws/ec2_instance_types.go
gjtempleton: Finally got around to spending more time on digging into this, and it looks like the Pricing API we're currently using can't be fully trusted. For one thing the correct API designation of these instances is `u-18tb1.metal`, not `u-18tb1` as returned by the pricing API.

I'm going to raise a separate issue to discuss whether/how we move away from the Pricing API as it seems it can't be fully trusted to build a proper picture of these nodes, in the meantime any problem with merging this as is? (I've updated it with a couple more changes generated when running today and double checked them against the info returned by instance descriptions using a boto3 client.)

Jeffwan: Em.. `u-18tb1.metal` is the right name according to release note. I don't have concern merging this since this instances are available from dedicated hosts with 3 yrs reservation. Most users may not even have access to it. Feel free to open a new issue and let's resolve the pricing api issue there. Some other projects uses nodes template as well like CNI, I will also have a check on reliable API.

PR #2565, opened 2019-11-26
Fix a bug in initializing scheduler

MaciekPytel:
By default scheduler now returns plugins instead of predicates.
However, plugins are not yet supported by CA, so we need to make
scheduler return predicates.

### DIFF
--- a/cluster-autoscaler/simulator/predicates.go
+++ b/cluster-autoscaler/simulator/predicates.go
@@ -32,7 +32,6 @@ import (
 
 	"k8s.io/kubernetes/pkg/scheduler"
 	"k8s.io/kubernetes/pkg/scheduler/algorithm/predicates"
-	schedulerconfig "k8s.io/kubernetes/pkg/scheduler/apis/config"
 	schedulerlisters "k8s.io/kubernetes/pkg/scheduler/listers"
 	schedulernodeinfo "k8s.io/kubernetes/pkg/scheduler/nodeinfo"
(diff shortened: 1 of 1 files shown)


MaciekPytel:
/hold
To give @ahg-g a chance to look at this.

MaciekPytel:
I tested this using pods with hostport conflict - without this CA won't scale for them, because it lacks relevant predicate and thinks those are schedulable. With this change the scale-up happens as expected.

cluster-autoscaler/simulator/predicates.go
```diff
@@ -125,6 +125,7 @@ func NewPredicateChecker(kubeClient kube_client.Interface, stop <-chan struct{})
 		NoOpEventRecorder{},
 		stop,
 		scheduler.WithAlgorithmSource(algorithmSource),
+		scheduler.WithFrameworkConfigProducerRegistry(nil),
```
ahg-g: you can remove line 127 above, by default the scheduler uses "schedulerconfig.SchedulerDefaultProviderName"

MaciekPytel: That's a fair point. I was wondering if we're better off not relying on default, but assuming that by default we'll get a DefaultProvider seems reasonable.

MaciekPytel:
/approve
It was already approved by @losipiuk, but apparently the bot needs the commands in separate lines. Also I'm curious if self approval will work :)

Issue #2473, opened 2019-10-22
Use apps/v1 in stead of post 1.16.x deprecated apis (extensions/v1beta1)


drewboswell:
TLDR; "extensions/v1beta1" bad, "apps/v1" good.

There are references to extensions/v1beta1 everywhere in this repository including in the go code.
These will not work for 1.16.x + k8s clusters.

using "apps/v1" allows for compatibility for clusters 1.9.x+.

bskiba:
The only references left are the ones in tests for object references, which should be fine.
Closing
/close

PR #2575, opened 2019-11-27
Switch to GA types for workloads, rbac

liggitt:
xref https://github.com/kubernetes/autoscaler/issues/2473


liggitt:
cc @wojtek-t

PR #2576, opened 2019-11-27
Ignore NominatedNodeName on Pod if node is gone

losipiuk:


losipiuk:
/lgtm
/approve
Thanks!

Issue #2597, opened 2019-11-28
[addon-resizer] Add support for scale up/down delay


44past4:
In case of fast changes in the cluster size it might be a good idea to delay resizing of the addon until the size of the cluster is stabilized. Therefore it would be good to have an option to specify a delay for both scale up and scale down operations.

44past4:
All required changes have been merged to master and 1.8 branches of addon-resizer.

PR #2598, opened 2019-11-28
Add support for scale up/down delay

44past4:
Fix for #2597 for 1.8

@bskiba

### DIFF
--- a/addon-resizer/nanny/nanny_lib.go
+++ b/addon-resizer/nanny/nanny_lib.go
@@ -21,6 +21,7 @@ and update a deployment based on that status.
 package nanny
 
 import (
+	"encoding/json"
 	"time"
 
 	log "github.com/golang/glog"
@@ -28,35 +29,59 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-// checkResource determines whether a specific resource needs to be over-written.
-func checkResource(threshold int64, actual, expected corev1.ResourceList, res corev1.ResourceName) bool {
+type operation int
+
+const (
+	unknown   operation = iota
+	scaleDown operation = iota
+	scaleUp   operation = iota
+)
+
+type updateResult int
+
+const (
+	noChange  updateResult = iota
+	postpone  updateResult = iota
+	overwrite updateResult = iota
+)
+
+// checkResource determines whether a specific resource needs to be over-written and determines type of the operation.
+func checkResource(threshold int64, actual, expected corev1.ResourceList, res corev1.ResourceName) (bool, operation) {
 	val, ok := actual[res]
 	expVal, expOk := expected[res]
 	if ok != expOk {
-		return true
+		return true, unknown
 	}
 	if !ok && !expOk {
-		return false
+		return false, unknown
 	}
 	q := new(inf.Dec).QuoRound(val.AsDec(), expVal.AsDec(), 2, inf.RoundDown)
 	lower := inf.NewDec(100-threshold, 2)
 	upper := inf.NewDec(100+threshold, 2)
-	if q.Cmp(lower) == -1 || q.Cmp(upper) == 1 {
-		return true
+	if q.Cmp(lower) == -1 {
+		return true, scaleUp
 	}
-	return false
+	if q.Cmp(upper) == 1 {
+		return true, scaleDown
+	}
+	return false, unknown
 }
 
-// shouldOverwriteResources determines if we should over-write the container's
-// resource limits. We'll over-write the resource limits if the limited
-// resources are different, or if any limit is violated by a threshold.
-func shouldOverwriteResources(threshold int64, limits, reqs, expLimits, expReqs corev1.ResourceList) bool {
-	return checkResource(threshold, limits, expLimits, corev1.ResourceCPU) ||
-		checkResource(threshold, limits, expLimits, corev1.ResourceMemory) ||
-		checkResource(threshold, limits, expLimits, corev1.ResourceStorage) ||
-		checkResource(threshold, reqs, expReqs, corev1.ResourceCPU) ||
-		checkResource(threshold, reqs, expReqs, corev1.ResourceMemory) ||
-		checkResource(threshold, reqs, expReqs, corev1.ResourceStorage)
+// shouldOverwriteResources determines if we should over-write the container's resource limits and determines type of the operation.
+// We'll over-write the resource limits if the limited resources are different, or if any limit is violated by a threshold.
+// Returned operation type (scale up/down or unknown) is calculated based on values taken from the first resource which requires overwrite.
+func shouldOverwriteResources(threshold int64, limits, reqs, expLimits, expReqs corev1.ResourceList) (bool, operation) {
+	for _, resourceType := range []corev1.ResourceName{corev1.ResourceCPU, corev1.ResourceMemory, corev1.ResourceStorage} {
+		overwrite, op := checkResource(threshold, limits, expLimits, resourceType)
+		if overwrite {
+			return true, op
+		}
+		overwrite, op = checkResource(threshold, reqs, expReqs, resourceType)
+		if overwrite {
+			return true, op
+		}
+	}
+	return false, unknown
 }
 
 // KubernetesClient is an object that performs the nanny's requisite interactions with Kubernetes.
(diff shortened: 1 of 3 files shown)


addon-resizer/nanny/estimator.go
```diff
@@ -98,6 +98,8 @@ func calculateResources(numNodes uint64, resources []Resource) *corev1.ResourceR
 		overhead := computeResourceOverheadValueString(numNodes, r)
 		newRes := r.Base
 		newRes.Add(resource.MustParse(overhead))
+		// Force initialization of newRes.s so that it could be logged
+		newRes.String()
```
wojtek-t: Seriously? You really need?
How this is logged?

44past4: Without this call this value can be logged as: 
```{i:{value:0 scale:0} d:{Dec:0xc4202b5d70} s: Format:DecimalSI}```
WIth this call this becomes:
```{i:{value:0 scale:0} d:{Dec:0xc4202b5d70} s:455m Format:DecimalSI}```

Comment (wojtek-t) [reply addon-resizer/nanny/estimator.go]: How do you print it? If you would log it with:
`klog.("%s", <your var>)`
if would work fine (as this is calling String() method.

Comment (wojtek-t) [reply addon-resizer/nanny/estimator.go]: And it seems like a better way to print it - things like `d:{Dec:0xc4202b5d70}` aren't particularly useful :)

File: addon-resizer/nanny/nanny_lib.go [hunk 2]
```diff
@@ -100,15 +118,24 @@ func PollAPIServer(k8s KubernetesClient, est ResourceEstimator, contName string,
...
 		}
 
+		now := time.Now()
+		if (op == scaleDown && now.Before(lastChange.Add(scaleDownDelay))) ||
+			(op == scaleUp && now.Before(lastChange.Add(scaleUpDelay))) {
+			log.V(3).Infof("Resources are not within the expected limits, Actual: %+v, accepted range: %+v. Skipping resource update because of scale up/down delay", *resources, *expResources)
+			continue
+		}
+
 		log.Infof("Resources are not within the expected limits, updating the deployment. Actual: %+v Expected: %+v", *resources, *expResources)
```
bskiba: nit: can you make this log consistent with the one above? Also could you add a log verbosity level?

addon-resizer/nanny/nanny_lib.go
```diff
@@ -28,35 +28,51 @@ import (
...
 
+type operation int
+
+const (
+	unknown   operation = iota
+	scaleDown operation = iota
+	scaleUp   operation = iota
+)
+
 // checkResource determines whether a specific resource needs to be over-written.
```
bskiba: Can you update the comment?

addon-resizer/nanny/nanny_lib.go
```diff
@@ -100,15 +118,24 @@ func PollAPIServer(k8s KubernetesClient, est ResourceEstimator, contName string,
...
 		// If there's a difference, go ahead and set the new values.
-		if !shouldOverwriteResources(int64(threshold), resources.Limits, resources.Requests, expResources.Limits, expResources.Requests) {
+		overwrite, op := shouldOverwriteResources(int64(threshold), resources.Limits, resources.Requests, expResources.Limits, expResources.Requests)
+		if !overwrite {
 			log.V(4).Infof("Resources are within the expected limits. Actual: %+v Expected: %+v", *resources, *expResources)
 			continue
 		}
 
+		now := time.Now()
+		if (op == scaleDown && now.Before(lastChange.Add(scaleDownDelay))) ||
```
mwielgus: Can you point me to the specific line in test where this thing is properly unit-tested. I don't see a line where we check what happens if last change is slightly before or slightly after scale****Delay. But maybe I'm missing something.

addon-resizer/nanny/estimator.go
44past4: This value is not printed directly but instead it is nested in k8s.io/api/core/v1/ResourceRequirements which is printed with `%+v`.
I will change it in a way that I will serialize ResourceRequirements to json before displaying it.

addon-resizer/nanny/nanny_lib.go
44past4: Done

addon-resizer/nanny/nanny_lib.go
44past4: Log message has been updated. I have not added log verbosity because I believe that this message will be logged only when actual resize will be performed and therefore it is ok to log this message without a verbosity level being set.

addon-resizer/nanny/nanny_lib.go
44past4: I have extracted logic for checking resources for deployment and updating them to a separate method and added unit tests for this.
When it comes E2E tests I have build an image with addon-resizer, deployed it to a Kubernetes cluster first without any delay being set and then with scale up and scale down being set and I have scaled cluster from 3 to 10 nodes and then back to 3 nodes.

addon-resizer/nanny/nanny_lib.go
```diff
@@ -74,41 +90,73 @@ type ResourceEstimator interface {
...
+	}
 
-		log.Infof("Resources are not within the expected limits, updating the deployment. Actual: %+v Expected: %+v", *resources, *expResources)
-		if err := k8s.UpdateDeployment(expResources); err != nil {
-			log.Error(err)
-			continue
+	if (op == scaleDown && now.Before(lastChange.Add(scaleDownDelay))) ||
+		(op == scaleUp && now.Before(lastChange.Add(scaleUpDelay))) {
+		if log.V(3) {
+			log.V(3).Infof("Resources are not within the expected limits, Actual: %+v, accepted range: %+v. Skipping resource update because of scale up/down delay", jsonOrValue(*resources), jsonOrValue(*expResources))
```
wojtek-t: I actually think this should be logged unconditionally - this is important log and lack of it may introduce a lot of misunderstanding - please change to simply log.Infof

addon-resizer/nanny/nanny_lib.go
```diff
@@ -21,42 +21,58 @@ and update a deployment based on that status.
...
+func shouldOverwriteResources(threshold int64, limits, reqs, expLimits, expReqs corev1.ResourceList) (bool, operation) {
+	for _, resourceType := range []corev1.ResourceName{corev1.ResourceCPU, corev1.ResourceMemory, corev1.ResourceStorage} {
+		override, op := checkResource(threshold, limits, expLimits, resourceType)
+		if override {
+			return true, op
+		}
+		override, op = checkResource(threshold, reqs, expReqs, resourceType)
+		if override {
+			return true, op
+		}
```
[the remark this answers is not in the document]
44past4: I have added a comment about the operation type to the function description.

addon-resizer/nanny/nanny_lib.go
44past4: Logging this unconditionally could spam logs with hundreds of the same messages were only the first one is really useful for the user. Therefore I have decided to change the implementation in a way that I will log this message unconditionally but only for the first time we when we are skipping the the override.

addon-resizer/nanny/nanny_lib.go
```diff
@@ -74,41 +99,74 @@ type ResourceEstimator interface {
...
-		// Query the apiserver for this pod's information.
-		resources, err := k8s.ContainerResources()
-		if err != nil {
-			log.Errorf("Error while querying apiserver for resources: %v", err)
-			continue
-		}
+// updateResources counts the number of nodes, estimates the expected
+// ResourceRequirements, compares them to the actual ResourceRequirements, and
+// updates the deployment with the expected ResourceRequirements if necessary.
+// It returns true if deployment has been updated.
```
wojtek-t: no longer true

addon-resizer/nanny/nanny_lib.go
```diff
@@ -74,41 +99,74 @@ type ResourceEstimator interface {
...
-		expResources := est.scaleWithNodes(num)
+	// Get the expected resource limits.
+	expResources := est.scaleWithNodes(num)
 
-		// If there's a difference, go ahead and set the new values.
-		if !shouldOverwriteResources(int64(threshold), resources.Limits, resources.Requests, expResources.Limits, expResources.Requests) {
-			log.V(4).Infof("Resources are within the expected limits. Actual: %+v Expected: %+v", *resources, *expResources)
-			continue
+	// If there's a difference, go ahead and set the new values.
+	overwrite, op := shouldOverwriteResources(int64(threshold), resources.Limits, resources.Requests, expResources.Limits, expResources.Requests)
```
wojtek-t: nit: consistency - you're using overwrite in some places and override in other
Can you please unify it?

44past4: I have changed all occurrences of override to overwrite.

addon-resizer/nanny/nanny_lib.go
44past4: Updated the information about the returned values.

PR #2606, opened 2019-12-03
Add FetchAvailableCpuPlatforms to AutoscalingGceClient + fix

losipiuk:


cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -312,6 +318,22 @@ func (client *autoscalingGceClientV1) FetchZones(region string) ([]string, error
 	return zones, nil
 }
 
+func (client *autoscalingGceClientV1) FetchAvailableCpuPlatforms() (map[string][]string, error) {
```
Jeffwan: This method is not being used yet?

losipiuk: It is not. We use autoscalingGceClientV1 also in GKE cloud provider which is not hosted on github.

PR #2610, opened 2019-12-03
Improved logging for errors during updating state

arjunrn:
In an earlier PR #1868 I added a memory saver mode which stores only pod info and container metrics for those pods which have a VPA. In later PR some logging was added when container metrics could not be stored regardless of the mode. This causes a lot of false error logs when the memory saver mode is on. So I've added a similar skip condition when the container does not have a matching VPA.

Also there are a couple of places where errors are not logged when the pod and container states cannot be added/updated. So I have added appropriate warnings in those places.

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
@@ -367,7 +367,9 @@ func (feeder *clusterStateFeeder) LoadPods() {
 		}
 		feeder.clusterState.AddOrUpdatePod(pod.ID, pod.PodLabels, pod.Phase)
 		for _, container := range pod.Containers {
-			feeder.clusterState.AddOrUpdateContainer(container.ID, container.Request)
+			if err = feeder.clusterState.AddOrUpdateContainer(container.ID, container.Request); err != nil {
+				klog.Warningf("Failed to add container %+v. Reason: %+v", container.ID, err)
+			}
 		}
 	}
 }
@@ -383,6 +385,10 @@ func (feeder *clusterStateFeeder) LoadRealTimeMetrics() {
 	for _, containerMetrics := range containersMetrics {
 		for _, sample := range newContainerUsageSamplesWithKey(containerMetrics) {
 			if err := feeder.clusterState.AddSample(sample); err != nil {
+				// Not all pod states are tracked in memory saver mode
+				if _, isKeyError := err.(model.KeyError); isKeyError && feeder.memorySaveMode {
+					continue
+				}
 				klog.Warningf("Error adding metric sample for container %v: %v", sample.Container, err)
 				droppedSampleCount++
 			} else {
@@ -396,7 +402,9 @@ Loop:
 		select {
 		case oomInfo := <-feeder.oomChan:
 			klog.V(3).Infof("OOM detected %+v", oomInfo)
-			feeder.clusterState.RecordOOM(oomInfo.ContainerID, oomInfo.Timestamp, oomInfo.Memory)
+			if err = feeder.clusterState.RecordOOM(oomInfo.ContainerID, oomInfo.Timestamp, oomInfo.Memory); err != nil {
+				klog.Warningf("Failed to record OOM %+v. Reason: %+v", oomInfo, err)
+			}
 		default:
 			break Loop
 		}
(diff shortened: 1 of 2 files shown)


vertical-pod-autoscaler/pkg/recommender/input/metrics/metrics_client.go
```diff
@@ -94,6 +96,7 @@ func newContainerMetricsSnapshot(containerMetrics v1beta1.ContainerMetrics, podM
 				PodName:   podMetrics.Name,
 			},
 		},
+		PodLabels:      podMetrics.Labels,
```
bskiba: I don't think this will work. From what I see looking at the PodMetrics objects in the example cluster, they don't have the same labels as the pods have. 
See example metrics for my hamster pod (this is part of the output when running `kubectl get --raw "/apis/metrics.k8s.io/v1beta1/pods"`):
```
{
      "metadata": {
        "name": "hamster-6d7f85f5d9-r9gn8",
        "namespace": "default",
        "selfLink": "/apis/metrics.k8s.io/v1beta1/namespaces/default/pods/hamster-6d7f85f5d9-r9gn8",
        "creationTimestamp": "2019-12-03T15:50:57Z"
      },
      ...
    }
```
While `kubectl describe pod hamster-6d7f85f5d9-r9gn8` gives me
```
...
Labels:   app=hamster
          pod-template-hash=6d7f85f5d9
...
```

arjunrn: I was trying out the changes now and just found out the same thing. Would it be ok if the recommender did not log warnings if there is a `KeyError` in the memory saver mode?

bskiba: Yeah, I think this is the easiest to do to not spam logs with this warning. 

Another thing is, that we could have memory saver mode query only for the metrics it actually needs , but it's out of scope of improving logging :)

arjunrn: I actually considered that and ran some tests. The problem with that approach is that it's much slower than doing it this way because of how the _api-server_ gets results given a label selector. All the objects from _etcd_ are retrieved and then matched with the label selector. So multiple label selector queries would mean all the metrics would be retrieved multiple times. Much faster if it's done this way.
I will go ahead and use the second approach. Thanks.

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -367,7 +367,10 @@ func (feeder *clusterStateFeeder) LoadPods() {
 		}
 		feeder.clusterState.AddOrUpdatePod(pod.ID, pod.PodLabels, pod.Phase)
 		for _, container := range pod.Containers {
-			feeder.clusterState.AddOrUpdateContainer(container.ID, container.Request)
+			err = feeder.clusterState.AddOrUpdateContainer(container.ID, container.Request)
```
bskiba: nit: You can shorten this into if err=..; err != nil (applies in the below changes as well)

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -383,6 +386,10 @@ func (feeder *clusterStateFeeder) LoadRealTimeMetrics() {
 	for _, containerMetrics := range containersMetrics {
 		for _, sample := range newContainerUsageSamplesWithKey(containerMetrics) {
 			if err := feeder.clusterState.AddSample(sample); err != nil {
+				// Not all pod states are tracked in memory saver mode
```
bskiba: nit: s/pod/container, since the keyError is raised for the container id in the metric

arjunrn: I did mean pod state because the `KeyError` will be from this check [here](https://github.com/kubernetes/autoscaler/blob/b232915c2485345d3ae8e587bc3fa933bc5c70d9/vertical-pod-autoscaler/pkg/recommender/model/cluster.go#L180)

bskiba: Ah, you're right, sorry!

Issue #2631, opened 2019-12-10
VPA - admission e2e test race condition


krzysied:
Current implementation of vpa admission test has a problem with race condition.

The scenario is:
1. VPA object is created.
2. Deployment object is create.
3. VPA watch on deployments is lagging so it takes ~1s to observe new deployment
4. New pods are sent to admission webhook and due to deployment not being observed they don't get the recommendations.
5. Test expects resource recommendations and flakes.

There was [a fix](https://github.com/kubernetes/autoscaler/pull/2504) created by @jbartosik that separates deployment creation and scale-up. It solved half of the problem. However, creating deployment and getting created object from etcd can still be faster then observing deployment with a watch.

bskiba:
@krzysied What is the current status of this issue? Is it resolved, or are we still missing pieces?

krzysied:
I think we can close it.
Current status:
- Tests will not be flaky.
- Race condition still exists, although it doesn't affect the correctness in normal clusters.

bskiba:
Thanks!
/close

PR #2632, opened 2019-12-10
Deflaking admission controller e2e test

krzysied:
"Fixing" admission controller test race condition.

ref #2631

### DIFF
--- a/vertical-pod-autoscaler/e2e/v1/admission_controller.go
+++ b/vertical-pod-autoscaler/e2e/v1/admission_controller.go
@@ -18,6 +18,7 @@ package autoscaling
 
 import (
 	"fmt"
+	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	autoscalingv1 "k8s.io/api/autoscaling/v1"
@@ -517,6 +518,9 @@ var _ = AdmissionControllerE2eDescribe("Admission-controller", func() {
 })
 
 func startDeploymentPods(f *framework.Framework, deployment *appsv1.Deployment) *apiv1.PodList {
+	// Apiserver watch can lag depending on cached object count and apiserver resource usage.
+	// We assume that watch can lag up to 5 seconds.
+	const apiserverWatchLag = 5 * time.Second
 	// In admission controller e2e tests a recommendation is created before deployment.
 	// Creating deployment with size greater than 0 would create a race between information
 	// about pods and information about deployment getting to the admission controller.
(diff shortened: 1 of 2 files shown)


vertical-pod-autoscaler/e2e/v1/admission_controller.go
```diff
@@ -517,6 +518,9 @@ var _ = AdmissionControllerE2eDescribe("Admission-controller", func() {
 })
 
 func startDeploymentPods(f *framework.Framework, deployment *appsv1.Deployment) *apiv1.PodList {
+	// Etcd watch can lag depending on etcd object count and etcd resource usage.
+	// We assume that ectd can lag up to 5 seconds.
+	const etcdWatchLag = 5 * time.Second
```
jbartosik: Why 5 second delay? Will it be enough?

vertical-pod-autoscaler/e2e/v1/admission_controller.go
```diff
@@ -532,6 +536,14 @@ func startDeploymentPods(f *framework.Framework, deployment *appsv1.Deployment)
 
 	err = framework_deployment.WaitForDeploymentComplete(c, deployment)
 	gomega.Expect(err).NotTo(gomega.HaveOccurred(), "when waiting for empty deployment to create")
+	// We want to ensure that test is not flaky due to the etcd lag.
```
jbartosik: I think a short explanation of the problem would be in place. Something like:

"If admission controller  recieves pod before controller it will not apply recommendation and test will fail. Wait after creating deployment to ensure VPA knows about it, then scale up. This is not a problem in terms of correctness:..."

vertical-pod-autoscaler/e2e/v1beta2/admission_controller.go
```diff
@@ -516,6 +517,9 @@ var _ = AdmissionControllerE2eDescribe("Admission-controller", func() {
 })
 
 func startDeploymentPods(f *framework.Framework, deployment *appsv1.Deployment) *apiv1.PodList {
+	// Etcd watch can lag depending on etcd object count and etcd resource usage.
+	// We assume that ectd can lag up to 5 seconds.
+	const etcdWatchLag = 5 * time.Second
```
jbartosik: Same comments as to v1 tests.

vertical-pod-autoscaler/e2e/v1/admission_controller.go
krzysied: Should be enough. Imho the lag should be  max 1s, so 5s is a little bit over the top.
In this case we want to focus on specific scenarios, so it's not a problem to wait little longer.

vertical-pod-autoscaler/e2e/v1/admission_controller.go
krzysied: Done.
I've also updated const description. It's about watch on apiserver, not a watch on etcd.

vertical-pod-autoscaler/e2e/v1beta2/admission_controller.go
krzysied: Done.

PR #2644, opened 2019-12-12
Change the regular expression to match the Fargate's ProviderIDs.

dggondol:
EKS now supports pods running on Fargate. The providerID has a different format:

"aws:///eu-west-1b/120328e333-369fe1b7ba8b4416ab87216be91dc4f1/fargate-ip-192-168-159-181.eu-west-1.compute.internal"

I just adapted the regex to match both patterns. Without this, the autoscaler will not work if you have one node (one pod) running on Fargate.


Jeffwan:
@dggondol do you use a different github id? Could you resolve the `commit missing Github user` issue? You can change username of the commit.

PR #2661, opened 2019-12-18
fix: properly parse azure cloud provider config

marwanad:
azure cloud provider config files aren't INI-formatted so the initial parsing logic would fail


feiskyer:
@marwanad Thanks for the fix. 

/area provider/azure
/kind bug

PR #2667, opened 2019-12-19
fix broken link: https://github.com/golang/lint/blob/master/README

CriaHu:
broken link : https://github.com/golang/lint/blob/master/README
correct link : https://github.com/golang/lint/blob/master/README.md


jbartosik:
Thanks for the PR but this isn't the right place to fix this. The files you're changing are in `vendor` directories. That means we don't own them, we just make a copy of files from the original repository (in this case https://github.com/google/gofuzz).

If we fix our copy then the next time we're updating our dependencies we would owrrite your fix with content of the original repository.

What you should do instead is sending a pull request to fix the link in the file in the original repository (https://github.com/google/gofuzz/blob/master/CONTRIBUTING.md). Then the link will be fixed for everyone (and the next time we update our dependencies we'll pull in the fix into our reopsitory).

PR #2694, opened 2019-12-27
Add UnregisteredNodeRemovalDelayAfterAttempt options when removeUnregisterNode

DasyDong:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it:**
In some conditions remove unregister nodes may be unexpected heavily.

- oldUnregisteredNode work well actually  

- remove nodes action very heavy, may be 5-10minutes, although I set maxnodeprovision time, I still don't want to delete all Unregistered nodes in same time 

Is it possible to add delay time to reduce the probability of false deletion?
So I add RemoveUnregisterDelay option, RemoveUnregisterDelay sets the delay time before the next remove unregister attempt ,default is 0. 

**Which issue(s) this PR fixes:**
None

sig /autoscaling 

cluster-autoscaler

### DIFF
--- a/cluster-autoscaler/config/autoscaling_options.go
+++ b/cluster-autoscaler/config/autoscaling_options.go
@@ -84,6 +84,8 @@ type AutoscalingOptions struct {
 	CloudProviderName string
 	// NodeGroups is the list of node groups a.k.a autoscaling targets
 	NodeGroups []string
+	// UnregisteredNodeRemovalDelayAfterAttempt sets the duration before the next remove unregister attempt
+	UnregisteredNodeRemovalDelayAfterAttempt time.Duration
 	// ScaleDownEnabled is used to allow CA to scale down the cluster
 	ScaleDownEnabled bool
 	// ScaleDownDelayAfterAdd sets the duration from the last scale up to the time when CA starts to check scale down options
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -63,15 +63,16 @@ type StaticAutoscaler struct {
 	// AutoscalingContext consists of validated settings and options for this autoscaler
 	*context.AutoscalingContext
 	// ClusterState for maintaining the state of cluster nodes.
-	clusterStateRegistry    *clusterstate.ClusterStateRegistry
-	startTime               time.Time
-	lastScaleUpTime         time.Time
-	lastScaleDownDeleteTime time.Time
-	lastScaleDownFailTime   time.Time
-	scaleDown               *ScaleDown
-	processors              *ca_processors.AutoscalingProcessors
-	processorCallbacks      *staticAutoscalerProcessorCallbacks
-	initialized             bool
+	clusterStateRegistry        *clusterstate.ClusterStateRegistry
+	startTime                   time.Time
+	lastScaleUpTime             time.Time
+	lastScaleDownDeleteTime     time.Time
+	lastScaleDownFailTime       time.Time
+	lastUnregisteredRemovalTime time.Time
+	scaleDown                   *ScaleDown
+	processors                  *ca_processors.AutoscalingProcessors
+	processorCallbacks          *staticAutoscalerProcessorCallbacks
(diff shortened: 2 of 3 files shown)


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -68,6 +68,7 @@ type StaticAutoscaler struct {
 	lastScaleUpTime         time.Time
 	lastScaleDownDeleteTime time.Time
 	lastScaleDownFailTime   time.Time
+	lastRemoveUnregister    time.Time
```
gjtempleton: The name of this variable is unclear on first reading to me. 

Something like `lastUnregisteredRemovalTime` would match the naming scheme of the existing variables used for similar purposes.

cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -84,6 +84,8 @@ type AutoscalingOptions struct {
 	CloudProviderName string
 	// NodeGroups is the list of node groups a.k.a autoscaling targets
 	NodeGroups []string
+	// RemoveUnregisterDelay sets the duration before the next remove unregister attempt
+	RemoveUnregisterDelay time.Duration
```
gjtempleton: Why not follow the existing naming scheme for options used for similar functionality here?

(e.g. something like `UnregisteredNodeRemovalDelayAfterAttempt`)

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -267,15 +269,26 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
-			klog.Warningf("Failed to remove unregistered nodes: %v", err)
-		}
-		if removedAny {
-			klog.V(0).Infof("Some unregistered nodes were removed, skipping iteration")
-			return nil
+
+		// Add protection for remove unregister nodes
+		if a.lastRemoveUnregister.Add(a.RemoveUnregisterDelay).After(currentTime) {
+			for _, unregisteredNode := range unregisteredNodes {
+				klog.V(4).Infof("Skipping remove unregister node: %v, lastRemoveUnregister=%s",
```
gjtempleton: It feels like this should have more explicit information about why the nodes were skipped. The time of the last removal without any further info wouldn't be very clear on first read.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -267,15 +269,26 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
+					unregisteredNode.Node.Name, a.lastRemoveUnregister)
+			}
+		} else {
+			removedAny, err := removeOldUnregisteredNodes(unregisteredNodes, autoscalingContext,
+				a.clusterStateRegistry, currentTime, autoscalingContext.LogRecorder)
+			// Set remove unregister time
+			a.lastRemoveUnregister = currentTime
+			// There was a problem with removing unregistered nodes. Retry in the next loop.
```
gjtempleton: This comment is now only true dependent on the value of `RemoveUnregisterDelay`

cluster-autoscaler/core/static_autoscaler.go
DasyDong: > It feels like this should have more explicit information about why the nodes were skipped. The time of the last removal without any further info wouldn't be very clear on first read.

Thanks  a lot @gjtempleton 
Excellent.I accept your ideas, I have made some changes .  

About the why the nodes were skipped, as my opinion it is similar with ScaleDownDelayAfterDelete . More info is not necessary if we know what is UnregisteredNodeRemovalDelay. 

 We see 'Skipping remove unregister node..', UnregisteredNodeRemovalDelay time not reached

1  UnregisteredNodeRemovalDelay could be optional to reduce the probability of wrong removal
when we have so many k8s clusters

2 If we have chance to trigger some custom behavior  during in node removal delay time,  eg: monitor alarm (if necessary), that also be nice

I add more comments in new commit

PR #2531, opened 2019-11-13
VPA: separate module for e2e tests

wozniakjan:
Effort to simplify go modules tracking even more than https://github.com/kubernetes/autoscaler/pull/2530 by removing the requirement for replace directives with filesystem absolute paths and separate go module for the e2e tests.

It is generally not recommended to depend on `k8s.io/kubernetes` https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-505725449, and opens doors to issues for people who may want to depend on VPA - https://github.com/kubernetes/kube-state-metrics/pull/949#discussion_r344429628

It looks like the e2e tests are the only large obstacle in removing the dependency on `k8s.io/kuberentes`. So I would like to propose separate go module for the e2e tests to simplify depending on VPA and only make life hard for those who decide to depend on VPA e2e package. Input from maintainers is greatly appreciated.

<details><summary>checking imports - grep output</summary>
<p>

```
$ grep k8s.io/autoscaler/vertical-pod-autoscaler --include=*.go -nre '[[:blank:]]"k8s.io/kubernetes' | grep -v "$vendor/k8s.io/kubernetes"
./pkg/utils/metrics/metrics.go:28:	_ "k8s.io/kubernetes/pkg/client/metrics/prometheus" // for client-go metrics registration
./e2e/v1/updater.go:33:	"k8s.io/kubernetes/test/e2e/framework"
./e2e/v1/updater.go:34:	framework_deployment "k8s.io/kubernetes/test/e2e/framework/deployment"
./e2e/v1/updater.go:35:	framework_job "k8s.io/kubernetes/test/e2e/framework/job"
./e2e/v1/updater.go:36:	framework_rs "k8s.io/kubernetes/test/e2e/framework/replicaset"
./e2e/v1/updater.go:37:	framework_ss "k8s.io/kubernetes/test/e2e/framework/statefulset"
./e2e/v1/updater.go:38:	testutils "k8s.io/kubernetes/test/utils"
./e2e/v1/autoscaling_utils.go:34:	api "k8s.io/kubernetes/pkg/apis/core"
./e2e/v1/autoscaling_utils.go:35:	"k8s.io/kubernetes/test/e2e/framework"
./e2e/v1/autoscaling_utils.go:36:	"k8s.io/kubernetes/test/e2e/framework/replicaset"
./e2e/v1/autoscaling_utils.go:37:	e2eservice "k8s.io/kubernetes/test/e2e/framework/service"
[...]

PR #2697, opened 2019-12-30
VPA: fix bash set options

wozniakjan:
The `set` in `hack/update-kubernetes-deps-in-e2e.sh` contains redundant options:
* `errexit` equals `-e`
* `nounset` equals `-u`

Also, I have misunderstood how to correctly set options in https://github.com/kubernetes/autoscaler/pull/2531, the `-o` can take only one argument and `pipefail` doesn't have short variant so it has to remain long.

Resource: https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

ptal @bskiba


bskiba:
Thanks!
/lgtm
/approve

Issue #2522, opened 2019-11-12
vpa: Misconfigured CRD can cause VPA to target ALL pods of a namespace.


mikkeloscar:
/kind bug

If you try to create a VPA for a CRD which has an incorrect/missing scale configuration as defined in: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#scale-subresource then you can end up with a VPA which targets ALL pods in a namespace.
This issue specifically happens if you misconfigure the `labelSelectorPath:` or simply provide an empty string in the labelSelector defined by the CRD scale subresource.

What happens is the selector is resolved and used here: https://github.com/kubernetes/autoscaler/blob/3da8964ead179c471f6a2d17dde9b35b935df2ab/vertical-pod-autoscaler/pkg/target/fetcher.go#L216
But if it was an empty string to begin with it will mean **select-all** and thus match all pods in a namespace.

I believe the VPA should validate the selector and report an error if the selector matches everything, because that should not be a valid label selector.

bskiba:
Thanks for reporting! 
I would argue that the validation should happen on the side of CRD, but [according to the docs](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#scale-subresource) the field is optional so that is out of the question since an empty selector here is correct from the CRD definition point of view.
Seems that the Horizontal Pod Autoscaler may have similar issues: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/podautoscaler/horizontal.go#L245

I agree that there should be a check for empty selector here, setting an error condition on the VPA object that scale subresource of the CRD is invalid for VPA needs.

mikkeloscar:
> I agree that there should be a check for empty selector here, setting an error condition on the VPA object that scale subresource of the CRD is invalid for VPA needs.

What do you mean by this? Should we NOT set an error condition?
If we should NOT set an error, how should we handle the empty selector as the code right now would want to use the selector to find the relevant pods.

bskiba:
Sorry for being unclear. I meant:
- check that the selector is not empty
- if the selector is empty, set error condition on VPA and not try to process it.

bskiba:
@arjunrn I believe this is fixed by #2711? 
Please reopen if I misunderstood.
/close

PR #2711, opened 2020-01-05
Ensure that the label selector in the scale subresource is not empty

arjunrn:
Fixes issue #2522


vertical-pod-autoscaler/pkg/target/fetcher.go
```diff
@@ -213,6 +213,9 @@ func (f *vpaTargetSelectorFetcher) getLabelSelectorFromResource(
 		groupResource := mapping.Resource.GroupResource()
 		scale, err := f.scaleNamespacer.Scales(namespace).Get(groupResource, name)
 		if err == nil {
+			if scale.Status.Selector == "" {
+				return nil, fmt.Errorf("Resource %s/%s has an empty selector for scale sub-resource", namespace, name)
```
bskiba: nit: please also log the resource type

bskiba: Actually, scratch that, the resource type is added to the error in the calling function

PR #2717, opened 2020-01-07
Add logging taints on template node on PodToleratesNodeTaints failure

aleksandra-malinowska:
Add logging taints on template node on PodToleratesNodeTaints failure.

### DIFF
--- a/cluster-autoscaler/simulator/predicates.go
+++ b/cluster-autoscaler/simulator/predicates.go
@@ -35,6 +35,7 @@ import (
 	schedulerlisters "k8s.io/kubernetes/pkg/scheduler/listers"
 	schedulernodeinfo "k8s.io/kubernetes/pkg/scheduler/nodeinfo"
 	schedulersnapshot "k8s.io/kubernetes/pkg/scheduler/nodeinfo/snapshot"
+
 	// We need to import provider to initialize default scheduler.
 	"k8s.io/kubernetes/pkg/scheduler/algorithmprovider"
 )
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/simulator/predicates.go
```diff
@@ -338,8 +342,18 @@ func (p *PredicateChecker) CheckPredicates(pod *apiv1.Pod, predicateMetadata pre
...
 			}
 		}
 	}
 	return nil
 }
+
+func (p *PredicateChecker) addDebugInfo(predInfo PredicateInfo, nodeInfo *schedulernodeinfo.NodeInfo) string {
+	switch predInfo.Name {
+	case "PodToleratesNodeTaints":
+		return fmt.Sprintf("taints on node: %#v", nodeInfo.Node().Spec.Taints)
```
MaciekPytel: The idea of introducing separate VerboseError that lazily generates error message (or, technically speaking, its previous implementation) was specifically to avoid surprisingly high cost of sprintfs in cases when we discard errors anyway (which is vast majority of times, for example in binpacking or drain simulations). Adding an unconditional sprintf here goes against this. 
I wonder how much it matters in practice - in binpacking you should never fail on taints (any pod that couldn't schedule on template node would be filtered out earlier on). That leaves scale-down - with heavy use of dedicated nodepools you'd expect a lot of taint-related failures, but since we only check so many nodes / loop it may not matter. So, my gut feeling is that we won't see any scalability impact from this, but I'm still a bit worried.

losipiuk: We could make debugInfo a `func() string` and only lazyily evalate that on call to `VerboseError()`. That creates a problems related to thread safety though. Also as `nodeInfo` is mutable we will not be sure if what we evalate on call to `VerboseError()` is the same as when predicates were checked.

aleksandra-malinowska: Yeah, I recall that :/ We could just store a copy of node.Spec.Taints and lazily evaluate the message, but this would complicate the code a bit. @MaciekPytel do you want me to do this?

PR #2723, opened 2020-01-09
CA: Add Simulator related flags to FAQ

gjtempleton:
Adds a quick stanza on overriding the node skipping behaviour for local storage and kube-system pods.

Also adds these flags to the table of CLI flags for the CA as they were previously missing.


MaciekPytel:
/lgtm
/approve
Thanks!

PR #2732, opened 2020-01-13
Decrease the number of calls to VMSS API

marwanad:
`FindForInstance` is called every autoscaler loop (default 10s). In the case of a large volume scale down, you end up with a bunch of nodes in `NotReady` state until they get cleaned up by cloud-provider. This would lead to several cache misses and a call to `asgCache. regenerate()` which could possibly lead to a VMSS List for every registered node group. This would be multiplied by the number of not ready nodes (cache misses) in the worst case, O(numberOfNodes). This should be O(numberOfVmss).

Move the cache regeneration to `forceRefresh()` which gets called at the start of the autoscaler loop so that the virtual machine details get fetched with one List call for every VMSS. The frequency is now once per VMSS/registered ASG, instead of the node multiplier. This would match the granularity of the `Nodes()` call with 15 seconds calls if vmss size changes and 5 minutes if no change instead of having the vmss calls number scale linearly with the number of not ready nodes or cache misses.

Over a period of 1h: With one nodepool, starting node count of 3 and the same workload. The benefit is amplified in the longer term and larger number of nodes:

Before:
```
GET/VIRTUALMACHINESCALESETS/VIRTUALMACHINES:

324 calls

GET/VIRTUALMACHINESCALESETS

260 calls
```

After:

```
GET/VIRTUALMACHINESCALESETS/VIRTUALMACHINES:

38 calls

GET/VIRTUALMACHINESCALESETS

276 calls
```

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_cache.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_cache.go
@@ -17,7 +17,6 @@ limitations under the License.
 package azure
 
 import (
-	"fmt"
 	"reflect"
 	"regexp"
 	"strings"
@@ -153,27 +152,12 @@ func (m *asgCache) FindForInstance(instance *azureRef, vmType string) (cloudprov
 		}
 	}
 
-	klog.V(4).Infof("FindForInstance: cache before refresh %v", m.instanceToAsg)
-
 	// Look up caches for the instance.
+	klog.V(4).Infof("FindForInstance: attempting to retrieve instance %v from cache", m.instanceToAsg)
 	if asg := m.getInstanceFromCache(inst.Name); asg != nil {
-		klog.V(4).Infof("FindForInstance: returns before refresh, asg: %s", asg.Id())
+		klog.V(4).Infof("FindForInstance: found asg %s in cache", asg.Id())
 		return asg, nil
 	}
-
-	// Not found, regenerate the cache and try again.
-	if err := m.regenerate(); err != nil {
-		klog.Errorf("FindForInstance: error while looking for ASG for instance %q, error: %v", instance.Name, err)
-		return nil, fmt.Errorf("error while looking for ASG for instance %q, error: %v", instance.Name, err)
-	}
-
-	klog.V(4).Infof("FindForInstance: cache after refresh %v", m.instanceToAsg)
-
-	if asg := m.getInstanceFromCache(inst.Name); asg != nil {
-		klog.V(4).Infof("FindForInstance: returns after refresh, asg: %s", asg.Id())
-		return asg, nil
-	}
-
 	klog.V(4).Infof("FindForInstance: Couldn't find NodeGroup of instance %q", inst)
 	return nil, nil
 }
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/cloudprovider/azure/azure_manager.go
```diff
@@ -316,6 +316,9 @@ func (m *AzureManager) Refresh() error {
 func (m *AzureManager) forceRefresh() error {
 	if err := m.fetchAutoAsgs(); err != nil {
```
marwanad: we should probably move this logic outside - separate concern.

cluster-autoscaler/cloudprovider/azure/azure_cache.go
```diff
@@ -51,8 +50,6 @@ func newAsgCache(asgCacheTTL int64) (*asgCache, error) {
 	}
 
 	go wait.Until(func() {
-		cache.mutex.Lock()
-		defer cache.mutex.Unlock()
 		if err := cache.regenerate(); err != nil {
```
marwanad: I think we should also get rid of this. This TTL is useless with the changes here/we already refresh the instances cache way faster than this.

cluster-autoscaler/cloudprovider/azure/azure_manager.go
marwanad: we should refactor this logic out. This does a List scale and can possibly be merged with the `Nodes()` call - only affects autodiscovery though.

feiskyer: makes sense, could you add a TODO for it? we could do later after this PR.

marwanad:
@feiskyer Validated the scale up/down operations work fine. Doing one more round of testing and capturing metrics on the calls volume.

marwanad:
Tesult with K8s 1.14.8: 2 clusters with 3 nodepools each. First nodepool has initial count of 3 nodes and the other two nodepools have initial count of 1 node. The max size for both nodepool is 100. Trigger a scaleup that would resize all the nodepools to 100 nodes, then scale down the replicas of that deployment to 1. 

The calls are bucketed by 30 min. In the short term, you might encounter more calls, based on the `vmssSizeRefreshInterval` of `15 seconds` but long term, when you encounter several unready nodes, you don't thrash vmss list with calls.

djsly:
What about the 3min and 30min range. does your test shows a possible throttling ? 

what is long term ? would things settle back to normal after ~1h ?

marwanad:
> What about the 3min and 30min range. does your test shows a possible throttling ?
> 
> what is long term ? would things settle back to normal after ~1h ?
@djsly you might end up with 20-30 more calls in the 30 min range but with normal conditions, assuming scale up and scale downs and not ready nodes getting cleared up properly - I didn't encounter any throttling. 

Sorry, by long term I meant extreme cases with large scale down volumes and orphaned nodes in etcd. That being said, given that you avoid that large volume of vmss list calls, eventually you could get unthrottled (I can't confirm that because there could be other factors).

PR #2733, opened 2020-01-14
Remove taints with specific prefix from template node

MaciekPytel:
This is the same functionality as what you get with --ignore-taint
flag, but it doesn't require updating CA config to add a new taint
to ignore.

### DIFF
--- a/cluster-autoscaler/core/utils/utils_test.go
+++ b/cluster-autoscaler/core/utils/utils_test.go
@@ -274,6 +274,11 @@ func TestSanitizeTaints(t *testing.T) {
 		Value:  "1",
 		Effect: apiv1.TaintEffectNoSchedule,
 	})
+	taints = append(taints, apiv1.Taint{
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/core/utils/utils_test.go
```diff
@@ -274,6 +274,11 @@ func TestSanitizeTaints(t *testing.T) {
 		Value:  "1",
 		Effect: apiv1.TaintEffectNoSchedule,
 	})
+	taints = append(taints, apiv1.Taint{
+		Key:    "ignore-taint.cluster-autoscaler.kubernetes.io/to-be-ignored",
+		Value:  "I-am-the-invisible-man-Incredible-how-you-can",
```
losipiuk: +1

losipiuk: OTOH do we need copyright for that? Maybe reach out to Sony entertainment. ;)

losipiuk:
/lgtm
/approve
Thanks!

PR #2734, opened 2020-01-14
VPA - Handling empty vpa observed containers annotation

krzysied:
Empty string should be a correct case of VpaObservedContainers annotation.


vertical-pod-autoscaler/pkg/utils/annotations/vpa_observed_containers.go
```diff
@@ -42,6 +42,9 @@ func GetVpaObservedContainersValue(pod *v1.Pod) string {
 // ParseVpaObservedContainersValue returns list of containers
 // based on a given vpa observed containers annotation value.
 func ParseVpaObservedContainersValue(value string) ([]string, error) {
+	if value == "" {
+		return []string{}, nil
+	}
 	containerNames := strings.Split(value, listSeparator)
```
bskiba: what happens if we pass empty label value here?

krzysied: Split will return array with one string [""] and this string will not be a valid pod name :/

Issue #2637, opened 2019-12-11
AWS: CA doesn't work with Fargate Nodes in the same EKS cluster.


gusampaio:
I have an EKS Cluster 1.14 with platform version eks.5 (which the new Fargate feature is already supported), and every time that I have a Fargate node running under this Cluster, the CA doesn't scale up/down. The first version that I tried for CA was 1.14.6  and with this version, I wasn't able to get any error from CA logs.

After upgrading CA to 1.17.0, the error logs retrieved were following:

`
E1211 11:41:19.979658       1 static_autoscaler.go:226] Failed to get node infos for groups: wrong id: expected format aws:///<zone>/<name>, got aws:///eu-west-1b/120328e333-369fe1b7ba8b4416ab87216be91dc4f1/fargate-ip-192-168-159-181.eu-west-1.compute.internal
`

By checking this error, it seems to be an issue with the ProviderID format that Fargate creates for the node.

Based on the following snippet, I believe that the ProviderID generated by Fargate is not matching the regex check:

https://github.com/kubernetes/autoscaler/blob/2d80cb340da0ab2c2041ea8019f506a2d90e9a69/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go#L166

Once I don't have any Fargate node running under my Cluster, the CA starts working properly. Then if a new Fargate node joins my Cluster, the CA stops again.

dggondol:
This PR should help https://github.com/kubernetes/autoscaler/pull/2644

shanhchen:
I am seeing the same behavior on my cluster, as soon as I delete the fargate node, the cluster started to auto scale. However, I do not see any error message when fargate node is there. My EKS cluster is 1.14.

briananstett:
Any updates on this? Will this be in the next release?

stijndehaes:
Making a new 1.14.x release for this would be nice as EKS latest version is 1.14 :)

Jeffwan:
https://github.com/kubernetes/autoscaler/issues/2988 All newer versions have this change

nibatandukar:
@gusampaio How did you resolve this issue?

nibatandukar:
Hi @gusampaio . How did you resolve this issue?

PR #2737, opened 2020-01-14
Change the regex to match the providerID used by Fargate on EKS.

dggondol:
EKS now supports pods running on Fargate. The providerID has a different format:

"aws:///eu-west-1b/120328e333-369fe1b7ba8b4416ab87216be91dc4f1/fargate-ip-192-168-159-181.eu-west-1.compute.internal"

I just adapted the regex to match both patterns. Without this, the autoscaler will not work if you have one node (one pod) running on Fargate.

Issue: https://github.com/kubernetes/autoscaler/issues/2637


Jeffwan:
@dggondol Could you sign CLA first?

PR #2740, opened 2020-01-15
Remove filterOutSchedulableSimple

losipiuk:


cluster-autoscaler/core/filter_out_schedulable.go
```diff
@@ -67,20 +65,11 @@ func (filterOutSchedulablePodListProcessor) Process(
 	filterOutSchedulableStart := time.Now()
 	var unschedulablePodsToHelp []*apiv1.Pod
 
-	if context.EstimatorName == estimator.BinpackingEstimatorName {
-		unschedulablePodsToHelp = filterOutSchedulableByPacking(unschedulablePods, upcomingNodes, allScheduledPods,
-			context.PredicateChecker, context.ExpendablePodsPriorityCutoff, false)
-	} else {
-		unschedulablePodsToHelp = unschedulablePods
-	}
+	unschedulablePodsToHelp = filterOutSchedulableByPacking(unschedulablePods, upcomingNodes, allScheduledPods,
```
MaciekPytel: Add some comment here. It looks really strange that we're calling this twice like that and if I didn't already knew how it works, I imagine I'd spent a while figuring it out.

MaciekPytel:
We should remember to add to release notes that we deleted this flag. Any suggestions on how we make sure we remember it? Maybe create draft release already and add the note there?

cluster-autoscaler/core/filter_out_schedulable.go
losipiuk: I will. It gets refactored with ClusterSnapshot anyway :)

losipiuk:
> We should remember to add to release notes that we deleted this flag. Any suggestions on how we make sure we remember it? Maybe create draft release already and add the note there?

I always go through all the commits in delta between releases when doing release notes. If everyone follows this pattern it should not slip.
The only issue is that person doing release notes may have little understanding - and not catch what should be added just based on commit message.

Possibly release notes preparation should be more collaborative effort. I have some past experiences from different projects. We can talk and discuss that offline.

PR #2742, opened 2020-01-16
Vpa - Component status

krzysied:
Adding component status implementation based on Leases.

### DIFF
--- a/vertical-pod-autoscaler/pkg/utils/status/status_object.go
+++ b/vertical-pod-autoscaler/pkg/utils/status/status_object.go
@@ -1,0 +1,169 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package status
+
+import (
+	"net"
+	"time"
+
+	apicoordinationv1 "k8s.io/api/coordination/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilnet "k8s.io/apimachinery/pkg/util/net"
+	"k8s.io/apimachinery/pkg/util/wait"
+	clientset "k8s.io/client-go/kubernetes"
+	typedcoordinationv1 "k8s.io/client-go/kubernetes/typed/coordination/v1"
+	"k8s.io/utils/pointer"
+)
+
+const (
+	// Parameters for retrying with exponential backoff.
+	retryBackoffInitialDuration = 100 * time.Millisecond
+	retryBackoffFactor          = 3
+	retryBackoffJitter          = 0
+	retryBackoffSteps           = 3
+)
+
+// Client for the status object.
+type Client struct {
+	client               typedcoordinationv1.LeaseInterface
+	leaseName            string
+	leaseNamespace       string
+	leaseDurationSeconds int32
+	holderIdentity       string
+}
+
+// NewClient returns a client for the status object.
+func NewClient(c clientset.Interface, leaseName, leaseNamespace string, leaseDuration time.Duration, holderIdentity string) *Client {
+	return &Client{
+		client:               c.CoordinationV1().Leases(leaseNamespace),
+		leaseName:            leaseName,
+		leaseNamespace:       leaseNamespace,
+		leaseDurationSeconds: int32(leaseDuration.Seconds()),
+		holderIdentity:       holderIdentity,
+	}
+}
+
+// UpdateStatus renews status object lease.
+// Status object will be created if it doesn't exist.
+func (c *Client) UpdateStatus() error {
+	updateFn := func() error {
+		lease, err := c.client.Get(c.leaseName, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			// Create lease if it doesn't exist.
+			return c.create()
+		} else if err != nil {
+			return err
+		}
+		lease.Spec.RenewTime = &metav1.MicroTime{Time: time.Now()}
+		_, err = c.client.Update(lease)
+		if apierrors.IsConflict(err) {
+			// Lease was updated by an another replica of the component.
+			// No error should be returned.
+			return nil
+		}
+		return err
+	}
+	return retryWithExponentialBackOff(updateFn)
+}
+
+func (c *Client) create() error {
+	_, err := c.client.Create(c.newLease())
+	if apierrors.IsAlreadyExists(err) {
+		// Lease was created by an another replica of the component.
+		// No error should be returned.
+		return nil
+	}
+	return err
+}
+
+func (c *Client) newLease() *apicoordinationv1.Lease {
+	return &apicoordinationv1.Lease{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      c.leaseName,
+			Namespace: c.leaseNamespace,
+		},
+		Spec: apicoordinationv1.LeaseSpec{
+			HolderIdentity:       pointer.StringPtr(c.holderIdentity),
+			LeaseDurationSeconds: pointer.Int32Ptr(c.leaseDurationSeconds),
+		},
+	}
+}
+
+// GetStatus returns status object.
+func (c *Client) GetStatus() (*apicoordinationv1.Lease, error) {
+	var lease *apicoordinationv1.Lease
+	getFn := func() error {
+		var err error
+		lease, err = c.client.Get(c.leaseName, metav1.GetOptions{})
+		return err
+	}
+	err := retryWithExponentialBackOff(getFn)
+	return lease, err
+}
+
+// IsStatusValid verifies if status was updated before lease timing out.
+func IsStatusValid(status *apicoordinationv1.Lease, leaseTimeout time.Duration) bool {
+	return isStatusValid(status, leaseTimeout, time.Now())
+}
+
+func isStatusValid(status *apicoordinationv1.Lease, leaseTimeout time.Duration, now time.Time) bool {
+	return status.CreationTimestamp.Add(leaseTimeout).After(now) ||
+		status.Spec.RenewTime.Add(leaseTimeout).After(now)
+}
+
+func isRetryableAPIError(err error) bool {
+	// These errors may indicate a transient error that we can retry.
+	if apierrors.IsInternalError(err) || apierrors.IsTimeout(err) || apierrors.IsServerTimeout(err) ||
+		apierrors.IsTooManyRequests(err) || utilnet.IsProbableEOF(err) || utilnet.IsConnectionReset(err) {
+		return true
+	}
+	// If the error sends the Retry-After header, we respect it as an explicit confirmation we should retry.
+	if _, shouldRetry := apierrors.SuggestsClientDelay(err); shouldRetry {
+		return true
+	}
+	return false
+}
+
+func isRetryableNetError(err error) bool {
+	if netError, ok := err.(net.Error); ok {
+		return netError.Temporary() || netError.Timeout()
+	}
+	return false
+}
+
+func retryWithExponentialBackOff(fn func() error) error {
+	backoff := wait.Backoff{
+		Duration: retryBackoffInitialDuration,
+		Factor:   retryBackoffFactor,
+		Jitter:   retryBackoffJitter,
+		Steps:    retryBackoffSteps,
+	}
+	retryFn := func(fn func() error) func() (bool, error) {
+		return func() (bool, error) {
+			err := fn()
+			if err == nil {
+				return true, nil
+			}
+			if isRetryableAPIError(err) || isRetryableNetError(err) {
+				return false, nil
+			}
+			return false, err
+		}
+	}
+	return wait.ExponentialBackoff(backoff, retryFn(fn))
+}
--- a/vertical-pod-autoscaler/pkg/utils/status/status_object_test.go
+++ b/vertical-pod-autoscaler/pkg/utils/status/status_object_test.go
@@ -1,0 +1,265 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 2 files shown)


krzysied:
Vendor update moved to #2743 
/hold

vertical-pod-autoscaler/pkg/utils/status/status_object.go
```diff
@@ -0,0 +1,163 @@
...
+	leaseDurationSeconds int32
+}
+
+// NewClient returns a client for the status object.
+func NewClient(c clientset.Interface, name, namespace string, leaseDuration time.Duration) *Client {
+	return &Client{
+		client:               c.CoordinationV1().Leases(namespace),
+		name:                 name,
+		namespace:            namespace,
+		leaseDurationSeconds: int32(leaseDuration) / int32(time.Second),
```
bskiba: or maybe int32(leaseDuration.Seconds())? 
https://golang.org/pkg/time/#Duration.Seconds

vertical-pod-autoscaler/pkg/utils/status/status_object.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+}
+
+// UpdateStatus renews status object lease.
+// Status object will be created if it doesn't exist.
+func (c *Client) UpdateStatus() error {
```
bskiba: How often will this be called in practice?

vertical-pod-autoscaler/pkg/utils/status/status_object.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func (c *Client) newLease() *apicoordinationv1.Lease {
+	return &apicoordinationv1.Lease{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      c.name,
+			Namespace: c.namespace,
+		},
+		Spec: apicoordinationv1.LeaseSpec{
+			HolderIdentity:       pointer.StringPtr(c.name),
```
bskiba: Shouldn't HolderIdentity be different for each holder? Then it can't be equal to lease object name.

vertical-pod-autoscaler/pkg/utils/status/status_object.go
```diff
@@ -0,0 +1,163 @@
...
+	retryBackoffInitialDuration = 100 * time.Millisecond
+	retryBackoffFactor          = 3
+	retryBackoffJitter          = 0
+	retryBackoffSteps           = 6
+)
+
+// Client for the status object.
+type Client struct {
+	client               typedcoordinationv1.LeaseInterface
+	name                 string
```
bskiba: Is this the name of the lease or the name of the client?

vertical-pod-autoscaler/pkg/utils/status/status_object.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// IsStatusValid verifies if status was updated before lease timing out.
+func IsStatusValid(status *apicoordinationv1.Lease, leaseTimeout time.Duration) bool {
+	return status.CreationTimestamp.Add(leaseTimeout).After(time.Now()) ||
+		status.Spec.RenewTime.Add(leaseTimeout).After(time.Now())
+}
+
+func isRetryableAPIError(err error) bool {
+	// These errors may indicate a transient error that we can retry in tests.
```
bskiba: Why does it say we can retry in tests?

vertical-pod-autoscaler/pkg/utils/status/status_object.go
krzysied: Good idea. Done.

vertical-pod-autoscaler/pkg/utils/status/status_object.go
krzysied: holderIdentity filed added.

vertical-pod-autoscaler/pkg/utils/status/status_object.go
krzysied: The name of the lease.

vertical-pod-autoscaler/pkg/utils/status/status_object.go
krzysied: Copy-paste error. I took list of retryable error from Cluster Loader code.

vertical-pod-autoscaler/pkg/utils/status/status_object.go
krzysied: I want it to use for Admission Controller. I want to call `UpdateStatus` every 10s. The same practice is used for node leases.
Assuming that `UpdateStatus` correlates with number of master, the generated qps should be negligible.

I also used explicit get for retrieving status instead of using watch due to correctness reasons. We are not generating heavy load on master.

vertical-pod-autoscaler/pkg/utils/status/status_object.go
bskiba: Then please make it explicit. It's name field of Client struct - my first guess was that this is client name.

vertical-pod-autoscaler/pkg/utils/status/status_object.go
bskiba: What correctness will we sacrifice with watch? 
We might try to create an already existing object, which will just cause the create to fail. (This may happen anyway if someone creates lease between our get and create)
For Updating we handle conflicts anyway in line 74.
Am I missing something?

I agree this might not be a huge load on master, but we are tight on out qps to API server anyway, so I wan't to make sure we're not making things worse.

vertical-pod-autoscaler/pkg/utils/status/status_object.go
```diff
@@ -0,0 +1,165 @@
...
+	clientset "k8s.io/client-go/kubernetes"
+	typedcoordinationv1 "k8s.io/client-go/kubernetes/typed/coordination/v1"
+	"k8s.io/utils/pointer"
+)
+
+const (
+	// Parameters for retrying with exponential backoff.
+	retryBackoffInitialDuration = 100 * time.Millisecond
+	retryBackoffFactor          = 3
+	retryBackoffJitter          = 0
```
bskiba: Why 0 Jitter?

vertical-pod-autoscaler/pkg/utils/status/status_object.go
```diff
@@ -0,0 +1,165 @@
...
+		},
+		Spec: apicoordinationv1.LeaseSpec{
+			HolderIdentity:       pointer.StringPtr(c.holderIdentity),
+			LeaseDurationSeconds: pointer.Int32Ptr(c.leaseDurationSeconds),
+		},
+	}
+}
+
+// GetStatus returns status object.
+func (c *Client) GetStatus() (*apicoordinationv1.Lease, error) {
```
bskiba: An alternative to retrying getting the lease object 6 times in GetStatus() is simply skipping actuation in updater if GetStatus fails. We will retry in 1 minute anyway.

vertical-pod-autoscaler/pkg/utils/status/status_object.go
krzysied: Watch can lag. Explicit get guaranties up-to-date data.
My assumption is that explicit get is always better than watch. Imho watch should be used if this could potentially reduce load on apiserver. This is not our case here.
Plus, explicit get is easier to implement and to understand than watch.

Implement explicit get and then if it necessary changing to watch is also an option ;)
Btw, I can reach to scalability folks and ask about their option.

vertical-pod-autoscaler/pkg/utils/status/status_object.go
krzysied: Done.

vertical-pod-autoscaler/pkg/utils/status/status_object.go
krzysied: Do we need  jitter?

vertical-pod-autoscaler/pkg/utils/status/status_object.go
krzysied: I agree, but I wanted to have generic module for status. Plus, retries are always a good idea.
Maybe 6 time is overkill, I will change it to 3...

vertical-pod-autoscaler/pkg/utils/status/status_object.go
bskiba: I cannot agree that explicit get is always better than watch but I agree with having it here.

vertical-pod-autoscaler/pkg/utils/status/status_object.go
bskiba: Can't agree retries are always a good idea. The retryWithBackoff function is used here for GetStatus and for UpdateStatus, which will be called every 10 seconds, right? Removing retrywithbackoff simplifies things a lot here. 
Might be worth considering. I won't be very stuborn if you want to keep it though.

vertical-pod-autoscaler/pkg/utils/status/status_object_test.go
```diff
@@ -0,0 +1,254 @@
...
+							return true, &apicoordinationv1.Lease{}, nil
+						default:
+							return true, nil, apierrors.NewConflict(schema.GroupResource{}, leaseName, nil)
+						}
+					}
+
+					return true, nil, fmt.Errorf("unsupported action")
+				}
+			}(),
+			wantErr: false,
```
bskiba: No case with wantErr = true?

vertical-pod-autoscaler/pkg/utils/status/status_object_test.go
```diff
@@ -0,0 +1,254 @@
...
+		})
+	}
+}
+
+func TestIsStatusValid(t *testing.T) {
+	tests := []struct {
+		name         string
+		lease        *apicoordinationv1.Lease
+		leaseTimeout time.Duration
+		want         bool
```
bskiba: s/want/expectedValid/

vertical-pod-autoscaler/pkg/utils/status/status_object_test.go
```diff
@@ -0,0 +1,254 @@
...
+					RenewTime: &metav1.MicroTime{Time: time.Now().Add(-time.Minute)},
+				},
+			},
+			leaseTimeout: 10 * time.Second,
+			want:         false,
+		},
+	}
+	for _, tc := range tests {
+		t.Run(fmt.Sprintf("test case: %s", tc.name), func(t *testing.T) {
+			assert.Equal(t, IsStatusValid(tc.lease, tc.leaseTimeout), tc.want)
```
bskiba: Please add a message for failed equality check, otherwise we get "expected: false got: true" only on test failure

vertical-pod-autoscaler/pkg/utils/status/status_object_test.go
```diff
@@ -0,0 +1,254 @@
...
+			lease: &apicoordinationv1.Lease{
+				ObjectMeta: metav1.ObjectMeta{
+					CreationTimestamp: metav1.Time{Time: time.Now()},
+				},
+			},
+			leaseTimeout: 10 * time.Second,
+			want:         true,
+		},
+		{
+			name: "Renew time check",
```
bskiba: I don't understand this test case name

vertical-pod-autoscaler/pkg/utils/status/status_object.go
krzysied: > Removing retrywithbackoff simplifies things a lot here.

I agree.

> GetStatus and for UpdateStatus, which will be called every 10 seconds, right?

UpdateStatus will be called every 10 seconds. if no update was made for 1m we assume that the component is not available. Therefore we have ~6 six attempts to update the status. Retrying every single update doesn't seem to be necessary.
The question is what approach we have for GetStatus?... If GetStatus fails, does it means that component is not available? It needs to be retried somehow. There are three solutions that I see:
-  GetStatus has retries. At the beginning of the eviction in Updater we call GetStatus. If it returns error or outdated lease, no eviction is made.
- We remove retries from GetStatus. We have single `GetStatus` call in Updater, so it happens that some evictions iterations are skipped due to random cluster errors.
- We remove retries form  GetStatus. We keep checking GetStatus every 10s and sore the result (when there is no error) in Updater. At the beginning of eviction loop we checked component availability based on the last stored status.

I don't see a problem with implementing third solution instead of the first one. I'm not sure if i'm happy with second though...
Wdyt?

krzysied: After offline discussion with @bskiba we decided to keep retries as they are.

vertical-pod-autoscaler/pkg/utils/status/status_object_test.go
krzysied: Test case added.

vertical-pod-autoscaler/pkg/utils/status/status_object_test.go
krzysied: Done.

vertical-pod-autoscaler/pkg/utils/status/status_object_test.go
krzysied: I've updated test names to be more self-explanatory. Imho simple "expected, got" is enough.

vertical-pod-autoscaler/pkg/utils/status/status_object_test.go
krzysied: Test names updated.

vertical-pod-autoscaler/pkg/utils/status/status_object_test.go
```diff
@@ -0,0 +1,264 @@
...
+				Spec: apicoordinationv1.LeaseSpec{
+					RenewTime: &metav1.MicroTime{Time: time.Now().Add(-time.Minute)},
+				},
+			},
+			leaseTimeout:  10 * time.Second,
+			expectedValid: false,
+		},
+	}
+	for _, tc := range tests {
+		t.Run(fmt.Sprintf("test case: %s", tc.name), func(t *testing.T) {
```
bskiba: These test cases are indeterministic. I think you want an injectable time.Now() if you's like to test this properly.

bskiba:
One more comment to the tests

vertical-pod-autoscaler/pkg/utils/status/status_object_test.go
krzysied: I added `innerClock` in status_object.go.
I'm not happy with package level variable, but I don't see another way to inject time to `IsStatusValid` method :/

bskiba: A simpler alternative is to have a helper function isStatusValid(lease, timeout, now)  test it instead and have IsStatusValid just call(isStatusValid, timeout, time.Now() )

GarbageCollection of container states does sth like this: https://github.com/kubernetes/autoscaler/blob/b232915c2485345d3ae8e587bc3fa933bc5c70d9/vertical-pod-autoscaler/pkg/recommender/model/cluster.go#L331

krzysied: Done.

PR #2745, opened 2020-01-16
Propagate scaling mode to containers. Correctly update ResourcePolicy and UpdatePolicy on update.

bskiba:


bskiba:
@kgolab could you review? I know it's XL but the biggest updates are to the tests. 
Also, not very urgent.

vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state.go
```diff
@@ -97,6 +97,7 @@ type AggregateContainerState struct {
 	LastRecommendation corev1.ResourceList
 	IsUnderVPA         bool
 	UpdateMode         *vpa_types.UpdateMode
+	scalingMode        *vpa_types.ContainerScalingMode
```
kgolab: Why is this one private while others are exported?

vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state.go
```diff
@@ -241,6 +249,17 @@ func (a *AggregateContainerState) isEmpty() bool {
 	return a.TotalSamplesCount == 0
 }
 
+// UpdateFromPolicy updates container state scaling mode based on reasource
```
kgolab: nit: s/reasource/resource/

vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state_test.go
```diff
@@ -208,3 +208,42 @@ func TestAggregateContainerStateIsExpired(t *testing.T) {
...
+			},
+			expected: &scalingModeAuto,
+		}, {
+			name: "Off scaling mode",
+			policy: &vpa_types.ContainerResourcePolicy{
+				Mode: &scalingModeOff,
+			},
+			expected: &scalingModeOff,
+		}, {
+			name:     "No mode specified - dafault to Auto",
```
kgolab: nit: s/dafault/default/

vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state_test.go
```diff
@@ -208,3 +208,42 @@ func TestAggregateContainerStateIsExpired(t *testing.T) {
...
+			policy:   &vpa_types.ContainerResourcePolicy{},
+			expected: &scalingModeAuto,
+		}, {
+			name:     "Nil policy - dafault to Auto",
```
kgolab: dafault again

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
```diff
@@ -235,13 +235,15 @@ func TestMissingKeys(t *testing.T) {
 }
 
 func addVpa(cluster *ClusterState, id VpaID, annotations vpaAnnotationsMap, selector string) *Vpa {
-	var apiObject vpa_types.VerticalPodAutoscaler
-	apiObject.Namespace = id.Namespace
-	apiObject.Name = id.VpaName
-	apiObject.Annotations = annotations
+	apiObject := test.VerticalPodAutoscaler().WithNamespace(id.Namespace).
+		WithName(id.VpaName).WithContainer(testContainerID.ContainerName).WithAnnotations(annotations).Get()
```
kgolab: You've added WithContainer but I don't immediately see why. Could you please explain (here, not in code) so we can make sure this was on purpose?

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
```diff
@@ -340,6 +342,120 @@ func TestUpdatePodSelector(t *testing.T) {
...
+		oldVpa              *vpa_types.VerticalPodAutoscaler
+		newVpa              *vpa_types.VerticalPodAutoscaler
+		resourcePolicy      *vpa_types.PodResourcePolicy
+		expectedScalingMode *vpa_types.ContainerScalingMode
+	}{
+		{
+			name:                "New VPA",
+			oldVpa:              nil,
+			newVpa:              testVpaBuilder.WithUpdateMode(vpa_types.UpdateModeOff).Get(),
+			expectedScalingMode: &scalingModeAuto,
```
kgolab: This look non-trivial. I understand that it's due to UpdateModeOff being on VPA level while the expectation is on container level but maybe a comment here would help future readers.

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
```diff
@@ -340,6 +342,120 @@ func TestUpdatePodSelector(t *testing.T) {
...
+				ContainerPolicies: []vpa_types.ContainerResourcePolicy{
+					{
+						ContainerName: "*",
+						Mode:          &scalingModeOff,
+					},
+				},
+			},
+			expectedScalingMode: &scalingModeOff,
+		}, {
+			name:   "New VPA with explicit scaling mode Off",
```
kgolab: Another test case could be to double-check that having scalingModeOff for another container does not break the one we're testing.

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
```diff
@@ -340,6 +342,120 @@ func TestUpdatePodSelector(t *testing.T) {
...
+				ContainerPolicies: []vpa_types.ContainerResourcePolicy{
+					{
+						ContainerName: testContainerID.ContainerName,
+						Mode:          &scalingModeOff,
+					},
+				},
+			},
+			expectedScalingMode: &scalingModeOff,
+		}, {
+			name:                "UpdateMode from Off to Auto",
```
kgolab: I'm not sure about the value of this and the following test case - the initial ("New VPA") already shows that we ignore VPA-level UpdateMode.
Or is something interesting happening here?

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
```diff
@@ -340,6 +342,120 @@ func TestUpdatePodSelector(t *testing.T) {
...
+			expectedScalingMode: &scalingModeAuto,
+		}, {
+			name:                "UpdateMode from Auto to Off",
+			oldVpa:              testVpaBuilder.WithUpdateMode(vpa_types.UpdateModeAuto).Get(),
+			newVpa:              testVpaBuilder.WithUpdateMode(vpa_types.UpdateModeOff).Get(),
+			expectedScalingMode: &scalingModeAuto,
+		}, {
+			name:   "Scaling mode to default Off",
+			oldVpa: testVpaBuilder.WithUpdateMode(vpa_types.UpdateModeAuto).Get(),
+			newVpa: testVpaBuilder.WithUpdateMode(vpa_types.UpdateModeOff).Get(),
```
kgolab: This will work even if you don't change UpdateMode, right? And it would make the test intent clearer.

vertical-pod-autoscaler/pkg/recommender/model/vpa_test.go
```diff
@@ -273,66 +276,136 @@ func TestUseAggregationIfMatching(t *testing.T) {
...
+						Mode:          &scalingModeOff,
+					},
+				},
+			},
+			updateMode:                  &modeAuto,
+			container:                   "second-container",
+			containerLabels:             testLabels,
+			expectedMatching:            true,
+			expectedAggregations:        []string{"test-container", "second-container"},
+			expectedNeedsRecommendation: false,
```
kgolab: Can we also (easily) check here that the "test-container" still has needsRecommendation == true?

vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state.go
```diff
@@ -115,12 +116,19 @@ func (a *AggregateContainerState) GetUpdateMode() *vpa_types.UpdateMode {
 	return a.UpdateMode
 }
 
+// GetScalingMode returns the container scaling mode of the container
+// represented byt his aggregator, nil if aggregator is not autoscaled.
+func (a *AggregateContainerState) GetScalingMode() *vpa_types.ContainerScalingMode {
+	return a.scalingMode
+}
+
 // MarkNotAutoscaled registers that this container state is not contorled by
```
kgolab: nit: s/contorled/controlled/

vertical-pod-autoscaler/pkg/utils/test/test_vpa.go
```diff
@@ -38,6 +38,7 @@ type VerticalPodAutoscalerBuilder interface {
 	WithLowerBound(cpu, memory string) VerticalPodAutoscalerBuilder
 	WithTargetRef(targetRef *autoscaling.CrossVersionObjectReference) VerticalPodAutoscalerBuilder
 	WithUpperBound(cpu, memory string) VerticalPodAutoscalerBuilder
+	WithAnnotations(map[string]string) VerticalPodAutoscalerBuilder
```
kgolab: Unless I missed something, this change looks unrelated and unused.

kgolab:
Sorry for such a delay.

The code looks good, I've left some minor comments and possible test improvements.
What I'm missing though is the context - why do we need this change?
And do we need so much data in AggregateContainerState? For example it looks like maybe we could get rid of UpdateMode, it does not affect how we treat containers in Recommender.

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
bskiba: Added comment

vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state.go
bskiba: Done

vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state.go
bskiba: Done

vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state_test.go
bskiba: Done

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
bskiba: VPA builder panics if you call get before setting container name.

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
bskiba: VPA builder panics if you call get before setting container name.

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
bskiba: Done

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
bskiba: This test also checks that we update updateMode correctly. Moved those cases to the end and added a comment.

vertical-pod-autoscaler/pkg/recommender/model/vpa_test.go
bskiba: Done

vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state.go
bskiba: Done

vertical-pod-autoscaler/pkg/utils/test/test_vpa.go
bskiba: This is due to the change in cluster_test addVpa function - I started using builder and that function sets annotations as well.

vertical-pod-autoscaler/pkg/recommender/model/cluster_test.go
bskiba: Done

bskiba:
Addressed comments.

This change is needed so that we can correctly report if container is autoscaled. This in turn is used by quality metrics to report missing recommendations correctly. 

This mostly comes from the fact that when we add a usage sample, we add it to container state and we don't have a direct link to the VPA object that is governing it (legacy of the fact that this used to be a 1:n relation - on aggregate could contribute to multiple VPA). Getting rid of this is possible if we refactor the code to reflect the 1:1 relationship, but it is a major change.

Let me double check why UpdateMode was needed. I can also add comments in aggregateContainerState why we need those fields

bskiba:
Update mode is needed for quality metrics as well. I've added a comment on this to AggregateContainerState

kgolab:
Thanks, LGTM.

Issue #1578, opened 2019-01-14
Why not to drain daemonset pods? 


calixwu:
I read the source code about the scaling-down part, and I found the cluster-autoscaler never drain the daemonset pods when scaling down. Wouldn't this cause problems?

aleksandra-malinowska:
There's no point evicting DaemonSet pods, as they can't be scheduled on another node anyway. You're correct this means in particular we won't attempt any sort of graceful termination other than what kubelet may provide when shutting down. I'm not really sure if there's any mechanism we could utilize for it.

MaciekPytel:
Note CA only respects graceful termination of up to 10 minutes anyway. Evicting daemonset pods if we're draining a node probably wouldn't be difficult, but the empty nodes case is a large issue. A lot of people rely on the fact that they're deleted in parallel and changing that would be a huge regression.

calixwu:
> Note CA only respects graceful termination of up to 10 minutes anyway. Evicting daemonset pods if we're draining a node probably wouldn't be difficult, but the empty nodes case is a large issue. A lot of people rely on the fact that they're deleted in parallel and changing that would be a huge regression.

@MaciekPytel 
Could we do these to solve the empty-nodes issue? :
1. We ignore the daemonset pods when we try to find out the empty nodes.
2. We do not ignore the  daemonset pods when we try to delete the empty nodes. Then add the "delete traint" to the empty nodes and evict the daemonset pods in parallel, after the empty nodes being actually empty (no daemonset pods on them), we call the cloud provider's deletion in parallel.

aleksandra-malinowska:
> Evicting daemonset pods if we're draining a node probably wouldn't be difficult

It may be harder than we think. We wait for pods to be scheduled back onto some other nodes, but with DaemonSet, each pod has a node selector for a particular node, and we don't want it to run anymore. These pods also seem to ignore the fact that node is not ready or unschedulable (which of course makes sense when starting a node that requires a DaemonSet to be ready), so they may just be scheduled back anyway. Finally, I recall some issues where an extra DaemonSet pod was left pending, so we'd have to make sure whatever we do is correctly handled by the controller.

Also, by default, `kubectl drain` doesn't try to remove DaemonSets, either. However, there's a flag `--ignore-daemonsets=false`, so it may be a starting point to investigate how it works.

calixwu:
[...]

PR #2759, opened 2020-01-21
cluster-autoscaler: support evict DaemonSet pods in scale down

oif:
## Background

Currently CA scale down ignore DaemonSet controlled pods, but in a production cluster, this may result in those pods killed not graceful. may lose data, etc.
And currently, kubelet will not handle with Pods graceful shutdown will be shutting down. And also cloud providers may(there is no restriction) kill node directly.

## Aim to

Handle with DaemonSets graceful shutdown while scaling down

ref #1578


joekohlsdorf:
I believe having this as a global setting is not good enough. Eviction might be necessary for some DaemonSets but not others, for example a log forwarder should be the last pod to go.

Instead of the proposed change I propose to evict DaemonSet pods if the `cluster-autoscaler.kubernetes.io/safe-to-evict` annotation is present.

oif:
@joekohlsdorf 

Maybe use `cluster-autoscaler.kubernetes.io/safe-to-evict=false` is more suitable for this once enabled daemonset eviction in my mind.

This PR keep pending for a long time :(

Issue #2738, opened 2020-01-15
VPA - Admission Controller status


krzysied:
It can potentially happen that Admission Controller is down. When this happens, we don't want to evict pods (recommendation will not be applied).
Admission Controller should have a status object. This status object should be observed by Updater. If status object has not been updated for a long time, Updater should stop evictions.

krzysied:
Implementation plan:
- [ ] Generic status util. (#2742)
- [ ] Status updating by Admission Controller
- [ ] Status checking by Updater

krzysied:
Tests seem to work https://testgrid.k8s.io/sig-autoscaling-vpa#autoscaling-vpa-updater.
Closing the issue.

hjkatz:
We were also affected by something similar to #2633 and #2715. Our solution was to restart the admission controller pod and then the recommendation targets seemed to be applied correctly. When will these merged changes be cut into a new release?

krzysied:
@bskiba Do we have any estimation for a new release?

PR #2763, opened 2020-01-22
VPA -  Adding status to Admission Controller

krzysied:
Adding status object (Lease) to Admission Controller.
Adding rbac entries for Leases manipulation.

ref #2738

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/logic/status_updater.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/logic/status_updater.go
@@ -1,0 +1,70 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package logic
+
+import (
+	"time"
+
+	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/status"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+)
+
+const (
+	// VpaAdmissionControllerStatusName is the name of
+	// the Admission Controller status object.
+	VpaAdmissionControllerStatusName = "vpa-admission-controller"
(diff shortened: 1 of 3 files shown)


vertical-pod-autoscaler/pkg/admission-controller/logic/status_updater.go
```diff
@@ -0,0 +1,77 @@
...
+// Start starts status updates.
+func (su *StatusUpdater) Start() {
+	go func() {
+		for {
+			select {
+			case <-su.stopCh:
+				return
+			case <-time.After(updateInterval):
+				if err := su.client.UpdateStatus(); err != nil {
+					klog.Errorf("Admission Controller status update failed: %v", err)
```
bskiba: nit: I think Warning should be enough

vertical-pod-autoscaler/pkg/admission-controller/logic/status_updater.go
```diff
@@ -0,0 +1,77 @@
...
+			VpaAdmissionControllerStatusNamespace,
+			updateInterval,
+			holderIdentity,
+		),
+		stopCh: make(chan struct{}),
+	}
+}
+
+// Start starts status updates.
+func (su *StatusUpdater) Start() {
```
bskiba: nit: isn't Run more popular for k8s controllers?

vertical-pod-autoscaler/go.mod
```diff
@@ -5,6 +5,7 @@ go 1.12
 require (
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/golang/mock v1.3.1
+	github.com/google/uuid v1.1.1
```
bskiba: Did you run go mod vendor? Not sure if this should cause a change in vendor forlder

vertical-pod-autoscaler/pkg/admission-controller/logic/status_updater.go
krzysied: Imho we want to have Updater logic to be dependent on this status, so error sounds reasonable.
Plus, there are retries implemented, so this situation shouldn't happen unless something is really off.

vertical-pod-autoscaler/pkg/admission-controller/logic/status_updater.go
krzysied: Changed to `Run(<-chan struct{})` signature.

vertical-pod-autoscaler/go.mod
krzysied: I realized that uuid was used during implementation, but I removed it from the final version...
I'm removing go.mod/go.sum changes.

Also, it seems that I messed some vendor update in one of the previous commit :/ The change is unrelated to current PR, so I will fix it in separate PR.

PR #2771, opened 2020-01-27
VPA - Updater timer ObserveTotal cleanup

krzysied:
Simplifying `ObserveTotal` logic. It will be called always when existing Updater's `RunOnce` method.


bskiba:
Thanks!
/lgtm
/approve

PR #2772, opened 2020-01-27
VPA - Checking Admission Controller status

krzysied:
Update check if there is valid Admission Controller status.
If not, eviction loop will be skipped.

### DIFF
--- a/vertical-pod-autoscaler/pkg/updater/logic/updater.go
+++ b/vertical-pod-autoscaler/pkg/updater/logic/updater.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/updater/eviction"
 	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/updater/priority"
 	metrics_updater "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/metrics/updater"
+	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/status"
 	vpa_api_util "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/vpa"
 	kube_client "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/fake"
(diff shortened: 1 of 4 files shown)


krzysied:
@bskiba Do we want to enable status logic at once and it's ok for tests to be red for a time 
or do we want to enable it step by step?

bskiba:
I'd like see fix for tests first.

bskiba:
It's actually not a lot of work - merge updater and actuation tests into one.

krzysied:
With [tests being merged](https://github.com/kubernetes/autoscaler/issues/2777), I think this PR is ready for a review.
@bskiba PTAL

vertical-pod-autoscaler/pkg/updater/logic/updater.go
```diff
@@ -51,32 +53,53 @@ type Updater interface {
...
+		eventRecorder:                newEventRecorder(kubeClient),
+		evictionFactory:              factory,
+		recommendationProcessor:      recommendationProcessor,
+		evictionRateLimiter:          evictionRateLimiter,
+		evictionAdmission:            evictionAdmission,
+		selectorFetcher:              selectorFetcher,
+		useAdmissionControllerStatus: useAdmissionControllerStatus,
+		statusClient: status.NewClient(
+			kubeClient,
+			amission_controller_logic.VpaAdmissionControllerStatusName,
```
bskiba: It seems weird that updater would import admission_controller_logic. Perhaps we can move the consts to the status lib which is common?

vertical-pod-autoscaler/pkg/updater/logic/updater.go
```diff
@@ -85,6 +108,18 @@ func (u *updater) RunOnce(ctx context.Context) {
 	timer := metrics_updater.NewExecutionTimer()
 	defer timer.ObserveTotal()
 
+	if u.useAdmissionControllerStatus {
+		s, err := u.statusClient.GetStatus()
+		if err != nil {
+			klog.Errorf("Admission Controller status getting error: %v. Skipping eviction loop", err)
```
bskiba: nit: "Error getting Admission Controller status: ..."

vertical-pod-autoscaler/pkg/updater/logic/updater.go
```diff
@@ -85,6 +108,18 @@ func (u *updater) RunOnce(ctx context.Context) {
...
+			return
+		}
+		if !status.IsStatusValid(s, time.Minute) {
```
bskiba: the time.Minute needs to be a const

vertical-pod-autoscaler/pkg/updater/logic/updater.go
```diff
@@ -85,6 +108,18 @@ func (u *updater) RunOnce(ctx context.Context) {
...
+			klog.Warningf("Admission Controller status %v is not valid. Skipping eviction loop", s)
```
bskiba: Do you think printing only the Lease will be enough for debugging? Maybe we can also return from isValidStatus an easily parsable reason why it is invalid?

vertical-pod-autoscaler/pkg/updater/logic/updater.go
```diff
@@ -85,6 +108,18 @@ func (u *updater) RunOnce(ctx context.Context) {
 	timer := metrics_updater.NewExecutionTimer()
 	defer timer.ObserveTotal()
 
+	if u.useAdmissionControllerStatus {
+		s, err := u.statusClient.GetStatus()
```
bskiba: This flow actually has me thinking, shouldn't statusClient just expose IsStatusValid? Updater doesn't care if status is implemented as a Lease object and it shouldn't have to parse it.
If you hide this behind an interface with just one method, then in the tests you can use a mock object that just returns a boolean.

vertical-pod-autoscaler/pkg/updater/main.go
```diff
@@ -53,6 +53,9 @@ var (
 	evictionRateBurst = flag.Int("eviction-rate-burst", 1, `Burst of pods that can be evicted.`)
 
 	address = flag.String("address", ":8943", "The address to expose Prometheus metrics.")
+
+	useAdmissionControllerStatus = flag.Bool("use-admission-controller-status", true,
+		"If Admission Contoller status should be used for making eviction decisions")
```
bskiba: More descriptive would be good:
If true, updater will only evict pods when admission controller status is valid.

vertical-pod-autoscaler/pkg/updater/logic/updater.go
krzysied: Good idea. I created https://github.com/kubernetes/autoscaler/pull/2785 to do this.

vertical-pod-autoscaler/pkg/updater/logic/updater.go
krzysied: I moved it to separated PR to not increase size of current PR: https://github.com/kubernetes/autoscaler/pull/2788

vertical-pod-autoscaler/pkg/updater/logic/updater.go
krzysied: Done.

vertical-pod-autoscaler/pkg/updater/logic/updater.go
krzysied: Currently we don't have access to the lease. I'm printing timeout duration.

vertical-pod-autoscaler/pkg/updater/main.go
krzysied: Done.

vertical-pod-autoscaler/pkg/updater/logic/updater.go
krzysied: Done.

vertical-pod-autoscaler/pkg/updater/logic/updater.go
```diff
@@ -85,6 +105,19 @@ func (u *updater) RunOnce(ctx context.Context) {
...
 	defer timer.ObserveTotal()
 
+	if u.useAdmissionControllerStatus {
+		isValid, err := u.statusValidator.IsStatusValid(status.AdmissionControllerStatusTimeout)
+		if err != nil {
+			klog.Errorf("Error getting Admission Controller status: %v. Skipping eviction loop", err)
+			return
+		}
+		if !isValid {
+			klog.Warningf("Admission Controller status has not been updated for %v. Skipping eviction loop",
```
bskiba: nit: Admission Controller status has been refreshed more than %v ago.
Also, how are Duration object printed? Will the log be readable?

vertical-pod-autoscaler/pkg/updater/logic/updater_test.go
```diff
@@ -136,12 +140,76 @@ func TestVPAOff(t *testing.T) {
...
+		selectorFetcher:              target_mock.NewMockVpaTargetSelectorFetcher(ctrl),
+		useAdmissionControllerStatus: true,
+		statusValidator:              newFakeValidator(true),
+	}
+
+	updater.RunOnce(context.Background())
+	eviction.AssertNumberOfCalls(t, "Evict", 0)
+}
+
+// TODO(krzysied): Merge tests so there is less copy-pasted code.
```
bskiba: How about extracting test setup to a separate function for now?

vertical-pod-autoscaler/pkg/updater/logic/updater.go
krzysied: Done.
Duration has String() method implemented. When %v is called it will print result of String() method if available. So it will be readable.

vertical-pod-autoscaler/pkg/updater/logic/updater_test.go
krzysied: Currently `TestRunOnce_InvalidStatus` is copy-paste of `TestRunOnce` with expected eviction count equaling 0.
The only reason I want to keep it this ways is the fact that I want to change it in separate PR. I assume that I'll be changing no only this test, but also all other tests in this file.

PR #2773, opened 2020-01-27
Do not compare StartTime for unschedulable pods

losipiuk:
Do not compare StartTime for unschedulable pods

Using standard MostImportantPod function for pods which do not have
Status.StartTime set result in Error log message which kills
scalability.


MaciekPytel:
/lgtm
/approve
Thanks, nice catch!

PR #2774, opened 2020-01-27
VPA - Optional RenewTime fix

krzysied:
RenewTime is optional Lease field. Requires not-nil check.

### DIFF
--- a/vertical-pod-autoscaler/pkg/utils/status/status_object.go
+++ b/vertical-pod-autoscaler/pkg/utils/status/status_object.go
@@ -124,7 +124,8 @@ func IsStatusValid(status *apicoordinationv1.Lease, leaseTimeout time.Duration) 
 
 func isStatusValid(status *apicoordinationv1.Lease, leaseTimeout time.Duration, now time.Time) bool {
 	return status.CreationTimestamp.Add(leaseTimeout).After(now) ||
-		status.Spec.RenewTime.Add(leaseTimeout).After(now)
+		(status.Spec.RenewTime != nil &&
+			status.Spec.RenewTime.Add(leaseTimeout).After(now))
 }
 
 func isRetryableAPIError(err error) bool {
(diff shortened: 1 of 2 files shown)


vertical-pod-autoscaler/pkg/utils/status/status_object.go
```diff
@@ -124,7 +124,8 @@ func IsStatusValid(status *apicoordinationv1.Lease, leaseTimeout time.Duration)
 
 func isStatusValid(status *apicoordinationv1.Lease, leaseTimeout time.Duration, now time.Time) bool {
 	return status.CreationTimestamp.Add(leaseTimeout).After(now) ||
```
bskiba: CreationTimestamp is not optional I presume?

vertical-pod-autoscaler/pkg/utils/status/status_object_test.go
```diff
@@ -230,6 +230,16 @@ func TestIsStatusValid(t *testing.T) {
 			leaseTimeout:  10 * time.Second,
 			expectedValid: true,
 		},
+		{
+			name: "Outdated CreationTimestamp",
```
bskiba: Useful to point out this actually tests nil RenewTime

bskiba: (Could be just by making RenewTime = nil explicit in the lease object or changing test case name to Outdated CreationTimestamp with no/nil RenewTime)

vertical-pod-autoscaler/pkg/utils/status/status_object.go
krzysied: AFIAK it is not.

vertical-pod-autoscaler/pkg/utils/status/status_object_test.go
krzysied: Test case name updated.

PR #2785, opened 2020-01-30
VPA - Exporting admission controller status consts to common

krzysied:


vertical-pod-autoscaler/pkg/common/admission_controller_status.go
```diff
@@ -0,0 +1,34 @@
+/*
```
bskiba: I'd rather see this in pkg/utils/status

krzysied: Done. Consts in pgk/utils/status/status_object.go.

vertical-pod-autoscaler/pkg/utils/status/status_object.go
```diff
@@ -31,6 +31,19 @@ import (
...
 const (
+	// AdmissionControllerStatusName is the name of
+	// the Admission Controller status object.
+	AdmissionControllerStatusName = "vpa-admission-controller"
+	// AdmissionControllerStatusNamespace is the namespace of
+	// the Admission Controller status object.
+	AdmissionControllerStatusNamespace = "kube-system"
+	// AdmissionControllerStatusUpdateInterval is a time after which
+	// the Admission Controller status object should be updated.
+	AdmissionControllerStatusUpdateInterval = 10 * time.Second
```
bskiba: Will timeout and interval be used by anyone else except for admission-controller? If not, they should be privete in the admission controller

krzysied: Timeout will be mostly used by Updater. Interval is Admission Controller only.
However, what I want to have here is a description of Admission Controller status properties. Instead of "Admission Controller updates status" it is rather "Admission Controller has to update status". That's why I promoted it to global constants.

bskiba: I don't see why it should be a requirement at what interval the status object is updated. If it's enough to update once 10 of 50 seconds, that's up to admission controller to decide, the only requirement is that after 1 minute without update the admission controller status is invalid.

krzysied: Fair point.
I kept `updateInterval` in Admission Controller unchanged.

PR #2788, opened 2020-01-31
VPA - Adding status object validator

krzysied:
Adding Validator interface that will be used by status readers.

### DIFF
--- a/vertical-pod-autoscaler/pkg/utils/status/status_object.go
+++ b/vertical-pod-autoscaler/pkg/utils/status/status_object.go
@@ -57,6 +57,11 @@ type Client struct {
 	holderIdentity       string
 }
 
+// Validator for the status object.
+type Validator interface {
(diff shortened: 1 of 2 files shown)


vertical-pod-autoscaler/pkg/utils/status/status_object.go
```diff
@@ -57,6 +57,11 @@ type Client struct {
 	holderIdentity       string
 }
 
+// Validator for the status object.
+type Validator interface {
+	IsStatusValid(leaseTimeout time.Duration) (bool, error)
```
bskiba: can you change leaseTimeout to something more implementation neutral? 
It doesn't have to be a lease backing the status.

krzysied: Done. `statusTimeout` it is.

bskiba:
:+1: 
/lgtm
/approve

PR #2790, opened 2020-01-31
Compute expansion options for additional created groups

aleksandra-malinowska:
Compute expansion options for additional created groups.

(I'm working on a test that would have caught the bug, TBA in another commit to this PR.)

### DIFF
--- a/cluster-autoscaler/core/scale_up.go
+++ b/cluster-autoscaler/core/scale_up.go
@@ -250,6 +250,36 @@ func maxResourceLimitReached(resources []string) *skippedReasons {
 	return &skippedReasons{[]string{fmt.Sprintf("max cluster %s limit reached", strings.Join(resources, ", "))}}
 }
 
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/core/scale_up.go
```diff
@@ -467,6 +475,12 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 					continue
 				}
 				nodeInfos[nodeGroup.Id()] = nodeInfo
+
+				option := computeExpansionOption(context, podEquivalenceGroups, nodeGroup, nodeInfo, upcomingNodes)
+
+				if len(option.Pods) > 0 && option.NodeCount > 0 {
```
losipiuk: Do we need this if? Why not put it always in the map. Even if pods are empty. No harm, right?

losipiuk: Edit - maybe this is safer. We are checking for size of expansion options later. Keep it.

aleksandra-malinowska:
Added commit with test - fails before the fix, passes with the fix.

PR #2792, opened 2020-02-04
VPA - Removing updater e2e test from scripts

krzysied:
Updater e2e tests are part of actuation tests now.
Updater tests should be removed from run-e2e scripts.

### DIFF
--- a/vertical-pod-autoscaler/hack/run-e2e-tests.sh
+++ b/vertical-pod-autoscaler/hack/run-e2e-tests.sh
@@ -23,7 +23,6 @@ function print_help {
   echo "ERROR! Usage: run-e2e-tests.sh <suite>"
   echo "<suite> should be one of:"
   echo " - recommender"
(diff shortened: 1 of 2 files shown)


vertical-pod-autoscaler/hack/deploy-for-e2e.sh
```diff
@@ -44,7 +43,7 @@ fi
 SUITE=$1
 
 case ${SUITE} in
-  recommender|updater|admission-controller)
+  recommender|admission-controller)
```
bskiba: While I totally agree with removing the e2e test suite, here it might be useful to have a script that just lets you deploy updater. I know it's called deploy-for-e2e, but it is useful just for ad-hock testing of components while changing them,

krzysied: Good idea.
Changes reverted.

PR #2800, opened 2020-02-06
VPA - Moving status updater to utils

krzysied:
Moving status_updater to utils/.
This is a setup for e2e tests.

ref #2738

### DIFF
--- a/vertical-pod-autoscaler/pkg/utils/status/status_updater.go
+++ b/vertical-pod-autoscaler/pkg/utils/status/status_updater.go
@@ -14,48 +14,46 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package logic
+package status
 
 import (
 	"time"
 
-	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/status"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
 )
 
-const (
-	updateInterval = 10 * time.Second
-)
-
-// StatusUpdater periodically updates Admission Controller status.
-type StatusUpdater struct {
-	client *status.Client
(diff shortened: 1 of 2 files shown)


vertical-pod-autoscaler/pkg/utils/status/status_updater.go
```diff
@@ -14,48 +14,46 @@ See the License for the specific language governing permissions and
...
 )
 
-const (
-	updateInterval = 10 * time.Second
-)
-
-// StatusUpdater periodically updates Admission Controller status.
-type StatusUpdater struct {
-	client *status.Client
+// Updater periodically updates Admission Controller status.
```
bskiba: Remove Admission Controller from comment

vertical-pod-autoscaler/pkg/utils/status/status_updater.go
```diff
@@ -14,48 +14,46 @@ See the License for the specific language governing permissions and
...
+type Updater struct {
```
bskiba: Soooo we already have an updater in VPA :)
Can you think of a similar name that will not cause clashes?

vertical-pod-autoscaler/pkg/utils/status/status_updater.go
krzysied: Done.

vertical-pod-autoscaler/pkg/utils/status/status_updater.go
krzysied: In normal cases the Updater will be referenced by the package name. Creating it looks like this:
```
status.NewUpdater(...)
```
So there should be no confusion there.

krzysied:
@bskiba PTAL

PR #2812, opened 2020-02-10
VPA - Removing e2e common tests

krzysied:
Removing unused k8s e2e common tests.

There should be only vpa e2e tests left.


bskiba:
/lgtm
a/pprove

PR #2813, opened 2020-02-10
VPA - Reducing copy-pasted code in updater tests

### LINKED ISSUE #2793: VPA - Updater logic tests clean-up
[Updater logic test](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/updater/logic/updater_test.go) should be simplified. Copy-pasted code should be reduced.

krzysied:
/fixes #2793

### DIFF
--- a/vertical-pod-autoscaler/pkg/updater/logic/updater_test.go
+++ b/vertical-pod-autoscaler/pkg/updater/logic/updater_test.go
@@ -42,7 +42,85 @@ func parseLabelSelector(selector string) labels.Selector {
 	return parsedSelector
 }
 
-func TestRunOnce(t *testing.T) {
+func TestRunOnce_Mode(t *testing.T) {
+	tests := []struct {
+		name                  string
+		updateMode            vpa_types.UpdateMode
+		expectFetchCalls      bool
+		expectedEvictionCount int
+	}{
+		{
+			name:                  "with Auto mode",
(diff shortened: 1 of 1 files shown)


vertical-pod-autoscaler/pkg/updater/logic/updater_test.go
```diff
@@ -209,11 +177,14 @@ func TestRunOnce_InvalidStatus(t *testing.T) {
 		recommendationProcessor:      &test.FakeRecommendationProcessor{},
 		selectorFetcher:              mockSelectorFetcher,
 		useAdmissionControllerStatus: true,
-		statusValidator:              newFakeValidator(false),
+		statusValidator:              statusValidator,
 	}
 
+	if expectFetchCalls {
```
bskiba: Is there value from verifying those calls are not made if not expected?

krzysied: 1. This verification was in original tests, so I want to preserve the same logic.
2. IIUC using gomock controller defines a list of calls that we expect to happen. If the call is registered and not used it throws exception at the end of the test.
This is the first time I'm seeing this mock controller so I can be wrong here...

bskiba: I see, thanks

PR #2815, opened 2020-02-10
Clarify Nodes() expectation

enxebre:
`Nodes()` is ambiguous since might cause the impression that is expected to return only the list of instances for the node group which also have a kubernetes node associated already. This tries to clarify that expectation.


MaciekPytel:
/lgtm
/approve
Thanks!

PR #2840, opened 2020-02-18
cluster-autoscaler/aws: batch launch config query and ttl cache

JulienBalestra:
This PR improve our caching strategy of the launch configuration name to instance type.
It adds a TTL and batch the query to populate this cache.

It happens that users change the instance type of an already running ASG and that become an issue forcing to restart the cluster-autoscaler.
Note that it's also an issue to have nodes compared to a template of nodes but that's an another topic I'd like to open in a dedicated thread.

In a cluster with more than 1300 ASGs, the query to the launch configuration API done one by one become a throttling situation.

Doing that by batching these call by page of 50 elements solved our throttling issue.

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/auto_scaling.go
+++ b/cluster-autoscaler/cloudprovider/aws/auto_scaling.go
@@ -18,10 +18,21 @@ package aws
 
 import (
 	"fmt"
+	"sync"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
+	"k8s.io/apimachinery/pkg/util/clock"
+	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
+)
+
+const (
+	launchConfigurationCachedTTL = time.Minute * 20
+	cacheMinTTL                  = 120
+	cacheMaxTTL                  = 600
 )
 
 // autoScaling is the interface represents a specific aspect of the auto-scaling service provided by AWS SDK for use in CA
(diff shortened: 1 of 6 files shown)


cluster-autoscaler/cloudprovider/aws/auto_scaling.go
```diff
@@ -36,12 +41,55 @@ type autoScaling interface {
...
+	sync.RWMutex
+}
+
+func newLaunchConfigurationInstanceTypeCache() *expirationStore {
+	jc := &jitterClock{}
+	return &expirationStore{
+		cache.NewExpirationStore(func(obj interface{}) (s string, e error) {
+			return obj.(instanceTypeCachedObject).name, nil
+		}, &cache.TTLPolicy{
+			TTL:   time.Minute * 20,
```
Jeffwan: I am thinking if 20 minutes work for most of the users. Not sure if user will change it if we add an command option for it.  Let's at least have const.

cluster-autoscaler/cloudprovider/aws/auto_scaling.go
```diff
@@ -36,12 +41,55 @@ type autoScaling interface {
...
+		jc,
+	}
+}
+
+func (c *jitterClock) Since(ts time.Time) time.Duration {
+	since := time.Since(ts)
+	c.RLock()
+	defer c.RUnlock()
+	if c.jitter {
+		return since + (time.Second * time.Duration(rand.IntnRange(120, 600)))
```
Jeffwan: Can you define some const for these magic numbers?

cluster-autoscaler/cloudprovider/aws/auto_scaling.go
```diff
@@ -94,6 +145,63 @@ func (m *autoScalingWrapper) getAutoscalingGroupsByNames(names []string) ([]*aut
...
+
+	// List expire old entries
+	_ = m.launchConfigurationInstanceTypeCache.List()
+	if len(launchConfigToQuery) == 0 {
+		klog.V(4).Infof("%d launch configurations already in cache", len(autoscalingGroups))
+		return nil
+	}
+	klog.V(4).Infof("%d launch configurations to query", len(launchConfigToQuery))
+
+	for i := 0; i < len(launchConfigToQuery); i += 50 {
```
Jeffwan: This looks duplicate as `getInstanceTypeByLCName`. Is it possible to refactor the `getInstanceTypeByLCName ` and make it accept a slice of LC names? Then we can reuse it.

cluster-autoscaler/cloudprovider/aws/auto_scaling.go
JulienBalestra: As you want, I've done the needed changes 👍

Jeffwan:
Changes look good to me

/lgtm
/approve

PR #2848, opened 2020-02-19
Add topology.kubernetes.io labels to be ignored when comparing similar node groups.

enxebre:
Without this, the autoscaler were using the labels in `compareLabels` and failing to match similar node groups in different zones. Starting in kube 1.17 `failure-domain.beta.kubernetes.io/*` are deprecated in favour of `topology.kubernetes.io/* ` https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#failure-domainbetakubernetesiozone


MaciekPytel:
/lgtm
/approve 
Thanks!

enxebre:
cc @hardikdr for awarness

hardikdr:
Yes, I think this will impact us, thanks for pointing out.

PR #2850, opened 2020-02-20
VPA - Fixing vpaObservedContainers annotation

krzysied:
Adding overriding  VpaObservedContainers annotation instead of all annotations.

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
@@ -35,6 +35,10 @@ import (
 	"k8s.io/klog"
 )
 
+const (
+	vpaAnnotationLabel = "vpaUpdates"
+)
+
 // AdmissionServer is an admission webhook server that modifies pod resources request based on VPA recommendation
 type AdmissionServer struct {
 	recommendationProvider RecommendationProvider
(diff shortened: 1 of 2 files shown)


vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
```diff
@@ -100,8 +100,8 @@ func (s *AdmissionServer) getPatchesForPodResourceRequest(raw []byte, namespace
 	vpaObservedContainersValue := annotations.GetVpaObservedContainersValue(&pod)
 	patches = append(patches, patchRecord{
 		Op:    "add",
-		Path:  "/metadata/annotations",
-		Value: map[string]string{annotations.VpaObservedContainersLabel: vpaObservedContainersValue},
+		Path:  fmt.Sprintf("/metadata/annotations/%s", annotations.VpaObservedContainersLabel),
```
bskiba: AFAIU this will fail if there are no annotations on the pod. Please either leave a comment on why we know annotations are not empty here or add an if for empty pod.Annotations as above with vpaUpdates annotation.

bskiba: (bonus point for extracting to a helper method :) )

krzysied: Ok. I changed the logic there:
- new we create patch to add empty annotations (if needed)
- we add resource annotation patch (if needed)
- we add observed containers annotation patch 

It creates additional patch compared to previous version, but IMHO the logic seems easier to understand now.

bskiba: Agreed, also harder to add another annotation in a wrong way.

PR #2462, opened 2019-10-16
Add AWS ignore labels and increase max memory difference.

cdmurph32:
AWS M5 instance types may differ in memory capacity by more than 128MB.

Whitelist additional node labels for AWS CNI custom networking and
EC2 lifecycle.

As I documented in https://github.com/kubernetes/autoscaler/issues/2020, the m5 instance type memory varies by more that 128MB .  I'm not sure what the basis of the max memory capacity difference is.  There are quite a few options if there are good reasons not to touch it.  Perhaps making it modifiable through a command argument.

PR #2851, opened 2020-02-20
Backporting AWS ignored labels and capacity memory changes to 1.16

maximerenou50:
This PR is about backporting changes from master to 1.16. This is required when using EKS 1.14 with node groups because:
- EKS 1.14 (which is currently the latest version) does not work with CA 1.17
- Sometimes the memory capacity between instances within worker groups can differ quite a lot
- EKS add a label that confused CA 

I was able to test this against a 3 AZs group set which now spread the instances across all AZs.

Finally, I would like to get a new release of 1.16, can you guide me through the process of doing that? (or is it just something that happens by itself periodically?)

### DIFF
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -293,6 +293,9 @@ func buildAutoscaler() (core.Autoscaler, error) {
 	if autoscalingOptions.CloudProviderName == cloudprovider.AzureProviderName {
 		processors.NodeGroupSetProcessor = &nodegroupset.BalancingNodeGroupSetProcessor{
 			Comparator: nodegroupset.IsAzureNodeInfoSimilar}
+	} else if autoscalingOptions.CloudProviderName == cloudprovider.AwsProviderName {
+		processors.NodeGroupSetProcessor = &nodegroupset.BalancingNodeGroupSetProcessor{
+			Comparator: nodegroupset.IsAwsNodeInfoSimilar}
 	}
 
 	opts := core.AutoscalerOptions{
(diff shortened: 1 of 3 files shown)


maximerenou50:
Hello,
Any update on this? I'm only backporting changes, not introducing anything new, so hopefully, it should be fairly straightforward. Please let me know if you need more information.
Thanks in advance!

piteur:
Heads up on that MR, I will benefit from it as we are facing the same issue!

Please consider :)

MaciekPytel:
/approve
cc: @Jeffwan - this is AWS specific, so I don't want to lgtm without pinging you first. If you're ok with this let's merge it.

Jeffwan:
@maximerenou50 Can you give original PR info? 

> Sometimes the memory capacity between instances within worker groups can differ quite a lot

Could you give nodeInfo, I'm fine to increase limit but we really want to resolve it internally. I do see some instance has memory difference. need to get more data points.

maximerenou50:
Hi @Jeffwan ,

This is the original PR: https://github.com/kubernetes/autoscaler/pull/2462/commits/7f0a42b023b6a5bb279161cee082664ffbc093c8
For my scenario, I noticed on AWS **m5.xlarge** instances a difference of 176MB. I do not know the reason for that, at first, I thought it was between AZ but actually it also happened within the same AZ.

Please let me know if you need more information.
Thanks

Jeffwan:
/lgtm

Yeah. I would suggest you to include original PR info in the cherry-pick. Like `cherry-pick #no. - PR title`. It would be helpful for us to manage back port changes for every release.

Jeffwan:
@MaciekPytel Changes looks good to me, could you help approve this PR?

Jeffwan:
original PR: https://github.com/kubernetes/autoscaler/pull/2462/files

PR #2854, opened 2020-02-21
Filter out expendable pods before initializing snapshot

aleksandra-malinowska:
Simpler than filtering them out at every step.

### DIFF
--- a/cluster-autoscaler/core/filter_out_schedulable.go
+++ b/cluster-autoscaler/core/filter_out_schedulable.go
@@ -20,9 +20,7 @@ import (
 	"sort"
 	"time"
 
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/autoscaler/cluster-autoscaler/context"
-	"k8s.io/autoscaler/cluster-autoscaler/core/utils"
 	"k8s.io/autoscaler/cluster-autoscaler/metrics"
 	"k8s.io/autoscaler/cluster-autoscaler/processors/pods"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator"
@@ -64,7 +62,7 @@ func (filterOutSchedulablePodListProcessor) Process(
 	var unschedulablePodsToHelp []*apiv1.Pod
 
 	unschedulablePodsToHelp, err := filterOutSchedulableByPacking(unschedulablePods, context.ClusterSnapshot,
-		context.PredicateChecker, context.ExpendablePodsPriorityCutoff)
+		context.PredicateChecker)
 
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


MaciekPytel:
A few general points (not applicable to any particular line in code):
* Removing filtering in scale_down.go is obviously the right thing to do.
  * Seems like it was dead code anyway - we already removed all the expendable pods in filterOutSchedulable PLP. I don't think we revert this change to snapshot anywhere in the code? Or is there some other way those pods get re-introduced that I don't see?
* I think filtering out in PLP was actually somewhat more elegant than doing that explicitly in static_autoscaler.go. Previously we tried to move all filtering logic out of static_autoscaler and into PLPs (ex. migrating filter_out_schedulable to PLP), specifically to clean up static_autoscaler and hide all sorts of filtering and scheduling heuristics behind a single call.
  * [Not directly related to this PR] Should all the heuristics in PLPs should apply to scale-down? With current implementation we implicitly apply them to scale-down via snapshot. It seems that for filtering out expendable and schedulable pods that's a good thing, but maybe there are other use-cases, where we'd want to revert? cc: @losipiuk

aleksandra-malinowska:
> * Removing filtering in scale_down.go is obviously the right thing to do.
>   
>   * Seems like it was dead code anyway - we already removed all the expendable pods in filterOutSchedulable PLP. I don't think we revert this change to snapshot anywhere in the code? Or is there some other way those pods get re-introduced that I don't see?
> * I think filtering out in PLP was actually somewhat more elegant than doing that explicitly in static_autoscaler.go. Previously we tried to move all filtering logic out of static_autoscaler and into PLPs (ex. migrating filter_out_schedulable to PLP), specifically to clean up static_autoscaler and hide all sorts of filtering and scheduling heuristics behind a single call.

One problem is that processors are called *after* initializing snapshot. As a result, pods are added and removed unnecessarily (and NodeInfo.AddPod() has non-zero cost, see benchmarks).

Elegance is a matter of taste, I suppose ;). For me, "filter out schedulable" silently discarding expendable pods combines two logically separate steps.

>   * [Not directly related to this PR] Should all the heuristics in PLPs should apply to scale-down? With current implementation we implicitly apply them to scale-down via snapshot. It seems that for filtering out expendable and schedulable pods that's a good thing, but maybe there are other use-cases, where we'd want to revert? cc: @losipiuk

Removing expendable pods is necessary. As for bin-packing and preemptions, I think right now we still disable scale-down if there are any schedulable pods, but if we want to ever make it possible (which may make sense in large clusters with constant churn), we will need those in snapshot.

MaciekPytel:
> One problem is that processors are called after initializing snapshot. As a result, pods are added and removed unnecessarily (and NodeInfo.AddPod() has non-zero cost, see snapshot benchmarks).

For filter out schedulable it's necessary - it needs working PredicateChecker, so it needs an initialized snapshot. I assume cost of removing pods is similar to adding, so technically the argument holds for filtering expendables, but that feels very much like a premature optimization.

> For me, "filter out schedulable" silently discarding expendable pods combines two logically separate steps.

I agree. My preferred state would be filtering out expendable in a separate PLP I think. Obviously that kind of cleanup never happens.
I don't think it breaks anything and I don't feel super strongly about this, but moving it to static_autoscaler is trading one solution I don't particularly like for another. I just don't think it's worth the delta it involves. But if you think it's much better I'm ok lgtm-ing it.

aleksandra-malinowska:
> > For me, "filter out schedulable" silently discarding expendable pods combines two logically separate steps.
> 
> I agree. My preferred state would be filtering out expendable in a separate PLP I think. Obviously that kind of cleanup never happens.
> I don't think it breaks anything and I don't feel super strongly about this, but moving it to static_autoscaler is trading one solution I don't particularly like for another. I just don't think it's worth the delta it involves. But if you think it's much better I'm ok lgtm-ing it.

I think it's much better for reasoning about autoscaler logic - e.g. it wasn't immediately obvious that the pods will be removed in the processor, resulting in TODO in scale-down. Different parts of the code had to know about expendable pods only to ignore them. If I were to implement support for evicting expendable pods, I'd probably first modify drain logic and happily write a passing unit test for it before realizing the pods wouldn't be in the snapshot in a real-world scenario. 

(BTW, adding support for non-expendable pods wouldn't be easier with the old version - we'd either have to store them in some random field in context, or change the pod list processor's interface to return them, or create a new snapshot (which is just as easy with this change)).

PR #2861, opened 2020-02-27
Cleanup todo

aleksandra-malinowska:
We've always used the same state for scale down, changing it doesn't seem related to scheduler migration.


MaciekPytel:
Seems like this todo is ~the same as my comment in https://github.com/kubernetes/autoscaler/pull/2854#issuecomment-593408738? Anyway - I think this is a useful thing to follow-up on, but the TODO itself doesn't seem too useful (and agreed - it's not really related to SF migration).
/lgtm
/approve

PR #2864, opened 2020-02-27
Add ability to add handlers for new resources to AdmissionController via RegisterHandler()

tweks:
This PR refactors AdmissionController and decouples its core logic from logic responsible for handling resources. Handlers for already supported resources are registered via the new mechanism.

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/logic/pod_resource_handler_test.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/logic/pod_resource_handler_test.go
@@ -23,11 +23,13 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"k8s.io/api/admission/v1beta1"
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1"
 	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/annotations"
-	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/limitrange"
 	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/test"
 	vpa_api_util "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/vpa"
 )
@@ -45,14 +47,6 @@ type fakePodPreProcessor struct {
 
 func (fpp *fakePodPreProcessor) Process(pod apiv1.Pod) (apiv1.Pod, error) {
 	return pod, fpp.e
-}
-
-type fakeVpaPreProcessor struct {
-	e error
-}
-
-func (fvp *fakeVpaPreProcessor) Process(vpa *vpa_types.VerticalPodAutoscaler, isCreate bool) (*vpa_types.VerticalPodAutoscaler, error) {
-	return vpa, fvp.e
 }
 
 type fakeRecommendationProvider struct {
@@ -152,7 +146,7 @@ func assertPatchOneOf(t *testing.T, got patchRecord, want []patchRecord) {
 	assert.Fail(t, msg)
 }
 
-func TestGetPatchesForResourceRequest(t *testing.T) {
+func TestGetPatches(t *testing.T) {
 	tests := []struct {
 		name                 string
 		podJson              []byte
@@ -344,12 +338,18 @@ func TestGetPatchesForResourceRequest(t *testing.T) {
 	for _, tc := range tests {
 		t.Run(fmt.Sprintf("test case: %s", tc.name), func(t *testing.T) {
 			fppp := fakePodPreProcessor{e: tc.podPreProcessorError}
-			fvpp := fakeVpaPreProcessor{}
 			fvm := fakeVpaMatcher{}
 			frp := fakeRecommendationProvider{tc.recommendResources, tc.recommendAnnotations, tc.recommendError}
-			lc := limitrange.NewNoopLimitsCalculator()
-			s := NewAdmissionServer(&frp, &fppp, &fvpp, lc, &fvm)
-			patches, err := s.getPatchesForPodResourceRequest(tc.podJson, tc.namespace)
+			h := newPodResourceHandler(&fppp, &frp, &fvm)
+			patches, err := h.GetPatches(&v1beta1.AdmissionRequest{
+				Resource: v1.GroupVersionResource{
+					Version: "v1",
+				},
+				Namespace: tc.namespace,
+				Object: runtime.RawExtension{
+					Raw: tc.podJson,
+				},
+			})
 			if tc.expectError == nil {
 				assert.NoError(t, err)
 			} else {
@@ -368,9 +368,8 @@ func TestGetPatchesForResourceRequest(t *testing.T) {
 	}
 }
 
-func TestGetPatchesForResourceRequest_TwoReplacementResources(t *testing.T) {
+func TestGetPatches_TwoReplacementResources(t *testing.T) {
 	fppp := fakePodPreProcessor{}
-	fvpp := fakeVpaPreProcessor{}
 	fvm := fakeVpaMatcher{}
 	recommendResources := []vpa_api_util.ContainerResources{
 		{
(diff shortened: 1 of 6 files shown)


vertical-pod-autoscaler/pkg/admission-controller/logic/resource_handler.go
```diff
@@ -0,0 +1,35 @@
...
+
+package logic
+
+import (
+	"k8s.io/api/admission/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/metrics/admission"
+)
+
+// ResourceHandler represents a handler for a resource in Admission Server
```
jbartosik: Please describe this interface a bit more. After reading the interface I don't know how I would use it. `GroupResourceType` and `AdmissionResource` (why one is `ResourceType` but the other is just `Resource`?)  tell me what resource types it accepts? `DismissIncorrectObjects` tells me whether it will discard invalid objects or not? `GetPatches` actually does the work described by other methods?

vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
```diff
@@ -285,22 +70,21 @@ func (s *AdmissionServer) admit(data []byte) (*v1beta1.AdmissionResponse, metric
...
+	handler, ok := s.resourceHandlers[admittedGroupResource]
+	if ok {
+		patches, err = handler.GetPatches(ar.Request)
+		resource = handler.AdmissionResource()
 	} else {
-		patches, err = nil, fmt.Errorf("expected the resource to be one of: %v, %v", podResource, vpaGroupResource)
+		patches, err = nil, fmt.Errorf("not supported resource type: %v", admittedGroupResource)
+	}
+
+	if handler.DismissIncorrectObjects() && err != nil {
```
jbartosik: 1. I think this doesn't work when ok is false
2. When we have the handler I think it would be easier to understand if `GetPatches` returned three values - `patches`, `allow` (`bool`), and `err` then here we would have

```
if !allow {
```

vertical-pod-autoscaler/pkg/admission-controller/logic/resource_handler.go
```diff
@@ -0,0 +1,35 @@
...
+type ResourceHandler interface {
```
jbartosik: I think we don't need to make this interface visible outside vertical-pod-autoscaler/pkg/admission-controller/logic/server.go

tweks: We need. The goal of this change is to be able to inject custom controllers into this Admission Server, which I will do in a next step.

vertical-pod-autoscaler/pkg/admission-controller/logic/pod_pre_processor.go
```diff
@@ -1,39 +0,0 @@
-/*
```
krzysied: What is the justification for moving `PodPreProcessor` from pod_pre_processor.go to pod_resource_handler.go?
Same question for `VpaPreProcessor`?

vertical-pod-autoscaler/pkg/admission-controller/logic/resource_handler.go
tweks: You got everything right :). Clarified in the comments. Does it look better now?

vertical-pod-autoscaler/pkg/admission-controller/logic/pod_pre_processor.go
krzysied: I don't find it cleaner. We move a struct/interface from a file dedicated purely for it to other unrelated file while this struct/interface still remains public.
If it was private structure then I would consider it justified.

For now, please move back `PodPreProcessor` and `VpaPreProcessor` to their files. This PR is still huge without these additional changes.

vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
tweks: 1. Good catch! Done.
2. Whether to dismiss incorrect objects is a global setting of a handler, not depending on an actual admission request. Eg. Pods should never be dismissed, as we do not own this resource type. So I find it mach cleaner to have it separately. WDYT?

vertical-pod-autoscaler/pkg/admission-controller/logic/vpa_resource_handler.go
```diff
@@ -0,0 +1,157 @@
...
+	return vpa, nil
+}
+
+// NewDefaultVpaPreProcessor creates a VpaPreProcessor that leaves VPAs unchanged and returns no error
+func NewDefaultVpaPreProcessor() VpaPreProcessor {
+	return &noopVpaPreProcessor{}
+}
+
+// VpaResourceHandler builds patches for VPAs
+type VpaResourceHandler struct {
```
krzysied: You can make `VpaResourceHandler` private and add
```
func NewVpaResourceHandler(vpaPreProcessor VpaPreProcessor) ResourceHandler {
    return &VpaResourceHandler{
        vpaPreProcessor: vpaPreProcessor,
    }
}
```
Now we know that this implements  ResourceHandler interface.
Same change can be done for PodResourceHandler

vertical-pod-autoscaler/pkg/admission-controller/logic/vpa_resource_handler.go
```diff
@@ -0,0 +1,157 @@
...
+
+// noopVpaPreProcessor leaves pods unchanged when processing
+type noopVpaPreProcessor struct{}
+
+// Process leaves the pod unchanged
+func (p *noopVpaPreProcessor) Process(vpa *vpa_types.VerticalPodAutoscaler, isCreate bool) (*vpa_types.VerticalPodAutoscaler, error) {
+	return vpa, nil
+}
+
+// NewDefaultVpaPreProcessor creates a VpaPreProcessor that leaves VPAs unchanged and returns no error
```
krzysied: nit: `.` here and in other comments.

vertical-pod-autoscaler/pkg/admission-controller/logic/pod_pre_processor.go
tweks: Done.

vertical-pod-autoscaler/pkg/admission-controller/logic/vpa_resource_handler.go
tweks: Added for all newly introduced methods.

vertical-pod-autoscaler/pkg/admission-controller/logic/vpa_resource_handler.go
tweks: Done.

tweks:
> I took a look at the code. I'm not sure which code changed and which just moved. Please:
> 
> * Address golint complaints

Done.

> * Break this into two changes (one moving code around, other making changes to the code) for easier review

Is it now clear enough, after putting preprocessors back into separate files?

krzysied:
/lgtm
Let's wait for @jbartosik  for the final approve.

Btw, I really like the fact that vertical-pod-autoscaler/pkg/admission-controller/logic/server.go logic will be split into smaller parts :+1:

vertical-pod-autoscaler/pkg/admission-controller/logic/pod_resource_handler_test.go
```diff
@@ -395,9 +396,19 @@ func TestGetPatchesForResourceRequest_TwoReplacementResources(t *testing.T) {
 				}`)
 	recommendAnnotations := vpa_api_util.ContainerToAnnotationsMap{}
 	frp := fakeRecommendationProvider{recommendResources, recommendAnnotations, "name", nil}
-	lc := limitrange.NewNoopLimitsCalculator()
-	s := NewAdmissionServer(&frp, &fppp, &fvpp, lc)
-	patches, err := s.getPatchesForPodResourceRequest(podJson, "default")
+	h := podResourceHandler{
```
krzysied: One more nit:  We can use `NewPodResourceHandler` to initialize structure. Here and everywhere else.

tweks: Done.

tweks:
Rebased to unfortunate concurrent refactor of Admission Controller.

bskiba:
I'd like to have a look as well
/hold

vertical-pod-autoscaler/pkg/admission-controller/logic/server.go
```diff
@@ -295,22 +72,21 @@ func (s *AdmissionServer) admit(data []byte) (*v1beta1.AdmissionResponse, metric
...
-		patches, err = s.getPatchesForVPADefaults(ar.Request.Object.Raw, ar.Request.Operation == v1beta1.Create)
-		resource = metrics_admission.Vpa
-		// we don't let in problematic VPA objects - late validation
-		if err != nil {
+	handler, ok := s.resourceHandlers[admittedGroupResource]
+	if ok {
+		patches, err = handler.GetPatches(ar.Request)
+		resource = handler.AdmissionResource()
+
+		if handler.DismissIncorrectObjects() && err != nil {
```
bskiba: nit: I think DisallowIncorrectObjects or BlockIncorrectObjects is more precise.

tweks: Done.

bskiba:
/lgtm
/approve

@jbartosik if you have concerns left, LMK

vertical-pod-autoscaler/pkg/admission-controller/logic/resource_handler.go
jbartosik: I think we don't need it. Even if an interface is private part of a module you can still implement the niterface outside the module

PR #2871, opened 2020-02-27
[Azure] Add fallback check for 'Promo' instance types

fooka03:
Currently, if a VMSS is created with a Promo instance type that's not contained in the static type list the autoscaler fails to manage it with the following errors:

```
Failed to get node infos for groups: instance type "<redacted>_Promo" not supported 
Unable to build proper template node for <redacted>: instance type "<redacted>_Promo" not supported
```

This error is invalid as there are no differences between standard and _Promo instance types outside of pricing.  In order to allow users to take advantage of Promo instances as they are made available without relying on new autoscaler images, this change will attempt to match a `_Promo` instance type against the non-promo types contained in the static Azure type list if an exact match is not found.

The change here is already realized in AKS - https://github.com/Azure/aks-engine/pull/1216

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
@@ -517,6 +517,22 @@ func (scaleSet *ScaleSet) buildNodeFromTemplate(template compute.VirtualMachineS
 			break
 		}
 	}
+
+	promoRe := regexp.MustCompile(`(?i)_promo`)
+	if promoRe.MatchString(*template.Sku.Name) {
+		if vmssType == nil {
+			// We didn't find an exact match but this is a promo type, check for matching standard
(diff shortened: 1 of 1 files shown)


fooka03:
I agree, dynamically keeping the list up to date (or not needing the list at all) would be ideal.  However we've found that Azure is quite sensitive to request counts (https://github.com/kubernetes/autoscaler/pull/2527) and we need to avoid DDoS'ing ourselves.

Perhaps a reasonable compromise here would be to keep the `Promo` instances in the list, leave the generator alone, but have an else if for the case where `Promo` isn't in the list but a corresponding standard type is found.  That would make a reasonable attempt to allow a cluster to continue running in between autoscaler releases when managing 'unlisted' promo scale sets.  Thoughts @marwanad ?

marwanad:
@fooka03 Fair enough. We can do dynamic with a long cache TTL but it might be overkill.

I like your proposition and I think it's a reasonable way to go.

PR #2121, opened 2019-06-12
feat: enable min/max on Azure VMSS autodiscovery

alexeldeib:
ref: https://github.com/kubernetes/autoscaler/issues/2051

Issue #2628, opened 2019-12-08
Failed to list *v1.CSINode


masterkain:
I deploy on AWS EKS, kubernetes 1.14.

I upgraded to image `k8s.gcr.io/cluster-autoscaler:v1.17.0` and updated my RBAC to avoid permission errors:

```
  - apiGroups: ["storage.k8s.io"]
    resources: ["storageclasses", "csinodes"]
```

```
  - apiGroups: ["coordination.k8s.io"]
    resources: ["leases"]
    verbs: ["create"]
  - apiGroups: ["coordination.k8s.io"]
    resourceNames: ["cluster-autoscaler"]
    resources: ["leases"]
    verbs: ["get", "update"]
```

However I'm getting spammed by this message:

```
│ I1208 09:27:14.811469       1 reflector.go:188] Listing and watching *v1.CSINode from k8s.io/client-go/informers/factory.go:135
│ E1208 09:27:14.812998       1 reflector.go:156] k8s.io/client-go/informers/factory.go:135: Failed to list *v1.CSINode: the server could not find the requested resource
```

left4taco:
Same error here.

zhuhesheng:
Same error here.

MaciekPytel:
I've seen the same on GCE.

@losipiuk I think you recently looked into CSINodes? Can you take a look?

losipiuk:
I am pretty sure that the reason you are seeing those is fact that k8s 1.14 has no `v1.storage.k8s.io.csinode` objects. AFAIK the API was promoted to `v1` from `v1beta1` in k8s 1.17

losipiuk:
CA 1.17 is not supposed to be used on k8s 1.14

michael-burt:
@losipiuk do you know what the latest CAS supported on EKS is? The latest EKS version is 1.14.

hectorj2f:
@michael-burt I'd say https://github.com/kubernetes/autoscaler/tree/cluster-autoscaler-1.14.7/cluster-autoscaler

michael-burt:
Thanks @hectorj2f 

I see on the CAS documentation for 1.17 that they list a bunch of notes related to AWS but I guess this is not for EKS.

PR #2877, opened 2020-03-02
Cherry pick label vmss

hichemaichour:
Cherry picking:
1. [enable min/max on Azure VMSS autodiscovery](https://github.com/kubernetes/autoscaler/pull/2121)
2. [Fix azure autodiscovery when max tags aren't specified](https://github.com/kubernetes/autoscaler/pull/2714)

/area provider/azure


marwanad:
hey @hichemaichour, thanks for the contribution.

Given the size of the changes and that it touches multiple cloud providers and also the nature of it being a feature), I don't think we will be cherry-picking this to earlier releases.

hichemaichour:
> It may take a couple minutes for the CLA signature to be fully registered; after that, please reply here with a new comment and we'll verify. Thanks.

Signed

hichemaichour:
> It may take a couple minutes for the CLA signature to be fully registered; after that, please reply here with a new comment and we'll verify. Thanks.

Signed

hichemaichour:
@marwanad , I understand

I was just looking forward to this feature, and with [this issue](https://github.com/kubernetes/autoscaler/issues/2628) (#2628) present for 1.17, I was hoping that the feature is introduced as well in 1.16

marwanad:
@hichemaichour yup, sorry about that. 

I think you can ignore the CSI node log errors, but regardless 1.17 is meant to be run on K8s 1.17 only.

/close

PR #2866, opened 2020-02-27
refactor updater for a better structured update priority processing

bskiba:
I believe this makes the code both more extensible and infinitely more testable.

- I moved the getPodPriority function to the priorityProcessor.GetPodPriority 1:1 
- I moved the getPodPrioriry tests to priorityProcessor's tests 1:1
- I converted some of the AddPod tests to priorityProcessor's (the ones that were only testing that we are calculating priority correctly and the sorting was actually secondary)

There's still room for improvement in the AddPod tests (hence the TODO to refactor them), but I didn't want to make this review gigantic

PR #2889, opened 2020-03-04
Make PriorityProcessor useful by changing return value to public

bskiba:
followup to #2866


krzysied:
Please address golint issue. Apart from that lgtm.

bskiba:
Done. I've also changed fields to public

Issue #2797, opened 2020-02-04
Scheduler Framework based PredicateChecker - followups


losipiuk:
This is an umbrella issue to list followup items to be addressed after merging in https://github.com/kubernetes/autoscaler/pull/2796

 - [x] Do not return errors from ClusterSnapshot methods for which it does not make sense (Cleare, Revert?, Get*, RemovePod?, RemoveNode?)
 - [x] Pass nodeName to RemovePod to avoid costly iteration over all nodes
 - [x] Refactor errors in basic_cluster_snapshot.go.
 - [x] Remove deprecated_scheduler_snapshot
 - [x] N/A - already resolved ~~Add list of nodes (node names?) to FitsAny. Use that to optimize number of PreFilter calls in binpacking.~~
 - [x] Remove SnapshotClusterState() and all related code (ex. keeping listers in PredicateChecker).
 - [x] Add unit tests for BasicClusterSnaphsot (share with DeltaClusterSnapshot)
   - [x] Assert cached lists are invalidated on Commit()/modify (including the scenario described in https://github.com/kubernetes/autoscaler/pull/2799#discussion_r378348879)
 - [x] Address TODOs in predicates_checker_interface.go
   - [x] Forget FakeNodeInfoForNodeName ever existed
 - [x] Address every TODO(scheduler_framework_migration).
 - [x] Cleanup mess in scale-down so we always use ClusterSnapshot as source of truth (see comments below)
 - [x] Clean up how we calculate podDestinations list so it works with snapshot.
 - [ ] Consider if we could do a single FitsAny call in findPlaceFor (instead of tryNodeForPod in a loop) .
 - [x] Investigate what happens if there is "pending" pod in cluster with nodeName pointing to node which does not exist in cluster; fix if needed
 - [x] Check if GPU hack still works after removing IsNodeReadyAndSchedulablePredicate
 - [ ] Pass context Estimate(), rather that PredicateChecker and ClusterSnapshot to EstimateBuilder
 - [x] Create issue for AWS about failing `TestGetRegion`
 - [ ] kubernetes/kubernetes#87822

gjtempleton:
I've already got a fix for the failing `TestGetRegion` in the closed PR #2750 - will open a separate PR for that now.

PR #2890, opened 2020-03-04
Use nodes and pods from snapshot in scale-down

aleksandra-malinowska:
Follow-up from #2797. Small refactors are only in the code already affected by this change.

### DIFF
--- a/cluster-autoscaler/core/scale_down.go
+++ b/cluster-autoscaler/core/scale_down.go
@@ -510,10 +510,10 @@ func (sd *ScaleDown) UpdateUnneededNodes(
 		emptyNodes[node.Name] = true
 	}
 
-	currentlyUnneededNonEmptyNodes := make([]*apiv1.Node, 0, len(currentlyUnneededNodeInfos))
+	currentlyUnneededNonEmptyNodes := make([]*schedulernodeinfo.NodeInfo, 0, len(currentlyUnneededNodeInfos))
 	for _, node := range currentlyUnneededNodeInfos {
 		if !emptyNodes[node.Node().Name] {
-			currentlyUnneededNonEmptyNodes = append(currentlyUnneededNonEmptyNodes, node.Node())
+			currentlyUnneededNonEmptyNodes = append(currentlyUnneededNonEmptyNodes, node)
 		}
 	}
 
@@ -687,15 +687,20 @@ func (sd *ScaleDown) markSimulationError(simulatorErr errors.AutoscalerError,
 // chooseCandidates splits nodes into current candidates for scale-down and the
 // rest. Current candidates are unneeded nodes from the previous run that are
 // still in the nodes list.
-func (sd *ScaleDown) chooseCandidates(nodes []*apiv1.Node) ([]*apiv1.Node, []*apiv1.Node) {
+func (sd *ScaleDown) chooseCandidates(nodes []*schedulernodeinfo.NodeInfo) ([]*apiv1.Node, []*apiv1.Node) {
 	// Number of candidates should not be capped. We will look for nodes to remove
 	// from the whole set of nodes.
 	if sd.context.ScaleDownNonEmptyCandidatesCount <= 0 {
-		return nodes, []*apiv1.Node{}
+		candidates := make([]*apiv1.Node, len(nodes))
+		for i, node := range nodes {
+			candidates[i] = node.Node()
+		}
+		return candidates, []*apiv1.Node{}
 	}
 	currentCandidates := make([]*apiv1.Node, 0, len(sd.unneededNodesList))
 	currentNonCandidates := make([]*apiv1.Node, 0, len(nodes))
-	for _, node := range nodes {
+	for _, nodeInfo := range nodes {
+		node := nodeInfo.Node()
 		if _, found := sd.unneededNodes[node.Name]; found {
 			currentCandidates = append(currentCandidates, node)
 		} else {
--- a/cluster-autoscaler/core/scale_down.go
+++ b/cluster-autoscaler/core/scale_down.go
@@ -752,7 +752,6 @@ func (sd *ScaleDown) TryToScaleDown(pdbs []*policyv1.PodDisruptionBudget, curren
 	nodesWithoutMaster := filterOutMasters(allNodeInfos)
 	nodesWithoutMasterNames := make(map[string]bool, len(nodesWithoutMaster))
 
-	candidates := make([]*schedulernodeinfo.NodeInfo, 0)
 	candidateNames := make([]string, 0)
 	readinessMap := make(map[string]bool)
 	candidateNodeGroups := make(map[string]cloudprovider.NodeGroup)
@@ -843,12 +842,11 @@ func (sd *ScaleDown) TryToScaleDown(pdbs []*policyv1.PodDisruptionBudget, curren
 			continue
 		}
 
-		candidates = append(candidates, nodeInfo)
 		candidateNames = append(candidateNames, node.Name)
 		candidateNodeGroups[node.Name] = nodeGroup
 	}
 
-	if len(candidates) == 0 {
+	if len(candidateNames) == 0 {
 		klog.V(1).Infof("No candidates for scale down")
 		scaleDownStatus.Result = status.ScaleDownNoUnneeded
 		return scaleDownStatus, nil
--- a/cluster-autoscaler/core/scale_down.go
+++ b/cluster-autoscaler/core/scale_down.go
@@ -659,22 +659,20 @@ func (sd *ScaleDown) markSimulationError(simulatorErr errors.AutoscalerError,
 // chooseCandidates splits nodes into current candidates for scale-down and the
 // rest. Current candidates are unneeded nodes from the previous run that are
 // still in the nodes list.
-func (sd *ScaleDown) chooseCandidates(nodes []string) ([]string, []string) {
+func (sd *ScaleDown) chooseCandidates(nodes []string) (candidates []string, nonCandidates []string) {
 	// Number of candidates should not be capped. We will look for nodes to remove
 	// from the whole set of nodes.
 	if sd.context.ScaleDownNonEmptyCandidatesCount <= 0 {
 		return nodes, nil
 	}
-	currentCandidates := make([]string, 0, len(sd.unneededNodesList))
-	currentNonCandidates := make([]string, 0, len(nodes))
 	for _, node := range nodes {
 		if _, found := sd.unneededNodes[node]; found {
-			currentCandidates = append(currentCandidates, node)
+			candidates = append(candidates, node)
 		} else {
-			currentNonCandidates = append(currentNonCandidates, node)
+			nonCandidates = append(nonCandidates, node)
 		}
 	}
-	return currentCandidates, currentNonCandidates
+	return candidates, nonCandidates
 }
 
 func (sd *ScaleDown) mapNodesToStatusScaleDownNodes(nodes []*apiv1.Node, nodeGroups map[string]cloudprovider.NodeGroup, evictedPodLists map[string][]*apiv1.Pod) []*status.ScaleDownNode {
--- a/cluster-autoscaler/core/scale_down.go
+++ b/cluster-autoscaler/core/scale_down.go
@@ -763,12 +763,17 @@ func (sd *ScaleDown) SoftTaintUnneededNodes(allNodes []*apiv1.Node) (errors []er
 
 // TryToScaleDown tries to scale down the cluster. It returns a result inside a ScaleDownStatus indicating if any node was
 // removed and error if such occurred.
-func (sd *ScaleDown) TryToScaleDown(allNodes []*apiv1.Node, pods []*apiv1.Pod, pdbs []*policyv1.PodDisruptionBudget,
+func (sd *ScaleDown) TryToScaleDown(
+	allNodes []*apiv1.Node,
+	pods []*apiv1.Pod,
+	pdbs []*policyv1.PodDisruptionBudget,
 	currentTime time.Time) (*status.ScaleDownStatus, errors.AutoscalerError) {
+
 	scaleDownStatus := &status.ScaleDownStatus{NodeDeleteResults: sd.nodeDeletionTracker.GetAndClearNodeDeleteResults()}
 	nodeDeletionDuration := time.Duration(0)
 	findNodesToRemoveDuration := time.Duration(0)
 	defer updateScaleDownMetrics(time.Now(), &findNodesToRemoveDuration, &nodeDeletionDuration)
+
 	nodesWithoutMaster := filterOutMasters(allNodes, pods)
 	candidates := make([]*apiv1.Node, 0)
 	readinessMap := make(map[string]bool)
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/core/scale_down.go
```diff
@@ -398,64 +399,79 @@ func (sd *ScaleDown) CleanUpUnneededNodes() {
...
-		node := nodeInfo.Node()
-		if node != nil {
-			nodeNameToNodeInfo[node.Name] = nodeInfo
+		if nodeInfo.Node() == nil {
+			// Do we need to check this?
+			klog.Errorf("Unexpected nil node in node info")
+			continue
+		}
+		if candidateNames[nodeInfo.Node().Name] {
+			candidateNodeInfos = append(candidateNodeInfos, nodeInfo)
```
MaciekPytel: Probably not a big deal, but previously we had a check for missing NodeInfo. Now if there is a node in candidates list that has no corresponding NI we will just ignore it silently.

cluster-autoscaler/simulator/cluster.go
```diff
@@ -113,14 +112,25 @@ type UtilizationInfo struct {
 
 // FindNodesToRemove finds nodes that can be removed. Returns also an information about good
 // rescheduling location for each of the pods.
-func FindNodesToRemove(candidates []*apiv1.Node, destinationNodes []*apiv1.Node, pods []*apiv1.Pod,
-	listers kube_util.ListerRegistry, clusterSnapshot ClusterSnapshot, predicateChecker PredicateChecker, maxCount int,
-	fastCheck bool, oldHints map[string]string, usageTracker *UsageTracker,
+func FindNodesToRemove(
+	candidates []*schedulernodeinfo.NodeInfo,
+	destinationNodes []*schedulernodeinfo.NodeInfo,
```
MaciekPytel: I think we already had this discussion, but I think passing along list of NodeInfos together with clusterSnapshot everywhere is really ugly. 
Snapshot should be the source of truth for NI. But we end up using both the snapshot and the list passed as parameter. I'm not saying building NI map at every step (ie. current state) is any better, but I think it would be more consistent to add getters based on node name to snapshot (GetNode(nodeName), GetPodsForNode(nodeName)) and just pass list of node names and snapshot. WDYT?
cc: @losipiuk for another opinion on this

cluster-autoscaler/core/scale_down.go
```diff
@@ -678,19 +690,20 @@ func (sd *ScaleDown) markSimulationError(simulatorErr errors.AutoscalerError,
 // chooseCandidates splits nodes into current candidates for scale-down and the
 // rest. Current candidates are unneeded nodes from the previous run that are
 // still in the nodes list.
-func (sd *ScaleDown) chooseCandidates(nodes []*apiv1.Node) ([]*apiv1.Node, []*apiv1.Node) {
+func (sd *ScaleDown) chooseCandidates(nodes []*schedulernodeinfo.NodeInfo) ([]*schedulernodeinfo.NodeInfo, []*schedulernodeinfo.NodeInfo) {
```
MaciekPytel: See my comment in simulator/cluster.go: there is really no need for this to be NIs. All we care about are node names.

cluster-autoscaler/simulator/cluster.go
aleksandra-malinowska: We already have 'Get(nodeName)', which returns nodeInfo, which includes pods ;) 

I didn't want to bundle too many changes, but if you think it's worth refactoring the node list processors to return names, I'm happy to do this.

cluster-autoscaler/core/scale_down.go
aleksandra-malinowska: Good point, will add logging error.

cluster-autoscaler/core/scale_down.go
```diff
@@ -398,64 +399,79 @@ func (sd *ScaleDown) CleanUpUnneededNodes() {
...
 		// This should never happen, List() returns err only because scheduler interface requires it.
 		return errors.ToAutoscalerError(errors.InternalError, err)
 	}
 
-	nodeNameToNodeInfo := make(map[string]*schedulernodeinfo.NodeInfo)
+	candidateNames := make(map[string]bool, len(scaleDownCandidates))
+	for _, candidate := range scaleDownCandidates {
+		if candidate == nil {
+			// Do we need to check this?
+			klog.Errorf("Unexpected nil node in node info")
```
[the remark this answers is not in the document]
aleksandra-malinowska: This is basically the check from https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/scale_down.go#L431

I strongly suspect if we ever get in that situation, we'd panic somewhere else in the code anyway. But I can remove the comment and keep the checks.

cluster-autoscaler/core/scale_down.go
```diff
@@ -398,64 +399,79 @@ func (sd *ScaleDown) CleanUpUnneededNodes() {
...
+			continue
+		}
+		candidateNames[candidate.Name] = true
+	}
+	destinationNames := make(map[string]bool, len(destinationNodes))
+	for _, destination := range destinationNodes {
+		if destination == nil {
+			// Do we need to check this?
+			klog.Errorf("Unexpected nil node in node info")
```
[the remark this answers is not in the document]
aleksandra-malinowska: Answered above

cluster-autoscaler/simulator/cluster.go
aleksandra-malinowska: I added the changes in new commits. The biggest drawback is IMO redundant error-checking, but it's not easy to get rid of since Get() can fail if a node is not in the snapshot.

losipiuk: I generally agree with Maciek's comment here. I would prefer to limit use of `NodeInfo` in the codebese. Preferably to 0. As this is internal scheduler structure.

cluster-autoscaler/core/scale_down.go
```diff
@@ -631,24 +618,18 @@ func (sd *ScaleDown) isNodeBelowUtilzationThreshold(node *apiv1.Node, utilInfo s
...
-	for name := range sd.unremovableNodes {
-		nodesToDelete[name] = struct{}{}
-	}
-	// Nodes that are in the cluster should not be deleted.
-	for _, node := range nodes {
-		if _, ok := nodesToDelete[node.Node().Name]; ok {
-			delete(nodesToDelete, node.Node().Name)
+	newUnremovableNodes := make(map[string]time.Time, len(sd.unremovableNodes))
+	for oldUnremovable, since := range sd.unremovableNodes {
+		if _, err := sd.context.ClusterSnapshot.NodeInfos().Get(oldUnremovable); err == nil {
```
losipiuk: Log on error here

cluster-autoscaler/core/scale_down.go
```diff
@@ -678,16 +659,16 @@ func (sd *ScaleDown) markSimulationError(simulatorErr errors.AutoscalerError,
 // chooseCandidates splits nodes into current candidates for scale-down and the
 // rest. Current candidates are unneeded nodes from the previous run that are
 // still in the nodes list.
-func (sd *ScaleDown) chooseCandidates(nodes []*apiv1.Node) ([]*apiv1.Node, []*apiv1.Node) {
+func (sd *ScaleDown) chooseCandidates(nodes []string) ([]string, []string) {
```
losipiuk: Nit: Could you please assign names for returned lists to increase readability?

cluster-autoscaler/core/scale_down.go
```diff
@@ -755,14 +736,24 @@ func (sd *ScaleDown) SoftTaintUnneededNodes(allNodes []*apiv1.Node) (errors []er
...
 	scaleDownStatus := &status.ScaleDownStatus{NodeDeleteResults: sd.nodeDeletionTracker.GetAndClearNodeDeleteResults()}
 	nodeDeletionDuration := time.Duration(0)
 	findNodesToRemoveDuration := time.Duration(0)
 	defer updateScaleDownMetrics(time.Now(), &findNodesToRemoveDuration, &nodeDeletionDuration)
-	nodesWithoutMaster := filterOutMasters(allNodes, pods)
-	candidates := make([]*apiv1.Node, 0)
+
+	allNodeInfos, errSnapshot := sd.context.ClusterSnapshot.NodeInfos().List()
+	if errSnapshot != nil {
+		// This should never happen, List() returns err only because scheduler interface requires it.
```
losipiuk: I do not believe this comment is helpful.

cluster-autoscaler/simulator/cluster.go
```diff
@@ -113,14 +111,20 @@ type UtilizationInfo struct {
 
 // FindNodesToRemove finds nodes that can be removed. Returns also an information about good
 // rescheduling location for each of the pods.
-func FindNodesToRemove(candidates []*apiv1.Node, destinationNodes []*apiv1.Node, pods []*apiv1.Pod,
-	listers kube_util.ListerRegistry, clusterSnapshot ClusterSnapshot, predicateChecker PredicateChecker, maxCount int,
-	fastCheck bool, oldHints map[string]string, usageTracker *UsageTracker,
+func FindNodesToRemove(
+	candidates []string,
+	destinations map[string]bool,
```
losipiuk: Using `map[string]bool` here while `[]string` for `candiates` is awkward. Please use `[]string` and compute map internally if needed.

cluster-autoscaler/core/scale_down.go
```diff
@@ -965,24 +963,38 @@ func updateScaleDownMetrics(scaleDownStart time.Time, findNodesToRemoveDuration
 	metrics.UpdateDuration(metrics.ScaleDownMiscOperations, miscDuration)
 }
 
-func (sd *ScaleDown) getEmptyNodesNoResourceLimits(candidates []*apiv1.Node, pods []*apiv1.Pod, maxEmptyBulkDelete int) []*apiv1.Node {
-	return sd.getEmptyNodes(candidates, pods, maxEmptyBulkDelete, noScaleDownLimitsOnResources())
+func (sd *ScaleDown) getEmptyNodesNoResourceLimits(candidates []string, maxEmptyBulkDelete int) []*apiv1.Node {
+	/*	candidateNodeInfos := make([]*schedulernodeinfo.NodeInfo, len(candidates))
```
losipiuk: remove comment

cluster-autoscaler/core/scale_down.go
```diff
@@ -755,14 +736,24 @@ func (sd *ScaleDown) SoftTaintUnneededNodes(allNodes []*apiv1.Node) (errors []er
...
+	allNodeInfos, errSnapshot := sd.context.ClusterSnapshot.NodeInfos().List()
+	if errSnapshot != nil {
+		// This should never happen, List() returns err only because scheduler interface requires it.
+		return scaleDownStatus, errors.ToAutoscalerError(errors.InternalError, errSnapshot)
+	}
+
+	nodesWithoutMaster := filterOutMasters(allNodeInfos)
+	nodesWithoutMasterNames := make(map[string]bool, len(nodesWithoutMaster))
+
+	candidates := make([]*schedulernodeinfo.NodeInfo, 0)
```
losipiuk: You do not use this one. You can just leave `candidateNames`

cluster-autoscaler/core/scale_down.go
```diff
@@ -755,14 +736,24 @@ func (sd *ScaleDown) SoftTaintUnneededNodes(allNodes []*apiv1.Node) (errors []er
...
-	nodesWithoutMaster := filterOutMasters(allNodes, pods)
-	candidates := make([]*apiv1.Node, 0)
+
+	allNodeInfos, errSnapshot := sd.context.ClusterSnapshot.NodeInfos().List()
+	if errSnapshot != nil {
+		// This should never happen, List() returns err only because scheduler interface requires it.
+		return scaleDownStatus, errors.ToAutoscalerError(errors.InternalError, errSnapshot)
+	}
+
+	nodesWithoutMaster := filterOutMasters(allNodeInfos)
```
losipiuk: Do we need `NodeInfo`s here or can we migrate to just `Node`?

cluster-autoscaler/core/scale_down.go
```diff
@@ -778,77 +769,85 @@ func (sd *ScaleDown) TryToScaleDown(allNodes []*apiv1.Node, pods []*apiv1.Pod, p
...
-			candidates = append(candidates, node)
-			candidateNodeGroups[node.Name] = nodeGroup
+		checkResult := scaleDownResourcesLeft.checkScaleDownDeltaWithinLimits(scaleDownResourcesDelta)
+		if checkResult.exceeded {
+			klog.V(4).Infof("Skipping %s - minimal limit exceeded for %v", node.Name, checkResult.exceededResources)
+			sd.addUnremovableNodeReason(node, simulator.MinimalResourceLimitExceeded)
+			continue
 		}
+
+		candidates = append(candidates, nodeInfo)
```
losipiuk: You can drop this one and just leave `candidateNames`

cluster-autoscaler/simulator/cluster.go
```diff
@@ -131,68 +135,71 @@ func FindNodesToRemove(candidates []*apiv1.Node, destinationNodes []*apiv1.Node,
 	newHints := make(map[string]string, len(oldHints))
 
 candidateloop:
-	for _, node := range candidates {
-		klog.V(2).Infof("%s: %s for removal", evaluationType, node.Name)
+	for _, nodeName := range candidates {
+		nodeInfo, err := clusterSnapshot.NodeInfos().Get(nodeName)
+		if err != nil {
+			klog.Errorf("Can't retrieve node %s from snapshot, err: %v", nodeName, err)
```
losipiuk: continue?

aleksandra-malinowska: Done

cluster-autoscaler/core/scale_down.go
aleksandra-malinowska: Done

cluster-autoscaler/core/scale_down.go
aleksandra-malinowska: Added logging on info level - this is supposed to happen once a node is removed.

cluster-autoscaler/core/scale_down.go
aleksandra-malinowska: Is this better?

cluster-autoscaler/core/scale_down.go
aleksandra-malinowska: Done

cluster-autoscaler/core/scale_down.go
aleksandra-malinowska: We've discussed it with @MaciekPytel before and it seems to be the least evil version for now. Ideally I'd prefer to have an error-less method to use internally, since none of the snapshot implementations can return error here:s

cluster-autoscaler/simulator/cluster.go
aleksandra-malinowska: Done

cluster-autoscaler/core/scale_down.go
```diff
@@ -167,7 +166,7 @@ type scaleDownResourcesDelta map[string]int64
 // used as a value in scaleDownResourcesLimits if actual limit could not be obtained due to errors talking to cloud provider
 const scaleDownLimitUnknown = math.MinInt64
 
-func computeScaleDownResourcesLeftLimits(nodes []*apiv1.Node, resourceLimiter *cloudprovider.ResourceLimiter, cp cloudprovider.CloudProvider, timestamp time.Time) scaleDownResourcesLimits {
+func computeScaleDownResourcesLeftLimits(nodes []*schedulernodeinfo.NodeInfo, resourceLimiter *cloudprovider.ResourceLimiter, cp cloudprovider.CloudProvider, timestamp time.Time) scaleDownResourcesLimits {
```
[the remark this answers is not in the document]
aleksandra-malinowska: Done

cluster-autoscaler/core/scale_down.go
```diff
@@ -225,15 +225,16 @@ func calculateScaleDownCoresMemoryTotal(nodes []*apiv1.Node, timestamp time.Time
 	return coresTotal, memoryTotal
 }
 
-func calculateScaleDownGpusTotal(nodes []*apiv1.Node, cp cloudprovider.CloudProvider, timestamp time.Time) (map[string]int64, error) {
+func calculateScaleDownGpusTotal(nodes []*schedulernodeinfo.NodeInfo, cp cloudprovider.CloudProvider, timestamp time.Time) (map[string]int64, error) {
```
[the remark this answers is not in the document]
aleksandra-malinowska: Done

cluster-autoscaler/core/scale_down.go
```diff
@@ -209,9 +208,10 @@ func computeAboveMin(total int64, min int64) int64 {
 
 }
 
-func calculateScaleDownCoresMemoryTotal(nodes []*apiv1.Node, timestamp time.Time) (int64, int64) {
+func calculateScaleDownCoresMemoryTotal(nodes []*schedulernodeinfo.NodeInfo, timestamp time.Time) (int64, int64) {
```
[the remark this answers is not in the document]
aleksandra-malinowska: Done

cluster-autoscaler/core/scale_down.go
aleksandra-malinowska: Passing around node names means checking errors every time we retrieve the node from the snapshot.

I'm leaning towards using node infos retrieved once from snapshot, but for now implemented it as requested.

cluster-autoscaler/core/scale_down.go
losipiuk: I like it much more. Thanks.

cluster-autoscaler/core/scale_down.go
losipiuk: Fair. I was just suggesting to remove the comment. But this is nit.

PR #2899, opened 2020-03-06
Fix `test-unit` make targets for VPA

jbartosik:
`admission-controller` and `updater` were still using `godep` and they
didn't work. `recommender` tried to run tests with `-race` flag, wich
failed with

`go test: -race requires cgo; enable cgo by setting CGO_ENABLED=1`

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/Makefile
+++ b/vertical-pod-autoscaler/pkg/admission-controller/Makefile
@@ -3,7 +3,8 @@ all: build
 TAG?=dev
 REGISTRY?=staging-k8s.gcr.io
 FLAGS=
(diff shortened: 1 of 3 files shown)


bskiba:
Thanks! Actually you can fix it by setting CGO_ENABLED=1 for tests only and keeping the -race

jbartosik:
> Thanks! Actually you can fix it by setting CGO_ENABLED=1 for tests only and keeping the -race

Done

bskiba:
This will break binaries for distroless.
We need this flag set for tests _only_

jbartosik:
> This will break binaries for distroless.
> We need this flag set for tests _only_

Done

bskiba:
:+1:  
/lgtm
/approve

PR #2900, opened 2020-03-06
Remove unused pod preprocessor

bskiba:
/assign @krzysied


krzysied:
I'm kinda lost here...
Why are we removing PodPreProcessor while keeping VpaPreProcessor?

Issue #2886, opened 2020-03-03
Fix flaky admission controller unit test


bskiba:
I've seen multiple instances of this on unrelated PRs: 

?   	k8s.io/autoscaler/vertical-pod-autoscaler/common	[no test files]
?   	k8s.io/autoscaler/vertical-pod-autoscaler/pkg/admission-controller	[no test files]
--- FAIL: TestGetPatchesForResourceRequest_TwoReplacementResources (0.00s)
    server_test.go:396: 
        	Error Trace:	server_test.go:396
        	Error:      	Should be true
        	Test:       	TestGetPatchesForResourceRequest_TwoReplacementResources
FAIL
FAIL	k8s.io/autoscaler/vertical-pod-autoscaler/pkg/admission-controller/logic	0.097s

bskiba:
Seen this again on #2895

jbartosik:
This flake happens very rarely when I try to reproduce it locally. I'm runing

```bash
$ for i in $(seq 1 1000); do echo $i; make test-unit; ret=$?; if [ $ret -ne 0 ]; then echo fail: $ret; break; fi; done
```

but saw the flake only once so far.

jbartosik:
In #2905 I just saw a failure with new [logs](https://travis-ci.org/kubernetes/autoscaler/jobs/660079355?utm_medium=notification&utm_source=github_status):

```
--- FAIL: TestGetPatchesForResourceRequest_TwoReplacementResources (0.00s)
    server_test.go:136: 
        	Error Trace:	server_test.go:136
        	            				server_test.go:410
        	Error:      	Should be true
        	Test:       	TestGetPatchesForResourceRequest_TwoReplacementResources
        	Messages:   	got {Op:add Path:/metadata/annotations/vpaUpdates Value:Pod resources updated by name: container 0: unobtanium request, cpu request}, want: {Op:add Path:/metadata/annotations/vpaUpdates Value:Pod resources updated by name: container 0: cpu request, unobtanium request}

```

krzysied:
This bug is fixed, but I would like to add additional context for why it happened:
This problem is caused by https://github.com/kubernetes/autoscaler/blob/4317838e1971d7c7cda6649f96c68c5a1720f160/vertical-pod-autoscaler/pkg/admission-controller/logic/server.go#L168
Golang assumes that map can be iterated in any order. This makes our test result non-deterministic.
This was a problem regarding our test assumptions, not a vpa admission controller issue.

PR #2901, opened 2020-03-06
Make flaky assertion more verbose

jbartosik:
This assertion is flaky:

https://github.com/kubernetes/autoscaler/issues/2886

Make it log more useful message when it fails. Also clean it up (an
asserion above makes sure the second half of the or is never true).


jbartosik:
@krzysied I'm changing assertion you recently added, please check if I missed something

/assign krzysied

jbartosik:
@krzysied like we talked offline I added helper(s) so other places that compare patches can have nice failure logs too

Issue #2859, opened 2020-02-26
verticalpodautoscaler CRD not compliant with openAPIV3


sebastien-prudhomme:
verticalpodautoscaler v0.6.3 with the patched CRD version from master branch (with some openAPIV3 fixes)

Just notice that there is still that error in the status of the verticalpodautoscaler CRD:

```
  - lastTransitionTime: "2020-02-26T10:44:51Z"
    message: 'spec.validation.openAPIV3Schema.type: Required value: must not be empty
      at the root'
    reason: Violations
    status: "True"
    type: NonStructuralSchema
```
With this patch, all is ok :

```diff
@ -24,6 +24,7 @@ spec:
       storage: true
   validation:
     openAPIV3Schema:
+      type: object
       properties:
         spec:
           type: object

```

bskiba:
Thanks for reporting! Which Kubernetes version are you observing the error on?

sebastien-prudhomme:
@bskiba Kubernetes v1.15.3

onesolpark:
fixed by #2904
/close

sebastien-prudhomme:
Thanks for the commit.

PR #2904, opened 2020-03-09
Make VPA Crds compliant with OpenAPIV3

onesolpark:
Resolves issue #2859 

Make CRDS compliant with OpenAPIV3 so it won't get NonStructuralSchema Validation Error


krzysied:
/lgtm
/approve

Thanks!

PR #2905, opened 2020-03-09
CLI flags for VPA recommender aggregations config.

amshuman-kr:
This PR exposes some of the aggregations configuration of VPA recommender as command-line flags in the VPA recommender. This can help change the default behaviour of the VPA recommender for different workload use-cases. The pre-existing constant values have been retained as the default values.

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/main.go
+++ b/vertical-pod-autoscaler/pkg/recommender/main.go
@@ -43,18 +43,22 @@ var (
 
 	storage = flag.String("storage", "", `Specifies storage mode. Supported values: prometheus, checkpoint (default)`)
 	// prometheus history provider configs
-	historyLength                        = flag.String("history-length", "8d", `How much time back prometheus have to be queried to get historical metrics`)
-	podLabelPrefix                       = flag.String("pod-label-prefix", "pod_label_", `Which prefix to look for pod labels in metrics`)
-	podLabelsMetricName                  = flag.String("metric-for-pod-labels", "up{job=\"kubernetes-pods\"}", `Which metric to look for pod labels in metrics`)
-	podNamespaceLabel                    = flag.String("pod-namespace-label", "kubernetes_namespace", `Label name to look for container names`)
-	podNameLabel                         = flag.String("pod-name-label", "kubernetes_pod_name", `Label name to look for container names`)
-	ctrNamespaceLabel                    = flag.String("container-namespace-label", "namespace", `Label name to look for container names`)
-	ctrPodNameLabel                      = flag.String("container-pod-name-label", "pod_name", `Label name to look for container names`)
-	ctrNameLabel                         = flag.String("container-name-label", "name", `Label name to look for container names`)
-	memoryAggregationInterval            = flag.Duration("memory-aggregation-interval", model.DefaultMemoryAggregationInterval, `The length of a single interval, for which the peak memory usage is computed. Memory usage peaks are aggregated in multiples of this interval. In other words there is one memory usage sample per interval (the maximum usage over that interval)`)
-	memoryAggregationWindowIntervalCount = flag.Int64("memory-aggregation-window-interval-count", model.DefaultMemoryAggregationWindowIntervalCount, `The number of consecutive memory-aggregation-intervals which make up the MemoryAggregationWindowLength which in turn is the period for memory usage aggregation by VPA. In other words, MemoryAggregationWindowLength = memory-aggregation-interval * memory-aggregation-window-interval-count.`)
-	memoryHistogramDecayHalfLife         = flag.Duration("memory-histogram-decay-half-life", model.DefaultMemoryHistogramDecayHalfLife, `The amount of time it takes a historical memory usage sample to lose half of its weight. In other words, a fresh usage sample is twice as 'important' as one with age equal to the half life period.`)
-	cpuHistogramDecayHalfLife            = flag.Duration("cpu-histogram-decay-half-life", model.DefaultCPUHistogramDecayHalfLife, `The amount of time it takes a historical CPU usage sample to lose half of its weight.`)
+	historyLength       = flag.String("history-length", "8d", `How much time back prometheus have to be queried to get historical metrics`)
+	podLabelPrefix      = flag.String("pod-label-prefix", "pod_label_", `Which prefix to look for pod labels in metrics`)
+	podLabelsMetricName = flag.String("metric-for-pod-labels", "up{job=\"kubernetes-pods\"}", `Which metric to look for pod labels in metrics`)
+	podNamespaceLabel   = flag.String("pod-namespace-label", "kubernetes_namespace", `Label name to look for container names`)
+	podNameLabel        = flag.String("pod-name-label", "kubernetes_pod_name", `Label name to look for container names`)
+	ctrNamespaceLabel   = flag.String("container-namespace-label", "namespace", `Label name to look for container names`)
+	ctrPodNameLabel     = flag.String("container-pod-name-label", "pod_name", `Label name to look for container names`)
+	ctrNameLabel        = flag.String("container-name-label", "name", `Label name to look for container names`)
+)
+
+// Aggregation configuration flags
+var (
+	memoryAggregationInterval      = flag.Duration("memory-aggregation-interval", model.DefaultMemoryAggregationInterval, `The length of a single interval, for which the peak memory usage is computed. Memory usage peaks are aggregated in multiples of this interval. In other words there is one memory usage sample per interval (the maximum usage over that interval)`)
+	memoryAggregationIntervalCount = flag.Int64("memory-aggregation-interval-count", model.DefaultMemoryAggregationIntervalCount, `The number of consecutive memory-aggregation-intervals which make up the MemoryAggregationWindowLength which in turn is the period for memory usage aggregation by VPA. In other words, MemoryAggregationWindowLength = memory-aggregation-interval * memory-aggregation-interval-count.`)
+	memoryHistogramDecayHalfLife   = flag.Duration("memory-histogram-decay-half-life", model.DefaultMemoryHistogramDecayHalfLife, `The amount of time it takes a historical memory usage sample to lose half of its weight. In other words, a fresh usage sample is twice as 'important' as one with age equal to the half life period.`)
+	cpuHistogramDecayHalfLife      = flag.Duration("cpu-histogram-decay-half-life", model.DefaultCPUHistogramDecayHalfLife, `The amount of time it takes a historical CPU usage sample to lose half of its weight.`)
 )
 
 func main() {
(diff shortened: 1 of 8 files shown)


amshuman-kr:
/assign @jbartosik

The failed test doesn't seem to have anything do with the change in this PR. It seems to be flakiness due to some randomness the order of keys in the resources map. How should I proceed?

```
--- FAIL: TestGetPatchesForResourceRequest_TwoReplacementResources (0.00s)
    server_test.go:136: 
        	Error Trace:	server_test.go:136
        	            				server_test.go:410
        	Error:      	Should be true
        	Test:       	TestGetPatchesForResourceRequest_TwoReplacementResources
        	Messages:   	got {Op:add Path:/metadata/annotations/vpaUpdates Value:Pod resources updated by name: container 0: unobtanium request, cpu request}, want: {Op:add Path:/metadata/annotations/vpaUpdates Value:Pod resources updated by name: container 0: cpu request, unobtanium request}
```

jbartosik:
The failure was caused by #2886. I think I have fixed the issue now.

/retest

vertical-pod-autoscaler/pkg/recommender/model/aggregations_config.go
```diff
@@ -17,22 +17,23 @@ limitations under the License.
...
 
 	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/recommender/util"
 )
 
 var (
 	// MemoryAggregationWindowLength is the length of the memory usage history
-	// aggregated by VPA, which is 8 days.
-	MemoryAggregationWindowLength = time.Hour * 8 * 24
+	// aggregated by VPA.
+	MemoryAggregationWindowLength = DefaultMemoryAggregationWindowLength
```
bskiba: The default assigned here is overriden anyway by the flag default.

Can we have one of two schemes (I'm using MemoryAggregationWindowLength as an example).

1. Directly here:
````
MemoryAggregationWindowLength = flag.Duration("memory-aggregation-window-length", DefaultMemoryAggregationWindowLength, 'The length of the memory usage history aggregated by VPA.')
````
change uses to *MemoryAggregationWindowLength


2. Separate flags defined here or in main.go
```
var(
RecommenderMemoryAggregationWindowLength = flag.Duration(&, "recommender-memory-aggregation-window-length", DefaultMemoryAggregationWindowLength, 'The length of the memory usage history aggregated by VPA.')
)
````
and here 
````
MemoryAggegationWindowLength = *RecommenderMemoryAggregationWindowLength
````
Although ideally we would have aggregationConfig struct with those values passed around.

File: vertical-pod-autoscaler/pkg/recommender/model/aggregations_config.go [hunk 2]
```diff
@@ -17,22 +17,23 @@ limitations under the License.
...
 	// MemoryAggregationInterval is the length of a single interval, for
 	// which the peak memory usage is computed.
 	// Memory usage peaks are aggregated in daily intervals. In other words
 	// there is one memory usage sample per day (the maximum usage over that
 	// day).
 	// Note: AggregationWindowLength must be integrally divisible by this value.
```
bskiba: There are configurations which will be invalid with the current algorithm. Perhaps we should consider exposing different parameters as flags - for example MemoryAggregationInterval and MemoryAggregationWindowIntervalCount. Then MemoryAggregationWindow = MemoryAggregationInterval *  MemoryAggregationWindowIntervalCount

vertical-pod-autoscaler/pkg/recommender/model/aggregations_config.go
```diff
@@ -83,3 +94,10 @@ func memoryHistogramOptions() util.HistogramOptions {
 	}
 	return options
 }
+
+func init() {
+	flag.DurationVar(&MemoryAggregationWindowLength, "memory-aggregation-window-length", DefaultMemoryAggregationWindowLength, `The length of the memory usage history aggregated by VPA.`)
+	flag.DurationVar(&MemoryAggregationInterval, "memory-aggregation-interval", DefaultMemoryAggregationInterval, `The length of a single interval, for which the peak memory usage is computed. Memory usage peaks are aggregated in daily intervals. In other words there is one memory usage sample per day (the maximum usage over that day). Note: AggregationWindowLength must be integrally divisible by this value.`)
```
bskiba: "Memory usage peaks are aggregated in daily intervals." is not true any more.

vertical-pod-autoscaler/pkg/recommender/main.go
```diff
@@ -22,6 +22,7 @@ import (
 
 	"k8s.io/autoscaler/vertical-pod-autoscaler/common"
 	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/recommender/input/history"
+	_ "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/recommender/model"
```
bskiba: I think you can remove this import

amshuman-kr:
> Do you already have a use-case where you would like to modify those values? If yes can you share it? I'd like to understand what problems there are with the default implementation we propose

Yes. We have clusters with non-batch workloads which still have load pattern cycles of less than 24h. Sometimes the workloads themselves might even last less than 24h (but still > 12h).

We are using VPA to scale these workloads vertically. This has worked well for scaling up. But not so much for scaling down. With the default values for aggregations config, scale down happens after 24h which might be too late or might not even happen as the next load cycle has started for the next day. We were thinking of tweaking these configurations to be less than 24h. Especially, the decay half-lives to something like 4h or 8h.

vertical-pod-autoscaler/pkg/recommender/model/aggregations_config.go
amshuman-kr: Sure. I could do that. I just went with the lexical change. I like the consistency of the first suggestion.

```
MemoryAggregationWindowLength = flag.Duration("memory-aggregation-window-length", DefaultMemoryAggregationWindowLength, 'The length of the memory usage history aggregated by VPA.')
```

I will make the changes.

vertical-pod-autoscaler/pkg/recommender/model/aggregations_config.go
amshuman-kr: Is this to make sure that `MemoryAggregationWindow` would be integrally divisible by `MemoryAggregationInterval`?

vertical-pod-autoscaler/pkg/recommender/model/aggregations_config.go
amshuman-kr: Copy-paste problem (copied from the variable comments) 😅 . I will change it.

vertical-pod-autoscaler/pkg/recommender/main.go
amshuman-kr: With the [above](https://github.com/kubernetes/autoscaler/pull/2905#discussion_r390289950) change, yes.

vertical-pod-autoscaler/pkg/recommender/model/aggregations_config.go
amshuman-kr: I bit the bullet and introduced the [`AggregationsConfig` struct](https://github.com/kubernetes/autoscaler/pull/2905/files#diff-7f21ec2575f476dc048bce486d6eec8dR25-R54), though still as a [singleton instance](https://github.com/kubernetes/autoscaler/pull/2905/files#diff-7f21ec2575f476dc048bce486d6eec8dR118) which is now initialised in the [main.go based on the command-line flags](https://github.com/kubernetes/autoscaler/pull/2905/files#diff-483c01ba6f34fd6f8c9d4944684cff98R67).

For the test-cases, this instance may not be initialised in which case it would be [initialised to the current defaults](https://github.com/kubernetes/autoscaler/pull/2905/files#diff-7f21ec2575f476dc048bce486d6eec8dR123).

vertical-pod-autoscaler/pkg/recommender/model/aggregations_config.go
amshuman-kr: I have changed the flag to [`memory-aggregation-window-interval-count`](https://github.com/kubernetes/autoscaler/pull/2905/files#diff-483c01ba6f34fd6f8c9d4944684cff98R55) and introduced a [func](https://github.com/kubernetes/autoscaler/pull/2905/files#diff-483c01ba6f34fd6f8c9d4944684cff98R55) to compute `MemoryAggregationWindowLength`.

vertical-pod-autoscaler/pkg/recommender/main.go
amshuman-kr: Now, I can't anymore because of the need to [initialise](https://github.com/kubernetes/autoscaler/pull/2905/files#diff-483c01ba6f34fd6f8c9d4944684cff98R67) aggregations config struct singleton instance.

vertical-pod-autoscaler/pkg/recommender/model/aggregations_config.go
amshuman-kr: [Fixed](https://github.com/kubernetes/autoscaler/pull/2905/files#diff-483c01ba6f34fd6f8c9d4944684cff98R54).

amshuman-kr:
@bskiba I have addressed all of your review comments. Could you PTAL?

amshuman-kr:
cc @bskiba @jbartosik @krzysied

vertical-pod-autoscaler/pkg/recommender/main.go
```diff
@@ -42,14 +43,18 @@ var (
...
+	podNamespaceLabel                    = flag.String("pod-namespace-label", "kubernetes_namespace", `Label name to look for container names`)
+	podNameLabel                         = flag.String("pod-name-label", "kubernetes_pod_name", `Label name to look for container names`)
+	ctrNamespaceLabel                    = flag.String("container-namespace-label", "namespace", `Label name to look for container names`)
+	ctrPodNameLabel                      = flag.String("container-pod-name-label", "pod_name", `Label name to look for container names`)
+	ctrNameLabel                         = flag.String("container-name-label", "name", `Label name to look for container names`)
+	memoryAggregationInterval            = flag.Duration("memory-aggregation-interval", model.DefaultMemoryAggregationInterval, `The length of a single interval, for which the peak memory usage is computed. Memory usage peaks are aggregated in multiples of this interval. In other words there is one memory usage sample per interval (the maximum usage over that interval)`)
```
bskiba, line 54: Can you put those flags in a separate block? The ones above are for prometheus.

vertical-pod-autoscaler/pkg/recommender/main.go
```diff
@@ -42,14 +43,18 @@ var (
...
+	memoryAggregationWindowIntervalCount = flag.Int64("memory-aggregation-window-interval-count", model.DefaultMemoryAggregationWindowIntervalCount, `The number of consecutive memory-aggregation-intervals which make up the MemoryAggregationWindowLength which in turn is the period for memory usage aggregation by VPA. In other words, MemoryAggregationWindowLength = memory-aggregation-interval * memory-aggregation-window-interval-count.`)
```
bskiba, line 55: Thanks for the great descriptions here, these are really helpful.

To simplify a bit, can we make this:
memoryAggregationIntervalCount = flag.Int64("memory-aggregation-interval-count", model.DefaultMemoryAggregationIntervalCount, `The number of consecutive memory-aggregation-intervals which make up the time window for memory usage aggregation. In other words, MemoryAggregationWindowLength = memory-aggregation-interval * memory-aggregation-interval-count.`)

vertical-pod-autoscaler/pkg/recommender/main.go
amshuman-kr: Done

vertical-pod-autoscaler/pkg/recommender/main.go
amshuman-kr: Thanks for the suggestion. Done.

amshuman-kr:
@bskiba Thanks for the review and I am sorry for the delay in response. I was not keeping well last week.

I have made the changes as you suggested. Can you PTAL?

amshuman-kr:
@bskiba @jbartosik @krzysied. Could you PTAL?

bskiba:
I'll take a look today, apolgies for the delay!

amshuman-kr:
@bskiba Thanks a lot for your reviews, support and approval!

PR #2906, opened 2020-03-09
Fix VPA admission controller server test

### LINKED ISSUE #2886: Fix flaky admission controller unit test

jbartosik:
It seems that order of CPU and unobtanium is somewhat random, which was
causing flaky tests:

fixes #2886

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/logic/server_test.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/logic/server_test.go
@@ -407,7 +407,9 @@ func TestGetPatchesForResourceRequest_TwoReplacementResources(t *testing.T) {
 		assertPatchOneOf(t, patches[1], []patchRecord{cpuUpdate, unobtaniumUpdate})
 		assert.False(t, eqPatch(patches[0], patches[1]))
 		assertEqPatch(t, patches[2], getAddEmptyAnnotationsPatch())
-		assertEqPatch(t, patches[3], addAnnotationRequest([][]string{{cpu, unobtanium}}, request))
+		cpuFirstUnobtaniumSecond := addAnnotationRequest([][]string{{cpu, unobtanium}}, request)
+		unobtaniumFirstCpuSecond := addAnnotationRequest([][]string{{unobtanium, cpu}}, request)
+		assertPatchOneOf(t, patches[3], []patchRecord{cpuFirstUnobtaniumSecond, unobtaniumFirstCpuSecond})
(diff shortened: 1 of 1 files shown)


krzysied:
Additional context: This problem is caused by https://github.com/kubernetes/autoscaler/blob/4317838e1971d7c7cda6649f96c68c5a1720f160/vertical-pod-autoscaler/pkg/admission-controller/logic/server.go#L168. Golang assumes that map can be iterate in any order. This makes our test result non-deterministic.

vertical-pod-autoscaler/pkg/admission-controller/logic/server_test.go
```diff
@@ -407,7 +407,9 @@ func TestGetPatchesForResourceRequest_TwoReplacementResources(t *testing.T) {
 		assertPatchOneOf(t, patches[1], []patchRecord{cpuUpdate, unobtaniumUpdate})
 		assert.False(t, eqPatch(patches[0], patches[1]))
 		assertEqPatch(t, patches[2], getAddEmptyAnnotationsPatch())
-		assertEqPatch(t, patches[3], addAnnotationRequest([][]string{{cpu, unobtanium}}, request))
+		cpuFirstUnobtaniumSecond := addAnnotationRequest([][]string{{cpu, unobtanium}}, request)
```
krzysied: This solution requires all n! permutation sets of n resources to be created. However in out tests there shouldn't be more than 2 resources per test so this shouldn't be a problem, right?

jbartosik: Yes, I went with this because there are only 2 values to rearrange here. If we ever have 3 or more values appearing in undeterministic order then it will probably be worth to implement matcher that's not sensitive to their order.

krzysied:
/lgtm
/approve

Thanks!

PR #2918, opened 2020-03-10
Separate admission controller logic for pods and vpas

bskiba:
No logic changes, just restructuring the code.


bskiba:
@tweks FYI

PR #2924, opened 2020-03-11
Add some more documentation to clarify how labels and GPUs work with the

AaronKalair:
Hi,

I experienced some confusion when trying to use the Cluster Autoscaler and so I've tried to document some of the things I ran into to help others in the future.

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/README.md
+++ b/cluster-autoscaler/cloudprovider/aws/README.md
@@ -162,6 +162,42 @@ If you'd like to scale node groups from 0, an `autoscaling:DescribeLaunchConfigu
 }
 ```
 
+### Gotchas
+
+* Without these tags, when the cluster autoscaler needs to increase the number of nodes, if a node group creates nodes with taints that the pending pod does not tolerate then the cluster autoscaler will only learn about this after the node has been created and it sees that it is tainted. From this point on this information will be cached and subsequent scaling operations will take this into account, but it means that the behaviour of the cluster autoscaler differs between the first and subsequent scale up requests and can lead to confusion.
+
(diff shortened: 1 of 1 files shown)


AaronKalair:
CLA Signed

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -162,6 +162,40 @@ If you'd like to scale node groups from 0, an `autoscaling:DescribeLaunchConfigu
 }
 ```
 
+### Gotchas
+
+Without these tags, when the cluster autoscaler needs to increase the number of nodes, if a node group creates nodes with taints that the pending pod does not tolerate then the cluster autoscaler will only learn about this after the node has been created and it sees that it is tainted. From this point on this information will be cached and subsequent scaling operations will take this into account, but it means that the behaviour of the cluster autoscaler differs between the first and subsequent scale up requests and can lead to confusion.
```
Jeffwan: Thanks! This is great. Do you want to tell the reason GPU takes time to start up here? Like device plugin takes time to advertise resources to APIServer which makes pod can not be scheduled on the ready node. That's the major reason CA have unnecessary scale up again. With the label, CA will wait for GPU to be ready.

AaronKalair: Added, and improved the formatting in the latest commit, thanks!

Jeffwan: Thanks. One more thing, could you squash the commits to 1?

AaronKalair: Have squashed the commits and force pushed

Jeffwan:
/lgtm
/approve

Thanks again for the contribution! @AaronKalair  Let me know if you have other GPU use case not implemented in CA

Issue #2888, opened 2020-03-04
Scale up windows nodegroups from 0 with AWS EKS Cluster 


sppwf:
Hi Guys,

I am using a Kubernetes cluster using AWS EKS Service. 
Cluster version 1.14, latest AWS offers
I am using Windows nodegroup with AWS supported vpc controller and webhook that adds required requests/limits to pods specs
I try to scale up windows ASG from 0 to 2. The idea is that it works for a couple of days and after that, it does not. 
the Workaround to get it working again is to change ASG desired setting to 1, node comes up, after that Cluster autoscaler does scale down (back to 0) and up (from 0 to 2) for a couple of days.

Here is the Pod output
```
$ kubectl describe pods win64-f0hql
Name:               win64-f0hql
Namespace:          default
Priority:           0
PriorityClassName:  <none>
Node:               <none>
Labels:             jenkins/label=windows
                    os=windows
Annotations:        kubernetes.io/psp: eks.privileged
Status:             Pending
IP:
Containers:
  jnlp:
    Image:      artifactory/win64:12.0
    Port:       <none>
    Host Port:  <none>
    Limits:
      cpu:                                   7
      memory:                                16Gi
      vpc.amazonaws.com/PrivateIPv4Address:  1
    Requests:
      cpu:                                   7
      memory:                                13Gi
      vpc.amazonaws.com/PrivateIPv4Address:  1
    Mounts:
      /home/jenkins/agent from workspace-volume (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-5qm8x (ro)
Conditions:
  Type           Status
  PodScheduled   False
Volumes:
  workspace-volume:
    Type:       EmptyDir (a temporary directory that shares a pod's lifetime)
    Medium:
    SizeLimit:  <unset>
QoS Class:       Burstable
Node-Selectors:  beta.kubernetes.io/os=windows
                 os=windows
Tolerations:     node.kubernetes.io/not-ready:NoExecute for 300s
                 node.kubernetes.io/unreachable:NoExecute for 300s
Events:
  Type     Reason             Age                  From                Message
  ----     ------             ----                 ----                -------
  Warning  FailedScheduling   61s (x2 over 2m15s)  default-scheduler   0/2 nodes are available: 2 Insufficient cpu, 2 Insufficient memory, 2 Insufficient vpc.amazonaws.com/PrivateIPv4Address, 2 node(s) didn't match node selector.
[...]

PR #2931, opened 2020-03-13
Support arbitrary custom resource building template

### LINKED ISSUE #2927: Deployment of a new daemonset cause CA "scale-from-zero" behavior unstable.
## Problem
Some ASGs are not able to scale from zero. Pods pending forever.

## Change
A device-plugin is deployed to cluster as K8s daemonset. Therefore for each pod, a new request is added to config:

```
Requests:
      cpu:                            16200m
     <some-new-device-resource>:  1 # This is a new resource request applied to each pod.
      memory:                         6G
      nvidia.com/gpu:                 0
```

## CA Logs
I0312 21:38:06.717094       6 utils.go:208] Pod `<some-pod> `can't be scheduled on `<some-asg>` predicate failed: PodFitsResources predicate mismatch, reason: Insufficient `<some-new-device-resource>`

## Notes
ASGs with gpu instances are working fine as expected. But ASGs with cpu instances are **NOT** able to scale up from zero.

Jeffwan:
Scale from 0 need to build node template from ASG tags. Currently, we only support `ephemeral -storage`. We'd like to extend to arbitrary custom resources. 

Resolves 
#2927 -> node level devices
#2888 -> EKS Windows IP resource. 

Signed-off-by: Jiaxin Shan <seedjeffwan@gmail.com>

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_manager_test.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_manager_test.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"reflect"
 	"sort"
+	"strconv"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


xinxingliu90:
The impact of this change is that it guarantees our asg instances to be terminated when idle, which saved a lot of our reserved capacity. Thank you @Jeffwan, and our team look forward to this change!

cluster-autoscaler/cloudprovider/aws/aws_manager_test.go
```diff
@@ -111,6 +112,74 @@ func TestExtractAllocatableResourcesFromAsg(t *testing.T) {
...
+		},
+	})
+	assert.NoError(t, observedErr)
+	esValue, esExist := observedNode.Status.Capacity[apiv1.ResourceName(ephemeralStorageKey)]
+	assert.True(t, esExist)
+	assert.Equal(t, int64(20), esValue.Value())
+	_, ipExist := observedNode.Status.Capacity[apiv1.ResourceName(vpcIPKey)]
+	assert.False(t, ipExist)
+
+	// Nod with labels
```
jaypipes: s/Nod/Node/

cluster-autoscaler/cloudprovider/aws/aws_manager_test.go
```diff
@@ -111,6 +112,74 @@ func TestExtractAllocatableResourcesFromAsg(t *testing.T) {
...
+	GPULabelValue := "nvidia-telsa-v100"
+	observedNode, observedErr = awsManager.buildNodeFromTemplate(asg, &asgTemplate{
+		InstanceType: c5Instance,
+		Tags: []*autoscaling.TagDescription{
+			{
+				Key:   aws.String(fmt.Sprintf("k8s.io/cluster-autoscaler/node-template/label/%s", GPULabel)),
```
jaypipes: I noticed that the singular "label" and "taint" are used in the Key's string here and below, but the plural "resources" is used in the Key string above. Might be nice in the future to make these consistently plural (preserving backwards-compat of course)

Jeffwan: Agree. They are inconsistent now and it would be great to have them unified.

cluster-autoscaler/cloudprovider/aws/aws_manager_test.go
Jeffwan: I will correct the spelling in future PRs. Seems words check scripts doesn't catch it..

Issue #2932, opened 2020-03-13
Let the CAPI group be overridden and discover the preferred version


enxebre:
We are currently hardcoding the CAPI groups/version supported by the provider https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go#L41

We should make the group overridable via env variable and let the version be discovered via discovery client.

enxebre:
@MaciekPytel would you please create a label for `area/provider/capi`, I don't have perms for that.

MaciekPytel:
I created area/provider/cluster-api. Is that ok? I can rename if you prefer, but I think it's more clear compared to capi.

enxebre:
thanks @MaciekPytel sgtm. I don't seem to have perms to add a label manually via GH UX. Is there any automation we can use for this? e.g:
/area provider/cluster-api

MaciekPytel:
Apparently yes :)
Looking at bot code it seems to just match anything starting with "area/".

PR #2934, opened 2020-03-13
Fixes 2932: let the capi version to be discovered

enxebre:
Add the ability to override CAPI group via env variable and discovers the API version.

This change adds detection for an environment variable to specify the group for the clusterapi resources. If the environment variable `CAPI_GROUP` is specified, then it will be used instead of the default. This also decouples the API group from the version and let the latter to be discovered dynamically.

Fixes https://github.com/kubernetes/autoscaler/issues/2932

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
+++ b/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
@@ -19,12 +19,14 @@ package clusterapi
 import (
 	"context"
 	"fmt"
+	"os"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/informers"
@@ -38,7 +40,9 @@ import (
 const (
 	machineProviderIDIndex = "machineProviderIDIndex"
 	nodeProviderIDIndex    = "nodeProviderIDIndex"
-	defaultMachineAPI      = "v1alpha2.cluster.x-k8s.io"
+	defaultCAPIGroup       = "cluster.x-k8s.io"
+	// CAPIGroupEnvVar contains the environment variable name which allows overriding defaultCAPIGroup.
+	CAPIGroupEnvVar = "CAPI_GROUP"
 )
 
 // machineController watches for Nodes, Machines, MachineSets and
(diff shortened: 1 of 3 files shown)


enxebre:
/hold
To do some manual validations
cc @frobware @elmiko

elmiko:
> Codewise looks good, just wondering whether we want to be calling all the variables etc Machine API or Cluster API, do we have a preference?

agreed, i think if we can use "cluster api" here it will reduce any confusion for someone reading this. at the very least we should name the external bits (eg env variable) to reflect cluster api.

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
```diff
@@ -281,15 +296,16 @@ func newMachineController(
...
-	// TODO(alberto): let environment variable to override defaultMachineAPI
-	machineDeploymentResource, _ := schema.ParseResourceArg(fmt.Sprintf("machinedeployments.%v", defaultMachineAPI))
+	specifiedMachineAPI := getAPIVersionGroup()
+
+	machineDeploymentResource, _ := schema.ParseResourceArg(fmt.Sprintf("machinedeployments.%v", specifiedMachineAPI))
 
-	machineSetResource, _ := schema.ParseResourceArg(fmt.Sprintf("machinesets.%v", defaultMachineAPI))
+	machineSetResource, _ := schema.ParseResourceArg(fmt.Sprintf("machinesets.%v", specifiedMachineAPI))
 	if machineSetResource == nil {
 		panic("MachineSetResource")
```
frobware: This should really return an error.

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller_test.go
```diff
@@ -991,3 +992,110 @@ func TestControllerMachineSetNodeNamesUsingStatusNodeRefName(t *testing.T) {
 		}
 	}
 }
+
+func TestControllerGetAPIVersionGroup(t *testing.T) {
+	expected := "myversion.mygroup"
+	err := os.Setenv(machineAPIVersionGroupEnvVar, expected)
```
frobware: nit: `if err := os.Setenv(machineAPIVersionGroupEnvVar, expected); err != nil` ...

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller_test.go
```diff
@@ -991,3 +992,110 @@ func TestControllerMachineSetNodeNamesUsingStatusNodeRefName(t *testing.T) {
...
+
+	machineDeployments, err := controller.listMachineDeployments(testNamespace, labels.Everything())
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if l := len(machineDeployments); l != 1 {
+		t.Fatalf("Incorrect number of MachineDeployments, expected 1, got %d", l)
+	}
+
+	machineSets, err := controller.listMachineSets(testNamespace, labels.Everything())
```
frobware: Is this belt and braces? We're testing Machine Deployments in this test; are just verifying that it doesn't error? If so could we collapse `TestControllerGetAPIVersionGroupWithMachineSets` and `TestControllerGetAPIVersionGroupWithMachineDeployments`.

enxebre:
>Codewise looks good, just wondering whether we want to be calling all the variables etc Machine API or Cluster API, do we have a preference?

>agreed, i think if we can use "cluster api" here it will reduce any confusion for someone reading this. at the very least we should name the external bits (eg env variable) to reflect cluster api.

I renamed to CAPI consistently and move code to fetch the preferred version into unit tested `getAPIGroupPreferredVersion`

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller_test.go
enxebre: `TestControllerGetAPIVersionGroupWithMachineSets` seems pretty much covered by `TestControllerGetAPIVersionGroupWithMachineDeployments` may be just drop the first one and rename the latter one to `TestControllerGetAPIVersionGroup`, cc @elmiko ?

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
enxebre: mm why? atm `MachineDeployment`, `MachineSets and `Machine` are intentionally assumed to exist so it seems reasonable to me to panic here?

frobware: Because it makes unit tests easier to write and assert for expected behaviour.

enxebre: ok make sense, let's address this separately from this PR. I'll move the capi resource expectations into its own unit tested `func`

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller_test.go
elmiko: that sounds fair to me, i originally added both tests since it wasn't clear to me if there were different mechanisms for machinesets and machinedeployments.

enxebre:
>in general it looks good to me, deduplicating the tests seems like a good idea.
 
Addressed comments and meld into one commit. PTAL @elmiko @frobware @JoelSpeed

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller_test.go
```diff
@@ -76,7 +79,19 @@ func mustCreateTestController(t *testing.T, testConfigs ...*testConfig) (*machin
 
 	kubeclientSet := fakekube.NewSimpleClientset(nodeObjects...)
 	dynamicClientset := fakedynamic.NewSimpleDynamicClient(runtime.NewScheme(), machineObjects...)
-	controller, err := newMachineController(dynamicClientset, kubeclientSet)
+	discoveryClient := &fakediscovery.FakeDiscovery{
+		Fake: &clientgotesting.Fake{
+			Resources: []*v1.APIResourceList{
+				{
+					GroupVersion: "machine.openshift.io/v1beta1",
```
frobware: Why do we reference openshift here?

frobware: OK, I see. We're testing discovery. But, I would still choose a more agnostic name for the case that should not be found.

enxebre: this is used for validating env variable overridability, let me update it to something like "custom" to avoid confusion.

enxebre: Updated to reference `customCAPIGroup = "custom.x-k8s.io"`

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
```diff
@@ -332,6 +357,28 @@ func newMachineController(
...
+		return "", fmt.Errorf("failed to get ServerGroups: %v", err)
+	}
+
+	foundAPIGroup := false
+	var APIVersion string
+	for _, group := range groupList.Groups {
+		if group.Name == APIGroup {
+			foundAPIGroup = true
+			APIVersion = group.PreferredVersion.Version
+			break
```
frobware: Why not just `return` at this point? You can drop some variables and conditionals below.

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller_test.go
```diff
@@ -136,7 +153,7 @@ func createTestConfigs(specs ...testSpec) []*testConfig {
 
 		config.machineSet = &MachineSet{
 			TypeMeta: v1.TypeMeta{
-				APIVersion: "cluster.x-k8s.io/v1alpha2",
+				APIVersion: "cluster.x-k8s.io/v1alpha3",
```
frobware: Can these strings literals not use `defaultCAPIGroup`?

enxebre:
>The issue (#2934) talks about discovering the version but, unless I'm mistaken, this only allows the group to be overridden. True?

This enable an env variable to override the default CAPI group and makes the discovers the API version (https://github.com/kubernetes/autoscaler/issues/2932) see https://github.com/kubernetes/autoscaler/pull/2934/files#diff-89363471d762322b93c9bcfec57b5af8R302. Let me update the PR desc with the commit message.

PR #2944, opened 2020-03-19
Properly propagate scale-up failures in scale sets

marwanad:
The following splits the scale-up operation into two parts:

- Synchronous short lived call that returns a future. This allows capturing quota failures and other early failures and properly propagate those tocore autoscaler so that the nodegroup can be backed-off. Otherwise, in the case of quota failures for example, retries will occur in every autoscaler loop leading to throttling.

- Asynchronous part that waits on the future and logs those failures.

/area provider/azure


marwanad:
@feiskyer added a test for the failure case. The success should be covered by earlier test cases.

feiskyer:
@marwanad how about other branches? let's merge them together

marwanad:
@feiskyer opened up the other PRs.

PR #2950, opened 2020-03-20
Let the controller move on if machineDeployments are not available

enxebre:
There might be adhoc environments where machineDeployments might not necessarily be available. This let the controller to remain functional for such scenarios.

/area provider/cluster-api

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
+++ b/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
@@ -186,7 +186,9 @@ func (c *machineController) run(stopCh <-chan struct{}) error {
 		c.nodeInformer.HasSynced,
 		c.machineInformer.Informer().HasSynced,
 		c.machineSetInformer.Informer().HasSynced,
-		c.machineDeploymentInformer.Informer().HasSynced,
+	}
+	if c.machineDeploymentResource != nil {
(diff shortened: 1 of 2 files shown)


enxebre:
cc @frobware @elmiko

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
```diff
@@ -509,26 +542,28 @@ func (c *machineController) nodeGroupForNode(node *corev1.Node) (*nodegroup, err
...
-		// to a nodegroup that has a scale size of at
-		// least 1.
-		if nodegroup.MaxSize()-nodegroup.MinSize() < 1 {
-			return nil, nil
+	if c.machineDeploymentResource != nil {
+		if ref := machineSetMachineDeploymentRef(machineSet); ref != nil {
+			key := fmt.Sprintf("%s/%s", machineSet.Namespace, ref.Name)
+			machineDeployment, err := c.findMachineDeployment(key)
+			if err != nil {
+				return nil, fmt.Errorf("unknown MachineDeployment %q: %v", key, err)
```
elmiko: should we differentiate this error text from the following clause?

maybe like `"Error finding MachineDeployment %q: %v"`

elmiko: thinking about this more, including the error text in this response is probably enough to differentiate with the other error. i just wanted to make sure that an operator who sees this error had a chance to figure out the root cause, but i think they will.

PR #2952, opened 2020-03-20
Add oom test

jbartosik:


vertical-pod-autoscaler/e2e/v1/full_vpa.go
```diff
@@ -120,6 +120,52 @@ var _ = FullVpaE2eDescribe("Pods under VPA", func() {
 	})
 })
 
+var _ = FullVpaE2eDescribe("OOMing pods under VPA", func() {
+	var (
+		vpaClientSet *vpa_clientset.Clientset
+		vpaCRD       *vpa_types.VerticalPodAutoscaler
+	)
+	replicas := 3
```
krzysied: const?

vertical-pod-autoscaler/e2e/v1/autoscaling_utils.go
```diff
@@ -427,3 +429,25 @@ func runServiceAndWorkloadForResourceConsumer(c clientset.Interface, ns, name st
...
+		Namespace:   ns,
+		Timeout:     timeoutRC,
+		Replicas:    replicas,
+		Annotations: make(map[string]string),
+	}
+
+	dpConfig := testutils.DeploymentConfig{
+		RCConfig: rcConfig,
+	}
+	ginkgo.By(fmt.Sprintf("creating deployment %s in namespace %s", dpConfig.Name, dpConfig.Namespace))
```
krzysied: nit: Capital letter.

bskiba:
Just a side note: Please remember to add to v1beta2 as well as long as we have it.

vertical-pod-autoscaler/e2e/v1/full_vpa.go
```diff
@@ -120,6 +120,52 @@ var _ = FullVpaE2eDescribe("Pods under VPA", func() {
...
+
+	// This schedules AfterEach block that needs to run after the AfterEach above and
```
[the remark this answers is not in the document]
jbartosik: Left over from c&p; removed

vertical-pod-autoscaler/e2e/v1/full_vpa.go
```diff
@@ -120,6 +120,52 @@ var _ = FullVpaE2eDescribe("Pods under VPA", func() {
...
+
+	})
+
+	ginkgo.It("have memory requests growing with OOMs", func() {
+		err := waitForResourceRequestInRangeInPods(
+			f, metav1.ListOptions{LabelSelector: "name=hamster"}, apiv1.ResourceMemory,
+			ParseQuantityOrDie("1400Mi"), ParseQuantityOrDie("10000Mi"))
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	})
+
```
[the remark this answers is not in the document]
jbartosik: Removed

vertical-pod-autoscaler/e2e/v1/full_vpa.go
```diff
@@ -120,6 +120,52 @@ var _ = FullVpaE2eDescribe("Pods under VPA", func() {
...
+			APIVersion: "v1",
+			Kind:       "ReplicationController",
+			Name:       "hamster",
+		})
+
+		vpaClientSet = vpa_clientset.NewForConfigOrDie(config)
+		vpaClient := vpaClientSet.AutoscalingV1()
+		_, err = vpaClient.VerticalPodAutoscalers(ns).Create(vpaCRD)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
```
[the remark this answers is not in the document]
jbartosik: Removed

vertical-pod-autoscaler/e2e/utils/ooming_resource_consumer/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM gcr.io/google-containers/stress:v1
+ENTRYPOINT ["/stress", "--mem-total", "10000000000", "-logtostderr"]
```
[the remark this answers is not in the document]
jbartosik: I changed to `--logtostderr` and it still works so I think it doesn't matter. I'll keep version with double dash to make it prettier

vertical-pod-autoscaler/e2e/v1/autoscaling_utils.go
```diff
@@ -427,3 +429,25 @@ func runServiceAndWorkloadForResourceConsumer(c clientset.Interface, ns, name st
 	framework.ExpectNoError(framework.WaitForServiceEndpointsNum(
 		c, ns, controllerName, 1, startServiceInterval, startServiceTimeout))
 }
+
+func runOomingReplicationController(c clientset.Interface, ns, name string, replicas int) {
+	ginkgo.By(fmt.Sprintf("Running fixed resource consuming RC %s with %v replicas", name, replicas))
```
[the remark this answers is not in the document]
jbartosik: Done

vertical-pod-autoscaler/e2e/v1/full_vpa.go
jbartosik: Done

vertical-pod-autoscaler/e2e/v1/full_vpa.go
```diff
@@ -120,6 +120,52 @@ var _ = FullVpaE2eDescribe("Pods under VPA", func() {
 	})
 })
 
+var _ = FullVpaE2eDescribe("OOMing pods under VPA", func() {
+	var (
+		vpaClientSet *vpa_clientset.Clientset
```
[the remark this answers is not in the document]
jbartosik: We don't need them. But if we want to have more tests on OOMing pods it will be easier

vertical-pod-autoscaler/e2e/v1/autoscaling_utils.go
jbartosik: Done.

jbartosik:
I forgot that Github sends my replies to comments immediately rather than batching them and waiting for me to send the batch; I now pushed fixed commits; I'll push v1beta2 test later today

jbartosik:
@bskiba I'm surprised - when I try to run the same test using `v1beta2` API it consistently fails (times out).

bskiba:
Weird, that should behave exactly the same...

jbartosik:
@krzysied please take a look.

Tests I'm adding pass now. They were failing because after OOMing pods sometimes stay around for longer time. The test wasn't filtering out failed pods and the old pods had wrong requests. I added filtering and now test consistently passes.

I also changed the loop to iterate over all pods when comparing requests rather than end after finding first pod that doesn't match. This change helped me a lot when I was debugging and I think it's worth keeping

jbartosik:
@bskiba , @krzysied please take a look. I added another commit to the PR - it changes timeout on the OOM test so that the test still passes. But the test modified to use an image that doesn't OOM (but uses enough memory to trigger growth) fails

vertical-pod-autoscaler/e2e/v1/full_vpa.go
```diff
@@ -135,18 +179,19 @@ func waitForPodsMatch(f *framework.Framework, listOptions metav1.ListOptions, ma
 			return false, nil
 		}
 
+		result := true
 		for _, pod := range podList.Items {
 			if !matcher(pod) {
-				return false, nil
+				result = false
```
krzysied: Why the return command was removed?

jbartosik: I moved it to after the loop. This is to evaluate all the pods (instead of exiting after the first pod that doesn't match). It makes debugging easier (for example test was failing because it didn't filter out failed pods. Which was hard to notice when we logged only about the first pod on the list)

krzysied: Please add comment why we want to have it this way.
TBH, it looks weird without a backstory.

jbartosik: Done

krzysied:
no comments
/hold cancel

PR #2968, opened 2020-03-23
Refactor admission controller patch logic

bskiba:
Only actual change that happens here is that empty annotations patch comes before any other patch.
Other than that, this only moves code around. 
I'm doing a restructuring of tests in followup PR.

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/resource/pod/recommendation/recommendation_provider.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/resource/pod/recommendation/recommendation_provider.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package pod
+package recommendation
 
 import (
 	"fmt"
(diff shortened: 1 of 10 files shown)


vertical-pod-autoscaler/pkg/admission-controller/resource/pod/recommendation/recommendation_provider.go
```diff
@@ -36,8 +36,8 @@ type recommendationProvider struct {
 	recommendationProcessor vpa_api_util.RecommendationProcessor
 }
 
-// NewRecommendationProvider constructs the recommendation provider that can be used to determine recommendations for pods.
-func NewRecommendationProvider(calculator limitrange.LimitRangeCalculator,
+// NewProvider constructs the recommendation provider that can be used to determine recommendations for pods.
+func NewProvider(calculator limitrange.LimitRangeCalculator,
 	recommendationProcessor vpa_api_util.RecommendationProcessor) *recommendationProvider {
```
krzysied: nit: I know it unrelated to this PR, but shouldn't this function return interface instead of a pointer?

krzysied: I'm not going to block the PR because of the unrelated nit. Other things look good.

Issue #2762, opened 2020-01-21
Workload controllers may time out possibly due to long timeout on VPA webhook


dharmab:
I'm investigating an issue where workload controllers like the replicaset controller and daemonset controller time out when attempting to create pods. The issue seems to be accompanied by log messages from kube-apiserver indicating that requests to the VPA mutating webhook are timing out. (I haven't yet determined why the VPA webhook times out in the first place.) Deleting the VPA webhook and then either waiting for the controller manager to try creating the pods again or killing the leading controller manager resolves the issue.

I wonder if the 30 second timeout on the VPA webhook is too long for the controller manager? If the workload controller request has a timeout <= 30s, it stands to reason that the controller request could time out before the apiserver ignores the VPA timeout and continues to create the Pod.

bskiba:
Sorry for the long response time, I missed this issue :(

Is it currently possible to change the timeout for admission controllers? I agree 30 seconds is too long, but last time I looked it was not configurable.

dharmab:
I believe the AC is created by vpa-admission-controller after launch- it doesn't seem to be in the manifest.

bskiba:
Yes, that's true, the vpa-admission-controller self-registers after startup. What I mean is do you know if the Admissionregistration Kubernetes API allows for specifying a custom timeout when registering a webhook (30 seconds is the default).

dharmab:
Yes, you can (and should) specify a shorter timeout: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#configure-admission-webhooks-on-the-fly

> Note: Default timeout for a webhook call is 10 seconds for webhooks registered created using admissionregistration.k8s.io/v1, and 30 seconds for webhooks created using admissionregistration.k8s.io/v1beta1. Starting in kubernetes 1.14 you can set the timeout and it is encouraged to use a small timeout for webhooks. If the webhook call times out, the request is handled according to the webhook’s failure policy.

dharmab:
@bskiba My team has hit this issue three times this weekend. Our workaround is to restart vpa-admission-controller when it happens, and we still haven't determined exactly _why_ it times out, but reducing the timeout would certainly help as a mitigation.

I'm also curious why the webhook is hardcoded in vpa-admission-controller and not a separate object in the manifest YAML. We had a similar issue with Open Policy Agent, but were able to mitigate it by changing one field in the YAML instead of going upstream.

dharmab:
[...]

PR #2949, opened 2020-03-20
Allow disabling webhook self registration

povilasv:
This allows users to create custom MutatingWebhookConfiguration objects, allowing users to edit fields like `objectSelector` or `namespaceSelector`.

PR #2969, opened 2020-03-23
Reduce VPA mutating webhook timeout from 30s to 10s

dharmab:
Fixes https://github.com/kubernetes/autoscaler/issues/2762

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/config.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/config.go
@@ -75,6 +75,7 @@ func selfRegistration(clientset *kubernetes.Clientset, caCert []byte, namespace,
 		RegisterClientConfig.URL = &url
 	}
 	sideEffects := v1beta1.SideEffectClassNone
+	timeoutSeconds := int32(10)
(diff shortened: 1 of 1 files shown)


dharmab:
A follow-up PR could make this timeout configurable, but reducing the timeout from the default 30s to the much saner 10s is an improvement and reflects the change in default between the admissionregistration.k8s.io/v1beta1 and admissionregistration.k8s.io/v1 types.

vertical-pod-autoscaler/pkg/admission-controller/config.go
```diff
@@ -75,6 +75,7 @@ func selfRegistration(clientset *kubernetes.Clientset, caCert []byte, namespace,
 		RegisterClientConfig.URL = &url
 	}
 	sideEffects := v1beta1.SideEffectClassNone
+	timeoutSeconds := int32(10)
```
bskiba: Can we make it a flag?

dharmab:
I've decided to close this in favor of using https://github.com/kubernetes/autoscaler/pull/2949 instead.

PR #2971, opened 2020-03-24
Add hack/update-vendor.sh script

losipiuk:
For some reason it was not present on cluster-autoscaler-release-1.15 branch


Jeffwan:
Thanks for the fix. 1.15-branch doesn't have `update-vendor.sh` even if it move to go modules. This helps to update dependency.

PR #2974, opened 2020-03-24
Move observed containers test to calculator

bskiba:


vertical-pod-autoscaler/pkg/admission-controller/resource/pod/patch/observed_containers_test.go
```diff
@@ -0,0 +1,66 @@
...
+			pod:           test.Pod().Get(),
+			expectedPatch: addVpaObservedContainersPatch([]string{}),
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			c := NewObservedContainersCalculator()
+			patches, err := c.CalculatePatches(tc.pod, nil)
+			assert.NoError(t, err)
+			if assert.Len(t, patches, 1, "Unexpected amount of patches.") {
```
krzysied: nit: s/amount/number/

vertical-pod-autoscaler/pkg/admission-controller/resource/pod/patch/util.go
```diff
@@ -39,3 +43,16 @@ func GetAddAnnotationPatch(annotationName, annotationValue string) resource_admi
...
+// EqPatch returns true if patches are equal by comparing their
+// marshalling result.
+func EqPatch(a, b resource_admission.PatchRecord) bool {
+	aJson, aErr := json.Marshal(a)
+	bJson, bErr := json.Marshal(b)
+	return string(aJson) == string(bJson) && aErr == bErr
+}
+
+// AssertEqPatch asserts patches are equal.
+func AssertEqPatch(t *testing.T, got, want resource_admission.PatchRecord) {
```
krzysied: nit: When golang package is vendored, the _test.go are omitted. IMHO it is good to keep code referencing "testing" separated from main code. I'm not sure what is golang official best practice for this case, though.

krzysied: If we consider keeping "AssertEqPatch" in util, maybe we should also have "assertPatchOneOf" there?

bskiba: Moved to test_utils file. I will move assertPatchOneOf with the next PR (resource updates calculator test)

vertical-pod-autoscaler/pkg/admission-controller/resource/pod/patch/observed_containers_test.go
bskiba: Done

Issue #2920, opened 2020-03-11
cluster-autoscaler 1.15.5 does not work with IAM Roles for Service Accounts


johnjeffers:
Just upgraded my EKS cluster to 1.15, and upgraded cluster-autoscaler along with it.

It appears that 1.15.5 has the same problem that we used to have with 1.14. It does not work with IAM Roles for Service Accounts.

Getting the following error:

```E0311 03:36:42.052777       1 aws_manager.go:259] Failed to regenerate ASG cache: cannot autodiscover ASGs: AccessDenied: User: arn:aws:sts::<my-account>:assumed-role/eks-worker/i-059d924ed6f52a032 is not authorized to perform: autoscaling:DescribeTags```

This should not be happening, as the `autoscaling:DescribeTags` permission is assigned to the service-account-based IAM Role, not to the `eks-worker` instance profile.

This problem was fixed in [1.14.6](https://github.com/kubernetes/autoscaler/releases/tag/cluster-autoscaler-1.14.6) but doesn't seem to have made it into 1.15.5

**Comment by rochacon** [comment]
I hit the same issue, downgrading to 1.14.7 seems to work fine. I believe we need to cherry-pick #2323 (or a more recent AWS SDK update).

**Comment by Jeffwan** [comment]
I will go fix this issue. I remember I did the cherry-pick. Need to double check.

**Comment by johnjeffers** [comment]
@Jeffwan Thank you!!

**Comment by Jeffwan** [comment]
@losipiuk @aleksandra-malinowska Trying to cherry-pick a change to v1.15.  that branch doesn't have `./hack/update-vendor.sh` and  `k8s/kubernetes` dependencies points to `/tmp/abc/kubernetes`. 



Another thing is I notice 

```
# Using go mod
build-binary: clean deps
        $(ENVVAR) GOOS=$(GOOS) go build -o cluster-autoscaler ${TAGS_FLAG}

# Seems godep
build-in-docker: clean docker-builder
        docker run -v `pwd`:/gopath/src/k8s.io/autoscaler/cluster-autoscaler/ autoscaling-builder:latest bash -c 'cd /gopath/src/k8s.io/autoscaler/cluster-autoscaler && BUILD_TAGS=${BUILD_TAGS} make build-binary'
```

How can I manage dependencies in this case?

ajohnstone:
Bumping version should solve #2829 with sts private endpoints too

canhnt:
[...]

Issue #2829, opened 2020-02-13
Cluster-autoscaler does not work with VPC-Endpoint (AWS/EKS)


maust:
We do have a quite locked-down network in AWS (no internet-connectivity at all). Access to AWS services only via VPC endpoints and on-premise systems via DirectConnect. At the same time we would like to use IAM roles.

Kubernetes: 1.14
Cluster-Autoscaler: 1.14.7

When using cluster-autoscaler it cannot fetch credentials via STS using the IAM role. To my understanding the issue is caused by cluster-autoscaler not using the regional STS endpoint (https://sts.eu-central-1.amazonaws.com) but instead the global (https://sts.amazonaws.com). With VPC-Endpoints it is not possible to replace the global enpoint.

With github.com/aws/aws-sdk-go v1.25.18 
(see https://github.com/aws/aws-sdk-go/blob/master/CHANGELOG.md) configuration of regional STS endpoints was introduced by setting env AWS_STS_REGIONAL_ENDPOINTS=regional

I tried setting the region as env AWS_REGION and AWS_STS_REGIONAL_ENDPOINTS but still the global endpoint is used. 

After looking at the 1.14 branch for cluster-autoscaler, it looks like v1.23.22 is used (see
https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-release-1.14/cluster-autoscaler/vendor/github.com/aws/aws-sdk-go/CHANGELOG.md)

I also checked the other cluster-autoscaler branches:
- 1.14 github.com/aws/aws-sdk-go v1.23.22
- 1.15 github.com/aws/aws-sdk-go v1.16.26
- 1.16 github.com/aws/aws-sdk-go v1.23.12
- 1.17 github.com/aws/aws-sdk-go v1.23.18
- master github.com/aws/aws-sdk-go v1.28.2

So I would assume that supporting such a use case would be possible by upgrading the aws-sdk-go version to >= v1.25.18 - let me know if I can be of help.

logs:
E0213 16:05:54.390164       1 aws_manager.go:259] Failed to regenerate ASG cache: cannot autodiscover ASGs: WebIdentityErr: failed to retrieve credentials
caused by: RequestError: send request failed
caused by: Post https://sts.amazonaws.com/: dial tcp 54.239.29.25:443: i/o timeout
[...]

PR #2978, opened 2020-03-24
Update aws sdk to v1.28.14

Jeffwan:
Resolve issue in 1.15 branch. 

https://github.com/kubernetes/autoscaler/issues/2920
https://github.com/kubernetes/autoscaler/issues/2829


losipiuk:
Thanks!

Generally looks good though I would prefer to revendor against either:
a) named 1.15 version (newest one is 1.15.11) `./hack/update-vendor.sh -rv1.15.11`
b) same revision which was used originally `./hack/update-vendor.sh -ref7808fec57284582c7bf3fc834570e5769df83e`

The tip of `k8s.io/kubernetes:release-1.15` probably is fine too - but still there is a chance that it has some bugs as it does not correspond to any released version of k8s right now.

I have no strong opinion on which of the options above is better. b) will results in smaller delta while a) may incorporate some fixes which may be good to have anyway. 
I think I would go for a)

Jeffwan:
> Thanks!
> 
> Generally looks good though I would prefer to revendor against either:
> a) named 1.15 version (newest one is 1.15.11) `./hack/update-vendor.sh -rv1.15.11`
> b) same revision which was used originally `./hack/update-vendor.sh -ref7808fec57284582c7bf3fc834570e5769df83e`
> 
> The tip of `k8s.io/kubernetes:release-1.15` probably is fine too - but still there is a chance that it has some bugs as it does not correspond to any released version of k8s right now.
> 
> I have no strong opinion on which of the options above is better. b) will results in smaller delta while a) may incorporate some fixes which may be good to have anyway.
> I think I would go for a)

Thanks for the advice and I will go for (a) and revendor dependencies.

PR #3003, opened 2020-03-30
Upgrade aws-sdk-go to v1.28.14

### LINKED ISSUE #2829: Cluster-autoscaler does not work with VPC-Endpoint (AWS/EKS)

Jeffwan:
Resolve #2829 

1. ~The reason I delete `github.com/google/go-querystring` in go.mod-extra~
```
Operating in /tmp/ca-update-vendor.HPnf
Sending logs to: /tmp/ca-update-vendor.HPnf/ca-update-vendor.log
Updating vendor against git@github.com:kubernetes/kubernetes.git:v1.16.8
Cloning git@github.com:kubernetes/kubernetes.git into /tmp/ca-update-vendor.HPnf/kubernetes
Extra dependency github.com/google/go-querystring already used by k8s in >= version v1.0.0
Extra dependencies found in one of go.mod-extra files
Fix errors and rerun script:
 ./hack/update-vendor.sh -d/tmp/ca-update-vendor.HPnf -fgit@github.com:kubernetes/kubernetes.git -rv1.16.8
```

As we change to use 1.16.0-beta2, go-querystring needs to be imported separately.

2. Another commit is to fix broken `update-vendor.sh`, it has invalid command
https://github.com/kubernetes/autoscaler/blob/66383d0c3b2725075cfbb6625e4fb37636f87a5e/cluster-autoscaler/hack/update-vendor.sh#L188

https://github.com/kubernetes/autoscaler/blob/66383d0c3b2725075cfbb6625e4fb37636f87a5e/cluster-autoscaler/hack/update-vendor.sh#L192

Check difference between 1.16 and 1.17
https://www.diffchecker.com/ziUDo49S

3. I use ` ./hack/update-vendor.sh -rv1.16.8` to update dependencies.


Jeffwan:
@mwielgus @jaypipes 

I used latest 1.16.x release version and that's the reason it brings many unrelated changes. To minimize the changes, I change back to previous version 1.16.0-beta2. PR only has AWS changes now.

offzale:
Many thanks for the work, looking forward to this to get support on regional STS endpoints and IAM Roles for Service Accounts with the SDK update

DzmitrySudnik:
When do you guys plan to release 1.16 with that fix in?

Issif:
1.17 is coming for EKS, I guess the same issue will happen.

PR #3009, opened 2020-04-01
Fix Prometheus CPU query

povilasv:
Issue https://github.com/kubernetes/autoscaler/issues/2508

Previously we did `rate(container_cpu_usage_seconds_total{%s}[8d])`, which returns you a per-second average rate of increase, which would be a single number per container. And I believe it isn't very useful as averaging out over 8days, won't account for for spikes.

The fix uses Prometheus subqueries (https://prometheus.io/blog/2019/01/28/subquery-support/), which calculates a same rate over 5 minutes over last 8 days with default step ( in my case it was 1min)

So result of executing that query for one container is:
```
0.08135529095444262 @1585684200
0.04520008871193183 @1585684260
0.04030128808668913 @1585684320
0.038564872259585437 @1585684380
0.037179121655268745 @1585684440
0.001468363844192067 @1585684500
0.0015794555333950588 @1585684560
```

Which is similiar to what memory query  (`container_memory_working_set_bytes{%s}[8d]`) returns:

```
1028096 @1585674210.87
995328 @1585674228.877
995328 @1585674259.042
995328 @1585674282.943
995328 @1585674310.786
995328 @1585674337.616
995328 @1585674367.744
995328 @1585674395.019
995328 @1585674429.124
995328 @1585674460.024
```

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
@@ -86,8 +86,10 @@ func (p *prometheusHistoryProvider) getContainerIDFromLabels(labels map[string]s
 	return &model.ContainerID{
 		PodID: model.PodID{
 			Namespace: namespace,
-			PodName:   podName},
-		ContainerName: containerName}, nil
+			PodName:   podName,
+		},
+		ContainerName: containerName,
+	}, nil
 }
 
 func (p *prometheusHistoryProvider) getPodIDFromLabels(labels map[string]string) (*model.PodID, error) {
@@ -131,7 +133,8 @@ func getContainerUsageSamplesFromSamples(samples []Sample, resource model.Resour
 		res = append(res, model.ContainerUsageSample{
 			MeasureStart: sample.Timestamp,
 			Usage:        resourceAmountFromValue(sample.Value, resource),
-			Resource:     resource})
+			Resource:     resource,
+		})
 	}
 	return res
 }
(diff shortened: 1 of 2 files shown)


vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
```diff
@@ -190,7 +193,7 @@ func (p *prometheusHistoryProvider) GetClusterHistory() (map[model.PodID]*PodHis
 	podSelector := fmt.Sprintf("job=\"%s\", %s=~\".+\", %s!=\"POD\", %s!=\"\"",
 		p.config.CadvisorMetricsJobName, p.config.CtrPodNameLabel,
 		p.config.CtrNameLabel, p.config.CtrNameLabel)
-	err := p.readResourceHistory(res, fmt.Sprintf("rate(container_cpu_usage_seconds_total{%s}[%s])", podSelector, p.config.HistoryLength), model.ResourceCPU)
+	err := p.readResourceHistory(res, fmt.Sprintf("rate(container_cpu_usage_seconds_total{%s}[5m])[%s:]", podSelector, p.config.HistoryLength), model.ResourceCPU)
```
bskiba, line 196: Can you add a comment for people not familiar with Prometheus (like me) what this query does? Otherwise LGTM

povilasv: Done, let me know if this reads ok :)

bskiba:
Thanks!
/lgtm
/approve

PR #3016, opened 2020-04-02
Rollback OOM tests

jbartosik:
Tests I added in #2952 are failing when running on prow because pods are stuck pending. Roll back the change

PR #3017, opened 2020-04-02
Squashed commits reverting #3016

jbartosik:
This reverts #3016 to add back OOM tests.

Squashed to make history shorter. The problem was with permission on the image that I use in the test. I verified that the problem is fixed by running the test on a cluster in a different project.

Messages of the squashed commits:

Revert "Revert "Add test for handling OOMing containers""

This reverts commit 644aa7a12cda8a7bc271c3c6716efcde3d8022de.

Revert "Revert "Add docekrfile for OOMing pods under VPA test""

This reverts commit 68c12d30eb8246894ee106eadbe7bb6dac1520d5.

Revert "Revert "Tweak OOM test""

This reverts commit c48f7f846c50824422fee6ef25834cdffdb203e3.

Revert "Revert "Enable tests for pods under VPA""

This reverts commit e3c38fa639179d622de1780e65530abb6e701820.


jbartosik:
I will merge this as as part of #3020

PR #3021, opened 2020-04-03
Run as non-root for kubernetes on VPA

Lujeni:
Thanks for your works !

Error with a rootless PSP 
```
  Type     Reason     Age                    From                  Message
  ----     ------     ----                   ----                  -------
  Normal   Scheduled  6m43s                  default-scheduler     Successfully assigned kube-system/vpa-recommender-654ddbd47b-xsmhw to k8s-xxx
  Normal   Pulled     5m17s (x8 over 6m40s)  kubelet, k8s-xxx  Successfully pulled image "k8s.gcr.io/vpa-recommender:0.6.3"
  Warning  Failed     5m17s (x8 over 6m40s)  kubelet, k8s-xxx  Error: container has runAsNonRoot and image will run as root
  Normal   Pulling    101s (x24 over 6m42s)  kubelet, k8s-xxx  Pulling image "k8s.gcr.io/vpa-recommender:0.6.3"
```

### DIFF
--- a/vertical-pod-autoscaler/deploy/admission-controller-deployment.yaml
+++ b/vertical-pod-autoscaler/deploy/admission-controller-deployment.yaml
@@ -15,6 +15,9 @@ spec:
         app: vpa-admission-controller
     spec:
       serviceAccountName: vpa-admission-controller
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534 # nobody
       containers:
         - name: admission-controller
           image: k8s.gcr.io/vpa-admission-controller:0.6.3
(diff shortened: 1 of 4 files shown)


bskiba:
With version 0.7.0 we have moved to distroless (https://github.com/GoogleContainerTools/distroless) which should solve the problem as well.
I will be moving the default version soon (so the hack/vpa-up.sh script uses 0.7.0). Would this be enough?

yashbhutwala:
I think if vpa used `gcr.io/distroless/static-debian10:nonroot` that would fix it, but currently on master it's running `gcr.io/distroless/static:latest`

bskiba:
I see. In that case
/lgtm
/approve

PR #3023, opened 2020-04-05
Ensure VMSS is not under updating before scaling out

### LINKED ISSUE #3018: CA should check the VMSS provisioning state before scaling up
Because of various reasons, the VM provisioning may need a long period and hence VMSS scaling requests would time out.

On such cases, the following scaling requests should check the provisioning state and only continue to scale after it has been finished. If the VMSS is still under updating, then CA should log an error.

/kind bug
/area provider/azure

feiskyer:
Because of various reasons, the VM provisioning may need a long period and hence VMSS scaling requests would time out.

On such cases, the following scaling requests should check the provisioning state and only continue to scale after it has been finished. If the VMSS is still under updating, then CA should log an error and abort the current scaling step.

Fix #3018 

/kind bug
/area provider/azure

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
@@ -40,6 +40,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
+	"github.com/Azure/go-autorest/autorest/to"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -253,27 +254,35 @@ func (scaleSet *ScaleSet) updateVMSSCapacity(future *azure.Future) {
...
 	}
 
+	// Abort scaling to avoid concurrent VMSS scaling if the VMSS is still under updating.
+	// Note that the VMSS provisioning state would be updated per scaleSet.sizeRefreshPeriod.
+	if vmssInfo.VirtualMachineScaleSetProperties != nil && strings.EqualFold(to.String(vmssInfo.VirtualMachineScaleSetProperties.ProvisioningState), string(compute.ProvisioningStateUpdating)) {
+		klog.Errorf("VMSS %q is still under updating, waiting for it finishes before scaling", scaleSet.Name)
+		return fmt.Errorf("VMSS %q is still under updating", scaleSet.Name)
+	}
+
+	// Proactively set the VMSS size so autoscaler makes better decisions.
```
marwanad, line 278: not a regression from this PR (master and 1.18) but we should move this to happen under `CreateOrUpdateAsync`, we shouldn't proactively update the size before sending the request.

What if `CreateOrUpdateAsync` fails because of a quota error, you'd have to wait for the next size refresh to clear.

feiskyer: good catch, this should be be moved after `CreateOrUpdateAsync()`, so it fails then the cache still the same as before

Issue #2543, opened 2019-11-19
vpa-recommender: Prometheus history provider doesn't work with large Prometheus/Kubernetes deployments


milesbxf:
Whilst working on a fix for #2508 I ran into another issue whilst testing the VPA recommender with the Prometheus history provider. On startup the history provider queries Prometheus for historical container CPU usage over 8 days. We run a fairly large cluster and this results in a non trivial amount of cardinality and large amount of data, which Prometheus sensibly rejects due to the query returning too much data:

```
1172 cluster_feeder.go:209] Cannot get cluster history: cannot get usage history: cannot get timeseries for cpu: retrying GetTimeseries unsuccessful: tried 10 times, last error: bad HTTP status: 400 Bad Request. Response text: {"status":"error","errorType":"bad_data","error":"exceeded maximum resolution of 11,000 points per timeseries. Try decreasing the query resolution (?step=XX)"}
```

Even if this limit wasn't present, a query of this magnitude can easily OOMKill a large enough Prometheus instance.

**What can we do to fix it?**

Given we already retrieve a list of pods/labels using Prometheus, we can use that to split up and shard the queries to retrieve and load a subset of data at a time. Alternatively, we could change the history provider so that instead of all data being retrieved in one go on startup, we could dynamically load historical data for each VPA object we observe added to the cluster. 

Separately, I think we should also switch to using the Prometheus client library rather than writing HTTP calls and marshalling to custom types. This will let us build up valid queries programmatically (rather than `fmt.Sprintf`) and use the built in Prometheus types. This isn't strictly related to the above, so I'm happy to raise a separate issue/PR if you think this would be a good idea.

What do you think? cc @bskiba

milesbxf:
An easier alternative I've just thought of is to make the step resolution time configurable, e.g. going from 30s resolution to 1h resolution. This will impact the accuracy of the history, but be a lot cheaper to run

bskiba:
Definitely support the idea to use Prometheus client libtrary.

For the cluster history being to big.
Sharding queries seems to be a reasonable idea.
Downsampling seems a bit risky, especially with memory but I guess downsampled data is better than no data at all.
You might also consider shortening the history horizon instead of downsampling - this will be better for jobs that have daily patterns. However this will still need to be a cluster-wide setting rather than per-workload one.
  
The idea to
> dynamically load historical data for each VPA object we observe added to the cluster.
seems interesting, might be a bit harder to implement though. 
[...]

PR #3030, opened 2020-04-07
VPA: Fix recommender's Prometheus CPU history

zuzzas:
There are some WIP pull requests that are quite large in scope. What if we use a subquery syntax to fix the issue with historical CPU metrics?

https://github.com/kubernetes/autoscaler/pull/2549
https://github.com/kubernetes/autoscaler/pull/2551
https://github.com/kubernetes/autoscaler/issues/2543

fixes https://github.com/kubernetes/autoscaler/issues/2508

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
@@ -18,6 +18,7 @@ package history
 
 import (
 	"fmt"
+	"math"
 	"net/http"
 	"sort"
 	"strings"
@@ -34,6 +35,7 @@ type PrometheusHistoryProviderConfig struct {
 	PodNamespaceLabel, PodNameLabel                    string
 	CtrNamespaceLabel, CtrPodNameLabel, CtrNameLabel   string
 	CadvisorMetricsJobName                             string
+	MetricsFetchInterval                               time.Duration
 }
 
 // PodHistory represents history of usage and labels for a given pod.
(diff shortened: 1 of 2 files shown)


povilasv:
looks like same pr as https://github.com/kubernetes/autoscaler/pull/3009 would be great if either one could get some review

zuzzas:
@povilasv 
My apologies, I thought I've all outstanding PRs before submitting mine. I'll leave mine just in case, because the bug is really "bugging" us. :)

vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
```diff
@@ -190,7 +192,7 @@ func (p *prometheusHistoryProvider) GetClusterHistory() (map[model.PodID]*PodHis
 	podSelector := fmt.Sprintf("job=\"%s\", %s=~\".+\", %s!=\"POD\", %s!=\"\"",
 		p.config.CadvisorMetricsJobName, p.config.CtrPodNameLabel,
 		p.config.CtrNameLabel, p.config.CtrNameLabel)
-	err := p.readResourceHistory(res, fmt.Sprintf("rate(container_cpu_usage_seconds_total{%s}[%s])", podSelector, p.config.HistoryLength), model.ResourceCPU)
+	err := p.readResourceHistory(res, fmt.Sprintf("rate(container_cpu_usage_seconds_total{%s}[%.fs])[%s:]", podSelector, math.Round(4*p.config.MetricsFetchInterval.Seconds()), p.config.HistoryLength), model.ResourceCPU)
```
bskiba, line 195: Why do we multiply the fetch interval by 4 exactly?

bskiba:
Actually I think we can move forward with #3009, is it OK if I close this one?

zuzzas:
@bskiba 
Aye, thanks for helping with this!

PR #3032, opened 2020-04-08
Add ControlledResources to VPA v1 API

tweks:
This field will allow to specify resources for which VPA generates recommendations.

### DIFF
--- a/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go
+++ b/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go
@@ -147,6 +147,11 @@ type ContainerResourcePolicy struct {
 	// for the container. The default is no maximum.
 	// +optional
 	MaxAllowed v1.ResourceList `json:"maxAllowed,omitempty" protobuf:"bytes,4,rep,name=maxAllowed,casttype=ResourceList,castkey=ResourceName"`
+
+	// Specifies the type of recommendations that will be computed
+	// (and possibly applied) by VPA.
+	// If not specified, the default of [ResourceCPU, ResourceMemory] will be used.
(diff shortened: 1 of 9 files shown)


krzysied:
nit: The title for the CL is "Add ControlledResources to VPA v1 API". For some reason it also contains fix for codegen script and adding hamster example.
This should be separate PRs - w don't want to revert codegen script fix when api doesn't work :)

vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go
```diff
@@ -147,6 +147,11 @@ type ContainerResourcePolicy struct {
 	// for the container. The default is no maximum.
 	// +optional
 	MaxAllowed v1.ResourceList `json:"maxAllowed,omitempty" protobuf:"bytes,4,rep,name=maxAllowed,casttype=ResourceList,castkey=ResourceName"`
+
+	// Specifies the type of recommendations that will be computed
+	// (and possibly applied) by VPA.
+	// If not specified, the default of [ResourceCPU, ResourceMemory] will be used.
+	ControlledResources *[]v1.ResourceName `json:"controlledResources,omitempty" patchStrategy:"merge" protobuf:"bytes,5,rep,name=controlledResources"`
```
krzysied, line 154: Why do we need a pointer to an array instead of just an array?

tweks: This is because we are adding a new field into API with non-zero default.

Specifically, to distinguish between those two situations:
- no value has been set (therefore Controller should interpret it as a default),
- field has been explicitly set to empty value.

krzysied: Ack.

PR #3036, opened 2020-04-10
Proactively decrement scale set count during deletion operations 

### LINKED ISSUE #3031: Deletion requests and stale cache can cause scale set to go below configured minimum
In some cases, where there's two deletion candidates ready for deletion in close proximity to one another, and there's a deletion request in flight, the `getVMSSInfo` would return a stale VMSS size and a deletion request would still happen for the second.

We should forceRefresh the caches after sending a delete request so as to block further requests that could possibly go below the min size.

/area provider/azure
/kind bug

marwanad:
Fixes #3031 

This is to avoid stale cache reads which might deem a node deletable and cause the scale-down operation to below minimum node count.

/area provider/azure
/kind bug

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
@@ -236,9 +236,7 @@ func (scaleSet *ScaleSet) updateVMSSCapacity(future *azure.Future) {
 		if err != nil {
 			klog.Errorf("Failed to update the capacity for vmss %s with error %v, invalidate the cache so as to get the real size from API", scaleSet.Name, err)
 			// Invalidate the VMSS size cache in order to fetch the size from the API.
-			scaleSet.sizeMutex.Lock()
-			defer scaleSet.sizeMutex.Unlock()
-			scaleSet.lastSizeRefresh = time.Now().Add(-1 * scaleSet.sizeRefreshPeriod)
+			scaleSet.invalidateStatusCacheWithLock()
 		}
 	}()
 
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -700,3 +709,13 @@ func (scaleSet *ScaleSet) invalidateInstanceCache() {
 	scaleSet.lastInstanceRefresh = time.Now().Add(-1 * vmssInstancesRefreshPeriod)
 	scaleSet.instanceMutex.Unlock()
 }
+
+func (scaleSet *ScaleSet) invalidateSizeCacheWithLock() {
+	scaleSet.sizeMutex.Lock()
+	scaleSet.lastSizeRefresh = time.Now().Add(-1 * scaleSet.sizeRefreshPeriod)
+	scaleSet.sizeMutex.Unlock()
+
+	scaleSetStatusCache.mutex.Lock()
```
marwanad, line 718: Pointing this diff out:

I don't think the initial intention was working as expected, from the comment:

```
// Invalidate the VMSS size cache in order to fetch the size from the API.
```

However, there's another safeguard in `getVMSSInfo` that would also read data from cache if `scaleSet.sizeRefreshPeriod` hasn't passed from the last `scaleSetStatusCache` refresh.

feiskyer: Right, this is not caught in our unit tests.  Could you help to add a unit test for the cache refreshing?

marwanad: yup, i'll look into adding one.

feiskyer:
@marwanad could you also cherry-pick the fix to older branches?

PR #3040, opened 2020-04-12
Ignore AWS NodeWithImpairedVolumes taint

johanneswuerbach:
The `NodeWithImpairedVolumes` taint is applied to node on AWS when a volume is stuck in attaching state for too long.

The taint was introduced in k8s a while ago https://github.com/kubernetes/kubernetes/pull/55558

Add it to the ignored node condition taints.

### DIFF
--- a/cluster-autoscaler/utils/taints/taints.go
+++ b/cluster-autoscaler/utils/taints/taints.go
@@ -17,11 +17,11 @@ limitations under the License.
 package taints
 
 import (
-	"k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
 	"strings"
 
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/deletetaint"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
 	cloudproviderapi "k8s.io/cloud-provider/api"
 
 	"k8s.io/klog"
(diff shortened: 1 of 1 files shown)


johanneswuerbach:
Sorry, maybe @Jeffwan would be a better assignee as this is AWS specific.

Jeffwan:
@johanneswuerbach 

Code changes look good to me

In your case, what's the root cause for the volume takes that long to attach?

Jeffwan:
@johanneswuerbach  Code changes look good to me

/lgtm

In your case, what's the root cause for the volume takes that long to attach?

johanneswuerbach:
Not sure, it looks like an issue in AWS, but it might also be k8s related. We currently run on 1.16 provisioned using kops and I requested a back port of a fix, which might be related https://github.com/kubernetes/kubernetes/pull/89894

Jeffwan:
@johanneswuerbach em, good to know, I will ask my team member to track this change and try to get it approved

johanneswuerbach:
@Jeffwan any update on this?

mwielgus:
@Jeffwan ping :)

Jeffwan:
/lgtm

needs someone to approve this change.  @mwielgus

edsonmarquezani:
I've seen a side effect of it.

I have an instance group whose applications are shut down at a certain scheduled. When there's a node tainted like this, this specific node won't be terminated and will be left as the only one for that instance group. The day after, when PODs are up again, the autoscaler WON'T scale nodes up because of this taint, showing a message like this.

```
I0622 12:35:50.236346 1 utils.go:196] Pod wololo-67ccdcb765-xd7dq can't be scheduled
on spot-nodes, predicate failed: PodToleratesNodeTaints predicate mismatch, cannot put
forno/wololo-67ccdcb765-xd7dq on template-node-for-spot-nodes-176367186976534907,
reason: node(s) had taints that the pod didn't tolerate
```

If I scale one node manually on AWS or terminate the tainted node, things get back to work again.

johanneswuerbach:
That is actually what the PR is supposed to solve and solved for us. 

As a workaround adding `--ignore-taint=NodeWithImpairedVolumes` should fix what you see above.

edsonmarquezani:
@joshbranham Thanks for the advice!

johanneswuerbach:
@mwielgus @Jeffwan anything missing here?

PR #3051, opened 2020-04-14
Fix eventRecorder throttling to include log lines

towca:
Previously, eventRecorder would throttle the actual events, but not
the information about them logged via klog.

### DIFF
--- a/cluster-autoscaler/utils/kubernetes/event_sink_logging_wrapper.go
+++ b/cluster-autoscaler/utils/kubernetes/event_sink_logging_wrapper.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kubernetes
+
+import (
+	clientv1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/klog"
+)
+
+type eventSinkLoggingWrapper struct {
+	actualSink record.EventSink
+}
+
+// Create wraps EventSink's Create().
+func (s eventSinkLoggingWrapper) Create(event *clientv1.Event) (*clientv1.Event, error) {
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/utils/kubernetes/factory.go
```diff
@@ -41,9 +39,13 @@ const (
 // CreateEventRecorder creates an event recorder to send custom events to Kubernetes to be recorded for targeted Kubernetes objects
 func CreateEventRecorder(kubeClient clientset.Interface) kube_record.EventRecorder {
 	eventBroadcaster := kube_record.NewBroadcasterWithCorrelatorOptions(getCorrelationOptions())
-	eventBroadcaster.StartLogging(klog.V(4).Infof)
 	if _, isfake := kubeClient.(*fake.Clientset); !isfake {
-		eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: v1core.New(kubeClient.CoreV1().RESTClient()).Events("")})
+		actualSink := &v1core.EventSinkImpl{Interface: v1core.New(kubeClient.CoreV1().RESTClient()).Events("")}
+		// EventBroadcaster has a StartLogging() method but the throttling options from getCorrelationOptions() get applied only to
```
losipiuk, line 44: Is there an issue created to fix the root cause? 
Can you add a `//TODO` here and reference the issue?

cluster-autoscaler/utils/kubernetes/factory.go
```diff
@@ -41,9 +39,13 @@ const (
...
+		// actual sinks, which makes it throttle the actual events, but not the corresponding log lines. As a workaround, event logging
```
losipiuk, line 45: I guess the problem here is performance-wise, right? Can you state that explicitly?

cluster-autoscaler/utils/kubernetes/event_sink_logging_wrapper.go
```diff
@@ -0,0 +1,54 @@
...
+
+package kubernetes
+
+import (
+	clientv1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/klog"
+)
+
+type eventSinkLoggingWrapper struct {
```
losipiuk, line 25: Normally I would suggest adding a test - but testing klog integration may be painfull. As you wish.

cluster-autoscaler/utils/kubernetes/factory.go
towca: Done.

cluster-autoscaler/utils/kubernetes/factory.go
towca: Added!

towca:
@losipiuk Comments addressed.

losipiuk:
/lgtm
/approve
Thanks! :)

PR #3052, opened 2020-04-14
Update aws readme for IAM roles & sts regional endpoints

maust:
1. with https://github.com/kubernetes/autoscaler/issues/2829 regional STS endpoints can be used since
https://github.com/kubernetes/autoscaler/releases/tag/cluster-autoscaler-1.15.6
https://github.com/kubernetes/autoscaler/releases/tag/cluster-autoscaler-1.18.1

2. recommend attaching IAM roles directly to the cluster-autoscaler pod instead of the node-group

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/README.md
+++ b/cluster-autoscaler/cloudprovider/aws/README.md
@@ -36,8 +36,6 @@ If you'd like the cluster autoscaler to [automatically
 discover](#auto-discovery-setup) EC2 AutoScalingGroups, the
 `autoscaling:DescribeTags` permission is also required.
 
-To attach the IAM role with the policy to the pod see 
-
 **NOTE**: You can restrict the target resources for the autoscaling actions by
 specifying autoscaling group ARNS. More information can be found
 [here](https://docs.aws.amazon.com/autoscaling/latest/userguide/control-access-using-iam.html#policy-auto-scaling-resources).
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -5,9 +5,9 @@ The cluster autoscaler on AWS scales worker nodes within any specified autoscali
 Cluster autoscaler must run on v1.3.0 or greater.
 
 ## Permissions
-The worker running the cluster autoscaler will need access to certain resources and actions. We always recommend you to attach IAM policy to nodegroup and avoid using AWS credentials directly unless you have special requirements.
+The pod running the cluster autoscaler will need access to certain resources and actions. We always recommend you to attach the IAM policy to the cluster austoscaler pod using [IAM roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) and avoid attaching it to the NodeGroup or using AWS credentials directly unless you have special requirements.
```
Jeffwan, line 8: Can we address that, recommend to use IRSA if user is using EKS? We have other self-managed AWS users as well.

Jeffwan:
@maust You need to sign CLA as well.

cluster-autoscaler/cloudprovider/aws/README.md
maust: Sure, what about the improved text?

Jeffwan: I think it's good. Thanks for the contribution

PR #3063, opened 2020-04-17
Allow max size to be zero when fetching from VMSS tags

oscar60310:
The max size should allow to be zero when min size is zero too.


marwanad:
Thanks for the PR :)

I'm curious what's the use-case for allowing a max size of 0? Min size is 0 in order to support scaling to/from zero, but a non-positive max doesn't make sense to have.

oscar60310:
We have multiple vm scale sets with different size and our service enable/disable them dynamically. Setting max size to zero can make sure a disabled scale set not be scaled up again after down to zero.

In AWS, we can simply set the max size of the [auto scaling groups](https://docs.aws.amazon.com/autoscaling/ec2/userguide/AutoScalingGroup.html) to zero, but changing tags on vmss seems to be the only way in Azure.

PR #3065, opened 2020-04-18
fix vpa yamls

yashbhutwala:
- [X] use vpa v1 crd
- [X] use `apply` instead of `create`
- [X] remove `poc.autoscaling.k8s.io` group permissions from rbac
- [X] organize rbac into groupings per recommender, updater and admission controller

### DIFF
--- a/vertical-pod-autoscaler/deploy/recommender-deployment.yaml
+++ b/vertical-pod-autoscaler/deploy/recommender-deployment.yaml
@@ -1,10 +1,4 @@
 ---
-apiVersion: v1
-kind: ServiceAccount
-metadata:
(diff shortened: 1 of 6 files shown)


bskiba:
As nice as a cleanup is, I am wary of making these many changes to the rbac file, this does introduce a lot of risk of missing something in the rewrite.

mwielgus:
Thank you for your contribution. 
Please split this PR into multiple smaller PRs that are easier to evaluate and review. In the past we've had quite a few production problems caused by incorrect rbacs, permissions and subtle config changes so people are afraid of breaking things up and are reluctant to review them. I'm closing this one.

PR #3073, opened 2020-04-20
digitalocean: fix min size bug

waynr:
Previously we were rejecting decreases down to the min size due to a logic error -- use of `<=` where we should have been using `<`.

### DIFF
--- a/cluster-autoscaler/cloudprovider/digitalocean/digitalocean_node_group.go
+++ b/cluster-autoscaler/cloudprovider/digitalocean/digitalocean_node_group.go
@@ -144,7 +144,7 @@ func (n *NodeGroup) DecreaseTargetSize(delta int) error {
 	}
 
 	targetSize := n.nodePool.Count + delta
-	if targetSize <= n.MinSize() {
+	if targetSize < n.MinSize() {
 		return fmt.Errorf("size decrease is too small. current: %d desired: %d min: %d",
 			n.nodePool.Count, targetSize, n.MinSize())
 	}
--- a/cluster-autoscaler/cloudprovider/digitalocean/digitalocean_node_group_test.go
+++ b/cluster-autoscaler/cloudprovider/digitalocean/digitalocean_node_group_test.go
@@ -182,6 +182,34 @@ func TestNodeGroup_DecreaseTargetSize(t *testing.T) {
 		assert.NoError(t, err)
 	})
 
+	t.Run("successful decrease to minimum", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


waynr:
meow

MorrisLaw:
/lgtm

We just need approval now.

cluster-autoscaler/cloudprovider/digitalocean/digitalocean_node_group.go
```diff
@@ -144,7 +144,7 @@ func (n *NodeGroup) DecreaseTargetSize(delta int) error {
 	}
 
 	targetSize := n.nodePool.Count + delta
-	if targetSize <= n.MinSize() {
+	if targetSize < n.MinSize() {
```
andrewsykim, line 147: For a future PR, I wonder if the target size should be adjusted to the min size if it goes below instead of returning error.

MorrisLaw: Like this @andrewsykim ?

```
if targetSize < n.MinSize() {
    targetSize = n.MinSize()
}
```

MorrisLaw: Maybe a log message too

andrewsykim: Yeah, I'm not sure what the other providers do but that seems more correct then returning an error here.

MorrisLaw: Talked to @waynr a bit. Maybe the error is relied upon somewhere else within the autoscaler and removing it may lead to undesired effects 🤔 

Not sure though. Maybe worth exploring.

waynr: Whoops, meant to respond to this earlier. Returning an error when we can't decrease by the requested amount seems to me like the most intuitive behavior. That being said, this is my first contribution to autoscaler and I only have a passing familiarity with the principles of its operation so maybe it is okay for functions like this to only perform partial/"best effort" mutations on the cluster.

andrewsykim: > Maybe the error is relied upon somewhere else within the autoscaler and removing it may lead to undesired effects thinking

Yeah I don't have the wholistic picture of this but I think worth investigating.

PR #3080, opened 2020-04-22
Add api-approved label to the Vertical Pod Autoscaler API

### LINKED ISSUE #3069: vpa and vpacheckpoint CRDs have KubernetesAPIApprovalPolicyConformant condition from MissingAnnotation
The CRDs for `vpa` and `vpacheckpoint` are missing annotations referring to their API review, which is causing the CRDs to have the condition `KubernetesAPIApprovalPolicyConformant` set to `false` with a reason of `MissingAnnotation`.

```
$ kubectl get crd verticalpodautoscalers.autoscaling.k8s.io -o yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  ..............8<...snip...8<........................
spec:
  ..............8<...snip...8<........................
status:
  acceptedNames:
    ..............8<...snip...8<........................
  conditions:
  - lastTransitionTime: "2020-04-16T06:49:00Z"
    message: protected groups must have approval annotation "api-approved.kubernetes.io",
      see https://github.com/kubernetes/enhancements/pull/1111
    reason: MissingAnnotation
    status: "False"
    type: KubernetesAPIApprovalPolicyConformant
```

You'll get similar output from `kubectl get crd verticalpodautoscalercheckpoints.autoscaling.k8s.io`

According to https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190612-crd-group-protection.md (which was created by the https://github.com/kubernetes/enhancements/pull/1111 PR referenced in the condition) an API review is required for these CRDs.

I spoke with @deads2k who said that if we have already done an API review, we just need to add the annotation to the CRD metadata with a link to the PR where the review was performed, but if we haven't, the API would need to get an API review and approval from the API team. Another option is to move the CRDs to groups to `x-8s.io`.

bskiba:
Resources are part of verticalpodautoscalers.autoscaling.k8s.io
- VerticalPodAutoscaler is an api resource for configuring vertical autoscaling for pods.
- VerticalPodAutoscalerCheckpoint is an API resource used internally by the Vertical Pod Autoscaler to store its additional state.

The API was discussed and approved in https://github.com/kubernetes/kubernetes/pull/63797
The design itself was discussed and approved in https://github.com/kubernetes/community/pull/338

Fixes #3069

### DIFF
--- a/vertical-pod-autoscaler/deploy/vpa-v1-crd.yaml
+++ b/vertical-pod-autoscaler/deploy/vpa-v1-crd.yaml
@@ -3,6 +3,8 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: verticalpodautoscalers.autoscaling.k8s.io
+  annotations:
+    "api-approved.kubernetes.io": "https://github.com/kubernetes/kubernetes/pull/63797"
 spec:
   group: autoscaling.k8s.io
   scope: Namespaced
@@ -66,6 +68,8 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: verticalpodautoscalercheckpoints.autoscaling.k8s.io
+  annotations:
+    "api-approved.kubernetes.io": "https://github.com/kubernetes/kubernetes/pull/63797"
 spec:
   group: autoscaling.k8s.io
   scope: Namespaced


bskiba:
@mwielgus @gjtempleton Can one of you approve as SIG-Autoscaling leads?

bskiba:
Getting following conditions on the CRD from this PR
```
 Conditions:
    Last Transition Time:  2020-04-22T15:36:55Z
    Message:               approved in https://github.com/kubernetes/kubernetes/pull/63797
    Reason:                ApprovedAnnotation
    Status:                True
```

gjtempleton:
This looks good to me. Won't work with prow, but /approve

bskiba:
Thanks @gjtempleton and @mwielgus

PR #3083, opened 2020-04-23
Update vendor & fix breaking scheduler changes

towca:
The following things changed in scheduler and needed to be fixed:
* NodeInfo was moved to schedulerframework
* Some fields on NodeInfo are now exposed directly instead of via getters
* NodeInfo.Pods is now a list of *schedulerframework.PodInfo, not *apiv1.Pod
* SharedLister and NodeInfoLister were moved to schedulerframework
* PodLister was removed


cluster-autoscaler/simulator/delta_cluster_snapshot.go
```diff
@@ -51,24 +49,24 @@ type deltaSnapshotPodLister DeltaClusterSnapshot
...
-	modifiedNodeInfoMap map[string]*schedulernodeinfo.NodeInfo
+	addedNodeInfoMap    map[string]*schedulerframework.NodeInfo
+	modifiedNodeInfoMap map[string]*schedulerframework.NodeInfo
 	deletedNodeInfos    map[string]bool
 
-	nodeInfoList         []*schedulernodeinfo.NodeInfo
+	nodeInfoList         []*schedulerframework.NodeInfo
 	podList              []*apiv1.Pod
-	havePodsWithAffinity []*schedulernodeinfo.NodeInfo
+	havePodsWithAffinity []*schedulerframework.NodeInfo
```
MaciekPytel, line 58: Is podList still used for anything after removing the pod lister functionality?

towca: No, good catch - fixed.

cluster-autoscaler/simulator/delta_cluster_snapshot.go
```diff
@@ -168,20 +164,11 @@ func (data *internalDeltaSnapshotData) addNodeInfo(nodeInfo *schedulernodeinfo.N
 		data.nodeInfoList = append(data.nodeInfoList, nodeInfo)
 	}
 
-	if len(nodeInfo.Pods()) > 0 {
-		data.clearPodCaches()
-	}
```
MaciekPytel, line 173: I think we still need to drop havePodsWithAffinity here.

MaciekPytel: cc: @aleksandra-malinowska

towca: Ohh right, that's my bad!

MaciekPytel:
lgtm modulo the cache invalidation comment. Thanks for doing this!

Also a minor general comment for future PRs - if trivial import changes happened in separate comment it would be easier to focus review on non-trivial parts.

PR #3094, opened 2020-04-24
Provider/Packet: Allow empty nodepool, Add support for multiple nodepools, Add price support

v-pap:
This PR adds several new features for the Packet provider in the cluster autoscaler.

1. Empty nodepool
2. Multiple nodepools
3. Price support

 Also, this PR makes an older PR for Packet redundant


cluster-autoscaler/cloudprovider/packet/packet_price_model.go
```diff
@@ -0,0 +1,75 @@
...
+	minutes := math.Ceil(float64(endTime.Sub(startTime)) / float64(time.Minute))
+	hours := minutes / 60.0
+	return hours
+}
+
+// PodPrice returns a theoretical minimum price of running a pod for a given
+// period of time on a perfectly matching machine.
+func (model *PacketPriceModel) PodPrice(pod *apiv1.Pod, startTime time.Time, endTime time.Time) (float64, error) {
+	// Always return 1.0 since Packet is only charging per Node.
+	return 1.0, nil
```
MaciekPytel, line 74: The is not about charging for pod. The price expander uses a ration between theoretical cost of resources used by a pod and the actual cost of infrastructure. 
This method is supposed to estimate cost of cpu, memory, etc used by the pod. In other words how much a custom-shaped VM that would have exactly the resources (cpu, memory, gpu, ...) used by the pod would cost if one was available.

MaciekPytel:
A general comment - with the growing number of cloudprovider implementations core developers aren't able to keep up with details of each one anymore. I'm happy to approve this change, but I don't have context to properly review it.
For every (?) other cloudprovider, we now have maintainers who have the right to approve changes to that cloudprovider and who take responsibility for reviewing provider-specific changes. I guess we're looking for a maintainer for packet too - do you know if there is some packet community we could approach about this?

cluster-autoscaler/cloudprovider/packet/packet_price_model.go
v-pap: Thank you for your explanation, I updated this function. As far as I understand, this function is used in the expander's heuristic. Since, there is no native support for custom VMs, or a pricing scheme that is per CPU/RAM usage in Packet (it is a bare metal provider after all), I calculated the constants (`cpuPricePerHour`, `memoryPricePerHourPerGb`) as the theoretical best value you can get per hour from the available instances. If you think that there is a better approach to this, please let me know.

v-pap:
> A general comment - with the growing number of cloudprovider implementations core developers aren't able to keep up with details of each one anymore. I'm happy to approve this change, but I don't have context to properly review it.
> For every (?) other cloudprovider, we now have maintainers who have the right to approve changes to that cloudprovider and who take responsibility for reviewing provider-specific changes. I guess we're looking for a maintainer for packet too - do you know if there is some packet community we could approach about this?

Packet is aware of this and is actively looking at it.

cluster-autoscaler/cloudprovider/packet/packet_price_model.go
MaciekPytel: I believe @mwielgus did the same for GCE - there are custom VMs there, but the pricing is not the same. It doesn't need to be perfect it's just used when trying to estimate how effective the packing of pods is wrt cost of underlying resources.

v-pap: Ok, I think I have done the same thing for Packet with the only difference being that I follow Packet's pricing model. Do you think that there is any other issue that needs to be addressed?

v-pap:
assign @feiskyer

MaciekPytel:
I have zero experience using packet (and I don't believe any other maintainer has any) and I don't really have the context to review this. I can approve it for permissions, but I'd like a review from someone with knowledge of packet. 
Ideally we'd need an owner for packet, we can't really maintain it without someone with context involved.

detiber:
@vielmetti can you recommend appropriate reviewers/approvers for the cluster-autoscaler packet provider?

thebsdbox:
@MaciekPytel I'm coming up to speed (at Packet) but I'm happy to also give this a review

vielmetti:
Suggest that @deitch might want to take a look from the Packet point of view on this.

deitch:
Happy to take a look as well.

I haven't dug into the autoscaler architecture quite as deeply, but I do have a general architecture question. Can providers be broken out into separate implementations in their own repos, each implementing an agreed API with a pluggable core autoscaler, the way cluster-api is? That allows each repo to have its own set of OWNERS to approve PRs, even as it is under `github.com/kubernetes-sigs/`. It would be even nicer if we could use the same providers as CAPI uses, but I might be asking a wee bit too much there. :-)

In the meantime, to the PR.

cluster-autoscaler/cloudprovider/packet/examples/cluster-autoscaler-secret.yaml
```diff
@@ -15,15 +15,30 @@ metadata:
...
...MCBrdWJlYWRtPTEuMTcuNC0wMCBrdWJlY3RsPTEuMTcuNC0wMAphcHQtbWFyayBob2xkIGt1YmVsZXQga3ViZWFkbSBrdWJlY3RsCmN1cmwgLWZzU0wgaHR0cHM6Ly9kb3dubG9hZC5kb2NrZXIuY29tL2xpbnV4L3VidW50dS9ncGcgfCBhcHQta2V5IGFkZCAtCmFkZC1hcHQtcmVwb3NpdG9yeSAiZGViIFthcmNoPWFtZDY0XSBodHRwczovL2Rvd25sb2FkLmRvY2tlci5jb20vbGludXgvdWJ1bnR1IGJpb25pYyBzdGFibGUiCmFwdCB1cGRhdGUKYXB0IGluc3RhbGwgLXkgZG9ja2VyLWNlPTE4LjA2LjJ+Y2V+My0wfnVidW50dQpjYXQgPiAvZXRjL2RvY2tlci9kYWVtb24uanNvbiA8PEVPRgp7CiAgImV4ZWMtb3B0cyI6IFsibmF0aXZlLmNncm91cGRyaXZlcj1zeXN0ZW1kIl0sCiAgImxvZy1kcml2ZXIiOiAianNvbi1maWxlIiwKICAibG9nLW9wdHMiOiB7CiAgICAibWF4LXNpemUiOiAiMTAwbSIKICB9LAogICJzdG9yYWdlLWRyaXZlciI6ICJvdmVybGF5MiIKfQpFT0YKbWtkaXIgLXAgL2V0Yy9zeXN0ZW1kL3N5c3RlbS9kb2NrZXIuc2VydmljZS5kCnN5c3RlbWN0bCBkYWVtb24tcmVsb2FkCnN5c3RlbWN0bCByZXN0YXJ0IGRvY2tlcgpzd2Fwb2ZmIC1hCm12IC9ldGMvZnN0YWIgL2V0Yy9mc3RhYi5vbGQgJiYgZ3JlcCAtdiBzd2FwIC9ldGMvZnN0YWIub2xkID4gL2V0Yy9mc3RhYgpjYXQgPDxFT0YgfCB0ZWUgL2V0Yy9kZWZhdWx0L2t1YmVsZXQKS1VCRUxFVF9FWFRSQV9BUkdTPS0tY2xvdWQtcHJvdmlkZXI9ZXh0ZXJuYWwKRU9GCmt1YmVhZG0gam9pbiAtLWRpc2NvdmVyeS10b2tlbi11bnNhZmUtc2tpcC1jYS12ZXJpZmljYXRpb24gLS10b2tlbiB7ey5Cb290c3RyYXBUb2tlbklEfX0ue3suQm9vdHN0cmFwVG9rZW5TZWNyZXR9fSB7ey5BUElTZXJ2ZXJFbmRwb2ludH19Cg==
```
deitch, line 30: How was this `cloudInit` generated? I always can run it through `base64 -D`, but it would be good if this were available in the source to inspect, and make it easier to change.

cluster-autoscaler/cloudprovider/packet/packet_cloud_provider.go
```diff
@@ -164,9 +175,7 @@ func BuildPacket(opts config.AutoscalingOptions, do cloudprovider.NodeGroupDisco
 		klog.Fatalf("Must specify at least one node group with --nodes=<min>:<max>:<name>,...")
 	}
 
-	if len(do.NodeGroupSpecs) > 1 {
-		klog.Fatalf("Packet autoscaler only supports a single nodegroup for now")
-	}
+	validNodepoolName := regexp.MustCompile(`^[a-z0-9A-Z]+[a-z0-9A-Z\-\.\_]*[a-z0-9A-Z]+$|^[a-z0-9A-Z]{1}$`)
```
deitch, line 178: I have no argument with validating the nodepool name, but why is this inside the packet provider? Is this not handled at a higher level?

cluster-autoscaler/cloudprovider/packet/packet_manager_rest.go
```diff
@@ -27,17 +27,140 @@ import (
...
+	GPU          int64
+}
+
+var (
+	maxHttpRetries int           = 3
+	httpRetryDelay time.Duration = 100 * time.Millisecond
+)
+
+// InstanceTypes is a map of packet resources
+var InstanceTypes = map[string]*instanceType{
```
deitch, line 58: I would be very hesitant to hard-code all of these in. The Packet API should provide a way to retrieve a list of all known device types (and if not, I will make sure it gets in).

cluster-autoscaler/cloudprovider/packet/packet_manager_rest.go
```diff
@@ -27,17 +27,140 @@ import (
 	"math/rand"
 	"net/http"
 	"os"
+	"strings"
```
deitch, line 30: This whole file seems quite duplicative. There is an [official packet go sdk](https://godoc.org/github.com/packethost/packngo). All of the work being done here duplicates it and requires maintenance. Why is this large file here instead of just leveraging packngo?

cluster-autoscaler/cloudprovider/packet/packet_manager_rest.go
deitch: Sure, here it is in [go sdk](https://godoc.org/github.com/packethost/packngo#Plan) and in [rest API](https://www.packet.com/developers/api/plans/)

cluster-autoscaler/cloudprovider/packet/README.md
```diff
@@ -53,3 +54,40 @@ This prevents the node that the autoscaler is running on from being scaled down.
 If you are deploying the autoscaler into a cluster which already has more than one node,
 it is best to deploy it onto any node which already has non-default kube-system pods,
 to minimise the number of nodes which cannot be removed when scaling.
+
+### Changes
```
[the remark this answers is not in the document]
v-pap: I don't mind this change, but this was requested by @enkelprifti98 and @jacobsmith928

cluster-autoscaler/cloudprovider/packet/README.md
```diff
@@ -53,3 +54,40 @@ This prevents the node that the autoscaler is running on from being scaled down.
...
+      - matchExpressions:
+        - key: beta.kubernetes.io/instance-type
+          operator: In
+          values:
+          - t1.small.x86
+```
+
+4. Cloud inits in the examples have pinned versions for Kubernetes in order to minimize potential incompatibilities as a result of nodes provisioned with different Kubernetes versions.
+
+5. In the provided cluster-autoscaler deployment example, we target the master node of the cluster. This change allows the cluster-autoscaler to stick to the master node and thus allowing the cluster-autoscaler to scale down all the worker nodes reliably to zero. Without this change there was a possibility for the cluster-autoscaler to be deployed in a worker node that could not be downscaled. In order to benefit from this change you have to label your master node like this: `kubectl label node <master_node> dedicated=master`
```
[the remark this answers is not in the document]
v-pap: Fixed

cluster-autoscaler/cloudprovider/packet/examples/cluster-autoscaler-deployment.yaml
```diff
@@ -149,13 +166,19 @@ spec:
...
+          # --nodes=0:10:pool2
           command:
             - ./cluster-autoscaler
             - --alsologtostderr
             - --cluster-name=cluster1
             - --cloud-config=/config/cloud-config
             - --cloud-provider=packet
-            - --nodes=1:10:pool1
+            - --expander=price
+            - --nodes=0:10:pool1
```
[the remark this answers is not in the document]
v-pap: Fixed

cluster-autoscaler/cloudprovider/packet/examples/cluster-autoscaler-secret.yaml
v-pap: To be honest, I don't really understand how this would be possible inside this yaml file. Could you provide a way on how to do that? Also keep in mind that this is only an example.

cluster-autoscaler/cloudprovider/packet/packet_cloud_provider.go
v-pap: AFAIK this is not handed at a higher level.

cluster-autoscaler/cloudprovider/packet/packet_manager_rest.go
v-pap: Indeed the Packet API could provide a way to do that, but this data is requested a lot. Hard coded values regarding cloud instances is a common practice with other providers as well (https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/ec2_instance_types.go, https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/alicloud/alicloud_instance_types.go, https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/azure/azure_instance_types.go). Another problem of the Packet API is that it doesn't provide core counts yet, which are needed for this case.

cluster-autoscaler/cloudprovider/packet/packet_manager_rest.go
v-pap: As for the packngo package I would be glad to use it, but as it has been mentioned in the first PR of the cluster autoscaler (https://github.com/kubernetes/autoscaler/pull/2101#issuecomment-499698555), this project doesn't allow for external dependencies other than those already in use within the project.

cluster-autoscaler/cloudprovider/packet/README.md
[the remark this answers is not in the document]
deitch: I agree very much; a changelog is a good thing. You want people to find it when looking for changes - in one of a changelog file, at the very end of a doc, or in the PR/release comments - but the basic functionality needs to be described in the doc.

cluster-autoscaler/cloudprovider/packet/README.md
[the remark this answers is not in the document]
deitch: Thanks.

cluster-autoscaler/cloudprovider/packet/examples/cluster-autoscaler-deployment.yaml
```diff
@@ -128,6 +131,20 @@ spec:
       labels:
         app: cluster-autoscaler
     spec:
+      # Node affinity is used to force cluster-autoscaler to stick
+      # to the master node. This allows the cluster to reliably downscale
+      # to zero worker nodes when needed. In order for this example to work
+      # you have to add a label on the master node with key: dedicated
+      # and value: master
```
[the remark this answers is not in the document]
deitch: :-)

cluster-autoscaler/cloudprovider/packet/packet_cloud_provider.go
deitch: I guess it needs to stay here then.

cluster-autoscaler/cloudprovider/packet/packet_manager_rest.go
deitch: > this data is requested a lot

Then we should cache it, even for hours or a day. We don't want to have to cut a new autoscaler release every time packet adds a new node type, or retires one, or bumps the memory on an offering, or any little change.

> Another problem of the Packet API is that it doesn't provide core counts yet, which are needed for this case

We do not? No, we do not, nor do we do the frequency. That is a good reason not to do it. I will open an issue on this for Packet.

cluster-autoscaler/cloudprovider/packet/packet_manager_rest.go
deitch: I missed that. I guess you are stuck then. Seems kind of sad, but so be it. Thanks for clarifying.

detiber: Might be good to bring that requirement back up at the next sig-autoscaling meeting, it seems like it introduces arbitrary complications, especially as support for cloud providers in k/k is in the process of being removed in favor of external cloud providers.

deitch: I won't argue with that @detiber :-)

Is autoscaler going into a per-cloud-provider pluggable format, like the core cloud providers are doing? Or could we even have autoscaler somehow talk to the ccm itself?

detiber: There are discussions around adopting a plugin-type model for cluster-autoscaler, such as https://github.com/kubernetes/autoscaler/pull/3140, but I don't know that there is consensus in that area currently.

I don't know that there could (or should) be direct integration with CCM, since CCM is generally scoped to a much smaller functional subset of cloud provider support than is used by cluster-autoscaler and adding support for scaling primitives needed for cluster-autoscaler seems a bit excessive for those components

deitch: Got it. Overall, a nice plugin model would be great for this, if we could make it manageable. Keep all of this functionality in separate repos, and the core `kubernetes/autoscaler` repo is clean. Since cluster-api v1alpha3 switched to work this way, it has been great, much easier to maintain.

cluster-autoscaler/cloudprovider/packet/README.md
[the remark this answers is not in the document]
v-pap: Done

cluster-autoscaler/cloudprovider/packet/examples/cluster-autoscaler-secret.yaml
deitch: As you point out, it is a yaml file, not code, so it cannot be generated. But how is this example intended to be used? Do people actually hand-craft the yaml and then handcraft the cloudinit and run it through base64? Or is this more of a template than an example, and the `cloudinit` field is largely static?

I notice that it is tied to a specific os and plan and facility, etc. So this really is a pretty narrow example.

In any case, this isn't changed in the PR, so no reason to hold it up for this. How customers use it, templates, generation, etc. really should be addressed more thoroughly, both in the `packet/README.md` and its tooling, in a follow-up

deitch:
I don't see anything holding this up from a Packet side, but take that as a grain of salt, as I am not really familiar with the background to this or its structure (my comments on using packngo vs raw REST API show how much I didn't know).

There is what to fix up here - the examples make it very difficult to use it for anything other than the example case, and need tooling and documentation to support users, maybe leverage installers and providers - but those can wait for a future PR.

mwielgus:
Inactivity warning. Please proceed on this PR or it will be closed.

PR #3105, opened 2020-04-30
[CA-1.18] #2934 cherry-pick: Fixes 2932: let the capi version to be discovered

xmudrii:
This PR proposes to cherry-pick #2934 to the `cluster-autoscaler-release-1.18` branch.

With Cluster Autoscaler v1.18.0, it's possible to use the Cluster-API provider. However, there is a bug that makes the provider unusable for some users.

The provider supports all Cluster-API groups and versions, but in the original PR, the group and version are hardcoded. Users that are not using v1alpha2 have two options, which are not optimal:
* Wait for Cluster Autoscaler v1.19.0, which will come up with K8s 1.19.0. The problem is that it's somewhere in August, so that could be a bit long to wait
* Build the image which includes the fix manually, which is not the greatest approach.

The PR (#2934) that fixes this could be looked at as a bugfix and it doesn't introduce any behavior changes (the default group is still `cluster.x-k8s.io` and version will be auto-discovered). I believe that it would make sense to cherry-pick it to the release-1.18 branch, so the bugfix is included in the next patch release. That would make the provider useful for all users, regardless of the group and version they are using.

Original PR description:
```
…API version.
This change adds detection for an environment variable to specify the group for the clusterapi resources. If the environment
variable `CAPI_GROUP` is specified, then it will
be used instead of the default.
This also decouples the API group from the version and let the latter to be discovered dynamically.
```

cc @elmiko

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
+++ b/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
@@ -19,12 +19,14 @@ package clusterapi
 import (
 	"context"
 	"fmt"
+	"os"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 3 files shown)


elmiko:
i think this is a good commit to cherry-pick, i'd like to get @enxebre 's thoughts also.

/approve

detiber:
+1 from me

enxebre:
I'm all good with this. But It should be validated by some autoscaler core owner for the release branching management and back ports policy.

MaciekPytel:
/approve
Based on PR description this seems like a reasonable fix to cherry-pick - CA not working seems like a good justification. I'll leave lgtm-ing to someone who is actually familiar with CAPI code.

PR #3114, opened 2020-05-05
cluster-autoscaler: Allow Azure Rate Limit Defaults to be set from the environment

marc-sensenich:
If you're using environment variables to configure the Cluster Autoscaler, you're unable to set the Azure cloud provider's rate limit QPS or bucket size and must switch your configuration to file based to do so. This PR enables the setting of 4 environment variables which allow for the rate limit QPS or bucket size to be configured without needing to switch to file based configuration 

Signed-off-by: Marc Sensenich <sensenichm91@gmail.com>

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_manager.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_manager.go
@@ -148,24 +148,63 @@ type Config struct {
 }
 
 // InitializeCloudProviderRateLimitConfig initializes rate limit configs.
-func InitializeCloudProviderRateLimitConfig(config *CloudProviderRateLimitConfig) {
+func InitializeCloudProviderRateLimitConfig(config *CloudProviderRateLimitConfig) error {
 	if config == nil {
-		return
+		return nil
 	}
 
 	// Assign read rate limit defaults if no configuration was passed in.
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/cloudprovider/azure/azure_manager.go
```diff
@@ -148,31 +148,72 @@ type Config struct {
 }
 
 // InitializeCloudProviderRateLimitConfig initializes rate limit configs.
-func InitializeCloudProviderRateLimitConfig(config *CloudProviderRateLimitConfig) {
+func InitializeCloudProviderRateLimitConfig(config *CloudProviderRateLimitConfig) error {
```
feiskyer, line 151: Thanks for adding this. Would you like to add some unit tests for the changes?

marc-sensenich: @feiskyer I've added some unit tests in https://github.com/kubernetes/autoscaler/pull/3114/commits/1d6f18ff3cb577b0c1f476c59359d75d20324e25

feiskyer: thanks a lot

PR #3119, opened 2020-05-07
Add prices for all machine types, adjust e2 fractional VM pricing

brett-elliott:
This change

* Adds prices for all machine types.

* Adjusts pricing for e2-micro and e2-small to be greater than e2-medium. See comment for more details.


MaciekPytel:
Thanks! It seems like it may also be worth cherry-picking into patch releases. WDYT?
/lgtm
/approve

One minor comment: maybe squash to one commit? It seems like a simple enough change and it will make cherry-picking easier.
/hold

PR #3127, opened 2020-05-12
Proposal for decoupling ClusterAutoScaler from CloudProvider implemen…

bhargav-naik:
Enabling external CloudProvider to inject their implementations in clusterAutoscaler.

### DIFF
--- a/cluster-autoscaler/proposals/cp_agnostic_ca.md
+++ b/cluster-autoscaler/proposals/cp_agnostic_ca.md
@@ -9,101 +9,150 @@ In this proposal, a different design approach is presented which will decouple t
 ### Issues
 - In the current design the ClusterAutoscaler core has dependency on the CloudProvider implementations.
 - For proprietary CloudProvider implementations it results in an development cycle to fork new version from public repo and port the CloudProvider implementation to the latest fork.
-- Also the ClusterAutoscaler library comes with a lot of unnecessary CloudProvider specific dependencies which might not be required for other CloudProviders.
+- Also the ClusterAutoscaler library comes with a lot of unnecessary CloudProvider specific dependencies which might not be required for other CloudProviders. (We dont want to address this)
 
 ### Solution
+Version 1:  (Not acceptable, as we don't want to disrupt the existing dev/build cycle)
 - Release ClusterAutoscalerCore as a separate module which won't have any Cloudprovider specific code/dependencies.
 - Release CloudProvider specific ClusterAutoscaler[PROVIDER] as separate module. This module depends on specific version of ClusterAutoscaler.
 - For proprietary CloudProvider implementations can depend on ClusterAutoscalerCore module and wont get any other CloudProvider specific dependencies.
 - Also the development cycle for CloudProvider implementation reduces to just a version bump of ClusterAutoscalerCore if the CloudProvider,Nodegroup interfaces are intact and CloudProvider doesn't intend to add any feature.
 
+Version 2:
+- Let the existing cloudProvider code remain in the current hierarchy.
+- Introduce external_cloud_provider.go under cloudprovider.external package: A shim layer which will give external_cloud_provider mechanism to register their implementation.
+
 ### Implementation
-- Introduce a Registration mechanism in ClusterAutoScalerCore which the CloudProvider can use to register the implementation at init.
 ```go
-package cloudprovider
+package external
 
-type CloudProviderFactory func(opts config.AutoscalingOptions, do NodeGroupDiscoveryOptions, rl *ResourceLimiter) CloudProvider
+import (
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
+	"k8s.io/autoscaler/cluster-autoscaler/config"
+	"k8s.io/klog"
+	"sync"
+)
 
-// All registered cloud providers.
+type CloudProviderFactory func(opts config.AutoscalingOptions, do cloudprovider.NodeGroupDiscoveryOptions, rl *cloudprovider.ResourceLimiter) cloudprovider.CloudProvider
+
 var (
 	mutex                         sync.Mutex
 	cloudProviderFactory          CloudProviderFactory
-	podListProcessor              pods.PodListProcessor
-	nodeGroupListProcessor        nodegroups.NodeGroupListProcessor
-	nodeGroupSetProcessor 		  nodegroupset.NodeGroupSetProcessor
-	scaleUpStatusProcessor 		  status.ScaleUpStatusProcessor
-	scaleDownNodeProcessor 		  nodes.ScaleDownNodeProcessor
-	scaleDownStatusProcessor 	  status.ScaleDownStatusProcessor
-	autoscalingStatusProcessor    status.AutoscalingStatusProcessor
-	nodeGroupManager 			  nodegroups.NodeGroupManager
-	nodeInfoProcessor 			  nodeinfos.NodeInfoProcessor
 )
 
-func RegisterCloudProviderFactory(cloudProviderFactory CloudProviderFactory) {
+func RegisterExternalCloudProvider(cpf CloudProviderFactory) {
 	mutex.Lock()
 	defer mutex.Unlock()
 	if cloudProviderFactory != nil {
-		klog.Fatalf("Cloud provider %q was registered twice", name)
+		klog.Fatalf("Cloud provider %q was registered twice")
 	}
-	cloudProviderFactory = cloudProviderFactory
+	cloudProviderFactory = cpf
 }
 
-func GetCloudProvider(name string, opts config.AutoscalingOptions, do NodeGroupDiscoveryOptions, rl *ResourceLimiter) CloudProvider {
-	cloudProvidersMutex.Lock()
-	defer cloudProvidersMutex.Unlock()
+func BuildExternalCloudProvider(opts config.AutoscalingOptions, do cloudprovider.NodeGroupDiscoveryOptions, rl *cloudprovider.ResourceLimiter) cloudprovider.CloudProvider {
+	mutex.Lock()
+	defer mutex.Unlock()
 	if cloudProviderFactory != nil {
-	    return cloudProviderFactory(opts, do, rl)
+		klog.Fatalf("external cloudProvider not registered")
+	}
+	return cloudProviderFactory(opts, do, rl)
+}
+```
+
+- Modify buidler_all.go
+```go
+// +build  !external !gce,!aws,!azure,!kubemark,!alicloud,!magnum,!digitalocean,!clusterapi
+package builder
+
+import (
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/alicloud"
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/aws"
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/azure"
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/baiducloud"
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/clusterapi"
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/digitalocean"
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/external"
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/gce"
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/magnum"
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/packet"
+	"k8s.io/autoscaler/cluster-autoscaler/config"
+)
+
+// AvailableCloudProviders supported by the cloud provider builder.
+var AvailableCloudProviders = []string{
+	cloudprovider.AwsProviderName,
+	cloudprovider.AzureProviderName,
+	cloudprovider.GceProviderName,
+	cloudprovider.AlicloudProviderName,
+	cloudprovider.BaiducloudProviderName,
+	cloudprovider.MagnumProviderName,
+	cloudprovider.DigitalOceanProviderName,
+	cloudprovider.ExternalCloudProviderName,
+	clusterapi.ProviderName,
+}
+
+// DefaultCloudProvider is GCE.
+const DefaultCloudProvider = cloudprovider.GceProviderName
+
+func buildCloudProvider(opts config.AutoscalingOptions, do cloudprovider.NodeGroupDiscoveryOptions, rl *cloudprovider.ResourceLimiter) cloudprovider.CloudProvider {
+	switch opts.CloudProviderName {
+	case cloudprovider.GceProviderName:
+		return gce.BuildGCE(opts, do, rl)
+	case cloudprovider.AwsProviderName:
+		return aws.BuildAWS(opts, do, rl)
+	case cloudprovider.AzureProviderName:
+		return azure.BuildAzure(opts, do, rl)
+	case cloudprovider.AlicloudProviderName:
+		return alicloud.BuildAlicloud(opts, do, rl)
+	case cloudprovider.BaiducloudProviderName:
+		return baiducloud.BuildBaiducloud(opts, do, rl)
+	case cloudprovider.DigitalOceanProviderName:
+		return digitalocean.BuildDigitalOcean(opts, do, rl)
+	case cloudprovider.MagnumProviderName:
+		return magnum.BuildMagnum(opts, do, rl)
+	case packet.ProviderName:
+		return packet.BuildPacket(opts, do, rl)
+	case clusterapi.ProviderName:
+		return clusterapi.BuildClusterAPI(opts, do, rl)
+	case cloudprovider.ExternalCloudProviderName:
+		return external.BuildExternalCloudProvider(opts, do, rl)
 	}
 	return nil
 }
-
-func RegisterPodListProcessor(podListProcessor pods.PodListProcessor) {
-	mutex.Lock()
-	defer mutex.Unlock()
-	podListProcessor = podListProcessor
-}
-
-func GetPodListProcessor() pods.PodListProcessor {
-	mutex.Lock()
-	defer mutex.Unlock()
-	return podListProcessor
-}
-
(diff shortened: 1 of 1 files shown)


MaciekPytel:
I think there are two major problems that need to be solved before this can happen:
- CA dependency management means importing CA (which would be required by cloudprovider implementations) is broken. We vendor fake commits, CA 1.17 uses a commit that is not even on k8s 1.17 branch, etc.
- The intended pattern is to allow implementation of advanced provider-specific or use-case specific implementation by using processors (https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/processors) to execute custom code in various critical places of CA logic. With processors provider-core integration goes far beyond just cloudprovider interface.
This isn't really covered by current version of proposal.

bhargav-naik:
Inline replies:
"CA dependency management means importing CA (which would be required by cloudprovider implementations) is broken. We vendor fake commits, CA 1.17 uses a commit that is not even on k8s 1.17 branch, etc."
- Can you please explain this further. CA can use vendoring and still release the CACore with all the required k8s code and move the CloudProvider out of the dependency.

"The intended pattern is to allow implementation of advanced provider-specific or use-case specific implementation by using processors (https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/processors) to execute custom code in various critical places of CA logic. With processors provider-core integration goes far beyond just cloudprovider interface."
- If I understood this correctly, you are pointing to setting up nodeInfoComparator which happens in the main.go file.
```
	nodeInfoComparatorBuilder := nodegroupset.CreateGenericNodeInfoComparator
	if autoscalingOptions.CloudProviderName == cloudprovider.AzureProviderName {
		nodeInfoComparatorBuilder = nodegroupset.CreateAzureNodeInfoComparator
	} else if autoscalingOptions.CloudProviderName == cloudprovider.AwsProviderName {
		nodeInfoComparatorBuilder = nodegroupset.CreateAwsNodeInfoComparator
	}
```
This can also by done in a similar pattern as described in the original proposal, by letting the Cloudprovider register specific processor implementation in init block and use default in case no specific implementation is provided.

MaciekPytel:
Regarding vendoring part - it's the other way round: cloudprovider implementation likely needs to import code from CA 'core' (not necessarily core/ directory in current codebase, but things that would be part of 'CA core' module). For example I did a quick grep in GCE and AWS providers and both import various CA files from outside of cloudprovider directory.
Unless we fix current dependency management it would get very painful to maintain any provider that want's to do that.

On the processors part - nodegroupset processor is the simplest case. The direction we took with customization / extensibility of CA is to create programming hooks that would allow provider-specific (or use-case specific) arbitrary code execution with almost unrestricted access to modify internal data structures, etc. The intent is to allow mixing in provider logic with core logic as much as possible and expose all sorts of internal state to providers to allow any specialized extensions that go far beyond what you can go with cloudprovider interface.
Given that this is something we've supported for a long time and we've been advertising tightly coupling provider and core logic in processors as the way to address various use-cases, I'd like to make sure we don't break this. Which would mean we'd need to make sure we explicitly expose every part of core datastructures that can currently be accessed and we fix dependency management to make it reasonably easy to have code that vendors both core and cloudprovider.

bhargav-naik:
@MaciekPytel , Thanks for your input. I had follow up questions on the design goal.

Does the following line items covers the requirements from extensibility perspective :
- CloudProvider should be able to register specific CloudProvider implementation
- CloudProvider should be able to override any AutoscalingProcessors at bootup time
- Only one CloudProvider will be needed in one runtime

Please give direction if more things needs to be covered.

MaciekPytel:
> Does the following line items covers the requirements from extensibility perspective

On a conceptual level maybe. But it doesn't really solve the problem of how to easily manage dependencies between core and cloudprovider. And, since the cloudprovider interface would need to stay in core, every implementation would need to import the core. 
And I expect a ton of problems for advanced processor use-cases where cyclic dependencies involving core, provider and processor interfaces are already an issue.
 
Maybe a good compromise would be to instead implement a 'shim' provider that acts as an adapter to implementation provided in a different module? Basically do what you're proposing, but instead of refactoring CA to separate provider from core, you'd just have a provider that handles the registration and forwards all calls to the implementation in separate module. You could compile CA with just the shim provider using the existing mechanism and get the same experience without affecting existing implementations. WDYT?

bhargav-naik:
I have updated the proposal to address the extensibility concern.
CloudProvider can override any AutoScaling Processor which will  be injected at the boot-up time.

MaciekPytel:
I don't think adding hook for registering processors addresses my worries. You call for removing existing provider-specific logic, but you don't say exactly how to replace it without breaking existing providers (and I don't think it's trivial). If your design aims to make such a large refactor to CA I'd expect it to tackle all the resulting problems.

https://github.com/kubernetes/autoscaler/pull/3140 proposes a different solution to the same problem and I think it has many benefits over this one:
- It doesn't require any significant refactors and doesn't affect existing providers.
- Cloudprovider interface changes almost every release anyway. Most of those changes are trivial, just changing import paths for NodeInfo or similar, but it's enough for any implementation to stop compiling. gRPC offers a much more stable API that has a more realistic chance of remaining the same across multiple versions.

bhargav-naik:
In-line replies:
_"I don't think adding hook for registering processors addresses my worries. You call for removing existing provider-specific logic, but you don't say exactly how to replace it without breaking existing providers (and I don't think it's trivial). If your design aims to make such a large refactor to CA I'd expect it to tackle all the resulting problems."_
- What kind of issues do we perceive if cloudProvider specific code is moved out to their own package?  Wouldn't the issue be caught in compile time and be addressed.

_"Cloudprovider interface changes almost every release anyway. Most of those changes are trivial, just changing import paths for NodeInfo or similar, but it's enough for any implementation to stop compiling. gRPC offers a much more stable API that has a more realistic chance of remaining the same across multiple versions."_
- Will the gRPC interface not change if the Cloudprovider interface changes in any release?
- Will the CloudProviders not need to go through the similar process everytime the interface changes?

Some observation about gRPC based approach:
- Is it a design goal to give flexibility to let cloudProvider to use any language/framework of choice for their implementation?
- Also in the gRPC based approach, only cloudProvider interface is externalized. If someone wants to override the processor they will still have to maintain their fork and add override the processor.
- Why do we want to go for a solve which will increase runtime complexity instead of making changes in compile time workflow ?

MaciekPytel:
Seems like you make a lot of good points about gRPC - maybe it would be worth to bring them up on the other proposal. 

> What kind of issues do we perceive if cloudProvider specific code is moved out to their own package? Wouldn't the issue be caught in compile time and be addressed.

Main issue is dependency management - it is really hard to import CA core. Especially if you also need to import other stuff from Kubernetes and you also need to bump the version of Kubernetes transitive dependency (which has been the case for AWS/Azure/GCP clients that were needed in CA in newer version than in Kubernetes). Keeping it in one package makes it slightly easier as you don't have to deal with fake commits in CA dependencies.

What about my idea to instead do your proposal via a wrapper cloudprovider that handles the registration and proxies any calls to separate package? That also solves the problems for existing providers, because there is no change to core.
I'm very worried about refactors to core, but as long as we can encapsulate new features there is little reason not to do it.

> Will the gRPC interface not change if the Cloudprovider interface changes in any release?

I think it's actually likely to change less. Part of existing golang interface are types coming from upstream kubernetes (Node, Taint, NodeInfo, Quantity). Those are moved around a lot and every other release there is a trivial change in import path of something. No change in behavior and so no need to update gRPC interface, but enough to require trivial fixes to most implementations.

bhargav-naik:
@MaciekPytel , Thank you for your inputs. I will try to spend some more cycle to evaluate the direction that you have provided _"What about my idea to instead do your proposal via a wrapper cloudprovider that handles the registration and proxies any calls to separate package?"_

bhargav-naik:
- Sincerely speaking, I would want to be in a state where not cloudProvider specific dependencies come in ClusterAutoscaler go.mod. 
- I always thought that taking cloudProvider implementation out of core would help achieve that.

MaciekPytel:
> Sincerely speaking, I would want to be in a state where not cloudProvider specific dependencies come in ClusterAutoscaler go.mod.
 
I think that's already mostly true? We request all providers to keep their dependencies (particularly client libraries) inside provider directory, except if they're already vendored by kubernetes/kubernetes (which is the case for AWS, Azure and GCE).

bhargav-naik:
@MaciekPytel , I have updated the proposal based on your inputs.

New proposal:
- Doesn't change the build/release workflow for existing cloudproviders
- Introduced external_cloud_provider which handles the registration/build flow external_cp
- No extra complexity introduced like communication through gRPC, running extra process

cluster-autoscaler/proposals/cp_agnostic_ca.md
```diff
@@ -0,0 +1,158 @@
+# Decoupling ClusterAutoScalerCore (separate library) from CloudProvider implementations
+##### Author: bhargav-naik
+
+### Introduction
+Current ClusterAutoscaler is released as a bundle which includes core and all the CloudProviders included.
+When someone intends to write implementation of the CloudProvider they need to fork the intended version ClusterAutoscaler and provide their implementation.
+In this proposal, a different design approach is presented which will decouple the ClusterAutoscaler core from CloudProviders and make ClusterAutoScalerCore available as a separate release.
```
hectorj2f, line 7: I don't think this is true anymore based on Version 2

bhargav-naik: @hectorj2f , thanks for the input. I will clean up the sections.

cluster-autoscaler/proposals/cp_agnostic_ca.md
```diff
@@ -0,0 +1,158 @@
...
+##### Author: bhargav-naik
+
+### Introduction
+Current ClusterAutoscaler is released as a bundle which includes core and all the CloudProviders included.
+When someone intends to write implementation of the CloudProvider they need to fork the intended version ClusterAutoscaler and provide their implementation.
+In this proposal, a different design approach is presented which will decouple the ClusterAutoscaler core from CloudProviders and make ClusterAutoScalerCore available as a separate release. 
+
+### Issues
+- In the current design the ClusterAutoscaler core has dependency on the CloudProvider implementations.
+- For proprietary CloudProvider implementations it results in an development cycle to fork new version from public repo and port the CloudProvider implementation to the latest fork.
```
hectorj2f, line 11: This is not true as mentioned in the gRPC proposal, you could use the new cluster-api provider

cluster-autoscaler/proposals/cp_agnostic_ca.md
```diff
@@ -0,0 +1,158 @@
...
+		return clusterapi.BuildClusterAPI(opts, do, rl)
+	case cloudprovider.ExternalCloudProviderName:
+		return external.BuildExternalCloudProvider(opts, do, rl)
+	}
+	return nil
+}
+```
+
+- Introduce builder_external.go
+```go
```
hectorj2f, line 126: how would processors work in the future per provider ?

cluster-autoscaler/proposals/cp_agnostic_ca.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Introduction
+Current ClusterAutoscaler is released as a bundle which includes core and all the CloudProviders included.
+When someone intends to write implementation of the CloudProvider they need to fork the intended version ClusterAutoscaler and provide their implementation.
+In this proposal, a different design approach is presented which will decouple the ClusterAutoscaler core from CloudProviders and make ClusterAutoScalerCore available as a separate release. 
+
+### Issues
+- In the current design the ClusterAutoscaler core has dependency on the CloudProvider implementations.
+- For proprietary CloudProvider implementations it results in an development cycle to fork new version from public repo and port the CloudProvider implementation to the latest fork.
+- Also the ClusterAutoscaler library comes with a lot of unnecessary CloudProvider specific dependencies which might not be required for other CloudProviders. (We dont want to address this)
```
hectorj2f, line 12: Why to add something that won't be addressed as part of this proposal.

cluster-autoscaler/proposals/cp_agnostic_ca.md
bhargav-naik: - Agreed, using cluster-api is definitely an option external_cp must consider.
- But CA might not want to enforce such user to go via CAPI path. CA should still have a way to enable external_cp without having them use CAPI.

cluster-autoscaler/proposals/cp_agnostic_ca.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Solution
+Version 1:  (Not acceptable, as we don't want to disrupt the existing dev/build cycle)
```
hectorj2f, line 15: To help the reader, I'd suggest to stick to one option

cluster-autoscaler/proposals/cp_agnostic_ca.md
bhargav-naik: - Sure now we are on same page :). That needs to be addressed. One problem at a time.

cluster-autoscaler/proposals/cp_agnostic_ca.md
```diff
@@ -0,0 +1,158 @@
+# Decoupling ClusterAutoScalerCore (separate library) from CloudProvider implementations
```
hectorj2f, line 1: Please, update the title accordingly to reflect the new approach

cluster-autoscaler/proposals/cp_agnostic_ca.md
bhargav-naik: Sure will do that.

cluster-autoscaler/proposals/cp_agnostic_ca.md
```diff
@@ -0,0 +1,158 @@
...
+		return baiducloud.BuildBaiducloud(opts, do, rl)
+	case cloudprovider.DigitalOceanProviderName:
+		return digitalocean.BuildDigitalOcean(opts, do, rl)
+	case cloudprovider.MagnumProviderName:
+		return magnum.BuildMagnum(opts, do, rl)
+	case packet.ProviderName:
+		return packet.BuildPacket(opts, do, rl)
+	case clusterapi.ProviderName:
+		return clusterapi.BuildClusterAPI(opts, do, rl)
+	case cloudprovider.ExternalCloudProviderName:
```
hectorj2f, line 118: can you elaborate how you plan to test this new provider ?

cluster-autoscaler/proposals/cp_agnostic_ca.md
```diff
@@ -0,0 +1,158 @@
...
+}
+```
+
+- External cloudprovider will register their implementation using init block. (in their own repo)
+```go
+package ecp
+//ecp: ExternalCloudProvider
+import "k8s.io/autoscaler/cluster-autoscaler/cloudprovider/external"
+
+func init() {
```
hectorj2f, line 155: Could you detail how this new custom-autoscaler would bootstrap ? Perhaps the tests should cover this scenario creating an example of an external cloud provider.

cluster-autoscaler/proposals/cp_agnostic_ca.md
bhargav-naik: Testing new provider (private provider) is outside the purview of CA.

cluster-autoscaler/proposals/cp_agnostic_ca.md
bhargav-naik: Done

cluster-autoscaler/proposals/cp_agnostic_ca.md
bhargav-naik: Removed.

cluster-autoscaler/proposals/cp_agnostic_ca.md
bhargav-naik: We will be able to add only test for the registration and wiring flow for external_cp. Actual execution of external_cp should be outside the purview of CA. More inputs are invited on this point.

hectorj2f: As a developer I suggest you add a sample-external-provider for two reasons:
* Test this new CA feature.
* Show to the community how to code a sample-external-provider. You can follow as example what was done in https://github.com/kubernetes/sample-controller.

cluster-autoscaler/proposals/cp_agnostic_ca.md
hectorj2f: Sorry, but I disagree any piece of code in the CA should be covered with tests. You could have a fake external provider.

cluster-autoscaler/proposals/cp_agnostic_ca.md
bhargav-naik: Sure we will try to take input on this if it needs to be addressed.

cluster-autoscaler/proposals/cp_agnostic_ca.md
bhargav-naik: - Agreed to that point of writing a sample dummy external_cp. A sample and readme on ho to add will be helpful.

On testing 2 test can be included:
- failure scenario in bootup when external_cp is not registered
- success scenario when the dummy_external_cp is registered

cluster-autoscaler/proposals/cp_agnostic_ca.md
bhargav-naik: Agreed, There should be testcases which cover the wiring of external_provider.
Whether or not to have a test to check execution of dummy_cp can be discussed.

cluster-autoscaler/proposals/cp_agnostic_ca.md
bhargav-naik: done.

cluster-autoscaler/proposals/cp_agnostic_ca.md
bhargav-naik: done.

cluster-autoscaler/proposals/cp_agnostic_ca.md
bhargav-naik: @hectorj2f, I have updated the current proposal to enable injection of autoscaling_processor in a similar manner handled in original proposal. Please have a look at it and provide inputs.

bhargav-naik:
Please let me know if I need to address anything.

mwielgus:
Inactivity warning. Please proceed on the pr or it will be closed.

mwielgus:
Closing due to inactivity.  Feel free to reopen if needed.

PR #3129, opened 2020-05-12
Add bunch of metrics to monitor VPA Updater impact in a cluster

kgolab:
New metrics to measure how VPA Updater impacts workloads controlled by VPA:
* number of Pods that are considered by the Updater,
* number of Pods that can be evicted,
* number of VPA objects with at least one Pod that can be evicted,
* number of VPA objects with at least one evicted Pod.

Additionally changes the existing metric (number of evicted Pods) by adding a label.

TODO: Manual testing.

### DIFF
--- a/vertical-pod-autoscaler/pkg/updater/logic/updater.go
+++ b/vertical-pod-autoscaler/pkg/updater/logic/updater.go
@@ -177,11 +177,29 @@ func (u *updater) RunOnce(ctx context.Context) {
 	}
 	timer.ObserveStep("AdmissionInit")
 
+	// wrappers for metrics which are computed every loop run
+	controlledPodsCounter := metrics_updater.NewControlledPodsCounter()
+	evictablePodsCounter := metrics_updater.NewEvictablePodsCounter()
+	vpasWithEvictablePodsCounter := metrics_updater.NewVpasWithEvictablePodsCounter()
+	vpasWithEvictedPodsCounter := metrics_updater.NewVpasWithEvictedPodsCounter()
+
+	// using defer to protect against 'return' after evictionRateLimiter.Wait
+	defer controlledPodsCounter.Observe()
+	defer evictablePodsCounter.Observe()
+	defer vpasWithEvictablePodsCounter.Observe()
+	defer vpasWithEvictedPodsCounter.Observe()
+
 	for vpa, livePods := range controlledPods {
+		vpaSize := len(livePods)
+		controlledPodsCounter.Add(vpaSize, vpaSize)
 		evictionLimiter := u.evictionFactory.NewPodsEvictionRestriction(livePods)
 		podsForUpdate := u.getPodsUpdateOrder(filterNonEvictablePods(livePods, evictionLimiter), vpa)
+		evictablePodsCounter.Add(vpaSize, len(podsForUpdate))
 
+		withEvictable := false
+		withEvicted := false
 		for _, pod := range podsForUpdate {
+			withEvictable = true
 			if !evictionLimiter.CanEvict(pod) {
 				continue
 			}
@@ -194,7 +212,17 @@ func (u *updater) RunOnce(ctx context.Context) {
 			evictErr := evictionLimiter.Evict(pod, u.eventRecorder)
 			if evictErr != nil {
 				klog.Warningf("evicting pod %v failed: %v", pod.Name, evictErr)
+			} else {
+				withEvicted = true
+				metrics_updater.AddEvictedPod(vpaSize)
 			}
+		}
+
+		if withEvictable {
+			vpasWithEvictablePodsCounter.Add(vpaSize, 1)
+		}
+		if withEvicted {
+			vpasWithEvictedPodsCounter.Add(vpaSize, 1)
 		}
 	}
 	timer.ObserveStep("EvictPods")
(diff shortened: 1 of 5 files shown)


kgolab:
@bskiba , please have a look.
There is one place which bothers me a little: I moved AddEvictedPod call and now we can lose the information on internal error in podsEvictionRestrictionImpl.Evict. I think it's not something likely to happen but I'd like your opinion there.

vertical-pod-autoscaler/pkg/utils/metrics/updater/updater.go
```diff
@@ -41,16 +85,77 @@ var (
...
+func NewVpasWithEvictedPodsCounter() *SizeBasedGauge {
+	return newSizeBasedGauge(vpasWithEvictedPodsCount)
+}
+
+func getVpaSizeLog2(vpaSize int) int {
+	if vpaSize >= maxVpaSize {
+		return maxVpaSizeLog
+	}
+
+	log2 := 0
```
bskiba, line 136: why not math.Log2?

vertical-pod-autoscaler/pkg/updater/logic/updater.go
```diff
@@ -177,11 +177,29 @@ func (u *updater) RunOnce(ctx context.Context) {
...
+
+	// using defer to protect against 'return' after evictionRateLimiter.Wait
+	defer controlledPodsCounter.Observe()
+	defer evictablePodsCounter.Observe()
+	defer vpasWithEvictablePodsCounter.Observe()
+	defer vpasWithEvictedPodsCounter.Observe()
+
 	for vpa, livePods := range controlledPods {
+		vpaSize := len(livePods)
+		controlledPodsCounter.Add(vpaSize, vpaSize)
```
bskiba, line 194: nit: I think controlledPods may include those with VPA in "off" or "initial", which makes them "controlled by VPA" not necessarily "controlled by updater" (as in description of the metric).

vertical-pod-autoscaler/pkg/utils/metrics/updater/updater.go
```diff
@@ -41,16 +85,77 @@ var (
...
 	return metrics.NewExecutionTimer(functionLatency)
 }
 
-// AddEvictedPod increases the counter of pods evicted by VPA
-func AddEvictedPod() {
-	evictedCount.Add(1)
+// newSizeBasedGauge provides a wrapper for counting items in a loop
+func newSizeBasedGauge(gauge *prometheus.GaugeVec) *SizeBasedGauge {
+	obj := SizeBasedGauge{
+		values: make(map[int]int),
```
bskiba, line 99: could just be an int array, initialized to 0 values

vertical-pod-autoscaler/pkg/utils/metrics/updater/updater.go
```diff
@@ -18,21 +18,65 @@ limitations under the License.
...
 	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/metrics"
 )
 
 const (
 	metricsNamespace = metrics.TopMetricsNamespace + "updater"
+
+	maxVpaSizeLog = 20
+	// maxVpaSize = 2 ^ maxVpaSizeLog
+	maxVpaSize = 1024 * 1024
```
bskiba, line 32: That seems to be a looot for one deployment. Which is a good thing I guess

vertical-pod-autoscaler/pkg/utils/metrics/updater/updater.go
```diff
@@ -41,16 +85,77 @@ var (
...
+	return newSizeBasedGauge(vpasWithEvictedPodsCount)
+}
+
+func getVpaSizeLog2(vpaSize int) int {
+	if vpaSize >= maxVpaSize {
+		return maxVpaSizeLog
+	}
+
+	log2 := 0
+	for vpaSize > 1 {
```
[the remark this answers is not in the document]
kgolab: Ugh, that's my rusty golang. Fixed (and moved to top-level metrics.go as I want to use it in an upcoming PR).

vertical-pod-autoscaler/pkg/updater/logic/updater.go
kgolab: If I understand the code correctly we only take VPAs in auto/recreate mode into account:
1. we filter such VPA objects to create `vpas` variable, lines 130+
2. we use `vpas` to filter controlled Pods, lines 166+

I added a comment for the future readers.

vertical-pod-autoscaler/pkg/utils/metrics/updater/updater.go
kgolab: Done.

vertical-pod-autoscaler/pkg/utils/metrics/updater/updater.go
kgolab: Yes, it's more than max number of Pods in a cluster (5K nodes * 110 pods/node) so I hope we don't have to change it anytime soon.

vertical-pod-autoscaler/pkg/updater/logic/updater.go
bskiba: Thanks, for some reason I thought this was unfiltered

PR #3136, opened 2020-05-14
Update README.md

emacdona:
Docs say "Cluster autoscaler supports four Kubernetes cluster options on Azure", and then goes on to list three options. Am I missing something obvious?


marwanad:
Thanks for the fix. There used to be an option for `ACS` but that support was retired earlier this year.

@emacdona can you sign the CLA?

emacdona:
Sorry, just saw your comment now... I closed my branch b/c I would need to talk to my employer before signing the CLA. I thought it might be easier for you to make the change instead of waiting to merge my "one word" commit :-)

PR #3138, opened 2020-05-14
report MaxNodesTotal count during scale up

elmiko:
This change adds a warning event to signal when a planned scale up
operation would go over the maximum total nodes. This is being proposed
for two primary reasons: as an event that can be watched during
end-to-end testing, and as a signal to users when this condition is
occurring.


elmiko:
cc @enxebre

PR #3140, opened 2020-05-17
doc: proposal custom cloud provider over gRPC

hectorj2f:
As discussed over Slack with @MaciekPytel, I share a proposal to make CA more plugable and allow calling custom cloud providers without having to fork the CA.


cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,127 @@
...
+        ...
+}
+...
+
+func (s *clusterAutoscalerCustomProviderServer) Refresh() error {
+        ...
+}
+...
+
+func (s *clusterAutoscalerCustomProviderServer) GetAvailableMachineTypes() ([]string, error) {
```
MaciekPytel, line 63: I don't think this is needed. We haven't officially deprecated this method, but I don't think CA uses it anymore. 

I've meant to clean up dead methods from cloudprovider interface for a long time, but it requires surprising amount of work and it just keeps sitting in backlog..

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,127 @@
...
+There are two alternatives on how to plug custom cloud providers:
+
+* **Option1:** Install the plugin as a binary that would be mounted into the CA container
+and invoked by CA server.
+In this option CA server launches each provider plugin as a subprocess and communicates with it over gRPC.
+
+* **Option2:** A custom cloud provider server is deployed along side CA and both communicates via gRPC with TLS/SSL.
+
+Regardless of the chosen option, both solutions have to expose a common gRPC API
+with the following operations:
```
MaciekPytel, line 41: You're missing all the NodeGroup methods, which are a critical part of any provider implementation (arguably the largest and most difficult part too).

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,127 @@
+# Plugable Cloud Provider over gRPC
+
+## Motivation
+
+CA is released as a bundle which includes a hardcoded list of supported cloud providers.
+Whenever users want to implement the logic of their own cloud provider, they need to fork the CA and add their own implementation.
```
elmiko, line 6: this is no longer true, a user could write a cluster-api provider without ever having to touch the autoscaler code.

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,127 @@
...
+to reach the remote gRPC server.
+
+A new flag, named
+`--cloud-provider-url=https://local.svc.io/mycloudprovider/server`, determines the URL to reach the custom provider implementation.
+In addition this approach reuses the existing flag that defines the name of the cloud provider `--cloud-provider=mycloudprovider`
+https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider.
+
+To connect the CA core with this new remote cloud provider, this approach needs to implement a new generic cloud provider 
+as part of the CA core code.
+This new provider, named `CustomCloudProvider` simply makes gRPC calls to the remote functions exposed by the custom cloud provider server. In other words, it forwards the calls and handle the errors analogously how done in other existing providers.
```
elmiko, line 114: imo, we should name this `gRPCCloudProvider` or something similar to be more explicit in the name.

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,127 @@
...
+
+Obviously, this new apprach needs to use TLS to ensure a secure communication between CA and this CA provider server.
+The flags need to be defined [TODO].
+
+## User Stories
+
+### Story1
+
+When using CA only a reduced list of cloud providers are supported, if users want to use their own private cloud provider (e.g. Openstack, OpenNebula,...), they need to implement its cloud provider interface for that environment.
+This locked design limits the extensibility of CA and goes against certain native Kubernetes primitives.
```
elmiko, line 124: i like how you described it in the introduction about not needing to fork and merge against the main CA repo, i think you should mention it here as well. i mean, realistically _anyone_ who wants to have their custom provider needs to implement it. the big difference that this provider exposes, imo, is that i can create experimental providers without needing to change the base CA repo.

cluster-autoscaler/proposals/plugable-provider-grpc.md
MaciekPytel: Agreed, but I'm not entirely comfortable having Cluster API as our default recommendation. There are certain problems that come with it's "cattle" approach to nodes differing from CA's design of operating on individual nodes (ex. best effort support for PDBs and annotations, no good intergration story for node upgrade).
Perhaps most importantly Cluster API seems to still be evolving, so it doesn't necessarily fit the goal of minimizing maintenance effort (which seems to be a driving factor here). All in all, I think it makes sense to support both Cluster API and this new provider and let the users choose.

cluster-autoscaler/proposals/plugable-provider-grpc.md
MaciekPytel: +1 for custom being too general. ExternalCloudProvider maybe? Naming stuff is hard :)

cluster-autoscaler/proposals/plugable-provider-grpc.md
elmiko: i agree. my point here was only to say that there are other options. i don't think it's fair to say that the only way for a user to add new providers is by forking the repo. there are other approaches, they are complex and have their own issues, but they do exist.

bhargav-naik:
@hectorj2f 
- I was working on similar usecase(#3127) where the aim was to decouple cloudProvider specific implementation from core and being able to plugin new cloudProvider and processors at boot-up time.

I went through your proposal and had a query:
- How  can  ExternalCloudProvider plugin nodeInfoComparator? (a very common requirement for cloudProvider implementation)

bhargav-naik:
Some more points:
- Is it a design goal to give flexibility to let cloudProvider to use any language/framework of choice for their implementation?
- If someone wants to override the autoscaling_processor, they will still have to maintain their fork and add override the processor.
- Why do we want to go for a solve which will increase runtime complexity (need of separate process for CloudProvider and have a gRPC communication between CA and CloudProvider process)  instead of making changes in compile time workflow ?

hardikdr:
We would also need to define the contract between gRPC client and server, basically the messages, error-codes, and so on. Wondering if that should be part of this proposal as well?
- Also, there could be a scope of learning a little from [here](https://github.com/container-storage-interface/spec/blob/master/spec.md) the csi-model, around using error-schemes, secret management, and so on, only if relatable.

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: > Perhaps most importantly Cluster API seems to still be evolving, so it doesn't necessarily fit the goal of minimizing maintenance effort (which seems to be a driving factor here). All in all, I think it makes sense to support both Cluster API and this new provider and let the users choose.

Yeah, I believe many people is still not able to transition to cluster-api and cluster-api is not mature enough yet.

@elmiko I'd mention that an alternative would be to use the cluster-api.

elmiko: > Yeah, I believe many people is still not able to transition to cluster-api and cluster-api is not mature enough yet.
> 

i agree, it is still working through alpha versions.

> @elmiko I'd mention that an alternative would be to use the cluster-api but I don't know if you have any example on how this could be possible.

i think that's all i'm asking for here, just changing the language so that it doesn't paint a black/white picture.  you might mention that cluster-api is another approach to this problem, i don't think it needs to be more than that.

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: I'd propose `http`, otherwise I'd go with `external`

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: gotcha! I'd change it accordingly

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,127 @@
...
+```
+
+In order to talk to the custom cloud provider server, this new cloud provider has to be registered
+when bootstrapping the CA. 
+Consequently, the CA needs to expose new flags to specify the cloud provider and all the required properties
+to reach the remote gRPC server.
+
+A new flag, named
+`--cloud-provider-url=https://local.svc.io/mycloudprovider/server`, determines the URL to reach the custom provider implementation.
+In addition this approach reuses the existing flag that defines the name of the cloud provider `--cloud-provider=mycloudprovider`
```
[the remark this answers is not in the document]
hectorj2f: It makes sense, I'd use `external`

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: I see your point, I'd avoid doing the cleanup now but it is true that function isn't used on any provider :S. 

I'll add all the functions

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: I'd add the NodeGroup functions, I'd also choose `Option2` as the default mechanism to use the plugin.

hectorj2f:
@bhargav-naik 

> Is it a design goal to give flexibility to let cloudProvider to use any language/framework of choice for their implementation?

I think a bit of common sense would help here, golang is the language commonly used in Kubernetes. If users want to code on a different language that would be up to them. Besides of that, this is not a requirement neither a motivation for this proposal.

> If someone wants to override the autoscaling_processor, they will still have to maintain their fork and add override the processor.

This external provider could be extended to satisfy that purpose. It should not be considered as the initial requirement in my opinion, (solving a problem at the time). We could iterate over it and leave the design open to expose the processors per provider in the future. 

> Why do we want to go for a solve which will increase runtime complexity (need of separate process for CloudProvider and have a gRPC communication between CA and CloudProvider process) instead of making changes in compile time workflow ?

I have to argue about this. The `runtime complexity` would depend on where this external provider would seat. I believe the purpose is being flexible and dynamic here. gRPC doesn't offer a bit penalty in terms of computation.

Regarding https://github.com/kubernetes/autoscaler/pull/3127, I think the development and maintenance cost of moving out all the cloud providers is really high. This proposal tries to add a new external provider without re-designing the whole CA. The impact is minimal and everyone could benefit from it.

bhargav-naik:
Inline replies:
_I think a bit of common sense would help here, golang is the language commonly used in Kubernetes. If users want to code on a different language that would be up to them. Besides of that, this is not a requirement neither a motivation for this proposal._
- In absence of this requirement, what is the motivation to decouple CloudProvider using gRPC.

_This external provider could be extended to satisfy that purpose. It should not be considered as the initial requirement in my opinion, (solving a problem at the time). We could iterate over it and leave the design open to expose the processors per provider in the future._
- I called out this as it was called out as a requirement in #3127. Also it came out as requirement primarily for nodeInfoComparator.

_I have to argue about this. The runtime complexity would depend on where this external provider would seat. I believe the purpose is being flexible and dynamic here. gRPC doesn't offer a bit penalty in terms of computation._
- Now user will have to solve on how to run 1 extra process and deal with issues/debugging of distributed process. If the cloudProvider is able to plugin it's implementation in the same runtime wouldn't it be simpler.

hectorj2f:
> I called out this as it was called out as a requirement in #3127. Also it came out as requirement primarily for nodeInfoComparator.

Does #3127 cover this scenario ?

> Now user will have to solve on how to run 1 extra process and deal with issues/debugging of distributed process. If the cloudProvider is able to plugin it's implementation in the same runtime wouldn't it be simpler.

You are making deployment assumptions that will depend on how and where the grpc server leaves. I suggest you have a look at grafana go plugins.

> If the cloudProvider is able to plugin it's implementation in the same runtime wouldn't it be simpler.

Honestly, I didn't like your initial approach because was unrealistic and hard to maintain in a longer term. However you just made some changes that might ease this integration, that makes me like it more now.

bhargav-naik:
_Does #3127 cover this scenario ?_
- I have updated the current proposal to enable injection of autoscaling_processor in a similar manner handled in the original proposal. Please have a look at it and provide inputs.

_Honestly, I didn't like your initial approach because was unrealistic and hard to maintain in a longer term._
- The motivation for earlier approach was to decouple the Cloudprovider completely from the core (not having client libraries even in a vendored form). Though I understand the apprehension on such a big change as it radically changes the dev/build/release cycles. I leave that call with the project maintainers. Though I believe that in the long term it will be benificial as core, cloudprovider modules can have their own independent release per CloudProvider. End user can just use the required go module (based on their CP) and not have to maitain their separate builds. They will not get any unnecessary dependencies (even in vendored form).

cluster-autoscaler/proposals/plugable-provider-grpc.md
michaelgugino: The integration with cluster-api could easily be made more generic.  The autoscaler could scale *something* and something else responds.  Today, that something is machineSets or machineDeployments, but it could easily be componentX.

Perhaps cluster-api implementation should be refactored to support a generic scalable resource.

mwielgus:
What is the status of this PR? Did you address the comments?

hectorj2f:
@mwielgus I didn't address them yet. I'd try to address them during this week.

Bessonov:
@hectorj2f great stuff, keep going! It can resolve a bunch of use cases: https://github.com/kubernetes/autoscaler/issues/3644

cluster-autoscaler/proposals/plugable-provider-grpc.md
Bessonov: May be ExternalgRPCCloudProvider?

hectorj2f:
@MaciekPytel @Bessonov @hardikdr @elmiko I finally managed to get some time to invest on this KEP to address the last comments. Let me know if you want me to specify the error codes as well.

hectorj2f:
Generally, the intention is to use gRPC to communicate with the external provider similarly how csi driver works. You could develop an experimental new provider gRPC server using a potential stub file available for Go.

elmiko:
thanks for the ping @hectorj2f , i'll give this another read through tomorrow or monday.

hectorj2f:
thanks @elmiko for having a look again, :).

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,608 @@
...
+to support custom providers as plugins.
+This approach is inspired based on [Hashicorp go-plugin](https://github.com/hashicorp/go-plugin) and [Grafana Go SDK for plugins](https://github.com/grafana/grafana-plugin-sdk-go).
+
+There are two alternatives on how to plug custom cloud providers:
+
+* **Option1:** Install the plugin as a binary that would be mounted into the CA container
+and invoked by CA server.
+In this option CA server launches each provider plugin as a subprocess and communicates with it over gRPC.
+
+* **Option2:** A custom cloud provider server is deployed along side CA and both communicates via gRPC with TLS/SSL.
```
elmiko, line 44: i just want to understand these options more thoroughly.

in option 1, would the CA be deploying the subprocess?
would it deployed in the same container?

in option 2, the plugin would be started as a separate container (eg from a deployment or similar)?

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,608 @@
...
+In order to talk to the custom cloud provider server, this new cloud provider has to be registered
+when bootstrapping the CA.
+Consequently, the CA needs to expose new flags to specify the cloud provider and all the required properties
+to reach the remote gRPC server.
+
+A new flag, named
+`--cloud-provider-url=https://local.svc.io/mycloudprovider/server`, determines the URL to reach the custom provider implementation.
+In addition this approach reuses the existing flag that defines the name of the cloud provider using a pre-defined value `--cloud-provider=external`
+https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider. `external` defines the usage
+of an external cloud provider whose interface is handled by a remote gRPC service.
```
elmiko, line 196: to help avoid future conflicts,  i think we should call this "externalgrpc" or something similar. what do you think?

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,608 @@
...
+
+To connect the CA core with this new external cloud provider, this approach needs to implement a new generic cloud provider
+as part of the CA core code.
+This new provider, named `ExternalCloudProvider`, makes gRPC calls to the remote functions exposed by the external cloud provider server. In other words, it forwards the calls and handle the errors analogously how done in other existing providers.
```
elmiko, line 200: similar to previous comment, we should probably name this `ExternalGrpcCloudProvider`

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,608 @@
...
+
+* Support custom cloud provider implementations without changing the current `CloudProvider` interface.
+* Make CA extendable, so users do not need to fork the CA repository.
+
+## Proposal
+
+There are couple of examples of pluggable designs using Go SDKs that would guide us on how to extend CA
+to support custom providers as plugins.
+This approach is inspired based on [Hashicorp go-plugin](https://github.com/hashicorp/go-plugin) and [Grafana Go SDK for plugins](https://github.com/grafana/grafana-plugin-sdk-go).
+
```
elmiko, line 37: i think it would be nice to add something here about the general architecture that is being proposed, just to help the reader understand better. as i understand it, regardless of the option chosen, this proposal will involve the CA and  a provider-specific gRPC server running as a separate binary.

elmiko (commented):
@hectorj2f in general this makes sense to me, i added a suggestion and a couple questions i have.

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: yeah, it sounds good to me.

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: yes, i'll change it accordingly.

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: yes, that is correct. I'll add a graph to the kep.

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: Yeah, I left this open but I am more inclined to Option2.

hectorj2f: I prefer Option2 at least as part of an initial version. wdyt ?

elmiko: i also prefer option 2. i think option 1 does not follow what i would expect for kubernetes applications with respect to deploying/managing multiple processes in the same container.

additionally, it makes the architecture much clearer to a user who might need to customize how these containers are deployed.

hectorj2f: I totally agree. I'll change that too.

hectorj2f: Done

hectorj2f:
@mwielgus @MaciekPytel Please, could you have a look again to this PR ?

elmiko (approved):
the updates look good to me @hectorj2f , i like the diagram i think it helps. i haven't done _too_ many enhancements here but this is looking good to me.
/lgtm

hectorj2f:
@mwielgus Please, could you review this PR?

hectorj2f:
@gjtempleton Please, could you review this PR? This will unblock us and other projects too. Thanks in advance.

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,604 @@
...
+  rpc PricingNodePrice(PricingNodePriceRequest)
+    returns (PricingNodePriceResponse) {}
+
+  rpc PricingPodPrice(PricingPodPriceRequest)
+    returns (PricingPodPriceResponse)
+
+  rpc NewNodeGroup(NewNodeGroupRequest)
+    returns (NewNodeGroupResponse) {}
+
+  rpc GetResourceLimiter(GetResourceLimiterRequest)
```
MaciekPytel, line 139: I would skip this. It's only useful if you want to dynamically override --cores-total, --memory-total or --gpu-total from cloudprovider. It works, but I don't think those flags are popular at all, much less that there is any demand for dynamically changing them.

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,604 @@
...
+```
+
+### NodeGroupForNode
+
+NodeGroupForNode returns the node group for the given node, nil if the node should not be processed by cluster autoscaler, or non-nil error if such occurred. Must be implemented.
+
+```protobuf
+message NodeGroupForNodeRequest {
+  // Node group for the given node
+  k8s.io.api.core.v1.Node node = 1;
```
MaciekPytel, line 241: Sending the whole node is probably unnecessary and likely to create performance problems (especially without caching). I would consider sending just name, provider id and maybe a few other fields that may be useful for implementation (labels and annotations?).

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,604 @@
...
+  rpc Name(NameRequest)
+    returns (NameResponse) {}
+
+  rpc NodeGroups(NodeGroupsRequest)
+    returns (GetNameResponse) {}    
+
+  rpc NodeGroupForNode(NodeGroupForNodeRequest)
+    returns (NodeGroupForNodeResponse) {}
+
+  rpc PricingNodePrice(PricingNodePriceRequest)
```
MaciekPytel, line 130: This is only needed for price expander and most existing providers don't implement it. I don't mind having it, but we should at least explicitly say implementing this is optional (at least beyond implementing an empty stub response).

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,604 @@
...
+ string name = 1;
+}
+```
+
+### NodeGroup
+
+```protobuf
+message NodeGroup {
+  // Name of the node group on the cloud provider
+  string name = 1;
```
MaciekPytel, line 430: Delete this? It seems to duplicate id below.

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,604 @@
...
+}
+```
+
+### NodeGroupDeleteNodes
+
+NodeGroupDeleteNodes deletes nodes from this node group.
+
+```protobuf
+message NodeGroupDeleteNodesRequest {
+ repeated k8s.io.api.core.v1.Node nodes = 1;
```
MaciekPytel, line 491: Again, I don't think there is any need to send the whole node object (which may be outdated, etc). I think we should consider defining some sort of node identifier and use this.

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,604 @@
...
+	// Id is instance id.
+	string id = 1;
+	// Status represents status of node. (Optional)
+	InstanceStatus status = 2;
+}
+
+// InstanceStatus represents instance status.
+message InstanceStatus {
+	// State tells if instance is running, being created or being deleted
+	int state = 1;
```
MaciekPytel, line 543: I wonder if this shouldn't be an enum?

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,604 @@
...
+
+import "k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto";
+import "k8s.io/apimachinery/pkg/api/resource/generated.proto";
+import "k8s.io/api/core/v1/generated.proto"
+
+option go_package = "clusterautoscaler.cloudprovider";
+
+service CloudProvider {
+  // CloudProvider specific RPC functions
+  rpc Name(NameRequest)
```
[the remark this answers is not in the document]
hectorj2f: I'll let it out of this initial scope.

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,604 @@
...
+  rpc NodeGroupForNode(NodeGroupForNodeRequest)
+    returns (NodeGroupForNodeResponse) {}
+
+  rpc PricingNodePrice(PricingNodePriceRequest)
+    returns (PricingNodePriceResponse) {}
+
+  rpc PricingPodPrice(PricingPodPriceRequest)
+    returns (PricingPodPriceResponse)
+
+  rpc NewNodeGroup(NewNodeGroupRequest)
```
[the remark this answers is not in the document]
hectorj2f: @MaciekPytel  I can see this function used by many providers. Is that correct ?

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: I will remove it from this first proposal.

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: I added `Optional` descriptions in the operation and the operation definition.

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: Yeah, it is a duplicate of the id, removing it!.

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,604 @@
...
+
+```protobuf
+message NodeGroupTemplateNodeInfoRequest {
+  // ID of the group node on the cloud provider
+  string id = 1;
+}
+
+message NodeGroupTemplateNodeInfoResponse {
+  // nodeInfo extracted of the template node on the cloud provider
+  NodeInfo nodeInfo = 1;
```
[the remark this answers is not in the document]
hectorj2f: For the sake of simplicity for this initial implementation, I'd say we can state that it won't scale-from-0 so it makes this initial development much easier.

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: Yes, you're right. I fix it now.

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: It makes sense, I'll update the message properties to reflect those.

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,604 @@
...
+
+service CloudProvider {
+  // CloudProvider specific RPC functions
+  rpc Name(NameRequest)
+    returns (NameResponse) {}
+
+  rpc NodeGroups(NodeGroupsRequest)
+    returns (GetNameResponse) {}    
+
+  rpc NodeGroupForNode(NodeGroupForNodeRequest)
```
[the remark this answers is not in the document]
hectorj2f: I agree, I'll add your concern in the proposal. Generally I'd vote for `moving this whole function to CA` but for now, we could start with the assumption of having some caching for the rpc responses.

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: I added the aforementioned properties, but this list could increase during the implementation phase.

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: It makes sense. Good call!

cluster-autoscaler/proposals/plugable-provider-grpc.md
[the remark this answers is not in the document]
hectorj2f: Okay, I changed my opinion. I decided to use for now the Node Kubernetes primitive. In my opinion, we could still collect relevant information from this resource type.

cluster-autoscaler/proposals/plugable-provider-grpc.md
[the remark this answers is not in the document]
MaciekPytel: After thinking about it a bit I think the problem with moving the whole function to CA is timing:
 * CA loop can easily take a minute or more in a large cluster. New nodes can show up in that time, so any datastructure precomputed by CA is likely to run into races. Especially in a scenario involving a rapid scale-up of hundreds of nodes, it's almost guaranteed that nodes created by CA in previous loops will register mid-loop.
 * There may be node objects that don't have matching VM on provider side (after VM is gone it takes time for the node object to be cleaned up). 

I don't see an easy way to generically implement NodeGroupForNode that works for those cases. Implementation based on calling Nodes() and doing lookups on result should work out eventually, but there may be failed loops until the cluster reconciles, which will impact autoscaling latency.

cluster-autoscaler/proposals/plugable-provider-grpc.md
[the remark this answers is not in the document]
MaciekPytel: It's not called at all by CA logic (except for some old unittests that call it on mock cloudprovider). So even if some provider were to implement it it would do nothing as the function will never be called.

Almost all existing providers return ErrNotImplemented for this method. The only exception is ovhcloud. However, since it was just added I'm guessing this is because we lack documentation for adding providers and the authors of the PR were unaware of the above. cc: @Kafei59 - can you confirm?

cluster-autoscaler/proposals/plugable-provider-grpc.md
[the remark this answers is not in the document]
hectorj2f: Then we don't have a choice and we should rely on the remote grpc operation to gather this information. This sounds as the only alternative based on your comments. Blocking the CA loop sounds like a bad practice as well as having inconsistent data due to the volatile nature of VMs.

cluster-autoscaler/proposals/plugable-provider-grpc.md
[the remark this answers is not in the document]
hectorj2f: Alright! I'll remove it.

cluster-autoscaler/proposals/plugable-provider-grpc.md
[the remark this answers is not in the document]
hectorj2f: @MaciekPytel Let me know if you see this as a blocker or we could move forward assuming there would be a cache for the rpc responses here. I removed the part that mentioned the initial proposal of moving this code to the CA which could increase the response time of the reconciliation loop.

hectorj2f:
It seems i hit the docker request limit in my travis build, ` toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading`.

cluster-autoscaler/proposals/plugable-provider-grpc.md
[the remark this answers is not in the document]
Kafei59: Yep, I only tried to follow the interface contract, but in fact, it if's not called ... well, it's useless :)

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,527 @@
...
+
+import "google/protobuf/descriptor.proto";
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/wrappers.proto";
+
+import "k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto";
+import "k8s.io/apimachinery/pkg/api/resource/generated.proto";
+import "k8s.io/api/core/v1/generated.proto"
+
+option go_package = "clusterautoscaler.cloudprovider";
```
Kafei59, line 117: I'm curious, you do not set the v1 in the go_package as well ?

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,527 @@
...
+GetAvailableGPUTypes handles all available GPU types cloud provider supports.
+
+```protobuf
+message GetAvailableGPUTypesRequest {
+  // Intentionally empty.
+}
+
+message GetAvailableGPUTypesResponse {
+ // GPU types passed in as opaque key-value pairs.
+  map<string, string> gpuTypes = 1;
```
Kafei59, line 288: actually, in the current interface it returns a map[string]struct{}, making it a string will break changes, imo

Kafei59 (approved):

Kafei59:
Hi @hectorj2f 👋
I really like the idea, I just implemented a cloudprovider for OVHcloud so I know the pain behind copy-pasting the go sdk for building the provider.
Also, I currently work on gRPC eco-system so, if you need any help, we should definitively check this out together !

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: Yes, it was not intentional but I didn't go into low implementation details. My main intention was to share a draft of a proto file.

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: I changed it now.

cluster-autoscaler/proposals/plugable-provider-grpc.md
hectorj2f: done.

hectorj2f:
@MaciekPytel I addressed the last minor comments.

cluster-autoscaler/proposals/plugable-provider-grpc.md
```diff
@@ -0,0 +1,528 @@
...
+```protobuf
+message NodeGroup {
+  // ID of the node group on the cloud provider
+  string id = 1;
+
+  // MinSize of the node group on the cloud provider
+  int32 minSize = 2;
+
+  // MaxSize of the node group on the cloud provider
+  int32 maxSize = 3;
```
MaciekPytel, line 389: I'm not sure if we should actually include min, max and debug. min/max can change (after Refresh() call), while id is the consistent group identifier. For something like NodeGroupForNode it seems more consistent to return permanent id and not some settings that can change later on.

MaciekPytel (approved):
/lgtm
/approve

Left some minor comments, but I don't think those are blocking for merging this.

MaciekPytel:
One more comment - in the meantime #3789 extended the cloudprovider API with another optional call. It should trivially convert into grpc (struct with a field for each option in NodeGroupAutoscalingOptions) or alternatively it can be skipped in initial version.

hectorj2f:
@MaciekPytel Good. I'll update the proposal to include that new api call, and address the minor comments.

migueleliasweb:
Hi all, is there any work being done to implement this provider? If not, would a PR be welcomed?

hectorj2f:
Hi @migueleliasweb, 
Unfortunately there is not any on-going work. Our company switched gears to focus on the CAPI autoscaler as a solution.

PR #3141, opened 2020-05-20
Avoid sending extra deletion calls for in-progress deletions

marwanad:
This utilizes the `cloudprovider.InstanceStatus` struct to prevent sending another deletion call for the cases that exceed the `MaxCloudProviderNodeDeletionTime` of 5 minutes (large SKUs) for instance.

Those calls are rejected with 409s but we should prevent sending those in the first place.

/area provider/azure

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
@@ -321,9 +321,7 @@ func (scaleSet *ScaleSet) IncreaseSize(delta int) error {
 }
 
 // GetScaleSetVms returns list of nodes for the given scale set.
-// Note that the list results is not used directly because their resource ID format
-// is not consistent with Get results.
-func (scaleSet *ScaleSet) GetScaleSetVms() ([]string, *retry.Error) {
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -682,16 +678,65 @@ func (scaleSet *ScaleSet) Nodes() ([]cloudprovider.Instance, error) {
...
+	}
+
+	status := &cloudprovider.InstanceStatus{}
+	switch *vm.ProvisioningState {
+	case string(compute.ProvisioningStateDeleting):
+		status.State = cloudprovider.InstanceDeleting
+	case string(compute.ProvisioningStateCreating):
+		status.State = cloudprovider.InstanceCreating
+	default:
+		status.State = cloudprovider.InstanceRunning
```
marwanad, line 736: We default to `Running` here. There's no action taken by core CA taken on those state except for `InstanceCreating` cases that also populate the `status.Error`. 

We can look into utilizing `instanceView` to surface CSE failures or OS provisioning errors and get CA to delete those VMs faster than waiting for `maxNodeProvisionTime`

PR #3149, opened 2020-05-21
Metrics-based addon-resizer

wojtek-t:
This PR adds an ability to get size of the cluster using kube-apiserver /metrics endpoint (in particular the etcd_object_count metric) as an opt-in option. The advantage from this is that for very large clusters, the memory usage (as well as cpu usage) is not as big as it is when node objects need to listed from kube-apiserver.

To use it, you need to:
- pass the `--use-metrics=true` flag
- potentially, add permissions to your cluster-role, as following:
```
- nonResourceURLs:
  - /metrics
  verbs:
  - get
```

/assign @mm4tt @bskiba


wojtek-t:
@mm4tt @bskiba - this is now ready for review

mm4tt:
Thanks, will take a look today

addon-resizer/nanny/kubernetes_client.go
```diff
@@ -53,6 +72,45 @@ func (k *kubernetesClient) CountNodes() (uint64, error) {
 	return uint64(len(result.Items)), err
 }
 
+func (k *kubernetesClient) countNodesThroughMetrics() (uint64, error) {
+	rawMetrics, err := k.clientset.Core().RESTClient().Get().RequestURI("/metrics").DoRaw()
+	if err != nil {
+		return 0, err
+	}
+
+	decoder := expfmt.SampleDecoder{
```
mm4tt, line 81: The code would be cleaner if you used `expfmt.TextParser.` You'd get a map from metric name to `dto.MetricFamily`, then you would have to just iterate over metrics to find sample with proper label.

addon-resizer/nanny/main/pod_nanny.go
```diff
@@ -60,6 +60,7 @@ var (
 	pollPeriod     = time.Millisecond * time.Duration(*flag.Int("poll-period", 10000, "The time, in milliseconds, to poll the dependent container."))
 	estimator      = flag.String("estimator", "linear", "The estimator to use. Currently supported: linear, exponential")
 	minClusterSize = flag.Uint64("minClusterSize", 16, "The smallest number of nodes resources will be scaled to. Must be > 1. This flag is used only when an exponential estimator is used.")
+	useMetrics     = flag.Bool("use-metrics", false, "Whether to use Kubernetes API or apiserver metrics to detect cluster size")
```
mm4tt, line 63: Leaving the @bskiba whether this should be documented in README.md

wojtek-t:
> I have no idea why this change is made against the addon-resizer-release-1.8 branch, but I assume is intended. Leaving to @bskiba to confirm.

@mm4tt 2.* releases contain some bug that noone never tried to debug. So we're sticking to 1.8 release in all our uses I'm aware of.
So yes - that was fully conscious decision.

addon-resizer/nanny/kubernetes_client.go
wojtek-t: I don't really see the difference - the code would be almost exactly the same - I just wouldn't need the catching of io.EOF.

Given that this is what is used in couple different places in both k/k and k/perf-tests (thus is validated approach) and TextParser isn't used anywhere, I'm going to stick to what there is now.

mm4tt: I see a huge difference between your code and this snippet

```
metrics := expfmt.TextParser{}.TextToMetricFamilies(rawMetrics)
if metricFamily, ok := metrics[objectCountMetricName]; ok {
   for _, m := metricFamily.Metric {
      for _, l := m.Label {
        if l.Name == "resource" && l.Value = nodeResourceName {
           value := uint64(m.Counter.Value)
		   if value < 0 {
		     return 0, fmt.Errorf("metric unknown")
		   }
		   return value, nil
        }
      }
   }
}
return 0, fmt.Errorf("metric unset")
```

addon-resizer/nanny/kubernetes_client.go
```diff
@@ -53,6 +72,45 @@ func (k *kubernetesClient) CountNodes() (uint64, error) {
...
+				break
+			}
+			continue
+		}
+		for _, metric := range v {
+			name := metric.Metric[model.MetricNameLabel]
+			if name != objectCountMetricName {
+				continue
+			}
+			resource := metric.Metric["resource"]
```
mm4tt, line 98: NIT: Given that we have a constant for "nodes" we should also introduce one for "resource"

addon-resizer/nanny/kubernetes_client.go
wojtek-t: TBH, I see bigger value in consistency with number of other places, than in saving 5 lines of simple code.

addon-resizer/nanny/kubernetes_client.go
wojtek-t: done

addon-resizer/nanny/kubernetes_client.go
mm4tt: If it's about consistency then `Decoder` with `FmtText` format calls `TextToMetricFamilies` underneath:

https://github.com/prometheus/common/blob/d0f7cd64bda49e08b22ae8a730aa57aa0db125d6/expfmt/decode.go#L127

It's not only about saving lines of code, it's also about readability.  Libraries evolve, copying-pasting solutions for sake of consistency is not always optimal.

That said, I don't see a point in arguing about a few lines of code. Your code is correct, I'd do it differently, but it looks good to me.

addon-resizer/nanny/main/pod_nanny.go
bskiba, line 63: If we use a boolean flag for this I would phrase it as: "Whether to use apiserver metrics to detect node count instead of the default method of listing node objects from the Kubernetes API".

bskiba: It would be nice to drop a line, yes.

addon-resizer/nanny/kubernetes_client.go
```diff
@@ -53,6 +74,45 @@ func (k *kubernetesClient) CountNodes() (uint64, error) {
 	return uint64(len(result.Items)), err
 }
 
+func (k *kubernetesClient) countNodesThroughMetrics() (uint64, error) {
+	rawMetrics, err := k.clientset.Core().RESTClient().Get().RequestURI("/metrics").DoRaw()
```
bskiba, line 78: Two questions here:
Should we be concerned about the size of the response here, since we are reading all metrics?
Is the API server's metrics endpoint always accessible from the nodes?

addon-resizer/nanny/kubernetes_client.go
```diff
@@ -18,23 +18,44 @@ package nanny
...
 }
 
 func (k *kubernetesClient) CountNodes() (uint64, error) {
+	if k.useMetrics {
+		return k.countNodesThroughMetrics()
+	}
+	return k.countNodesThroughAPI()
+}
+
+func (k *kubernetesClient) countNodesThroughAPI() (uint64, error) {
```
bskiba, line 58: Can you leave a one line comment on the two different methods of getting the node count here or in the CountNodes() method?

wojtek-t: done

addon-resizer/nanny/kubernetes_client.go
wojtek-t: 1. It may be larger for very small clusters, but it will be smaller for larger clusters (comparing to listing nodes). We still need to get exact numbers.

2. if they have permissions (similarly as for listing nodes or anything else) - then yes. Added comment.

addon-resizer/nanny/main/pod_nanny.go
wojtek-t: done

wojtek-t: done

bskiba:
perfect, thanks
/lgtm
/approve

PR #3152, opened 2020-05-24
Exclude the virtual node from nodes not created by auto scaling group

### LINKED ISSUE #3151: cluster autoscaler may not scale up the cluster if virtual kublet is deployed
Root cause:

The virtual node created by virtual kubelet usually has a lots of cpu cores and memory bytes, and may exceed the cpu cores limit and memory limit passed to the cluster autoscaler, therefore cluster autoscaler fails to scale up the cluster automatically.

Possible solution:
Exclude the virtual node when calculating total cores and memory of `nodesFromNotAutoscaledGroups` based on label `type=virtual-kubelet`
https://github.com/kubernetes/autoscaler/blob/0255c1152773ecfcc532fe6ca494414351a946df/cluster-autoscaler/core/scale_up.go#L133-L137

Or exclude the virtual node in function `FilterOutNodesFromNotAutoscaledGroups`
https://github.com/kubernetes/autoscaler/blob/0255c1152773ecfcc532fe6ca494414351a946df/cluster-autoscaler/core/utils/utils.go#L191-L206

knight42:
Fix #3151

### DIFF
--- a/cluster-autoscaler/core/utils/utils.go
+++ b/cluster-autoscaler/core/utils/utils.go
@@ -188,12 +188,21 @@ func GetNodeInfoFromTemplate(nodeGroup cloudprovider.NodeGroup, daemonsets []*ap
 	return sanitizedNodeInfo, nil
 }
 
+// isVirtualNode determines if the node is created by virtual kubelet
+func isVirtualNode(node *apiv1.Node) bool {
+	return node.ObjectMeta.Labels["type"] == "virtual-kubelet"
+}
+
 // FilterOutNodesFromNotAutoscaledGroups return subset of input nodes for which cloud provider does not
 // return autoscaled node group.
 func FilterOutNodesFromNotAutoscaledGroups(nodes []*apiv1.Node, cloudProvider cloudprovider.CloudProvider) ([]*apiv1.Node, errors.AutoscalerError) {
 	result := make([]*apiv1.Node, 0)
 
 	for _, node := range nodes {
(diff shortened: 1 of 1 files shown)


MaciekPytel:
Is the problem here that you're trying to use core/memory limits and think they shouldn't include virtual nodes, or just that virtual node has so much resources that it exceeds default value for limits? 

In the first case - why shouldn't those resources be included? The idea is to provide a mechanism that allows to control total size of the cluster which has many different NodeGroups. It seems inconsistent to ignore resources from virtual nodes if we count other non-autoscaled nodes in cluster.

In the second case - maybe it would be easier to just bump default limits in https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/config/const.go#L21? This way we don't have to think about all the questions I listed above.

knight42:
@MaciekPytel I think the both cases matter.

In the first case, the allocatable resources specified in the virtual node is up to the virtual kubelet and its resources has no relation with the size of the cluster. Worse still, it will interfere with the scaling up process:
```
Skipping node group <group-id-redacted>; maximal limit exceeded for [cpu memory]
```

In the second case, as mentiond above, the allocatable resources specified in the virtual node is up to the virtual kubelet and different cloud provider has different implementation, it may not be feasible to specify large enough default limits. Simply excluding the virtual node seems to be more easier IMHO.

knight42:
@MaciekPytel just a friendly ping. Did you consider #3151 is an issue that needs to be fixed?

mwielgus (approved):
/lgtm
/approve

PR #3155, opened 2020-05-26
Support Magnum node groups

### LINKED ISSUE #2011: [cluster-autoscaler][magnum] Support /resize API
Now Magnum has a /resize API which can accept the list of nodes to be removed and a target node count after the resize. That's a better option than the current way by calling Heat to remove the node and then calling Magnum to update the correct node count.

### LINKED ISSUE #2548: Support magnum node groups
With the magnum "train" release node groups are added. Old clusters will be migrated to the new system with two default node groups ("default-master" and "default-worker"), and the autoscaler will continue to work for existing clusters if no extra node groups are added to the cluster.

I have finished writing the gophercloud API for node groups, and reviewing has started https://github.com/gophercloud/gophercloud/issues/1767.

I'm now adapting the magnum cloud provider to use this API.

This will also take care of #2011.

### LINKED ISSUE #2802: Support for NodeGroups in Magnum
Now that Magnum Train has NodeGroups support, can the cluster-autoscaler work with new nodegroups that are added to a cluster?
https://github.com/kubernetes/autoscaler/blob/1d2f183dc739b3ba03496dee6f458c4c8990a130/cluster-autoscaler/cloudprovider/magnum/examples/cluster-autoscaler-deployment.yaml#L31

tghartland:
It's taken me longer than I would have liked to get this ready, but here we go.

* The vendored version of gophercloud is updated to include the magnum node groups API.
  * It's now also only vendoring the files that are actually used.
  * https://github.com/tghartland/gophercloud-vendorer
* Updated the manager to use the cluster resize API for changing the size of the cluster.
* Support for Magnum node groups.
  * Also adding node group autodiscovery using the Role property of node groups.
* Implemented the Nodes method of the NodeGroup interface.
  * It's a bit of a mess trying to map the possible states in Magnum to the cloudprovider's three Creating/Running/Deleting states, but I think I've handled every edge case I've seen.
  * The most difficult is that UPDATE_IN_PROGRESS is ambiguous, it could be Creating or Running, but it can be differentiated with a few more API calls to Heat.
* Updated the README.
  * Add a note about CA/Magnum compatibility, assuming that this can be merged for v1.19.
  * Documented how node group autodiscovery works for Magnum.

Closes #2548 
Closes #2011
Closes #2802


MaciekPytel:
I don't really have enough context to do any meaningful review here. You're the owner of magnum provider, feel free to merge changes to it.
If there are any changes to core needed, I'd prefer if you extracted them to a separate PR - I'm worried about missing something in one as big as this.

tghartland:
I haven't made any changes outside of the magnum cloud provider.

I did think about if it would be worth it to add a new method `Name()` to the NodeGroup interface. For the `Id()` I'm having to make a joined string like `ng-name-d67f5d4d` that combines the user-defined name and the UUID, to have something that is both human readable and recognisable, but satisfying the requirement that it must be unique. Since the `Id()` is used as a key for caching but also in the logs and status configmap.

I'm fine keeping things are they are for now though.

I'll aim to get this merged soon but I do want to do a bit more testing first.

tghartland:
I changed the GPULabel for magnum cloudprovider to `magnum.openstack.org/gpu`, matching the format of other magnum labels.

There is no real support for GPUs in openstack so nothing will be using this yet, it will just be a placeholder for now.

tghartland:
I'm still waiting to be able to test this with our internal setup for Kubernetes GPUs from Openstack, then I'll merge.

tghartland:
No progress on getting GPUs to test with, I'll merge this now and if I am able to get some GPUs then any fixes that I find can be done in another PR.

tghartland:
/assign @vivekbagade

@MaciekPytel or @vivekbagade I'm ready to merge this but not able to /approve or /lgtm this for myself.

vivekbagade:
/hold.
I lgtm-ed and approved. Remove the hold label when you're ready to merge

tghartland:
Thanks @vivekbagade. I had to fix a string formatting error that the linter caught, so I lost the lgtm label. If you can add that back then I'm ready to merge.

tghartland:
Thanks @vivekbagade. I see the first beta release of v1.19 was two days ago, will this be included in beta.2 / the actual release?

vivekbagade:
I think it should be.

openstacker:
@tghartland  Could this work with k8s v1.16.x?

PR #3156, opened 2020-05-26
Count live Pods per VPA

kgolab:
Replaces a map of bools which were computed based on aggregated states with a map of ints, computed based on raw data.

One potentially heavy operation is a full scan of Pods whenever new VPA object is created, apart from that I don't see any major performance impact.
The full scan is in-memory only, we have all the necessary data already fetched.

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/model/cluster.go
+++ b/vertical-pod-autoscaler/pkg/recommender/model/cluster.go
@@ -46,9 +46,8 @@ type ClusterState struct {
 	// time we've noticed the recommendation missing or last time we logged
 	// a warning about it.
 	EmptyVPAs map[VpaID]time.Time
-	// VpasWithMatchingPods contains information if there exist live pods that
-	// this VPAs selector matches.
-	VpasWithMatchingPods map[VpaID]bool
+	// VpaPodCount contains number of live Pods matching a given VPA object.
+	VpaPodCount map[VpaID]int
 	// Observed VPAs. Used to check if there are updates needed.
 	ObservedVpas []*vpa_types.VerticalPodAutoscaler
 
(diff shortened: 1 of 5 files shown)


kgolab:
@bskiba , please take a look.
If you're OK with this change, I'll base more metrics off it.

vertical-pod-autoscaler/pkg/recommender/model/cluster.go
```diff
@@ -124,18 +123,44 @@ func (cluster *ClusterState) AddOrUpdatePod(podID PodID, newLabels labels.Set, p
...
 	}
 	pod.Phase = phase
 }
 
+// addPodToItsVpa increases the count of Pods associated with a VPA object.
+// Does a scan similar to findOrCreateAggregateContainerState so could be optimized if needed.
+func (cluster *ClusterState) addPodToItsVpa(pod *PodState) {
+	for _, vpa := range cluster.Vpas {
+		if vpa_utils.PodLabelsMatchVPA(pod.ID.Namespace, cluster.labelSetMap[pod.labelSetKey], vpa.ID.Namespace, vpa.PodSelector) {
+			cluster.VpaPodCount[vpa.ID]++
```
bskiba, line 150: Since we are currently in the state where only one VPA object should match pod, should we return early?

vertical-pod-autoscaler/pkg/recommender/routines/recommender.go
```diff
@@ -103,16 +103,18 @@ func (r *recommender) UpdateVPAs() {
...
 			klog.V(4).Infof("VPA dump")
 			klog.V(4).Infof("%+v", vpa)
 			klog.V(4).Infof("HasMatchingPods: %v", hasMatchingPods)
+			podCount := r.clusterState.VpaPodCount[vpa.ID]
+			klog.V(4).Infof("VpaPodCount: %v", podCount)
 			pods := r.clusterState.GetMatchingPods(vpa)
 			klog.V(4).Infof("MatchingPods: %+v", pods)
-			if len(pods) > 0 != hasMatchingPods {
+			if len(pods) != podCount {
 				klog.Errorf("Aggregated states and matching pods disagree for vpa %v/%v", vpa.ID.Namespace, vpa.ID.VpaName)
```
bskiba, line 118: nit: ClusterState pod count and matching pods disagree

vertical-pod-autoscaler/pkg/recommender/model/cluster.go
kgolab: I was wondering the same and left it as it since we have the same construct in multiple places and I didn't want to end up with an outlier.
Or do you think it's still worth changing it?

vertical-pod-autoscaler/pkg/recommender/routines/recommender.go
kgolab: Thanks, fixed.

vertical-pod-autoscaler/pkg/recommender/model/cluster.go
bskiba: I guess it's fine then, though that other places come from the time when we had selector-based API.

PR #3161, opened 2020-05-28
Update group identifier to use for Cluster API annotations

detiber:
Builds on top of #3203 

- Also add backwards compatibility for the previously used deprecated annotations


cluster-autoscaler/cloudprovider/clusterapi/clusterapi_nodegroup.go
```diff
@@ -25,9 +25,13 @@ import (
 )
 
 const (
-	machineDeleteAnnotationKey = "cluster.k8s.io/delete-machine"
-	machineAnnotationKey       = "cluster.k8s.io/machine"
-	debugFormat                = "%s (min: %d, max: %d, replicas: %d)"
+	// deprecatedMachineDeleteAnnotationKey should not be removed until minimum cluster-api support is v1alpha3
+	deprecatedMachineDeleteAnnotationKey = "cluster.k8s.io/delete-machine"
+	// TODO: determine what currently relies on deprecatedMachineAnnotationKey to determine when it can be removed
```
detiber, line 30: It wasn't actually clear to me where this is being used. @elmiko do you know? I would expect any v1alpha3 compatible infrastructure providers to be setting ProviderID. We'll still likely need to keep this for backward compatibility, but would like to have a target removal in mind.

elmiko: not sure i completely understand, but afaik, we _should_ be using the const variable names and not the actual string values anywhere. as for how it is used in the context of the autoscaler, the `DeleteNodes` function wraps this all to be used during the scale down attempts from the core.

i was under the impression that these annotations are then used to help inform during the scale down, but perhaps  we don't need them from CAPI side? (if so, it certainly seems like we could investigate removing them)

detiber: I think where I added the comment with the github UI might be messing up the context I was trying to set a bit. My comment (and the TODO is for the `deprecatedMachineAnnotationKey` rather than `deprecatedMachineDeleteAnnotationKey`.

From what I can tell we are using that annotation for matching a Node to a Machine (it's set on the Node) when a providerID is not present on the Machine to be able to use for matching. It is currently being used by some tests in this repo, but I wasn't sure if there is some known usage externally from pre-v1alpha3 cluster-api infrastructure providers.

elmiko: ok, that makes so much more sense lol =)

i think you are correct, this might be an artifact from previous generations. definitely seems like we are using it for storage of the id. as for the external usage, i'm not 100% sure but i can do a little research. certainly seems like a good candidate to cleanup.

mwielgus:
Where we are with this PR? Do you still need it?

ncdc:
I think we need #3203 to go in first

elmiko:
> Where we are with this PR? Do you still need it?

yes, we will need it

> I think we need #3203 to go in first

+1, and some rebase love as i think we have merged some of the commits in this.

detiber:
Rebased, apologies about the delay

benmoss:
Oops, I misread who was on the real approvers list, @hardikdr you're welcome to review too but I think maybe @MaciekPytel 
can review since he also reviewed #3203 :smile:

ncdc:
@detiber is this worth rebasing to drop off the commit from #3203?

ncdc (approved):
/lgtm

benmoss:
/assign @feiskyer 

sorry to spam just hoping to find someone with some cycles for a review since we have some changes outside of the CAPI provider

feiskyer (commented):
/approve

PR #3162, opened 2020-05-29
a misspelling

wangxy518:
functionallity->functionality
There is one 'l'
From https://dictionary.cambridge.org/spellcheck/english-chinese-simplified/?q=functionallity


bskiba:
This is vendored code (coming from an imported dependency).
Please open a PR against the original library.

/close

Issue #3104, opened 2020-04-30
capi: Replicas can go stale in nodeGroup causing erratic behaviour


enxebre:
DeleteNodes() is sometimes called in subroutines https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/scale_down.go#L1067-L1090

But our implementation does not account for that properly https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_nodegroup.go#L84-L149. And subroutines might compete to call setSize() with different values and getting stale replicas(). 

Also when we call `setSize()` we leave stale the replicas number in the machineSet for the scalableResource struct used by `replicas()` https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_machineset.go#L63-L85.

We need to:
- Ensure replicas() does never return an stale number and unit test it.
- Ensure multiple subroutines calls to deleteNodes() never use an stale replicas number. We should probably lock deleteNodes so any upcoming routine pick a valid replicas().
- Also we need to make the deleteNodes() logic more robust by comparing against minSize here https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_nodegroup.go#L106 rather than with 1.

/area provider/cluster-api

enxebre:
cc @elmiko

elmiko:
++, thanks Alberto!

PR #3177, opened 2020-06-02
Fix stale replicas issue with cluster-autoscaler CAPI provider

elmiko:
This change brings in a series of patches to remediate the issues with MachineSet and MachineDeployment replicas becoming stale. These changes add protection around the deletion mechanisms by adding a mutex to this operation and also changing the check versus 0 to use the minimum size for that group. Additionally, the operations to get the replicas are now using API server calls to ensure that the freshest information is available during replica count checks. Lastly, a minor change to the unit tests is added to address the underlying changes with respect to API server queries for replicas.

fixes: #3104 

/area provider/cluster-api

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
+++ b/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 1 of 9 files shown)


elmiko:
cc @enxebre @JoelSpeed @detiber

enxebre:
thanks!
/approve

PR #3178, opened 2020-06-03
Cherry-pick #3141: Avoid sending extra deletion calls for in-progress deletions 

marwanad:
This utilizes the cloudprovider.InstanceStatus struct to prevent sending another deletion call for the cases that exceed the MaxCloudProviderNodeDeletionTime of 5 minutes (large SKUs) for instance.

Those calls are rejected with 409s but we should prevent sending those in the first place.

/area provider/azure

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_cloud_provider_test.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_cloud_provider_test.go
@@ -22,6 +22,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
 	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2017-05-10/resources"
 	"github.com/Azure/go-autorest/autorest/azure"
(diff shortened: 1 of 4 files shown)


detiber:
This backport appears to have broken building on the `cluster-autoscaler-release-1.18` branch:
```
cloudprovider/azure/azure_scale_set.go:334:3: cannot use rerr (type *retry.Error) as type error in return argument:
        *retry.Error does not implement error (wrong type for Error method)
                have Error() error
                want Error() string
cloudprovider/azure/azure_scale_set.go:672:30: cannot use rerr (type error) as type *retry.Error in argument to isAzureRequestsThrottled
cloudprovider/azure/azure_scale_set.go:678:25: cannot use rerr.Error() (type string) as type error in return argument:
        string does not implement error (missing Error method)
```

detiber:
Looks like HEAD is using azure-sdk-for-go v40.2.0+incompatible compared to `cluster-autoscaler-release-1.18` using v35.0.0+incompatible

PR #3186, opened 2020-06-03
Use golang1.14.4 to compile AddonResizer

jkaniuk:
More aggressive GC should help with memory usage

/cc wojtek-t bskiba

### DIFF
--- a/addon-resizer/Makefile
+++ b/addon-resizer/Makefile
@@ -21,7 +21,7 @@ ALL_ARCH = amd64 arm arm64 ppc64le s390x
 ARCH ?= amd64
 
 GOARM=7
-GOLANG_VERSION = 1.12.1
+GOLANG_VERSION = 1.14.3
 REGISTRY = gcr.io/k8s-staging-autoscaling
 IMGNAME = addon-resizer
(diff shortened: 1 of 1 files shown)


addon-resizer/Makefile
```diff
@@ -21,7 +21,7 @@ ALL_ARCH = amd64 arm arm64 ppc64le s390x
 ARCH ?= amd64
 
 GOARM=7
-GOLANG_VERSION = 1.12.1
+GOLANG_VERSION = 1.14.1
```
wojtek-t, line 24: IIUC, 1.14.1 had some bug.
@mm4tt - do you remember?

In k8s we use 1.14.3 - so maybe we could stick to that version here too:
https://github.com/kubernetes/release/blob/master/images/build/cross/Makefile#L23

bskiba:
Please change to 1.14.3 and I'll approve

addon-resizer/Makefile
jkaniuk: That supposed to be 1.14.4, wrongly amended :-/

Changed to 1.14.3

PR #3187, opened 2020-06-03
1.15: Fix bad cherry-pick from #3141

marwanad:
1.17 and below are using the old clients not vendored from cloud-provider


marwanad:
CI status: https://travis-ci.org/github/kubernetes/autoscaler/builds/694341060

marwanad:
Seems like Travis keeps insisting on running the VPA suite instead but I did a local run to verify the unit tests aren't failing.

PR #3188, opened 2020-06-03
1.16: Fix bad cherry pick from #3141

marwanad:
1.17 and below are using the old clients not vendored from cloud-provider which don't have that error type


marwanad:
https://travis-ci.org/github/kubernetes/autoscaler/jobs/694405756

PR #3189, opened 2020-06-03
Fix bad cherry pick #3141 for 1.17

marwanad:
1.17 and below are using the old clients not vendored from cloud-provider which don't have that error type

/area provider/azure


marwanad:
https://travis-ci.org/github/kubernetes/autoscaler/jobs/694396568

PR #3191, opened 2020-06-03
fix gofmt and lint issues

marwanad:
Seems like CI status isn't reported on PRs anymore. This fixes gofmt and lint issues.


marwanad:
https://travis-ci.org/github/kubernetes/autoscaler/builds/694466745

seems like this brings master back to a happy state.

/assign @mwielgus

Issue #2786, opened 2020-01-30
Unclear documentation on permissible alternatives for AWS ASG MixedInstancesPolicy


ari-becker:
The [AWS README](https://github.com/kubernetes/autoscaler/blob/f9535479f2ca7316eb062c93918027926223c3fc/cluster-autoscaler/cloudprovider/aws/README.md#using-autoscalinggroup-mixedinstancespolicy) currently advises:

> Note that the instance types should have the same amount of RAM and number of CPU cores, since this is fundamental to CA's scaling calculations. Using mismatched instances types can produce unintended results.

The README also provides an example:

> Set LaunchTemplateOverrides to include the 'base' instance type r5.2xlarge and suitable alternatives, e.g. r5d.2xlarge, i3.2xlarge, r5a.2xlarge and r5ad.2xlarge.

This raises two questions for me:

a) While `r5.2xlarge` has 64 GB of RAM, the `i3.2xlarge` has less: 61 GB of RAM. Wouldn't the 3 fewer GB of RAM play havoc with the CA's scaling calculations, as documented? Am I missing something here, or is `i3.2xlarge` erroneously included?

b) Would it be permissible to list as an alternative an instance type with slightly more CPU and/or RAM, accepting that the extra CPU and/or RAM will not be recognized/utilized/exploited by the CA's scheduler? For example, permitting the `C5n` family to be used as an alternative for the `C5` family? If so, then the documentation language should be changed from "the same amount" and "mismatched" to language that makes it clear that larger alternatives are acceptable. And if not, then the documentation should clarify that larger alternatives are unacceptable, because at least to this naive perspective of somebody unfamiliar with the specifics of the scheduling algorithm, it seems as though it should be fine.

ari-becker:
Paging @drewhemm , who wrote that section of the documentation

drewhemm:
Hi @ari-becker ,

Perhaps the `i3` family is not the best example, but personally I have not had any issues with them. It depends on the use case in question, particularly around resource requests...

Typically CA will just add more nodes until the current requests are satisfied. However, there is a theoretical edge case where if a request is for 62GB of RAM and CA adds `i3.2xlarge` instances, then the request may never be satisfied. I have not personally tested this edge case, but `i3.2xlarge` remains a "permissible alternative" in many cases. 

[...]

PR #3198, opened 2020-06-05
AWS: Update documentation

otterley:
Update AWS README document for:

1. Clarity
2. Document fully the use of tags on Auto Scaling Groups to advertise resources to Cluster Autoscaler
3. Emphasize the use of IAM Roles for Service accounts where available for security
4. Fix some grammatical/spelling errors
5. Wrap lines

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/README.md
+++ b/cluster-autoscaler/cloudprovider/aws/README.md
@@ -1,14 +1,22 @@
 # Cluster Autoscaler on AWS
-The cluster autoscaler on AWS scales worker nodes within any specified autoscaling group. It will run as a `Deployment` in your cluster. This README covers the steps required to configure and run the cluster autoscaler.
+On AWS, Cluster Autoscaler utilizes Amazon EC2 Auto Scaling Groups to manage node
+groups. Cluster Autoscaler typically runs as a `Deployment` in your cluster.
 
-## Kubernetes Version
-Cluster autoscaler must run on v1.3.0 or greater.
+## Requirements
+Cluster Autoscaler requires Kubernetes v1.3.0 or greater.
 
 ## Permissions
-The pod running the cluster autoscaler will need access to certain resources and actions. If using AWS EKS it is recommend to attach the IAM policy to the cluster austoscaler pod using [IAM roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html). For non-EKS kubernetes clusters attaching the IAM policy to the NodeGroup is recommended instead of using AWS credentials directly unless you have special requirements.
+Cluster Autoscaler requires the ability to examine and modify EC2 Auto Scaling
+Groups. We recommend using [IAM roles for Service
+Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html)
+to associate the Service Account that the Cluster Autoscaler Deployment runs as
+with an IAM role that is able to perform these functions. If you are unable to
+use IAM Roles for Service Accounts, you may associate an IAM service role with
+the EC2 instance on which the Cluster Autoscaler pod runs.
 
-### Attach IAM policy to Service Account
-A minimum IAM policy would look like:
+### IAM Policy
+The following policy provides the minimum privileges necessary for Cluster Autoscaler to run:
+
 ```json
 {
     "Version": "2012-10-17",
@@ -22,26 +30,35 @@ A minimum IAM policy would look like:
                 "autoscaling:SetDesiredCapacity",
                 "autoscaling:TerminateInstanceInAutoScalingGroup"
             ],
-            "Resource": "*"
+            "Resource": ["*"]
         }
     ]
 }
 ```
-If you'd like to scale node groups from 0, an
-`autoscaling:DescribeLaunchConfigurations` or
-`ec2:DescribeLaunchTemplateVersions` permission is required depending on if you
-made your ASG with Launch Configuration or Launch Template.
 
-If you'd like the cluster autoscaler to [automatically
-discover](#auto-discovery-setup) EC2 AutoScalingGroups, the
-`autoscaling:DescribeTags` permission is also required.
+If you'd like Cluster Autoscaler to [automatically
+discover](#auto-discovery-setup) EC2 Auto Scaling Groups **(recommended)**, add
+`autoscaling:DescribeTags` to the `Action` list. Also add
+`autoscaling:DescribeLaunchConfigurations` (if you created your ASG using a
+Launch Configuration) and/or `ec2:DescribeLaunchTemplateVersions` (if you
+created your ASG using a Launch Template) to the `Action` list.
 
-**NOTE**: You can restrict the target resources for the autoscaling actions by
-specifying autoscaling group ARNS. More information can be found
+If you prefer, you can restrict the target resources for the autoscaling actions
+by specifying Auto Scaling Group ARNs in the `Resource` list of the policy. More
+information can be found
 [here](https://docs.aws.amazon.com/autoscaling/latest/userguide/control-access-using-iam.html#policy-auto-scaling-resources).
 
 ### Using AWS Credentials
-For on premise users wishing to scale out to AWS, the above approach of attaching policy to a nodegroup role won't work. Instead, you can create an aws secret manually and add following environment variables to cluster-autoscaler deployment manifest. Cluster autoscaler will use credential to authenticate and authorize itself. Please make sure your role has above permissions.
+
+**NOTE** The following is not recommended for Kubernetes clusters running on
+AWS. If you are using Amazon EKS, consider using [IAM roles for Service
+Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html)
+instead.
+
+For on-premise clusters, you may create an IAM user subject to the above policy
+and provide the IAM credentials as environment variables in the Cluster
+Autoscaler deployment manifest. Cluster Autoscaler will use these credentials to
+authenticate and authorize itself.
 
 ```yaml
 apiVersion: v1
@@ -53,7 +70,9 @@ data:
   aws_access_key_id: BASE64_OF_YOUR_AWS_ACCESS_KEY_ID
   aws_secret_access_key: BASE64_OF_YOUR_AWS_SECRET_ACCESS_KEY
 ```
-Please check [guidance](https://kubernetes.io/docs/concepts/configuration/secret/#creating-a-secret-manually) for creating a secret manually.
+Please refer to the [relevant Kubernetes
+documentation](https://kubernetes.io/docs/concepts/configuration/secret/#creating-a-secret-manually)
+for creating a secret manually.
 
 ```yaml
 env:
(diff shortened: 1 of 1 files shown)


arhea:
Is it worth mentioning how Cluster Autoscaler relates to Managed Node Groups (cluster autoscaler tags are pre-configured) and AWS Fargate (not needed)?

Jeffwan:
I was off last week and I will review it by end of day. Thanks for contribution!

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -22,26 +32,35 @@ A minimum IAM policy would look like:
...
-discover](#auto-discovery-setup) EC2 AutoScalingGroups, the
-`autoscaling:DescribeTags` permission is also required.
+If you'd like Cluster Autoscaler to [automatically
+discover](#auto-discovery-setup) EC2 Auto Scaling Groups **(recommended)**, add
+`autoscaling:DescribeTags` to the `Action` list.
 
-**NOTE**: You can restrict the target resources for the autoscaling actions by
-specifying autoscaling group ARNS. More information can be found
+In addition, we also recommend adding `autoscaling:DescribeLaunchConfigurations`
+(if you created your ASG using a Launch Configuration) and/or
```
Jeffwan, line 46: This is confusing. usually, user just need LC or LT permission

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -71,153 +92,158 @@ env:
   value: YOUR_AWS_REGION
 ```
 
-## Deployment Specification
-Auto-Discovery Setup is always preferred option to avoid multiple, potentially different configuration for min/max values. If you want to adjust minimum and maximum size of the group, please adjust size on ASG directly, CA will fetch latest change when talking to ASG.
+## Auto-Discovery Setup
```
Jeffwan, line 95: Do we need to make this change?  Probably change to following makes more sense? They should belong to one header 2

Deployment Specification
- Auto-Discovery
- Manual Configuration
  - One ASG
  - Multiple ASG
- Master Node Setup

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -71,153 +92,158 @@ env:
...
+`<resource-name>` is the name of the resource, such as `ephemeral-storage`. The
+value of each tag specifies the amount of resource provided. The units are
+identical to the units used in the `resources` field of a Pod specification.
 
-```
-kubectl apply -f examples/cluster-autoscaler-autodiscover.yaml
-```
+Example tags:
 
-## Scaling a node group to 0
```
Jeffwan, line 112: This is only being used by `scale from 0` case. Why do you delete header?

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -71,153 +92,158 @@ env:
...
+kubectl apply -f examples/cluster-autoscaler-autodiscover.yaml
 ```
 
-### Auto-Discovery Setup
+Cluster Autoscaler will respect the minimum and maximum values of each Auto
+Scaling Group. It will will only adjust the desired value.
 
-To run a cluster-autoscaler which auto-discovers ASGs with nodes use the `--node-group-auto-discovery` flag. For example, `--node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,k8s.io/cluster-autoscaler/<YOUR CLUSTER NAME>` will find the ASGs where those tag keys
-_exist_. It does not matter what value the tags have.
+Each Auto Scaling Group should be comprised of instance types that provide
```
Jeffwan, line 114: Even mixed instance policy is supported. This is not something we recommend users to use. It never guarantee all the simulation is accurate because if instances ASG bring up are unknown to CA. I don't recommend we talk about this here. It would be better to move to MixedInstancePolicy section. The best practice is still one ASG -> one Launch Template -> One Instance type.

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -71,153 +92,158 @@ env:
...
-                "autoscaling:SetDesiredCapacity",
-                "autoscaling:TerminateInstanceInAutoScalingGroup",
-                "ec2:DescribeLaunchTemplateVersions"
-            ],
-            "Resource": "*"
-        }
-    ]
-}
-```
+Cluster Autoscaler < 1.15: `cloud.google.com/gke-accelerator=<gpu-type>`
```
Jeffwan, line 175: Need to change line here.

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -241,30 +267,89 @@ spec:
...
-- By default, cluster autoscaler will wait 10 minutes between scale down operations, you can adjust this using the `--scale-down-delay-after-add`, `--scale-down-delay-after-delete`, and `--scale-down-delay-after-failure` flag. E.g. `--scale-down-delay-after-add=5m` to decrease the scale down delay to 5 minutes after a node has been added.
-- If you're running multiple ASGs, the `--expander` flag supports three options: `random`, `most-pods` and `least-waste`. `random` will expand a random ASG on scale up. `most-pods` will scale up the ASG that will schedule the most amount of pods. `least-waste` will expand the ASG that will waste the least amount of CPU/MEM resources. In the event of a tie, cluster autoscaler will fall back to `random`.
-- If you're managing your own kubelets, they need to be started with the `--provider-id` flag. The provider id has the format `aws:///<availability-zone>/<instance-id>`, e.g. `aws:///us-east-1a/i-01234abcdef`.
-- If you want to use regional STS endpoints (e.g. when using VPC endpoint for STS) the env `AWS_STS_REGIONAL_ENDPOINTS=regional` should be set.
+* The `/etc/ssl/certs/ca-bundle.crt` should exist by default on ec2 instance in
```
Jeffwan, line 310: It's kind of hard for me to check line by line. Can you summaries what's your change? 
Basically, our principle is to minimize the changes..

otterley: It might be easier if you review the full README rather than review as a line by line patch, because I've completely reorganized the document for clarity. https://github.com/kubernetes/autoscaler/blob/3c046e7233d0d1dcb9cb46fd57f37c5dd189344e/cluster-autoscaler/cloudprovider/aws/README.md

cluster-autoscaler/cloudprovider/aws/README.md
otterley: I agree that using a single ASG with a single instance type is the simplest case for most customers, and we can come up with verbiage to emphasize that. However, we do recommend using Mixed Instance policies under several circumstances: (1) customers using Spot Instances, and (2) very large customers who benefit from instance diversity to address node shortages in certain regions/AZs. 

I disagree that a single ASG is "best practice" however, since we even support multiple ASGs with Managed Node Groups out of the box and encourage customers to use it where it makes sense for them.

cluster-autoscaler/cloudprovider/aws/README.md
otterley: The previous language suggested that if there were no pods scheduled on an ASG that there might be a reason for the ASG size not to scale down to zero, and that it could be corrected somehow by tagging.  Or, that scaling up from zero was an important special case. Having re-read the documentation and the source code a number of times, I'm not sure that scaling down to zero or up from zero is a special case. Am I mistaken?

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -71,153 +92,158 @@ env:
...
+Scaling Group. It will will only adjust the desired value.
 
-To run a cluster-autoscaler which auto-discovers ASGs with nodes use the `--node-group-auto-discovery` flag. For example, `--node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,k8s.io/cluster-autoscaler/<YOUR CLUSTER NAME>` will find the ASGs where those tag keys
-_exist_. It does not matter what value the tags have.
+Each Auto Scaling Group should be comprised of instance types that provide
+approximately equal capacity. For example, ASG "xlarge" could be comprised of
+m5a.xlarge, m4.xlarge, m5.xlarge, and m5d.xlarge instance types, because each of
+those provide 4 vCPUs and 16GiB RAM. Separately, ASG "2xlarge" could be
+comprised of m5a.2xlarge, m4.2xlarge, m5.2xlarge, and m5d.2xlarge instance
+types, because each of those provide 8 vCPUs and 32GiB RAM.
```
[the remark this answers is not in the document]
otterley: s/comprised of/composed of/g

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -71,153 +92,158 @@ env:
...
+provided in an ASG's Mixed Instances Policy. If any such overrides are found,
+only the first instance type found will be used.  See [Using Mixed Instances
+Policies and Spot Instances](#Using-Mixed-Instances-Policies-and-Spot-Instances)
+for details.
 
-* It is recommended to use a second tag like `k8s.io/cluster-autoscaler/<YOUR CLUSTER NAME>` when `k8s.io/cluster-autoscaler/enabled` is used across many clusters to prevent ASGs from different clusters recognized as the node groups
-* There are no `--nodes` flags passed to cluster-autoscaler because the node groups are automatically discovered by tags
-* No min/max values are provided when using Auto-Discovery, cluster-autoscaler will respect the current min and max values of the ASG being targeted, and it will adjust only the "desired" value.
+From version 1.14, Cluster Autoscaler can also determine the resources provided
+by each Auto Scaling group via tags. The tag is of the format
```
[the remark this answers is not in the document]
otterley: Fixed

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -71,153 +92,158 @@ env:
...
-```json
-{
-    "ResourceType": "auto-scaling-group",
-    "ResourceId": "foo.example.com",
-    "PropagateAtLaunch": true,
-    "Value": "bar",
-    "Key": "k8s.io/cluster-autoscaler/node-template/label/foo"
-}
-```
+Notes:
```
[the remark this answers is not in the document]
otterley: Fixed

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -71,153 +92,158 @@ env:
...
-        }
-    ]
-}
-```
+Cluster Autoscaler < 1.15: `cloud.google.com/gke-accelerator=<gpu-type>`
+Cluster Autoscaler >= 1.15: `k8s.amazonaws.com/accelerator=<gpu-type>`
 
-### Gotchas
+`<gpu-type>` varies by instance type. On P2 instances, for example, the
+value is `nvidia-tesla-k80`.
```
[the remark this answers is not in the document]
otterley: That's a good question. I lack this knowledge, personally.

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -1,14 +1,24 @@
...
+## Requirements
+Cluster Autoscaler requires Kubernetes v1.3.0 or greater.
 
 ## Permissions
-The pod running the cluster autoscaler will need access to certain resources and actions. If using AWS EKS it is recommend to attach the IAM policy to the cluster austoscaler pod using [IAM roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html). For non-EKS kubernetes clusters attaching the IAM policy to the NodeGroup is recommended instead of using AWS credentials directly unless you have special requirements.
+Cluster Autoscaler requires the ability to examine and modify EC2 Auto Scaling
+Groups. We recommend using [IAM roles for Service
+Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html)
+to associate Cluster Autoscaler with an IAM role that is able to perform these
+functions. If you are unable to use IAM Roles for Service accounts, you may
```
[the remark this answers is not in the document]
otterley: Fixed

cluster-autoscaler/cloudprovider/aws/README.md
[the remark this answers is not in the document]
Jeffwan: This is reserved for future GPU optimization, user doesn't have to use `nvidia-tesla-k80` at this moment, any identifier will work. While, we plan to ask MNG to apply these labels by default.

cluster-autoscaler/cloudprovider/aws/README.md
Jeffwan: Mixed Instance policy is not mature enough as a solution and user needs to take the risk and have to know all prerequisites to use it. It's just one presentation of spot, there're group of users using spot separately without MIP. I may not explain it clearly, my point is single instance per ASG is best practice, not single ASG. You can definitely has more ASGs/node groups.

otterley:
Resolves #2786

ari-becker:
> Resolves #2786

Yeah, this PR would certainly do this. Thanks @otterley - looks fantastic to me.

Not that it matters so much for this PR, but I would draw your attention to the partial solution (partial because it doesn't cover every EC2 instance type, only the ones we use) we wrote in Dhall and open-sourced for the problem of trying to compute the correct list of instances that can be put into a Mixed Instance Policy. This was essential to allow downstream to specify a desired image type, without having to provide all of the potential alternatives as well.

https://github.com/coralogix/dhall-aws/blob/55db0fb33d95e8a5e99b943bda1f74d23c388bba/ec2/InstanceType.dhall#L301

It surprisingly turned out to be one of the more profitable uses of our time - we prefer to use a lowest-cost allocation policy, and we would've never considered adding instance families like `m5dn` as their on-demand prices are higher than corresponding `m5` instances. Apparently, most of AWS's users suffer the same fallacy, and we found AWS launching `m5dn` spot instances which, surprisingly, were sometimes cheaper than `m5` spot instances. So, I applaud explicitly pointing out that instances like `m5d` can be added alongside `m5`.

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -71,153 +92,158 @@ env:
...
 
-### One ASG Setup (min: 1, max: 10, ASG Name: k8s-worker-asg-1)
-```
-kubectl apply -f examples/cluster-autoscaler-one-asg.yaml
-```
+To enable this, provide the `--node-group-auto-discovery` flag as an argument
+whose value is a list of tag keys that should be looked for. For example,
+`--node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,k8s.io/cluster-autoscaler/<cluster-name>`
+will find the ASGs where those tag keys _exist_. It does not matter what value
+the tags have.
```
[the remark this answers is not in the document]
dwaiba: This should also mention that launch templates are not supported but only node groups.
 autoscaler ups fine until v1.16.5 ups fine on EKS 1.16.8 with mixed policy for launch template but post deployment would throw
`I0617 10:22:37.411836       1 static_autoscaler.go:147] Starting main loop
I0617 10:22:37.412013       1 utils.go:626] No pod using affinity / antiaffinity found in cluster, disabling affinity predicate for this loop
I0617 10:22:37.412026       1 static_autoscaler.go:303] Filtering out schedulables
I0617 10:22:37.412147       1 static_autoscaler.go:320] No schedulable pods
I0617 10:22:37.412165       1 static_autoscaler.go:328] No unschedulable pods
I0617 10:22:37.412177       1 static_autoscaler.go:375] Calculating unneeded nodes
I0617 10:22:37.412188       1 utils.go:574] Skipping ip-10-1-2-40.ec2.internal - no node group config
I0617 10:22:37.412197       1 utils.go:574] Skipping ip-10-2-3-111.ec2.internal - no node group config
I0617 10:22:37.412204       1 utils.go:574] Skipping ip-10-1-2-135.ec2.internal - no node group config
I0617 10:22:37.412212       1 utils.go:574] Skipping ip-10-0-1-177.ec2.internal - no node group config
I0617 10:22:37.412219       1 utils.go:574] Skipping ip-10-2-3-143.ec2.internal - no node group config
I0617 10:22:37.412226       1 utils.go:574] Skipping ip-10-0-1-61.ec2.internal - no node group config
`

Jeffwan:
/lgtm

@jaypipes any other feedbacks or concerns?

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -1,14 +1,22 @@
 # Cluster Autoscaler on AWS
-The cluster autoscaler on AWS scales worker nodes within any specified autoscaling group. It will run as a `Deployment` in your cluster. This README covers the steps required to configure and run the cluster autoscaler.
+On AWS, Cluster Autoscaler utilizes Amazon EC2 Auto Scaling Groups to manage node
+groups. Cluster Autoscaler runs as a `Deployment` in your cluster.
```
ellistarn, line 3: Theoretically, couldn't it be deployed elsewhere? I might soften this language to "Cluster Autoscaler can be easily run"

otterley: How about "typically runs"? I haven't heard of any situations in which a customer might run it elsewhere; have you?

Jeffwan:
@otterley Can you check why you have code changes in this PR? There're lots of dependency conflicts.I assume this is only doc change.

Jeffwan:
@otterley One more thing, could you squash 3 commits to 1 commit? This repo doesn't use squash based. One commit for one change makes commit history clean. Thanks!

otterley:
@jeffwan Done!

cluster-autoscaler/cloudprovider/aws/README.md
ellistarn: You could imagine running it on the kubernetes master or in a cloud provider's control plane :)

PR #3200, opened 2020-06-05
DigitalOcean: do not use provider ID for node ID

timoreimann:
The DO API requires the proper node ID and otherwise returns an error. Thus, we can localize the error.


MorrisLaw:
LGTM

PR #3205, opened 2020-06-07
AWS: Use smallest instance type found in MixedInstancesPolicy

otterley:
When determining the capacity of an Auto Scaling Group that has a Mixed Instances Policy, use the smallest instance of all the instance types in it. Otherwise, we run the risk of launching an instance that is too small for a pending pod to actually fit in.

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_manager.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_manager.go
@@ -331,9 +331,11 @@ func (m *AwsManager) buildInstanceType(asg *asg) (string, error) {
 	} else if asg.LaunchTemplate != nil {
 		return m.ec2Service.getInstanceTypeByLT(asg.LaunchTemplate)
 	} else if asg.MixedInstancesPolicy != nil {
-		// always use first instance
-		if len(asg.MixedInstancesPolicy.instanceTypesOverrides) != 0 {
-			return asg.MixedInstancesPolicy.instanceTypesOverrides[0], nil
+		// Use smallest instance of all the instance types. Otherwise, we run
+		// the risk of launching an instance that is too small for the pending
+		// pod to actually fit in.
+		if len(asg.MixedInstancesPolicy.instanceTypesOverrides) > 0 {
+			return smallestInstanceType(asg.MixedInstancesPolicy.instanceTypesOverrides), nil
 		}
 
 		return m.ec2Service.getInstanceTypeByLT(asg.MixedInstancesPolicy.launchTemplate)
--- a/cluster-autoscaler/go.mod-extra
+++ b/cluster-autoscaler/go.mod-extra
@@ -3,7 +3,6 @@ go 1.12
 
 require (
   github.com/rancher/go-rancher v0.1.0
-  github.com/google/go-querystring v1.0.0
   github.com/aws/aws-sdk-go v1.23.12
 )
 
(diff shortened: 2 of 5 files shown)


cluster-autoscaler/hack/update-vendor.sh
```diff
@@ -17,7 +17,7 @@ fi
 
 SCRIPT_NAME=$(basename "$0")
 K8S_FORK="git@github.com:kubernetes/kubernetes.git"
-K8S_REV="master"
+K8S_REV="release-1.16"
```
Jeffwan, line 20: This can be overrided from argument. You don't need to make this change.

cluster-autoscaler/go.mod-extra
```diff
@@ -3,7 +3,6 @@ go 1.12
 
 require (
   github.com/rancher/go-rancher v0.1.0
-  github.com/google/go-querystring v1.0.0
```
Jeffwan, line 6: I don't see any pkg deletion. Do you needs this change?

cluster-autoscaler/cloudprovider/aws/aws_manager.go
```diff
@@ -331,9 +331,11 @@ func (m *AwsManager) buildInstanceType(asg *asg) (string, error) {
 	} else if asg.LaunchTemplate != nil {
 		return m.ec2Service.getInstanceTypeByLT(asg.LaunchTemplate)
 	} else if asg.MixedInstancesPolicy != nil {
-		// always use first instance
-		if len(asg.MixedInstancesPolicy.instanceTypesOverrides) != 0 {
-			return asg.MixedInstancesPolicy.instanceTypesOverrides[0], nil
+		// Use smallest instance of all the instance types. Otherwise, we run
+		// the risk of launching an instance that is too small for the pending
+		// pod to actually fit in.
+		if len(asg.MixedInstancesPolicy.instanceTypesOverrides) > 0 {
```
Jeffwan, line 337: em. I understand the point that CA uses smallest instance for simulation. The problem is most of the time, ASG will spawn new instances based on override sequence order. If the instance is not even close, that would be a big problem. 

One case is c5.4xlarge, c5.2xlarge, c5.xlarge. It will use c5.xlarge as template, it probably spin up 4 instances. however, one c5.4xlarge can fit all pending pods.

otterley: Sorry, I'm not understanding. Why would it be a problem?

First, we do make it pretty clear in the README PR (#3198) that customers shouldn't be placing instance types of such vastly different sizes in the same ASG. 

But let's suppose for the sake of argument that they do; and a c5.4xlarge comes up. But so what? The pending pod fits, so there's no harm. The worst case is that there could be a bit of temporary waste in spinning up more instances than is strictly needed, but in that case, the excess instances will eventually be scaled in.

The question, I think, is what do we care about more: wasted spend or a pod not fitting? Should we allow customers to specify that preference? And what should we do in the case of a pod not fitting? I think preventing scheduling delays is more important.

Also, the example you gave had specific ordering. The result would be different if it were c5.xlarge, c5.2xlarge, and c5.4xlarge. The fact that it's order depending is a pretty strong argument in favor of merging, because otherwise it makes the situation even harder to reason about.

cluster-autoscaler/hack/update-vendor.sh
otterley: I did this to replicate the work already done in the 1.15 branch (see #2971). This is a PR against the 1.16 branch, not master.

cluster-autoscaler/go.mod-extra
otterley: When I ran `hack/update-vendor.sh`, it said that this was removed. I had to remove it to get the script to finish successfully.

mwielgus:
Inactivity warning. Please proceed with the PR or it will be closed.

mwielgus:

PR #3122, opened 2020-05-09
Add history resolution and use prometheus client

yashbhutwala:
Follow-up from PR #2549; only includes:
- [X] add history resolution
- [X] use official prometheus client; remove previous client

PR #3206, opened 2020-06-08
port from PR 2551; shard prometheus history queries by day

### LINKED ISSUE #2543: vpa-recommender: Prometheus history provider doesn't work with large Prometheus/Kubernetes deployments

yashbhutwala:
Now that #3122 is merged, this is another follow-up PR from #2551  

Resolves #2543

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
@@ -19,7 +19,7 @@ package history
 import (
 	"context"
 	"fmt"
-	"k8s.io/klog"
+	"math"
 	"sort"
 	"strings"
 	"time"
@@ -29,7 +29,11 @@ import (
 	prommodel "github.com/prometheus/common/model"
 
 	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/recommender/model"
+	"k8s.io/klog"
 )
+
+// Day is a shortcut for the Prometheus-defined day (which is always 24 hours, unlike real days
+const Day = 24 * time.Hour
 
 // PrometheusHistoryProviderConfig allow to select which metrics
 // should be queried to get real resource utilization.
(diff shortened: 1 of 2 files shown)


estahn:
@yashbhutwala Have you had a chance to review @bskiba comments?

povilasv (commented):
I feel like this shouldn't live in VPA repo. 

There are proxies that do this for you also cache results and do other nice things. No need to reimplement this as a client library in VPA ;) 

Example:
https://thanos.io/components/query-frontend.md/
https://cortexmetrics.io/docs/configuration/prometheus-frontend/

cliles:
> I feel like this shouldn't live in VPA repo.
> 
> There are proxies that do this for you also cache results and do other nice things. No need to reimplement this as a client library in VPA ;)
> 
> Example:
> https://thanos.io/components/query-frontend.md/
> https://cortexmetrics.io/docs/configuration/prometheus-frontend/

In the current state the prometheus integration is not feature rich and has multiple issues when deployed in a large environment. I would hope that by introducing additional features we could solve some of the outstanding issues without adding another later of infrastructure. 

If the intent is that the VPA should work out of the box then this feature seems useful as it makes the client more intelligent and resilient.

povilasv:
> > I feel like this shouldn't live in VPA repo.
> > There are proxies that do this for you also cache results and do other nice things. No need to reimplement this as a client library in VPA ;)
> > Example:
> > https://thanos.io/components/query-frontend.md/
> > https://cortexmetrics.io/docs/configuration/prometheus-frontend/
> 
> In the current state the prometheus integration is not feature rich and has multiple issues when deployed in a large environment. I would hope that by introducing additional features we could solve some of the outstanding issues without adding another later of infrastructure.
> 
> If the intent is that the VPA should work out of the box then this feature seems useful as it makes the client more intelligent and resilient.

Why more resilient? I think less resilient, u split the query to multiple queries, means more of them can fail as they hit back-end. 
On large environments u should definitely add some caching proxy like I mentioned above.

I feel having thick custom made client in here is not a right place

mwielgus:
Closing due to inactivity. Feel free to reopen if/once needed.

PR #3210, opened 2020-06-08
rename and flip return value for actOnEmptyCluster

dwradcliffe:
This conditional was confusing and the comment for the method was wrong. This shouldn't change the behavior but makes the code easier to reason about.

The "new" method `shouldActOnEmptyCluster` will now return `true` if the autoscaler should continue to evaluate and scale up. If ScaleUpFromZero is disabled or if the cluster is not really empty, it will return false.

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -239,7 +239,7 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 		return errors.ToAutoscalerError(errors.ApiCallError, err)
 	}
 
-	if a.actOnEmptyCluster(allNodes, readyNodes, currentTime) {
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -719,23 +719,23 @@ func (a *StaticAutoscaler) obtainNodeLists(cp cloudprovider.CloudProvider) ([]*a
 	return allNodes, readyNodes, nil
 }
 
-// actOnEmptyCluster returns true if the cluster was empty and thus acted upon
-func (a *StaticAutoscaler) actOnEmptyCluster(allNodes, readyNodes []*apiv1.Node, currentTime time.Time) bool {
```
mwielgus, line 723: The behavior is still confusing.  Now `shouldActOnEmptyCluster` returns true if the cluster is not empty.

mwielgus (changes requested):

mwielgus:

PR #3221, opened 2020-06-15
switch scalesets to delete asynchronously without waiting on future

marwanad:
Splits `DeleteInstances` into an operation call + async wait so we don't block main thread when the main loop is deleting unregistered instances.

/area provider/azure


marwanad:
test failures are on the AWS side and unrelated

PR #3234, opened 2020-06-22
Added a make target to support immutable images when developing

ellistarn:
Added a basic make target that automatically dates and patches images while developing.

1) Some repositories recommend immutable tags. Requiring a fixed tag like "dev" creates eventual consistency challenges with remote storage. 
2) Automatically patches cluster autoscaler with the new image.


ellistarn:
I signed the CLA!

PR #3254, opened 2020-06-30
full vpa e2e - bumping up cpu request tolerance

### LINKED ISSUE #3249: Full vpa e2e test is flaky
Full vpa test is flaky.
https://testgrid.k8s.io/sig-autoscaling-vpa#autoscaling-vpa-full

The main contributor is "Pods under VPA have cpu requests growing with usage" (both v1 and v1beta2).
Recommender is epxected to provide recommendation between 500 and 900[mCPU], however it often is above 900.

krzysied:
Bumping up upper limit for cpu recommendation in "Pods under VPA have cpu requests growing with usage" test case.

/fixes #3249

### DIFF
--- a/vertical-pod-autoscaler/e2e/v1/full_vpa.go
+++ b/vertical-pod-autoscaler/e2e/v1/full_vpa.go
@@ -101,7 +101,7 @@ var _ = FullVpaE2eDescribe("Pods under VPA", func() {
 		rc.ConsumeCPU(600 * replicas)
 		err = waitForResourceRequestInRangeInPods(
 			f, pollTimeout, metav1.ListOptions{LabelSelector: "name=hamster"}, apiv1.ResourceCPU,
-			ParseQuantityOrDie("500m"), ParseQuantityOrDie("900m"))
+			ParseQuantityOrDie("500m"), ParseQuantityOrDie("1000m"))
 		gomega.Expect(err).NotTo(gomega.HaveOccurred())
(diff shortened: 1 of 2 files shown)


vertical-pod-autoscaler/e2e/v1/full_vpa.go
```diff
@@ -101,7 +101,7 @@ var _ = FullVpaE2eDescribe("Pods under VPA", func() {
 		rc.ConsumeCPU(600 * replicas)
 		err = waitForResourceRequestInRangeInPods(
 			f, pollTimeout, metav1.ListOptions{LabelSelector: "name=hamster"}, apiv1.ResourceCPU,
-			ParseQuantityOrDie("500m"), ParseQuantityOrDie("900m"))
+			ParseQuantityOrDie("500m"), ParseQuantityOrDie("1000m"))
```
bskiba, line 104: That is 1, right?

krzysied: Yes.
Is this a suggestion to change it?

krzysied: IMHO is easier to read if we have the same units on each side of the range.

bskiba: IMHO 1000m looks weird, but I won't block.

PR #3261, opened 2020-07-01
Reduce instance lock scope in scale sets

marwanad:
Currently, `DeleteInstances` can take a lock on `instanceMutex` and get blocked on `GetAsgForInstance` if a `manager.regenerate` call happens and grab the `azure_cache` mutex lock. Both `DeleteInstances` and `Nodes` will then be deadlocked.

/area provider/azure


cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -228,14 +228,6 @@ func (scaleSet *ScaleSet) GetScaleSetSize() (int64, error) {
 }
 
 func (scaleSet *ScaleSet) waitForDeleteInstances(future *azure.Future, requiredIds *compute.VirtualMachineScaleSetVMInstanceRequiredIDs) {
-	var err error
```
marwanad, line 231: we already log an error with the same info, so this line is useless.

marwanad:
seems the cncf-cla label is stuck?

PR #3269, opened 2020-07-02
Update golang version used in docker builds

MaciekPytel:
This is a minor cleanup after https://github.com/kubernetes/autoscaler/pull/3267 to fix building in docker.


MaciekPytel:
The travis test blocking this has actually passed and just haven't updated the status for some reason. Force merging this as the project is currently not building.

PR #3277, opened 2020-07-04
Decrement curSize by the number of instances to be deleted

marwanad:
DeleteInstances can take in an array of instances to delete so we should decrement by the length of instances to be deleted.

Note: It's not really an issue now because all code paths used in CA send a single instance to Delete (except `deleteCreatedNodesWithErrors` which isn't used by the Azure provider) but just in case the signatures change in the future

/area provider/azure

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_scale_set_test.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_scale_set_test.go
@@ -59,16 +59,19 @@ func newTestVMSSList(cap int64, name, loc string) []compute.VirtualMachineScaleS
 	}
 }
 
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/cloudprovider/azure/azure_scale_set_test.go
```diff
@@ -427,8 +436,10 @@ func TestScaleSetNodes(t *testing.T) {
 	assert.NotNil(t, ss)
 	instances, err := group.Nodes()
 	assert.NoError(t, err)
-	assert.Equal(t, len(instances), 1)
```
marwanad, line 430: This test case had VMSS capacity as 3 and VMSS VMs as 1.

Updated to make both consistent as 3.

PR #3308, opened 2020-07-09
Fix priority expander falling back to random although higher priority matches

bruecktech:
When an option matches in a low priority together with another option both get added to the `best` list where the random expander fallback strategy finally selects one option randomly although one of the options also matches in a higher priority.
This PR ensures all priorities are evaluated for an option and higher ones take precedence by removing the `break` in the loop. The root cause is that the map of priorities is unordered and therefore we have to walk to the end

### DIFF
--- a/cluster-autoscaler/expander/priority/priority.go
+++ b/cluster-autoscaler/expander/priority/priority.go
@@ -148,7 +148,6 @@ func (p *priority) BestOption(expansionOptions []expander.Option, nodeInfo map[s
 			}
 			best = append(best, option)
 			found = true
-			break
 		}
 		if !found {
 			msg := fmt.Sprintf("Priority expander: node group %s not found in priority expander configuration. "+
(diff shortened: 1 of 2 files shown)


MaciekPytel:
/lgtm
/approve

Sorry for the long wait. Re-opening to trigger presubmits re-run (since the previous run was quite old).

MaciekPytel:
/cancel lgtm
I just realized this PR is to 1.16 and not master branch. The problem with this is that it will regress in any future version and so the way to fix this would be to make a PR to master and later cherry-pick it to all the way down through branches (ie. to 1.18, 1.17 and 1.16).

bruecktech:
@MaciekPytel I changed the PR to be against master now

nitrag:
This didn't make the 1.15.7 release :(

Jeffwan:
@nitrag Maybe we can bring to next release.

MaciekPytel:
/lgtm
/approve
I see no reason not to bring it to next round of patch releases. Please feel free to open cherry-pick PRs and I'm happy to accept them. As I mentioned before my only requirement is that it needs to be cherry-picked to all branches newer than the one you're targeting (ie. if you need it in 1.16 you should also cherry-pick to 1.17-1.19). The reason is to avoid regressions as people upgrade their cluster.

PR #3313, opened 2020-07-10
Cherry-pick: #2064 Use distroless base image for 1.15

marwanad:
Ref: #2064

Tested a built image on Azure but would love confirmation from other providers.


MaciekPytel:
That seems like a pretty large change to cherry-pick. Similarly to kubernetes itself we generally aim to only cherry-pick critical bugfixes. We've been quite relaxed about this policy when it comes to provider-specific changes (trusting owners to make their own calls), but this is a very significant change to core. On top of that I know there were multiple problems with migrating k8s itself itself to distroless and the change was reverted multiple times in main repo.

With all of that I think this one is too risky to cherry-pick, unless you have a strong reason why this is needed so badly to justify the risk.

marwanad:
That's reasonable. There's no absolute pushing need for that - mostly for the sake of addressing CVE if we're doing one final release of 1.15.

But yeah - I understand your point, people can always move to 1.16 if they need those CVEs fixed so we can close this since it could have the potential to break things.

PR #3314, opened 2020-07-13
[cluster-autoscaler][clusterapi] Add support for node autodiscovery to clusterapi provider

detiber:
Builds on #3312

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_autodiscovery.go
+++ b/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_autodiscovery.go
@@ -1,0 +1,99 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package clusterapi
+
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+
+	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
+)
+
+type clusterAPIAutoDiscoveryConfig struct {
+	clusterName   string
+	namespace     string
+	labelSelector labels.Selector
+}
+
+func parseAutoDiscoverySpec(spec string) (*clusterAPIAutoDiscoveryConfig, error) {
+	cfg := &clusterAPIAutoDiscoveryConfig{
+		labelSelector: labels.NewSelector(),
+	}
+
+	tokens := strings.Split(spec, ":")
+	if len(tokens) != 2 {
+		return cfg, errors.NewAutoscalerError(errors.ConfigurationError, fmt.Sprintf("spec \"%s\" should be discoverer:key=value,key=value", spec))
+	}
+	discoverer := tokens[0]
+	if discoverer != autoDiscovererTypeClusterAPI {
+		return cfg, errors.NewAutoscalerError(errors.ConfigurationError, fmt.Sprintf("unsupported discoverer specified: %s", discoverer))
+	}
+
+	for _, arg := range strings.Split(tokens[1], ",") {
+		if len(arg) == 0 {
+			continue
+		}
+		kv := strings.Split(arg, "=")
+		if len(kv) != 2 {
+			return cfg, errors.NewAutoscalerError(errors.ConfigurationError, fmt.Sprintf("invalid key=value pair %s", kv))
+		}
+		k, v := kv[0], kv[1]
+
+		switch k {
+		case autoDiscovererClusterNameKey:
+			cfg.clusterName = v
+		case autoDiscovererNamespaceKey:
+			cfg.namespace = v
+		default:
+			req, err := labels.NewRequirement(k, selection.Equals, []string{v})
+			if err != nil {
+				return cfg, errors.NewAutoscalerError(errors.ConfigurationError, fmt.Sprintf("failed to create label selector; %v", err))
+			}
+			cfg.labelSelector = cfg.labelSelector.Add(*req)
+		}
+	}
+	return cfg, nil
+}
+
+func parseAutoDiscovery(specs []string) ([]*clusterAPIAutoDiscoveryConfig, error) {
+	result := make([]*clusterAPIAutoDiscoveryConfig, 0, len(specs))
+	for _, spec := range specs {
+		autoDiscoverySpec, err := parseAutoDiscoverySpec(spec)
+		if err != nil {
+			return result, err
+		}
+		result = append(result, autoDiscoverySpec)
+	}
+	return result, nil
+}
+
+func allowedByAutoDiscoverySpec(spec *clusterAPIAutoDiscoveryConfig, r *unstructured.Unstructured) bool {
+	switch {
+	case spec.namespace != "" && spec.namespace != r.GetNamespace():
+		return false
+	case spec.clusterName != "" && spec.clusterName != clusterNameFromResource(r):
+		return false
+	case !spec.labelSelector.Matches(labels.Set(r.GetLabels())):
+		return false
+	default:
+		return true
+	}
+}
--- a/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
+++ b/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
@@ -37,6 +37,8 @@ import (
 	"k8s.io/client-go/scale"
 	"k8s.io/client-go/tools/cache"
 	klog "k8s.io/klog/v2"
+
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
 )
 
 const (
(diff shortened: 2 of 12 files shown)


cluster-autoscaler/cloudprovider/clusterapi/clusterapi_autodiscovery.go
```diff
@@ -0,0 +1,97 @@
...
+	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
+)
+
+type clusterAPIAutoDiscoveryConfig struct {
+	clusterName   string
+	namespace     string
+	labelSelector labels.Selector
+}
+
+func parseAutoDiscoverySpec(spec string) (*clusterAPIAutoDiscoveryConfig, error) {
```
detiber, line 36: Need to add tests around node autodiscovery parsing

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_autodiscovery.go
```diff
@@ -0,0 +1,97 @@
...
+		if err != nil {
+			return result, err
+		}
+		result = append(result, autoDiscoverySpec)
+	}
+	return result, nil
+}
+
+
+func allowedByAutoDiscoverySpec(spec *clusterAPIAutoDiscoveryConfig, r *unstructured.Unstructured) bool {
```
detiber, line 86: Need to add a test around this function

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
```diff
@@ -98,31 +107,20 @@ func indexNodeByProviderID(obj interface{}) ([]string, error) {
...
+}
 
-	return machine, nil
+func (c *machineController) findMachineDeployment(id string) (*unstructured.Unstructured, error) {
+	return c.findResourceByKey(c.machineDeploymentInformer.Informer().GetStore(), id)
 }
 
-func (c *machineController) findMachineDeployment(id string) (*MachineDeployment, error) {
-	item, exists, err := c.machineDeploymentInformer.Informer().GetStore().GetByKey(id)
+func (c *machineController) findResourceByKey(store cache.Store, key string) (*unstructured.Unstructured, error) {
```
detiber, line 122: Need to add test cases for this method to verify that autodiscovery config is doing the right thing.

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
```diff
@@ -428,190 +425,128 @@ func getAPIGroupPreferredVersion(client discovery.DiscoveryInterface, APIGroup s
...
 
-	return nodegroups, nil
+	klog.V(4).Infof("nodegroup %s has nodes %v", scalableResource.GetName(), providerIDs)
+
+	return providerIDs, nil
 }
 
-func (c *machineController) machineDeploymentNodeGroups() ([]*nodegroup, error) {
-	machineDeployments, err := c.listMachineDeployments(metav1.NamespaceAll, labels.Everything())
+func (c *machineController) nodeGroups() ([]*nodegroup, error) {
```
detiber, line 508: Need to add test cases for this method to verify that autodiscovery config is doing the right thing.

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
```diff
@@ -428,190 +425,128 @@ func getAPIGroupPreferredVersion(client discovery.DiscoveryInterface, APIGroup s
...
-		if err != nil {
-			return nil, err
-		}
-		machineSets = append(machineSets, machineDeployments...)
-	}
-
-	return machineSets, nil
-}
-
 func (c *machineController) nodeGroupForNode(node *corev1.Node) (*nodegroup, error) {
```
detiber, line 529: Need to add test cases for this method to verify that autodiscovery config is doing the right thing.

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
```diff
@@ -639,80 +574,109 @@ func (c *machineController) findNodeByProviderID(providerID normalizedProviderID
...
 	}
-	return newMachineDeploymentFromUnstructured(u.DeepCopy()), nil
+	return scalableResources, nil
 }
 
-func (c *machineController) listMachines(namespace string, selector labels.Selector) ([]*Machine, error) {
-	objs, err := c.machineInformer.Lister().ByNamespace(namespace).List(selector)
-	if err != nil {
-		return nil, err
+func (c *machineController) listResources(lister cache.GenericLister) ([]*unstructured.Unstructured, error) {
```
detiber, line 622: Need to add test cases for this method to verify that autodiscovery config is doing the right thing.

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
```diff
@@ -639,80 +574,109 @@ func (c *machineController) findNodeByProviderID(providerID normalizedProviderID
...
 
-	return machineSets, nil
+	return results, nil
 }
 
-func (c *machineController) listMachineDeployments(namespace string, selector labels.Selector) ([]*MachineDeployment, error) {
-	objs, err := c.machineDeploymentInformer.Lister().ByNamespace(namespace).List(selector)
-	if err != nil {
-		return nil, err
+func (c *machineController) allowedByAutoDiscoverySpecs(r *unstructured.Unstructured) bool {
```
detiber, line 668: Need to add test cases for this method to verify that autodiscovery config is doing the right thing.

elmiko:
nice Jason! i know we talked about the node discovery stuff, but i'm a little unfamiliar with the process. i'll give this a test drive.

detiber:
@elmiko The main goal was to try to leverage the existing functionality of node autodiscovery to be able to limit which MachineSets/MachineDeployments are allowed to be used for NodeGroups.

If they are not specified this falls back to the previous behavior of allowing all discovered MachineSets and MachineDeployments to be used, but we may want to reconsider that at somepoint in the future.

elmiko:
thanks @detiber, i appreciate the explanation =)

MaciekPytel:
Close + reopen to trigger travis rerun after fixing tests at HEAD.

ncdc:
Does this need to be rebased to drop the the commits from 3312?

detiber:
> Does this need to be rebased to drop the the commits from 3312?

Rebased

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
```diff
@@ -80,7 +86,7 @@ func indexMachineByProviderID(obj interface{}) ([]string, error) {
 		return nil, nil
 	}
 
-	providerID, found, err := unstructured.NestedString(u.Object, "spec", "providerID")
+	providerID, found, err := unstructured.NestedString(u.UnstructuredContent(), "spec", "providerID")
```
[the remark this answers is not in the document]
detiber: Not directly, I think I originally hit an issue while building out the tests where `u.Object` was nil, and it seemed like it would be good to guard against that in general.

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
```diff
@@ -497,21 +515,13 @@ func (c *machineController) nodeGroups() ([]*nodegroup, error) {
...
-		// add nodegroup iff it has the capacity to scale
-		if ng.MaxSize()-ng.MinSize() > 0 {
-			replicas, found, err := unstructured.NestedInt64(r.Object, "spec", "replicas")
-			if err != nil {
-				return nil, err
-			}
-
-			if found && replicas > 0 {
-				nodegroups = append(nodegroups, ng)
-			}
```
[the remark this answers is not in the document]
detiber: The size check has been pushed down into `newNodeGroupFromScalableResource()`, which will return `nil` if it does not have capacity to scale or has 0 replicas.

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
```diff
@@ -611,10 +621,31 @@ func (c *machineController) listScalableResources() ([]*unstructured.Unstructure
...
 func (c *machineController) listResources(lister cache.GenericLister) ([]*unstructured.Unstructured, error) {
-	return listResources(lister.ByNamespace(metav1.NamespaceAll), labels.Everything())
+	if len(c.autoDiscoverySpecs) == 0 {
+		return listResources(lister.ByNamespace(metav1.NamespaceAll), "", labels.Everything())
+	}
+
+	var results []*unstructured.Unstructured
+	tracker := map[string]bool{}
+	for _, spec := range c.autoDiscoverySpecs {
+		resources, err := listResources(lister.ByNamespace(spec.namespace), spec.clusterName, spec.labelSelector)
```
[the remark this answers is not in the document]
detiber: spec.namespace is a string, so it will default to "" when not set, which is equivalent to all namespaces when passed to a lister. For additional context, if one uses `metav1.NamespaceAll`, it is defined as the empty string: https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#pkg-constants

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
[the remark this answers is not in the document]
detiber: I'm happy to break this change out separately if you prefer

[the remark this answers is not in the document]
elmiko: hmm, i'm torn. on one hand it makes it easier to follow the unstructured stuff if we keep it separate, and on the other this is a small change to pick up a few more references.

honestly, my question was just to increase my understanding of the code here. i guess i'm ok to leave it here if no one else has issues.

elmiko:
hey @detiber , i finally got my test stand working again and gave this PR a drive today. everything seemed to work as expected. i tested the namespace and label options, attempting both inclusion and exclusion.

my one comment is that the node group autodiscovery stuff took me a few to understand that the constraints its looking for are on the MachineSet/MachineDeployment resources. i think you mention scalable resources in the readme, but i wonder if we should be more specific?

/approve

PR #3320, opened 2020-07-14
Add sample deployment/service account manifest for CAPI

benmoss:
Based on https://notes.elmiko.dev/2020/05/22/kubernetes-autoscaler-capd.html

@detiber @elmiko

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/examples/deployment.yaml
+++ b/cluster-autoscaler/cloudprovider/clusterapi/examples/deployment.yaml
@@ -1,0 +1,167 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: cluster-autoscaler
+  namespace: ${AUTOSCALER_NS}
+  labels:
+    app: cluster-autoscaler
+spec:
+  selector:
+    matchLabels:
+      app: cluster-autoscaler
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: cluster-autoscaler
+    spec:
+      containers:
+      - image: ${AUTOSCALER_IMAGE}
+        name: cluster-autoscaler
+        command:
+        - /cluster-autoscaler
+        args:
+        - --cloud-provider=clusterapi
+      serviceAccountName: cluster-autoscaler
+      terminationGracePeriodSeconds: 10
+      tolerations:
+      - effect: NoSchedule
+        key: node-role.kubernetes.io/master
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: cluster-autoscaler-workload
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-autoscaler-workload
+subjects:
+- kind: ServiceAccount
+  name: cluster-autoscaler
+  namespace: ${AUTOSCALER_NS}
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: cluster-autoscaler-management
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-autoscaler-management
+subjects:
+- kind: ServiceAccount
+  name: cluster-autoscaler
+  namespace: ${AUTOSCALER_NS}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: cluster-autoscaler
+  namespace: ${AUTOSCALER_NS}
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: cluster-autoscaler-workload
+rules:
+  - apiGroups:
+    - ""
+    resources:
+    - persistentvolumeclaims
+    - persistentvolumes
+    - pods
+    - replicationcontrollers
+    verbs:
+    - get
+    - list
+    - watch
+  - apiGroups:
+    - ""
+    resources:
+    - nodes
+    verbs:
+    - get
+    - list
+    - update
+    - watch
+  - apiGroups:
+    - ""
+    resources:
+    - pods/eviction
+    verbs:
+    - create
+  - apiGroups:
+    - policy
+    resources:
+    - poddisruptionbudgets
+    verbs:
+    - list
+    - watch
+  - apiGroups:
+    - storage.k8s.io
+    resources:
+    - csinodes
+    - storageclasses
+    verbs:
+    - get
+    - list
+    - watch
+  - apiGroups:
+    - batch
+    resources:
+    - jobs
+    verbs:
+    - list
+    - watch
+  - apiGroups:
+    - apps
+    resources:
+    - daemonsets
+    - replicasets
+    - statefulsets
+    verbs:
+    - list
+    - watch
+  - apiGroups:
+    - ""
+    resources:
+    - events
+    verbs:
+    - create
+    - patch
+  - apiGroups:
+    - ""
+    resources:
+    - configmaps
+    verbs:
+    - create
+    - delete
+    - get
+    - update
+  - apiGroups:
+    - coordination.k8s.io
+    resources:
+    - leases
+    verbs:
+    - create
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/cloudprovider/clusterapi/deployment.yaml
```diff
@@ -0,0 +1,158 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: cluster-autoscaler
+  namespace: kube-system
```
detiber, line 6: This should probably not be kube-system by default. It might be good to use a variable for this and expect users to run envsubst or similar against this prior to deploying.

cluster-autoscaler/cloudprovider/clusterapi/deployment.yaml
```diff
@@ -0,0 +1,158 @@
...
+        command:
+        - /cluster-autoscaler
+        args:
+        - --cloud-provider=clusterapi
+        - --namespace=default
+        - --scale-down-enabled=true
+        - --scale-down-delay-after-add=10s
+        - --scale-down-delay-after-delete=10s
+        - --scale-down-delay-after-failure=10s
+        - --scale-down-unneeded-time=23s
```
elmiko, line 31: we might want to ease these scale down times back a little. i have used these aggressive timings for demonstration purposes but they are not suited for production. imo, we could drop these in favor of the default values (which are in the minutes), or perhaps change just put the defaults here so people can see them.

detiber, line 31: I believe we should probably use a variable for these as well, since these will likely need to be tweaked for "real world" type of environments.

cluster-autoscaler/cloudprovider/clusterapi/deployment.yaml
```diff
@@ -0,0 +1,158 @@
...
+        - --balance-similar-node-groups
+        - --max-node-provision-time=10m
```
detiber, line 33: I believe we should probably use a variable for this as well, since this will likely need to be tweaked depending on which provider is being used for Cluster API.

cluster-autoscaler/cloudprovider/clusterapi/deployment.yaml
```diff
@@ -0,0 +1,158 @@
...
+        - --expendable-pods-priority-cutoff=-10
+        - --max-nodes-total=24
```
detiber, line 35: I believe we should probably use a variable for this as well, since this will likely need to be tweaked depending on the scale of workload clusters a user might be deploying.

cluster-autoscaler/cloudprovider/clusterapi/deployment.yaml
```diff
@@ -0,0 +1,158 @@
...
+  name: cluster-autoscaler
+  namespace: kube-system
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: cluster-autoscaler
+  namespace: kube-system
+---
+kind: ClusterRole
```
detiber, line 61: Can you break this into two separate ClusterRoles? One for workload cluster resources and one for management cluster resources?

That will make it simpler to break out into separate manifests for each use case once we have support for connecting to separate workload and management clusters.

cluster-autoscaler/cloudprovider/clusterapi/deployment.yaml
```diff
@@ -0,0 +1,158 @@
...
+    matchLabels:
+      app: cluster-autoscaler
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: cluster-autoscaler
+    spec:
+      containers:
+      - image: us.gcr.io/k8s-artifacts-prod/autoscaling/cluster-autoscaler:v1.18.1
```
detiber, line 20: I'm not quite sure how we want to handle it yet, but we probably want to consider taking into account for overriding either the image repository/name and the version tag (or both), since users may want to override the registry/image name for testing non-released versions and the version tag will need to correspond to the same minor kubernetes release that is running on the managed workload cluster.

elmiko:
+1 to variable usage where possible, nice suggestion Jason.

cluster-autoscaler/cloudprovider/clusterapi/deployment.yaml
elmiko: i like this, great idea

cluster-autoscaler/cloudprovider/clusterapi/deployment.yaml
elmiko: this makes way too much sense ;)

benmoss:
I removed the default flags, moved the file to under `cloudprovider/clusterapi/examples`, and split the `ClusterRoles` into management cluster permissions and workload cluster permissions. I just guessed about which permissions belong to which cluster though.

cluster-autoscaler/cloudprovider/clusterapi/examples/deployment.yaml
```diff
@@ -0,0 +1,167 @@
...
+metadata:
+  name: cluster-autoscaler-workload
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-autoscaler-workload
+subjects:
+- kind: ServiceAccount
+  name: cluster-autoscaler
+  namespace: kube-system
```
detiber, line 43:
```suggestion
  namespace: ${AUTOSCALER_NS}
```

cluster-autoscaler/cloudprovider/clusterapi/examples/deployment.yaml
```diff
@@ -0,0 +1,167 @@
...
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: cluster-autoscaler
+  namespace: kube-system
```
detiber, line 62:
```suggestion
  namespace: ${AUTOSCALER_NS}
```

cluster-autoscaler/cloudprovider/clusterapi/examples/deployment.yaml
```diff
@@ -0,0 +1,167 @@
...
+metadata:
+  name: cluster-autoscaler-management
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-autoscaler-management
+subjects:
+- kind: ServiceAccount
+  name: cluster-autoscaler
+  namespace: kube-system
```
detiber, line 56:
```suggestion
  namespace: ${AUTOSCALER_NS}
```

cluster-autoscaler/cloudprovider/clusterapi/examples/deployment.yaml
```diff
@@ -0,0 +1,167 @@
...
+metadata:
+  name: cluster-autoscaler-management
+rules:
+  - apiGroups:
+    - ""
+    resources:
+    - events
+    verbs:
+    - create
+    - patch
```
detiber, line 139: I believe this should be moved to `cluster-autoscaler-workload`

cluster-autoscaler/cloudprovider/clusterapi/examples/deployment.yaml
```diff
@@ -0,0 +1,167 @@
...
+  - apiGroups:
+    - ""
+    resources:
+    - configmaps
+    verbs:
+    - create
+    - delete
+    - get
+    - update
```
detiber, line 148: I believe this should be moved to cluster-autoscaler-workload

cluster-autoscaler/cloudprovider/clusterapi/examples/deployment.yaml
```diff
@@ -0,0 +1,167 @@
...
+  - apiGroups:
+    - coordination.k8s.io
+    resources:
+    - leases
+    verbs:
+    - create
+    - get
+    - update
```
detiber, line 156: I believe this should be moved to cluster-autoscaler-workload

cluster-autoscaler/cloudprovider/clusterapi/examples/deployment.yaml
benmoss: This was the one I was least sure of, but it makes sense

PR #3321, opened 2020-07-15
Gce pricing fixes

MaciekPytel:


MaciekPytel:
/hold
So it doesn't merge before I do some more e2e testing.

cluster-autoscaler/cloudprovider/gce/gce_price_model.go
```diff
@@ -283,9 +307,27 @@ func (model *GcePriceModel) NodePrice(node *apiv1.Node, startTime time.Time, end
 			price = price * preemptibleDiscount
 		}
 	}
-	// TODO: handle SSDs.
 
-	price += getAdditionalPrice(node.Status.Capacity, startTime, endTime)
```
aleksandra-malinowska, line 288: Do we want to keep the old logic in PodPrice()?

cluster-autoscaler/cloudprovider/gce/gce_price_model.go
```diff
@@ -283,9 +307,27 @@ func (model *GcePriceModel) NodePrice(node *apiv1.Node, startTime time.Time, end
...
+	// GPUs
+	if gpuRequest, found := node.Status.Capacity[gpu.ResourceNvidiaGPU]; found {
+		gpuPrice := gpuPricePerHour
+		if node.Labels != nil {
```
aleksandra-malinowska, line 314: nit: can we log sth about fallback to default pricing if there's no label? (or error if this isn't supposed to happen)

cluster-autoscaler/cloudprovider/gce/gce_price_model_test.go
```diff
@@ -29,73 +30,118 @@ import (
...
+			cheaperNode:                testNode(t, "custom", "custom-8", 8000, 32*units.GiB, "", 0, false),
+			expensiveNode:              testNode(t, "n1", "n1-standard-8", 8000, 32*units.GiB, "", 0, false),
+			priceComparisonCoefficient: 1.2,
+		},
+		"custom node price scales linearly": {
+			cheaperNode:                testNode(t, "small_custom", "custom-1", 1000, 3.75*units.GiB, "", 0, false),
+			expensiveNode:              testNode(t, "large_custom", "custom-8", 8000, 32*units.GiB, "", 0, false),
+			priceComparisonCoefficient: 1.0 / 7.9,
+		},
+		"custom node price scales linearly 2": {
```
aleksandra-malinowska, line 86: If I understand this test case correctly, it checks that `price(custom-1) < 8.1 * price(custom-8)`. I'm guessing cheaper/expensive node were supposed to be reversed in this test case?

cluster-autoscaler/cloudprovider/gce/gce_price_model_test.go
```diff
@@ -29,73 +30,118 @@ import (
...
+			expensiveNode:              testNode(t, "large_custom", "custom-8", 8000, 32*units.GiB, "", 0, false),
+			cheaperNode:                testNode(t, "small_custom", "custom-1", 1000, 3.75*units.GiB, "", 0, false),
+			priceComparisonCoefficient: 8.1,
+		},
+		// GPUs
+		"accelerators are expensive": {
+			cheaperNode: testNode(t, "no_accelerators", "n1-standard-8", 8000, 32*units.GiB, "", 0, false),
+			// #DadJoke #NotFunny
```
aleksandra-malinowska, line 94: nit: it feels more like a nerd joke. I'd remove #DadJoke hashtag, as it can be misleading.

cluster-autoscaler/cloudprovider/gce/gce_price_model_test.go
```diff
@@ -29,73 +30,118 @@ import (
...
+			expensiveNode:              testNode(t, "large hadron collider", "n1-standard-8", 8000, 32*units.GiB, "nvidia-tesla-v100", 1, false),
+			priceComparisonCoefficient: 2,
```
aleksandra-malinowska, line 96: This also looks like either the nodes could be swapped, or coefficient changed to 1.0/2.0

cluster-autoscaler/cloudprovider/gce/gce_price_model.go
MaciekPytel: I think so:
- It's much more difficult to tell what GPU a pod is requesting, since with NodeAffinity you can request any GPU from a given subset. What price should we assign in that case? 
- I don't think getting accurate pod price estimate is very important. If all expansion options have the same set of pods it doesn't matter at all. It's needed as a rough approximation for comparing options that help different set of pods, but even in this case its impact is likely limited in most cases (if some pods don't fit because a node is too small, preferred cluster shape part of expander formula is likely to determine result anyway). 
  - We already do something very similar for CPU - an e2 CPU and a c2 one have a very different price and pod is free to request whatever it wants via NodeAffinity.

cluster-autoscaler/cloudprovider/gce/gce_price_model_test.go
MaciekPytel: Yup, thanks for catching that.

cluster-autoscaler/cloudprovider/gce/gce_price_model.go
MaciekPytel: I'm not sure about this - in that case falling back to the default seems to be the clearly right thing to do (as opposed to situation where we don't have pricing info for some hardware). Those logs will get quite spammy (as it's likely to be repeated many times each loop) and not having any labels on node is unexpected, but not necessarily wrong.

cluster-autoscaler/cloudprovider/gce/gce_price_model.go
```diff
@@ -301,34 +305,72 @@ var (
...
 	basePriceFound := false
+	isPreemptible := false
+
+	// Base instance price
 	if node.Labels != nil {
+		isPreemptible = (node.Labels[preemptibleLabel] == "true")
 		if machineType, found := node.Labels[apiv1.LabelInstanceType]; found {
 			var priceMapToUse map[string]float64
-			if node.Labels[preemptibleLabel] == "true" {
+			if isPreemptible {
```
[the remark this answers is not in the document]
BigDarkClown: Agreed

PR #3259, opened 2020-07-01
Add c5a EC2 instance type for AWS

hajowieland:
Just added newly created AWS **c5a** EC2 instances type on EC2 catalog.

See the AWS announcement (04 JUN 2020): https://aws.amazon.com/blogs/aws/new-amazon-ec2-c5a-instances-powered-by-2nd-gen-amd-epyc-processors/

PR #3250, opened 2020-06-29
Adding r6g EC2 instance types

blee-wd:
Hoping to add r6g EC2 instance types

PR #3328, opened 2020-07-20
CA - AWS CloudProvider - Static Instance List update

gjtempleton:
Adds a number of new instance types/families to the static list

Doesn't include the `c5a` family as #3259 already exists, however does include the `r6g` family as #3250 seems to miss the smaller instance types in this family.

Also brings in the new `c6g` family [announced last month](https://aws.amazon.com/ec2/instance-types/c6/).


MaciekPytel:
Close + reopen to trigger travis rerun after fixing tests at HEAD.

PR #3028, opened 2020-04-06
VPA: Configurable container limit scaling

johanneswuerbach:
Make the proportional limit scaling configurable.

Follow up of https://github.com/kubernetes/autoscaler/issues/2359, https://github.com/kubernetes/autoscaler/pull/2362 and https://github.com/kubernetes/autoscaler/issues/2387

I'm happy to update the generated files and add tests if the approach looks acceptable.

//cc @Avanpourm @bskiba

PR #3340, opened 2020-07-21
Only cap to limit in RequestOnly mode

bskiba:


bskiba:
/assign @jbartosik 
Joachim, would you be OK to review?

For context: RequestsOnly mode reintroduced the previous logic of treating limits - leave them alone and only cap request to limit. However, unintentionally, #3028 introduced capping to all VPA objects, including those that have limits proportional to requests. This broke e2e tests.
@johanneswuerbach FYI only, I think we're good with this fix.

bskiba:
Tests: https://k8s-testgrid.appspot.com/sig-autoscaling-vpa#autoscaling-vpa-admission-controller

bskiba:
I've run admission-controller tests with this change and they passed.

PR #3343, opened 2020-07-23
Fix for VPA e2e vendor update

bskiba:
Currently e2e tests are failing with
go: inconsistent vendoring in /go/src/k8s.io/autoscaler/vertical-pod-autoscaler/e2e


krzysied:
I'm wondering why presubmit tests didn't catch it?
https://github.com/kubernetes/autoscaler/blob/master/hack/for-go-proj.sh#L68 sh

bskiba:
Depends on which go version Travis is using.

bskiba:
Hm, it's using 1.14 so it should have caught it...

PR #3344, opened 2020-07-23
cluster-autoscaler: ignore nodegroups with min/max tag issues

JulienBalestra:
While using Azure vmss, we discovered if we just have a single vmss missing the min/max tag on it, it failed to build the entire nodegroup cache.

The autoscaler is running without any potential candidates and the cluster can't be scaled at all.

I just added some warning logs, let me know if you want me to do differently and if the logic is fine like this.

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_manager.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_manager.go
@@ -671,28 +671,38 @@ func (m *AzureManager) listScaleSets(filter []labelAutoDiscoveryConfig) ([]cloud
 			if minSize, err := strconv.Atoi(*val); err == nil {
 				spec.MinSize = minSize
 			} else {
-				return asgs, fmt.Errorf("invalid minimum size specified for vmss: %s", err)
+				klog.Warningf("ignoring nodegroup %q because of invalid minimum size specified for vmss: %s", *scaleSet.Name, err)
+				continue
 			}
 		} else {
-			return asgs, fmt.Errorf("no minimum size specified for vmss: %s", *scaleSet.Name)
+			klog.Warningf("ignoring nodegroup %q because of no minimum size specified for vmss", *scaleSet.Name)
+			continue
 		}
 		if spec.MinSize < 0 {
-			return asgs, fmt.Errorf("minimum size must be a non-negative number of nodes")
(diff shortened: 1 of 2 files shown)


marwanad:
I'm honestly indifferent about this. Should we error and fail or should we just log?

For the explicit spec definition cases, for all providers - it would fail if the min/max were invalid on any nodegroup

https://github.com/kubernetes/autoscaler/blob/fdfd9e673ca700bea163d371c7b0cdcc9a224f5a/cluster-autoscaler/config/dynamic/node_group_spec.go#L68-L85


For autodiscovery, looking at GCE and it seems its consistent with the explicit case and also fails there?

https://github.com/kubernetes/autoscaler/blob/fdfd9e673ca700bea163d371c7b0cdcc9a224f5a/cluster-autoscaler/cloudprovider/gce/gce_manager.go#L187-L189


I'm totally fine with this fix - just wanted to get your opinion on what the expectations should be.

cluster-autoscaler/cloudprovider/azure/azure_manager.go
```diff
@@ -671,28 +671,38 @@ func (m *AzureManager) listScaleSets(filter []labelAutoDiscoveryConfig) ([]cloud
...
 		if spec.MaxSize < spec.MinSize {
-			return asgs, fmt.Errorf("maximum size must be greater than minimum size")
+			klog.Warningf("ignoring nodegroup %q because of maximum size must be greater than minimum size: max=%d < min=%d", *scaleSet.Name, spec.MaxSize, spec.MinSize)
+			continue
 		}
 
-		asg, _ := NewScaleSet(spec, m)
+		asg, err := NewScaleSet(spec, m)
+		if err != nil {
+			klog.Warningf("ignoring nodegroup %q %s", *scaleSet.Name, err)
```
marwanad, line 703: one more %q 😄

JulienBalestra:
@marwanad thanks for the answer.
On my end I believe that if we have some misconfiguration it should only impacts the object involved instead of the entire cluster.

In an ideal situation, we could report metrics around these issues to be able to easily surface this rather with logs only.

marwanad:
That's reasonable. Feel free to fix the uni tests and I'll lgtm.

JulienBalestra:
@marwanad I fixed the unit tests, let me know if it looks okay 👌

marwanad:
/lgtm
/approve

Thanks!

PR #3351, opened 2020-07-24
Add an annotation identifying upcoming nodes

towca:


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -57,6 +57,9 @@ const (
 	unschedulablePodWithGpuTimeBuffer = 30 * time.Second
 	// How long should Cluster Autoscaler wait for nodes to become ready after start.
 	nodesNotReadyAfterStartTimeout = 10 * time.Minute
+
+	// NodeUpcomingAnnotation is an annotation CA adds to nodes which are upcoming.
+	NodeUpcomingAnnotation = "cloud.google.com/cluster-autoscaler-node-upcoming"
```
MaciekPytel, line 62: This is in no way specific to GCP. Please use a more generic CA prefix.

towca: My bad, changed!

PR #3352, opened 2020-07-24
Add the ability to check if a pod fits onto any node matching a function

towca:


cluster-autoscaler/simulator/predicates_checker_interface.go
```diff
@@ -18,10 +18,12 @@ package simulator
 
 import (
 	apiv1 "k8s.io/api/core/v1"
+	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
 )
 
 // PredicateChecker checks whether all required predicates pass for given Pod and Node.
 type PredicateChecker interface {
 	FitsAnyNode(clusterSnapshot ClusterSnapshot, pod *apiv1.Pod) (string, error)
+	FitsAnyNodeMatching(clusterSnapshot ClusterSnapshot, pod *apiv1.Pod, nodeMatches func(*schedulerframework.NodeInfo) bool) (string, error)
```
MaciekPytel, line 27: I don't really like offloading the check to the function. What if the NodeInfo doesn't exist in the snapshot? I think that would fail somewhere in scheduler framework. 
The thing is we don't really know what are the criteria it uses to match those NodeInfos. I'd feel safer just passing a list of node names and taking NodeInfos for those directly from snapshot. WDYT?

MaciekPytel: Please ignore, I misread function signature.

PR #3335, opened 2020-07-21
Automated cherry pick of #3185: cluster-autoscaler: use generated instance types #3222: Fix AWS CA tests for InstanceType generation changes

gjtempleton:
Cherry pick of #3185 #3222 on cluster-autoscaler-release-1.16.

#3185: cluster-autoscaler: use generated instance types
#3222: Fix AWS CA tests for InstanceType generation changes

For details on the cherry pick process, see the [cherry pick requests](https://git.k8s.io/community/contributors/devel/sig-release/cherry-picks.md) page.

PR #3365, opened 2020-07-27
Cluster-Autoscaler - Resolve failing tests

gjtempleton:
Resolves failing tests from faulty cherry-pick #3335 - restores to previous args to creation of `autoScalingWrapper` in AWS manager tests


marwanad:
@MaciekPytel are we good to merge those to unblock the release?

MaciekPytel:
I'm not particularly happy to merge PRs to provider I don't own, as I generally don't have the context and I'd rather leave decisions to people who do. That being said I agree this one seems like a reasonable exception: it's a simple unittest fix and it impacts whole CA not just AWS provider.

/lgtm
/approve

PR #3373, opened 2020-07-29
[cluster-autoscaler] Skip nodes that are not Kubemark nodes

ellistarn:
For clusters with both real and hollow nodes, CA fails at static_autoscaler.go:268: "core_utils.GetNodeInfosForGroups(" since the Kubemark Cloud Provider doesn't know how to find pods for real nodes. This change will simply skip nodes that are not Kubemark nodes, as they cannot be scaled by CA's Kubemark implementation anyways.


ellistarn:
I've been asked to be OWNER of this provider by @MaciekPytel. This is pending https://github.com/kubernetes/org/issues/2068 if anyone is interested in sponsoring me before he gets back from vacation.

Issue #3378, opened 2020-07-30
Failed test: [VPA] [admission-controller] [v1beta2] Admission-controller accepts valid and rejects invalid VPA object


bskiba:
Corresponding V1 test passes.
/assign @bskiba

bskiba:
This is fixed
/close

PR #3382, opened 2020-07-30
VPA v1beta2 tests use v1beta2 client

bskiba:


bskiba:
fixes #3378

krzysied:
What about https://github.com/kubernetes/autoscaler/blob/51b767f43c24a51b89b5646574f03af47abc0643/vertical-pod-autoscaler/e2e/v1beta2/common.go#L388? It also has V1. Shouldn't it be changed?

bskiba:
Yeah you're right. Fixed.

PR #3391, opened 2020-07-31
VPA Add controlledValues to CRD

johanneswuerbach:
This was missed in https://github.com/kubernetes/autoscaler/pull/3028.

I updated the CRD definition manually, not sure whether there is a better way.


bskiba:
Thanks!
/lgtm
/approve

PR #3395, opened 2020-07-31
Add scale down protection for containers

evnsio:
At Monzo, we use the VPA on many of our microservices, which are pods containing a service container and an envoy sidecar.  In the vast majority of cases, the VPA works as expected and the requests are set to sensible values.  

image

We do however see cases where pods which typically see very little load, receive a surge in traffic, and the behaviour of the VPA results in all pods in the deployment being evicted in quick succession (each yellow line below is an eviction). Typically this happens when the request is close to the lower bound, and a small uptick in resource usage causes the recommender to suggest marginally higher resources. After the event which generated load, the VPA gradually decreases the requests until the next time.

image

Historically, we've relied on manually increasing the lower bound for the VPA recommendation to prevent these kind of pods from being sized too small and evicted under load, but with 1800+ independent microservices, we'd like the VPA to be able to increase the size of the containers only. 

To address this we've introduced an additional flag on the ContainerResourcePolicy - `PreventScaleDown` - which can be used to prevent the resources of a container from being reduced. The flag is used by:
- the Updater to prevent it evicting pods if the resulting change would mean scaling down
- the Admission Controller to ensure the requests are only set the recommended values if doing so would increase them
- the Recommender to prevent it reducing the recommendation values for these containers

> The overall behaviour is that a container with this flag will get the higher of it's static request, or the highest recommendation that has been produced over its lifetime.

## Example VPA Spec
```
apiVersion: autoscaling.k8s.io/v1beta2
kind: VerticalPodAutoscaler
metadata:
  labels:
    app: foo
  name: foo
spec:
  resourcePolicy:
    containerPolicies:
    - containerName: '*'
      maxAllowed:
        cpu: 1
        memory: 1Gi
      minAllowed:
        cpu: 10m
        memory: 50Mi
      preventScaleDown: true   # 👈 This is new!
  targetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: foo
  updatePolicy:
    updateMode: Auto
```

## Change Details
- Adds `PreventScaleDown` as an option on a container resource policy.  This allows the scale down protection to be applied on specific containers in a pod.
- Updates the recommender capping behaviour to prevent recommendations from being reduced for containers with `PreventScaleDown` enabled.  This ensures the recommended resources will remain static or increase over time.
- Updates the VPA Updater priority calculator so it doesn't flag a `PreventScaleDown` container which is requesting more than the upper bound as outside the recommended range. It is acceptable for these containers to request more resources than the VPA recommends. 
- Updates the capping behaviour in the admission controller to prevent the requests of any containers with `PreventScaleDown` from being reduced.  This prevents the 

### Testing changes
- Updates the VPABuilder test harness to make the setup of Recommendations and ContainerResourcePolicies explicit rather than implicit (previously, many of the `With` commands on the VPA builder were creating recommendations as side effects) The change aimed to make this more explicit and support multi-container testing more easily.
- Adds a new `ContainerResourcePolicyBuilder` to support adding these explicitly to tests.

## Remaining Work
- The documentation for this setting is missing.  If the approach is agreed, I'll write it before merging.

### DIFF
--- a/vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go
+++ b/vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go
@@ -90,9 +90,6 @@ func (calc *UpdatePriorityCalculator) AddPod(pod *apiv1.Pod, now time.Time) {
 	}
 
 	hasObservedContainers, vpaContainerSet := parseVpaObservedContainers(pod)
-
-	updatePriority := calc.priorityProcessor.GetUpdatePriority(pod, calc.vpa, processedRecommendation)
-
 	quickOOM := false
 	for i := range pod.Status.ContainerStatuses {
 		cs := &pod.Status.ContainerStatuses[i]
@@ -119,9 +116,11 @@ func (calc *UpdatePriorityCalculator) AddPod(pod *apiv1.Pod, now time.Time) {
 		}
 	}
 
+	updatePriority := calc.priorityProcessor.GetUpdatePriority(pod, calc.vpa, processedRecommendation)
+
 	// The update is allowed in following cases:
 	// - the request is outside the recommended range for some container.
-	// - the pod lives for at least 24h and the resource diff is >= MinChangePriority.
+	// - the pod lives for at least 12h and the resource diff is >= MinChangePriority.
 	// - a vpa scaled container OOMed in less than evictAfterOOMThreshold.
 	if !updatePriority.OutsideRecommendedRange && !quickOOM {
 		if pod.Status.StartTime == nil {
@@ -144,6 +143,7 @@ func (calc *UpdatePriorityCalculator) AddPod(pod *apiv1.Pod, now time.Time) {
 		klog.V(2).Infof("not updating pod %v/%v because resource would not change", pod.Namespace, pod.Name)
 		return
 	}
+
 	klog.V(2).Infof("pod accepted for update %v/%v with priority %v", pod.Namespace, pod.Name, updatePriority.ResourceDiff)
 	calc.pods = append(calc.pods, prioritizedPod{
 		pod:            pod,
--- a/cluster-autoscaler/cloudprovider/magnum/gophercloud/openstack/orchestration/v1/stacks/utils.go
+++ b/cluster-autoscaler/cloudprovider/magnum/gophercloud/openstack/orchestration/v1/stacks/utils.go
@@ -9,8 +9,8 @@ import (
 	"reflect"
 	"strings"
 
-	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/magnum/gophercloud"
 	yaml "gopkg.in/yaml.v2"
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/magnum/gophercloud"
 )
 
 // Client is an interface that expects a Get method similar to http.Get. This
--- a/cluster-autoscaler/cloudprovider/magnum/gophercloud/pagination/linked.go
+++ b/cluster-autoscaler/cloudprovider/magnum/gophercloud/pagination/linked.go
@@ -40,7 +40,7 @@ func (current LinkedPageBase) NextPageURL() (string, error) {
 	}
 
 	for {
-		key, path = path[0], path[1:len(path)]
+		key, path = path[0], path[1:]
 
 		value, ok := submap[key]
 		if !ok {
--- a/vertical-pod-autoscaler/pkg/admission-controller/resource/pod/handler_test.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/resource/pod/handler_test.go
@@ -58,7 +58,9 @@ func (c *fakePatchCalculator) CalculatePatches(_ *apiv1.Pod, _ *vpa_types.Vertic
 }
 
 func TestGetPatches(t *testing.T) {
-	testVpa := test.VerticalPodAutoscaler().WithName("name").WithContainer("testy-container").Get()
+	testVpa := test.VerticalPodAutoscaler().WithName("name").
+		AppendRecommendation(test.Recommendation().WithContainer("testy-container").GetContainerResources()).
+		AppendContainerResourcePolicy(test.ContainerResourcePolicy().WithContainer("testy-container").Get()).Get()
 	testPatchRecord := resource_admission.PatchRecord{
 		Op:    "add",
 		Path:  "some/path",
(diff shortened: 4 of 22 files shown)


evnsio:
Thinking about it, I'm not sure whether I'd made an incorrect assumption in the admission controller change.  I've assumed that the a pod being scheduled will come in with it's request set to the previous recommendation, and that value will be compared to the current recommendation.  If that holds true then the change is good.

If, however, the pod reaches the admission controller with it's request set to the fixed value from the deployment/pod spec then all we're preventing is the recommendation being scaled lower than _that_.  If this is the case, then for this to work I'd need to store somewhere the last recommendation that applied to the target pods rather than the request.

I'd appreciate some feedback on this aspect especially.

vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go
```diff
@@ -119,9 +116,11 @@ func (calc *UpdatePriorityCalculator) AddPod(pod *apiv1.Pod, now time.Time) {
 		}
 	}
 
+	updatePriority := calc.priorityProcessor.GetUpdatePriority(pod, calc.vpa, processedRecommendation)
+
 	// The update is allowed in following cases:
 	// - the request is outside the recommended range for some container.
-	// - the pod lives for at least 24h and the resource diff is >= MinChangePriority.
+	// - the pod lives for at least 12h and the resource diff is >= MinChangePriority.
```
evnsio, line 123: Fixing this to align to the definition above:
```
podLifetimeUpdateThreshold = time.Hour * 12
```

bskiba:
Apologies for the long delay in getting back to you.

I'd like to understand a couple things at first. It would be great to understand why the regular recommendation algorithm does not work well in this case before we try to work around it.
  - can you share roughly for a problematic workload what is the time share that it spends in the "low usage" vs "high usage" phases?
  - do I understand correctly that this is mostly a problem with cpu recommendations?
  - looking at the first graph, is the y axis cpu cores? So we are talking about changes between 100mCpu and 150 mCPU? Is this typical for the problematic workloads or happens regardless of the recommendation value?
  - aside from the low/high usage pattern is there anything about the pods that experience this problem that is different from the ones that VPA handles well?

On the approach itself, ideally I would expect the prevent scale down as purely an actuation mechanism, i.e. changes would only be needed in updater and admission controller. This would mean that live pods would never be shrinked, however new pods could be created with the lower than max recommendation.

mwielgus:
Ping :)

evnsio:
Thank you @mwielgus!

> can you share roughly for a problematic workload what is the time share that it spends in the "low usage" vs "high usage" phases?

We have a number of workloads that are hit hard once per month.  The VPA behaviour results in containers being set to the appropriate size for the low period, and then evicted in quick succession when the load increases.

> do I understand correctly that this is mostly a problem with cpu recommendations?

It's both actually, the graphs only show CPU for brevity.

> looking at the first graph, is the y axis cpu cores? So we are talking about changes between 100mCpu and 150 mCPU? Is this typical for the problematic workloads or happens regardless of the recommendation value?

Yes, 1-200 millicores is pretty typical for our workloads.  These are small microservices running in golang, and we have close to 2000 separate microservices running, all with a least 3 replicas, and some with many more.

> aside from the low/high usage pattern is there anything about the pods that experience this problem that is different from the ones that VPA handles well?

No, all of our workloads look the same.  Each pod is made up from:

- The main application microservice container
- A sidecar envoy proxy which make up our service mesh

In terms of autoscaling configurations:

- We have some pods that are solely using the VPA for both containers, with separate container resource policies for each.
- We have some pods where the HPA scales the replicas based on custom metrics tracking the CPU of the service container, and the VPA scales the envoy sidecar.

The general problem is just pods that see low usage for a large time, and then have a period of increased usage.

evnsio:
> On the approach itself, ideally I would expect the prevent scale down as purely an actuation mechanism, i.e. changes would only be needed in updater and admission controller. This would mean that live pods would never be shrinked, however new pods could be created with the lower than max recommendation.

It's been a while since I looked at the code, but this does sound like a reasonable approach.  I'll remind myself why I included the recommender and come back to you!

evnsio:
Right, I've refreshed my memory but I'd love your thoughts on whether my logic holds up for needing the recommender involved.  To my understanding, this is how the admission process works:

- New pod is being admitted into K8s
- Its requests will be those that are specified in the deployment manifest (assuming it's from a deployment)
- Pod is processed by the VPA admission controller
- The admission controller uses the recommendation to update the requests (and optionally limits)

Critically, the only values we have to work with at the point of admission are the recommended ones, and the ones that are statically defined on the pod itself. 

If we _don't_ include the recommender in the `preventScaleDown` option, then when new pods are admitted, there's no way to prevent them reducing their request.  We have nothing to compare the static request as specified in the deployment to.  

If we _do_ include the recommender in the `preventScaleDown` option, then newly admitted pods can be compared to this and guarantee that they wont have their requests reduced.

evnsio:
@bskiba 👋 ☝️

bskiba:
@evnsio Thanks a lot for the detailed information.

I agree that once in a month spikes are not something we handle well. There is a number of ways we can approach this and I am still thinking about which one to take.

One downside of the approach that you propose is that there is no way to get rid of an unintentional high recommendation. For example if you mistakenly overload a deployment or if there is a bug in the recommender logic, there is no way to shrink the recommendation other than adding an artificial maximum or creating a new VPA object. (This second option might actually be a valid countermeasure).

Do allow me a couple more days to dwell on this, I'll get back to you again next week.

bskiba:
Thanks for your patience, I have a couple more questions.

1. When the pods are hit with the once in a month traffic, what would happen if they were not scaled up by VPA? Would they be able to handle the load with maybe some level of degradation?
2. In the second graph from https://github.com/kubernetes/autoscaler/pull/3395#issue-460439494, can you tell us what is on y axis and what exactly is plotted? If I read the graph right, VPA would react in a matter of minutes, and this is not what we typically observe.
3. Do you have an example of a recommendation value as observed before a problematic spike? I would be interested in target, uncapped target, lower and upper bound values.
4. Have you considered alternatives to this approach?

For the context, my many questions come from the fact that I would like every addition to the API well thought out and motivated so I'd like to understand the gist of the problem it solves. Also in general VPA is not well suited to workloads with patterns that occur less frequently than a week and not well suited to react fast to sudden spikes.

mwielgus:
Ping. Inactivity warning.

mwielgus:
Closing due to inactivity. Feel free to reopen once needed.

PR #3393, opened 2020-07-31
fix(charts): Renaming chart to avoid tag name conflict

sc250024:
# Description

Quick fix after it was discovered that the build triggered on `master` after https://github.com/kubernetes/autoscaler/pull/3341 was merged failed because of the conflicting tags.

Essentially, this renames the folder `charts/cluster-autoscaler` to `charts/cluster-autoscaler-chart` so that, when a Helm chart change is released, it has the separate tag `cluster-autoscaler-chart-1.X.X` to keep releases separate.

# IMPORTANT

**_Please make sure that the `gh-pages` branch exists before merging this!_**. Here are the commands for someone with write access:

```shell
$ git checkout --orphan gh-pages
$ git rm -rf .
$ git add index.html
$ git commit -m "Initial commit"
$ git push origin gh-pages
```

<details>
<summary>index.html</summary>

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="viewport" content="width=device-width, initial-scale=1">

    <title>Autoscaler Helm Charts</title>

    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css">
    <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.0.10/css/all.css">

    <!-- HTML5 Shim and Respond.js IE8 support of HTML5 elements and media queries -->
    <!-- Leave those next 4 lines if you care about users using IE8 -->
    <!--[if lt IE 9]>
        <script src="https://oss.maxcdn.com/html5shiv/3.7.2/html5shiv.min.js"></script>
        <script src="https://oss.maxcdn.com/respond/1.4.2/respond.min.js"></script>
      <![endif]-->
  </head>

  <body>
    <h1>Autoscaler Helm Charts</h1>
    <p>Use the following commands to install the Autoscaler Helm repository:</p>
    <pre>helm repo add autoscaler https://kubernetes.github.io/autoscaler
helm repo update</pre>
    <!-- Including Bootstrap JS (with its jQuery dependency) so that dynamic components work -->
    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js"></script>
    <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/js/bootstrap.min.js"></script>
  </body>
</html>
```

</details>

# Testing

[...]

PR #3403, opened 2020-08-03
CA-Chart - Move from helm chart-releaser-action to scripts

gjtempleton:
Moves away from the helm chart-releaser-action to scripts due to `gh-pages` branch being protected.

Post merging of #3393 the attempted publish action failed due to the branch protection on the `gh-pages` branch.

Unfortunately due to the issues documented by https://github.com/helm/chart-releaser-action/issues/33 we can't use the official chart-releaser action currently as it relies on the ability to push directly to the `gh-pages` branch.

This workflow and the associated script are lifted liberally from the kubernetes/dashboard repo's flow for publishing their helm charts: https://github.com/kubernetes/dashboard/blob/master/.github/workflows/cd-helm-workflow.yml albeit changing some regex and custom functions.

I'm hopeful that this will be a short term fix until the official action resolves this issue.

### Questions still to be addressed

- [ ] How to clean up the branch created for raising PRs `gh-pages-pull-request` as this existing will block further raising of PRs - I briefly considered randomly generating the branch names, but that just puts the problem off and we'd end up with many stale branches, not ideal, would prefer this to be automated if at all possible
It looks like the maintainers of the dashboard repo are currently manually cleaning this up post merging - not ideal.

- [ ] Whether the script should be made more generic now to handle possible extension of charts hosted in this repo - I'm hopeful that the official action should address the issue raised above and as such we won't need to use this long term

- [ ] Whether branch protection will automatically be stuck on the new `gh-pages-pull-request` branch preventing a push to it - I don't think this should be the case looking at `kubernetes/dashboard`, but worth checking

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -4,16 +4,39 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Fetch history
-        run: git fetch --prune --unshallow
+        with:
+          fetch-depth: 0
+
       - name: Configure Git
         run: |
           git config user.name "${GITHUB_ACTOR}"
           git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
-      - env:
-          CR_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        name: Run chart-releaser
-        uses: helm/chart-releaser-action@v1.0.0-rc.2
+
+      - name: Package
+        shell: bash
+        run: hack/helm-release-chart.sh
+
+      - name: Create Pull Request branch
+        run: |
+          git checkout -b gh-pages-pull-request
+
+      # Note that this will fail if branch already exists - i.e. we can only have one bot raised PR at a time
+      - name: Push Git Branch
+        uses: ad-m/github-push-action@v0.6.0
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          branch: gh-pages-pull-request
+
+      # Note that this will silently fail if the PR already exists
+      - name: Create Pull Request
+        uses: repo-sync/pull-request@v2
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          source_branch: gh-pages-pull-request
+          destination_branch: gh-pages
+          pr_title: "Helm Chart: Update Helm repository"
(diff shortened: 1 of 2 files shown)


gjtempleton:
@sc250024 Would be great to get your input on this

gjtempleton:
/hold
Sticking a hold to prevent accidental merging until questions are addressed

sc250024:
Why is `gh-pages` a protected branch if it was just created?

gjtempleton:
Test-infra enforces it via [branchprotector](https://github.com/kubernetes/test-infra/blob/8c8546d74168a57d0a3d22b17d98eb09bbf70601/prow/cmd/branchprotector/README.md).

As per: https://github.com/kubernetes/test-infra/blob/master/config/prow/config.yaml#L143

So once the first push has been made it'll become protected.

sc250024:
If we're doing it this way, then I would say the only change in the PR would be to rename the chart back to `cluster-autoscaler`.

From `cluster-autoscaler-chart` => `cluster-autoscaler`

gjtempleton:
@sc250024 I thought about that, however I hope that we'll be able to move back to the official chart-releaser action at some point, and worried about potentially then changing the naming scheme when switching back.

sc250024:
> @sc250024 I thought about that, however I hope that we'll be able to move back to the official chart-releaser action at some point, and worried about potentially then changing the naming scheme when switching back.

Alternatively, has the [kubernetes](https://github.com/kubernetes) org simply thought about creating a dedicated `helm-charts` repo like other organizations do?

- https://github.com/vmware-tanzu/helm-charts
- https://github.com/bitnami/charts
- https://github.com/argoproj/argo-helm
- https://aws.github.io/eks-charts

Seems like a common pattern?

gjtempleton:
@sc250024 It's been discussed with the relevant working group a few months back, and the feedback seemed to be that it's for individual repos to maintain their own charts if they want to adopt them for now.

gjtempleton:
I've been giving the workflow currently used by this PR some more thought and would like to avoid committing the tarballs to the repo itself if at all possible and follow the chart releaser action's workflow of publishing a release through the GH API instead. I'll try to find some time to rework it to follow that flow this week.

gjtempleton:
Closing as superseded by #3679

PR #3404, opened 2020-08-04
Fix wrong default value for expendable-pods-priority-cutoff in FAQ

bruecktech:
The default is `-10`


MaciekPytel:
/lgtm
/approve
Thanks for catching this

PR #3405, opened 2020-08-04
Update OWNERS

mgoodness:
I'm no longer using the autoscaler chart, so I'll bow out of maintainership.


gjtempleton:
Thanks for all your hard work maintaining it to this point!
/lgtm
/approve

PR #3411, opened 2020-08-05
Fix label selector in chart post-install notes

tghartland:
Fixes the label selectors in the `kubectl get pods` command printed by the post-install notes.

The `app` and `release` labels were replaced by this commit https://github.com/helm/charts/commit/5e4e493bbb27037c54fdeeab7397de2452310d52 in the old chart location. The new labels come from

https://github.com/kubernetes/autoscaler/blob/852ea800914cae101824687a71236f7688ee653d/charts/cluster-autoscaler-chart/templates/_helpers.tpl#L36-L39

which have the same values as the old `app` and `release` labels.


gjtempleton:
@tghartland can you give the version a patch version bump please?

gjtempleton:
Thanks!
/lgtm
/approve

Issue #3266, opened 2020-07-02
vpa-recommender: lots of duplicate scale calls


chenchun:
All workloads are extended workloads "tapp" in my cluster. Each time vpa-recommender runs a single loop of recommendation, it calls scale api three times for each tapp. I have 2400+ tapp in my cluster, and most time of a loop are spent in [get selector and validate target ref of vpa](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go#L475-L477).  I have set `--kube-api-qps=1000` `--kube-api-burst=2000` to speed the loop, but it increases the pressure for kube-apiserver.

Since workloads may run for a long time, I think it's reasonable to cache scale resource with cache time tunnable.
And it seems we can't list-watch scale resource via client-go, can we add a cache for scale resource ourself?

![image](https://user-images.githubusercontent.com/985861/86310210-0a092b00-bc50-11ea-9be5-2f6459f38206.png)

bskiba:
@chenchun Could you summarise your takeaways from the discussions in the PRs/issues around supporting watch and list for subresources? I gather the idea did not go through but I am interested in the reasons behind that.

chenchun:
Because they think 
> `GET /apis/apps.tkestack.io/v1/namespaces/default/tapps//scale` This doesn't follow Kubernetes API conventions. The subresource endpoints belong to one resource instance, and do not support LIST operation. This is true for both built-in types and custom resources.

In my opinion, `GET /apis/apps.tkestack.io/v1/namespaces/default/tapps//scale` represents listing all tapps scale objects or `GET /apis/apps.tkestack.io/v1/watch/namespaces/default/tapps//scale` for watching all tapps scale objects  is indeed not a good idea and I can't think of a better restful uri than this. `//` is valid but don't use it. Refer [Is a URL with // in the path-section valid?](https://stackoverflow.com/questions/20523318/is-a-url-with-in-the-path-section-valid) But I do think supporting list and watch subresource for custom resources is valuable and I don't know why we can't break this convention in a way that is compatible with current usage.

PR #3412, opened 2020-08-06
cache crd objects to speed up VpaTargetSelectorFetcher and ControllerFetcher

chenchun:
Fixes https://github.com/kubernetes/autoscaler/issues/3266
Without this patch, it takes **1m41s** (from `I0806 07:50:08.676461` to `I0806 07:51:49.018659`) to get all selectors for each of 3111 vpas (mostly crd workload targets).
image

With this patch, it takes **1s** (from `I0806 08:02:49.137257` to `I0806 08:02:50.266574`).
image

We also save 3111*3/2/60 qps of scale http calls to kube-apiserver and speed up vpa admission.


krzysied:
Wow! The performance improvement looks amazing.
Can I review this?

bskiba:
/assign @krzysied 
Krzysztof please do take a look. 
My main concern with the approach is that we need get, list and watch permissions on all objects in the cluster.

krzysied:
> My main concern with the approach is that we need get, list and watch permissions on all objects in the cluster.

I'm not sure if we can do it differently. The best solution would be having watch on scale subresource, but k8s doesn't prove such option.
These permissions give a lot of access, but it's only read access.

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -108,7 +109,8 @@ func NewClusterStateFeeder(config *rest.Config, clusterState *model.ClusterState
 	kubeClient := kube_client.NewForConfigOrDie(config)
 	podLister, oomObserver := NewPodListerAndOOMObserver(kubeClient, namespace)
 	factory := informers.NewSharedInformerFactoryWithOptions(kubeClient, defaultResyncPeriod, informers.WithNamespace(namespace))
-	controllerFetcher := controllerfetcher.NewControllerFetcher(config, kubeClient, factory)
+	crdCache := crdcache.NewCrdCache(config, make(chan struct{}), defaultResyncPeriod)
```
krzysied, line 112: stop channel should provided via function params.

vertical-pod-autoscaler/pkg/crdcache/crdcache.go
```diff
@@ -0,0 +1,167 @@
...
+
+// NewCrdCache returns new instance of CrdCache
+func NewCrdCache(config *rest.Config, stop chan struct{}, resyncTime time.Duration) CrdCache {
+	dynamicClient := dynamic.NewForConfigOrDie(config)
+	c := &crdCache{
+		stop:             stop,
+		dynamicFactory:   dynamicinformer.NewDynamicSharedInformerFactory(dynamicClient, resyncTime),
+		startedInformers: map[schema.GroupVersionResource]bool{},
+	}
+	// lazy watch of CustomResourceDefinition
```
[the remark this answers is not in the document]
chenchun: Because someone may not run extended workload or creating vpa for them.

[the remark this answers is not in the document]
krzysied: That's true.
However, having simpler code in exchange for one additional watch can be a good trade-off.
Also, this is only watching CRDs, so it shouldn't generate a heavy load on the master.

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
chenchun: I don't know what you mean. I deleted stop chan param and follow the rules in other package to create stop chan in crdcache.

vertical-pod-autoscaler/pkg/crdcache/crdcache.go
```diff
@@ -0,0 +1,161 @@
...
+	c.extensionLister = extensionInformer.Lister()
+	return c
+}
+
+type crdCache struct {
+	dynamicFactory  dynamicinformer.DynamicSharedInformerFactory
+	extensionLister extensionlister.CustomResourceDefinitionLister
+	// startedInformers caches started GroupVersionResource informers
+	startedInformers map[schema.GroupVersionResource]bool
+	lock             sync.Mutex
```
krzysied, line 81: nit: put `lock` above `startedInformers`

vertical-pod-autoscaler/pkg/crdcache/crdcache.go
```diff
@@ -0,0 +1,161 @@
...
+	return unstructuredObj.GetOwnerReferences(), nil
+}
+
+// getLister gets the lister for the given GroupVersionResource and makes sure the informer have started and synced
+func (c *crdCache) getLister(gvr schema.GroupVersionResource) cache.GenericLister {
+	informer := c.dynamicFactory.ForResource(gvr)
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	if _, ok := c.startedInformers[gvr]; !ok {
+		klog.Infof("watching custom resource definition %s", gvr.String())
```
krzysied, line 147: nit: logs start with capital letters.

- fmt.Error: lower case letter - error will be concatenated.
- klog.Info: capital letter - responsible for printing

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -266,9 +260,8 @@ func (f *controllerFetcher) isWellKnownOrScalable(key *ControllerKeyWithAPIVersi
 	}
 
 	for _, mapping := range mappings {
-		groupResource := mapping.Resource.GroupResource()
-		scale, err := f.scaleNamespacer.Scales(key.Namespace).Get(context.TODO(), groupResource, key.Name, metav1.GetOptions{})
-		if err == nil && scale != nil {
+		isScale, err := f.crdCache.IsScalable(mapping.Resource, key.Namespace, key.Name)
```
krzysied, line 263: s/isScale/isScalable

vertical-pod-autoscaler/pkg/crdcache/crdcache_test.go
```diff
@@ -0,0 +1,147 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/test"
+	dynamic "k8s.io/client-go/dynamic/fake"
+)
+
+var (
+	scalelabelFields   = []string{"status", "scaleLabelSelector"}
```
krzysied, line 37: nit: s/scalelabelFields/scaleLabelFields

vertical-pod-autoscaler/pkg/crdcache/crdcache_test.go
```diff
@@ -0,0 +1,147 @@
...
+		obj       *unstructured.Unstructured
+		expectErr error
+	}{
+		{gvr: fooGvr, obj: foo1},
+		{gvr: fooGvr, obj: foo2},
+		{gvr: barGvr, obj: bar1},
+		{gvr: notScalableGvr, obj: notScalable1, expectErr: fmt.Errorf("resource %s hasn't scale sub-resource", notScalableGvr.String())},
+		{gvr: fooGvr, obj: objNotExist1, expectErr: fmt.Errorf(`foos.test.org "obj-not-exist1" not found`)},
+	} {
+		msg := fmt.Sprintf("case %d", i)
```
krzysied, line 86: Please use `t.Run` as test case wrapper.
```
t.Run(testCase.name, func(t *testing.T) {
    (...)
})
```
so adding `msg := fmt.Sprintf("case %d", i)` everywhere should not be required.

vertical-pod-autoscaler/pkg/crdcache/crdcache_test.go
```diff
@@ -0,0 +1,147 @@
...
+		expectErr      error
+		expectScalable bool
+	}{
+		{gvr: fooGvr, obj: foo1, expectScalable: true},
+		{gvr: fooGvr, obj: foo2, expectScalable: true},
+		{gvr: barGvr, obj: bar1, expectScalable: true},
+		{gvr: notScalableGvr, obj: notScalable1},
+		{gvr: fooGvr, obj: objNotExist1, expectErr: fmt.Errorf(`foos.test.org "obj-not-exist1" not found`)},
+	} {
+		msg := fmt.Sprintf("case %d", i)
```
krzysied, line 115: Same case as TestFetchSelector.

vertical-pod-autoscaler/pkg/crdcache/crdcache_test.go
```diff
@@ -0,0 +1,147 @@
...
+	owners, err := crdCache.GetOwnerReferences(fooGvr, objHasOwner.GetNamespace(), objHasOwner.GetName())
+	assert.NoError(t, err)
+	assert.Equal(t, objHasOwner.GetOwnerReferences(), owners)
+}
+
+func newFakeCrdCache() CrdCache {
+	dynamicClient := dynamic.NewSimpleDynamicClient(runtime.NewScheme(), foo1, foo2, bar1, notScalable1, objHasOwner)
+	extensionClient := extensionclient.NewSimpleClientset(fooCrd, barCrd, notScalableCrd)
+	c := NewCrdCache(dynamicClient, extensionClient, time.Minute*10)
+	return c
```
krzysied, line 138: nit: no need for a new variable.

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
krzysied: I was thinking about having stop channel as an input parameter. This would be helpful if cluster feeder was recreated many times during recommender life. 
However, I looked at the recommender code and it didn't seem like a use case. Moreover other informers are also not propagating stop channels.
Deleting stop channel seems ok.

krzysied:
/hold
I talked with security folks and having rbac that allows to list the whole universe doesn't seems to be a great idea.
I will look for some alternatives/ways to restrict the access.

bskiba:
Form investigation by @krzysied, the current understanding is that giving VPA read access to all objects is a no-go from security perspective (gives access to secrets) and there is no way to express "give access to everything but secrets". Unfortunately, I don't see how we can make this PR work without that so I'm closing it, but feel free to reopen if you see an alternative.
/close

chenchun:
> there is no way to express "give access to everything but secrets"

I find https://github.com/kubernetes/kubernetes/issues/85963 talks about this.
@bskiba Do you think it's worth adding a flag to make this feature turned off by default and can be turned on for specific clusters by adding permissions for certain workloads instead of all objects to RBAC ?

bskiba:
Sadly, looks like https://github.com/kubernetes/kubernetes/issues/85963 is rejected. We're looking currently at a different way of tackling this, I'm hoping https://github.com/kubernetes/autoscaler/pull/3589 will help.

Issue #3318, opened 2020-07-14
Adopt Cluster Autoscaler Helm chart


sc250024:
# Description

Since the Helm stable channel will be deprecated eventually, I'm wondering if it's possible that this repository can adopt the [`stable/cluster-autoscaler`](https://hub.helm.sh/charts/stable/cluster-autoscaler) Helm chart into this repository?

Other projects are doing the same, notably NGINX Ingress (https://github.com/kubernetes/ingress-nginx) and Kubernetes dashboard (https://github.com/kubernetes/dashboard) as simple examples.

gjtempleton:
Hi @sc250024 thanks for raising this

We've discussed it at the SIG Autoscaling meeting earlier today and we don't have any objection to adopting the chart going forward as long as there are people willing to maintain it going forward. I'm willing to take this on, but if there are any others who want to contribute to this please let us know.

We'll need to discuss the specifics of how we do this in more detail.

sc250024:
@gjtempleton I would be willing to maintain it. I can open a PR for it soon.

I've noticed that the current CI process for the autoscaler is using Travis. Do you mind if the Helm chart uses GitHub actions? The reason is because there are already automated actions which handle this quite well.

- https://github.com/helm/chart-testing-action
- https://github.com/helm/chart-releaser-action

Other notable Kubernetes projects (dashboard, nginx-ingress) use these actions as well.

gjtempleton:
I don't have an objection to Github actions given they're being used elsewhere in the k8s org already, and as you point out those seem to be well supported paths for this.

Best to get input from @MaciekPytel as well.

MaciekPytel:
I don't have any experience with github actions, but based on the comment above I don't see any reason against using them.

sc250024:
@MaciekPytel Cool. I'll open a PR.

sc250024:
PR here: https://github.com/kubernetes/autoscaler/pull/3341

stevehipwell:
We've done the same thing on the descheduler project and had the same issue around chart naming and CI (waiting on tide to support the testing action).

https://github.com/kubernetes-sigs/descheduler/issues/248

hakman:
Hi. Quick question in case I am missing something. Should adding the `autoscaler` repo work?
```bash
$ helm version
[...]

PR #3413, opened 2020-08-06
CA-Chart Manual Release

gjtempleton:
Manual creation of index.yaml for 1.0.0 and 1.0.1 releases of Cluster-autoscaler Helm chart

This manually addresses the number of people running into 404s trying to use the repository for now (see #3318 comments for some of these) #3403 is an attempt to move to an automated process for this in the future

Also removes @mgoodness from the `OWNERS` file in `gh-pages` branch as per #3405 - adds in same approvers and reviewers from repo's default branch to prevent dropping to one person.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,6 +1,10 @@
 approvers:
-- mgoodness
 - gjtempleton
+- mwielgus
+- maciekpytel
+- bskiba
 reviewers:
-- mgoodness
 - gjtempleton
+- mwielgus
+- maciekpytel
+- bskiba
--- a/index.yaml
+++ b/index.yaml
@@ -1,0 +1,50 @@
+apiVersion: v1
+entries:
+  cluster-autoscaler-chart:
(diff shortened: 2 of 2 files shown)


gjtempleton:
/hold

Sticking a hold on until all mentioned parties have OKed this

mwielgus (approved):
/lgtm
/approve

eaterm:
Just tried installing the chart and the application cluster-autoscaler doesn't exist.
The autoscaler application is named: cluster-autoscaler-chart inside the chart.

To install you would need to run: helm install my-release autoscaler/cluster-autoscaler-chart

Not sure if the is correct.

gjtempleton:
It is, unfortunately we need to distinguish the naming from the already published `cluster-autoscaler` binaries due to the Github action we're using for publishing. 

See https://github.com/helm/chart-releaser/issues/65

PR #3420, opened 2020-08-10
Use shallow clone in update-vendor.sh

tghartland:
Pulling the complete kubernetes/kubernetes repo can take a while especially on a slow connection.

Instead of pulling everything, this change pulls only the tip of the master branch and then fetches the single revision that is requested, checking it out using `FETCH_HEAD`.

It works with `--k8srev=branch-name`, `--k8srev=tags/tag-name` and `--k8srev=commit-sha`, which should be all the ways this could already be used.

The total download size is reduced from ~750MB to ~50MB. The disk space used by the cloned kubernetes repo is also reduced from 950MB to 250MB.


mwielgus:
cc: @MaciekPytel

mwielgus (approved):
/lgtm
/approve

PR #3424, opened 2020-08-11
Add magnum support in cluster autoscaler helm chart 

### LINKED ISSUE #3399: Support Magnum installation with cluster autoscaler helm chart
The main change will be to add the autodiscovery configuration for when the magnum provider is selected. Possibly some changes for how the cloud-config file is loaded but I'm not sure about that yet.

tghartland:
Fixes #3399

@gjtempleton This is everything that I think needs to be changed, but I've marked this as WIP to discuss a few things. Mostly things where I'm assuming that other providers will also want to add to the chart.

First, I added
```yaml
autoDiscovery:
  # autoDiscovery.roles -- Magnum node group roles to match
  roles: []
```
for configuring `--node-group-auto-discovery` based on group roles. It serves the same purpose as the existing `autoDiscovery.tags` that AWS uses, but for keeping naming consistency within each provider I think it makes sense.

It would be possible to separate it as
```yaml
autoDiscovery:
  magnum:
    # autoDiscovery.magnum.roles -- Magnum node group roles to match
    roles: []
```
and this might keep things more organised if there are going to be other providers adding to `autoDiscovery`.

Second, where I have this `or`
```
{{- if or (eq .Values.cloudProvider "gce") (eq .Values.cloudProvider "magnum") }}
  - --cloud-config={{ .Values.cloudConfigPath }}
{{- end }}
```
would that be better as a second `if` with the same content? To keep separation of the providers and to keep the condition from getting too long to be easily readable, if other providers add themselves in as well.

There are also some things that can be updated in the documentation, where it is saying that autodiscovery is limited to the aws or gce provider, I can look at rewording those later though.

### DIFF
--- a/charts/cluster-autoscaler-chart/Chart.yaml
+++ b/charts/cluster-autoscaler-chart/Chart.yaml
@@ -17,4 +17,4 @@ name: cluster-autoscaler-chart
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 1.0.5
+version: 1.1.0
--- a/charts/cluster-autoscaler-chart/README.md
+++ b/charts/cluster-autoscaler-chart/README.md
@@ -45,10 +45,15 @@ You must provide some minimal configuration, either to specify instance groups o
 
 Either:
 
-- Set `autoDiscovery.clusterName` and tag your autoscaling groups appropriately (`--cloud-provider=aws` only) **or**
-- Set at least one ASG as an element in the `autoscalingGroups` array with its three values: `name`, `minSize` and `maxSize`.
+- Set `autoDiscovery.clusterName` and provide additional autodiscovery options if necessary **or**
+- Set static node group configurations for one or more node groups (using `autoscalingGroups` or `autoscalingGroupsnamePrefix`).
 
-To install the chart with the release name `my-release`:
+To create a valid configuration, follow instructions for your cloud provider:
+
+* [AWS](#aws---using-auto-discovery-of-tagged-instance-groups)
+* [GCE](#gce)
+* [Azure AKS](#azure-aks)
+* [OpenStack Magnum](#openstack-magnum)
 
 ### AWS - Using auto-discovery of tagged instance groups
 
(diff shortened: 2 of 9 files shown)


mwielgus:
Inactivity warning. Please finish the PR and resolve conflicts.

tghartland:
@gjtempleton can you take a look at this?

gjtempleton:
@tghartland Sorry, this fell through the cracks. 

Thanks for the PR, I'll make time to give this a look over this weekend.

gjtempleton (commented):
I think this mostly looks good to me, thanks for your work on this. 

My preference if we were starting fresh would be to do as you suggest and go with 

```yaml
autoDiscovery:
  magnum:
    roles: []
  aws:
    tags: []
```

However given where the the tags key already is, I'm happy to go with `roles: []` where it is.

I'm also happy to keep the `or` logic for now, if a third cloud provider starts making use of that then I think we can revisit at that point.

On roughly the same topic I wonder if we can reduce the duplication of the cloudconfig volume with something like:

```yaml
      {{- if or (eq .Values.cloudProvider "gce") (eq .Values.cloudProvider "magnum") }}
      volumes:
        - name: cloudconfig
          hostPath:
            path: {{ .Values.cloudConfigPath }}
      {{- if .Values.magnumCABundlePath }}
        - name: ca-bundle
          hostPath:
            path: {{ .Values.magnumCABundlePath }}
      {{- end }}
      {{- end }}
```

Finally, there's the lines you've already mentioned in the `Values.yaml` which we should update at the same time.

tghartland:
> On roughly the same topic I wonder if we can reduce the duplication of the cloudconfig volume with something like:

I removed the duplication of the cloudconfig volume. Your suggestion needed a small change for the magnum CA bundle because I'm setting that value by default in the values file (it's the default location and shouldn't have to be changed). It just needs another check that the cloudprovider is magnum for that volume, otherwise it would be added for gce as well.

```yaml
{{- if and (eq .Values.cloudProvider "magnum") (.Values.magnumCABundlePath) }}
  - name: ca-bundle
    mountPath: {{ .Values.magnumCABundlePath }}
    readOnly: true
{{- end }}
```

For the autodiscovery comments in values.yaml I've gone through and added magnum where it refers to which cloud providers support autodiscovery. I'll take a look in the README next.

tghartland:
@gjtempleton I've made some changes in the README and values comments where things now apply to multiple providers.

I'll remove the WIP status from this PR, if you're happy with the changes then you can merge.

gjtempleton (changes requested):
I've realised that currently the PR workflow doesn't validate that running [helm-docs](https://github.com/norwoodj/helm-docs) results in no changes, I'll raise a PR to ensure that's the case going forward.

There's a number of small changes introduced by running it against your branch currently including the loss of the cloud provider instruction index, could you fix that up please?

tghartland:
That's my mistake, I didn't even realise the README file was generated. I was manually editing both the file and the template.

I've just pushed the README as generated with helm-docs. But even as I was editing the file and template, the pre-commit hook was never being run.
```
helmlint.............................................(no files to check)Skipped
Helm Docs............................................(no files to check)Skipped
```
Also, the pre-commit config specifies that helm-docs v0.13.0 is used. If the PR workflow will use that same version then it should be specified [here](https://github.com/kubernetes/autoscaler/tree/master/charts#install-hook-dependencies) when installing the dependency. I installed the latest version of helm-docs (using `go get`) by default and there were some changed section titles and formatting in the README, which would have failed the PR if it was being validated using the older version of helm-docs.

gjtempleton:
No worries, as you say, it should have been picked up by the pre-commit and PR hooks.

Thanks for making the changes.

/lgtm
/approve

tghartland:
@gjtempleton I pushed to rebase and fix the chart version conflict, you'll have to re-approve.

@vivekbagade we also need approval from an owner in the top level owners file because I've changed documentation in the magnum cloud provider as well.

gjtempleton (approved):
LGTM

PR #3428, opened 2020-08-13
Update autoscaler faq

gokula-krishna-dev:


gokula-krishna-dev:
I signed it

cluster-autoscaler/FAQ.md
```diff
@@ -701,7 +701,9 @@ CA doesn't remove underutilized nodes if they are running pods [that it shouldn'
 * there was a failed attempt to remove this particular node, in which case Cluster Autoscaler
   will wait for extra 5 minutes before considering it for removal again,
 
-* using large custom value for `--scale-down-delay-after-delete` or `--scan-interval`, which delays CA action.
+* using large custom value for `--scale-down-delay-after-delete` or `--scan-interval`, which delays CA action,
+
+* node is requested with high CPU/memory.
```
MaciekPytel, line 706: Can you clarify what you mean? In general scale-down depends on what type of pods run on the node and it shouldn't directly depend on how much CPU / memory the node has.

gokula-krishna-dev: @MaciekPytel understood

I observed this behavior on GKE. I have set the minimum nodes as 2. I deployed few pods with lower CPU/Memory request. Some pods were failing because they require more resource to startup after the startup they consume very less resource. I tried increasing the CPU/Memory request and deployed again, my pods started fine and cluster scaled up during the pod initialization. After the scale up it never scaled down (for days) unless I requested less Memory/CPU for all my pods even though they were consuming very less resources. I thought it is a feature that is not documented.

MaciekPytel: * Cluster Autoscaler operates only on resource requests. It doesn't look at all at real CPU/memory utilization at all. So the fact that it will not scale-down unless you change request is expected. If you want to autoscale based on real resource utilization you need to combine CA with HPA/VPA.
 * In GKE you can check why a node is not scaled-down by looking at https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-autoscaler-visibility#noscaledown-event.

MaciekPytel: One more thing that may come in useful - you can use `kubectl describe <node>` to see node utilization in terms of resource requests. That is the utilization that CA operates on.

gokula-krishna-dev: Thank you for your reply @MaciekPytel. I'll check the docs again.

PR #3431, opened 2020-08-14
Regenerate list of EC2 instances

kemra102:
Re-generate list of available EC2 instances from available metadata. This primarily pulls in support for `c5ad` & various `xxgd` instances.


Jeffwan:
I will need to compare EC2 lists. I will come back to you later

Jeffwan:
Verified it's correct.

/lgtm
/approve

Thanks!

PR #3435, opened 2020-08-18
Expose max-nodes-total as a metric

### LINKED ISSUE #3425: cluster-autoscaler expose max-nodes-total as a metric
Currently it is not easy to include the use `max-nodes-total`cli option within any metrics, except for putting it there manually.

It would be nice if the metrics endpoint also exposes the value passed to `max-nodes-total` as it could then easily be queried in prometheus e.g. to generate alerts if the amount of nodes is coming close to that value?

Why is the manual configuration not enough? `max-nodes-total` is used as a safe-guard from the cluster to go crazy and provision too many resources. But sometimes the newly demanded resources are really needed and one will tune that setting. Meaning it always need to be set in two places.

habibrosyad:
Closes #3425

/assign @MaciekPytel

### DIFF
--- a/cluster-autoscaler/metrics/metrics.go
+++ b/cluster-autoscaler/metrics/metrics.go
@@ -125,6 +125,14 @@ var (
 		},
 	)
 
+	maxNodesCount = k8smetrics.NewGauge(
+		&k8smetrics.GaugeOpts{
+			Namespace: caNamespace,
+			Name:      "max_nodes_count",
+			Help:      "Maximum number of nodes in all node groups",
+		},
+	)
+
 	/**** Metrics related to autoscaler execution ****/
 	lastActivity = k8smetrics.NewGaugeVec(
 		&k8smetrics.GaugeOpts{
@@ -257,6 +265,7 @@ func RegisterAll() {
 	legacyregistry.MustRegister(nodesCount)
 	legacyregistry.MustRegister(nodeGroupsCount)
 	legacyregistry.MustRegister(unschedulablePodsCount)
+	legacyregistry.MustRegister(maxNodesCount)
 	legacyregistry.MustRegister(lastActivity)
 	legacyregistry.MustRegister(functionDuration)
 	legacyregistry.MustRegister(functionDurationSummary)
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/metrics/metrics.go
```diff
@@ -125,6 +125,14 @@ var (
 		},
 	)
 
+	maxNodesTotal = k8smetrics.NewGauge(
+		&k8smetrics.GaugeOpts{
+			Namespace: caNamespace,
+			Name:      "max_nodes_total",
```
MaciekPytel, line 131: Can you rename to max_nodes_count? IIRC in prometheus metric convention (which CA follows) having '_total' in name is generally expected for counters and gauges should be suffixed with '_count'.
Also this way it will be consistent with nodes_count.

habibrosyad: Done

habibrosyad:
@MaciekPytel it seems the Travis CI is failing and I don't know how to restart the test...

habibrosyad:
Will try to close and reopen again.

/close

elmiko:
@habibrosyad it appears there is a non-deterministic bug in the unit test. i do not think it is related to your chage, but it seems to happen in about 5% of test runs.

habibrosyad:
@elmiko yes, I've checked the logs but nothing seems related to my changes. Just found out that closing and reopening PR could trigger the test again, so it's ok.

elmiko:
@habibrosyad ok, just wanted to make sure you knew =)

habibrosyad:
@elmiko thanks anyway, it's good to know why tho 😀

infa-anisrini:
@habibrosyad Is this metric available in v1.17 of official CA docker image ? Can you merge this to v1.17 branch of CA ?

infa-anisrini:
@habibrosyad  i cherry-picked your commit and merged to v1.17 CA, built a image. Now i see the max_nodes_count being exposed but its value is coming out to zero. Any idea ?

habibrosyad:
@infa-anisrini did you set the `max-nodes-total`? (default value is 0)

yasinlachiny:
It's not working

PR #3437, opened 2020-08-18
Avoid unwanted VMSS VMs caches invalidations

bpineau:
`fetchAutoAsgs()` is called at regular intervals, fetches a list of VMSS, then call `Register()` to cache each of those. That registration function will tell the caller wether that vmss' cache is outdated (when the provided VMSS, supposedly fresh, is different than the one held in cache) and will replace existing cache entry by the provided VMSS (which in effect will require a forced refresh since that ScaleSet struct is passed by fetchAutoAsgs with a nil lastRefresh time and an empty instanceCache).

To detect changes, `Register()` uses an `reflect.DeepEqual()` between the provided and the cached VMSS. Which does always find them different: cached VMSS were enriched with instances lists (while the provided one is blank, fresh from a simple vmss.list call). That DeepEqual is also fragile due to the compared structs containing mutexes (that may be held or not) and refresh timestamps, attributes that shouldn't be relevant to the comparison.

As a consequence, all Register() calls causes indirect cache invalidations and a costly refresh (VMSS VMS List). The number of Register() calls is directly proportional to the number of VMSS attached to the cluster, and can easily triggers ARM API throttling.

With a large number of VMSS, that throttling prevents `fetchAutoAsgs` to ever succeed (and cluster-autoscaler to start). ie.:
```
I0807 16:55:25.875907     153 azure_scale_set.go:344] GetScaleSetVms: starts
I0807 16:55:25.875915     153 azure_scale_set.go:350] GetScaleSetVms: scaleSet.Name: a-testvmss-10, vmList: []
E0807 16:55:25.875919     153 azure_scale_set.go:352] VirtualMachineScaleSetVMsClient.List failed for a-testvmss-10: &{true 0 2020-08-07 17:10:25.875447854 +0000 UTC m=+913.985215807 azure cloud provider throttled for operation VMSSVMList with reason "client throttled"}
E0807 16:55:25.875928     153 azure_manager.go:538] Failed to regenerate ASG cache: Retriable: true, RetryAfter: 899s, HTTPStatusCode: 0, RawError: azure cloud provider throttled for operation VMSSVMList with reason "client throttled"
F0807 16:55:25.875934     153 azure_cloud_provider.go:167] Failed to create Azure Manager: Retriable: true, RetryAfter: 899s, HTTPStatusCode: 0, RawError: azure cloud provider throttled for operation VMSSVMList with reason "client throttled"
```

From [`ScaleSet` struct attributes](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go#L74-L89) (manager, sizes, mutexes, refreshes timestamps) only sizes are relevant to that comparison. `curSize` is not strictly necessary, but comparing it will provide early instance caches refreshes (please tell me if you'd rather not have this here).

Example on a cluster with 219 attached VMSS (prev version was unthrottled at 12:29, modified CA rolled out at 12:38):
image
image

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_cache.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_cache.go
@@ -17,7 +17,6 @@ limitations under the License.
 package azure
 
 import (
-	"reflect"
 	"regexp"
 	"strings"
 	"sync"
@@ -62,7 +61,9 @@ func (m *asgCache) Register(asg cloudprovider.NodeGroup) bool {
 
 	for i := range m.registeredAsgs {
 		if existing := m.registeredAsgs[i]; strings.EqualFold(existing.Id(), asg.Id()) {
-			if reflect.DeepEqual(existing, asg) {
+			e := existing.(*ScaleSet)
+			a := asg.(*ScaleSet)
+			if e.minSize == a.minSize && e.maxSize == a.maxSize && e.curSize == a.curSize {
 				return false
 			}
 
@@ -181,7 +182,7 @@ func (m *asgCache) regenerate() error {
 
 	m.instanceToAsg = newCache
 
-	// Incalidating unowned instance cache.
+	// Invalidating unowned instance cache.
 	m.invalidateUnownedInstanceCache()
 
 	return nil
--- a/cluster-autoscaler/cloudprovider/azure/azure_manager.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_manager.go
@@ -513,7 +513,7 @@ func (m *AzureManager) buildAsgFromSpec(spec string) (cloudprovider.NodeGroup, e
 	case vmTypeStandard:
 		return NewAgentPool(s, m)
 	case vmTypeVMSS:
-		return NewScaleSet(s, m)
+		return NewScaleSet(s, m, -1)
 	case vmTypeAKS:
 		return NewAKSAgentPool(s, m)
 	default:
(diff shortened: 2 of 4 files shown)


marwanad:
@bpineau Thanks for this fix! On that note, the implementation of the caches within `azure_scale_set.go` isn't ideal because essentially every scaling group will have its own cache of the size, which makes the batched List VMSS calls that retrieve the size redundant. I'm planning on redesigning this flow .


/lgtm
/approve

bpineau:
Thanks @marwanad  !
Very true, that makes two sources of truth for vmss sizes.

One good aspect of the VMSS List (as op. to the many VMSS VMs Lists) is it's constant time (one call every vmssCacheTTL seconds), while we're suffering from the many VMSS VM List calls, O(number of vmss).
In that sense, it's a cheaper way to check for VMSSes freshness. Is that something informing the redesign?

I'm working on adding an optional jitter to spread those vmssvm.list calls, should help us survive until the redesign lands in.

marwanad:
@bpineau 

Yup, so the redesign will address [VMSS LIST](https://docs.microsoft.com/en-us/rest/api/compute/virtualmachinescalesets/list) and have the size source of truth in a single global cache. This is going to be O(1) every vmssCacheTTL. Today, unfortunately it's O(numberOfVMSS) every `vmssCacheTTL`.

This acts as guard to the VMSS VM cache in a sense as well because if the sizes are the same, there's no need to fetch new VM instances.

bpineau:
@marwanad Amazing! That redesign will be very useful for us, and a great relieve!

PR #3440, opened 2020-08-19
Azure: optional jitter on initial VMSS VM cache refresh

bpineau:
On (re)start, cluster-autoscaler will refresh all VMSS instances caches at once, and set those cache TTL to 5mn. All VMSS VM List calls (for VMSS discovered at boot) will then continuously hit ARM API at the same time, potentially causing regular throttling bursts.

Exposing an optional jitter subtracted from the initial first scheduled refresh delay will splay those calls (except for the first one, at start), while keeping the predictable (max. 5mn, unless the VMSS changed) refresh interval after the first refresh.

While at it, also expose an option to configure that VMSS cache TTL.

image


marwanad:
/lgtm

I think failures are from CAPI, closing and reopening should retrigger travis.

feiskyer (commented):
Thanks for the improvement.

/lgtm
/approve

PR #3443, opened 2020-08-20
Bypass comparing node group resources if the hardware config is the same

leonsodhi-lf:
https://github.com/Leadfeeder/autoscaler/pull/1 for Autoscaler 1.17.3


leonsodhi-lf:
Sorry, wrong repo.

PR #3446, opened 2020-08-21
Cherry-pick onto 1.16: Azure Instance generation changes

marwanad:
Cherry-pick #3311 and #2871


marwanad:
Seems we need owner approval. I'll drop the `gen` code changes since they shouldn't be needed

PR #3456, opened 2020-08-24
VPA: Ignore admission hook failures

johanneswuerbach:
When switching the VPA admission controller to the v1 admissionregistration API, the FailurePolicy was changed from Ignore to Fail.

[v1beta1 type](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/admissionregistration/v1beta1/types.go#L212-L215) vs [v1 type](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/admissionregistration/v1/types.go#L194-L197) (the change is the "Defaults to")

Using Fail makes pod creation depend on VPA and can cause a deadlock when the admission controller pod itself can't be scheduled again.


bskiba:
Thanks! 
/lgtm
/approve

PR #3459, opened 2020-08-25
Call nodeGroup API to avoid type assertion errors

marwanad:
Previously, it would assume every ASG is of `ScaleSet` and this would possibly for panic for the `vmType` that's not VMSS.

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_cache.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_cache.go
@@ -55,27 +55,25 @@ func newAsgCache() (*asgCache, error) {
 }
 
 // Register registers a node group if it hasn't been registered.
-func (m *asgCache) Register(asg cloudprovider.NodeGroup) bool {
+func (m *asgCache) Register(newAsg cloudprovider.NodeGroup) bool {
 	m.mutex.Lock()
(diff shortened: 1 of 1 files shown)


marwanad:
/hold 

@bpineau Does that change look reasonable to you? Just want to confirm - The `.TargetSize()` call could possibly lead to an API call past the cache TTL.

bpineau:
Thanks for the fix @marwanad !

I wonder if we shouldn't just ditch that "current size" comparison altogether?
It doesn't represents the "ASG" envelope (ie. matter for registration).
I added it as a cheap way to (potentially) speed up refresh because that value was already there handy.

But that comparison is actually already done by `getCurSize()` and [`Nodes()` calls](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go#L695-L698). Everything reads the scalesets size through `getCurSize` (like `TargetSize` does) which also does instances cache invalidation on size change.

marwanad:
@bpineau very good point. Registration should be considered with configurable specs. Name, MinSize and MaxSize. I'll update the PR to drop it.

marwanad:
@bpineau updated.

/hold cancel

PR #3460, opened 2020-08-26
Break up the logic in azure_manager 

marwanad:
A lot of the logic in `azure_manager.go` concerns creating the config object. I moved this to a separate.

Similarly, the autodiscovery stuff was moved out. There's one logic change related to autodiscovery which I highlighted below.


cluster-autoscaler/cloudprovider/azure/azure_manager.go
```diff
@@ -643,21 +254,11 @@ func (m *AzureManager) getFilteredAutoscalingGroups(filter []labelAutoDiscoveryC
...
-	case vmTypeAKS:
-		return nil, nil
-	default:
-		err = fmt.Errorf("vmType %q not supported", m.config.VMType)
-	}
-	if err != nil {
-		return nil, err
+	if m.config.VMType == vmTypeVMSS {
+		return m.listScaleSets(filter)
 	}
```
marwanad, line 259: @feiskyer I wanted to highlight this diff - was intentional to have the VM standard type as a valid type for autodisocery?

It seems that the `filter` was never used in listAgentPools so how do the min and max get populated? That's the only logic change in this PR - basically error out when trying to use autodiscovery for a non-supported `VMType`

feiskyer: It should be a bug, only vmss support autodisocery.

feiskyer (commented):
/lgtm
/approve

PR #3462, opened 2020-08-26
[WIP] Add functionality to dynamically reload the VPA certs

surajnarwade:
This PR adds the functionality of dynamically reloading the tls certificates for VPA admission controller.

**How does it work?**

* cert-manager or any other mechanism updates the certs in secrets, `fsnotify` watcher will watch at `/etc/tls-certs` (where secrets are mounted) and triggers an event if certs are updated

* Once, certs are updated,  [reload](https://github.com/kubernetes/autoscaler/pull/3462/files#diff-163de8317f97c4ca27159e13ae6ce383R111) function will again reload new certs

* callback hook `GetCertificate` in `tls.Config` is used to get the new certs

**Reference**

* https://stackoverflow.com/a/40883377/7804592

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/certs.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/certs.go
@@ -17,10 +17,23 @@ limitations under the License.
 package main
 
 import (
+	"crypto/tls"
 	"io/ioutil"
+	"sync"
 
+	"github.com/fsnotify/fsnotify"
 	"k8s.io/klog"
 )
+
+// KeypairReloader structs holds cert path and certs
+type KeypairReloader struct {
+	certMu   sync.RWMutex
+	cert     *tls.Certificate
+	caCert   []byte
+	certPath string
+	keyPath  string
+	caPath   string
+}
 
 type certsContainer struct {
 	caCert, serverKey, serverCert []byte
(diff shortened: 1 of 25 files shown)


vertical-pod-autoscaler/pkg/admission-controller/certs.go
```diff
@@ -41,10 +54,75 @@ func readFile(filePath string) []byte {
...
+		return nil, err
+	}
+
+	defer func() {
+		if err != nil {
+			watcher.Close()
+		}
+	}()
+
+	if err := watcher.Add("/etc/tls-certs"); err != nil {
```
arjunrn, line 82: You probably don't want this hardcoded. You should get the directories of both the cert and key and set watches on them.

arjunrn:
How about a test where an `KeyReloader` is the configuration files are changed and then the test verifies that the key and cert have been reloaded correctly.

vertical-pod-autoscaler/pkg/admission-controller/certs.go
bskiba: +1 I think you want those as flags passed in to the KeyPairReloader

vertical-pod-autoscaler/pkg/admission-controller/certs.go
```diff
@@ -41,10 +54,78 @@ func readFile(filePath string) []byte {
...
+	}
+
+	go func() {
+		for {
+			select {
+			// watch for events
+			case event := <-watcher.Events:
+				// fsnotify.create events will tell us if there are new certs
+				if event.Op&fsnotify.Create == fsnotify.Create {
+					klog.Info("Reloading certs")
```
bskiba, line 93: Can you add verbosity level to the log? And the log could be "New certificate found, reloading certs"

vertical-pod-autoscaler/pkg/admission-controller/certs.go
```diff
@@ -41,10 +54,78 @@ func readFile(filePath string) []byte {
...
+					if err := result.reload(); err != nil {
+						klog.Infof("Could not load new certs: %v", err)
```
bskiba, line 95: This should probably be Warning

vertical-pod-autoscaler/pkg/admission-controller/certs.go
```diff
@@ -99,7 +98,7 @@ func NewKeypairReloader(config certsConfig) (*KeypairReloader, error) {
 
 				// watch for errors
 			case err := <-watcher.Errors:
-				klog.Infof("error", err)
+				klog.Infof("error: %v", err)
```
bskiba, line 101: We need more descriptive log here. Probably also warning or error level. 
Is there any action to be taken on observing errors from the watcher?

vertical-pod-autoscaler/pkg/admission-controller/main.go
```diff
@@ -77,7 +78,11 @@ func main() {
 	metrics.Initialize(*address, healthCheck)
 	metrics_admission.Register()
 
-	certs := initCerts(*certsConfiguration)
+	// load certs
+	kpr, err := NewKeypairReloader(*certsConfiguration)
+	if err != nil {
+		klog.Fatal(err)
```
bskiba, line 84: More info in the log needed, like "Failed to load certificate on startup: %v", err

vertical-pod-autoscaler/pkg/admission-controller/main.go
```diff
@@ -127,13 +132,16 @@ func main() {
 	})
 	clientset := getClient()
 	server := &http.Server{
-		Addr:      fmt.Sprintf(":%d", *port),
-		TLSConfig: configTLS(clientset, certs.serverCert, certs.serverKey),
+		Addr: fmt.Sprintf(":%d", *port),
 	}
+	// this will check if there are new certs before every tls handshake
+	t := &tls.Config{GetCertificate: kpr.GetCertificateFunc()}
```
bskiba, line 138: server.TLSConfig = &tls.Config{GetCertificate: kpr.GetCertificateFunc()}

you can also do it above:
server := &http.Server{
		Addr:      fmt.Sprintf(":%d", *port),			
		TLSConfig: &tls.Config{GetCertificate: kpr.GetCertificateFunc()},	
}

One more thing is that the KeypairReloader could have a GetTLSConfig function and then we could simply have:

server := &http.Server{
		Addr:      fmt.Sprintf(":%d", *port),			
		TLSConfig: kpr.GetTLSConfig(),
}

vertical-pod-autoscaler/pkg/admission-controller/main.go
```diff
@@ -127,13 +132,16 @@ func main() {
 	})
 	clientset := getClient()
 	server := &http.Server{
-		Addr:      fmt.Sprintf(":%d", *port),
-		TLSConfig: configTLS(clientset, certs.serverCert, certs.serverKey),
+		Addr: fmt.Sprintf(":%d", *port),
 	}
+	// this will check if there are new certs before every tls handshake
```
bskiba, line 137: Actually, this will return up-to-date cert before every handshake, right?

vertical-pod-autoscaler/pkg/admission-controller/certs.go
```diff
@@ -41,10 +54,75 @@ func readFile(filePath string) []byte {
...
+	}
+	kpr.certMu.Lock()
+	defer kpr.certMu.Unlock()
+	kpr.cert = &newCert
+	return nil
+}
+
+// GetCertificateFunc will return function which will be used as tls.Config.GetCertificate
+func (kpr *KeypairReloader) GetCertificateFunc() func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
+	return func(clientHello *tls.ClientHelloInfo) (*tls.Certificate, error) {
```
bskiba, line 123: Since clientHello is unused, you can replace it with underscore

bskiba (changes requested):
Left some comments. I would also like this new behavior hidden behind the flag, so that the default behavior would still be not starting a watch for certs.

Can you describe how you tested this change?

mwielgus:
Ping.

vertical-pod-autoscaler/pkg/admission-controller/certs.go
surajnarwade: sure, I will update it :+1:

vertical-pod-autoscaler/pkg/admission-controller/certs.go
surajnarwade: sure :+1:

surajnarwade:
@bskiba , @mwielgus thanks for all the reviews, sorry could not get back to PR earlier. I will update the PR now

vertical-pod-autoscaler/pkg/admission-controller/main.go
surajnarwade: yeah, that's correct, I've updated the comment :)

mwielgus:
Closing due to inactivity. The PR has been WIP for 3 months.

mwielgus:
Feel free to reopen if needed.

PR #3463, opened 2020-08-26
Fix limit rounding

### LINKED ISSUE #3319: [VPA] Pod scheduled with memory limit above limitrange
Hi there,

We're seeing an issue with pods that have more than one container which VPA recommends and updates memory requests/limits for. The pods are being scheduled with memory limits that are above the pod memory limit configured in the namespace's limitrange.

We don't see this issue when a pod only has one container - VPA updates the memory limit to be within the limitrange. The issue only seems to occur when there are multiple containers in a pod which VPA updates requests/limits for.

Also, I should mention that we're not configuring any `minAllowed` or `maxAllowed` limits via container policies in the VPA resource policy, so we don't expect there to be any conflicts which would cause VPA to set limits above the limit range. In case it makes a difference, we are setting a request/limit ratio of 1:1 when initially deploying the pod. 

When looking at the replicaset when the pod fails to schedule, the pod is scheduled with a total memory limit which is 1 byte over the limit imposed by the limitrange.

E.g:

Namespace limitrange has a memory limit of 115Gi (== 123480309760 bytes)
```
Type        Resource  Min  Max    Default Request  Default Limit  Max Limit/Request Ratio
----        --------  ---  ---    ---------------  -------------  -----------------------
Pod         memory    5Mi  115Gi  -                -              -
Container   memory    5Mi  -      256Mi            512Mi          -
```
VPA recommends memory request for two containers in a pod which totals above the limitrange limit (as expected)
```
    Container Recommendations:
      Container Name:  oomps
      Lower Bound:
        Cpu:     407m
        Memory:  54651563825
      Target:
        Cpu:     813m
        Memory:  130488397494
      Uncapped Target:
        Cpu:     813m
        Memory:  130488397494
      Upper Bound:
        Cpu:           195933m
        Memory:        31447703796054
      Container Name:  oomps-2
      Lower Bound:
        Cpu:     838m
        Memory:  1823741057
      Target:
        Cpu:     1168m
        Memory:  2539377048
      Uncapped Target:
        Cpu:     1168m
        Memory:  2539377048
      Upper Bound:
        Cpu:     211408m
        Memory:  459627245688
Events:          <none>
```
Pod seems to be updated with requests/limits that attempt to fall more in-line with the limitrange, but fails to schedule due to total pod memory limit being 1 byte above the limitrange:
```
Events:
  Type     Reason            Age                From                   Message
  ----     ------            ----               ----                   -------
<truncated>
  Warning  FailedCreate      66s                replicaset-controller  Error creating: pods "rhys-oomps-6676478b6-vj89d" is forbidden: maximum memory usage per Pod is 115Gi, but limit is 123480309761
  Warning  FailedCreate      26s (x9 over 65s)  replicaset-controller  (combined from similar events): Error creating: pods "rhys-oomps-6676478b6-zb8fv" is forbidden: maximum memory usage per Pod is 115Gi, but limit is 123480309761

```

As the pods are consistently scheduled with only 1 byte over the limitrange, it makes me think there may be an error in the calculation that updates requests/limits for multiple containers in a pod, while trying to keep the total requests of the pod within the limitrange. I'm not sure though, so if I am missing something I'd be really grateful for any help!

jbartosik:
fixes #3319

### DIFF
--- a/vertical-pod-autoscaler/pkg/utils/vpa/capping.go
+++ b/vertical-pod-autoscaler/pkg/utils/vpa/capping.go
@@ -371,9 +371,11 @@ func applyPodLimitRange(resources []vpa_types.RecommendedContainerResources,
 	if minLimit.Cmp(sumRecommendation) > 0 && !sumLimit.IsZero() {
 		for i := range pod.Spec.Containers {
 			request := (*fieldGetter(resources[i]))[resourceName]
-			cappedContainerRequest, _ := scaleQuantityProportionally(&request, &sumRecommendation, &minLimit)
+			var cappedContainerRequest *resource.Quantity
 			if resourceName == apiv1.ResourceMemory {
-				cappedContainerRequest.RoundUp(resource.Scale(0))
+				cappedContainerRequest, _ = scaleQuantityProportionally(&request, &sumRecommendation, &minLimit, roundUpToFullUnit)
+			} else {
+				cappedContainerRequest, _ = scaleQuantityProportionally(&request, &sumRecommendation, &minLimit, noRounding)
 			}
 			(*fieldGetter(resources[i]))[resourceName] = *cappedContainerRequest
 		}
@@ -393,11 +395,11 @@ func applyPodLimitRange(resources []vpa_types.RecommendedContainerResources,
 	}
 	for i := range pod.Spec.Containers {
 		limit := (*fieldGetter(resources[i]))[resourceName]
-		cappedContainerRequest, _ := scaleQuantityProportionally(&limit, &sumLimit, &targetTotalLimit)
+		var cappedContainerRequest *resource.Quantity
 		if resourceName == apiv1.ResourceMemory {
-			// Round down
-			cappedContainerRequest.Sub(*resource.NewMilliQuantity(999, cappedContainerRequest.Format))
-			cappedContainerRequest.RoundUp(resource.Scale(0))
+			cappedContainerRequest, _ = scaleQuantityProportionally(&limit, &sumLimit, &targetTotalLimit, roundDownToFullUnit)
+		} else {
+			cappedContainerRequest, _ = scaleQuantityProportionally(&limit, &sumLimit, &targetTotalLimit, noRounding)
 		}
 		(*fieldGetter(resources[i]))[resourceName] = *cappedContainerRequest
 	}
--- a/vertical-pod-autoscaler/pkg/utils/vpa/capping.go
+++ b/vertical-pod-autoscaler/pkg/utils/vpa/capping.go
@@ -372,6 +372,9 @@ func applyPodLimitRange(resources []vpa_types.RecommendedContainerResources,
 		for i := range pod.Spec.Containers {
 			request := (*fieldGetter(resources[i]))[resourceName]
 			cappedContainerRequest, _ := scaleQuantityProportionally(&request, &sumRecommendation, &minLimit)
+			if resourceName == apiv1.ResourceMemory {
+				cappedContainerRequest.RoundUp(resource.Scale(0))
+			}
 			(*fieldGetter(resources[i]))[resourceName] = *cappedContainerRequest
 		}
 		return resources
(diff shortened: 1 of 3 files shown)


jbartosik:
/assign @bskiba 

@rhysemmas FYI

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
```diff
@@ -391,6 +394,11 @@ func applyPodLimitRange(resources []vpa_types.RecommendedContainerResources,
 	for i := range pod.Spec.Containers {
 		limit := (*fieldGetter(resources[i]))[resourceName]
 		cappedContainerRequest, _ := scaleQuantityProportionally(&limit, &sumLimit, &targetTotalLimit)
+		if resourceName == apiv1.ResourceMemory {
+			// Round down
```
jbartosik, line 398: I couldn't find a better way to round down, if you know one I welcome suggestions :)

bskiba: Wouldn't AsScale work? https://github.com/kubernetes/autoscaler/blob/02d67527d779beed73916362bcb3e334959690e8/vertical-pod-autoscaler/vendor/k8s.io/apimachinery/pkg/api/resource/quantity.go#L502

bskiba (commented):
Wouldn't it be cleaner if scaleQuantityProportionally was aware of resource type and worked on Value() instead of MIlliValue() for memory?

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
jbartosik: `AsScale` rounds up too

https://github.com/kubernetes/autoscaler/blob/02d67527d779beed73916362bcb3e334959690e8/vertical-pod-autoscaler/vendor/k8s.io/apimachinery/pkg/api/resource/quantity.go#L500

jbartosik:
> Wouldn't it be cleaner if scaleQuantityProportionally was aware of resource type and worked on Value() instead of MIlliValue() for memory?

I added a commit moving this logic inside the `scaleQuantityProportionally` function. I like this approach better. I also considered passing 2 pieces of information (resource and direction for rounding) separately but passing rounding details as an enum look cleaner

vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go
```diff
@@ -103,21 +103,37 @@ func GetBoundaryRequest(originalRequest, originalLimit, boundaryLimit, defaultLi
...
+	noRounding roundingMode = iota
+	roundUpToFullUnit
+	roundDownToFullUnit
+)
+
 // scaleQuantityProportionally returns value which has the same proportion to scaledQuantity as scaleResult has to scaleBase
 // It also returns a bool indicating if it had to cap result to MaxInt64 milliunits.
-func scaleQuantityProportionally(scaledQuantity, scaleBase, scaleResult *resource.Quantity) (*resource.Quantity, bool) {
+func scaleQuantityProportionally(scaledQuantity, scaleBase, scaleResult *resource.Quantity, rounding roundingMode) (*resource.Quantity, bool) {
 	originalMilli := big.NewInt(scaledQuantity.MilliValue())
```
bskiba, line 121: If we worked on Value() for memory instead of MilliValue, would we still need rounding?

vertical-pod-autoscaler/pkg/utils/vpa/capping.go
bskiba: I see, that's a bit weird 🤷‍♀️

vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go
jbartosik: We wouldn't have to worry about rounding down. But we'd have to worry about rounding up.

jbartosik (commented):
I saw another comment in an email notification but don't see it on the PR. I presume you closed it

jbartosik:
@bskiba please take a look

PR #3470, opened 2020-08-28
Add Exoscale provider Cluster Autoscaler

pierre-emmanuelJ:
This PR introduce [Exoscale](https://exoscale.com) Cloud Provider to Cluster Autoscaler (CA) Controller.

If you need some resources to test the CA at Exoscale, let us know.

For more details you can follow the [Readme](https://github.com/exoscale/autoscaler-1/blob/master/cluster-autoscaler/cloudprovider/exoscale/README.md) and ask some question in this PR.


pierre-emmanuelJ:
I'm trying to update the vendor using `cluster-autoscaler/hack/update-vendor.sh` to add `github.com/exoscale/egoscale v0.31.1` dependencies, if you have any guidelines on it, it would be a pleasure :)

cluster-autoscaler/go.mod-extra
```diff
@@ -4,6 +4,7 @@ go 1.12
 require (
   github.com/digitalocean/godo v1.27.0
   github.com/rancher/go-rancher v0.1.0
+  github.com/exoscale/egoscale v0.31.1
```
mwielgus, line 7: Please do not add new dependencies to Cluster Autoscaler. Please inline all necessary libraries in the cloud provider directory. Example:

https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/baiducloud

mwielgus (changes requested):

cluster-autoscaler/go.mod-extra
pierre-emmanuelJ: Hello :) Since DigitalOcean is in the vendoring:
https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/go.mod-extra#L5.
Do you think it's possible to do the same for us only for `github.com/exoscale/egoscale`?

Because moving the `github.com/exoscale/egoscale` package into the `cloudprovider/exoscale` directory is a bit heavy, (modifying import path in the whole project, addition of the license header on each file...etc ). I think you can understand It's not very maintainable for us if we want to update something...etc.

Thank you @mwielgus

mwielgus: Sorry, but we cannot accept more dependencies. The main reason is that at some point we will (and we actually had in the past) have an unsolvable case when provider A requires library in version X, provider B requires library in version Y and  X is incompatible with Y. For that reason we ask all new cloud providers to vendor whatever they need into their directory.

pierre-emmanuelJ: okay, I understand, so we'll vendorise it into the Exoscale directory.

falzm: All right @mwielgus, we've amended our PR to ship our internal dependencies under our cloudprovider directory – please let us know if that's OK for you.

mwielgus (approved):
/lgtm
/approve

PR #3473, opened 2020-08-28
CA: fix integer to string conversion

### LINKED ISSUE #3472: cluster-autoscaler: go test fails on go 1.15
Go 1.15 enabled _stringintconv_ as additional `go vet` check. It checks if a int is converted into a string expecting the number as string instead of getting the representing rune.

These check is violated within:

https://github.com/kubernetes/autoscaler/blob/d9e8fa3cbba3c3c76adf96e70bbb1277fad60ea7/cluster-autoscaler/cloudprovider/alicloud/alibaba-cloud-sdk-go/sdk/auth/signers/signer_ecs_ram_role.go#L126

and

https://github.com/kubernetes/autoscaler/blob/d9e8fa3cbba3c3c76adf96e70bbb1277fad60ea7/cluster-autoscaler/cloudprovider/huaweicloud/huawei-cloud-sdk-go/errors.go#L164-L166

Caused by these failing checks it's impossible to run `hack/update-vendor.sh` anymore ash the invoked command `go test -mod=vendor ./...` is failing. IMHO `go vet` is completely right and in both cases the author intended to receive the int as string instead of getting the rune which is represented by the integer code-point.

tobiaskohlbau:
Fixes bug reported by go vet check stringintconv. In go casting an
integer to string does not result in the string representation of
the number, instead the result is a rune representing
the codepoint of that number.

Fixes #3472

### DIFF
--- a/cluster-autoscaler/cloudprovider/alicloud/alibaba-cloud-sdk-go/sdk/auth/signers/signer_ecs_ram_role.go
+++ b/cluster-autoscaler/cloudprovider/alicloud/alibaba-cloud-sdk-go/sdk/auth/signers/signer_ecs_ram_role.go
@@ -19,13 +19,14 @@ package signers
 import (
 	"encoding/json"
 	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
 	"github.com/jmespath/go-jmespath"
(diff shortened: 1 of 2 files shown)


mwielgus (approved):
/lgtm
/approve

PR #3498, opened 2020-09-09
Don't fetch nodes

jbartosik:
I see pods that have Nodes as their owners in K8s `1.17` and `1.18`. We shouldn't try to fetch nodes, we know it's not a controller

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
@@ -45,13 +45,14 @@ import (
 type wellKnownController string
 
 const (
+	cronJob               wellKnownController = "CronJob"
 	daemonSet             wellKnownController = "DaemonSet"
 	deployment            wellKnownController = "Deployment"
-	replicaSet            wellKnownController = "ReplicaSet"
-	statefulSet           wellKnownController = "StatefulSet"
-	replicationController wellKnownController = "ReplicationController"
+	node                  wellKnownController = "Node"
 	job                   wellKnownController = "Job"
-	cronJob               wellKnownController = "CronJob"
+	replicaSet            wellKnownController = "ReplicaSet"
+	replicationController wellKnownController = "ReplicationController"
+	statefulSet           wellKnownController = "StatefulSet"
 )
 
 const (
@@ -251,6 +252,9 @@ func (f *controllerFetcher) isWellKnownOrScalable(key *ControllerKeyWithAPIVersi
 	if f.isWellKnown(key) {
 		return true
 	}
+	if gk, err := key.groupKind(); err != nil && wellKnownController(gk.Kind) == node {
+		return false
+	}
 
 	//if not well known check if it supports scaling
 	groupKind, err := key.groupKind()
@@ -276,11 +280,16 @@ func (f *controllerFetcher) isWellKnownOrScalable(key *ControllerKeyWithAPIVersi
 }
 
 func (f *controllerFetcher) getOwnerForScaleResource(groupKind schema.GroupKind, namespace, name string) (*ControllerKeyWithAPIVersion, error) {
+	if wellKnownController(groupKind.Kind) == node {
+		// Some pods specify nods as their owners. This causes performance problems
+		// in big clusters when VPA tries to get all nodes. We know nodes aren't
+		// valid controllers so we can skip trying to fetch them.
+		return nil, fmt.Errorf("node is not a valid owner")
+	}
 	mappings, err := f.mapper.RESTMappings(groupKind)
 	if err != nil {
 		return nil, err
 	}
-
 	var lastError error
 	for _, mapping := range mappings {
 		groupResource := mapping.Resource.GroupResource()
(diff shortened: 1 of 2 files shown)


vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -45,13 +45,14 @@ import (
 type wellKnownController string
 
 const (
+	cronJob               wellKnownController = "CronJob"
 	daemonSet             wellKnownController = "DaemonSet"
 	deployment            wellKnownController = "Deployment"
+	node                  string              = "Node"
```
kgolab, line 51: Why string and not wellKnownController here?

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -267,6 +268,9 @@ func (f *controllerFetcher) isWellKnownOrScalable(key *ControllerKeyWithAPIVersi
 
 	for _, mapping := range mappings {
 		groupResource := mapping.Resource.GroupResource()
+		if gk, err := key.groupKind(); err != nil && gk.Kind == node {
```
kgolab, line 271: Why here and not above (around line 256, where the comment "if not well known" is)?
It seems that key does not change between there & here.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -284,6 +288,10 @@ func (f *controllerFetcher) getOwnerForScaleResource(groupKind schema.GroupKind,
 	var lastError error
 	for _, mapping := range mappings {
 		groupResource := mapping.Resource.GroupResource()
+		if groupKind.Kind == node {
```
kgolab, line 291: Same as above, why inside the loop?

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
jbartosik: Done

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
jbartosik: Done

jbartosik:
@kgolab  @bskiba PTAL

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
```diff
@@ -340,8 +353,31 @@ func TestControllerFetcher(t *testing.T) {
...
+		{
+			name: "pod, parent is node",
+			objects: []runtime.Object{&appsv1.Deployment{
+				TypeMeta: metav1.TypeMeta{
+					Kind: "Deployment",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-deployment",
+					Namespace: "test-namesapce",
+					// Parent that support scale subresource and is not well known
```
bskiba, line 365: nit: s/support/supports/

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -276,11 +280,14 @@ func (f *controllerFetcher) isWellKnownOrScalable(key *ControllerKeyWithAPIVersi
 }
 
 func (f *controllerFetcher) getOwnerForScaleResource(groupKind schema.GroupKind, namespace, name string) (*ControllerKeyWithAPIVersion, error) {
+	if wellKnownController(groupKind.Kind) == node {
+		// Node is not a proper owner
```
bskiba, line 284: Can you add more context? I.e. that this is actually a scalability issue in bigger clusters. Also is it not a problem that we return no error but parent is nil? Shouldn't this be an error, since we failed to find a proper controller?

jbartosik:
@kgolab @bskiba please take a look

bskiba:
I left some comments today

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher_test.go
jbartosik: I replaced the comment with a more accurate one

jbartosik (commented):
> I left some comments today

I didn't notice, Github only shows those comments to me only when I go to the Files Changes tab. Thanks for the review and please excuse the unnecessary ping.

bskiba:
Thanks!
/lgtm
/approve

bskiba:
> > I left some comments today
> 
> I didn't notice, Github only shows those comments to me only when I go to the Files Changes tab. Thanks for the review and please excuse the unnecessary ping.

No problem, I do that all the time 🙄

PR #3505, opened 2020-09-11
Cherry-pick #2207 add kops instance group label to ignore list for similar node group identification.

whwalter:
Cherry-picking ignored labels that were lost from [1.15.5 to 1.16](https://github.com/kubernetes/autoscaler/blob/c6929a2fa9632eb10eaad71621096f35eee7a736/cluster-autoscaler/processors/nodegroupset/compare_nodegroups.go#L100). 

The MaxCapacityMemoryDifferenceRatio changes were needed to correct balance-similar-node-groups behavior on AWS but the removal of the labels prevents it working properly with kops. I don't think BalancingExtraIgnoredLabels is available into 1.19 so we are stuck with this unless that gets backported.

I didn't add tests for the labels to keep this as just a cherry-pick of existing commits.


mwielgus (approved):
/lgtm
/approve

PR #3509, opened 2020-09-14
correctly label failed scale ups error

marwanad:
Errors due to `IncreaseSize` should be of type "cloudProviderError" and not "apiError".


mwielgus (approved):
/lgtm
/approve

PR #3512, opened 2020-09-15
Add recommendation cpu and memory percentile flags

chenchun:


vertical-pod-autoscaler/pkg/recommender/logic/recommender.go
```diff
@@ -20,12 +20,20 @@ import (
...
 	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/recommender/model"
+	"k8s.io/klog"
 )
 
 var (
 	safetyMarginFraction = flag.Float64("recommendation-margin-fraction", 0.15, `Fraction of usage added as the safety margin to the recommended request`)
 	podMinCPUMillicores  = flag.Float64("pod-recommendation-min-cpu-millicores", 25, `Minimum CPU recommendation for a pod`)
 	podMinMemoryMb       = flag.Float64("pod-recommendation-min-memory-mb", 250, `Minimum memory recommendation for a pod`)
+
+	targetCPUPercentile        = flag.Float64("recommendation-target-cpu-percentile", 0.9, `Percentile of cpu usage histogram for target recommendation for a pod`)
```
bskiba, line 31: nit: since VPA recommends per container, it's more accurate to say "recommendation for a container" here and below. (The podMin... flags above do target a whole pod, hence the "for a pod" in descriptions above)

vertical-pod-autoscaler/pkg/recommender/logic/recommender.go
```diff
@@ -99,13 +107,19 @@ func FilterControlledResources(estimation model.Resources, controlledResources [
...
-	lowerBoundMemoryPeaksPercentile := 0.5
-	upperBoundMemoryPeaksPercentile := 0.95
+	if lowerBoundCPUPercentile <= 0 || lowerBoundCPUPercentile > targetCPUPercentile || targetCPUPercentile > upperBoundCPUPercentile {
+		klog.Fatalf("Please make sure lowerBoundCPUPercentile > 0 and lowerBoundCPUPercentile <= targetCPUPercentile <= upperBoundCPUPercentile")
+	}
+	targetMemoryPeaksPercentile := *targetMemoryPercentile
+	lowerBoundMemoryPeaksPercentile := *lowerBoundMemoryPercentile
+	upperBoundMemoryPeaksPercentile := *upperBoundMemoryPercentile
+	if lowerBoundMemoryPeaksPercentile <= 0 || lowerBoundMemoryPeaksPercentile > targetMemoryPeaksPercentile || targetMemoryPeaksPercentile > upperBoundMemoryPeaksPercentile {
+		klog.Fatalf("Please make sure lowerBoundMemoryPercentile > 0 and lowerBoundMemoryPercentile <= targetMemoryPercentile <= upperBoundMemoryPercentile")
```
bskiba, line 121: Can you add this information in flag descriptions? For example
lowerBoundCPUPercentile    = flag.Float64("recommendation-lower-bound-cpu-percentile", 0.5, `Percentile of cpu usage histogram for lower bound recommendation for a container. Needs to be lower than recommendation-target-cpu-percentile`)

bskiba (commented):
Left some comments. 

In general changing the percentiles of the recommender may result in a completely different recommendation behavior and we never tested for that. If we merge this PR, we'll need some well-exposed information that changing these settings leaves you in an untested territory and there is little guarantees we can give. Can you give some context on why you want to modify these parameters?

mwielgus:
Closing due to inactivity (review comments not answered). Feel free to reopen once the comments are addressed.

PR #3532, opened 2020-09-22
Azure: support allocatable resources overrides via VMSS tags

bpineau:
This allows specifying effective nodes resources capacities using Scale
Sets tags, preventing wrong CA decisions and infinite upscale when pods
requests are within instance type capacity but over k8s nodes allocatable
(which might comprise system and kubelet's reservations), and when using
node-infos inferred from instances templates (ie. scaling from 0).

This is similar to what AWS (with launch configuration tags) and
GCP (with instances templates metadata) cloud providers offers,
ensuring the tags format is similar to AWS' for consistency.

See also:
https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/proposals/min_at_zero_gcp.md


feiskyer (commented):
/lgtm
/approve

PR #3546, opened 2020-09-28
Extend GetContainersResources with an option to emit all containers.

kgolab:
Includes some minor clean-up in test of the public interface.

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/resource/pod/recommendation/recommendation_provider.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/resource/pod/recommendation/recommendation_provider.go
@@ -46,23 +46,29 @@ func NewProvider(calculator limitrange.LimitRangeCalculator,
 }
 
 // GetContainersResources returns the recommended resources for each container in the given pod in the same order they are specified in the pod.Spec.
+// If addAll is set to true, containers w/o a recommendation are also added to the list, otherwise they're skipped (default behaviour).
 func GetContainersResources(pod *core.Pod, vpaResourcePolicy *vpa_types.PodResourcePolicy, podRecommendation vpa_types.RecommendedPodResources, limitRange *core.LimitRangeItem,
-	annotations vpa_api_util.ContainerToAnnotationsMap) []vpa_api_util.ContainerResources {
+	addAll bool, annotations vpa_api_util.ContainerToAnnotationsMap) []vpa_api_util.ContainerResources {
 	resources := make([]vpa_api_util.ContainerResources, len(pod.Spec.Containers))
 	for i, container := range pod.Spec.Containers {
 		recommendation := vpa_api_util.GetRecommendationForContainer(container.Name, &podRecommendation)
 		if recommendation == nil {
-			klog.V(2).Infof("no matching recommendation found for container %s", container.Name)
-			continue
+			if !addAll {
+				klog.V(2).Infof("no matching recommendation found for container %s, skipping", container.Name)
+				continue
+			}
+			klog.V(2).Infof("no matching recommendation found for container %s, using Pod request", container.Name)
+			resources[i].Requests = container.Resources.Requests
+		} else {
+			resources[i].Requests = recommendation.Target
 		}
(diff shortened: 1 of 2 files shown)


vertical-pod-autoscaler/pkg/admission-controller/resource/pod/recommendation/recommendation_provider.go
```diff
@@ -46,23 +46,30 @@ func NewProvider(calculator limitrange.LimitRangeCalculator,
...
 		recommendation := vpa_api_util.GetRecommendationForContainer(container.Name, &podRecommendation)
 		if recommendation == nil {
-			klog.V(2).Infof("no matching recommendation found for container %s", container.Name)
-			continue
+			if addAll {
+				klog.V(2).Infof("no matching recommendation found for container %s, using Pod request", container.Name)
+				resources[i].Requests = container.Resources.Requests
+			} else {
+				klog.V(2).Infof("no matching recommendation found for container %s, skipping", container.Name)
+				continue
```
bskiba, line 61: We should either not continue at all or continue for both addAll=true and addAll=false (after the if addAll, else) and skip the else in line 63.

vertical-pod-autoscaler/pkg/admission-controller/resource/pod/recommendation/recommendation_provider.go
```diff
@@ -46,23 +46,30 @@ func NewProvider(calculator limitrange.LimitRangeCalculator,
...
 		}
-		resources[i].Requests = recommendation.Target
 		defaultLimit := core.ResourceList{}
 		if limitRange != nil {
 			defaultLimit = limitRange.Default
 		}
 		containerControlledValues := vpa_api_util.GetContainerControlledValues(container.Name, vpaResourcePolicy)
 		if containerControlledValues == vpa_types.ContainerControlledValuesRequestsAndLimits {
-			proportionalLimits, limitAnnotations := vpa_api_util.GetProportionalLimit(container.Resources.Limits, container.Resources.Requests, recommendation.Target, defaultLimit)
+			proportionalLimits, limitAnnotations := vpa_api_util.GetProportionalLimit(container.Resources.Limits, container.Resources.Requests, resources[i].Requests, defaultLimit)
```
bskiba, line 72: Why this change?

vertical-pod-autoscaler/pkg/admission-controller/resource/pod/recommendation/recommendation_provider.go
kgolab: The idea is that addAll controls whether we emit all containers, regardless of them having a recommendation or not.
If we always continue, there is no real change in behaviour.
Or did I misunderstand you comment?

vertical-pod-autoscaler/pkg/admission-controller/resource/pod/recommendation/recommendation_provider.go
kgolab: So we set request back into the resources - resources[i].Requests is either recommendation.Target (as previously) or container.Resources.Requests, if addAll allows.

bskiba: makes sense

vertical-pod-autoscaler/pkg/admission-controller/resource/pod/recommendation/recommendation_provider.go
bskiba: Ah sorry, I misread how big the else part is. 

Then we could do 
if !addAll {
    ...
    continue
}
so that you can drop the else

but it's not very important

bskiba (approved):

vertical-pod-autoscaler/pkg/admission-controller/resource/pod/recommendation/recommendation_provider.go
kgolab: Changed, I trust your golang instincts way more than mine.

PR #3549, opened 2020-09-28
Allow custom timeout value for webhook calls

renan:
Because webhooks add to API request latency, they should evaluate as
quickly as possible. timeoutSeconds allows configuring how long the
API server should wait for a webhook to respond before treating the call
as a failure.

The default values from Kubernetes are, however, too large. For
admissionregistration.k8s.io/v1 the default value is 10 seconds while
for admissionregistration.k8s.io/v1beta1 is 30 seconds!

This custom/default value of 5 seconds should be more than enough for
standard installs.

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/config.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/config.go
@@ -57,7 +57,7 @@ func configTLS(clientset *kubernetes.Clientset, serverCert, serverKey []byte) *t
 
 // register this webhook admission controller with the kube-apiserver
 // by creating MutatingWebhookConfiguration.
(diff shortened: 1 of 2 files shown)


bskiba:
Thanks, super useful. Can you set the default timeout to 30s? I know it's a ridiculously high value, but this is what the originally unconfigurable webhook timeout was set to and then the default behavior does not change unless users want to override it.

renan:
@bskiba Makes sense. I amended my commit!  :)

renan:
The failing test has been fixed 👍

bskiba:
Thanks!
/lgtm
/approve

renan:
@bskiba Awesome, thank you! Oh, and it would be totally awesome if you could release a v0.8.1 or something. My clusters could use this feature 😅

PR #3572, opened 2020-10-01
added a new NodeGroupDoesNotExistError in errors.go

jayantjain93:
This is to support no nodepool exists conditions, which crashes Autoscaler

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/mig_target_sizes_provider.go
+++ b/cluster-autoscaler/cloudprovider/gce/mig_target_sizes_provider.go
@@ -17,9 +17,9 @@ limitations under the License.
 package gce
 
 import (
-	"fmt"
-	klog "k8s.io/klog/v2"
 	"sync"
+
+	klog "k8s.io/klog/v2"
 )
 
 // MigTargetSizesProvider allows obtaining target sizes of MIGs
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/cloudprovider/gce/mig_target_sizes_provider.go
```diff
@@ -55,8 +55,10 @@ func (c *cachingMigTargetSizesProvider) GetMigTargetSize(migRef GceRef) (int64,
 	}
 
 	newTargetSizes, err := c.fillInMigTargetSizeCache()
-	if err != nil {
-		// fallback to querying for single mig
+
+	// if we still do not have value here return an error
```
MaciekPytel, line 59: that is not what this code does, the previous comment described the actual logic better (ie. you fallback to FetchMigTargetSize and only return an error if you fail it)

cluster-autoscaler/utils/test/test_utils.go
```diff
@@ -237,3 +237,25 @@ func (l *HttpServerMock) handle(url string) string {
...
+// NewHttpServerMockWithStatusCode creates new HttpServerMock.
+func NewHttpServerMockWithStatusCode() *HttpServerMock {
+	httpServerMock := &HttpServerMock{}
+	mux := http.NewServeMux()
+	mux.HandleFunc("/",
+		func(w http.ResponseWriter, req *http.Request) {
+			code, result := httpServerMock.handleWithStatusCode(req.URL.Path)
+			w.WriteHeader(code)
+			w.Write([]byte(result))
+			//w.Write([]byte("status: " + string(code) + "\n" + result))
```
MaciekPytel, line 250: delete this line please

jayantjain93: Done.

cluster-autoscaler/cloudprovider/gce/mig_target_sizes_provider.go
jayantjain93: Changed to the previous comment

PR #3573, opened 2020-10-01
simplify DeleteNode logic by removing an extra Mutex

ysy2020:
Previously, two mutexes are used in the logic of removing a node from a node pool. Now only one mutex is used to simplify the logic and the time for removing a node can be shortened with this logic since it doesn't always have to wait for a certain time length before being deleted.

### DIFF
--- a/cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_cloud_provider.go
@@ -185,7 +185,6 @@ func getAutoscaleNodePools(manager *huaweicloudCloudManager, opts config.Autosca
 	}
 
 	clusterUpdateLock := sync.Mutex{}
-	deleteMux := sync.Mutex{}
 
 	// Given our current implementation just support single node pool,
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_cloud_provider.go
```diff
@@ -185,7 +185,6 @@ func getAutoscaleNodePools(manager *huaweicloudCloudManager, opts config.Autosca
 	}
 
 	clusterUpdateLock := sync.Mutex{}
-	deleteMux := sync.Mutex{}
 
 	// Given our current implementation just support single node pool,
```
RainbowMango, line 189: No problem here, but need to point out that the `clusterUpdateLock` and `deleteMux` should be defined in the loop so that we can ensure the same mutex can't be shared with different node groups.

ysy2020: Yeah you're right. I'll move the clusterUpdateLock inside the for loop below. Thank you for the comment!

PR #3586, opened 2020-10-07
vpa/hack/vpa-process-yamls.sh add print action

icy:
The print action will print yaml contents for all resources
that would be understood by the next kubectl subcommand
(e.g., kubectl diff). This would be helpful for kustomization
user who wants to generate complete yaml content and
store them in their source tree.

Action delete, diff and print will include crd information.
The print action will not include secret information
that is supposed to be generated by gencerts.sh script


vertical-pod-autoscaler/README.md
```diff
@@ -112,6 +112,16 @@ in the `kube-system` namespace. It also generates
 and uploads a secret (a CA cert) used by VPA Admission Controller when communicating
 with the API server.
 
+To print YAML contents with all resources that would be understood by
+`kubectl diff|apply|...` commands, you can use
+
+```
+./hack/vpa-process-yamls.sh noop
```
mwielgus, line 119: `Noop` is not prarticularly intuitive. What about `print`?

mwielgus (changes requested):

vertical-pod-autoscaler/README.md
icy: Hi @mwielgus,

yes `print` is a lot better. I have updated the Pr (commits, subject and description with `push --force` command). 

Please let me know if there is something else I need to change or improve (I am not a native English speaker.)

Thanks a lot.

mwielgus (approved):
/lgtm
/approve

PR #3589, opened 2020-10-08
Cache controllers

jbartosik:
Cache results of attempting to determine the parent a controller.

This is to improve VPAs performance in big clusters, where we need to throttle queries we're making.

How caching is supposed to work:

* When we first try to determine parent of a controller try to get it (and cache the result).
  * This should happen rarely (when new controllers appear and when VPA starts). So it shouldn't affect VPAs performance heavily
  * It let's us avoid incorrect state
* On consequent attempts to get parent of a controller return the cached result
  * We do this frequently, so this should significantly improve our performance
* Refresh content of cache and remove unused entries in parallel

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
@@ -49,7 +49,13 @@ import (
 	resourceclient "k8s.io/metrics/pkg/client/clientset/versioned/typed/metrics/v1beta1"
 )
 
-const defaultResyncPeriod time.Duration = 10 * time.Minute
+const (
+	scaleCacheLoopPeriod         time.Duration = 7 * time.Second
+	scaleCacheEntryLifetime      time.Duration = time.Hour
+	scaleCacheEntryFreshnessTime time.Duration = 10 * time.Minute
+	scaleCacheEntryJitterFactor  float64       = 1.
+	defaultResyncPeriod          time.Duration = 10 * time.Minute
+)
 
 // ClusterStateFeeder can update state of ClusterState object.
 type ClusterStateFeeder interface {
@@ -108,7 +114,8 @@ func NewClusterStateFeeder(config *rest.Config, clusterState *model.ClusterState
 	kubeClient := kube_client.NewForConfigOrDie(config)
 	podLister, oomObserver := NewPodListerAndOOMObserver(kubeClient, namespace)
 	factory := informers.NewSharedInformerFactoryWithOptions(kubeClient, defaultResyncPeriod, informers.WithNamespace(namespace))
-	controllerFetcher := controllerfetcher.NewControllerFetcher(config, kubeClient, factory)
+	controllerFetcher := controllerfetcher.NewControllerFetcher(config, kubeClient, factory, scaleCacheEntryFreshnessTime, scaleCacheEntryLifetime, scaleCacheEntryJitterFactor)
+	controllerFetcher.Start(context.TODO(), scaleCacheLoopPeriod)
 	return ClusterStateFeederFactory{
 		PodLister:           podLister,
 		OOMObserver:         oomObserver,
--- a/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
@@ -34,6 +34,7 @@ type scaleCacheKey struct {
 	groupResource schema.GroupResource
 	name          string
 }
+
 type scaleCacheEntry struct {
 	refreshAfter time.Time
 	deleteAfter  time.Time
@@ -133,8 +134,10 @@ func (cc *controllerCacheStorage) RemoveExpired() {
 	cc.mux.Lock()
 	defer cc.mux.Unlock()
 	now := now()
+	removed := 0
 	for k, v := range cc.cache {
 		if now.After(v.deleteAfter) {
+			removed += 1
 			delete(cc.cache, k)
 		}
 	}
--- a/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
@@ -1,0 +1,165 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllerfetcher
+
+import (
+	"sync"
+	"time"
+
+	autoscalingapi "k8s.io/api/autoscaling/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog"
+)
+
+// Allows tests to inject their time.
+var now = time.Now
+
+type scaleCacheKey struct {
+	namespace     string
+	groupResource schema.GroupResource
+	name          string
+}
+type scaleCacheEntry struct {
+	refreshAfter time.Time
+	deleteAfter  time.Time
+	resource     *autoscalingapi.Scale
+	err          error
+}
+
+// Cache for responses to get queries on controllers. Thread safe.
+// Usage:
+// - `Get` cached response. If there is one use it, otherwise make query and
+// - `Insert` the response you got into the cache.
+// When you create a `controllerCacheStorage` you should start two go routines:
+// - One for refreshing cache entries, which calls `GetKeysToRefresh` then for
+//   each key makes query to the API server and calls `Refresh` to update
+//  content of the cache.
+// - Second for removing stale entries which periodically calls `RemoveExpired`
+// Each entry is refreshed after duration
+// `validityTime` * (1 + `jitterFactor`)
+// passes and is removed if there are no reads for it for more than `lifeTime`.
+//
+// Sometimes refreshing might take longer than refreshAfter (for example when
+// VPA is starting in a big cluster and tries to fetch all controllers). To
+// handle such situation lifeTime should be longer than refreshAfter so the main
+// VPA loop can do its work quickly, using the cached information (instead of
+// getting stuck on refreshing the cache).
+// TODO(jbartosik): Add a way to detect when we don't refresh cache frequently
+// enough.
+// TODO(jbartosik): Add a way to learn how long we keep entries around so we can
+// decide if / how we want to optimize entry refreshes.
+type controllerCacheStorage struct {
+	cache        map[scaleCacheKey]scaleCacheEntry
+	mux          sync.Mutex
+	validityTime time.Duration
+	jitterFactor float64
+	lifeTime     time.Duration
+}
+
+// Returns bool indicating whether the entry was present in the cache and the cached response.
+// Updates deleteAfter for the element.
+func (cc *controllerCacheStorage) Get(namespace string, groupResource schema.GroupResource, name string) (ok bool, controller *autoscalingapi.Scale, err error) {
+	key := scaleCacheKey{namespace: namespace, groupResource: groupResource, name: name}
+	cc.mux.Lock()
+	defer cc.mux.Unlock()
+	r, ok := cc.cache[key]
+	if ok {
+		r.deleteAfter = now().Add(cc.lifeTime)
+		cc.cache[key] = r
+	}
+	return ok, r.resource, r.err
+}
+
+// If key is in the cache, refresh updates the cached value, error and refresh
+// time (but not time to remove).
+// If the key is missing from the cache does nothing (relevant when we're
+// concurrently updating cache and removing stale entries from it, to avoid
+// adding back an entry which we just removed).
+func (cc *controllerCacheStorage) Refresh(namespace string, groupResource schema.GroupResource, name string, controller *autoscalingapi.Scale, err error) {
+	key := scaleCacheKey{namespace: namespace, groupResource: groupResource, name: name}
+	cc.mux.Lock()
+	defer cc.mux.Unlock()
+	old, ok := cc.cache[key]
+	if !ok {
+		return
+	}
+	// We refresh entries that are waiting to be removed. So when we refresh an
+	// entry we mustn't change entries deleteAfter time (otherwise we risk never
+	// removing entries that are not being read).
+	cc.cache[key] = scaleCacheEntry{
+		refreshAfter: now().Add(wait.Jitter(cc.validityTime, cc.jitterFactor)),
+		deleteAfter:  old.deleteAfter,
+		resource:     controller,
+		err:          err,
+	}
+}
+
+// If the key is missing from the cache, updates the cached value, error and refresh time (but not deleteAfter time).
+// If key is in the cache, does nothing (to make sure updating element doesn't change its deleteAfter time).
+func (cc *controllerCacheStorage) Insert(namespace string, groupResource schema.GroupResource, name string, controller *autoscalingapi.Scale, err error) {
+	key := scaleCacheKey{namespace: namespace, groupResource: groupResource, name: name}
+	cc.mux.Lock()
+	defer cc.mux.Unlock()
+	if _, ok := cc.cache[key]; ok {
+		return
+	}
+	now := now()
+	cc.cache[key] = scaleCacheEntry{
+		refreshAfter: now.Add(wait.Jitter(cc.validityTime, cc.jitterFactor)),
+		deleteAfter:  now.Add(cc.lifeTime),
+		resource:     controller,
+		err:          err,
+	}
+}
+
+// Removes entries which we didn't read in a while from the cache.
+func (cc *controllerCacheStorage) RemoveExpired() {
+	klog.V(1).Info("Removing entries from controllerCacheStorage")
+	cc.mux.Lock()
+	defer cc.mux.Unlock()
+	now := now()
+	for k, v := range cc.cache {
+		if now.After(v.deleteAfter) {
+			delete(cc.cache, k)
+		}
+	}
+	klog.V(1).Infof("Removed %d entries from controllerCacheStorage", removed)
+}
+
+// Returns a list of keys for which values need to be refreshed
+func (cc *controllerCacheStorage) GetKeysToRefresh() []scaleCacheKey {
+	result := make([]scaleCacheKey, 0)
+	cc.mux.Lock()
+	defer cc.mux.Unlock()
+	now := now()
+	for k, v := range cc.cache {
+		if now.After(v.refreshAfter) {
+			result = append(result, k)
+		}
+	}
+	return result
+}
+
+func newControllerCacheStorage(validityTime, lifeTime time.Duration, jitterFactor float64) controllerCacheStorage {
+	return controllerCacheStorage{
+		cache:        make(map[scaleCacheKey]scaleCacheEntry),
+		validityTime: validityTime,
+		jitterFactor: jitterFactor,
+		lifeTime:     lifeTime,
+	}
+}
--- a/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage_test.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage_test.go
@@ -1,0 +1,215 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllerfetcher
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	autoscalingapi "k8s.io/api/autoscaling/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func getKey(key string) scaleCacheKey {
+	return scaleCacheKey{
+		namespace: "ns",
+		groupResource: schema.GroupResource{
+			Group:    "group",
+			Resource: "resource",
+		},
+		name: key,
+	}
+}
+
+func getScale() *autoscalingapi.Scale {
+	return &autoscalingapi.Scale{}
+}
+
+func TestControllerCache_InitiallyNotPresent(t *testing.T) {
+	c := newControllerCacheStorage(time.Second, 10*time.Second, 1)
+	key := getKey("foo")
+	present, _, _ := c.Get(key.namespace, key.groupResource, key.name)
+	assert.False(t, present)
+}
+
+func TestControllerCache_Refresh_NotExisting(t *testing.T) {
+	key := getKey("foo")
+	c := newControllerCacheStorage(time.Second, 10*time.Second, 1)
+	present, _, _ := c.Get(key.namespace, key.groupResource, key.name)
+	assert.False(t, present)
+
+	// Refreshing key that isn't in the cache doesn't insert it
+	c.Refresh(key.namespace, key.groupResource, key.name, getScale(), nil)
+	present, _, _ = c.Get(key.namespace, key.groupResource, key.name)
+	assert.False(t, present)
+}
+
+func TestControllerCache_Insert(t *testing.T) {
+	key := getKey("foo")
+	c := newControllerCacheStorage(time.Second, 10*time.Second, 1)
+	present, _, _ := c.Get(key.namespace, key.groupResource, key.name)
+	assert.False(t, present)
+
+	c.Insert(key.namespace, key.groupResource, key.name, getScale(), nil)
+	present, val, err := c.Get(key.namespace, key.groupResource, key.name)
+	assert.True(t, present)
+	assert.Equal(t, getScale(), val)
+	assert.Nil(t, err)
+}
+
+func TestControllerCache_InsertAndRefresh(t *testing.T) {
+	key := getKey("foo")
+	c := newControllerCacheStorage(time.Second, 10*time.Second, 1)
+	present, _, _ := c.Get(key.namespace, key.groupResource, key.name)
+	assert.False(t, present)
+
+	c.Insert(key.namespace, key.groupResource, key.name, getScale(), nil)
+	present, val, err := c.Get(key.namespace, key.groupResource, key.name)
+	assert.True(t, present)
+	assert.Equal(t, getScale(), val)
+	assert.Nil(t, err)
+
+	c.Refresh(key.namespace, key.groupResource, key.name, nil, fmt.Errorf("err"))
+	present, val, err = c.Get(key.namespace, key.groupResource, key.name)
+	assert.True(t, present)
+	assert.Nil(t, val)
+	assert.Errorf(t, err, "err")
+}
+
+func TestControllerCache_InsertExistingKey(t *testing.T) {
+	key := getKey("foo")
+	c := newControllerCacheStorage(time.Second, 10*time.Second, 1)
+	present, _, _ := c.Get(key.namespace, key.groupResource, key.name)
+	assert.False(t, present)
+
+	c.Insert(key.namespace, key.groupResource, key.name, getScale(), nil)
+	present, val, err := c.Get(key.namespace, key.groupResource, key.name)
+	assert.True(t, present)
+	assert.Equal(t, getScale(), val)
+	assert.Nil(t, err)
+
+	// We might overwrite old values or keep them, either way should be fine.
+	c.Insert(key.namespace, key.groupResource, key.name, nil, fmt.Errorf("err"))
+	present, _, _ = c.Get(key.namespace, key.groupResource, key.name)
+	assert.True(t, present)
+}
+
+func TestControllerCache_GetRefreshesDeleteAfter(t *testing.T) {
+	oldNow := now
+	defer func() { now = oldNow }()
+	startTime := oldNow()
+	timeNow := startTime
+	now = func() time.Time {
+		return timeNow
+	}
+
+	key := getKey("foo")
+	c := newControllerCacheStorage(time.Second, 10*time.Second, 1)
+	c.Insert(key.namespace, key.groupResource, key.name, nil, nil)
+	assert.Equal(t, startTime.Add(10*time.Second), c.cache[key].deleteAfter)
+
+	timeNow = startTime.Add(5 * time.Second)
+	c.Get(key.namespace, key.groupResource, key.name)
+	assert.Equal(t, startTime.Add(15*time.Second), c.cache[key].deleteAfter)
+}
+
+func assertTimeBetween(t *testing.T, got, expectAfter, expectBefore time.Time) {
+	assert.True(t, got.After(expectAfter), "expected %v to be after %v", got, expectAfter)
+	assert.False(t, got.After(expectBefore), "expected %v to not be after %v", got, expectBefore)
+}
+
+func TestControllerCache_GetChangesLifeTimeNotFreshness(t *testing.T) {
+	oldNow := now
+	defer func() { now = oldNow }()
+	startTime := oldNow()
+	timeNow := startTime
+	now = func() time.Time {
+		return timeNow
+	}
+
+	key := getKey("foo")
+	c := newControllerCacheStorage(time.Second, 10*time.Second, 1)
+	c.Insert(key.namespace, key.groupResource, key.name, nil, nil)
+	cacheEntry := c.cache[key]
+	// scheduled to delete 10s after insert
+	assert.Equal(t, startTime.Add(10*time.Second), cacheEntry.deleteAfter)
+	// scheduled to refresh (1-2)s after insert (with jitter)
+	firstRefreshAfter := cacheEntry.refreshAfter
+	assertTimeBetween(t, firstRefreshAfter, startTime.Add(time.Second), startTime.Add(2*time.Second))
+
+	timeNow = startTime.Add(5 * time.Second)
+	c.Get(key.namespace, key.groupResource, key.name)
+	cacheEntry = c.cache[key]
+	// scheduled to delete 10s after get (15s after insert)
+	assert.Equal(t, startTime.Add(15*time.Second), cacheEntry.deleteAfter)
+	// refresh the same as before calling Get
+	assert.Equal(t, firstRefreshAfter, cacheEntry.refreshAfter)
+}
+
+func TestControllerCache_GetKeysToRefresh(t *testing.T) {
+	oldNow := now
+	defer func() { now = oldNow }()
+	startTime := oldNow()
+	timeNow := startTime
+	now = func() time.Time {
+		return timeNow
+	}
+
+	key1 := getKey("foo")
+	c := newControllerCacheStorage(time.Second, 10*time.Second, 1)
+	c.Insert(key1.namespace, key1.groupResource, key1.name, nil, nil)
+	cacheEntry := c.cache[key1]
+	// scheduled to refresh (1-2)s after insert (with jitter)
+	refreshAfter := cacheEntry.refreshAfter
+	assertTimeBetween(t, refreshAfter, startTime.Add(time.Second), startTime.Add(2*time.Second))
+
+	timeNow = startTime.Add(5 * time.Second)
+	key2 := getKey("bar")
+	c.Insert(key2.namespace, key2.groupResource, key2.name, nil, nil)
+	cacheEntry = c.cache[key2]
+	// scheduled to refresh (1-2)s after insert (with jitter)
+	refreshAfter = cacheEntry.refreshAfter
+	assertTimeBetween(t, refreshAfter, startTime.Add(6*time.Second), startTime.Add(7*time.Second))
+
+	assert.ElementsMatch(t, []scaleCacheKey{key1}, c.GetKeysToRefresh())
+}
+
+func TestControllerCache_Clear(t *testing.T) {
+	oldNow := now
+	defer func() { now = oldNow }()
+	startTime := oldNow()
+	timeNow := startTime
+	now = func() time.Time {
+		return timeNow
+	}
+
+	key1 := getKey("foo")
+	c := newControllerCacheStorage(time.Second, 10*time.Second, 1)
+	c.Insert(key1.namespace, key1.groupResource, key1.name, nil, nil)
+	assert.Equal(t, startTime.Add(10*time.Second), c.cache[key1].deleteAfter)
+
+	timeNow = startTime.Add(15 * time.Second)
+	key2 := getKey("bar")
+	c.Insert(key2.namespace, key2.groupResource, key2.name, nil, nil)
+	assert.Equal(t, startTime.Add(25*time.Second), c.cache[key2].deleteAfter)
+
+	c.RemoveExpired()
+	assert.Equal(t, 1, len(c.cache))
+	assert.Contains(t, c.cache, key2)
+}
--- a/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
+	autoscalingapi "k8s.io/api/autoscaling/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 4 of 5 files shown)


jbartosik:
/hold
I think it's ready for review but want to do some more tests before merging

mwielgus (changes requested):
The PR contains non trivial amount of code, but yet it doesn't have any description. Could you please provide some information what it does and why is it needed?

jbartosik:
> The PR contains non trivial amount of code, but yet it doesn't have any description. Could you please provide some information what it does and why is it needed?

I added a description

jbartosik:
/hold cancel

I ran `full-vpa` and `recommender` tests with this PR and they passed.

bskiba:
Looking at this now.

bskiba:
OK, I had a first look and looks good, thanks for splitting it into nicely separated commits, makes for a way easier review.
I'll need to take another look on Monday for a deeper review. So far I'd like to confirm some basic things.

1. When testing with 200 controllers, did you observe api call throttling?
2. Did you watch the memory consumption? Would be good to confirm we're not leaking, i.e. once we delete the controllers we should see a drop in memory after ttl passes.
3. With this solution, don't we still issue scale subresource calls when checking if the controller is scalable? https://github.com/kubernetes/autoscaler/blob/8711a1e37aa32eeda5ba7f3abb02295ab6c223db/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go#L262

jbartosik:
> OK, I had a first look and looks good, thanks for splitting it into nicely separated commits, makes for a way easier review.
> I'll need to take another look on Monday for a deeper review. So far I'd like to confirm some basic things.
> 
> 1. When testing with 200 controllers, did you observe api call throttling?

No, I don't see any logs about throttling.

> 2. Did you watch the memory consumption? Would be good to confirm we're not leaking, i.e. once we delete the controllers we should see a drop in memory after ttl passes.

I didn't see significant change in memory usage (which is not surprising, we're storing only a couple of controller keys per controller). I'll run another test where I create a bunch of controllers to see memory usage increase, then wait to see if it drops after time.

> 3. With this solution, don't we still issue scale subresource calls when checking if the controller is scalable? https://github.com/kubernetes/autoscaler/blob/8711a1e37aa32eeda5ba7f3abb02295ab6c223db/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go#L307

We do. I'll add caching for that too. Is it ok if I do this in a separate PR?

jbartosik:
> > 1. Did you watch the memory consumption? Would be good to confirm we're not leaking, i.e. once we delete the controllers we should see a drop in memory after ttl passes.
> 
> I didn't see significant change in memory usage (which is not surprising, we're storing only a couple of controller keys per controller). I'll run another test where I create a bunch of controllers to see memory usage increase, then wait to see if it drops after time.

I started a a few hundreds of pods. As a result memory usage increased from `23Mi` to t o`31Mi`. When I deleted pods memory usage decreased to `26Mi` and later (not sure after how much time) down to `25Mi`.  I'll repeat and see what happens

jbartosik:
@kgolab  PTAL

jbartosik:
> > 1. With this solution, don't we still issue scale subresource calls when checking if the controller is scalable? https://github.com/kubernetes/autoscaler/blob/8711a1e37aa32eeda5ba7f3abb02295ab6c223db/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go#L307
> 
> We do. I'll add caching for that too. Is it ok if I do this in a separate PR?

Done

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,149 @@
...
+import (
+	"math/rand"
+	"sync"
+	"time"
+
+	autoscalingapi "k8s.io/api/autoscaling/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var now = time.Now
```
kgolab, line 28: Could you please add a comment that this is done to allow controlling time from unit tests? Or is this obvious for anyone coding in golang?

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,149 @@
...
+	deleteAfter  time.Time
+	resource     *autoscalingapi.Scale
+	err          error
+}
+
+// Cache for responses to get queries on controllers. Thread safe.
+// Usage:
+// - `Get` cached response. If there is one use it, otherwise make query and
+// - `Insert` the response you got into the cache.
+// Start two go routines:
```
kgolab, line 47: Please make it clear who should start these go routines - it's not immediately clear that the caller has to do it (and I actually find it slightly strange that it's organised this way).

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// Returns a list of keys for which values need to be refreshed
+func (cc *controllerCacheStorage) GetKeysToRefresh() []scaleCacheKey {
+	result := make([]scaleCacheKey, 0)
+	cc.mux.Lock()
+	defer cc.mux.Unlock()
+	now := now()
+	for k, v := range cc.cache {
+		if now.After(v.refreshAfter) {
```
kgolab, line 135: I'd like to understand the idea behind refreshing & clearing stale entries a little better.

IIUC when an entry becomes idle, we'd refresh it few times before deleting which seems counterintuitive.
If this is true, shouldn't we throttle refreshes only to entries which were read recently? Then Get should probably return false to force a refresh upon reading an idle (& not refreshed) entry.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,149 @@
...
+	defer cc.mux.Unlock()
+	old, ok := cc.cache[key]
+	if !ok {
+		return
+	}
+	jitter := time.Duration(rand.Float64()*float64(cc.refreshJitter.Nanoseconds())) * time.Nanosecond
+	now := now()
+	cc.cache[key] = scaleCacheEntry{
+		refreshAfter: now.Add(cc.betweenRefreshes).Add(jitter),
+		deleteAfter:  old.deleteAfter,
```
kgolab, line 91: IIUC this is critical for an entry to get removed, could you please comment why it's here so nobody 'fixes' it?
I know the function comment says it's done on purpose but adding 'WHY' would help.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,149 @@
...
+// If the key is missing from the cache updates the cached value, error and refresh time (but not time to remove).
+// If key is in the cache it does nothing (to make sure updating element doesn't change its deleteAfter time).
+func (cc *controllerCacheStorage) Insert(namespace string, groupResource schema.GroupResource, name string, controller *autoscalingapi.Scale, err error) {
+	key := scaleCacheKey{namespace: namespace, groupResource: groupResource, name: name}
+	cc.mux.Lock()
+	defer cc.mux.Unlock()
+	if _, ok := cc.cache[key]; ok {
+		return
+	}
+	jitter := time.Duration(rand.Float64()*float64(cc.refreshJitter.Nanoseconds())) * time.Nanosecond
```
kgolab, line 106: How about extracting this to a function?

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage_test.go
```diff
@@ -0,0 +1,217 @@
...
+	present, _, _ := c.Get(key.namespace, key.groupResource, key.name)
+	assert.False(t, present)
+
+	c.Insert(key.namespace, key.groupResource, key.name, getScale(), nil)
+	present, val, err := c.Get(key.namespace, key.groupResource, key.name)
+	assert.True(t, present)
+	assert.Equal(t, getScale(), val)
+	assert.Nil(t, err)
+
+	// Doesn't overwrite previous values
```
kgolab, line 107: I'm not sure why we want this behaviour.
IIUC the result from second Insert is fresher so it should overwrite the old value (maybe just not deleteAfter).
If there is a valid reason, please add a comment about it.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -82,10 +83,41 @@ type controllerFetcher struct {
 	scaleNamespacer scale.ScalesGetter
 	mapper          apimeta.RESTMapper
 	informersMap    map[wellKnownController]cache.SharedIndexInformer
+	controllerCache controllerCacheStorage
+}
+
+func (f *controllerFetcher) periodicallyRemoveExpired(ctx context.Context, period time.Duration) {
```
kgolab, line 89: Here and below - since these functions are part of controllerFetcher and not controllerCache, could you please rename them so they include 'cache' or 'controllerCache' in their names?

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -49,7 +49,14 @@ import (
 	resourceclient "k8s.io/metrics/pkg/client/clientset/versioned/typed/metrics/v1beta1"
 )
 
-const defaultResyncPeriod time.Duration = 10 * time.Minute
+const (
+	betweenScaleSubresourceRefreshes time.Duration = 10 * time.Minute
```
kgolab, line 53: Maybe you could consider running this more frequently so the load (GET calls to API server) is spread more evenly in time? If GetKeysToRefresh is lightweight (and it seems so), it might be beneficial to run the whole refresher subroutine more often than main loop of the recommender.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -82,10 +83,41 @@ type controllerFetcher struct {
...
+	}
+}
+
+func (f *controllerFetcher) periodicallyRefresh(ctx context.Context, period time.Duration) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(period):
+			for _, item := range f.controllerCache.GetKeysToRefresh() {
```
kgolab, line 106: Could you please add some logs here marking start & finish of this procedure? Ideally with the number of entries refreshed after the loop.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -82,10 +83,41 @@ type controllerFetcher struct {
...
+	controllerCache controllerCacheStorage
+}
+
+func (f *controllerFetcher) periodicallyRemoveExpired(ctx context.Context, period time.Duration) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(period):
+			f.controllerCache.RemoveExpired()
```
kgolab, line 95: Maybe it would be useful here to log start & finish (or inside RemoveExpired, to get the number of entries removed).

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
kgolab: Actually this should belong below.

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -49,7 +49,14 @@ import (
...
+	containerCacheCleanupLoopPeriod  time.Duration = time.Hour
+	containerCacheRefreshLoopPeriod  time.Duration = time.Minute
```
kgolab, line 55: Seeing how these two are used I'd say they need renaming.

And seeing that I got some of these consts confused, maybe they could be commented to explain their meaning.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,149 @@
...
+// - One for refereshing cache entries, which calls `GetKeysToRefresh` then for
+//   each key makes query to the API server and calls `Refresh` to update
+//  content of the cache.
+// - Second from removing stale entries which periodically calls `RemoveExpired`
+// Each entry is refreshed after duration (`betweenRefreshes` + `jitter`) passes
+// and is removed if there are no reads for it for more than `lifeTime`.
+type controllerCacheStorage struct {
+	cache            map[scaleCacheKey]scaleCacheEntry
+	mux              sync.Mutex
+	betweenRefreshes time.Duration
```
kgolab, line 57: I'd say it's rather validityTime not to confuse it with period between executing refresh loop but please feel free to ignore.

kgolab:
> @kgolab PTAL

Many questions and some nits but I guess it's worth discussing how exactly the case is expected to behave with regards to refreshes and removal of stale entries.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
jbartosik: I chose the simplest approach I thought would work (periodically refresh all entries, remove entries that weren't read in a while).

> If this is true, shouldn't we throttle refreshes only to entries which were read recently? Then Get should probably return false to force a refresh upon reading an idle (& not refreshed) entry.

If `Get` returns false then the entry is effectively removed. So if I understand your idea correctly it's effectively the same as setting shorter lifetime for cache entries (less than 2 refresh durations?)

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
jbartosik: Done

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage_test.go
jbartosik: I don't expect this to come up in how we use the cache so I think this won't affect VPAs performance. And I find this easier to understand.

Please let me know if I should change the behavior

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
jbartosik: Done

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
jbartosik: Done.

jbartosik:
> > @kgolab PTAL
> 
> Many questions and some nits but I guess it's worth discussing how exactly the case is expected to behave with regards to refreshes and removal of stale entries.

I pushed changes with comments I applied. I had questions about a few and left them open. I'll add explanation why I want the cache to work this way and a TODO we discussed later today.

jbartosik:
> > > @kgolab PTAL
> > 
> > 
> > Many questions and some nits but I guess it's worth discussing how exactly the case is expected to behave with regards to refreshes and removal of stale entries.
> 
> I pushed changes with comments I applied. I had questions about a few and left them open. I'll add explanation why I want the cache to work this way and a TODO we discussed later today.

@kgolab I added the explanation (above `controllerCacheStorage`) and TODO.  Please take a look

jbartosik:
@kgolab @bskiba Please take a look

bskiba:
I've set aside some time tomorrow to take a look 👀

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,169 @@
...
+// handle such situation lifeTime should be longer than refreshAfter so the main
+// VPA loop can do its work quickly, using the cached information (instead of
+// getting stuck on refreshing the cache).
+// TODO(jbartosik): Add a way to detect when we don't refresh cache frequently
+// enough.
+type controllerCacheStorage struct {
+	cache         map[scaleCacheKey]scaleCacheEntry
+	mux           sync.Mutex
+	validityTime  time.Duration
+	refreshJitter time.Duration
```
bskiba, line 68: Might consider jitterFactor instead of jitter expressed as duration, similarly to Kuberentes wait functionality:
https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go#L132

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
bskiba: We can also use the Jitter function from here: https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go#L193:6

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	return ok, r.resource, r.err
+}
+
+func generateJitter(maxJitter time.Duration) time.Duration {
+	return time.Duration(rand.Float64()*float64(maxJitter.Nanoseconds())) * time.Nanosecond
+}
+
+// If key is in the cache it updates the cached value, error and refresh time (but not time to remove).
+// If the key is missing from the cache does nothing (relevant when we asynchronously updating cache and
```
bskiba, line 91: maybe: (relevant when we're concurrently updating cache and removing stale entries from it..)
Also, missing commas, see comment below.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,169 @@
...
+	cc.mux.Lock()
+	defer cc.mux.Unlock()
+	old, ok := cc.cache[key]
+	if !ok {
+		return
+	}
+	now := now()
+	// We refresh entries that are waiting to be removed. So when we refresh an
+	// entry we mustn't change entries deleteAfter time (otherwise we risk never
+	// removing entries that none is reading)
```
bskiba, line 104: "that noone/nobody is reading" or "that are not being read"

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,169 @@
...
+// removing stale entries from it, to avoid adding back an entry which we just removed).
+func (cc *controllerCacheStorage) Refresh(namespace string, groupResource schema.GroupResource, name string, controller *autoscalingapi.Scale, err error) {
+	key := scaleCacheKey{namespace: namespace, groupResource: groupResource, name: name}
+	cc.mux.Lock()
+	defer cc.mux.Unlock()
+	old, ok := cc.cache[key]
+	if !ok {
+		return
+	}
+	now := now()
```
bskiba, line 101: nit: we can simply use now() in line 106

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -49,7 +49,13 @@ import (
 	resourceclient "k8s.io/metrics/pkg/client/clientset/versioned/typed/metrics/v1beta1"
 )
 
-const defaultResyncPeriod time.Duration = 10 * time.Minute
+const (
+	scaleCacheLoopPeriod time.Duration = time.Hour
```
bskiba, line 53: I would like to differentiate between constants affecting whole cache (like the loop period) and cache entries (like refresh)

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -82,10 +83,44 @@ type controllerFetcher struct {
...
+		}
+	}
+}
+
+func (f *controllerFetcher) periodicallyRefreshCache(ctx context.Context, period time.Duration) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(period):
```
bskiba, line 105: Our refresh period for an entry is 10 minutes, but we check which entries should be refreshed every 1 hour. Is this intentional?

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -49,7 +49,13 @@ import (
...
+	scaleCacheCleanup    time.Duration = time.Hour
```
bskiba, line 54: Would be nice to decide on one of lifetime and cleanup (both are used to describe the cache entry ttl AFAIU)

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -82,10 +83,44 @@ type controllerFetcher struct {
...
+			for _, item := range keysToReferesh {
+				scale, err := f.scaleNamespacer.Scales(item.namespace).Get(context.TODO(), item.groupResource, item.name, metav1.GetOptions{})
+				f.controllerCache.Refresh(item.namespace, item.groupResource, item.name, scale, err)
+			}
+			klog.Infof("Finished refreshing %d entries in controllerFetchers controllerCache", len(keysToReferesh))
+		}
+	}
+}
+
+func (f *controllerFetcher) Start(ctx context.Context, removePeriod, refreshPeriod time.Duration) {
```
bskiba, line 117: Could we actually for simplicity have one cache refresh function that would delete items that should be deleted and refresh the ones that are stale but still valid?

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,163 @@
...
+
+// Cache for responses to get queries on controllers. Thread safe.
+// Usage:
+// - `Get` cached response. If there is one use it, otherwise make query and
+// - `Insert` the response you got into the cache.
+// When you create a `controllerCacheStorage` you should start two go routines:
+// - One for refereshing cache entries, which calls `GetKeysToRefresh` then for
+//   each key makes query to the API server and calls `Refresh` to update
+//  content of the cache.
+// - Second from removing stale entries which periodically calls `RemoveExpired`
```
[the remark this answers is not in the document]
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,163 @@
...
+// Each entry is refreshed after duration (`validityTime` + `jitter`) passes
+// and is removed if there are no reads for it for more than `lifeTime`.
+//
+// Sometimes refreshing might take longer than refreshAfter (for example when
+// VPA is starting in a big cluster and tries to fetch all controllers). To
+// handle such situation lifeTime should be longer than refreshAfter so the main
+// VPA loop can do its work quickly, using the cached information (instead of
+// getting stuck on refreshing the cache).
+// TODO(jbartosik): Add a way to detect when we don't refresh cache frequently
```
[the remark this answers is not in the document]
jbartosik: I was thinking about a metric reporting how much time elapsed between entries being ready for refresh and actually being refreshed. I think it would be better to implement it in a separate PR to make reviews more focused.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,163 @@
...
+	cache         map[scaleCacheKey]scaleCacheEntry
+	mux           sync.Mutex
+	validityTime  time.Duration
+	refreshJitter time.Duration
+	lifeTime      time.Duration
+}
+
+// Returns bool indicating whether the entry was present in the cache and the cached response.
+// Updates deleteAfter for the element.
+func (cc *controllerCacheStorage) Get(namespace string, groupResource schema.GroupResource, name string) (ok bool, controller *autoscalingapi.Scale, err error) {
```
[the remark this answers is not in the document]
jbartosik: I prefer this signature because `ok` is the most important of the return values (it tells you if it makes sense to use the other two)

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,169 @@
...
+	// removing entries that none is reading)
+	cc.cache[key] = scaleCacheEntry{
+		refreshAfter: now.Add(cc.validityTime).Add(generateJitter(cc.refreshJitter)),
+		deleteAfter:  old.deleteAfter,
+		resource:     controller,
+		err:          err,
+	}
+}
+
+// If the key is missing from the cache updates the cached value, error and refresh time (but not time to remove).
```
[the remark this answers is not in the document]
jbartosik: Done.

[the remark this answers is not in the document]
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,169 @@
...
+		refreshAfter: now.Add(cc.validityTime).Add(jitter),
+		deleteAfter:  now.Add(cc.lifeTime),
+		resource:     controller,
+		err:          err,
+	}
+}
+
+// Removes entries from the cache which we didn't read in a while.
+func (cc *controllerCacheStorage) RemoveExpired() {
+	klog.Infof("Removing entries from controllerCacheStorage")
```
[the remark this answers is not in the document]
jbartosik: Done. We'll log it ~1/minute.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,169 @@
...
+	now := now()
+	cc.cache[key] = scaleCacheEntry{
+		refreshAfter: now.Add(cc.validityTime).Add(jitter),
+		deleteAfter:  now.Add(cc.lifeTime),
+		resource:     controller,
+		err:          err,
+	}
+}
+
+// Removes entries from the cache which we didn't read in a while.
```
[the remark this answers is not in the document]
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,169 @@
...
+	defer cc.mux.Unlock()
+	now := now()
+	removed := 0
+	for k, v := range cc.cache {
+		if now.After(v.deleteAfter) {
+			removed += 1
+			delete(cc.cache, k)
+		}
+	}
+	klog.Infof("Removed %d entries from controllerCacheStorage", removed)
```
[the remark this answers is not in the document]
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -248,6 +284,15 @@ func (f *controllerFetcher) isWellKnown(key *ControllerKeyWithAPIVersion) bool {
 	return exists
 }
 
+func (f *controllerFetcher) getScaleForResource(namespace string, groupResource schema.GroupResource, name string) (controller *autoscalingapi.Scale, err error) {
+	if ok, scale, err := f.controllerCache.Get(namespace, groupResource, name); ok {
+		return scale, err
+	}
+	scale, err := f.scaleNamespacer.Scales(namespace).Get(context.TODO(), groupResource, name, metav1.GetOptions{})
```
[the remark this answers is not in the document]
jbartosik: I agree that it would be nicer to use if cache took care of fetching entries. But I think it would make code more complicated. Since this is the only place I expect to use the cache I think it's overall better to keep things simpler.

I named it cache *storage* to emphasise that it's doing less than one might hope.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
jbartosik: Changed to 1 minute

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
jbartosik: Done.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
jbartosik: Done.

jbartosik:
@bskiba @kgolab I pushed changes, please take a look

jbartosik:
@bskiba I can't reply directly to the comment about using jitter function from API machinery but I applied it

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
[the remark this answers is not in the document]
bskiba: ack

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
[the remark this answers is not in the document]
bskiba: ack

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,164 @@
...
+	defer cc.mux.Unlock()
+	r, ok := cc.cache[key]
+	if ok {
+		r.deleteAfter = now().Add(cc.lifeTime)
+		cc.cache[key] = r
+	}
+	return ok, r.resource, r.err
+}
+
+// If key is in the cache it updates the cached value, error and refresh time (but not time to remove).
```
bskiba, line 86: nit: If key is in the cache, refresh updates...

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,164 @@
...
+// If the key is missing from the cache does nothing (relevant when we're
```
bskiba, line 87: nit: If the key is missing from the cache, does nothing

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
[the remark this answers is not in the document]
bskiba: That makes sense, I'm not gonna be supper stubborn about it :)

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,164 @@
...
+// Cache for responses to get queries on controllers. Thread safe.
+// Usage:
+// - `Get` cached response. If there is one use it, otherwise make query and
+// - `Insert` the response you got into the cache.
+// When you create a `controllerCacheStorage` you should start two go routines:
+// - One for refreshing cache entries, which calls `GetKeysToRefresh` then for
+//   each key makes query to the API server and calls `Refresh` to update
+//  content of the cache.
+// - Second for removing stale entries which periodically calls `RemoveExpired`
+// Each entry is refreshed after duration (`validityTime` + `jitter`) passes
```
bskiba, line 54: > duration (`validityTime` + `jitter`)
This is no longer valid

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
```diff
@@ -0,0 +1,164 @@
...
+	key := scaleCacheKey{namespace: namespace, groupResource: groupResource, name: name}
+	cc.mux.Lock()
+	defer cc.mux.Unlock()
+	old, ok := cc.cache[key]
+	if !ok {
+		return
+	}
+	// We refresh entries that are waiting to be removed. So when we refresh an
+	// entry we mustn't change entries deleteAfter time (otherwise we risk never
+	// removing entries are not being read).
```
bskiba, line 100: nit: entries that/which are not being read

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -49,7 +49,13 @@ import (
 	resourceclient "k8s.io/metrics/pkg/client/clientset/versioned/typed/metrics/v1beta1"
 )
 
-const defaultResyncPeriod time.Duration = 10 * time.Minute
+const (
+	scaleCacheLoopPeriod         time.Duration = time.Minute
```
bskiba, line 53: If we're refreshing every minute, for the usecase where there is one pod per controller, won't we have the same scalability as we have now ?
i.e. currently, if you have 600 pods, one custom controller for each, in the current solution this eats up 600 qpm = 10 qps
with the caching configured like in this PR we still take up  600 qpm = 10 qps

bskiba (commented):
Some nits left.

Overall I think we are close to what we want, would like to make sure the cache constants we choose work in our favor.

@kgolab Would you be able to take another look today? I would like you to explicitly approve as well once you're happy with this PR

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
jbartosik: updated

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
jbartosik: Cache entry lives for `scaleCacheEntryFreshnessTime` + jitter (15 minutes on average). Having short `scaleCacheLoopPeriod` is how often we check which entries need a refresh (so it should be shorter than `scaleCacheEntryFreshnessTime` or we'll just refresh everything every loop).

jbartosik:
@bskiba  @kgolab PTAL ≤

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
bskiba: Ah of course you're right, sorry

bskiba (approved):
I'm happy, I'll LGTM once kgolab takes one more look (I think more eyes are better here as caching is always tricky :) )

bskiba:
And thanks a lot for the work that went into this!

jbartosik:
@kgolab please take a look

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
kgolab, line 53: If you really want this to be spread independently from the VPA main loop, maybe choose some other period, preferably relatively prime to main loop's 60s period?
IIUC it might be a relatively short one, e.g. 13s, as the refresh loop is pretty lightweight except for scaleNamespacer.Get calls which won't change in number but instead get spread more evenly.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage.go
kgolab: If it's about simplicity then why not just:
* Insert sets validity time (let's say it's current scaleCacheEntryFreshnessTime),
* Get checks validity time and returns false if the entry is expired; the caller has to call Insert then,
* there is no background refresh,
* there is only background garbage collection?

This reduces number of calls to scaleNamespacer.Get as we never call it for entires that are no longer present.
I understand that the drawback of this solution is that all Gets are executed in the main VPA loop instead of being spread evenly over time.

So basically the discussion boils down to a question: "how long do we actively use an entry, on average"; if it's much longer than scaleCacheEntryLifetime, the amortised cost of multiple refreshes of a no-longer-used entry is low and then the gain from spreading scaleNamespacer.Get calls is likely more important.

If you think we'd indeed have long-lived entries (seems plausible) let's leave the solution as it is now.
I'd love to see some metric (or logs with counter) added at some point so we can assess the impact of this change and maybe fine-tune scaleCacheEntryLifetime vs scaleCacheEntryFreshnessTime, likely lowering the former.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_cache_storage_test.go
kgolab: I don't have very strong feelings regarding the behaviour itself but if we simply don't care about whether it's one way of the other, maybe let's not have a test for this to basically keep it undefined?

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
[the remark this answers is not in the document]
kgolab: Actually it might be useful to rename the field as well, so it's clear at call site it's not a full cache but rather a storage.

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -82,10 +83,33 @@ type controllerFetcher struct {
...
+	controllerCache controllerCacheStorage
+}
+
+func (f *controllerFetcher) periodicallyRefreshCache(ctx context.Context, period time.Duration) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(period):
+			keysToReferesh := f.controllerCache.GetKeysToRefresh()
```
kgolab, line 95: nit: keysToRefresh

kgolab:
Thanks!

I added more comments but this already looks good to me.
Please note though that some of this comments are in older conversations so might be more tricky to find, I don't know how to make github show them all together.

And I'm really sorry for taking so much time to get back to this. Please ping me earlier next time.

kgolab:
@bskiba , I cannot execute the command so letting you know here: LGTM

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
jbartosik: Done. 7s because it's  the shortest that's in full seconds and relatively prime

vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
jbartosik: Done

jbartosik:
@kgolab replies to comments I can't find here but got email notifications for:

* we don't want to refresh entries in the main loop so it wont' get stuck so I'm keeping as is (but I'm making refresh loop execute more frequently)
* I added a TODO to add something that will let us optimize performance. I'll probably do it in two stages. First I'll add something to let us determine if we need to improve performance. If we think we need performance improvements I think it's best to use something more detailed than average lifetime (e.g. I think we might want to check for changes in young collections and error responses more quickly than in collections which were around for long time, unchanged)
* I changed the test for inserting over an old value to not care which value it keeps
* renamed `controllerCache` to `scaleSubresourceCacheStorage`

bskiba:
Thanks a lot Joachim
/lgtm
/approve

PR #3595, opened 2020-10-09
add volume mount label option to cluster-autoscaler makefile

elmiko:
This change adds an option to the `build-in-docker` and
`test-in-docker` make targets which allows usage on linux systems with
selinux enabled.  Currently these make targets use a volume mount to the
local directory for creating the binary output. On systems which have
selinux enforcing, the container runtime will disallow the mount to
occur. This change enables these make targets for selinux systems by
adding the `:Z` option to the volume flags.

On systems which do not have selinux enforcing this change will not have
any effect.


elmiko:
@MaciekPytel @mwielgus ptal

PR #3597, opened 2020-10-09
Merge pull request #3570 from towca/jtuznik/scale-down-after-delete-fix

ryaneorth:
Remove ScaleDownNodeDeleted status since we no longer delete nodes synchronously

PR #3600, opened 2020-10-09
Merge pull request #3570 from towca/jtuznik/scale-down-after-delete-fix

ryaneorth:
Remove ScaleDownNodeDeleted status since we no longer delete nodes synchronously


ryaneorth:
Cherry picking changes from #3570  into 1.16

mwielgus (approved):
/lgtm
/approve

MaciekPytel:
Note - I don't think this should have been merged before the issues on https://github.com/kubernetes/autoscaler/pull/3597 are resolved. Arguably we should only merge to 1.16 after we've merged to newer branches (to avoid a situation where we end up not merging to those branches and there is a regression in a later version).

Issue #3574, opened 2020-10-02
CRD: bad map key type: map keys must be strings


tamalsaha:
I tried to use kubebuilder controller-tools to generate CRD YAML for VPA but I am getting the error
`bad map key type: map keys must be strings`.

The problem is VPA uses `map[int]uint32` as seen here:
https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L355-L356

It seems that this will require an api data type change. 

xref: https://github.com/kubernetes/kubernetes/issues/95250

bskiba:
Thanks a lot for reporting.
Is there a Kubernetes version in which this will stop working in the current form?
If we want to change this, we need to do it carefully, possibly avoiding issues with reading the old objects after upgrade.

tamalsaha:
I think it will affect from k8s 1.16 and later versions if users want to use CRD openapi validation.

bskiba:
I'm not seeing any errors in 1.16, is CRD openapiv3 validation with structural schema an opt-in feature? If yes, how do you turn it on?

tamalsaha:
From https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema

`With apiextensions.k8s.io/v1 the definition of a structural schema is mandatory for CustomResourceDefinitions. In the beta version of CustomResourceDefinition, the structural schema was optional.`

bskiba:
I see, so does #3606 sort of mitigate the issue? Changing the histogram type at this point might cause data loss for some users if we're not very careful.

tamalsaha:
I replied in the pr .

Issue #3508, opened 2020-09-14
Upgrade VPA CRD version from apiextensions.k8s.io/v1beta1 to apiextensions.k8s.io/v1


Ladicle:
CRD apiextensions.k8s.io/v1beta1 has already been deprecated in v1.16. And now that 3 versions have passed, I'd like to migrate it to v1.

bskiba:
Sounds reasonable. 
1. Do you know if we need an upgrade story when we switch? i.e. what will happen with existing objects?
2. Are you interested in contributing? :)

Ladicle:
> 1. Do you know if we need an upgrade story when we switch? i.e. what will happen with existing objects?

CRD v1beta1 will be removed v1.22+, so we need to support v1 before that. 
For upgrades, applying the rewritten v1 manifests does not affect an existing object. (at this time `preseveUnknownFields` is true.) This is because of k8s support API conversion. Of course, it's better to keep the v1beta1 manifest for backwards compatibility.

However, `preseveUnknownFields` became false by default at GA, so we need to set `true` or define an `openAPIV3Scheme`. Considering the future, the latter is recommended. In that case, if someone misses specifying the field, the value will be prune. I think it's better to generate `openAPIV3Schema` from `types.go` using `controller-tools`.

> 2. Are you interested in contributing? :)

Yes! I'd like to contribute. :)

bskiba:
If you're interested, we welcome contributions :) Let me know if you need help.

PR #3606, opened 2020-10-12
Generate VPA CRD v1 from types.go

Ladicle:
ref: https://github.com/kubernetes/autoscaler/issues/3508
In this PR, I generate CRD v1 from types.go using `controller-gen@v0.4.0`.

Here are some things to note:
- Generates only the API version that was served in the previous manifest.
- The not recommended float64 is used as the CRD value, so `allowDangerousType` option is enabled. (ref: https://github.com/kubernetes-sigs/controller-tools/issues/245#issuecomment-518465214)
- `viacheckpoint` have int map key that is not supported in JSON, so disable validation only for that field. Along with this, generation script ignores error returned by controller-gen. (ref: https://github.com/kubernetes/autoscaler/issues/3574)
- `metav1.ObjectMeta` does not accept optional tags, so remove it from types.go.

### DIFF
--- a/vertical-pod-autoscaler/deploy/kustomization.yaml
+++ b/vertical-pod-autoscaler/deploy/kustomization.yaml
@@ -3,4 +3,4 @@ resources:
 - recommender-deployment.yaml
 - updater-deployment.yaml
 - vpa-rbac.yaml
-- vpa-v1-crd.yaml
+- vpa-v1-crd-gen.yaml
--- a/vertical-pod-autoscaler/hack/deploy-for-e2e.sh
+++ b/vertical-pod-autoscaler/hack/deploy-for-e2e.sh
@@ -73,7 +73,7 @@ for i in ${COMPONENTS}; do
   ALL_ARCHITECTURES=amd64 make --directory ${SCRIPT_ROOT}/pkg/${i} release
 done
 
-kubectl create -f ${SCRIPT_ROOT}/deploy/vpa-v1-crd.yaml
+kubectl create -f ${SCRIPT_ROOT}/deploy/vpa-v1-crd-gen.yaml
 kubectl create -f ${SCRIPT_ROOT}/deploy/vpa-rbac.yaml
 
 for i in ${COMPONENTS}; do
--- a/vertical-pod-autoscaler/hack/generate-crd-yaml.sh
+++ b/vertical-pod-autoscaler/hack/generate-crd-yaml.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPOSITORY_ROOT=$(realpath $(dirname ${BASH_SOURCE})/..)
+CRD_OPTS=crd:trivialVersions=false,allowDangerousTypes=true
+APIS_PATH=${REPOSITORY_ROOT}/pkg/apis
+OUTPUT=${REPOSITORY_ROOT}/deploy/vpa-v1-crd-gen.yaml
+WORKSPACE=$(mktemp -d)
+
+function cleanup() {
+    rm -r ${WORKSPACE}
+}
+trap cleanup EXIT
+
+if [[ -z $(which controller-gen) ]]; then
+    (
+        cd $WORKSPACE
+	      go mod init tmp
+	      go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.0
+    )
+    CONTROLLER_GEN=${GOBIN:-$(go env GOPATH)/bin}/controller-gen
+else
+    CONTROLLER_GEN=$(which controller-gen)
+fi
+
+# The following commands always returns an error because controller-gen does not accept keys other than strings.
+${CONTROLLER_GEN} ${CRD_OPTS} paths="${APIS_PATH}/..." output:crd:dir=${WORKSPACE} >& ${WORKSPACE}/errors.log ||:
+grep -v -e 'map keys must be strings, not int' -e 'not all generators ran successfully' -e 'usage' ${WORKSPACE}/errors.log \
+    && { echo "Failed to generate CRD YAMLs."; exit 1; }
+
+cd ${WORKSPACE}
+cat <<EOF > kustomization.yaml
+resources:
+- autoscaling.k8s.io_verticalpodautoscalers.yaml
+- autoscaling.k8s.io_verticalpodautoscalercheckpoints.yaml
+commonAnnotations:
+  "api-approved.kubernetes.io": "https://github.com/kubernetes/kubernetes/pull/63797"
+EOF
+kubectl kustomize . > ${OUTPUT}
--- a/vertical-pod-autoscaler/hack/vpa-process-yamls.sh
+++ b/vertical-pod-autoscaler/hack/vpa-process-yamls.sh
@@ -40,7 +40,7 @@ if [ $# -gt 2 ]; then
 fi
 
 ACTION=$1
-COMPONENTS="vpa-v1-crd vpa-rbac updater-deployment recommender-deployment admission-controller-deployment"
+COMPONENTS="vpa-v1-crd-gen vpa-rbac updater-deployment recommender-deployment admission-controller-deployment"
 case ${ACTION} in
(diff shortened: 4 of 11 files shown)


vertical-pod-autoscaler/deploy/kustomization.yaml
```diff
@@ -3,4 +3,4 @@ resources:
 - recommender-deployment.yaml
 - updater-deployment.yaml
 - vpa-rbac.yaml
-- vpa-v1-crd.yaml
+- vpa-v1-crd-v1.yaml
```
bskiba, line 6: maybe vpa-v1-crd-gen.yaml

bskiba (commented):
Thanks, this looks interesting.
My main question is, what will happen for an existing vpa deployment that used the old yaml? Is the generated yaml compatible, so that the existing vpa objects will continue working?

vertical-pod-autoscaler/hack/generate-crd-yaml.sh
```diff
@@ -0,0 +1,54 @@
...
+	      go mod init tmp
+	      go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.0
+    )
+    CONTROLLER_GEN=${GOBIN:-$(go env GOPATH)/bin}/controller-gen
+else
+    CONTROLLER_GEN=$(which controller-gen)
+fi
+
+# The following commands always returns an error because controller-gen does not accept keys other than strings.
+${CONTROLLER_GEN} ${CRD_OPTS} paths="${APIS_PATH}/..." output:crd:dir=${WORKSPACE} ||:
```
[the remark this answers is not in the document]
Ladicle: Currently, controller-gen does not support feature to check for specific error or ignore only the specific error, but we can find other errors in the output.  If the generator fails for a different error, this script will also output other than `map keys must be strings`.

```
$ hack/generate-crd-yaml.sh
/Users/ladicle/src/k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go:360:20: map keys must be strings, not int
/Users/ladicle/src/k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2/types.go:340:20: map keys must be strings, not int
Error: not all generators ran successfully
run `controller-gen crd:trivialVersions=false,allowDangerousTypes=true paths=/Users/ladicle/src/k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/... output:crd:dir=/var/folders/p_/hzm_gclx7vs5cdbfbzpl0s4w0000gn/T/tmp.Cytp8OtyCn -w` to see all available markers, or `controller-gen crd:trivialVersions=false,allowDangerousTypes=true paths=/Users/ladicle/src/k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/... output:crd:dir=/var/folders/p_/hzm_gclx7vs5cdbfbzpl0s4w0000gn/T/tmp.Cytp8OtyCn -h` for usage
```

Another way is to grep the controller-gen output and display a warning message.

Ladicle:
> what will happen for an existing vpa deployment that used the old yaml? Is the generated yaml compatible, so that the existing vpa objects will continue working?

Yes, there is no change in the yaml field, so it's compatible. If you applied to a cluster which has old vpa, `.spec.preserveUnknownFileds` will be true, so the behaviour also will not change. Also, I've confirmed that vpa objects continue to work. However, v1.16- cluster does not support CRD v1, so the new yaml can not be applied. 

When the first time to install vpa, `.spec.preserveUnknownFields` will be false,  so if you apply an invalid object to the cluster, you will get an error message.

like this:
```
$ kubectl apply -f examples/hamster.yaml
error: error validating "examples/hamster.yaml": error validating data: ValidationError(VerticalPodAutoscaler.spec.resourcePolicy.containerPolicies[0]): unknown field "controlledResources" in io.k8s.autoscaling.v1beta2.VerticalPodAutoscaler.spec.resourcePolicy.containerPolicies; if you choose to ignore these errors, turn validation off with --validate=false
```

vertical-pod-autoscaler/hack/vpa-process-yamls.sh
```diff
@@ -39,8 +39,17 @@ if [ $# -gt 2 ]; then
   exit 1
 fi
 
+SEMVER_RE='v[^0-9]*\([0-9]*\)[.]\([0-9]*\)[.]\([0-9]*\)\([0-9A-Za-z-]*\)'
+K8S_VERSION=$(kubectl version --short=true|grep Server|awk '{print $3}')
```
bskiba, line 43: I don't think we need to check the k8s version - the 0.9 VPA requires 1.16 anyway: https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler#compatibility

vertical-pod-autoscaler/hack/vpa-process-yamls.sh
```diff
@@ -39,8 +39,17 @@ if [ $# -gt 2 ]; then
...
 ACTION=$1
-COMPONENTS="vpa-v1-crd vpa-rbac updater-deployment recommender-deployment admission-controller-deployment"
+COMPONENTS="vpa-rbac updater-deployment recommender-deployment admission-controller-deployment"
+if [ $K8S_MINOR_VERSION -lt 16 ]; then
+    COMPONENTS="vpa-v1-crd ${COMPONENTS}"
+else
+    COMPONENTS="vpa-v1-crd-v1-gen ${COMPONENTS}"
+fi
 case ${ACTION} in
 delete|diff|print) COMPONENTS+=" vpa-beta2-crd" ;;
```
bskiba, line 54: What I meant is that we need to add vpa-v1-crd here to delete it on vpa-down.sh

vertical-pod-autoscaler/hack/generate-crd-yaml.sh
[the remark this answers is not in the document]
bskiba: If there is a not-overly complicated way to indicate that something other went wrong, I would really like to see it. Otherwise it might get really confusing for someone encountering issues.

bskiba (commented):
Thanks, left a couple more minor comments.

tamalsaha:
> viacheckpoint have int map key that is not supported in JSON, so disable validation only for that field. Along with this, generation script ignores error returned by controller-gen. (ref: #3574)

I don't think this is the right approach, since we are preserving a mistake. I recommend that @kubernetes/sig-api-machinery-api-reviews is consulted for this one. Thanks.

vertical-pod-autoscaler/hack/vpa-process-yamls.sh
Ladicle: Remove version-specific processing and specify only vpa-v1-crd-v1-gen; although the versions of the CRD are different, they define the same VPA resources, we cand delete CR using either crd-v1-crd or crd-v1-crd-gen files.

Ladicle:
I agree with @tamalsaha that we should finally modify the VPA APIs to use string as the map key. Disabling validation of the `BucketWeights` field is just a workaround. 
However, updating both the CRD API version and the VPA API version at once requires a lot of changes. IMHO, updating the CRD version and modifying VPA field will have less impact on the user if we work on them separately.

bskiba:
I agree with @Ladicle on not mixing too many changes at once. I think there is already a lot of benefit from moving to the controller-gen here and the issue with incorrectly defined field is a separate one, which needs to be approached with a lot of care to not break existing users.

vertical-pod-autoscaler/hack/vpa-process-yamls.sh
bskiba: Makes sence, thank you

vertical-pod-autoscaler/deploy/kustomization.yaml
```diff
@@ -3,4 +3,4 @@ resources:
 - recommender-deployment.yaml
 - updater-deployment.yaml
 - vpa-rbac.yaml
-- vpa-v1-crd.yaml
+- vpa-v1-crd-v1-gen.yaml
```
bskiba, line 6: vpa-v1-crd-gen

vertical-pod-autoscaler/hack/deploy-for-e2e.sh
```diff
@@ -73,7 +73,7 @@ for i in ${COMPONENTS}; do
   ALL_ARCHITECTURES=amd64 make --directory ${SCRIPT_ROOT}/pkg/${i} release
 done
 
-kubectl create -f ${SCRIPT_ROOT}/deploy/vpa-v1-crd.yaml
+kubectl create -f ${SCRIPT_ROOT}/deploy/vpa-v1-crd-v1-gen.yaml
```
bskiba, line 76: vpa-v1-crd-gen

bskiba (approved):
Thanks, this looks good.
Small naming change and we should be good to merge.

vertical-pod-autoscaler/deploy/kustomization.yaml
Ladicle: Recheck all file names and unified to vpa-v1-crd-gen. Also, recreated the CRD manifest.

bskiba:
Looks good, thanks a lot!

Can you squash the commits? Then it's ready to go :)

Ladicle:
Thank you for reviewing :) I've squashed all commits.

mwielgus (approved):
/lgtm
/approve

Issue #2892, opened 2020-03-04
Imbalance with balance similar node groups


JulienBalestra:
In the current AWS implementation, for example, the comparaison of node templates vs real world nodes produce imbalance scale ups when using the option `--balance-similar-node-groups`.

For example, when having 3 ASGs (1 per zone) with one ASG sets to 0, the autoscaler is always going  to scale the ASG with real world nodes instead of balancing the under provisioned ASG (the one at 0).

I suppose this is due to how we build node template and compare it to the real world ones.

To avoid this issue, we've been running these _experimental_ changes:
* https://github.com/DataDog/autoscaler/blob/datadog-master/cluster-autoscaler/core/utils.go#L248-L356
* https://github.com/DataDog/autoscaler/commit/9da4ff41cbab0bd25012fc70e5a9ce9995c6c3ef

These _experimental_ changes have been running properly for few months and gave us the expected results.
I'd like to open a conversation, why we initially used a mix comparaison of **real and templated nodes**.

This is also a potential issue when users change the instance type of the running ASG (https://github.com/kubernetes/autoscaler/pull/2840).

Let me know if you need more details I might have forgotten.

MaciekPytel:
At a glance, this basically always uses TemplateNodeInfo, right? This was discussed at length in https://github.com/kubernetes/autoscaler/pull/1021.

tl;dr TemplateNodeInfo is always wrong. Even in controlled and uniform env like GKE it's impossible to get it 100% right (and, believe me, I tried). If you ever have a pod that barely fits or barely doesn't fit on a node (within the margin of error of TemplateNodeInfo) and you use a patch that always uses TemplateNodeInfo the effects are devastating (including infinite scale-up and / or pods remaining pending forever). 

I think a possibly better solution to the original problem is to add logic, where if you have multiple ~identical NodeGroup you use an existing node in one of the groups as a template for all the groups (overriding only the zone label).

PR #3608, opened 2020-10-13
Fix BalanceSimilarNodeGroups when scaling from zero

bpineau:
Use real-world nodeInfo from similar groups when BalanceSimilarNodeGroups.

Where available, the CA evaluates node groups capacities and labels using real-world nodes, for optimal accuracy (accounting for kubelet and system reserved resources, for instance).

It fallbacks to synthetic nodeInfos inferred from ASG/MIG/VMSs templates when no real nodes are available yet (ie. upscaling from zero/empty node group).

That asymmetric evaluation can prevent `BalanceSimilarNodeGroups` from working reliably when upscaling from zero:
* The first upscaled nodeGroup will get a node (capacity initially evaluated from ASG/MIG/VMSS template, then from the first real node, runtime labels)
* Other empty nodeGroups will likely be considered dissimilar to the first one: we're comparing ASG templates with real-world nodeInfo
* In the case of `least-waste` expander for instance, CA might prefer the nodegroup from real-world node, accounting for reserved resources (best usage) over the others

This change set implements [Maciek Pytel suggestion](https://github.com/kubernetes/autoscaler/issues/2892#issuecomment-595769765) (from a previous attempt at fixing this issue): we try to use real-world nodes examples from ~similar node groups, before a last resort fallback to synthetic nodeInfos built from ASG templates.

We compare nodeInfos using the `FindSimilarNodeGroups` processor primitives in order to leverage per cloud provider comparators (eg. label exclusions).

We look for similar nodegroups using synthetic nodeInfo from ASG/MIG (where available): comparing their nodeInfo from real-world nodes to a synthetic nodeInfo from the empty nodegroup wouldn't work (that's the root cause of this issue). When ~similar nodegroups are found that way, we use their nodeInfos built from real-world nodes as models for empty nodegroups (but keeping original location related labels).

Tested with AWS ASGs, GCP MIGs and Azure VMSS.

### DIFF
--- a/cluster-autoscaler/core/scale_up_test.go
+++ b/cluster-autoscaler/core/scale_up_test.go
@@ -522,7 +522,8 @@ func runSimpleScaleUpTest(t *testing.T, config *scaleTestConfig) *scaleTestResul
 	}
 	context.ExpanderStrategy = expander
 
-	nodeInfos, _ := utils.GetNodeInfosForGroups(nodes, nil, provider, listers, []*appsv1.DaemonSet{}, context.PredicateChecker, nil)
+	processors := NewTestProcessors()
+	nodeInfos, _ := utils.GetNodeInfosForGroups(nodes, nil, nil, provider, listers, []*appsv1.DaemonSet{}, context.PredicateChecker, nil, processors, &context)
 	clusterState := clusterstate.NewClusterStateRegistry(provider, clusterstate.ClusterStateRegistryConfig{}, context.LogRecorder, newBackoff())
 	clusterState.UpdateNodes(nodes, nodeInfos, time.Now())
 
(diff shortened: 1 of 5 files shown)


MaciekPytel:
I think this should be implemented as a processor:
 * A custom implementation of NodeGroupSetProcessor (which may internally call to BalancingNodeGroupSetProcessor or reuse some of its code). 
 * Or, possibly, a NodeInfoProcessor (if you actually want to use template node from 'similar' group for binpacking and not just balancing).

Those processors were added specifically to allow to use custom logic to replace balancing (NodeGroupSetProcessor) and allow custom overrides to NodeInfo map produced by GetNodeInfosForGroups (NodeInfoProcessor). My reasons are:
* The current implementation adds even more complexity to what are already most complex pieces of code (and it spreads the logic across multiple parts of the code). Encapsulating all this logic in one place would improve readability.
  * The cache can be held in processor. There is no reason why a processor couldn't be stateful (and there is an existing example of a processor using internal cache: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/filter_out_schedulable.go#L34)
* Core logic should not assume any particular implementation of a processor. You can replace the whole balancing logic with something completely different by switching NodeGroupSetProcessor (ex. balance all nodegroups that have a certain label, or for a more concrete example - balance all MIGs in a GKE nodepool).

mwielgus:
Ping. Please respond to @MaciekPytel comment.

bpineau:
@mwielgus : Maciek comment makes perfect sense, I want to secure a bit of time to try implementing the suggested options (I suspect I'll come back with questions if you don't mind).

bpineau:
Implemented @MaciekPytel suggestion in PR #3761 ; closing in favour of the new PR.

PR #3609, opened 2020-10-13
Flag for exclusive usage of template infos (with warnings)

bpineau:
This is a second attempt at #1021 (might also provides an alternate solution for #2892 and #3608).

Per #1021 discussion, a flag might be acceptable, if defaulting to false and describing the limitations (not all cloud providers) and the risk of using it (loss of accuracy, risk of upscaling unusable nodes or leaving pending pods).

Some uses cases includes:
* Balance Similar when uspscaling from zero (but I believe #3608 is a better approach)
* Edited/updated ASGs/MIGs taints and labels
* Updated instance type

Per previous discussion, the later two cases could be covered in the long term by custom node-shape discovery Processors (open to discuss that option too).

### DIFF
--- a/cluster-autoscaler/core/utils/utils.go
+++ b/cluster-autoscaler/core/utils/utils.go
@@ -42,9 +42,30 @@ import (
 func GetNodeInfosForGroups(nodes []*apiv1.Node, nodeInfoCache map[string]*schedulerframework.NodeInfo, cloudProvider cloudprovider.CloudProvider, listers kube_util.ListerRegistry,
 	// TODO(mwielgus): This returns map keyed by url, while most code (including scheduler) uses node.Name for a key.
 	// TODO(mwielgus): Review error policy - sometimes we may continue with partial errors.
-	daemonsets []*appsv1.DaemonSet, predicateChecker simulator.PredicateChecker, ignoredTaints taints.TaintKeySet) (map[string]*schedulerframework.NodeInfo, errors.AutoscalerError) {
+	daemonsets []*appsv1.DaemonSet, predicateChecker simulator.PredicateChecker, ignoredTaints taints.TaintKeySet, useTemplatesOnly bool) (map[string]*schedulerframework.NodeInfo, errors.AutoscalerError) {
 	result := make(map[string]*schedulerframework.NodeInfo)
 	seenGroups := make(map[string]bool)
+
+	if useTemplatesOnly {
+		for _, nodeGroup := range cloudProvider.NodeGroups() {
+			id := nodeGroup.Id()
+			if _, found := result[id]; found {
+				continue
+			}
+			nodeInfo, err := GetNodeInfoFromTemplate(nodeGroup, daemonsets, predicateChecker, ignoredTaints)
+			if err != nil {
+				if err == cloudprovider.ErrNotImplemented {
+					klog.Warningf("Running in template only mode, but cloud provider template isn't implemented for group %s", id)
+					continue
+				} else {
+					klog.Errorf("Unable to build proper template node for %s: %v", id, err)
+					return map[string]*schedulerframework.NodeInfo{}, errors.ToAutoscalerError(errors.CloudProviderError, err)
+				}
+			}
+			result[id] = nodeInfo
+		}
+		return result, nil
+	}
 
 	podsForNodes, err := getPodsForNodes(listers)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


bpineau:
@losipiuk and @MaciekPytel as you took part to the previous attempt  discussion.

cluster-autoscaler/core/utils/utils.go
```diff
@@ -95,12 +95,12 @@ func GetNodeInfosForGroups(nodes []*apiv1.Node, nodeInfoCache map[string]*schedu
...
 		id := nodeGroup.Id()
 		seenGroups[id] = true
-		if _, found := result[id]; found {
+		if _, found := result[id]; found && !useTemplatesOnly {
 			continue
 		}
 
 		// No good template, check cache of previously running nodes.
-		if nodeInfoCache != nil {
+		if nodeInfoCache != nil && !useTemplatesOnly {
```
mwielgus, line 103: NodeInfoCache is populated but not used when useTemplatesOnly is provided.

cluster-autoscaler/core/utils/utils.go
```diff
@@ -42,7 +42,7 @@ import (
 func GetNodeInfosForGroups(nodes []*apiv1.Node, nodeInfoCache map[string]*schedulerframework.NodeInfo, cloudProvider cloudprovider.CloudProvider, listers kube_util.ListerRegistry,
 	// TODO(mwielgus): This returns map keyed by url, while most code (including scheduler) uses node.Name for a key.
 	// TODO(mwielgus): Review error policy - sometimes we may continue with partial errors.
-	daemonsets []*appsv1.DaemonSet, predicateChecker simulator.PredicateChecker, ignoredTaints taints.TaintKeySet) (map[string]*schedulerframework.NodeInfo, errors.AutoscalerError) {
+	daemonsets []*appsv1.DaemonSet, predicateChecker simulator.PredicateChecker, ignoredTaints taints.TaintKeySet, useTemplatesOnly bool) (map[string]*schedulerframework.NodeInfo, errors.AutoscalerError) {
 	result := make(map[string]*schedulerframework.NodeInfo)
```
mwielgus, line 46: There is lots of half done interactions with cache/result when useTemplatesOnly is used. Like populating cache and not using it. Using "last resort" nodes for if no teplate is provided. It would be cleaner to have a IF userTemplateOnly { } and the very beginning of the function, try generate stuff based on templates only and return.

mwielgus (changes requested):

cluster-autoscaler/core/utils/utils.go
bpineau: Right that suggestion makes it more clean and clear, while avoiding processing things we don't use (I tried to minimize the needed changes but that's not a great criterion indeed).

Updated to implement it as suggested. While doing so, had to make those two choices that could benefit from second opinion:
* Not using nodeInfoCache. We weren't using it for templates before either, but I suppose it was likely omitted to allow adoption of a real-world node as a model whenever possible (so might not be optimal here). On Azure and GCE GetNodeInfoFromTemplate->TemplateNodeInfo hits caches but didn't dig for other providers. Wdyt?
* Left cloud-providers the option to selectively return ErrNotImplemented depending on the group (otherwise error handling could be simpler)

cluster-autoscaler/core/utils/utils.go
```diff
@@ -95,12 +95,12 @@ func GetNodeInfosForGroups(nodes []*apiv1.Node, nodeInfoCache map[string]*schedu
 	for _, nodeGroup := range cloudProvider.NodeGroups() {
 		id := nodeGroup.Id()
 		seenGroups[id] = true
-		if _, found := result[id]; found {
+		if _, found := result[id]; found && !useTemplatesOnly {
```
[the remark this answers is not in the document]
bpineau: Removed that chunk while implementing the other suggested change

mwielgus:
Do we still need this PR? If yes, please resolve conflicts.

bpineau:
@mwielgus yes, rebased

MaciekPytel:
I wonder what's the relationship between this and #3761? It seems that you'd want to use one or the other? If so maybe we can consider packing this into another NodeInfo processor? I think the implementation would be pretty much a cut-paste of the change in this PR.
I know this one doesn't add as much complexity as #3761, but core/ is already one huge code spaghetti and I'm trying to limit changes to it as much as possible to avoid bloating it even further.

bpineau:
Yes, using #3671 should progressively replace this feature for us (once find alternatives to the habit of updating instance types in place, which came as a side effect of this initially "fix balancing + scale from zero" change, ie. as for #1021 original motivation, but as a second effect here). I can definitely look at making this a Nodeinfo processor.

mwielgus:
Ok then let's use #3671 and do it as NodeInfo processor/function. Closing this PR in wait for the new one, but feel free to reopen if you prefer to refactor this one.

bpineau:
Re-did it  it as NodeInfo processor/function, in PR #4000, would you mind having a look?

PR #3610, opened 2020-10-13
add Packet provider links to the root README.md

displague:
This PR adds links to the Packet provider README.md to the root README.md.  It is otherwise not obvious that Packet support is included.

The Packet provider will need to be renamed to Equinix Metal in the future. I'm using the existing naming in this PR.


mwielgus (approved):
/lgtm
/approve

PR #3618, opened 2020-10-15
Remove obsolete comments

Ladicle:
LastUpdateTime no longer exists and it is not used by this function.
The field was removed in API v1beta2 for scaling issue.

ref: https://github.com/kubernetes/autoscaler/pull/968


mwielgus (approved):
/lgtm
/approve

PR #3621, opened 2020-10-16
CA huaweicloud: implements node group by AS

RainbowMango:
This PR is part of refactoring work. Could you please take a look? @Dingshujie @ysy2020


ysy2020:
The changes make sense to me. No problem with the ECS and AS support parts. Just what to check if we can still keep the support for CCE node groups, like supporting CCE and ECS autoscaling groups at the same time?

ysy2020:
My thought is that whether we can provide two ways of building CloudProvider in BuildHuaweiCloud function. By default, we can support ECS and AS implementation, but we still keep the version for CCE, which we can change to support CCE easily after we clone the repo for our current testing. After our project is done, we can move to ECS and AS completely.

RainbowMango:
Technically, it's easy to support both CCE cluster and `self-build` cluster. There is a strong appeal about the `self-build` one.
This PR barely removed the origin node group, so I'd like to move forward and discuss it later.

PR #3622, opened 2020-10-16
Ensure CA simulator takes existing nodes into account

vivekbagade:
In findPlaceFor method, the nodes passed aren't taken into account.

### DIFF
--- a/cluster-autoscaler/simulator/cluster.go
+++ b/cluster-autoscaler/simulator/cluster.go
@@ -294,6 +294,10 @@ func findPlaceFor(removedNode string, pods []*apiv1.Pod, nodes map[string]bool,
 		return fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
 	}
 
+	isCandidateNode := func(nodeName string) bool {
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/simulator/cluster.go
```diff
@@ -294,6 +294,10 @@ func findPlaceFor(removedNode string, pods []*apiv1.Pod, nodes map[string]bool,
 		return fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
 	}
 
+	nodeExists := func(nodeName string) bool {
```
MaciekPytel, line 297: It's not really what the function is checking - the nodes you're filtering out do exist. Maybe rename it to isCandidateNode or something along those lines?

MaciekPytel:
Thanks for catching that!
Left super small comment, otherwise lgtm.

/approve

cluster-autoscaler/simulator/cluster.go
vivekbagade: Done

PR #3624, opened 2020-10-19
proactively set instance status to deleting on deletions

marwanad:
To help in cases where the VMSSVM cache TTL is set for too long and avoid unnecessary deletion calls

/area provider/azure


feiskyer (approved):
/lgtm

PR #3630, opened 2020-10-19
Allow for the leader election resourcelock to have a configurable name

marc-sensenich:
This is the start to allow for multiple deployments of the Cluster Autoscaler to be running in the same Namespace


mwielgus (approved):
/lgtm
/approve

PR #3635, opened 2020-10-20
Remove redundant log lines and reduce severity around parsing kubeEnv

towca:
The removed log lines had a very misleading ERROR severity, which
caused confusion. Furthermore, they were completely redundant when
no error happens - the same information can be deduced from the log
line at the end of extractAutoscalerVarFromKubeEnv.

Not finding a value in kubeEnv having a WARNING severity also seems
like too much.


mwielgus (approved):
/lgtm
/approve

PR #3636, opened 2020-10-21
Revert base price changes from PR #3603

MaciekPytel:
It turns out changing those values has unexpected impact on
the results of pricing expander. I think that a different base
value implicitly changes the weight of pricing to the
other parts of expander formula which is some cases changes CA
scaling decisions.
While not necessarilly wrong we shouldn't make changes like this
without analyzing implications and testing, hence the revert
of this part.


MaciekPytel:
cc: @towca @BigDarkClown

BigDarkClown (approved):

PR #3637, opened 2020-10-22
Configurable --network and --rm flags for docker

avorima:
The value of `DOCKER_NETWORK` is used for `--network` during `docker build`, default is `default`.
Setting `DOCKER_RM` to any non-empty value adds `--rm` for `docker run`.

### DIFF
--- a/cluster-autoscaler/Makefile
+++ b/cluster-autoscaler/Makefile
@@ -6,6 +6,7 @@ LDFLAGS?=-s
 ENVVAR=CGO_ENABLED=0 GO111MODULE=off
 GOOS?=linux
 REGISTRY?=staging-k8s.gcr.io
+DOCKER_NETWORK?=default
 ifdef BUILD_TAGS
   TAGS_FLAG=--tags ${BUILD_TAGS}
   PROVIDER=-${BUILD_TAGS}
@@ -19,6 +20,11 @@ ifdef LDFLAGS
   LDFLAGS_FLAG=--ldflags "${LDFLAGS}"
 else
   LDFLAGS_FLAG=
+endif
+ifdef DOCKER_RM
+  RM_FLAG=--rm
+else
+  RM_FLAG=
 endif
 
 build: clean deps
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/Makefile
```diff
@@ -58,10 +59,10 @@ format:
     test -z "$$(find . -path ./vendor -prune -type f -o -name '*.go' -exec gofmt -s -w {} + | tee /dev/stderr)"
 
 docker-builder:
-	docker build -t autoscaling-builder ../builder
+	docker build --network=${DOCKER_NETWORK} -t autoscaling-builder ../builder
 
 build-in-docker: clean docker-builder
-	docker run -v `pwd`:/gopath/src/k8s.io/autoscaler/cluster-autoscaler/:Z autoscaling-builder:latest bash -c 'cd /gopath/src/k8s.io/autoscaler/cluster-autoscaler && BUILD_TAGS=${BUILD_TAGS} LDFLAGS="${LDFLAGS}" make build-binary'
+	docker run --rm -v `pwd`:/gopath/src/k8s.io/autoscaler/cluster-autoscaler/:Z autoscaling-builder:latest bash -c 'cd /gopath/src/k8s.io/autoscaler/cluster-autoscaler && BUILD_TAGS=${BUILD_TAGS} LDFLAGS="${LDFLAGS}" make build-binary'
```
detiber, line 65: Not a blocker, but might be nice to also make this optional through a env variable, since it could help in troubleshooting issues to keep the container around.

avorima: Ok, I made it configurable. PTAL

cluster-autoscaler/Makefile
```diff
@@ -20,6 +21,9 @@ ifdef LDFLAGS
 else
   LDFLAGS_FLAG=
 endif
+ifdef DOCKER_CLEANUP
+  RM_FLAG=--rm
```
mwielgus, line 25: To be consistent with other flags, can you set `RM_FLAG` to empty in `else`? Moreover, can you rename it to `DOCKER_CLENUP_FLAG`. Or rename  `DOCKER_CLENUP` to `DOCKER_RM` so that we don't have both `CLEANUP`and `RM`.

mwielgus (changes requested):

avorima:
Should I squash my commits?

cluster-autoscaler/Makefile
avorima: I went with `DOCKER_RM` to be consistent with `DOCKER_NETWORK`

mwielgus:
Yes, please squash them.

feiskyer (commented):
/lgtm
/approve

PR #3642, opened 2020-10-23
Reduce logging in scale calculation

mcristina422:
I noticed this log message on a cluster with a large amount of nodes and many scale events. Especially in cases where a node isNodeBelowUtilizationThreshold() this reduces the number of lines logged. This log message is redundant since the same information is found in the isNodeBelowUtilizationThreshold check, so moving this log after it reduces the total number of messages 

Since this function is called for each node, reducing the number of logs in total is desired


feiskyer (commented):
/lgtm
/approve

PR #3646, opened 2020-10-27
Update Cluster Autoscaler (CA) volume mount path to /etc/ssl/certs/ca-bundle.crt

dmakeroam:
To support deploying CA on AWS EKS, we would need to update the volume mount path to '/etc/ssl/certs/ca-bundle.crt'
As the current volume mount path which is /etc/ssl/certs/ca-certificates.crt is not found on AWS EKS worker nodes with the error below:

Screenshot 2563-10-27 at 22 58 13

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
+++ b/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
@@ -157,7 +157,7 @@ spec:
             - --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,k8s.io/cluster-autoscaler/<YOUR CLUSTER NAME>
           volumeMounts:
             - name: ssl-certs
-              mountPath: /etc/ssl/certs/ca-bundle.crt
+              mountPath: /etc/ssl/certs/ca-certificates.crt #/etc/ssl/certs/ca-bundle.crt for EKS Amazon Linux Worker Nodes
               readOnly: true
           imagePullPolicy: "Always"
       volumes:
--- a/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
+++ b/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
@@ -157,7 +157,7 @@ spec:
             - --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,k8s.io/cluster-autoscaler/<YOUR CLUSTER NAME>
           volumeMounts:
             - name: ssl-certs
-              mountPath: /etc/ssl/certs/ca-certificates.crt
+              mountPath: /etc/ssl/certs/ca-bundle.crt
               readOnly: true
           imagePullPolicy: "Always"
       volumes:
--- a/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
+++ b/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
@@ -157,7 +157,7 @@ spec:
             - --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,k8s.io/cluster-autoscaler/<YOUR CLUSTER NAME>
           volumeMounts:
             - name: ssl-certs
-              mountPath: /etc/ssl/certs/ca-certificates.crt #/etc/ssl/certs/ca-bundle.crt for EKS Amazon Linux Worker Nodes
+              mountPath: /etc/ssl/certs/ca-certificates.crt #/etc/ssl/certs/ca-bundle.crt for Amazon Linux Worker Nodes
               readOnly: true
           imagePullPolicy: "Always"
       volumes:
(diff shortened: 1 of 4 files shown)


dmakeroam:
@jaypipes Could you help to review this PR, this would be good for AWS EKS users?

dmakeroam:
@jaypipes Any updates on this review?

jaypipes:
> @jaypipes Any updates on this review?

Apologies, I'm looking into this @dmakeroam.

cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
```diff
@@ -157,7 +157,7 @@ spec:
             - --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,k8s.io/cluster-autoscaler/<YOUR CLUSTER NAME>
           volumeMounts:
             - name: ssl-certs
-              mountPath: /etc/ssl/certs/ca-certificates.crt
+              mountPath: /etc/ssl/certs/ca-bundle.crt
```
jaypipes, line 160: I need to determine whether the previous default `ca-certificates.crt` was used for a specific reason. If that is, for example, the correct path on kops-installed clusters on EC2, then I think it would be better to add a new `cluster-autoscaler-autodiscover-eks.yaml` file instead of modifying this one.

dmakeroam: @jaypipes As you defined the deployment volume host path is "/etc/ssl/certs/ca-bundle.crt" as below:
```
      volumes:
        - name: ssl-certs
          hostPath:
            path: "/etc/ssl/certs/ca-bundle.crt"
```

So, to solve this issue, I would think that we are required to update the volume mount path to "/etc/ssl/certs/ca-bundle.crt" in

```
 volumeMounts:
            - name: ssl-certs
              from "mountPath: /etc/ssl/certs/ca-certificates.crt"
              to     "mountPath: /etc/ssl/certs/ca-bundle.crt"
```

 to match the volume you defined. In my opinion, it is not about EKS or KOPS right now.

mmerkes: From what I can see, it looks like the location of the SSL certs primarily has to do with the operating system. [Here's the kops guidance for setting up cluster autoscaler.](https://github.com/kubernetes/kops/blob/9885df83ad1dc0e07e921d9864f25c696cba5a77/addons/cluster-autoscaler/README.md#L17)

Perhaps we should just follow their example and just add a comment? As you can see [here](https://serverfault.com/questions/62496/ssl-certificate-location-on-unix-linux), looks like there's a lot of variation across Linux distros. Even in the EKS case, if you set up non-AmazonLinux worker nodes, the cert could be at a different location.

dmakeroam: @mmerkes So, could we add another comment in the aws example, so anyone would know if they use EKS with Amazon Linux, they should config like this.

dmakeroam:
@mmerkes I added another comment to the aws example already

cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
```diff
@@ -157,7 +157,7 @@ spec:
             - --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,k8s.io/cluster-autoscaler/<YOUR CLUSTER NAME>
           volumeMounts:
             - name: ssl-certs
-              mountPath: /etc/ssl/certs/ca-certificates.crt
+              mountPath: /etc/ssl/certs/ca-certificates.crt #/etc/ssl/certs/ca-bundle.crt for EKS Amazon Linux Worker Nodes
```
mmerkes, line 160: I believe this will be the case for any AmazonLinux worker nodes, so you can probably just remove the EKS part.

cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
mmerkes: @dmakeroam Makes sense, though I believe it would be any AmazonLinux worker nodes, not just EKS.

cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
dmakeroam: @mmerkes I updated as your suggestion already.

cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
dmakeroam: @mmerkes Agreed with this

mmerkes:
LGTM

dmakeroam:
Any updates on this review? @jaypipes

cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
foriequal0: Shouldn't this comment be moved to `volumes: hostPath:` part? IIUC, the difference comes from the host OS, not from the container.

PR #3651, opened 2020-10-29
Add e2e test framework

benmoss:
Builds off of #3566.

Adds adapted tests from the e2e suite. Right now there only a subset but ideally over time we can backfill additional coverage as we figure out how to do it in vendor agnostic ways.

cc @elmiko


elmiko:
thanks @benmoss , this is gonna take some for me to go through. just a heads up ;)

benmoss:
The main changes are in `test/framework`, the rest is just noise. We might want to consider dropping the `vendor/` directory as part of #3566 since it generates a bunch of commit noise

elmiko:
> We might want to consider dropping the `vendor/` directory as part of #3566 since it generates a bunch of commit noise

that makes sense to me.

i want to spend some time thinking about how we integrate this into the wider testing infra. i like that this is self-contained and we could just build by importing, i guess the next steps will be to figure out if we should promote subdirectories in the cloudproviders to contain their implementations.

elmiko:
hey Ben, just wanted to give a heads up. i am still working my way through this pr. currently trying to get everything working with the cluster-api-autoscaler-provider. it's going slow but i think most of it is making sense.

mwielgus:
There seems to be no progress on the PR. Closing. Feel free to reopen once needed.

elmiko:
@mwielgus i would like to continue with this work, but at this point i think it's best to just open a new pr when i am ready to move forward.

PR #3652, opened 2020-10-29
refactor CAPI controller unit test to use PollImmediate

elmiko:
This change removes the `PollImmediateInfinite` calls in the cluster-api
controller unit tests in favor of `PollImmediate`. It is being proposed
to prevent an edge case where the polling calls would become blocked
indefinitely. As we are using fake clients within the unit tests there
should be no delay getting a return value, but just in case there is a
miss on the poll function the new `PollImmediate` will timeout after 15
seconds.

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller_test.go
+++ b/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller_test.go
@@ -70,6 +70,7 @@ type testSpec struct {
 }
 
 const customCAPIGroup = "custom.x-k8s.io"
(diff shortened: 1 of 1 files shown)


elmiko:
@detiber ptal, @JoelSpeed and i were discussing the use of the `PollImmediateInfinite` in these tests and i think we can use `PollImmediate` with a short timeout just to avoid the potential corner case. i wasn't sure if you had another reason for using the infinite version here, appreciate any advice =)

i did run stress against the test suite with these changes in place and saw no issue. i will try running them in a memory/cpu constrained container just to make sure we don't regress on the low resource situation.

PR #3653, opened 2020-10-30
Include taints by condition when determining if a node is unready/still starting

towca:
Conditions and their corresponding taints can sometimes skew, which
can cause unnecessary scale-up. CA thinks nodes are ready because it
looks only at the conditions, but scheduler predicates fail because they
consider the taints as well. CA adds nodes, even though the existing
nodes are still starting. This commit brings CA behavior in line
with scheduler predicates behavior, eliminating the unnecessary
scale-up.

NOTE: NoSchedule taints don't have TimeAdded set, which is why I
implemented `isNodeStillStarting` this way.

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -860,22 +860,34 @@ func buildScaleDownStatusClusterwide(candidates map[string][]string, lastProbed 
 	return condition
 }
 
+func hasTaint(node *apiv1.Node, taintKey string) bool {
+	for _, taint := range node.Spec.Taints {
+		if taint.Key == taintKey {
+			return true
+		}
+	}
+	return false
+}
+
 func isNodeStillStarting(node *apiv1.Node) bool {
 	for _, condition := range node.Status.Conditions {
-		if condition.Type == apiv1.NodeReady &&
-			condition.Status != apiv1.ConditionTrue &&
-			condition.LastTransitionTime.Time.Sub(node.CreationTimestamp.Time) < MaxStatusSettingDelayAfterCreation {
-			return true
+		if condition.Type == apiv1.NodeReady {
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -860,21 +860,30 @@ func buildScaleDownStatusClusterwide(candidates map[string][]string, lastProbed
...
+	return false
+}
+
 func isNodeStillStarting(node *apiv1.Node) bool {
 	for _, condition := range node.Status.Conditions {
 		if condition.Type == apiv1.NodeReady &&
-			condition.Status != apiv1.ConditionTrue &&
-			condition.LastTransitionTime.Time.Sub(node.CreationTimestamp.Time) < MaxStatusSettingDelayAfterCreation {
+			condition.LastTransitionTime.Time.Sub(node.CreationTimestamp.Time) < MaxStatusSettingDelayAfterCreation &&
+			(condition.Status != apiv1.ConditionTrue || hasTaint(node, apiv1.TaintNodeNotReady)) {
```
MaciekPytel, line 876: nit: this is hard to read, maybe break it into sub-conditions? ex.

```
if condition.Type == apiv1.NodeReady {
  isNotReady := condition.Status != apiv1.ConditionTrue || hasTaint(node, apiv1.TaintNodeNotReady)
  if isNotReady && condition.LastTransitionTime.Time.Sub(node.CreationTimestamp.Time) < MaxStatusSettingDelayAfterCreation {
    return true
  }
}
```

MaciekPytel:
/lgtm
/approve
/hold
Feel free to cancel hold if you disagree with my comment.

cluster-autoscaler/clusterstate/clusterstate.go
towca: Done.

PR #3662, opened 2020-11-02
[CA-1.17] cherry-pick: Improve Cluster API tests to work better with constrained resources

elmiko:


MaciekPytel:
/lgtm
/approve
I think this was already fixed in 1.18 and 1.19, is that correct?

PR #3675, opened 2020-11-09
Pr fix scaledown issue

wangshiqi308:
Fixed after scaling down spefic numbers of node, the auto scaling group will automatically create corresponding amount of nodes issue.

### DIFF
--- a/cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_auto_scaling_group.go
+++ b/cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_auto_scaling_group.go
@@ -158,7 +158,17 @@ func (asg *AutoScalingGroup) Nodes() ([]cloudprovider.Instance, error) {
 // capacity and allocatable information as well as all pods that are started on
 // the node by default, using manifest (most likely only kube-proxy). Implementation optional.
 func (asg *AutoScalingGroup) TemplateNodeInfo() (*schedulerframework.NodeInfo, error) {
-	return nil, cloudprovider.ErrNotImplemented
+	template, err := asg.cloudServiceManager.getAsgTemplate(asg.groupID)
+	if err != nil {
+		return nil, err
+	}
+	node, err := asg.cloudServiceManager.buildNodeFromTemplate(asg.groupName, template)
+	if err != nil {
+		return nil, err
+	}
+	nodeInfo := schedulerframework.NewNodeInfo(cloudprovider.BuildKubeProxy(asg.groupName))
+	nodeInfo.SetNode(node)
+	return nodeInfo, nil
 }
 
 // Exist checks if the node group really exists on the cloud provider side. Allows to tell the
--- a/cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_service_manager.go
+++ b/cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_service_manager.go
@@ -18,8 +18,14 @@ package huaweicloud
 
 import (
 	"fmt"
+	"math/rand"
+	"strconv"
+	"strings"
 	"time"
 
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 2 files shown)


cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_service_manager.go
```diff
@@ -62,6 +62,9 @@ type AutoScalingService interface {
 	// IncreaseSizeInstance wait until instance number is updated.
 	IncreaseSizeInstance(groupID string, delta int) error
 
+	// Scale down cluster by updating auto scaling group instances by instanceIDs.
+	ScaleDownCluster(groupID string, instanceIds []string) error
```
RainbowMango, line 66: I'd prefer to take `DeleteScalingInstances(groupID string, instanceIDs []string)` as method name.

The `AutoScalingService` designed to mask implementation details about SDK, but it no `cluster` concept here.

RainbowMango: Your comments should start with the method name so that it will satisfy `golint` check.

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_service_manager.go
```diff
@@ -218,6 +221,29 @@ func (csm *cloudServiceManager) GetInstances(groupID string) ([]cloudprovider.In
...
+	opts := &huaweicloudsdkasmodel.UpdateScalingGroupInstanceRequest{
+		ScalingGroupId: groupID,
+		Body: &huaweicloudsdkasmodel.UpdateScalingGroupInstanceRequestBody{
+			InstancesId:    instanceIds,
+			InstanceDelete: &instanceDelete,
+			Action:         huaweicloudsdkasmodel.GetUpdateScalingGroupInstanceRequestBodyActionEnum().REMOVE,
+		},
+	}
+
+	_, err := asClient.UpdateScalingGroupInstance(opts)
```
RainbowMango, line 237: The reason you don't use [DeleteScalingInstance](https://github.com/huaweicloud/huaweicloud-sdk-go-v3/blob/8733c595be33e3f957c7532bfa0c091fc2731571/services/as/v1/as_client.go#L143) because it used for delete `one` instance in each call, right?

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_auto_scaling_group.go
```diff
@@ -94,26 +94,25 @@ func (asg *AutoScalingGroup) IncreaseSize(delta int) error {
...
+		}
 	}
 
-	err = asg.cloudServiceManager.DeleteServers(servers)
+	err = asg.cloudServiceManager.ScaleDownCluster(asg.groupID, instanceIds)
 	if err != nil {
-		klog.Warningf("failed to delete nodes. error: %v", err)
+		klog.Warningf("failed to scale down cluster. error: %v", err)
 		return err
 	}
```
RainbowMango, line 115: Don't we have to wait for the node group size update?

RainbowMango:
I guess you need to remove the merge commit `Merge pull request #1 from kubernetes/master ` from this PR.

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_service_manager.go
wangshiqi308: for Delete Scaling Instance it will only delete those instances but won't modify desire instance number of auto scaling group, so that why I'm using this API

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_auto_scaling_group.go
wangshiqi308: No, after calling the api, the node group size (I'm assuming the group size is the desire instance number, we can discuss this later) will change first, then it's going to remove those instances.

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_service_manager.go
RainbowMango: That's ok.

PR #3679, opened 2020-11-09
Return cluster-autoscaler-chart Chart name to cluster-autoscaler

### LINKED ISSUE #3475: cluster-autoscaler helm chart has '-chart' in its name
It seems bad form for a chart to include the word 'chart' in its name. It seems worse to then name every object the chart creates with 'chart' in the name. And it seems especially bad to RENAME the chart when moving it from one repo to another (i.e. the helm/stable chart was called 'cluster-autoscaler', not 'cluster-autoscaler-chart'), causing all existing named k8s objects to need to be destroyed and re-created, for no good reason.

Yes, I could set the 'nameOverride' and/or 'fullnameOverride' variables to work around this issue, but I really should not need to do so.

Please rename the chart to 'cluster-autoscaler'.

gjtempleton:
Fixes #3475 

Update helm/chart-releaser action to leverage new functionality of the chart releaser action allowing custom overrides of the published chart artifact separate from the naming of the chart

Tested successfully on my fork here: https://github.com/gjtempleton/autoscaler/releases/tag/cluster-autoscaler-chart-2.0.0

This will keep the names of the build artifacts consistent with the existing ones for the 1.X.X releases of the chart, but update the naming in the published chart to be neater and consistent with best practices.

I've also included a note in the readme to warn users.


llamahunter:
Yeah, that looks like it will do the trick.  Many things in _helpers.tpl are driven off .Chart.Name, and so long as that is set the same as it was before migration from the legacy to new repo location, all the resources should not go through a destroy/create cycle.

Thx very much!

charts/cluster-autoscaler/Chart.yaml
```diff
@@ -13,8 +13,8 @@ maintainers:
     name: gjtempleton
   - email: scott.crooks@gmail.com
     name: sc250024
-name: cluster-autoscaler-chart
+name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 1.1.1
+version: 2.0.0
```
llamahunter, line 20: Hmm.. unclear if this should be 9.0.0 or something, tho.  Since the last version was 8.0.0 before the chart took a detour through the other name.

stevehipwell:
@gjtempleton you could modify you checkout step to look like the following so you don't need a seperate fetch step.

```yaml
- name: Checkout
  uses: actions/checkout@v2
  with:
    fetch-depth: 0
```

Also the [example](https://github.com/helm/chart-releaser-action#example-workflow) has a additional install helm step that may be required?

charts/cluster-autoscaler/Chart.yaml
gjtempleton: Yeah, that's a good point and would probably make migration from the deprecated chart easier on users.

I'll update this to 9.0.0 and update the docs to make clear why this is happening as well.

stevehipwell:
> @gjtempleton you could modify you checkout step to look like the following so you don't need a seperate fetch step.
> 
> ```yaml
> - name: Checkout
>   uses: actions/checkout@v2
>   with:
>     fetch-depth: 0
> ```
> 
> Also the [example](https://github.com/helm/chart-releaser-action#example-workflow) has a additional install helm step that may be required?

I can confirm that the additional helm step is required.

llamahunter (approved):

stevehipwell:
Is there a roadmap for when this might be released?

gjtempleton:
@stevehipwell I can't commit to a timeline as it'll need approval from one of the other repo owners

/assign @MaciekPytel

llamahunter:
Thx for the update.  Our upgrades to k8s 1.18 are sort of on hold until this gets merged.

sc250024:
Thank you @gjtempleton !

llamahunter:
still waiting on this

stevehipwell:
The parallel https://github.com/kubernetes-sigs/descheduler/pull/436 was merged yesterday.

mwielgus (approved):
/lgtm
/approve

PR #3682, opened 2020-11-10
[CA]delete nodes from cluster after scale down

RainbowMango:
This PR :
- make sure all nodes belong to the specific node group before scale down.
- remove nodes from cluster after scale down.
- update example service account, CA requires node delete.

some logs from local testing:
```
I1110 05:03:59.033880       1 delete.go:103] Successfully added ToBeDeletedTaint on node k8s-node-components-1ney296p
I1110 05:03:59.158326       1 huaweicloud_auto_scaling_group.go:123] going to remove node from scaling group. group: e7c3f6ff-d6b3-4b22-b809-47a22c180276, node: eb81661f-ad84-4f8e-8a19-c31b514e3d12
I1110 05:03:59.372705       1 huaweicloud_auto_scaling_group.go:250] deleted one node from cluster. node name: k8s-node-components-1ney296p
```

### DIFF
--- a/cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_auto_scaling_group.go
+++ b/cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_auto_scaling_group.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
 	huaweicloudsdkasmodel "k8s.io/autoscaler/cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud-sdk-go-v3/services/as/v1/model"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_auto_scaling_group.go
```diff
@@ -202,3 +226,29 @@ func (asg *AutoScalingGroup) Autoprovisioned() bool {
...
+		klog.Warningf("Failed to delete nodes from cluster due to can not get kube-client. error: %v", err)
+		return err
+	}
+
+	// delete nodes one by one, return immediately when any error happens.
+	for _, nodeName := range nodeNames {
+		err := kubeClient.CoreV1().Nodes().Delete(context.TODO(), nodeName, metav1.DeleteOptions{})
+		if err != nil {
+			klog.Warningf("Failed to delete node from cluster. node: %s, error: %s", nodeName, err)
+			return err
```
Dingshujie, line 248: continue to delelte node when some error happen?  Instances is deleted, we need try our best to delete node record.

RainbowMango: That makes sense. Done.

MaciekPytel:
Note: Cluster Autoscaler is not really supposed to delete Node objects (and it doesn't do it for other providers). CA is responsible for removing underlying VMs, but it's the kubelet and NodeController that manage lifecycle of the Node object. 
In particular, assuming that the call to delete VM is not blocking, I don't think there is anything stopping NodeController and/or kubelet from adding the node again after you delete it, but before the VM is fully gone.

This comment is meant as a friendly advice, I don't want to interfere in your provider.

RainbowMango:
@MaciekPytel Thanks for your advice.

Yes, there is a `CloudNodeLifecycleController` runs by CCM which will delete nodes from the cluster after the underlying VM be deleted. But what if users haven't install CCM? CA will keep printing the warning log, [the code is here](https://github.com/kubernetes/autoscaler/blob/343dd75a048e963912a08a9edb0dc546257965c7/cluster-autoscaler/processors/nodes/pre_filtering_processor.go#L53).

Maybe we can add a parameter or provider scoped config field, give the user a chance to decide whether CA should delete node or not. How do you think?

MaciekPytel:
> But what if users haven't install CCM?

My point is that node may be deleted for multiple reasons not involving Cluster Autoscaler (manual scale-up, node upgrade, VM failure, ...). Having CA delete node object doesn't really solve the issue of lingering node objects, because CA can only remove node objects for VMs it deleted. So you need node lifecycle controller anyway. I think it's a vital part of kubernetes setup and should not be considered optional.

That being said I'm not at all familiar with details of running k8s on huawei cloud, so I may be easily missing something here.

RainbowMango:
> So you need node lifecycle controller anyway. I think it's a vital part of kubernetes setup and should not be considered optional.

Yeah, I agree. 

I guess there is no harm in letting CA delete node objects and it will make it easier for the user to use CA without CCM.
So, I would like to keep the current implementation for a while, hope we can get some feedback later.

Thanks @MaciekPytel for your comments and your time, I appreciate it.

PR #3685, opened 2020-11-10
added cache for instances

wangshiqi308:
This PR,

1. Added cache for all nodes to reduce API call.
2. Initialize node group from config file, instead of just get all node groups.

### DIFF
--- a/cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_auto_scaling_groups.go
+++ b/cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_auto_scaling_groups.go
@@ -1,0 +1,97 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package huaweicloud
+
+import (
+	"sync"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
+)
+
+type autoScalingGroupCache struct {
+	registeredAsgs           []*AutoScalingGroup
+	instanceToAsg            map[string]*AutoScalingGroup
+	cacheMutex               sync.Mutex
+	instancesNotInManagedAsg map[string]struct{}
+}
+
+func newAutoScalingGroupCache() *autoScalingGroupCache {
+	registry := &autoScalingGroupCache{
+		instanceToAsg:            make(map[string]*AutoScalingGroup),
+		instancesNotInManagedAsg: make(map[string]struct{}),
+	}
+
+	return registry
+}
+
+// Register asg in HuaweiCloud Manager.
+func (asg *autoScalingGroupCache) Register(autoScalingGroup *AutoScalingGroup) {
+	asg.cacheMutex.Lock()
+	defer asg.cacheMutex.Unlock()
+
+	asg.registeredAsgs = append(asg.registeredAsgs, autoScalingGroup)
+}
+
+// FindForInstance returns AsgConfig of the given Instance
+func (asg *autoScalingGroupCache) FindForInstance(instanceId string, csm *cloudServiceManager) (*AutoScalingGroup, error) {
+	asg.cacheMutex.Lock()
+	defer asg.cacheMutex.Unlock()
+	if config, found := asg.instanceToAsg[instanceId]; found {
+		return config, nil
+	}
+	if _, found := asg.instancesNotInManagedAsg[instanceId]; found {
+		return nil, nil
+	}
+	if err := asg.regenerateCache(csm); err != nil {
+		return nil, err
+	}
+	if config, found := asg.instanceToAsg[instanceId]; found {
+		return config, nil
+	}
+	// instance does not belong to any configured ASG
+	asg.instancesNotInManagedAsg[instanceId] = struct{}{}
+	return nil, nil
+}
+
+func (asg *autoScalingGroupCache) generateCache(csm *cloudServiceManager) {
+	go wait.Forever(func() {
+		asg.cacheMutex.Lock()
+		defer asg.cacheMutex.Unlock()
+		if err := asg.regenerateCache(csm); err != nil {
+			klog.Errorf("failed to do regenerating ASG cache,because of %s", err.Error())
+		}
+	}, time.Hour)
+}
+
+func (asg *autoScalingGroupCache) regenerateCache(csm *cloudServiceManager) error {
+	newCache := make(map[string]*AutoScalingGroup)
+
+	for _, registeredAsg := range asg.registeredAsgs {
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_auto_scaling_groups.go
```diff
@@ -0,0 +1,81 @@
+package huaweicloud
```
RainbowMango, line 1: Need to add a boilerplate header.

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_cloud_provider.go
```diff
@@ -65,17 +66,19 @@ func newCloudProvider(opts config.AutoscalingOptions, do cloudprovider.NodeGroup
...
 
 	csm := newCloudServiceManager(cloudConfig)
-	sgs, err := csm.ListScalingGroups()
+
+	hcp := &huaweicloudCloudProvider{
+		cloudServiceManager: csm,
+		resourceLimiter:     rl,
+		autoScalingGroup:    make([]AutoScalingGroup, 0),
+	}
+	err = hcp.buildAsgs(do.NodeGroupSpecs)
```
RainbowMango, line 75: We should check if specified the node groups, if not, return an error.

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_cloud_provider.go
```diff
@@ -65,17 +66,19 @@ func newCloudProvider(opts config.AutoscalingOptions, do cloudprovider.NodeGroup
...
 	if err != nil {
 		klog.Errorf("failed to list scaling groups. error: %v", err)
```
RainbowMango, line 77: Log content also needs to be updated.

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_cloud_provider.go
```diff
@@ -184,3 +201,24 @@ func BuildHuaweiCloud(opts config.AutoscalingOptions, do cloudprovider.NodeGroup
 
 	return newCloudProvider(opts, do, rl)
 }
+
+func buildAsgFromSpec(value string, asgs []AutoScalingGroup, manager CloudServiceManager) (*AutoScalingGroup, error) {
```
RainbowMango, line 205: s/value/spec/ or 
s/value/groupName/

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_cloud_provider.go
```diff
@@ -184,3 +201,24 @@ func BuildHuaweiCloud(opts config.AutoscalingOptions, do cloudprovider.NodeGroup
...
+	spec, err := dynamic.SpecFromString(value, true)
+
```
RainbowMango, line 207: Remove blank line.

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_cloud_provider.go
```diff
@@ -176,6 +161,38 @@ func (hcp *huaweicloudCloudProvider) Refresh() error {
...
+			klog.Warningf("failed to add node group to huaweicloud provider with spec: %s", spec)
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (hcp *huaweicloudCloudProvider) addNodeGroup(spec string) error {
+	asgs, err := hcp.cloudServiceManager.ListScalingGroups()
```
RainbowMango, line 176: One request for each spec seems a little bit expensive if the user specifies many specs. 
How do you think?

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_service_manager.go
```diff
@@ -86,6 +92,7 @@ type cloudServiceManager struct {
 	cloudConfig      *CloudConfig
 	getECSClientFunc func() *huaweicloudsdkecs.EcsClient
 	getASClientFunc  func() *huaweicloudsdkas.AsClient
+	asgs             *autoScalingGroups
```
RainbowMango, line 95: I'm not sure if the cloud service manager should hold a cache here.

A scaling group already holds a service manager, and the service manager also holds a scaling group cache, the cache holds all scaling groups. Here is a pointer loop, don't you think?

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_auto_scaling_groups.go
wangshiqi308: yep

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_cloud_provider.go
wangshiqi308: Make sense, I will move it to upper function.

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_service_manager.go
wangshiqi308: I don't really think it is a pointer loop. A scaling group doesn't hold a service manager, scaling groups and service manager are blong to provider.

Why I created another scaling group under manager mainly because scaling groups hold by provider were created by specs, which should be an attribute of provider, while scaling groups hold by service manager is actually used for caching instance, and those groups were reconstructed by us and have specific usages, so it should be hold by manager I think.

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_cloud_provider.go
wangshiqi308: yes, will do

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_service_manager.go
RainbowMango: OK.

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_auto_scaling_groups.go
```diff
@@ -0,0 +1,97 @@
...
+
+import (
+	"sync"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
+)
+
+type autoScalingGroups struct {
```
RainbowMango, line 27: Do you think `autoScalingGroupCache` would more meaningful?

cluster-autoscaler/cloudprovider/huaweicloud/huaweicloud_cloud_provider.go
```diff
@@ -184,3 +211,23 @@ func BuildHuaweiCloud(opts config.AutoscalingOptions, do cloudprovider.NodeGroup
...
+			return &AutoScalingGroup{
+				cloudServiceManager: manager,
+				groupName:           asg.groupName,
+				groupID:             asg.groupID,
+				minInstanceNumber:   asg.minInstanceNumber,
+				maxInstanceNumber:   asg.maxInstanceNumber,
+			}, nil
+		}
+	}
+	return nil, fmt.Errorf("No auto scaling group found, spec: %s", spec.Name)
```
RainbowMango, line 232: Error string should not be capitalized or end with punctuation.

More details please refer to [CodeReviewComments](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

PR #3690, opened 2020-11-13
Add unremovable_nodes_count metric

evgenii-petrov-arrival:
I want to set up alerting for cases where nodes are unremovable due to user errors, like lack of "safe to deschedule" annotation, and so this Pull Requests adds a Gauge sliced by `fmt.Sprintf("%s", simulator.UnremovableReason)`.

I wasn't sure if there is any concurrent access involved, so added a mutex for the counter map, please let me know if it is unnecessary.

I also wasn't sure how to test this, any suggestions are welcome.

### DIFF
--- a/cluster-autoscaler/core/scale_down.go
+++ b/cluster-autoscaler/core/scale_down.go
@@ -659,6 +659,16 @@ func (sd *ScaleDown) addUnremovableNode(unremovableNode *simulator.UnremovableNo
 	sd.unremovableNodeReasons[unremovableNode.Node.Name] = unremovableNode
 }
 
+func (sd *ScaleDown) getUnremovableNodesCount() map[simulator.UnremovableReason]int {
+	reasons := make(map[simulator.UnremovableReason]int)
+
+	for _, node := range sd.unremovableNodeReasons {
+		reasons[node.Reason]++
+	}
+
+	return reasons
+}
+
 // markSimulationError indicates a simulation error by clearing  relevant scale
 // down state and returning an appropriate error.
 func (sd *ScaleDown) markSimulationError(simulatorErr errors.AutoscalerError,
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -526,6 +526,7 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 			metrics.UpdateLastTime(metrics.ScaleDown, scaleDownStart)
 			scaleDownStatus, typedErr := scaleDown.TryToScaleDown(currentTime, pdbs)
 			metrics.UpdateDurationFromStart(metrics.ScaleDown, scaleDownStart)
+			metrics.UpdateUnremovableNodesCount(scaleDown.getUnremovableNodesCount())
 
 			scaleDownStatus.RemovedNodeGroups = removedNodeGroups
 
--- a/cluster-autoscaler/metrics/metrics.go
+++ b/cluster-autoscaler/metrics/metrics.go
@@ -17,6 +17,8 @@ limitations under the License.
 package metrics
 
 import (
+	"fmt"
+	"k8s.io/autoscaler/cluster-autoscaler/simulator"
 	"time"
 
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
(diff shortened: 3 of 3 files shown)


mwielgus (changes requested):
A related test is failing:
```
--- FAIL: TestFindUnneededNodes (0.02s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x3b7791f]
```
Please fix your code.

evgenii-petrov-arrival:
>Please fix your code.

Done.

Sorry for not running the gofmt/golint/tests before create a pull request, I haven't figured out how vendoring is structured in the repo and was depending on CI to catch any errors while I'm figuring that out.

evgenii-petrov-arrival:
@mwielgus , as far as I can tell, there is no way for me to mark your request for changes as addressed, so just in case I'll duplicate a request for re-review as a comment.

I think I've addressed your request for changes and the tests are passing now.

cluster-autoscaler/core/scale_down.go
```diff
@@ -364,21 +364,25 @@ type ScaleDown struct {
 	usageTracker           *simulator.UsageTracker
 	nodeDeletionTracker    *NodeDeletionTracker
 	unremovableNodeReasons map[string]*simulator.UnremovableNode
+
+	mx                        sync.Mutex
```
mwielgus, line 368: Mutex. All other fields are full name.

mwielgus (changes requested):

cluster-autoscaler/core/scale_down.go
evgenii-petrov-arrival: Fixed.

MaciekPytel:
It seems to me that this is duplicating the logic for gathering unready reasons that we already have for [ScaleDownStatusProcessor](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/processors/status/scale_down_status_processor.go). I think implementing this as a processor should be preferred to avoid duplication, keep the aggregation logic encapsulated and avoid bloating scale-down logic even more.

cluster-autoscaler/core/scale_down.go
```diff
@@ -608,6 +612,7 @@ func (sd *ScaleDown) UpdateUnneededNodes(
 	sd.nodeUtilizationMap = utilizationMap
 	sd.clusterStateRegistry.UpdateScaleDownCandidates(sd.unneededNodesList, timestamp)
 	metrics.UpdateUnneededNodesCount(len(sd.unneededNodesList))
+	metrics.UpdateUnremovableNodesCount(sd.unremovableReasonCounters)
```
mwielgus, line 615: I guess you may need to lock the mutex here.

evgenii-petrov-arrival:
>It seems to me that this is duplicating the logic for gathering unready reasons that we already have for ScaleDownStatusProcessor.

I've looked at ScaleDownStatusProcessor after your comment (wasn't aware of it before), and I don't think that adding a processor here would reduce complexity. Currently NoOpScaleDownStatusProcessor is the default ScaleDownStatusProcessor. And changing from NoOp to SomeOp seems to be a much larger change than adding a metric next to another similar metric.

That said, moving both `metrics.UpdateUnneededNodesCount` and `metrics.UpdateUnremovableNodesCount` to ScaleDownStatusProcessor seems like a good idea, but that refactoring is a much larger change than this one, and I think shouldn't block addition of this feature.

cluster-autoscaler/core/scale_down.go
evgenii-petrov-arrival: I think you are right, I'll try to get to this tomorrow.

evgenii-petrov-arrival: Done.

evgenii-petrov-arrival:
@mwielgus , is there something I can do to advance this pull request through the review?

towca:
Adding a processor only for updating one metric does seem like a bit of an overkill to me. This PR isn't actually duplicating any existing logic, since we don't pivot the `unremovableNodeReasons` map into per-reason counters anywhere. That said, I see two major problems with the proposed approach.

First of all, unremovable reasons are also set in `TryToScaleDown()`, which is called after `UpdateUnneededNodes()`. If you set the metric at the end of `UpdateUnneededNodes()`, you'll miss all of the reasons found in `TryToScaleDown()`. I think a good place to update this metric is just after `TryToScaleDown()` is called in `RunOnce()`.

Secondly, I think you're missing the fact that the `unremovableNodeReasons` map is cleared every loop, at the beginning of the scale-down logic in `RunOnce()` - in `scaleDown.CleanUp()`. So you don't need to keep track of how many reasons were added in the current loop in another structure. I'd recommend adding a `getUnremovableNodeReasonCounters() map[simulator.UnremovableReason]int` method to ScaleDown, which would go through all nodes in `unremovableNodeReasons` and count how many times each reason occurs. Then you can just pass its result to the function which updates the metric.

evgenii-petrov-arrival:
Thanks @towca , both very good points. I think I've addressed them know, please take a look.

towca:
Sorry for missing your reply, thanks for addressing my comments!

/lgtm

mwielgus:
Please squash commits to just 1-2 and we are good to merge.

evgenii-petrov-arrival:
Squashed to one commit, please take another look.

mwielgus (approved):
/lgtm
/approve

PR #3691, opened 2020-11-13
Add `extraEnvConfigMaps` and document some values

avorima:
There already is `envFromSecret` and `extraEnvSecrets`, so I made `envFromConfigMap` behave like `envFromSecret` and added `extraEnvConfigMaps` to replace it.

This pattern should make the intended use of the values clearer.

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -17,4 +17,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.0.0
+version: 9.1.0
--- a/charts/cluster-autoscaler/README.md
+++ b/charts/cluster-autoscaler/README.md
@@ -46,6 +46,10 @@ Releases of the chart from `9.0.0` onwards return the naming of the chart to `cl
 
 To migrate from a 1.X release of the chart to a `9.0.0` or later release, you should first uninstall your `1.X` install of the `cluster-autoscaler-chart` chart, before performing the installation of the new `cluster-autoscaler` chart.
 
+## Migration from 9.0 to 9.1
+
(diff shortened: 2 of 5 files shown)


avorima:
@gjtempleton I'm unsure what version increase to use, since this "only" is a breaking change for people who use the undocumented `envFromConfigMap`.

charts/cluster-autoscaler-chart/templates/deployment.yaml
```diff
@@ -156,10 +156,15 @@ spec:
                   name: {{ default (include "cluster-autoscaler.fullname" $) $value.name }}
                   key: {{ required "Must specify key!" $value.key }}
           {{- end }}
-          {{- if .Values.envFromSecret }}
+          {{- if or .Values.envFromSecret .Values.envFromConfigMap }}
           envFrom:
+          {{- if .Values.envFromSecret }}
             - secretRef:
                 name: {{ .Values.envFromSecret }}
+          {{- else if .Values.envFromConfigMap }}
```
gjtempleton, line 164: Do we want this to be an `else if` could a user not conceivably want to define both `envFromSecret` and `envFromConfigMap`?

gjtempleton (changes requested):
Hi, thanks for the PR!

Good point about the versioning and breaking of the existing functionality being undocumented.

I've already raised #3679 bumping the major version from 1.X to 9.X and I think this change is best included at the same time as also potentially breaking functionality. If you could add an entry to the Readme via the `.gotmpl` file to highlight the change happening and how users can maintain their existing behaviour I think that's the best approach. I've also got one quick nitpick.

charts/cluster-autoscaler-chart/templates/deployment.yaml
avorima: Allowing both is better, you're right. I'll change it.

avorima:
@gjtempleton I've added a short migration guide, but I still felt like this change should at least be a minor version increase. WDYT?

gjtempleton:
@avorima Agreed on the minor release, thanks for the rebase. 

Sorry, I should have got this merged last week before the 9.0.0 release.

PR #3692, opened 2020-11-13
Add extraVolumeSecrets chart value

avorima:
This value can be used to render `volumes` and `volumeMounts`.

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -17,4 +17,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.1.0
+version: 9.2.0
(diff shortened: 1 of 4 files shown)


charts/cluster-autoscaler-chart/values.yaml
```diff
@@ -150,6 +150,9 @@ extraArgs:
 # extraEnv -- Additional container environment variables.
 extraEnv: {}
 
+# extraVolumeSecrets -- Additional volumes to mount from Secrets.
```
gjtempleton, line 153: Can you include a full example of using this commented out given there's a number of sub-options here please.

gjtempleton:
Hey, thanks for the PR!

One request for you from a first look over it.

foriequal0:
I've found that my PR will conflict with this: https://github.com/kubernetes/autoscaler/pull/3725

avorima:
@foriequal0 My intent is that extra volumes and volumeMounts can be added for any cloud provider, so I'd prefer not to move it into an `else if`-block

charts/cluster-autoscaler/Chart.yaml
```diff
@@ -17,4 +17,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.1.0
+version: 9.1.1
```
gjtempleton, line 20: I'd prefer if we stuck to SemVer here and bumped the minor version as new functionality is being added

PR #3694, opened 2020-11-16
Fixes for update-vendor.sh on OSX and zsh

avorima:
The update-vendor.sh script needed some fixes to work with zsh on OSX.

Various binaries are expected to be GNU utils, which are not installed by default on OSX. I've added suggestions to install ported binaries using homebrew if they are not found in PATH.

Skip `shopt -s lastpipe` when `$BASH` is not defined.

### DIFF
--- a/cluster-autoscaler/hack/update-vendor.sh
+++ b/cluster-autoscaler/hack/update-vendor.sh
@@ -3,7 +3,36 @@
 set -o errexit
 set -o pipefail
 set -o nounset
-shopt -s lastpipe
+if [[ -n "${BASH}" ]]; then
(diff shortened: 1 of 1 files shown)


mwielgus (approved):
/lgtm
/approve

towca:
#3702 should fix the flaky test for good, but for this PR a retest should be enough.

/ok-to-test

avorima:
Thank you @towca 

/retest

towca:
I don't think you can restart Travis this way. Could you do a `git commit --amend --no-edit`? I think this'll trigger another build.

PR #3700, opened 2020-11-16
vpa: Replace component-base restclient metrics with built-in copy

adammw:
restclient metrics from controller-base were not showing up as vertical-pod-autoscaler uses prometheus metrics directly instead of via k8smetrics wrappers. therefore, the restclient hooks get attached to the legacyregistry but never used. 

this PR creates a copy of the restclient hooks which are compatible with the prometheus interface used throughout vpa to enable capturing rest_client_* metrics alongside the vpa_* metrics.


bskiba:
Thanks for the contribution. To make sure I understand the change, would it be possible to actually use k8smetrics wrappers correctly in vpa instead of copying the metric definitions into our repo?

adammw:
I think so, but it would require changes to every other metric in verticalpodautoscaler as it currently doesn't use them.

bskiba:
Sorry for the very long delay here.

Do you have an example how the metrics would have to be configured so that we don't have to keep our own copy of them? 

Copying here seems like a maintenance burden, so I'd like to avoid that if possible (even if it means extra work).

jbartosik:
How hard would it be to avoid duplicating code for registering metrics?

mwielgus:
Closing as inactive. Feel free to reopen if needed.

PR #3701, opened 2020-11-17
add daemonset eviction for non-empty nodes

yaroslava-serdiuk:
Modified DrainNode function: add daemonset eviction. The process of ds eviction is the same as for other pods, but in case of not proper eviction of ds pod the draining process isn't interrupted.


towca (changes requested):

cluster-autoscaler/core/scale_down_test.go
```diff
@@ -553,6 +554,7 @@ func TestDeleteNode(t *testing.T) {
 		{
 			name:               "successful attempt to delete empty node",
 			pods:               []string{},
+			daemonSetPods:      []string{"d1"},
```
towca, line 557: Why did you add it to this case specifically?

cluster-autoscaler/utils/drain/drain_test.go
```diff
@@ -588,5 +611,12 @@ func TestDrain(t *testing.T) {
 		if len(pods) != len(test.expectPods) {
 			t.Fatalf("Wrong pod list content: %v", test.description)
 		}
+
+		if len(daemonSetPods) != len(test.expectDaemonSetPods) {
```
towca, line 615: I think you can replace both these ifs with assert.ElementsMatch(t, test.expectDaemonSetPods, daemonSetPods).

cluster-autoscaler/core/scale_down_test.go
```diff
@@ -794,6 +806,47 @@ func TestDrainNodeWithRetries(t *testing.T) {
...
+		}
+		if eviction.Name == "d2" {
+			return true, nil, e2
+		}
+		return true, nil, nil
+	})
+	evictionResults, err := drainNode(n1, []*apiv1.Pod{p1, p2}, []*apiv1.Pod{d1, d2}, fakeClient, kube_util.CreateEventRecorder(fakeClient), 20, 0*time.Second, 0*time.Second, PodEvictionHeadroom)
+	assert.NoError(t, err)
+	assert.Equal(t, 2, len(evictionResults))
+	assert.Equal(t, *p1, *evictionResults["p1"].Pod)
```
towca, line 843: Pointers are automatically dereferenced in assert.Equal, so no need for the * operator here and below.

cluster-autoscaler/core/scale_down_test.go
```diff
@@ -794,6 +806,47 @@ func TestDrainNodeWithRetries(t *testing.T) {
...
+	assert.Equal(t, *p2, *evictionResults["p2"].Pod)
+	assert.NoError(t, evictionResults["p1"].Err)
```
towca, line 845: Why are you checking Err, TimedOut and WasEvictionSuccessful for 1 pod only? If the assertions hold true for both, please assert both because otherwise the reader is wondering what's different between the 2 pods.

cluster-autoscaler/core/scale_down_test.go
```diff
@@ -782,7 +794,7 @@ func TestDrainNodeWithRetries(t *testing.T) {
 			return true, nil, fmt.Errorf("too many concurrent evictions")
 		}
 	})
-	_, err := drainNode(n1, []*apiv1.Pod{p1, p2, p3}, fakeClient, kube_util.CreateEventRecorder(fakeClient), 20, 5*time.Second, 0*time.Second, PodEvictionHeadroom)
+	_, err := drainNode(n1, []*apiv1.Pod{p1, p2, p3}, []*apiv1.Pod{}, fakeClient, kube_util.CreateEventRecorder(fakeClient), 20, 5*time.Second, 0*time.Second, PodEvictionHeadroom)
```
towca, line 797: Could you add actual DS pods to this test as well?

towca (changes requested):

cluster-autoscaler/core/scale_down_test.go
```diff
@@ -643,7 +649,7 @@ func TestDeleteNode(t *testing.T) {
 			sd := NewScaleDown(&context, clusterStateRegistry)
 
 			// attempt delete
-			result := sd.deleteNode(n1, pods, provider.GetNodeGroup("ng1"))
+			result := sd.deleteNode(n1, pods, daemonSetPods, provider.GetNodeGroup("ng1"))
```
towca, line 652: If you're not adding any DS pods to any of the test cases, I'd just put an empty list here and remove the changes above.

towca (changes requested):

cluster-autoscaler/simulator/drain.go
```diff
@@ -29,18 +29,17 @@ import (
 	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
 )
 
-// FastGetPodsToMove returns a list of pods that should be moved elsewhere if the node
+// FastGetPodsToMove returns a list of pods that should be moved elsewhere and a list of DaemonSet pods that should be evicted if the node
```
vivekbagade, line 32: Nit: the line length in the func comments seems to extend a bit long. See if you can wrap it.

cluster-autoscaler/simulator/drain.go
```diff
@@ -51,27 +50,26 @@ func FastGetPodsToMove(nodeInfo *schedulerframework.NodeInfo, skipNodesWithSyste
...
-		return []*apiv1.Pod{}, pdbBlockingPod, err
+		return []*apiv1.Pod{}, []*apiv1.Pod{}, pdbBlockingPod, err
 	}
 
-	return pods, nil, nil
+	return pods, daemonSetPods, nil, nil
 }
 
-// DetailedGetPodsForMove returns a list of pods that should be moved elsewhere if the node
+// DetailedGetPodsForMove returns a list of pods that should be moved elsewhere and a list of DaemonSet pods that should be evicted if the node
```
vivekbagade, line 62: Nit: the line length in the func comments seems to extend a bit long. See if you can wrap it.

PR #3704, opened 2020-11-17
gcp: faster startup and refreshes with many MIGs

bpineau:
With 1.5k MIGs attached to a cluster, cluster-autoscaler needs about
40mn to start. Refreshing MIGs+ITs concurrently brings that down to
about 5mn.

While bulk GCE API calls (triggered at startup and on Refresh() calls)
and a few stateless functions (called by GetMigInstanceTemplate) become
concurrent, cache accesses remains lock protected. To that effect:
* Set RegenerateInstancesCache to run parallels RegenerateInstanceCacheForMig
  (slightly adapted so the slow GceService.FetchMigInstances call isn't locked)
* Set fetchAutoMigs to run parallels registerMig: rework GetMigInstanceTemplate
  so the slow InstanceGroupManagers.Get+InstanceTemplates.Get calls aren't locked

Tested on a large k8s cluster (> 1k MIGs) with intense scaling activity,
and tested on live clusters with "go build -race" cluster-autoscaler builds.

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/cache.go
+++ b/cluster-autoscaler/cloudprovider/gce/cache.go
@@ -17,12 +17,14 @@ limitations under the License.
 package gce
 
 import (
+	"context"
 	"fmt"
 	"reflect"
 	"strings"
 	"sync"
 
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
+	"k8s.io/client-go/util/workqueue"
 
 	gce "google.golang.org/api/compute/v1"
 	klog "k8s.io/klog/v2"
@@ -71,13 +73,14 @@ type GceCache struct {
 	migTargetSizeCache       map[GceRef]int64
 	migBaseNameCache         map[GceRef]string
 	instanceTemplatesCache   map[GceRef]*gce.InstanceTemplate
+	concurrentGceRefreshes   int
 
 	// Service used to refresh cache.
 	GceService AutoscalingGceClient
 }
 
 // NewGceCache creates empty GceCache.
-func NewGceCache(gceService AutoscalingGceClient) *GceCache {
+func NewGceCache(gceService AutoscalingGceClient, concurrentGceRefreshes int) *GceCache {
 	return &GceCache{
 		migs:                     map[GceRef]Mig{},
 		instanceRefToMigRef:      map[GceRef]GceRef{},
@@ -87,6 +90,7 @@ func NewGceCache(gceService AutoscalingGceClient) *GceCache {
 		migBaseNameCache:         map[GceRef]string{},
 		instanceTemplatesCache:   map[GceRef]*gce.InstanceTemplate{},
 		GceService:               gceService,
+		concurrentGceRefreshes:   concurrentGceRefreshes,
 	}
 }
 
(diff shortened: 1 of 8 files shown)


MaciekPytel:
Sorry I missed it earlier. I'd like to run some tests to understand the impact of this change on the rate of API call quota. In particular whether it will hit default quota in the case of 100 or so MIGs.
I won't be able to run the test today, but I'll try to do it early next week (either myself or get one of the other GCP devs to do it).

vivekbagade:
Hey @bpineau Thanks for submitting the fix. I've picked up the review and will do so this week. Sorry for the delay overall.

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -222,9 +224,28 @@ func (gc *GceCache) getMigNoLock(migRef GceRef) (mig Mig, found bool) {
...
+		klog.V(4).Infof("Failed MIG info request for %s: %v", migRef.String(), err)
+		return err
+	}
+
 	gc.cacheMutex.Lock()
 	defer gc.cacheMutex.Unlock()
-	return gc.regenerateInstanceCacheForMigNoLock(migRef)
+
+	// cleanup old entries
+	gc.removeInstancesForMigs(migRef)
```
vivekbagade, line 239: this code(239-248) is shared with regenerateInstanceCacheForMigNoLock. Consider moving it to one func used that would be used in both funcs.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -51,6 +54,7 @@ const (
 	migAutoDiscovererKeyPrefix   = "namePrefix"
 	migAutoDiscovererKeyMinNodes = "min"
 	migAutoDiscovererKeyMaxNodes = "max"
+	concurrentGceRefreshes       = 10
```
vivekbagade, line 57: This really needs to a configurable via an flag with a default of 1. This can affect project quota in major ways.

cluster-autoscaler/cloudprovider/gce/mig_instance_templates_provider.go
```diff
@@ -52,14 +52,13 @@ func NewCachingMigInstanceTemplatesProvider(cache *GceCache, gceClient Autoscali
...
 	p.mutex.Lock()
-	defer p.mutex.Unlock()
-
 	if !p.lastRefresh.Add(migInstanceCacheRefreshInterval).After(time.Now()) {
 		p.cache.InvalidateAllMigInstanceTemplates()
 		p.lastRefresh = time.Now()
 	}
 
 	instanceTemplate, found := p.cache.GetMigInstanceTemplate(migRef)
+	p.mutex.Unlock()
```
vivekbagade, line 61: releasing and acquire locks multiple times in a function can cause bugs where a refactor can break things. Consider making moving each snippet into new funcs where each func begins with lock and defer unlock

vivekbagade (commented):
Thanks for the CL. Left some comments.

bpineau:
@vivekbagade thanks a lot for the review! Just pushed an update addressing your comments.

PR #3718, opened 2020-11-24
Update generic labels for GCE

yaroslava-serdiuk:


cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -215,6 +215,8 @@ func BuildGenericLabels(ref GceRef, machineType string, nodeName string, os Oper
 	}
 	result[apiv1.LabelZoneRegion] = ref.Zone[:ix]
```
towca, line 216: Could you do the same for apiv1.LabelInstanceTypeStable above?

cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -215,6 +215,8 @@ func BuildGenericLabels(ref GceRef, machineType string, nodeName string, os Oper
...
 	result[apiv1.LabelZoneFailureDomain] = ref.Zone
+	result[apiv1.LabelZoneRegionStable] = ref.Zone[:ix]
```
towca, line 218: nit: I'd prefer to have the stable version right after the non-stable one (see LabelArch and LabelArchStable above).

towca (changes requested):
Could you squash the commits into just one, with an appropriate message? You can read about it here: https://levelup.gitconnected.com/how-to-squash-git-commits-9a095c1bc1fc.

In general, we prefer fixing comments on a PR by amending to an existing commit (git commit --amend) and then force-pushing (git push -f), instead of introducing additional commits just for the fixes.

PR #3719, opened 2020-11-25
CA-Chart - Mark Cluster-Autoscaler-Chart as Deprecated

gjtempleton:
**There will be a follow-up PR to remove these files as only one release marked as deprecated is required.**

This diff looks very big, however this is due to copying back in the files for the `cluster-autoscaler-chart` Helm chart as it was pre the renaming to `cluster-autoscaler` performed by #3679 .

The actual changes performed by this PR to the last previous release of the `cluster-autoscaler-chart` chart marked by https://github.com/kubernetes/autoscaler/tree/cluster-autoscaler-chart-1.1.1 can be seen by running 

```bash
git diff cluster-autoscaler-chart-1.1.1 charts/cluster-autoscaler-chart/

diff --git a/charts/cluster-autoscaler-chart/Chart.yaml b/charts/cluster-autoscaler-chart/Chart.yaml
index 4f86dcc4d..b288ae4a8 100644
--- a/charts/cluster-autoscaler-chart/Chart.yaml
+++ b/charts/cluster-autoscaler-chart/Chart.yaml
@@ -17,4 +17,5 @@ name: cluster-autoscaler-chart
 sources:
 type: application
-version: 1.1.1
+version: 2.0.0
+deprecated: true
diff --git a/charts/cluster-autoscaler-chart/README.md b/charts/cluster-autoscaler-chart/README.md
...
```
The changes to the readme are expected as pre-commit hooks are now enforced and have regenerated the readme from the template.

Performs a major version bump of cluster-autoscaler-chart to mark it as deprecated
in favour of `cluster-autoscaler` chart

This is required to hide the deprecated old version of the chart from the artifact hub UI, as currently users are shown both the old and new charts: https://artifacthub.io/packages/search?page=1&ts_query_web=cluster-autoscaler

Marking the chart as deprecated in the `Chart.yaml` will hide all releases of the old chart version from the UI. Performs a major version bump of the old version of the chart to 2.0.0 to mark the deprecation.

### DIFF
--- a/charts/cluster-autoscaler-chart/.helmignore
+++ b/charts/cluster-autoscaler-chart/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
--- a/charts/cluster-autoscaler-chart/Chart.yaml
+++ b/charts/cluster-autoscaler-chart/Chart.yaml
@@ -1,0 +1,21 @@
+apiVersion: v2
+appVersion: 1.18.1
+description: Scales Kubernetes worker nodes within autoscaling groups.
+engine: gotpl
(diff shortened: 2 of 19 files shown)


mwielgus:
Is this still needed/worked on?

gjtempleton:
Yep, had some discussions and the preference is for the option that makes this PR look rather messy unfortunately. I've updated the PR description to include instructions for seeing the changes from the last release of the chart as it used to be named here, the only meaningful changes being to bump the major version and mark the chart as deprecated.

mwielgus (approved):
/lgtm
/approve

PR #3722, opened 2020-11-25
update generic labels for GCE

yaroslava-serdiuk:


JimmyMaksymiw:
Hi @yaroslava-serdiuk,
Could we please get this fix with the [stable topology labels](https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#failure-domainbetakubernetesiozone) backported to 1.17 since the old once are now deprecated?

We are having issues in GKE land with kubernetes 1.17. The cluster autoscaler v1.7.x does not trigger any scale up when a node pool have 0 nodes and using the new topology labels.

MaciekPytel:
GKE uses its own fork of autoscaler. Backporting this PR will have no impact on it. 
I'll check how this looks in GKE fork and if some backporting is needed there.

towca:
I actually think we still use the same `BuildGenericLabels` function while creating template nodes in GKE.

JimmyMaksymiw:
Thanks for the info!
Could you get back if some backporting is needed and when that would go out to GKE?

yaroslava-serdiuk:
@JimmyMaksymiw The update on this topic: Stable topology labels are backported for clusters 1.19 and 1.18 in GKE, so stable topology labels are available for rapid and regular channel.

JimmyMaksymiw:
@yaroslava-serdiuk thanks for the info!
Just to make sure we are on the same page, the topology labels are backported to GKE regular channel which is 1.17? 😄 
image

yaroslava-serdiuk:
@JimmyMaksymiw Not really.
1.17 is moving to GKE stable channel and 1.18 is moving to GKE regular channel:
https://cloud.google.com/kubernetes-engine/docs/release-notes-stable
https://cloud.google.com/kubernetes-engine/docs/release-notes-regular

JimmyMaksymiw:
Ahh cool, that sounds good. Totally missed that, thank you! 👍

PR #3723, opened 2020-11-25
Improve error message: Not all error are due to "it wouldn't fit if a new node is added"

crgarcia12:
If we try to deploy a pod that is bigger than the node itself, currently we get an error message like this one:

**cluster-autoscaler pod didn't trigger scale-up (it wouldn't fit if a new node is added): 1 Insufficient cpu**

the problem is when the error is due to another reason. For example:
**cluster-autoscaler pod didn't trigger scale-up (it wouldn't fit if a new node is added): 1 in backoff after failed scale-up**

In that case, the problem has nothing to do with _"it wouldn't fit if a new node is added"_, so I think that only the second part of the error message is relevant in such cases, and the first brings some confusion.

The proposed change for the same case is :
**pod didn't trigger scale-up: 1 in backoff after failed scale-up**

### DIFF
--- a/cluster-autoscaler/processors/status/eventing_scale_up_processor.go
+++ b/cluster-autoscaler/processors/status/eventing_scale_up_processor.go
@@ -40,8 +40,8 @@ func (p *EventingScaleUpStatusProcessor) Process(context *context.AutoscalingCon
 	if status.Result != ScaleUpSuccessful && status.Result != ScaleUpError {
 		for _, noScaleUpInfo := range status.PodsRemainUnschedulable {
 			context.Recorder.Event(noScaleUpInfo.Pod, apiv1.EventTypeNormal, "NotTriggerScaleUp",
-				fmt.Sprintf("pod didn't trigger scale-up (it wouldn't fit if a new node is"+
-					" added): %s", ReasonsMessage(noScaleUpInfo, consideredNodeGroupsMap)))
+				fmt.Sprintf("pod didn't trigger scale-up: %s",
+					ReasonsMessage(noScaleUpInfo, consideredNodeGroupsMap)))
 		}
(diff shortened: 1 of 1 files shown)


MaciekPytel:
I have mixed feelings about this one - on one hand you're absolutely correct not all the reasons are related to pod "fitting" on a node. On the other hand I'm aware a lot people out there have filters set up for the current message and it may impact existing monitoring. IIRC when we introduced the reason part of the message we explicitly decided to leave the old, slightly misleading wording for the sake of backward compatibility.

I'm not feeling very strongly either way, maybe someone with knowledge of different providers can comment on how popular monitioring using the current wording is on their platform?

crgarcia12:
> I have mixed feelings about this one - on one hand you're absolutely correct not all the reasons are related to pod "fitting" on a node. On the other hand I'm aware a lot people out there have filters set up for the current message and it may impact existing monitoring. IIRC when we introduced the reason part of the message we explicitly decided to leave the old, slightly misleading wording for the sake of backward compatibility.
> 
> I'm not feeling very strongly either way, maybe someone with knowledge of different providers can comment on how popular monitioring using the current wording is on their platform?

That is a very good point!
I only work with Azure, in there it will not be a problem.

The reason for the PR is that I already got several people trying to create bigger nodes or splitting pods because they believe that the deployment does not fit. When you are doing a big deployment with multiple pods/services then it is even harder because there is no reference to which pod is creating the problem

I am confident than on Azure this change would no be a problem, but I cannot guarantee for other providers, so I am also fine either way

mwielgus:
Let's have this issue fixed and add the information about log text change to release notes.

mwielgus (approved):
/lgtm
/approve

Issue #3724, opened 2020-11-26
Cluster Autoscaler examples and helm chart is different


foriequal0:
I wanted to deploy cloud-autoscaler using Helm Chart to ease the maintance.
So I've tweaked helm values to match with the AWS Cloud Autoscaler manual ([here](https://docs.aws.amazon.com/ko_kr/eks/latest/userguide/cluster-autoscaler.html)) and the example ([here](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml))
```js
{
  fullnameOverride: "cluster-autoscaler",
  image: {
    // Updated as the manual says
    tag: "v1.18.3",
    pullPolicy: "Always"
  },
  
  // As in the example
  resources: {
    limits: {
      cpu: "100m",
      memory: "300Mi"
    },
    requests: {
      cpu: "100m",
      memory: "300Mi"
    }
  },
  // As in the example
  podDisruptionBudget: null,

  cloudProvider: "aws",
  awsRegion: "<YOUR CLUSTER REGION>",

  autoDiscovery: {
    clusterName: "<YOUR CLUSTER NAME>"
  },
  extraArgs: {
    // Helm chart default values
    // logtostderr: true, // Doesn't exist in the example
    stderrthreshold: "info",
    v: 4,
    // Exists in the example
    "skip-nodes-with-local-storage": false,
    expander: "least-waste",
    // Exists in the manual
    "balance-similar-node-groups": true,
    "skip-nodes-with-system-pods": false
  }
}
```
However there are still subtle differences:
1. The chart has extra `ClusterRole` for batch, batch/extensions, but the example doesn't:
```yaml
  - apiGroups: ["batch"],
    resources: ["jobs", "cronjobs"]
    verbs: ["watch", "list", "get"]
```
2. The example has extra `Role` for some resources
The chart has a role only for
```yaml
  - apiGroups: [""],
    resources: ["configmaps"]
    verbs: ["create"]
  - apiGroups: [""],
    resources: ["cluster-autoscaler-status"]
    verbs: ["delete", "get", "update"]
```
However the example has more roles:
```yaml
  - apiGroups: [""]
    resources: ["configmaps"]
    verbs: ["create","list","watch"]
  - apiGroups: [""]
    resources: ["configmaps"]
    resourceNames: ["cluster-autoscaler-status", "cluster-autoscaler-priority-expander"]
[...]

PR #3725, opened 2020-11-26
Fix cluster-autoscaler chart to match with the examples

foriequal0:
Partially resolves: https://github.com/kubernetes/autoscaler/issues/3724

Following lint error is well known issue:
```
templates/: template: cluster-autoscaler/templates/role.yaml:15:7: executing "cluster-autoscaler/templates/role.yaml" at <eq (default .Values.extraArgs.expander "") "priority">: error calling eq: invalid type for comparison
```
https://github.com/helm/helm/issues/6376
The workaround doesn't seem to work.

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -17,4 +17,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.3.2
+version: 9.4.0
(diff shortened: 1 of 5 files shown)


foriequal0:
I've overlooked https://github.com/kubernetes/autoscaler/pull/3692 that it would solve the issue, but it wasn't enough. I should've read the code carefully and commented to it.

gjtempleton:
I've had a play around with the code, and I think the issue you're seeing boils down to this: https://github.com/helm/helm/issues/7091

The current default is an empty string, which YAML is then viewing as an unquoted `nil`.

I think there's a couple of options here:

1. We pipe the default to `quote`, i.e. `{{- if eq (default .Values.extraArgs.expander "" | quote ) "priority" }}`
1. We set the default to a dummy string/other expander, but don't set a default for the values in the values file.
1. We set a default value of `expander:` in the top level values

foriequal0:
Oh. I've mistakenly swapped `default`'s argument. `default` takes a default value as a first argument. Swapping two arguments fixes that. I'll rebase and fix that.

foriequal0:
@gjtempleton Can you review the fix?

gjtempleton:
Sorry for the delay, thanks for fixing that!

/lgtm
/approve

PR #3729, opened 2020-11-27
Cache controllers 09

jbartosik:
Cherry pick #3589 to vpa 0.9


jbartosik:
@bskiba please take a look

PR #3730, opened 2020-11-27
Update vendor dependencies

BigDarkClown:


MaciekPytel:
/lgtm
/approve

@elmiko @detiber FYI - this includes some small unittest fixes to clusterapi provider that were required because of the changes in upstream test framework. I think this is fairly mechanical change and hopefully it shouldn't cause any problems for you.

elmiko:
thanks for the update @MaciekPytel !

PR #3737, opened 2020-12-03
classify configmap update failures as apiError failures so they factor into the liveness checks

marwanad:
There are two issues here that this PR is trying to address by signaling failure to the liveness probe:

edit: leader election is set to false, I believe that would help with case 2 since it will force a restart, not sure about case 1.

### Issue 1 with Possible Stale Connection Lingering on 
Seen with both 1.17 and 1.18 with an in-cluster kubeconfig. It seems that a high churn of rolling apiserver deployments, can cause a weird failure mode with reflectors failing to update the store and reading from a stale cache. It seems to be stuck making calls to a stale endpoint.

Below are the logs from the suspected stale connection run (unfortunately no repro but prior to that the apiserver was bounced off nodes a few times):
```
"I1130 02:37:59.152285       1 reflector.go:211] Listing and watching *v1.ReplicaSet from k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes/listers.go:347
"
"E1130 02:37:59.152592       1 reflector.go:178] k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes/listers.go:347: Failed to list *v1.ReplicaSet: Get https://kube-apiserver.svc.cluster.local:443/apis/apps/v1/replicasets?resourceVersion=1976811: write tcp 172.31.6.93:51230->172.30.146.11:443: write: broken pipe
"
"E1130 02:37:56.528825       1 reflector.go:178] k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes/listers.go:246: Failed to list *v1.Node: Get https://kube-apiserver.svc.cluster.local:443/api/v1/nodes?resourceVersion=2376223: write tcp 172.31.6.93:51230->172.30.146.11:443: write: broken pipe
"
"I1130 02:37:56.528166       1 reflector.go:211] Listing and watching *v1.Node from k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes/listers.go:246
"
"I1130 02:37:53.726440       1 scale_down.go:867] No candidates for scale down
"
"I1130 02:37:53.726295       1 pre_filtering_processor.go:66] Skipping pool1-vmss node group min size reached
"
"I1130 02:37:53.726131       1 static_autoscaler.go:391] No unschedulable pods
"
"E1130 02:37:53.726744       1 status.go:123] Failed to retrieve status configmap for update: Get https://kube-apiserver.svc.cluster.local:443/api/v1/namespaces/kube-system/configmaps/cluster-autoscaler-status: write tcp 172.31.6.93:51230->172.30.146.11:443: write: broken pipe
"
```
I believe the above ties to a plethora of upstream with watches and recovering from connections which may be solved with the change to enabled HTTP 2 health checks in client-go that went in 1.20 and 1.19. The problem with this case is that CA never recovers. It will continue with those broken pipe errors until the pod is restarted.

### Issue 2 with No Api Server Running
Upon further investigations, I discovered that even with no apiserver running - CA doesn't report liveness probe failures and continues normal operations.

```
E1203 03:32:35.655334       1 reflector.go:178] k8s.io/client-go/informers/factory.go:135: Failed to list *v1.PersistentVolumeClaim: Get https://kube-apiserver.svc.cluster.local:443/api/v1/persistentvolumeclaims?limit=500&resourceVersion=0: dial tcp 172.30.40.244:443: i/o timeout
I1203 03:32:41.652798       1 trace.go:116] Trace[545566877]: "Reflector ListAndWatch" name:k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes/listers.go:213 (started: 2020-12-03 03:32:11.65219098 +0000 UTC m=+5391.186317723) (total time: 30.000573274s):
Trace[545566877]: [30.000573274s] [30.000573274s] END
E1203 03:32:41.652824       1 reflector.go:178] k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes/listers.go:213: Failed to list *v1.Pod: Get https://kube-apiserver.svc.cluster.local:443/api/v1/pods?fieldSelector=spec.nodeName%21%3D%2Cstatus.phase%21%3DFailed%2Cstatus.phase%21%3DSucceeded&limit=500&resourceVersion=0: dial tcp 172.30.40.244:443: i/o timeout
I1203 03:32:42.299129       1 reflector.go:211] Listing and watching *v1.Node from k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes/listers.go:247
I1203 03:32:48.052918       1 trace.go:116] Trace[1559053748]: "Reflector ListAndWatch" name:k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes/listers.go:312 (started: 2020-12-03 03:32:18.052253979 +0000 UTC m=+5397.586380622) (total time: 30.000638637s):
Trace[1559053748]: [30.000638637s] [30.000638637s] END
E1203 03:32:48.052944       1 reflector.go:178] k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes/listers.go:312: Failed to list *v1beta1.PodDisruptionBudget: Get https://kube-apiserver.svc.cluster.local:443/apis/policy/v1beta1/poddisruptionbudgets?limit=500&resourceVersion=0: dial tcp 172.30.40.244:443: i/o timeout
I1203 03:32:48.659621       1 trace.go:116] Trace[1185234412]: "Reflector ListAndWatch" name:k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes/listers.go:189 (started: 2020-12-03 03:32:18.658965721 +0000 UTC m=+5398.193092464) (total time: 30.000621397s):
Trace[1185234412]: [30.000621397s] [30.000621397s] END
E1203 03:32:48.659651       1 reflector.go:178] k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes/listers.go:189: Failed to list *v1.Pod: Get https://kube-apiserver.svc.cluster.local:443/api/v1/pods?fieldSelector=spec.nodeName%3D%2Cstatus.phase%21%3DFailed%2Cstatus.phase%21%3DSucceeded&limit=500&resourceVersion=0: dial tcp 172.30.40.244:443: i/o timeout
I1203 03:32:49.972905       1 trace.go:116] Trace[211996317]: "Reflector ListAndWatch" name:k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes/listers.go:332 (started: 2020-12-03 03:32:19.972369022 +0000 UTC m=+5399.506495665) (total time: 30.000512002s):
Trace[211996317]: [30.000512002s] [30.000512002s] END
E1203 03:32:49.972929       1 reflector.go:178] k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes/listers.go:332: Failed to list *v1.ReplicationController: Get https://kube-apiserver.svc.cluster.local:443/api/v1/replicationcontrollers?limit=500&resourceVersion=0: dial tcp 172.30.40.244:443: i/o timeout
I1203 03:32:52.660054       1 node_instances_cache.go:156] Start refreshing cloud provider node instances cache
I1203 03:32:52.660092       1 node_instances_cache.go:168] Refresh cloud provider node instances cache finished, refresh took 10.201µs
E1203 03:32:56.035436       1 status.go:123] Failed to retrieve status configmap for update: Get https://kube-apiserver.svc.cluster.local:443/api/v1/namespaces/kube-system/configmaps/cluster-autoscaler-status: dial tcp 172.30.40.244:443: i/o timeout
````
This case recovers when the api server comes up so it's less concerning.

It seems that there are no errors triggered along the paths so the CA main loop runs to completion with stale or false assumptions:
https://github.com/kubernetes/autoscaler/blob/19582d95b016d09b585a0e1d9c4738dbbc98e26e/cluster-autoscaler/core/static_autoscaler.go#L233-L244
 
The errors here are always nil: https://github.com/kubernetes/autoscaler/blob/19582d95b016d09b585a0e1d9c4738dbbc98e26e/cluster-autoscaler/utils/kubernetes/listers.go#L256-L270

https://github.com/kubernetes/kubernetes/blob/5b8c3b90f3c74ba79f4bfa5be560d011694a7da0/staging/src/k8s.io/client-go/listers/core/v1/node.go#L51-L56

I believe a problem with keeping CA alive in both cases is that it may unnecessarily see unregistered nodes and attempt to delete those. Let's say your starting state on the cloudprovider side is 10 VMs, you kick-off CA with an unavailable apiserver, you lose those 10 VMs as unregistered nodes because your main loop runs to completion with those failures. It's a bit unrealistic but it seems that [the expectations for this codepath to fail](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/static_autoscaler.go#L239-L243) doesn't actually holds, unless I'm missing something.
 
I'm open to other suggestions on how to get CA to recover from such cases. It seems that those errors won't be visible to the main loop so the only way to have this factor in the liveness check is by using the configmap write failure signal (which may or may not be a 100% indicator because missing RBAC permissions can cause configmap write failures). But then again, if we classify [those pod listing](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/static_autoscaler.go#L239-L243) as non-transient autoscaler errors, then I don't see why we shouldn't do the same for the configmap writes if the configmap writes are enabled.

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -215,7 +215,7 @@ func (a *StaticAutoscaler) initializeClusterSnapshot(nodes []*apiv1.Node, schedu
 }
 
 // RunOnce iterates over node groups and scales them up/down if necessary
-func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError {
+func (a *StaticAutoscaler) RunOnce(currentTime time.Time) (typedErr errors.AutoscalerError) {
 	a.cleanUpIfRequired()
 	a.processorCallbacks.reset()
 	a.clusterStateRegistry.PeriodicCleanup()
@@ -293,8 +293,12 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 		// Update status information when the loop is done (regardless of reason)
 		if autoscalingContext.WriteStatusConfigMap {
 			status := a.clusterStateRegistry.GetStatus(currentTime)
-			utils.WriteStatusConfigMap(autoscalingContext.ClientSet, autoscalingContext.ConfigNamespace,
+			_, writeErr := utils.WriteStatusConfigMap(autoscalingContext.ClientSet, autoscalingContext.ConfigNamespace,
 				status.GetReadableString(), a.AutoscalingContext.LogRecorder)
+			// If no other errors occurred during the run and the configmap write failed, set the typedErr
+			if typedErr == nil && writeErr != nil {
+				typedErr = errors.NewAutoscalerError(errors.ApiCallError, "%v", writeErr)
+			}
 		}
 
 		// This deferred processor execution allows the processors to handle a situation when a scale-(up|down)

### CODE REVIEW

**Comment by marwanad** [comment]
/hold

@MaciekPytel @mwielgus curious what are your thoughts here.

**Comment by marwanad** [comment]
I wonder if issue 1 is actually related to: https://github.com/kubernetes/kubernetes/pull/89652

**Comment by marwanad** [comment]
I guess we can also have some dummy call that doesn't use listers to fetch deployments, or something and fail early if this fails

**Review by feiskyer — COMMENTED** [review]
LGTM

**Comment by feiskyer** [comment]
even https://github.com/kubernetes/kubernetes/pull/89652 is related, I think the fix here still makes sense as other errors may also happen.

File: cluster-autoscaler/core/static_autoscaler.go [hunk 1]
```diff
@@ -293,8 +293,12 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 		// Update status information when the loop is done (regardless of reason)
 		if autoscalingContext.WriteStatusConfigMap {
 			status := a.clusterStateRegistry.GetStatus(currentTime)
-			utils.WriteStatusConfigMap(autoscalingContext.ClientSet, autoscalingContext.ConfigNamespace,
+			_, writeErr := utils.WriteStatusConfigMap(autoscalingContext.ClientSet, autoscalingContext.ConfigNamespace,
 				status.GetReadableString(), a.AutoscalingContext.LogRecorder)
+			// If no other errors occurred during the run and the configmap write failed, set the typedErr
+			if typedErr == nil && writeErr != nil {
+				typedErr = errors.NewAutoscalerError(errors.ApiCallError, "%v", writeErr)
```
bpineau, line 300: Perhaps specifically on timeouts errors?
Writing that ConfigMap often fails on large clusters (where the status might exceed 1MiB). ie. seeing right now:
```
Failed to write status configmap: ConfigMap "cluster-autoscaler-status" is invalid: []: Too long: must have at most 1048576 characters
```
which is currently harmless

marwanad: good call

marwanad:
@mwielgus @MaciekPytel friendly ping.

PR #3738, opened 2020-12-03
CA-Chart - 9.2.0 Release

gjtempleton:
* Brings in new functionality from #3692 

Addition/re-organisation of `type: application` labels on all previous charts due to update of local helm version being used to generate the `index.yaml`


mwielgus (approved):
/lgtm
/approve

Issue #3742, opened 2020-12-04
Add EC2 m5zn instance types


koenvo:
The new m5zn instance types are missing in `ec2_instance_types.go` which results in `uses the unknown EC2 instance type "m5zn.3xlarge"` messages.

AWS article: https://aws.amazon.com/about-aws/whats-new/2020/12/introducing-amazon-ec2-m5zn-instances/

I'll try to create a pull request for it.

PR #3746, opened 2020-12-04
Add new aws instance types: r5b, m5zn, d3, d3en

dbonfigli:
Update the aws instance types map with the recently added new instance types.


mcristina422:
Can you use this tool to generate this file? https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/ec2_instance_types/gen.go 

When I do I see more instances like `t4*`, `p4d.24xlarge`

dbonfigli:
You are right, I missed that command from the doc, I updated the list with it now.

koenvo:
> Cover #3742 too.

Thanks! I was still struggling to get the `gen.go` script working.

aianus:
Anything left to do here besides the approval?

Would love to switch our CI cluster to m5zn

gjtempleton:
As you say, should just be the approval from one of the AWS cloud provider owners required now.

I've just noticed that @Jeffwan is no longer at AWS, tagging in @jaypipes as well just in case.

jaypipes:
/lgtm
/approve

thank you @aianus!

PR #3751, opened 2020-12-07
Add comment for exported type InstanceType in gen template

dbonfigli:
The "gen" tool used to generate ec2_instance_types.go creates the file without a comment for the exported type InstanceType so golint complains in the CI, this patch adds the same comment that currently is in ec2_instance_types.go in the template too.


mwielgus (approved):
/lgtm
/approve

PR #3754, opened 2020-12-09
Unreachable code removed

koef:
Looks like there is [unreachable code](https://github.com/kubernetes/autoscaler/blob/0fb897b83992787fd75f262499c71b8ff3bfd14c/cluster-autoscaler/core/scale_up.go#L82) in `if errCoresMem != nil` condition under the `switch`.

Because this case was handled above:
```
if (resource == cloudprovider.ResourceNameCores || resource == cloudprovider.ResourceNameMemory) && errCoresMem != nil
```


mwielgus:
CLA is missing. We won't be able to merge the change without CLA.

PR #3755, opened 2020-12-09
Add max pods per node config to build node

vivekbagade:


cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -37,6 +37,11 @@ import (
 // GceTemplateBuilder builds templates for GCE nodes.
 type GceTemplateBuilder struct{}
 
+// MaxPodsPerNode contains info on max pods per node.
+type MaxPodsPerNode struct {
```
MaciekPytel, line 41: Why make this a struct and not just *int64? Do we expect more fields in future?

vivekbagade: Added it just to be safe. But ya, don't think it is needed

cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -52,10 +52,15 @@ func (t *GceTemplateBuilder) getAcceleratorCount(accelerators []*gce.Accelerator
 }
 
 // BuildCapacity builds a list of resource capacities given list of hardware.
-func (t *GceTemplateBuilder) BuildCapacity(cpu int64, mem int64, accelerators []*gce.AcceleratorConfig, os OperatingSystem) (apiv1.ResourceList, error) {
+func (t *GceTemplateBuilder) BuildCapacity(cpu int64, mem int64, accelerators []*gce.AcceleratorConfig, os OperatingSystem, pods *int64) (apiv1.ResourceList, error) {
 	capacity := apiv1.ResourceList{}
 	// TODO: get a real value.
-	capacity[apiv1.ResourcePods] = *resource.NewQuantity(110, resource.DecimalSI)
+	if pods == nil {
```
MaciekPytel, line 58: nit: maybe remove TODO above? It's pretty clear that this code no longer needs to 'get a real value' since it would now get it as a parameter.

vivekbagade: Done

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -485,7 +485,7 @@ func (m *gceManagerImpl) GetMigTemplateNode(mig Mig) (*apiv1.Node, error) {
 	if err != nil {
 		return nil, err
 	}
-	return m.templates.BuildNodeFromTemplate(mig, template, cpu, mem)
+	return m.templates.BuildNodeFromTemplate(mig, template, cpu, mem, nil)
```
ryanmcnamara, line 488: Is pods always going to be nil then? It doesn't seem it's possible to actually set this by configuring the mig...

PR #3757, opened 2020-12-11
Reduce logging when creating placeholder instances

mcristina422:
On clusters with large pools this log line is extremely excessive. The message can be printed once and the names of the created nodes can be inferred if desired


mwielgus (approved):
/lgtm
/approve

PR #3758, opened 2020-12-11
Bugfix: Expander Priority warns misleading log.

### LINKED ISSUE #3735: Priority expander: {node-group} not found in priority expander configuration
While using `cluster-autoscaler:v1.18.3` with kubernetes 1.18.10 we are seeing in the logs that it cannot find some node groups

priority-expander configMap:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: cluster-autoscaler-priority-expander
data:
  priorities: |-
    10:
      - .*
    20:
      - .*nodes-a-ondemand.*
      - .*nodes-b-ondemand.*
    30:
      - .*nodes-a-spot.*
      - .*nodes-b-spot.*
```

cluster-autoscaler container arguments:
```
          command:
            - ./cluster-autoscaler
            - --v=4
            - --stderrthreshold=info
            - --cloud-provider=aws
            - --expander=priority
            - --skip-nodes-with-system-pods=false
            - --nodes=1:2:nodes-cifs.cluster.k8s.tld
            - --nodes=0:12:nodes-a-ondemand.cluster.k8s.tld
            - --nodes=2:12:nodes-a-spot.cluster.k8s.tld
            - --nodes=0:12:nodes-b-ondemand.cluster.k8s.tld
            - --nodes=2:12:nodes-b-spot.cluster.k8s.tld
            - --nodes=0:4:nodes-ssd-spot.cluster.k8s.tld
            - --balance-similar-node-groups=true
```

We see these messages in the logs:
```
1 priority.go:91] Priority expander: node group nodes-a-ondemand.cluster.k8s.tld not found in priority expander configuration. The group won't be used.
1 priority.go:91] Priority expander: node group nodes-b-ondemand.cluster.k8s.tld not found in priority expander configuration. The group won't be used.
```

Originally we did not have the wildcard regex at priority 10 however after reading the suggestion [HERE](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/expander/priority/readme.md#configuration) we added it. However even with the wildcard regex it still does not match the node group and the message persists in the logs.
> Note that if a group name doesn't match any of the regular expressions in the priority list it will not be considered for expansion. To ensure that all of your groups are autoscaled you might want to add a "catch-all" regex of .* (with a low priority) to your priorities list.

Previously on `cluster-autoscaler:v1.18.2` we did not see this message however we did see the issue that was fixed in #3308

umialpha:
Fixes #3735.

### DIFF
--- a/cluster-autoscaler/expander/priority/priority.go
+++ b/cluster-autoscaler/expander/priority/priority.go
@@ -135,8 +135,10 @@ func (p *priority) BestOption(expansionOptions []expander.Option, nodeInfo map[s
 	for _, option := range expansionOptions {
 		id := option.NodeGroup.Id()
 		found := false
+		skipped := false
 		for prio, nameRegexpList := range priorities {
 			if prio < maxPrio {
+				skipped = true
 				continue
 			}
 			if !p.groupIDMatchesList(id, nameRegexpList) {
@@ -149,7 +151,7 @@ func (p *priority) BestOption(expansionOptions []expander.Option, nodeInfo map[s
 			best = append(best, option)
 			found = true
 		}
-		if !found {
+		if !skipped && !found {
 			msg := fmt.Sprintf("Priority expander: node group %s not found in priority expander configuration. "+
 				"The group won't be used.", id)
 			p.logConfigWarning(cm, "PriorityConfigMapNotMatchedGroup", msg)
--- a/cluster-autoscaler/expander/priority/priority.go
+++ b/cluster-autoscaler/expander/priority/priority.go
@@ -135,13 +135,12 @@ func (p *priority) BestOption(expansionOptions []expander.Option, nodeInfo map[s
 	for _, option := range expansionOptions {
 		id := option.NodeGroup.Id()
 		found := false
-		skipped := false
 		for prio, nameRegexpList := range priorities {
-			if prio < maxPrio {
-				skipped = true
+			if !p.groupIDMatchesList(id, nameRegexpList) {
 				continue
 			}
-			if !p.groupIDMatchesList(id, nameRegexpList) {
+			found = true
+			if prio < maxPrio {
 				continue
 			}
 			if prio > maxPrio {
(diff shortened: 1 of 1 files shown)


mwielgus:
Please sign the CLA. We won't be able to merge the PR without it.

cluster-autoscaler/expander/priority/priority.go
```diff
@@ -135,8 +135,10 @@ func (p *priority) BestOption(expansionOptions []expander.Option, nodeInfo map[s
 	for _, option := range expansionOptions {
 		id := option.NodeGroup.Id()
 		found := false
+		skipped := false
 		for prio, nameRegexpList := range priorities {
 			if prio < maxPrio {
+				skipped = true
```
mwielgus, line 141: If i understand the code correctly, this can only be set if maxPrio != -1, and when it happens found=true. So we have the following combinations possible:

found = false skipped = false -> log should be printed
found = true skipped = false -> log should not be printed 
found = true skipped = true -> log should not be printed

found = false and skipped=true is not possible

thus, line 154 can simply be left as it was and skipped (and the whole PR) is not needed

mwielgus (changes requested):

cluster-autoscaler/expander/priority/priority.go
umialpha: `if prio > maxPrio {
				maxPrio = prio
				best = nil
			}`
 when it finds a higher priority, it updates maxPrio, so the node group with lower priority will be skipped.
But when skipped, it warns "node group %s not found in priority expander configuration.". It confuses customers. 

I change the conditions to: when the node group is not skipped and not matched, it throws warnings.

umialpha: Here is a example.
suppose the priority list is [[1, "\*"], [10, "A"], [11, "C"]], the nodegroup list is [Node("A"), Node("B")]. Node("A") will be the best option. Node("B") matches [1, "*"] but it still warns "**node group B not found in priority expander configuration.**"

umialpha: You are right, I fixed it in a wrong way. But the problem is still there. Here is an example.

Suppose the priority list is [[1, "\*"], [10, "A"], [11, "C"]], the nodegroup list is [Node("A"), Node("B")]. Node("A") will be the best option. Node("B") matches [1, "*"] but it warns "node group B not found in priority expander configuration."

Could we remove the block `if prio < maxPrio {
				continue
			}`. Yes it has some performance cost but we have to compare all the regex in order to warn misconfiguration to customers.

vinayan3:
Any updates on getting this reviewed and closed? I'm also seeing the misleading log statements.

mwielgus (approved):
/lgtm
/approve

PR #3761, opened 2020-12-14
NodeInfo processor to refine template-based NodeInfos

bpineau:
Comparing synthetic NodeInfos obtained from nodegroup's TemplateInfo() to NodeInfos obtained from real-world nodes is bound to fail, even with kube reservations provided through nodegroups labels/annotations (for instance: kernel mem reservation is hard to predict, in particular on AWS instances).

This makes `balance-similar-node-groups` likely to misbehave when `scale-up-from-zero` is enabled (and a first nodegroup gets a real node), for instance.

Following [Maciek suggestion](https://github.com/kubernetes/autoscaler/pull/3608#issuecomment-709308255) (from discussions on a previous attempt at solving this), we can implement a NodeInfo Processor that would improve template-generated NodeInfos whenever a node was created off a similar nodegroup.

We're storing node's virtual origin through machineid, which works fine but is a bit ugly (suggestions welcome).

Tested this solves balance-similar-node-groups + scale-up-from-zero, with various instance types on AWS and GCP.

Previous attempts to solve that issue/discussions:
* https://github.com/kubernetes/autoscaler/issues/2892#issuecomment-595769765
* https://github.com/kubernetes/autoscaler/pull/3608#issuecomment-709308255


MaciekPytel:
A general comment - this strongly assumes that two NodeGroups that are "similar" according to NodeGroupSet processor are _identical_ for scheduling purposes (except for location related labels). I think this is likely to be true in most cases, but it's still a pretty significant assumption. 
If the NodeGroups in question are not identical this assumption can lead to incorrect scale-up decision (since the template used in binpacking would differ from actual node). This makes me think that enabling the processor should probably be controlled via a flag. WDYT?

cluster-autoscaler/processors/nodeinfos/refine_node_infos_processor.go
```diff
@@ -0,0 +1,174 @@
...
+
+type nodeInfoCache struct {
+	sync.RWMutex
+	nodeInfos   map[string]*schedulerframework.NodeInfo
+	lastUpdated time.Time
+}
+
+// Process refines nodeInfos obtained by TemplateInfos when possible
+func (p *RefineNodeInfosProcessor) Process(ctx *context.AutoscalingContext, nodeInfosForNodeGroups map[string]*schedulerframework.NodeInfo) (map[string]*schedulerframework.NodeInfo, error) {
+	innacurateNodeInfos := make(map[string]*schedulerframework.NodeInfo)
```
umialpha, line 52: typo: innacurate?

cluster-autoscaler/processors/nodeinfos/refine_node_infos_processor.go
```diff
@@ -0,0 +1,174 @@
...
+
+	nodeGroups := make(map[string]cloudprovider.NodeGroup)
+	for _, nodeGroup := range ctx.CloudProvider.NodeGroups() {
+		nodeGroups[nodeGroup.Id()] = nodeGroup
+	}
+
+	if err := p.nodeInfoCache.update(nodeGroups); err != nil {
+		return nodeInfosForNodeGroups, err
+	}
+	p.nodeInfoCache.RLock()
```
mwielgus, line 63: Why do you need this lock?

mwielgus (changes requested):

cluster-autoscaler/processors/nodeinfos/refine_node_infos_processor.go
```diff
@@ -0,0 +1,174 @@
...
+
+func (n *nodeInfoCache) update(nodeGroups map[string]cloudprovider.NodeGroup) error {
+	n.Lock()
+	defer n.Unlock()
+
+	if len(n.nodeInfos) == 0 {
+		n.nodeInfos = make(map[string]*schedulerframework.NodeInfo)
+		n.lastUpdated = time.Now()
+	}
+	needsRefresh := n.lastUpdated.Add(templateNodeInfoCacheTTL).Before(time.Now())
```
umialpha, line 144: Why do we need cache TTL?

umialpha, line 144: I think we could remove cache TTL. Instead we could,
1. If the new nodeinfo is real node, update the cache.
2. Else, use cached real nodeinfo.

Assume we never or  rarely change instance type within a nodegroup. So in most time, the cache works well.
Anyway if we do change , our cache will be updated later(when scaling up from 1 or more).

umialpha (commented):
IIUC, this PR tries to find **real** or **similar real** nodeinfo as much as possible. Could we try the following routine.

1. If it is a real nodeinfo, update nodecache.
2. Else, 
   - If nodecache has cached real nodeinfo, use cached nodeinfo.
   - Else, use the most similar real nodeinfo and update the cache.

cluster-autoscaler/processors/nodeinfos/refine_node_infos_processor.go
bpineau: Thanks; innacurate was indeed an inaccurate spelling!

cluster-autoscaler/processors/nodeinfos/refine_node_infos_processor.go
bpineau: Not really right now. Removed.

cluster-autoscaler/processors/nodeinfos/refine_node_infos_processor.go
bpineau: Not sure to understand the suggestion. We don't cache nodeInfos from real-world nodes, only templated nodeInfos which we can then use as keys for comparisons/findsimilar (then once similar groups ids are found, we use real-world nodeinfos as provided as an arg to that processor (no need to cache those)).

Cache TTL is a distinct matter; it's not essential and can be removed from the current implememtation if we prefer so; it's meant to eventually catch up if a nodegroup is updated (or deleted+recreated): scale-up might never happen with stale infos.

bpineau:
@MaciekPytel agreed, added a (default false) flag.

umialpha (commented):
Sorry for misunderstanding. I list my understanding as follows. Please point out if anything wrong.

This PR tries to do,
If a NodeInfo is a template, find the most "similar" RealNode to replace it. But instead of comparing template node with real nodes, it compare template node with template nodes.  Right?

IIUC, if this logic only influences "BalanceSimilarNodeGroups", would it be better to make it closer to "BalanceSimilarNodeGroups" logic, e.g. put this logic into `processors.NodeGroupSetProcessor.FindSimilarNodeGroups(context, bestOption.NodeGroup, nodeInfos)`

This logic takes place at the very beginning of RunOnce logic. As @MaciekPytel said,

>  this strongly assumes that two NodeGroups that are "similar" according to NodeGroupSet processor are identical for scheduling purposes

If the assumption is incorrect, it will lead to  incorrect scale-up decision.

WDYT? Also @MaciekPytel .

bpineau:
@umialpha Yes, that PR logic description is correct. We're following suggestions from previous attempts at solving this (in particular https://github.com/kubernetes/autoscaler/issues/2892#issuecomment-595769765 and https://github.com/kubernetes/autoscaler/pull/3608#issuecomment-709308255 ).

You're right about the assumption, hence the new flag to be enabled when using scale-up-from-zero (otherwise we already have real nodes and this change is no-op) and where nodeGroups having similar nodeInfos templates are similar nodeGroups.

Having more accurately evaluated nodeGroups capacities (the intent of this PR) is useful in other situations than just keeping nodegroups balanced with FindSimilarNodeGroups during upscale; for instance when evaluating options to re-pack undersubscribed nodes.

MaciekPytel:
Agreed, this goes beyond just balancing. The fact that TemplateNodeInfo generally doesn't get resources exactly right is a problem that can have more impact that not matching NodeGroups for balancing purposes. Consider an example where TemplateNodeInfo overestimates memory by a little bit and a pod that won't really fit on the node would look like it fits - in this case CA would trigger an incorrect scale-up. NodeGroupSetProcessor doesn't really apply to this situation.

Using a template from a real node is generally more likely to be correct and avoid problems like this _if you are sure_ the new node will really be the same (except for location-related stuff). As per my previous comment I don't think we should make this assumption for the users, but if you happen to know that it's true in your env what this PR does makes a lot of sense to me. As such I'm ok with adding this as an opt-in feature (note: I haven't done a detailed review, just talking about general approach).

bpineau:
@mwielgus (and/or @MaciekPytel ) mind taking an other look? it's updated according to the comments

cluster-autoscaler/main.go
```diff
@@ -323,6 +326,11 @@ func buildAutoscaler() (core.Autoscaler, error) {
 		Comparator: nodeInfoComparatorBuilder(autoscalingOptions.BalancingExtraIgnoredLabels),
 	}
 
+	opts.Processors.NodeInfoProcessor = &nodeinfos.RefineNodeInfosProcessor{
```
MaciekPytel, line 329: I would rather just not install processor at all if the setting is disabled (instead of having the processor do no-op based on the setting). My reasons are:
 * When the setting is disabled it doesn't do anything, so why call it at all?
 * It (marginally) reduces blast radius of any bugs.
 * It's clear to the author of next NodeInfoProcessor implementation that it's safe to ignore this processor if they're not interested in the functionality. The same is not true for some other processors (ex. filterOutSchedulablePodListProcessor).

cluster-autoscaler/core/utils/utils.go
```diff
@@ -181,13 +183,27 @@ func GetNodeInfoFromTemplate(nodeGroup cloudprovider.NodeGroup, daemonsets []*ap
...
 	if typedErr != nil {
 		return nil, typedErr
 	}
 	return sanitizedNodeInfo, nil
 }
 
+// SetNodeInfoBuiltFromTemplate marks a NodeInfo as generated from a cloud provider's TemplateInfo()
+func SetNodeInfoBuiltFromTemplate(nodeInfo *schedulerframework.NodeInfo) {
+	node := nodeInfo.Node()
+	node.Status.NodeInfo.MachineID = templatedMachineID
```
MaciekPytel, line 197: I have mixed feelings about using a semi-random field in node status for this. It's hard to guarantee we'll never have support cloudprovider that use this field for something. Maybe an annotation would be more appropriate? I don't have a strong feeling here though.

cluster-autoscaler/processors/nodeinfos/refine_node_infos_processor.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// CleanUp cleans up processor's internal structures.
+func (p *RefineNodeInfosProcessor) CleanUp() {
+}
+
+func mirrorNodeInfoFromSimilarGroup(oldInfo, newInfo *schedulerframework.NodeInfo, nodeGroupName string, ignoredTaints taints.TaintKeySet) (*schedulerframework.NodeInfo, errors.AutoscalerError) {
+	newInfoCopy := newInfo.Clone()
+
+	// keep locality labels (region, zone, failure domain, ...) from original template
```
MaciekPytel, line 126: Is this really all we need from template? I think it would probably be good to have things like node name be different between templates. Another example are daemonset pods (that I think will be part of nodeInfo) - how do we know there are no zonal DS? Not sure about other stuff, but it would be good to double check if there is anything else worth randomizing or keeping from old template.

MaciekPytel:
Sorry for delay, I left a few minor comments. Overall, looks good, they're all pretty minor comments.

cluster-autoscaler/main.go
```diff
@@ -177,6 +178,7 @@ var (
 	clusterAPICloudConfigAuthoritative = flag.Bool("clusterapi-cloud-config-authoritative", false, "Treat the cloud-config flag authoritatively (do not fallback to using kubeconfig flag). ClusterAPI only")
 	cordonNodeBeforeTerminate          = flag.Bool("cordon-node-before-terminating", false, "Should CA cordon nodes before terminating during downscale process")
 	daemonSetEvictionForEmptyNodes     = flag.Bool("daemonset-eviction-for-empty-nodes", false, "DaemonSet pods will be gracefully terminated from empty nodes")
+	enableRefineUsingSimilarNodeGroups = flag.Bool("refine-using-similar-nodegroups", false, "Refine empty nodegoups capacities evaluations using real-world nodes from similar nodegroups when available.")
```
MaciekPytel, line 181: I'm not sure how helpful that description is to end user who's not already very familiar with CA internals. What this flag does is pretty hard to explain without going into details, but maybe we can try to focus on how it impacts users?

cluster-autoscaler/core/utils/utils.go
bpineau: Thanks, an annotation is much cleaned indeed, just did that!

cluster-autoscaler/processors/nodeinfos/refine_node_infos_processor.go
bpineau: That's a very good call. Changed for a conservative approach: start from (clone) the original/template NodeInfo, and only pick the node from the model/real world NodeInfo (keeping the original node's name, uid, hostname, locality labels)

cluster-autoscaler/main.go
bpineau: Updated following those guidelines (still not perfect, but I tried to remain concise)

bpineau:
Thanks for the review @MaciekPytel ! Updated the PR accordingly, please take an other look

MaciekPytel:
@bpineau I suspect the motivation for https://github.com/kubernetes/autoscaler/pull/4191 is to replace this with a version based on a new interface. Is that correct? Or do you still want to include this one?

bpineau:
@MaciekPytel you guessed right; I think we can close this one

PR #3763, opened 2020-12-14
Ephemeral Storage support for BuildNodeFromTemplate in GCE

jayantjain93:
This change will allow us to scale from zero nodepools for pods requiring ephemeral storage

### DIFF
--- a/cluster-autoscaler/simulator/delegating_shared_lister.go
+++ b/cluster-autoscaler/simulator/delegating_shared_lister.go
@@ -18,6 +18,7 @@ package simulator
 
 import (
 	"fmt"
+
 	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
 )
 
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/simulator/delegating_shared_lister.go
```diff
@@ -18,6 +18,7 @@ package simulator
 
 import (
 	"fmt"
+
```
MaciekPytel, line 21: This is obviously trivial so I'm ok with it, but in future I think it's better to avoid mixing changes in cloudprovider/ with changes in any other parts of CA. The rationale is that cloudprovider and the rest of CA usually have a different set of owners and the changes often need to be reviewed by a different set of people.
Unless the changes are strictly connected it's better to split them into 2 PRs that can be reviewed by owners of the given piece of code.

jayantjain93: Sure. Thanks for the note.

PR #3768, opened 2020-12-15
Make arch-specific releases use separate images instead of tags on the same image

towca:
This seems to be the current convention in k8s.

### DIFF
--- a/cluster-autoscaler/Makefile
+++ b/cluster-autoscaler/Makefile
@@ -28,6 +28,9 @@ ifdef DOCKER_RM
 else
   RM_FLAG=
 endif
+IMAGE=$(REGISTRY)/cluster-autoscaler$(PROVIDER)
+
+export DOCKER_CLI_EXPERIMENTAL := enabled
 
 build: build-arch-$(GOARCH)
 
@@ -53,11 +56,11 @@ make-image: make-image-arch-$(GOARCH)
 make-image-arch-%:
 ifdef BASEIMAGE
 	docker build --pull --build-arg BASEIMAGE=${BASEIMAGE} \
-		-t ${REGISTRY}/cluster-autoscaler${PROVIDER}:${TAG}-$* \
+		-t ${IMAGE}-$*:${TAG} \
 		-f Dockerfile.$* .
 else
 	docker build --pull \
-		-t ${REGISTRY}/cluster-autoscaler${PROVIDER}:${TAG}-$* \
+		-t ${IMAGE}-$*:${TAG} \
 		-f Dockerfile.$* .
 endif
 	@echo "Image ${TAG}${FOR_PROVIDER}-$* completed"
@@ -65,12 +68,12 @@ endif
 push-image: push-image-arch-$(GOARCH)
 
 push-image-arch-%:
-	./push_image.sh ${REGISTRY}/cluster-autoscaler${PROVIDER}:${TAG}-$*
+	./push_image.sh ${IMAGE}-$*:${TAG}
 
 push-manifest:
-	DOCKER_CLI_EXPERIMENTAL=enabled docker manifest create ${REGISTRY}/cluster-autoscaler${PROVIDER}:${TAG} \
-		$(addprefix $(REGISTRY)/cluster-autoscaler$(PROVIDER):$(TAG)-, $(ALL_ARCH))
-	DOCKER_CLI_EXPERIMENTAL=enabled docker manifest push --purge ${REGISTRY}/cluster-autoscaler${PROVIDER}:${TAG}
+	docker manifest create ${IMAGE}:${TAG} \
+	    $(addprefix $(REGISTRY)/cluster-autoscaler$(PROVIDER)-, $(addsuffix :$(TAG), $(ALL_ARCH)))
+	docker manifest push --purge ${IMAGE}:${TAG}
 
 execute-release: $(addprefix make-image-arch-,$(ALL_ARCH)) $(addprefix push-image-arch-,$(ALL_ARCH)) push-manifest
 	@echo "Release ${TAG}${FOR_PROVIDER} completed"


towca:
@hakman, you introduced arch-specific builds to the Makefile. Could you please take a look?
/assign @hakman

hakman:
@towca In general this is true, but not sure if there is support support for pushing manifests using cross repo references. If you really want the `-amd64` and `-arm64` images pushed to the corresponding repos, may be safer to retag with the corresponding suffixed images.

If you look at what is actually pushed to the prod repo from staging, you will see that there are no tags for the `-amd64` and `-arm64` images, just the main one, so cleanup is done by the promoter.
https://console.cloud.google.com/gcr/images/k8s-artifacts-prod/EU/autoscaling/cluster-autoscaler

towca:
The process of promoting images from staging to prod is completely manual. We haven't promoted any arch-specific `cluster-autoscaler` images yet. I want to promote an arm64 image right now, hence the changes. The promotion shouldn't be an issue, the same thing has already been done for Vertical Pod Autoscaler: https://github.com/kubernetes/k8s.io/pull/1318.

cluster-autoscaler/Makefile
```diff
@@ -53,24 +56,24 @@ make-image: make-image-arch-$(GOARCH)
...
-	./push_image.sh ${REGISTRY}/cluster-autoscaler${PROVIDER}:${TAG}-$*
+	./push_image.sh ${IMAGE}-$*:${TAG}
 
 push-manifest:
-	DOCKER_CLI_EXPERIMENTAL=enabled docker manifest create ${REGISTRY}/cluster-autoscaler${PROVIDER}:${TAG} \
-		$(addprefix $(REGISTRY)/cluster-autoscaler$(PROVIDER):$(TAG)-, $(ALL_ARCH))
-	DOCKER_CLI_EXPERIMENTAL=enabled docker manifest push --purge ${REGISTRY}/cluster-autoscaler${PROVIDER}:${TAG}
+	docker manifest create --amend ${IMAGE}:${TAG} $(shell echo $(ALL_ARCH) | sed -e "s~[^ ]*~$(IMAGE)\-&:$(TAG)~g")
+	@for arch in $(ALL_ARCH); do docker manifest annotate --arch $${arch} ${IMAGE}:${TAG} ${IMAGE}-$${arch}:${TAG}; done
+	docker manifest push --purge ${IMAGE}:${TAG}
```
hakman, line 76: There is no need to annotate the image with the arch.
```suggestion
	docker manifest create ${REGISTRY}/cluster-autoscaler${PROVIDER}:${TAG} \
		$(addprefix $(REGISTRY)/cluster-autoscaler$(PROVIDER)-, $(addsuffix :$(TAG), $(ALL_ARCH)))
	docker manifest push --purge ${REGISTRY}/cluster-autoscaler${PROVIDER}:${TAG}
```

cluster-autoscaler/Makefile
```diff
@@ -53,24 +56,24 @@ make-image: make-image-arch-$(GOARCH)
 make-image-arch-%:
 ifdef BASEIMAGE
 	docker build --pull --build-arg BASEIMAGE=${BASEIMAGE} \
-		-t ${REGISTRY}/cluster-autoscaler${PROVIDER}:${TAG}-$* \
+		-t ${IMAGE}-$*:${TAG} \
 		-f Dockerfile.$* .
 else
 	docker build --pull \
-		-t ${REGISTRY}/cluster-autoscaler${PROVIDER}:${TAG}-$* \
+		-t ${IMAGE}-$*:${TAG} \
```
hakman, line 63:
```suggestion
		-t ${REGISTRY}/cluster-autoscaler${PROVIDER}-$*:${TAG} \
```

cluster-autoscaler/Makefile
```diff
@@ -53,24 +56,24 @@ make-image: make-image-arch-$(GOARCH)
 make-image-arch-%:
 ifdef BASEIMAGE
 	docker build --pull --build-arg BASEIMAGE=${BASEIMAGE} \
-		-t ${REGISTRY}/cluster-autoscaler${PROVIDER}:${TAG}-$* \
+		-t ${IMAGE}-$*:${TAG} \
```
hakman, line 59:
```suggestion
		-t ${REGISTRY}/cluster-autoscaler${PROVIDER}-$*:${TAG} \
```

cluster-autoscaler/Makefile
```diff
@@ -28,6 +28,9 @@ ifdef DOCKER_RM
 else
   RM_FLAG=
 endif
+IMAGE=$(REGISTRY)/cluster-autoscaler$(PROVIDER)
+
```
hakman, line 32:
```suggestion
```

hakman (changes requested):
Simplified things a bit. My concern was that the push to staging won't work.
Also, if you want to push the ARM64 image in https://github.com/kubernetes/k8s.io/pull/1485, you should also push the AMD64 one.

cluster-autoscaler/Makefile
towca: Done.

towca:
I've already tested pushing to staging, so that shouldn't be a concern (gcr.io/k8s-staging-autoscaling/cluster-autoscaler-arm64:v1.20.0).

I'm building and promoting the arm64 image for a previous release specifically because somebody asked for it during the SIG meeting. All next releases will have all 3 images released. But sure, I can include the amd64 image as well.

cluster-autoscaler/Makefile
hakman: Seemed easier to read for me like without using the variable and doesn't seem to be that much more room for mistakes in this case. In any case, it's just a preference.

hakman:
Thanks. I tested this also and worked fine for me.
/lgtm

cluster-autoscaler/Makefile
towca: Ack

towca:
Thanks for the review!
/assign @MaciekPytel

hakman:
Thanks for the releases @towca :)

mwielgus (approved):
/lgtm
/approve

PR #2970, opened 2020-03-24
Allow out-of-cluster operation of VPA

JensErat:
Because of how we operate our Kubernetes clusters, we need to run vertical pod autoscaler with an external kubeconfig. Also aligned and factored out redundant configuration code.

PR #3769, opened 2020-12-16
Allow out-of-cluster operation of VPA

JensErat:
Because of how we operate our Kubernetes clusters, we need to run vertical pod autoscaler with an external kubeconfig. Also aligned and factored out redundant configuration code.

This is follow-up for now closed #2970. Regarding the open review comments:

- `Paths` -> `Path` is fixed for both occurrences
- Updated the copyright year
- I don't think the `var limitRangeCalculator limitrange.LimitRangeCalculator` line (now [vertical-pod-autoscaler/pkg/updater/main.go#86](https://github.com/kubernetes/autoscaler/pull/3769/files#diff-a64cbe814d071ca8df41b2890583bd9e92c3fb39fbbe57e17f013e434b2d108aR86)) can be removed, since it enforces the interface type. By removing this line, assigning `limitrange.NewNoopLimitsCalculator()` in line 90 would fail.

<sup>Jens Erat <jens.erat@daimler.com>, Daimler TSS GmbH, [imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sup>


JensErat:
/cc bskiba

Thank you for your review some time ago, and sorry for the long delay. I had to recreate the PR, as `/reopen` was not possible because I force-pushed after rebasing against master after the rather long time.

bskiba:
/lgtm
/approve

Apologies for the delay.

PR #3778, opened 2020-12-21
add DaemonSet eviction option for empty nodes

yaroslava-serdiuk:
We consider a node to be empty if it doesn’t run any regular pod, i.e. node runs only Mirror and DaemonSet pods.

Cluster Autoscaler checks for empty nodes in the cluster and deletes them if they exist. The process deletes a chunk of empty nodes concurrently without DaemonSet pods eviction (meaning those pods have no chance to cleanly exit)

This PR is aimed to add DaemonSet eviction functionality for empty nodes.

More info in the discussion: https://github.com/kubernetes/autoscaler/issues/1578

### DIFF
--- a/cluster-autoscaler/core/scale_down.go
+++ b/cluster-autoscaler/core/scale_down.go
@@ -1065,7 +1065,7 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
 			return deletedNodes, errors.ToAutoscalerError(errors.ApiCallError, taintErr)
 		}
 		deletedNodes = append(deletedNodes, node)
-		go func(nodeToDelete *apiv1.Node, nodeGroupForDeletedNode cloudprovider.NodeGroup) {
+		go func(nodeToDelete *apiv1.Node, nodeGroupForDeletedNode cloudprovider.NodeGroup, isEvictDaemonSetPods bool) {
 			sd.nodeDeletionTracker.StartDeletion(nodeGroupForDeletedNode.Id())
 			defer sd.nodeDeletionTracker.EndDeletion(nodeGroupForDeletedNode.Id())
 			var result status.NodeDeleteResult
@@ -1081,7 +1081,12 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
 					sd.context.LogRecorder.Eventf(apiv1.EventTypeNormal, "ScaleDownEmpty", "Scale-down: empty node %s removed", nodeToDelete.Name)
 				}
 			}()
-
+			if isEvictDaemonSetPods {
+				err := evictDaemonSetPods(sd.context.ClusterSnapshot, nodeToDelete, client, 10*time.Second, recorder, sd.context.MaxGracefulTerminationSec)
+				if err != nil {
+					klog.Warningf(err.Error())
+				}
+			}
 			deleteErr = waitForDelayDeletion(nodeToDelete, sd.context.ListerRegistry.AllNodeLister(), sd.context.AutoscalingOptions.NodeDeletionDelayTimeout)
 			if deleteErr != nil {
 				klog.Errorf("Problem with empty node deletion: %v", deleteErr)
@@ -1101,9 +1106,50 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
 				metrics.RegisterScaleDown(1, gpu.GetGpuTypeForMetrics(sd.context.CloudProvider.GPULabel(), sd.context.CloudProvider.GetAvailableGPUTypes(), nodeToDelete, nodeGroupForDeletedNode), metrics.Unready)
 			}
 			result = status.NodeDeleteResult{ResultType: status.NodeDeleteOk}
-		}(node, nodeGroup)
+		}(node, nodeGroup, sd.context.DaemonSetEvictionForEmptyNodes)
 	}
 	return deletedNodes, nil
+}
+
+// create eviction object for all DaemonSet pods on the node
+func evictDaemonSetPods(clusterSnapshot simulator.ClusterSnapshot, nodeToDelete *apiv1.Node, client kube_client.Interface, dsEvictionTimeout time.Duration,
+	recorder kube_record.EventRecorder, maxGracefulTerminationSec int) error {
+	nodeInfo, err := clusterSnapshot.NodeInfos().Get(nodeToDelete.Name)
+	if err != nil {
+		return fmt.Errorf("Failed to get node info for %s", nodeToDelete.Name)
+	}
+	_, daemonSetPods, _, err := simulator.FastGetPodsToMove(nodeInfo, true, true, []*policyv1.PodDisruptionBudget{}, time.Now())
+	if err != nil {
+		return fmt.Errorf("Failed to get DaemonSet pods for %s", nodeToDelete.Name)
+	}
+
+	dsEviction := make(chan status.PodEvictionResult, len(daemonSetPods))
+
+	// Perform eviction of DaemonSet pods. We don't want to raise an error if daemonsetPod wasn't evict properly
+	waitForDsEviction := time.Now().Add(dsEvictionTimeout)
+	for _, daemonSetPod := range daemonSetPods {
+		go func(podToEvict *apiv1.Pod) {
+			dsEviction <- evictPod(podToEvict, true, client, recorder, maxGracefulTerminationSec, waitForDsEviction, 5*time.Second)
+		}(daemonSetPod)
+	}
+	// wait for creating eviction of DaemonSet pods
+	allCreatedEviction := true
+	var failedPodName string
+	for range daemonSetPods {
+		select {
+		case status := <-dsEviction:
+			if status.Err != nil {
+				allCreatedEviction = false
+				failedPodName = status.Pod.Name
+			}
+		case <-time.After(dsEvictionTimeout + 5*time.Second):
+			return fmt.Errorf("Failed to create DaemonSet eviction for %v seconds on the %s", dsEvictionTimeout, nodeToDelete.Name)
+		}
+	}
+	if !allCreatedEviction {
+		return fmt.Errorf("%s failed to created eviction on the %s", failedPodName, nodeToDelete.Name)
+	}
+	return nil
 }
 
 func (sd *ScaleDown) deleteNode(node *apiv1.Node, pods []*apiv1.Pod, daemonSetPods []*apiv1.Pod,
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/core/scale_down.go
```diff
@@ -1065,7 +1065,34 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
...
+				for range daemonSetPods {
+					select {
+					case <-dsEviction:
+						counter++
+					case <-time.After(waitForDsEviction.Sub(time.Now()) + 5*time.Second):
+						klog.V(1).Infof("Failed to create DaemonSet eviction for 10 seconds seconds")
+					}
+				}
+			}
+
```
mwielgus, line 1095: Why are you starting the deletion tracking **AFTER** the deamon set is handled?

mwielgus (changes requested):

cluster-autoscaler/core/scale_down.go
```diff
@@ -1065,7 +1065,34 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
 			return deletedNodes, errors.ToAutoscalerError(errors.ApiCallError, taintErr)
 		}
 		deletedNodes = append(deletedNodes, node)
-		go func(nodeToDelete *apiv1.Node, nodeGroupForDeletedNode cloudprovider.NodeGroup) {
+		go func(nodeToDelete *apiv1.Node, nodeGroupForDeletedNode cloudprovider.NodeGroup, daemonSetEvictionForEmptyNodes bool) {
```
towca, line 1068: Same for the name of the parameter here.

cluster-autoscaler/core/scale_down.go
```diff
@@ -1065,7 +1065,34 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
...
+
+				// wait for creating eviction of DaemonSet pods
+				waitForDsEviction := time.Now().Add(10 * time.Second)
+				counter := 0
+				for range daemonSetPods {
+					select {
+					case <-dsEviction:
+						counter++
+					case <-time.After(waitForDsEviction.Sub(time.Now()) + 5*time.Second):
+						klog.V(1).Infof("Failed to create DaemonSet eviction for 10 seconds seconds")
```
towca, line 1091: seconds x2

cluster-autoscaler/core/scale_down.go
```diff
@@ -1065,7 +1065,34 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
...
+				}
+
+				// wait for creating eviction of DaemonSet pods
+				waitForDsEviction := time.Now().Add(10 * time.Second)
+				counter := 0
+				for range daemonSetPods {
+					select {
+					case <-dsEviction:
+						counter++
+					case <-time.After(waitForDsEviction.Sub(time.Now()) + 5*time.Second):
```
towca, line 1090: You're specifying a deadline (`time.Now().Add(10 * time.Second)`), and then substracting `time.Now()` from that deadline to get a timeout value - seems redundant. Also, this value seems repeated from `retryUntil`.

WDYT about defining `dsEvictionTimeout := 10*time.Second` and `retryUntil := time.Now().Add(dsEvictionTimeout)` at the top, and passing `dsEvictionTimeout + 5*time.Second` to `time.After()`?

cluster-autoscaler/core/scale_down.go
towca: +1

cluster-autoscaler/core/scale_down.go
```diff
@@ -1065,7 +1065,34 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
 			return deletedNodes, errors.ToAutoscalerError(errors.ApiCallError, taintErr)
 		}
 		deletedNodes = append(deletedNodes, node)
-		go func(nodeToDelete *apiv1.Node, nodeGroupForDeletedNode cloudprovider.NodeGroup) {
+		go func(nodeToDelete *apiv1.Node, nodeGroupForDeletedNode cloudprovider.NodeGroup, daemonSetEvictionForEmptyNodes bool) {
+			if daemonSetEvictionForEmptyNodes {
+				nodeInfo, _ := sd.context.ClusterSnapshot.NodeInfos().Get(nodeToDelete.Name)
```
towca, line 1070: WDYT about extracting this logic to a separate function/method? This would allow us to easily test it, leaving only minimal changes here - 1 line if.

towca (changes requested):

cluster-autoscaler/core/scale_down.go
yaroslava-serdiuk: Do you suggest to add a new method to scale down object?

umialpha:
Can somebody clarify why "evict DaemonSet" is needed?

cluster-autoscaler/core/scale_down.go
```diff
@@ -876,7 +876,7 @@ func (sd *ScaleDown) TryToScaleDown(
 	emptyNodes := sd.getEmptyNodes(candidateNames, sd.context.MaxEmptyBulkDelete, scaleDownResourcesLeft)
 	if len(emptyNodes) > 0 {
 		nodeDeletionStart := time.Now()
-		deletedNodes, err := sd.scheduleDeleteEmptyNodes(emptyNodes, sd.context.ClientSet, sd.context.Recorder, readinessMap, candidateNodeGroups)
+		deletedNodes, err := sd.scheduleDeleteEmptyNodes(emptyNodes, sd.context.DaemonSetEvictionForEmptyNodes, sd.context.ClientSet, sd.context.Recorder, readinessMap, candidateNodeGroups)
```
umialpha, line 879: I would suggest to pass`sd.context` instead of `sd.context.DaemonSetEvictionForEmptyNodes` for latter features.

cluster-autoscaler/core/scale_down.go
umialpha: +1

cluster-autoscaler/core/scale_down.go
```diff
@@ -1101,11 +1103,38 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
...
 			}
 			result = status.NodeDeleteResult{ResultType: status.NodeDeleteOk}
-		}(node, nodeGroup)
+		}(node, nodeGroup, sd.context.DaemonSetEvictionForEmptyNodes)
 	}
 	return deletedNodes, nil
 }
 
+// create eviction object for all DaemonSet pods on the node
+func daemonSetEviction(clusterSnapshot simulator.ClusterSnapshot, nodeToDelete *apiv1.Node, client kube_client.Interface, dsEvictionTimeout time.Duration,
```
towca, line 1112: Function names usually work better as verbs - WDYT about `evictDaemonSetPods`?

cluster-autoscaler/core/scale_down.go
```diff
@@ -1101,11 +1103,38 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
...
+	recorder kube_record.EventRecorder, maxGracefulTerminationSec int) {
+	nodeInfo, _ := clusterSnapshot.NodeInfos().Get(nodeToDelete.Name)
```
towca, line 1114: I don't think we should ignore the returned error here or in the next line. I'd return an error from this function, check for it in `scheduleDeleteEmptyNodes` and just log a warning/error if anything's wrong.

cluster-autoscaler/core/scale_down.go
```diff
@@ -1101,11 +1103,38 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
...
+		go func(podToEvict *apiv1.Pod) {
+			dsEviction <- evictPod(podToEvict, true, client, recorder, maxGracefulTerminationSec, waitForDsEviction, 5*time.Second)
+		}(daemonSetPod)
+	}
+	// wait for creating eviction of DaemonSet pods
+	for range daemonSetPods {
+		select {
+		case <-dsEviction:
+		case <-time.After(dsEvictionTimeout + 5*time.Second):
+			klog.V(1).Infof("Failed to create DaemonSet eviction for %v seconds on the %v", dsEvictionTimeout, nodeToDelete.Name)
```
towca, line 1131: Same here, I'd return this as an error.

cluster-autoscaler/core/scale_down.go
```diff
@@ -1101,11 +1103,38 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
...
+			return
+		}
+	}
+	klog.V(1).Infof("all DaemonSet pods created eviction on the %v", nodeToDelete.Name)
```
towca, line 1135: I'd remove this line - we'll log something if something goes wrong anyway, so this line doesn't carry much additional information. It does, however, have potential to be spammy, if a lot of empty nodes are deleted in parallel.

cluster-autoscaler/core/scale_down_test.go
```diff
@@ -1080,6 +1080,58 @@ var defaultScaleDownOptions = config.AutoscalingOptions{
 	MaxMemoryTotal:                   config.DefaultMaxClusterMemory * units.GiB,
 }
 
+func TestDaemonSetEvictionForEmptyNodes(t *testing.T) {
+	options := config.AutoscalingOptions{
```
towca, line 1084: WDYT about making this test table-driven and adding a case e.g. for when creating evictions times out?

towca (changes requested):

towca (changes requested):

umialpha:
Hi, I found [Implement Graceful Node Shutdown in Kubelet ](https://github.com/kubernetes/kubernetes/pull/96129). Is it related to the problem we are trying to solve?

yaroslava-serdiuk:
Hi @umialpha, I attached the discussion about DaemonSet eviction for empty nodes in the description

cluster-autoscaler/core/scale_down_test.go
```diff
@@ -1080,6 +1080,119 @@ var defaultScaleDownOptions = config.AutoscalingOptions{
...
+
+			if scenario.nodeInfoSuccess {
+				simulator.InitializeClusterSnapshotOrDie(t, context.ClusterSnapshot, []*apiv1.Node{n1}, dsPods)
+			} else {
+				simulator.InitializeClusterSnapshotOrDie(t, context.ClusterSnapshot, []*apiv1.Node{}, []*apiv1.Pod{})
+			}
+
+			err = evictDaemonSetPods(context.ClusterSnapshot, n1, fakeClient, dsWaitingTime, kube_util.CreateEventRecorder(fakeClient), 3)
+			if err != nil {
+				assert.Contains(t, err.Error(), scenario.err.Error())
```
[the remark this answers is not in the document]
yaroslava-serdiuk: Fair point, thanks

PR #3782, opened 2020-12-23
chart: fix reference to additionalLabels in deployment template

mgarstecki:
Because of incorrect casing (.values instead of .Values), the Deployment template fails when the additionalLabels block is defined.


gjtempleton:
Thanks for the PR!

/lgtm
/approve

PR #3783, opened 2020-12-24
adding support to block ephemeral storage from boot disk

jayantjain93:
This will allow ephemeral storage from boot disk to be skipped and not picked when BuildNodeFromTemplate is used to create a node.

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/templates.go
+++ b/cluster-autoscaler/cloudprovider/gce/templates.go
@@ -165,10 +165,13 @@ func (t *GceTemplateBuilder) BuildNodeFromTemplate(mig Mig, template *gce.Instan
 		return nil, fmt.Errorf("could not obtain os-distribution from kube-env from template metadata")
 	}
 
-	ephemeralStorage, err := getEphemeralStorageFromInstanceTemplateProperties(template.Properties)
-	if err != nil {
-		klog.Errorf("could not fetch ephemeral storage from instance template. %s", err)
-		return nil, err
+	var ephemeralStorage int64 = -1
+	if !isEphemeralStorageWithInstanceTemplateDisabled(kubeEnvValue) {
+		ephemeralStorage, err = getEphemeralStorageFromInstanceTemplateProperties(template.Properties)
+		if err != nil {
+			klog.Errorf("could not fetch ephemeral storage from instance template. %s", err)
+			return nil, err
+		}
 	}
 
 	capacity, err := t.BuildCapacity(cpu, mem, template.Properties.GuestAccelerators, os, osDistribution, ephemeralStorage, pods)
@@ -223,6 +226,17 @@ func (t *GceTemplateBuilder) BuildNodeFromTemplate(mig Mig, template *gce.Instan
 	// Ready status
 	node.Status.Conditions = cloudprovider.BuildReadyConditions()
 	return &node, nil
+}
+
+// isEphemeralStorageWithInstanceTemplateDisabled will allow bypassing Disk Size of Boot Disk from being
+// picked up from Instance Template and used as Ephemeral Storage, in case other type of storage are used
+// as ephemeral storage
+func isEphemeralStorageWithInstanceTemplateDisabled(kubeEnvValue string) bool {
+	v, found, err := extractAutoscalerVarFromKubeEnv(kubeEnvValue, "BLOCK_EPH_STORAGE_BOOT_DISK")
+	if err == nil && found && v == "true" {
+		return true
+	}
+	return false
 }
 
 func getEphemeralStorageFromInstanceTemplateProperties(instanceProperties *gce.InstanceProperties) (ephemeralStorage int64, err error) {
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/cloudprovider/gce/templates_test.go
```diff
@@ -322,6 +326,96 @@ func TestBuildAllocatableFromKubeEnv(t *testing.T) {
 		}
 	}
 }
+func TestBlockedBootDiskForEphemeralStorage(t *testing.T) {
```
MaciekPytel, line 329: This feels like it's almost a copy of TestBuildNodeFromTemplateSetsResources, which makes sense given that you're testing an condition checked in BuildNodeFromTemplate that is related to setting resources. At a glance it looks like maybe a better way to do it would be to add isEpemeralStorageBlocked and the related if statement to TestBuildNodeFromTemplateSetsResources and just add cases related to this logic there?

cluster-autoscaler/cloudprovider/gce/templates_test.go
```diff
@@ -322,6 +326,96 @@ func TestBuildAllocatableFromKubeEnv(t *testing.T) {
...
+	type testCase struct {
+		scenario                 string
+		kubeEnv                  string
+		mig                      Mig
+		physicalCpu              int64
+		physicalMemory           int64
+		physicalEphemeralStorage int64
+		expectedCpu              string
+		expectedMemory           string
+		expectedEphemeralStorage string
```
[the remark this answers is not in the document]
jayantjain93: Done

cluster-autoscaler/cloudprovider/gce/templates_test.go
jayantjain93: Done. Removed it.
I just have a concern, should we look to not add more cases to a single test func TestBuildNodeFromTemplateSetsResources, rather have smaller test functions that test specific functionalities.

cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -165,10 +165,13 @@ func (t *GceTemplateBuilder) BuildNodeFromTemplate(mig Mig, template *gce.Instan
 		return nil, fmt.Errorf("could not obtain os-distribution from kube-env from template metadata")
 	}
 
-	ephemeralStorage, err := getEphemeralStorageFromInstanceTemplateProperties(template.Properties)
-	if err != nil {
-		klog.Errorf("could not fetch ephemeral storage from instance template. %s", err)
-		return nil, err
+	var ephemeralStorage int64 = -1
```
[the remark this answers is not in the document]
jayantjain93: I have used -1 as a default value where ephemeralStorage is not used.
I feel it is a better "bad" value, since 0 could still an eligible value to be used, but -1 storage is just not possible.
Hence, keeping it consistent, I would like to keep it -1.

cluster-autoscaler/cloudprovider/gce/templates_test.go
```diff
@@ -137,13 +159,16 @@ func TestBuildNodeFromTemplateSetsResources(t *testing.T) {
 				assert.NotNil(t, node.Status)
 				assert.NotNil(t, node.Status.Capacity)
 				assert.NotNil(t, node.Status.Allocatable)
-				capacity, err := tb.BuildCapacity(tc.physicalCpu, tc.physicalMemory, tc.accelerators, OperatingSystemLinux, OperatingSystemDistributionCOS, -1, tc.pods)
+				if tc.isEphemeralStorageBlocked {
+					tc.physicalEphemeralStorage = 0
```
MaciekPytel, line 163: nit: I don't think it changes anything, but modifying test case definition from within test feels inelegant. I would use a local variable instead.

cluster-autoscaler/cloudprovider/gce/templates_test.go
```diff
@@ -98,6 +103,23 @@ func TestBuildNodeFromTemplateSetsResources(t *testing.T) {
 			kubeReserved:   false,
 			expectedErr:    false,
 		},
+		{
+			scenario: "BLOCK_EPH_STORAGE_BOOT_DISK in kube-env",
```
MaciekPytel, line 107: can we also add a case where BLOCK_EPH_STORAGE_BOOT_DISK is in kube-env, but with a value of 'false' or some other non-true value?

cluster-autoscaler/cloudprovider/gce/templates.go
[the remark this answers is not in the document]
MaciekPytel: Fair enough.

cluster-autoscaler/cloudprovider/gce/templates_test.go
MaciekPytel: In general I agree - we should aim to have multiple unittests covering separate pieces of code in isolation. However, in this specific case the test was almost an exact copy case of the other test and vast majority of logic executed by test was the same. I don't see the benefit of artificially splitting very similar tests and I think it makes it harder for future maintainers who will need to read ~the same thing twice and spend time figuring what the differences are.

cluster-autoscaler/cloudprovider/gce/templates_test.go
jayantjain93: Done

cluster-autoscaler/cloudprovider/gce/templates_test.go
jayantjain93: Done

PR #3790, opened 2020-12-30
cloudprovider: linode cloud provider

dbonfigli:
This PR add the linode cloud provider for the cluster autoscaler.

I tagged it as WIP because I would like to involve someone from linode to test it more deeply than I could.

### DIFF
--- a/cluster-autoscaler/cloudprovider/linode/README.md
+++ b/cluster-autoscaler/cloudprovider/linode/README.md
@@ -1,0 +1,62 @@
+# Cluster Autoscaler for Linode
+
+The cluster autoscaler for Linode scales nodes in a LKE cluster.
+
+## Linode Kubernetes Engine
+
+Linode Kubernetes Engine ([LKE](https://www.linode.com/docs/guides/deploy-and-manage-a-cluster-with-linode-kubernetes-engine-a-tutorial/)) is the managed K8s solution provided by Linode.
+
+LKE lets users create Node Pools, i.e. groups of nodes (also called Linodes) each of the same type.
+
+The size of a Node Pool can be configured at any moment. The user cannot select specific nodes to be deleted when downsizing a Node Pool, rather, Linode will randomly select nodes to be deleted to reach the defined size, even if a node is not healthy or has been manually deleted.
+
+Nodes in a Node Pool are considered disposable: they can be deleted and recreated at any moment, deleting a single node or using the *recycle* feature, on these cases the node will be recreated by Linode after a small amount of time.
+
+Node Pools do not support user defined tags or labels.
+
+There is no limitation on the number of Node Pool a LKE Cluster can have, limited to the maximum number of nodes an LKE Cluster can have.
+
+## Cluster Autoscaler
+
+A cluster autoscaler node group is composed of multiple LKE Node Pools of the same Linode type (e.g. g6-standard-1, g6-standard-2), each holding a *single* Linode.
+
+At every scan interval the cluster autoscaler reviews every LKE Node Pool and if:
+* it is not among the ones to be excluded as defined in the configuration file;
+* it holds a single Linode;
+
+then it becomes part of the node group holding LKE Node Pools of the Linode type it has, or a node group is created with this LKE Node Pool inside if there are no node groups with that Linode type at the moment.
+
+Scaling is achieved adding LKE Node Pools to node groups, *not* increasing the size of a LKE Node Pool, that must stay 1. The reason behind this is that Linode does not provide a way to selectively delete a Linode from a LKE Node Pool and decrease the size of the pool with it.
+
+This is also the reason we cannot use the standard `nodes` and `node-group-auto-discovery` cluster autoscaler flag (no labels could be used to select a specific node group), and the reason why there can be no node group of the same type.
+
+## Configuration
+
+The cluster autoscaler automatically select every LKE Node Pool that is part of a LKE cluster, so there is no need define the `node-group-auto-discovery` or `nodes` flags, see [examples/cluster-autoscaler-autodiscover.yaml](examples/cluster-autoscaler-autodiscover.yaml) for an example of a kubernetes deployment.
+
+It is mandatory to define the cloud configuration file `cloud-config`.
+You can see an example of the cloud config file at [examples/cluster-autoscaler-secret.yaml](examples/cluster-autoscaler-secret.yaml), it is an INI file with the following fields:
+
+| Key | Value | Mandatory | Default |
+|-----|-------|-----------|---------|
+| global/linode-token | Linode API Token with Read/Write permission for Kubernetes and Linodes | yes | none |
+| global/lke-cluster-id | ID of the LKE cluster (numeric of the form: 12345, you can get this via `linode-cli` or looking at the first number of a linode in a pool, e.g. for lke15989-19461-5fec9212fad2 the lke-cluster-id is "15989") | yes | none |
+| global/defaut-min-size-per-linode-type | minimum size of a node group (must be > 0) | no | 1 |
+| global/defaut-max-size-per-linode-type | maximum size of a node group | no | 254 |
+| global/do-not-import-pool-id | Pool id (numeric of the form: 12345) that will be excluded from the pools managed by the cluster autoscaler; can be repeated | no | none
+| nodegroup \"linode_type\"/min-size" | minimum size for a specific node group | no | global/defaut-min-size-per-linode-type |
+| nodegroup \"linode_type\"/max-size" | maximum size for a specific node group | no | global/defaut-min-size-per-linode-type |
+
+Log levels of intertest for the Linode provider are:
+* 1 (flag: ```--v=1```): basic logging at start;
+* 2 (flag: ```--v=2```): logging of the node group composition at every scan;
+
+## Development
+
+Make sure you are inside the `cluster-autoscaler` path of the [autoscaler repository](https://github.com/kubernetes/autoscaler).
+
+Create the docker image:
+```
+make container
+```
+tag the generated docker image and push it to a registry.
--- a/cluster-autoscaler/cloudprovider/linode/linode_api_client.go
+++ b/cluster-autoscaler/cloudprovider/linode/linode_api_client.go
@@ -1,0 +1,52 @@
+/*
+Copyright 2016 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package linode
+
+import (
+	"context"
+	"net/http"
+	"time"
+
+	"golang.org/x/oauth2"
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/linode/linodego"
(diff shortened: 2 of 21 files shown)


cluster-autoscaler/cloudprovider/linode/linode_cloud_config.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	max := defaultMax
+	if len(maxStr) != 0 {
+		max, err = strconv.Atoi(maxStr)
+		if err != nil {
+			return 0, 0, fmt.Errorf("could not parse max size for node group: %v", err)
+		}
+	}
+	if min >= max {
+		return 0, 0, fmt.Errorf("min size for a node group must be strictly less than its max size (got min: %d, max: %d)",
```
displague, line 160: Why can't min be equal to max?

cluster-autoscaler/go.sum
```diff
@@ -191,6 +191,8 @@ github.com/go-openapi/swag v0.19.5/go.mod h1:POnQmlKehdgb5mhVOsnJFsivZCEZ/vjK9gh
 github.com/go-openapi/validate v0.19.5/go.mod h1:8DJv2CVJQ6kGNpFW6eV9N3JviE1C85nY1c2z52x1Gk4=
 github.com/go-ozzo/ozzo-validation v3.5.0+incompatible h1:sUy/in/P6askYr16XJgTKq/0SZhiWsdg4WZGaLsGQkM=
 github.com/go-ozzo/ozzo-validation v3.5.0+incompatible/go.mod h1:gsEKFIVnabGBt6mXmxK0MoFy+cZoTJY6mu5Ll3LVLBU=
+github.com/go-resty/resty/v2 v2.1.1-0.20191201195748-d7b97669fe48 h1:JVrqSeQfdhYRFk24TvhTZWU0q8lfCojxZQFi3Ou7+uY=
```
displague, line 194: The packngo client had external depencies that were not already vetted, so the Packet provider implements a minimal REST client in autoscaler.  I would expect something similar in this case.  gopkg.in/resty.v1 is already in go.mod, but for Linode v2 would be needed.

cluster-autoscaler/cloudprovider/linode/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+A cluster autoscaler node group is composed of multiple LKE Node Pools of the same Linode type (e.g. g6-standard-1, g6-standard-2), each holding a *single* linode.
+
+At every scan interval the cluster autoscaler review every LKE Node Pool and if:
+* it is not among the the ones to be excluded as defined in the configuration file;
+* it holds a single linode;
+
+then it becomes part of the node group holding LKE Node Pools of the linode type it has, or a node group is created with this LKE Node Pool inside if there are no node groups with that linode type at the moment.
+
+Scaling is achieved adding LKE Node Pools to node groups, *not* increasing the size of a LKE Node Pool, that must stay 1. The reason behind this is that Linode does not provide a way to selectively delete a linode from a  LKE Node Pool and decrease the size of the pool with it.
```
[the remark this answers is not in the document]
displague: I appreciate the level of detail you're putting into this.

displague (commented):
This is a cursory pass over the provider, nothing authoritative. I'm familiar with the Linode APIs and can likely offer some level of review on those matters.

Static-Flow:
Hello everyone,

I am very much interested in this PR since I am planning on using Linode for a large distributed workload and wanted a clean way to autoscale!

I did have a question about the implementation. You mentioned that there isn't a way to remove specific nodes using the Node Pool API, which is true. My question is, since auto-scaler uses node allocation metrics to determine when to scale down, could you use that insight to get the id of the node that's underutilized and then call the regular Linode delete api to remove it followed by a call to the resize API to make the node pool count match?

I ran some quick tests on a cluster in my account and it seems like a feasible solution. I spun up 4 nodes in a cluster and assigned them to a service using all 4. Then, in the console, I hard deleted one of the nodes and downsized the cluster. Linode didn't try to spin up a new 4th one for me when I hard deleted it and my service continued working properly. 

There very well might be some edge cases I'm not thinking of that makes that solution non-universal or I might be over estimating what knowledge about the underutilized node auto-scaler can pass to us but I just wanted to toss the idea out there while it's still in WIP mode. Regardless, thank you for taking the time to build this PR and I look forward to using it! 😄

dbonfigli:
> My question is, since auto-scaler uses node allocation metrics to determine when to scale down, could you use that insight to get the id of the node that's underutilized and then call the regular Linode delete api to remove it followed by a call to the resize API to make the node pool count match?

> I ran some quick tests on a cluster in my account and it seems like a feasible solution. I spun up 4 nodes in a cluster and assigned them to a service using all 4. Then, in the console, I hard deleted one of the nodes and downsized the cluster. Linode didn't try to spin up a new 4th one for me when I hard deleted it and my service continued working properly.

Hi, I tried this approach at first by but it was not useful for what cluster-autoscaler needs to do: you are right that you can delete a single Linode in a LKE Cluster Pool, and that node will remain in the LKE Cluster Pool even if deleted, but when you shrink the size of the Pool, Linode API will take down a random node, not the one you have deleted. What Linode API lack at the momento is something like this for AWS: [TerminateInstanceInAutoScalingGroup](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_TerminateInstanceInAutoScalingGroup.html).

Static-Flow:
Hmm, when I hard deleted the node it was removed from the pool and k8, though it wasn't instantly. That   delay might cause some issues with how auto-scaler functions but might be possible to compensate for in the Linode custom code.   
  
  
  
    
  

  
  
>

cluster-autoscaler/go.sum
dbonfigli: I reviewed the PR for the Packet provider and now I know that external dependencies are not allowed anymore, I missed that information at first. I only use 3 API endpoints so I hope to substitute the linodego SDK with an internal implementation, I will work on it and update this PR.

dbonfigli:
> Hmm, when I hard deleted the node it was removed from the pool and k8, though it wasn't instantly. That delay might cause some issues with how auto-scaler functions but might be possible to compensate for in the Linode custom code.

Thank you for the new information, I will do more tests to verify if this approach is feasible.

cluster-autoscaler/cloudprovider/linode/linode_cloud_config.go
dbonfigli: I guessed that if min size == max size then it didn't make sense to use the cluster-autoscaler for this node group, since the size is fixed then, but yes, you are right, if the user want to configure it that way why not, I will change this.

dbonfigli:
> > Hmm, when I hard deleted the node it was removed from the pool and k8, though it wasn't instantly. That delay might cause some issues with how auto-scaler functions but might be possible to compensate for in the Linode custom code.
> 
Hi @Static-Flow, I took the time to do more tests when, after manually deleting a Linode in a LKE Node Pool, then on shrinking the size of this pool Linode would select the dead node. I did at least 20 tests: the Linode would delete the dead node most of the times, but not always, especially when you have more node pools. There is no particular logic in this, sometimes Linode would delete the "correct" dead node for even 6 or 7 times in a row, and then would delete an healthy node while there are deleted nodes around. This is consistent with what the Linode API documentation says, it would remove a random node. So I think that using LKE Node Pools of size 1 is the only implementation possible, at least for the moment. Feel free to do more testing and report back so that I can improve this PR, thank you.

Static-Flow:
Hey @dbonfigli could you share your testing suite? I created this and it seems to properly delete the specific node targeted and scales down its node pool without deleting further nodes. https://gist.github.com/Static-Flow/8c913415cdd2a5fe4e4720d06169d7fa

dbonfigli:
> Hey @dbonfigli could you possibly share your testing suite? I created this and it seems to properly delete the specific node targeted and scales down its node pool without deleting further nodes. https://gist.github.com/Static-Flow/8c913415cdd2a5fe4e4720d06169d7fa

Hi @Static-Flow, thank you for the time you put to clarify your concerns. I feel you misunderstood the issue: at step 6 of your test code you check the size of the down scaled LKE Pool and compare it to the expected size, that is not the real issue, and I agree with you that the size of a pool stays consistent, and that when a node is manually deleted and then the size of a pool is scaled down at a specific size, the size will stay consistent with the one we selected. The real issue is that Linode will select a random node when scaling down a pool, not the one we manually deleted, and for this reason you cannot use this approach (deleting a linode in a pool and after that downsizing the pool) on implementing this function:
```
// DeleteNodes deletes nodes from this node group (and also increasing the size
// of the node group with that). Error is returned either on failure or if the
// given node doesn't belong to this node group. This function should wait
// until node group size is updated. Implementation required.
func (n *NodeGroup) DeleteNodes(nodes []*apiv1.Node) error
```
that is a requirement for a new provider. 

To show you what I'm saying I modified your gist and replaced step 6 with what I feel is the real issue (and made several other changes to make the code compile): https://gist.github.com/dbonfigli/0ee7d064a580369aeeb94196bbecda56.

cluster-autoscaler/cloudprovider/linode/linode_cloud_config.go
dbonfigli: I've updated the PR allowing min == max.

cluster-autoscaler/go.sum
dbonfigli: I've updated the PR implementing a minimal rest client.

cluster-autoscaler/cloudprovider/linode/linode_api_client.go
```diff
@@ -0,0 +1,44 @@
...
+
+// buildLinodeAPIClient returns the struct ready to perform calls to linode API
+func buildLinodeAPIClient(linodeToken string) linodeAPIClient {
+	tokenSource := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: linodeToken})
+	oauth2Client := &http.Client{
+		Transport: &oauth2.Transport{
+			Source: tokenSource,
+		},
+	}
+	client := linodego.NewClient(oauth2Client)
```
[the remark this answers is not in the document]
dbonfigli: I've added user agent headers to the requests.

cluster-autoscaler/cloudprovider/linode/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Configuration
+
+The cluster autoscaler automatically select every LKE Node Pool that is part of a LKE cluster, so there is no need define the `node-group-auto-discovery`  or `nodes` flags, see [examples/cluster-autoscaler-autodiscover.yaml](examples/cluster-autoscaler-autodiscover.yaml) for an example of a kubernetes deployment.
```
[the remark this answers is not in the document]
dbonfigli: I've updated the README with more information on the reasons behind the choices. At the moment I feel using the LKE cluster tags is no better choice than the INI file, and we cannot use the tags in a node as you explained because nodes are disposable.

cluster-autoscaler/cloudprovider/linode/README.md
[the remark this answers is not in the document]
dbonfigli: I've updated the README with more information on the reasons behind the choices, hopefully is clearer now.

dbonfigli:
I removed the WIP tag, The PR is ready to be reviewed @towca @aleksandra-malinowska.

cluster-autoscaler/cloudprovider/linode/README.md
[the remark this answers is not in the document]
displague: Very clear and explicit, thanks.

mwielgus:
Please squash the commits to no more than 3.

dbonfigli:
> Please squash the commits to no more than 3.

done.

mwielgus (approved):
/lgtm
/approve

displague:
Congrats on the merge, @dbonfigli!

You might be interested in pursuing an issue and PR like #3399 and #3424 which adds Helm chart install instructions.

PR #3797, opened 2021-01-06
aws: Don't pile up successive full refreshes during AWS scaledowns

bpineau:
Force refreshing everything at every `DeleteNodes` calls causes slow down
and throttling on large clusters with lots of ASGs and activity.

That function might be called many times in a row during scale-down.
Each time the forced refresh will re-discover all ASGs, all LaunchConfigurations,
then re-list all instances from discovered ASGs.

That immediate refresh isn't required anyway, as the cache's DeleteInstances
concrete implementation will decrement the nodegroup size, and we can
schedule a grouped refresh for the next loop iteration.

As a later step, we can consider splitting the asgCache.generate() function
to support per ASG refreshes (and maybe per ASG caches TTLs + jitter, to
spread API calls). But that should address the current issue for now.

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider_test.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider_test.go
@@ -469,7 +469,7 @@ func TestDeleteNodes(t *testing.T) {
 	err = asgs[0].DeleteNodes([]*apiv1.Node{node})
 	assert.NoError(t, err)
 	service.AssertNumberOfCalls(t, "TerminateInstanceInAutoScalingGroup", 1)
-	service.AssertNumberOfCalls(t, "DescribeAutoScalingGroupsPages", 2)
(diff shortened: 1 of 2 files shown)


gjtempleton:
Thanks for the PR. 

My only concern is whether we could somehow make the log messages clearer, in particular: https://github.com/kubernetes/autoscaler/pull/3797/files#diff-22984a3a02b16ff49b2a94a43b49f3aa61c856483c3612b06b69dd51e347746fL269 Though it's already a bit misleading.

bpineau:
This might be clearer but still not great perhaps (or do you have a suggestion @gjtempleton )?:
```
DeleteInstances was called: scheduling an ASG list refresh for next accesses
```

gjtempleton:
@bpineau that reads far better to me, one minor nit, maybe?:

`DeleteInstances was called: scheduling an ASG list refresh for next main loop evaluation`

bpineau:
Thanks, that's better indeed! Updated accordingly.

PR #3744, opened 2020-12-04
Watch only pods in state Running and Unknown

jbartosik:
Stop watching pods in `Succeeded` and `Failed` states. I ran `full-vpa` e2e tests on this and they passed (including the OOM test).

I grabbed VPA logs about OOMs and one was discarded.  But it was because the OOM was too old, not because we removed the container (which was my worry about the change):

```
$ kubectl logs -n kube-system vpa-recommender-f7c794c46-swx8l|grep -i oom
I1204 14:57:45.110882       1 cluster_feeder.go:419] OOM detected {Timestamp:2020-12-04 14:57:35 +0000 UTC Memory:1073741824 ContainerID:{PodID:{Namespace:vertical-pod-autoscaling-7787 PodName:hamster-69d67ccbd4-lghrd} ContainerName:hamster}}
I1204 14:57:45.110956       1 cluster_feeder.go:419] OOM detected {Timestamp:2020-12-04 14:57:37 +0000 UTC Memory:1073741824 ContainerID:{PodID:{Namespace:vertical-pod-autoscaling-7787 PodName:hamster-69d67ccbd4-j6gsz} ContainerName:hamster}}
I1204 14:57:45.110981       1 cluster_feeder.go:419] OOM detected {Timestamp:2020-12-04 14:57:37 +0000 UTC Memory:1073741824 ContainerID:{PodID:{Namespace:vertical-pod-autoscaling-7787 PodName:hamster-69d67ccbd4-967k8} ContainerName:hamster}}
I1204 14:59:45.111255       1 cluster_feeder.go:419] OOM detected {Timestamp:2020-12-04 14:58:50 +0000 UTC Memory:1073741824 ContainerID:{PodID:{Namespace:vertical-pod-autoscaling-7787 PodName:hamster-69d67ccbd4-lghrd} ContainerName:hamster}}
I1204 14:59:45.111350       1 cluster_feeder.go:419] OOM detected {Timestamp:2020-12-04 14:59:35 +0000 UTC Memory:1555165137 ContainerID:{PodID:{Namespace:vertical-pod-autoscaling-7787 PodName:hamster-69d67ccbd4-lm6km} ContainerName:hamster}}
[...]

PR #3806, opened 2021-01-12
Revert 3744

jbartosik:
* Revert #3744 because I no longer think we need to make the change.
* Add a comment explaining why we filter pods the way we do so if someone in the future wants to change the behaviour they need to do less archeologi

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
@@ -172,6 +172,13 @@ func watchEvictionEvents(evictedEventChan <-chan watch.Event, observer oom.Obser
 
 // Creates clients watching pods: PodLister (listing only not terminated pods).
 func newPodClients(kubeClient kube_client.Interface, resourceEventHandler cache.ResourceEventHandler, namespace string) v1lister.PodLister {
+	// We are interested in pods which are Running or Unknown (in case the pod is
+	// running but there are some transient errors we don't want to delete it from
+	// our model).
+	// We don't want to watch Pending pods because they didn't generate any usage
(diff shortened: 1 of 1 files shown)


vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -175,11 +175,11 @@ func newPodClients(kubeClient kube_client.Interface, resourceEventHandler cache.
...
 	// running but there are some transient errors we don't want to delete it from
 	// our model).
-	// We don't want to watch Pending, Succeeded, or Failed failed pods because we
-	// know they're not generating any usage.
-	// Filter
-	statusFilter := fmt.Sprintf("status.phase!=%s,status.phase!=%s,status.phase!=%s", apiv1.PodPending, apiv1.PodSucceeded, apiv1.PodFailed)
-	selector := fields.ParseSelectorOrDie(statusFilter)
+	// We don't want to watch Pending pods because they didn't generate any usage
+	// yet.
+	// Succeeded and Failed failed pods don't generate any usage anymore but they
```
bskiba, line 180: Can you rephrase the comment to say what data we lose when we stopped listing those pods? 
It's also a bit worrying since I'm not sure if API server gives any guarantees on how long the failed and succeeded pods are kept for.

jbartosik: Rephrased. I didn't see any deleted data (that we wouldn't want to delete). I think we won't delete any. I just want to highlight here that one needs to be careful if they decide to change this

jbartosik:
@bskiba please take a look. I updated cherry pick  (#3809) too.

PR #3817, opened 2021-01-14
Implement round-robin for SchedulerBasedPredicateChecker

BigDarkClown:
Removal of randomness in 1.18 resulted in many pods using the same destination nodes during scaleDown. That in turn caused major regression in scaleDown velocity, decreasing the number of under-utilised nodes removal drastically. Former solution used shuffle to guarantee better variety of destination nodes, this implementation uses deterministic round-robin for the same purpose.

### DIFF
--- a/cluster-autoscaler/simulator/scheduler_based_predicates_checker.go
+++ b/cluster-autoscaler/simulator/scheduler_based_predicates_checker.go
@@ -41,6 +41,7 @@ type SchedulerBasedPredicateChecker struct {
 	delegatingSharedLister *DelegatingSchedulerSharedLister
 	nodeLister             v1listers.NodeLister
 	podLister              v1listers.PodLister
+	lastIndex              int
 }
 
 // NewSchedulerBasedPredicateChecker builds scheduler based PredicateChecker.
@@ -106,7 +107,8 @@ func (p *SchedulerBasedPredicateChecker) FitsAnyNodeMatching(clusterSnapshot Clu
 		return "", fmt.Errorf("error running pre filter plugins for pod %s; %s", pod.Name, preFilterStatus.Message())
 	}
 
-	for _, nodeInfo := range nodeInfosList {
+	for i := range nodeInfosList {
+		nodeInfo := nodeInfosList[(p.lastIndex+i)%len(nodeInfosList)]
 		if !nodeMatches(nodeInfo) {
 			continue
 		}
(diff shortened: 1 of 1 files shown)


mwielgus (commented):
How was it tested?

BigDarkClown:
The change was tested in e2e tests on 2k node cluster. The tests covered scale-up from 0 to 2k nodes, scale-down of empty nodes and scale-down of under-utilised nodes. There was no noticeable impact on the first two tests, while scale-down of under-utilised increased from ~6 nodes per hour to ~100 nodes per hour.

cluster-autoscaler/simulator/scheduler_based_predicates_checker.go
```diff
@@ -106,7 +107,8 @@ func (p *SchedulerBasedPredicateChecker) FitsAnyNodeMatching(clusterSnapshot Clu
 		return "", fmt.Errorf("error running pre filter plugins for pod %s; %s", pod.Name, preFilterStatus.Message())
 	}
 
-	for _, nodeInfo := range nodeInfosList {
+	for i := range nodeInfosList {
+		nodeInfo := nodeInfosList[(p.lastIndex+i)%len(nodeInfosList)]
```
mwielgus, line 111: I cannot see any random anymore. Where did it go :) ?

BigDarkClown: Good point, this approach is no longer based on the randomness, as it is replaced by deterministic round-robin. I am gonna change the commit name to better reflect that.

BigDarkClown: The purpose of both the previous approach and round-robin is to ensure, that in the consecutive function calls pods are assigned to destination nodes somewhat evenly, so the group of chosen destination nodes is as big as possible - this ensures scale down works better.

Round-robin is better in terms of time complexity and repetitive memory allocation than the previous approach - shuffling nodes in each function call takes additional time and memory, while keeping round-robin index is relatively cost free. 

Other than that both these approaches are expected to work similarly - the difference is that in case of randomisation we rely on expected outcome while in case of round-robin we have more control over it.

BigDarkClown: It's additionally worth to mention that I tested both proposed solutions, round-robin and shuffling of nodes, and found that the latter may have slightly negative impact on scale-up speed, as scale-up uses this function as well. Implementing the shuffling correctly would then require some additional code refactoring, which further confirmed my conviction to round-robin approach.

mwielgus (approved):
/lgtm
/approve

PR #3823, opened 2021-01-16
added cluster-name parameter for AWS provider

vglafirov:
Adding cluster-name parameter for AWS config in helm chart.


gjtempleton:
Hi, thanks for the PR!

What problem is this trying to solve and have you tested this on a test cluster, as I believe this won't actually change the behaviour, as the AWS cloud provider doesn't make use of the `ClusterName` option anywhere currently:
https://sourcegraph.com/github.com/kubernetes/autoscaler@a9cf777cfb591b1244eb94633bfa248d77c41557/-/blob/cluster-autoscaler/main.go#L235:3&tab=references

If I'm right, we should update the values.yaml as it does mean that the comment in the values file is misleading.

vglafirov:
> Hi, thanks for the PR!
> 
> What problem is this trying to solve and have you tested this on a test cluster? I believe this won't actually change the behaviour, as the AWS cloud provider doesn't make use of the `ClusterName` option anywhere currently:
> https://sourcegraph.com/github.com/kubernetes/autoscaler@a9cf777cfb591b1244eb94633bfa248d77c41557/-/blob/cluster-autoscaler/main.go#L235:3&tab=references
> 
> If I'm right, we should update the values.yaml as it does mean that the comment in the values file is misleading.

@gjtempleton I spend the whole day trying to figure this out :) Apparently, autoscaler does add nodes to ASG based just on tags, but they don't appear in EKS control plane without cluster name as a parameter.

gjtempleton:
Thanks for resolving the merge conflict however I'm still not sure what problem this is solving, as I can seem to see nodes in the AWS console for EKS clusters without this flag being set.

Could you give me some reproduction steps so I can see exactly what this change is aiming to achieve?

mwielgus:

PR #3824, opened 2021-01-18
add DaemonSet eviction option for empty nodes

yaroslava-serdiuk:
We consider a node to be empty if it doesn’t run any regular pod, i.e. node runs only Mirror and DaemonSet pods.

Cluster Autoscaler checks for empty nodes in the cluster and deletes them if they exist. The process deletes a chunk of empty nodes concurrently without DaemonSet pods eviction (meaning those pods have no chance to cleanly exit)

This PR is aimed to add DaemonSet eviction functionality for empty nodes.

More info in the discussion: #1578

### DIFF
--- a/cluster-autoscaler/core/scale_down.go
+++ b/cluster-autoscaler/core/scale_down.go
@@ -71,6 +71,10 @@ const (
 	// PodEvictionHeadroom is the extra time we wait to catch situations when the pod is ignoring SIGTERM and
 	// is killed with SIGKILL after MaxGracefulTerminationTime
 	PodEvictionHeadroom = 30 * time.Second
+	// DaemonSetEvictionEmptyNodeTimeout is the time to evict all DaemonSet pods on empty node
+	DaemonSetEvictionEmptyNodeTimeout = 10 * time.Second
+	// DeamonSetTimeBetweenEvictionRetries is a time between retries to create eviction that uses for DaemonSet eviction for empty nodes
+	DeamonSetTimeBetweenEvictionRetries = 3 * time.Second
 )
 
 // NodeDeletionTracker keeps track of node deletions.
@@ -1065,7 +1069,7 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
 			return deletedNodes, errors.ToAutoscalerError(errors.ApiCallError, taintErr)
 		}
 		deletedNodes = append(deletedNodes, node)
-		go func(nodeToDelete *apiv1.Node, nodeGroupForDeletedNode cloudprovider.NodeGroup) {
+		go func(nodeToDelete *apiv1.Node, nodeGroupForDeletedNode cloudprovider.NodeGroup, evictDaemonSetPodsBool bool) {
 			sd.nodeDeletionTracker.StartDeletion(nodeGroupForDeletedNode.Id())
 			defer sd.nodeDeletionTracker.EndDeletion(nodeGroupForDeletedNode.Id())
 			var result status.NodeDeleteResult
@@ -1081,7 +1085,11 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
 					sd.context.LogRecorder.Eventf(apiv1.EventTypeNormal, "ScaleDownEmpty", "Scale-down: empty node %s removed", nodeToDelete.Name)
 				}
 			}()
-
+			if evictDaemonSetPodsBool {
+				if err := evictDaemonSetPods(sd.context.ClusterSnapshot, nodeToDelete, client, sd.context.MaxGracefulTerminationSec, time.Now(), DaemonSetEvictionEmptyNodeTimeout, DeamonSetTimeBetweenEvictionRetries, recorder); err != nil {
+					klog.Warningf(err.Error())
+				}
+			}
 			deleteErr = waitForDelayDeletion(nodeToDelete, sd.context.ListerRegistry.AllNodeLister(), sd.context.AutoscalingOptions.NodeDeletionDelayTimeout)
 			if deleteErr != nil {
 				klog.Errorf("Problem with empty node deletion: %v", deleteErr)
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/core/scale_down.go
```diff
@@ -1081,7 +1081,13 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
 					sd.context.LogRecorder.Eventf(apiv1.EventTypeNormal, "ScaleDownEmpty", "Scale-down: empty node %s removed", nodeToDelete.Name)
 				}
 			}()
-
+			if evictDaemonSetPodsBool {
+				dsEvictionTimeout := 10 * time.Second
```
mwielgus, line 1085: Can you define a constant for these two?

cluster-autoscaler/core/scale_down.go
```diff
@@ -1101,11 +1107,51 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
...
+	// wait for creating eviction of DaemonSet pods
+	var failedPodNames []string
+	for range daemonSetPods {
+		select {
+		case status := <-dsEviction:
+			if status.Err != nil {
+				failedPodNames = append(failedPodNames, status.Pod.Name)
+			}
+		// adding waitBetweenRetries in order to have a bigger time interval than evictPod()
+		case <-time.After(dsEvictionTimeout + waitBetweenRetries):
```
mwielgus, line 1145: Why waitBetweenRetries is added here?

cluster-autoscaler/core/scale_down.go
```diff
@@ -1101,11 +1107,51 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
...
+			return fmt.Errorf("Failed to create DaemonSet eviction for %v seconds on the %s", dsEvictionTimeout, nodeToDelete.Name)
+		}
+	}
+	if len(failedPodNames) > 0 {
+		return fmt.Errorf("%s failed to created eviction on the %s", failedPodNames, nodeToDelete.Name)
```
mwielgus, line 1150: failed to evict %s on node %s.
Moreover it would be good to return or log a sample error.

mwielgus (changes requested):

cluster-autoscaler/core/scale_down.go
```diff
@@ -71,6 +71,10 @@ const (
 	// PodEvictionHeadroom is the extra time we wait to catch situations when the pod is ignoring SIGTERM and
 	// is killed with SIGKILL after MaxGracefulTerminationTime
 	PodEvictionHeadroom = 30 * time.Second
+	// DaemonSetEvictionEmptyNodeTimeout is the time to evict all DaemonSet pods on empty node
+	DaemonSetEvictionEmptyNodeTimeout = 10 * time.Second
+	// TimeBetweenRetries is a time between retries to create eviction that uses for DaemonSet eviction for empty nodes
+	TimeBetweenRetries = 3 * time.Second
```
mwielgus, line 77: DeamonSetTimeBetweenEvictionRetries

cluster-autoscaler/core/scale_down.go
```diff
@@ -1101,11 +1109,51 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
...
+// Create eviction object for all DaemonSet pods on the node
+func evictDaemonSetPods(clusterSnapshot simulator.ClusterSnapshot, nodeToDelete *apiv1.Node, client kube_client.Interface, maxGracefulTerminationSec int, timeNow time.Time, dsEvictionTimeout time.Duration, waitBetweenRetries time.Duration,
+	recorder kube_record.EventRecorder) error {
+	nodeInfo, err := clusterSnapshot.NodeInfos().Get(nodeToDelete.Name)
+	if err != nil {
+		return fmt.Errorf("failed to get node info for %s", nodeToDelete.Name)
+	}
+	_, daemonSetPods, _, err := simulator.FastGetPodsToMove(nodeInfo, true, true, []*policyv1.PodDisruptionBudget{}, time.Now())
+	if err != nil {
+		return fmt.Errorf("failed to get DaemonSet pods for %s", nodeToDelete.Name)
```
mwielgus, line 1126: Include the original error.

cluster-autoscaler/core/scale_down.go
```diff
@@ -1101,11 +1109,51 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
...
+				failedPodErrors = append(failedPodErrors, status.Err.Error())
+			}
+		// adding waitBetweenRetries in order to have a bigger time interval than evictPod()
+		case <-time.After(dsEvictionTimeout):
+			return fmt.Errorf("failed to create DaemonSet eviction for %v seconds on the %s", dsEvictionTimeout, nodeToDelete.Name)
+		}
+	}
+	if len(failedPodErrors) > 0 {
+
+		return fmt.Errorf("following DaemonSet pod failed to evict on the %s:\n%s", nodeToDelete.Name, fmt.Errorf(strings.Join(failedPodErrors, "\n")))
```
mwielgus, line 1152: You don't mention the pod names now, only names errors are passed.

mwielgus (changes requested):

cluster-autoscaler/core/scale_down.go
yaroslava-serdiuk: Pod names are included in the errors. Current errors have format: fmt.Errorf("failed to evict pod %s/%s within allowed timeout (last error: %v)", podToEvict.Namespace, podToEvict.Name, lastError)

mwielgus (approved):
/lgtm
/approve

cluster-autoscaler/core/scale_down.go
```diff
@@ -1065,7 +1069,7 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
 			return deletedNodes, errors.ToAutoscalerError(errors.ApiCallError, taintErr)
 		}
 		deletedNodes = append(deletedNodes, node)
-		go func(nodeToDelete *apiv1.Node, nodeGroupForDeletedNode cloudprovider.NodeGroup) {
+		go func(nodeToDelete *apiv1.Node, nodeGroupForDeletedNode cloudprovider.NodeGroup, evictDaemonSetPodsBool bool) {
```
towca, line 1072: You don't need to include bool in the name, since it's implied by the type.

cluster-autoscaler/core/scale_down.go
```diff
@@ -1081,7 +1085,11 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
 					sd.context.LogRecorder.Eventf(apiv1.EventTypeNormal, "ScaleDownEmpty", "Scale-down: empty node %s removed", nodeToDelete.Name)
 				}
 			}()
-
+			if evictDaemonSetPodsBool {
+				if err := evictDaemonSetPods(sd.context.ClusterSnapshot, nodeToDelete, client, sd.context.MaxGracefulTerminationSec, time.Now(), DaemonSetEvictionEmptyNodeTimeout, DeamonSetTimeBetweenEvictionRetries, recorder); err != nil {
+					klog.Warningf(err.Error())
```
towca, line 1090: I'd add a prefix here, e.g. `klog.Warningf("error while evicting DS pods from an empty node: %v", err)`

cluster-autoscaler/core/scale_down.go
```diff
@@ -1101,11 +1109,51 @@ func (sd *ScaleDown) scheduleDeleteEmptyNodes(emptyNodes []*apiv1.Node, client k
...
 }
 
+// Create eviction object for all DaemonSet pods on the node
+func evictDaemonSetPods(clusterSnapshot simulator.ClusterSnapshot, nodeToDelete *apiv1.Node, client kube_client.Interface, maxGracefulTerminationSec int, timeNow time.Time, dsEvictionTimeout time.Duration, waitBetweenRetries time.Duration,
+	recorder kube_record.EventRecorder) error {
+	nodeInfo, err := clusterSnapshot.NodeInfos().Get(nodeToDelete.Name)
+	if err != nil {
+		return fmt.Errorf("failed to get node info for %s", nodeToDelete.Name)
+	}
+	_, daemonSetPods, _, err := simulator.FastGetPodsToMove(nodeInfo, true, true, []*policyv1.PodDisruptionBudget{}, time.Now())
```
towca, line 1124: You should probably pass the timeNow variable here, right?

cluster-autoscaler/core/scale_down_test.go
```diff
@@ -1080,6 +1080,121 @@ var defaultScaleDownOptions = config.AutoscalingOptions{
...
+
+			if scenario.nodeInfoSuccess {
+				simulator.InitializeClusterSnapshotOrDie(t, context.ClusterSnapshot, []*apiv1.Node{n1}, dsPods)
+			} else {
+				simulator.InitializeClusterSnapshotOrDie(t, context.ClusterSnapshot, []*apiv1.Node{}, []*apiv1.Pod{})
+			}
+
+			err = evictDaemonSetPods(context.ClusterSnapshot, n1, fakeClient, options.MaxGracefulTerminationSec, timeNow, dsEvictionTimeout, waitBetweenRetries, kube_util.CreateEventRecorder(fakeClient))
+			if scenario.err != nil {
+				assert.Contains(t, err.Error(), scenario.err.Error())
```
towca, line 1185: You also have to assert that `err` is not nil before this assertion, otherwise this can still panic.

cluster-autoscaler/core/scale_down_test.go
```diff
@@ -1080,6 +1080,121 @@ var defaultScaleDownOptions = config.AutoscalingOptions{
...
+	for _, scenario := range testScenarios {
+		t.Run(scenario.name, func(t *testing.T) {
+			options := config.AutoscalingOptions{
+				ScaleDownUtilizationThreshold:  0.5,
+				ScaleDownUnneededTime:          time.Minute,
+				MaxGracefulTerminationSec:      1,
+				DaemonSetEvictionForEmptyNodes: true,
+			}
+			deletedPods := make(chan string, len(scenario.dsPods)+2)
+			dsEvictionTimeout := 100 * time.Millisecond
```
towca, line 1136: This should probably be set to a high value for all cases which don't test the timeout mechanism, otherwise it can flake, right?

cluster-autoscaler/core/scale_down.go
yaroslava-serdiuk: I already have evictDaemonSetPods function

cluster-autoscaler/core/scale_down_test.go
```diff
@@ -1080,6 +1080,121 @@ var defaultScaleDownOptions = config.AutoscalingOptions{
 	MaxMemoryTotal:                   config.DefaultMaxClusterMemory * units.GiB,
 }
 
+func TestDaemonSetEvictionForEmptyNodes(t *testing.T) {
+	timeNow := time.Now()
```
towca, line 1084: Could we replace this with a set time?

cluster-autoscaler/core/scale_down.go
towca: evictDsPods?

yaroslava-serdiuk: I find existed name more explicit, otherwise we will have quite similar name for bool value and function name

towca: Adding type information to variable names is generally discouraged in go, so I'd really like to see a different name here, but it's not super important.

PR #3828, opened 2021-01-19
Change VPA default version to 0.9.2 #3827

jbartosik:


jbartosik:
/close
I created against wrong branch, now it's stuck with needs-rebase tag

PR #3832, opened 2021-01-20
Fill in the LastUpdateTime Field of VpaCheckpoint Status with Correct Time.

PhdLoLi:
Currently, the "LastUpdateTime" field of the Status of VpaCheckpoint is not set, which is always "NULL" now.
Actually, we found that this filed is very helpful for us to judge whether the vpa Recommender component is working efficiently.
The logic is as below:

(1) If the Recommender component can completely update all vpas in one "RunOnce" cycle, the vpacheckpoints will also be updated in time. In other words, if there are so many vpas to update in one operating cycle (or the "metricsFetcherInterval" is set too short), which cause context timeout, then the update of vpacheckpoints may be delayed for a long time.

(2) If there is no change of any filed of vpa (target, lowerbound and etc), the vpa itself will not be updated. So we can not rely on the update of vpa to detect if the Recommender process vpas and vpacheckpoints efficiently.

(3) In our production environment, the situation described in (1) happened very often, we want to detect this situation and make adjustment accordingly as soon as possible.
And finally we found that the "LastUpdateTime" field of VpaCheckpoint Status can help solve such problem, which is not set yet Now.
We fill in this filed by ourself and use it for a while, and is not particularly convenient when we want to upgrade the code base. So we decide to make this pull request to merge this tiny but very helpful change.
Thank you very much.


vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state_test.go
```diff
@@ -141,6 +141,7 @@ func TestAggregateContainerStateSaveToCheckpoint(t *testing.T) {
 
 	assert.NoError(t, err)
 
+	assert.Equal(t, time.Now().Round(1*time.Second), checkpoint.LastUpdateTime.Time.Round(1*time.Second))
```
mwielgus, line 144: This is flaky, there may be 1 second delay in test execution. Use bigger window.

mwielgus (changes requested):

vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state_test.go
PhdLoLi: Thank you for your suggestion. I have expanded the time window to 10 seconds.

PhdLoLi: > This is flaky, there may be 1 second delay in test execution. Use bigger window.

Hi,  I am wondering if I should make any other changes to solve this review.  Or maybe this "changes requested" review can be approved? Looking forward to you and Thanks a lot .   @mwielgus

PhdLoLi:
Two months have past since I made the changes according to your advice. Can anyone help review this PR?  Thanks a lot. @mwielgus @krzysied @jbartosik

PR #3839, opened 2021-01-21
Allow to balance nodes using specified labels

### LINKED ISSUE #3615: Provide option to balance label based on the given labels
It would be nice to have a flag that could be passed to the autoscaler to chose the flags to compare the nodes if they are same.

This request would be the opposite of [https://github.com/kubernetes/autoscaler/pull/2493](url).

At the moment, in our project we are passing a lot of labels to this flag to be ignored. This can be simplified by passing in the opt-in labels vs exclude labels using a different flag say, **--balancing-node-labels**

rtnpro:
Fixes #3615


cluster-autoscaler/processors/nodegroupset/compare_nodegroups.go
```diff
@@ -160,5 +167,10 @@ func IsCloudProviderNodeInfoSimilar(n1, n2 *schedulerframework.NodeInfo, ignored
 		return false
 	}
 
+	fmt.Println(includedLabels, len(includedLabels))
+	if len(includedLabels) > 0 && !compareLabels(nodes, includedLabels) {
```
bpineau, line 171: Wasn't the #3615 intent to shortcut the full labels comparison (the `if !compareLabels(nodes, ignoredLabels)`  above) with a comparison over a smaller set of `includedLabels`?

rtnpro: You are correct @bpineau . I will fix the PR

cluster-autoscaler/processors/nodegroupset/compare_nodegroups.go
```diff
@@ -105,6 +106,19 @@ func CreateGenericNodeInfoComparator(extraIgnoredLabels []string) NodeInfoCompar
...
+		for _, l := range labels {
+			includedLabels[l] = true
+		}
+		if !compareLabels([]*schedulerframework.NodeInfo{n1, n2}, includedLabels, make(map[string]bool)) {
+			return false
+		}
+		return IsCloudProviderNodeInfoSimilar(n1, n2, make(map[string]bool))
+	}
+}
+
```
rtnpro, line 121: As suggested, I created a new `comparator` to accept `labels` instead of `extraIgnoredLabels`, and compare nodes based on the specified labels. This `--balancing-labels` arg here bypasses the `--balancing-ignored-labels` flag. This also ignores the cloud provider-specific basic ignored flags. This calls for some discussion.

Rather than creating a new `comparator`, I'd personally prefer to edit the signature of the current comparator creators to accept an extra argument for `labels` in addition to `extraIgnoredLabels`, and compare the nodes for matching labels specified in `labels` if `labels` is not empty. If check fails, then we return, else we proceed to call `IsCloudProviderNodeInfoSimilar`, which takes care of ignored labels.

@MaciekPytel please share your thoughts and review here.

MaciekPytel: I think the second implementation idea doesn't change current behavior at all. Currently every label that is not on ignored-labels list will cause the node groups to compare as different. Checking `--balancing-labels` first doesn't change the ultimate result, it just fails the comparison slightly earlier.

My understanding of #3615 is that specifying `--balancing-labels` makes comparison treat any label not on the list as ignored-label (current implementation treats any label that is not on ignored list as "balancing labels"). I think that:
* Specifying both `--balancing-labels` and `--balancing-ignored-labels` makes no sense. What would be the behavior for any label not on either list?
    * (if you agree with this logic) we should add validation in main.go that prevents setting both flags at the same time.
* Ignoring cloudprovider labels seems irrelevant when using `--balancing-labels` (since we would be ignoring all labels not on the list). 
* On the implementation side - calling IsCloudProviderNodeInfoSimilar as-is will not work as far as I can see, since it already has the label check embedded. I think the easiest way is to extract the resources check to a separate function and call that new function if `--balancing-labels are specified`, but there are plenty of other ways to refactor and I don't feel particularly strongly about which is the right one.

rtnpro: 👍 to

- Making `--balancing-labels` and `--balancing-ignored-labels` mutually exclusive
- Ignoring cloudprovider labels irrelevant when using `--balancing-labels`
- Yes, we'll need to refactor a underlying implementation of `IsCloudProviderNodeInfoSimilar` as you said. Since this is my first issue on CA, I was a bit hesitant to do much refactoring. Thanks for asserting my thoughts.

rtnpro:
@MaciekPytel I pushed some changes to this PR today, as per your last review comments: https://github.com/kubernetes/autoscaler/pull/3839/commits/fd5434f67d4abb829c2bad460220d7aa26046801

Need a review here. Thanks

rtnpro:
@MaciekPytel could I get a review here?

Issue #3836, opened 2021-01-21
[clusterapi] weird behavior when decresae workload 


jichenjc:
**Which component are you using?**:

**What version of the component are you using?**:

clusterapi provider with openstack ,have 2 VMs 
cluster.k8s.io/cluster-api-autoscaler-node-group-max-size=2
cluster.k8s.io/cluster-api-autoscaler-node-group-min-size=1
```
| 1b30a2a8-659b-4ba1-8ac1-58787d0f3572 | capi-openstack-control-plane-2v44j | ACTIVE | -          | Running     | k8s-clusterapi-cluster-default-capi-openstack-3=10.6.0.192, 172.24.4.2 |
| 0b303bdc-cbae-470d-9379-9d3b28762372 | capi-openstack-md-0-ns7vb          | ACTIVE | -          | Running     | k8s-clusterapi-cluster-default-capi-openstack-3=10.6.0.65              |
```

create pods and one is pending
```
# k get pods
NAME                                READY   STATUS    RESTARTS   AGE
nginx-deployment-779b9ccfcb-n6lrv   1/1     Running   0          4h25m
nginx-deployment-779b9ccfcb-x4896   0/1     Pending   0          4h25m
```

auto scale done , that's good 
```
| 1b30a2a8-659b-4ba1-8ac1-58787d0f3572 | capi-openstack-control-plane-2v44j | ACTIVE | -          | Running     | k8s-clusterapi-cluster-default-capi-openstack-3=10.6.0.192, 172.24.4.2 |
| 0b303bdc-cbae-470d-9379-9d3b28762372 | capi-openstack-md-0-ns7vb          | ACTIVE | -          | Running     | k8s-clusterapi-cluster-default-capi-openstack-3=10.6.0.65              |
| 7f29bbd0-9f5c-486c-ac28-349e0e108a2c | capi-openstack-md-0-sm8mp          | ACTIVE | -          | Running     | k8s-clusterapi-cluster-default-capi-openstack-3=10.6.0.28              |
```

delete the 2 pods (the deployment) and wait for auto scale to decrease node

```
| 1b30a2a8-659b-4ba1-8ac1-58787d0f3572 | capi-openstack-control-plane-2v44j | ACTIVE | -          | Running     | k8s-clusterapi-cluster-default-capi-openstack-3=10.6.0.192, 172.24.4.2 |
[...]

PR #3840, opened 2021-01-22
Add log for registered/not registered nodes

jichenjc:
fixes: #3836


MaciekPytel:
In a cluster with thousands of nodes it will print all of them every loop. I think this is too much volume, especially for a log that will print the same thing over and over for the most time.

In general CA is already generating huge logs volume in large clusters. I'm against adding any log that will print all pods or all nodes every loop.

jichenjc:
ok, that make sense ..thanks for the review comments

PR #3848, opened 2021-01-26
aws: support arm64 instances

bpineau:
Sets the `kubernetes.io/arch` (and legacy `beta.kubernetes.io/arch`) label to the proper instance architecture.
While at it, re-gen the instance types list (adding new instance types that were missing)


gjtempleton:
Thanks!

Sorry for the delay in approving this.

/lgtm
/approve

PR #3853, opened 2021-01-28
Expose backoff time parameters

cwdsuzhou:
```
	InitialNodeGroupBackoffDuration
	MaxNodeGroupBackoffDuration
	NodeGroupBackoffResetTimeout
```

are hard codes now, maybe not reasonable for all cloud providers. 
we need expose these backoff time parameters

### DIFF
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -62,7 +62,7 @@ import (
 	componentbaseconfig "k8s.io/component-base/config"
 	"k8s.io/component-base/config/options"
 	"k8s.io/component-base/metrics/legacyregistry"
-	klog "k8s.io/klog/v2"
+	"k8s.io/klog/v2"
 )
 
 // MultiStringFlag is a flag for passing multiple parameters using same flag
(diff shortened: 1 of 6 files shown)


cwdsuzhou:
/assign @feiskyer  


Could you help review this PR?

cluster-autoscaler/main.go
```diff
@@ -177,6 +178,13 @@ var (
 	clusterAPICloudConfigAuthoritative = flag.Bool("clusterapi-cloud-config-authoritative", false, "Treat the cloud-config flag authoritatively (do not fallback to using kubeconfig flag). ClusterAPI only")
 	cordonNodeBeforeTerminate          = flag.Bool("cordon-node-before-terminating", false, "Should CA cordon nodes before terminating during downscale process")
 	daemonSetEvictionForEmptyNodes     = flag.Bool("daemonset-eviction-for-empty-nodes", false, "DaemonSet pods will be gracefully terminated from empty nodes")
+
+	initialNodeGroupBackoffDuration = flag.Duration("initial-node-group-backoff-duration", clusterstate.InitialNodeGroupBackoffDuration,
```
mwielgus, line 182: Let's remove constants from clusterstate and put them here.

mwielgus (changes requested):

cluster-autoscaler/main.go
cwdsuzhou: seems  we can not , some unit tests also dependent on the constants.

therc:
@cwdsuzhou any chance you could rebase this? Or should someone else take over? We have hit similar issues

cwdsuzhou:
> @cwdsuzhou any chance you could rebase this? Or should someone else take over? We have hit similar issues

sure

cwdsuzhou:
PR rebased

@mwielgus  @feiskyer @vivekbagade  


could you help  review  this PR

nsherak1 (approved):

therc:
Looks like formatting changes needed on main.go?

cwdsuzhou:
@therc PR rebased

could anyone can merge this PR

@mwielgus @nsherak1

cluster-autoscaler/main.go
towca: I see that the parameters are used to initialize backoff in tests in two packages: core and  clusterstate.

In principle the default parameters shouldn't be the only ones we're testing. The tests should be written in a way that allows using any valid set of parameters. Frequently in these situations dummy values like "1 minute" are used.

We are where we are with the current tests, but I see no reason they should be using the default values specifically. WDYT about changing the parameters to literals/local non-exported constants in these 2 packages, and moving the defaults to the flags themselves? We could change the test parameters to some dummy values if feasible, but leaving the current values also SGTM if the tests are depending on the specific values.

Compare this with e.g. `scaleDownCandidatesPoolMinCount` - the default value `50` is in the flag, and different tests are using different values for it. Or `unremovableNodeRecheckTimeout` - the default value `5*time.Minute` is in the flag, and tests are just using the `5*time.Minute` literal.

towca (changes requested):
Hi, sorry this PR has taken such a long time. I'm committed to iterating on it and merging quickly, but I agree with the existing comment. I added some thoughts, could you take a look?

cwdsuzhou:
@towca hi, sorry for delay. PR has modified. 

PTAL, thanks

towca:
Thanks for the changes!
/lgtm
/approve

towca (approved):

nakamume:
@towca I see these changes have been merged to [cluster-autoscaler-1.22.3](https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-1.22.3/cluster-autoscaler/main.go#L195) release branch but are not yet available in the corresponding docker images.

In what release would this be made possible?

nakamume:
Thanks @towca I've made a proposal here https://github.com/kubernetes/autoscaler/pull/4996

PR #3860, opened 2021-02-01
Test and verify using Git Actions

### LINKED ISSUE #3825: Travis-ci.org Shutdown - Migration to Travis-ci.com
N/A

N/A

N/A

N/A

Travis CI builds should be processed through https://travis-ci.com

**What happened instead?**:

The deprecated https://travis-ci.org is still used, potentially breaking our builds once the old domain is switched off.

Visit a travis build from a PR to this repository:
https://travis-ci.org/github/kubernetes/autoscaler/builds/752730430

Travis' announcement of the intended shutdown of travis-ci.org:
https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom

Travis' instructions for moving to travis-ci.com:
https://docs.travis-ci.com/user/migrate/open-source-repository-migration

mwielgus:
As Travis or is about to shut down, we are migrating to some alternative. 
Fixes #3825

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,36 @@
+name: Tests
+
+on:
+  - push
+  - pull_request
+
+env:
+  GOPATH: ${{ github.workspace }}/go
+
+jobs:
+  test-and-verify:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.15
+
+      - uses: actions/checkout@v2
+        with:
+          path: ${{ env.GOPATH }}/src/k8s.io/autoscaler
+
+      - name: Apt-get
+        run: sudo apt-get install libseccomp-dev -qq
+
+      - name: Prepare
+        working-directory: ${{ env.GOPATH }}/src/k8s.io/autoscaler
+        run: hack/install-verify-tools.sh
+
+      - name: Verify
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,36 @@
...
+  GOPATH: ${{ github.workspace }}/go
+
+jobs:
+  test-and-verify:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.14
```
MaciekPytel, line 17: I think we're at 1.15 now (at least in CA).

MaciekPytel (commented):
Are you planning to remove travis config in a follow-up PR? If so this looks good to me, though I have little expertise in github workflows.

@gjtempleton do you want to also review? I think you're more qualified than I am.

.github/workflows/ci.yaml
mwielgus: Travis is at 1.14

https://github.com/kubernetes/autoscaler/blob/master/.travis.yml#L9

Let's go with 1.14. Then we can update to 1.15.

MaciekPytel: I think travis being on 1.14 is a bug (we must have forgotten - again - to bump it when bumping go version). https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/go.mod#L9

mwielgus: Ok, whatever. Flipped to 1.15.

PR #3862, opened 2021-02-01
gce: spread instance template refreshes deadlines

bpineau:
`FetchMigTemplate()` is costly, causing two cloud API calls (InstanceGroupManagersService.Get and InstanceTemplatesService.Get, at about ~400ms/call).

Due to that, expiring all instance templates at once results in significant latencies spikes, and exposes to throttling, in particular for clusters having many MIGs attached.

On this example medium sized cluster, RunOnce peaked at 1m40s every 30mn, now stable below 30s per iteration:
avg main   getnodeinfosforgroups functions durations
Main loop - p99 duration

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/cache.go
+++ b/cluster-autoscaler/cloudprovider/gce/cache.go
@@ -22,6 +22,7 @@ import (
 	"reflect"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


bpineau:
Closing in favor of  #4423

PR #3863, opened 2021-02-01
remove extra build command from cluster-autoscaler

elmiko:
This change removes an extra build command from the `build-arch-%`
target. The extra build command has been in the Makefile since the
beginning but doesn't seem to be needed any more.

This was referenced in a comment on slack[0], and it doesn't seem to affect the binary building inside or outside of a container build.

[0] https://kubernetes.slack.com/archives/C09R1LV8S/p1612003154063600

### DIFF
--- a/cluster-autoscaler/Makefile
+++ b/cluster-autoscaler/Makefile
@@ -35,12 +35,6 @@ export DOCKER_CLI_EXPERIMENTAL := enabled
 build: build-arch-$(GOARCH)
 
 build-arch-%: clean-arch-%
-	$(ENVVAR) GOOS=$(GOOS) GOARCH=$* go build ${LDFLAGS_FLAG} ${TAGS_FLAG} ./...
-	$(ENVVAR) GOOS=$(GOOS) GOARCH=$* go build -o cluster-autoscaler-$* ${LDFLAGS_FLAG} ${TAGS_FLAG}
-
-build-binary: build-binary-arch-$(GOARCH)
-
-build-binary-arch-%: clean-arch-%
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/Makefile
```diff
@@ -35,7 +35,6 @@ export DOCKER_CLI_EXPERIMENTAL := enabled
 build: build-arch-$(GOARCH)
 
 build-arch-%: clean-arch-%
-	$(ENVVAR) GOOS=$(GOOS) GOARCH=$* go build ${LDFLAGS_FLAG} ${TAGS_FLAG} ./...
```
mwielgus, line 38: With this change `build-arch-%` is the same as `build-binary-arch-%`. Do we need both?

mwielgus (changes requested):

cluster-autoscaler/Makefile
elmiko: good question, i don't _think_ so as they all seem to create the same binary target

elmiko: i looked into this a little more and i think we need to keep the build-arch target to satisfy the all target in https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/Makefile#L2

i have a feeling we will break a build _somewhere_ if we remove the build-arch target

elmiko: i think you are correct though, we could standardize around either build/build-arch or build-binary/build-binary-arch. i'll try out a few different combos.

elmiko:
after some experimentation, i have removed the `build-binary-arch` and `build-binary-arch-%` targets, and replaced their usage within the commands in favor of `build` and `build-arch-%`.

mwielgus (approved):
/lgtm
/approve

PR #3880, opened 2021-02-10
Helm Chart Image Update (v1.18.1 → v1.20.0)

jonioni:
This PR updates the default image tag in the Helm Chart, from v1.18.1 to v1.20.0.

The reason for the default tag change is to enable the fixes contained in [Cluster Autoscaler 1.20.0](https://github.com/kubernetes/autoscaler/releases/tag/cluster-autoscaler-1.20.0) by default for all future Helm installation of the cluster autoscaler.

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-appVersion: 1.18.1
+appVersion: 1.20.0
 description: Scales Kubernetes worker nodes within autoscaling groups.
 engine: gotpl
 home: https://github.com/kubernetes/autoscaler
@@ -17,4 +17,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.4.0
+version: 9.5.0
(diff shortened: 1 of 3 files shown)


jonioni:
I signed it.

mwielgus (changes requested):
Why https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/Chart.yaml is not updated as well?

gjtempleton:
I'll need to look into why the tests haven't blocked this, as the Chart version (and app version) should be being bumped for this.

jonioni:
Thanks. `Chart.yml` has been updated.

gjtempleton:
You'll also need to bump the `version` in the `Chart.yaml`, not just the `appVersion` as it'll currently fail to publish a release.

jonioni:
Thanks for the notice. `version` has been bumped.

charts/cluster-autoscaler/Chart.yaml
```diff
@@ -17,4 +17,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.4.0
+version: 9.4.1
```
gjtempleton, line 20: Sorry to be a pain, but given we're bumping the default version of the CA, I'd rather make this bump a minor version rather than a patch, should increase the likelihood of those who aren't pinning it checking what's changed. 

(That's the last one, I promise, and thanks for your patience with this!)

gjtempleton (changes requested):

Eyjafjallajokull:
Shouldn't you also publish version 1.19 of the chart? I would like to update my Kubernetes cluster from 1.18 to 1.19, and currently, there is no matching chart version.

charts/cluster-autoscaler/Chart.yaml
jonioni: Yes agree, a minor version bump is more appropriate. Thanks for point it out.

jonioni:
Chart version corrected and commits squashed.

gjtempleton:
Thanks!
/lgtm
/approve

PR #3884, opened 2021-02-12
CA-Chart - Update index with final ca-chart chart release

gjtempleton:
* Pushes the release caused by #3719 to the index.yaml for indexing by artifact hub to complete marking of prior chart release name as deprecated


mwielgus (approved):
/lgtm
/approve

PR #3885, opened 2021-02-12
Final removal of ca-chart chart post deprecation

gjtempleton:
* Final removal of `cluster-autoscaler-chart` chart after marking it as deprecated in artifact hub via #3719 and #3884 - this chart was renamed by #3679 and this code was only reinstated to ensure artifact hub marked it as deprecated


mwielgus (approved):
/lgtm
/approve

Issue #3891, opened 2021-02-19
Cluster autoscaler crashing with AWS self managed nodes and node group autodiscovery


danthegoodman1:
**Which component are you using?**: Cluster Autoscaler

**What version of the component are you using?**: 1.20.0

**What k8s version are you using (`kubectl version`)?**:

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ v1.18.9-eks-d1db3c
</pre></details>

**What environment is this in?**: AWS EKS

**What did you expect to happen?**:

Cluster autoscaler will identify self managed node groups that have appropriate tagging.

**What happened instead?**:

It crashes

**How to reproduce it (as minimally and precisely as possible)**:

Here is the following section of the CA that changes it crashing vs working:

```yaml
command:
            - ./cluster-autoscaler
            - --v=4
            - --stderrthreshold=info
            - --cloud-provider=aws
            - --skip-nodes-with-local-storage=false
            - --expander=least-waste
            # - --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,k8s.io/cluster-autoscaler/dantestingcluster
            - --nodes=1:20:gggggg-NodeGroup1-1KVYZBBPT7CIT
            - --nodes=1:20:gggggg-NodeGroup2-16KZFF54VE5IZ
            - --nodes=1:20:gggggg-NodeGroup3-1N3XO1S6HIKGA
            - --max-node-provision-time=5m0s
```
This configuration works, it is when I comment out the `- --nodes=...` lines and uncomment the `- --node-group-auto-discovery=...` line it crashes.

It also prints out lots of duplicates of the following:

```go
goroutine 255 [chan receive]:
k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/client-go/tools/cache.(*sharedProcessor).run(0xc000129e30, 0xc001226cc0)
	/gopath/src/k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/client-go/tools/cache/shared_informer.go:628 +0x53
k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/apimachinery/pkg/util/wait.(*Group).StartWithChannel.func1()
	/gopath/src/k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:56 +0x2e
[...]

PR #3893, opened 2021-02-21
Add Fetch Error Output to Warning Log

danthegoodman1:
Add the output of the error to the log so users can identify why the autoscaler was unable to fetch the url.

See https://github.com/kubernetes/autoscaler/issues/3891


gjtempleton:
Thanks for the contribution!

/lgtm
/approve

PR #3894, opened 2021-02-22
Helm Chart image update to 1.19.1

Eyjafjallajokull:
This is in line with https://github.com/kubernetes/autoscaler/pull/3880, I think there should be also a release of 1.19 for anyone who wants to run on Kubernetes 1.19.

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-appVersion: 1.18.1
+appVersion: 1.19.1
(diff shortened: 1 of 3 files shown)


gjtempleton:
Hi, thanks for the PR!

I don't think this PR is required given the existence of #3880 , we don't need a 1:1 mapping of releases of the chart to each release of the autoscaler, as we already allow users to override the version of the CA they use via overriding the default value of the tag

```yaml
image:
  tag: v1.19.1
```

Is there a reason you can't pass in value overrides to the chart that means a full release of the chart with this new default image tag is required?

Eyjafjallajokull:
Ah, you are right, I forgot I can do that.

PR #3900, opened 2021-02-23
feat: add prometheus rule support

eddycharly:
This PR adds prometheus rule support.

It is complementary with `serviceMonitor`.

### DIFF
--- a/charts/cluster-autoscaler/templates/prometheusrule.yaml
+++ b/charts/cluster-autoscaler/templates/prometheusrule.yaml
@@ -1,0 +1,15 @@
+{{- if .Values.prometheusRule.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: {{ include "cluster-autoscaler.fullname" . }}
+  {{- if .Values.prometheusRule.namespace }}
+  namespace: {{ .Values.prometheusRule.namespace }}
+  {{- end }}
+  labels: {{- toYaml .Values.prometheusRule.additionalLabels | nindent 4 }}
+spec:
+  groups:
+    - name: {{ include "cluster-autoscaler.fullname" . }}
+      interval: {{ .Values.prometheusRule.interval }}
+      rules: {{- tpl (toYaml .Values.prometheusRule.rules) . | nindent 8 }}
+{{- end }}
--- a/charts/cluster-autoscaler/values.yaml
+++ b/charts/cluster-autoscaler/values.yaml
@@ -298,6 +298,21 @@ serviceMonitor:
   # serviceMonitor.path -- The path to scrape for metrics; autoscaler exposes `/metrics` (this is standard)
   path: /metrics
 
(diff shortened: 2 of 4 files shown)


gjtempleton:
Hi, thanks for the PR!

Can you please make a minor version bump to the chart's `version` under `Chart.yml` as well.

eddycharly:
Sure, I bumped the version to `9.6.0`.

charts/cluster-autoscaler/values.yaml
```diff
@@ -298,6 +298,19 @@ serviceMonitor:
...
+## The value is evaluated as a template, so, for example, the value can depend on .Release or .Chart
+## ref: https://github.com/coreos/prometheus-operator#customresourcedefinitions
+prometheusRule:
+  # prometheusRule.enabled -- If true, creates a Prometheus Operator PrometheusRule.
+  enabled: false
+  # prometheusRule.additionalLabels -- Additional labels to be set in metadata.
+  additionalLabels: {}
+  # prometheusRule.namespace -- Namespace which Prometheus is running in.
+  namespace: monitoring
+  # prometheusRule.rules -- Rules spec template.
```
gjtempleton, line 311: Is it worth linking to the API spec for the rule object here? I.e. https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#rule

charts/cluster-autoscaler/templates/prometheusrule.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: {{ include "cluster-autoscaler.fullname" . }}
+  {{- if .Values.prometheusRule.namespace }}
+  namespace: {{ .Values.prometheusRule.namespace }}
+  {{- end }}
+  labels: {{- toYaml .Values.prometheusRule.additionalLabels | nindent 4 }}
+spec:
+  groups:
+    - name: {{ include "cluster-autoscaler.fullname" . }}
+      rules: {{- tpl (toYaml .Values.prometheusRule.rules) . | nindent 8 }}
```
gjtempleton, line 13: Do we want to support users setting the interval as well? (Potentially the partial_response_strategy as well, though that's probably less useful, so not as sold on including.) 

https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#rulegroup

eddycharly: Good catch ! , i'll add that.

charts/cluster-autoscaler/values.yaml
eddycharly: Added the reference doc link. I'm thinking about adding an example too, what do you think ?

charts/cluster-autoscaler/templates/prometheusrule.yaml
eddycharly: Done for `interval`. Not sure about `partial_response_strategy` either.

Issue #3902, opened 2021-02-25
[VPA] Pod creation fails if VPA has only one controlledResources


Ladicle:
**Which component are you using?**: Vertical Pod Autoscaler

**What version of the component are you using?**:

Component version: 0.9.2

**What k8s version are you using (`kubectl version`)?**:

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.2", GitCommit:"faecb196815e248d3ecfb03c680a4507229c2a56", GitTreeState:"clean", BuildDate:"2021-01-14T05:14:17Z", GoVersion:"go1.15.6", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.1", GitCommit:"c4d752765b3bbac2237bf87cf0b1c2e307844666", GitTreeState:"clean", BuildDate:"2020-12-18T12:00:47Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What environment is this in?**:

both minikube & on-prem k8s cluster

**What did you expect to happen?**:

Deployment with limits can be controlled only one resource by VPA.

**What happened instead?**:

VPA resource only has cpu as a controlledResource, but a vpa-admission-controller sends patch for non controlled resource limits and the value is 0. (like this`{add /spec/containers/0/resources/limits/memory 0}`) For that reason, replicaset-controller gets `FailedCreate` error.

vpa-admission-controller patch log: 
```
vpa-admission-controller-6cd546c4f-zckmj admission-controller I0225 05:53:42.693154       1 server.go:110] Sending patches: [{add /metadata/annotations map[]} {add /spec/containers/0/resources/requests/cpu 587m} {add /spec/containers/0/resources/limits/cpu 5870m} {add /spec/containers/0/resources/limits/memory 0} {add /metadata/annotations/vpaUpdates Pod resources updated by hamster-vpa: container 0: cpu request, cpu limit, memory limit} {add /metadata/annotations/vpaObservedContainers hamster}
```

replicaset-controller warning event:
```
tainers[0].resources.requests: Invalid value: "256Mi": must be less than or equal to memory limit
[...]

PR #3903, opened 2021-02-25
Do not set limits when recommendation is 0

Ladicle:
In this PR changed not to set a limit when a recommendation is empty. This solves #3902 issue of sending patches for the not-controlled resource by VPA when Pod has resource limits.


jbartosik:
Looks good to me. I want to check if this with instructions from  #3902 before approving

Ladicle:
@jbartosik the following gist is a log when I run this fixed vpa-admission-controller on the local minikube.
https://gist.github.com/Ladicle/fcab44279992eebb8ed7ab9a42a706b0

jbartosik (approved):
/lgtm
/approve

Issue #3798, opened 2021-01-07
Cluster Autoscaler Deployment link in README for GCE is wrong?


jatin-pathangi:
**Which component are you using?**:
Cluster Autoscaler

**What version of the component are you using?**:
1.16.x

Component version:

**What k8s version are you using (`kubectl version`)?**:
1.16
<details><summary><code>kubectl version</code> Output</summary><br><pre>

</pre></details>

**What environment is this in?**:
GCE

**What did you expect to happen?**:

**What happened instead?**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
The link for GCE here https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler#deployment is this: https://kubernetes.io/docs/tasks/administer-cluster/, which doesn't seem to contain any relevant information about deploying cluster autoscaler on a GCE cluster, whereas every other supported Cloud Provider has their own README page. I'm wondering why the discrepancy and if this is an oversight? I'm trying to set up Cluster Autoscaler on a kubernetes cluster in GCE (not GKE, but a self-managed cluster), and was looking targeted documentation similar to what's present for AWS/Azure etc.

unixfox:
I'm also trying to deploy autoscaler on GCE (not GKE) and I just don't find any documentation for that.

Did you manage to install it anyway?

MoSheikh:
I'm experiencing difficulties with the same issue as well. It would be extremely helpful having some documentation for GCE since all other major cloud providers do. Thank you!

MiguelLucas:
Would also appreciate if some kind of documentation was provided, as there is tons of information on deploying the cluster autoscaler on GKE but not on GCE only. Thank you

js-max:
Checking https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/gce I'm not able to see any README or related info.

Am I missing something?

PR #3907, opened 2021-02-26
Fix GCE where no cloud-config-path is required

samos123:
GCE doesn't require loading /etc/gce.conf from a hostpath. In-fact this file isn't available at all on GCE nodes nor on GKE nodes. This was a left-over from the way GKE bundled autoscaler seems to work. There have been multiple reports of GCE deployment being broken due to this. For example see: helm/charts#10292 and helm/charts#15509

Another example of people not being able to deploy GCE because the README points to no longer existing docs: https://github.com/kubernetes/autoscaler/issues/3798

After the PR is merged this blog post will work: https://samos-it.com/posts/gke-custom-oss-cluster-autoscaler.html however if you try to run the blogpost steps today it would fail because it tries to mount /etc/gce.conf with no way of disabling that behaviour.

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -17,4 +17,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.8.0
+version: 9.9.0
--- a/charts/cluster-autoscaler/templates/deployment.yaml
+++ b/charts/cluster-autoscaler/templates/deployment.yaml
@@ -70,7 +70,7 @@ spec:
             - --cluster-name={{ .Values.magnumClusterName }}
             {{- end }}
           {{- end }}
-          {{- if or (eq .Values.cloudProvider "gce") (eq .Values.cloudProvider "magnum") }}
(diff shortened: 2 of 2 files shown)


samos123:
@gjtempleton and @MaciekPytel could you please review? I've tested that this works on a GKE cluster the original helm chart doesn't work. There are several reports of others hitting the same issue. For example see: https://github.com/helm/charts/issues/10292

gjtempleton:
Hey, thanks for the PR

Could you please rebase against the repo's default branch as you're currently running an old version of the Helm chart PR CI, this was updated in #3932.

You'll need to increment the chart's `version` in the `Chart.yaml` to enable the CD to publish a new version of the chart.

samos123:
@gjtempleton Thanks for reviewing, sorry about not providing more context initially. I've added the description to this PR with examples of other people getting stuck. I also linked a blog post which allows you to reproduce the issue on GCE. The reason why not many people hit this issue is because most GKE customers use the bundled cluster autoscaler. Most K8s users on GCP use GKE. Other cloud providers don't always bundle the cluster autoscaler so that's why you see a lot more AWS users of the helm chart.

gjtempleton:
Thanks for clarifying, makes sense now!
/lgtm
/approve

PR #3910, opened 2021-02-26
Correct type error in readme chart

allthatilk:
`expanderPriorities` is incorrectly set as an object type when the chart requires a string type for this key.

I know this might count as a trivial change but it wasted a lot of my time today and is an easily missed mistake in the docs.


gjtempleton:
Hey,
Thanks for the PR and raising this, I can see that it does error as its currently set up. There are a few things you'll need to do to get this merged:

1. Can you follow the CLA instructions linked to by the bot above
1. The Readme is autogenerated from the value types specified in the `Values.yaml`, the pre-commit hooks [mentioned here](https://github.com/kubernetes/autoscaler/tree/master/charts#pre-commit-hooks) should take care of regenerating the Readme once you update the Values file
1. It would be good to have an example of using the priorities to make validation/users adopting this easier, could you add that as a commented-out example to the values?

gjtempleton:
@allthatilk Are you able to address the feedback comments already provided?

mwielgus:
CLA not signed, comments not addressed. Closing. Feel free to reopen once fixed.

Issue #3913, opened 2021-03-01
Need support for customized recommender in Vertical Pod Autoscaler (VPA)


wangchen615:
**Which component are you using?**:
Vertical Pod Autoscaler

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:
Today, the current VPA recommends CPU/Mem requests and limits based on one recommender, 
which recommends the future requests/limits based on the historical usage observed in a 
rolling time window. As there is no universal recommendation policy that applies to all 
types of workload, this feature requests supporting multiple customized recommenders in VPA.

**Describe the solution you'd like.**:
The support of a customized recommender can be implemented via a first-citizen approach. Namely, a dedicated field `recommenderName` can be added to the VPA object to indicate which recommender to use.

**Describe any alternative solutions you've considered.**:
- Develop a plugin framework for customizable recommenders.
- Develop a label selector approach.

**Additional context.**:
This issue is created to obtain a KEP number.

bskiba:
Shouldn't we keep open for implementation? Or do you want to track separately?

wangchen615:
> Shouldn't we keep open for implementation? Or do you want to track separately?

Yes, let's keep it open until the implementation is merged.

Jeffwan:
@wangchen615 What's the status of this VPA KEP? Is there a PR going on?

wangchen615:
@Jeffwan , I am still working on the implementation. Will create the PR soon, in 2 weeks.

Jeffwan:
/remove-lifecycle stale

We are still working on this

PR #3914, opened 2021-03-01
Add enhancement proposal for feature request #3913

wangchen615:
Added the enhancement proposal for the support of a customized recommender in Vertical Pod Autoscaler.

@bskiba


vertical-pod-autoscaler/enhancements/3913-customized-recomender-vpa.md
```diff
@@ -0,0 +1,272 @@
...
+Containers used in monitoring may receive load periodically to process but need to be long-running 
+to listen to incoming traffic. Thus, these containers usually exhibit cyclic patterns, alternating 
+between usage spikes and idling. Resizing containers according to usage observed in the previous 
+time window may always lead to under-provision for a short period when the load spikes just arrive. 
+The problem will even lead to frequent pod evictions when the pod was resized according to the idling 
+period and the host resource has been taken by other pods.
+
+#### Story 2 
+
+- Containers with Different but Recurrent Behaviors in Resource Usage
```
bskiba, line 73: I'm not sure I fully understand this one. If this is a single job, we don't have the ability to resize the workload between the phases. Currently default VPA will provision for the busiest phase, unless the phases are ~days long.

vertical-pod-autoscaler/enhancements/3913-customized-recomender-vpa.md
```diff
@@ -0,0 +1,272 @@
...
+           type: object
+```
+
+Correspondingly, the `VerticalPodAutoscalerSpec` in `pkg/apis/autoscaling.k8s.io/v1/types.go` 
+should be updated to include the `recommenderName` field.
+
+```golang
+// VerticalPodAutoscalerSpec is the specification of the behavior of the autoscaler.
+type VerticalPodAutoscalerSpec struct {
+  // Denote the recommender to be used
```
bskiba, line 113: // Name of the recommender responsible for generating recommendation for this object.

vertical-pod-autoscaler/enhancements/3913-customized-recomender-vpa.md
```diff
@@ -0,0 +1,272 @@
...
+  RecommenderName string `json:"recommenderName,omitempty" protobuf:"bytes,1,opt,name=recommenderName"`
```
bskiba, line 114: nit: colision in protobuf field numbering (1 is in TargetRef as well)

vertical-pod-autoscaler/enhancements/3913-customized-recomender-vpa.md
```diff
@@ -0,0 +1,272 @@
...
+}
+```
+
+Therefore, when loading VPA objects to the `clusterStateFeeder`, it can use the field selector to select VPA CRDs that 
+have `recommenderName` equal to the current clusterState’s `RecommenderName`.
+```golang
+// Fetch VPA objects and load them into the cluster state.
+func (feeder *clusterStateFeeder) LoadVPAs() {
+  // List VPA API objects.
+  allVpaCRDs, err := feeder.vpaLister.List(labels.Everything())
```
bskiba, line 191: Is it possible to list only the VPA objects that match condition in line 200?

bskiba: I think it should be possible to use selector by field values, but needs to be checked.

vertical-pod-autoscaler/enhancements/3913-customized-recomender-vpa.md
```diff
@@ -0,0 +1,272 @@
...
+## Design Details
+
+### Test Plan
+- Add e2e test demonstrating the default recommender ignores a VPA which specifies an alternate recommender.
+
+### Upgrade / Downgrade Strategy
+For cluster upgrades, the VPA from the previous version will continue working as before. 
+There is no change in behavior or flags which have to be enabled or disabled.
+
+## Alternatives
```
bskiba, line 238: Can you add a "rejected because" sentence to the alternatives considered?

vertical-pod-autoscaler/enhancements/3913-customized-recomender-vpa.md
```diff
@@ -0,0 +1,272 @@
...
+Add a label for the CRD object to denote the recommender’s name. When making 
+recommendations in the recommender, only the VpaCrds with the label 
+`recommender=default` will be loaded and updated by the existing recommender. 
+A label selector approach is too powerful and users can easily ignore those labels 
+and misconfigure the VPA objects.
+
+## Out of Scope
+
+Although this proposal will enable alternate recommenders, no alternate recommenders 
+will be created as part of this proposal.
```
bskiba, line 256: Can you add multiple recommendations as out of scope? So that we have record that we discussed those.

bskiba (commented):
One concern that I did not think of before and @kgolab brought it to my attention:

What do we do with VPA objects that specify a recommender that does not exist? In the current proposal they would get no recommendations, without any way to signal that to the user. Maybe we should actually produce default recommendations for them? Can you check how this is done for the custom scheduling feature? It seems they would have the same problem.

bskiba:
@jbartosik FYI

vertical-pod-autoscaler/enhancements/3913-customized-recomender-vpa.md
```diff
@@ -0,0 +1,272 @@
...
+(CPU and Memory) predictions can significantly reduce over-provisioning and OOM kills in VPA.
+
+### Goals
+
+- Allow the VPA object to specify a customized recommender to use. 
+- Allow the VPA object to use the default recommender when no recommender is specified.
+
+### Non-Goals
+
+- We assume no pod uses two recommenders at the same time.
```
kgolab, line 53: I wonder how much we need this assumption.

Just to be clear: I'm not suggesting the we handle multiple recommenders already but if we could just make sure we don't close the way towards that goal w/o introducing much extra work, I'd be for it.

There are at least two use-cases which might require multiple recommenders:
* recommendation depending on pod life-cycle phase - admittedly this requires in-place updates to be fully useful
* separate recommenders for separate resources (e.g. you could have few dedicated recommender for CPU patterns, a separate set for memory patterns and let users mix them up according to their needs, w/o deploying all the combinations).

vertical-pod-autoscaler/enhancements/3913-customized-recomender-vpa.md
```diff
@@ -0,0 +1,272 @@
...
+Add a webhook system for customized recommendations. The default VPA recommender would 
+call any available recommendation webhooks, and if any of them make a recommendation, 
+the recommender would use that recommendation instead of making its own. If none make 
+a recommendation, it would make its recommendation as it currently does.
+
+### Develop a label selector approach.
+Add a label for the CRD object to denote the recommender’s name. When making 
+recommendations in the recommender, only the VpaCrds with the label 
+`recommender=default` will be loaded and updated by the existing recommender. 
+A label selector approach is too powerful and users can easily ignore those labels
```
kgolab, line 250: Would it still be a problem if we said that VpaCrds w/o any label are still handled by the default recommender?

vertical-pod-autoscaler/enhancements/3913-customized-recomender-vpa.md
```diff
@@ -0,0 +1,272 @@
...
+
+Containers running spark/deep learning training workloads are known to show recurring and repeating 
+patterns in resource usage. Prior research has shown that different but recurrent behaviors occur 
+for different containerized tasks, such as Spark or deep learning training. These common patterns can 
+be represented by phases, which display similar resource usage of computational resources over time. 
+There are common sequences of patterns for different executions of the workload and they can be used 
+to proactively predict future resource usage more accurately. The default recommender in the current 
+VPA adopts a reactive approach so a more proactive recommender is needed for these types of workload.
+
+### Implementation Details
```
kgolab, line 83: I think it's clear from the context, especially from rejected alternatives, but it still would be good to explicitly state that the customised recommenders are expected to be running outside VPA binary (or at least that they can run separately).

Maybe an image/description showing how VPA controller would look like (multiple recommender binaries, still a single Updater & web-hook)?

wangchen615:
@bskiba @kgolab could you review again?

kgolab:
@wangchen615 , I reviewed once more. I'm fine with the latest changes - no new comments there. However my old comments still stand:
* one is purely editorial: I think it would be useful to state how we expect the additional recommenders to be deployed
* the other was about not closing the way towards multiple recommenders. I understand you want to keep multiple recommenders but let's not close the possibility with this change. For example: what if recommenderName was not a string but an array? We could still say that for now only one recommender is supported (e.g. we don't change status) but if we ever come to multiple recommenders, the field won't need to be changed (or rather accompanied by a new one).

wangchen615 (commented):
> @wangchen615 , I reviewed once more. I'm fine with the latest changes - no new comments there. However my old comments still stand:
> 
> * one is purely editorial: I think it would be useful to state how we expect the additional recommenders to be deployed
> * the other was about not closing the way towards multiple recommenders. I understand you want to keep multiple recommenders but let's not close the possibility with this change. For example: what if recommenderName was not a string but an array? We could still say that for now only one recommender is supported (e.g. we don't change status) but if we ever come to multiple recommenders, the field won't need to be changed (or rather accompanied by a new one).

@kgolab , I added a section called deployment details according to your comments. https://github.com/wangchen615/autoscaler/blob/customizable_recommender_kep/vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa/README.md#deployment-details

Could you help take a final review?

vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa/README.md
```diff
@@ -0,0 +1,290 @@
...
+The customized recommender is supposed to be deployed as a separate deployment that is chosen 
+by different sets of VPA objects.Each VPA object is supposed to choose only one recommender at a time.
+The way how the default recommender and the customized recommender are running and interacting with VPA objects 
+are shown in the following drawing.
+
+<img src="images/deployment.png" alt="deployment" width="720" height="360"/>
+
+Though we do not support a VPA object to use multiple recommenders in this proposal, we leave the possibility of necessary  
+changes of using multiple recommenders in the future. Namely, in the future, we may come to multiple recommenders for one VPA
+object and the field of `recommenderName` can be an array instead of a string.
```
kgolab, line 241: This might not be so easy - please see https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#on-compatibility

So maybe we could instead already make it an array and explicitly state that only the first entry is used for now (as multiple recommenders are not supported with this KEP)?

kgolab:
[...]
> @kgolab , I added a section called deployment details according to your comments. https://github.com/wangchen615/autoscaler/blob/customizable_recommender_kep/vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa/README.md#deployment-details
> 
> Could you help take a final review?

Thank you, this diagram and its description clarifies a lot.

wangchen615:
@kgolab @bskiba Thanks a lot for all your comments. Is this KEP ready to be merged?

kgolab:
> @kgolab @bskiba Thanks a lot for all your comments. Is this KEP ready to be merged?

It looks fine from my perspective.

I'd still slightly prefer to make recommenderName an array already (and just use the first item for now) but since you marked similar comment as resolved, I guess your preference is to stay with a string field and solve the API change separately, if/when required.

Leaving to @bskiba for final approval.

vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa/README.md
```diff
@@ -0,0 +1,290 @@
...
+```
+
+Correspondingly, the `VerticalPodAutoscalerSpec` in `pkg/apis/autoscaling.k8s.io/v1/types.go` 
+should be updated to include the `recommenderName` field.
+
+```golang
+// VerticalPodAutoscalerSpec is the specification of the behavior of the autoscaler.
+type VerticalPodAutoscalerSpec struct {
+  // Name of the recommender responsible for generating recommendation for this object.
+  RecommenderName string `json:"recommenderName,omitempty" protobuf:"bytes,1,opt,name=recommenderName"`
```
bskiba, line 116: nit: Not sure if that's what is happening here, but for backward compatibility you should not re-index the fields in spec, i.e. if TargetRef was 
`protobuf:"bytes,1,name=targetRef"` 
it should stay that way

vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa/README.md
bskiba: I propose to do as Karol suggests, and have this be an array, with one element only. 
It is indeed not backwards-compatible to change API field type so we would not be able to do this. 

It should be relatively easy to have the admission controller validate that the array has <=1 elements for now.

bskiba (approved):
Looks good, thanks a lot.

Last comment before LGTM is that I would like to once more to consider making recommender an array (the change Karol suggested), for which we will support only first element as part of this KEP. Otherwise, we will not be able to change that field later due to backwards incompatibility.

I can support you with adding validation for now that the array contains one item only.

One more nit: can you add white background to the image? I'm viewing github in dark mode and the image text is not visible in that setting.

bskiba:
/lgtm
/approve

Thanks!

PR #3921, opened 2021-03-03
add arch specific cluster-autoscaler targets to gitignore

elmiko:
adding entries for the amd64 and arm64 targets.


mwielgus (approved):
/lgtm
/approve

PR #3922, opened 2021-03-03
add a metric for unregistered nodes removed by cluster autoscaler

elmiko:
This change adds a new metric which counts the number of nodes removed
by the cluster autoscaler due to being unregistered with kubernetes.

User Story

As a cluster-autoscaler user, I would like to know when the autoscaler
is cleaning up nodes that have failed to register with kubernetes. I
would like to monitor the rate at which failed nodes are being removed
so that I can better alert on infrastructure issues which may go
unnoticed elsewhere.

### DIFF
--- a/cluster-autoscaler/metrics/metrics.go
+++ b/cluster-autoscaler/metrics/metrics.go
@@ -18,8 +18,9 @@ package metrics
 
 import (
 	"fmt"
-	"k8s.io/autoscaler/cluster-autoscaler/simulator"
 	"time"
+
+	"k8s.io/autoscaler/cluster-autoscaler/simulator"
 
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/gpu"
@@ -246,6 +247,14 @@ var (
 		},
 	)
 
+	oldUnregisteredNodesRemovedCount = k8smetrics.NewCounter(
+		&k8smetrics.CounterOpts{
+			Namespace: caNamespace,
+			Name:      "old_unregistered_nodes_removed_count",
+			Help:      "Number of unregistered nodes removed by CA.",
+		},
+	)
+
 	/**** Metrics related to NodeAutoprovisioning ****/
 	napEnabled = k8smetrics.NewGauge(
 		&k8smetrics.GaugeOpts{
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/metrics/metrics.go
```diff
@@ -18,9 +18,10 @@ package metrics
 
 import (
 	"fmt"
-	"k8s.io/autoscaler/cluster-autoscaler/simulator"
 	"time"
 
+	"k8s.io/autoscaler/cluster-autoscaler/simulator"
```
elmiko, line 23: i think my go fmt did this

michaelgugino (approved):
This would be quite helpful!

cluster-autoscaler/metrics/metrics.go
```diff
@@ -426,3 +437,9 @@ func UpdateScaleDownInCooldown(inCooldown bool) {
 		scaleDownInCooldown.Set(0.0)
 	}
 }
+
+// RegisterOldUnregisteredNodesRemoved records number of old unregistered
+// nodes that have been removed by the cluster autoscaler
+func RegisterOldUnregisteredNodesRemoved(nodesCount int, nodegroupId string) {
+	oldUnregisteredNodesRemovedCount.WithLabelValues(nodegroupId).Add(float64(nodesCount))
```
mwielgus, line 444: Please don't add labels that have potentially high cardinality (many times new node pools are created by scripts with random name). It may cause problems in many metric storage systems. A quote from prometheus doc:

```
CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.
```

mwielgus (changes requested):

cluster-autoscaler/metrics/metrics.go
elmiko: that's fair, i can imagine the cardinality getting out of control in those situations. it is really helpful from a debugging perspective, but i suppose the metric could lead some to inspect the autoscaler logs or events. i'll remove the label.

michaelgugino: I think have many small nodepools is not standard behavior.  One compromise is to make this metric optional and keep it off by default.

For the vast majority of clusters, this metric would be highly useful.

elmiko: an interesting thought about adding a flag to enable/disable this metric. i'd be ok with that, my impression is that we have been trying to limit adding more cli flags though. curious to hear what @mwielgus thinks.

mwielgus: We have other ways of providing detailed statistics in CA - there is a status configmap and logs. I would rather not copy their functionality in metrics. Or at least I would not add group name labels in just single place - cluster size, etc could also be split across node groups. If you feel strongly about labels, please merge this PR without labels and start a new PR/discussion about consistent metrics labeling.

elmiko: thanks Marcin, appreciate the discussion. i'm going to repost without the label because i think this is useful as a metric source regardless of that detail.

elmiko:
updated to remove label

mwielgus (approved):
/lgtm
/approve

PR #3924, opened 2021-03-05
Fix bug where a node that becomes ready after 2

vivekbagade:
Fix bug where a node that becomes ready after 2
mins can be treated as unready. Deprecated LongNotStarted

In cases where node n1 would:
1) Be created at t=0min
2) Ready condition is true at t=2.5min
3) Not ready taint is removed at t=3min
the ready node is counted as unready

Tested cases after fix:
1) Case described above
2) Nodes not starting even after 15mins still
treated as unready
3) Nodes created long ago that suddenly become unready are
counted as unready.

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -43,10 +43,6 @@ const (
 	// MaxNodeStartupTime is the maximum time from the moment the node is registered to the time the node is ready.
 	MaxNodeStartupTime = 15 * time.Minute
 
-	// MaxStatusSettingDelayAfterCreation is the maximum time for node to set its initial status after the
-	// node is registered.
-	MaxStatusSettingDelayAfterCreation = 2 * time.Minute
-
 	// MaxNodeGroupBackoffDuration is the maximum backoff duration for a NodeGroup after new nodes failed to start.
 	MaxNodeGroupBackoffDuration = 30 * time.Minute
 
@@ -399,7 +395,7 @@ func (csr *ClusterStateRegistry) IsNodeGroupHealthy(nodeGroupName string) bool {
 
 	if unjustifiedUnready > csr.config.OkTotalUnreadyCount &&
 		float64(unjustifiedUnready) > csr.config.MaxTotalUnreadyPercentage/100.0*
-			float64(readiness.Ready+readiness.Unready+readiness.NotStarted+readiness.LongNotStarted) {
+			float64(readiness.Ready+readiness.Unready+readiness.NotStarted) {
 		return false
 	}
 
@@ -452,7 +448,7 @@ func (csr *ClusterStateRegistry) getProvisionedAndTargetSizesForNodeGroup(nodeGr
 		}
 		return 0, target, true
 	}
-	provisioned = readiness.Registered - readiness.NotStarted - readiness.LongNotStarted
+	provisioned = readiness.Registered - readiness.NotStarted
 
 	return provisioned, target, true
 }
@@ -531,8 +527,6 @@ type Readiness struct {
 	// Number of nodes that are being currently deleted. They exist in K8S but
 	// are not included in NodeGroup.TargetSize().
 	Deleted int
-	// Number of nodes that failed to start within a reasonable limit.
-	LongNotStarted int
 	// Number of nodes that are not yet fully started.
 	NotStarted int
 	// Number of all registered nodes in the group (ready/unready/deleted/etc).
@@ -554,12 +548,10 @@ func (csr *ClusterStateRegistry) updateReadinessStats(currentTime time.Time) {
 		current.Registered++
 		if deletetaint.HasToBeDeletedTaint(node) {
 			current.Deleted++
-		} else if stillStarting := isNodeStillStarting(node); stillStarting && node.CreationTimestamp.Time.Add(MaxNodeStartupTime).Before(currentTime) {
-			current.LongNotStarted++
-		} else if stillStarting {
-			current.NotStarted++
 		} else if ready {
 			current.Ready++
+		} else if node.CreationTimestamp.Time.Add(MaxNodeStartupTime).After(currentTime) {
+			current.NotStarted++
 		} else {
 			current.Unready++
 		}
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -554,9 +554,7 @@ func (csr *ClusterStateRegistry) updateReadinessStats(currentTime time.Time) {
 		current.Registered++
 		if deletetaint.HasToBeDeletedTaint(node) {
 			current.Deleted++
-		} else if stillStarting := isNodeStillStarting(node); stillStarting && node.CreationTimestamp.Time.Add(MaxNodeStartupTime).Before(currentTime) {
-			current.LongNotStarted++
-		} else if stillStarting {
+		} else if !ready && node.CreationTimestamp.Time.Add(MaxNodeStartupTime).After(currentTime) {
```
MaciekPytel, line 557: nit: I think with the new logic the order of conditions would be more readable if we handled all unready conditions one after another (like below). The logic is the same and I'm not feeling super strongly about this, it just looks more structured this way.

```
} else if ready {
	current.Ready++
} else if node.CreationTimestamp.Time.Add(MaxNodeStartupTime).After(currentTime) {
         current.NotStarted++
} else {
         current.Unready++
}
```

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -554,9 +554,7 @@ func (csr *ClusterStateRegistry) updateReadinessStats(currentTime time.Time) {
 		current.Registered++
 		if deletetaint.HasToBeDeletedTaint(node) {
 			current.Deleted++
-		} else if stillStarting := isNodeStillStarting(node); stillStarting && node.CreationTimestamp.Time.Add(MaxNodeStartupTime).Before(currentTime) {
-			current.LongNotStarted++
```
MaciekPytel, line 558: I think it's better to completely remove LongNotStarted from Readiness if it's unused anyway. It's confusing to keep it in places like upcoming nodes calculation (*I* know it doesn't change anything since it cannot take value other than 0, but it's likely to surprise someone less familiar with clusterstate).

cluster-autoscaler/clusterstate/clusterstate_test.go
```diff
@@ -319,6 +319,76 @@ func TestTooManyUnready(t *testing.T) {
...
+	fakeClient := &fake.Clientset{}
+	fakeLogRecorder, _ := utils.NewStatusMapRecorder(fakeClient, "kube-system", kube_record.NewFakeRecorder(5), false)
+	clusterstate := NewClusterStateRegistry(provider, ClusterStateRegistryConfig{
+		MaxTotalUnreadyPercentage: 10,
+		OkTotalUnreadyCount:       1,
+	}, fakeLogRecorder, newBackoff())
+	err := clusterstate.UpdateNodes([]*apiv1.Node{ng1_1, ng2_1}, nil, now)
+	assert.NoError(t, err)
+	assert.Equal(t, 1, clusterstate.GetClusterReadiness().Unready)
+	assert.Equal(t, 0, clusterstate.GetClusterReadiness().LongNotStarted)
```
MaciekPytel, line 347: You should assert NotStarted == 0. Non-zero LongNotStarted wouldn't have any practical implications (other than misreporting in configmap and, possibly, metrics). Non-zero NotStarted would result in unready node being treated as upcoming (as discussed offline).

Also maybe assert other readiness states too and Upcoming == 0 (upcoming == 0 is skirting the definition of unittest, but it is the most likely negative consequence of a bug in Readiness calculation and so I think worth checking).

cluster-autoscaler/clusterstate/clusterstate_test.go
```diff
@@ -768,57 +838,6 @@ func TestUpdateScaleUp(t *testing.T) {
 	assert.Nil(t, clusterstate.scaleUpRequests["ng1"])
 }
 
-func TestIsNodeStillStarting(t *testing.T) {
```
MaciekPytel, line 771: Not really part of this PR, but I noticed that there is no equivalent test for GetReadinessState (which has very similar logic and you effectively use it as a replacement). Maybe instead of deleting this test cut/paste it as a test for GetReadinessState (minus the recent/long part)?

cluster-autoscaler/clusterstate/clusterstate.go
vivekbagade: Done

cluster-autoscaler/clusterstate/clusterstate.go
vivekbagade: Done

cluster-autoscaler/clusterstate/clusterstate_test.go
vivekbagade: Ya. This was mistake. Fixed

cluster-autoscaler/clusterstate/clusterstate_test.go
vivekbagade: Added tests

cluster-autoscaler/utils/kubernetes/ready_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+			return node
+		}
+		t.Run("recent "+tc.desc, func(t *testing.T) {
+			node := createTestNode(1 * time.Minute)
+			isReady, _, err := GetReadinessState(node)
+			assert.NoError(t, err)
+			assert.Equal(t, tc.expectedResult, isReady)
+		})
+		t.Run("long "+tc.desc, func(t *testing.T) {
```
MaciekPytel, line 59: I think that part doesn't make sense, since GetReadinessState() doesn't care about age of the nodes? That's what I meant by  "(minus the recent/long part)" in my previous comment (sorry for not being more clear).

vivekbagade: Done

MaciekPytel:
I think MaxStatusSettingDelayAfterCreation is no longer used. Please remove it.

cluster-autoscaler/clusterstate/clusterstate_test.go
```diff
@@ -319,6 +319,78 @@ func TestTooManyUnready(t *testing.T) {
...
+	assert.NotNil(t, provider)
+	fakeClient := &fake.Clientset{}
+	fakeLogRecorder, _ := utils.NewStatusMapRecorder(fakeClient, "kube-system", kube_record.NewFakeRecorder(5), false, "some-map")
+	clusterstate := NewClusterStateRegistry(provider, ClusterStateRegistryConfig{
+		MaxTotalUnreadyPercentage: 10,
+		OkTotalUnreadyCount:       1,
+	}, fakeLogRecorder, newBackoff())
+	err := clusterstate.UpdateNodes([]*apiv1.Node{ng1_1, ng2_1}, nil, now)
+	assert.NoError(t, err)
+	assert.Equal(t, 1, clusterstate.GetClusterReadiness().NotStarted)
```
MaciekPytel, line 377: nit: Below you assert on both notStarted and ready. Maybe do it here too for consistency?

cluster-autoscaler/utils/kubernetes/ready_test.go
```diff
@@ -0,0 +1,75 @@
...
+				node.Spec.Taints = []apiv1.Taint{{
+					Key:       tc.taintKey,
+					Effect:    apiv1.TaintEffectNoSchedule,
+					TimeAdded: &metav1.Time{Time: testedTime},
+				}}
+			}
+
+			return node
+		}
+		t.Run("recent "+tc.desc, func(t *testing.T) {
```
MaciekPytel, line 68: nit: remove "recent" from description? it doesn't really apply in this context

MaciekPytel:
/lgtm
/approve
/hold

Left a few nits. Feel free to remove hold after addressing those.

cluster-autoscaler/utils/kubernetes/ready_test.go
vivekbagade: Done

cluster-autoscaler/clusterstate/clusterstate_test.go
vivekbagade: Done

MaciekPytel:
/lgtm
/hold cancel

Thanks, that was a tricky one.

PR #3926, opened 2021-03-07
Allow generic labels to be overwritten by tags coming from the ASG

sylr:
This simple patch allowed me to scaled ASG from 0 to X for a deployment having a NodeSelector label which is part of the ones `buildGenericLabels()` sets.

In my case it was `beta.kubernetes.io/arch`.


gjtempleton:
Nice find, thanks for the PR!

/lgtm
/approve

PR #3933, opened 2021-03-09
updating the vendor for new release tag `v1.21.0-beta.1`

jayantjain93:


cluster-autoscaler/go.mod
```diff
@@ -10,30 +10,32 @@ require (
 	github.com/Azure/go-autorest/autorest/date v0.3.0
```
towca, line 10: The go version is not updated by the `update-vendor.sh` script, even though it should be. I'll fix the script, but for this PR could you manually bump the go version in `go.mod` to 1.16, to keep it in line with k8s?

towca (changes requested):
Just one thing to change, otherwise LGTM.

One nit for the future: keeping the actual cluster-autoscaler changes (i.e. the scheduler-framework-related changes here) in a separate commit makes them much easier to review (now and in the future in `git log`). I know it was just a couple of lines for this update, but worth keeping in mind for the future.

cluster-autoscaler/go.mod
jayantjain93: Some of the changes didn't add properly, I have rerun.

towca (approved):

PR #3939, opened 2021-03-12
Support for reporting authorization errors during scale up

brett-elliott:


cluster-autoscaler/core/scale_up.go
```diff
@@ -666,7 +666,11 @@ func executeScaleUp(context *context.AutoscalingContext, clusterStateRegistry *c
 	increase := info.NewSize - info.CurrentSize
 	if err := info.Group.IncreaseSize(increase); err != nil {
 		context.LogRecorder.Eventf(apiv1.EventTypeWarning, "FailedToScaleUpGroup", "Scale-up failed for group %s: %v", info.Group.Id(), err)
-		clusterStateRegistry.RegisterFailedScaleUp(info.Group, metrics.CloudProviderError, now)
+		reason := metrics.CloudProviderError
+		if aerr, ok := err.(errors.AutoscalerError); ok && aerr.Type() == errors.AuthorizationError {
```
MaciekPytel, line 670: nit: I think it's more elegant to hide the type assertion using the functions in our errors library - we can replace NewAutoscalerError that we do few lines below with ToAutoscalerError and AddPrefix. ToAutoscalerError internally does the type assertion and it will preserve original error type, so we can just use that. WDYT?

MaciekPytel: After some more thought on this - a lot of FailedScaleUpReasons match 1-1 to error types. Maybe it would be nice to add a util function (in metrics.go?) to do a convertion from AutoscalerError to FailedScaleUpReason? This way we could actually correctly report other types of errors too.

PR #3940, opened 2021-03-12
Release leader election lock on shutdown

mcristina422:
```go
        // ReleaseOnCancel should be set true if the lock should be released
	// when the run context is cancelled. If you set this to true, you must
	// ensure all code guarded by this lease has successfully completed
	// prior to cancelling the context, or you may have two processes
	// simultaneously acting on the critical path.
```


mwielgus (approved):
/lgtm
/approve

PR #3951, opened 2021-03-17
Update description of CA in main README

jacksgt:
Just mentioning AWS, Azure and GCP makes it sounds like only those are supported.
Now however, many more platforms are supported.


mwielgus (approved):
/lgtm
/approve

PR #3954, opened 2021-03-18
adds status-config-map-name to values.yaml of helm chart

joaovitor:
Supplementary documentation to help anyone using the #3888 functionality

PR #3960, opened 2021-03-19
Fix typo in values.yaml example for extraVolumeMounts

Kixa:
Incredibly minor, but there's a typo in the Charts default values; the field for volume mounts when mounted read only is `readOnly`, not `readonly`, so if you directly uncomment it, `helm install` will fail. This fixes that.

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -17,4 +17,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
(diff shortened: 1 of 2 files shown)


Kixa:
@gjtempleton Happy to bump the helm chart version if you want, but since this is pretty much a documentation change, I didn't think it would need it.

gjtempleton:
Hey, thanks for the PR, good catch.

As this is just a fix in an example I'm happy to bundle this in with another change ( #3954 ) and let the version bump in that take care of publishing a release to include both docs improvements.

/lgtm
/approve

gjtempleton:
Thanks!
/lgtm
/approve

PR #3962, opened 2021-03-21
Allow users to change thresholds for evicting pods in VPA

maksim-paskal:
Allow users to change thresholds for evicting pods that use resources lower that recomended. Some pods needs to lower their request less than 12h.

This changes only allow users to change hardcoded values

New flags that will added with this PR
| Name | Default | Usage |
| --- | --- | --- |
| `pod-update-threshold` | 0.1 | Ignore updates that have priority lower than the value of this flag |
| `in-recommendation-bounds-eviction-lifetime-threshold` | 12h | Pods that live for at least that long can be evicted even if their request is within the [MinRecommended...MaxRecommended] range |

### DIFF
--- a/vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go
+++ b/vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go
@@ -29,15 +29,11 @@ import (
 	"k8s.io/klog"
 )
 
-const (
-	// Ignore change priority that is smaller than 10%.
-	defaultUpdateThreshold = 0.10
-	// Pods that live for at least that long can be evicted even if their
-	// request is within the [MinRecommended...MaxRecommended] range.
-	podLifetimeUpdateThreshold = time.Hour * 12
-)
-
 var (
+	defaultUpdateThreshold = flag.Float64("pod-update-threshold", 0.1, "Ignore updates that have priority lower than the value of this flag")
+
+	podLifetimeUpdateThreshold = flag.Duration("in-recommendation-bounds-eviction-lifetime-threshold", time.Hour*12, "Pods that live for at least that long can be evicted even if their request is within the [MinRecommended...MaxRecommended] range")
+
 	evictAfterOOMThreshold = flag.Duration("evict-after-oom-threshold", 10*time.Minute,
 		`Evict pod that has only one container and it OOMed in less than
 		evict-after-oom-threshold since start.`)
@@ -72,7 +68,7 @@ func NewUpdatePriorityCalculator(vpa *vpa_types.VerticalPodAutoscaler,
 	recommendationProcessor vpa_api_util.RecommendationProcessor,
 	priorityProcessor PriorityProcessor) UpdatePriorityCalculator {
 	if config == nil {
-		config = &UpdateConfig{MinChangePriority: defaultUpdateThreshold}
+		config = &UpdateConfig{MinChangePriority: *defaultUpdateThreshold}
 	}
 	return UpdatePriorityCalculator{
 		vpa:                     vpa,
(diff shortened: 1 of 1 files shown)


maksim-paskal:
I signed it

vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go
```diff
@@ -29,12 +29,13 @@ import (
 	"k8s.io/klog"
 )
 
-const (
+var (
 	// Ignore change priority that is smaller than 10%.
-	defaultUpdateThreshold = 0.10
+	defaultUpdateThreshold = flag.Float64("priority-update-threshold", 0.10, "Ignore change priority that is smaller this value")
```
jbartosik, line 34: Please make the flag description more descriptive.

Ignore updates that have priority lower than the value of this flag.

Short explanation of what is update priority would be useful too.

vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go
```diff
@@ -29,12 +29,13 @@ import (
...
+
 	// Pods that live for at least that long can be evicted even if their
 	// request is within the [MinRecommended...MaxRecommended] range.
-	podLifetimeUpdateThreshold = time.Hour * 12
+	podLifetimeUpdateThreshold = flag.Duration("priority-lifetime-threshold", time.Hour*12, "Pods that live for at least that long can be evicted even if their request is within the [MinRecommended...MaxRecommended] range")
```
jbartosik, line 38: I think current name of this flag could be more descriptive. What do you think about `in-recommendation-bounds-eviction-lifetime-threshold`?

vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go
```diff
@@ -29,12 +29,13 @@ import (
...
 )
 
 var (
```
jbartosik, line 41: Please merge those two `var` sections

vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go
maksim-paskal: I think prefix `in-recommendation` is unnecessary

what do you think about this names with pod prefix?

| Name | Default | Usage |
| --- | --- | --- |
| `pod-update-threshold` | 0.1 | Ignore updates that have priority lower than the value of this flag |
| `pod-eviction-lifetime-threshold` | 12h | Pods that live for at least that long can be evicted even if their request is within the [MinRecommended...MaxRecommended] range |

jbartosik: I think that the `in-recommendatiotion-bounds` part is necessary (we will evict younger pods, if they are out of the [MinRecommended...MaxRecommended]  interval). 

`pod-eviction-lifetime-threshold` sounds like we wouldn't evict younger pods no matter what to me

maksim-paskal: ok - finally we got this flags. Ok?

| Name | Default | Usage |
| --- | --- | --- |
| `in-recommendation-update-threshold` | 0.1 | Ignore updates that have priority lower than the value of this flag |
| `in-recommendation-bounds-eviction-lifetime-threshold` | 12h | Pods that live for at least that long can be evicted even if their request is within the [MinRecommended...MaxRecommended] range |

jbartosik: Sounds good. Please update the PR

maksim-paskal: PR updated

jbartosik: Looks like I didn't explain myself clearly enough, sorry about that. I think that we need the `in-recommendation-bounds` prefix for the flag that controls after how long VPA will evict pods which have requests in the  `\[`MinRecommended`...`MaxRecommended`] interval.

We don't need the prefix for the flag which controls priority threshold, below which we will ignore updates.


So I'm suggesting

| Name | Default | Usage |
| --- | --- | --- |
| pod-update-threshold | 0.1 | Ignore updates that have priority lower than the value of this flag |
| in-recommendation-bounds-eviction-lifetime-threshold | 12h | Pods that live for at least that long can be evicted even if their request is within the [MinRecommended...MaxRecommended] range|

maksim-paskal: LGTM - PR updated

jbartosik (approved):
/lgtm
/approve

Thank you

Issue #3873, opened 2021-02-04
Add support for other kinds of Daemonset Kube Resource (CRD)


clamoriniere:
**component: cluster-autoscaler**

## Issue description

The cluster-autoscaler uses the Daemonsets spec to simulate the resources needed for the Daemonset’s pods on a new node.
We are looking  to extend the capability of the cluster-autoscaler in terms of watching for other resources that implement an equivalent of a Daemonset (one Pod deployed on each Node). 

We have the following established open-source project [github.com/DataDog/ExtendedDaemonset](https://github.com/DataDog/ExtendedDaemonset) that defines a CRD (ExtendedDaemonset) that replaces the Daemonset object for some of our applications that need to be deployed on all nodes. One of the reasons that we have worked on this project is to allow a canary deployment on a subset of nodes,as well as have more control over the rollout of an application (velocity, pausing…).

We made this previous [contribution](https://github.com/kubernetes/autoscaler/pull/2483) that allows the cluster-autoscaler controller to recognize the Pods created by the ExtendedDaemonset as Daemonset’s Pod.

Now the idea is to provide a way for the controller to consider other kinds of Daemonset resources when it estimates the Daemonsets’ pod resources usage in the [GetDaemonSetPodsForNode() function](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/utils/daemonset/daemonset.go#L31)

## Proposed solutions

To make it possible, we identified two ways to do it:

1. Add the possibility for the cluster-autoscaler controller to look at `PodTemplates` resources in addition to the Daemonsets. To avoid querying PodTemplates that are not linked to our use case, we can limit the query with a LabelSelector on a dedicated label like `cluster-autoscaler.kubernetes.io/daemonset-pod`. Then we can simulate a pod creation like for the Daemonset.

2. Add the possibility to list any kind of resource that contains a PodTemplate in its spec section.

   We can imagine providing a list of resource metadata (kind, apiVersion) as controller parameters. Attach to each resource, also the path to the PodTemplate in the spec section.

   ```yaml
   daemonsetResources:
   - kind: CustomeDaemonset
     apiVersion: foo.io/v1
     podTemplate:
       path: .spec.template
       kind: PodTemplate
       apiVersion: v1
   ```

   Then, using the dynamic goclient, we can query the api-server to get the list of resources, and then decoding the object, access the PodTemplate section.

## Questions

* What do you think about this feature?

[...]

PR #3964, opened 2021-03-22
Add PodTemplates support as NodeInfoProcessor in Cluster-Autoscaler

clamoriniere:
This Pull request is implementing feature request #3873

Solution1 describe in #3873 has been implemented. Base on the feedback and advice received during the last [SIG meeting: March 15 2021](https://docs.google.com/document/d/1RvhQAEIrVLHbyNnuaT99-6u9ZUMp7BfkPupT2LAZK7w/edit#heading=h.oh2koj9sbr3x), the feature has been wrapped inside a `Processor` and disabled by default.

Thanks to this feature, it is possible to provide to the cluster-autoscaler, `PodTemplates` that will be considered as `Daemonset`.

## Changes:

* [e29a8f1](https://github.com/kubernetes/autoscaler/pull/3964/commits/e29a8f16b266d35c1af11ffab146a6cb8a9e29b4): Initial implementation Add a new processor:
   `PodTemplatesProcessor` which aims to simulate Daemonset instances from 
   PodTemplates. This implementation extended the Daemonset list used during scale-up operations.
   * Add new `podTemplates` processor with 2 implementations: `noOpPodTemplateListProcessor ` (default) and
     `activePodTemplateListProcessor`. 
   * Update `cluster-autoscaler` helm chart with `corev1.PodTemplate` required RBAC.

* [6294791](https://github.com/kubernetes/autoscaler/pull/3964/commits/6294791642f2de5bed5a4768bd502d98cde4f0b3): Reimplement PodTemplateProcessor
  behind the `NodeInfoProcessor` interface. The new `nodeInfoWithPodTemplateProcessor` updates for each `NodeInfo` the `NodeInfo.Pods` with `Pods` generated from the `PodTemplate`.

  Thanks to this code structure it will be possible to enable `NodeInfoProcessor` implementations behind build flags.


MaciekPytel:
What I had in mind during the sig-meeting was to implement a NodeInfoProcessor that would inject those pods directly into template nodes. 

My reasoning is that NodeInfoProcessor allows all sort of modifications to template nodes or pods that run on them. The new processor introduced in this PR seems tailored to this single use-case, handling custom DS controllers is the only reason I can think of for modifying the list of DS. I'd argue that this defeats the point of having a processor:
 * The annotation alone seems to give enough flexibility to handle custom DS controllers;
 * The interface is not generic enough to cover other use-cases.

Is there some reason why injecting into list of daemonsets would work better than injecting directly into template nodes? Or is there some other possible use-case where this processor could be useful that I'm missing?

clamoriniere:
Hi @MaciekPytel 

Sorry, I didn't fully understood your recommendation to use the `NodeInfoProcessor` for this use case. I had focused on the `Processor` pattern only. I will improve my change in this direction.

> Is there some reason why injecting into the list of daemonsets would work better than injecting directly into template nodes? Or is there some other possible use-case where this processor could be useful that I'm missing?

I focused too much on my use case which is to recognize other resources as daemonsets. That is why updating the Daemonsets list seemed to be a good approach to not modify the workflows that use this list.

But I can definitely work on a solution that implements the `NodeInfoProcessor` interface and create pods in the `NodeInfo.Pods` slice.

MaciekPytel:
Thanks. Just for the record - I'm not saying I'm 100% certain that NodeInfoProcessor based solution is the best one, just that I'd like to consider it as an option and, if it's not good enough understand why that is. I think it should cover scale-up (though I may be missing something).
It may be worth looking into scale-down though - we handle DS pods differently when draining nodes and that logic has nothing to do with template nodes or NodeInfoProcessor. I think the question is what the desired behavior is in that case and whether we need a change that goes beyond template nodes to achieve it.

clamoriniere:
@MaciekPytel 
I agree with you. I will update my PR with `NodeInfoProcessor` approach. 

I have a question about the `NodeInfoProcessor` if I understand properly (and correct me if I'm wrong) the current approach is to only have one instance of the `NodeInfoProcessor` at the time (like for other processor). But could we imaging to have severals `NodeInfoProcessor` at the same time? For example several extra pods sources; In this case run sequentially the different `NodeInfoProcessor` implementations.

For the scale down. I think I have already cover the needs in https://github.com/kubernetes/autoscaler/pull/2483. 
The Daemonset list that the `podTemplates` processor was modifying is only used for scale up.

MaciekPytel:
Regarding running multiple processors:
 * In many cases ordering is very important. In my experience if we have 2 processors A and B it's safer to assume A(B(x)) != B(A(x)).
 * I don't think we have enough processors to justify refactoring processors to lists of processors or similar. If needed it's easy enough to just write a processor that wraps some other processors and calls them.
   * A processor that just holds a list of processors and calls them in sequence is a simple generic solution, but I found it starts to be pretty hard to reason about it once you have 3+ processors that you want to call. I now think having a dedicated processor that calls a specific sequence of other processors is actually a better solution. Given the importance in ordering I mentioned in my first comment I think there is little value in a generic solution and having a specialized processor that chains a set of processors in a way that addresses a given use-case is the way to go.

clamoriniere:
Hi @MaciekPytel 

I have updated the PR based on our previous discussion. I added two new commits:

* [0356b02](https://github.com/kubernetes/autoscaler/pull/3964/commits/0356b02cd87751a8fc7a4a16fe89843959451314): Reimplement PodTemplateProcessor
  behind the `NodeInfoProcessor` interface. The new `nodeInfoWithPodTemplateProcessor` updates for each `NodeInfo` the `NodeInfo.Pods` with `Pods` generated from the `PodTemplate`.

* [3d0c914](https://github.com/kubernetes/autoscaler/pull/3964/commits/3d0c914f7f1e1bd158689c962b1fad15cf0e213c): Add new package k8s.io/autoscaler/cluster-autoscaler/processors/nodeinfos/builder which contains a "NodeInfoProcessor" factory:
  * `NodeInfoProcessor` Implementations can register into the factory with the `builder.Register()` function.
  * `builder.Build()` function is used to instantiate the `NodeInfoProcessor` implementation base on `AutoscalerOptions`.
  
  Thanks to this code structure it will be possible to enable `NodeInfoProcessor` implementations behind build flags.

cluster-autoscaler/processors/nodeinfos/podtemplates/node_info_with_podtemplates_processors.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func getNodeInfoWithPodTemplates(baseNodeInfo *schedulerframework.NodeInfo, podTemplates []*apiv1.PodTemplate, clusterSnapshot *simulator.BasicClusterSnapshot, predicateChecker simulator.PredicateChecker) (*schedulerframework.NodeInfo, error) {
+	node := baseNodeInfo.Node()
+	var pods []*apiv1.Pod
+
+	for _, podInfo := range baseNodeInfo.Pods {
+		pods = append(pods, podInfo.Pod)
+	}
+	if err := clusterSnapshot.AddNodeWithPods(node, pods); err != nil {
```
elmiko, line 125: this seems like a side effect of calling this function, i would expect from the name that it just returns the nodeinfo with the pod templates. why do we add the node/pods to the cluster snapshot here?

elmiko (commented):
i am not overly familiar with the processors code, but this generally makes sense to me. i did have one question though.

cluster-autoscaler/processors/nodeinfos/podtemplates/node_info_with_podtemplates_processors.go
clamoriniere: Thanks @elmiko  for the review.

I used the same logic that is present in the Daemonset support [here](https://github.com/kubernetes/autoscaler/blob/2542e8c884a8e25634d3b8f43243fa8706007f30/cluster-autoscaler/utils/daemonset/daemonset.go#L43) 
I need to check more deeply the purpose of this function.

elmiko: that makes sense following the daemonset pattern. i was mainly asking because i don't this code that well and was curious, i appreciate any further findings you might share =)

clamoriniere: Hello @elmiko,

Sorry for the late reply,

I checked why it is used like this in the `GetDaemonSetPodsForNode()` function. 

So for what I understand, it is because in this specific case the `Node` doesn't exist yet (we are in the edgecase where we scale from 0), and we don't want to add this Node the `context.AutoscalingContext.ClusterSnapshot` yet. 

The other solution could have been to use "context.AutoscalingContext.ClusterSnapshot.Fork()" but since we are not interested by the others `NodeInfo` here, creating a new `ClusterSnapshot` with `simulator.NewBasicClusterSnapshot()` limit the memory copy.

clamoriniere:
Hi @MaciekPytel,

Would it be possible to make a new review of this PR 🙏 . 

Thanks

cluster-autoscaler/processors/nodeinfos/podtemplates/node_info_with_podtemplates_processors.go
elmiko: thanks @clamoriniere ! i appreciate the extra context =)

elmiko (commented):
this is looking good to me, i have one minor suggestion but it's not a blocker.
/lgtm

cluster-autoscaler/processors/nodeinfos/podtemplates/node_info_with_podtemplates_processors.go
MaciekPytel: Using an empty snapshot implicitly makes nodes that are already in the clusters invisible to scheduling, in other words you implicitly assume such pods wouldn't impact the result of scheduling. There are some scheduling constraints (podAffinity, podTopologySpreading) that look at multiple nodes. Admittedly using any of those on DS or DS-like pod seems like a really bad idea, but technically I think it's more correct to go with Fork/Revert route. 

Also - Fork/Revert is used across codebase, but I don't think we actually use BasicClusterSnapshot anywhere in CA. So you're exposing yourself to code that is not tested in production.

cluster-autoscaler/processors/nodeinfos/podtemplates/node_info_with_podtemplates_processors.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil, err
+	}
+	// clone baseNodeInfo to not modify the input object.
+	newNodeInfo := baseNodeInfo.Clone()
```
MaciekPytel, line 142: DeepCopyTemplateNode from utils/scheduler? IIRC there were some issues with nodeInfo.Clone() being shallower copy than expected.

cluster-autoscaler/processors/nodeinfos/podtemplates/node_info_with_podtemplates_processors.go
clamoriniere: Hi @MaciekPytel 
thanks for the details. 
The edge-case that I would like to solve is the "scale from 0", for which the cluster-autoscaler is looking at Daemonset to simulate an existing Node. Thanks is why in this case, knowing the other existing Nodes doesn't seems useful. It is what I understood from code use to generate Pod based on Daemonset https://github.com/kubernetes/autoscaler/blob/2542e8c884a8e25634d3b8f43243fa8706007f30/cluster-autoscaler/utils/daemonset/daemonset.go#L43 which also use an empty snapshot. I thought it was an optimisation. 

But I don't see any reason why I can't change PodTemplate NodeInfoProcessor to use a Fork.

cluster-autoscaler/processors/nodeinfos/podtemplates/node_info_with_podtemplates_processors.go
```diff
@@ -0,0 +1,172 @@
...
+		PodTemplateDaemonSetLabelKey: PodTemplateDaemonSetLabelValueTrue,
+	}
+	return labels.Set(daemonsetPodTemplateLabels)
+}
+
+func newPod(pt *apiv1.PodTemplate, nodeName string) *apiv1.Pod {
+	newPod := &apiv1.Pod{Spec: pt.Template.Spec, ObjectMeta: pt.Template.ObjectMeta}
+	newPod.Namespace = pt.Namespace
+	newPod.Name = fmt.Sprintf("%s-pod", pt.Name)
+	newPod.Spec.NodeName = nodeName
```
MaciekPytel, line 170: Setting nodeName means the pod is already scheduled on a given node. Here you want to check if it would be able to schedule, so I don't think it makes sense to set this field.

cluster-autoscaler/processors/nodeinfos/podtemplates/node_info_with_podtemplates_processors.go
```diff
@@ -0,0 +1,172 @@
...
+
+	result := make(map[string]*schedulerframework.NodeInfo, len(nodeInfosForNodeGroups))
+	var errs []error
+	for id, nodeInfo := range nodeInfosForNodeGroups {
+		var newNodeInfo *schedulerframework.NodeInfo
+
+		// only runs getNodeInfoWithPodTemplates() for NodeInfo generated from NodeTemplate
+		// If not a NodeTemplate, Pods from PodTemplates should alreadt be already present
+		// in the PodList attached to the Node.
+		if strings.HasPrefix(nodeInfo.Node().Name, "template-node-for-") {
```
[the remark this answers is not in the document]
MaciekPytel: I'm not sure I understand what this check is trying to achieve. nodeInfosForNodeGroups only contain 'template nodes' that will be used to represent potential new nodes, nodeInfos that represent actual nodes in the cluster live in ClusterSnapshot.

Or do you mean to only trigger the logic for template nodes created using cloudprovider.TemplateNodeInfo and not by copying another node? In this case I don't think this check will work: quickly looking at the code 'template-node-for-' name is set in sanitizeTemplateNode which seems to be called (via sanitizeNodeInfo) both on the path where node is created from template and one where it's created by copying existing node.

cluster-autoscaler/main.go
```diff
@@ -179,6 +181,7 @@ var (
 	cordonNodeBeforeTerminate          = flag.Bool("cordon-node-before-terminating", false, "Should CA cordon nodes before terminating during downscale process")
 	daemonSetEvictionForEmptyNodes     = flag.Bool("daemonset-eviction-for-empty-nodes", false, "DaemonSet pods will be gracefully terminated from empty nodes")
 	userAgent                          = flag.String("user-agent", "cluster-autoscaler", "User agent used for HTTP calls.")
+	nodeInfoProcessorName              = flag.String("node-info-processor", "", "node-info processor type. Available values: ["+strings.Join(nodeinfobuilder.GetAvailableNodeInfoProcessors(), ",")+"]")
```
MaciekPytel, line 184: This is solving the problem of processor configuration half-way. 

My thinking is that processors are intended as a way to implement support for advanced use-cases (that are likely relevant to a very few users) without further complicating core. Historically the assumption was that they would mostly be implemented in forks, but there is a growing interest in adding them to the main repo. As the number of processors grows the reviews will increasingly be a problem (this is exactly what happened with cloudproviders).  

I think we ultimately want to solve it the same way we did for providers - create a directory and start delegating ownership of individual processors. I'd like to make sure any solution for managing processors is going towards this ultimate goal. Some thoughts on this:
 * We want a consistent way to manage all processors, not just NodeInfoProcessor.
 * We need a way to compile CA with only relevant processors, similarly to how it can be compiled with just a single provider. 
 * I'm not sure I like the idea of dynamic registration via init(). Having a static configuration seems like a good balance - any new processor would have to be initially reviewed by core maintainer (owner of static config) to verify the general idea, but following that initial review the processor would be managed by its owner. Once again, this is basically the same model we use for cloudproviders (cloudprovider/builder plays that role there). 

--- 

Solving the general problem of processors obviously goes far beyond the scope of this PR. If you want to volunteer to tackle it I'm happy to help, but I'm obviously not going to ask you to do it. However, I'd rather not include a partial solution that I don't think is going in the direction we want to ultimately go to. 
For now maybe just add a specific flag for this feature? Something like --add-pod-info-to-template-nodes and just use it in a simple if statement in main.go.

elmiko: > Historically the assumption was that they would mostly be implemented in forks, but there is a growing interest in adding them to the main repo. 

i did not realize this, thanks for the clarification.

> I think we ultimately want to solve it the same way we did for providers - create a directory and start delegating ownership of individual processors.

this makes a lot of sense to me. the unfortunate side effect is that this PR would slow down a little until we can agree on the subpoints, but i think it would be nice to have this for future node processors.

MaciekPytel: Yeah, I don't think it's fair to block this PR until we figure out the details. What I think we should do is merge the processor part of this PR as-is (modulo review comments) and move it to this new directory once we set it up.

clamoriniere: Hi @MaciekPytel and @elmiko

Thanks for the detailed and constructive reviews.

I agree with you the problem of processor configuration is “halfway” solved, I am glad we are on the same page and I am happy to implement a more generic feature.

What do you think about the following approach?

> We want a consistent way to manage all processors, not just NodeInfoProcessor.

Would you accept extending the “factory/builder” idea to other Processor interfaces? If each Processor has a unique ID, each Process implementation can register a builder function that will be executed if the user wants to enabled it.

> I'm not sure I like the idea of dynamic registration via init().

I also agree with you that we should be able to compile the CA with only relevant processors, my approach here with the `func init()` was to invert the registration logic. With to this pattern, all the implementation files of my Processors are in a dedicated `package`. Then, if we add `go build tags` on each processor implementation, the user would be able to control which processor implementation they want to include in their build.

In the `main.go` file, the only reference to the processor package implementation would be:

```golang
import (
    //...
    _ "k8s.io/autoscaler/cluster-autoscaler/processors/nodeinfos/podtemplates"
)
```

and the flag will only contains options that was part of the build

```golang
flag.String("node-info-processor", "", "node-info processor type. Available values: ["+strings.Join(nodeinfobuilder.GetAvailableNodeInfoProcessors(), ",")+"]")
```

> We need a way to compile CA with only relevant processors, similarly to how it can be compiled with just a single provider.

That's interesting, I don't have a strong opinion and I am happy to use the same pattern as in the `cloudprovider/builder package`. If we still want to support the fact that we can remove some implementation at build time, we can define two build functions in 2 different files, with different build tags.

```golang
// +build podtemplateprocessor

package podtemplateprocessor

func BuildProcessor(opts *core.AutoscalerOptions) NodeInfoProcessor {
    // Real implementation
}
```

```golang
// +build !podtemplateprocessor

package podtemplateprocessor

func BuildProcessor(opts *core.AutoscalerOptions) NodeInfoProcessor {
    return nil
}
```

and then 

```golang
package nodeinfos

// Build returns a new nodeinfos.NodeInfoProcessor instance with the requested implementation
func BuildNodeInfoProcessor(opts *core.AutoscalerOptions) (nodeinfos.NodeInfoProcessor, error) {
    switch opts.NodeInfosProcessorName {
    case podtemplateprocessor.Name
        return podtemplateprocessor.BuildProcessor(opts)
    }
    return nil
}

```

I can create an issue to start a separate discussion about this topic. I am more than happy to work on it, basically build this framework to register and build processors, merge it and then rebase and rework this PR to implement our use case.

let my know what you think.

elmiko: > Would you accept extending the “factory/builder” idea to other Processor interfaces? If each Processor has a unique ID, each Process implementation can register a builder function that will be executed if the user wants to enabled it.

i don't have an objection, but i would like to clarify a little. when you say factory/builder, are you saying that the factory could build a node processor given the unique ID? (eg make me a Foo processor)

> I also agree with you that we should be able to compile the CA with only relevant processors, my approach here with the func init() was to invert the registration logic. ...

your logic generally makes sense to me, but i am also in the "not liking init()" camp. perhaps we could follow the metrics approach and just have a registration function that gets called once from main?

> That's interesting, I don't have a strong opinion and I am happy to use the same pattern as in the cloudprovider/builder package.  ...

i agree with @MaciekPytel about re-using the build tag pattern that we have for the cloud providers. that would be really nice for the node processors too.

thanks for taking the time to write this up and for driving it forward @CharlyF !

cluster-autoscaler/processors/nodeinfos/podtemplates/node_info_with_podtemplates_processors.go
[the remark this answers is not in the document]
clamoriniere: Thanks @MaciekPytel for all the information

> Or do you mean to only trigger the logic for template nodes created using cloudprovider.TemplateNodeInfo and not by copying another node?

Yes it is exactly what I wanted to achieve. Do you know a way to make a distinction between the 2 path (from template or from copy)?

charts/cluster-autoscaler/templates/clusterrole.yaml
```diff
@@ -48,6 +48,7 @@ rules:
     - ""
     resources:
       - pods
+      - podtemplates
```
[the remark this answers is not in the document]
clamoriniere: I have updated all the `ClusterRole` to add the "PodTemplates" resource

cluster-autoscaler/processors/nodeinfos/podtemplates/node_info_with_podtemplates_processors.go
clamoriniere: good to know.
I will use the `DeepCopyTemplateNode` instead.

cluster-autoscaler/processors/nodeinfos/podtemplates/node_info_with_podtemplates_processors.go
clamoriniere: Here I want to simulate a Pod that was/will be create by a Daemonset, it is why I force the scheduling on this specific node by setting the NodeName. I used the same logic used in `GetDaemonSetPodsForNode` with the `newPod` function ([link](https://github.com/kubernetes/autoscaler/blob/a839343bbd737534133d110e7decf295f8281baa/cluster-autoscaler/utils/daemonset/daemonset.go#L68))

But it is true that I can use the  `pod.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution` to fully simulate what the daemonset-controller is doing.

mwielgus:
No activity, closing. Feel free to reopen if needed.

PR #3968, opened 2021-03-23
Proposal to circumvent 50 tag ASG limit for EKS ManagedNodegroups

MyannaHarris:
This proposal is related to this issue that AWS EKS has open: https://github.com/aws/containers-roadmap/issues/724

The proposal was originally in kubernetes/enhancements. See the previous feedback cycle here: https://github.com/kubernetes/enhancements/pull/2561

@gjtempleton 
@jaypipes

### DIFF
--- a/cluster-autoscaler/proposals/circumvent-tag-limit-aws.md
+++ b/cluster-autoscaler/proposals/circumvent-tag-limit-aws.md
@@ -1,0 +1,142 @@
+# Circumvent 50 tag ASG limit for EKS ManagedNodegroups
+##### Author: MyannaHarris
+
+## Table of Contents
+
+<!-- toc -->
+- [Summary](#summary)
+- [IAM](#iam)
+- [Design](#design)
+- [Implementation Proposal](#implementation-proposal)
+- [Testing](#testing)
+- [Risks](#risks)
+- [Other Solutions Considered](#other-solutions-considered)
+<!-- /toc -->
+
+## Summary
+
+Currently an EC2 Autoscaling group can only have 50 tags. Many tags are already added to the ASG by standard components like the AWS cloudprovider for Kubernetes and by customers for billing and cost association purposes. Adding labels and taints to the ASG as tags will run into this 50 tag limit. The primary focus of this proposal is to get around the 50 tag limit for customers scaling to/from 0 nodes using Cluster Autoscaler on AWS EKS ManagedNodegroups in a way that will not limit the ManagedNodegroups service.
+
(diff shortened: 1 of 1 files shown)


gjtempleton:
Nice work! All of my comments from the previous discussion in kubernetes/enhancements#2561 have been answered/addressed, so:
/lgtm

cluster-autoscaler/proposals/circumvent-tag-limit-aws.md
```diff
@@ -0,0 +1,142 @@
...
+- [TC3] Test pulling values from cache and calling DescribeNodegroup API if no data exists
+- [TC4] Test pulling values from cache and calling DescribeNodegroup API if TTL has expired
+- [TC5] Test missing permissions and falling back to just ASG tags
+- [TC6] Test duplicate label keys - selecting the value from ASG tags response
+
+## Risks
+
+Latency and throttling:
+
+By default, Cluster Autoscaler runs every 10 seconds. Our best practices documentation notes that this short interval can cause throttling because Cluster Autoscaler already makes AWS API calls during each loop. Our documentation recommends that customers increase the interval, so adding this API shouldn’t cause latency problems for customers. ([EKS Best Practices][]) Also, the API call will only happen for EKS ManagedNodegroups. If this increase in latency is too much for even one run of the loop, we will look into moving the API calls into parallel goroutines.
```
elmiko, line 130: i'm just curious if there is precedent for using parallel goroutines in cluster-autoscaler to do similar types of API calls?

elmiko (commented):
i am not intimately familiar with the AWS provider code, but this proposal seems reasonable to me. i have one minor question, but it's more to help my understanding.

MaciekPytel:
/approve

Sorry for delay on approving this. The proposal is strictly specific to AWS and @gjtempleton is an owner of AWS cloudprovider, so I think his opinion (and not mine) is really the one that is relevant here. That being said this proposal looks very reasonable to me.

cluster-autoscaler/proposals/circumvent-tag-limit-aws.md
MyannaHarris: Thank you for taking a look at my proposal.
There's no precedent for using parallel goroutines to do similar types of API calls in the AWS cloud-provider code. I'm less familiar with the rest of the code, but I don't think there's a precedent in it either.

MaciekPytel: There is a precedent in GCE provider: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/gce/cache.go#L277.

PR #3971, opened 2021-03-23
Separate and refactor custom resources logic

BigDarkClown:
Currently the logic regarding custom resources (other than cpu and memory) in Cluster Autoscaler is limited to GPU resources. Because nodes need to install GPU drivers, the GPU resources may not become allocatable immediately after the node creation. This required implementing  additional functionality (GPU hack) to better handle this case, predicting the type/count of GPUs in case the drivers were not yet installed.

The refactor was suggested because:
1. This logic is dependent on the underlying implementation of driver installation in the cluster, so it would be useful to have it fine-tunable for specific cases. This was done by creating a separate processor and moving the current implementation there.
2. In the future it may be applied for other custom resources, thus we would greatly benefit from generalisation of the underlying code. This was done by removing the dependency on GPU constants outside of the implementation.

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -40,7 +40,6 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/utils/backoff"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/deletetaint"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
-	"k8s.io/autoscaler/cluster-autoscaler/utils/gpu"
 	scheduler_utils "k8s.io/autoscaler/cluster-autoscaler/utils/scheduler"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/taints"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/tpu"
@@ -719,7 +718,7 @@ func (a *StaticAutoscaler) obtainNodeLists(cp cloudprovider.CloudProvider) ([]*a
 	// Treat those nodes as unready until GPU actually becomes available and let
 	// our normal handling for booting up nodes deal with this.
 	// TODO: Remove this call when we handle dynamically provisioned resources.
-	allNodes, readyNodes = gpu.FilterOutNodesWithUnreadyGpus(cp.GPULabel(), allNodes, readyNodes)
+	allNodes, readyNodes = a.processors.CustomResourcesProcessor.FilterOutNodesWithUnreadyResources(a.AutoscalingContext, allNodes, readyNodes)
 	allNodes, readyNodes = taints.FilterOutNodesWithIgnoredTaints(a.ignoredTaints, allNodes, readyNodes)
 	return allNodes, readyNodes, nil
 }
--- a/cluster-autoscaler/processors/customresources/custom_resources_processor.go
+++ b/cluster-autoscaler/processors/customresources/custom_resources_processor.go
@@ -1,0 +1,48 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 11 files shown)


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -719,7 +719,8 @@ func (a *StaticAutoscaler) obtainNodeLists(cp cloudprovider.CloudProvider) ([]*a
 	// Treat those nodes as unready until GPU actually becomes available and let
 	// our normal handling for booting up nodes deal with this.
 	// TODO: Remove this call when we handle dynamically provisioned resources.
-	allNodes, readyNodes = gpu.FilterOutNodesWithUnreadyGpus(cp.GPULabel(), allNodes, readyNodes)
+	allNodes, readyNodes = a.processors.CustomResourceProcessor.FilterOutNodesWithUnreadyResource(a.AutoscalingContext,
+		gpu.ResourceNvidiaGPU, cp.GPULabel(), allNodes, readyNodes)
```
MaciekPytel, line 723: I don't particularly like this, it feels like doing the refactor half-way - I still couldn't use a processor to support "GPU hack" for my own custom resource, because it's only ever going to be called for GPU. I think it would be better to dump the gpu related parameters from the method signature, so it looks like this:
```
FilterOutNodesWithUnreadyResource(a.AutoscalingContext, allNodes, readyNodes)
```

This way it's the implementation of the processor that decides what resources need to be initialized to consider the node ready. The default implementation could than be renamed to GpuCustomResourceProcessor and use ResourceNvidiaGPU label as well as GPULabel() provided by cloudprovider.

---

I think the difference in our approaches is what logic should be customizable by processor - your PR goes into "how GPUs are handled direction", but I don't think there are actually many different ways to handle this. I'd much rather assume the processor implements the current "GPU hack" or something very similar and allow customizing "what resource are we doing the hack for / how do we recognize those resources". I could even imagine the implementation of the processor not looking at resource request, but something else entirely - ex. we use basically the same "GPU hack" to handle "ignore-taints" now. With the processor implemented the way I describe that logic could be re-implemented as a combination of NodeInfoProcessor and CustomResourceProcessor without any explicit support in core (note: I'm not asking you to reimplement it, just using it as example :) ).

NodeGroupSetProcessor is a similar example with mulitple implementations that use fundamentally the same logic and only differ in details (what labels they look at, etc).

BigDarkClown: I took a long hard look at my implementation and I must agree that your approach would be superior here - it would be much more flexible and customisable. It will require a significant bit more refactoring of the context in which these processor methods are executed, but it is definitely doable.

Regarding `GetNodeTargetResource`, I think that the implementation returning something like a slice with triples `(resource, resourceType, resourceCount)`, eg. `("nvidia.com/gpu", "nvidia-tesla-k80", 2)` should be flexible enough to work in all current contexts - I would not want to "mix" our approaches into one PR, it would be better to follow yours strictly.

cluster-autoscaler/processors/customresources/custom_resources_processor.go
```diff
@@ -0,0 +1,145 @@
...
+	// GetNodeResourceTargets returns mapping of resource names to their targets.
+	GetNodeResourceTargets(context *context.AutoscalingContext, node *apiv1.Node, nodeGroup cloudprovider.NodeGroup) ([]CustomResourceTarget, errors.AutoscalerError)
+	// CleanUp cleans up processor's internal structures.
+	CleanUp()
+}
+
+// GpuCustomResourcesProcessor handles only the GPU custom resource. It assumes,
+// that the GPU may not become allocatable immediately after the node creation.
+// It uses additional hacks to predict the type/count of GPUs in that case.
+type GpuCustomResourcesProcessor struct {
```
MaciekPytel, line 51: nit: I'd slightly prefer moving this to a separate file (ex. customeresources/gpu.go).

MaciekPytel (commented):
Great stuff. Left some minor nits, but overall LGTM - it's a very clean refactor.

MaciekPytel:
/approve
(will lgtm after the nits are addressed)

PR #3972, opened 2021-03-24
fix: add missing call to fetch autodiscovered nodegroups

marwanad:
The refactor in https://github.com/kubernetes/autoscaler/pull/3717 missed adding this method to fetch and register auto-discovered groups.

/area provider/azure

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_cloud_provider_test.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_cloud_provider_test.go
@@ -135,6 +135,7 @@ func TestNodeGroupForNode(t *testing.T) {
 
 	registered := provider.azureManager.RegisterNodeGroup(
 		newTestScaleSet(provider.azureManager, "test-asg"))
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/cloudprovider/azure/azure_cloud_provider_test.go
```diff
@@ -135,6 +135,7 @@ func TestNodeGroupForNode(t *testing.T) {
 
 	registered := provider.azureManager.RegisterNodeGroup(
 		newTestScaleSet(provider.azureManager, "test-asg"))
+	provider.azureManager.explicitlyConfigured["test-asg"] = true
```
marwanad, line 138: this is needed because the manager setup logic skips `createAzureManagerInternal` and thus by the time `fetchAutoNodegroups` is called `explicitlyConfigured` will be empty and the nodegroup will be unregistered on L217 in azure_manager.

marwanad:
@nilo19 @feiskyer this should be good to go

PR #3998, opened 2021-04-08
cloudprovider: add Bizflycloud provider

lmq1999:
This adds a new cluster autoscaler for [Bizflycloud Kubernetes Engine](https://bizflycloud.vn/kubernetes-engine)

Add gobizfly, Bizflycloud API package at vendor


mwielgus (changes requested):
We don't accept any new dependencies. Please inline them in your cloud provider package (as others).

PR #4004, opened 2021-04-09
Addon resizer uses patch for deployment update

### LINKED ISSUE #3567: addon-resizer does not preserve new seccompProfile field
I tried using the GA level seccompProfile with Kubernetes v1.19 - removed the alpha annotation and added seccompProfile to the container securityContext for a metrics-server deployment with addon-resizer.

After resizing neither the seccompProfile nor the alpha seccomp annotation appear in the containers.

Kubernetes v1.19.2
addon-resizer:1.8.1​1

bskiba:
This is a permanent fix for addon resizer removing field values in deployments that it does not know about due to using an old Kubernetes client library. Only appropriate resource requests and limits should be changed.

Fixes #3567

### DIFF
--- a/addon-resizer/nanny/kubernetes_client.go
+++ b/addon-resizer/nanny/kubernetes_client.go
@@ -17,6 +17,7 @@ limitations under the License.
 package nanny
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 
@@ -25,6 +26,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	metav1beta1 "k8s.io/apimachinery/pkg/apis/meta/v1beta1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	scheme "k8s.io/client-go/kubernetes/scheme"
 )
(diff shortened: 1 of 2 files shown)


addon-resizer/nanny/kubernetes_client.go
```diff
@@ -151,16 +153,54 @@ func (k *kubernetesClient) UpdateDeployment(resources *corev1.ResourceRequiremen
 	// Modify the Deployment object with our ResourceRequirements.
 	for i, container := range dep.Spec.Template.Spec.Containers {
 		if container.Name == k.container {
-			// Update the deployment.
-			dep.Spec.Template.Spec.Containers[i].Resources = *resources
-			_, err = k.clientset.AppsV1().Deployments(k.namespace).Update(dep)
-			return err
+			return k.patchDeployment(getContainerResourcesPatch(i, mergeResources(&container.Resources, resources)))
```
liggitt, line 156: Consider what happens if the deployment spec changes between the Get and the Patch request.

Previously, the Update would conflict and force a retry. Now, the patch will be applied to the container at the specified index, even if it is not the expected container.

I'd suggest adding a test to the patch the make sure the name of the container at index %d is still the expected container.

bskiba: Thanks! Makes a lot of sense.

Implemented and tested in all the variants. Looks good.

addon-resizer/nanny/kubernetes_client.go
```diff
@@ -152,15 +154,65 @@ func (k *kubernetesClient) UpdateDeployment(resources *corev1.ResourceRequiremen
...
+		return fmt.Errorf("Cannot marshal deployment patches %+v. Reason: %+v", patches, err)
+	}
+
+	_, err = k.clientset.AppsV1().Deployments(k.namespace).Patch(k.deployment, types.JSONPatchType, bytes)
+	return err
+}
+
+type patchRecord struct {
+	Op    string      `json:"op,inline"`
+	Path  string      `json:"path,inline"`
```
liggitt, line 177: what is the `inline` bit for?

liggitt:
one question about the inline json annotations, lgtm otherwise

did you try this out in a built container?

liggitt:
would like a lgtm from @wojtek-t as well

addon-resizer/nanny/kubernetes_client.go
wojtek-t: I agree we should remove it (for both fields).

wojtek-t:
LGTM modulo this single comment (please ping me once fixed and I will lgtm).

addon-resizer/nanny/kubernetes_client.go
bskiba: Done

PR #4009, opened 2021-04-12
cloudprovider: add Bizflycloud provider

lmq1999:
This adds a new cluster autoscaler for [Bizflycloud Kubernetes Engine](https://bizflycloud.vn/kubernetes-engine)

P/s: 
Remove new dependencies and add inline cloud provider package
Make new pull request since old pull request commited by un-emailed user


lmq1999:
I need help with my sub package, it didn't go though test-and-verify process

lmq1999:
@mwielgus Hi, can you review my change again? 
I have remove dependencies and add inline package as other cloud provider.

I already edit verify file but my API package didn't pass the verifier

mwielgus:
Looks good, please also add your cloud provider here and we are good to go:

https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/README.md

lmq1999:
> Looks good, please also add your cloud provider here and we are good to go:
> 
> https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/README.md

Added as you say

mwielgus:
Actually, there is one serious problem I have just noticed (sorry for not checking it earlier). Your cloud provider library  is released under GPL(https://github.com/kubernetes/autoscaler/pull/4009/files#diff-d85a67df5ab779ac4287080d3d00b64aded31bf33207f0594c546ae51a32482c) which is not compatible with Apache license under which Cluster Autoscaler is released (https://github.com/kubernetes/autoscaler/blob/master/LICENSE).

Quote from https://www.apache.org/licenses/GPL-compatibility.html:

> [...] However, GPLv3 software cannot be included in Apache projects. The licenses are incompatible in one direction only, and it is a result of ASF's licensing philosophy and the GPLv3 authors' interpretation of copyright law. [...]

So we cannot accept any Bizflycloud dependency as long as it stays under GPL. If your are employed in Bizflycloud please talk to your lawyers about possible license change.

lmq1999:
> Actually, there is one serious problem I have just noticed (sorry for not checking it earlier). Your cloud provider library is released under GPL(https://github.com/kubernetes/autoscaler/pull/4009/files#diff-d85a67df5ab779ac4287080d3d00b64aded31bf33207f0594c546ae51a32482c) which is not compatible with Apache license under which Cluster Autoscaler is released (https://github.com/kubernetes/autoscaler/blob/master/LICENSE).
> 
> Quote from https://www.apache.org/licenses/GPL-compatibility.html:
> 
> > [...] However, GPLv3 software cannot be included in Apache projects. The licenses are incompatible in one direction only, and it is a result of ASF's licensing philosophy and the GPLv3 authors' interpretation of copyright law. [...]
> 
> So we cannot accept any Bizflycloud dependency as long as it stays under GPL. If your are employed in Bizflycloud please talk to your lawyers about possible license change.

Can we use MIT license for Bizfly Cloud dependency instead of GPL. I can't ask him right now but I will talk to him later

lmq1999:
We have manage to change license to MIT since it also compatible with Apache 2.0 license

lmq1999:
@mwielgus Can you check the PR again, it quite long since your last review

mwielgus (approved):
/lgtm
/approve

Issue #4007, opened 2021-04-11
Provide new/old count for ScaledUpGroup event


tomkerkhove:
**Which component are you using?**:
cluster-autoscaler

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:

Have a consistent experience for `ScaleDown` to include new/old node count, following `ScaledUp` based on https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-events-are-emitted-by-ca

**Describe the solution you'd like.**:
A similar approach to `ScaledUp` for pods based on https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-events-are-emitted-by-ca because the current `ScaledUpGroup` does not include the old amount of nodes.

PR #4014, opened 2021-04-14
Improve ScaledUpGroup event info to include current & max nodes

tomkerkhove:
Closes https://github.com/kubernetes/autoscaler/issues/4007

### DIFF
--- a/cluster-autoscaler/core/scale_up.go
+++ b/cluster-autoscaler/core/scale_up.go
@@ -674,7 +674,7 @@ groupsloop:
 func executeScaleUp(context *context.AutoscalingContext, clusterStateRegistry *clusterstate.ClusterStateRegistry, info nodegroupset.ScaleUpInfo, gpuType string, now time.Time) errors.AutoscalerError {
 	klog.V(0).Infof("Scale-up: setting group %s size to %d", info.Group.Id(), info.NewSize)
 	context.LogRecorder.Eventf(apiv1.EventTypeNormal, "ScaledUpGroup",
-		"Scale-up: setting group %s size to %d", info.Group.Id(), info.NewSize)
+		"Scale-up: setting group %s size to %d instead of %d (max: %d)", info.Group.Id(), info.NewSize, info.CurrentSize, info.MaxSize)
 	increase := info.NewSize - info.CurrentSize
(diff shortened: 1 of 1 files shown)


tomkerkhove:
Any thoughts @vivekbagade ?

tomkerkhove:
Bump @vivekbagade

cluster-autoscaler/core/scale_up.go
```diff
@@ -674,7 +674,7 @@ groupsloop:
 func executeScaleUp(context *context.AutoscalingContext, clusterStateRegistry *clusterstate.ClusterStateRegistry, info nodegroupset.ScaleUpInfo, gpuType string, now time.Time) errors.AutoscalerError {
 	klog.V(0).Infof("Scale-up: setting group %s size to %d", info.Group.Id(), info.NewSize)
 	context.LogRecorder.Eventf(apiv1.EventTypeNormal, "ScaledUpGroup",
-		"Scale-up: setting group %s size to %d", info.Group.Id(), info.NewSize)
+		"Scale-up: setting group %s size to %d instead of %d (max: %d)", info.Group.Id(), info.NewSize, info.CurrentSize, info.MaxSize)
```
mwielgus, line 677: It would be cleaner if you used something like "setting group % size to %d (previously: %d max %d)". "Instead of" suggests that something altered the decision and info.currentSize was the preferred size. 

@MaciekPytel - do you know if anyone is heavily dependent on the exact stings in events?

mwielgus (changes requested):

cluster-autoscaler/core/scale_up.go
tomkerkhove: I am with http://docs.k8s-event-grid-bridge.io/ 😅 

Happy to change it if you want, but it wouldn't really be backwards compatible.

Spoiler alert - Kubernetes needs to adopt CloudEvents with unique event types.

tomkerkhove:
Any thoughts on this @mweibel @MaciekPytel?

tomkerkhove:
Any thoughts @towca?

mwielgus (approved):
/lgtm
/approve

PR #4015, opened 2021-04-14
cloudprovider: ClusterAPIProviderName spelling

jackfrancis:
This is a housekeeping PR that updates the cluster-api provider const reference to `ClusterAPIProviderName` (from `ClusterAPIProiverName`).

No functional change should be observed as a result of updating this const.


mwielgus (approved):
/lgtm
/approve

PR #4017, opened 2021-04-15
logs: set clusterapi providerID warning to v=3

jackfrancis:
This PR increases the log verbosity of the warning message outputted if a machine does not have a providerID value to `3`.

The justification for such a change is that the providerID value is a "late", asynchronous arrival by design: the Machine resource is always created first, and then it is up to the cloud provider to populate the `providerID` value when the cloud provider-specific string reference is ready to go, which depends upon cloud provider-specific IaaS bootstrap behaviors.

tl;dr we don't want to see a stream of warning messages like these for every scale out event (unless we have explicitly opted into more verbose logging with the `--v=3` or greater runtime configuration):

```
W0415 19:01:45.629763       1 clusterapi_controller.go:454] Machine "default-template-md-0-75cf99b64-psglk" has no providerID
W0415 19:01:45.629837       1 clusterapi_controller.go:454] Machine "default-template-md-0-75cf99b64-pc2xp" has no providerID
W0415 19:01:45.629851       1 clusterapi_controller.go:454] Machine "default-template-md-0-75cf99b64-p7sxr" has no providerID
W0415 19:01:45.629859       1 clusterapi_controller.go:454] Machine "default-template-md-0-75cf99b64-kdqfj" has no providerID
W0415 19:01:45.629867       1 clusterapi_controller.go:454] Machine "default-template-md-0-75cf99b64-fgdld" has no providerID
W0415 19:01:45.629878       1 clusterapi_controller.go:454] Machine "default-template-md-0-75cf99b64-6l45k" has no providerID
W0415 19:01:45.629885       1 clusterapi_controller.go:454] Machine "default-template-md-0-75cf99b64-7jc76" has no providerID
```


jackfrancis:
cc @shysank @CecileRobertMichon @elmiko for thoughts on this

jackfrancis:
Also cc @detiber who may have thoughts

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
```diff
@@ -452,7 +452,9 @@ func (c *machineController) scalableResourceProviderIDs(scalableResource *unstru
 			}
 		}
 
-		klog.Warningf("Machine %q has no providerID", machine.GetName())
+		if klog.V(3).Enabled() {
```
elmiko, line 455: could we make this log level 4 to keep it consistent with the other debug log statement in the capi provider?

elmiko (commented):
i agree that this could be more noise for the logs, but i do think it's a useful signal. perhaps we should check the machine creation time and if it's older than the max provision time for the CA, then we could print the warning at `V(1)`, otherwise print at `V(4)`. wdyt?

hardikdr:
Is there any update here?

elmiko:
i think this is a good patch, just would like to make the log levels consistent. otherwise i'm lgtm

mwielgus:
Relatively simple comment still not addressed, closing due to inactivity. Feel free to reopen.

PR #4019, opened 2021-04-16
Brightbox cluster autoscaler provider

NeilW:
Hello,

Here is the implementation of the [Brightbox Cloud](https://www.brightbox.com) cloud provider . We've been running this provider [as a fork](https://github.com/brightbox/autoscaler) for several versions now and we feel it is now stable enough to be submitted to the mainline.

Comments appreciated


cluster-autoscaler/go.mod
```diff
@@ -10,6 +10,8 @@ require (
 	github.com/Azure/go-autorest/autorest/date v0.3.0
 	github.com/Azure/go-autorest/autorest/to v0.2.0
 	github.com/aws/aws-sdk-go v1.35.24
+	github.com/brightbox/gobrightbox v0.5.7
```
mwielgus, line 13: We do not accept any additional dependencies that are not in kubernetes/kubernetes. Please in-line your deps as in other cloud providers do.

mwielgus (changes requested):

cluster-autoscaler/go.mod
NeilW: In-lined the dependencies as requested.

BigDarkClown:
This PR was reversed in #4410. The code added here does not pass lint & spell checking unit tests, causing unit-test failures in repository. Please clean up the your cloudprovider implementation before adding it again.

PR #4022, opened 2021-04-21
[cluster-autoscaler] Publish node group min/max metrics

amrmahdi:
This change publishes the node groups min and max count set on the cluster autoscaler. The current use case that these metrics are needed for is to create alerts when the number of nodes in the cluster reach a certain percentage of the max allowed nodes for example.

### DIFF
--- a/cluster-autoscaler/core/autoscaler.go
+++ b/cluster-autoscaler/core/autoscaler.go
@@ -27,7 +27,6 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/estimator"
 	"k8s.io/autoscaler/cluster-autoscaler/expander"
 	"k8s.io/autoscaler/cluster-autoscaler/expander/factory"
-	"k8s.io/autoscaler/cluster-autoscaler/metrics"
 	ca_processors "k8s.io/autoscaler/cluster-autoscaler/processors"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/backoff"
@@ -67,13 +66,6 @@ func NewAutoscaler(opts AutoscalerOptions) (Autoscaler, errors.AutoscalerError) 
 	if err != nil {
 		return nil, errors.ToAutoscalerError(errors.InternalError, err)
 	}
-
-	// These metrics should be published only once.
-	for _, nodeGroup := range opts.CloudProvider.NodeGroups() {
-		metrics.UpdateNodeGroupMin(nodeGroup.Id(), nodeGroup.MinSize())
-		metrics.UpdateNodeGroupMax(nodeGroup.Id(), nodeGroup.MaxSize())
-	}
-
 	return NewStaticAutoscaler(
 		opts.AutoscalingOptions,
 		opts.PredicateChecker,
--- a/cluster-autoscaler/core/autoscaler.go
+++ b/cluster-autoscaler/core/autoscaler.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/estimator"
 	"k8s.io/autoscaler/cluster-autoscaler/expander"
 	"k8s.io/autoscaler/cluster-autoscaler/expander/factory"
+	"k8s.io/autoscaler/cluster-autoscaler/metrics"
 	ca_processors "k8s.io/autoscaler/cluster-autoscaler/processors"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/backoff"
@@ -66,6 +67,13 @@ func NewAutoscaler(opts AutoscalerOptions) (Autoscaler, errors.AutoscalerError) 
 	if err != nil {
 		return nil, errors.ToAutoscalerError(errors.InternalError, err)
 	}
+
+	// These metrics should be published only once.
+	for _, nodeGroup := range opts.CloudProvider.NodeGroups() {
+		metrics.UpdateNodeGroupMin(nodeGroup.Id(), nodeGroup.MinSize())
+		metrics.UpdateNodeGroupMax(nodeGroup.Id(), nodeGroup.MaxSize())
+	}
+
 	return NewStaticAutoscaler(
 		opts.AutoscalingOptions,
 		opts.PredicateChecker,
--- a/cluster-autoscaler/metrics/metrics.go
+++ b/cluster-autoscaler/metrics/metrics.go
@@ -330,7 +330,7 @@ var (
 )
 
 // RegisterAll registers all metrics.
-func RegisterAll() {
+func RegisterAll(emitPerNodeGroupMetrics bool) {
 	legacyregistry.MustRegister(clusterSafeToAutoscale)
 	legacyregistry.MustRegister(nodesCount)
 	legacyregistry.MustRegister(nodeGroupsCount)
@@ -340,8 +340,6 @@ func RegisterAll() {
 	legacyregistry.MustRegister(cpuLimitsCores)
 	legacyregistry.MustRegister(memoryCurrentBytes)
 	legacyregistry.MustRegister(memoryLimitsBytes)
-	legacyregistry.MustRegister(nodesGroupMinNodes)
-	legacyregistry.MustRegister(nodesGroupMaxNodes)
 	legacyregistry.MustRegister(lastActivity)
 	legacyregistry.MustRegister(functionDuration)
 	legacyregistry.MustRegister(functionDurationSummary)
@@ -359,6 +357,11 @@ func RegisterAll() {
 	legacyregistry.MustRegister(napEnabled)
 	legacyregistry.MustRegister(nodeGroupCreationCount)
 	legacyregistry.MustRegister(nodeGroupDeletionCount)
+
+	if emitPerNodeGroupMetrics {
+		legacyregistry.MustRegister(nodesGroupMinNodes)
+		legacyregistry.MustRegister(nodesGroupMaxNodes)
+	}
 }
 
 // UpdateDurationFromStart records the duration of the step identified by the
--- a/cluster-autoscaler/metrics/metrics.go
+++ b/cluster-autoscaler/metrics/metrics.go
@@ -170,6 +170,22 @@ var (
 		}, []string{"direction"},
 	)
 
+	nodesGroupMinNodes = k8smetrics.NewGaugeVec(
+		&k8smetrics.GaugeOpts{
+			Namespace: caNamespace,
+			Name:      "node_group_min_count",
+			Help:      "Minimum number of nodes in the node group",
+		}, []string{"node_group"},
+	)
+
+	nodesGroupMaxNodes = k8smetrics.NewGaugeVec(
+		&k8smetrics.GaugeOpts{
+			Namespace: caNamespace,
+			Name:      "node_group_max_count",
+			Help:      "Maximum number of nodes in the node group",
+		}, []string{"node_group"},
+	)
+
 	/**** Metrics related to autoscaler execution ****/
 	lastActivity = k8smetrics.NewGaugeVec(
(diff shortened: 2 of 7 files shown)


cluster-autoscaler/metrics/metrics.go
```diff
@@ -138,6 +138,22 @@ var (
 		},
 	)
 
+	nodesGroupMinNodes = k8smetrics.NewGaugeVec(
+		&k8smetrics.GaugeOpts{
+			Namespace: caNamespace,
+			Name:      "node_group_min_count",
+			Help:      "Minimum number of nodes in the node group",
+		}, []string{"node_group"},
```
elmiko, line 146: one issue here is that the cardinality of these metrics will essentially be unbounded. i have been advised against doing this type of thing in the past, see this discussion https://github.com/kubernetes/autoscaler/pull/3922#discussion_r586939682

elmiko (commented):
although i like these kind of metrics, i think there will be issues with the cardinality around the node group IDs.

cluster-autoscaler/metrics/metrics.go
dntosas: for the majority of users this should be totally ok. if a cluster is huge enough to have dozens of dozens of nodegroups, its Prometheus should either way handle another bunch of time series.

CA is a very usable component to expose such metrics and provide data for alerting for NodeGroups stuff

elmiko: i generally agree with you. i just wanted to highlight that i proposed something similar with respect to adding node group IDs, although in that case the metrics were re-calculated frequently. in this case they are just published once, so maybe that is more applicable.

i am curious those, what about node groups that get added after the autoscaler has started?

dntosas: >i am curious those, what about node groups that get added after the autoscaler has started?

i think it will discover them automatically if configured to do so via tags etc

cluster-autoscaler/core/autoscaler.go
```diff
@@ -66,6 +67,13 @@ func NewAutoscaler(opts AutoscalerOptions) (Autoscaler, errors.AutoscalerError)
 	if err != nil {
 		return nil, errors.ToAutoscalerError(errors.InternalError, err)
 	}
+
+	// These metrics should be published only once.
+	for _, nodeGroup := range opts.CloudProvider.NodeGroups() {
+		metrics.UpdateNodeGroupMin(nodeGroup.Id(), nodeGroup.MinSize())
+		metrics.UpdateNodeGroupMax(nodeGroup.Id(), nodeGroup.MaxSize())
+	}
```
dntosas, line 75: can we maybe include current nodeGroup size? smth like `nodeGroup.TargetSize()` WDYT ?

dntosas (commented):
very nice addition, one minor comment :+1:

cluster-autoscaler/metrics/metrics.go
elmiko: yeah, seems like new node groups would get picked up, but changes to the min/max would not.

cluster-autoscaler/core/autoscaler.go
elmiko: adding the current size would require this metric to updated more frequently for it to be accurate. which brings us back to the cardinality issue.

cluster-autoscaler/metrics/metrics.go
amrmahdi: I understand your concern about cardinality. But node groups are bounded. Compare that to something like `kube_node_info` which has the node label.

cluster-autoscaler/core/autoscaler.go
amrmahdi: The current ready nodes can calculated from other kube metrics as well.

feiskyer: agreed, ready notes metrics is not required in CA

feiskyer:
+1 as well to add those metrics. could you fix the CI failures?

feiskyer:
@mwielgus @maciekpytel any ideas the tests didn't run now for this PR?

elmiko:
> +1 as well to add those metrics.

i'm curious if we could get come consensus around using node group IDs in the metrics. i think there are other metrics that could benefit from using these identifiers, but i have been warned against using them in the past.

i think if we are going to agree that node group IDs are acceptable to use as metrics labels, then we should also establish guidance as to when it is appropriate. for example, this PR proposes a metric that will only get updated once (although this  misses the notion of node groups being adjusted after initialization), is this usage of node group IDs acceptable? if this usage is acceptable, when is it unacceptable to use them?

amrmahdi:
> @mwielgus @MaciekPytel any ideas the tests didn't run now for this PR?

Can we get help on the CI? It seems it is waiting for a maintainer.

image

feiskyer:
@mwielgus @MaciekPytel could you help to approve the github actions? The new changes from github actions make it required of approval first from maintainers (this is also true for other PRs).

feiskyer (approved):
LGTM
/approve

feiskyer:
@mwielgus @MaciekPytel could you help to approve the github actions?

cluster-autoscaler/core/autoscaler.go
```diff
@@ -66,6 +67,13 @@ func NewAutoscaler(opts AutoscalerOptions) (Autoscaler, errors.AutoscalerError)
 	if err != nil {
 		return nil, errors.ToAutoscalerError(errors.InternalError, err)
 	}
+
+	// These metrics should be published only once.
```
towca, line 71: This doesn't take into account node groups which get added after Cluster Autoscaler starts.

cluster-autoscaler/metrics/metrics.go
towca: @mwielgus you had strong opinions on label cardinality in the past, could you weigh in here?

towca (changes requested):
I approved the github workflows. I'll defer to @mwielgus on cardinality.

cluster-autoscaler/core/autoscaler.go
MaciekPytel: Both the list of nodegroups and min/max size of existing nodegroups can change during cloudprovider.Refresh() call that happens every loop. So even for nodegroups that are already registered at this point the value may change later.

cluster-autoscaler/metrics/metrics.go
MaciekPytel: I definitely argued against using identifiers as metric labels due to cardinality issues. I broadly disagree with the assumption that if something is useful but doesn't scale well it's fine since most users don't run clusters with thousands of nodes and so most users will benefit from it. And I disagree that number of nodegroups can be assumed to be small. 

That being said those metrics are obviously very specific to monitoring individual nodegroups. It's not really stopping anyone from using existing metrics in a large cluster - they can just not use the new metrics. So I'm generally fine with adding those new metrics. My issue is that prometheus will start collecting those metrics by default and you'll need to add relabel_configs to stop doing it. Someone upgrading CA to a newer version is likely to miss it. And that will also be true if we add more nodegroup labeled metrics in future.
For that reason I'd like to consider hiding it behind a CA flag (ex. --emit-per-nodegroup-metrics). That way a user could easily opt-in/out of all such metrics.

amrmahdi: @MaciekPytel I'm fine adding this behind a flag. But I'd like to understand more the cardinality scope here. Even for clusters with thousands of nodes, how many node groups are we talking about here? A hundred, a couple of hundred? Given that such metrics are static by nature, I'd like to understand how this can lead to cardinality explosion. Even if node groups come and go, in a single TSDB block you should not have a high cardinality. 

Again, kubernetes itself emits metrics with node name labels, which have far higher cardinality.

elmiko: i like the notion of adding a flag which would enable node group labels for some metrics. i think there are other metrics which could benefit from having the node group ID as a label and it would be nice to empower users who understand the operational load of adding this extra data to their metrics solutions.

MaciekPytel: Re: cardinality issue - thousands of nodegroups can happen (ex. #4058 is an optimization specifically for such case). It's hard to imagine the number any higher given that kubernetes itself only supports 5k nodes I think. However, it's not uncommon for large organization to run dozens/hundreds of clusters and I think joining the metrics from multiple clusters is where the cardinality really becomes an issue. I do agree this specific metric is much less of an issue than other metrics, since it's going to be very static. I think having a flag is still the way to go if we want to add more metrics labeled by nodegroup. Once we have such flag I see little reason why we couldn't add more metrics like this. I'm less sure about adding labels to existing metrics - variable number of metrics on a label seems like an easy source of confusion. Instead I'd propose creating an equivalent metric with slightly different name and emitting one or the other based on the flag. WDYT?

elmiko: >  I'm less sure about adding labels to existing metrics - variable number of metrics on a label seems like an easy source of confusion. Instead I'd propose creating an equivalent metric with slightly different name and emitting one or the other based on the flag. WDYT?

you make a good point about the possible confusion for users. i'm not sure how i feel about creating multiple metrics (with and without the labels), it seems a little onerous but that might just be my initial reaction. in the absence of another option, it seems like it might the best path forward. i'm happy to hear more conversation on this topic though. thanks @MaciekPytel !

amrmahdi: Thanks @MaciekPytel for the elaboration. Sounds good. I'll add the flag and update the PR.

cluster-autoscaler/core/autoscaler.go
amrmahdi: This is now addressed by publishing the metrics after each cloud provider refresh in the static autoscaler `RunOnce`

elmiko (approved):
i like the addition of the flag, i'm happy with this solution.
/lgtm

elmiko (changes requested):
actually, i just thought about something. @amrmahdi would you mind adding something to the [FAQ](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca) to mention the new flag please?

/lgtm cancel

amrmahdi:
> i just thought about something. @amrmahdi would you mind adding som

Yes I missed that. Thanks

cluster-autoscaler/metrics/metrics.go
```diff
@@ -422,6 +443,16 @@ func UpdateMemoryLimitsBytes(minMemoryCount int64, maxMemoryCount int64) {
 	memoryLimitsBytes.WithLabelValues("maximum").Set(float64(maxMemoryCount))
 }
 
+// UpdateNodeGroupMin records the node group minimum allowed number of nodes
+func UpdateNodeGroupMin(nodeGroup string, minNodes int) {
+	nodesGroupMinNodes.WithLabelValues(nodeGroup).Set(float64(minNodes))
```
MaciekPytel, line 448: What happens if we set value of unregistered metric? Is this supported? Or are we relying on some undocumented implementation detail here?

MaciekPytel:
Left one more comment, otherwise lgtm.

cluster-autoscaler/metrics/metrics.go
amrmahdi: Setting the value is independent from registering the metrics for collection. In the standard Prometheus client, the value will be set, and can be read to what it was set to, but it will not be exposed to collection unless it is registered. In the current wrapper implementation used, internal the Prometheus gauge is not created unless it is registered, and hence the set operation is a no-op.

elmiko (approved):
thanks for the update
/lgtm

amrmahdi:
> I approved the github workflows. I'll defer to @mwielgus on cardinality.

@towca Any more concerns?

amrmahdi:
> > I approved the github workflows. I'll defer to @mwielgus on cardinality.
> 
> @towca Any more concerns?

@towca @mwielgus can we close on this PR? Thanks

PR #4023, opened 2021-04-21
Enable custom k8s fork in update-vendor.sh

BigDarkClown:
`update-vendor.sh` was using hardcoded open-source Kubernetes repository, which made it impossible to update vendor dependencies based on a Kubernetes fork, which was the case prior to the refactor of `update-vendor.sh`. This PR implements that by moving from `curl` command to `git` based solution which works also in closed source repositories.


mwielgus:
Please provide more description of the change.

PR #4034, opened 2021-04-24
fix(docs): corrects various typos in project documentation

plan-do-break-fix:
Scope of changes is restricted to docs only. Changes made in accordance with Standard English. CLA previously signed.


plan-do-break-fix:
Didn't realize I had pulled typos in the vendor dependency folder. Closing PR.

PR #3649, opened 2020-10-28
Adding functionality to cordon the node before destroying it. 

atulaggarwal:
This helps load balancer to remove the node from healthy hosts (ALB does have this support).

This won't fix the issue of 502 completely as there is some time node to live even after cordoning as to serve In-Flight request.
Shamelessly copied from https://github.com/kubernetes/autoscaler/pull/3014/files

PR #4044, opened 2021-04-28
Adding functionality to cordon the node before destroying it. This helps load balancer to remove the node from healthy hosts (ALB does have this support).

atulaggarwal:
This won't fix the issue of 502 completely as there is some time node has to live even after cordoning as to serve In-Flight request but load balancer can be configured to remove Cordon nodes from healthy host list.
This feature is enabled by cordon-node-before-terminating flag with default value as false to retain existing behavior.


atulaggarwal:
This for merging PR #3649  into 1.20 release branch.

mwielgus:
Please recreate it as an official cherry-pick of #3649.

PR #4046, opened 2021-04-28
Improve misleading log

### LINKED ISSUE #4045: Misleading logs in the cluster-autoscaler

cluster-autoscaler

Component version: master

In the logs of the cluster autoscaler for the AWS provider we can see those lines:

```
W0428 15:47:20.900817       1 aws_manager.go:315] Found multiple availability zones for ASG "eks-10bc8bf8-2e7d-cf1b-5aca-8a0e281160d4"; using eu-west-1b
W0428 15:47:20.901176       1 aws_manager.go:315] Found multiple availability zones for ASG "eks-febc8bee-dfd9-2d97-b3cf-5d5526e60c87"; using eu-west-1b
```

It is my understanding that the cluster-autoscaler has no means to select an Availability Zone when scaling so I don't understand the meaning of this log (the `using eu-west-1b` part) that can be misleading.

Regards.

sylr:
Fix #4045

Signed-off-by: Sylvain Rabot <sylvain@abstraction.fr>

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_manager.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_manager.go
@@ -312,7 +312,7 @@ func (m *AwsManager) getAsgTemplate(asg *asg) (*asgTemplate, error) {
 	region := az[0 : len(az)-1]
 
 	if len(asg.AvailabilityZones) > 1 {
-		klog.Warningf("Found multiple availability zones for ASG %q; using %s\n", asg.Name, az)
+		klog.V(4).Infof("Found multiple availability zones for ASG %q; using %s for %s label\n", asg.Name, az, apiv1.LabelFailureDomainBetaZone)
 	}
 
 	instanceTypeName, err := m.buildInstanceType(asg)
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/cloudprovider/aws/aws_manager.go
```diff
@@ -312,7 +312,7 @@ func (m *AwsManager) getAsgTemplate(asg *asg) (*asgTemplate, error) {
 	region := az[0 : len(az)-1]
 
 	if len(asg.AvailabilityZones) > 1 {
-		klog.Warningf("Found multiple availability zones for ASG %q; using %s\n", asg.Name, az)
+		klog.V(4).Infof("Found multiple availability zones for ASG %q; using %s for %s label\n", asg.Name, az, apiv1.LabelFailureDomainBetaZone)
```
gjtempleton, line 315:
```suggestion
		klog.V(4).Infof("Found multiple availability zones for ASG %q; using %s for %s label\n", asg.Name, az, apiv1.LabelZoneFailureDomain)
```

We should stick to the same constant as is being used by the `buildGenericLabels` which is processing the AZ value returned by this function, though it will result in the same result in this case.

Is there a reason from moving from a warning to a high verbosity level Info message? I feel like this is generally an indicator to cluster operators that they'd want to be warned about.

gjtempleton:
This may also come into conflict with #4053

cluster-autoscaler/cloudprovider/aws/aws_manager.go
sylr: I agree but this warning is emitted each run and becomes an annoyance more than being useful.

gjtempleton:
/approve
/lgtm
Thanks!

Issue #2434, opened 2019-10-09
Autoscaler not respecting the taint tag in AWS


dcherniv:
It seems that cluster autoscaler is ignoring the taints in some cases and tries to spin up a node pool that is tainted to schedule a pod with no toleration for the taint.

Example pod spec:
```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    kubernetes.io/psp: eks.privileged
  labels:
    app: dev-pay-stub-services
    pod-template-hash: 64f54bbcd
  name: dev-pay-stub-services-64f54bbcd-d5z2r
  namespace: dev
  ownerReferences:
  - apiVersion: apps/v1
    blockOwnerDeletion: true
    controller: true
    kind: ReplicaSet
    name: dev-pay-stub-services-64f54bbcd
spec:
  containers:
  - env:
REDACTED
    image: REDACTED/pay-stub-services:0.9.1-5-g3d1358f
    imagePullPolicy: Always
    livenessProbe:
      failureThreshold: 3
      httpGet:
        path: /
        port: http
        scheme: HTTP
      initialDelaySeconds: 30
      periodSeconds: 3
      successThreshold: 1
      timeoutSeconds: 30
    name: pay-stub-services
    ports:
    - containerPort: 5000
      name: http
      protocol: TCP
    readinessProbe:
      failureThreshold: 3
      httpGet:
        path: /ready
        port: http
        scheme: HTTP
      initialDelaySeconds: 30
      periodSeconds: 3
      successThreshold: 1
      timeoutSeconds: 30
    resources:
      limits:
        cpu: "2"
        memory: 4Gi
      requests:
        cpu: 1500m
        memory: 3Gi
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    volumeMounts:
REDACTED
  restartPolicy: Always
  schedulerName: default-scheduler
  securityContext: {}
  serviceAccount: default
  serviceAccountName: default
  terminationGracePeriodSeconds: 30
  tolerations:
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    operator: Exists
    tolerationSeconds: 300
  - effect: NoExecute
    key: node.kubernetes.io/unreachable
    operator: Exists
    tolerationSeconds: 300
  volumes:
  - configMap:
      defaultMode: 420
      items:
      - key: newrelic.ini
        path: newrelic.ini
      name: dev-pay-stub-services
    name: newrelic-config
  - name: default-token-nj7fb
[...]

PR #4050, opened 2021-04-29
Add example to AWS readme if taint has value

afirth:
/area provider/aws
Maybe helps https://github.com/kubernetes/autoscaler/issues/2434 - TBD in the issue

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/README.md
+++ b/cluster-autoscaler/cloudprovider/aws/README.md
@@ -143,9 +143,10 @@ be labeled or tainted when they join the cluster, such as:
 
 * `k8s.io/cluster-autoscaler/node-template/label/foo`: `bar`
 * `k8s.io/cluster-autoscaler/node-template/taint/dedicated`: `NoSchedule`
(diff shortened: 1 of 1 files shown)


gjtempleton:
Thanks for the PR!

I'm not clear on how this resolves #2434 though.

afirth:
> Thanks for the PR!
> 
> I'm not clear on how this resolves #2434 though.

As far as I could tell, all comments in that issue were a) expecting CAS to taint new nodes b) failing to specify the `Effect` or c) using a taint which is ignored by scheduler.
Basically, working as designed I think. For most/all of those comments I think seeing this doc would solve their issue

afirth:
/remove-lifecycle stale

Not sure what's against merging this.. want me to remove the /fix?

gjtempleton:
Yeah, let's remove the `fixes`, merge this, and confirm whether it does resolve the issue with those who had commented in the issue rather than assuming it resolves all of the user queries in there.

afirth:
Done :)

gjtempleton:
Thanks! And thanks for persisting with this.

/lgtm
/approve

Issue #4001, opened 2021-04-08
Provide observability for unregistered nodes


dharmab:
**Which component are you using?**:

Cluster Autoscaler

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:

The metric `cluster_autoscaler_nodes_count` can be used to track if VMs created by the Cluster Autoscaler are not joining the clusters as Nodes in a timely manner. However, the metric and logs only report the number of unregistered nodes. In a large cluster with many ASGs/scale sets, it is difficult to find the specific VMs which are not joining the cluster.

**Describe the solution you'd like.**:

The cluster autoscaler should provide a way to determine which VMs are not registered. This could be a log containing the VM names, or a metric with a label for each Node and its state.

**Describe any alternative solutions you've considered.**:

We could write a script that uses the cloud provider API and compares the VMs in the cloud provider to those in Kubernetes. This would have to be different for each cloud provider, and it seems that CA has all the necessary access and information to do this anyway.

**Additional context.**:

https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/proposals/metrics.md#cluster-autoscaler-operations

gjtempleton:
Hey, thanks for raising this as a feature request!

I'd be worried about adding metrics with node names/identities as labels given the potential for high cardinality in large clusters. To me using logs seems like the sensible way to go instead to provide this functionality.

You're welcome to raise a PR adding this functionality if you want to.

PR #4060, opened 2021-05-05
Log names of longUnregistered Nodes

dharmab:
Log the names of longUnregistered Nodes to make it easier to debug issues where new Nodes are unable to join the cluster.

Resolves https://github.com/kubernetes/autoscaler/issues/4001

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -576,6 +576,7 @@ func (csr *ClusterStateRegistry) updateReadinessStats(currentTime time.Time) {
 		total = update(total, node, ready)
 	}
 
+	var longUnregisteredNodeNames []string
 	for _, unregistered := range csr.unregisteredNodes {
 		nodeGroup, errNg := csr.cloudProvider.NodeGroupForNode(unregistered.Node)
 		if errNg != nil {
@@ -588,6 +589,7 @@ func (csr *ClusterStateRegistry) updateReadinessStats(currentTime time.Time) {
 		}
 		perNgCopy := perNodeGroup[nodeGroup.Id()]
 		if unregistered.UnregisteredSince.Add(csr.config.MaxNodeProvisionTime).Before(currentTime) {
+			longUnregisteredNodeNames = append(longUnregisteredNodeNames, unregistered.Node.Name)
 			perNgCopy.LongUnregistered++
 			total.LongUnregistered++
 		} else {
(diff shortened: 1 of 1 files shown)


dharmab:
Amended commit - `longUnregisteredNodeNames` is now declared as mutable

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -596,6 +598,9 @@ func (csr *ClusterStateRegistry) updateReadinessStats(currentTime time.Time) {
 		}
 		perNodeGroup[nodeGroup.Id()] = perNgCopy
 	}
+	if total.LongUnregistered > 0 {
+		klog.Warningf("Found longUnregistered Nodes %s", longUnregisteredNodeNames)
```
mwielgus, line 602: The list can get quite long on larger clusers (in case of some outage) and printing it every iteration may be problematic. Please use higher lever of info for printing this.  Frequenlyy printend warnings/errors should contain only relatively brief messages (like the number of nodes, not their list).

mwielgus (changes requested):

cluster-autoscaler/clusterstate/clusterstate.go
dharmab: Sure, I can log it at Info level instead.

In #4001 discussion, it was explained that this information is very difficult to obtain from any other source. Asking cluster operators to query this information from the cloud API and compare to the Kubernetes API is cumbersome, and the data already exists in `clusterstate.go`. Adding this node-level data to metrics has a performance impact on metrics. A long log line is the least impactful way to make this information available to cluster operators. In my organization we get a lot of feedback from our ops teams on the difficulty of troubleshooting unregistered nodes without this information. I think the tradeoff of a long log line in the uncommon case of an outage is worth the human time and effort saved in the common case where a relatively small number of nodes are unregistered.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -596,6 +598,9 @@ func (csr *ClusterStateRegistry) updateReadinessStats(currentTime time.Time) {
 		}
 		perNodeGroup[nodeGroup.Id()] = perNgCopy
 	}
+	if total.LongUnregistered > 0 {
+		klog.Infof("Found longUnregistered Nodes %s", longUnregisteredNodeNames)
```
mwielgus, line 602: Please use some verbosity level (preferably 2+). klog.V(3).Infof(......)

mwielgus:
Thank you, please squash commits to just 1 and we are good to go.

mwielgus (approved):
/lgtm
/approve

PR #4053, opened 2021-05-03
Also set new (non-beta/non-deprecated) labels in buildGenericLabels

codablock:
This fixes issues when an ASG has 0 nodes yet and node labels are built
using buildGenericLabels and the node-template labels.

Issues include (anti-)affinity and nodeSelectors for the given labels,
giving false-negative results for candidate nodes, which leads to ASGs
never scaling up.

PR #4061, opened 2021-05-06
add stable zone labels in azure template generation

marwanad:
add stable zone labels in template regions

/area provider/azure


feiskyer (commented):
/lgtm
/approve

Thanks for adding this.

codablock:
fyi, https://github.com/kubernetes/autoscaler/pull/4053 solves the same issue but for all stable/beta/deprecated labels and for all providers

PR #4067, opened 2021-05-07
Document that TLS bootstrapping may be necessary for scale-up

### LINKED ISSUE #4065: Mention in documentation that CA does not automatically register nodes into the cluster.
**Documentation suggestion**

Could you mention on the main `cluster-autoscaler` README, and scaling up FAQ, that CA is meant to scale the VMs, but it is not responsible for registering the nodes into the cluster?

I don't think this information is expressed anywhere in this repository. The FAQ does mention that when removing the nodes you have to remove them yourself, but no such thing is mentioned with scaling up. My experience was that I went through configuration, and setup, only to see the VM scale up, but nothing happened after. I checked the logs only to see "2 unregistered nodes", and only in a raised issue did I find out that CA does not register the nodes. This could be avoided.

I dont mind writing a PR, but I believe you would be faster, and it would be less of a hassle for, but do let me know!

dharmab:
Resolves #4065. Kubernetes extensively documents what needs to happen for a Kubelet to register with the apiserver, so linking to the main doc is the simplest resolution.

### DIFF
--- a/cluster-autoscaler/FAQ.md
+++ b/cluster-autoscaler/FAQ.md
@@ -466,8 +466,9 @@ If there are multiple node groups that, if increased, would help with getting so
 different strategies can be selected for choosing which node group is increased. Check [What are Expanders?](#what-are-expanders) section to learn more about strategies.
 
 It may take some time before the created nodes appear in Kubernetes. It almost entirely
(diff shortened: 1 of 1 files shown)


mwielgus (approved):
/lgtm
/approve

PR #4073, opened 2021-05-11
cache ASG InstanceTypes for AWS

drmorr0:
## Summary

This change incorporates a cache for `LaunchTemplates` in addition to `LaunchConfigurations` for the AWS cloud provider.  Previously for ASGs with no instances (`DesiredCapacity = 0`), cluster-autoscaler would repeatedly hit the AWS API with `DescribeLaunchTemplateVersions` calls many times/second, resulting in throttling by the AWS API.  After this change, LaunchTemplates are cached just like LaunchConfigurations, which reduces the number of API calls to a manageable level, even with large numbers of ASGs.

The following changes were made:
1. Instead of having the launch configuration cache stored with the AutoScaling wrapper, the cache object is stored with the `asgCache` (this is because `DescribeLaunchTemplateVersions` is an EC2 call, not an AutoScaling call).
2. The AutoScaling wrapper and the EC2 wrapper are refactored into a single AWS wrapper, which means fewer parameters that have to get passed around
3. `buildLaunchTemplateFromSpec` is removed from the `asgCache` since it doesn't depend on any of the internals of that struct

## Testing done
- Unit tests added/fixed
- Manual testing with a Kubernetes cluster.  The following chart shows the number of `DescribeLaunchTemplateVersions` calls made by cluster-autoscaler before and after the change.  The Kubernetes cluster had 300 ASGs with no instances in them; from 13:30 to 14:45 shows the number of API calls made by the unpatched autoscaler; from 14:45 to 16:30 shows the number of API calls made by the patched autoscaler (note that this change also significantly improves the cycle time of CA when it has a large number of ASGs):

image


cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
```diff
@@ -18,83 +18,33 @@ package aws
...
-const (
-	launchConfigurationCachedTTL = time.Minute * 20
-	cacheMinTTL                  = 120
-	cacheMaxTTL                  = 600
+	"github.com/aws/aws-sdk-go/service/ec2"
 )
 
 // autoScaling is the interface represents a specific aspect of the auto-scaling service provided by AWS SDK for use in CA
-type autoScaling interface {
+type autoScalingI interface {
```
elmiko, line 28: maybe rename this to `autoScaler` to be more in line with interface type names?

adding an `I` makes sense to me, but also can be missed easily.

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
```diff
@@ -18,83 +18,33 @@ package aws
...
-	name         string
-	instanceType string
-}
-
-type jitterClock struct {
-	clock.Clock
-
-	jitter bool
-	sync.RWMutex
+type ec2I interface {
```
elmiko, line 36: similar here, i would expect this be named `ec2Templater` or something similar

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
```diff
@@ -261,3 +149,58 @@ func (m *autoScalingWrapper) getAutoscalingGroupNamesByTags(kvs map[string]strin
 
 	return asgNames, nil
 }
+
+func (m *awsWrapper) getInstanceTypeByLT(launchTemplate *launchTemplate) (string, error) {
```
elmiko, line 153: i think it would be nice to just name this `getInstanceTypeByLaunchTemplate`, just to make it clear

elmiko (commented):
this seems like a really nice upgrade, good work on showing the effect too!

i am not overly familiar with the aws provider code base, but this generally makes sense to me. i have a few small suggestions, but no blockers.

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
drmorr0: So originally we had an `autoScaling` interface and an `ec2I` interface, because without the `I` the `ec2` interface name conflicted with the AWS SDK service (see [here](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/ec2.go#L26) on the master branch).  I added the `I` here to make the two different interfaces consistent, but maybe there's a better solution?  I'll think about it some.

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
drmorr0: ++, also renamed `getInstaceTypeByLCNames`

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
elmiko: makes sense, i was just trying to think about verbs that describe the action for the interfaces. but, as always, naming is hard ;)

drmorr0:
@gjtempleton bump?  Any progress on this PR?  Thanks!

gjtempleton:
Hey @drmorr0,

Thanks for the nudge, I've been prioritising closing off some of the easier ones to this point (hence the need to rebase this PR now), however, I've also been giving some time to wrap my head around this one given its scale.

cluster-autoscaler/cloudprovider/aws/instance_type_cache.go
```diff
@@ -0,0 +1,158 @@
...
+	"k8s.io/apimachinery/pkg/util/clock"
+	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/client-go/tools/cache"
+	klog "k8s.io/klog/v2"
+)
+
+const (
+	asgInstanceTypeCacheTTL = time.Minute * 20
+	cacheMinTTL             = 120
+	cacheMaxTTL             = 600
```
bpineau, line 18: (only taking note for a potential follow-up, as this was already there (in an other file)):
perhaps asgInstanceTypeCacheTTL would deserve a cli arg so users can balance between throttling risks vs. freshness (depending on their context). And cacheMinTTL could be named something less confusing, like minCacheTTLJitter (resp. maxCacheTTLJitter).

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
```diff
@@ -261,3 +149,58 @@ func (m *autoScalingWrapper) getAutoscalingGroupNamesByTags(kvs map[string]strin
...
 	return asgNames, nil
 }
+
+func (m *awsWrapper) getInstanceTypeByLaunchTemplate(launchTemplate *launchTemplate) (string, error) {
+	params := &ec2.DescribeLaunchTemplateVersionsInput{
+		LaunchTemplateName: aws.String(launchTemplate.name),
+		Versions:           []*string{aws.String(launchTemplate.version)},
+	}
+
+	describeData, err := m.DescribeLaunchTemplateVersions(params)
```
bpineau, line 159: Would be nice to export metric for that API call (`start := time.Now(); describeData, err := m.DescribeLaunchTemplateVersions(params); observeAWSRequest("DescribeLaunchTemplateVersions", err, start)`).  Those metrics were introduced after this PR was opened so most will be picked up with a rebase, but that function is new.

bpineau:
For info in case that helps, I looked at the code (which LGTM minus a couple details), and successfully tested that PR (a rebased version) in the following conditions:
* On a cluster having ~50 AGS+LC and no LT, saw no regressions (in either discovery, upscaling/downscaling behaviour, or API calls rate)
* Attached 100 ASGs+LT to that cluster; saw LT were properly discovered, and DescribeLT API calls rate cooled down (after the initial discovery burst) showing the cache in effect. Hot-added/removed ASGs+LT were properly registered/unregistered. Saw no issues or regressions.

drmorr0:
> Hey @drmorr0,
> 
> Thanks for the nudge, I've been prioritising closing off some of the easier ones to this point (hence the need to rebase this PR now), however, I've also been giving some time to wrap my head around this one given its scale.

No worries!  It's taken me a bit to carve out some time to come back to this.  I just successfully rebased, am going to go through the other comments now.

cluster-autoscaler/cloudprovider/aws/instance_type_cache.go
```diff
@@ -0,0 +1,158 @@
...
+		})
+	}
+	klog.V(4).Infof("Successfully query %d launch configurations", len(launchConfigsToQuery))
+
+	// Have to query LaunchTemplates one-at-a-time, since there's no way to query <lt, version> pairs in bulk
+	for asgName, launchTemplateSpec := range launchTemplatesToQuery {
+		launchTemplate := buildLaunchTemplateFromSpec(launchTemplateSpec)
+		instanceType, err := m.awsService.getInstanceTypeByLaunchTemplate(launchTemplate)
+		if err != nil {
+			klog.Error("Failed to query launch tempate %s", launchTemplate.name)
```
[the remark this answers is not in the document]
drmorr0: Thanks, good catch!

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
drmorr0: Fixed.

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
```diff
@@ -88,6 +90,15 @@ func newASGCache(service autoScalingWrapper, explicitSpecs []string, autoDiscove
 	return registry, nil
 }
 
+// Use a function variable for ease of testing
+var getCachedInstanceTypeForAsg = func(m *asgCache, asg *asg) (string, error) {
+	if obj, found, _ := m.asgInstanceTypeCache.GetByKey(asg.AwsRef.Name); found {
+		return obj.(instanceTypeCachedObject).instanceType, nil
+	}
+
+	return "", fmt.Errorf("Could not find instance type for %s", asg.AwsRef.Name)
```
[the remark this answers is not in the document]
drmorr0: Oh good catch, thanks.  I think I've fixed this now.

drmorr0:
OK I think I've handled all the feedback I've gotten thus far.  One significant change from the previous version is that I moved the `instanceTypeCache.populate` call later in the `regenerate` function so I could use the internal `asg` type in both places that we query AWS.

I have not yet fixed up the tests nor repeated my experiments to make sure there aren't any regressions, but that's a tomorrow task :)

drmorr0:
OK @gjtempleton @bpineau @elmiko I have updated the tests and also repeated the experiment above to make sure that the launch-template caching works as expected.  I think this is ready for another look when you have a chance?  Thanks!

cluster-autoscaler/cloudprovider/aws/instance_type_cache.go
```diff
@@ -0,0 +1,112 @@
...
+	since := time.Since(ts)
+	c.RLock()
+	defer c.RUnlock()
+	if c.jitter {
+		return since + (time.Second * time.Duration(rand.IntnRange(cacheMinTTL, cacheMaxTTL)))
+	}
+	return since
+}
+
+func (m instanceTypeExpirationStore) populate(autoscalingGroups []*asg) error {
```
gjtempleton, line 71: Is there a reason for naming this `m`?

gjtempleton (commented):
Generally, this looks good to me, thanks for responding to all of the feedback. 

There's still some consistency issues around variable naming, mostly in tests, that I'd like to see cleaned up as you've mostly addressed this in `aws_cloud_provider_tests` but not fully in other places.

cluster-autoscaler/cloudprovider/aws/instance_type_cache.go
drmorr0: Nope; probably a copy-paste error.

drmorr0:
@gjtempleton I've done another pass through the tests and addressed your comments and a couple other cases that I found.  I think/hope that's all of them!  Let me know if you see anything else.

PR #4085, opened 2021-05-17
Print out error when unable to download module

brett-elliott:
Use shell syntax instead of pipe so error messages are printed out. Before
this change, if it could not download a module, it would just exit without
printing an error message due to the pipe. With this change it now prints
out the error message(unable to download) from the underlying process.

### DIFF
--- a/cluster-autoscaler/hack/update-vendor.sh
+++ b/cluster-autoscaler/hack/update-vendor.sh
@@ -50,8 +50,14 @@ rm -rf ${WORKDIR}
 
 for MOD in "${MODS[@]}"; do
     V=$(
-        go mod download -json "${MOD}@kubernetes-${VERSION}" |
-        sed -n 's|.*"Version": "\(.*\)".*|\1|p'
+        GOMOD="${MOD}@kubernetes-${VERSION}"
+        JSON=$(go mod download -json "${GOMOD}")
+        retval=$?
+        if [ $retval -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/hack/update-vendor.sh
```diff
@@ -50,8 +50,13 @@ rm -rf ${WORKDIR}
 
 for MOD in "${MODS[@]}"; do
     V=$(
-        go mod download -json "${MOD}@kubernetes-${VERSION}" |
-        sed -n 's|.*"Version": "\(.*\)".*|\1|p'
+        GOMOD=$(go mod download -json "${MOD}@kubernetes-${VERSION}")
+        retval=$?
+        if [ $retval -ne 0 ]; then
+            echo "Error downloading module."
```
mwielgus, line 56: Can we include the module name in the message?

mwielgus (changes requested):

brett-elliott:
It's already printed out when the $(...) command fails. When I did log it
explicitly it seemed a bit verbose and it looked odd to see the error
message twice. If you'd like I can include it again though. :)

cluster-autoscaler/hack/update-vendor.sh
brett-elliott: Hmm added the module in the message just how, doesn't looks better.

mwielgus (approved):
/lgtm
/approve

PR #4089, opened 2021-05-19
Fix templated nodeinfo names collisions in BinpackingNodeEstimator

bpineau:
Since a couple of months, both upscale's `getUpcomingNodeInfos` and the
binpacking estimator uses the same shared `DeepCopyTemplateNode` function
and inherits its naming pattern, which is great and fixes a long standing bug.

`getUpcomingNodeInfos` will now enrich the cluster snapshots with
generated nodeinfos and nodes having predictable names (using the nodegroup's
template name, and an incremental ordinal starting at 0) for upcoming nodes.

Later, when it looks for fitting nodes for unschedulable pods, the binpacking
estimator will also build virtual nodes and place them in a snapshot fork to
evaluate scheduler predicates.

Those temporary virtual nodes are built using the same pattern (template name
and an index ordinal also starting at 0) as the one previously used by
`getUpcomingNodeInfos`, which means it will generate the same nodeinfos/nodes
names for nodegroups having upcoming nodes.

But adding nodes by the same name in an existing cluster snapshot [isn't allowed](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/simulator/delta_cluster_snapshot.go#L152-L154),
and the evaluation attempt will fail. So no nodeInfo from node groups having
upcoming nodes can ever be considered as fitting unschedulable pods.

Practically this blocks re-upscales for nodegroups having upcoming nodes,
which can cause a significant delay (or suboptimal choices among the
possible node groups options). This wasn't the case until recently, as
`getUpcomingNodeInfos` and the estimator used different naming schemes.

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -799,7 +799,7 @@ func getUpcomingNodeInfos(registry *clusterstate.ClusterStateRegistry, nodeInfos
 			// Ensure new nodes have different names because nodeName
 			// will be used as a map key. Also deep copy pods (daemonsets &
 			// any pods added by cloud provider on template).
-			upcomingNodes = append(upcomingNodes, scheduler_utils.DeepCopyTemplateNode(nodeTemplate, i))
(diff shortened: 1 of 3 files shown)


MaciekPytel:
Nice catch! Sorry for introducing that issue :(

/lgtm
/approve

PR #4094, opened 2021-05-20
Update vendor to v1.22.0-alpha.1

brett-elliott:


brett-elliott:
Note: the most interesing thing is I copied/pasted code that was deleted by the scheduler team into:

utils/scheduler/scheduler.go

here is where they removed the code:

https://github.com/kubernetes/kubernetes/commit/7286f9712a924307c0403f9de7b0335b4597d378#

MaciekPytel:
Can you fix the gofmt so the presubmit passes? Otherwise lgtm.
Thanks for doing this!

PR #4103, opened 2021-05-26
Addon resizer sync

jbartosik:
We need to update client-go if we want it to continue working in K8s 1.22+ ([issue in k/k](https://github.com/kubernetes/kubernetes/issues/101286)).


bskiba:
I think you need to tun 'go mod vendor'.

Otherwise let me know how the testing went

PR #4112, opened 2021-06-02
addon-resizer: Fix --poll-period is not working in 1.8.13

nonylene:
In Addon-resizer 1.8.13 or earlier, pod_nanny has a problem that `--poll-period` flag does not change the poll period.

This is caused by a bug while parsing the flag: pollPeriod value from `flag.Int` is dereferenced too early (before `flag.Parse()`)
and it always be assigned the default value.

This commit resolves this issue by getting the pollPeriod value after `flag.Parse()`.

### Example

```
$ ./pod_nanny --poll-period 300000 --cpu ...
```

pod_nanny should crawl Pod state every 300 seconds but it now does every 10 seconds.


bskiba:
Thanks!
/lgtm
/approve

Issue #3965, opened 2021-03-22
VPA: invalid memory recommedation unit


gaurav1086:
**Which component are you using?**:
vertical pod scaler

**What version of the component are you using?**:
v0.9.2

Component version: v0.9.2

**What k8s version are you using (`kubectl version`)?**:

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.6", GitCommit:"fbf646b339dc52336b55d8ec85c181981b86331a", GitTreeState:"clean", BuildDate:"2020-12-18T12:09:30Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.8", GitCommit:"fd5d41537aee486160ad9b5356a9d82363273721", GitTreeState:"clean", BuildDate:"2021-02-17T12:33:08Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}

</pre></details>

**What environment is this in?**:
linux

**What did you expect to happen?**:

The manifest contains:

```
        resources:
          limits:
            cpu: 1000m
            memory: 1024Mi
          requests:
            cpu: 100m
            memory: 512Mi
```

**What happened instead?**:

The current usage stats:

```
$kubectl -n app-namespace describe pod my-pod-name

    Restart Count:  0
    Limits:
      cpu:     25390m
      memory:  731815514666m
    Requests:
      cpu:      2539m
      memory:   548861636
```

The limit memory units is showing with suffix 'm' (731815514666**m**) . As per the kubernetes doc, the memory resource is measured in bytes. You can express memory as a plain integer or a fixed-point integer with one of these suffixes: E, P, T, G, M, K, Ei, Pi, Ti, Gi, Mi, Ki. (https://kubernetes.io/docs/tasks/configure-pod-container/assign-memory-resource/#Memory%20units) . Is the 'm' suffix valid ?

**How to reproduce it (as minimally and precisely as possible)**:

Create a deployment and have it do a vpa-opt-in.

**Anything else we need to know?**:

sibucan:
[...]

PR #4113, opened 2021-06-02
Separate limits scaling between CPU & memory

sibucan:
See the [comment](https://github.com/kubernetes/autoscaler/issues/3965#issuecomment-851024267) I made for issue #3965

When the ratio between the original memory requests/limits on the Deployment is a floating point number instead of an integer, the calculated limit is expressed in millivalues, (https://play.golang.org/p/dR4SaXRBGmB) which is valid for CPU values, but not necessarily for memory values.

The purpose of this PR is to ensure it is expressed as bytes.

### DIFF
--- a/vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go
+++ b/vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go
@@ -85,7 +85,15 @@ func getProportionalResourceLimit(resourceName core.ResourceName, originalLimit,
 		result := *recommendedRequest
 		return &result, ""
 	}
-	result, capped := scaleQuantityProportionally( /*scaledQuantity=*/ originalLimit /*scaleBase=*/, originalRequest /*scaleResult=*/, recommendedRequest, noRounding)
+	if resourceName == core.ResourceCPU {
+		result, capped := scaleQuantityProportionallyCPU( /*scaledQuantity=*/ originalLimit /*scaleBase=*/, originalRequest /*scaleResult=*/, recommendedRequest, noRounding)
+		if !capped {
+			return result, ""
+		}
+		return result, fmt.Sprintf(
+			"%v: failed to keep limit to request ratio; capping limit to int64", resourceName)
+	}
+	result, capped := scaleQuantityProportionallyMem( /*scaledQuantity=*/ originalLimit /*scaleBase=*/, originalRequest /*scaleResult=*/, recommendedRequest, noRounding)
 	if !capped {
 		return result, ""
 	}
@@ -95,7 +103,7 @@ func getProportionalResourceLimit(resourceName core.ResourceName, originalLimit,
 
 // GetBoundaryRequest returns the boundary (min/max) request that can be specified with
 // preserving the original limit to request ratio. Returns nil if no boundary exists
-func GetBoundaryRequest(originalRequest, originalLimit, boundaryLimit, defaultLimit *resource.Quantity) *resource.Quantity {
+func GetBoundaryRequest(resourceName core.ResourceName, originalRequest, originalLimit, boundaryLimit, defaultLimit *resource.Quantity) *resource.Quantity {
 	if originalLimit == nil || originalLimit.Value() == 0 && defaultLimit != nil {
 		originalLimit = defaultLimit
 	}
@@ -107,7 +115,14 @@ func GetBoundaryRequest(originalRequest, originalLimit, boundaryLimit, defaultLi
 	if originalRequest == nil || originalRequest.Value() == 0 {
 		return boundaryLimit
 	}
-	result, _ := scaleQuantityProportionally(originalRequest /* scaledQuantity */, originalLimit /*scaleBase*/, boundaryLimit /*scaleResult*/, noRounding)
+
+	// Determine which scaling function to use based on resource type.
+	var result *resource.Quantity
+	if resourceName == core.ResourceCPU {
+		result, _ = scaleQuantityProportionallyCPU(originalRequest /* scaledQuantity */, originalLimit /*scaleBase*/, boundaryLimit /*scaleResult*/, noRounding)
+		return result
+	}
+	result, _ = scaleQuantityProportionallyMem(originalRequest /* scaledQuantity */, originalLimit /*scaleBase*/, boundaryLimit /*scaleResult*/, noRounding)
 	return result
 }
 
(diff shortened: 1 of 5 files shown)


vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go
```diff
@@ -107,7 +115,25 @@ func GetBoundaryRequest(originalRequest, originalLimit, boundaryLimit, defaultLi
...
+func GetBoundaryRequestMem(originalRequest, originalLimit, boundaryLimit, defaultLimit *resource.Quantity) *resource.Quantity {
+	if originalLimit == nil || originalLimit.Value() == 0 && defaultLimit != nil {
+		originalLimit = defaultLimit
+	}
+	// originalLimit not set, no boundary
+	if originalLimit == nil || originalLimit.Value() == 0 {
+		return &resource.Quantity{}
+	}
+	// originalLimit set but originalRequest not set - K8s will treat the pod as if they were equal
+	if originalRequest == nil || originalRequest.Value() == 0 {
```
jbartosik, line 133: Why this is not in `GetBoundaryRequestCPU`?  I think it should be there.

vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go
```diff
@@ -107,7 +115,25 @@ func GetBoundaryRequest(originalRequest, originalLimit, boundaryLimit, defaultLi
...
 		return boundaryLimit
 	}
-	result, _ := scaleQuantityProportionally(originalRequest /* scaledQuantity */, originalLimit /*scaleBase*/, boundaryLimit /*scaleResult*/, noRounding)
+	result, _ := scaleQuantityProportionallyCPU(originalRequest /* scaledQuantity */, originalLimit /*scaleBase*/, boundaryLimit /*scaleResult*/, noRounding)
+	return result
+}
+
+// GetBoundaryRequestMem returns the boundary (min/max) memory request that can be specified with
+// preserving the original limit to request ratio. Returns nil if no boundary exists
+func GetBoundaryRequestMem(originalRequest, originalLimit, boundaryLimit, defaultLimit *resource.Quantity) *resource.Quantity {
```
jbartosik, line 124: `GetBoundaryRequestMem` is (and should be) very similar to `GetBoundaryRequestCPU` (it looks like the only difference is that CPU operates on milli units and Memory operates on whole values). I'd rather avoid having two very similar pieces of code

jbartosik (changes requested):
Thanks & sorry for slow reply.

Please add some tests that cover cases when we expect this to behave differently

vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go
sibucan: It should be there, since `GetBoundaryRequestCPU` is just the original function, but renamed. Nothing was removed from it, rather a copy of it was created and named `GetBoundaryRequestMem` to get the scaled memory values (as opposed to scaled CPU values).

vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go
sibucan: That's fair -- I could revert these changes and keep the original function name, and instead add a new input parameter to the function (e.g. "cpu" or "memory") which will determine what scaling function to use. How does that sound?

jbartosik: Sounds good to me

sibucan:
@jbartosik made changes based on feedback & added tests for the changes made in the scaling function, lmk if you find any other issues

jbartosik (changes requested):
A couple more small changes

vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go
```diff
@@ -141,3 +156,26 @@ func scaleQuantityProportionally(scaledQuantity, scaleBase, scaleResult *resourc
 	}
 	return resource.NewMilliQuantity(math.MaxInt64, scaledQuantity.Format), true
 }
+
+// scaleQuantityProportionallyMem returns value which has the same proportion to scaledQuantity as scaleResult has to scaleBase
+// It also returns a bool indicating if it had to cap result to MaxInt64 milliunits.
```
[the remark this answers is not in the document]
sibucan: Good catch, updated.

vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go
```diff
@@ -119,9 +134,9 @@ const (
 	roundDownToFullUnit
 )
 
-// scaleQuantityProportionally returns value which has the same proportion to scaledQuantity as scaleResult has to scaleBase
+// scaleQuantityProportionallyCPU returns value which has the same proportion to scaledQuantity as scaleResult has to scaleBase
```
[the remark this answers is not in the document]
sibucan: Mentioned how the value returned can be either in units or in milliunits depending on the function being called.

jbartosik (commented):
Thank you. Looks good to me now. Please squash then I'll approve this PR

jbartosik (approved):

sibucan:
Commits have been squashed together to a single one.

sibucan:
@jbartosik I noticed that some tests were failing -- one of the tests failures was a simple change due to how memory resources are now displayed correctly, but the other failure was a bit more complicated to debug and it's related to the [rounding](https://github.com/kubernetes/autoscaler/blob/68a7d27b84d803eceadb2d5e82aff4dc517318d2/vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go#L133) logic.

It seems that previously, the rounding for memory values worked by taking advantage of the millivalue representation:
```
result: 357913941333m
round down: 357913941
result: 715827882666m
round down: 715827882

result: 1431655765333m
round up: 1431655766
result: 2863311530666m
round up: 2863311531
```

However, after these changes, the results just don't round the same way anymore:
```
result: 357913941
round down: 357913941
result: 715827882
round down: 715827882

result: 1431655765
round up: 1431655765
result: 2863311530
round up: 2863311530
```

This causes some of the tests to fail. I'm not sure how to fix this, since changing the [rounding scale](https://github.com/kubernetes/autoscaler/blob/68a7d27b84d803eceadb2d5e82aff4dc517318d2/vertical-pod-autoscaler/pkg/utils/vpa/limit_and_request_scaling.go#L134) only makes the rounded result differ from current tests even more, e.g. using a [scale of 1](https://github.com/kubernetes/apimachinery/blob/a644435e2c133d990b624858d9c01985d7f59adf/pkg/api/resource/quantity_test.go#L516) would make `1431655765` round up to `1431655770`, which is technically correct, but the tests expect `1431655766`.

I changed the expected values to reflect the new actuals, but I'm not sure of the implications of doing this. Any advice would be appreciated.

jbartosik:
@sibucan I still see two commits. Please squash

sibucan:
@jbartosik Hi! I just thought you'd want to re-review that other commit by itself before I squashed commits, but I went ahead and squashed again in the interest of time.

jbartosik (approved):
/lgtm
/approve

PR #4116, opened 2021-06-03
Ignore deleted nodes in upcoming calculation

ksmiley:
Nodes with the `ToBeDeletedByClusterAutoscaler` taint are currently counted as "upcoming" nodes even though they are not schedulable and are expected to be removed from the cluster. This can cause cluster-autsocaler to deadlock when there are pending pods: the autoscaler will not scale up because it thinks there is already a node starting that the pod will fit on, and the autoscaler will not terminate the tainted node because pending pods disable scaledown.

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -917,7 +917,7 @@ func (csr *ClusterStateRegistry) GetUpcomingNodes() map[string]int {
 		readiness := csr.perNodeGroupReadiness[id]
 		ar := csr.acceptableRanges[id]
 		// newNodes is the number of nodes that
-		newNodes := ar.CurrentTarget - (readiness.Ready + readiness.Unready + readiness.LongUnregistered)
+		newNodes := ar.CurrentTarget - (readiness.Ready + readiness.Unready + readiness.Deleted + readiness.LongUnregistered)
 		if newNodes <= 0 {
 			// Negative value is unlikely but theoretically possible.
 			continue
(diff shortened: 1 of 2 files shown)


MaciekPytel:
I think this will cause too much scale-up. The expectation is that nodes that are counted as Deleted in readiness have already been removed on cloudprovider side and they don't count towards NodeGroup.TargetSize(). In cases where this assumption is correct (which I think are the majority of cases) this change will lead to deleted nodes effectively being subtracted twice. This means CA will underestimate the number of upcoming nodes. 

The reason why we have this logic is that CA deletes VM on provider side, but it does not remove k8s Node object. It takes a while for k8s NodeController to actually delete the node object and so clusterstate needs to account for that. 

I can see 2 possible root causes for the issue you observe:
 * We expect that if CA taints the node it will either delete it or, in case of a failure, remove the taint. It's possible the error handling is not removing the taint as expected.
 * I'm not familiar with all cloudprovider implementations. Perhaps the provider you're using implements TargetSize() based on Node objects or otherwise counts the VM after CA issues a deletion request?

I think the way to tackle this would be to fix the underlying issue. 

---

All above being said - I think that part of cloudprovider logic comes from before we implemented asynchronous drain. Back than CA would just block on scale-down. However, currently it continues the logic and so I suspect the condition you describe could also happen temporarily while the node is being drained. That would be especially problematic in clusters that run pods with long graceful termination as those will make drain take much longer. If true that is clearly a bug in clusterstate, but I think the way to fix it would have to involve devising a way to distinguish nodes that are have taint and haven't been deleted yet from those that have already been deleted on cloudprovider side and subtracting just the first category from upcoming nodes calculation.

PR #4119, opened 2021-06-06
annotate fakeNodes so that cloudprovider implementations can identify them

marwanad:
This adds an annotation to the "fakeNode" used by CA for unregistered node. This can be useful for cloudproviders in the `DeleteNode()` method to have some custom error handling logic and also identify the reason for the `fakeNode`. (For now, it seems only GCE would make use of the nodes created with error path)

An example of how we want to leverage that: we don't want to decrement the cached NodeGroup capacity in Azure because the capacity reported by APIs won't include Failed instances so the cache will end up drifting unnecessarily from the real capacity and can inadvertently lead to a scale down if some conditions align. 

I've noticed the packet provider tries to identify some via providerId but I don't think this is a reliable way. Similar with the magnum provider via the object UID. I think a namespaced annotation would be a convenient way for identification.

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -51,6 +51,14 @@ const (
 
 	// NodeGroupBackoffResetTimeout is the time after last failed scale-up when the backoff duration is reset.
 	NodeGroupBackoffResetTimeout = 3 * time.Hour
+
+	// FakeNodeReasonAnnotation is an annotation added to the fake placeholder nodes CA has created
+	// Note that this don't map to real nodes in k8s and are merely used for error handling
+	FakeNodeReasonAnnotation = "k8s.io/cluster-autoscaler/fake-node-reason"
+	// FakeNodeUnregistered represents a node that is identified by CA as unregistered
+	FakeNodeUnregistered = "unregistered"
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -51,6 +51,14 @@ const (
 
 	// NodeGroupBackoffResetTimeout is the time after last failed scale-up when the backoff duration is reset.
 	NodeGroupBackoffResetTimeout = 3 * time.Hour
+
+	// FakeNodeReasonAnnotation is an annotation added to the fake placeholder nodes CA has created
+	// Note that this don't map to real nodes in k8s and are merely used for error handling
+	FakeNodeReasonAnnotation = "k8s.io/cluster-autoscaler/fake-node-reason"
```
MaciekPytel, line 57: Given that the stated goal is to use those annotations in cloudprovider, won't declaring them here risk cyclic dependency problems? I think it's much cleaner to aim at situation where cloudprovider doesn't need to import core CA modules. 
Maybe we can declare those consts in cloudprovider.go instead? This has an added benefit of making them easier to discover for anyone implementing new provider.

marwanad: yeah - good call on that!

PR #4130, opened 2021-06-10
dont proactively decrement azure cache for unregistered nodes

marwanad:
This is because for unregistered failed nodes, the capacity field on the VMSS already excludes and so this avoids a scenario where the cached capacity drifts significantly from the VMSS one (if cache TTL is long) and inadvertently lead to a scale down on the next scale up operation.

/area provider/azure


marwanad:
gonna add a UT tomorrow.

feiskyer (commented):
/lgtm
/approve

feiskyer:
@marwanad do we need cherry-pick this to older releases?

PR #4131, opened 2021-06-10
Added changes to support alternative recommender

### LINKED ISSUE #3913: Need support for customized recommender in Vertical Pod Autoscaler (VPA)

wangchen615:
This update is to support customized recommender in VPA.
The KEP is here: https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa

resolve #3913 
close #3913


bskiba:
@kgolab for visibility

vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go
```diff
@@ -90,6 +90,9 @@ type VerticalPodAutoscalerSpec struct {
 	// resources for all containers in the pod, without additional constraints.
 	// +optional
 	ResourcePolicy *PodResourcePolicy `json:"resourcePolicy,omitempty" protobuf:"bytes,3,opt,name=resourcePolicy"`
+
+	// Denote the recommender to be used
+	RecommenderName []string `json:"recommenderName,omitempty" protobuf:"bytes,1,opt,name=recommenderName"`
```
kgolab, line 95: 1 is already used for TargetRef, it should 4 as in KEP.

vertical-pod-autoscaler/examples/hamster-default-recommender.yaml
```diff
@@ -0,0 +1,56 @@
...
+# It also creates a corresponding Vertical Pod Autoscaler that adjusts the
+# requests.
+# Note that the update mode is left unset, so it defaults to "Auto" mode.
+---
+apiVersion: "autoscaling.k8s.io/v1"
+kind: VerticalPodAutoscaler
+metadata:
+  name: hamster-default-recommender-vpa
+spec:
+  recommenderName: ['default']
```
kgolab, line 13: Why do we need this example?

If this is to show that the default recommender might be specified explicitly, I think it would be good to state here (and maybe in the hamster.yaml) that recommenderName does not have to be set.
But maybe it would be enough to simply modify hamster.yaml with the comment how to specify the recommender.

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -330,11 +331,44 @@ func (feeder *clusterStateFeeder) GarbageCollectCheckpoints() {
...
+				return true
+			}
+		}
+		return false
+	}
+
+	klog.V(3).Infof("Start selecting the vpaCRDs.")
+	var vpaCRDs []*vpa_types.VerticalPodAutoscaler
+	for _, vpaCRD := range allVpaCRDs {
+		currentRecommenderName := feeder.clusterState.RecommenderName
```
kgolab, line 353: Could you please explain the approach here?

From the KEP I understood that the alternative recommenders would be run as separate binaries. Do you see these binaries still based on default VPA code (how would this work then)?
If the alternative recommenders would not share code base with default VPA I think the change here can be simplified to only look for 'empty' and "default", without the need of dynamically storing recommenderName in ClusterState.

vertical-pod-autoscaler/pkg/recommender/model/cluster.go
```diff
@@ -38,6 +38,8 @@ const (
 // events (container OOMs).
 // All input to the VPA Recommender algorithm lives in this structure.
 type ClusterState struct {
+	// Denote the recommender name
+	RecommenderName string
```
kgolab, line 42: As commented above I don't yet understand why we need to store this dynamically in the default recommender.

kgolab:
Thank you for the PR.

I'd like to understand if we really need to propagate recommenderName throughout the code of default VPA:
- if we don't the change could be simplified to yaml files, types.go and (parts of) cluster_feeder.go,
- if we do: could please explain why (so likely how you plan to use the code)?

Looking at the KEP I think we're missing an update in validateVPA. Or have you already decided to support multiple recommenders (changes in cluster_feeder.go might suggest so)?

vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go
Jeffwan, line 95: for the api part, I am thinking we can leave more complexity here like defining recommender strategies for multiple recommenders, but that can be left for future use cases.

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -330,11 +331,44 @@ func (feeder *clusterStateFeeder) GarbageCollectCheckpoints() {
...
 	// List VPA API objects.
-	vpaCRDs, err := feeder.vpaLister.List(labels.Everything())
+	allVpaCRDs, err := feeder.vpaLister.List(labels.Everything())
 	if err != nil {
 		klog.Errorf("Cannot list VPAs. Reason: %+v", err)
 		return
 	}
+
+	// Define a findElement function to check if clusterState.RecommenderName is in the vpaCRD.Spec.RecommenderName
+	findElement := func(slice []string, val string) bool {
```
Jeffwan, line 341: Do you want to limit recommender to only one at this moment?

vertical-pod-autoscaler/pkg/recommender/checkpoint/checkpoint_writer_test.go
```diff
@@ -40,6 +40,7 @@ var (
 		model.ResourceCPU:    model.CPUAmountFromCores(3.14),
 		model.ResourceMemory: model.MemoryAmountFromBytes(3.14e9),
 	}
+	recommenderName = "default"
```
Jeffwan, line 43: Seem we have a few places need this default name. (cluster_feeder_test.go#60) can we have a const defined somewhere and all tests can share same reference.

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
Jeffwan: I am a little bit confused here. Is the design target for `algorithm recommender` or  `controller level recommender`?  
I thought we want to have better abstraction on the algorithm and metrics part so that we can still leverage reconcile logic. But if that's the controller level. We should start different recommender runloop at the beginning. 

@kgolab @wangchen615

vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go
```diff
@@ -90,6 +90,9 @@ type VerticalPodAutoscalerSpec struct {
 	// resources for all containers in the pod, without additional constraints.
 	// +optional
 	ResourcePolicy *PodResourcePolicy `json:"resourcePolicy,omitempty" protobuf:"bytes,3,opt,name=resourcePolicy"`
+
+	// Denote the recommender to be used
```
bskiba, line 94: Should we mention that for now only a single recommender is supported?

bskiba, line 94: So the comment could be: names of recommenders that should generate recommendations for this VPA object. Only single recommender per object is supported as of now.

vertical-pod-autoscaler/examples/hamster-default-recommender.yaml
wangchen615: @kgolab , yes, I will just add comments in the default `hamster.yaml`

vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go
wangchen615: Yes, it should be 4. Thanks.

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
wangchen615: @kgolab , yes, the alternative recommenders are supposed to run as separate binaries. However, I do not want to exclude the possibility that some other recommender is reusing part of the default recommenders' code. Especially for functions like `loadVPAs`, which will apparently be reusable in other recommender binaries.

> Could you please explain the approach here?
> 
> From the KEP I understood that the alternative recommenders would be run as separate binaries. Do you see these binaries still based on default VPA code (how would this work then)?
> If the alternative recommenders would not share code base with default VPA I think the change here can be simplified to only look for 'empty' and "default", without the need of dynamically storing recommenderName in ClusterState.

wangchen615: @Jeffwan , The KEP is about supporting `controller level recommender`. We do start different recommender runloop at the beginning by defining `const RecommenderName = "default"` in `pkg/recommender/main.go`

> I am a little bit confused here. Is the design target for `algorithm recommender` or `controller level recommender`?
> I thought we want to have better abstraction on the algorithm and metrics part so that we can still leverage reconcile logic. But if that's the controller level. We should start different recommender runloop at the beginning.
> 
> @kgolab @wangchen615

wangchen615:
@kgolab Thanks so much for your review.
- As I mentioned above, keeping `recommenderName` will make the default recommender code easier to be reused when others developed their own alternative recommenders. 
  - Some functions like `loadVPAs` in reconcilers are definitely reusable.
  - However, other recommenders may rely on completely different metric collectors, such as Prometheus, and run different logics to come up with the target, lower bound and upper bound for requests.
  - For example, I am writing a customized recommender that will reuse the ClusterStateFeederFactory but rewrite the metric client.
  - At the same time, we provide support for the controller-level alternative recommender, so developers also have the freedom to develop their own recommender from scratch.

> Thank you for the PR.
> 
> I'd like to understand if we really need to propagate recommenderName throughout the code of default VPA:
> * if we don't the change could be simplified to yaml files, types.go and (parts of) cluster_feeder.go,
> * if we do: could please explain why (so likely how you plan to use the code)?
> 
> Looking at the KEP I think we're missing an update in validateVPA. Or have you already decided to support multiple recommenders (changes in cluster_feeder.go might suggest so)?

Yes, sorry for missing the validateVPA. I added back.

And yes, we want to support multiple VPA recommenders in the future. However, we may need a new KEP on how to cache the RecommendedPodResources for different recommenders in VPAStatus and how to select one recommendation among multiple recommenders when evicting and admitting the pod. We could initiate discussions once this first step PR is merged.

vertical-pod-autoscaler/examples/hamster-default-recommender.yaml
```diff
@@ -0,0 +1,56 @@
...
+  selector:
+    matchLabels:
+      app: hamster-default-recommender
+  replicas: 2
+  template:
+    metadata:
+      labels:
+        app: hamster-default-recommender
+    spec:
+      securityContext:
```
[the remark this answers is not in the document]
wangchen615: @Jeffwan , there is no particular reason. It was just copied from an old example with added recommenderName field. According to @kgolab's comment, I will delete this YAML and just add comments in the default `hamster.yaml`.

vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go
wangchen615: Yes, I agree. If we will later contribute more recommendation algorithms to the default recommender, we can use recommenderName as a field to select one among candidate algorithms.

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
wangchen615: Yes, the current solution can only support one recommender. Otherwise, two recommenders will overwrite the RecommendedPodResources.

vertical-pod-autoscaler/pkg/recommender/checkpoint/checkpoint_writer_test.go
wangchen615: > Seem we have a few places need this default name. (cluster_feeder_test.go#60) can we have a const defined somewhere and all tests can share same reference.

@Jeffwan , do you have a suggestion on where to define it?
I can only think of adding a new go file in utils to define the name. Do you prefer to add a new file just for this constant or do you prefer the current implementation?

vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go
wangchen615: Yes, we should. Thanks.

vertical-pod-autoscaler/pkg/recommender/checkpoint/checkpoint_writer_test.go
Jeffwan: This is minor. Current implementation is fine. Whenever we have more to support, we can refactor it later

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
wangchen615: The logic to check if only one recommender is defined in VPA is added in the handler of the admission controller.

vertical-pod-autoscaler/pkg/recommender/main.go
```diff
@@ -32,6 +32,9 @@ import (
 	"k8s.io/klog"
 )
 
+// RecommenderName denotes the current recommender name as the "default" one.
+const RecommenderName = "default"
```
bskiba, line 36: How about DefaultRecommenderName

bskiba (commented):
Some minor comments on my side. @kgolab waiting for your input on the current approach as explained by Chen

Jeffwan:
@wangchen615 It would be great to squash your commits. This repo doesn't squash it automatically.

wangchen615:
@mwielgus @bskiba @kgolab , this PR had no new comments for almost 2 months and all comments have been resolved. Could you help approve this?

jbartosik:
> @mwielgus, @schylek, @kgolab, @jbartosik, @krzysied, this PR had no new comments for almost 2 months and all comments have been resolved. Could you help approve this?

Sorry for the long delay. I'm looking at this right now

jbartosik (commented):
I'm trying to understand this and there is one thing I'm not sure about.

The idea I got during SIG was that the intent of this change is to allow A/B testing of recommenders. So I expected to be able to:

* Set two (or more) recommenders that will expose their recommendations (so I can compare those recommendations)
* Choose which recommendation to apply (when I see which one looks better)

Instead this looks to me like I can choose one recommender to provide a recommendation (that I will see and which will be applied).

I think I'm confused about something here.

I'll try to figure this out tomorrow. But if can see what I'm getting wrong I'd appreciate help

wangchen615:
> I'm trying to understand this and there is one thing I'm not sure about.
> 
> The idea I got during SIG was that the intent of this change is to allow A/B testing of recommenders. So I expected to be able to:
> 
> * Set two (or more) recommenders that will expose their recommendations (so I can compare those recommendations)
> * Choose which recommendation to apply (when I see which one looks better)
> 
> Instead this looks to me like I can choose one recommender to provide a recommendation (that I will see and which will be applied).
> 
> I think I'm confused about something here.
> 
> I'll try to figure this out tomorrow. But if can see what I'm getting wrong I'd appreciate help

Hi, @jbartosik , there are two proposals that allow A/B testing eventually in two steps:
1. Support using one recommender for one VPA object, but the recommender can use a customized one. This is the initial proposal [Support Customized Recommenders for Vertical Pod Autoscalers](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa)
2. Support multiple recommenders for a single VPA object. The proposal slides are [here](https://docs.google.com/presentation/d/1TtFNsCcEnbKki9xe31E8beNMq0fjm2Ek_-jXiGWst-8/edit?usp=sharing).

Once this PR is merged, we will move forward to add the second KEP and create a new PR accordingly.

jbartosik:
@wangchen615 Thank you for the explanation.

I'm not sure if this is the best way to get an API that allows A/B testing of recommenders.

If I understand the proposal in the presentation correctly we'd make a breaking API change (remove `VerticalPodAutoscalerStatus.Recommendation` field and replace it with `VerticalPodAutoscalerStatus.Recommendations`). This means we'll need to introduce a new API version. If we do that then `v1` API will be a bit weird. It will have a repeated `VerticalPodAutoscalerSpec.RecommenderNames` field but only one value will matter.

I think it would be better to:
1. Allow A/B testing without breaking API changes, or 
2. If that's not feasible then in the `v1` API support only one recommender in a clearer way. Add a `string` field `VerticalPodAutoscalerSpec.RecommenderName` (instead of `RecommenderNames []string`)

I'll think if 1. is feasible and get back to you later this week.

Also I want to make sure we don't make extending the API too hard for the future contributors. For example if in the future we might want to allow users to use CPU recommendation from one recommender but memory recommendation from another. I'd like to make it possible without a breaking API change.

wangchen615:
@jbartosik Changing the API from the string of `VerticalPodAutoscalerStatus.Recommendation` to the list of `VerticalPodAutoscalerStatus.Recommendations` was exactly what was requested by @bskiba and @kgolab. I believe the code now is supporting a list of `VerticalPodAutoscalerStatus.Recommendations` as the API but only one recommender for this PR.

In the next proposal where we support multiple recommenders, we do not need to further change the API but only support more than 1 recommender per VPA object.

Supporting CPU and Memory recommendations using different recommenders is a new use case to be added in the multiple recommender's support proposal. I can try to add some slides for that as well. However, I believe if the recommender is only updating CPU or memory, the list of recommenders in the API will work.

wangchen615:
@jbartosik , I updated the multiple recommender support slides here: https://docs.google.com/presentation/d/1TtFNsCcEnbKki9xe31E8beNMq0fjm2Ek_-jXiGWst-8/edit?usp=sharing

The `AllRecommendations` field is introduced in `VerticalPodAutoscalerStatus` so we do not need to change the existing `Recommendation` field but can track multiple recommenders' recommendations.

The `SelectedRecommender` is introduced in `VerticalPodAutoscalerSpec` so users can denote which recommender's recommendation is selected. If it is not specified, it will either use the `default` recommender's recommendation or if `default` not exists, it will use the first recommender specified in the list of `RecommenderNames`.

kgolab:
> @jbartosik Changing the API from the string of `VerticalPodAutoscalerStatus.Recommendation` to the list of `VerticalPodAutoscalerStatus.Recommendations` was exactly what was requested by @bskiba and @kgolab. I believe the code now is supporting a list of `VerticalPodAutoscalerStatus.Recommendations` as the API but only one recommender for this PR.

Just double-checking if we're not mixing up two things - IIRC we asked for introducing RecommenderNames (string[]) instead of RecommenderName (string) to support multiple recommenders in the future.

What @jbartosik asked is that we should align this change with what happens to Recommendations field.

When discussing the KEP (#3914) we excluded discussion about multiple recommender / statuses for the time being but with the new proposal you showed (https://docs.google.com/presentation/d/1TtFNsCcEnbKki9xe31E8beNMq0fjm2Ek_-jXiGWst-8/edit?usp=sharing), it's becoming clearer which directions are possible.

It would be great if this PR was aligned with the direction chosen for multiple recommenders, that's why we're trying to address the future proposal (slides) at the same time as the PR, to reduce uncertainty and make sure we don't end up with a half-way API (so either move all the changes to v2 or make them all in v1, compatible with current API).

Finally: I'm sorry for the too-long a delay before my comments - it's been a mix of very busy times for me but this should be over now. Thank you for your patience.

wangchen615:
@kgolab , yes, that's correct.
> Just double-checking if we're not mixing up two things - IIRC we asked for introducing RecommenderNames (string[]) instead of RecommenderName (string) to support multiple recommenders in the future.

Yes, so in the new proposal, we introduce the `VerticalPodAutoscalerStatus.AllRecommendations` field and `VerticalPodAutoscalerSpec.SelectedRecommender` field. So `VerticalPodAutoscalerStatus.Recommendation` will keep the same and will only store the recommendations from `VerticalPodAutoscalerSpec.SelectedRecommender`. Please refer to the [slides](https://docs.google.com/presentation/d/1TtFNsCcEnbKki9xe31E8beNMq0fjm2Ek_-jXiGWst-8/edit?usp=sharing). In this way, we can make this PR aligned with future changes.

> It would be great if this PR was aligned with the direction chosen for multiple recommenders, that's why we're trying to address the future proposal (slides) at the same time as the PR, to reduce uncertainty and make sure we don't end up with a half-way API (so either move all the changes to v2 or make them all in v1, compatible with current API).

jbartosik (commented):
I thought about this and talked to a few people and there is one change I'd like to make to the [Support Customized Recommenders for Vertical Pod Autoscalers](https://github.com/kubernetes/autoscaler/tree/680a09400a3cc0eecd000eaf8ffd0747b2acafdb/vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa) to make it more future proof.

Instead of having `string` as recommender identifier we should have a struct. For now the struct can will have only one field (Name string) but in the future we can use it to pass arguments to recommenders.

For example: It looks like [In-Place Update of Pod Resources](https://github.com/kubernetes/enhancements/issues/1287) will be available. We might want a recommender that allows configuring how aggressive it should be (since it will be cheaper to apply recommendations).

This would require some changes to this PR.

I'll soon send a PR to modify Support Customized Recommenders for Vertical Pod Autoscalers.

wangchen615:
@jbartosik , I made the changes according to the suggestions in [PR#4329](https://github.com/kubernetes/autoscaler/pull/4329)

jbartosik (changes requested):
Please add a test that checks that:

1. default recommender acts when no recommender is specified
2. default recommender acts when it's explicitly configured
3. default recommender doesn't act when a different recommender is configured

Also our test dashboards are currently broken, please run e2e tests (recommender and full suites should be enough). Please use K8s older than 1.23 (dashboard are red because of problems in 1.23. (only `full` and `recommender` tests are broken so there is no need to run other test suites)

EDIT: [link to the dashboard](https://k8s-testgrid.appspot.com/sig-autoscaling-vpa#autoscaling-vpa-full)

vertical-pod-autoscaler/pkg/recommender/model/cluster.go
jbartosik: Please split this change (adding `RecommenderName` field) to a separate PR.

It will make this PR easier to review. It will also make it easier to see how we'd use the field in alternative recommenders

jbartosik: Please move this change (adding `RecommenderName` field) to a separate PR.

It will make this PR easier to review. It will also make it easier to see how we'd use RecommenderName field

jbartosik (commented):
Some more comments

wangchen615:
@jbartosik Just a clarification question about your comment "Please move this change (adding RecommenderName field) to a separate PR.
It will make this PR easier to review. It will also make it easier to see how we'd use the RecommenderName field". 

Do you mean for this PR, we do not pass the `RecommenderName` through `ClusterState` struct and only checks if `DefaultRecommenderName` exists or if VPA objects choose other recommender names?

And in a new PR, we add `RecommenderName` in `ClusterState` structs so `cluster.go` code in default recommender can be reused by others who develop their own recommenders?

jbartosik:
> @jbartosik Just a clarification question about your comment "Please move this change (adding RecommenderName field) to a separate PR. It will make this PR easier to review. It will also make it easier to see how we'd use the RecommenderName field".
> 
> Do you mean for this PR, we do not pass the `RecommenderName` through `ClusterState` struct and only checks if `DefaultRecommenderName` exists or if VPA objects choose other recommender names?
> 
> And in a new PR, we add `RecommenderName` in `ClusterState` structs so `cluster.go` code in default recommender can be reused by others who develop their own recommenders?

I think in this PR we should only have changes that add support for alternative recommenders to the default recommender. And in the follow up PR changes to allow reusing code in alternative recommenders.

So yes to both questions :)

wangchen615:
@jbartosik , thanks so much for your advice. I did add any tests to v1beta2, that's why I was confusing.

Your comment on the cluster size helped a lot and it seemed the cluster I was using had too small nodes to test VPA and all tests eventually failed. Please see the full log [here](https://drive.google.com/file/d/1kI1TuFmWp_8IWbnhsN7glYLEQSTrYhHx/view?usp=sharing).

According to your comment, I created another GKE cluster (1.20.10-gke.301) with one large node (e2-standard-16), and then all tests passed.

```
grep \"msg\" alt_recommender_e2e.log
{"msg":"Test Suite starting","total":3,"completed":0,"skipped":0,"failed":0}
{"msg":"PASSED [sig-autoscaling] [VPA] [full-vpa] [v1beta2] Pods under VPA have memory requests growing with usage","total":3,"completed":1,"skipped":4,"failed":0}
{"msg":"PASSED [sig-autoscaling] [VPA] [full-vpa] [v1beta2] Pods under VPA have cpu requests growing with usage","total":3,"completed":2,"skipped":23,"failed":0}
{"msg":"PASSED [sig-autoscaling] [VPA] [full-vpa] [v1beta2] OOMing pods under VPA have memory requests growing with OOMs","total":3,"completed":3,"skipped":35,"failed":0}
{"msg":"Test Suite completed","total":3,"completed":3,"skipped":38,"failed":0}
{"msg":"Test Suite starting","total":5,"completed":0,"skipped":0,"failed":0}
{"msg":"PASSED [sig-autoscaling] [VPA] [full-vpa] [v1] Pods under VPA have cpu requests growing with usage","total":5,"completed":1,"skipped":13,"failed":0}
{"msg":"PASSED [sig-autoscaling] [VPA] [full-vpa] [v1] Pods under VPA with default recommender explicitly configured have cpu requests growing with usage","total":5,"completed":2,"skipped":26,"failed":0}
{"msg":"PASSED [sig-autoscaling] [VPA] [full-vpa] [v1] Pods under VPA with non-recognized recommender explicitly configured deployment not updated by non-recognized recommender","total":5,"completed":3,"skipped":27,"failed":0}
{"msg":"PASSED [sig-autoscaling] [VPA] [full-vpa] [v1] OOMing pods under VPA have memory requests growing with OOMs","total":5,"completed":4,"skipped":28,"failed":0}
{"msg":"PASSED [sig-autoscaling] [VPA] [full-vpa] [v1] Pods under VPA have memory requests growing with usage","total":5,"completed":5,"skipped":36,"failed":0}
{"msg":"Test Suite completed","total":5,"completed":5,"skipped":39,"failed":0}
``` 

More details of full log message can also be seen [here](https://drive.google.com/file/d/1SCC8GpnHSB5dHHH915vq8kN6sO8myBeJ/view?usp=sharing).

I think we are ready to merge. Please take a final look and see if there are some last changes we want to make. Thanks.

> @wangchen615
> 
> To speed things up. There is one guess I have. If you're trying to add `Pods under VPA with default recommender explicitly configured have cpu requests growing with usage` and `Pods under VPA with non-recognized recommender explicitly configured deployment not updated by non-recognized recommender` tests for API `v1beta2` then I don't think they will work. This PR only extends `v1` API.
> 
> If this is the problem then:
> 
> * I think we shouldn't change `v1beta2` API but I'll check
> * So please let me know so I know that I should check,
> * If I'm correct about this then we can submit the PR with tests only for the `v1` API
>   
>   * So for now please don't add those tests for `v1beta2` API.
> 
> After I check
> 
> * If I'm correct about us not needing to change `v1beta2` API then I'll review new changes and this PR should be ready to submit soon.
> * If I'm wrong then we'll need to change `v1beta2` API too.
> 
> If this is not the problem you're facing then please give me error messages you see and how you get them and I'll do my best to help.
> 
> Thank you.

jbartosik (approved):
Looks good to me. Please squash changes

wangchen615:
> Looks good to me. Please squash changes

@jbartosik , thanks for the review. Yes, I did the rebase and squash.
The EasyCLA issue was due to this issue: https://github.com/kubernetes/org/issues/3068
It should be automatically merged if you do `/lgtm` and `/approve`

jbartosik (approved):
/lgtm
/approve

PR #4144, opened 2021-06-15
digitalocean: do not Refresh() on startup

timoreimann:
If the API is temporarily unavailable, cluster-autoscaler will be crash-looping on startup during the initial call to `Refresh()`. This makes for a bad user/operator experience since it aggravates differentiating between API and cluster/workload problems.

Let autoscaler start up and retry fetching node pool information from the API as part of the pre-existing, periodic sync. This should be no different to experiencing transient API problems during runtime.


MorrisLaw (approved):
lgtm

PR #4145, opened 2021-06-15
digitalocean: remove tag references from README

timoreimann:
An initial version of the DigitalOcean cloud provider implementation relied on tags to define the behavior but has since been transitioned to using the public DOKS API. Update the README accordingly.


MorrisLaw (approved):
:+1:

PR #4146, opened 2021-06-15
improve addon-resizer deployment example

pgier:
This fixes the comment at the beginning of the addon-resizer deployment example
so that it describes the correct key names to use for configuring the cpu
and memory parameters.


bskiba:
Thanks!
/lgtm
/approve

PR #4151, opened 2021-06-18
Watch Action instead of polling Server Status

fhofherr:
We have recevied an issue from one of our customers that their
autoscaler pod regularly crashes due to a nil pointer dereference panic.
Analyzing the code we found out that the autoscaler polls the server
status to find out if a server is running.

The Hetzner Cloud Go client is implemented in such a way that it does
not return an error if a resource could not be found. Instead it returns
nil for the error and the resource. Ususally this is not an issue.
However, in case the server creation fails the server gets deleted from
Hetzner Cloud. This in turn leads to nil being returned and the
abovementioned panic.

The Hetzner Cloud API implements a concept called Actions. Whenever a
long running process is triggered we return an object which can
be used to get information about the progress of the task. The Action
object reliably allows to detect if a server has been created and
provides access to any error that may have occured.

This commit replaces polling the server status with using the action
object.

### DIFF
--- a/cluster-autoscaler/cloudprovider/hetzner/hetzner_node_group.go
+++ b/cluster-autoscaler/cloudprovider/hetzner/hetzner_node_group.go
@@ -18,6 +18,10 @@ package hetzner
 
 import (
 	"fmt"
+	"math/rand"
+	"sync"
+	"time"
+
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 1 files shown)


LKaemmerling (approved):

blindpirate:
Thanks @fhofherr .

I've already built my custom CA image on top of this PR. It works perfectly and I no longer see crashing.

mwielgus (approved):
/lgtm
/approve

PR #4153, opened 2021-06-20
Add the ability to spin up Hetzner servers from custom snapshots

marvinpinto:
This comes in handy when using tools such as Packer to generate customized images.

I have a working docker image built off this feature and I've not run into issues thus far. The image below is linked from my `autoscaler` fork.

```bash
docker pull ghcr.io/marvinpinto/autoscaler:ba42c328
```

### DIFF
--- a/cluster-autoscaler/cloudprovider/hetzner/README.md
+++ b/cluster-autoscaler/cloudprovider/hetzner/README.md
@@ -5,11 +5,17 @@ The cluster autoscaler for Hetzner Cloud scales worker nodes.
 # Configuration
 
 `HCLOUD_TOKEN` Required Hetzner Cloud token.
+
 `HCLOUD_CLOUD_INIT` Base64 encoded Cloud Init yaml with commands to join the cluster, Sample [examples/cloud-init.txt for (Kubernetes 1.20.1)](examples/cloud-init.txt)
-`HCLOUD_IMAGE` Defaults to `ubuntu-20.04`, @see https://docs.hetzner.cloud/#images
+
+`HCLOUD_IMAGE` Defaults to `ubuntu-20.04`, @see https://docs.hetzner.cloud/#images. You can also use an image ID here (e.g. `15512617`), or a label selector associated with a custom snapshot (e.g. `customized_ubuntu=true`).
+
 `HCLOUD_NETWORK` Default empty , The name of the network that is used in the cluster , @see https://docs.hetzner.cloud/#networks
+
 `HCLOUD_SSH_KEY` Default empty , This SSH Key will have access to the fresh created server, @see https://docs.hetzner.cloud/#ssh-keys
+
 Node groups must be defined with the `--nodes=<min-servers>:<max-servers>:<instance-type>:<region>:<name>` flag.
+
 Multiple flags will create multiple node pools. For example:
 ```
 --nodes=1:10:CPX51:FSN1:pool1
--- a/cluster-autoscaler/cloudprovider/hetzner/README.md
+++ b/cluster-autoscaler/cloudprovider/hetzner/README.md
@@ -8,7 +8,7 @@ The cluster autoscaler for Hetzner Cloud scales worker nodes.
 
 `HCLOUD_CLOUD_INIT` Base64 encoded Cloud Init yaml with commands to join the cluster, Sample [examples/cloud-init.txt for (Kubernetes 1.20.1)](examples/cloud-init.txt)
 
-`HCLOUD_IMAGE` Defaults to `ubuntu-20.04`, @see https://docs.hetzner.cloud/#images. You can also use an image ID here (e.g. `15512617`), or a label selector associated with a custom snapshot (e.g. `customized_ubuntu=true`).
+`HCLOUD_IMAGE` Defaults to `ubuntu-20.04`, @see https://docs.hetzner.cloud/#images. You can also use an image ID here (e.g. `15512617`), or a label selector associated with a custom snapshot (e.g. `customized_ubuntu=true`). The most recent snapshot will be used in the latter case.
 
 `HCLOUD_NETWORK` Default empty , The name of the network that is used in the cluster , @see https://docs.hetzner.cloud/#networks
 
--- a/cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go
+++ b/cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go
@@ -72,18 +72,16 @@ func newManager() (*hetznerManager, error) {
 	// (e.g. "ubuntu-20.04"), or a label selector associated with an image
 	// snapshot. In the latter case it will use the most recent snapshot.
 	image, _, err := client.Image.Get(ctx, imageName)
-	if err != nil || image == nil {
-		labelSelector := strings.Split(imageName, "=")
-		if len(labelSelector) != 2 {
-			return nil, fmt.Errorf("unable to find image %s: invalid label selector", imageName)
-		}
-
+	if err != nil {
+		return nil, fmt.Errorf("unable to find image %s: %v", imageName, err)
+	}
+	if image == nil {
 		images, err := client.Image.AllWithOpts(ctx, hcloud.ImageListOpts{
 			Type:   []hcloud.ImageType{hcloud.ImageTypeSnapshot},
 			Status: []hcloud.ImageStatus{hcloud.ImageStatusAvailable},
 			Sort:   []string{"created:desc"},
 			ListOpts: hcloud.ListOpts{
-				LabelSelector: fmt.Sprintf("%s=%s", labelSelector[0], labelSelector[1]),
+				LabelSelector: imageName,
 			},
 		})
 
--- a/cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go
+++ b/cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go
@@ -39,7 +39,7 @@ type hetznerManager struct {
 	nodeGroups     map[string]*hetznerNodeGroup
 	apiCallContext context.Context
 	cloudInit      string
-	image          string
+	image          *hcloud.Image
 	sshKey         *hcloud.SSHKey
 	network        *hcloud.Network
 }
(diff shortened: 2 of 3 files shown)


mwielgus:
cc: @LKaemmerling

LKaemmerling:
Hey @marvinpinto,

thank you for your contribution. I would suggest using the ID of the image instead of the description. The description is not unique and can be changed really fast. The ID is stable. 

Another point:
You changed the vendored hcloud-go, this should be avoided because those changes will be overwritten when the hcloud-go is updated here. I would suggest you implement your feature in the manager using the hcloud-go List/All method for images within the Manager https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go#L47. This could also be done via labels (& label selectors)

hcloud-go All Method with Filtering: https://pkg.go.dev/github.com/hetznercloud/hcloud-go/hcloud#ImageClient.AllWithOpts

marvinpinto:
Hey @LKaemmerling, thank you for taking time to review this!


> You changed the vendored hcloud-go, this should be avoided because those changes will be overwritten when the hcloud-go is updated here. I would suggest you implement your feature in the manager using the hcloud-go List/All method for images within the Manager

Understood, I will move all logic out of hcloud-go :+1:


> I would suggest using the ID of the image instead of the description. The description is not unique and can be changed really fast. The ID is stable.

The original lookup for an image was done purely using the image name, e.g. "ubuntu-20.04". This of course works well for pre-baked images but not so well for custom snapshots because it seems that snapshots don't get a `name` attribute, I verified this in the API.

In the lookup function I modified, it first attempts to look up an image by its ID (e.g. if someone supplies an int), then if it doesn't find anything it attempts a lookup using the `name` attribute (e.g. "ubuntu-20.04"), and finally if it still doesn't find anything it attempts a lookup using the `description` attribute. I kept it this way so it would not be a breaking change.

In the final case, it returns the most recent (custom snapshot) image, which is useful in scenarios where these images are churned regularly and one only wishes to use the newest.

Do you still have concerns about using the `description` attribute?

LKaemmerling:
> Hey @LKaemmerling, thank you for taking time to review this!
> 
> 
> 
> 
> 
> > You changed the vendored hcloud-go, this should be avoided because those changes will be overwritten when the hcloud-go is updated here. I would suggest you implement your feature in the manager using the hcloud-go List/All method for images within the Manager
> 
> 
> 
> Understood, I will move all logic out of hcloud-go :+1:
> 
> 
> 
> 
> 
> > I would suggest using the ID of the image instead of the description. The description is not unique and can be changed really fast. The ID is stable.
> 
> 
> 
> The original lookup for an image was done purely using the image name, e.g. "ubuntu-20.04". This of course works well for pre-baked images but not so well for custom snapshots because it seems that snapshots don't get a `name` attribute, I verified this in the API.
> 
> 
> 
> In the lookup function I modified, it first attempts to look up an image by its ID (e.g. if someone supplies an int), then if it doesn't find anything it attempts a lookup using the `name` attribute (e.g. "ubuntu-20.04"), and finally if it still doesn't find anything it attempts a lookup using the `description` attribute. I kept it this way so it would not be a breaking change.
> 
> 
> 
> In the final case, it returns the most recent (custom snapshot) image, which is useful in scenarios where these images are churned regularly and one only wishes to use the newest.
> 
> 
> 
> Do you still have concerns about using the `description` attribute?
> 
> 

The description is basically just a description (this sentence should get an Oscar!). You should not query against it as it may change easily. For your use case I would suggest using labels on the snapshot and then a label selector as a way to query all images with this selector. Ideally this should be only one image, but if there are more images with a specific label you can use the most recent one.

So as I said using the description is not recommended. Also you already (without any change) are able to use the ID of the snapshot.

LKaemmerling:
What needs to be taken into account as well is that every request you send against the API counts towards your Ratelimit. So you shouldn't to this on every node creation (because this might be a problem when you need a lot of nodes in a short time)

marvinpinto:
> For your use case I would suggest using labels on the snapshot and then a label selector as a way to query all images with this selector

Sounds good to me :+1:


> What needs to be taken into account as well is that every request you send against the API counts towards your Ratelimit

I was thinking the logic to retrieve the image ID could go somewhere [in this block](https://github.com/kubernetes/autoscaler/blob/1d4e6dff1bcf42b87fac9430e3958be60b60fc67/cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go#L47-L111). Which I believe would only run on startup and use the "found" image ID until the autoscaler is restarted.

Does this sound reasonable to you, or would you prefer I implement caching + periodic updating (possibly via the [Refresh](https://github.com/kubernetes/autoscaler/blob/1d4e6dff1bcf42b87fac9430e3958be60b60fc67/cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go#L113-L117) mechanism)?

LKaemmerling:
> 
> Does this sound reasonable to you, or would you prefer I implement caching + periodic updating (possibly via the [Refresh](https://github.com/kubernetes/autoscaler/blob/1d4e6dff1bcf42b87fac9430e3958be60b60fc67/cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go#L113-L117) mechanism)?
> 
> 
Personally from a hcloud provider perspective I'm fine with both 😀 but I guess using the refresh would be the best option

cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go
```diff
@@ -55,18 +55,47 @@ func newManager() (*hetznerManager, error) {
...
+	// variable. This value can either be an image ID itself (an int), a name
+	// (e.g. "ubuntu-20.04"), or a label selector associated with an image
+	// snapshot. In the latter case it will use the most recent snapshot.
+	image, _, err := client.Image.Get(ctx, imageName)
+	if err != nil || image == nil {
+		images, err := client.Image.AllWithOpts(ctx, hcloud.ImageListOpts{
+			Type:   []hcloud.ImageType{hcloud.ImageTypeSnapshot},
+			Status: []hcloud.ImageStatus{hcloud.ImageStatusAvailable},
+			Sort:   []string{"created:desc"},
+		})
```
marvinpinto, line 80: Note that I cannot use a `label_selector` to query against the Hetzner API here since it has not been added as an [option](https://github.com/kubernetes/autoscaler/blob/1d4e6dff1bcf42b87fac9430e3958be60b60fc67/cluster-autoscaler/cloudprovider/hetzner/hcloud-go/hcloud/image.go#L129-L138). This can be refactored once that option is available.

marvinpinto: Darn this is not true, need to re-do this.

marvinpinto:
I've rebased this PR against `master` and updated it implementing all of @LKaemmerling's suggestions. Let me know if I missed anything!

My updated docker image is available at:
```bash
docker pull ghcr.io/marvinpinto/autoscaler:d234dc55
```

cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go
```diff
@@ -55,18 +55,45 @@ func newManager() (*hetznerManager, error) {
...
+		if len(labelSelector) != 2 {
+			return nil, fmt.Errorf("unable to find image %s: invalid label selector", imageName)
+		}
+
+		images, err := client.Image.AllWithOpts(ctx, hcloud.ImageListOpts{
+			Type:   []hcloud.ImageType{hcloud.ImageTypeSnapshot},
+			Status: []hcloud.ImageStatus{hcloud.ImageStatusAvailable},
+			Sort:   []string{"created:desc"},
+			ListOpts: hcloud.ListOpts{
+				LabelSelector: fmt.Sprintf("%s=%s", labelSelector[0], labelSelector[1]),
```
LKaemmerling, line 86: You should pass the raw `imageName` variable here instead of generating it again

cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go
```diff
@@ -55,18 +55,45 @@ func newManager() (*hetznerManager, error) {
...
+		imageName = "ubuntu-20.04"
+	}
+
+	// Search for an image ID corresponding to the supplied HCLOUD_IMAGE env
+	// variable. This value can either be an image ID itself (an int), a name
+	// (e.g. "ubuntu-20.04"), or a label selector associated with an image
+	// snapshot. In the latter case it will use the most recent snapshot.
+	image, _, err := client.Image.Get(ctx, imageName)
+	if err != nil || image == nil {
+		labelSelector := strings.Split(imageName, "=")
```
LKaemmerling, line 76: I would suggest to not check this here because the label selectors can not only be `key=value`: https://docs.hetzner.cloud/#label-selector

cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go
```diff
@@ -55,18 +55,45 @@ func newManager() (*hetznerManager, error) {
...
+	if imageName == "" {
+		imageName = "ubuntu-20.04"
+	}
+
+	// Search for an image ID corresponding to the supplied HCLOUD_IMAGE env
+	// variable. This value can either be an image ID itself (an int), a name
+	// (e.g. "ubuntu-20.04"), or a label selector associated with an image
+	// snapshot. In the latter case it will use the most recent snapshot.
+	image, _, err := client.Image.Get(ctx, imageName)
+	if err != nil || image == nil {
```
LKaemmerling, line 75: I would suggest checking if err != nil earlier and not together with image == nil, because:

When the image is not found via ID or name the result will be `image=nil, err=nil`. When now an error occurs (Sample: API token is invalid) it will be `image=nil, err=<notnil>`, you would then try to access the API again, so I would recommend returning the error independently.

cluster-autoscaler/cloudprovider/hetzner/README.md
```diff
@@ -5,11 +5,17 @@ The cluster autoscaler for Hetzner Cloud scales worker nodes.
 # Configuration
 
 `HCLOUD_TOKEN` Required Hetzner Cloud token.
+
 `HCLOUD_CLOUD_INIT` Base64 encoded Cloud Init yaml with commands to join the cluster, Sample [examples/cloud-init.txt for (Kubernetes 1.20.1)](examples/cloud-init.txt)
-`HCLOUD_IMAGE` Defaults to `ubuntu-20.04`, @see https://docs.hetzner.cloud/#images
+
+`HCLOUD_IMAGE` Defaults to `ubuntu-20.04`, @see https://docs.hetzner.cloud/#images. You can also use an image ID here (e.g. `15512617`), or a label selector associated with a custom snapshot (e.g. `customized_ubuntu=true`).
+
```
LKaemmerling, line 12: line above: I suggest to add a note about the usage of the most recent snapshot when label selectors are used and there is more than one image associated with the label selector.

LKaemmerling:
> I've rebased this PR against `master` and updated it implementing all of @LKaemmerling's suggestions. Let me know if I missed anything!
> 
> My updated docker image is available at:
> 
> ```shell
> docker pull ghcr.io/marvinpinto/autoscaler:d234dc55
> ```

Hey @marvinpinto,

thank you this looks good, I just have a few smaller minor things but than from my side everything is ready :) Good job!

cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go
marvinpinto: Ahh, I was not aware a label selector did all of this, TIL!

marvinpinto:
Thank you again for taking the time @LKaemmerling, really appreciate it 😊

LKaemmerling (approved):

LKaemmerling:
Hey @marvinpinto,

this looks good to me! Great job :)

mwielgus (approved):
/lgtm
/approve

PR #4154, opened 2021-06-21
fix: add missing RBAC permissions to autoscaler chart

### LINKED ISSUE #4114: Autoscaler chart is missing some rules
Cluster Autoscaler Helm chart

Latest

v1.20.4-eks-6b7464 & v1.19.6-eks-49a6c0

EKS on AWS

Autoscaler can access to all required resource in the cluster or doesn't try to access to unneeded resources

**What happened instead?**:
cluster-autoscaler logs report a permission denied when trying to access to some csi related resources:
```
k8s.io/client-go/informers/factory.go:134: Failed to watch *v1.CSIDriver: failed to list *v1.CSIDriver: csidrivers.storage.k8s.io is forbidden: User "system:serviceaccount:REDACTED:REDACTED-cluster-autoscaler" cannot list resource "csidrivers" in API group "storage.k8s.io" at the cluster scope
````
and for `csistoragecapacities`

Launch cluster-autoscaler from the helm chart with service account and RBAC enabled

We fixed it by changing the [cluster role](https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/templates/clusterrole.yaml#L104) from:
```
- apiGroups:
  - storage.k8s.io
  resources:
  - storageclasses
  - csinodes
  verbs:
  - watch
  - list
  - get
``` 
to 
```
- apiGroups:
  - storage.k8s.io
  resources:
  - storageclasses
  - csinodes
  - csidrivers
  - csistoragecapacities
  verbs:
  - watch
  - list
  - get
```

But I'm not sure this change is the best/correct as I don't know the minimum permissions required for `csidrivers` and `csistoragecapacities`

MarcusNoble:
Fixes #4114 

Added RBAC permissions to watch, list and get `csidrivers` and `csistoragecapacities`

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -17,4 +17,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.9.2
+version: 9.9.3
--- a/charts/cluster-autoscaler/templates/clusterrole.yaml
+++ b/charts/cluster-autoscaler/templates/clusterrole.yaml
@@ -106,6 +106,8 @@ rules:
     resources:
     - storageclasses
     - csinodes
+    - csidrivers
+    - csistoragecapacities
     verbs:
     - watch
     - list


gjtempleton:
Sorry for the delay in reviewing.
Thanks for the PR!

/lgtm
/approve

stevehipwell:
I think the chart version here `9.10.01` is a typo and could cause issues due to not being a valid SemVer. This should probably be resolved before updating the `gh-pages` branch with the chart releases, I'm happy to open a PR to raise this to `9.10.2`.

gjtempleton:
@stevehipwell Good catch, you're 100% correct.

If you could raise that PR that would be great. I've stuck a hold on the PR to update the index.yaml for now.
I'll have a look into why the chart version not being SemVer didn't lead to a failed CI run.

MarcusNoble:
Ah yes! Good catch indeed. Looks like I made a typo when handling the merge conflict.

stevehipwell:
I've added PR https://github.com/kubernetes/autoscaler/pull/4183.

gjtempleton:
After a bit of digging, found why this slipped through linting. It seems both `chart-testing` and `helm`'s semver functionalities use `https://github.com/Masterminds/semver#parsing-semantic-versions` and use the  `NewVersion` rather than `StrictNewVersion` functionality.

With a bit of testing in a cluster this non-compliant version will still successfully install, as it seems all of the code dealing with SemVer handles these versions in the same way.

stevehipwell:
That's my understanding too, but other tools using `StrictNewVersion` will error. There is also a precedence for Helm itself to start erroring as it does from v3 with versions not even close to SemVer (the official Flink repo has this issue).

PR #4105, opened 2021-05-27
Addon resizer release 1.8 update deps

jbartosik:
We need to update client-go if we want it to continue working in K8s 1.22+ ([issue in k/k](https://github.com/kubernetes/kubernetes/issues/101286)).

Bump version so we can release and use version with updated dependencies

PR #4157, opened 2021-06-22
Addon resizer release 1.8 update deps 2

jbartosik:
I'm recreating #4105 because `continuous-integration/travis-ci/pr` got stuck.

We need to update client-go if we want it to continue working in K8s 1.22+ ([issue in k/k](https://github.com/kubernetes/kubernetes/issues/101286)).

Bump version so we can release and use version with updated dependencies


mwielgus (approved):
/lgtm
/approve

PR #4162, opened 2021-06-24
Add a flag to control DaemonSet eviction on non-empty nodes

x13n:


cluster-autoscaler/core/scale_down.go
```diff
@@ -1247,6 +1249,13 @@ func (sd *ScaleDown) deleteNode(node *apiv1.Node, pods []*apiv1.Pod, daemonSetPo
 	return status.NodeDeleteResult{ResultType: status.NodeDeleteOk}
 }
 
+func selectDSPodsForEviction(pods []*apiv1.Pod, shouldEvict bool) []*apiv1.Pod {
```
towca, line 1252: nit: I think dsPods would be a better name for the argument, right now it looks like this function would select DS pods from a wider list of pods.

towca (approved):
/lgtm
/approve
/hold
Feel free to unhold if you don't agree with the nit.

cluster-autoscaler/core/scale_down.go
x13n: I agree this name was confusing. Changed to "podsToEvict", since the purpose of this function is to select pods that should be evicted from the provided list.

towca: LGTM, thanks!

PR #4167, opened 2021-06-27
Increase the server creation timeout to give the initialization process some leeway

marvinpinto:
Currently, if the Hetzner Cloud initialization process takes longer than a minute, perhaps due to spinning up images from large snapshots (or other reasons), the server is deleted via the API and the autoscaler tries again. This results in a constant deletion/creation loop which could potentially get expensive.

I have a working docker image built off this fix and I've not run into issues thus far. The image below is linked from my autoscaler fork.

```
docker pull ghcr.io/marvinpinto/autoscaler:628f187a
```

cc: @LKaemmerling @fhofherr

### DIFF
--- a/cluster-autoscaler/cloudprovider/hetzner/hetzner_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/hetzner/hetzner_cloud_provider.go
@@ -40,7 +40,7 @@ const (
 	nodeGroupLabel         = hcloudLabelNamespace + "/node-group"
 	hcloudLabelNamespace   = "hcloud"
 	drainingNodePoolId     = "draining-node-pool"
-	serverCreateTimeout    = 1 * time.Minute
+	serverCreateTimeout    = 5 * time.Minute
 	serverRegisterTimeout  = 10 * time.Minute
 	defaultPodAmountsLimit = 110
 )


cluster-autoscaler/cloudprovider/hetzner/hetzner_cloud_provider.go
```diff
@@ -40,7 +40,7 @@ const (
 	nodeGroupLabel         = hcloudLabelNamespace + "/node-group"
 	hcloudLabelNamespace   = "hcloud"
 	drainingNodePoolId     = "draining-node-pool"
-	serverCreateTimeout    = 1 * time.Minute
+	serverCreateTimeout    = 5 * time.Minute
```
LKaemmerling, line 43: i guess it would make sense to make this configurable 🤔

LKaemmerling:
Hey @marvinpinto,

Good catch! I would prefer to make this configurable with a default of 5 minutes, then we are on the save side when for what ever reason someone has a really big snapshot and tries to create it in another location.

marvinpinto:
Sounds good, I considered making it configurable but didn't want to assume you all wanted to maintain more knobs. 

How does using the environment variable `HCLOUD_SERVER_CREATION_TIMEOUT` sound? I will default it to 5 mins. Do you also want this documented in the README?

LKaemmerling:
> Sounds good, I considered making it configurable but didn't want to assume you all wanted to maintain more knobs. 
> 
> 
> 
> How does using the environment variable `HCLOUD_SERVER_CREATION_TIMEOUT` sound? I will default it to 5 mins. Do you also want this documented in the README?
> 
> 

@marvinpinto in case you missed my reaction. I like the name 😀

marvinpinto:
Hello @LKaemmerling :wave:

Got side-tracked with a few other things, I'll try and wrap this PR up sometime next week :+1:

mwielgus:
Closing in favor of #4345.

PR #4170, opened 2021-06-28
added vpa hack for case of 1 replica

ofaakye:


jbartosik (changes requested):
Please explain what's purpose of this PR

PR #4172, opened 2021-06-29
Allow DaemonSet pods to opt in/out from eviction

x13n:
Changes to documentation will be done in a follow-up PR.

### DIFF
--- a/cluster-autoscaler/core/scale_down.go
+++ b/cluster-autoscaler/core/scale_down.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/processors/customresources"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator"
 	"k8s.io/autoscaler/cluster-autoscaler/utils"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/daemonset"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/deletetaint"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/core/scale_down.go
```diff
@@ -1249,11 +1252,17 @@ func (sd *ScaleDown) deleteNode(node *apiv1.Node, pods []*apiv1.Pod, daemonSetPo
 	return status.NodeDeleteResult{ResultType: status.NodeDeleteOk}
 }
 
-func podsToEvict(pods []*apiv1.Pod, shouldEvict bool) []*apiv1.Pod {
-	if shouldEvict {
-		return pods
+func podsToEvict(pods []*apiv1.Pod, evictByDefault bool) (evictable []*apiv1.Pod) {
```
MaciekPytel, line 1255: Can you rename this to make it clear it's related to ds and not 'regular' pods? Also maybe consider moving it to cluster-autoscaler/utils/daemonset?

MaciekPytel:
This PR only covers empty node deletion path. To respect annotation consistently I think you need to add filtering logic in drainNode() as well.

x13n:
It covers both paths, `podsToEvict` was already used before the call to drainNode.

cluster-autoscaler/core/scale_down.go
x13n: I think moving this to `cluster-autoscaler/utils/daemonset` will suffice as a naming change, since it will then become `daemonset.PodsToEvict`. Will move it there.

PR #4173, opened 2021-06-29
Document DaemonSet eviction opt in/out behavior

x13n:
/hold until https://github.com/kubernetes/autoscaler/pull/4172 is merged


cluster-autoscaler/FAQ.md
```diff
@@ -434,6 +435,30 @@ spec:
...
+
+```
+"cluster-autoscaler.kubernetes.io/enable-ds-eviction": "true"
+```
+
+It is also possible to disable DaemonSet pods eviction expicitly:
+
+
+```
+"cluster-autoscaler.kubernetes.io/enable-ds-eviction": "true"
```
MaciekPytel, line 453: s/true/false

x13n: Good catch, thanks!

Issue #4165, opened 2021-06-25
Redesign balance-similar-node-groups logic


jsravn:
**Which component are you using?**:
cluster-autoscaler

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:

Fix --balance-similar-node-groups to work by default for most people.

**Describe the solution you'd like.**:

There are dozens of issues in this repo about how `balance-similar-node-groups` doesn't work correctly. As of now, it seems it still doesn't work correctly with the least-waste expander. This is problematic - as least-waste is the only reasonable expander to use with large diversified machinedeployments that have zone affinity requirements (such as EBS).

The sample node templates, as far as I can tell, that are generated by cloud providers can never actually match real node infos. As a result, at least for the AWS cloudprovider, nodes are generally never created in empty node groups for the least-waste expander.

The details behind why this happens is mostly covered in https://github.com/kubernetes/autoscaler/issues/1676#issuecomment-568189444.

To summarize:

1. The AWS cloudprovider overestimates memory capacity significantly for instance types. In my testing, m5.4xlarge for instance is almost 2Gi off. The PR that was made a while ago to add a 1.5% tolerance is simply not enough. The memory capacity can also vary based on the Linux kernel used, so it's not as simple as picking an arbitrary tolerance.
2. There is a strange "free" concept which I still don't really understand the reasons behind. In my experience, this seems to always cause template node infos to not match, especially if you use static pods.
3. kubelet adds several fields to capacity/allocatable that none of the cloudproviders add as far as I can tell, like "hugepages-*" and "ephemeral-storage". The node balancing logic only considers node infos the same if they have all the same fields - so again, it will never consider the template node infos equivalent.

The entire logic seems overly complex to me. It could be there are some edge cases due to some of the other cloudproviders. But I think as is, it is far too restrictive right now. We should give users the ability to control this via labels exclusively in my opinion, which gives people the flexibility to define what things are considered similar, rather than a bunch of esoteric rules that are difficult to debug and understand.

So my proposal would be to move towards ONLY using labels as a way of comparing similar node groups - with the ability to ignore labels as needed. This provides a truly generic mechanism that should work for most people out of the box. In my particular use case, I want all nodes with the same AWS instance type to be considered similar.

**Describe any alternative solutions you've considered.**:

**Additional context.**:

jsravn:
[...]

PR #4174, opened 2021-06-29
Allow balancing by labels exclusively

jsravn:
Adds a new flag `--balance-label` which allows users to balance between node groups exclusively via labels for when `--balance-similar-node-groups` is being used.

This gives users the flexibility to specify the similarity logic themselves rather than trying to work within the existing heuristics.

This is a POC based on my idea in https://github.com/kubernetes/autoscaler/issues/4165.

An example use case is a heavily diversified AWS deployment with per-AZ node groups to accommodate zone affinity restrictions of EBS volumes. The current heuristic logic makes this difficult to accomplish in combination with the least-waste expander, as the template node info for empty node groups rarely matches existing nodes (for many reasons, as discussed in the linked issue and elsewhere). This change makes it very simple - just specify the labels you want to match on, such as `node.kubernetes.io/instance-type`, and you're done.

### DIFF
--- a/cluster-autoscaler/processors/nodegroupset/label_nodegroups.go
+++ b/cluster-autoscaler/processors/nodegroupset/label_nodegroups.go
@@ -1,0 +1,41 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package nodegroupset
+
+import (
+	klog "k8s.io/klog/v2"
+	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
+)
+
+// CreateLabelNodeInfoComparator returns a comparator that checks for node group similarity using the given labels.
+func CreateLabelNodeInfoComparator(labels []string) NodeInfoComparator {
+	return func(n1, n2 *schedulerframework.NodeInfo) bool {
+		return areLabelsSame(n1, n2, labels)
+	}
+}
+
+func areLabelsSame(n1, n2 *schedulerframework.NodeInfo, labels []string) bool {
+	for _, label := range labels {
+		val1 := n1.Node().ObjectMeta.Labels[label]
+		val2 := n2.Node().ObjectMeta.Labels[label]
+		if val1 != val2 {
+			klog.V(4).Infof("%s label did not match. %s: %s, %s: %s", label, n1.Node().Name, val1, n2.Node().Name, val2)
+			return false
+		}
+	}
+	return true
+}
--- a/cluster-autoscaler/config/autoscaling_options.go
+++ b/cluster-autoscaler/config/autoscaling_options.go
@@ -148,6 +148,9 @@ type AutoscalingOptions struct {
 	// BalancingExtraIgnoredLabels is a list of labels to additionally ignore when comparing if two node groups are similar.
 	// Labels in BasicIgnoredLabels and the cloud provider-specific ignored labels are always ignored.
 	BalancingExtraIgnoredLabels []string
+	// BalancingLabels is a list of labels to use when comparing if two node groups are similar.
+	// If this is set, only labels are used to compare node groups. It is mutually exclusive with BalancingExtraIgnoredLabels.
+	BalancingLabels []string
 	// AWSUseStaticInstanceList tells if AWS cloud provider use static instance type list or dynamically fetch from remote APIs.
 	AWSUseStaticInstanceList bool
 	// ConcurrentGceRefreshes is the maximum number of concurrently refreshed instance groups or instance templates.
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -172,6 +172,7 @@ var (
 
 	ignoreTaintsFlag                   = multiStringFlag("ignore-taint", "Specifies a taint to ignore in node templates when considering to scale a node group")
 	balancingIgnoreLabelsFlag          = multiStringFlag("balancing-ignore-label", "Specifies a label to ignore in addition to the basic and cloud-provider set of labels when comparing if two node groups are similar")
+ 	balancingLabelsFlag                = multiStringFlag("balancing-label", "Specifies a label to use for comparing if two node groups are similar, rather than the built in heuristics. Setting this flag disables all other comparison logic, and cannot be combined with --balancing-ignore-label.")
 	awsUseStaticInstanceList           = flag.Bool("aws-use-static-instance-list", false, "Should CA fetch instance types in runtime or use a static list. AWS only")
 	concurrentGceRefreshes             = flag.Int("gce-concurrent-refreshes", 1, "Maximum number of concurrent refreshes per cloud object type.")
 	enableProfiling                    = flag.Bool("profiling", false, "Is debug/pprof endpoint enabled")
@@ -248,6 +249,7 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 		NewPodScaleUpDelay:                 *newPodScaleUpDelay,
 		IgnoredTaints:                      *ignoreTaintsFlag,
 		BalancingExtraIgnoredLabels:        *balancingIgnoreLabelsFlag,
+		BalancingLabels:                    *balancingLabelsFlag,
 		KubeConfigPath:                     *kubeConfigFile,
 		NodeDeletionDelayTimeout:           *nodeDeletionDelayTimeout,
 		AWSUseStaticInstanceList:           *awsUseStaticInstanceList,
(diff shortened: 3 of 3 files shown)


bpineau (commented):
That's a great idea!
Unpredictable kernel memory reservations on AWS instances is a real issue for `balance-similar-node-groups`. We unsuccessfully tried to other approaches [here](https://github.com/kubernetes/autoscaler/pull/4000), [here](https://github.com/kubernetes/autoscaler/pull/3761), [here](https://github.com/kubernetes/autoscaler/issues/2892) and a few more times.

cluster-autoscaler/processors/nodegroupset/label_nodegroups.go
```diff
@@ -0,0 +1,41 @@
...
+		return areLabelsSame(n1, n2, labels)
+	}
+}
+
+func areLabelsSame(n1, n2 *schedulerframework.NodeInfo, labels []string) bool {
+	for _, label := range labels {
+		val1 := n1.Node().ObjectMeta.Labels[label]
+		val2 := n2.Node().ObjectMeta.Labels[label]
+		if val1 != val2 {
+			klog.V(4).Infof("%s label did not match. %s: %s, %s: %s", label, n1.Node().Name, val1, n2.Node().Name, val2)
```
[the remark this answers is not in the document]
jsravn: I thought it would be helpful to debug why things are not being balanced. Mostly from my struggles trying to debug the existing balancing code :).

jsravn:
I've been testing this in my dev clusters (1000+ nodes) and it is working really well. I'm getting perfect balancing across node groups with same instance types.

tsunamishaun:
I've had to disable balance similar with priority expander as it ignores the priority list. I have two node groups one for spot, one for on demand, same taint tolerations and same workergroup labels. Only difference is market option label which I dont use for affinity etc. I'm pretty sure this would fix my problem letting me re-enable balance similar and use -balance-label set to my workergroup label/topology?

jsravn:
> I've had to disable balance similar with priority expander as it ignores the priority list. I have two node groups one for spot, one for on demand, same taint and workergroup labels. Only difference is market option label which I dont use for affinity etc. I'm pretty sure this would fix my problem letting me re-enable balance similar and use -balance-label set to my workergroup label/topology?

I haven't tested it, but assuming balancing works w/ the priority expander, then yes this should help your usecase.

syscod3:
Hello. I was wondering if this PR can be merged now? It would solve a lot of people's issues.

cluster-autoscaler/processors/nodegroupset/label_nodegroups.go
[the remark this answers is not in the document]
jsravn: What level would be good? Or should I remove it completely?

jsravn:
I pushed changes for the review comments. I suppose it needs to be discussed at the SIG meeting?

cluster-autoscaler/processors/nodegroupset/label_nodegroups.go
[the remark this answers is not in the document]
dlaidlaw: I would actually like to see this loop through all labels and log all mismatches. I really like to see all the reasons, not just the first one. This often saves time.

jaypipes (changes requested):
@jsravn I support this addition. I think it's a simple solution to the problem you describe in #4165 and simple == good, at least for me!

Couple things I'd like to see, however:

1) Please add at least a unit test for the new functionality.
2) A discussion, either in the `areLabelsSame` code comment or in the [AWS FAQ section about multiple instance type node groups](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md#using-mixed-instances-policies-and-spot-instances) about how this technique will not work with ASGs with a mixed instance policy because the NodeInfo object will be returned for the first instance in the InstanceGroup and that instance will obviously only have a single `instance-type` label and therefore depending on which instance type different NodeGroups happened to launch first, the `instance-type` label may be different for them, which will cause the `areLabelsSame` call to return `false`.

jsravn:
@jaypipes I've addressed your comment and rebased. Let me know what you think.

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -304,6 +304,10 @@ spec:
                   - i3.2xlarge
 ```
 
+Similarly, if using the `balancing-label` flag, you should only choose labels which have the same value for all nodes in
+the node group.  Otherwise you may get unexpected results, as the flag values will vary based on the nodes created by
+the ASG.
```
jaypipes, line 309: :+1: thanks for adding this note, @jsravn

cluster-autoscaler/processors/nodegroupset/label_nodegroups_test.go
```diff
@@ -0,0 +1,68 @@
...
+			isSimilar:   true,
+		},
+	} {
+		t.Run(tc.description, func(t *testing.T) {
+			node1.ObjectMeta.Labels = tc.labels1
+			node2.ObjectMeta.Labels = tc.labels2
+			checkNodesSimilar(t, node1, node2, comparator, tc.isSimilar)
+		})
+	}
+}
```
jaypipes, line 68: ++ nice tests.

jaypipes (approved):
Great stuff, @jsravn, much appreciated!

TBBle:
To avoid any confusion for users, perhaps there should there be some kind of failure or complaint if both `balancing-ignore-label` and `balancing-label` are provided, even if it's just a log warning that the former will be ignored. A start-up failure with error seems reasonable to me too. The semantic of both flags provided is pretty clear (since 'ignored labels' includes all labels not listed in balancing-labels by default, so the ignore is redundant) except if the user provides the same label in both lists for some reason, which implies a misunderstanding of the intended behaviour.

cluster-autoscaler/processors/nodegroupset/label_nodegroups.go
```diff
@@ -0,0 +1,41 @@
...
+func CreateLabelNodeInfoComparator(labels []string) NodeInfoComparator {
+	return func(n1, n2 *schedulerframework.NodeInfo) bool {
+		return areLabelsSame(n1, n2, labels)
+	}
+}
+
+func areLabelsSame(n1, n2 *schedulerframework.NodeInfo, labels []string) bool {
+	for _, label := range labels {
+		val1 := n1.Node().ObjectMeta.Labels[label]
+		val2 := n2.Node().ObjectMeta.Labels[label]
```
TBBle, line 34: What happens if the specified label is not present on either NodeInfo? Wouldn't this _panic_?

This seems like an easy use-case to hit if I'm using a custom `example.com/balance-group-name` label for when I have duplicated a nodegroup across AZs, but am not using it for nodegroups which are not set up that way as they are AZ-local for reasons other than EBS-selection. The latter would still be fed into this comparison function if they otherwise matched the pod-to-be-created.

And once this isn't panicking, should the behaviour of 'balancing label is not present' be:
* "Don't use this nodeset", i.e. the balancing should not even be attempted if the pod-match-chosen `n1` doesn't have the desired label, and if the potential-match `n2` lacks the label, it always returns false; or
* "Absence is also a value", i.e. only match if both NodeInfos lack the label, or if both have the label and the same value for that label; or
* something else?

TBBle: e.g., to implement the "missing label excludes from balancing completely", I guess it'd be:
```suggestion
		val1, ok1 := n1.Node().ObjectMeta.Labels[label]
		val2, ok2 := n2.Node().ObjectMeta.Labels[label]
		if !ok1 {
			klog.V(8).Infof("%s label not present on %s.", label, n1.Node().Name)
			return false
		}
		if !ok2 {
			klog.V(8).Infof("%s label not present on %s.", label, n2.Node().Name)
			return false
		}
```

TBBle: Either way, it'd be good to see the behaviour in the case of one, the other, or both node groups lacking the relevant label in the tests, so we know what was intended.

artificial-aidan:
@jsravn are you still working on getting this PR in? Anything I can do to help with the comments from TBBle?

jsravn:
@artificial-aidan I've yet to have a single OWNER review this. I'm hesitant to keep working on this PR until I get some official feedback first. ping @towca

njablonski:
This, as written, would solve for some challenges I have with predicting and controlling the existing similarity logic.

cluster-autoscaler/processors/nodegroupset/label_nodegroups.go
jsravn: Good point! My initial thinking is that if the label is missing completely, it wouldn't be be considered similar to another node group with also a missing label. The absence of the label would indicate "never consider this similar".

jsravn: I've added tests and your suggestion. I made a slight change to use an existence check instead of comparing on the empty string - so if a user does explicitly set an empty label value, it will be considered, allow users to use label keys alone as ways to group nodes.

jsravn:
I've rebased and addressed the edge cases @TBBIe pointed out. @mwielgus would you or another reviewer be able to have a look?

jsravn:
/assign @mwielgus 

(just noticed towca is no longer in the OWNERS file)

grosser:
FYI used this patch and it kinda works

I created pending pods that need ~7 new nodes:

```
{"level":"INFO","message":"Estimated 1 nodes needed in sandbox-usw2c-AutoscalingGroup-C","pattern":"unknown"}
{"level":"INFO","message":"Splitting scale-up between 3 similar node groups: {sandbox-usw2c-AutoscalingGroup-C, sandbox-node-usw2a-AutoscalingGroup-A, sandbox-usw2b-AutoscalingGroup-B}","pattern":"unknown"}
{"level":"INFO","message":"Final scale-up plan: [{sandbox-usw2c-AutoscalingGroup-C 1->2 (max: 10)}]","pattern":"unknown"}
{"level":"INFO","message":"Scale-up: setting group sandbox-usw2c-AutoscalingGroup-C size to 2","pattern":"unknown"}
{"level":"INFO","message":"Setting asg sandbox-usw2c-AutoscalingGroup-C size to 2","pattern":"unknown"}
```

... but the math seems to not work out, it scaled up 1 and then next iteration again 1
and `master` scales up 7 at once
```
{"level":"INFO","message":"Best option to resize: sandbox-usw2a-AutoscalingGroup-A","pattern":"unknown"}
{"level":"INFO","message":"Estimated 7 nodes needed in sandbox-usw2a-AutoscalingGroup-A","pattern":"unknown"}
{"level":"INFO","message":"Final scale-up plan: [{sandbox-usw2a-AutoscalingGroup-A 1->8 (max: 10)}]","pattern":"unknown"}
```

elmiko (commented):
this seems like a nice feature, we should probably discuss it at a SIG meeting if we can't get more review here.

elmiko:
we talked about this PR at today's meeting, hopefully we'll get a few more reviews. i think it's a nice feature, code looks good to me.

/lgtm

towca:
Everything looks good to me here, thanks for a very useful contribution!

/lgtm
/approve

jbilliau-rcd:
I just switched to this, since my existing use of `--balance-ignore-labels` apparently broke at some point (it just spun up 11 nodes in one node group, RIP ). Using `balancing-label: eks.amazonaws.com/nodegroup-image`, I am not getting proper balancing across 3 node groups, 3 AZ's again. Thanks @jsravn !

PR #4175, opened 2021-07-01
update cluster-autoscaler chart to 1.21.0

mweibel:


gjtempleton:
Hey, thanks for the PR!

Can you please make the chart version bump a minor version bump rather than a patch version bump? It matches what we've done with [past PRs to bump the default image](https://github.com/kubernetes/autoscaler/pull/3880) and flags it as a larger change to users who are potentially using the default image tag.

mweibel:
sure, done :)

gjtempleton:
Brilliant, thanks again!
/lgtm
/approve

Issue #3979, opened 2021-03-30
Addon-resizer uses 'etcd_object_counts' metric which is deprecated starting from k8s 1.22


mborsz:
**Which component are you using?**:
addon-resizer

**What version of the component are you using?**:
1.8.12

Component version:

**What k8s version are you using (`kubectl version`)?**:

1.22
**What environment is this in?**:
all

**What did you expect to happen?**:

addon-resizer uses a metric that exists.

**What happened instead?**:

addon-resizer uses a metric that is no longer available:
https://github.com/kubernetes/autoscaler/blob/addon-resizer-release-1.8/addon-resizer/nanny/kubernetes_client.go#L34

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

The alpha metric 'etcd_object_counts' has been replaced by 'apiserver_storage_objects' and will be deprecated in 1.22. In 1.21 both metrics are available. We need to migrate addon-resizer to a new metric to make sure that it will work for a new metric.

Most likely we need to implement a logic where we try to use 'apiserver_storage_objects' with fallback to 'etcd_object_counts' to keep addon-resizer k8s-version agnostic. If we simply start using 'apiserver_storage_objects' this will work only in 1.21+ k8s versions

See https://github.com/kubernetes/kubernetes/issues/100621 for an example what happens if we do not migrate.

PR #4176, opened 2021-07-01
Fix addon resizer node metric

jbartosik:
Addon Resizer will use available metric for node count
    
Kubernetes 1.21 and newer have `apiserver_storage_objects` (1.20 and older don't have the metric).
Kubernetes 1.21 and older have `etcd_object_counts` (1.22 and newer won't have the metric).
    
To let addon resizer work with all K8s versions use whichever metric is available, if both are available then prefer the newer metric.
    
See also https://github.com/kubernetes/autoscaler/issues/3979

Fix a small bug where we were checking if unsigned int is less than 0.

### DIFF
--- a/addon-resizer/nanny/kubernetes_client.go
+++ b/addon-resizer/nanny/kubernetes_client.go
@@ -33,9 +33,11 @@ import (
 )
 
 const (
-	// objectCountMetricName is the metric to be used to get number of nodes
-	objectCountMetricName = "etcd_object_counts"
-	// nodeResourceName is the label value for Nodes in objectCountMetricName metric.
+	// objectCountMetricName is the preferred metric to be used to get number of nodes (present in Kubernetes 1.21 and higher)
+	objectCountMetricName = "apiserver_storage_objects"
+	// objectCountFallbackMetricName is the metric to be used to get number of nodes if objectCountMetricName metric is missing
+	objectCountFallbackMetricName = "etcd_object_counts"
+	// nodeResourceName is the label value for Nodes in objectCountFallbackMetricName and objectCountMetricName metrics.
 	nodeResourceName = "nodes"
 	// resourceLabel is the label name for resource.
 	resourceLabel = "resource"
@@ -83,17 +85,36 @@ func hasEqualValues(a string, b *string) bool {
 	return b != nil && a == *b
 }
 
-func (k *kubernetesClient) countNodesThroughMetrics() (uint64, error) {
-	// Similarly as for listing nodes, permissions for /metrics endpoint are needed.
-	// Other than that, endpoint is visible from everywhere.
-	reader, err := k.clientset.CoreV1().RESTClient().Get().RequestURI("/metrics").Stream(context.Background())
-	if err != nil {
-		return 0, err
+func extractMetricValueForNodeCount(mf dto.MetricFamily) (uint64, error) {
+	for _, metric := range mf.Metric {
+		hasLabel := false
+		for _, label := range metric.Label {
+			if hasEqualValues(resourceLabel, label.Name) && hasEqualValues(nodeResourceName, label.Value) {
+				hasLabel = true
+				break
+			}
+		}
+		if !hasLabel {
+			continue
+		}
+		if metric.Gauge == nil || metric.Gauge.Value == nil {
+			continue
+		}
+		if *metric.Gauge.Value < 0 {
+			return 0, fmt.Errorf("metric unknown")
+		}
+		value := uint64(*metric.Gauge.Value)
+		return value, nil
 	}
+	return 0, fmt.Errorf("no valid metric values")
+}
 
-	decoder := expfmt.NewDecoder(reader, expfmt.FmtText)
-
+func getNodeCountFromDecoder(decoder expfmt.Decoder) (uint64, error) {
 	var mf dto.MetricFamily
+	var fallbackNodeCountMetricValue uint64
+	var fallbackMetricError error
+	useNodeCountFromFallbackMetric := false
+
 	for {
 		if err := decoder.Decode(&mf); err != nil {
 			if err == io.EOF {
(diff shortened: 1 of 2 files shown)


jbartosik:
/hold

Tests pass and images build but when I tried to test an image in cluster I'm getting `standard_init_linux.go:211: exec user process caused "exec format error"` on the image.

jbartosik:
I was tired, made a mistake when editing config for deployment and used image for a different architecture, so it couldn't work.

I confirmed that Addon Resizer with changes from this PR works in a cluster running K8s ` 1.19.10-gke.1600`.

I created deployment from [this template](https://github.com/kubernetes/autoscaler/blob/a7ddc7e2c2410cd1f9e4fa428ed4b473918be489/addon-resizer/deploy/example.yaml) (modified to use image I built from repo with changes from this PR) in a cluster with 2 nodes.
The deployment started, the following requests:

```
          cpu: 340m
          memory: 220Mi
```

I resized cluster to 5 nodes, and requests of the deployment changed to

```
          cpu: 400m
          memory: 250Mi
```

jbartosik:
I confirmed that Addon Resizer with this changes worked on a cluster running K8s `1.21.1-gke.1800` (this played out exactly the same as in cluster running `1.19.10-gke.1600`)

jbartosik:
I have problems setting up a cluster with 1.22 so I added some more logs to see which metrics Addon Resizer is using.

Turns out it wasn't using metrics at all, I had to set `--use-metrics=true` flag and add permissions to read metrics. With changed configuration I verified that in cluster running K8s 1.21 Addon Resizer:

- Reads both preferred and fallback metrics and
- It works correctly (changed request from 360m CPU, 230 Mi memory to 400m CPU, 250Mi memory after I resized cluster from 3 to 5 nodes)

jbartosik:
I verified that in 1.19 cluster Addon Resizer configured to use metrics was getting only the fallback metric and was working (on resize from 3 to 5 nodes request changed  from 360m CPU, 230 Mi memory to 400m CPU, 250Mi memory).

jbartosik:
@bskiba @mwielgus please take a look. I looks like we will need to merge this one manually too

addon-resizer/nanny/kubernetes_client.go
```diff
@@ -83,54 +85,70 @@ func hasEqualValues(a string, b *string) bool {
...
 				break
 			}
 			return 0, fmt.Errorf("decoding error: %v", err)
 		}
-
-		if !hasEqualValues(objectCountMetricName, mf.Name) {
-			continue
+		if hasEqualValues(objectCountMetricName, mf.Name) {
+			// Preferred metric is present - return immediately
+			return extractMetricValueForNodeCount(mf)
```
bskiba, line 127: Should we fall back if extractMetricValueForNodeCount returns error?

addon-resizer/nanny/kubernetes_client.go
```diff
@@ -83,54 +85,70 @@ func hasEqualValues(a string, b *string) bool {
...
+			// Fallback metric is present - store values to return later (in case the preferred metric is missing)
+			fallbackNodeCountMetricValue, fallbackMetricError = extractMetricValueForNodeCount(mf)
+			useNodeCountFromFallbackMetric = true
 		}
 	}
 
+	if useNodeCountFromFallbackMetric {
+		return fallbackNodeCountMetricValue, fallbackMetricError
+	}
 	return 0, fmt.Errorf("metric unset")
```
bskiba, line 139: nit: can we explicitly say which metrics were unset?

addon-resizer/nanny/kubernetes_client_test.go
```diff
@@ -100,3 +103,185 @@ func TestMergeResources(t *testing.T) {
...
+				Value: &labelValue,
+			},
+		},
+		Gauge: &dto.Gauge{
+			Value: &value,
+		},
+	}
+}
+
+func TestExtractMetricValueForNodeCount(t *testing.T) {
```
bskiba, line 140: Should we test:

Multiple labels on a metric?
Multiple metrics in metricFamily?

bskiba (commented):
I think the comment in the second commit has names of matrics mixed up (i.e. etcd_object_counts is being deprecated and apiserver_storage_objects is the new one ahile the commit states the other way round.

addon-resizer/nanny/kubernetes_client.go
wojtek-t: +1

jbartosik: I thought it wouldn't be worth it (since we only expect both metrics in 1.21 and it makes code more complicated for all versions)

Done.

addon-resizer/nanny/kubernetes_client.go
jbartosik: Done.

addon-resizer/nanny/kubernetes_client_test.go
jbartosik: Done

jbartosik:
@bskiba @mwielgus please take a look

addon-resizer/nanny/kubernetes_client_test.go
```diff
@@ -273,6 +273,63 @@ func TestGetNodeCountFromDecoder(t *testing.T) {
 			finalResult: fmt.Errorf("oops"),
 			expectErr:   fmt.Errorf("decoding error: oops"),
 		},
+		{
+			name:        "fallbacks on error preferred on preferred metric",
```
bskiba, line 277: I can't parse this test case's name

bskiba (commented):
One nit. Also the commit description and PR description still have the metric names switched, right?

addon-resizer/nanny/kubernetes_client_test.go
jbartosik: Changed to "falls back on error in preferred metric" (the test name I tried to write)

jbartosik:
> I think the comment in the second commit has names of matrics mixed up (i.e. etcd_object_counts is being deprecated and apiserver_storage_objects is the new one ahile the commit states the other way round.

Done

jbartosik:
> One nit. Also the commit description and PR description still have the metric names switched, right?

Done

PR #4179, opened 2021-07-05
Metrics for AWS API calls

bpineau:
Provide metrics for AWS API calls; helps identifying slowness, throttling causes, and errors bursts.
That's similar to what [GCE provider does](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/gce/gce_metrics.go).

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_metrics.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_metrics.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package aws
+
+import (
+	"time"
+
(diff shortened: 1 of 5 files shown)


bpineau:
@gjtempleton may I request your input on that one?
/assign @gjtempleton

cluster-autoscaler/cloudprovider/aws/aws_metrics.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+var (
+	/**** Metrics related to AWS API usage ****/
+	requestSummary = k8smetrics.NewHistogramVec(
+		&k8smetrics.HistogramOpts{
+			Namespace: caNamespace,
+			Name:      "aws_request_duration_seconds",
+			Help:      "Time taken by AWS requests, by method and status code, in seconds",
+			Buckets:   []float64{0.05, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 2.0, 5.0, 10.0, 20.0, 30.0, 60.0},
```
gjtempleton, line 38: I'm very slightly worried about the number of histogram buckets here, though the low number of `endpoint` label values is making me less concerned.

Do we have any idea of the potential cardinality of the `status` label?

gjtempleton (commented):
Thanks for this, this looks really useful!

One small query for you.

cluster-autoscaler/cloudprovider/aws/aws_metrics.go
bpineau: Yes, I run (a slightly older version of) this patch for over one year all my clusters (managing several thousand ASGs):
saw those 10 distinct status over last 12 months:
image

gjtempleton: Great, thanks for that.

PR #4184, opened 2021-07-08
Skipping metrics tests added in #4022

MaciekPytel:
Each test works in isolation, but they cause panic when the entire
suite is run (ex. make test-in-docker), because the underlying
metrics library panics when the same metric is registered twice.


MaciekPytel:
@amrmahdi 

I'm disabling the tests added in #4022 for now. Do you have any idea on how to better isolate them? We could wrap the registration with an interface and swap it for test, but that feels like an overkill - looks like the changes required to test code would be more complex than the logic itself. Maybe unregistering metrics can be a workaround?

PR #4191, opened 2021-07-12
Move GetNodeInfosForGroups to it's own processor

bpineau:
Supports providing different NodeInfos sources (either upstream or in
local forks, eg. to properly implement variants like in #4000).

The already existing NodeInfosProcessor doesn't cover that use case: it's
actually a "post processor", and processes the provided nodeInfos set.

This also moves a large and specialized code chunk out of core, and removes
the need to maintain and pass the GetNodeInfosForGroups() cache from the side,
as processors can hold their states themselves.

No functional changes to GetNodeInfosForGroups(), outside mechanical changes
due to the move: remotely call a few utils functions in core/utils package,
pick context attributes (the processor takes the context as arg rather than
ListerRegistry + PredicateChecker + CloudProvider), and use the builtin cache
rather than receiving it from arguments.

### DIFF
--- a/cluster-autoscaler/core/utils/utils.go
+++ b/cluster-autoscaler/core/utils/utils.go
@@ -31,139 +31,10 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/utils/daemonset"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/gpu"
-	kube_util "k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/labels"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/taints"
 	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
-
-	klog "k8s.io/klog/v2"
 )
-
-// GetNodeInfosForGroups finds NodeInfos for all node groups used to manage the given nodes. It also returns a node group to sample node mapping.
-func GetNodeInfosForGroups(nodes []*apiv1.Node, nodeInfoCache map[string]*schedulerframework.NodeInfo, cloudProvider cloudprovider.CloudProvider, listers kube_util.ListerRegistry,
-	// TODO(mwielgus): This returns map keyed by url, while most code (including scheduler) uses node.Name for a key.
-	// TODO(mwielgus): Review error policy - sometimes we may continue with partial errors.
-	daemonsets []*appsv1.DaemonSet, predicateChecker simulator.PredicateChecker, ignoredTaints taints.TaintKeySet) (map[string]*schedulerframework.NodeInfo, errors.AutoscalerError) {
-	result := make(map[string]*schedulerframework.NodeInfo)
-	seenGroups := make(map[string]bool)
-
-	podsForNodes, err := getPodsForNodes(listers)
-	if err != nil {
-		return map[string]*schedulerframework.NodeInfo{}, err
-	}
-
-	// processNode returns information whether the nodeTemplate was generated and if there was an error.
-	processNode := func(node *apiv1.Node) (bool, string, errors.AutoscalerError) {
-		nodeGroup, err := cloudProvider.NodeGroupForNode(node)
-		if err != nil {
-			return false, "", errors.ToAutoscalerError(errors.CloudProviderError, err)
-		}
-		if nodeGroup == nil || reflect.ValueOf(nodeGroup).IsNil() {
-			return false, "", nil
-		}
-		id := nodeGroup.Id()
-		if _, found := result[id]; !found {
-			// Build nodeInfo.
-			nodeInfo, err := simulator.BuildNodeInfoForNode(node, podsForNodes)
-			if err != nil {
-				return false, "", err
-			}
-			sanitizedNodeInfo, err := sanitizeNodeInfo(nodeInfo, id, ignoredTaints)
-			if err != nil {
-				return false, "", err
-			}
-			result[id] = sanitizedNodeInfo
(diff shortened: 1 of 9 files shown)


MaciekPytel:
I feel like this approach is just better than the existing NodeInfoProcessor. Unless I'm missing something any existing NodeInfoProcessor could be reimplemented to a NodeInfoProvider by just calling the default implementation internally and applying whatever postprocessing logic it's already doing to the result. The opposite is not true as there are some important parameters to GetNodeInfosForGroups() that are not passed to NodeInfoProcessors. 

I don't particularly like having both though. They're called one after another with no logic in between them and the intended use-cases are pretty much identical. I wonder if there is any way we can make it less confusing? Ideally we'd just get rid of the old one, but we don't have any deprecation policy for processors (maybe it's time we get one?).
An alternative may be making those 2 methods on a single processor. I like that less, but it may still be less confusing compared to 2 different processors doing ~the same thing. While not technically backward compatible we have been adding new methods to cloudprovider interface and expecting any for to add an empty implementation, so there is some precedent here. 

Any thoughts on this @bpineau @towca?

bpineau:
> I don't particularly like having both though. They're called one after another with no logic in between them and the intended use-cases are pretty much identical. I wonder if there is any way we can make it less confusing? Ideally we'd just get rid of the old one, but we don't have any deprecation policy for processors (maybe it's time we get one?).

Ver true! I felt the same about it; didn't dare altering the existing interface because I was in doubt about this.
Also I confirm the NodeInfoProvider interface can cover both uses cases but not the current NodeInfo (after I attempted using it to that effect).

How would a deprecation look like? Having both for one or two release, with the old one annotated as deprecated? Or do we have more formal comm channels?

> An alternative may be making those 2 methods on a single processor. I like that less, but it may still be less confusing compared to 2 different processors doing ~the same thing. While not technically backward compatible we have been adding new methods to cloudprovider interface and expecting any for to add an empty implementation, so there is some precedent here.

Wouldn't the two methods roles/semantics (and the successive calls from core to that processor) look a bit weird for a long term change?

I suppose it would probably requires about he same effort for an external processor maintainer  to update to an unified processor vs adding a new (non empty in that case) method; except for the relatable precedent's interface changes scope in cloudproviders (but there I suppose many -if not all- are in-tree, while no alternate processors are).

MaciekPytel:
Since there is no opposition to replacing the processor, I propose the following:

 * Let's merge this as-is, including calling existing NodeInfoProcessor just after producing templates.
   * One minor suggestion: I'd consider renaming the processor. Current name seems to be chosen to be both similar to and distinct from the old processor which I think is the way to go if we have both. But since we want to get rid of the old one I'd consider renaming it to something that makes the role of the processor more clear. Maybe TemplateNodeInfoProvider? This doesn't really process anything and there is already a precedent for a processor-type interface that doesn't have "Processor" suffix in the name (NodeGroupManager). All that being said - I don't feel strongly about this and I'm ok to go with the current name if you think it's better.
 * Let's add a deprecation warning comment to NodeInfoProcessor (I can do this in follow-up PR) and log message warning about deprecation. 
   * The warning should include the version in which we intend to remove the existing processor. 
   * Since we don't have a deprecation policy I propose we follow the [Kubernetes policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/). I believe this falls under "Rule 5b" in said policy and as such requires '6 months or 1 release (whichever is longer)' before we remove the feature. Given that 1.22 was just cut I propose to remove the old processor from 1.24.

Since this is effectively setting a policy I think this should be discussed in sig meeting. I'll add it to agenda.

Does the above make sense to you @bpineau?

cluster-autoscaler/processors/nodeinfosprovider/node_info_provider_processor.go
```diff
@@ -0,0 +1,53 @@
...
+
+// NodeInfoProviderProcessor is provides the initial nodeInfos set.
+type NodeInfoProviderProcessor interface {
+	// Process returns a map of nodeInfos for node groups.
+	Process(nodes []*apiv1.Node, ctx *context.AutoscalingContext, daemonsets []*appsv1.DaemonSet, ignoredTaints taints.TaintKeySet) (map[string]*schedulerframework.NodeInfo, errors.AutoscalerError)
+	// CleanUp cleans up processor's internal structures.
+	CleanUp()
+}
+
+// NoOpNodeInfoProviderProcessor doesn't change nodeInfos.
```
towca, line 37: Is this NoOp implementation actually used anywhere? Do we need it?

towca (commented):
I'd definitely get rid of the existing NodeInfoProcessor if we're introducing this, the deprecation plan Maciek outlined SGTM. The change also looks good to me, I just have 2 questions.

cluster-autoscaler/core/utils/utils.go
```diff
@@ -173,8 +43,6 @@ func GetNodeInfoFromTemplate(nodeGroup cloudprovider.NodeGroup, daemonsets []*ap
 		return nil, errors.ToAutoscalerError(errors.CloudProviderError, err)
 	}
 
-	labels.UpdateDeprecatedLabels(baseNodeInfo.Node().ObjectMeta.Labels)
```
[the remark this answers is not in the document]
bpineau: This new chunk was missed on the prev rebase (I also just noticed and am restoring it, as I'm re-rebasing), sorry.

cluster-autoscaler/processors/nodeinfosprovider/node_info_provider_processor.go
bpineau: It's not, removed - thanks

bpineau:
@MaciekPytel I also prefer your suggested name, updated accordingly. I didn't place deprecation warnings (taking on your offer to update it in a follow-up pr, but can do if you prefer).

PR #4199, opened 2021-07-15
Optimise generate ec2

aidy:
The pricing json for us-east-1 is currently 129MB. Currently fetching
this into memory and parsing results in a large memory footprint on
startup, and can lead to the autoscaler being OOMKilled.

Change the ReadAll/Unmarshal logic to a stream decoder to significantly
reduce the memory use.

Should hopefully address #3506

PR #4207, opened 2021-07-19
[Examples] Update the AWS examples to have 600Mi of memory

MyannaHarris:
This change updates the AWS examples to have 600Mi of memory because CAS downloads a pricing file that contains EC2 instance info at startup which grows each time there's new EC2 instance information available. Currently the largest region is hitting the 300Mi limit when downloading that file, so we are increasing the memory limit in our examples for customers.

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
+++ b/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
@@ -143,10 +143,10 @@ spec:
           resources:
             limits:
               cpu: 100m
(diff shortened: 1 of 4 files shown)


gjtempleton:
Thanks for the PR, it won't solve the problems for everyone as we can't cherry-pick it back everywhere, but have you had a look at #4199 for an approach to optimising how we do the reading in as well?

I guess we can revisit these examples once every supported version is using a more efficient method so can merge this as-is for now, as sad as it makes me that the CA needs this much memory.

/lgtm
/approve

MyannaHarris:
Thank you! Yes, hopefully we can optimize this more going forward.

PR #4211, opened 2021-07-21
Subtract toBeDeleted nodes from number of upcoming nodes; cleanup toBeDeleted taints from all nodes, not only ready ones

### LINKED ISSUE #3949: Node groups get "stuck" during node deletion

**What version of the component are you using?**: 1.17.4

Component version: 1.17.4

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.4", GitCommit:"e87da0bd6e03ec3fea7933c4b5263d151aafd07c", GitTreeState:"clean", BuildDate:"2021-02-20T02:22:41Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17+", GitVersion:"v1.17.12-eks-7684af", GitCommit:"7684af4ac41370dd109ac13817023cb8063e3d45", GitTreeState:"clean", BuildDate:"2020-10-20T22:57:40Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What environment is this in?**: AWS EKS with self-managed node groups

**What happened?**:

The leading cluster-autoscaler pod was apparently disrupted in the middle of draining and deleting a node. The affected node had both the `ToBeDeletedByClusterAutoscaler` and `DeletionCandidateOfClusterAutoscaler` taints but there were still pods running and it was not deleted from EC2. Afterwards, the node group that this node is part of was apparently kind of "stuck" as cluster-autoscaler didn't scale it up anymore although there was a pending pod that would have been able to run on it. According to the logs, cluster-autoscaler believed that there was a node coming up to run this pod, but this was not the case, the Auto Scaling Group was not scaled up, all nodes in the ASG were registered and ready in Kubernetes:

```
I0317 10:43:37.632876       1 static_autoscaler.go:194] Starting main loop
I0317 10:43:38.604818       1 filter_out_schedulable.go:66] Filtering out schedulables
I0317 10:43:38.616421       1 filter_out_schedulable.go:125] Pod my-pod-that-is-pending marked as unschedulable can be scheduled on upcoming node template-node-for-my-autoscalinggroup-name-1923849245781316735-0. Ignoring in scale up.
I0317 10:43:38.616448       1 filter_out_schedulable.go:131] 1 other pods marked as unschedulable can be scheduled.
I0317 10:43:38.622681       1 filter_out_schedulable.go:131] 0 other pods marked as unschedulable can be scheduled.
I0317 10:43:38.622713       1 filter_out_schedulable.go:88] Schedulable pods present
I0317 10:43:38.622731       1 static_autoscaler.go:343] No unschedulable pods
I0317 10:43:38.622762       1 static_autoscaler.go:390] Calculating unneeded nodes
```

The `cluster-autoscaler-status` config map in `kube-system` showed the following for the affected node group:
```
Health:      Healthy (ready=18 unready=0 notStarted=0 longNotStarted=0 registered=19 longUnregistered=0 cloudProviderTarget=19 (minSize=0, maxSize=50))
```

So it didn't consider the affected node as `ready` and (I assume) thought that there was a node coming up because `cloudProviderTarget - ready > 0`. At the same time, it didn't delete the node because there were still pods running on it.

Manually removing the `ToBeDeletedByClusterAutoscaler` taint from the affected node resolved the situation, causing the node group to be scaled up as expected.

alfredkrohmer:
Fixes #3949

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -917,7 +917,7 @@ func (csr *ClusterStateRegistry) GetUpcomingNodes() map[string]int {
 		readiness := csr.perNodeGroupReadiness[id]
 		ar := csr.acceptableRanges[id]
 		// newNodes is the number of nodes that
-		newNodes := ar.CurrentTarget - (readiness.Ready + readiness.Unready + readiness.LongUnregistered)
+		newNodes := ar.CurrentTarget - (readiness.Ready + readiness.Unready + readiness.LongUnregistered + readiness.Deleted)
 		if newNodes <= 0 {
 			// Negative value is unlikely but theoretically possible.
 			continue
--- a/cluster-autoscaler/clusterstate/clusterstate_test.go
+++ b/cluster-autoscaler/clusterstate/clusterstate_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package clusterstate
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
@@ -27,6 +28,7 @@ import (
 	testprovider "k8s.io/autoscaler/cluster-autoscaler/cloudprovider/test"
 	"k8s.io/autoscaler/cluster-autoscaler/clusterstate/api"
 	"k8s.io/autoscaler/cluster-autoscaler/clusterstate/utils"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/deletetaint"
 	. "k8s.io/autoscaler/cluster-autoscaler/utils/test"
 	"k8s.io/client-go/kubernetes/fake"
 	kube_record "k8s.io/client-go/tools/record"
@@ -481,6 +483,22 @@ func TestUpcomingNodes(t *testing.T) {
 	provider.AddNodeGroup("ng4", 1, 10, 1)
 	provider.AddNode("ng4", ng4_1)
 
+	// One node is already there, for a second node deletion / draining was already started.
+	ng5_1 := BuildTestNode("ng5-1", 1000, 1000)
+	SetNodeReadyState(ng5_1, true, now.Add(-time.Minute))
+	ng5_2 := BuildTestNode("ng5-2", 1000, 1000)
+	SetNodeReadyState(ng5_2, true, now.Add(-time.Minute))
+	ng5_2.Spec.Taints = []apiv1.Taint{
+		{
+			Key:    deletetaint.ToBeDeletedTaint,
+			Value:  fmt.Sprint(time.Now().Unix()),
+			Effect: apiv1.TaintEffectNoSchedule,
+		},
+	}
+	provider.AddNodeGroup("ng5", 1, 10, 2)
+	provider.AddNode("ng5", ng5_1)
+	provider.AddNode("ng5", ng5_2)
+
 	assert.NotNil(t, provider)
 	fakeClient := &fake.Clientset{}
 	fakeLogRecorder, _ := utils.NewStatusMapRecorder(fakeClient, "kube-system", kube_record.NewFakeRecorder(5), false, "my-cool-configmap")
@@ -488,7 +506,7 @@ func TestUpcomingNodes(t *testing.T) {
 		MaxTotalUnreadyPercentage: 10,
 		OkTotalUnreadyCount:       1,
 	}, fakeLogRecorder, newBackoff())
-	err := clusterstate.UpdateNodes([]*apiv1.Node{ng1_1, ng2_1, ng3_1, ng4_1}, nil, now)
+	err := clusterstate.UpdateNodes([]*apiv1.Node{ng1_1, ng2_1, ng3_1, ng4_1, ng5_1, ng5_2}, nil, now)
 	assert.NoError(t, err)
 	assert.Empty(t, clusterstate.GetScaleUpFailures())
 
@@ -497,6 +515,7 @@ func TestUpcomingNodes(t *testing.T) {
 	assert.Equal(t, 1, upcomingNodes["ng2"])
 	assert.Equal(t, 2, upcomingNodes["ng3"])
 	assert.NotContains(t, upcomingNodes, "ng4")
+	assert.NotContains(t, upcomingNodes, "ng5")
 }
 
 func TestIncorrectSize(t *testing.T) {
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -180,14 +180,14 @@ func (a *StaticAutoscaler) cleanUpIfRequired() {
 	}
 
 	// CA can die at any time. Removing taints that might have been left from the previous run.
-	if readyNodes, err := a.ReadyNodeLister().List(); err != nil {
-		klog.Errorf("Failed to list ready nodes, not cleaning up taints: %v", err)
+	if nodes, err := a.AllNodeLister().List(); err != nil {
+		klog.Errorf("Failed to list nodes, not cleaning up taints: %v", err)
 	} else {
-		deletetaint.CleanAllToBeDeleted(readyNodes,
+		deletetaint.CleanAllToBeDeleted(nodes,
 			a.AutoscalingContext.ClientSet, a.Recorder, a.CordonNodeBeforeTerminate)
 		if a.AutoscalingContext.AutoscalingOptions.MaxBulkSoftTaintCount == 0 {
 			// Clean old taints if soft taints handling is disabled
-			deletetaint.CleanAllDeletionCandidates(readyNodes,
+			deletetaint.CleanAllDeletionCandidates(nodes,
 				a.AutoscalingContext.ClientSet, a.Recorder)
 		}
 	}


x13n (changes requested):
Can you add some unit tests that would fail without this change?

alfredkrohmer:
@x13n Added a unit test in `TestUpcomingNodes`. Without the fix, the test is failing with:
```
--- FAIL: TestUpcomingNodes (0.00s)
    clusterstate_test.go:518: 
                Error Trace:    clusterstate_test.go:518
                Error:          "map[ng1:%!s(int=6) ng2:%!s(int=1) ng3:%!s(int=2) ng5:%!s(int=1)]" should not contain "ng5"
                Test:           TestUpcomingNodes
```

x13n:
I read the original issue and while I understand the need to clean up taints on all nodes instead of just ready ones, I don't understand why we're reducing the number of nodes that should come up by subtracting the nodes that are being deleted.

Suppose:
1. there are 10 nodes in the node group
2. one of them starts being scaled down, target size is now 9
3. before all pods are successfully evicted, new unschedulable pods appear in the cluster, increasing the target size back to 10
4. new node won't get scheduled now, because we're going to subtract the deleted node from the target size
5. pods will stay pending until the old node is deleted - only after that happens, the new node will get provisioned

alfredkrohmer:
> I don't understand why we're reducing the number of nodes that should come up by subtracting the nodes that are being deleted

That's not the case here, this is not being subtracted from the nodes that _should_ come up, it's subtracted from the number of nodes that are _supposedly coming up right now_ (that cluster-autoscaler _thinks_ are coming up right now).

alfredkrohmer:
> The number of nodes that are supposedly coming up right now will later impact the number of nodes that should come up

Exactly

> CA will think that there's no need to create more nodes, since they are coming up already.

My PR is _reducing_ the number of nodes that CA _thinks_ is coming up by subtracting the number of nodes being deleted - otherwise CA thinks these nodes that are being deleted are actually coming up right now which is not true. This actually _increases_ (not decreases) the number of nodes that _should_ come up.

x13n:
Hm, you're right, this will increase, not decrease the number of upcoming nodes. That brings about another concern though: if some nodes are tainted for deletion, but don't actually end up deleted, we may exceed the max size. Nodes that are being deleted can end up staying around e.g. if CA is restarted in the meantime.

alfredkrohmer:
> Nodes that are being deleted can end up staying around e.g. if CA is restarted in the meantime.

That's the other part of the PR where deletion taints are _properly_ removed on CA startup.

x13n:
> That's the other part of the PR where deletion taints are _properly_ removed on CA startup.

Proper removal of deletion taints is not my concern here. We may still exceed max # of nodes in the following scenario:

1. Node X is tainted for deletion
2. Another node Y starts getting created as a part of scale up
3. CA crashes before X is actually deleted
4. New CA process clears deletion taint from node X

Now we have both X and Y up and running even though they might be exceeding the limit.

alfredkrohmer:
I don't see how this behavior could be triggered by my PR. My PR doesn't touch the scale-up logic; again, it _only_ changes the number of nodes CA _thinks_ are coming up right now, I don't see how this would cause exceeding the max number of nodes in a NG.

alfredkrohmer:
@x13n I think I found what you mean. Here:

https://github.com/kubernetes/autoscaler/blob/86068ba9a92b8ed3099cad96c49743e68549ff86/cluster-autoscaler/core/scale_up.go#L476-L478
the number of upcoming nodes is considered to check if the maximum number of nodes in the whole cluster would be exceeded. If you want, I can add/subtract the deleted nodes here as well, then it's the same behavior as before. However, I'm not sure what the best course of action would actually be here (allow to exceed the max while nodes are being deleted vs risk that scale up is not possible when max would be exceeded).

As for the per-node group max number of nodes, this shouldn't be affected by my change, as this check considers only the actual number of nodes in the node group (independent of the state of the node in k8s):
https://github.com/kubernetes/autoscaler/blob/86068ba9a92b8ed3099cad96c49743e68549ff86/cluster-autoscaler/processors/nodegroupset/balancing_processor.go#L87-L113

x13n:
Apologies for late response. Yes, I meant the logic in https://github.com/kubernetes/autoscaler/blob/86068ba9a92b8ed3099cad96c49743e68549ff86/cluster-autoscaler/core/scale_up.go#L476-L478

I think deleted nodes should also be considered here. In edge cases when deletion takes arbitrarily long time, we could easily exceed the maximum if deleted nodes are not accounted for in the scale up logic.

alfredkrohmer:
I tried to somehow get the number of deleted nodes at this point in the code to add / subtract them and so I went through the call graph up to where `ScaleUp` is called and I suspect I found more errors in the surrounding implementation.

One thing upfront that I found: `len(nodes)` ([here](https://github.com/kubernetes/autoscaler/blob/86068ba9a92b8ed3099cad96c49743e68549ff86/cluster-autoscaler/core/scale_up.go#L476-L478) in the `ScaleUp` function) _does_ contain the deleted nodes - but only if they are not cordoned; hence:
* for not-cordoned "deleted" nodes: the current calculation is actually broken, it's off by `n` (number of deleted nodes) in the other direction, i.e. already preventing scale up at `max-n` because they would be contained twice in the sume, in `len(nodes)` and in `len(upcomdingNodes)`, my change would fix this.
* for cordoned "deleted" nodes: the current calculation is correct and my change would break it by allowing to scale up to `max+n`; see the following points for a proposed solution

In the `RunOnce` function here: https://github.com/kubernetes/autoscaler/blob/86068ba9a92b8ed3099cad96c49743e68549ff86/cluster-autoscaler/core/static_autoscaler.go#L235
`allNodes` and `readyNodes` are retrieved (`allNodes` includes unready and cordoned nodes, `readyNodes` doesn't)

In the same function, `readyNodes` is used three times in a relevant way.

First time to derive node group infos from the ready nodes: https://github.com/kubernetes/autoscaler/blob/86068ba9a92b8ed3099cad96c49743e68549ff86/cluster-autoscaler/core/static_autoscaler.go#L283

Second time for another check that the max number of nodes in the cluster is not yet reached: https://github.com/kubernetes/autoscaler/blob/86068ba9a92b8ed3099cad96c49743e68549ff86/cluster-autoscaler/core/static_autoscaler.go#L420

Third time to finally call the `ScaleUp` function: https://github.com/kubernetes/autoscaler/blob/86068ba9a92b8ed3099cad96c49743e68549ff86/cluster-autoscaler/core/static_autoscaler.go#L435

I believe it would be more correct to pass `allNodes` instead in these three uses of `readyNodes`.
1. In the first use, I think it makes sense to build node group info from unready nodes (e.g. if it's the only one from this node group).
2. In the second use, as per your argument, cordoned and unready nodes should count into the max number of nodes.
3. In the third use (the actual problem I was trying to solve), this would make `len(nodes)` ([here](https://github.com/kubernetes/autoscaler/blob/86068ba9a92b8ed3099cad96c49743e68549ff86/cluster-autoscaler/core/scale_up.go#L476-L478)) contain the cordoned nodes and hence they would also be considered correctly for the max number of nodes in the cluster - so this would also make my previous change correct for cordoned "deleted" nodes.

Any thoughts?

x13n:
Apologies again for super slow response. I had to take some time to go through the existing code. I think the proposal makes sense, I'm just a bit worried I didn't think about some edge case though - historically, clusterstate changes were a source of "interesting" bugs - and would prefer to keep this change behind a feature flag because of this. Can be on by default and can be removed in a few versions if there are no issues, but it might be safer to have a quick rollback mechanism. WDYT?

x13n:
/remove-lifecycle stale

Maybe a feature flag is an overkill, but would appreciate another pair of eyes on this. @MaciekPytel - can you take a look?

MaciekPytel:
Sorry for missing this for so long. I understand the problem being solved. I'm worried that this PR may fix it, but at the same time introduce an 'opposite' issue.

1. Today we count a node that is intended to be deleted as "Deleted". Because Deleted is mutually exclusive with Ready/Unready and because of how upcoming nodes calculation work, CA overestimates the number of upcoming nodes and doesn't trigger scale-up.
2. It can take time for node controller to remove kubernetes node object after the underlying VM has already been deleted. During this time the node also shows up in clusterstarte as Deleted. If we subtracted Deleted from TargetSize as proposed in this PR we would underestimate the number of upcoming nodes (we're subtracting a node that no longer counts towards TargetSize, as the VM no longer exists on cloudprovider side). Underestimating the number of upcoming nodes would most likely lead to too much scale-up.

The [description of Deleted in Readiness struct](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/clusterstate/clusterstate.go#L529) makes it pretty clear to me that the intent of 'Deleted' node in Readiness is the case 2 above. And I think the current implementation addresses this problem correctly. If Deleted nodes are defined as the nodes that no longer exist on provider side, than we shouldn't subtract them from TargetSize.

I think the bug is that we're incorrectly identifying nodes as Deleted and the way to fix both problems 1 and 2 would be to fix that. Unfortunately that would probably require a slightly bigger code change: I think we should check if a particular instance exists on cloudprovider side as part of readiness calculation. This is very similar to how we identify unregistered nodes, just doing the check in other direction (check if there is a matching instance for a given node). 

WDYT @alfredkrohmer @x13n ?

fookenc:
Hi @MaciekPytel, thank you for taking a look!  I wouldn't want to introduce new issues into the autoscaler, but I'm concerned that the readiness calculation isn't aware of nodes being drained & deleted.

To be exact, the readiness calculation doesn't consider taints that may exist on the nodes.  If a node has the ToBeDeleted taint, it has an effect of NoSchedule.  Pods wouldn't be capable of being scheduled against the node, but the autoscaler believes that enough resources exist to handle the pending pods.  Unfortunately this blocks scale ups indefinitely while the node is being drained and deleted.  This can be observed in the logs obtained in the bug #4456.

>I1105 11:07:58.593642 1 filter_out_schedulable.go:157] Pod default.app-1-cf767bfc7-m5672 marked as unschedulable can be scheduled on node template-node-for-node-group-2-az2-6211936831103862438-0. Ignoring in scale up.
I1105 11:07:58.593798 1 filter_out_schedulable.go:157] Pod default.app-1-cf767bfc7-jtcc4 marked as unschedulable can be scheduled on node template-node-for-node-group-2-az2-6211936831103862438-0. Ignoring in scale up.
I1105 11:07:58.593818 1 filter_out_schedulable.go:170] 0 pods were kept as unschedulable based on caching
I1105 11:07:58.593825 1 filter_out_schedulable.go:171] 2 pods marked as unschedulable can be scheduled.

>CA seem to be thinking that the above node can be used to schedule the un-schedulable pod but it is actually being deleted due to under-utilisation. Due to a max termination grace period of 3 hours (and cause the pods already running on the node failed to grace-fully terminate before that), CA had to wait 3 hours for all pods to be removed from the affected node. During this interval, there was no scale up triggered by CA even though there were pods that needed to be scheduled.

I have found that scaling up the pods enough to exceed the capacity/resources available of the tainted node will result in a new node being added.  Given that the Deleted field is the count of nodes with the ToBeDeleted taint, I assumed that this value could only be zero or 1 in typical scenarios.

One alternative that I've been exploring is to have a separate controller monitor for the ToBeDeleted taint, and triggering an over provision a new node with Pause pods to prevent scale ups from being blocked.  This does move the burden of overprovisioning outside of the autoscaler instead of having the readiness check adjusted to remove upcoming nodes that shouldn't be considered viable.

Please let me know your thoughts!  Thanks!

MaciekPytel:
>  Given that the Deleted field is the count of nodes with the ToBeDeleted taint

I think this is the problem. It's not intended to be the number of nodes with ToBeDeleted taint. It's intended to be a number of nodes that don't have a matching VM on cloudprovider side, because the VM has already been deleted. 

My guess is that looking at a taint seemed like an easy proxy for nodes that have already been deleted by CA [^1]. As we can see, it isn't. I think the right way to fix this issue without introducing new problems is to stop identifying nodes that are _being_ deleted as _already_ deleted. This should fix your scenario, as the node that is being deleted would not be counted as Deleted, meaning that it would be counted as either Ready or Unready, both of which are subtracted from upcoming nodes.

It also means that nodes that have actually been deleted (in the sense that VM no longer exists and it's just a kubernetes object that haven't been cleaned up yet) would still not be included in upcoming nodes calculation. This part is correct - those nodes no longer count towards NodeGroup.TargetSize() and so should not be subtracted from this number.

We already get list of cloudprovider instances in [UpdateNodes()](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/clusterstate/clusterstate.go#L303) before the [updateReadinessStats() call](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/clusterstate/clusterstate.go#L318). I think we could use this data in updateReadinessStats() to only count node as Deleted if there is no corresponding instance.

The potentially tricky part would be understanding if there is any negative interaction with node_instances_cache (do we need to call InvalidateNodeInstancesCacheEntry() on scale-down? I'm guessing yes, but I haven't really dived into the code deeply enough to know).

[^1]: This makes more sense if you consider when this logic was implemented. Back than node deletion was synchronous (ie. CA would completely block for the entire process), so from CA perspective tainting and deleting the node was effectively atomic.

fookenc:
I made a commit based on the suggestions provided, but not certain that I got it right.  Within the updateReadinessStats, I switched to using the Cluster State Registry of the latest nodes unregistered in the cloud provider.

https://github.com/fookenc/autoscaler/blob/049e8ad0fc79bf2d6b9ef18b933efbb3d7d8415f/cluster-autoscaler/clusterstate/clusterstate.go#L541-L548

I also made a change to the GetReadinessState function to include the ToBeDeleted taint key to mark any node as Unready.  I don't know if this is necessary, but felt that it made sense to mark it that way.

https://github.com/fookenc/autoscaler/blob/049e8ad0fc79bf2d6b9ef18b933efbb3d7d8415f/cluster-autoscaler/utils/kubernetes/ready.go#L72-L77

After building and deploying the changes from this commit[^1] to a test cluster, the scale ups were no longer being blocked when a node was tainted with ToBeDeleted.  However, I'm not sure that these code changes exactly match to what was described.  Additionally, I didn't make any attempts to address the scale down that may need to call InvalidateNodeInstancesCacheEntry.

[^1]:https://github.com/fookenc/autoscaler/commit/049e8ad0fc79bf2d6b9ef18b933efbb3d7d8415f#

fookenc:
Hi @MaciekPytel & @x13n,

I've created a new pull request that incorporates some of the changes (tests for UpcomingNodes, clearing taints from ALL nodes on startup) and new changes to identify those nodes that no longer have a corresponding VM on the cloud provider.  Please let me know your feedback on the proposed changes.  If possible, I'd like to have a discussion regarding the InvalidateNodeInstancesCacheEntry during scale down.

Thanks!

x13n:
This has been obsoleted by https://github.com/kubernetes/autoscaler/pull/5054 now.

/close

alfredkrohmer:
@x13n Correct me if I'm wrong, but I belive #5054 doesn't actually address the main problem described in #3949. Our problem is not nodes deleted in the cloud provider and still having a k8s node, but rather that nodes that are being drained by cluster-autoscaler for binpacking are counted as "upcoming" nodes. Or is the PR that you linked fixing the problem by rewiring a bunch of logic so that it will be fixed "on the way, too"?

PR #4216, opened 2021-07-22
Update go version to 1.16 in Dockerfile and Github Tests

jayantjain93:
Updating the builder in Dockerfile to use go version 1.16, since vendoring to k8s v1.23.0-alpha.0 uses go1.16 packages which will fail otherwise.

GO111MODULE has to turned to `auto` due to the default being changed to `on` in go1.16:
https://golang.org/doc/go1.16#go-command
https://blog.golang.org/go116-module-changes

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,7 +14,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.15
+          go-version: 1.16
 
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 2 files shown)


hack/for-go-proj.sh
```diff
@@ -43,9 +43,6 @@ esac
 
 for project_name in ${PROJECT_NAMES[*]}; do
   (
-    if [[ $project_name == cluster-autoscaler ]];then
```
towca, line 46: The previous value used was "auto" for every project but cluster-autoscaler, and "off" for cluster-autoscaler, right? And now we're setting it to "off" for every project but vertical-pod-autoscaler/e2e, for which we're setting it to "on". So the delta is:
cluster-autoscaler: off -> off
vertical-pod-autoscaler/e2e: auto -> on
everything else: auto -> off

If we want to just make it work like before, shouldn't we be aiming for no delta? Also, if that's the goal (as opposed to updating the scripts), won't setting GO111MODULE=auto for all of the scripts be the easiest option? From the go docs you linked:
>More precisely, the GO111MODULE environment variable now defaults to on. To switch to the previous behavior, set GO111MODULE to auto.

jayantjain93: Great catch.
Changes this replicate earlier behaviour of using `auto`

PR #4225, opened 2021-07-26
GCE: CreateInstances() should use BaseInstanceName

bpineau:
The new `CreateInstances()` upscale method replacing `Resize()` API
calls generates new instances names based on the MIG's name (from
`mig.GceRef()`).

Before that change, `Resize()`-initiated upscales were prompting MIGs to
spawn instances named after MIG's `BaseInstanceName` attribute.

Accordingly, `GetMigForInstance()` still uses MIG's `BaseInstanceName`
to map instances to their parent MIG and discover which MIGs needs an
immediate refresh.

Down the line the `clusterstate.updateReadinessStats()` periodic
goroutines won't be able to map new ready nodes to their parent MIGs
(until the cache is backfilled upward from k8s node's providerid, ie.
from an hourly goroutine), and those MIGs will be considered non-ready
(because MIG's size>0 while the MIG has no known ready instances).

So after a first upscale, MIGs (having a BaseInstanceName that is not
the MIG's Name) won't be re-upscalable for a while. Example symptoms:

```
cluster-autoscaler W0719 12:35:43.166563 6 clusterstate.go:447] Failed to find readiness information for https://www.googleapis.com/compute/v1/projects/REDACTED-PROJECT/zones/europe-west3-b/instanceGroups/REDACTED-MIGNAME
cluster-autoscaler W0719 12:35:43.193469 6 clusterstate.go:626] Readiness for node group https://www.googleapis.com/compute/v1/projects/REDACTED-PROJECT/zones/europe-west3-b/instanceGroups/REDACTED-MIGNAME not found
```

Beside mapping cache issue, this changed the instance names prefixes for
some users, while it might make sense to keep using basenames when
explicitely provided (might have an use for eg. identification, or name
length limits) and avoid a breaking change before `CreateInstances` hits
a release.

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
+++ b/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
@@ -40,7 +40,6 @@ const (
 	defaultOperationWaitTimeout          = 20 * time.Second
 	defaultOperationPollInterval         = 100 * time.Millisecond
 	defaultOperationDeletionPollInterval = 1 * time.Second
-	instanceGroupNameSuffix              = "-grp"
 	// ErrorCodeQuotaExceeded is an error code used in InstanceErrorInfo if quota exceeded error occurs.
 	ErrorCodeQuotaExceeded = "QUOTA_EXCEEDED"
 
@@ -77,7 +76,7 @@ type AutoscalingGceClient interface {
 	// modifying resources
 	ResizeMig(GceRef, int64) error
 	DeleteInstances(migRef GceRef, instances []GceRef) error
-	CreateInstances(GceRef, int64, []string) error
+	CreateInstances(GceRef, string, int64, []string) error
 }
 
 type autoscalingGceClientV1 struct {
@@ -198,7 +197,7 @@ func (client *autoscalingGceClientV1) ResizeMig(migRef GceRef, size int64) error
 	return client.waitForOp(op, migRef.Project, migRef.Zone, false)
 }
 
-func (client *autoscalingGceClientV1) CreateInstances(migRef GceRef, delta int64, existingInstances []string) error {
+func (client *autoscalingGceClientV1) CreateInstances(migRef GceRef, baseName string, delta int64, existingInstances []string) error {
 	registerRequest("instance_group_managers", "create_instances")
 	req := gce.InstanceGroupManagersCreateInstancesRequest{}
 	instanceNames := map[string]bool{}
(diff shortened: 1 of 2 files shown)


olagacek:
Can you update the tests in `gce_cloud_provider_test.go`  and `gce_manager_test.go`? Thanks!

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -369,15 +369,15 @@ func isInstanceNotRunningYet(gceInstance *gce.ManagedInstance) bool {
 	return gceInstance.InstanceStatus == "" || gceInstance.InstanceStatus == "PROVISIONING" || gceInstance.InstanceStatus == "STAGING"
 }
 
-func generateInstanceName(migRef GceRef, existingNames map[string]bool) string {
+func generateInstanceName(baseName string, existingNames map[string]bool) string {
 	for i := 0; i < 100; i++ {
-		name := fmt.Sprintf("%v-%v", strings.TrimSuffix(migRef.Name, instanceGroupNameSuffix), rand.String(4))
+		name := fmt.Sprintf("%v-%v", strings.TrimSuffix(baseName, instanceGroupNameSuffix), rand.String(4))
```
olagacek, line 374: I guess we no longer need to trim suffix? I believe it could event produce incorrect behaviour if base name contains `instanceGroupNameSuffix`

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -303,8 +303,15 @@ func (m *gceManagerImpl) CreateInstances(mig Mig, delta int64) error {
 	for _, ins := range instances {
 		instancesNames = append(instancesNames, ins.Id)
 	}
+	baseName, found := m.cache.GetMigBasename(mig.GceRef())
+	if !found {
+		baseName, err = m.GceService.FetchMigBasename(mig.GceRef())
+		if err != nil {
+			return fmt.Errorf("can't upscale %s: failed to collect BaseInstanceName: %w", mig.GceRef(), err)
+		}
```
MaciekPytel, line 311: Maybe consider m.cache.SetMigBasename() here? I don't think it will change all that much, but I see no obvious reason not to update the cache if we're fetching the value anyway.

MaciekPytel:
Good catch, thanks for spotting this. I left a single minor comment, but overall lgtm.

/approve

bpineau:
@MaciekPytel thanks for the suggestion, filling the cache makes sense indeed; updated accordingly.

@olagacek CreateInstances() don't set or return anything related to instances names; do you have something specific in mind with regards to tests update? (we might be able to rework the http mock framework to catch and validate API calls POST arguments, but it seems that'd require some surgery, is it worth the added complexity? or do I just entirely miss the point?)

olagacek:
I don't think we need to update the tests, you may disregard this comment :)

PR #4208, opened 2021-07-19
CA-Chart - Update link to readme in notes.txt

gjtempleton:
Updates link provided to users on unsuccessful install to new home of chart instead of old helm/charts repository.

PR #4226, opened 2021-07-26
[chart] Add multi string arg support

### LINKED ISSUE #3673: Helm chart limits setting balancing-ignore-label flag to a single label
The snippet in the chart to set extra arguments is in the format of key=value pairs.  For arguments that can be passed multiple times, this limits you to only setting one value.

```
          {{- range $key, $value := .Values.extraArgs }}
            - --{{ $key }}={{ $value }}
          {{- end }}
```

Switching it from a hash to an array would solve the problem, but would break backwards compatibility and complexity to the existing configuration.  I think accepting a separate block of multi use arguments would keep the configuration simple.

```
extraMultiUseArgs:
  balancing-ignore-label:
    - some-label
    - and-another
```

stevehipwell:
Enable `extraArgs` to contain multi-string arguments via an ignored `_x` suffix.

Fixes #3673.

### DIFF
--- a/charts/cluster-autoscaler/values.yaml
+++ b/charts/cluster-autoscaler/values.yaml
@@ -128,6 +128,7 @@ expanderPriorities: {}
 # extraArgs -- Additional container arguments.
 # Refer to https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca for the full list of cluster autoscaler
 # parameters and their default values.
+# Everything after the first _ will be ignored allowing the use of multi-string arguments.
 extraArgs:
   logtostderr: true
   stderrthreshold: info
@@ -149,6 +150,8 @@ extraArgs:
   # scale-down-delay-after-failure: 3m
   # scale-down-unneeded-time: 10m
   # skip-nodes-with-system-pods: true
+  # balancing-ignore-label_1: first-label-to-ignore
+  # balancing-ignore-label_2: second-label-to-ignore
 
 # extraEnv -- Additional container environment variables.
 extraEnv: {}
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -17,4 +17,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.10.2
+version: 9.10.3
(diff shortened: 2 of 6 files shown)


stevehipwell:
I'm not sure why the Helm Docs action failed as I haven't touched the chart README?

gjtempleton:
It'll be what I mentioned here, unfortunately: https://github.com/kubernetes/autoscaler/issues/3673#issuecomment-886524517 (I'd planned to address it in #4208, but that's not been merged yet.)

stevehipwell:
@gjtempleton I've contributed before (I think) and the other actions ran fine. I've grabbed your changes from PR https://github.com/kubernetes/autoscaler/pull/4208 but the Helm Docs action is still failing?

gjtempleton:
You'll need to re-run `helm-docs` against the chart, your change to the comment block on the `extraArgs` will be pulled into the README by it. What I get when I run it:

```
-| extraArgs | object | `{"logtostderr":true,"stderrthreshold":"info","v":4}` | Additional container arguments. Refer to https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca for the full list of cluster autoscaler parameters and their default values. |
+| extraArgs | object | `{"logtostderr":true,"stderrthreshold":"info","v":4}` | Additional container arguments. Refer to https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca for the full list of cluster autoscaler parameters and their default values. everything after the first _ will be ignored allowing the use of multi-string arguments |
```

charts/cluster-autoscaler/values.yaml
```diff
@@ -128,6 +128,7 @@ expanderPriorities: {}
 # extraArgs -- Additional container arguments.
 # Refer to https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca for the full list of cluster autoscaler
 # parameters and their default values.
+# everything after the first _ will be ignored allowing the use of multi-string arguments
```
gjtempleton, line 131:
```suggestion
# Everything after the first _ will be ignored allowing the use of multi-string arguments
```

stevehipwell:
Thanks @gjtempleton, I'd just figured it out myself although adding `--exit-code` to the `git diff` command in the action might make it easier to figure out the desired behaviour. A contributing file might also help or maybe a PR template?

gjtempleton:
Thanks, I'm having a look into what the flow with multiple PR templates for different components within the repo might look like at the moment, but good call on the workflow exit code, will add that.
/lgtm
/approve

stevehipwell:
@gjtempleton is it still a manual process to get a chart released after the PR is merged?

gjtempleton:
🤦 This'll be available once #4227 is merged.

Issue #4206, opened 2021-07-19
[cluster-autoscaler] Additional expander options


ryanmcnamara:
Users are able to specify the `expander` [option](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-expanders) to change which nodegroup (e.g. ASG) is selected when there are multiple that satisfy the scheduling requirements for a pod. The feature request is to allow selecting an additional expanders in the event that the chosen expander *still* arbitrarily ties multiple nodegroups. This is very similar to this closed ticket https://github.com/kubernetes/autoscaler/issues/2564.

For our specific use case, we are currently using the `least-waste` expander, but we want to have some nodegroups which are preferred. What we would then like is to use the `priority` expander but fall back to `least-waste` when multiple still match.

I see several ways this feature could be added, in my order of preference
1. Use a `list` of expanders to use, with some validation of course very similar to what @adrienjt mentions [here](https://github.com/kubernetes/autoscaler/issues/2564#issuecomment-825116554). Satisfying my above use case would look like setting `expander=priority,least-waste`.
1. Add a `priority-expander` option, specifically only used in the case where `expander=priority`. Satisfying the above use case would look like setting `expander=priority` and `priority-expander=least-waste`

I think the first option is cleaner / easier to understand but is perhaps more complicated to implement. I'd be happy to submit a PR for this but wanted to get any feedback first, thanks!

cc @Jeffwan @bpineau @adrienjt

x13n:
`mostpods` and `waste` expanders, similarly to `priority`, have a notion of fallback strategy. Option 1 is much cleaner from API point of view, and also fairly easy to implement - we'd just need to update expander factory to allow chaining (and return an error if there's an expander configured after a one that doesn't need fallback).

ryanmcnamara:
Thanks @x13n 
I'll probably take a shot at this over the next several days and push something else. Would you like to review or know who I should get for a review?

x13n:
I'm happy to review the change, but you'll still need a repo owner to approve.

x13n:
This was done in https://github.com/kubernetes/autoscaler/pull/4233

/close

stevehipwell:
@x13n which versions of CA have this code?

I think there is a bigger question here on backporting vs recommending newer CA builds for supported K8s versions, does this need a separate issue opening?

x13n:
[...]

PR #4233, opened 2021-07-30
Allow specification of multiple expanders

ryanmcnamara:
Multiple expanders can now be specified, expanders now "filter to the
tied for best" instead of "selecting the best" so the output of one
expander is now fed to the input of the next. Each expander may only
be used once to disallow bad configuration. This should not be a change
in functionality as in the event of a tie the random expander is still
used.

### DIFF
--- a/cluster-autoscaler/expander/factory/chain.go
+++ b/cluster-autoscaler/expander/factory/chain.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package factory
+
+import (
+	"k8s.io/autoscaler/cluster-autoscaler/expander"
+
+	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
+)
+
+type chainStrategy struct {
+	filters  []expander.Filter
+	fallback expander.Strategy
+}
+
+func newChainStrategy(filters []expander.Filter, fallback expander.Strategy) expander.Strategy {
+	return &chainStrategy{
+		filters:  filters,
+		fallback: fallback,
+	}
+}
+
+func (c *chainStrategy) BestOption(options []expander.Option, nodeInfo map[string]*schedulerframework.NodeInfo) *expander.Option {
+	filteredOptions := options
+	for _, filter := range c.filters {
+		filteredOptions = filter.BestOptions(filteredOptions, nodeInfo)
+		if len(filteredOptions) == 1 {
+			return &filteredOptions[0]
+		}
+	}
+	return c.fallback.BestOption(filteredOptions, nodeInfo)
+}
--- a/cluster-autoscaler/expander/factory/expander_factory.go
+++ b/cluster-autoscaler/expander/factory/expander_factory.go
@@ -31,30 +31,48 @@ import (
 	kube_client "k8s.io/client-go/kubernetes"
 )
 
-// ExpanderStrategyFromString creates an expander.Strategy according to its name
-func ExpanderStrategyFromString(expanderFlag string, cloudProvider cloudprovider.CloudProvider,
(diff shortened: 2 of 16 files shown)


ryanmcnamara:
fixes https://github.com/kubernetes/autoscaler/issues/4206

ryanmcnamara:
Built and deployed this locally and confirmed I was able to use `--expander=priority,least-waste` and it worked as expected

cluster-autoscaler/expander/factory/expander_factory.go
```diff
@@ -31,30 +31,41 @@ import (
...
+	seenExpanders := map[string]struct{}{}
+	for _, expanderFlag := range expanderFlags {
+		if _, ok := seenExpanders[expanderFlag]; ok {
+			return nil, errors.NewAutoscalerError(errors.InternalError, "Expander %s was specified multiple times, each expander must not be specified more than once", expanderFlag)
+		}
+		seenExpanders[expanderFlag] = struct{}{}
+
+		switch expanderFlag {
+		case expander.RandomExpanderName:
+			filters = append(filters, random.NewFilter())
```
x13n, line 48: Shouldn't it be an error to specify an expander that's guaranteed to provide a single option with a fallback? E.g. `random,least-waste` is not really a reasonable config.

cluster-autoscaler/expander/factory/chain.go
```diff
@@ -0,0 +1,27 @@
...
+	return &chainStrategy{
+		filters:  filters,
+		fallback: fallback,
+	}
+}
+
+func (c *chainStrategy) BestOption(options []expander.Option, nodeInfo map[string]*schedulerframework.NodeInfo) *expander.Option {
+	filteredOptions := options
+	for _, filter := range c.filters {
+		filteredOptions = filter.BestOptions(filteredOptions, nodeInfo)
```
x13n, line 24: There is non-trivial logic behind BestOptions call sometime, so maybe worth returning earlier if there's just one option left?

ryanmcnamara: +1, will do

cluster-autoscaler/expander/factory/expander_factory.go
ryanmcnamara: True that the configuration is not a reasonable one, but I felt that catching this case wasn't worth the effort as it would require disambiguating between expanders that

1. Always return one result and
2. Those that can return more than one.

My thinking is that since it won't create problems then we can keep the simplicity of allowing this. Happy to be overruled on this though.

ryanmcnamara: Here's what it would look like: https://github.com/ryanmcnamara/autoscaler/compare/rm/expander-chain...ryanmcnamara:rm/expander-chain-demo?expand=1

cluster-autoscaler/expander/factory/chain.go
ryanmcnamara: done

cluster-autoscaler/expander/factory/expander_factory.go
x13n: Yup, something like this. I'm not insisting on this one, but since you're generating an error for one class of invalid configs (duplicated entries), I thought it may make sense to go a step further and disallow bad ordering as well.

ryanmcnamara: Ok addressed in this pr now!

ryanmcnamara:
Looking for a review from a code owner, perhaps @MaciekPytel or if not can you redirect? Thanks!

cluster-autoscaler/expander/factory/expander_factory.go
x13n: Neat, thanks! Maybe worth adding a comment about this behavior? May be surprising for anyone adding a new expander in the future. Please also squash commits before merging.

x13n:
I'm about to go on vacation tomorrow, so to make sure it's not blocked on me in the meantime:

/lgtm
/hold until the commits are squashed
/assign @MaciekPytel for final approval

x13n:
Just realized this is still open. Can you squash the commits?

ryanmcnamara:
@x13n done!

mwielgus (approved):
/lgtm
/approve

mwielgus:
Please fix:
```
Boilerplate header is wrong for: /home/runner/work/autoscaler/autoscaler/go/src/k8s.io/autoscaler/cluster-autoscaler/expander/factory/chain_test.go
```

ryanmcnamara:
@mwielgus done, thanks for the review

jonathanbbai:
I don't want to upgrade my cluster to 1.23 but i really like this feature, You think if I upgrade the CA version to 1.23 while on a 1.20 cluster, what are the chances it'll work?

PR #4236, opened 2021-08-03
implement GetOptions for GCE

bpineau:
Support per-MIG (scaledown) settings as permited by the
cloudprovider's interface `GetOptions()` method.

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/gce_manager.go
+++ b/cluster-autoscaler/cloudprovider/gce/gce_manager.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"reflect"
 	"regexp"
 	"strconv"
 	"strings"
@@ -30,6 +31,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
+	"k8s.io/autoscaler/cluster-autoscaler/config"
 	"k8s.io/autoscaler/cluster-autoscaler/config/dynamic"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/units"
 	"k8s.io/client-go/util/workqueue"
@@ -90,6 +92,8 @@ type GceManager interface {
 	GetResourceLimiter() (*cloudprovider.ResourceLimiter, error)
 	// GetMigSize gets MIG size.
 	GetMigSize(mig Mig) (int64, error)
+	// GetMigOptions returns MIG's NodeGroupAutoscalingOptions
+	GetMigOptions(mig Mig, defaults config.NodeGroupAutoscalingOptions) *config.NodeGroupAutoscalingOptions
 
 	// SetMigSize sets MIG size.
 	SetMigSize(mig Mig, size int64) error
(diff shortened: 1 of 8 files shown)


bpineau:
Maciej, would you mind having a look? This is entirely standing on your amazing "per Nodegroups configs" work!
/assign @MaciekPytel

cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -131,6 +134,38 @@ func getKubeEnvValueFromTemplateMetadata(template *gce.InstanceTemplate) (string
...
+	if err != nil {
+		return nil, fmt.Errorf("could not obtain kube-env from template metadata; %v", err)
+	}
+
+	options, err := extractAutoscalingOptionsFromKubeEnv(kubeEnvValue)
+	if err != nil {
+		return nil, err
+	}
+
+	if opt, ok := getFloat64Option(options, template.Name, "scaledownutilizationthreshold"); ok {
```
mwielgus, line 153: These strings occur in other cloud provider PRs as well. Is it possible to make them global constants?

mwielgus (changes requested):

cluster-autoscaler/cloudprovider/gce/templates.go
MaciekPytel: Given the way to pass the config is different between providers do we really see value in shared string consts? I can easily see providers striving to make them consistent with the rest of config for given provider (ex. use the same prefix), so we can't guarantee they would be portable anyway. 

That being said I would also prefer if the strings were consts instead (whether defined at provider level or as something like DefaultScaleDownUtilizationThresholdKey next to NodeGroupAutoscalingOptions definition).

cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -452,6 +487,51 @@ func extractOperatingSystemDistributionFromKubeEnv(kubeEnv string) OperatingSyst
...
+		klog.Warningf("error while obtaining autoscaling_options from AUTOSCALER_ENV_VARS; %v", err)
+		return nil, err
+	}
+
+	if !found {
+		klog.V(5).Info("no autoscaling_options defined in AUTOSCALER_ENV_VARS")
+		return make(map[string]string), nil
+	}
+
+	return parseKeyValueListToMap(optionsAsString)
```
MaciekPytel, line 532: I'd consider logging the values obtained from template (the actual parsed value).

cluster-autoscaler/cloudprovider/gce/templates.go
mwielgus: Currently we have in-code constants that the author defined (I assume) that happen to be exactly the same as on other cloud providers. We could try to push it more towards a standard. 
But if you prefer local constants then I'm fine with this solution as well. It will still be better than magical strings hidden in code :).

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -522,6 +525,15 @@ func (m *gceManagerImpl) findMigsInRegion(region string, name *regexp.Regexp) ([
 	return links, nil
 }
 
+// GetMigOptions merges default options with user-provided options as specified in the MIG's instance template metadata
+func (m *gceManagerImpl) GetMigOptions(mig Mig, defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
+	template, err := m.migInstanceTemplatesProvider.GetMigInstanceTemplate(mig.GceRef())
+	if err != nil {
+		return nil, err
+	}
+	return m.templates.BuildAutoscalingOptions(template, defaults)
```
MaciekPytel, line 534: I don't think this should be called here directly:
 * GetOptions is similar to MinSize() and MaxSize(): CA code calls it at any point in the loop and we generally assume that:
   * The value returned should not change during CA loop. 
   * The call is "free" from performance point of view (ie. no API calls should happen as a result).

The latter is probably already true due to caching in migInstanceTemplatesProvider, but I think the way to update any provider level settings should be to fetch it in Refresh() call and only return cached value here. In this case I propose to move extractAutoscalingOptionsFromKubeEnv and associated logic into forceRefresh() call and only apply the options to defaults in GetMigOptions.

WDYT?

cluster-autoscaler/cloudprovider/gce/templates.go
bpineau: Made that shared constants (we don't often have opportunities to enforce some consistency ;), and changed AWS and Azure PRs to draft until this gets in (so I'll refresh them to pick those symbols + feedback from this PR).

cluster-autoscaler/cloudprovider/gce/templates.go
bpineau: Good point, added a log entry as part of the refactoring suggested in your other comment.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
bpineau: Updated as suggested

bpineau:
@mwielgus and @MaciekPytel Thanks for the reviews; updated following the suggestion + re-tested, please PTAL.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -321,11 +325,40 @@ func (m *gceManagerImpl) forceRefresh() error {
...
+			klog.Warningf("Not evaluating autoscaling options for %q MIG: failed to find corresponding instance template", mig.GceRef(), err)
+			continue
+		}
+		if template.Properties == nil {
+			klog.Warningf("Failed to extract autoscaling options from %q metadata: instance template is incomplete", template.Name)
+			continue
+		}
+		kubeEnvValue, err := getKubeEnvValueFromTemplateMetadata(template)
+		if err != nil {
+			klog.Warningf("Failed to extract KubeEnv from %q instance template's metadata: %v", template.Name, err)
```
MaciekPytel, line 347: nit: this is a very generic warning (we get a lot of stuff from kube env in template). I like the ones above more, as they make it clear where the error is coming from.

MaciekPytel:
/lgtm
/approve
/hold
Left hold for a minor nit. Please feel free to remove it if you disagree or fix.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
bpineau: Right, updated to make it clear this comes from extracting autoscaling options.

PR #4237, opened 2021-08-03
implement GetOptions for Azure

bpineau:
Support per-VMSS (scaledown) settings as exposed by the cloudprovider's interface `GetOptions()` method.

For ref, providing similar GetOptions support for other providers:
* [GCE version](https://github.com/kubernetes/autoscaler/pull/4236)
* [AWS version](https://github.com/kubernetes/autoscaler/pull/4238/)

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_cache.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_cache.go
@@ -17,6 +17,7 @@ limitations under the License.
 package azure
 
 import (
+	"reflect"
 	"regexp"
 	"strings"
 	"sync"
@@ -53,6 +54,7 @@ type azureCache struct {
 	registeredNodeGroups []cloudprovider.NodeGroup
 	instanceToNodeGroup  map[azureRef]cloudprovider.NodeGroup
 	unownedInstances     map[azureRef]bool
+	autoscalingOptions   map[azureRef]map[string]string
 }
 
 func newAzureCache(client *azClient, cacheTTL time.Duration, resourceGroup, vmType string) (*azureCache, error) {
@@ -67,6 +69,7 @@ func newAzureCache(client *azClient, cacheTTL time.Duration, resourceGroup, vmTy
 		registeredNodeGroups: make([]cloudprovider.NodeGroup, 0),
 		instanceToNodeGroup:  make(map[azureRef]cloudprovider.NodeGroup),
 		unownedInstances:     make(map[azureRef]bool),
+		autoscalingOptions:   make(map[azureRef]map[string]string),
 	}
 
 	if err := cache.regenerate(); err != nil {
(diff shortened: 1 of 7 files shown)


bpineau:
Converted to draft until the corresponding GCE pull request gets in, so I can backport improvement suggested during GCE review there, and use global symbols defined in that other PR.

bpineau:
Updated with feedback and constants from the GCE equivalent PR, ready for review.
/assign @marwanad

cluster-autoscaler/cloudprovider/azure/azure_manager.go
```diff
@@ -246,6 +247,29 @@ func (m *AzureManager) GetNodeGroupForInstance(instance *azureRef) (cloudprovide
 	return m.azureCache.FindForInstance(instance, m.config.VMType)
 }
 
+// GetScaleSetOptions parse options extracted from VMSS tags and merges them with provided defaults
+func (m *AzureManager) GetScaleSetOptions(scaleSetName string, defaults config.NodeGroupAutoscalingOptions) *config.NodeGroupAutoscalingOptions {
```
marwanad, line 251: should we propagate an error back from here? How are you handling those errors for other cloudprovider implementations?

cluster-autoscaler/cloudprovider/azure/azure_cache.go
```diff
@@ -117,10 +120,22 @@ func (m *azureCache) regenerate() error {
...
 
+	// Regenerate VMSS to autoscaling options mapping.
+	newAutoscalingOptions := make(map[azureRef]map[string]string)
+	for _, vmss := range m.scaleSets {
+		ref := azureRef{Name: *vmss.Name}
+		options := extractAutoscalingOptionsFromScaleSetTags(vmss.Tags)
+		if !reflect.DeepEqual(m.getAutoscalingOptions(ref), options) {
+			klog.V(4).Infof("Extracted autoscaling options from %q ScaleSet tags: %v", *vmss.Name, options)
+		}
+		newAutoscalingOptions[ref] = options
```
marwanad, line 131: nit: should this be done in the above if check?

marwanad (commented):
lgtm, just a small comment and a question

cluster-autoscaler/cloudprovider/azure/azure_manager.go
bpineau: Intent is not to impact operations on errors, but make them visible and auditable. Missing options tags is just fine (and likely). Likewise for incomplete/partial tags set. We log and ignore bogus (or similar-looking yet different, etc) tag values. That's similar for aws and gce cloudproviders.

cluster-autoscaler/cloudprovider/azure/azure_cache.go
bpineau: We're rebuilding that option cache from scratch (`newAutoscalingOptions` is empty before that loop) so we have to store both changed and previously known options there.

marwanad: oh gotcha

PR #4238, opened 2021-08-03
implement GetOptions for AWS

bpineau:
Support per-ASG (scaledown) settings as specified by the cloudprovider's interface GetOptions() method.

For ref, merged PRs providing similar GetOptions support for other providers:
* [GCE version](https://github.com/kubernetes/autoscaler/pull/4236)
* [Azure version](https://github.com/kubernetes/autoscaler/pull/4237)

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/README.md
+++ b/cluster-autoscaler/cloudprovider/aws/README.md
@@ -162,6 +162,20 @@ Example tags:
 
 - `k8s.io/cluster-autoscaler/node-template/resources/ephemeral-storage`: `100G`
 
+ASG labels can specify autoscaling options, overriding the global cluster-autoscaler
+settings for the labeled ASGs. Those labels takes the same values format as the
+cluster-autoscaler command line flags they override (a float or a duration, encoded
+as string). Currently supported autoscaling options (and example values) are:
+
+* `k8s.io/cluster-autoscaler/node-template/autoscaling-options/scaledownutilizationthreshold`: `0.5`
+  (overrides `--scale-down-utilization-threshold` value for that specific ASG)
+* `k8s.io/cluster-autoscaler/node-template/autoscaling-options/scaledowngpuutilizationthreshold`: `0.5`
+  (overrides `--scale-down-gpu-utilization-threshold` value for that specific ASG)
+* `k8s.io/cluster-autoscaler/node-template/autoscaling-options/scaledownunneededtime`: `10m0s`
+  (overrides `--scale-down-unneeded-time` value for that specific ASG)
+* `k8s.io/cluster-autoscaler/node-template/autoscaling-options/scaledownunreadytime`: `20m0s`
+  (overrides `--scale-down-unready-time` value for that specific ASG)
+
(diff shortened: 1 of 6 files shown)


cluster-autoscaler/cloudprovider/aws/aws_manager.go
```diff
@@ -419,6 +459,18 @@ func extractLabelsFromAsg(tags []*autoscaling.TagDescription) map[string]string
 	return result
 }
 
+func extractAutoscalingOptionsFromTags(tags []*autoscaling.TagDescription) map[string]string {
+	options := make(map[string]string)
+	for _, tag := range tags {
+		splits := strings.Split(aws.StringValue(tag.Key), "k8s.io/cluster-autoscaler/node-template/autoscaling-options/")
```
younseunghyun, line 465: I'm not the reviewer, but this logic can occur a not expected behavior.

If the key is `[space]k8s.io/cluster-autoscaler/node-template/autoscaling-options/` , then key is `space` and value is `k8s.io/cluster-autoscaler/node-template/autoscaling-options/`.

Not a crucial problem, but that situation is not intended.

And I think this logic may be not performance critical, so we just can check the key has suffix `k8s.io/cluster-autoscaler/node-template/autoscaling-options/` and then parse the value.

This is not crucial, just a opinion.

bpineau: Note that was mirroring the way other ASG labels parsers (`extractLabelsFromAsg`, `extractAllocatableResourcesFromAsg`) are implemented, and such bogus labels are detected, ignored and logged at parsing time. I don't mind either way though, so updated to filter out on prefix as suggested.

bpineau:
Converted to draft until the corresponding GCE pull request gets in, so I can port improvement suggested during GCE review there, and leverage global symbols defined in that other PR.

bpineau:
Rebased and updated on GCE equivalent, it's ready for review
/assign @gjtempleton

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -162,6 +162,20 @@ Example tags:
 
 - `k8s.io/cluster-autoscaler/node-template/resources/ephemeral-storage`: `100G`
 
+ASG labels can specify autoscaling options, overriding the global cluster-autoscaler
+settings for the labeled ASGs. Those labels takes the same values format as the
+cluter-autoscaler command line flags they override (a float or a duration, encoded
```
gjtempleton, line 167:
```suggestion
cluster-autoscaler command line flags they override (a float or a duration, encoded
```

gjtempleton (commented):
Sorry for the delay in getting this reviewed!

One tiny spelling nitpick from me, but other than that, this looks good to go for me.

cluster-autoscaler/cloudprovider/aws/README.md
bpineau: Thanks for catching this, updated & rebased accordingly!

gjtempleton:
Thanks!
/lgtm
/approve

PR #4253, opened 2021-08-11
Extend ScaleUpStatus structure with ScaleUpError field.

olagacek:
Extend ScaleUpStatus structure with ScaleUpError field in order to provide more information about failed scale up.


cluster-autoscaler/core/scale_up.go
```diff
@@ -477,10 +482,11 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 			klog.V(1).Infof("Capping size to max cluster total size (%d)", context.MaxNodesTotal)
 			newNodes = context.MaxNodesTotal - len(nodes) - len(upcomingNodes)
 			context.LogRecorder.Eventf(apiv1.EventTypeWarning, "MaxNodesTotalReached", "Max total nodes in cluster reached: %v", context.MaxNodesTotal)
+			err = errors.NewAutoscalerError(
```
MaciekPytel, line 485: Shouldn't that happen inside an if and not outside?

cluster-autoscaler/core/scale_up.go
```diff
@@ -341,7 +341,8 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 
 	nodesFromNotAutoscaledGroups, err := utils.FilterOutNodesFromNotAutoscaledGroups(nodes, context.CloudProvider)
 	if err != nil {
-		return &status.ScaleUpStatus{Result: status.ScaleUpError}, err.AddPrefix("failed to filter out nodes which are from not autoscaled groups: ")
+		err = err.AddPrefix("failed to filter out nodes which are from not autoscaled groups: ")
+		return &status.ScaleUpStatus{Result: status.ScaleUpError, ScaleUpError: &err}, err
```
MaciekPytel, line 345: nit: this feels a bit boilerplate-y. Maybe a helper function would be nice? I'm thinking something like `return scaleUpError(err.AddPrefix())`.

I don't feel strongly about this at all, please feel free to ignore this comment if you don't like the idea.

MaciekPytel (commented):
/approve
/lgtm

/hold
Hold to address a minor comment. Please feel free to cancel it once done.

cluster-autoscaler/core/scale_up.go
olagacek: Agree with your suggestion, I created scaleUpError() func which takes ScaleUpStatus (it is required as sometimes we pass additional information to ScaleUpStatus) and error as arguments

cluster-autoscaler/core/scale_up.go
olagacek: Moved it

cluster-autoscaler/core/scale_up.go
```diff
@@ -350,24 +350,25 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
...
 			errors.CloudProviderError,
-			errCP)
+			errCP))
 	}
 
 	scaleUpResourcesLeft, errLimits := computeScaleUpResourcesLeftLimits(context, processors, nodeGroups, nodeInfos, nodesFromNotAutoscaledGroups, resourceLimiter)
 	if errLimits != nil {
-		return &status.ScaleUpStatus{Result: status.ScaleUpError}, errLimits.AddPrefix("Could not compute total resources: ")
+		err = errLimits.AddPrefix("Could not compute total resources: ")
+		return &status.ScaleUpStatus{Result: status.ScaleUpError, ScaleUpError: &err}, err
```
MaciekPytel, line 361: nit: I think you missed converting this one to scaleUpError()?

Issue #3802, opened 2021-01-11
Cluster Autoscaler does not start new nodes when Taints and NodeSelector are used in EKS


dschunack:
Hi,

we use EKS with kubernetes 1.18 and the Cluster Autoscaler. With kubernetes 1.17 the "beta.kubernetes.io/Instance-type" is [deprecated](https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/). We use instead the new "node.kubernetes.io/instance-type" as NodeSelector. This is working for autoscaling groups without taints. For the autoscaling groups with taints is the new "node.kubernetes.io/instance-type" selector not working and the cluster autoscaler doesn't start new nodes. If we switch back to the old and [deprecated](https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/) "beta.kubernetes.io/instance-type" Selector the cluster autoscaler starts a new Node. We see this behavior on all of our EKS.

Events output for both Test PODs with beta and node.kubernetes.io as NodeSelector.
POD with node.kubernetes.io selector was started first.

```
% kubectl get pods
NAME                READY   STATUS    RESTARTS   AGE
test-4xlarge-beta   0/1     Pending   0          41s
test-4xlarge-node   0/1     Pending   0          72s

% kubectl describe pod test-4xlarge-node
Name:         test-4xlarge-node
Namespace:    default
Priority:     0
Node:         <none>
Labels:       <none>
Annotations:  kubernetes.io/psp: eks.privileged
Status:       Pending
IP:           
IPs:          <none>
Containers:
  test-4xlarge-node:
Conditions:
  Type           Status
  PodScheduled   False 
Volumes:
  default-token-lzknk:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  default-token-lzknk
    Optional:    false
QoS Class:       BestEffort
Node-Selectors:  node.kubernetes.io/instance-type=c5a.4xlarge
Tolerations:     disk=true:NoSchedule
                 node.kubernetes.io/not-ready:NoExecute op=Exists for 300s
                 node.kubernetes.io/unreachable:NoExecute op=Exists for 300s
Events:
  Type     Reason             Age               From                Message
  ----     ------             ----              ----                -------
  Normal   NotTriggerScaleUp  88s               cluster-autoscaler  pod didn't trigger scale-up (it wouldn't fit if a new node is added): 12 node(s) didn't match node selector
[...]

Issue #4052, opened 2021-04-30
predicate checking error: node(s) didn't match Pod's node affinity; predicateName=NodeAffinity; reasons: node(s) didn't match Pod's node affinity; debugInfo=


imriss:
**Which component are you using?**:

`cluster-autoscaler`

**What version of the component are you using?**:

Component version: `9.9.2` (via `Helm` chart)

**What k8s version are you using (`kubectl version`)?**:

<details><summary><code>kubectl version</code>
</summary><br><pre>
</pre></details>

```
Server Version: version.Info{Major:"1", Minor:"19+", GitVersion:"v1.19.6-eks-49a6c0", GitCommit:"49a6c0bf091506e7bafcdb1b142351b69363355a", GitTreeState:"clean", BuildDate:"2020-12-23T22:10:21Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"
```

**What environment is this in?**:

Cluster: `EKS` (`k8s` 1.19)
original ASG's node type: `r5.2xlarge`
Other ASGs' node types: `m5.2xlarge` and `m5.xlarge`

**What did you expect to happen?**:

After a node from an ASG was drained and terminated, the evicted pod indefinitely went in the `Pending` state with the following error:
```
predicate checking error: node(s) didn't match Pod's node affinity; predicateName=NodeAffinity; reasons: node(s) didn't match Pod's node affinity; debugInfo=
```
Note: The ASG group, which had that pod hosted in one of its nodes, was at max, and cannot spin new nodes. While the other ASGs could spin new nodes. Also, I tried to increase the number of nodes in the other ASGs, but the pod still could not be assign to those nodes with the same error. The CPU/RAM request and limit of the pod is less than of those of all ASGs. 

**What happened instead?**:

The evicted pod should allocated to other nodes, or a new node from other ASGs should be added so that pod can be allocated.

**How to reproduce it (as minimally and precisely as possible)**:

Possible scenario to reproduce: Have all the nodes of an ASG cordoned. Then, drain one of them, and then terminate it while reducing the max of that ASG by one at the same time.

**Anything else we need to know?**:

tsunamishaun:
seems related to [#3802](https://github.com/kubernetes/autoscaler/issues/3802#issuecomment-853382182) ?

imriss:
any update?

sidewinder12s:
[...]

Issue #3780, opened 2020-12-21
Cluster Autoscaler scaling from 0 on AWS EKS stops working after CA pod is restarted


marwan116:
**Which component are you using?**:  cluster-autoscaler

**What version of the component are you using?**: 

Component version: `1.18.3`

**What k8s version are you using (`kubectl version`)?**: 1.18

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version

</pre></details>

**What environment is this in?**: 
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"clean", BuildDate:"2020-11-14T14:49:35Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.9-eks-d1db3c", GitCommit:"d1db3c46e55f95d6a7d3e5578689371318f95ff9", GitTreeState:"clean", BuildDate:"2020-10-20T22:18:07Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

**What did you expect to happen?**: I have a `cluster-autoscaler` deployed on AWS EKS. It is scaling three non-managed nodegroups from 0 and is running smoothly. However after a period of time (which seems to be somewhere between 14 and 20 days) the autoscaler seems to "lose" visibility of some of the nodegroups and starts failing to find a place to schedule the pods. 

Please note that this behavior has been consistent over the past three months at least (over different versions of the austocaler on different versions of kubernetes - I have tried this also on kuberentes 1.17 with cluster-autoscaler version 1.17.4 and 1.17.3). 

Also please note that no modifications have been made to the pod spec or the nodes that I have been using (the pods get deployed as part of a scheduled job).

To resolve this issue - I have to manually scale the nodegroup to a number of nodes that is non-zero and different from the "desired capacity" - the nodegroups then become visible again to the autoscaler and it resumes functioning properly.

Please see the following logs of when the autoscaler fails to fit a pod

```
I1221 18:43:28.403351       1 filter_out_schedulable.go:60] Filtering out schedulables
I1221 18:43:28.403424       1 filter_out_schedulable.go:116] 0 pods marked as unschedulable can be scheduled.
[...]

Issue #3916, opened 2021-03-02
ratelimit on cluster scale-up


sulixu:
**Which component are you using?**: cluster autoscaler 1.16.7

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:
I would like have a sort of ratelimit mechanism on cluster scale-up, for example if cluster autoscaler decides to scale up 500 nodes at a time because of a submission from 1 batch workload, we don't want to scale up 500 nodes all at once, it would be nice to have a few controlled and gradual scale-ups to alleviate the sudden load to kube api server. 

**Describe the solution you'd like.**:

I have a draft on https://github.com/sulixu/autoscaler/commit/3e53de7caeb1002fe1e8a6e7026ebaaf632222cf and tested the scale-up ratelimiting, it seems to be working as expected.

for example I set 5 nodes per batch per min (with burst at 10 nodes per min)

to scale up 25 nodes: (with 40 secs scan-interval)

t1 -> t2 -> t3 -> t4 -> t5 -> t6 -> t7
10 -> idle -> 5 -> idle -> 5 -> idle -> 5

I understand this ratelimiting won't be fair for all clients, for example at the same time 1 batch workload has 500 pods pending, and 1 long running service have 2 pods pending, if all the pending pods have the same priority, the 2 pods from the long running services will always have a lower chance of getting scheduled, and will likely be pending until most of batch pods have scheduled. however we would set the pod priority to let the long running service's pod always have a higher priority than others.

**Describe any alternative solutions you've considered.**:

increasing the scan-interval won't help as it cannot control the number of new nodes in a single scale-up

**Additional context.**:

We have seen that a sudden massive increase in nodes during cluster scale-up can overwhelm the cluster. the control plane got into a busy state and it overloaded the api server.

sonowz:
> We have seen that a sudden massive increase in nodes during cluster scale-up can overwhelm the cluster. the control plane got into a busy state and it overloaded the api server.

I second this

MaciekPytel:
For anyone looking to implement this - I think it could probably be implemented fairly easily as a NodeGroupSetProcessor (one that rate limits scale-up rather than just splitting it across zones).

anson627:
Hi @MaciekPytel , can you elaborate on how to use NodeGroupSetProcessor to implement rate limits scale-up? Also I'm exploring ways to circuit break scale up when cluster reaching certain size (in terms of # of pods and # of nodes)

[...]

PR #4258, opened 2021-08-12
Added node readiness grace time & node-info cache expiration

dany74q:
Fixes https://github.com/kubernetes/autoscaler/issues/3802, https://github.com/kubernetes/autoscaler/issues/4052 and potentially some other scale-up-from-0 scenarios (a-la https://github.com/kubernetes/autoscaler/issues/3780).

Quick context - In order to scale up a specific node group, the cluster autoscaler needs to be able to do the following:
a. Fetch all available node groups of given cluster, by querying the cloud provider
b. Build a virtual Kubernetes Node representation for each such node group (i.e. - how would a node from such a node group would look like once alive)
c. Match pending pods against every such representation, and scale-up the matching node-group of the matched node

The autoscaler currently builds such node representations by first looking at alive Kubernetes nodes in the cluster - and only if a node group had no current Kubernetes node alive (when it's scaled-down to 0), 
would the autoscaler query the cloud provider and inferring the node-template from the tags/labels/capacity the cloud provider's API provides; 
that is because the autoscaler works under the premise that live nodes might hold extra information not available in the provider (for instance, additional labels which are added at runtime, by kubelet or otherwise).

Keeping a solid node-template view in-memory for node groups which scale to and fro from zero and upwards is a bit tricky - the current implementation opts for the following:
a. On each autoscaling cycle, all visible & ready Kubernetes nodes are cached to a non-expiring cache
b. If a node-group node representation was already cached, it's overridden
c. The first Kubernetes-supplied node from a given node-group is chosen to represent nodes in that group

This could potentially lead to the following race conditions:
1. Stale reads - when a node group scales back to zero, the autoscaler would continue using the last surviving Kubernetes node representation of that group available in cache (as entries don't expire); 
this could be problematic on two fronts:
1.1. Underlying node group changes would not be reflected until the autoscaler restarts (e.g. new labels / tags are added, or the instance type is changed)
1.2. When the last surviving node cached had some partial view of how a node from the given group should look like, for instance - if there is some runtime un-labeling going on pre-termination - the autoscaler might deem this group not fit for pods it can, in fact host

2. Partial view during node initialization - There are several workloads which modify the node's attributes at runtime, consider for instance - a Daemonset which installs an agent or a kernel module on a given node, and then labels it ready to accept workload; one might purposefully 'trick' the autoscaler to accept such node-groups for scheduling by adding node-template tags to the underlying node group,
thus having their desired view of a node in their cloud provider, which may take several minutes to reflect in Kubernetes nodes themselves.

In such and similar circumstances, what may happen in case of a quick ramp-of of nodes from the given group,
is that Kubernetes might continuously return nodes which are still initializing as the first representative of the node-group;
this would cause the autoscaler to have a prolonged undesired view of its node-templates, until a node which is fully ready had stabilized and is then used as the node-group representative.

I propose the following two additions to the autoscaling loop, which I believe would eliminate such race conditions:
1. Per @lsowen's [suggestion](https://github.com/kubernetes/autoscaler/issues/3802#issuecomment-853382182) - We can introduce a slight grace period for the use of Kubernetes nodes as a node-group representative until they've hit a point we'd consider fully initialized; This means that for the first n minutes, we'd use the cloud provider's representation, or a more mature node from the node group as a representative.

2. Have the node info cache expire stale entries for node groups which scaled down to zero

This PR introduces the above two, with what I believe would be sensible defaults - a 5 minute node-readiness grace period, and a 10 minute cache expiration time.

I think that it'd be worthwhile to have non-zero defaults (meaning - to have the grace period and expiration on by default), as it could benefit managed autoscaler cloud provider solutions and new users without further tinkering with the runtime parameters, while having minimal impact (a few more cloud provider API hits) in cases where scaling to zero isn't utilized.

Would love to hear what you think - Thanks !

Changes:
autoscaling_options.go, main.go
- Added NodeReadinessGraceTime, which gives kubernetes nodes an extra grace period before basing node templates on them
- Added NodeReadinessGraceTime, which controls expiration for the k8s node group -> node info cache

FAQ.md
- Documented node-readiness-grace-time, node-cache-expiration-time and their usefulness

node_info_cache.go
- Introduced, small expiring cache wrapper for node info entries

scale_up_test.go, utils_test.go
- Fixed tests based on new GetNodeInfosForGroups signature
- Introduced sanity tests for the node grace time & cache expiration

static_autoscaler.go
- Will now use the new NodeInfoCache
- Will now pass the NodeReadinessGraceTime

utils.go
- Node info cache will now only populate nodes that are ready for a given readiness grace time


MaciekPytel:
The reason for taking an existing node over one used in scale-from-zero is that scale-from-zero is just an approximation that is always wrong. Things like memory or ephemeral-storage capacity are determined by kubelet at runtime and depends on things like operating system, disk formatting, etc making them effectively impossible to predict. Similarly, automatically assigned system labels are hard to keep track of - there are plenty of issues in this repo reporting CA not working with some system label or another when scaling from zero.

Note that the reason is not runtime changes like suggested above (unless you mean runtime changes during initialization?). CA is build on assumption that nodes don't change over time and we don't support adding or removing taints or labels once the node has started up (it may work in some cases, but there are no promises there). 

I think that a grace period is actually very dangerous. I've seen plenty of issues where CA incorrectly predicts the memory capacity of a node when scaling-from-0 and thinks the pod will fit on the node, while in reality it won't. In this case it will trigger an incorrect scale-up, but will only do so once. If we keep using provider template CA will just keep scaling up. It's also a reason why the cache exists in the first place - if it did CA will just keep retrying the incorrect scale-up every time it expires. Dropping the cache when nodepool is scaled down to 0 mostly defeats the purpose of having it at all. 

To answer specific problems:
1. CA generally assumes NodeGroups are immutable when it comes to node configuration. Hence the cache and the current logic. I'm not opposed to a flag that allows cache expiration or completely disables it, but I think the default should be the current behavior as dropping the cache will expose the problems I mentioned above (and is not necessary if NodeGroups are immutable).
2. The problems with partial initialization go far beyond this issue. If the pending pods depend on label / custom resource added by DS or similar you will still see additional, incorrect scale-up during node initialization. As I said above CA assumes node don't change and once the node becomes ready (based on kubernetes status conditions) and it doesn't expect a partially-initialized node to change anymore. If pending pods can't schedule on partially-initialized node CA will try to create another one. We already have a solution for this:
    * If partial initialization is specifically installing GPU drivers, it should just work as long as the node has provider-specific gpu label.
    * For any other custom initialization you can use "ignore-taint". Any node that has a taint prefixed with [ignore-taint.cluster-autoscaler.kubernetes.io/](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/utils/taints/taints.go#L34) will be considered still starting up until the taint is removed. So the way to do custom initialization is to start a node with ignore-taint and have your DS remove it at the same time it makes whatever other changes it does to the node (setting labels, adding custom resources, ...).
        * This should take care of all the issues. An upcoming node will not be used as template for new scale-up (which I think is what you are trying to achieve with grace period) and CA will know not to trigger another scale-up.
        * One caveat is that CA will timeout scale-up if the node takes too long to initialize. The default is 15 minutes, but IIRC you can change it with a flag.

lsowen:
@MaciekPytel This is great information, thank you for putting this together.

The issue seems to be that not all node labels which are set by kubelet are available at the time the node is reported as "ready".  For example, using autoscaling on AWS utilizes the `--node-labels` option to kubelet (https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/), and without a delay the "cached" nodeInfo does not include those labels.   

(some additional info https://github.com/kubernetes/autoscaler/issues/3802#issuecomment-846615442 and https://github.com/kubernetes/autoscaler/issues/3802#issuecomment-853382182)

dany74q:
Hey @MaciekPytel, thanks for the elaborate response !

Following up on what you said, I was wondering -

```CA is build on assumption that nodes don't change over time and we don't support adding or removing taints or labels once the node has started up (it may work in some cases, but there are no promises there).```

I've seen this mentioned in the docs, but how does this reconcile with the fact that on each scale-up cycle - [we override the cache](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/utils/utils.go#L90-L93) with the first node-from-node-group returned from k8s (and this behavior is cloud-agnostic) ?

Overriding cache entries on each scale-up attempt effectively updates the returned node info object, thus making the returned template dynamic in nature, meaning that if labels or taints are added / removed at runtime (on that first returned node from the given node group) - the returned node-info will always reflect that.

This brings me back to the point where any node-template modifications made at runtime - are in fact reflected and returned from the cache, which might break autoscaling - even permanently (until restart), if the node group just scaled down to zero - as the cache could now be in a state where it holds an incorrect view from the previously running node.


Putting that aside - I like the idea of using the ignore taint - this definitely mitigates the need for a grace period; one could also dynamically set the taint on any custom teardown logic (if labels are removed dynamically), so that the cache would not be overridden with the torn-down node-info.

What I still think could be beneficial to address in this PR though - is the cache-overriding behavior described above, I'm not sure if it's entirely aligned with the notion that CA doesn't support dynamic node modifications; 
maybe we should keep the first record instead of continuously update the cache, or alternatively - explain that a bit better in the docs ?

Thanks again !

MaciekPytel:
Ah, I wasn't aware of the labels being applied after node becomes ready. That's unfortunate :/ 

I think the way to address this should be similar to how GPU and ignore-taint work: treat the node as upcoming until all the labels have been added. The question is whether there is some way to look at the node and recognize if it already has all the labels? If so, it should be pretty simple to implement a [CustomResourcesProcessor](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/processors/customresources/custom_resources_processor.go) that would mark such nodes as upcoming.

As to your point on the behavior when node is updated: you're right, CA doesn't handle this well. Not only in cache / scale-from-0 case: even without cache CA selects a random node as a template, so if some nodes have been modified scale-up decision will be random. Essentially if you start modifying nodes you get undefined behavior (note: this doesn't apply to kubernetes status changes such as taint by condition).

> What I still think could be beneficial to address in this PR though - is the cache-overriding behavior described above, I'm not sure if it's entirely aligned with the notion that CA doesn't support dynamic node modifications;

If the assumption that nodes don't change is true it shouldn't make any difference. In practice - if anything I suspect the first template CA grabs that is more likely to be wrong (it's likely to be a freshly started node, meaning it will be wrong ex. in case of custom initialization without ignore-taint). It seems that unless we have smarter logic to validate, any strategy for selecting template to cache is susceptible to caching wrong template. At least with updates it's recoverable without restarting CA (by creating a new node). That being said, I'm not feeling strongly that updating is right strategy and I'm happy to discuss it further.

> one could also dynamically set the taint on any custom teardown logic (if labels are removed dynamically)

ignore-taint (and the GPU support it's based on) is essentially a hack that I added to allow existing CA logic deal with custom initialization. It works by making CA think the node is unready. This works fine for its intended use-case (nodes are expected to be unready during initial startup), but if you add an ignore-taint to a node that was not recently created CA will handle it like any unready node (ex. using `--scale-down-unready-time` instead of `--scale-down-unneeded-time`, counting the node as unhealthy when determining NodeGroup health, etc). 
Depending on the details of your use-case that may or may not be a problem, but you'd be entering uncharted territory.

> explain that a bit better in the docs ?

Admittedly, this stuff is pretty much not covered in the docs at all. There is huge room for improvement here.

MaciekPytel:
> one could also dynamically set the taint on any custom teardown logic (if labels are removed dynamically)

Actually, I'm thinking maybe having an equivalent of ignore-taint for this could be useful? I think it should be pretty easy to implement - just treat it the same as ignore-taint in template creation logic, but don't treat node as unready if the taint is there? It should cover any use-cases of taints added after the node is created.
If we went this way we should probably add a new name for ignore-taint (current one is historical and doesn't really describe it well) - I'm thinking maybe initialization-taint and status-taint (with ignore-taint equivalent to initialization-taint for backward compatibility)?

ezh:
Hi @dany74q. Thank you for an amazing job 👍 . Is it possible to introduce at least `node-readiness-grace-time`? IMO it will fix an issue with using extended resources for autoscaler. https://github.com/kubernetes/autoscaler/issues/3852

There is an actual behavior explanation https://github.com/kubernetes/autoscaler/issues/3852#issuecomment-769722674

For now, autoscaler is broken because of multiple scale-up requests.

ezh:
As I see, extended resources issue is tracking at https://github.com/kubernetes/autoscaler/issues/3916

PR #4261, opened 2021-08-13
Use highest available magnum microversion

### LINKED ISSUE #4254: [magnum] Scaling down nodegroups to 0 requires microversion bump
**Which component are you using?**: autoscaler with magnum

**What version of the component are you using?**: v1.22.0

Component version:

**What k8s version are you using (`kubectl version`)?**: 1.21.1

$ kubectl version

</pre></details>

**What environment is this in?**: testing with magnum from wallaby openstack release

**What did you expect to happen?**: when nodegroup has min_node_count set to 0, autoscaler should be able to remove all nodes from the group properly.

**What happened instead?**: even with 1.22.0 release, which adds support for empty nodegroups, scaling down fails due to microversion mismatch:
```E0811 10:55:11.158565       1 scale_down.go:1146] Problem with empty node deletion: failed to delete processing-tests-processin-3pxorznq65oy-node-0: manager error deleting nodes: could not resize cluster: Bad request with: [POST http://ctrl:9511/v1/clusters/23d8d1c6-d713-47ff-92f1-471092b96a7d/actions/resize], error message: {"errors": [{"request_id": "", "code": "client", "status": 400, "title": "Resizing a nodegroup to zero is not supported in the provided microversion", "detail": "Resizing a nodegroup to zero is not supported in the provided microversion.", "links": []}]}```

According to https://github.com/openstack/magnum/blob/master/magnum/api/rest_api_version_history.rst empty nodegroups are supported with 1.10+ microapi version.

**How to reproduce it (as minimally and precisely as possible)**: Create 

tghartland:
Fixes #4254 

When I enabled scale to 0 for the magnum provider, I was looking at this patch which allows node groups of size 0 with a new microversion, but did not require the microversion for making resize requests.
https://opendev.org/openstack/magnum/commit/f46923cc5ea74da663640bb0c470a4b1039c777e

The microversion requirement for resizing to 0 was added separately, later on.
https://opendev.org/openstack/magnum/commit/0e6d178939647e61fee1485e7b3611f093adb7c6

This patch makes the magnum client send "latest" as its microversion request, and on the server side magnum will replace it with the highest microversion that it supports.

When microversion 1.10 (or higher) is available, it will be used which allows the autoscaler to scale node groups to 0.

On magnum versions older than Wallaby, microversion 1.9 will still be used.
Attempting to scale to 0 would be an error, but magnum will also not allow the minimum node count of a group to be set to 0, so as long as node group autodiscovery is used then that error will never occur.

### DIFF
--- a/cluster-autoscaler/cloudprovider/magnum/README.md
+++ b/cluster-autoscaler/cloudprovider/magnum/README.md
@@ -16,6 +16,8 @@ using the cluster autoscaler v1.18 or lower.
 
 ## Updates
 
+* CA 1.22
+  * Allow scaling node groups to 0 nodes, if supported (requires Magnum Wallaby).
 * CA 1.19
   * Update to support Magnum node groups (introduced in Magnum Train).
     * Add node group autodiscovery based on the group's role property.
--- a/cluster-autoscaler/cloudprovider/magnum/magnum_manager.go
+++ b/cluster-autoscaler/cloudprovider/magnum/magnum_manager.go
@@ -29,6 +29,10 @@ import (
 const (
 	// Magnum microversion that must be requested to use the node groups API.
 	microversionNodeGroups = "1.9"
+	// Magnum microversion that must be requested to support scaling node groups to 0 nodes.
+	microversionScaleToZero = "1.10"
+	// Magnum interprets "latest" to mean the highest available microversion.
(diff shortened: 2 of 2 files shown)


mwielgus (approved):
/lgtm
/approve

PR #4269, opened 2021-08-16
Automated cherry pick of #3860 to upstream/cluster-autoscaler-release-1.19

gjtempleton:
As per https://github.com/kubernetes/autoscaler/issues/4251#issuecomment-899603689 we want the Github actions set up to test PRs for all supported release branches

PR #4260, opened 2021-08-12
cherry pick #3649 - Adding functionality to cordon the node before destroying it.

matthias50:
cherry pick #3649 (Adding functionality to cordon the node before destroying it) to 1.19

PR #4274, opened 2021-08-18
Cloud provider[Packet] fixes

ipochi:
This PR address certain issues found when deploying Cluster Autoscaler on Equinix Metal, formerly Packet.

1. fix panic on entry in nil map

    In the latest version of cluster-autoscaler (cloudprovider: packet), the
    code panics and the pods go into CrashLoopBackoff due to an entry
    assignment on a nil map.
    
2. Consider the string prefix `equinixmetal://`

    When K8s API is queried to get providerID from Node Spec, some machines
    return `packet://<uuid>`, whereas some return `equinixmetal://`, this
    creates error as the string is not trimmed properly and hence results in
    a  404 when an untrimmed string is queried to Equinix Metal API for
    device information.
    
3. Make controller node label configurable

    Currently the label to identify controller/master node is hard coded to
    `node-role.kubernetes.io/master`.

    There have been some conversations centered around replacing the label
    with `node-role.kubernetes.io/control-plane`.

    In [Lokomotive](https://github.com/kinvolk/lokomotive), the label to identify
    the controller/master node is `node.kubernetes.io/master`, the reasons
    for this is mentioned in this [issue](https://github.com/kinvolk/lokomotive/issues/227)

    Make the label configurable by setting an env variable in
    the deployment `CONTROLLER_NODE_IDENTIFIER_LABEL`, if set then the value
    in the env variable is used for identifying controller/master nodes, if
    not set/passed, then the existing behavior is followed choosing the
    existing label.

### DIFF
--- a/cluster-autoscaler/cloudprovider/packet/packet_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/packet/packet_cloud_provider.go
@@ -37,6 +37,11 @@ const (
 	ProviderName = "packet"
 	// GPULabel is the label added to nodes with GPU resource.
 	GPULabel = "cloud.google.com/gke-accelerator"
+	// DefaultControllerNodeLabelKey is the label added to Master/Controller to identify as
+	// master/controller node.
+	DefaultControllerNodeLabelKey = "node-role.kubernetes.io/master"
+	// ControllerNodeIdentifierEnv is the string for the environment variable.
+	ControllerNodeIdentifierEnv = "PACKET_CONTROLLER_NODE_IDENTIFIER_LABEL"
 )
 
 var (
@@ -94,7 +99,13 @@ func (pcp *packetCloudProvider) AddNodeGroup(group packetNodeGroup) {
 //
 // Since only a single node group is currently supported, the first node group is always returned.
 func (pcp *packetCloudProvider) NodeGroupForNode(node *apiv1.Node) (cloudprovider.NodeGroup, error) {
-	if _, found := node.ObjectMeta.Labels["node-role.kubernetes.io/master"]; found {
+	controllerNodeLabel := os.Getenv(ControllerNodeIdentifierEnv)
+	if controllerNodeLabel == "" {
+		klog.V(3).Infof("env %s not set, using default: %s", ControllerNodeIdentifierEnv, DefaultControllerNodeLabelKey)
+		controllerNodeLabel = DefaultControllerNodeLabelKey
+	}
+
+	if _, found := node.ObjectMeta.Labels[controllerNodeLabel]; found {
 		return nil, nil
 	}
 	nodeGroupId, err := pcp.packetManager.NodeGroupForNode(node.ObjectMeta.Labels, node.Spec.ProviderID)
--- a/cluster-autoscaler/cloudprovider/packet/packet_manager_rest.go
+++ b/cluster-autoscaler/cloudprovider/packet/packet_manager_rest.go
@@ -292,7 +292,12 @@ func Contains(a []string, x string) bool {
 // createPacketManagerRest sets up the client and returns
 // an packetManagerRest.
 func createPacketManagerRest(configReader io.Reader, discoverOpts cloudprovider.NodeGroupDiscoveryOptions, opts config.AutoscalingOptions) (*packetManagerRest, error) {
-	var cfg ConfigFile
+	// Initialize ConfigFile instance
+	cfg := ConfigFile{
+		DefaultNodegroupdef: ConfigNodepool{},
+		Nodegroupdef:        map[string]*ConfigNodepool{},
+	}
+
 	if configReader != nil {
 		if err := gcfg.ReadInto(&cfg, configReader); err != nil {
 			klog.Errorf("Couldn't read config: %v", err)
--- a/cluster-autoscaler/cloudprovider/packet/packet_manager_rest.go
+++ b/cluster-autoscaler/cloudprovider/packet/packet_manager_rest.go
@@ -48,6 +48,8 @@ import (
 const (
 	userAgent                    = "kubernetes/cluster-autoscaler/" + version.ClusterAutoscalerVersion
 	expectedAPIContentTypePrefix = "application/json"
+	packetPrefix                 = "packet://"
+	equinixMetalPrefix           = "equinixmetal://"
 )
 
 type instanceType struct {
@@ -436,7 +438,11 @@ func (mgr *packetManagerRest) NodeGroupForNode(labels map[string]string, nodeId 
 	if nodegroup, ok := labels["pool"]; ok {
 		return nodegroup, nil
 	}
-	device, err := mgr.getPacketDevice(context.TODO(), strings.TrimPrefix(nodeId, "packet://"))
+
+	trimmedNodeId := strings.TrimPrefix(nodeId, packetPrefix)
+	trimmedNodeId = strings.TrimPrefix(trimmedNodeId, equinixMetalPrefix)
+
+	device, err := mgr.getPacketDevice(context.TODO(), trimmedNodeId)
 	if err != nil {
 		return "", fmt.Errorf("Could not find group for node: %s %s", nodeId, err)
 	}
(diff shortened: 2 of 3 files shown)


cluster-autoscaler/cloudprovider/packet/packet_manager_rest.go
```diff
@@ -592,7 +603,7 @@ func (mgr *packetManagerRest) getNodes(nodegroup string) ([]string, error) {
 
 	for _, d := range devices.Devices {
 		if Contains(d.Tags, "k8s-cluster-"+mgr.getNodePoolDefinition(nodegroup).clusterName) && Contains(d.Tags, "k8s-nodepool-"+nodegroup) {
-			nodes = append(nodes, fmt.Sprintf("packet://%s", d.ID))
+			nodes = append(nodes, fmt.Sprintf("equinixmetal://%s", d.ID))
```
detiber, line 606: If I'm not mistaken, wouldn't this cause the autoscaler to think that any Nodes w/ the 'packet://' provider id prefix to be unregistered (and as a result, eventually deleted)?

https://github.com/kubernetes/autoscaler/blob/c64e97570a2f06dabe685e6f0387495431466217/cluster-autoscaler/clusterstate/clusterstate.go#L295-L329

I've hit similar complications around the Node provider IDs with cluster-api-provider-packet around migration from an older version of the packet-ccm to a newer version of cloud-provider-equinix-metal, and the migration process can be a bit messy, especially since the ProviderID field on the Node is immutable.

There are probably a couple of options available:
- Switch everything to use the new `equinixmetal:;//` provider id prefix and require using a newer version of the cloud-provider
- Make the prefix configurable, so that it can work with either prefix, but the user must configure it correctly

Either way, we should probably document some type of migration strategy, since I'm not sure there is a way to stay flexible without changes to the core autoscaling code, or forcing node/device churn in the cluster.

For cluster-api-provider-packet, the current plan is to do something along the lines of:
- Update packet-ccm to cloud-provider-equinix-metal
- Upgrade the cluster api provider to a version that uses the new provider id prefix
- Delete and Recreate the Node resources to change the providerID.

A similar approach should be possible with the autoscaler, thought it might be better to disable the autoscaler during the upgrade process to avoid any potential of confusion that could cause workload disruption during the node deletion/recreation process.

cluster-autoscaler/cloudprovider/packet/packet_manager_rest.go
```diff
@@ -293,6 +293,13 @@ func Contains(a []string, x string) bool {
 // an packetManagerRest.
 func createPacketManagerRest(configReader io.Reader, discoverOpts cloudprovider.NodeGroupDiscoveryOptions, opts config.AutoscalingOptions) (*packetManagerRest, error) {
 	var cfg ConfigFile
+
+	// Initialize ConfigFile instance
+	cfg = ConfigFile{
+		DefaultNodegroupdef: ConfigNodepool{},
+		Nodegroupdef:        map[string]*ConfigNodepool{},
+	}
```
invidian, line 301: Code can be simplified:
```suggestion
	cfg := ConfigFile{
		DefaultNodegroupdef: ConfigNodepool{},
		Nodegroupdef:        map[string]*ConfigNodepool{},
	}
```

cluster-autoscaler/cloudprovider/packet/packet_cloud_provider.go
```diff
@@ -37,6 +37,11 @@ const (
 	ProviderName = "packet"
 	// GPULabel is the label added to nodes with GPU resource.
 	GPULabel = "cloud.google.com/gke-accelerator"
+	// DefaultControllerNodeLabelKey is the label added to Master/Controller to identify as
+	// master/controller node.
+	DefaultControllerNodeLabelKey = "node-role.kubernetes.io/master"
+	// ControllerNodeIdentifierEnv is the string for the environment variable.
+	ControllerNodeIdentifierEnv = "CONTROLLER_NODE_IDENTIFIER_LABEL"
```
invidian, line 44: This is Packet-specific, so maybe reflect that in env variable name?
```suggestion
	ControllerNodeIdentifierEnv = "PACKET_CONTROLLER_NODE_IDENTIFIER_LABEL"
```

cluster-autoscaler/cloudprovider/packet/packet_manager_rest.go
invidian: Good point @detiber. Thanks for the feedback :)

@ipochi as far as I understand, it is not possible to get mixed `packet://` and `equnixmetal://` on a single cluster, right? Those are set by the CCM, which will either set it to one or another, or? If that's the case, we should indeed make it configurable.

ipochi: @detiber Thanks for your valuable feedback. I think I misunderstood about the importance of `packet://` in
```nodes = append(nodes, fmt.Sprintf("packet://%s", d.ID))
```

Also @invidian @detiber 

Its is quite possible to get mixed `packet://` and `equinixmetal://` on a single cluster. If ta cluster is created with packet-ccm and then packet-ccm is upgraded to cloud-provider-equinix-metal, then any new nodes created will have `equinixmetal://`, whilst the old nodes would still continue to have `packet://`

Example:
```
$ kubectl get nodes
NAME                         STATUS   ROLES    AGE     VERSION
theta-anotherpool-worker-0   Ready    <none>   9m29s   v1.21.1 - After upgrade to cloud-provider-equinix-metal
theta-controller-0           Ready    <none>   86m     v1.21.1 - Before upgrade
theta-small-worker-0         Ready    <none>   86m     v1.21.1 - Before upgrade
```

```
$ kubectl get nodes -o jsonpath='{.items[*].spec.providerID}'
equinixmetal://536e117a-6d6e-4d9c-b502-f67170662aef packet://b7057a83-732e-4ae6-b5a7-d28516556ba6 packet://4815edc6-b0e2-4bdb-bbb0-daca73e72c1a⏎
```

invidian: >Its is quite possible to get mixed packet:// and equinixmetal:// on a single cluster. If ta cluster is created with packet-ccm and then packet-ccm is upgraded to cloud-provider-equinix-metal, then any new nodes created will have equinixmetal://, whilst the old nodes would still continue to have packet://

Yes, but this is the situation which can be gated. You can disable the autoscaler, do CCM -> CCEM upgrade, re-register all nodes, then upgrade the autoscaler I guess.

ipochi: The scenario I described above didn't involve autoscaler.

Merely upgrading from CCM TO CCEM and adding a node to the cluster.


Can deregistering and  registering a node be done in case of a single controller node ?

invidian: >The scenario I described above didn't involve autoscaler.

> Merely upgrading from CCM TO CCEM and adding a node to the cluster.

It doesn't matter who adds new node :) But the autoscaler can do it unattended. 

>Can deregistering and registering a node be done in case of a single controller node ?

That seem Lokomotive specific question and I don't know the answer :smile:

ipochi: Q: We should't worry about controller nodes if I understand correctly ?

invidian: Assuming they will be correctly skipped, yes. Good point!

ipochi:
@invidian addressed your feedback.

detiber:
/lgtm
/approve

Since the autoscaler is versioned in lock step with kubernetes minor versions, we'll need to also backport this to the appropriate release branches as well.

cluster-autoscaler/cloudprovider/packet/packet_manager_rest.go
```diff
@@ -595,9 +601,30 @@ func (mgr *packetManagerRest) getNodes(nodegroup string) ([]string, error) {
...
+	// `equinixmetal://`. This is now needed as `packet-ccm` is now deprecated and renamed in favor of
+	// `cloud-provider-equinix-metal`.
+	// This code checks if the INSTALLED_CCM env var is set or not. If set to `cloud-provider-equinix-metal`,
+	// the prefix is set to `equinixmetal://` and any other case the prefix is `packet://`.
+	// At a later point in time, there would be a need to make `equinixmetal://` prefix as the default or do away
+	// with `packet://` prefix entirely. This should happen presumably when the packet code in autoscaler is
+	// renamed from packet to equinixmetal.
+	prefix := packetPrefix
+
+	switch installedCCM := os.Getenv("INSTALLED_CCM"); installedCCM {
```
invidian, line 616: Ditto PACKET_ prefix + env variable name could be a constant :)

invidian:
@ipochi can branch be removed now?

ipochi:
> /lgtm
> /approve
> 
> Since the autoscaler is versioned in lock step with kubernetes minor versions, we'll need to also backport this to the appropriate release branches as well.

@detiber 

Can you please provide guidance on how to backport this to the release branches and till which autoscaler release ?

detiber:
> @detiber
> 
> Can you please provide guidance on how to backport this to the release branches and till which autoscaler release ?

You should be able to cherry-pick the commit and create a PR against the `cluster-autoscaler-release-*` branches that are appropriate.

I don't see a need to backport to anything that isn't currently supported, so I think 1.19, 1.20, 1.21, and 1.22 are probably the appropriate branches we'd want to backport to.

You might be able to leverage the "cherry-pick script" from the kubernetes release team: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-release/cherry-picks.md, but since I don't do it frequently, I generally just do it manually:

```sh
git checkout -b backport-1.22 upstream/cluster-autoscaler-release-1.22 # assuming the upstream k/autoscaler repo is configured as the remote 'upstream'
git cherry-pick <sha for commit 1> <sha for commit 2> ...
```
And repeating for the other branches.

https://github.com/kubernetes/autoscaler/pull/4260 is an example of a manual cherry-pick PR
https://github.com/kubernetes/autoscaler/pull/4269 is an example of an automated cherry-pick PR

PR #4295, opened 2021-08-30
azure: support CLI credentials instead of service principal

pohly:
During development it is simpler to reuse the existing credentials. Also, some
setups with Active Directory make it hard to create a service principal because
that depends on additional privileges.

/cc @bpineau


bpineau:
(not an Azure maintainer here)
Thanks, that's a sweet improvement!

Perhaps we could consider `s/UseCLI bool/AuthMethod string/g`, anticipating support for other methods (without ending up with a maze of mutualy exclusive flags)?

For instance MSI (Azure equivalent for IMDS/instance profiles, ie. auth using instance's metadata server) would be a straightforward addition to your improved `newAuthorizerFromCredentials`, probably just (untested):

```go
if config.AuthMethod == "cli" {
        return auth.NewAuthorizerFromCLI()
}

if config.AuthMethod == "msi" {
        config := auth.NewMSIConfig()
        config.Resource = env.ResourceManagerEndpoint
        return config.Authorizer()
}
```

What do you think?

cluster-autoscaler/cloudprovider/azure/azure_client.go
```diff
@@ -206,13 +207,24 @@ func newServicePrincipalTokenFromCredentials(config *Config, env *azure.Environm
 	return nil, fmt.Errorf("no credentials provided for AAD application %s", config.AADClientID)
 }
 
+func newAuthorizerFromCredentials(config *Config, env *azure.Environment) (autorest.Authorizer, error) {
```
bpineau, line 210: nit (naming): with CLI support, that's not necessarily `FromCredentials` anymore right?

pohly:
> Perhaps we could consider s/UseCLI bool/AuthMethod string/g, anticipating support for other methods (without ending up with a maze of mutually exclusive flags)?

Makes sense. I wasn't sure whether there are others. I can change that once it has been determined that this PR can be merged despite the additional dependencies (small, but still not something that would be needed for "normal" deployments).

pohly:
/cc @feiskyer @nilo19 @marwanad

Is it okay to pull in additional Azure credential providers into cluster-autoscaler?

cluster-autoscaler/cloudprovider/azure/azure_client.go
pohly: I removed `FromCredentials`.

pohly:
> > Perhaps we could consider s/UseCLI bool/AuthMethod string/g, anticipating support for other methods (without ending up with a maze of mutually exclusive flags)?
> 
> Makes sense. I wasn't sure whether there are others.

I've changed it so that (for now) "cli" and "principal" are valid authentication methods, with "principal" being the default when nothing is set (i.e. the current behavior).

marwanad:
@pohly yup the PR adding the functionality is reasonable. I'll take a look at the code tomorrow. Sorry for the delay!

pohly:
/assign @towca

For final approval.

feiskyer (approved):
/lgtm
/approve

PR #4299, opened 2021-08-30
Cluster-Autoscaler - Update hardcoded AWS instance list

gjtempleton:
Bring AWS fallback hardcoded list of instances up to date to 30/08/21

* Brings in new [m6i](https://aws.amazon.com/ec2/instance-types/m6i/) and [x2gd](https://aws.amazon.com/ec2/instance-types/x2/) instance family data as well as other assorted new sizes in existing families

PR #4300, opened 2021-08-31
Update AWS SDK to support m6i instances

der-eismann:
We want to make use of the new AWS m6i instances, but unfortunately cluster-autoscaler doesn't support them yet.


der-eismann:
Didn't see #4299, also it seems the SDK update is not needed since the info is pulled from an AWS endpoint.

PR #4302, opened 2021-09-01
Rate limit garbage collection inside `ClusterState`

jbartosik:
It looks like it belongs inside the class, not with code using it.


jbartosik:
@krzysied @mwielgus PTAL

kgolab:
Thanks, LGTM.

PR #4303, opened 2021-09-01
[CA] [AWS examples] Add priorityClassName & securityContext & upgrade image

damienleger:
Hello, some improvements on AWS examples.

To ensure the pod is scheduled
```
priorityClassName: system-cluster-critical
```

To run pod as non-root (security best practice)
```
securityContext:
  runAsNonRoot: true
  runAsUser: 65534
```

And use the last image (1.22.0).

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
+++ b/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
@@ -142,7 +142,7 @@ spec:
         runAsUser: 65534
       serviceAccountName: cluster-autoscaler
       containers:
-        - image: k8s.gcr.io/autoscaling/cluster-autoscaler:v1.22.0
+        - image: k8s.gcr.io/autoscaling/cluster-autoscaler:v1.21.0
           name: cluster-autoscaler
           resources:
             limits:
--- a/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
+++ b/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
@@ -136,9 +136,13 @@ spec:
         prometheus.io/scrape: 'true'
         prometheus.io/port: '8085'
     spec:
+      priorityClassName: system-cluster-critical
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
       serviceAccountName: cluster-autoscaler
       containers:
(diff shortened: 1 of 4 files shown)


damienleger:
I signed it

cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
```diff
@@ -136,9 +136,13 @@ spec:
...
         prometheus.io/port: '8085'
     spec:
+      priorityClassName: system-cluster-critical
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
       serviceAccountName: cluster-autoscaler
       containers:
-        - image: k8s.gcr.io/autoscaling/cluster-autoscaler:v1.17.3
+        - image: k8s.gcr.io/autoscaling/cluster-autoscaler:v1.22.0
```
gjtempleton, line 145: As this is [linked to](https://docs.aws.amazon.com/eks/latest/userguide/cluster-autoscaler.html#ca-deploy) by the AWS EKS docs, can we instead bump this to the latest patch version of 1.21 instead as that's the newest minor version currently supported by AWS.

You've also bumped this in 3/4 of the examples, can we bump it in the last of the examples as well to keep things consistent.
```suggestion
        - image: k8s.gcr.io/autoscaling/cluster-autoscaler:v1.21.0
```

gjtempleton (commented):
Hey, thanks for the PR.

One quick comment for you for now.

I also want to double-check all the examples still work as intended with the new security settings and won't be able to do that until next week.

cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
damienleger: Fixed 🙂

damienleger:
Hello @gjtempleton, thanks for reply

> I also want to double-check all the examples still work as intended with the new security settings and won't be able to do that until next week.

The modified `cluster-autoscaler-autodiscover.yaml` with 1.22.0 (*) image, priorityClassName, and securityContext runs in my 1.21 EKS. I haven't tested the other.

(*) I use the 1.22 image in EKS 1.21 since there is [no API deprecated in 1.22 used by cluster-autoscaler](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22) but I agree with you, it's simpler to stick with the 1:1 version mapping with EKS and cluster-autoscaler.

gjtempleton:
Thanks for that.

/lgtm
/approve

PR #4309, opened 2021-09-07
Use in-place slice deletion for unregister and add a test

suzaku:
Use in-place slice deletion for `unregister` and add a test to avoid allocating new slices.

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
+++ b/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
@@ -115,7 +115,7 @@ func (m *asgCache) register(asg *asg) *asg {
 			klog.V(4).Infof("Updating ASG %s", asg.AwsRef.Name)
 
 			// Explicit registered groups should always use the manually provided min/max
-			// values and the not the ones returned by the API
+			// values and not the ones returned by the API
 			if !m.explicitlyConfigured[asg.AwsRef] {
 				existing.minSize = asg.minSize
 				existing.maxSize = asg.maxSize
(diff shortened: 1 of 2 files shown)


suzaku:
I've signed up CLA with my GitHub primary email, but still, the CLA check is failing, what might be wrong here?

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
```diff
@@ -141,17 +141,21 @@ func (m *asgCache) register(asg *asg) *asg {
...
 	var changed *asg
+	var i int
 	for _, existing := range m.registeredAsgs {
 		if existing.AwsRef == a.AwsRef {
 			klog.V(1).Infof("Unregistered ASG %s", a.AwsRef.Name)
 			changed = a
 			continue
 		}
-		updated = append(updated, existing)
+		m.registeredAsgs[i] = existing
```
bpineau, line 152: I might be mistaken but I assumed the previous version was copying/replacing (but not updating in place) because the cluster-autoscaler don't lock-protect that cache's content (slice items).

regenerate(), register() and unregister() operates under m.mutex.Lock and regenerates a new m.registeredAsgs (the containing slice), while other threads might have obtained an older version via asgCache.Get() and still access a stale/previous m.registeredAsgs versions (which should be gc'ed later).

If that's the case then updating in place wouldn't be safe; what do you think?

suzaku: Thanks for reviewing.

Yes, you are right, although `asgCache` is guarded by `mutex.Lock`, the returned `registeredAsgs` is out of the lock's control. So I can't just reuse the slice here.

PR #4313, opened 2021-09-07
reimplement linode autoscaler to use official API bindings

Charliekenney23:
These changes reimplement the Linode CAS implementation to allow autoscaling individual LKE NodePools. NodePool min/max sizes can now be configured via the Linode API.

Previously, the CAS could only be configured by setting min/max nodes per Linode instance types in a global `.ini` file.

The `cloud-config` is now an optional (all required fields can instead be specified via environment variables) JSON file.


Charliekenney23:
I signed it

dbonfigli:
I am not an official reviewer (I only created the first version of this provider implementation), I added a couple of comments.

The most important thing that I see is the use of specific Linode APIs to configure pools (enable/disable scaling, min/max size) when before it was done via a config file: please clearly document this change, and whether Linode will indeed host the cluster autoscaler, so that current users of this implementation can plan for it.

dbonfigli (commented):
I am not an official reviewer (I only created the first version of this provider implementation), I added a couple of comments.

The most important thing that I see is the use of specific Linode APIs to configure pools (enable/disable scaling, min/max size) when before it was done via a config file: please document this change more clearly, and whether Linode will indeed host the cluster autoscaler, so that current users of this implementation can plan for it.

mwielgus:
Comments not addressed, closing due to inactivity.

PR #4317, opened 2021-09-08
Update to take into account some doc changes

MarkSRobinson:
Take into account some changes from https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-does-cluster-autoscaler-work-with-pod-priority-and-preemption.  The overprovision should be set to -10 since that's the minimum that will trigger the auto-scaler


mwielgus:
CLA still not signed, closing.

PR #4326, opened 2021-09-13
Processor policy

MaciekPytel:
Start a README file for processor documenting the deprecation policy,
as well as existing code review practices.
The deprecation policy has been discussed in
#4191 (comment) and
sig-meeting, where it gained an approval from both top-level OWNERS and
the sig in general.
Add deprecation warnings for NodeInfoProcessor.

### DIFF
--- a/cluster-autoscaler/FAQ.md
+++ b/cluster-autoscaler/FAQ.md
@@ -55,6 +55,7 @@ this document:
   * [What events are emitted by CA?](#what-events-are-emitted-by-ca)
   * [What happens in scale-up when I have no more quota in the cloud provider?](#what-happens-in-scale-up-when-i-have-no-more-quota-in-the-cloud-provider)
 * [Developer](#developer)
+  * [How should I implement new features?](#how-should-i-implement-new-features)
   * [How can I run e2e tests?](#how-can-i-run-e2e-tests)
   * [How should I test my code before submitting PR?](#how-should-i-test-my-code-before-submitting-pr)
   * [How can I update CA dependencies (particularly k8s.io/kubernetes)?](#how-can-i-update-ca-dependencies-particularly-k8siokubernetes)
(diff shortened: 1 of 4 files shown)


MaciekPytel:
cc: @mwielgus @gjtempleton

cluster-autoscaler/processors/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+Processors are the *preferred way* for introducing new features to core CA
+logic. If a feature can be implemented via an existing processor interface, the
+reviewers may ask for a PR to be implemented this way.
+
+## Changes and deprecations
+
+Adding new processors, adding new methods to existing processors or adding
+additional parameters to existing methods may all happen in minor releases of
+Cluster Autoscaler.
```
bpineau, line 21: Open question: should we also name the cloudprovider interface here (ie. can that interface follow the same evolution process - and if not, should this FAQ document it)?

cluster-autoscaler/FAQ.md
```diff
@@ -898,6 +899,15 @@ Depending on how long scale-ups have been failing, it may wait up to 30 minutes
 
 # Developer:
 
+### How should I implement new features?
+
+The preferred way of implementing new features is via [processor
+interfaces](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/processors).
+Before opening a PR with changes (and especially significant changes) to core CA
+logic please check if those changes can instead be encapsulated behind on of the
```
bpineau, line 907: behind one of

cluster-autoscaler/processors/README.md
MaciekPytel: That is a very good question. I think it is how we've been treating cloudprovider historically (at least the addition part - ex. introduction of GetOptions). It could also apply to other interfaces we expect people to commonly implement in forks (expander?).

@mwielgus @gjtempleton Any opinions on this one? I'm happy to move this to a top-level deprecation policy doc that covers all of the above.

PR #4327, opened 2021-09-13
Update k8s dependencies to 1.23.0-alpha.1 in CA.

olagacek:
On top of that update gce client tests in line with new GCE client API


cluster-autoscaler/version/version.go
```diff
@@ -17,4 +17,4 @@ limitations under the License.
 package version
 
 // ClusterAutoscalerVersion contains version of CA.
-const ClusterAutoscalerVersion = "1.23.0-alpha.0"
```
MaciekPytel, line 20: Actually this is a version of Cluster Autoscaler, not the kubernetes vendored in. CA matches the minor release number with k8s, but not patch release numbers and since we're not releasing CA we should not be bumping this.

MaciekPytel:
/approve

Thanks for picking this up. I left a single comment, after addressing it this lgtm.

cluster-autoscaler/version/version.go
olagacek: Done, rollbacked this file.

PR #4328, opened 2021-09-14
Use custom spam filtering function in event recorder.

olagacek:


olagacek:
assign @MaciekPytel

cluster-autoscaler/utils/kubernetes/factory.go
```diff
@@ -57,5 +59,26 @@ func getCorrelationOptions() kube_record.CorrelatorOptions {
 		QPS:          defaultQPS,
 		BurstSize:    defaultBurstSize,
 		LRUCacheSize: defaultLRUCache,
+		SpamKeyFunc:  getCustomSpamKeyFunc(),
+	}
+}
+
+func getCustomSpamKeyFunc() kube_record.EventSpamKeyFunc {
```
MaciekPytel, line 66: Can you add a comment describing what this does and why? I know it, but without having all the context beforehand I think this would be very hard to understand.

olagacek: Added a comment with description

PR #4331, opened 2021-09-15
gce: refresh instances templates in background

bpineau:
This is an alternate (maybe less intrusive) approach to PR #3862 : avoid blocking head of line when instance templates cache expires (all at once).

The GCE instance template's Get API is not fast (on my projects, it takes about 400ms to get a single template), and for each template we also need to get an instance group manager from the API (also a couple hundred milliseconds).

Due to that, even a cluster with just a dozen attached ITs experiences latencies at each cache expiration deadline. That latency grows linearly with the number of instance templates.

Contrary to the other PR, this runs the refreshes entirely in background, but doesn't spread refreshes over cache TTL to be kind with GCE API; I can rework to combine both approaches if that's preferred.


bpineau:
Obsoleted by #4423 , closing.

PR #4308, opened 2021-09-06
Update VPA e2e dependencies

jbartosik:
To Kubernetes 1.20.4

PR #4329, opened 2021-09-14
Adjust KEP for customized recommender

jbartosik:
Use a list of structs instead of a list of strings to choose
recommenders.

This will allow us to pass parameters to recommenders (if we decide we
want to allow that).

Related to #3913

PR #4341, opened 2021-09-17
Fix cluster feeder test

jbartosik:
I didn't update one call to `model.NewClusterState` in #4308. Fix it.

I wonder why presubmit didn't complain on the PR which introduced the problem


krzysied:
/lgtm
IMHO we want to verify if presubmits are actually running the unit tests..

jbartosik:
I saw this test failing to compile on a different PR ([relevant test run](https://github.com/kubernetes/autoscaler/pull/4329/checks?check_run_id=3633580821)).

So at least sometimes they do. I'll try to look into how this happened

PR #4354, opened 2021-09-27
fix expanders unit tests

bpineau:
The github action hooks didn't run the unit tests on the expanders
update in #4233 (likely the known "tests don't run automatically for
new contributor" limitation).

NewStrategy became NewFilter, with a BestOptions (plural) method,
and returns an Option slice instead of a single Option.


x13n:
/lgtm

Can we do something about the "tests don't run automatically for new contributor" issue? Even if we don't want to run tests for every PR from new people, merging without testing shouldn't be allowed.

PR #4357, opened 2021-09-28
remove nodes stuck creating even if this takes us below nodeGroup minSize

### LINKED ISSUE #4351: Cluster Autoscaler should remove oldUnregistered placeholder nodes, even if this would go below minSize
cluster autoscaler

1.18.3, but the issue is still present on `master`

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.0", GitCommit:"c2b5237ccd9c0f1d600d3072634ca66cefdf272f", GitTreeState:"clean", BuildDate:"2021-08-04T17:56:19Z", GoVersion:"go1.16.6", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.15-airbnb0", GitCommit:"7e9eabf6e12c4c4c2cffdc72802ed41d75bb4010", GitTreeState:"clean", BuildDate:"2021-01-27T19:18:08Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
WARNING: version difference between client (1.22) and server (1.18) exceeds the supported minor version skew of +/-1
</pre></details>

AWS

When nodes aren't available for a particular nodeGroup, and the node group is below its minimum size, the node group should be placed into the backoff queue.

**What happened instead?**:
If the node group is below its minimum size and the cloud provider doesn't have capacity, the oldUnregistered nodes will all be placeholder nodes.  Cluster Autoscaler will [refuse to delete these nodes](
https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/static_autoscaler.go#L622), which means that the `ScaleUp` object for that nodeGroup will not get deleted (since a `ScaleUp` object only gets deleted when the target capacity <= actual capacity).  This means that in the future, if more load appears, Cluster Autoscaler will try to edit the existing `ScaleUp` object, which will reset its expiration time.  This leads to a self-perpetuating cycle where the node group never gets added to the backoff list, load continues to build up on the cluster, and Cluster Autoscaler will never scale up (if, for example, you're using the priority expander and the unavailable nodeGroup is the highest priority).

1. Use the priority expander, and set the highest-priority node group to an ASG for which the cloud provider has no capacity.
2. Set the minimum size for the node group to be greater than its current size.
3. Every five minutes or so, add load to the cluster by scaling up some deployments
4. Observe that the cluster never scales up and the chosen node group never goes into backoff.

N/A

drmorr0:
This PR is an attempt to resolve #4351.  The AWS cloudprovider doesn't currently set the `InstanceStatus` field on `cloudprovider.Instance`, so in this change we start populating this field.  We set it to `InstanceRunning` except when the node is a placeholder node, in which case we set it to `InstanceCreating`.  Then, in the `removeOldUnregisteredNodes` function, if removing the node would take us below the node group's `minSize`, but the status is "creating", we go ahead and remove the node anyways.  The logic here is that if the node is stuck in creating for 15 minutes (or whatever the max node provision time is), we're _already_ below the node group's minimum size, but CA just doesn't think it is.

I've added a case to the `TestStaticAutoscalerRunOnceWithLongUnregisteredNodes` test case to confirm that this works as expected.

I _think_ this is a relatively simple/safe change.  My only slight concern is that I'm not sure how other cloud providers that use the `InstanceStatus` field will respond to this change.  It [looks like](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/kubernetes/autoscaler%24+file:%5Ecluster-autoscaler/cloudprovider/.*/.*+InstanceCreating&patternType=literal) the other cloud providers that use the `InstanceCreating` value are:

* Azure
* Bizfly
* DigitalOcean
* Exoscale
* GCE
* Hetzner
* Huawei
* Ionos
* Magnum
* OVH

I don't know a good way to test the behaviour of these other cloud providers.  :\

The other thing I was unsure about was what to do in the `ClusterStateRegistry` when an instance changes state.  My intuition was that if a node transitions from (say) "creating" to "running", that it would make sense to reset the node's "unregistered-since" timer, but since we currently don't track the state transitions at all, this would be a change from the existing behaviour that might be unexpected.

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
+++ b/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
@@ -31,8 +31,7 @@ import (
 )
 
 const (
-	scaleToZeroSupported          = true
-	placeholderInstanceNamePrefix = "i-placeholder"
+	scaleToZeroSupported = true
 )
 
 type asgCache struct {
@@ -264,7 +263,7 @@ func (m *asgCache) DeleteInstances(instances []*AwsInstanceRef) error {
 	for _, instance := range instances {
 		// check if the instance is a placeholder - a requested instance that was never created by the node group
 		// if it is, just decrease the size of the node group, as there's no specific instance we can remove
-		if m.isPlaceholderInstance(instance) {
+		if instance.isPlaceholder() {
 			klog.V(4).Infof("instance %s is detected as a placeholder, decreasing ASG requested size instead "+
 				"of deleting instance", instance.Name)
 			m.decreaseAsgSizeByOneNoLock(commonAsg)
@@ -286,11 +285,6 @@ func (m *asgCache) DeleteInstances(instances []*AwsInstanceRef) error {
 		}
 	}
 	return nil
-}
-
-// isPlaceholderInstance checks if the given instance is only a placeholder
-func (m *asgCache) isPlaceholderInstance(instance *AwsInstanceRef) bool {
-	return strings.HasPrefix(instance.Name, placeholderInstanceNamePrefix)
 }
 
 // Fetch automatically discovered ASGs. These ASGs should be unregistered if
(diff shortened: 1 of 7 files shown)


drmorr0:
This is superseded by #4489

PR #4392, opened 2021-10-12
Extend ScaleUpStatus with node groups that failed scale up.

olagacek:
Add information about Node Groups that failed creation and/or resize to ScaleUpStatus structure.

### DIFF
--- a/cluster-autoscaler/core/scale_up.go
+++ b/cluster-autoscaler/core/scale_up.go
@@ -489,7 +489,9 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 			oldId := bestOption.NodeGroup.Id()
 			createNodeGroupResult, err := processors.NodeGroupManager.CreateNodeGroup(context, bestOption.NodeGroup)
 			if err != nil {
-				return scaleUpError(&status.ScaleUpStatus{}, err)
+				return scaleUpError(
+					&status.ScaleUpStatus{FailedCreationNodeGroups: []cloudprovider.NodeGroup{bestOption.NodeGroup}},
+					err)
 			}
 			createNodeGroupResults = append(createNodeGroupResults, createNodeGroupResult)
 			bestOption.NodeGroup = createNodeGroupResult.MainCreatedNodeGroup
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/core/scale_up.go
```diff
@@ -734,8 +741,10 @@ func applyScaleUpResourcesLimits(
 	return newNodes, nil
 }
 
-func scaleUpError(s *status.ScaleUpStatus, err errors.AutoscalerError) (*status.ScaleUpStatus, errors.AutoscalerError) {
+func scaleUpError(s *status.ScaleUpStatus, err errors.AutoscalerError, failedCreationNodeGroup cloudprovider.NodeGroup, failedResizeNodeGroup cloudprovider.NodeGroup) (*status.ScaleUpStatus, errors.AutoscalerError) {
```
towca, line 744: Do you know why we have this function? It would made sense if it was for adding an error to an existing status object, but it looks like all uses actually create a new status object right in the invocation. Wouldn't it be easier to just directly set all the fields and get rid of the function?

cluster-autoscaler/core/scale_up.go
```diff
@@ -581,13 +588,13 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
...
 		if typedErr != nil {
-			return scaleUpError(&status.ScaleUpStatus{CreateNodeGroupResults: createNodeGroupResults}, typedErr)
+			return scaleUpError(&status.ScaleUpStatus{CreateNodeGroupResults: createNodeGroupResults}, typedErr, nil, bestOption.NodeGroup)
 		}
 		klog.V(1).Infof("Final scale-up plan: %v", scaleUpInfos)
 		for _, info := range scaleUpInfos {
 			typedErr := executeScaleUp(context, clusterStateRegistry, info, gpu.GetGpuTypeForMetrics(gpuLabel, availableGPUTypes, nodeInfo.Node(), nil), now)
 			if typedErr != nil {
-				return scaleUpError(&status.ScaleUpStatus{CreateNodeGroupResults: createNodeGroupResults}, typedErr)
+				return scaleUpError(&status.ScaleUpStatus{CreateNodeGroupResults: createNodeGroupResults}, typedErr, nil, info.Group)
```
towca, line 597: This is the only place where actual resize happens. I'd either only set `FailedResizeNodeGroups` here, or change the name somehow. If you're going for "this should be set for every error after `bestOption` is known and exists", maybe it'd be enough to add `bestOption` to the status and use ScaleUpError/Result to detect cases where it failed?

cluster-autoscaler/core/scale_up.go
towca: Ok, I see. If we want to keep the function, I'd rather keep just the error as an argument (since the repetition is in setting `s.Result = status.ScaleUpError`), and set everything else directly on the status object passed to the function. This way we don't have the `nil, nil` added everywhere where we're not setting these fields, and we don't have to change the function's signature every time we want to set a new field in just one place. WDYT?

PR #4398, opened 2021-10-14
[AWS EKS - Scale-to-0] Update conditional to check cluster-name as well

MyannaHarris:
This change updates the conditional to check for the cluster-name label as well since we need both for the DescribeNodegroup API call and a customer can accidentally delete either.


MyannaHarris:
Related to https://github.com/kubernetes/autoscaler/pull/3968

PR #4411, opened 2021-10-20
CA - AWS - Update StaticListLastUpdateTime on re-generating instance …

gjtempleton:
…list

Moves the generation of the `lastUpdateTime` string used to warn users of the last time the static instance list was updated to be generated by the same code as that which generates the instance list data. Currently, the fact that this is a separate string in a separate file makes it easy to forget to update this when running `make generate` to update the instance data.


jaypipes (approved):
Nice, @gjtempleton! Much better than a static last updated string!

PR #4417, opened 2021-10-22
Upgrade image tag version for helm chart

Sebor:
Because version 1.21.1 has been [released](https://github.com/kubernetes/autoscaler/releases/tag/cluster-autoscaler-1.21.1) it would be nice to update helm chart


gjtempleton:
Thanks!
/lgtm
/approve

PR #4423, opened 2021-10-26
Refactor MIG info providers in GCE cloud provider

BigDarkClown:
### Code cleanup

Managing MIG related information kept in `GceCache` had been previously split between `MigInstanceTemplateProvider`, `MigTargetSizesProvider` and `GceManager`. This entire logic has been moved to `MigInfoProvider`, which now works as a middle-man between `GceCache` and `AutoscalingGceClient`. This interface is  responsible for keeping MIG target size, basename and instance template up to date. 

### Instance template caching

In previous implementation instance templates were updated when a) Cluster Autoscaler noticed a new MIG b) templates have not been updated for 30 minutes. These periodic updates can cause latency spikes and throttling, especially if the number of MIGs is significant. This has been described in #3862 in more detail. 

Another issue is that, while instance templates are immutable resources, it is still possible to assign different instance template to the MIG. Current logic does not handle that well - Cluster Autoscaler can take up to 30 minutes to notice that a new template was assigned to MIG, resulting in incorrect scale up decisions for empty MIGs.

Instance templates are immutable, so you cannot modify an existing template - you can only create a new template and then assign it to MIG. Because of that we can assume that if any property in template changes, then it is a new template and the template name maintained in MIG should reflect that. We fetch these template names each loop anyway, so we can refresh templates only if we detect their names changed.

Thanks to that, instead of performing `2 * [number of MIGs]` calls every 30 minutes, we will perform a single call each time we detect a template name change. This limits the number of API calls significantly **and** lets us fetch instance templates as soon as they are changed, solving both described issues.


cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
```diff
@@ -0,0 +1,211 @@
...
+
+	template, err = c.gceClient.FetchMigTemplate(migRef, templateName)
+	if err != nil {
+		return nil, err
+	}
+	c.cache.SetMigInstanceTemplate(migRef, template)
+	return template, nil
+}
+
+func (c *cachingMigInfoProvider) fillMigInfoCache() error {
```
MaciekPytel, line 153: nit: maybe mention in comment this needs to be called with mutex locked?

cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
```diff
@@ -0,0 +1,211 @@
...
+	templateName, err := c.GetMigInstanceTemplateName(migRef)
+	if err != nil {
+		return nil, err
+	}
+
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	template, found := c.cache.GetMigInstanceTemplate(migRef)
+	if found && template.Name == templateName {
```
MaciekPytel, line 141: Maybe it's worth logging if instance template has changed? This should be a very rare event (so no impact on log volume) and it may be quite useful when debugging (ex. see which template was used in particular scale-up).

cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
```diff
@@ -0,0 +1,211 @@
...
+	"path"
+	"sync"
+
+	gce "google.golang.org/api/compute/v1"
+	"k8s.io/client-go/util/workqueue"
+	klog "k8s.io/klog/v2"
+)
+
+// MigInfoProvider allows obtaining information about MIGs
+type MigInfoProvider interface {
```
MaciekPytel, line 32: This refactor could be an opportunity to add some unittests here. I think it should be fairly easy given that the only external dependency seems to be AutoscalingGceClient, which is an interface and so could be easily mocked. 

I'm not going to block this PR given that we didn't have tests for previous implementation, but I think tests would be a really nice addition if you're up for it.

MaciekPytel (commented):
This is a very nice refactor, thank you. Left some minor comments, but overall this looks very good.

cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
BigDarkClown: Done

cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
BigDarkClown: Done

cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
BigDarkClown: That's a good point, I'll add a commit with tests to this PR. I'm already immersed in this code so postponing it for later seems like a waste of context.

BigDarkClown: Done

cluster-autoscaler/cloudprovider/gce/mig_info_provider_test.go
```diff
@@ -0,0 +1,519 @@
...
+			name:               "target size from cache fill",
+			cache:              emptyCache(),
+			fetchMigs:          fetchMigsConst([]*gce.InstanceGroupManager{instanceGroupManager}),
+			expectedTargetSize: targetSize,
+		},
+		{
+			name:               "cache fill without mig, fallback success",
+			cache:              emptyCache(),
+			fetchMigs:          fetchMigsConst([]*gce.InstanceGroupManager{}),
+			fetchMigTargetSize: fetchMigTargetSizeConst(targetSize),
```
MaciekPytel, line 144: nit: That feels needlessly clever. It feels like simple targetSize and targetSizeError fields in your struct would accomplish the same thing without the need to rout the interface methods through function references, create closures etc.

I don't think this is worth the effort to refactor and I'm fine merging this as-is. The comment is only meant as something to consider on future PRs.

MaciekPytel (approved):
/lgtm
/approve

PR #4426, opened 2021-10-27
Allow multiStringFlag arguments using a slice

artificial-aidan:
This allows you to do things like

```
extraArgs:
  balancing-ignore-label:
    - label1
    - label2
```

And it will template into

```
--balance-ignore-label=label1
--balance-ignore-label=label2
```

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -17,4 +17,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.10.8
+version: 9.11.0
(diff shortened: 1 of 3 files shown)


gjtempleton:
Hi, thanks for the PR, this is a useful addition.

You'll need to update the chart version as I've just merged a separate PR to the chart.

I think it would also be good to include a commented-out demonstration as in the PR description of how users would use this support in the values file.

artificial-aidan:
> Hi, thanks for the PR, this is a useful addition.
> 
> You'll need to update the chart version as I've just merged a separate PR to the chart.
> 
> I think it would also be good to include a commented-out demonstration as in the PR description of how users would use this support in the values file.

Good idea. I'll add that and bump the version.

artificial-aidan:
I don't think I realized that this was added: https://github.com/artificial-aidan/autoscaler/commit/c8c9a3eea308d93c6796115ebd5d333603fbb33b

I prefer this method, but having duplicate ways to do something could be confusing.

gjtempleton:
I've given this a bit more thought, and whilst I agree with you that this is probably an easier implementation for users to grok, I'd rather avoid duplicating ways to achieve this.

As such, any objection to closing this for now?

artificial-aidan:
No objections.

PR #4443, opened 2021-11-04
[clusterapi] Rely on replica count found in unstructuredScalableResource

codablock:
Instead of retrieving it each time from k8s, which easily causes client-side
throttling, which in turn causes each autoscaler run to take multiple
seconds even if only a small number of NodeGroups is involved and nothing
is to do.

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_unstructured.go
+++ b/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_unstructured.go
@@ -80,19 +80,14 @@ func (r unstructuredScalableResource) ProviderIDs() ([]string, error) {
 }
 
 func (r unstructuredScalableResource) Replicas() (int, error) {
-	gvr, err := r.GroupVersionResource()
+	replicas, found, err := unstructured.NestedInt64(r.unstructured.UnstructuredContent(), "spec", "replicas")
 	if err != nil {
-		return 0, err
+		return 0, errors.Wrap(err, "error getting replica count")
 	}
-
-	s, err := r.controller.managementScaleClient.Scales(r.Namespace()).Get(context.TODO(), gvr.GroupResource(), r.Name(), metav1.GetOptions{})
-	if err != nil {
-		return 0, err
+	if !found {
+		replicas = 0
 	}
-	if s == nil {
-		return 0, fmt.Errorf("unknown %s %s/%s", r.Kind(), r.Namespace(), r.Name())
-	}
-	return int(s.Spec.Replicas), nil
+	return int(replicas), nil
 }
 
 func (r unstructuredScalableResource) SetSize(nreplicas int) error {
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/cloudprovider/clusterapi/clusterapi_unstructured.go
```diff
@@ -119,6 +114,11 @@ func (r unstructuredScalableResource) SetSize(nreplicas int) error {
 
 	s.Spec.Replicas = int32(nreplicas)
 	_, updateErr := r.controller.managementScaleClient.Scales(r.Namespace()).Update(context.TODO(), gvr.GroupResource(), s, metav1.UpdateOptions{})
+
+	if updateErr == nil {
```
jackfrancis, line 118: So I see that we're taking an accounting of the replica count and storing it every time we change the size from within cluster-autoscaler.

What if an out-of-band (i.e., not cluster-autoscaler-originating) replica change occurs. Won't our local replica count be stale in that event?

Lemme know if I'm not understanding everything here.

cc @elmiko @marwanad

codablock: NodeGroups and underlying objects are temporary and re-queries/re-built every time they are needed, e.g. by calling `machineController.nodeGroups()`. This means that the next run of the autoscaler loop will definitely pick up up-to-date objects. Internally, watches/caches/informers are used to keep the local cache up-to-date.

This is based on my understanding after a few hours of debugging cluster-autoscaler. I'm not very familiar with the code-base and also far away from being a full-blown go-coder, so take my analysis/understanding with a grain of salt :)

elmiko:
thanks for the ping @jackfrancis , i'm just starting to take a look at this.

also, thanks for the contribution @codablock =)

elmiko (commented):
i think this change generally makes sense to me, although i found the unit tests a little confusing at first.

i have some similar concerns to what @jackfrancis raised, but i would like to test drive this patch a little to see how it behaves.

codablock:
@elmiko I agree on the confusing nature of the unit test changes...but I couldn't find a better way to make the tests reliable, especially with my limited knowledge about the code base and architecture.

elmiko:
i've been running some local tests with a capi-kubemark cluster and i think this PR is probably safe for us to merge, and also will reduce the calls to the API server as @codablock noted.

here's what i did
1. create a machinedeployment for inclusion in autoscaling, min size 1, max size 5
2. manually increase replicas to 100
3. watch
in this case the autoscaler became very unhappy, it did see that the node group was too large but had trouble scaling it down because of the size differential and the amount of time i was willing to wait. this probably bears more investigation in a scalability sense.

so, i ran a second test with more reasonable limits
1. create a machinedeployment for inclusion in autoscaling, min size 1, max size 5
2. manually increase replicas to 6
3. watch
the autoscaler properly detected the replica size change (on the next node group acquisition), and then brought the replicas size down to 5.

i also tried the same procedure with an unmodified autoscaler and it seems to behave in a very similar manner, albeit with far fewer client side throttling events.

i'm fine to accept this change, thanks for giving us a chance to test it @codablock . i would like to get @jackfrancis to take another look as well.
/lgtm

elmiko:
fwiw, i think we have a bigger issue when it comes to a user changing the replicas outside of the autoscaler. it appears that if the user creates a large enough difference that the autoscaler stops being able to properly set the desired replicas because it is beyond the maximum setting.

for example
```
E1123 21:43:51.183228  644237 scale_down.go:1146] Problem with empty node deletion: failed to delete km-wl-kubemark-md-0-gh8jm: size increase too large - desired:99 max:5
```

hardikdr:
Thanks @codablock for the PR, looks good. 

/lgtm

PR #4451, opened 2021-11-08
expand CAPI_GROUP usage to cover other capi group variables

elmiko:
This change updates the logic for the clusterapi autoscaler provider so
that the `CAPI_GROUP` environment variable will also affect the
annotations keys for minimum and maximum node group size, the machine
annotation, machine deletion, and the cluster name label. It also addes
unit tests and an update to the readme.

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/README.md
+++ b/cluster-autoscaler/cloudprovider/clusterapi/README.md
@@ -121,6 +121,15 @@ some situations, such as testing or prototyping, you may wish to change this
 group variable. For these situations you may use the environment variable
 `CAPI_GROUP` to change the group that the provider will use.
 
+Please note that setting the `CAPI_GROUP` environment variable will also cause the
+annotations for minimum and maximum size to change.
+This behavior will also affect the machine annotation on nodes, the machine deletion annotation,
+and the cluster name label. For example, if `CAPI_GROUP=test.k8s.io`
+then the minimum size annotation key will be `test.k8s.io/cluster-api-autoscaler-node-group-min-size`,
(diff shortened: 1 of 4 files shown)


enxebre:
should this also affect https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_nodegroup.go#L35-L36 and https://github.com/kubernetes/autoscaler/pull/4451/files#diff-8266b9d377a9df07645393a09139352c01636b7c84b6678586245e72c69dfe45R32 ?

enxebre:
slightly related I created https://github.com/kubernetes/autoscaler/issues/4453

elmiko:
> should this also affect https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_nodegroup.go#L35-L36 and https://github.com/kubernetes/autoscaler/pull/4451/files#diff-8266b9d377a9df07645393a09139352c01636b7c84b6678586245e72c69dfe45R32 ?

yeah, i can clean those up as well. i was mainly focused on the min/max annotations, but i think it makes sense to set all of these.

elmiko:
update
* add machine annotation key
* add machine deletion annotation key
* add cluster name label

randomvariable:
this is mostly a lgtm from me, but defer to @enxebre for final lgtm.

enxebre:
@elmiko although changing the group is recommended for testing purposes this is technically a breaking change for the annotations we are now including within the swappable apigroup. Please let's make sure we include a proper release note.

elmiko:
ack, thanks for the heads up @enxebre

PR #4134, opened 2021-06-10
CA expander plugin proposal

evansheng:
Proposal for an expander plugin to allow users to iterate on custom expander strategies out of band with Cluster Autoscaler

Heavily based off of https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/proposals/plugable-provider-grpc.md for a plugable cloud provider.

PR #4452, opened 2021-11-09
Add gRPC expander plugin

evansheng:
Adds a new expander `grpc`, implementing the gRPC plugin outlined in: https://github.com/kubernetes/autoscaler/pull/4134

This PR adds a gRPC client as an expander in CA, and calls out to a gRPC server with config specified by two new flags `--grpcExpanderCert` and `--grpcExpanderUrl`. Should initialization of the client fail, or any returned response be invalid or malformed, the expander fails open, and doesn't filter any Options.


evansheng:
I've built this and deployed it locally, alongside an internal gRPC server - verified that this is working e2e successfully. I'm able to return options successfully via `bestOptions`. 
I've also successfully tested failure modes in which the gRPC server doesn't exist, we fall back to appropriate specified expanders in the list.

cluster-autoscaler/core/autoscaler.go
```diff
@@ -48,6 +49,7 @@ type AutoscalerOptions struct {
 	EstimatorBuilder       estimator.EstimatorBuilder
 	Processors             *ca_processors.AutoscalingProcessors
 	Backoff                backoff.Backoff
+	GRPCOpts               grpcplugin.GRPCOpts
```
MaciekPytel, line 52: I don't think this is needed here. It's not used by anything except the grpc plugin and I think it's just leaking implementation details of this specific expander.

cluster-autoscaler/core/autoscaler.go
```diff
@@ -101,9 +103,12 @@ func initializeDefaultOptions(opts *AutoscalerOptions) error {
 	if opts.CloudProvider == nil {
 		opts.CloudProvider = cloudBuilder.NewCloudProvider(opts.AutoscalingOptions)
 	}
+	if strings.Contains(opts.ExpanderNames, expander.GRPCExpanderName) && opts.GRPCExpanderURL != "" && opts.GRPCExpanderCert != "" {
+		opts.GRPCOpts = grpcplugin.GRPCOpts{opts.GRPCExpanderCert, opts.GRPCExpanderURL}
```
MaciekPytel, line 107: Similar to the comment above - I don't think code in core/ should contain any logic specific to a particular expander. I'd much rather pass AutoscalingOptions to expander factory (so it has access to all flags) or just define expander-specific flags directly in the expander.

cluster-autoscaler/go.mod
```diff
@@ -25,6 +25,8 @@ require (
 	golang.org/x/crypto v0.0.0-20210513164829-c07d793c2f9a
 	golang.org/x/oauth2 v0.0.0-20210514164344-f6687ab2804c
 	google.golang.org/api v0.46.0
+	google.golang.org/grpc v1.38.0
```
MaciekPytel, line 28: Kubernetes already uses grpc in a different version: https://github.com/kubernetes/kubernetes/blob/master/go.mod#L102, which makes it a transitive dependency of CA. In my experience is that different grpc versions don't always play well together. Right now the difference in version number is pretty small, but is there some way in which we can guarantee the same grpc version is used by CA and kubernetes going forward?

cluster-autoscaler/expander/grpcplugin/grpc_client.go
```diff
@@ -0,0 +1,129 @@
...
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/autoscaler/cluster-autoscaler/expander"
+	"k8s.io/autoscaler/cluster-autoscaler/expander/grpcplugin/protos"
+	"k8s.io/klog/v2"
+	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
```
MaciekPytel, line 15: That is a big dependency to push on anyone forking CA (which is how a lot of cloudproviders work), especially given that they're quite likely to already use grpc, possibly in some other way. Can we try and make it easy to get rid of this dependency by tightly controlling where we import this expander from (ideally try and avoid importing it directly anywhere outside of expander/)?

cluster-autoscaler/go.mod
evansheng: I can use the same version that that kubernetes repo is using pretty easily, but do you have suggestions on how to keep it synced? I did a cursory search across other grpc uses in the k8s repo and found many repos seem to use many different versions:

1. https://github.com/kubernetes/node-problem-detector/blob/fc0edbd222e16b775ec58a253366a41822ecb3d5/vendor/modules.txt#L348
2. https://github.com/kubernetes/ingress-gce/blob/b1b62298cfc7adb08154402f2b26607d1a94d65e/vendor/modules.txt#L191
3. https://github.com/kubernetes/kops/blob/e40fd308b26676ef3f054e702640d32fecaa9271/vendor/modules.txt#L946
4. https://github.com/kubernetes/cloud-provider-openstack/blob/197604cb32c5d1850aeb8c903b7ea280f9bc0e1e/go.mod#L32
5. etc...

cluster-autoscaler/expander/grpcplugin/grpc_client.go
evansheng: good call - removing the grpcOpts struct and just passing in Autoscaler Options to the expander factory directly fixes this

cluster-autoscaler/core/autoscaler.go
evansheng: Opted for passing the grpc expander flags as two more args, instead of either creating an import cycle (core & expander factory), or moving the AutoscalerOptions struct to `config`

evansheng:
added the autogenerated protobuf code to exclusion from golint

evansheng:
@MaciekPytel I believe this PR is ready for another round of review when you get the chance

elmiko (commented):
hey @evansheng , i'm still reviewing this code, but one thing i am wondering about is documenting how to use this expander. is the enhancement meant to be the main documentation or have you considered adding detailed usage instructions to one of our READMEs?

i would like to try this out if possible, but i am still understanding how to setup the grpc expander. perhaps a small "fake" expander would be nice to demonstrate how a user could build one. or is the mock expander meant to be an example?

evansheng:
@elmiko added a README on how to use this new expander, and added some starter code
PTAL when you get the change

elmiko (commented):
thanks @evansheng , that readme and example are exactly what i was looking for. i'd like to test drive this a little more, but i'm +1 for merging.

evansheng:
@MaciekPytel friendly ping for another review on this pr

cluster-autoscaler/expander/grpcplugin/protos/expander.proto
```diff
@@ -0,0 +1,31 @@
...
+service Expander {
+
+  rpc BestOptions (BestOptionsRequest)
+    returns (BestOptionsResponse) {}
+}
+
+message BestOptionsRequest {
+  repeated Option options = 1;
+  // key is node id from options
+  map<string, k8s.io.api.core.v1.Node> nodeInfoMap = 2;
```
bpineau, line 20: So over gRPC `nodeInfoMap` is really a Node map: is that required to save rpc calls throughput and simplify the proto?

cluster-autoscaler/expander/grpcplugin/protos/expander.proto
```diff
@@ -0,0 +1,31 @@
+syntax = "proto3";
+
+package grpcplugin;
+import "k8s.io/api/core/v1/generated.proto";
+//import "google/protobuf/struct.proto";
```
MaciekPytel, line 5: Please remove if unused.

cluster-autoscaler/expander/grpcplugin/grpc_client.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	return &grpcclientstrategy{grpcClient: client}
+}
+
+func createGRPCClient(expanderCert string, expanderUrl string) protos.ExpanderClient {
+	var dialOpt grpc.DialOption
+
+	// if no Cert file specified, use insecure
+	if expanderCert == "" {
+		dialOpt = grpc.WithInsecure()
```
MaciekPytel, line 52: That seems like an unnecessary risk. Kubernetes itself moved away from allowing insecure connections and I think we also shouldn't allow them.

cluster-autoscaler/expander/grpcplugin/grpc_client.go
```diff
@@ -0,0 +1,140 @@
...
+	} else {
+		creds, err := credentials.NewClientTLSFromFile(expanderCert, "")
+		if err != nil {
+			log.Fatalf("Failed to create TLS credentials %v", err)
+			return nil
+		}
+		dialOpt = grpc.WithTransportCredentials(creds)
+	}
+	klog.V(2).Info("Dialing ", expanderUrl, " dialopt: ", dialOpt)
```
MaciekPytel, line 61: nit: Please use Infof

cluster-autoscaler/expander/grpcplugin/grpc_client.go
```diff
@@ -0,0 +1,140 @@
...
+	conn, err := grpc.Dial(expanderUrl, dialOpt)
+	if err != nil {
+		log.Fatalf("fail to dial server: %v", err)
```
MaciekPytel, line 64: nit: Here Fatalf is lowercase, but for failure to create credentials it's upper case

cluster-autoscaler/expander/grpcplugin/grpc_client.go
```diff
@@ -0,0 +1,140 @@
...
+		return nil
+	}
+	return protos.NewExpanderClient(conn)
+}
+
+func (g *grpcclientstrategy) BestOptions(expansionOptions []expander.Option, nodeInfo map[string]*schedulerframework.NodeInfo) []expander.Option {
+	if g.grpcClient == nil {
+		log.Fatalf("Incorrect gRPC client config, filtering no options")
```
MaciekPytel, line 72: Fatalf() is fine at initialization time, but it generally shouldn't be used to handle runtime issues (logging at Fatal exits the process). Especially given that in this case there is a reasonable fallback (return all options), so logging at Error level and continuing seems like the right choice.

cluster-autoscaler/expander/grpcplugin/grpc_client.go
```diff
@@ -0,0 +1,140 @@
...
+		return expansionOptions
+	}
+
+	// Transform inputs to gRPC inputs
+	nodeGroupIDOptionMap := make(map[string]expander.Option)
+	grpcOptionsSlice := []*protos.Option{}
+	populateOptionsForGRPC(expansionOptions, nodeGroupIDOptionMap, &grpcOptionsSlice)
```
MaciekPytel, line 79: It seems much more natural to create datastructures in the function and return them rather than mutate them this way. Is there any particular reason for the current approach?

cluster-autoscaler/expander/grpcplugin/grpc_client.go
```diff
@@ -0,0 +1,140 @@
...
+	grpcNodeInfoMap := make(map[string]*v1.Node)
+	populateNodeInfoForGRPC(nodeInfo, grpcNodeInfoMap)
+
+	// call gRPC server to get BestOption
+	klog.V(2).Info("GPRC call of best options to server with ", len(nodeGroupIDOptionMap), " options")
```
MaciekPytel, line 84: I think this sort of logs are generally logged in CA at V(4). V(2) seems like a very high level for a debug log like this. Same goes for other logs in this PR.

cluster-autoscaler/expander/grpcplugin/grpc_client.go
```diff
@@ -0,0 +1,140 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
MaciekPytel, line 85: Consider making 5s a const

cluster-autoscaler/expander/grpcplugin/grpc_client.go
```diff
@@ -0,0 +1,140 @@
...
+
+// populateOptionsForGRPC creates a map of nodegroup ID and options, as well as a slice of Options objects for the gRPC call
+func populateOptionsForGRPC(expansionOptions []expander.Option, nodeGroupIDOptionMap map[string]expander.Option, grpcOptionsSlice *[]*protos.Option) {
+	for _, option := range expansionOptions {
+		nodeGroupIDOptionMap[option.NodeGroup.Id()] = option
+		*grpcOptionsSlice = append(*grpcOptionsSlice, newOptionMessage(option.NodeGroup.Id(), int32(option.NodeCount), option.Debug, option.Pods))
+	}
+}
+
+// populateNodeInfoForGRPC modifies the nodeInfo object, and replaces it with the v1.Node to pass through grpc
```
MaciekPytel, line 114: That comment is misleading, original nodeInfos are not modified (and they must not be!).

cluster-autoscaler/expander/grpcplugin/grpc_client.go
```diff
@@ -0,0 +1,140 @@
...
+		grpcNodeInfoMap[nodeId] = nodeInfo.Node()
+	}
+}
+
+func transformAndSanitizeOptionsFromGRPC(bestOptionsResponseOptions []*protos.Option, nodeGroupIDOptionMap map[string]expander.Option) []expander.Option {
+	var options []expander.Option
+	for _, option := range bestOptionsResponseOptions {
+		if option == nil {
+			klog.Errorf("gRPC server returned nil Option")
+			return nil
```
MaciekPytel, line 126: Wouldn't continue make more sense here? Other options may not be nil.

cluster-autoscaler/expander/grpcplugin/grpc_client.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+		if _, ok := nodeGroupIDOptionMap[option.NodeGroupId]; ok {
+			options = append(options, nodeGroupIDOptionMap[option.NodeGroupId])
+		} else {
+			klog.Errorf("gRPC server returned invalid nodeGroup ID: ", option.NodeGroupId)
+			return nil
```
MaciekPytel, line 132: Same as above, continue seems more natural.

cluster-autoscaler/expander/grpcplugin/grpc_client_test.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+	grpcEoM44XLarge = protos.Option{
+		NodeGroupId: eoM44XLarge.NodeGroup.Id(),
+		NodeCount:   int32(eoM44XLarge.NodeCount),
+		Debug:       eoM44XLarge.Debug,
+		Pod:         eoM44XLarge.Pods,
+	}
+)
+
+func TestPopulateOptionsForGrpc(t *testing.T) {
```
MaciekPytel, line 89: Please consider a table-based test to cover various inputs, including edge cases (ex. empty list of options).

cluster-autoscaler/expander/grpcplugin/fake_grpc_server.go
```diff
@@ -0,0 +1,104 @@
...
+	"k8s.io/autoscaler/cluster-autoscaler/expander/grpcplugin/protos"
+)
+
+// This code is meant to be used as starter code, deployed as a separate app, not in Cluster Autoscaler.
+// This serves as the gRPC Expander Server counterpart to the client which lives in this repo
+// main.go of said application should simply pass in paths to (optional)cert, (optional)private key, and port, and call Serve to start listening
+// copy the protos/expander.pb.go to your other application's repo, so it has access to the protobuf definitions
+
+// Serve should be called by the main() function in main.go of the Expander Server repo to start serving
+func Serve(certPath string, keyPath string, port uint) {
```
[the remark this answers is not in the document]
evansheng: including a simple main function but leaving out the makefile, as it is up to the user how they would like to structure their project repo

evansheng:
@MaciekPytel addressed your comments, PTAL!

cluster-autoscaler/expander/factory/expander_factory.go
```diff
@@ -67,6 +70,9 @@ func ExpanderStrategyFromStrings(expanderFlags []string, cloudProvider cloudprov
 			stopChannel := make(chan struct{})
 			lister := kubernetes.NewConfigMapListerForNamespace(kubeClient, stopChannel, configNamespace)
 			filters = append(filters, priority.NewFilter(lister.ConfigMaps(configNamespace), autoscalingKubeClients.Recorder))
+		case expander.GRPCExpanderName:
+			klog.V(1).Info("GRPC expander chosen")
```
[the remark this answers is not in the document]
evansheng: done

cluster-autoscaler/core/autoscaler.go
evansheng: done

cluster-autoscaler/expander/grpcplugin/protos/expander.proto
evansheng: called out in readme, and renamed variable to be more appropriate

cluster-autoscaler/expander/grpcplugin/grpc_client.go
evansheng: removed this option

cluster-autoscaler/expander/grpcplugin/grpc_client.go
evansheng: done

cluster-autoscaler/expander/grpcplugin/grpc_client.go
evansheng: good catch, fixed

cluster-autoscaler/expander/grpcplugin/grpc_client_test.go
evansheng: added some tests, as requested

evansheng:
@elmiko @MaciekPytel - addressed all comments, and squashed into two commits.
I think this PR is ready for another round of review!

Issue #4436, opened 2021-11-01
Use EC2 DescribeInstanceTypes API instead of parsing pricing API


AustinSiu:
**Which component are you using?**: 
Cluster Autoscaler

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:

CA maintains code to handle [instance filtering and parsing](https://github.com/kubernetes/autoscaler/blame/d09b8931bb1949018f9b502368bb5f35c57bb08e/cluster-autoscaler/cloudprovider/aws/aws_util.go#L63-L126) from EC2 [pricing API](https://github.com/kubernetes/autoscaler/blame/d09b8931bb1949018f9b502368bb5f35c57bb08e/cluster-autoscaler/cloudprovider/aws/aws_util.go#L40).

**Describe the solution you'd like.**:

Update [existing API call](https://github.com/kubernetes/autoscaler/blame/d09b8931bb1949018f9b502368bb5f35c57bb08e/cluster-autoscaler/cloudprovider/aws/aws_util.go#L40) with a [filterable list](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstanceTypes.html) through EC2's DescribeInstanceTypes API to simplify the logic.

**Describe any alternative solutions you've considered.**:

**Additional context.**:

sbs2001:
@AustinSiu  are you working on a patch for this ? If not I'd like to submit one :)

AustinSiu:
> @AustinSiu are you working on a patch for this ? If not I'd like to submit one :)

I'm actively working on a patch for this. Thanks for checking!

jaypipes:
@gjtempleton honestly, I don't know why we didn't go this route when the EC2 DescribeInstanceTypes API was released in November 22, 2019. I'm supportive of this approach and @AustinSiu's PR looks solid to me.

gjtempleton:
Addressed by #4468 which has now been merged and released.
/close

PR #4468, opened 2021-11-15
Use DescribeInstanceTypes API to get EC2 instance type details

AustinSiu:
This change addresses #4436 by replacing the existing pricing API call used to resolve instance types with EC2's DescribeInstanceTypes API. It also adds unit tests and updates the readme.

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/README.md
+++ b/cluster-autoscaler/cloudprovider/aws/README.md
@@ -32,7 +32,8 @@ The following policy provides the minimum privileges necessary for Cluster Autos
         "autoscaling:DescribeAutoScalingInstances",
         "autoscaling:DescribeLaunchConfigurations",
         "autoscaling:SetDesiredCapacity",
-        "autoscaling:TerminateInstanceInAutoScalingGroup"
+        "autoscaling:TerminateInstanceInAutoScalingGroup",
+        "ec2:DescribeInstanceTypes"
       ],
       "Resource": ["*"]
     }
--- a/cluster-autoscaler/cloudprovider/aws/ec2_instance_types/gen.go
+++ b/cluster-autoscaler/cloudprovider/aws/ec2_instance_types/gen.go
@@ -1,7 +1,7 @@
 // +build ignore
 
 /*
-Copyright 2017 The Kubernetes Authors.
+Copyright 2021 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


AustinSiu:
@aleksandra-malinowska @feiskyer may I ask when you will be able to review this PR? Thanks!

feiskyer:
+ @Jeffwan @jaypipes @gjtempleton

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -32,7 +32,8 @@ The following policy provides the minimum privileges necessary for Cluster Autos
         "autoscaling:DescribeAutoScalingInstances",
         "autoscaling:DescribeLaunchConfigurations",
         "autoscaling:SetDesiredCapacity",
-        "autoscaling:TerminateInstanceInAutoScalingGroup"
+        "autoscaling:TerminateInstanceInAutoScalingGroup",
+        "ec2:DescribeInstanceTypes"
```
jaypipes, line 36: OK, so like I said on #4436, I very much support this approach. I think it's much simpler and cleaner than the existing pricing API-based strategy.

The only thing that is going to be a bit tricky is this part here which requires a new IAM policy for the running cluster-autoscaler binary. This new required IAM privilege isn't onerous per-se, but what it does mean is that we can no longer pre-generate instance type information using an anonymous user (something that was previously possible since we were basically just curl'ing a public HTTPS endpoint for pricing JSON files).

I am OK with this limitation, but we should point out (maybe in the [gen.go](https://github.com/kubernetes/autoscaler/blob/01a32390344755ba4339f5bb72b8c6f2b2ae1966/cluster-autoscaler/cloudprovider/aws/ec2_instance_types/gen.go) file code comments?) that the user running the static instance types generator will need to be a non-anonymous IAM user with privileges to call the DescribeInstanceTypes EC2 API.

jaypipes (changes requested):
Requesting a bit of documentation in the [gen.go](https://github.com/kubernetes/autoscaler/blob/01a32390344755ba4339f5bb72b8c6f2b2ae1966/cluster-autoscaler/cloudprovider/aws/ec2_instance_types/gen.go) file (see inline comment) but otherwise this approach is excellent and MUCH cleaner than the existing strategy. Well done, @AustinSiu.

bwagner5 (approved):
Apart from Jay's comment on docs for the permissions, this LGTM! Nice work!

cluster-autoscaler/cloudprovider/aws/README.md
AustinSiu: Understood, thank you for the comments and suggestions, Jay! I'll add the requested documentation shortly.

jaypipes (approved):
:+1: great work on this @AustinSiu! :)

cluster-autoscaler/cloudprovider/aws/ec2_instance_types/gen.go
```diff
@@ -1,3 +1,4 @@
+//go:build ignore
```
jaypipes, line 1: @AustinSiu looks like the CI's Go version doesn't like this change. Best to remove it.

AustinSiu:
> 👍 great work on this @AustinSiu! :)

Thank you @jaypipes! May I ask, are you able/willing to apply the `lgtm` label? Assuming that is the final remaining step for merging.

PR #4469, opened 2021-11-16
Increase upper bound for expected VPA CPU recommendations

jbartosik:
Tests are flaky with VPA sometimes generating recommendations higher
than 1000 mCPU.

I think this is a reasonable behavior - we're asking resoirce consumer
to use 1800 mCPU between 3 pods, if it gets unevenly distributed we can
end up with some pods using 1000 mCPU.


mwielgus (approved):
/lgtm
/approve

jbartosik:
For example in [this run](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-autoscaling-vpa-full/1460541017580113920) we see that `have cpu requests growing with usage` fails waiting for pods in range 500 - 1000 mCPU, when actual pod requests are in range 1101 - 1311 mCPU.

jbartosik:
@wangchen615 FYI

Issue #4464, opened 2021-11-12
CA failed to load Instance Type list unless configured with hostNetworking


adaam:
**Which component are you using?**:
cluster-autoscaler

**What version of the component are you using?**:
Helm chart 9.10.8
cluster-autoscaler v1.21.1

Component version:

**What k8s version are you using (`kubectl version`)?**:
v1.21
<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.1", GitCommit:"632ed300f2c34f6d6d15ca4cef3d3c7073412212", GitTreeState:"clean", BuildDate:"2021-08-19T15:38:26Z", GoVersion:"go1.16.6", Compiler:"gc", Platform:"darwin/arm64"}
Server Version: version.Info{Major:"1", Minor:"21+", GitVersion:"v1.21.2-eks-06eac09", GitCommit:"5f6d83fe4cb7febb5f4f4e39b3b2b64ebbbe3e97", GitTreeState:"clean", BuildDate:"2021-09-13T14:20:15Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What environment is this in?**:
AWS EKS

**What did you expect to happen?**:
It will load instance type list normally and keep running.

**What happened instead?**:
It keep CrashLoopBack and exit with error 255

**How to reproduce it (as minimally and precisely as possible)**:

Set environment variable with
`AWS_REGION: ap-northeast-3`

**Anything else we need to know?**:

Part of logs:

```
1112 07:23:25.974866       1 main.go:391] Cluster Autoscaler 1.21.1
I1112 07:23:25.996783       1 leaderelection.go:243] attempting to acquire leader lease kube-system/cluster-autoscaler...
I1112 07:23:26.016572       1 leaderelection.go:253] successfully acquired lease kube-system/cluster-autoscaler
[...]

Issue #4220, opened 2021-07-23
My pod is in the CrashLoopBackOff state after configuring  cluster-autoscaler 


ibata:
**Which component are you using?**:
 cluster-autoscaler

**What version of the component are you using?**:
v1.17.3

Component version:

**What k8s version are you using (`kubectl version`)?**:

kubectl -n kube-system version --short
Client Version: v1.21.1
Server Version: v1.17.17-eks-c5067d
WARNING: version difference between client (1.21) and server (1.17) exceeds the supported minor version skew of +/-1

**What environment is this in?**:

aws eks

**What did you expect to happen?**:

I'm expecting to get my first cluster-autoscaler to be setup and working. Meaning replace my ASG.

**What happened instead?**:

Getting exactly this error reported by https://aws.amazon.com/premiumsupport/knowledge-center/eks-pod-status-troubleshooting/ 

$ kubectl describe po crash-app-6847947bf8-28rq6

Name:               crash-app-6847947bf8-28rq6
Namespace:          default
Priority:           0
PriorityClassName:  <none>
Node:               ip-192-168-6-51.us-east-2.compute.internal/192.168.6.51
Start Time:         Wed, 22 Jan 2020 08:42:20 +0200
Labels:             pod-template-hash=6847947bf8
                    run=crash-app
Annotations:        kubernetes.io/psp: eks.privileged
Status:             Running
IP:                 192.168.29.73
Controlled By:      ReplicaSet/crash-app-6847947bf8
Containers:
  main:
    Container ID:  docker://6aecdce22adf08de2dbcd48f5d3d8d4f00f8e86bddca03384e482e71b3c20442
    Image:         alpine
    Image ID:      docker-pullable://alpine@sha256:ab00606a42621fb68f2ed6ad3c88be54397f981a7b70a79db3d1172b11c4367d
    Port:          80/TCP
    Host Port:     0/TCP
    Command:
      /bin/sleep
      1
    State:          Waiting
      Reason:       CrashLoopBackOff
...
Events:
  Type     Reason     Age                From                                                 Message
  ----     ------     ----               ----                                                 -------
  Normal   Scheduled  47s                default-scheduler                                    Successfully assigned default/crash-app-6847947bf8-28rq6 to ip-192-168-6-51.us-east-2.compute.internal
  Normal   Pulling    28s (x3 over 46s)  kubelet, ip-192-168-6-51.us-east-2.compute.internal  Pulling image "alpine"
[...]

Issue #4036, opened 2021-04-26
Crash loop: Failed to generate AWS EC2 Instance Types: unable to load EC2 Instance Type list


fitchtech:
I0426 16:22:12.584855       1 flags.go:52] FLAG: --add-dir-header="false"
I0426 16:22:12.584904       1 flags.go:52] FLAG: --address=":8085"
I0426 16:22:12.584910       1 flags.go:52] FLAG: --alsologtostderr="false"
I0426 16:22:12.584915       1 flags.go:52] FLAG: --aws-use-static-instance-list="false"
I0426 16:22:12.584919       1 flags.go:52] FLAG: --balance-similar-node-groups="false"
I0426 16:22:12.584923       1 flags.go:52] FLAG: --balancing-ignore-label="[]"
I0426 16:22:12.584928       1 flags.go:52] FLAG: --cloud-config=""
I0426 16:22:12.584936       1 flags.go:52] FLAG: --cloud-provider="aws"
I0426 16:22:12.584941       1 flags.go:52] FLAG: --cloud-provider-gce-l7lb-src-cidrs="130.211.0.0/22,35.191.0.0/16"
I0426 16:22:12.584947       1 flags.go:52] FLAG: --cloud-provider-gce-lb-src-cidrs="130.211.0.0/22,209.85.152.0/22,209.85.204.0/22,35.191.0.0/16"
I0426 16:22:12.584953       1 flags.go:52] FLAG: --cluster-name=""
I0426 16:22:12.584963       1 flags.go:52] FLAG: --clusterapi-cloud-config-authoritative="false"
I0426 16:22:12.584968       1 flags.go:52] FLAG: --cores-total="0:320000"
I0426 16:22:12.584972       1 flags.go:52] FLAG: --estimator="binpacking"
I0426 16:22:12.584995       1 flags.go:52] FLAG: --expander="random"
I0426 16:22:12.585000       1 flags.go:52] FLAG: --expendable-pods-priority-cutoff="-10"
I0426 16:22:12.585005       1 flags.go:52] FLAG: --gpu-total="[]"
I0426 16:22:12.585010       1 flags.go:52] FLAG: --ignore-daemonsets-utilization="false"
I0426 16:22:12.585014       1 flags.go:52] FLAG: --ignore-mirror-pods-utilization="false"
I0426 16:22:12.585018       1 flags.go:52] FLAG: --ignore-taint="[]"
[...]

PR #4480, opened 2021-11-28
CA - AWS CloudProvider - Fallback to Static EC2 list rather than fatal error

gjtempleton:
Related to #4464 

Currently, if the CA is unable to dynamically load instances the CA immediately crashes fatally. Instead, we should gracefully fall back to using the bundled static list of EC2 instance types, warning the user this is what we're doing.

I'm not 100% sure this is the right move to make, as it will potentially mask more errors in people's configs by running with degraded functionality, though #4468 should help alleviate this, by returning more meaningful AWS errors from `GenerateEC2InstanceTypes`. 

In this case, if someone was running this way (e.g. due to insufficient IAM permissions) and tried to use a newer instance type not included in the fallback static list, they would then receive an error. It may not be clear from the distance between where this fallback message is, and the error being generated why this would be happening: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/aws_manager.go#L328

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
@@ -366,7 +366,7 @@ func BuildAWS(opts config.AutoscalingOptions, do cloudprovider.NodeGroupDiscover
 
 		generatedInstanceTypes, err := GenerateEC2InstanceTypes(region)
 		if err != nil {
-			klog.Fatalf("Failed to generate AWS EC2 Instance Types: %v", err)
+			klog.Warningf("Failed to generate AWS EC2 Instance Types: %v, falling back to using static Instance Types. Last update time: %s", err, lastUpdateTime)
 		}
 		// fallback on the static list if we miss any instance types in the generated output
 		// credits to: https://github.com/lyft/cni-ipvlan-vpc-k8s/pull/80
@@ -385,7 +385,7 @@ func BuildAWS(opts config.AutoscalingOptions, do cloudprovider.NodeGroupDiscover
 			keys = append(keys, key)
 		}
 
-		klog.Infof("Successfully load %d EC2 Instance Types %s", len(keys), keys)
+		klog.Infof("Successfully loaded %d EC2 Instance Types %s", len(keys), keys)
 	}
 
 	manager, err := CreateAwsManager(config, do, instanceTypes)


gjtempleton:
@Jeffwan I would appreciate your input on this.

/assign @jaypipes

MyannaHarris (commented):
This change makes sense since random issues out of the user's control could cause the failure to dynamically load the information, but, correct me if I'm wrong, the instance type information only gets loaded when Cluster Autoscaler starts up. Since the instance type information never gets updated, defaulting to the static information could definitely mask issues on the user's side. And since it only gets called when Cluster Autoscaler starts-up, the warning could easily be missed by a user.

I think this change would work great if there was also a periodic regeneration of the instance types information. Then, if it's a user issue, this warning would be printed multiple times in the logs. Or, if it's an issue out of their control, the map will be fixed soon automatically.

If I read the code wrong and there is in fact already a periodic regeneration of the instance type information, then this change looks good.

jaypipes:
@gjtempleton @MyannaHarris yeah, I think we should revisit the whole dynamic instance type fetching. It's cause OOM issues (#4220, #4036, #3044, #3506) and I believe a better, more stable, approach to solving this problem would be to add a periodic CI job, initially set to run, say, every night or something (but potentially being triggered off some AWS-sourced event) that calls the `DescribeInstanceTypes` API call and regenerates the static instance type structs, automatically creating a pull request that updates the master branch. Potentially we could write some automation that does the same for release branches as well... thoughts?

gjtempleton:
Thanks for the feedback and thoughts.

No arguments here with revisiting the dynamic instance type fetching wholesale. There are already some incremental improvements in flight/implemented, though they don't materially change the model of a hardcoded fallback list or a default dynamically generated one on startup. 

There has been an improvement in the memory use of the current approach since the merging of #4199 to move to stream processing of the API. We also now have a PR (which I haven't had time to have a look at yet) to move from the current JSON implementation to using `DescribeInstanceTypes` in #4468. 

As much as I love the proposal to automate the regeneration of the static list, we'd need to move to an automated process of also cutting and promoting the image releases to make it usable enough for users to move away from the dynamic list generation on startup being the default behaviour, there's currently far too much manual action/friction involved in that pipeline to make the process sufficiently fast in my view.

gjtempleton:
Closing as superseded by the larger improvements brought in by #4468

PR #4487, opened 2021-11-30
Update cluster-autoscaler FAQ to define correct resource-lock options

g1bl3t:
The cluster-autoscaler FAQ has outdated options documented for the `leader-elect-resource-lock` parameter. The [upstream LeaderElectionConfiguration](https://github.com/kubernetes/kubernetes/blob/c1153d3353bd4f4b68d85245d53d2745586be474/staging/src/k8s.io/component-base/config/options/leaderelectionconfig.go#L43) specifies the following as options for this field:
```
 endpoints, configmaps, leases, endpointsleases, and configmapsleases
```

The default is now specified [here](https://github.com/kubernetes/autoscaler/blob/222f71ea73dcd3bf1d331b9bbddfa060137249ba/cluster-autoscaler/main.go#L471) as the option "leases".


feiskyer (approved):
/lgtm

PR #4494, opened 2021-12-06
switch azure clients to non-legacy repo

marwanad:
This switches the azure cloudprovider clients to not use the legacy repo since its no longer maintained.

Dependency is now pulled from sig.k8s.io/https://github.com/kubernetes-sigs/cloud-provider-azure


cluster-autoscaler/cloudprovider/azure/azure_config.go
```diff
@@ -157,12 +157,6 @@ func BuildAzureConfig(configReader io.Reader) (*Config, error) {
 		cfg.ClusterName = os.Getenv("AZURE_CLUSTER_NAME")
 		cfg.NodeResourceGroup = os.Getenv("AZURE_NODE_RESOURCE_GROUP")
 
-		subscriptionID, err := getSubscriptionIdFromInstanceMetadata()
```
marwanad, line 160: I had to get rid of that because I was getting:

```
'sigs.k8s.io/cloud-provider-azure/pkg/provider' imports package with invalid import comment 
```

on importing. Looking back at it - I highly doubt this was ever used since most people would either provide AZ config or explicitly sub ID from env vars.

feiskyer (commented):
/lgtm
/approve

PR #4321, opened 2021-09-09
[HELM] Fix missing yaml conversion for expander priorities

bwhaley:
This should fix an error with Helm when passing expanderPriorities:

```
Error: template: cluster-autoscaler/templates/priority-expander-configmap.yaml:11:39: executing "cluster-autoscaler/templates/priority-expander-configmap.yaml" at <4>: wrong type for value; expected string; got map[string]interface {}
```

PR #4503, opened 2021-12-07
Helm Chart bug, can not use a cascading expander with priority in it

### LINKED ISSUE #4502: Bug: Helm Chart bug, can not use a cascading expander with priority in it, configmap missing
# Issue
The conditional creation logic for the [priority-expander-configmap.yaml](https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/templates/priority-expander-configmap.yaml#L2):
```
{{- if hasKey .Values.extraArgs "expander" }}
{{- if and (.Values.expanderPriorities) (eq .Values.extraArgs.expander "priority") -}}
```

Directly conflicts with ability to pass multiple expanders to cascade, info [from the FAQ](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-expanders):
```
Multiple expanders may be passed, i.e. .cluster-autoscaler --expander=priority,least-waste
```

which results in it not being possible to set...
```yaml
  extraArgs:
    expander: priority,least-waste  # <--- this
  expanderPriorities: |-
    10:
      - '.*g2\..*'
```

And have it still set the `expanderPriorities` configmap.  Right now I can set this expander value above and it works properly, except the configmap is not created because of this flawed conditional logic.

# Bugfix
Possible bugfix is to change the "eq" to contains in that second line, I'm submitting an MR now.

AndrewFarley:
Fixes #4502

### DIFF
--- a/charts/cluster-autoscaler/templates/priority-expander-configmap.yaml
+++ b/charts/cluster-autoscaler/templates/priority-expander-configmap.yaml
@@ -1,5 +1,5 @@
 {{- if hasKey .Values.extraArgs "expander" }}
-{{- if and (.Values.expanderPriorities) (eq .Values.extraArgs.expander "priority") -}}
+{{- if and (.Values.expanderPriorities) (contains "priority" .Values.extraArgs.expander) -}}
 apiVersion: v1
 kind: ConfigMap
 metadata:
--- a/charts/cluster-autoscaler/templates/role.yaml
+++ b/charts/cluster-autoscaler/templates/role.yaml
@@ -12,7 +12,7 @@ rules:
       - configmaps
     verbs:
       - create
-{{- if eq (default "" .Values.extraArgs.expander) "priority" }}
+{{- if contains "priority" (default "" .Values.extraArgs.expander) }}
       - list
       - watch
 {{- end }}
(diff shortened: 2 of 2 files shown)


AndrewFarley:
Related to: #4321 

With both of these fixed, proper support for expander priority will be fully capable in the helm chart

AndrewFarley:
I'm fully CLA'd, please @ me if you need anything else.  This is a pretty bad bug, blocking me to use upstream charts at the moment.  Please fix, accept, and publish new Helm Charts ASAP if possible.  Thanks

AndrewFarley:
Found a bug, fixed role.yaml which also used the old eq logic.  Fixed to use contains logic, now the output is exactly the same.

gjtempleton:
Hi @AndrewFarley, thanks for the PR!

The chart version will need a bump [here](https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/Chart.yaml#L20) before we can merge it.

/assign @gjtempleton

AndrewFarley:
Note: Safe to merge now, but this feature won't work until #4504 is resolved.

AndrewFarley:
FYI, it looks like someone else fixed this in another MR/PR in https://github.com/kubernetes/autoscaler/commit/e4cf3b0f4833ebd2d01dd8172d32485b44f6d8ef but I think perhaps theirs was incomplete and missing some stanzas I included to handle all edge cases.

gjtempleton:
Sorry, my bad in terms of reviewing order. I assume from the merging of #4665 and your update to the comment there, we can close this PR now?

PR #4508, opened 2021-12-08
Cluster Autoscaler: GCE: check the result of the operation

aledbf:
Suppose the instance group update mode is not `Opportunistic`. In that case, the operation ends without error; the GCP console shows no increase in the number of nodes (but it shows the updating clock image) and tries to repeat the operation multiple times.

If the operation is `DONE` but contains an error, return it.

Note: I am aware this is not an error in the autoscaler but the only place to find the reason why the scale-up operation is not being done is in the console https://console.cloud.google.com/compute/operations (even there, the operation appears as OK)

Screenshot from 2021-12-08 18-12-23


aledbf:
@jbartosik what can I do/is missing to get this change merged? Thanks

jbartosik:
Hi,
I contribute to VPA. @MaciekPytel will know more about Cluster Autoscaler.

I tagged the PR because I try to keep all PRs and Issues tagged (to make them easier to find).

You can try slack (linked from [README](https://github.com/kubernetes/autoscaler#contact-info)) or attending a SIG meeting ([link](https://github.com/kubernetes/community/tree/master/sig-autoscaling#meetings)).

mwielgus (approved):
/lgtm
/approve

aledbf:
@mwielgus thanks!

PR #4511, opened 2021-12-09
Move gce reserved to interface

yaroslava-serdiuk:


cluster-autoscaler/cloudprovider/gce/reserved.go
```diff
@@ -16,275 +16,11 @@ limitations under the License.
...
 package gce
 
-import (
-	"fmt"
-	"math"
-	"strconv"
-	"strings"
+// Reserved calculates the OS reserved values.
+type Reserved interface {
+	// CalculateKernelReserved computes how much memory Linux kernel will reserve.
```
MaciekPytel, line 21: nit: s/Linux/OS? It also covers windows nodes.

cluster-autoscaler/cloudprovider/gce/reserved.go
```diff
@@ -16,275 +16,11 @@ limitations under the License.
 
 package gce
 
-import (
-	"fmt"
-	"math"
-	"strconv"
-	"strings"
+// Reserved calculates the OS reserved values.
+type Reserved interface {
```
MaciekPytel, line 20: I'd prefer a more descriptive name. Maybe 'NodeReservedCalculator'? Or 'ReservedResoucesCalculator'?

cluster-autoscaler/cloudprovider/gce/reserved.go
```diff
@@ -16,275 +16,11 @@ limitations under the License.
...
+	// CalculateKernelReserved computes how much memory Linux kernel will reserve.
+	CalculateKernelReserved(physicalMemory int64, os OperatingSystem, osDistribution OperatingSystemDistribution, nodeVersion string) int64
```
MaciekPytel, line 22: Current implementation doesn't use the nodeVersion parameter and the code calling it passes empty string. Should the interface (via docstring) specify some expectations here? Ex. by explicitly saying nodeVersion is optional and clarifying the expected behavior if not specified?

cluster-autoscaler/cloudprovider/gce/reserved.go
yaroslava-serdiuk: Renamed to OsReservedCalculator

cluster-autoscaler/cloudprovider/gce/reserved.go
yaroslava-serdiuk: Added a comment that it optional. 
>Ex. by explicitly saying nodeVersion is optional and clarifying the expected behavior if not specified?
behavior depends on implementation, so not sure what to say.

cluster-autoscaler/cloudprovider/gce/os_reserved.go
```diff
@@ -0,0 +1,28 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package gce
+
+// OsReservedCalculator calculates the OS reserved values.
+type OsReservedCalculator interface {
+	// CalculateKernelReserved computes how much memory OS kernel will reserve.
+	// NodeVersion parameter may be optional if it is not used in calculation.
```
MaciekPytel, line 22: As a user of the interface I don't know if the field is used in calculation or not. What I had in mind was along the lines of 'NodeVersion parameter is optional. If left empty a result calculated using default node version will be returned', ie. setting requirements for how any implementation is expected to handle empty value.

PR #4512, opened 2021-12-09
Refresh instances to mig cache in MigInfoProvider

BigDarkClown:
Refreshing `GceCache` should be done by other components, similarly to how the basenames are refreshed by `MigInfoProvider` or how `Migs` are registered by `GceManager`. This provides additional level of isolation in the code, cleans up dependencies and simplifies reasoning.

Currently instances to `Migs` cache is being refreshed by `GceCache` itself. This introduces unnecessary dependancy on the `GceClient`. This change moves that refreshing logic to `MigInfoProvider`, simplifying the resulting dependency network.

The changes are aimed to maintain the same functionality as the current one. The only significant difference is checking all the Migs in `RefillAllMigInstances` and `fillCandidateMigInstances` functions in `MigInfoProvider`. Thanks to that the CA should be able to handle fetches returning errors better - such cases will no longer block cache updating for other Migs.

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/cache.go
+++ b/cluster-autoscaler/cloudprovider/gce/cache.go
@@ -17,14 +17,10 @@ limitations under the License.
 package gce
 
 import (
-	"context"
-	"fmt"
 	"reflect"
-	"strings"
 	"sync"
 
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
-	"k8s.io/client-go/util/workqueue"
 
 	gce "google.golang.org/api/compute/v1"
 	klog "k8s.io/klog/v2"
@@ -44,30 +40,22 @@ type machinesCacheValue struct {
 // GceCache is used for caching cluster resources state.
 //
 // It is needed to:
-// - keep track of autoscaled MIGs in the cluster,
-// - keep track of instances and which MIG they belong to,
+// - keep track of MIGs in the cluster,
+// - keep track of MIGs to instances mapping,
+// - keep track of MIGs configuration such as target size and basename,
+// - keep track of resource limiters and machine types,
 // - limit repetitive GCE API calls.
 //
-// Cached resources:
-// 1) MIG configuration,
-// 2) instance->MIG mapping,
-// 3) resource limits (self-imposed quotas),
-// 4) machine types.
-//
-// How it works:
-// - migs (1), resource limits (3) and machine types (4) are only stored in this cache,
-// not updated by it.
-// - instanceRefToMigRef (2) is based on registered migs (1). For each mig, its instances
-// are fetched from GCE API using gceService.
-// - instanceRefToMigRef (2) is NOT updated automatically when migs field (1) is updated. Calling
-// RegenerateInstancesCache is required to sync it with registered migs.
+// Cache keeps these values and gives access to getters, setters and
+// invalidators all guarded with mutex. Cache does not refresh the data by
+// itself - it just provides an interface enabling access to this data.
 type GceCache struct {
 	cacheMutex sync.Mutex
 
 	// Cache content.
 	migs                      map[GceRef]Mig
-	instanceRefToMigRef       map[GceRef]GceRef
-	instancesFromUnknownMigs  map[GceRef]struct{}
+	instancesToMig            map[GceRef]GceRef
+	instancesFromUnknownMig   map[GceRef]bool
 	resourceLimiter           *cloudprovider.ResourceLimiter
 	autoscalingOptionsCache   map[GceRef]map[string]string
 	machinesCache             map[MachineTypeKey]machinesCacheValue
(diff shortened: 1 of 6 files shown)


cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
```diff
@@ -41,21 +46,95 @@ type MigInfoProvider interface {
...
+	}
+	return nil, false, nil
+}
+
+// RefillAllMigInstances refills MIGs to instances mapping cache
+func (c *cachingMigInfoProvider) RefillAllMigInstances() error {
+	c.cache.InvalidateAllInstancesToMig()
+	migs := c.cache.GetMigs()
+	errors := make([]error, len(migs))
+	workqueue.ParallelizeUntil(context.Background(), c.concurrentGceRefreshes, len(migs), func(piece int) {
```
MaciekPytel, line 103: Previous version would cancel upon receiving an error. Just to clarify - this change is by design, right? I'm assuming the intent here is to allow for better error handling?

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -147,151 +136,80 @@ func (gc *GceCache) GetMigs() []Mig {
...
 }
 
-func (gc *GceCache) getMigNoLock(migRef GceRef) (mig Mig, found bool) {
-	mig, found = gc.migs[migRef]
-	return
-}
-
-// RegenerateInstanceCacheForMig triggers instances cache regeneration for single MIG under lock.
-func (gc *GceCache) RegenerateInstanceCacheForMig(migRef GceRef) error {
-	klog.V(4).Infof("Regenerating MIG information for %s", migRef.String())
```
MaciekPytel, line 233: Do we think those were useless? I'm not saying they weren't, just want to make sure this is a conscious decision.

cluster-autoscaler/cloudprovider/gce/mig_info_provider_test.go
```diff
@@ -107,11 +109,284 @@ func (client *mockAutoscalingGceClient) CreateInstances(_ GceRef, _ string, _ in
 	return nil
 }
 
+func TestGetMigForInstanceProvider(t *testing.T) {
```
MaciekPytel, line 112: nit: 'Provider' makes it look like there is some 'InstanceProvider' being tested.

cluster-autoscaler/cloudprovider/gce/mig_info_provider_test.go
```diff
@@ -28,25 +29,26 @@ import (
...
-		gceRef: gceRef,
+		gceRef: GceRef{
+			Project: "project",
+			Zone:    "us-test1",
+			Name:    "mig",
+		},
 	}
 )
 
 type mockAutoscalingGceClient struct {
```
MaciekPytel, line 47: Not related to this PR, but I'd love to see that extracted to a separate file, so we could start using it in other tests. For 90% of tests this is so much better than the Http mock we use.
Though with this refactor maybe a MockMigInfoProvider would be even better?

MaciekPytel (commented):
I think this is a very nice refactor, the responsibilities are now split much more cleanly between components. Left a bunch of minor comments (mainly for naming stuff and docstrings).

cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
```diff
@@ -41,21 +46,95 @@ type MigInfoProvider interface {
...
 }
 
+// GetMigForInstance returns MIG ref for a given instance
+func (c *cachingMigInfoProvider) GetMigForInstance(instanceRef GceRef) (Mig, error) {
+	mig, found, err := c.getCachedMigForInstance(instanceRef)
+	if found {
+		return mig, err
+	}
+
+	foundCandidate, err := c.fillCandidateMigInstances(instanceRef)
```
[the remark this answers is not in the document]
BigDarkClown: That's a really good suggestion, done.

cluster-autoscaler/cloudprovider/gce/cache.go
BigDarkClown: I've added the line back, did not mean to remove it in the first place.

cluster-autoscaler/cloudprovider/gce/mig_info_provider_test.go
BigDarkClown: Done. I had to make a function name longer because of a conflict with `TestGetMigForInstance` from `autoscaling_gce_client_test.go`.

cluster-autoscaler/cloudprovider/gce/mig_info_provider_test.go
BigDarkClown: I think we have some more refactors for that to be possible - still there are some client functions which are called by manager directly. But a bit more and I think we could stop mocking both client and cache, and instead mock only various providers.

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -147,151 +136,80 @@ func (gc *GceCache) GetMigs() []Mig {
...
+	klog.V(5).Infof("Instances to migs cache invalidated")
+	gc.instancesToMig = make(map[GceRef]GceRef)
+	gc.instancesFromUnknownMig = make(map[GceRef]bool)
+}
+
+func (gc *GceCache) removeMigInstances(migRef GceRef) {
+	for instanceRef, instanceMigRef := range gc.instancesToMig {
+		if migRef == instanceMigRef {
+			delete(gc.instancesToMig, instanceRef)
+			delete(gc.instancesFromUnknownMig, instanceRef)
```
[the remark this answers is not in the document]
MaciekPytel: Agreed, makes perfect sense to keep delete just in case we missed something. I'd consider adding a comment mentioning this so that someone reading the code in future doesn't have to wonder why this is here.

MaciekPytel (commented):
/lgtm
/approve

PR #4513, opened 2021-12-09
Re-add default expander

dbonfigli:
https://github.com/kubernetes/autoscaler/pull/4233  introduced the possibility to add multiple expanders but removed the default definition so now it must be explicitly defined otherwise a user gets the not user-friendly error:

```
F1209 16:45:32.229497       1 main.go:350] Failed to create autoscaler: Expander  not supported
```

This PR reintroduces a default expander, with value "random" as it was before, so that the `--expander` flag does not need to be explicitly defined if a user does not care about it.


gjtempleton:
Thanks for catching this, this brings the functionality back to what the docs already claim, so:

/lgtm

For final approval:
/assign @towca

feiskyer (approved):
LGTM

PR #4514, opened 2021-12-10
Use DescribeAutoscalingGroups call

uthark:


gjtempleton:
Hi @uthark , thanks for the placeholder PR, this looks very similar to the work already done in #4424 and #4506 though, or am I missing some extra context here?

uthark:
@gjtempleton Yeah, looks like dupe.

PR #4519, opened 2021-12-13
Introduce the scale down processor that picks the final scale down candidates

kisieland:
This PR introduces a new processor (proposed name is: `ScaleDownSetProcessor`), that is used at ScaleDown operations to pick a final set of candidates (after sanity checks size, pods etc.).
This may be useful to implement a variety of features, for example, users can introduce a processor that has a specific priorities for which node group to scale down first.

In this PR I also simplified the `FindNodesToRemove` function to not include the `maxCount` and `fastCheck` parameters.
The latter one is no longer needed due to the Fast and Detailed functions being backed by the same checks.
Rather than return a limited number of candidates the function returns a slice with all possible (so that the processor can pick a best suited one).

The PR also includes a default implementation of the processor that picks first x candidates (current behaviour).

### DIFF
--- a/cluster-autoscaler/core/scale_down.go
+++ b/cluster-autoscaler/core/scale_down.go
@@ -509,11 +509,11 @@ func (sd *ScaleDown) UpdateUnneededNodes(
 		klog.V(1).Infof("Scale-down calculation: ignoring %v nodes unremovable in the last %v", skipped, sd.context.AutoscalingOptions.UnremovableNodeRecheckTimeout)
 	}
 
-	emptyNodesList := sd.getEmptyNodesNoResourceLimits(currentlyUnneededNodeNames, len(currentlyUnneededNodeNames), timestamp)
+	emptyNodesToRemove := sd.getEmptyNodesToRemoveNoResourceLimits(currentlyUnneededNodeNames, timestamp)
 
 	emptyNodes := make(map[string]bool)
-	for _, node := range emptyNodesList {
-		emptyNodes[node.Name] = true
+	for _, empty := range emptyNodesToRemove {
+		emptyNodes[empty.Node.Name] = true
 	}
 
 	currentlyUnneededNonEmptyNodes := make([]string, 0, len(currentlyUnneededNodeNames))
@@ -535,11 +535,9 @@ func (sd *ScaleDown) UpdateUnneededNodes(
 	nodesToRemove, unremovable, newHints, simulatorErr := simulator.FindNodesToRemove(
 		currentCandidates,
 		destinations,
-		nil,
+		sd.context.ListerRegistry,
 		sd.context.ClusterSnapshot,
 		sd.context.PredicateChecker,
-		len(currentCandidates),
-		true,
 		sd.podLocationHints,
 		sd.usageTracker,
 		timestamp,
@@ -567,17 +565,18 @@ func (sd *ScaleDown) UpdateUnneededNodes(
 			simulator.FindNodesToRemove(
 				currentNonCandidates[:additionalCandidatesPoolSize],
 				destinations,
-				nil,
+				sd.context.ListerRegistry,
 				sd.context.ClusterSnapshot,
 				sd.context.PredicateChecker,
-				additionalCandidatesCount,
-				true,
 				sd.podLocationHints,
 				sd.usageTracker,
 				timestamp,
 				pdbs)
 		if simulatorErr != nil {
 			return sd.markSimulationError(simulatorErr, timestamp)
+		}
+		if len(additionalNodesToRemove) > additionalCandidatesCount {
+			additionalNodesToRemove = additionalNodesToRemove[:additionalCandidatesCount]
 		}
 		nodesToRemove = append(nodesToRemove, additionalNodesToRemove...)
 		unremovable = append(unremovable, additionalUnremovable...)
@@ -586,8 +585,8 @@ func (sd *ScaleDown) UpdateUnneededNodes(
 		}
 	}
 
-	for _, node := range emptyNodesList {
-		nodesToRemove = append(nodesToRemove, simulator.NodeToBeRemoved{Node: node, PodsToReschedule: []*apiv1.Pod{}})
+	for _, empty := range emptyNodesToRemove {
+		nodesToRemove = append(nodesToRemove, simulator.NodeToBeRemoved{Node: empty.Node, PodsToReschedule: []*apiv1.Pod{}})
 	}
 
 	// Update the timestamp map.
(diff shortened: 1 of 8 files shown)


cluster-autoscaler/simulator/cluster.go
```diff
@@ -117,79 +117,56 @@ func FindNodesToRemove(
 	listers kube_util.ListerRegistry,
 	clusterSnapshot ClusterSnapshot,
 	predicateChecker PredicateChecker,
-	maxCount int,
-	fastCheck bool,
```
towca, line 121: How are Fast and Detailed backed by the same checks? Fast calls `GetPodsForDeletionOnNodeDrain` with `checkReferences=false`, while Detailed calls it with `checkReferences=true`. `checkReferences` determines whether `GetPodsForDeletionOnNodeDrain` does GETs on pod controllers. I'm not sure what the actual difference in performance this implies, but it looks like a material change - or are you saying that the GETs don't matter?

cluster-autoscaler/core/scale_down.go
```diff
@@ -567,27 +566,26 @@ func (sd *ScaleDown) UpdateUnneededNodes(
...
-				additionalCandidatesCount,
-				true,
 				sd.podLocationHints,
 				sd.usageTracker,
 				timestamp,
 				pdbs)
 		if simulatorErr != nil {
 			return sd.markSimulationError(simulatorErr, timestamp)
 		}
+		additionalNodesToRemove = sd.processors.ScaleDownSetProcessor.GetNodesToRemove(additionalNodesToRemove, additionalCandidatesCount)
```
towca, line 579: I think that inside  `UpdateUnneededNodes`, `maxCount` was supposed not to actually limit anything, since it's equal or higher than the list of candidates that `FindNodesToRemove` iterates over. So this call wouldn't actually be needed, unless the processor wants to return less nodes than it can.

Which brings me to my second point - by putting the processor call here (and inside `getEmptyNodesToRemoveNoResourceLimits`), we're allowing the processor to arbitrarily limit the number of nodes returned by `FindNodesToRemove` inside `UpdateUnneededNodes`, even if there are fewer than `maxCount` nodes passed to it. It's not clear to me if this won't introduce any unexpected issues e.g. with `TryToScaleDown` later on, since it's so tightly coupled with  `UpdateUnneededNodes`.

Is there a use case for calling the same method in two completely different contexts (looking for unneeded nodes and actually choosing nodes to scale down)? The method wouldn't be able to distinguish between the two. If there is a use case, I'd at least introduce two different methods to the processor. But I still think it'd require a bit more thought about how `UpdateUnneededNodes` interacts with `TryToScaleDown`.

cluster-autoscaler/core/scale_down.go
```diff
@@ -1086,35 +1083,34 @@ func (sd *ScaleDown) getEmptyNodes(candidates []string, maxEmptyBulkDelete int,
...
+			})
 		}
 	}
-	limit := maxEmptyBulkDelete
-	if len(result) < limit {
-		limit = len(result)
-	}
-	return result[:limit]
+
+	return sd.processors.ScaleDownSetProcessor.GetNodesToRemove(nodesToRemove, maxEmptyBulkDelete)
```
towca, line 1092: Why does this method call the processor inside itself, but `FindNodesToRemove` doesn't? If the purpose of the processor is to pick final scale-down candidates, I'd expect it to be called in exactly two places - ~right before `scheduleDeleteEmptyNodes`, and ~right before `deleteNode` in `TryToScaleDown`.

cluster-autoscaler/simulator/cluster.go
x13n: I don't think this is actually generating any GET requests, CA is only fetching cached objects from listers wheck `checkReferences=true`. To be honest I'm not sure if that flag makes sense anymore.

cluster-autoscaler/core/scale_down.go
kisieland: You are right about those calls inside of `UpdateUnneededNodes` refactored the code to use the `GetNodesToRemove` only just before the scaledowns. So there is no need for the second method.

cluster-autoscaler/core/scale_down.go
kisieland: Moved the calls to the `GetNodesToRemove` just before the actual deletion.

cluster-autoscaler/simulator/cluster.go
kisieland: The listers are backed up by the cache so there should not be a noticeable number of new GET calls.
I did run preliminary 5k test to verify and I did not see any regressions.

cluster-autoscaler/core/scale_down.go
kisieland: Although, when I looked at the tests it seems this filtering out is needed as it may return more than `additionalCandidatesCount` candidates. This part of the logic aims to limit the number of considered nodes to the `ScaleDownNonEmptyCandidatesCount` limit.
I would say that we should not call the `GetNodesToRemove` here as it picks best candidates and we should take just prefix as in previous version.

cluster-autoscaler/simulator/cluster.go
kisieland: I've re-run the tests with current version and still did not find any increase in number of GAT calls.

towca: Ok, thanks for verifying that! We should get rid of the parameter altogether then, but that's probably out-of-scope of this PR.

cluster-autoscaler/core/scale_down.go
towca: Thanks, reads much better now!

towca (approved):
Thanks for addressing the comments, looks very clean now! I just have one question, but let's not block on answering.

MaciekPytel:
Sorry, I know I'm late to the party, but I wanted to ask if we can have a quick follow up to add AutoscalingContext as a parameter to GetNodesToRemove()? My reasons are:
1. Consistency. Every processor function in CA takes context as first parameter.
2. Not having access to a context is potentially very limiting. With the current interface the GetNodesToRemove() can only look at the candidate nodes and compare them, but it has no access to the view of the entire cluster. An example use case would be balancing zone sizes in scale-down: that processor is an obvious candidate to implement such logic, but without access to all nodes I don't know what the zone sizes are.

I think it should be a very simple change, passing the context to TryToScaleDown() and the processor call. WDYT @towca @kisieland

towca:
Oh yes, that's a very good idea that I missed, thanks Maciek. @kisieland are you up for adding that in a follow-up?

kisieland:
Sure, I will do it next week.

towca:
Thank you!

cluster-autoscaler/core/scale_down.go
```diff
@@ -967,6 +965,7 @@ func (sd *ScaleDown) TryToScaleDown(
 		scaleDownStatus.Result = status.ScaleDownError
 		return scaleDownStatus, err.AddPrefix("Find node to remove failed: ")
 	}
+	nodesToRemove = sd.processors.ScaleDownSetProcessor.GetNodesToRemove(nodesToRemove, 1)
```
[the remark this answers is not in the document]
kisieland: > I assume your scalability tests measured scale-down speed at scale, right?

Yes, this was a scale-down test, both for empty and underutilised nodes.

I also went through the metrics of scale down loop and did not find a noticeable changes in the times.

PR #4520, opened 2021-12-13
CA - AWS - Update Static Instance List

gjtempleton:
Re-running AWS static instance list generation post merging of #4468 - run as if in `us-east-1` region to capture as many instance types as possible.

This results in the deletion of non-existent instance types which were placeholders for the family (e.g. `c4` and `a1`) but didn't represent real instance types. Also corrects some faulty data (e.g. number of GPUs in `dl1.24xlarge` instances and adds in a number of new instance types.


mwielgus (approved):
/lgtm
/approve

PR #4524, opened 2021-12-14
Charts - CA - Default to system-cluster-critical priority class

gjtempleton:
Sets the default priority class of the CA to `system-cluster-critical`. We should default to safe behaviour, and risking the CA being deployed by default without a priority class risks users having their CAs be evicted and not being able to be rescheduled, thus losing scaling in their clusters is not a safe default.

This matches the examples for a number of cloud providers already in this repo, e.g. [AWS](https://github.com/kubernetes/autoscaler/blob/9e63892f86a389b337dc77558df433921237254c/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-one-asg.yaml#L140) and [AliCloud](https://github.com/kubernetes/autoscaler/blob/2542e8c884a8e25634d3b8f43243fa8706007f30/cluster-autoscaler/cloudprovider/alicloud/README.md).

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -17,4 +17,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.10.9
+version: 9.11.0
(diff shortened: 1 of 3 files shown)


gjtempleton:
@stevehipwell would appreciate your eyes on this if you have the time?

stevehipwell:
@gjtempleton I completely agree with your summary and solution.

mwielgus (approved):
/lgtm
/approve

PR #4541, opened 2021-12-20
improve logging for scale set size changes

marwanad:
This adds logging around previous + refreshed VMSS size and lowers the verbosity for one other log.

It also removes some fields that are no longer used after #3717 move to global azure_manager cache

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
@@ -50,10 +50,8 @@ type ScaleSet struct {
 	minSize int
 	maxSize int
 
-	sizeMutex         sync.Mutex
-	curSize           int64
-	lastSizeRefresh   time.Time
-	sizeRefreshPeriod time.Duration
+	sizeMutex sync.Mutex
+	curSize   int64
 
 	instancesRefreshPeriod time.Duration
 	instancesRefreshJitter int
@@ -118,7 +116,7 @@ func (scaleSet *ScaleSet) Autoprovisioned() bool {
 func (scaleSet *ScaleSet) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
 	template, err := scaleSet.getVMSSFromCache()
 	if err != nil {
-		return nil, err.Error()
+		return nil, err
 	}
 	return scaleSet.manager.GetScaleSetOptions(*template.Name, defaults), nil
 }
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -50,10 +50,8 @@ type ScaleSet struct {
 	minSize int
 	maxSize int
 
-	sizeMutex         sync.Mutex
-	curSize           int64
-	lastSizeRefresh   time.Time
-	sizeRefreshPeriod time.Duration
```
marwanad, line 56: The `sizeRefreshPeriod` was always 0 after #3717. The TTL for VMSS cache refreshes is in honoured there where we depend on the CA core `cloudprovider.Refresh()` calls to update the caches.

https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/azure/azure_manager.go#L166-L168

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -142,35 +140,30 @@ func (scaleSet *ScaleSet) getCurSize() (int64, error) {
...
 
 	vmssSizeMutex.Lock()
 	curSize := *set.Sku.Capacity
 	vmssSizeMutex.Unlock()
 
-	klog.V(5).Infof("Getting scale set (%q) capacity: %d\n", scaleSet.Name, curSize)
-
 	if scaleSet.curSize != curSize {
 		// Invalidate the instance cache if the capacity has changed.
+		klog.V(5).Infof("VMSS %q size changed from: %d to %d, invalidating instance cache", scaleSet.Name, scaleSet.curSize, curSize)
```
marwanad, line 161: keeping it as v=5 intentionally because it can be inferred from L64.

cluster-autoscaler/cloudprovider/azure/azure_scale_set_test.go
```diff
@@ -138,12 +136,9 @@ func TestIncreaseSize(t *testing.T) {
...
 	assert.NoError(t, err)
 
-	ss := newTestScaleSet(provider.azureManager, "test-asg")
-	ss.lastSizeRefresh = time.Now()
-	ss.sizeRefreshPeriod = 1 * time.Minute
-	ss.curSize = -1
+	ss := newTestScaleSet(provider.azureManager, "test-asg-doesnt-exist")
 	err = ss.IncreaseSize(100)
-	expectedErr := fmt.Errorf("the scale set test-asg is under initialization, skipping IncreaseSize")
+	expectedErr := fmt.Errorf("could not find vmss: test-asg-doesnt-exist")
```
marwanad, line 141: technically with the changes in #3717, we'll never end with the "-1" initialization case.

PR #4543, opened 2021-12-20
Cache Mig instances in GceCache

BigDarkClown:
I've added caching of the Mig instances to the `GceCache` + utilised the existing logic from `MigInfoProvider` to update and refresh this cache. This way we pass all the Mig info-related data from `GceClient` to `GceManager` exclusively via `MigInfoProvider`. This gives us better control over the client and cache, while also improving the separation of these components.

Additionally we may utilise the Mig instance cache in the future. It is refreshed once per loop, so any new functionalities that would require fetching instances will now no longer increase the QPS.

It is worth noting that we do not guarantee that `instances` and `instanceToMig` caches are in sync. They are modified independently and at different times, so it is possible that they get out of sync for a couple loops. This is mostly caused by `InvalidateXYZ` methods, which are used to force the refresh of cache.

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/gce_manager.go
+++ b/cluster-autoscaler/cloudprovider/gce/gce_manager.go
@@ -283,11 +283,12 @@ func (m *gceManagerImpl) GetMigForInstance(instance GceRef) (Mig, error) {
 
 // GetMigNodes returns mig nodes.
 func (m *gceManagerImpl) GetMigNodes(mig Mig) ([]cloudprovider.Instance, error) {
-	return m.GceService.FetchMigInstances(mig.GceRef())
+	return m.migInfoProvider.GetMigInstances(mig.GceRef())
 }
 
 // Refresh triggers refresh of cached resources.
 func (m *gceManagerImpl) Refresh() error {
+	m.cache.InvalidateAllMigInstances()
 	m.cache.InvalidateAllMigTargetSizes()
 	m.cache.InvalidateAllMigBasenames()
 	m.cache.InvalidateAllMigInstanceTemplateNames()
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/cloudprovider/gce/gce_manager.go
```diff
@@ -283,11 +283,12 @@ func (m *gceManagerImpl) GetMigForInstance(instance GceRef) (Mig, error) {
 
 // GetMigNodes returns mig nodes.
 func (m *gceManagerImpl) GetMigNodes(mig Mig) ([]cloudprovider.Instance, error) {
-	return m.GceService.FetchMigInstances(mig.GceRef())
+	return m.migInfoProvider.GetMigInstances(mig.GceRef())
```
MaciekPytel, line 286: That makes a lot of sense, but the caching slightly changes the contract for this method (getting fresh vs cached list). I think it's fine for the current CA implementation, as long as the cache is refreshed each loop (as opposed to, for example, in forceRefresh()). Your implementation is correct in that sense, but you need to be increasingly aware what is cached for how long. 
Maybe consider adding a comment somewhere (not sure what is the best place) about expected lifetime of all the caches? WDYT?

MaciekPytel:
/lgtm
/approve
/hold
Adding hold for a single comment. Please feel free to cancel if you disagree or prefer to do it in a separate PR.

cluster-autoscaler/cloudprovider/gce/gce_manager.go
BigDarkClown: Yeah, I think a comment above the `GceCache` would be a good place for that info. There are some other caches that are dependant on the same logic - base names, target sizes and instance template names are also maintained in cache and refreshed after each loop.

Issue #4546, opened 2021-12-21
VPA e2e tests should use images from official repos


jbartosik:
VPA e2e OOM tests are using image `gcr.io/jbartosik-gke-dev/stress:0.10` ([v1](https://github.com/kubernetes/autoscaler/blob/259c1acdeed58047980a772343af60ab6b695225/vertical-pod-autoscaler/e2e/v1/autoscaling_utils.go#L65), [v1beta2](https://github.com/kubernetes/autoscaler/blob/259c1acdeed58047980a772343af60ab6b695225/vertical-pod-autoscaler/e2e/v1beta2/autoscaling_utils.go#L65)), we should move the image from my dev project (it already caused problems in the past).

Steps:
* Build the image and put it in an official repo,
* Change tests to use the new image.

PR #4547, opened 2021-12-21
Add a Makefile for building image VPA uses in e2e tests

jbartosik:
Related to issue #4546


jbartosik:
@piotrnosek please take a look

kgolab:
This LGTM but I'd like to double-check if we could use gcr.io/google-containers/stress:v1 (referenced from Dockerfile) directly - it would be one indirection layer less.

jbartosik:
I'm doing #4556 instead

PR #4550, opened 2021-12-21
ignore azure csi topology label for similarity checks and populate it for scale from zero

marwanad:
For balancing similar nodegroups, this label should be ignored when comparing nodegroups that may be in different zones.

This also adds the label to the set of zonal labels predicted by azure template for the scale from zero case.


andyzhangx (approved):
/lgtm

mwielgus (approved):
/lgtm
/approve

PR #4552, opened 2021-12-22
Adding support for Debugging Snapshot

### LINKED ISSUE #4346: Proposal: Cluster Autoscaler Debugging Snapshot
## Cluster Autoscaler Debugging Snapshot [Toil Reduction]

Author:
- Jayant Jain (@jayantjain93)

## Introduction
With the growing number of large, autoscaled clusters we are increasingly dealing with complex cases that are very hard to debug. One major difficulty is that we log information about what decision Cluster Autoscaler (CA) took as well as the results of various intermediate steps that lead to this decision, but we don't log the data this decision was based on. The reason for this is that CA is internally simulating the behavior of the entire cluster and to fully understand any given decision we need to know the exact state of relevant nodes and pods (possibly all nodes and pods in the cluster) as well as certain k8s objects. The volume of logging required to capture all that data would be prohibitive.

This document proposed introducing a new "snapshot" feature to Cluster Autoscaler. This feature would allow an engineer debugging an autoscaler issue to manually trigger CA to dump its internal state.

## Proposal
The snapshot tool will use a manual trigger mechanism using **HTTP endpoint** which will collect the debugging information in the following run-cycle of CA and return the information as a **JSON-parsable HTTP response**.
Screenshot 2021-09-20 at 7 34 02 PM

### Trigger
The debugging snapshot is captured only when it receives a trigger from the user. Manual trigger is used instead of automated data collection is because the process of capturing the snapshot can be long in a large cluster thereby affecting the performance of CA and increase in latency. Using the HTTP request as a trigger. This is done by creating a HTTP endpoint in CA which would receive the trigger as an API call. This will allow passing of parameters for better extension of the trigger. It allows for easy return of error code if the request fails.

### Method of Formatting
The data in the snapshot will be very large and not all of the fields are always relevant. This poses the problem of choosing between what data we want to include and how readable the data be. We want to avoid a situation where we have a debugging snapshot of which fields are missing, but that may not always be possible.
The proposed way is to make the snapshot a **parsable json**.
All the relevant data collected to be encapsulated as a json field. The snapshot will contain all the elements that can be captured.
This file may not be easily human readable. But with some additional tooling (e.g. jq) to extract a readable format from this “full” snapshot. This should bridge the gap between how much data can be pushed in the snapshot and the readability of it. It also gives a quick turnaround time to create a “new readable” snapshot which won’t require any changes to the code or long waiting times for kubernetes releases. It also gives the ability to regenerate a new readable snapshot for older data.

### Data Collection and Workflow

There are two data points proposed (so far) to be collected.

1. **List of all NodeInfo**. This will be collected in static_autoscaler:RunOnce(), after we have added all the upcoming nodes. This will contain all the nodes, incl. properties not limited to resources, labels, taints. It will also have all pods scheduled on each of the nodes and all the properties related to each pod.

2. **List of all UnschedulablePodsCanBeScheduled**. They are collected in filter_out_schedulable:Process(). This list of Pods are what CA considers will fit on an existing / upcoming node(s), and hence does not consider it as part of the scale up.

The final list of data points and the location where they are set may change based on implementation details.

There are design decisions also made on how the interface will be used. We need to have an adaptive interface to capture all data fields and be extensible for each cloud provider.

There will be one interface in-core which cloud-providers will have to extend to add extra values from in-cloud-provider. And for each data field added, a new function is created in-cloud-provider interface with the correct data type of the data field.

```
type Snapshot_interface interface {
       func SetData1(Data1)
}

type Cloud_provider_snapshot_interface interface {    
    Snapshot_interface
    func SetData2(Data2)
    func SetData3(Data3)
}

// implementation of the interface
func (someClass *b) SetData2(Data2 d) {
          // do something
}
```
This is **actively trying to avoid** using a generic function using `interface{}` as an argument as:
`func AddExtraData(interface{} d) { }`

This is done to increase readability of the code and keep type-augmented functions.

jayantjain93:
Cluster-Autoscaler

#### What type of PR is this?
/kind feature

This PR introduces a new Debugging Tool, which would make it easier to analyse larger clusters.

Fixes #4346

NONE

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -23,6 +23,7 @@ import (
 
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/autoscaler/cluster-autoscaler/debuggingsnapshot"
 	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
 
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
@@ -119,7 +120,8 @@ func NewStaticAutoscaler(
 	cloudProvider cloudprovider.CloudProvider,
 	expanderStrategy expander.Strategy,
 	estimatorBuilder estimator.EstimatorBuilder,
-	backoff backoff.Backoff) *StaticAutoscaler {
+	backoff backoff.Backoff,
+	debuggingSnapshotter debuggingsnapshot.DebuggingSnapshotter) *StaticAutoscaler {
 
 	processorCallbacks := newStaticAutoscalerProcessorCallbacks()
 	autoscalingContext := context.NewAutoscalingContext(
@@ -130,7 +132,8 @@ func NewStaticAutoscaler(
 		cloudProvider,
 		expanderStrategy,
 		estimatorBuilder,
-		processorCallbacks)
+		processorCallbacks,
+		debuggingSnapshotter)
 
 	clusterStateConfig := clusterstate.ClusterStateRegistryConfig{
 		MaxTotalUnreadyPercentage: opts.MaxTotalUnreadyPercentage,
@@ -220,6 +223,8 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 	a.cleanUpIfRequired()
 	a.processorCallbacks.reset()
 	a.clusterStateRegistry.PeriodicCleanup()
+	a.DebuggingSnapshotter.StartDataCollection()
+	defer a.DebuggingSnapshotter.Flush()
 
 	unschedulablePodLister := a.UnschedulablePodLister()
 	scheduledPodLister := a.ScheduledPodLister()
@@ -407,6 +412,13 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 			klog.Errorf("Failed to add upcoming node %s to cluster snapshot: %v", upcomingNode.Node().Name, err)
 			return errors.ToAutoscalerError(errors.InternalError, err)
 		}
+	}
+
+	l, err := a.ClusterSnapshot.NodeInfos().List()
+	if err != nil {
+		klog.Errorf("Unable to fetch NodeInfo List for Debugging Snapshot, %v", err)
+	} else {
+		a.AutoscalingContext.DebuggingSnapshotter.SetNodeGroupInfo(l)
 	}
 
 	unschedulablePodsToHelp, _ := a.processors.PodListProcessor.Process(a.AutoscalingContext, unschedulablePods)
@@ -715,6 +727,7 @@ func (a *StaticAutoscaler) filterOutYoungPods(allUnschedulablePods []*apiv1.Pod,
 // ExitCleanUp performs all necessary clean-ups when the autoscaler's exiting.
 func (a *StaticAutoscaler) ExitCleanUp() {
 	a.processors.CleanUp()
+	a.DebuggingSnapshotter.Cleanup()
 
 	if !a.AutoscalingContext.WriteStatusConfigMap {
 		return
--- a/cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
+++ b/cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
@@ -1,0 +1,150 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package debuggingsnapshot
+
+import (
+	"encoding/json"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+	"k8s.io/kubernetes/pkg/scheduler/framework"
+)
+
+// NodeInfo captures a single entity of nodeInfo. i.e. Node specs and all the pods on that node.
+type NodeInfo struct {
+	Node *v1.Node             `json:"Node"`
+	Pods []*framework.PodInfo `json:"Pods"`
+}
+
+// DebuggingSnapshot is the interface used to define any debugging snapshot
+// implementation, incl. any custom impl. to be used by DebuggingSnapshotter
+type DebuggingSnapshot interface {
+	// SetNodeGroupInfo is a setter to capture all the NodeInfo
+	SetNodeGroupInfo([]*framework.NodeInfo)
+	// SetUnscheduledPodsCanBeScheduled is a setter for all pods which are unscheduled
+	// but they can be scheduled. i.e. pods which aren't triggering scale-up
+	SetUnscheduledPodsCanBeScheduled([]*v1.Pod)
+	// SetErrorMessage sets the error message in the snapshot
+	SetErrorMessage(string)
+	// SetEndTimestamp sets the timestamp in the snapshot,
+	// when all the data collection is finished
+	SetEndTimestamp(time.Time)
+	// SetStartTimestamp sets the timestamp in the snapshot,
+	// when all the data collection is started
+	SetStartTimestamp(time.Time)
+	// GetOutputBytes return the output state of the Snapshot with bool to specify if
+	// the snapshot has the error message set
+	GetOutputBytes() ([]byte, bool)
+	// Cleanup clears the internal data obj of the snapshot, readying for next request
+	Cleanup()
+}
+
+// DebuggingSnapshotImpl is the struct used to collect all the data to be output.
+// Please add all new output fields in this struct. This is to make the data
+// encoding/decoding easier as the single object going into the decoder
+type DebuggingSnapshotImpl struct {
+	NodeInfo                      []*NodeInfo `json:"NodeList"`
+	UnscheduledPodsCanBeScheduled []*v1.Pod   `json:"UnscheduledPodsCanBeScheduled"`
+	Error                         string      `json:"Error,omitempty"`
+	StartTimestamp                time.Time   `json:"StartTimestamp"`
+	EndTimestamp                  time.Time   `json:"EndTimestamp"`
+}
+
+// SetUnscheduledPodsCanBeScheduled is the setter for UnscheduledPodsCanBeScheduled
+func (s *DebuggingSnapshotImpl) SetUnscheduledPodsCanBeScheduled(podList []*v1.Pod) {
+	if podList == nil {
+		return
+	}
+
+	s.UnscheduledPodsCanBeScheduled = nil
+	for _, pod := range podList {
+		s.UnscheduledPodsCanBeScheduled = append(s.UnscheduledPodsCanBeScheduled, pod)
+	}
+}
+
+// SetNodeGroupInfo is the setter for Node Group Info
+// All filtering/prettifying of data should be done here.
+func (s *DebuggingSnapshotImpl) SetNodeGroupInfo(nodeInfos []*framework.NodeInfo) {
+	if nodeInfos == nil {
+		return
+	}
+
+	var NodeInfoList []*NodeInfo
+
+	for _, n := range nodeInfos {
+		nClone := n.Clone()
+		node := nClone.Node()
+
+		nodeInfo := &NodeInfo{
+			Node: node,
+			Pods: nClone.Pods,
+		}
+
+		NodeInfoList = append(NodeInfoList, nodeInfo)
+	}
+	s.NodeInfo = NodeInfoList
+}
+
+// SetEndTimestamp is the setter for end timestamp
+func (s *DebuggingSnapshotImpl) SetEndTimestamp(t time.Time) {
+	s.EndTimestamp = t
+}
+
+// SetStartTimestamp is the setter for end timestamp
+func (s *DebuggingSnapshotImpl) SetStartTimestamp(t time.Time) {
+	s.StartTimestamp = t
+}
+
+// GetOutputBytes return the output state of the Snapshot with bool to specify if
+// the snapshot has the error message set
+func (s *DebuggingSnapshotImpl) GetOutputBytes() ([]byte, bool) {
+	errMsgSet := false
+	if s.Error != "" {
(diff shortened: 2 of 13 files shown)


cluster-autoscaler/debuggingsnapshot/debugging_snapshotter.go
```diff
@@ -0,0 +1,222 @@
...
+	// If the debugging snapshot is disabled return 404.
+	// This is a preventive measure in case the handler is listening without the flag
+	// Although in the main we don't listen if the flag is not set
+	if *d.State == SNAPSHOTTER_DISABLED {
+		w.WriteHeader(http.StatusNotFound)
+		return
+	}
+
+	// checks if the handler is in the correct State to accept a new snapshot request
+	if *d.State != LISTENING {
```
x13n, line 116: Checking the state should be done with the mutex locked. Right now there is a race condition between one request checking the state here and another request setting the state back to LISTENING below. In fact, the entire ResponseHandler function would probably be better off in a separate object that interacts with the snapshotter via functions. It would be the handler object then to control whether there are any requests in flight. Snapshotter interface could then have a trigger function accepting a channel that would then be used to return (a pointer to) the result.

cluster-autoscaler/debuggingsnapshot/debugging_snapshotter.go
```diff
@@ -0,0 +1,222 @@
...
+		Mutex:             &sync.Mutex{},
+		DebuggingSnapshot: &DebuggingSnapshotImpl{},
+		FlushTrigger:      make(chan bool),
+	}
+}
+
+// ResponseHandler is the impl for request handler
+func (d *DebuggingSnapshotterImpl) ResponseHandler(w http.ResponseWriter, r *http.Request) {
+	// If the debugging snapshot is disabled return 404.
+	// This is a preventive measure in case the handler is listening without the flag
```
x13n, line 108: nit: I don't think this kind of defensive programming is necessary.

cluster-autoscaler/debuggingsnapshot/debugging_snapshotter.go
```diff
@@ -0,0 +1,222 @@
...
+// before doing extra processing.
+// e.g. If you want to pre-process a particular State in cloud-provider for snapshot
+// you should check this func in the loop before doing that extra processing
+func (d *DebuggingSnapshotterImpl) IsDataCollectionAllowed() bool {
+	statesAllowed := map[DebuggingSnapshotterState]bool{
+		DATA_COLLECTED:        true,
+		START_DATA_COLLECTION: true,
+	}
+
+	return statesAllowed[*d.State]
```
x13n, line 158: nit: Having a map just to compare one value with two constants seems to be an overkill.

cluster-autoscaler/debuggingsnapshot/debugging_snapshot_test.go
```diff
@@ -0,0 +1,86 @@
...
+	var nodeGroups []*framework.NodeInfo
+	nodeGroups = append(nodeGroups, nodeInfo)
+	nodeGroups[0].SetNode(node)
+	timestamp := time.Now().In(time.UTC)
+	snapshot.SetNodeGroupInfo(nodeGroups)
+	snapshot.SetTimestamp(timestamp)
+	op, err := snapshot.GetAll()
+	assert.False(t, err)
+
+	gkeSnapshot := &DebuggingSnapshotImpl{
```
x13n, line 67: Why is this called gkeSnapshot?

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -715,6 +727,7 @@ func (a *StaticAutoscaler) filterOutYoungPods(allUnschedulablePods []*apiv1.Pod,
 // ExitCleanUp performs all necessary clean-ups when the autoscaler's exiting.
 func (a *StaticAutoscaler) ExitCleanUp() {
 	a.processors.CleanUp()
+	a.DebuggingSnapshotter.Cleanup()
```
x13n, line 730: `Cleanup()` resets the underlying snapshot, which isn't really necessary during CA shutdown. At the same time, it is not doing anything about a HTTP request that might be hanging.

cluster-autoscaler/debuggingsnapshot/debugging_snapshot_test.go
```diff
@@ -0,0 +1,86 @@
...
+			{
+				Node: node,
+				Pods: pod,
+			},
+		},
+		Timestamp: timestamp,
+	}
+	expected, err1 := json.Marshal(gkeSnapshot)
+	assert.NoError(t, err1)
+	assert.EqualValues(t, expected, *op)
```
x13n, line 78: This test is essentially comparing json.Marshal output with itself. Maybe instead it would be more useful to Unmarshal op into a `map[string]interface{}` and check if it contains the fields you expect it to contain?

cluster-autoscaler/debuggingsnapshot/debugging_snapshotter.go
```diff
@@ -0,0 +1,222 @@
+/*
```
x13n, line 1: This file the core part of this PR. Please add some unit tests to confirm it is working as intended.

x13n (changes requested):

cluster-autoscaler/debuggingsnapshot/debugging_snapshot_test.go
jayantjain93: changed

cluster-autoscaler/debuggingsnapshot/debugging_snapshot_test.go
jayantjain93: done

cluster-autoscaler/debuggingsnapshot/debugging_snapshotter.go
x13n: You don't need if-else, just `return *d.State == START_DATA_COLLECTION || *d.State == DATA_COLLECTED`. Up to you though.

cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
```diff
@@ -0,0 +1,150 @@
...
+
+	// this error captures if the snapshot couldn't be marshalled, hence we create a new object
+	// and return the error message
+	if err != nil {
+		klog.Error("Unable to json marshal the debugging snapshot,", err)
+		errorSnapshot := DebuggingSnapshotImpl{}
+		errorSnapshot.SetErrorMessage("Unable to marshal the snapshot, " + err.Error())
+		errorSnapshot.SetEndTimestamp(s.EndTimestamp)
+		errorSnapshot.SetStartTimestamp(s.StartTimestamp)
+		errorMarshal, _ := json.Marshal(errorSnapshot)
```
x13n, line 134: What if this marshalling fails as well? At the very least we should log an error.

cluster-autoscaler/debuggingsnapshot/debugging_snapshot_test.go
```diff
@@ -0,0 +1,89 @@
...
+	timestamp := time.Now().In(time.UTC)
+	snapshot.SetNodeGroupInfo(nodeGroups)
+	snapshot.SetEndTimestamp(timestamp)
+	op, err := snapshot.GetOutputBytes()
+	assert.False(t, err)
+
+	var parsed map[string]interface{}
+	er := json.Unmarshal(op, &parsed)
+	assert.NoError(t, er)
+	pNodeInfo := parsed["NodeList"].([]interface{})[0].(map[string]interface{})
```
x13n, line 70: Can you check if these type assertions are ok so that if they fail, we can get a test error instead of panic?

cluster-autoscaler/debuggingsnapshot/debugging_snapshotter.go
```diff
@@ -0,0 +1,225 @@
...
+	klog.Infof("UnscheduledPodsCanBeScheduled is being set for the debugging snapshot")
+	d.DebuggingSnapshot.SetUnscheduledPodsCanBeScheduled(podList)
+	*d.State = DATA_COLLECTED
+}
+
+// Cleanup clears the internal data sets of the cluster
+func (d *DebuggingSnapshotterImpl) Cleanup() {
+	// The new data bean is set by value, to avoid changing the reference
+	d.DebuggingSnapshot.Cleanup()
+	d.Trigger <- false
```
x13n, line 224: This is going to hang until there's an incoming request. I'd suggest using `r.Context().WithCancel()` in the request handler, `select{}` on both trigger and `ctx.Done()`, and just calling the cancel function here.

cluster-autoscaler/debuggingsnapshot/debugging_snapshotter.go
```diff
@@ -0,0 +1,222 @@
...
+type DebuggingSnapshotterImpl struct {
+	// State captures the internal state of the snapshotter
+	State *DebuggingSnapshotterState
+	// DebuggingSnapshot is the data bean for the snapshot
+	DebuggingSnapshot DebuggingSnapshot
+	// Mutex is the synchronisation used to the methods/states in the critical section
+	Mutex *sync.Mutex
+	// FlushTrigger is the channel on which the Response Handler waits on to know
+	// when there is data to be flushed back to the channel from the Snapshot
+	FlushTrigger chan bool
```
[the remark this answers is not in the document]
x13n: I think selector would be more readable (and would allow you to avoid a bug I just mentioned in another comment). If you want to pass states through this channel though, an enum would be easier to follow.

cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
```diff
@@ -0,0 +1,150 @@
...
+		errMsgSet = true
+	}
+
+	klog.Infof("debug snapshot flush ready")
+	marshalOutput, err := json.Marshal(s)
+
+	// this error captures if the snapshot couldn't be marshalled, hence we create a new object
+	// and return the error message
+	if err != nil {
+		klog.Error("Unable to json marshal the debugging snapshot,", err)
```
x13n, line 129: nit: missing space at the end of string. Also, the usual convention is to use a colon instead of a comma before embedding underlying error.

x13n (changes requested):

cluster-autoscaler/debuggingsnapshot/debugging_snapshotter.go
jayantjain93: done

cluster-autoscaler/debuggingsnapshot/debugging_snapshotter.go
```diff
@@ -0,0 +1,234 @@
...
+		*d.State = LISTENING
+		d.CancelRequest = nil
+
+		d.Mutex.Unlock()
+	case <-ctx.Done():
+		d.Mutex.Lock()
+		klog.Infof("Received terminate trigger, aborting ongoing snapshot request")
+		w.WriteHeader(http.StatusServiceUnavailable)
+
+		d.DebuggingSnapshot.Cleanup()
```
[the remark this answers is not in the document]
jayantjain93: Yes. It does happen within the debugging_snapshot:116, as part of `GetOutputBytes()`.
I think your opinion makes more sense. And adding it to snapshotter might make it more clear and consistent. Done

cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
```diff
@@ -0,0 +1,151 @@
...
+// the snapshot has the error message set
+func (s *DebuggingSnapshotImpl) GetOutputBytes() ([]byte, bool) {
+	defer s.Cleanup()
+	errMsgSet := false
+	if s.Error != "" {
+		klog.Errorf("Debugger Found with error message set when GetOutputBytes() is called. - ", s.Error)
+		errMsgSet = true
+	}
+
+	klog.Infof("debug snapshot flush ready")
```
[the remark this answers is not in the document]
jayantjain93: done

cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
```diff
@@ -0,0 +1,151 @@
...
+	s.StartTimestamp = t
+}
+
+// GetOutputBytes return the output state of the Snapshot with bool to specify if
+// the snapshot has the error message set
+func (s *DebuggingSnapshotImpl) GetOutputBytes() ([]byte, bool) {
+	defer s.Cleanup()
+	errMsgSet := false
+	if s.Error != "" {
+		klog.Errorf("Debugger Found with error message set when GetOutputBytes() is called. - ", s.Error)
```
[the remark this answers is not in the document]
jayantjain93: done

cluster-autoscaler/debuggingsnapshot/debugging_snapshotter.go
```diff
@@ -0,0 +1,238 @@
...
+	state := SNAPSHOTTER_DISABLED
+	if isDebuggerEnabled {
+		klog.Infof("Debugging Snapshot is enabled")
+		state = LISTENING
+	}
+	return &DebuggingSnapshotterImpl{
+		State:             &state,
+		Mutex:             &sync.Mutex{},
+		DebuggingSnapshot: &DebuggingSnapshotImpl{},
+		Trigger:           make(chan struct{}, 1),
```
x13n, line 105: I don't think this needs to be buffered anymore.

cluster-autoscaler/debuggingsnapshot/debugging_snapshotter_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	req := httptest.NewRequest(http.MethodGet, "/", nil)
+	w := httptest.NewRecorder()
+
+	go func() {
+		snapshotter.ResponseHandler(w, req)
+		wg.Done()
+	}()
+
+	for !snapshotter.IsDataCollectionAllowed() {
```
x13n, line 74: I know this is only a test, but let's avoid busy waiting and add some sleep() here (and in other tests below) instead of burning cpus.

cluster-autoscaler/debuggingsnapshot/debugging_snapshotter.go
jayantjain93: This needs to be buffered or else putting on the channel becomes a blocking call (instead of non-blocking) on when the channel is emptied. [this was evident in the test]

cluster-autoscaler/debuggingsnapshot/debugging_snapshotter_test.go
jayantjain93: I purposefully avoided using Sleep in these tests. (Did consider Sleep as an alternative). Since the whole test execution is in microseconds I don't really see any substantial gains here.

Also, this loop should ideally only execute once, since the conditional statement is waiting for a lock, that will be freed by ResponseHandler and it should directly be able to execute the command, inside the loop, when the conditional statement fails.

For the more indefinite / unknown time-period wait, I am using a Waitgroup anyways.

mwielgus (approved):
/lgtm
/approve

PR #4560, opened 2021-12-22
Introduce per-VPA minReplicas to allow overriding global setting.

### LINKED ISSUE #3986: VPA configuration to evict pods
**Which component are you using?**: vertical pod autoscaler

At the moment VPA can assign resources for pods during start up only. And it doesn't restart pods with single replica. There is a configuration option `--min-replicas` but it's not always available (e.g. on GKE). It would be nice to have an option to allow pods eviction even with single replica without necessity to configure VPA itself.

Cluster autoscaler supports pod annotation `cluster-autoscaler.kubernetes.io/safe-to-evict` so any specific pod can be marked as a safe to evict. VPA can respect this annotation as well. Another option can be either annotation to override `minReplicas` or VPA resource configuration like it was proposed in https://github.com/kubernetes/autoscaler/issues/1828

So possible solution can be one or all of:
1. VPA to respect `cluster-autoscaler.kubernetes.io/safe-to-evict` pod annotation;
2. VPA to support `vertical-pod-autoscaler.kubernetes.io/min-replicas: N` pod annotation;
3. VPA resource to support `minReplicas` parameter.

Create extra replica for every single pod. But it's just a waste of resources.

**Additional context.**:

kgolab:

vertical-pod-autoscaler (VPA API & VPA Updater, to be exact)

#### What type of PR is this?

/kind feature
/kind api-change

Addresses issue #3986 by adding a per-VPA-object setting, overriding global `--min-replicas` flag of VPA Updater.

API is extended in `spec.updatePolicy` as the setting applies to VPA Updater only.

Needed for managed VPA installations or to avoid setting global flag to an unsafe value.
Also allows to specify a more strict update policy, even in absence of PodDisruptionBudget.

/kind feature
/kind api-change

Fixes #3986

```release-note
Adds new API field: `spec.updatePolicy.minReplicas`.
If set it will override the global `--min-replicas` flag for this particular VPA object.
If not set the old behaviour remains and the global flag is used.
```


jbartosik (commented):
Code change look good to me.

Two things to do before I can approve:
- I think we need someone to check API changes we plan to make.
- Since we're changing API we should add e2e test(s).

kgolab:
@mwielgus , could you please take a look at the API change?

As described in the initial PR comment, the idea is to add a single field (MinReplicas) to PodUpdatePolicy struct.

I will update the initial comment with more info, including about testing that @jbartosik asked for, once I'm done with all the test-related changes.

vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml
```diff
@@ -383,6 +383,12 @@ spec:
                   pods. If not specified, all fields in the `PodUpdatePolicy` are
                   set to their default values.
                 properties:
+                  minReplicas:
+                    description: Overrides global '--min-replicas' flag, specifying
```
jbartosik, line 387: I'd change the order of information here.

Minimal number of replicas needed for Updater to attempt pod eviction for pods controlled by this VPA object. Overrides `--min-replicas` flag.

I expect some users of VPA API won't know about VPA flags.

vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml
```diff
@@ -383,6 +383,12 @@ spec:
...
+                      minimal number of replicas which need to be alive for Updater
+                      to attempt pod eviction (pending other checks like PDB).
```
jbartosik, line 389: Maybe we should add a validation to ensure that this is a positive value?  (I think 0 and negative values don't make sense).

If we do then we should document that

vertical-pod-autoscaler/e2e/v1/actuation.go
```diff
@@ -445,6 +493,33 @@ func testEvictsPods(f *framework.Framework, controller *autoscaling.CrossVersion
 	gomega.Expect(err).NotTo(gomega.HaveOccurred())
 }
 
+func testEvictsSingletonPod(f *framework.Framework, controller *autoscaling.CrossVersionObjectReference) {
```
jbartosik, line 496: Please rename or break into two helpers (setup & verify no eviction + update & verify eviction).

I wanted to suggest adding a test checking that we don't evict singleton pod unless configured to do it.

Then I read the body of this function and learned that we do test that.

`testEvictsSingletonPodOnlyAfterConfiguredTo` ?

vertical-pod-autoscaler/e2e/v1/common.go
```diff
@@ -321,6 +320,21 @@ func SetupVPAForNHamsters(f *framework.Framework, n int, cpu string, mode vpa_ty
...
+// PatchVpaUpdatePolicy modifies existing VPA object 'hamster-vpa' with new UpdatePolicy.
+func PatchVpaUpdatePolicy(f *framework.Framework, vpa *vpa_types.VerticalPodAutoscaler, policy *vpa_types.PodUpdatePolicy) {
+	newSpec := vpa.Spec.DeepCopy()
+	newSpec.UpdatePolicy = policy
+	bytes, err := json.Marshal([]patchRecord{{
+		Op:    "replace",
+		Path:  "/spec",
+		Value: *newSpec,
+	}})
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
```
jbartosik, line 335: Please provide a short description to make debugging this test easier if it fails. Something like "when marshalling VPA UpdatePolicy patch" will be enough

https://pkg.go.dev/github.com/onsi/gomega#Assertion

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go
```diff
@@ -203,10 +206,15 @@ func (f *podsEvictionRestrictionFactoryImpl) NewPodsEvictionRestriction(pods []*
 	creatorToSingleGroupStatsMap := make(map[podReplicaCreator]singleGroupStats)
 
 	for creator, replicas := range livePods {
+		// Use per-VPA minReplicas if present, fall back to the global setting.
```
jbartosik, line 209: I don't think we should do this in every iteration of the loop

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go
```diff
@@ -203,10 +206,15 @@ func (f *podsEvictionRestrictionFactoryImpl) NewPodsEvictionRestriction(pods []*
...
+		required := f.minReplicas
+		if vpa.Spec.UpdatePolicy != nil && vpa.Spec.UpdatePolicy.MinReplicas != nil {
+			required = int(*vpa.Spec.UpdatePolicy.MinReplicas)
+		}
 		actual := len(replicas)
-		if actual < f.minReplicas {
-			klog.V(2).Infof("too few replicas for %v %v/%v. Found %v live pods",
-				creator.Kind, creator.Namespace, creator.Name, actual)
+		if actual < required {
+			klog.V(2).Infof("too few replicas for %v %v/%v. Found %v live pods, needs %v",
```
jbartosik, line 216: It might be useful to log global vs override too

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction_test.go
```diff
@@ -60,11 +69,13 @@ func TestEvictReplicatedByController(t *testing.T) {
 	testCases := []struct {
 		replicas           int32
 		evictionTollerance float64
+		vpa                *vpa_types.VerticalPodAutoscaler
 		pods               []podWithExpectations
 	}{
 		{
 			replicas:           3,
 			evictionTollerance: 0.5,
+			vpa:                generateVpa(),
```
jbartosik, line 78: I'd just put `test.VerticalPodAutoscaler().WithContainer("any").Get()` here. I think we need `generatePod` because we want pods to have different names but we're not doing that with VPAs here

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction_test.go
```diff
@@ -223,6 +240,32 @@ func TestEvictReplicatedByController(t *testing.T) {
 				},
 			},
 		},
+		{
+			// Cannot evict a single Pod under default settings.
```
jbartosik, line 244: Even better would be to add `name` string to test case and then use that instead of index in the loop. Then:

- It works as a comment
- Makes it easy to find which test case failed

jbartosik (changes requested):
This is a big PR. I think the review will be easier if you make changes as new commits and then we squash at the end of the review (rather than squashing immediately and force-pushing)

vertical-pod-autoscaler/e2e/v1/common.go
kgolab: Removed this helper completely.

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go
kgolab: Added info here & separate log when choosing the value.

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go
kgolab: True, fixed.

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction_test.go
kgolab: Done.

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction_test.go
kgolab: OK, I moved newly added & pre-existing comments to TC name.

vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml
kgolab: Done.

vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml
kgolab: Good idea. I coupled CRD changes and validation into a separate commit.

vertical-pod-autoscaler/e2e/v1/actuation.go
kgolab: I split into two tests & moved to parametrised test suite (I didn't want to copy & paste 3 tests for each controller). PTAL.

kgolab:
Addressed all the comments with separate commits. PTAL.

jbartosik (approved):
One small suggestion.

Looks good to me. Please squash & I'll label so it can merge (unless you're waiting for more reviews; in that case please let me me know that I should wait).

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler.go
```diff
@@ -115,6 +115,11 @@ func validateVPA(vpa *vpa_types.VerticalPodAutoscaler, isCreate bool) error {
 		if _, found := possibleUpdateModes[*mode]; !found {
 			return fmt.Errorf("unexpected UpdateMode value %s", *mode)
 		}
+
+		minReplicas := vpa.Spec.UpdatePolicy.MinReplicas
+		if minReplicas != nil && *minReplicas <= 0 {
```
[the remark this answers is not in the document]
kgolab: Updated.

kgolab:
@jbartosik , squashed and rebased. I'm not sure why CA tests are still failing, the fix is already included.

jbartosik (changes requested):
Please rebase the PR on top of current master. PR currently has 4 commits, I think only 1 of them should be here, others look like artifacts of rebasing

kgolab:
Sorry for the extra commits. Rebased now so it's only the expected one.

jbartosik (approved):
/lgtm

PR #4567, opened 2021-12-28
Update dependencies to kubernetes 1.23

MaciekPytel:
Includes go.mod update, simple import fixes and moving to go 1.17. The move to 1.17 is required, because k8s itself has moved and the dependencies no longer compile using go1.16. 

Notably, the go version tag in go.mod is purposefully left at go1.16 (even though it's not what we actually use). This is needed, because go1.17 changes how go.mod work and the upstream kubernetes haven't transitioned to the new go.mod yet (they misrepresent go version in go.mod exactly the same way).
Reference: https://github.com/kubernetes/kubernetes/pull/105563#issuecomment-960915506

Also added go:build tags matching existing +build tags for all cluster-autoscaler files across different providers. This should have no effect, but it is a part of golang migration to new syntax and both tags is expected by go fmt in go1.17. Added verify-gofmt.sh exceptions for some autogenerated VPA files instead, as VPA shouldn't be built anyway from a CA release branch. Updated verify-boilerplate tooling to be compatible with new tags.


mwielgus (approved):
/lgtm
/approve

Issue #4230, opened 2021-07-29
[VPA] Ability to control amount of acceptable throttling


Niksko:
**Which component are you using?**:
VPA

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:
We have a number of workloads (cluster-autoscaler) being a good example, where we have to constantly tweak the request:limit ratio in order to avoid the workload being throttled. This work is toil, and we would like an automated way of handling this. These workloads are cluster critical, and as such, throttling is unacceptable.

**Describe the solution you'd like.**:
As proposed in [this old issue](https://github.com/kubernetes/autoscaler/issues/3091), some way of configuring the VPA to set limits such that a workload is not throttled at least some percentage of the time.

Something like:

```
apiVersion: "autoscaling.k8s.io/v1"
kind: VerticalPodAutoscaler
metadata:
  name: hamster-vpa
spec:
  targetRef:
    apiVersion: "apps/v1"
    kind: Deployment
    name: hamster
  resourcePolicy:
    containerPolicies:
      - containerName: '*'
        controlledResources: ["cpu", "memory"]
        noThrottlingTarget: 0.99
```

The `noThrottlingTarget: 0.99` instructs the VPA to choose CPU limits such that the container's CPU is throttled roughly 99% of the time.

**Describe any alternative solutions you've considered.**:
The current solution is to just set the request:limit ratio very high. e.g. `request: 100m`, `limit: 1`. However this feels hacky, because the goal of setting limits is to ensure that runaway processes don't exhaust the CPU of a node.

alex-berger:
I think this is still a valid feature request and should be reopened.

Issue #4135, opened 2021-06-11
Support custom metrics in Vertical Pod Autoscaler


Jeffwan:
We have a brief chat in this week's sig-autoscaling meeting. We think it's hard to find relationship between metrics and desired pod resources.  This issue is created to track story and attract people has similar requirements. 

**Which component are you using?**:
VPA

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:

Some domains like DB doesn't support horizontal scaling and the application throughput is major determined by the power of the instance. In this case, we need to scale up/down the individual pod.

Currently, VPA only  support scaling up/down the pod by leveraging container resource metrics (cpu, memory). With the custom metrics support, VPA can use application level metrics to make scaling decisions. 

**Describe the solution you'd like.**:

This is a very early idea and lots of code are highly binding to resource metrics. I will try to summarize a KEP to address the metrics support. 

**Describe any alternative solutions you've considered.**:

[Support Customized Recommenders for Vertical Pod Autoscalers](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa)

can be used together with custom metric. A custom recommender tailored for sepcified custom metrics is needed. 

**Additional context.**:

This is very early stage and this issue is used to track the process.

Issue #3975, opened 2021-03-25
VPA not producing reasonable recommendations on memory for Cronjobs


rdubya16:
**Which component are you using?**: VPA Recommender
**What version of the component are you using?**: 0.8.1
**What k8s version are you using (`kubectl version`)?**:

➜  vpa kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T18:43:01Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.9-eks-d1db3c", GitCommit:"d1db3c46e55f95d6a7d3e5578689371318f95ff9", GitTreeState:"clean", BuildDate:"2020-10-20T22:18:07Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}

**What environment is this in?**: EKS
**What did you expect to happen?**: Recommendations that make sense
**What happened instead?**: Recommendations dont match recorded metrics from prometheus

We are noticing after configuring VPA recommender for 100+ cronjobs the results we are getting dont seem to match the memory used by the pods. Many of our cronjobs are very short lived and ive found a few notes around that say it doesnt work well for jobs that run under a minute. I think this explains the majority of our cases but im still left with multiple examples of jobs that run for multiple minutes but their recommendations are much higher or lower than the expected values. 

One other thing I noticed is that multiple cronjobs all seem to get the exact same target memory value. Should this be possible? These are all different cronjobs that run different workloads. Not sure how they could end up with all the same target values. Is the issue here the cronjobs only run for a few minutes a day and there isnt enough data to work with?

```
628694953 job1
628694953 job2
628694953 job3
628694953 job4
628694953 job5
628694953 job6
628694953 job7
628694953 job8
628694953 job8
628694953 job9
628694953 job10
628694953 job11
628694953 job12

  name: job1
  namespace: default
  resourceVersion: "48748421"
  selfLink: /apis/autoscaling.k8s.io/v1/namespaces/default/verticalpodautoscalers/analytics-extracts-perform
  uid: 9e13d83f-87af-47c8-9e86-df99b0144818
spec:
  resourcePolicy:
    containerPolicies:
    - containerName: '*'
      controlledResources:
[...]

PR #4574, opened 2022-01-03
add TargetCPUPercentile to VerticalPodAutoscalerSpec

### LINKED ISSUE #4293: Make the target configurable
**Which component are you using?**: Recommender

The target is based on the percentile 90th. It is a good assumption, but sometimes there are some workloads we can be more aggressive about: `P85`, `P80`, ...

[Harness](https://docs.harness.io/article/ikxjmkqi03-recommendations) for instance has 2 standard profiles:
- Cost Optimized: for cpu it uses the `P50*1.15` (which is roughly equivalent to the `P65` in my own tests)
- Performance Optimized: uses the `P95`

The `VerticalPodAutoscaler` should allow to configure the percentile value used for the target (`P90` being the default) in the `ContainerResourcePolicy`:

```
kind: VerticalPodAutoscaler
metadata:
  name: my-app-vpa
spec:
  targetRef:
    apiVersion: "apps/v1"
    kind:       Deployment
    name:       my-app
  updatePolicy:
    updateMode: "Auto"
  resourcePolicy:
    containerPolicies:
    - containerName: "prometheus"
      targetCPUPercentile: 0.80
...
```

This would have the effect of changing the value set in the recommender [initialization](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/recommender/logic/recommender.go#L102).

The `RecommendedContainerResources` should display multiple percentiles, and not juste the `lowerBound`, `target` and `upperBound`. It could always display all the percentiles from 65 to 95, or display the few values that a field in the spec would list: `calculatedCPUPercentiles: [0.8, 0.85]`. Then the recommendation would look like:

```
...
  recommendation:
    containerRecommendations:
    - containerName: my-rec-container
      lowerBound:
        cpu: 25m
        memory: 262144k
      target:
        cpu: 200m
        memory: 262144k
      p80:
        cpu: 100m
        memory: 262144k
      p85:
        cpu: 150m
        memory: 262144k
      upperBound:
        cpu: 7931m
        memory: 8291500k
...
```

Based on this recommendation, the SRE could decide to switch to either the `P80` or the `P85`, depending on the workload profile (performance or cost optimized).

Use harness, or develop our own recommendation engine based on prometheus queries.

**Additional context.**:

matthyx:
vertical-pod-autoscaler

#### What type of PR is this?
/kind feature

Fixes #4293

```release-note
add TargetCPUPercentile to VerticalPodAutoscalerSpec to override default per VPA
```

```docs

```

### DIFF
--- a/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go
+++ b/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go
@@ -103,6 +103,11 @@ type VerticalPodAutoscalerSpec struct {
 	// recommendation) or contain exactly one recommender.
 	// +optional
 	Recommenders []*VerticalPodAutoscalerRecommenderSelector `json:"recommenders,omitempty" protobuf:"bytes,4,opt,name=recommenders"`
+
+	// Specifies the target CPU percentile to consider when computing recommendations.
+	// The default is 0.9
+	// *optional
+	TargetCPUPercentile float64 `json:"targetCPUPercentile,omitempty" protobuf:"bytes,5,rep,name=targetCPUPercentile"`
 }
 
 // PodUpdatePolicy describes the rules on how changes are applied to the pods.
--- a/vertical-pod-autoscaler/pkg/recommender/logic/estimator.go
+++ b/vertical-pod-autoscaler/pkg/recommender/logic/estimator.go
@@ -95,9 +95,13 @@ func (e *constEstimator) GetResourceEstimation(s *model.AggregateContainerState)
 
 // Returns specific percentiles of CPU and memory peaks distributions.
 func (e *percentileEstimator) GetResourceEstimation(s *model.AggregateContainerState) model.Resources {
+	cpuPercentile := e.cpuPercentile
+	if s.TargetCPUPercentile != nil {
+		cpuPercentile = *s.TargetCPUPercentile
+	}
 	return model.Resources{
 		model.ResourceCPU: model.CPUAmountFromCores(
-			s.AggregateCPUUsage.Percentile(e.cpuPercentile)),
+			s.AggregateCPUUsage.Percentile(cpuPercentile)),
 		model.ResourceMemory: model.MemoryAmountFromBytes(
 			s.AggregateMemoryPeaks.Percentile(e.memoryPercentile)),
 	}
--- a/vertical-pod-autoscaler/pkg/recommender/logic/estimator_test.go
+++ b/vertical-pod-autoscaler/pkg/recommender/logic/estimator_test.go
@@ -51,6 +51,7 @@ func TestPercentileEstimator(t *testing.T) {
 	CPUPercentile := 0.2
 	MemoryPercentile := 0.5
 	estimator := NewPercentileEstimator(CPUPercentile, MemoryPercentile)
+	dotNinety := 0.9
 
 	resourceEstimation := estimator.GetResourceEstimation(
 		&model.AggregateContainerState{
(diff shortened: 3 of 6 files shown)


matthyx:
@jbartosik I am adding a field to the CRD... do I need a KEP or something?

jbartosik:
> @jbartosik I am adding a field to the CRD... do I need a KEP or something?

Yes :)

To change API we need a KEP first. Since the VPA API is in our repo we're doing a little lighter process.

Our KEPs are in [vertical-pod-autoscaler/enhancements](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/enhancements). So far we have one submitted (and implemented) KEP and [one more in review](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa).

Writing a KEP is a bit of work so it might be a good idea to talk first.


A few things I think the KEP should address (probably worth discussing before writing the KEP):

1. I'd like to understand what use case(s) this intends to address.
2. "Target percentile" looks to me like it exposes an implementation detail of current VPA algorithm. I'd rather not expose implementation details in API for two reasons:
  - To be able to improve implementation in the future,
  - We now support alternative recommenders ([KEP](https://github.com/kubernetes/autoscaler/tree/23902caec6670ac8ebe939845ff36af8b5689045/vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa)) so any fields we add to the API should make sense for other recommenders too

jbartosik (changes requested):
I remembered one more thing to consider when writing the KEP.

There are some other ideas for changes to VPA algorithm[1] we should take a look at them to make sure the API change we'll make will fit with any other changes we will implement in the future.


[1]
* https://github.com/kubernetes/autoscaler/issues/4583
* https://github.com/kubernetes/autoscaler/issues/4293
* https://github.com/kubernetes/autoscaler/issues/4230
* https://github.com/kubernetes/autoscaler/issues/4135
* https://github.com/kubernetes/autoscaler/issues/3986
* https://github.com/kubernetes/autoscaler/issues/3975
* https://github.com/kubernetes/autoscaler/issues/3766

matthyx:
@jbartosik thanks for taking the time to discuss this :)
I work with @vsevel and this PR is the first step of an implementation for #4293
The higher level goal is to have workload profiles (normal, cost optimized, high performance, ...) that use a different CPU percentile to compute the CPU recommendations.

I agree that leaking implementation details might not be the best solution for the future evolutions, we based our analysis on [a feature of Harness](https://docs.harness.io/article/ikxjmkqi03-recommendations).

Actually, if [your KEP merges](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa), we could implement this use case with different recommenders...

jan-kantert:
We are interested in this as well. It would also be useful to have this for memory. A few services (i.e. prometheus) spike on memory on startup but generally do not need that much memory during runtime. Currently, we are over-reserving memory at factor 2 or 3 for some services.

jbartosik:
Sorry for slow reply.

I thought about this and talked this over. I see two ways forward with this (and similar requests for tuning VPA):

1. Using [customized recommenders](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa):
  1.1 Add flags that allow tuning VPA parameters like this PR does
  1.2 Add a flag that allows VPA to run under a different name than "" and "default"
  1.3 Add instructions for running two VPA recommender instances in one cluster, under different names
  1.4 Now you can run two VPA recommender instances in a cluster - "default" / "" and "lower cpu", each VPA object can specify which one to use
  1.5 Profit
2. Extending VPA API - this one is not so straight forward. We'd need to define some autoscaling profiles, add [enhancement](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/enhancements), extend API, use them in the default recommender to change percentiles we use.

@matthyx  @jan-kantert  what do you think?

matthyx:
> @matthyx @jan-kantert what do you think?

I'd say go for 1 !
If you agree I can work on it, if you provide some structure: do we need a KEP, separate issues, ...

matthyx:
Superceded by #4799

PR #4587, opened 2022-01-10
Update Cluster Autoscaler to go 1.17

MaciekPytel:
This is a partial cherry-pick of a #4567. 

Kubernetes has moved to go 1.17 and no longer compiles on go 1.16 (currently used by CA). This bumps CA to go1.17 and is a prerequisite to updating deps (which I'll do in a follow-up PR).
Some gotchas:
 * Go version tag in go.mod is purposefully left at go1.16. This is needed, because go1.17 changes how go.mod work and the upstream kubernetes haven't transitioned to the new go.mod yet (they misrepresent go version in go.mod exactly the same way). https://github.com/kubernetes/kubernetes/pull/105563#issuecomment-960915506
 * Added go:build tags matching existing +build tags for all cluster-autoscaler files across different providers. This should have no effect, but it is a part of golang migration to new syntax and both tags is expected by go fmt in go1.17.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,7 +14,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.16
+          go-version: 1.17.5
 
       - uses: actions/checkout@v2
         with:
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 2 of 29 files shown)


towca:
/lgtm
/hold

nit: It might be worth pointing out the difference in comments somewhere (e.g. add a comment to the Dockerfile saying that this should be the authoritative version and why it's different than go.mod + an opposite comment in go.mod). I know I've been using the version from go.mod as the authoritative one so far, others might be surprised as well.

MaciekPytel:
/hold cancel
@towca Added FAQ entry, can you review?

PR #4588, opened 2022-01-10
Support attribute-based instance selection for AWS

### LINKED ISSUE #4479: support AWS "Instance type requirements"

Cluster-Autoscaler

AWS added a new way of specifying instance types wanted.
they do this by specifying a set of instance characteristics - requirements.
which obsolete the need of specifying instance types, and in turn, breaks the CA when fetching the ASG info

CA needs to support not knowing which instance type is used by the ASG

CA will reed all the instance requirements from the ASG and calculate which instance type ASG will use when needed

**Additional context.**:

using CA v1.22.1 , error is:
`E1128 15:47:36.224041       1 static_autoscaler.go:286] Failed to get node infos for groups: ASG "......." uses the unknown EC2 instance type ""`

new settings in console looks:
Screen Shot 2021-11-28 at 17 50 31

AustinSiu:
#### What type of PR is this?

/kind feature

Support attribute-based instance selection using instance requirements.

This PR updates aws-sdk-go from v1.38.49 to v1.42.25 in order to pick up instance requirements feature changes.

Fixes #4479 

Testing done using an ASG w/ pre-existing instances:
1. Created an ASG w/ Spot and CapacityOptimized allocation strategy
2. Set InstanceRequirements Overrides w/ vcpu min: 10 max: 100
3. Update Cluster Autoscaler Min size to 1, Max 10
4. Waited for an instance to be added with Min 10 vcpu
5. Update InstanceRequirements Overrides to vcpu min: 2, max: 100
6. Cordon the 1st Node that came up in the ASG so that we can differentiate between nodes from step 2 vs step 5
7. Create a deployment with pods with resource request of 1 vcpu each
8. Scale deployment to 5 replicas
9. Expectation and observed: CA raises Desired Capacity to 6 under newer requirements, vs 1-2 expected nodes if under step 2 requirements 

```release-note
Added support for attribute-based instance type selection for AWS using available instance requirements.
```

```docs

```


AustinSiu:
@aleksandra-malinowska and @feiskyer may I ask if you are available to help review this PR?

AustinSiu:
@gjtempleton I haven't been able to get a response from the reviewers unfortunately. I've posted on the #pr-reviews Slack channel, but may I ask if there's anyone in particular I should reach out to for help reviewing?

feiskyer:
+AWS owners @Jeffwan @jaypipes @gjtempleton could you have a look?

cluster-autoscaler/cloudprovider/aws/aws_wrapper_test.go
```diff
@@ -64,11 +64,21 @@ type ec2Mock struct {
 	mock.Mock
 }
 
+func (e *ec2Mock) DescribeImages(input *ec2.DescribeImagesInput) (*ec2.DescribeImagesOutput, error) {
```
bwagner5, line 67: Can you update the IAM policy for the 2 new API calls?

cluster-autoscaler/cloudprovider/aws/aws_manager.go
```diff
@@ -395,6 +396,33 @@ func (m *AwsManager) buildNodeFromTemplate(asg *asg, template *asgTemplate) (*ap
 	node.Status.Capacity[gpu.ResourceNvidiaGPU] = *resource.NewQuantity(template.InstanceType.GPU, resource.DecimalSI)
 	node.Status.Capacity[apiv1.ResourceMemory] = *resource.NewQuantity(template.InstanceType.MemoryMb*1024*1024, resource.DecimalSI)
 
+	if asg.MixedInstancesPolicy != nil && asg.MixedInstancesPolicy.instanceRequirementsOverrides != nil {
```
bwagner5, line 399: This needs to handle regular LT Instance Requirements (ie. no overrides are present in the Mixed Instances Policy) as well.

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
```diff
@@ -255,10 +258,84 @@ func (m *awsWrapper) getInstanceTypeByLaunchTemplate(launchTemplate *launchTempl
 	return aws.StringValue(instanceType), nil
 }
 
+func (m *awsWrapper) getInstanceTypesFromInstanceRequirements(policy *mixedInstancesPolicy) ([]string, error) {
```
bwagner5, line 261: This needs to handle regular LT Instance Requirements (ie. no overrides are present in the Mixed Instances Policy) as well.

bwagner5 (changes requested):
Great work! Very clean addition to this rather complicated API add-on :D 

A few comments on handling the case where a user specified the InstanceRequirements within the LT rather than within an ASG MixedInstancesPolicy Override. In that case, the InstanceRequirement will not show up in the MixedInstancePolicy, so the LT will need to be inspected to retrieve the Requirements.

cluster-autoscaler/cloudprovider/aws/aws_manager.go
```diff
@@ -419,6 +449,36 @@ func (m *AwsManager) buildNodeFromTemplate(asg *asg, template *asgTemplate) (*ap
 	return &node, nil
 }
 
+func (m *AwsManager) getInstanceRequirementsFromMixedInstancesPolicy(policy *mixedInstancesPolicy) (*ec2.InstanceRequirements, error) {
+	instanceRequirements := &ec2.InstanceRequirements{}
+	if policy.launchTemplate != nil {
```
bwagner5, line 454: a mixed instances policy requires an LT, so this should always be non-nil 

the check should be:

```
if policy.instanceRequirementsOverrides != nil {

} else if lt.LaunchTemplateData.InstanceRequirements != nil {

}
```

Basically if an LT override is specified in the MixedInstancesPolicy, then it should always be used (it overrides the LT parameter).
If there is no instance requirement override, then check if there are InstanceRequirements defined in the LT.

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -33,7 +33,9 @@ The following policy provides the minimum privileges necessary for Cluster Autos
         "autoscaling:DescribeLaunchConfigurations",
         "autoscaling:SetDesiredCapacity",
         "autoscaling:TerminateInstanceInAutoScalingGroup",
-        "ec2:DescribeInstanceTypes"
+        "ec2:DescribeImages",
+        "ec2:DescribeInstanceTypes",
+        "ec2:GetInstanceTypesFromInstanceRequirementsPages"
```
bwagner5, line 38: remove the `Pages` part since that's just a aws-go-sdk helper

cluster-autoscaler/cloudprovider/aws/aws_manager.go
```diff
@@ -395,6 +396,35 @@ func (m *AwsManager) buildNodeFromTemplate(asg *asg, template *asgTemplate) (*ap
 	node.Status.Capacity[gpu.ResourceNvidiaGPU] = *resource.NewQuantity(template.InstanceType.GPU, resource.DecimalSI)
 	node.Status.Capacity[apiv1.ResourceMemory] = *resource.NewQuantity(template.InstanceType.MemoryMb*1024*1024, resource.DecimalSI)
 
+	if asg.MixedInstancesPolicy != nil {
+		instanceRequirements, err := m.getInstanceRequirementsFromMixedInstancesPolicy(asg.MixedInstancesPolicy)
+		if err != nil {
+			klog.Error("error while building node template using instance requirements: (%s)", err)
```
bwagner5, line 402: in the case of an err returned in `getInstanceRequirementsFromMixedInstancesPolicy(...)` instanceRequirements will be nil. So I believe the next group of if's would panic on a nil pointer, right?

```
if instanceRequirements.VCpuCount != nil {
...
```

AustinSiu: You're right, thanks for catching. Will fix and make this section more readable as well.

cluster-autoscaler/cloudprovider/aws/aws_manager.go
```diff
@@ -419,6 +422,70 @@ func (m *AwsManager) buildNodeFromTemplate(asg *asg, template *asgTemplate) (*ap
...
+	instanceRequirements := &ec2.InstanceRequirements{}
+	if policy.instanceRequirementsOverrides != nil {
+		var err error
+		instanceRequirements, err = m.awsService.getRequirementsRequestFromAutoscalingToEC2(policy.instanceRequirementsOverrides)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		params := &ec2.DescribeLaunchTemplateVersionsInput{
+			LaunchTemplateName: aws.String(policy.launchTemplate.name),
```
jaypipes, line 467: This assumes `policy.launchTemplate` is non-nil. You will want to add a nil-guard to the `else` statement above on line 465:

```go
     } else if policy.launchTemplate != nil {
```

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
```diff
@@ -233,32 +236,151 @@ func (m *awsWrapper) getInstanceTypeByLaunchTemplate(launchTemplate *launchTempl
...
+func (m *awsWrapper) getRequirementsRequestFromAutoscaling(requirements *autoscaling.InstanceRequirements) (*ec2.InstanceRequirementsRequest, error) {
+	requirementsJson, err := json.Marshal(*requirements)
+	if err != nil {
+		return nil, err
+	}
+
+	var requirementsRequest *ec2.InstanceRequirementsRequest
+	if err := json.Unmarshal(requirementsJson, &requirementsRequest); err != nil {
+		return nil, err
+	}
```
jaypipes, line 346: heh. sneaky. But prone to breakage if and when the shapes in the EC2 and EC2 autoscaling APIs inevitably change... recommend rewriting this using a more verbose constructor and field setting approach instead of the JSON-to-JSON approach.

jaypipes (changes requested):
Great start on this @AustinSiu! Thank you!

I've left a number of comments inline, please take a look.

bwagner5 (approved):
nice!

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
```diff
@@ -483,17 +484,27 @@ func (m *asgCache) buildAsgFromAWS(g *autoscaling.Group) (*asg, error) {
...
 			return res
 		}
 
+		getInstanceTypeRequirements := func(overrides []*autoscaling.LaunchTemplateOverrides) *autoscaling.InstanceRequirements {
+			if len(overrides) == 1 && overrides[0].InstanceRequirements != nil {
+				return overrides[0].InstanceRequirements
+			}
+			return nil
+		}
+
```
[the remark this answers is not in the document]
AustinSiu: My reasoning was that a MixedInstancesPolicy with both InstanceType *and* InstanceRequirements overrides configured would be an invalid setup, as the API only allows one or the other, but not both: https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_LaunchTemplateOverrides.html

I've added a check for both being present on line 510, to make it more clear to Autoscaler readers what to expect.

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
```diff
@@ -233,32 +236,151 @@ func (m *awsWrapper) getInstanceTypeByLaunchTemplate(launchTemplate *launchTempl
...
+	if len(instanceType) == 0 {
+		return "", fmt.Errorf("unable to find instance type using launch template")
+	}
+	return instanceType, nil
+}
 
+func (m *awsWrapper) getInstanceTypeFromRequirementsOverrides(policy *mixedInstancesPolicy) (string, error) {
+	describeTemplateInput := &ec2.DescribeLaunchTemplateVersionsInput{
+		LaunchTemplateName: aws.String(policy.launchTemplate.name),
+		Versions:           []*string{aws.String(policy.launchTemplate.version)},
```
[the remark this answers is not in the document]
AustinSiu: Addressed as above.

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
```diff
@@ -233,32 +236,151 @@ func (m *awsWrapper) getInstanceTypeByLaunchTemplate(launchTemplate *launchTempl
...
+	}
+	if len(instanceType) == 0 {
+		return "", fmt.Errorf("unable to find instance type using launch template")
+	}
+	return instanceType, nil
+}
 
+func (m *awsWrapper) getInstanceTypeFromRequirementsOverrides(policy *mixedInstancesPolicy) (string, error) {
+	describeTemplateInput := &ec2.DescribeLaunchTemplateVersionsInput{
+		LaunchTemplateName: aws.String(policy.launchTemplate.name),
```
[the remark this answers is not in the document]
AustinSiu: Thank you for catching, added a nil-check at beginning of function.

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
```diff
@@ -233,32 +236,151 @@ func (m *awsWrapper) getInstanceTypeByLaunchTemplate(launchTemplate *launchTempl
...
 		return "", err
 	}
+	if len(describeData.LaunchTemplateVersions) == 0 {
+		return "", fmt.Errorf("unable to find template versions")
+	}
+	lt := describeData.LaunchTemplateVersions[0]
+	instanceType := ""
+	if lt.LaunchTemplateData.InstanceType != nil {
+		instanceType = *lt.LaunchTemplateData.InstanceType
+	} else if lt.LaunchTemplateData.InstanceRequirements != nil {
```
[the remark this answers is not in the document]
AustinSiu: Added on line 249

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
```diff
@@ -233,32 +235,455 @@ func (m *awsWrapper) getInstanceTypeByLaunchTemplate(launchTemplate *launchTempl
...
+
+	start := time.Now()
+	describeData, err := m.DescribeLaunchTemplateVersions(describeTemplateInput)
+	observeAWSRequest("DescribeLaunchTemplateVersions", err, start)
+	if err != nil {
+		return "", err
+	}
 	if len(describeData.LaunchTemplateVersions) == 0 {
 		return "", fmt.Errorf("unable to find template versions")
 	}
```
jaypipes, line 291: you need to place the same nil-guard on `lt.LaunchTemplateData` that you did above on new line 249. You might consider DRY'ing this up into a helper method since you are calling `m.DescribeLaunchTemplateVersions` numerous times in both `aws_wrapper.go` and `aws_manager.go`

jaypipes (changes requested):
@AustinSiu this is very close! just a couple remaining nil guards to address inline, otherwise this is looking good! :)

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
AustinSiu: Makes sense! Will do 👍

cluster-autoscaler/cloudprovider/aws/aws_manager.go
```diff
@@ -419,6 +424,69 @@ func (m *AwsManager) buildNodeFromTemplate(asg *asg, template *asgTemplate) (*ap
...
+		observeAWSRequest("DescribeLaunchTemplateVersions", err, start)
+		if err != nil {
+			return nil, err
+		}
+		if len(describeData.LaunchTemplateVersions) == 0 {
+			return nil, fmt.Errorf("unable to find template versions")
+		}
+
+		lt := describeData.LaunchTemplateVersions[0]
+		if lt.LaunchTemplateData.InstanceRequirements != nil {
```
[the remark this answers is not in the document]
AustinSiu: Thank you for catching! Consolidated the template data logic to reduce the discrepancies in nil guarding.

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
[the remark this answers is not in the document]
AustinSiu: I think the below documentation is a little more explicit about disallowing both fields to be present: https://docs.aws.amazon.com/cli/latest/reference/autoscaling/update-auto-scaling-group.html
https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_LaunchTemplate.html

> The overrides can include either one or more instance types or a set of instance requirements, but not both.

I have also tried to set both on a Launch Template through the Console, and it doesn't allow setting both when creating or editing, just to try and confirm.

jaypipes (approved):
Thanks for the revisions and your patience on this, @AustinSiu :)

AustinSiu:
@aleksandra-malinowska @feiskyer @towca can you please help approve?

cluster-autoscaler/go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/Azure/go-autorest/autorest/azure/auth v0.5.8
 	github.com/Azure/go-autorest/autorest/date v0.3.0
 	github.com/Azure/go-autorest/autorest/to v0.4.0
-	github.com/aws/aws-sdk-go v1.38.49
```
mwielgus, line 13: As we import lots of code from kubernetes/kubernetes want to be in sync with it (k/k uses github.com/aws/aws-sdk-go v1.38.49) to avoid compilation and runtime issues. So we cannot accept any changes that can make future sync problematic. Please update k/k deps or roll back this change.

mwielgus (changes requested):

cluster-autoscaler/go.mod
AustinSiu: Understood, opened a PR to update dependencies in kubernetes/kubernetes: https://github.com/kubernetes/kubernetes/pull/108897

AustinSiu: @mwielgus the reviewers on the above PR have some pushback against further changes to `aws-sdk-go` as a legacy cloud provider in k/k: https://github.com/kubernetes/kubernetes/pull/108897#issuecomment-1083689535. What should be the next step here?

For some context, I updated the `github.com/aws/aws-sdk-go` version from v1.38.49 because this integration is dependent on API support for attribute-based instance requirements, introduced in [v1.41.12](https://github.com/aws/aws-sdk-go/releases/tag/v1.41.12).

AustinSiu: @mwielgus ping :)

sauryadas: @feiskyer Possible for you to do look into this ? we need to make progress

AustinSiu: @gjtempleton may I ask if you have an opinion on where to go from here, with regards to k/k's stance on provider updates and changes like this in Autoscaler?

AustinSiu:
@mwielgus @gjtempleton I've updated this change to handle the dependencies differently, does this address your concerns?

gjtempleton (approved):
Thanks for your persistence with this, great job.

/lgtm
/approve

PR #4610, opened 2022-01-14
Fix code formatting

Fedosin:

All

#### What type of PR is this?

/kind cleanup

It seems like the code formatting is broken. This commit fixes it using standard golang tooling.

Ran:
./hack/update-gofmt.sh


elmiko:
it looks like this fixes things for golang 1.17, thanks Mike
/lgtm

@MaciekPytel ptal

jbartosik (changes requested):
I don't think we should change autogenerated files (for example vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/zz_generated.deepcopy.go)

PR #4617, opened 2022-01-17
Add AutoscalingContext to the scale-down post-processor

kisieland:
This is a followup PR to #4519 that extends the `ScaleDownSetProcessor` to include the `context.AutoscalingContext` parameter in the `GetNodesToRemove` function.

Edit: unlike the https://github.com/kubernetes/autoscaler/pull/4519#issuecomment-1012967211 suggested I propose we use the context that is already part of `ScaleDown`


kisieland:
The tests are also broken on master branch, seems like this is unrelated.
As soon as fix hits the master I will rebase and it should be good.

MaciekPytel:
/lgtm
/approve
Thanks! The unittest is indeed a know issue tracked by https://github.com/kubernetes/autoscaler/issues/4592.

PR #4620, opened 2022-01-17
azure: change a flacky test

bpineau:
It seems that test gets varying error messages which prompted
Bartłomiej previous fix, but I'm now seeing the original error
message string back ("Server failed to authenticate [...]"),
so that `TestDeleteBlob` test is failing again (other PRs' tests
failures suggest that's not just my laptop).

Let's assume this was meant to check for an error, until someone
can confirm, that might be better than potentially hidding other
PRs real tests failures.

Azure cloud provider.

#### What type of PR is this?
/kind flake

Please let us know if that was really meant to catch specific error messages, and if so, which message.

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_util_test.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_util_test.go
@@ -19,7 +19,6 @@ package azure
 import (
 	"fmt"
 	"net/http"
-	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


marwanad:
I think we should remove this test completely till we figure out what exactly is being tested and avoid making that API call. @nilo19 

This affects all release branches so we should remove it to unblock people on their cherry-picks too.

PR #4621, opened 2022-01-17
gce: rm unnecessary GetMigInstanceTemplate locking

bpineau:
The lock held from (and covering most of) `GetMigInstanceTemplate()` prevents
`InstanceTemplates.Get()` concurrent executions, while it's not really needed
as all caches access functions used here already take care of locking cache
accesses (and it's reading and filling one single cache entry per call).

`InstanceTemplates.Get()` isn't fast (at about 0.5s/call in my test env),
running all calls sequentially slows down startup significantly:

```
forceRefresh()
  fetchAutoMigs()
    workqueue.ParallelizeUntil(..., func() { <- can launch concurrent registerMigs
      registerMig()
        GetMigTemplateNode()
          GetMigInstanceTemplate() <- lock held
            FetchMigTemplate()
              client.gceService.InstanceTemplates.Get() <- slow API call
```

When running this change on a cluster having 800 MIGs, startup went from 6min to 1.5min.
Also tested with a `-race` build.

Cluster-autoscaler's GCE cloud provider.

#### What type of PR is this?
/kind cleanup
/kind feature

Remove unnecessary locking to speed-up startup time on large clusters (many MIGs).


MaciekPytel:
@BigDarkClown can you take a look? I think you're the most familiar with this part of the code.

BigDarkClown:
I think this change is reasonable. As you noticed the cache setters and getters are already guarded by a different lock, so there should be no issues with concurrency there. 

In other `MigInfoProvider` functions the lock held here prevents excessive querying, as usually it takes only one a single call to fill the underlying cache. But for `GetMigInstanceTemplate` this is not really a case, for each MIG first call we will require a single API call either way. So removing this lock here seems like a good idea.

MaciekPytel:
Oops, it completely slipped my radar.
/lgtm
/approve
Sorry for delay :(

PR #4628, opened 2022-01-19
Explain how to prevent deletion of non-empty nodes

aleb:

cluster-autoscaler

#### What type of PR is this?

/kind documentation

The Google Kubernetes clusters scale the node pools automatically per each node-pool settings, all is good. In AWS EKS the "managed nodegroups" are misleadingly called "managed" as they are not actually scaled. We're migrating to EKS so that's how we ended up setting up CA. I lost too much time debugging why pods are killed and how to prevent this, so hopefully this small addition will clarify things for future adventurers, saving them time.

```release-note
NONE
```


mwielgus (approved):
/lgtm
/approve

PR #4633, opened 2022-01-21
CA: Debugging snapshot adding a new field for TemplateNode.

jayantjain93:
This captures all the templates for nodegroups present for debugging snapshot

/kind cluster-autoscaler

#### What type of PR is this?
/kind feature

Fixes #

```release-note

```

```docs

```


cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
```diff
@@ -58,11 +61,12 @@ type DebuggingSnapshot interface {
...
-	UnscheduledPodsCanBeScheduled []*v1.Pod   `json:"UnscheduledPodsCanBeScheduled"`
-	Error                         string      `json:"Error,omitempty"`
-	StartTimestamp                time.Time   `json:"StartTimestamp"`
-	EndTimestamp                  time.Time   `json:"EndTimestamp"`
+	NodeInfo                      []*NodeInfo           `json:"NodeList"`
+	UnscheduledPodsCanBeScheduled []*v1.Pod             `json:"UnscheduledPodsCanBeScheduled"`
+	Error                         string                `json:"Error,omitempty"`
+	StartTimestamp                time.Time             `json:"StartTimestamp"`
+	EndTimestamp                  time.Time             `json:"EndTimestamp"`
+	TemplateNode                  []*framework.NodeInfo `json:"TemplateNode"`
```
MaciekPytel, line 69: Making it a list looses information about which TemplateNode is for which NodeGroup. It seems to me that if you have a hundred NodeGroups, trying to manually match template nodes to each group would be a very painful exercise (some providers may have a label that includes NodeGroup name, but we shouldn't assume that).
Is there any reason why we can't just use a map here?

also nit: TemplateNode implies a singular node which is very confusing, please pluralize the name.

cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
```diff
@@ -58,11 +61,12 @@ type DebuggingSnapshot interface {
 // Please add all new output fields in this struct. This is to make the data
 // encoding/decoding easier as the single object going into the decoder
 type DebuggingSnapshotImpl struct {
-	NodeInfo                      []*NodeInfo `json:"NodeList"`
-	UnscheduledPodsCanBeScheduled []*v1.Pod   `json:"UnscheduledPodsCanBeScheduled"`
-	Error                         string      `json:"Error,omitempty"`
-	StartTimestamp                time.Time   `json:"StartTimestamp"`
-	EndTimestamp                  time.Time   `json:"EndTimestamp"`
+	NodeInfo                      []*NodeInfo           `json:"NodeList"`
```
MaciekPytel, line 64: nit: I know this is not introduced in this PR, but why is the field name different that the corresponding field in json? This seems needlessly confusing.
Also - can you make the name imply it's multiple nodes/nodeInfos? NodeList is good, but NodeInfo seems like it would be just a single object and not a list.

cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
```diff
@@ -77,6 +81,21 @@ func (s *DebuggingSnapshotImpl) SetUnscheduledPodsCanBeScheduled(podList []*v1.P
...
 
+// SetTemplateNode is the setter for TemplateNode
+func (s *DebuggingSnapshotImpl) SetTemplateNode(templates []*framework.NodeInfo) {
+	if templates == nil {
+		return
+	}
+
+	var TemplateNodeInfo []*framework.NodeInfo
+	for _, template := range templates {
+		t := template.Clone()
```
MaciekPytel, line 92: Why do you clone the nodeInfo here and create a new NodeInfo instead in SetNodeGroupInfo? It seems like the logic should be the same in both cases. Maybe consider extracting the copying logic to a helper function?

cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
```diff
@@ -77,6 +81,21 @@ func (s *DebuggingSnapshotImpl) SetUnscheduledPodsCanBeScheduled(podList []*v1.P
...
+		t.UsedPorts = nil
```
MaciekPytel, line 93: Why is this needed? Why do it here, but not in SetNodeGroupInfo?

cluster-autoscaler/debuggingsnapshot/debugging_snapshotter.go
```diff
@@ -230,6 +233,17 @@ func (d *DebuggingSnapshotterImpl) SetUnscheduledPodsCanBeScheduled(podList []*v
 	*d.State = DATA_COLLECTED
 }
 
+// SetTemplateNode is the setter for TemplateNode
+func (d *DebuggingSnapshotterImpl) SetTemplateNode(templates []*framework.NodeInfo) {
+	if !d.IsDataCollectionAllowed() {
+		return
+	}
+	klog.Infof("TemplateNode is being set for the debugging snapshot")
```
MaciekPytel, line 241: Please use V() for any Info logs (here and elsewhere in this function). I'd consider a V(4) - on by default, we tend to use it as default log level for most logs.

MaciekPytel (commented):
Github doesn't allow me to make comments in lines that weren't modified in this PR, but I'd like to ask you to rename SetNodeGroupInfo method. I think the name is very confusing: it captures nodeInfos for exising nodes. It has nothing to do with NodeGroups - the nodes are not grouped by the NodeGroup they belong to and they contain no information about NodeGroup. Some of the nodes can come up from not-autoscaled ASG/nodepool/whatever, in which case they don't even belong to any NodeGroup known to CA.

I'd much prefer something like SetClusterNodes() or similar.

cluster-autoscaler/debuggingsnapshot/debugging_snapshotter.go
jayantjain93: done

cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
jayantjain93: we also clone in SetNodeGroupInfo along with transforming it into a new data type. The new data type debugging_snaphot.NodeInfo, contains *framework.Node and []*framework.Pod but it doesn't contain the "template", that is framwork.NodeInfo.

cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
jayantjain93: There are fields in that obj which cannot be json marshalled, so they have been removed.
The object being captured in SetNodeGroupInfo is different to the one here. As mentioned above.

cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
jayantjain93: We can change it to a map with a key of nodegroup.
Done.

cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
MaciekPytel: Oh, I missed that it's using a locally defined "NodeInfo" struct. Can we rename that? Masking name like that makes it super confusing (as demonstrated by the fact that I didn't realize it was a different NodeInfo type).

Also - why the different representation? As far as I know framework.NodeInfo is a wrapper for a node + set of pods and everything else is just some precomputation to optimize scheduler functions. Is there any data we're missing by using the node+pods representation? If so, why is it ok to lose this data for nodes, but not ok to lose it for template nodes?

jayantjain93: Renamed.
The idea is to remove excess / duplicate information for "ClusterNodes". Since we already have the information in the node spec, we don't need another representation / copy in the Template. (At least for the fields I thought would of interest). This also has scalability implications, having duplicate template for every node in that NodeGroup.

cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
jayantjain93: Done. Changed Names.

cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
MaciekPytel: I'm confused. framework.NodeInfo is just a wrapper for node+pods used by scheduler. I don't think there is anything more to it (sure, there are other fields, but AFAIK those are just some precomputed values that are purely based on the node+pods). The template itself is just node + ds + static pods.
For convenience CA re-uses framework.NodeInfo to internally represent node+pods. We do it for both template nodes and actual nodes. As far as I know the only difference between 'cluster node' NodeInfo and 'template node' NodeInfo is that one represents a node that really exist and the other an in-memory fake node, but functionally they should be the same.

I don't think there is anything in NodeInfo that we need in snapshot other than node+pods. And if there were, I don't understand why we'd need it only for template nodes.

jayantjain93: Based on offline discussion, it is concluded that template node should also be of internal debugging data struct `ClusterNode`, as data needed would be available in the internal struct and there would be similarities of type usage.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -291,6 +291,10 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 		return autoscalerError.AddPrefix("failed to build node infos for node groups: ")
 	}
 
+	if a.DebuggingSnapshotter.IsDataCollectionAllowed() {
```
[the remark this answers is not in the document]
jayantjain93: This was a spillover from multiple patches. Removed IsDataCollectionAllowed() call from autoscaler as it's not needed, and it is being checked by the snapshotter.

cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
```diff
@@ -77,27 +81,41 @@ func (s *DebuggingSnapshotImpl) SetUnscheduledPodsCanBeScheduled(podList []*v1.P
...
+	s.TemplateNodes = nil
+	for ng, template := range templates {
+		s.TemplateNodes[ng] = GetClusterNodeCopy(template)
+	}
+}
+
+// GetClusterNodeCopy is an util func to copy template node and filter values
+func GetClusterNodeCopy(template *framework.NodeInfo) *ClusterNode {
+	cNode := &ClusterNode{}
+	clone := template.Clone()
```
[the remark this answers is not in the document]
jayantjain93: Great catch. Fixed it.

cluster-autoscaler/debuggingsnapshot/debugging_snapshot.go
```diff
@@ -73,31 +77,48 @@ func (s *DebuggingSnapshotImpl) SetUnscheduledPodsCanBeScheduled(podList []*v1.P
...
+	}
+}
+
+// GetClusterNodeCopy is an util func to copy template node and filter values
+func GetClusterNodeCopy(template *framework.NodeInfo) *ClusterNode {
+	cNode := &ClusterNode{}
+	cNode.Node = template.Node().DeepCopy()
+	var pods []*framework.PodInfo
+	for _, p := range template.Pods {
+		pods = append(pods, p.DeepCopy())
```
MaciekPytel, line 102: I just realized this is PodInfo, which is another scheduler wrapper (similarly to how NodeInfo wraps Node). Maybe we should consider switching to PodInfo.Pod.DeepCopy()? I'm pretty sure we don't need anything in PodInfo other than the actual pod.

jayantjain93: Done.

PR #4641, opened 2022-01-24
Don't cache NodeInfo for recently Ready nodes

x13n:

cluster-autoscaler

#### What type of PR is this?

/kind bug

There's a race condition between DaemonSet pods getting scheduled to a
new node and Cluster Autoscaler caching that node for the sake of
predicting future nodes in a given node group. We can reduce the risk of
missing some DaemonSet by providing a grace period before accepting nodes in the
cache. 1 minute should be more than enough, except for some pathological
edge cases.

Fixes #

```release-note
NONE
```

```docs

```

/assign @MaciekPytel

### DIFF
--- a/cluster-autoscaler/processors/nodeinfosprovider/mixed_nodeinfos_processor.go
+++ b/cluster-autoscaler/processors/nodeinfosprovider/mixed_nodeinfos_processor.go
@@ -18,6 +18,7 @@ package nodeinfosprovider
 
 import (
 	"reflect"
+	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	apiv1 "k8s.io/api/core/v1"
@@ -32,6 +33,8 @@ import (
 
 	klog "k8s.io/klog/v2"
 )
+
+const stabilizationDelay = 1 * time.Minute
 
 // MixedTemplateNodeInfoProvider build nodeInfos from the cluster's nodes and node groups.
 type MixedTemplateNodeInfoProvider struct {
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/processors/nodeinfosprovider/mixed_nodeinfos_processor.go
```diff
@@ -56,6 +59,7 @@ func (p *MixedTemplateNodeInfoProvider) Process(ctx *context.AutoscalingContext,
 	// TODO(mwielgus): Review error policy - sometimes we may continue with partial errors.
 	result := make(map[string]*schedulerframework.NodeInfo)
 	seenGroups := make(map[string]bool)
+	now := time.Now()
```
MaciekPytel, line 62: nit: This is called early enough in CA loop that it probably doesn't matter as much, but maybe passing currentTime from RunOnce() would be more consistent? Since CA loop operates on a snapshot we generally use timestamp of loop start as 'now' for this type of checks.

cluster-autoscaler/processors/nodeinfosprovider/mixed_nodeinfos_processor.go
```diff
@@ -90,7 +94,7 @@ func (p *MixedTemplateNodeInfoProvider) Process(ctx *context.AutoscalingContext,
 
 	for _, node := range nodes {
 		// Broken nodes might have some stuff missing. Skipping.
-		if !kube_util.IsNodeReadyAndSchedulable(node) {
+		if !isNodeGoodForCaching(node, now) {
```
MaciekPytel, line 97: You're using this check to see if a node can be used as a template node for this loop (processNode call that you're skipping adds the node to the result), not just if it can be cached. That may actually be the right thing to do (node without DS doesn't make for a good template), but the function name is misleading. I'd either rename the function or, if we think this check should only apply to caching, move it check to L104 where we actually handle caching.

MaciekPytel (commented):
Overall lgtm, left a few minor comments.

cluster-autoscaler/processors/nodeinfosprovider/mixed_nodeinfos_processor.go
x13n: Done.

cluster-autoscaler/processors/nodeinfosprovider/mixed_nodeinfos_processor.go
x13n: Done.

PR #4651, opened 2022-01-27
CA: Debugging snapshotter locking optimisation for better transactions

jayantjain93:
Cluster autoscaler

#### What type of PR is this?
/kind feature

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/debuggingsnapshot/debugging_snapshotter.go
+++ b/cluster-autoscaler/debuggingsnapshot/debugging_snapshotter.go
@@ -172,6 +172,13 @@ func (d *DebuggingSnapshotterImpl) ResponseHandler(w http.ResponseWriter, r *htt
 func (d *DebuggingSnapshotterImpl) IsDataCollectionAllowed() bool {
 	d.Mutex.Lock()
 	defer d.Mutex.Unlock()
+	return d.IsDataCollectionAllowedNoLock()
+}
+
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/debuggingsnapshot/debugging_snapshotter.go
```diff
@@ -172,6 +172,13 @@ func (d *DebuggingSnapshotterImpl) ResponseHandler(w http.ResponseWriter, r *htt
 func (d *DebuggingSnapshotterImpl) IsDataCollectionAllowed() bool {
 	d.Mutex.Lock()
 	defer d.Mutex.Unlock()
+	return d.IsDataCollectionAllowedNoLock()
+}
+
+// IsDataCollectionAllowedNoLock encapsulated the check to know if data collection is currently active
+// The need for NoLock implementation is for cases when the caller funcs have procured the lock
+// for a single transactional execution
+func (d *DebuggingSnapshotterImpl) IsDataCollectionAllowedNoLock() bool {
```
x13n, line 181: I don't think the NoLock version should be public: it should only be used internally by the snapshotter.

jayantjain93: This isn't part of the interface, which is public.
However, CP snapshotter extensions should have access to this function, which is what we achieve from it being public method.

PR #4657, opened 2022-02-01
Add more buckets to admission_latency_seconds metric

jbartosik:
I want more resolution in [0.2, 1]s range. [K8s Pod Creation Latency
SLO](https://github.com/kubernetes/community/blob/master/sig-scalability/slos/pod_startup_latency.md) has a target of 5s so it makes a difference for the SLO how the
delay we introduce is distributed. Right now I think buckets are too
wide in the [0.1, 1]s range.

/kind feature

```release-note
NONE
```


jbartosik:
@kgolab

josephburnett (approved):

PR #4669, opened 2022-02-07
Add expire time for nodeInfo cache items

yaroslava-serdiuk:

cluster-autoscaler

#### What type of PR is this?

/kind bug

Adding expire time for nodeInfo cache items. Expire time could be set via flag `--node-info-cache-expire-time`, by default there is no expire time (old behavior).

Currently, item in the node info cache is updated only if there are any nodes in the node group. If node group is scaled down to 0, node info cache will store last node and use it in scale up algorithm. Thus if the last node is a wrong representative of the node group (for example it has custom taints), then Cluster Autoscaler may struggle to scale up this node group.

```release-note
Added `--node-info-cache-expire-time` flag for Cluster Autoscaler to specify the time to store node info cache items.
```

### DIFF
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -46,6 +46,7 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/metrics"
 	ca_processors "k8s.io/autoscaler/cluster-autoscaler/processors"
 	"k8s.io/autoscaler/cluster-autoscaler/processors/nodegroupset"
+	"k8s.io/autoscaler/cluster-autoscaler/processors/nodeinfosprovider"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
 	kube_util "k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
@@ -185,6 +186,7 @@ var (
 
 	emitPerNodeGroupMetrics  = flag.Bool("emit-per-nodegroup-metrics", false, "If true, emit per node group metrics.")
 	debuggingSnapshotEnabled = flag.Bool("debugging-snapshot-enabled", false, "Whether the debugging snapshot of cluster autoscaler feature is enabled")
+	nodeInfoCacheExpireTime  = flag.Duration("node-info-cache-expire-time", 87600*time.Hour, "Node Info cache expire time for each item. Default value is 10 years.")
 )
 
 func createAutoscalingOptions() config.AutoscalingOptions {
@@ -322,6 +324,7 @@ func buildAutoscaler(debuggingSnapshotter debuggingsnapshot.DebuggingSnapshotter
 	}
 
 	opts.Processors = ca_processors.DefaultProcessors()
+	opts.Processors.TemplateNodeInfoProvider = nodeinfosprovider.NewDefaultTemplateNodeInfoProvider(nodeInfoCacheExpireTime)
 	opts.Processors.PodListProcessor = core.NewFilterOutSchedulablePodListProcessor()
 
 	nodeInfoComparatorBuilder := nodegroupset.CreateGenericNodeInfoComparator
--- a/cluster-autoscaler/processors/nodeinfosprovider/mixed_nodeinfos_processor.go
+++ b/cluster-autoscaler/processors/nodeinfosprovider/mixed_nodeinfos_processor.go
@@ -35,18 +35,34 @@ import (
 )
 
 const stabilizationDelay = 1 * time.Minute
+const maxCacheExpireTime = 87660 * time.Hour
+
+type cacheItem struct {
+	*schedulerframework.NodeInfo
+	added time.Time
+}
 
 // MixedTemplateNodeInfoProvider build nodeInfos from the cluster's nodes and node groups.
 type MixedTemplateNodeInfoProvider struct {
-	nodeInfoCache map[string]*schedulerframework.NodeInfo
+	nodeInfoCache map[string]cacheItem
+	ttl           time.Duration
 }
 
 // NewMixedTemplateNodeInfoProvider returns a NodeInfoProvider processor building
 // NodeInfos from real-world nodes when available, otherwise from node groups templates.
-func NewMixedTemplateNodeInfoProvider() *MixedTemplateNodeInfoProvider {
-	return &MixedTemplateNodeInfoProvider{
-		nodeInfoCache: make(map[string]*schedulerframework.NodeInfo),
+func NewMixedTemplateNodeInfoProvider(t *time.Duration) *MixedTemplateNodeInfoProvider {
+	ttl := maxCacheExpireTime
+	if t != nil {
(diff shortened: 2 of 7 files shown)


cluster-autoscaler/processors/nodeinfosprovider/mixed_nodeinfos_processor.go
```diff
@@ -36,17 +36,31 @@ import (
...
+func NewMixedTemplateNodeInfoProvider(ttl *time.Duration) *MixedTemplateNodeInfoProvider {
 	return &MixedTemplateNodeInfoProvider{
-		nodeInfoCache: make(map[string]*schedulerframework.NodeInfo),
+		nodeInfoCache: make(map[string]cacheItem),
+		ttl:           ttl,
+	}
+}
+
+func (p *MixedTemplateNodeInfoProvider) isCacheItemExpired(added time.Time) bool {
+	if p.ttl == nil || *p.ttl < 0 {
```
x13n, line 60: Why do we need both nil duration and negative duration to achieve the same thing as setting the TTL to an arbitrarily large value?

cluster-autoscaler/processors/processors.go
```diff
@@ -77,7 +77,6 @@ func DefaultProcessors() *AutoscalingProcessors {
 		NodeInfoProcessor:          nodeinfos.NewDefaultNodeInfoProcessor(),
 		NodeGroupConfigProcessor:   nodegroupconfig.NewDefaultNodeGroupConfigProcessor(),
 		CustomResourcesProcessor:   customresources.NewDefaultCustomResourcesProcessor(),
-		TemplateNodeInfoProvider:   nodeinfosprovider.NewDefaultTemplateNodeInfoProvider(),
```
x13n, line 80: Why did you remove this line?

cluster-autoscaler/processors/nodeinfosprovider/mixed_nodeinfos_processor.go
```diff
@@ -36,17 +36,31 @@ import (
 
 const stabilizationDelay = 1 * time.Minute
 
+type cacheItem struct {
```
x13n, line 39: nit: You could embed `*schedulerframework.NodeInfo` here.

cluster-autoscaler/main.go
```diff
@@ -185,6 +186,7 @@ var (
 
 	emitPerNodeGroupMetrics  = flag.Bool("emit-per-nodegroup-metrics", false, "If true, emit per node group metrics.")
 	debuggingSnapshotEnabled = flag.Bool("debugging-snapshot-enabled", false, "Whether the debugging snapshot of cluster autoscaler feature is enabled")
+	nodeInfoCacheExpireTime  = flag.Duration("node-info-cache-expire-time", -1*time.Minute, "Node Info cache expire time for each item. If the value isn't set or negative, the cache items won't be expired")
```
x13n, line 189: nit: You could default to something like 100 years to avoid special casing negative values.

x13n (changes requested):

cluster-autoscaler/processors/processors.go
yaroslava-serdiuk: I don't think it is default processors after introducing an ttl time parameter. In order to not pass a ttl time to DefaultProcessors() function I initialise TemplateNodeInfoProvider separately in main.go

cluster-autoscaler/processors/nodeinfosprovider/mixed_nodeinfos_processor.go
```diff
@@ -36,17 +36,31 @@ import (
...
+func NewMixedTemplateNodeInfoProvider(ttl *time.Duration) *MixedTemplateNodeInfoProvider {
 	return &MixedTemplateNodeInfoProvider{
-		nodeInfoCache: make(map[string]*schedulerframework.NodeInfo),
+		nodeInfoCache: make(map[string]cacheItem),
+		ttl:           ttl,
+	}
+}
+
+func (p *MixedTemplateNodeInfoProvider) isCacheItemExpired(added time.Time) bool {
+	if p.ttl == nil {
```
x13n, line 60: nit: Why do you use a pointer at all? If `ttl` was a regular value you wouldn't have to do this.

yaroslava-serdiuk: I left pointer for test that mocks MixedTemplateNodeInfoProvider, so there is no need to pass a specific parameter there. However it should be handled during provider creation.

cluster-autoscaler/processors/nodeinfosprovider/mixed_nodeinfos_processor_test.go
```diff
@@ -187,10 +191,10 @@ func TestGetNodeInfosForGroupsCache(t *testing.T) {
 	// Check cache
 	cachedInfo, found := niProcessor.nodeInfoCache["ng1"]
 	assert.True(t, found)
-	assertEqualNodeCapacities(t, ready1, cachedInfo.Node())
+	assertEqualNodeCapacities(t, ready1, cachedInfo.NodeInfo.Node())
```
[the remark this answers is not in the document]
yaroslava-serdiuk: Yes, sorry missed few places.

cluster-autoscaler/processors/processors.go
x13n: I understand, but now you've changed the default implementation so that it always has a nil `TemplateNodeInfoProvider`, so anyone who used to rely on this will suddenly get a nil pointer (and most likely `panic` down the road). I suggest either keeping it initialized with some default TTL value (like 10 years) or mention breaking change in a release note.

cluster-autoscaler/processors/nodeinfosprovider/mixed_nodeinfos_processor.go
x13n: So you have different state for tests and a different one for actual execution. If you used a const like `veryLongTime = 123456 * time.Hour` in the tests, you wouldn't need special casing `nil`.

yaroslava-serdiuk: I wouldn't say that state for test is different, a nil pointer correspond to default value — the behavior is identical as if --node-info-cache-expire-time flag is not set. https://github.com/kubernetes/autoscaler/pull/4669/files#diff-ad9d6ea8aa127e0dc3b271f9b5761df7ba18f9fd63ffd0a7f69a6886c86ef317R54
I think that if new test is introduced that doesn't test nodeInforCache directly but use it, then the writer shouldn't come up with a new veryLongTime variable. 
Please, explain if my line of thought is wrong. 
(Also, after your argument I used nil pointer for initialising in DefaultProcessor(), which is used not only for test.)

cluster-autoscaler/processors/processors.go
yaroslava-serdiuk: Got it, thanks. I returned TemplateNodeInfoProvider.

PR #4674, opened 2022-02-09
Expose nodes with unready GPU in CA status

x13n:

cluster-autoscaler

#### What type of PR is this?

/kind feature

This change simplifies debugging GPU issues: without it, all nodes can
be Ready as far as Kubernetes API is concerned, but CA will still report
some of them as unready if they are missing e.g. GPU resource. Explicitly calling
them out in the status ConfigMap will point into the right direction.

Fixes #

```release-note
Status ConfigMap now contains additional information about nodes with unready resources.
```

```docs

```

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -537,6 +537,10 @@ type Readiness struct {
 	Unregistered int
 	// Time when the readiness was measured.
 	Time time.Time
+	// Number of nodes that are Unready due to missing resources.
+	// This field is only used for exposing information externally and
+	// doesn't influence CA behavior.
+	ResourceUnready int
 }
 
 func (csr *ClusterStateRegistry) updateReadinessStats(currentTime time.Time) {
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -537,30 +537,35 @@ type Readiness struct {
 	Unregistered int
 	// Time when the readiness was measured.
 	Time time.Time
+	// Number of nodes that are either Unready or NotStarted due to missing resources.
```
MaciekPytel, line 540: Looking at implementation it seems to only count unready nodes and not notStarted? Also maybe it's worth mentioning that this is only used for reporting purposes and not in actual calculations?

MaciekPytel:
lgtm except for the minor nit above
/approve

cluster-autoscaler/clusterstate/clusterstate.go
x13n: Good catch, thanks. That was a leftover from the initial implementation, which used to count notStarted too. I ultimately decided it is simpler to make it a subset of a single existing category. All notStarted nodes will change their state within 15m anyway.

PR #4688, opened 2022-02-17
make cloud provider node deletion and kubernetes empty node deletion configurable

marwanad:
make those parameters configurable instead of hardcoded since not every cloudprovider will have the same upper bound. This is to avoid scenario where autoscaler will retry sending the delete call too early.


marwanad:
lol....

Issue #4692, opened 2022-02-18
Log Verbosity is Too High for VPA Updater


git-it127:
**Which component are you using?**: 
The vpa updater

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**: 
The logs for the vpa updater are too verbose. There is a lot of log output about actions not taken. This clogs up the logs and makes it hard to find the relevant information. 

example:
```go
I0218 21:36:49.818992       1 update_priority_calculator.go:133] not updating a short-lived pod kube-system/etcd-0, request within recommended range
I0218 21:36:49.819061       1 update_priority_calculator.go:137] not updating pod kube-system/ccm-86f5f869b6-fs9n8, resource diff too low: {false false 0}
```

**Describe the solution you'd like.**: 
The "not updating" log messages are currently verbosity level 2. I believe they should be verbosity level 4 for debug.

**Describe any alternative solutions you've considered.**:

**Additional context.**:

git-it127:
The PR to address this has been merged.

PR #4693, opened 2022-02-18
Turn VPA Logging Level to V4 for Not Updating Messages

git-it127:
The VPA updater

#### What type of PR is this?
/kind cleanup

This reduces the verbosity of logs for the VPA updater with "not updating" log messages to a debug level. The VPA updater logs are clogged up with a lot of messages like this and it makes it difficult to find valuable information.

example:
```go
I0218 21:36:49.818992       1 update_priority_calculator.go:133] not updating a short-lived pod kube-system/etcd-0, request within recommended range
I0218 21:36:49.819061       1 update_priority_calculator.go:137] not updating pod kube-system/ccm-86f5f869b6-fs9n8, resource diff too low: {false false 0}
```

Fixes # https://github.com/kubernetes/autoscaler/issues/4692

```release-note
Lower the verbosity for "not updating" log messages for the VPA updater to debug level

```

```docs

```

### DIFF
--- a/vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go
+++ b/vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go
@@ -122,22 +122,22 @@ func (calc *UpdatePriorityCalculator) AddPod(pod *apiv1.Pod, now time.Time) {
 	if !updatePriority.OutsideRecommendedRange && !quickOOM {
 		if pod.Status.StartTime == nil {
 			// TODO: Set proper condition on the VPA.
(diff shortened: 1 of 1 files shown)


g1bl3t:
@jbartosik you could still get these logs at a higher verbosity level, though. I feel like if you find yourself filtering these logs out often (as we do), it might be a good opportunity to cut down on some noise

git-it127:
> 

@jbartosik how are you filtering out the logs? That is the real problem I am trying to solve. When loading the logs via kubectl, they take a long time to load because these `not updating` messages are added every minute or so.

I was looking to filter the `not updating` messages out by changing the verbosity to a lower setting. But the `accepted for update` and the `not updating` messages are the same verbosity level. So I am not able to filter them out.

mwielgus (approved):
/lgtm
/approve

Issue #3216, opened 2020-06-09
Failed to generate AWS EC2 Instance Types: unable to load EC2 Instance Type list


mrichman:
I'm not sure if my worker node role is missing a permission:

`aws_cloud_provider.go:363] Failed to generate AWS EC2 Instance Types: unable to load EC2 Instance Type list`

I currently have the following applied:

- AmazonEKSWorkerNodePolicy 
- AmazonEC2ContainerRegistryReadOnly 
- AmazonEKS_CNI_Policy 

and this inline policy:

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "autoscaling:DescribeAutoScalingGroups",
                "autoscaling:DescribeAutoScalingInstances",
                "autoscaling:DescribeTags",
                "autoscaling:SetDesiredCapacity",
                "autoscaling:TerminateInstanceInAutoScalingGroup"
            ],
            "Resource": "*"
        }
    ]
}
```

gjtempleton:
The code path that currently fails is actually using the AWS Pricing API over HTTP, here (that's on master as you've not mentioned which version of the CA you're using):
https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/aws_util.go#L55:1

mrichman:
I updated to CA 1.16.5 and added these additional permissions and it's now working:

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "autoscaling:DescribeAutoScalingGroups",
                "autoscaling:DescribeAutoScalingInstances",
                "autoscaling:DescribeInstances",
                "autoscaling:DescribeLaunchConfigurations",
                "autoscaling:DescribeTags",
                "autoscaling:SetDesiredCapacity",
                "autoscaling:TerminateInstanceInAutoScalingGroup",
                "ec2:DescribeLaunchTemplateVersions",
                "ec2:DescribeInstanceTypes"
            ],
            "Resource": "*"
        }
    ]
}
```

git-heera:
My deployment failing with below error. Any particular reason it is going to pricing.us-east-1.amazonaws.com?

W0806 10:18:39.046022       1 aws_util.go:71] Error fetching https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonEC2/current/ap-east-1/index.json skipping...
[...]

PR #4701, opened 2022-02-21
Charts - CA - Remove AWS IAM Docs - Replace With Link

gjtempleton:

helm-charts - Cluster Autoscaler

#### What type of PR is this?

/kind documentation

Addresses the fact we currently specify the AWS IAM permissions required by the CA running in AWS in both the chart's README and the cloudprovider implementation's docs, making it easy for the two to fall out of sync (as seen by the multiple comments on issue #3216. Just updating the README in the chart wouldn't fully address this, as users could use an old version of the chart, but update the CA's image tag to one requiring new IAM permissions. This addresses this by removing the duplication, and instead linking to the README of the AWS cloudprovider implementation as a single source of truth.

None

N/A

```release-note
NONE
```

```docs

```

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -17,4 +17,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.16.1
+version: 9.16.2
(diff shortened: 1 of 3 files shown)


stijnbrouwers:
This PR would solve my issue with #3216 since I was following instructions in the readme.md file to configure my IAM policy. Being redirected to the correct page is an acceptable solution for me.

mwielgus (approved):
/lgtm
/approve

PR #4711, opened 2022-02-24
[CA] [HELM] fix templating of secret used for CAPI

MaxRink:
cluster-autoscaler

#### What type of PR is this?

/kind bug

Fixes the templating of the kubeconfig secret used

```release-note
fix usage of `clusterAPIKubeconfigSecret``in cluster-autoscaler helm chart
```


MaxRink:
cc @gjtempleton

gjtempleton:
Hey @MaxRink 

Thanks for the PR.

I believe this issue is already being addressed (albeit in a slightly different way) along with some other bits and pieces by #4691

MaxRink:
yup, seems that i overlooked that one :S

PR #4714, opened 2022-02-24
added cce verion and move sdk to go mod

wangshiqi308:
cluster-autoscaler

#### What type of PR is this?
/kind feature
/kind cleanup

1. We had to copy the SDK file before because the go module was not enabled(see https://github.com/kubernetes/autoscaler/issues/3146), but in this PR we put our SDK in the go.mod file
2. Added CCE version CA to servicestage folder.

N/A

```release-note
NONE
```

```docs

```


RainbowMango:
I guess after putting SDK to `go.mod`, the original SDL should be removed(actually moved to vendor), why not?

wangshiqi308:
it couldn't be compiled without running `go mod vendor`

RainbowMango:
> it couldn't be compiled without running go mod vendor

Why not vendor the SDK? Is there something wrong?

wangshiqi308:
it could be built with flag `-mod=mod`, but it is not in the `Makefile`, I'm not sure if it could pass the Test, I will have a try.

wangshiqi308:
It couldn't pass test https://github.com/kubernetes/autoscaler/runs/5342369544?check_suite_focus=true

PR #4716, opened 2022-02-25
add user configurable cluster api version

elmiko:

cluster-autoscaler

#### What type of PR is this?

/kind feature

This change introduces an environment variable, `CAPI_VERSION`, through
which a user can set the API version for the group they are using. This
change is being added to address situations where a user might have
multiple API versions for the cluster api group and wishes to be
explicit about which version is selected.

Also adds unit tests and documentation for the new behavior. This change
does not break the existing behavior.

n/a

```release-note
Users can now specify the Cluster API group version by setting the `CAPI_VERSION` environment variable to the desired version string.
```

Updated the README file to include usage documentation

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/README.md
+++ b/cluster-autoscaler/cloudprovider/clusterapi/README.md
@@ -180,6 +180,23 @@ the machine annotation on nodes will be `test.8s.io/machine`, the machine deleti
 annotation will be `test.k8s.io/delete-machine`, and the cluster name label will be
 `test.k8s.io/cluster-name`.
 
+## Specifying a Custom Resource Version
+
+When determining the group version for the Cluster API types, by default the autoscaler
+will look for the latest version of the group. For example, if `MachineDeployments`
+exist in the `cluster.x-k8s.io` group at versions `v1alpha1` and `v1beta1`, the
+autoscaler will choose `v1beta1`.
+
+In some cases it may be desirable to specify which version of the API the cluster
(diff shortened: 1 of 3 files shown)


arunmk:
This looks good. My basic question is if it will be clear from context that `CAPI_GROUP` applies to the autoscaler, of it we need something like `CAPI_GROUP_AUTOSCALER`.

elmiko:
> This looks good. My basic question is if it will be clear from context that `CAPI_GROUP` applies to the autoscaler, of it we need something like `CAPI_GROUP_AUTOSCALER`.

i think that's a good thought @arunmk, my primary concern is that we have already had the `CAPI_GROUP` variable available for quite some time now, i wonder if we would need some sort of deprecation policy if we change these?

like perhaps, we could add a patch that would recognize either `CAPI_GROUP` or `CAPI_GROUP_AUTOSCALER`, and `CAPI_VERSION` or `CAPI_VERSION_AUTOSCALER`, then we could provide a deprecation window for the old behavior while promoting the new behavior. wdyt?

arunmk:
Thanks for the explanation @elmiko.

If the variable was already there then I think it is overkill to phase it out. If we document the increased meaning it should be fine IMO.

elmiko:
we do have these sections, https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/clusterapi#specifying-a-custom-resource-group

hopefully that is sufficient, but i'm always open to docs PRs =)

PR #4721, opened 2022-03-02
[CA Helm] fix RBAC for namespaced mode

MaxRink:
Cluster Autoscaler helm chart
Cluster
/kind bug

 This fixed the RBAC for namespaced mode of the cluster autoscaler

```release-note
fix RBAC for namespaced mode of cluster autoscaler helm chart
```

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -17,4 +17,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.16.0
+version: 9.16.1
--- a/charts/cluster-autoscaler/templates/role.yaml
+++ b/charts/cluster-autoscaler/templates/role.yaml
@@ -43,7 +43,7 @@ rules:
       - get
       - update
 {{- end }}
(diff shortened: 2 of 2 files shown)


gjtempleton:
cc @ugiraud As this is reverting a change you made in #4691

/lgtm
Based on my understanding.
Just as a check that I'm grokking it fully before merging though, these RBAC permissions are required when the CA is running in the cluster that is the management cluster, as then it's modifying the CAPI `machine*` objects in that cluster to let the management cluster do the required scaling?

MaxRink:
Correct.
If CA ( or more likely multiple instances, thus the namespaced mode) is running in the management cluster and is watching the target cluster via kubeconfig it needs to interact with machines, machinesets and machinedeployments in the mgmt cluster.

ugiraud:
Hi,
The values.yaml file describe the property this way :
Syntax: workloadClusterMode-ManagementClusterMode
Si, incluster-kubeconfig should mean CA is deployed on the manages cluster ans should not deploy rbac rules.
Etheir the condition is inverted or the comment and most of the conditions

MaxRink:
No, you're wrong here.
Kubeconfig-incluster is correct here, using incluster mode for the mgmt and using kubeconfig for the workload cluster

Incluster-kubeconfig means using incluster mode for the workload and then a kubeconfig file for the mgmt cluster (where Capi resides)

PR #4726, opened 2022-03-06
Make OOM bump up ratio and min bump up configurable.

weiwu-sre:
This PR applies to vertical-pod-autoscaler recommender component.

#### What type of PR is this?

/kind feature

Right now `OOMBumpUpRatio` and `OOMMinBumpUp` are hardcoded. Making it configurable via command line args.

Fixes #

I think this is first step to make `OOMBumpUpRatio` and `OOMMinBumpUp` configurable via command line args. After this, it will be nice to make both variables configurable per VPA. Because different workloads may require different bump up ratio during OOM event, make it per VPA configuration will provide extra flexibility.  This means changes to CRDs and only default to those variable if per VPA configurations are not set.

No. User can keep using existing configuration. And this change will not break any existing configuration.

For any users, who care about configuring `OOMBumpUp` related configuration they can change via command line args. Will be nice to mention that in future release note to indicate the change.

```release-note

```

```docs

```

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/model/container.go
+++ b/vertical-pod-autoscaler/pkg/recommender/model/container.go
@@ -17,6 +17,7 @@ limitations under the License.
 package model
 
 import (
+	"flag"
 	"fmt"
 	"time"
 
@@ -26,10 +27,19 @@ import (
 )
 
 const (
+	// DefaultOOMBumpUpRatio specifies by default how much memory will be added after observing OOM.
+	DefaultOOMBumpUpRatio float64 = 1.2
+	// DefaultOOMMinBumpUp specifies by default minimal increase of memory after observing OOM.
+	DefaultOOMMinBumpUp float64 = 100 * 1024 * 1024 // 100MB
+)
+
+var (
 	// OOMBumpUpRatio specifies how much memory will be added after observing OOM.
-	OOMBumpUpRatio float64 = 1.2
+	// Ideally bump up ratio should be per VPA configuration.
+	OOMBumpUpRatio = flag.Float64("oom-bump-up-ratio", DefaultOOMBumpUpRatio, `Specifies how much memory will be added after observing OOM.`)
 	// OOMMinBumpUp specifies minimal increase of memory after observing OOM.
-	OOMMinBumpUp float64 = 100 * 1024 * 1024 // 100MB
+	// Ideally min bump up should be per VPA configuration.
+	OOMMinBumpUp = flag.Float64("oom-min-bump-up-ratio", DefaultOOMMinBumpUp, `Specifies minimal increase of memory after observing OOM.`)
 )
 
 // ContainerUsageSample is a measure of resource usage of a container over some
@@ -198,8 +208,8 @@ func (container *ContainerState) RecordOOM(timestamp time.Time, requestedMemory 
 	// Get max of the request and the recent usage-based memory peak.
 	// Omitting oomPeak here to protect against recommendation running too high on subsequent OOMs.
 	memoryUsed := ResourceAmountMax(requestedMemory, container.memoryPeak)
-	memoryNeeded := ResourceAmountMax(memoryUsed+MemoryAmountFromBytes(OOMMinBumpUp),
-		ScaleResource(memoryUsed, OOMBumpUpRatio))
+	memoryNeeded := ResourceAmountMax(memoryUsed+MemoryAmountFromBytes(*OOMMinBumpUp),
+		ScaleResource(memoryUsed, *OOMBumpUpRatio))
 
 	oomMemorySample := ContainerUsageSample{
 		MeasureStart: timestamp,


jbartosik (changes requested):
Thanks for the PR.

I think we should discuss what changes do we want to make and why:

1. I'd like to lear a bit about examples where current values don't work.

2. I'd like to discuss how we should proceed.

If we'll decide to introduce per-VPA-object configuration then I'm not sure I adding flags makes sense.

For example let's consider a scenario:
 - VPA runs with current values (bump ratio 1.2, min bump 100 MB),
 - I create an VPA object `A` I don't set OOM parameters directly,
 - I change flag values (let's say to bump ratio 1.4, min bump 200 MB),

What values is the VPA object using? Those which were default when it was created (1.2, 100MB)? Or the new ones (1.4, 200 MB)?

The first behavior sounds better to me - existing VPA doesn't change behavior without any notification (and we can set values in the VPA object on creation so API looks nice / user can know those values without checking what flags are set). I think this is especially nice in clusters where separate teams are managing VPA (and setting flag values) than using VPA (creating VPA objects).

But it's in tension with what would happen after this change (changing flag values will change behavior of all existing VPA objects).

3. We're getting some requests for tuning VPA parameters / defaults. Before we start making code changes I want to discuss our stategy here. We have at least 3 ways to do that:
- Flags, 
- [Parameters to the chosen recommender](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa#implementation-details),
- Per-VPA-object configuration like ([here](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/enhancements/4566-min-replicas#proposal)).

weiwu-sre:
> 1. I'd like to lear a bit about examples where current values don't work.

We are using VPA for Prometheus and noticed sometimes when it hits OOM and it will stuck in crashloop because VPA doesn't bump up enough memory for it to recover. Some workloads like Prometheus may require more bump up ratio than 20%. I think making it configurable can provide the fixability for us to tune and adjust to our needs. 

> 2. I'd like to discuss how we should proceed.

I am new to this project. I am happy to go though any process to make this feature better.

> If we'll decide to introduce per-VPA-object configuration then I'm not sure I adding flags makes sense.
> 
> For example let's consider a scenario:
> 
> * VPA runs with current values (bump ratio 1.2, min bump 100 MB),
> * I create an VPA object `A` I don't set OOM parameters directly,
> * I change flag values (let's say to bump ratio 1.4, min bump 200 MB),
> 
> What values is the VPA object using? Those which were default when it was created (1.2, 100MB)? Or the new ones (1.4, 200 MB)?
> 
> The first behavior sounds better to me - existing VPA doesn't change behavior without any notification (and we can set values in the VPA object on creation so API looks nice / user can know those values without checking what flags are set). I think this is especially nice in clusters where separate teams are managing VPA (and setting flag values) than using VPA (creating VPA objects).
> 
> But it's in tension with what would happen after this change (changing flag values will change behavior of all existing VPA objects).

For per VPA configuration, by default if no value set for existing VPA resource, or it will keep using default value 1.2 or value set from this flag.

Adding this flag is to solve short term needs for adjusting OOM bump up ratio. Then we can plan and discuss for a long term implementation to introduce per VPA OOM bump up configurations.

If you think we should just plan for the long term, I am happy to close this PR and draft a proposal.
 
> 3. We're getting some requests for tuning VPA parameters / defaults. Before we start making code changes I want to discuss our stategy here. We have at least 3 ways to do that:
> 
> * Flags,
> * [Parameters to the chosen recommender](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa#implementation-details),
> * Per-VPA-object configuration like ([here](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/enhancements/4566-min-replicas#proposal)).

If that's the standard process to introduce a feature, I can create a page like that.

jbartosik:
@weiwu-sre 

Sorry for slow reply.

I thought about this and talked this over. I see two ways forward with this (and similar requests for tuning VPA):

1. Using [customized recommenders](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa):
  1.1 Add flags that allow tuning VPA parameters like this PR does
  1.2 Add a flag that allows VPA to run under a different name than "" and "default"
  1.3 Add instructions for running two VPA recommender instances in one cluster, under different names
  1.4 Now you can run two VPA recommender instances in a cluster - "default" / "" and "higher mem", each VPA object can specify which one to use
  1.5 Profit
2. Extending VPA API - this one is not so straight forward. We'd need to define some autoscaling profiles, add [enhancement](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/enhancements), extend API, use them in the default recommender to change percentiles we use.

Please let me know if you're still interested in this.

PR #4728, opened 2022-03-09
feat: use non-root user for base-image

### LINKED ISSUE #3705: Run cluster-autoscaler container with non-root user
To comply with CIS benchmarking, can we run cluster autoscaler container as non-root user?
We can create cluster-autoscaler image by mentioning below in Dockerfile:
`USER username`

moolen:
There is no need to use the root user. This allows the user to comply with pod security standards.

It turns out that the `nonroot` image has the `cwd` set to `/home/nonroot` whereas `latest` is rooted in `/` :shrug: . 
This is important because the helm chart uses `./cluster-autoscaler` as command. Keeping the same cwd seems like the safest option to me. 

cluster-autoscaler

#### What type of PR is this?
/kind feature

This PR makes infosec happy :smiley: 

Fixes #3705

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/Dockerfile.amd64
+++ b/cluster-autoscaler/Dockerfile.amd64
@@ -11,9 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-ARG BASEIMAGE=gcr.io/distroless/static:latest-amd64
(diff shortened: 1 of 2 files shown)


maltejk:
Is there something left to do on this?

moolen:
Hey @maltejk 😁! Actually i need to set the uid to make it work. I'll commit that later and mark the pr as ready.

moolen:
@feiskyer did you find some time to look into this yet?

PR #4735, opened 2022-03-14
Limit caching pods per owner reference

### LINKED ISSUE #4724: Pod spec equivalency checks can break Cluster Autoscaler scalability
The logic in `buildPodEquivalenceGroups` and `filterOutSchedulable` groups pods by their scheduling requirements, as a scalability optimization. This is done by first grouping by the controller UID, and then comparing pod specs for pods from one controller. If there's something in the pod spec that's unique to a single pod within a controller, every pod ends up in a group of its own, and the optimization breaks.

In extreme cases when there are a lot of such pods (a couple thousand can be enough), CA can spend such a long time in one loop that it fails health-checks and is killed by kubelet. Then everything repeats once it gets back up, and CA is effectively broken until the pods are scheduled or deleted.

One trigger for pod specs being different is the `BoundServiceAccountTokenVolume` feature, which injects uniquely-named projected volumes into each pod's spec. This was taken into account by CA in https://github.com/kubernetes/autoscaler/pull/4441.

We've just run into another one - Jobs using `completionMode: Indexed`. In this mode, each pod gets a unique, indexed hostname in its spec. This is documented here: https://kubernetes.io/docs/concepts/workloads/controllers/job/#completion-mode. AFAIU the hostname shouldn't affect scheduling, so sanitizing it in `PodSpecSemanticallyEqual` should be enough to fix this particular issue.

However, this approach of "fixing" single fields as issues pop up doesn't scale very well. We should come up with a more generic solution to these kinds of problems. One idea could be having a cutoff for the number of groups within one controller, proposed in https://github.com/kubernetes/autoscaler/pull/4441#discussion_r773014959.

x13n:

cluster-autoscaler

#### What type of PR is this?

/kind bug

This limits the number of pods that are searched in O(N) time per owner reference to a small fixed constant (10), to avoid future regressions when PodSpec is extended with new fields that differ between pods owned by the same controller. I added a new metric for tracking this, so that we can monitor & fix cases when that happens. This should keep the caching optimization maintainable in the future.

Fixes #4724

```release-note
NONE
```

```docs

```

/assign @MaciekPytel

### DIFF
--- a/cluster-autoscaler/core/equivalence_groups.go
+++ b/cluster-autoscaler/core/equivalence_groups.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/autoscaler/cluster-autoscaler/processors/status"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/drain"
+	pod_utils "k8s.io/autoscaler/cluster-autoscaler/utils/pod"
 )
 
 type podEquivalenceGroup struct {
@@ -54,6 +55,7 @@ type equivalenceGroup struct {
 const maxEquivalenceGroupsByController = 10
 
 // groupPodsBySchedulingProperties groups pods based on scheduling properties. Group ID is meaningless.
+// TODO(x13n): refactor this to have shared logic with PodSchedulableMap.
 func groupPodsBySchedulingProperties(pods []*apiv1.Pod) map[equivalenceGroupId][]*apiv1.Pod {
 	podEquivalenceGroups := map[equivalenceGroupId][]*apiv1.Pod{}
 	equivalenceGroupsByController := make(map[types.UID][]equivalenceGroup)
@@ -61,7 +63,7 @@ func groupPodsBySchedulingProperties(pods []*apiv1.Pod) map[equivalenceGroupId][
 	var nextGroupId equivalenceGroupId
 	for _, pod := range pods {
 		controllerRef := drain.ControllerRef(pod)
-		if controllerRef == nil {
+		if controllerRef == nil || pod_utils.IsDaemonSetPod(pod) {
 			podEquivalenceGroups[nextGroupId] = []*apiv1.Pod{pod}
 			nextGroupId++
 			continue
--- a/cluster-autoscaler/core/equivalence_groups.go
+++ b/cluster-autoscaler/core/equivalence_groups.go
@@ -51,6 +51,8 @@ type equivalenceGroup struct {
 	representant *apiv1.Pod
 }
 
+const maxEquivalenceGroupsByController = 10
+
 // groupPodsBySchedulingProperties groups pods based on scheduling properties. Group ID is meaningless.
 func groupPodsBySchedulingProperties(pods []*apiv1.Pod) map[equivalenceGroupId][]*apiv1.Pod {
 	podEquivalenceGroups := map[equivalenceGroupId][]*apiv1.Pod{}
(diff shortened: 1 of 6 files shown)


x13n:
/hold 

I'll add one more commit to prevent DaemonSets (incl. custom ones) from using this logic at all.

cluster-autoscaler/core/equivalence_groups.go
```diff
@@ -51,39 +53,54 @@ type equivalenceGroup struct {
...
-				representant: pod,
-			}
-			equivalenceGroupsByController[controllerRef.UID] = append(equivalenceGroupsByController[controllerRef.UID], newGroup)
-			podEquivalenceGroups[newGroup.id] = append(podEquivalenceGroups[newGroup.id], pod)
-			nextGroupId++
+		newGroup := equivalenceGroup{
+			id:           nextGroupId,
+			representant: pod,
 		}
+		equivalenceGroupsByController[controllerRef.UID] = append(egs, newGroup)
```
MaciekPytel, line 82: I think the current implementation works, but it's surprising to me that we use the maxEquivalenceGroupsByController to limit the lookup through equivalence groups list, not to limit the size of the list itself. I think the code would be easier to understand if we just added a more explicit check + continue, rather than keep extending the list and bypass comparison like we do now.

Also I highly doubt it makes any real performance difference, but it feels pointless to create all those equivalenceGroup structs if we're never going to return them or look into them.

cluster-autoscaler/core/utils/pod_schedulable.go
```diff
@@ -73,11 +77,19 @@ func (podMap PodSchedulableMap) Get(pod *apiv1.Pod) (*simulator.PredicateError,
...
 func (podMap PodSchedulableMap) Set(pod *apiv1.Pod, err *simulator.PredicateError) {
 	ref := drain.ControllerRef(pod)
-	if ref == nil {
+	if ref == nil || pod_utils.IsDaemonSetPod(pod) {
 		return
 	}
 	uid := string(ref.UID)
-	podMap[uid] = append(podMap[uid], PodSchedulableInfo{
+	pm := podMap[uid]
+	if len(pm) >= maxPodsPerOwnerRef {
```
MaciekPytel, line 85: The end result is roughly the same<sup>1</sup>, but it's implemented in a completely different way (limiting list size vs short-circuiting lookups). Wouldn't it be better to use a similar implementation? Arguably it makes code more readable (more clear the logic is the same) and (admittedly, minimally) reduces the risk of diverging implementations.

<sup>1</sup> technically with this version we still do 10*#pods comparisons and with the other we don't, but I'm not sure this makes practical difference).

cluster-autoscaler/metrics/metrics.go
```diff
@@ -304,6 +304,14 @@ var (
 		},
 	)
 
+	podsPerOwnerRefOverflowCount = k8smetrics.NewCounter(
+		&k8smetrics.CounterOpts{
+			Namespace: caNamespace,
+			Name:      "pods_per_owner_ref_overflow_count",
+			Help:      "Number of pods that were rejected from internal caching, due to high number of different pods under the same owner reference.",
```
MaciekPytel, line 311: I find this definition of metric a bit surprising:
 * If we're after pod count wouldn't a gauge be a better choice? With counter it's inherently a function of both pod count and how often we run the logic (so CA loop time).
 * I guess the first 10 pods weren't technically 'rejected from internal caching', but the fact that we're reporting the number of affected pods - 10 feels arbitrary.
 * Aren't we double-counting the pods in any loop where we trigger scale-up?

Maybe having a gauge counting controllers with enough pods to disable caching would be easier to count precisely and provide a more meaningful value? Admittedly, it may not matter very much, since the main use of this seems to be checking if the number is >0 or not.

MaciekPytel (commented):
Really glad to see this addressed. I left a few nitpicks, but overall the implementation looks good.

/approve

cluster-autoscaler/core/utils/pod_schedulable.go
MaciekPytel: +1 to not worth the time :)

cluster-autoscaler/metrics/metrics.go
MaciekPytel: Good point on scraping interval, I haven't thought about that. Since we count controllers in filterOutSchedulable, the value should not go down for the duration of node startup which I think should be long enough for scraping? I guess you agree, since I see you changed it to a gauge? You've much more experience with metrics than I do, so I'll defer to your judgement.

PR #4738, opened 2022-03-15
Fix CI script to use GO111MODULE auto as default

yankay:
The cluster-autoscaler use godep, and the  cluster-autoscaler vertical-pod-autoscaler use gomod.

So should we set GO111MODULE=auto as default. and GO111MODULE=off only for the  cluster-autoscaler project.

Signed-off-by: Kay Yan <kay.yan@daocloud.io>


yankay:
HI @losipiuk, would you please review the PR?

mwielgus (approved):
/lgtm
/approve

PR #4753, opened 2022-03-21
cluster-autoscaler/cloudstack: Identify node by name and id

davidjumani:

cluster-autoscaler / cloudprovider / cloudstack

#### What type of PR is this?

/kind feature

Previously, nodes were identified by their system UUID. This sometimes causes issues as the hypervisor has a VM UUID, which differs from the one CloudStack uses. In such a case, identify the node by name


mwielgus (approved):
/lgtm
/approve

mwielgus:
PR approved based on the maintainer/owner status of @davidjumani.

Issue #4784, opened 2022-04-05
cluster-autoscaler caches workload cluster kubeconfig


charlie-haley:
**Which component are you using?**: cluster-autoscaler

**What version of the component are you using?**: 1.21.2

**What k8s version are you using (`kubectl version`)?**:

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.5", GitCommit:"c285e781331a3785a7f436042c65c5641ce8a9e9", GitTreeState:"archive", BuildDate:"2022-03-17T21:14:47Z", GoVersion:"go1.18", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21+", GitVersion:"v1.21.5-eks-bc4871b", GitCommit:"5236faf39f1b7a7dabea8df12726f25608131aa9", GitTreeState:"clean", BuildDate:"2021-10-29T23:32:16Z", GoVersion:"go1.16.8", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What environment is this in?**: EKS managed by Cluster API

**What did you expect to happen?**:

The autoscaler should use the renewed kubeconfig for the workload cluster.

**What happened instead?**:

After around ~10 minutes the token renews in the kubeconfig secret for the workload cluster and the autoscaler no longer works, the pod has to be killed for it to pick up the new token.

**How to reproduce it (as minimally and precisely as possible)**:

Deploy cluster-autoscaler on a CAPI management cluster, pointing at a workload cluster.

Example Helm values:
```yaml
additionalLabels:
  app: autoscaler
autoDiscovery:
  clusterName: mycluster-name
cloudProvider: clusterapi
clusterAPIConfigMapsNamespace: kube-system
clusterAPIKubeconfigSecret: mycluster-name-kubeconfig
clusterAPIMode: kubeconfig-incluster
extraArgs:
  balance-similar-node-groups: true
  expander: least-waste
  leader-elect: false
  logtostderr: true
  stderrthreshold: error
  v: 1
```

Skarlso:
@charlie-haley Hello 👋 
So I'm about to do something about this thing. I see you did some annotation thing in the attached PR. Were you planning on reloading the whole thing every 10 minutes?

[...]

PR #4791, opened 2022-04-06
feat: add support for deployment annotations to helm chart

charlie-haley:
helm charts

#### What type of PR is this?
/kind feature

Due to this issue https://github.com/kubernetes/autoscaler/issues/4784, we need access to annotations on the Deployment so we can reload the pod every time the workload cluster kubeconfig updates.

```release-note
new property in values: added 'annotations' property to allow setting annotations on the Deployment
```

### DIFF
--- a/charts/cluster-autoscaler/README.md
+++ b/charts/cluster-autoscaler/README.md
@@ -287,7 +287,6 @@ Though enough for the majority of installations, the default PodSecurityPolicy _
 |-----|------|---------|-------------|
 | additionalLabels | object | `{}` | Labels to add to each object of the chart. |
 | affinity | object | `{}` | Affinity for pod assignment |
-| annotations | object | `{}` | Annotations to add to the Deployment object. |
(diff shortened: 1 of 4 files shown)


charts/cluster-autoscaler/README.md
```diff
@@ -287,6 +287,7 @@ Though enough for the majority of installations, the default PodSecurityPolicy _
 |-----|------|---------|-------------|
 | additionalLabels | object | `{}` | Labels to add to each object of the chart. |
 | affinity | object | `{}` | Affinity for pod assignment |
+| annotations | object | `{}` | Annotations to add to the Deployment object. |
```
gjtempleton, line 290: I'm not keen on having a root value adding annotations only to the deployment. This should either be `deployment.annotations` (like `rbac.serviceAccount.annotations`) or apply to all objects in the chart as `additionalLabels` does and maybe move to being called `additionalAnnotations` to match.

gjtempleton (changes requested):
Hey, thanks for the PR.

One request, and one inform on how to get the PR checks passing.

gjtempleton:
Thanks!

/lgtm
/approve

PR #4798, opened 2022-04-07
Add resource validation

### LINKED ISSUE #4790: VPA should validate that resources are specified with supported resolution
Follow up to #4774

VPA should validate that all fields that specify memory and CPU have supported resolution:

* CPU should be a whole number of milli CPUs ([documentation](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu)),
* Memory should be a whole number of bytes ([documentation](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory)).

Currently one can specify memory and CPU with higher resolution than K8s supports, with undesirable results (see #4774 for details).

VPA admission controller already does [validation](https://github.com/kubernetes/autoscaler/blame/1fa0716db97a0ce2b9785a954ffce5611c7a2d67/vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler.go#L109) of VPA objects.

In the validation we should add checks all fields which specify CPU or memory. Currently we have 2 such fields:

* [MinAllowed](https://github.com/kubernetes/autoscaler/blob/1fa0716db97a0ce2b9785a954ffce5611c7a2d67/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L170) and
* [MaxAllowed](https://github.com/kubernetes/autoscaler/blob/1fa0716db97a0ce2b9785a954ffce5611c7a2d67/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L174).

mowangdk:

vertical-pod-autoscaler

#### What type of PR is this?

/kind bug

Add Validation for VPA

Fixes #4790

```release-note
None
```

```docs
None
```

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler.go
@@ -19,7 +19,10 @@ package vpa
 import (
 	"encoding/json"
 	"fmt"
+
 	v1 "k8s.io/api/admission/v1"
+	corev1 "k8s.io/api/core/v1"
+	apires "k8s.io/apimachinery/pkg/api/resource"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/admission-controller/resource"
@@ -133,9 +136,18 @@ func validateVPA(vpa *vpa_types.VerticalPodAutoscaler, isCreate bool) error {
 				}
 			}
 			for resource, min := range policy.MinAllowed {
+				if err := validateResourceResolution(resource, min); err != nil {
+					return fmt.Errorf("MinAllowed: %v", err)
+				}
 				max, found := policy.MaxAllowed[resource]
 				if found && max.Cmp(min) < 0 {
 					return fmt.Errorf("max resource for %v is lower than min", resource)
+				}
+			}
+
+			for resource, max := range policy.MaxAllowed {
+				if err := validateResourceResolution(resource, max); err != nil {
+					return fmt.Errorf("MaxAllowed: %v", err)
 				}
 			}
 			ControlledValues := policy.ControlledValues
@@ -155,5 +167,29 @@ func validateVPA(vpa *vpa_types.VerticalPodAutoscaler, isCreate bool) error {
 		return fmt.Errorf("The current version of VPA object shouldn't specify more than one recommenders.")
 	}
 
+	return nil
+}
+
+func validateResourceResolution(name corev1.ResourceName, val apires.Quantity) error {
+	switch name {
+	case corev1.ResourceCPU:
+		return validateCPUResolution(val)
+	case corev1.ResourceMemory:
+		return validateMemoryResolution(val)
+	}
+	return nil
+}
+
+func validateCPUResolution(val apires.Quantity) error {
+	if _, precissionPreserved := val.AsScale(apires.Milli); !precissionPreserved {
+		return fmt.Errorf("CPU [%s] must be a whole number of milli CPUs", val.String())
+	}
+	return nil
+}
+
+func validateMemoryResolution(val apires.Quantity) error {
+	if _, precissionPreserved := val.AsScale(0); !precissionPreserved {
+		return fmt.Errorf("Memory [%v] must be a whole number of bytes", val)
+	}
 	return nil
 }
--- a/vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler_test.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler_test.go
@@ -27,7 +27,8 @@ import (
 )
 
 const (
-	cpu = apiv1.ResourceCPU
+	cpu    = apiv1.ResourceCPU
+	memory = apiv1.ResourceMemory
 )
 
 func TestValidateVPA(t *testing.T) {
@@ -36,6 +37,7 @@ func TestValidateVPA(t *testing.T) {
 	badMinReplicas := int32(0)
 	validMinReplicas := int32(1)
 	badScalingMode := vpa_types.ContainerScalingMode("bad")
+	badCPUResource := resource.MustParse("187500u")
 	validScalingMode := vpa_types.ContainerScalingModeAuto
 	scalingModeOff := vpa_types.ContainerScalingModeOff
 	controlledValuesRequestsAndLimits := vpa_types.ContainerControlledValuesRequestsAndLimits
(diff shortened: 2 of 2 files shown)


mowangdk:
@jbartosik ptal

mowangdk:
@jbartosik PTAL

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler.go
```diff
@@ -157,3 +170,19 @@ func validateVPA(vpa *vpa_types.VerticalPodAutoscaler, isCreate bool) error {
 
 	return nil
 }
+
+func validateResource(name corev1.ResourceName, val apires.Quantity) error {
```
jbartosik, line 174: We don't share any code between CPU and memory. I think it would be simpler to have two functions (for example `validateCPUResolution` and `validateMemoryResolution`)

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler.go
```diff
@@ -157,3 +170,19 @@ func validateVPA(vpa *vpa_types.VerticalPodAutoscaler, isCreate bool) error {
...
+	switch name {
+	case corev1.ResourceCPU:
+		strValue := val.String()
```
jbartosik, line 177: I think using [`AsScale`](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity.AsScale) will work better.
```
if _, precissionPreserved := val.AsScale(resource.Milli); !precissionPreserved {
 //...
```

For memory we're rounding to whole units, so there is no prefix ([docs](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Scale)), so you should use `AsScale(0)`.

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler_test.go
```diff
@@ -150,6 +152,50 @@ func TestValidateVPA(t *testing.T) {
...
+				Spec: vpa_types.VerticalPodAutoscalerSpec{
+					ResourcePolicy: &vpa_types.PodResourcePolicy{
+						ContainerPolicies: []vpa_types.ContainerResourcePolicy{
+							{
+								ContainerName: "loot box",
+								MinAllowed: apiv1.ResourceList{
+									cpu: badCPUValue,
+								},
+								MaxAllowed: apiv1.ResourceList{
+									cpu: resource.MustParse("275m"),
```
jbartosik, line 167: `275m` is a whoe number of milli CPUs

jbartosik (changes requested):
Thank you for the change. I left a few comments. (Also I'll respond faster in the future)

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler.go
mowangdk: okay

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler.go
mowangdk: gotcha, I'm not familiar with this quantity object. thank you for your comments

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler_test.go
mowangdk: This test case will exit due to memory minAllow invalidate(100m)  and will not go to the maxcpu validation

mowangdk:
@jbartosik comments fixed, ptal

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler_test.go
```diff
@@ -150,6 +152,71 @@ func TestValidateVPA(t *testing.T) {
...
+								},
+							},
+						},
+					},
+				},
+			},
+			expectError: fmt.Errorf("Memory [%v] must be a whole number of bytes", resource.MustParse("100m")),
+		},
+		{
+			name: "bad manAllowed cpu value",
```
jbartosik, line 200: typo: s/manAllowed/maxAllowed/

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler_test.go
```diff
@@ -36,6 +37,7 @@ func TestValidateVPA(t *testing.T) {
 	badMinReplicas := int32(0)
 	validMinReplicas := int32(1)
 	badScalingMode := vpa_types.ContainerScalingMode("bad")
+	badCPUValue := resource.MustParse("187500u")
```
jbartosik, line 40: Please use resource.MustParse("187500u") directly, like with other values.

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler_test.go
```diff
@@ -150,6 +152,71 @@ func TestValidateVPA(t *testing.T) {
...
+									cpu: badCPUValue,
+								},
+							},
+						},
+					},
+				},
+			},
+			expectError: fmt.Errorf("CPU [%s] must be a whole number of milli CPUs", badCPUValue.String()),
+		},
 		{
```
jbartosik, line 220: Please add a case for invalid maxAllowed memory

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler.go
```diff
@@ -133,9 +136,17 @@ func validateVPA(vpa *vpa_types.VerticalPodAutoscaler, isCreate bool) error {
...
 			}
 			for resource, min := range policy.MinAllowed {
+				if err := validate(resource, min); err != nil {
+					return err
+				}
 				max, found := policy.MaxAllowed[resource]
-				if found && max.Cmp(min) < 0 {
-					return fmt.Errorf("max resource for %v is lower than min", resource)
+				if found {
+					if err := validate(resource, max); err != nil {
```
jbartosik, line 144: We should check if max has correct resolution in a loop:

```
for resource, min := range policy.MaxAllowed {
  if err := validate(resource, max); err != nil {
    //...
```

current implementation doesn't validate max when min is not specified.

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler.go
```diff
@@ -133,9 +136,17 @@ func validateVPA(vpa *vpa_types.VerticalPodAutoscaler, isCreate bool) error {
 				}
 			}
 			for resource, min := range policy.MinAllowed {
+				if err := validate(resource, min); err != nil {
+					return err
```
jbartosik, line 140: It would be nice to mention in the error message which field has improper value.

You can add this information here:

```
  return fmt.Errorf("MinAllowed: %s", err)
```

or pass the name of the field to `validate` so it can include it in the error message it returns.

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler.go
```diff
@@ -157,3 +168,27 @@ func validateVPA(vpa *vpa_types.VerticalPodAutoscaler, isCreate bool) error {
 
 	return nil
 }
+
+func validate(name corev1.ResourceName, val apires.Quantity) error {
```
jbartosik, line 172: Please make the name more specific. For example `validateResourceResolution`

jbartosik (changes requested):
Thank you. I think this is close to being ready for submit.

There is one change needed (validate resolution of max even if there is no corresponding min). I also left a few requests for smaller improvements.

mowangdk:
@jbartosik ptal, There is no way to call string() directly on ```resource.MustParse("187500u")``` return value , so i kept badCPUResource

jbartosik (approved):
Looks good, only a few small comments left:

- No need to change if
- Consistent approach to amounts in unit tests
- Please squash commits

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler.go
```diff
@@ -133,9 +136,20 @@ func validateVPA(vpa *vpa_types.VerticalPodAutoscaler, isCreate bool) error {
 				}
 			}
 			for resource, min := range policy.MinAllowed {
+				if err := validateResourceResolution(resource, min); err != nil {
+					return fmt.Errorf("MinAllowed: %v", err)
+				}
 				max, found := policy.MaxAllowed[resource]
-				if found && max.Cmp(min) < 0 {
-					return fmt.Errorf("max resource for %v is lower than min", resource)
+				if found {
```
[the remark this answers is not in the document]
mowangdk: gotcha, i forgot to recover it

mowangdk:
@jbartosik make some changes according to your comments, ptal

jbartosik (approved):
Thank you

/lgtm
/approve

PR #4804, opened 2022-04-08
Support for dynamic SKUs for scaling from zero scenario

gandhipr:
cluster-autoscaler

#### What type of PR is this?
/kind feature

Currently, cluster autoscaler(azure) uses hard-coded (static) list of instanceTypes to scale from zero as there is no node to build blueprint of the information from. This static list needs to updated every-time a new SKU is added in Azure which is not feasible. This PR enables fetching InstanceType information dynamically for scale form zero scenario.

Fixes #

NONE

```release-note

```

```docs

```


marwanad:
nit on description: This static list needs to updated every-time a new VMSS should be "every-time a new SKU is added in Azure"

cluster-autoscaler/cloudprovider/azure/azure_instance_types.go
```diff
@@ -78,18 +78,6 @@ var InstanceTypes = map[string]*InstanceType{
 		MemoryMb:     1024,
 		GPU:          0,
 	},
-	"Standard_A10": {
```
marwanad, line 81: any reason those were removed?

cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -169,4 +169,6 @@ type AutoscalingOptions struct {
 	DaemonSetEvictionForOccupiedNodes bool
 	// User agent to use for HTTP calls.
 	UserAgent string
+	// AzureEnableDynamicInstanceList enables dynamic instance workflow for vmss check
+	AzureEnableDynamicInstanceList bool
```
marwanad, line 173: usually not a fan of having provider-specific flags on the global flags. 

@gjtempleton thoughts on merging this with the AWS one? We could just have a generic `EnableDynamicInstanceInformation` flag and deprecate the AWS one eventually since I forsee other providers benefiting from that too.

gjtempleton: I'd be all in favour of merging it to a single flag, only challenge may be co-ordination with other cloud providers, I'm not sure what the current split is around default behaviours being dynamic or hard coded, and how many (if any) others support both.

cluster-autoscaler/cloudprovider/azure/azure_instance_types.go
gandhipr: Running go gen and regnerated this file.

cluster-autoscaler/cloudprovider/azure/azure_config.go
```diff
@@ -128,6 +128,9 @@ type Config struct {
 	CloudProviderBackoffExponent float64 `json:"cloudProviderBackoffExponent,omitempty" yaml:"cloudProviderBackoffExponent,omitempty"`
 	CloudProviderBackoffDuration int     `json:"cloudProviderBackoffDuration,omitempty" yaml:"cloudProviderBackoffDuration,omitempty"`
 	CloudProviderBackoffJitter   float64 `json:"cloudProviderBackoffJitter,omitempty" yaml:"cloudProviderBackoffJitter,omitempty"`
+
+	// AzureEnableDynamicInstanceList defines whether to enable dynamic instance workflow for instance information check
+	AzureEnableDynamicInstanceList bool `json:"azureEnableDynamicInstanceList,omitempty" yaml:"azureEnableDynamicInstanceList,omitempty"`
```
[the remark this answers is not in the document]
gandhipr: but main.go needs to have azure... and kept it in-sync with that name.

cluster-autoscaler/cloudprovider/azure/azure_instance_types.go
marwanad: you're likely running it against a subscription that doesn't have those SKUs or those SKUs are deprecated. Let's stick to an append-only for this file without removing entries.

cluster-autoscaler/cloudprovider/azure/azure_config.go
```diff
@@ -59,6 +59,9 @@ const (
 	// auth methods
 	authMethodPrincipal = "principal"
 	authMethodCLI       = "cli"
+
+	// toggle
+	dynamicInstanceListDefault = false
```
marwanad, line 64: you can get rid of this and set to false directly.

marwanad: like the `EnableDynamicInstanceList` boolean will always default to false if unspecified so no need.

marwanad: I just saw the other fields do that but it feels odd lol

cluster-autoscaler/cloudprovider/azure/azure_instance.go
```diff
@@ -0,0 +1,107 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package azure
+
+import (
+	"context"
+	"fmt"
+	compute20190701 "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
```
marwanad, line 22: why can't we use the API version below? 2020-12-01? Is it because of the `skewer` dependency?

cluster-autoscaler/cloudprovider/azure/azure_config.go
gandhipr: Yes, that's why I followed this pattern. In that case, I won't make any changes here?

cluster-autoscaler/cloudprovider/azure/azure_instance.go
gandhipr: Yes, skewer depends on 2019-07-01

cluster-autoscaler/cloudprovider/azure/azure_config.go
```diff
@@ -212,6 +218,15 @@ func BuildAzureConfig(configReader io.Reader) (*Config, error) {
 			}
 		}
 
+		if enableDynamicInstanceList := os.Getenv("AZURE_ENABLE_DYNAMIC_INSTANCE_LIST"); enableDynamicInstanceList != "" {
```
marwanad, line 221: nit: can we have this as `AUTOSCALER_ENABLE_DYNAMIC_INSTANCE_LIST` and prefix the json and yaml keys with `autoscalerEnableDynamicInstanceList`.

I presume a lot of ppl may have that secret mounted to both CA and cloudprovider so it might be worth having a prefix.

marwanad (approved):
/lgtm

just one minor comment. You might need Pengfei's approval for the vendor changes.

feiskyer (commented):
/lgtm
/approve

PR #4807, opened 2022-04-11
Introduce extraction of System Architecture from AutoscalerVars 

jayantjain93:
Introduce extraction of System Architecture from AutoscalerVars  in GCE cloud-provider

cluster-autoscaler

#### What type of PR is this?
/kind feature

Fixes #

None

```release-note

```

```docs

```

/assign @x13n

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/templates.go
+++ b/cluster-autoscaler/cloudprovider/gce/templates.go
@@ -59,7 +59,8 @@ func (t *GceTemplateBuilder) getAcceleratorCount(accelerators []*gce.Accelerator
 }
 
 // BuildCapacity builds a list of resource capacities given list of hardware.
-func (t *GceTemplateBuilder) BuildCapacity(cpu int64, mem int64, accelerators []*gce.AcceleratorConfig, os OperatingSystem, osDistribution OperatingSystemDistribution, ephemeralStorage int64, ephemeralStorageLocalSSDCount int64, pods *int64, version string, r OsReservedCalculator) (apiv1.ResourceList, error) {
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -531,6 +536,40 @@ func extractOperatingSystemDistributionFromImageType(imageType string) Operating
...
+
+const (
+	// UnknownArch is used if the Architecture is Unknown
+	UnknownArch SystemArchitecture = ""
+	// Amd64 is used if the Architecture is x86_64
+	Amd64 SystemArchitecture = "amd64"
+	// Arm64 is used if the Architecture is ARM
+	Arm64 SystemArchitecture = "arm64"
+	// DefaultArch is used if the Architecture is used as a fallback if not passed by AUTOSCALER_ENV_VARS
+	DefaultArch SystemArchitecture = Amd64
```
x13n, line 550: `DefaultArch` currently lives in cloudprovider/utils. It isn't really GCE-specific, so why move it here?

jayantjain93: It is similar to `DefaultOS`. We were using common CA defaults. But since GCE now explicitly support arch, it seems appropriate to use cloudprovider-specific default value (even if the values are the same). We now have cloudspecific ability to control default as well.

This is also similar impl to OS, distribution, image-type.

x13n: Ok, I buy the consistency argument, let's keep it here for now. I'd consider refactoring non-GCP-specific parts (OS, arch) to some common dir in the future though.

PR #4810, opened 2022-04-12
Continue CA loop when unregistered nodes were removed

yaroslava-serdiuk:

Cluster Autoscaler

#### What type of PR is this?

Add one of the following kinds:
/kind bug

If LongUnregistered nodes are present and couldn't be removed for long time they block Cluster Autoscaler. 

Basic Scale Up and Scale Down are tested with current change. 

```release-note
LongUnregistered nodes don't block Cluster Autoscaler. 
```


FeLvi-zzz:
I'd like this fix, I'm going to cherry-pick this PR into 1.21, 1.22, 1.23.

PR #4812, opened 2022-04-13
Remove newline indentation from CA deployment 

iainbowman:

cluster-autoscaler

#### What type of PR is this?

/kind bug

The `extraVolumeMounts` section of the deployment template currently adds an unnecessary blank line (whitespace) in the rendered yaml output 

Current rendered output example:

```yaml
          volumeMounts:
            
            - mountPath: /etc/ssl/certs/ca-certificates.crt
              name: ssl-certs
              readOnly: true
```

Rendered output example with this fix applied:

```yaml
          volumeMounts:
            - mountPath: /etc/ssl/certs/ca-certificates.crt
              name: ssl-certs
              readOnly: true
```

Fixes #

```release-note
Removes whitespace from the templated deployment output when the `extraVolumeMounts` option is used.
```

```docs

```

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -17,4 +17,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.17.0
+version: 9.17.1
--- a/charts/cluster-autoscaler/templates/deployment.yaml
+++ b/charts/cluster-autoscaler/templates/deployment.yaml
@@ -229,7 +229,7 @@ spec:
               mountPath: {{ .Values.clusterAPIWorkloadKubeconfigPath | trimSuffix "/value" }}
           {{- end }}
           {{- if .Values.extraVolumeMounts }}
-            {{ toYaml .Values.extraVolumeMounts | nindent 12 }}
+            {{ toYaml .Values.extraVolumeMounts | indent 12 }}
           {{- end }}
           {{- end }}
     {{- if .Values.affinity }}


gjtempleton:
Thanks for the PR.

You'll need to rebase the chart version bump as I've just merged another chart PR.

gjtempleton:
Thanks, one last thing, could you squash the commits down to a single commit please?

gjtempleton:
Thanks!
/lgtm
/approve

stevehipwell:
@iainbowman @gjtempletonI I don't think this is the correct fix and has broken the chart due to removing the correct indentation logic provided by `nindent`. The fix here was to make sure that the directive had a `-` to make the `n` of `nindent` not print, so the correct code should be as follows.

```yaml
          {{- if .Values.extraVolumeMounts }}
            {{- toYaml .Values.extraVolumeMounts | nindent 12 }}
          {{- end }}
```

A better pattern would be as follows as it is easier to grep.

```yaml
          {{- with .Values.extraVolumeMounts }}
            {{- toYaml . | nindent 12 }}
          {{- end }}
```

gjtempleton:
@stevehipwell Thanks for raising, have just merged #4819 to get a fix out, though it makes the minimal change necessary as suggested in your top block.

Going to look into improving the testing of these optional blocks around values that default to empty, have you got any examples of enforcing that kind of testing at PR on any of your charts?

stevehipwell:
@gjtempleton I currently use the manual review process for the charts I manage as I know what to look out for by having made all of these mistakes myself at least once. This is a "Brent" solution and isn't scalable so I'm working on an automation design which can be added to idiomatic charts to cover the general cases and easily be customised for new cases which use an idiomatic pattern.

I still don't have @kubernetes membership so I can't be an official maintainer here but I'm more than happy to review some of your changes before you have more automation coverage.

PR #4818, opened 2022-04-18
update azure instance types

Prashanth684:
with the latest az 2.35 to include the new arm64 instances

/area provider/azure


marwanad:
There's a lot of instances that got removed with your change. Can you revert those?

We try and keep it as an append-only list since the subscription that you ran the instance generation on might not have all instances.

There's a PR in flight to switch to dynamic instance fetching so hopefully that step will no longer be needed.

Prashanth684:
> There's a lot of instances that got removed with your change. Can you revert those?
> 
> We try and keep it as an append-only list since the subscription that you ran the instance generation on might not have all instances.
> 
> There's a PR in flight to switch to dynamic instance fetching so hopefully that step will no longer be needed.

thanks @marwanad . indeed, i'd forgotten that my specific subscription was not all inclusive. added back the removed instances

marwanad:
Thank you!

/lgtm
/approve

Issue #4764, opened 2022-03-24
VPA e2e tests are all failing


jbartosik:
VPA actuation, admission controller, full, recommender, and updater e2e are all failing (and have been failing as fart as the dashboard goes (two weeks back)):

https://k8s-testgrid.appspot.com/sig-autoscaling-vpa

jbartosik:
I took a look at a few failures from different test suites and they all look like [this one](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-autoscaling-vpa-updater/1506887456606130176):

```
I0324 07:06:46.206] STEP: Creating a kubernetes client
I0324 07:06:46.207] Mar 24 07:06:46.204: INFO: >>> kubeConfig: /workspace/.kube/config
I0324 07:06:46.207] STEP: Building a namespace api object, basename vertical-pod-autoscaling
I0324 07:08:46.265] Mar 24 07:08:46.262: FAIL: Unexpected error:
I0324 07:08:46.265]     <*errors.errorString | 0xc0001ce100>: {
I0324 07:08:46.265]         s: "timed out waiting for the condition",
I0324 07:08:46.266]     }
I0324 07:08:46.266]     timed out waiting for the condition
I0324 07:08:46.266] occurred
I0324 07:08:46.266] 
I0324 07:08:46.266] Full Stack Trace
I0324 07:08:46.266] k8s.io/kubernetes/test/e2e/framework.(*Framework).BeforeEach(0xc00043b1e0)
I0324 07:08:46.266] 	/go/src/k8s.io/autoscaler/vertical-pod-autoscaler/e2e/vendor/k8s.io/kubernetes/test/e2e/framework/framework.go:236 +0x7ff
I0324 07:08:46.267] command-line-arguments.RunE2ETests(0x48dc37?)
I0324 07:08:46.267] 	/go/src/k8s.io/autoscaler/vertical-pod-autoscaler/e2e/v1beta2/e2e.go:121 +0x61b
I0324 07:08:46.267] command-line-arguments.TestE2E(0x0?)
I0324 07:08:46.267] 	/go/src/k8s.io/autoscaler/vertical-pod-autoscaler/e2e/v1beta2/e2e_test.go:83 +0x19
I0324 07:08:46.268] testing.tRunner(0xc000503860, 0x291e280)
I0324 07:08:46.268] 	/usr/local/go/src/testing/testing.go:1439 +0x102
[...]

PR #4821, opened 2022-04-19
Work around VPA e2e test failures

jbartosik:
/kind bug
/kind failing-test

Workaround for #4764


jbartosik:
With this change I can see that tests are at least running in my cluster. I want to submit this so we can see if there are any problems masked underneath

krzysied:
\lgtm

PR #4723, opened 2022-03-04
Fix lint issues in vertical pod autoscaller

Fedosin:
#### Which component this PR applies to?

vertical-pod-autoscaler

#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

This is a code refactoring for vertical pod autoscaler aimed to meet current golang code formatting requirements.

#### Does this PR introduce a user-facing change?

No

```release-note
None
```

PR #4824, opened 2022-04-20
Fix updater main

jbartosik:
/kind bug
/kind regression

This should fix at least some e2e test failures, maybe all of them (#4764)

#4723 moved context creation from inside loop (correct) to doing it once, before loop starts (incorrect, context will be done in all loop iterations after the first one).

Fix a bug we had in the old version (using `defer` in a loop would call the deferred function only after we exit `main` not as expected on each loop iteration). Thanks @krzysied for pointing out the bug

@Fedosin FYI

### DIFF
--- a/vertical-pod-autoscaler/pkg/updater/main.go
+++ b/vertical-pod-autoscaler/pkg/updater/main.go
@@ -113,10 +113,10 @@ func main() {
 		klog.Fatalf("Failed to create updater: %v", err)
 	}
 	ticker := time.Tick(*updaterInterval)
-	ctx, cancel := context.WithTimeout(context.Background(), *updaterInterval)
-	defer cancel()
 	for range ticker {
+		ctx, cancel := context.WithTimeout(context.Background(), *updaterInterval)
 		updater.RunOnce(ctx)
 		healthCheck.UpdateLastActivity()
+		cancel()
 	}
 }


jbartosik:
I verified that this fixes at least some problems by:
- Running VPA updater v1 "evicts pods when Admission Controller status available" tests with updater built from master and observing it fail,
- Running the same tests with updater from master + this change and observing it pass.

vertical-pod-autoscaler/pkg/updater/main.go
```diff
@@ -113,9 +113,9 @@ func main() {
 		klog.Fatalf("Failed to create updater: %v", err)
 	}
 	ticker := time.Tick(*updaterInterval)
-	ctx, cancel := context.WithTimeout(context.Background(), *updaterInterval)
-	defer cancel()
 	for range ticker {
+		ctx, cancel := context.WithTimeout(context.Background(), *updaterInterval)
+		defer cancel()
```
krzysied, line 118: What is defer vs for interaction. Are defers run at the end of a for iteration or after exiting a for?

jbartosik: They're run when the function returns. So here we would never run the defer. Thanks for catching the bug!

I'll check if we have similar bug repeated elsewhere (using defer in main / possible a different function from which we don't return)

PR #4831, opened 2022-04-27
Add KEP: Control VPA eviction behavior based on scaling direction and resource

voelzmo:

vertical-pod-autoscaler
#### What type of PR is this?

/kind feature
/kind api-change

Discuss introduction of a new `UpdateMode` for VPA as previously discussed in the SIG meeting.

```release-note
NONE
```

```docs

```


jbartosik:
Thanks for the PR. I've been a way for a while. I'll try to take a look at this PR soon.

elmiko (commented):
this generally makes sense to me, i do have a question though. what would happen if a user downgrades their kubernetes version with this option in use on their VPAs, is there any provision or plan for this scenario?

voelzmo:
@elmiko 
> what would happen if a user downgrades their kubernetes version with this option in use on their VPAs

I'm not sure how this is related to kubernetes versions. Are you asking what would happen when users downgrade their VPA? I don't think there is a good way to make this work, as any fallback logic would require a new VPA release as well, right?

elmiko:
> Are you asking what would happen when users downgrade their VPA?

@voelzmo yeah, that's more what i was asking about. apologies for the confusion.

> I don't think there is a good way to make this work, as any fallback logic would require a new VPA release as well, right?

that makes sense to me. i am trying to understand the scope of things that a user might have to deal with.

vertical-pod-autoscaler/enhancements/4831-upscale-only/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+
+### Goals
+* For a specific VPA object, allow to turn off active downscaling entirely while still allow for upscaling
+* Allow for changes of `UpdateMode` during the lifetime of a VPA object
+### Non-goals
+* Have VPA take care of dynamically changing the `UpdateMode`: Reasons and already existing mechanism achieving a similar functionality will vary widely, this is not what VPA should be concerned with
+
+## Proposal
+Add a new `UpdateMode` called `UpscaleOnly`, which works similar to `Auto` when scaling up, allowing the VPA to evict Pods and increase their resource requests, but prevents the VPA from evicting Pods when the new recommendation is smaller than the current one. Similarly to how the `Initial` mode works, this means that a smaller recommendation can still be applied to a Pod if it is recreated due to other reasons.
```
jbartosik, line 28: I see one thing that could be confusing here:
- Mode is called "UpscaleOnly"
- VPA will still scale down if something else evicts a pod

I'll try to think of a different name to better describe the behavior (I don't think we can ensure we won't scale down - we don't know where the new pod is coming from (and if it's an old recreated pod what was its size before))

What I think we're doing here is "initial  but evict for scale ups".  If you have an idea for a better name (or argument for why this one is good) please share it.

vertical-pod-autoscaler/enhancements/4831-upscale-only/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Since the change is backward-compatible the suggestion is to extend `v1` version of VPA API, avoiding the hassle of introducing a new API version.
+
+## Design Details
+### Test Plan
+Add automated E2E tests with `spec.updatePolicy.updateMode: "UpscaleOnly"` verify that the VPA Updater does evict for higher recommendations, but not for lower ones.
+
+## Alternatives
+### Run VPA in recommendation only mode
+We currently achieve a similar functionality by running VPA in `spec.updatePolicy.updateMode: "Off"` and having a different controller inspect the recommendations and apply them based on certain criteria like scaling direction and time.
```
jbartosik, line 38: Why not use Initial mode?

I'm asking because:
- With Initial you'd have to reimplement less of VPA
- But behavior would be different than my understanding of "UpscaleOnly" mode described here (if something else evicts pod you'd apply the lower recommendation)

I want to make sure I understand your use case well enough (and that the proposal addresses it).

vertical-pod-autoscaler/enhancements/4831-upscale-only/README.md
jbartosik, line 28: I want to make sure: we're comparing recommendation (target?) to current pod request? And evicting only if target > current request ?

jbartosik, line 28: What if:
- For CPU recommendation target > current pod request but
- For memory recommendation target < current pod request?

Do we evict the pod or not? (does it matter if it's the other way around)?

jbartosik (commented):
Thanks and sorry for taking so long. I have a few questions

vertical-pod-autoscaler/enhancements/4831-upscale-only/README.md
voelzmo: Oh, that's a great question I indeed didn't cover! The main use-case for this mode is to avoid disruptions that aren't necessary to help the workload perform its task, so I'd propose as long as one resource has `target > current request` we would evict.

voelzmo: Correct! I tried to change the text to make this more clear.

jbartosik (changes requested):
I think using `updateMode` for the change you want not expressive enough for what you're trying to do here.

I propose we:
* add `EvictionRequirements` field to the [`PodUpdatePolicy`](https://github.com/kubernetes/autoscaler/blob/2542e8c884a8e25634d3b8f43243fa8706007f30/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L96). It will have type `[]EvictionRequirement`
* eviction will happen only if all the current requirements are met and all requirements specified in the `EvictionRequirements` field met (so by default, when field is empty VPA will work as it does now)
* `EvictionRequirement` will specify which `Resource` it applies to (`Any` / `Cpu` / `Memory`) and how it must change to meet the requirement (`ChangeRequirement`) (`TargetLowerThanRequest`, `TargetLowerThanOrEqualToRequest`, `TargetHigherThanOrEqualToRequest`, `TargetHigherThanRequest`).

So to specify the behavior you described one would write something like:

```
PodUpdatePolicy: {
  EvictionRequirements: {
    EvictionRequirement {
      Resource: "Any",
      ChangeRequirement: "TargetHigherThanRequest",
    },
  },
}
```

Or you could ensure that memory is not being reduced and at least one resource is being increased:

```
PodUpdatePolicy: {
  EvictionRequirements: {
    EvictionRequirement {
      Resource: "Any",
      ChangeRequirement: "TargetHigherThanRequest",
    },
  EvictionRequirement {
      Resource: "Memory",
      ChangeRequirement: "TargetHigherThanOrEqualToRequest",
    },
  },
}
```

---

We could make it even more general and have something like


```
PodUpdatePolicy: {
  EvictionRequirements: {
    EvictionRequirement {
      Resource: "Memory",
      Left: "RecommendationTarget",
      Operator: ">=",
      Right: "CurrentRequest",
    },
  },
}
```
To ensure that we're not evicting to lower memory but I'm worried this would make things harder to understand.

jbartosik (commented):
One more idea: we could have a separate config in [ContainerResourcePolicy](https://github.com/kubernetes/autoscaler/blob/2542e8c884a8e25634d3b8f43243fa8706007f30/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L140) to ensure that Admission Controller isn't lowering pods request.

Please let me know if you're interested in that.

voelzmo:
Thanks for your thoughts on this, @jbartosik!

My understanding is that you're talking about two different alternatives to my proposal here, so let me try to address them one-by-one.

> I propose we: add `EvictionRequirements` field to the [`PodUpdatePolicy`](https://github.com/kubernetes/autoscaler/blob/2542e8c884a8e25634d3b8f43243fa8706007f30/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L96)(...)

This seems to solve a more sophisticated problem than initially proposed. This proposal was meant to reduce the negative effects of evictions for certain workloads which are very sensitive to that. If a workload needs more resources (regardless which kind), it should still get evicted such that it can continue do its work – but if we're trying to save cost (i.e. `recommendation < requests`), we don't evict. If `memory` and `CPU` have different trends (let's say `recommendation < requests` for memory, but `recommendation > requests` for CPU), then we would still want to evict to get the higher requests for CPU, so the workload gets the processing power it needs.

All of this probably wouldn't be necessary, if [In-place Update of Pod Resources](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources) would be a thing – but I guess that will still take a while ;)

As I mentioned above: We currently have a different controller in place that helps us achieve this goal using a different method and currently we don't have the need to distinguish between resources.

> we could have a separate config in [`ContainerResourcePolicy`](https://github.com/kubernetes/autoscaler/blob/2542e8c884a8e25634d3b8f43243fa8706007f30/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L140) to ensure that Admission Controller isn't lowering pods request.

Are you proposing to add a new `spec.resourcePolicy.containerPolicies[].mode: upscaleOnly` (or called whatever) alongside `Auto` and `Off` instead of adding a new `spec.updatePolicy.updateMode: upscaleOnly`? If I'm misunderstanding this, please disregard my comments below.

I'm not sure if that semantically fits into what `spec.resourcePolicy.containerPolicies[].mode` is doing (Autoscaler being enabled or not) and the granularity it works on (per Container, not per VPA).
Additionally, I'm not entirely sure what it would mean for Pods with multiple containers: The admission-webhook wouldn't evict the Pod if a single container has a `mode: upscaleOnly` and for that specific container we would see `recommendation < requests`? Or would it still evict, but not touch this Container's requests? In this case, this wouldn't serve the original goal: reduce evictions for Pods that don't need more resources to work properly.

jbartosik:
> My understanding is that you're talking about two different alternatives to my proposal here, so let me try to address them one-by-one.

Yes. This part (above a horizontal line) is about changes I want to make to API in this KEP.

> 
> > I propose we: add `EvictionRequirements` field to the [`PodUpdatePolicy`](https://github.com/kubernetes/autoscaler/blob/2542e8c884a8e25634d3b8f43243fa8706007f30/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L96)(...)
> 
> This seems to solve a more sophisticated problem than initially proposed. This proposal was meant to reduce the negative effects of evictions for certain workloads which are very sensitive to that. If a workload needs more resources (regardless which kind), it should still get evicted such that it can continue do its work – but if we're trying to save cost (i.e. `recommendation < requests`), we don't evict. If `memory` and `CPU` have different trends (let's say `recommendation < requests` for memory, but `recommendation > requests` for CPU), then we would still want to evict to get the higher requests for CPU, so the workload gets the processing power it needs.

The API is a little bigger (adding a new message instead of a new enum value) but I hope that implementation should take about the same amount of work. And it's more future proof. Let's say we add another enum value to encode "evict only if at least 1 resource will be increased, in addition to all the usual conditions, apply recommendations to containers when they start". Then it's reasonable to get another enum value for evicting only when memory is being increased. And another one for a resource would be decreased (when someone is setting high starting values)...

> 
> All of this probably wouldn't be necessary, if [In-place Update of Pod Resources](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources) would be a thing – but I guess that will still take a while ;)

Yes, I looking into that but I hear SIG node has a lot of work so it will take a while.

> As I mentioned above: We currently have a different controller in place that helps us achieve this goal using a different method and currently we don't have the need to distinguish between resources.

I want to support your use case. But I also need to make sure that our API remains in good shape and can support needs of all our users. Like I wrote above I'm worried that supporting your use case by adding a new mode will lead to explosion in number of modes we're supporting and make the API confusing.


---
This is about a different improvement which was inspired by this one. I thought I'll mention it in case you're interested in it. But if you're not then let's focus on the change you want to contribute (evicting pods only when VPA wants to increase their requests / part above the horizontal line in this comment).

> > we could have a separate config in [`ContainerResourcePolicy`](https://github.com/kubernetes/autoscaler/blob/2542e8c884a8e25634d3b8f43243fa8706007f30/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L140) to ensure that Admission Controller isn't lowering pods request.
> 
> Are you proposing to add a new `spec.resourcePolicy.containerPolicies[].mode: upscaleOnly` (or called whatever) alongside `Auto` and `Off` instead of adding a new `spec.updatePolicy.updateMode: upscaleOnly`? If I'm misunderstanding this, please disregard my comments below.
> 
> I'm not sure if that semantically fits into what `spec.resourcePolicy.containerPolicies[].mode` is doing (Autoscaler being enabled or not) and the granularity it works on (per Container, not per VPA). Additionally, I'm not entirely sure what it would mean for Pods with multiple containers: The admission-webhook wouldn't evict the Pod if a single container has a `mode: upscaleOnly` and for that specific container we would see `recommendation < requests`? Or would it still evict, but not touch this Container's requests? In this case, this wouldn't serve the original goal: reduce evictions for Pods that don't need more resources to work properly.

No. I meant that in addition to what you proposed in this KEP we could also make a change to `spec.ResourcePolicy.ContainerPolicies`, add a field that will tell VPA Admission Controller that it shouldn't set requests lower than they were when a pod came in. This is kind of independent from the change you described in this KEP (just inspired by it so I thought I'd mention it in case you're interested).

voelzmo:
> The API is a little bigger (adding a new message instead of a new enum value) but I hope that implementation should take about the same amount of work.

I was thinking a bit on how far we would want to take the validation bit of the implementation: Which kind of check would we build to avoid the conditions can never be true? We could
* restrict to a single condition per resource, to avoid having e.g. `recommendation > resources` and `recommendation < resources`
* ensure that `Any` cannot be specified when there is a condition in place for both resources

This would probably be enough? It still is additional logic we should (need?) to have in place that we otherwise wouldn't need.

> I want to support your use case. But I also need to make sure that our API remains in good shape and can support needs of all our users. Like I wrote above I'm worried that supporting your use case by adding a new mode will lead to explosion in number of modes we're supporting and make the API confusing.

I absolutely understand this concern and I'm not saying we shouldn't implement it this way. I can see how this API design makes sense if you're expecting additional scenarios in the future that you'd want to support here. Do you have thoughts on the use-cases for any of the other scenarios or even concrete input/feedback that this is something that people want to configure in a more fine-granular way?

>> Are you proposing to add a new spec.resourcePolicy.containerPolicies[].mode: upscaleOnly (or called whatever) alongside Auto and Off instead of adding a new spec.updatePolicy.updateMode: upscaleOnly? If I'm misunderstanding this, please disregard my comments below.
> (...)
>
> No. I meant that in addition to what you proposed in this KEP we could also make a change to spec.ResourcePolicy.ContainerPolicies, add a field that will tell VPA Admission Controller that it shouldn't set requests lower than they were when a pod came in. This is kind of independent from the change you described in this KEP (just inspired by it so I thought I'd mention it in case you're interested).

Thanks for helping me understand! From my point-of-view, we don't need a mechanism like this in the admission-controller that would prevent `recommendation < requests` to be updated on the Pods.

voelzmo:
@jbartosik Should I change this KEP to reflect our state of discussion, so we can get this in?

jbartosik:
> Joachim Should I change this KEP to reflect our state of discussion, so we can get this in?

Please do. I think this mostly ready.

voelzmo:
@jbartosik Does this update version work for you? Once this is merged I could start implementing it.

vertical-pod-autoscaler/enhancements/4831-upscale-only/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Add validation to prevent users from adding `EvictionRequirements` which can never evaluate to `true`:
+* Reject if more than one `EvictionRequirement` for `Resource: CPU` or `Resource: Memory` is found
+* Reject if `Resource: Any` is specified together with `Resource: CPU` or `Resource: Memory`
+
+## Design Details
+### Test Plan
+* Add automated E2E tests verifying that VPA only evicts when `ChangingRequirements` evaluate to `true`
+* Add tests verifying the validation
+## Alternatives
```
jbartosik, line 37: Nit: empty line before the header?

jbartosik (commented):
Looks good to me.

We still need someone to check the API.

I think it will be helpful to add code snippet with definition of `EvictionRequirement` to make it easier to check the API, something like (ChangeRequirementType needs a better name)

```go
type EvictionRequirement struct {
  Resource                    ResourceName
  ChangeRequirement ChangeRequirementType
}

type ResourceName string
type ChangeRequirementType string
const(
 ResourceNameCpu = ...
```

vertical-pod-autoscaler/enhancements/4831-upscale-only/README.md
voelzmo: fixed, thanks!

voelzmo:
> Looks good to me.
> 
> We still need someone to check the API.
> 
> I think it will need to add code snippet with definition API changes to make it feasible to check the API, something like (ChangeRequirementType needs a better name)
> 
> ```go
> type EvictionRequirement struct {
>   Resource                    ResourceName
>   ChangeRequirement ChangeRequirementType
> }
> 
> type ResourceName string
> type ChangeRequirementType string
> const(
>  ResourceNameCpu = ...
> ```

Does https://github.com/kubernetes/autoscaler/pull/5176/files help to review the API changes?

jbartosik:
> > Looks good to me.
> > We still need someone to check the API.
> > I think it will need to add code snippet with definition API changes to make it feasible to check the API, something like (ChangeRequirementType needs a better name)
> > ```go
> > type EvictionRequirement struct {
> >   Resource                    ResourceName
> >   ChangeRequirement ChangeRequirementType
> > }
> > 
> > type ResourceName string
> > type ChangeRequirementType string
> > const(
> >  ResourceNameCpu = ...
> > ```
> 
> Does https://github.com/kubernetes/autoscaler/pull/5176/files help to review the API changes?

Probably.

/lgtm

@mwielgus please take a look at the API changes

vertical-pod-autoscaler/enhancements/4831-control-eviction-behavior/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Goals
+* For a specific VPA object, allow to turn off eviction for downscaling entirely while still allow for upscaling
+* Allow for resource-specific decisions: The desired policy may be different for CPU and Memory
+
+## Proposal
+Add a new field `EvictionRequirements` to [`PodUpdatePolicy`](https://github.com/kubernetes/autoscaler/blob/2f4385b72e304216cf745893747da45ef314898f/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L109) of type `[]EvictionRequirement`. A single `EvictionRequirement` defines a condition which must be `true` to allow eviction for the corresponding `Pod`. When multiple `EvictionRequirements` are specified for a `Pod`, all of them must evaluate to `true` to allow eviction.
+
+A single `EvictionRequirement` refers to a `Resource` and defines a `ChangeRequirement` comparing the new recommendation (`Target`) with the existing requests on a Pod (`Requests`). Possible values for `Resource` are `CPU`, `Memory` or `Any`. Possible values for `ChangeRequirement` are `TargetHigherThanRequests`, `TargetHigherThanOrEqualToRequests`, `TargetLowerThanRequests` and `TargetLowerThanOrEqualToRequests`.
```
jbartosik, line 27: I noticed 's' at the end of TargetHigherThanRequests`, `TargetHigherThanOrEqualToRequests`, `TargetLowerThanRequests` and `TargetLowerThanOrEqualToRequests`.

I think the name works better without it (we're comparing one target to one reqest)

jbartosik (commented):
I noticed that API here is different than in #5176:

- This PR says in `ChangeRequirement` you pick on of  {`CPU`, `Memory` or `Any`}
- In the other PR you can pick 1 or 2 or {`CPU`, `Memory`} (and it does "or" if you pick more than 1)

Please reconcile those two PRs (especially if you plan to use #5176 to demonstrate what API will look like)

jbartosik:
I noticed that API here is different than in #5176:

- This PR says in `ChangeRequirement` you pick on of  {`CPU`, `Memory` or `Any`}
- In the other PR you can pick 1 or 2 or {`CPU`, `Memory`} (and it does "or" if you pick more than 1)

> I noticed that API here is different than in #5176:
> 
> * This PR says in `ChangeRequirement` you pick on of  {`CPU`, `Memory` or `Any`}
> * In the other PR you can pick 1 or 2 or {`CPU`, `Memory`} (and it does "or" if you pick more than 1)

vertical-pod-autoscaler/enhancements/4831-control-eviction-behavior/README.md
voelzmo: I think the reasoning behind that was that the field in the Pod's Spec is called `requests`? We can certainly rename this if you feel it should be changed. So, semantically this condition compares what is in `target` to what is in `requests`.

voelzmo:
> I noticed that API here is different than in #5176:
> 
> * This PR says in `ChangeRequirement` you pick on of  {`CPU`, `Memory` or `Any`}
> * In the other PR you can pick 1 or 2 or {`CPU`, `Memory`} (and it does "or" if you pick more than 1)
> 
> Please reconcile those two PRs (especially if you plan to use #5176 to demonstrate what API will look like)

done.

vertical-pod-autoscaler/enhancements/4831-control-eviction-behavior/README.md
voelzmo: @jbartosik Do you think I should rename to the singular `request` or should we keep `requests` because it refers to the `requests` in the Pod spec?

voelzmo:
@jbartosik Can we merge this or are there still things you want me to address?

jbartosik:
https://github.com/kubernetes/autoscaler/pull/4831#pullrequestreview-1099346882

> We still need someone to check the API. 

@mwielgus can you please take a look?

mwielgus (approved):
/lgtm
/approve

PR #4837, opened 2022-04-28
[CA][Helm] Allow skipping service creation

avorima:

helm charts

#### What type of PR is this?

/kind feature

The service is the only resource that can't be disabled. As I see it its only purpose is to expose metrics and some optional things like profiling and snapshots. If none of these are used a service shouldn't be required.

```release-note
Optionally exclude service from templated resources
```


gjtempleton:
Hey, thanks for the PR!

Can you please give it a minor rather than patch version bump as its adding a new piece of functionality?

PR #4839, opened 2022-04-29
Move the condition for ScaleDownInProgress to legacy scaledown code

x13n:

cluster-autoscaler

#### What type of PR is this?

/kind cleanup

Step towards preparing the code for another ScaleDown implementation.

Fixes #

```release-note
NONE
```

```docs
https://github.com/kubernetes/autoscaler/blob/48273b7affaa9463b37c72146a3b3de2ac06fc06/cluster-autoscaler/proposals/parallel_drain.md
```

/assign @towca

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -523,32 +523,28 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 			a.lastScaleUpTime.Add(a.ScaleDownDelayAfterAdd).After(currentTime) ||
 			a.lastScaleDownFailTime.Add(a.ScaleDownDelayAfterFailure).After(currentTime) ||
 			a.lastScaleDownDeleteTime.Add(a.ScaleDownDelayAfterDelete).After(currentTime)
-		// TODO(x13n): Move nonEmptyDeletionsCount > 0 condition to the legacy scaledown implementation.
-		_, nonEmptyDeletionsInProgress := actuationStatus.DeletionsInProgress()
-		nonEmptyDeletionsCount := len(nonEmptyDeletionsInProgress)
-		// In dry run only utilization is updated
-		calculateUnneededOnly := scaleDownInCooldown || nonEmptyDeletionsCount > 0
 
-		klog.V(4).Infof("Scale down status: unneededOnly=%v lastScaleUpTime=%s "+
-			"lastScaleDownDeleteTime=%v lastScaleDownFailTime=%s scaleDownForbidden=%v "+
-			"nonEmptyDeletionsCount=%v scaleDownInCooldown=%v",
-			calculateUnneededOnly, a.lastScaleUpTime,
-			a.lastScaleDownDeleteTime, a.lastScaleDownFailTime, a.processorCallbacks.disableScaleDownForLoop,
-			nonEmptyDeletionsCount, scaleDownInCooldown)
+		klog.V(4).Infof("Scale down status: lastScaleUpTime=%s lastScaleDownDeleteTime=%v "+
+			"lastScaleDownFailTime=%s scaleDownForbidden=%v scaleDownInCooldown=%v",
+			a.lastScaleUpTime, a.lastScaleDownDeleteTime, a.lastScaleDownFailTime,
+			a.processorCallbacks.disableScaleDownForLoop, scaleDownInCooldown)
 		metrics.UpdateScaleDownInCooldown(scaleDownInCooldown)
 
 		if scaleDownInCooldown {
 			scaleDownStatus.Result = status.ScaleDownInCooldown
-		} else if nonEmptyDeletionsCount > 0 {
-			scaleDownStatus.Result = status.ScaleDownInProgress
 		} else {
 			klog.V(4).Infof("Starting scale down")
 
 			// We want to delete unneeded Node Groups only if there was no recent scale up,
 			// and there is no current delete in progress and there was no recent errors.
-			removedNodeGroups, err := a.processors.NodeGroupManager.RemoveUnneededNodeGroups(autoscalingContext)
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/core/scaledown/legacy/wrapper.go
```diff
@@ -47,7 +51,15 @@ func NewScaleDownWrapper(sd *ScaleDown) *ScaleDownWrapper {
 func (p *ScaleDownWrapper) UpdateClusterState(podDestinations, scaleDownCandidates []*apiv1.Node, actuationStatus scaledown.ActuationStatus, pdbs []*policyv1.PodDisruptionBudget, currentTime time.Time) errors.AutoscalerError {
 	p.sd.CleanUp(currentTime)
 	p.pdbs = pdbs
-	return p.sd.UpdateUnneededNodes(podDestinations, scaleDownCandidates, currentTime, pdbs)
+	err := p.sd.UpdateUnneededNodes(podDestinations, scaleDownCandidates, currentTime, pdbs)
+	if err != nil {
+		return err
+	}
+	_, drained := actuationStatus.DeletionsInProgress()
+	if len(drained) > 0 {
```
MaciekPytel, line 59: Hmm, wouldn't it be easier to do this check in StartDeletion instead? This way you wouldn't need to check error type in static_autoscaler. With that solution you could also move `scaleDownStatus.Result = status.ScaleDownInProgress` into the wrapper (IIRC we don't think we want to retain ScaleDownInProgress status in new implementation?)

x13n: There's also a second difference - putting this in Actuator would cause `processors.NodeGroupManager.RemoveUnneededNodeGroups` to be called when there's an ongoing drain, which would change the existing behavior of legacy scale down.

MaciekPytel: I hoped we may be able to make static_autoscaler not aware of legacy / new scale-down and just use the interface. I'd rather have a permanent actuator wrapper for legacy scale-down or something along those lines if that meant we can clean up static_autoscaler. Ideally we'd encapsulate all logic related to legacy inside the broader scale-down package.

If we need to spill over to static_autoscaler for other reasons than I guess I'm fine with leaving this as-is.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -523,32 +523,28 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
-			scaleDownStatus.Result = status.ScaleDownInProgress
 		} else {
 			klog.V(4).Infof("Starting scale down")
 
 			// We want to delete unneeded Node Groups only if there was no recent scale up,
 			// and there is no current delete in progress and there was no recent errors.
-			removedNodeGroups, err := a.processors.NodeGroupManager.RemoveUnneededNodeGroups(autoscalingContext)
-			if err != nil {
-				klog.Errorf("Error while removing unneeded node groups: %v", err)
+			e, d := actuationStatus.DeletionsInProgress()
```
towca, line 540: nit: I get striving for conciseness in naming, but isn't this a bit too "concise" :sweat_smile:?

cluster-autoscaler/core/scaledown/legacy/wrapper.go
```diff
@@ -81,6 +81,12 @@ func (p *ScaleDownWrapper) NodeUtilizationMap() map[string]utilization.Info {
 
 // StartDeletion triggers an actual scale down logic.
 func (p *ScaleDownWrapper) StartDeletion(empty, needDrain []*apiv1.Node, currentTime time.Time) (*status.ScaleDownStatus, errors.AutoscalerError) {
+	_, drained := p.sd.nodeDeletionTracker.DeletionsInProgress()
```
towca, line 84: Have you thought about putting this part inside `TryToScaleDown`? Maybe then after extracting the actuation logic from `TryToScaleDown`, `TryToScaleDown` could become `NodesToDelete`, and we could get rid of this wrapper?

cluster-autoscaler/core/static_autoscaler.go
x13n: As a rule of thumb, the broader the scope, the more descriptive the variable name should be. Since we're using `a` which is scoped to the entire function, I didn't see anything wrong with another one-letter variable scoped to an even narrower scope. But if you're saying it hurts readability, then sure, let's use a longer name :slightly_smiling_face:

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -523,32 +523,28 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
+			var removedNodeGroups []cloudprovider.NodeGroup
+			if len(e)+len(d) == 0 {
+				var err error
+				removedNodeGroups, err = a.processors.NodeGroupManager.RemoveUnneededNodeGroups(autoscalingContext)
```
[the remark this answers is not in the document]
x13n: I know this changes the behavior, but that looks more like a bug than a feature to me: if we're still deleting nodes, maybe we shouldn't delete the node groups yet. Checking both empty and drained nodes is also in line with the comment above. On the other hand, changing the logic in this PR isn't probably the best idea, so updated to match the existing behavior.

cluster-autoscaler/core/scaledown/legacy/wrapper.go
x13n: Moving to `TryToScaleDown` sounds good (done), but I'm not sure we can easily make it into `NodesToDelete` and remove the wrapper, since `TryToScaleDown` has a lot of logic for setting the scale down status which won't make sense in `NodesToDelete`. We'll still need the wrapper (although not necessarily this whole file) around the common actuation part.

PR #4842, opened 2022-04-30
cluster-autoscaler: Add --max-pod-eviction-time flag to configure MaxPodEvictionTime

### LINKED ISSUE #4841: Allow user to configure maxPodEvictionTime through a flag on the cluster autoscaler
Cluster Autoscaler

There are cases when the eviction of a pod may take more than the default / hard-coded duration of 2 minutes.

We could expose the `MaxPodEvictionTime` as a flag, with default value `2 minutes`, that the user can configure if needed.

None.

**Additional context.**:
`MaxPodEvictionTime` is the maximum time CA tries to evict a pod before giving up.

gregth:
Signed-off-by: Grigoris Thanasoulas <gregth@arrikto.com>

cluster-autoscaler

#### What type of PR is this?
/kind feature

This PR introduces a  `--max-pod-eviction-time` flag that allows the user to configure the `MaxPodEvictionTime`, which is currently hard coded and set to `2m`. There are cases when a pod may take (by design) more than 2 minutes to be evicted, and the user may wish to configure the autoscaler to wait for more time for the eviction to complete.

Fixes #4841.

```release-note
Introduce a `--max-pod-eviction-time` flag to allow configuration of `MaxPodEvictionTime`, i.e., the maximum time the cluster autoscaler tries to evict a pod.
```

```docs

```

### DIFF
--- a/cluster-autoscaler/config/autoscaling_options.go
+++ b/cluster-autoscaler/config/autoscaling_options.go
@@ -147,6 +147,8 @@ type AutoscalingOptions struct {
 	MaxBulkSoftTaintCount int
 	// MaxBulkSoftTaintTime sets the maximum duration of single run of PreferNoSchedule tainting.
 	MaxBulkSoftTaintTime time.Duration
+	// MaxPodEvictionTime sets the maximum time CA tries to evict a pod before giving up.
+	MaxPodEvictionTime time.Duration
 	// IgnoredTaints is a list of taints to ignore when considering a node template for scheduling.
 	IgnoredTaints []string
 	// BalancingExtraIgnoredLabels is a list of labels to additionally ignore when comparing if two node groups are similar.
--- a/cluster-autoscaler/core/scaledown/legacy/legacy.go
+++ b/cluster-autoscaler/core/scaledown/legacy/legacy.go
@@ -68,8 +68,6 @@ const (
 	MaxKubernetesEmptyNodeDeletionTime = 3 * time.Minute
 	// MaxCloudProviderNodeDeletionTime is the maximum time needed by cloud provider to delete a node.
 	MaxCloudProviderNodeDeletionTime = 5 * time.Minute
-	// MaxPodEvictionTime is the maximum time CA tries to evict a pod before giving up.
-	MaxPodEvictionTime = 2 * time.Minute
 	// EvictionRetryTime is the time after CA retries failed pod eviction.
 	EvictionRetryTime = 10 * time.Second
 	// PodEvictionHeadroom is the extra time we wait to catch situations when the pod is ignoring SIGTERM and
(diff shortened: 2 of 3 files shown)


mwielgus (approved):
/lgtm
/approve

psharik1:
How would this work with respect to // MaxGracefulTerminationSec which is still hard coded to 10 minutes. Does this mean max-pod-eviction-time can be maximum of 10 minutes?

gregth:
@psharik1 To my understanding, the `MaxGracefulTerminationSec` is not hard-coded anymore, see:
https://github.com/kubernetes/autoscaler/blob/b5e4edc8933a1e8f04dcbf5f62fb44a59a5cc5d5/cluster-autoscaler/main.go#L138

However, let's make clear what the difference between `MaxGracefulTerminationSec` and `maxPodEvictionTimeout` is, and how it should be used.

https://github.com/kubernetes/autoscaler/blob/b5e4edc8933a1e8f04dcbf5f62fb44a59a5cc5d5/cluster-autoscaler/core/scaledown/legacy/legacy.go#L1176-L1192

From this snippet of code, I get that:
- `maxPodEvictionTimeout` is used to calculate the time until (`retryUntil`) which the CA will retry to evict the pod.
- We define the max termination of a Pod to the minimum of the argument configured `maxTermination` = min(`maxGracefulTermination`, `podToEvict.Spec.TerminationGracePeriodSeconds`) 
- The `maxTermination` is passed to the Eviction request, whereas the `maxPodEviction` defines how long the autoscaler tries to send an eviction request. 

https://github.com/kubernetes/autoscaler/blob/b5e4edc8933a1e8f04dcbf5f62fb44a59a5cc5d5/cluster-autoscaler/core/scaledown/legacy/legacy.go#L1139-L1168

Moreover, a numbered list of steps that take place when `evictPod()` is called:
1. Calculate  `maxTermination` = min(`maxGracefulTermination`, `podToEvict.Spec.TerminationGracePeriodSeconds`) 
2. Send a request to the API Server to for the pod to be evicted, with `GracePeriodSeconds` set to`maxTermination`. 
3. If the eviction is not permitted (for example a PDB is configured preventing the eviction to take place), then the request will return immediately and the  `GracePeriodSeconds` is irrelevant. The autoscaler will retry with a new eviction request, as many times as it can, till `maxPodEvictionTimeout` duration elapses. 
4. Otherwise, i.e., if the eviction is permitted, the  `GracePeriodSeconds` will be considered, so that the API Server deletes the pod after `GracePeriodSeconds`.

The summary of my understanding:
- The `maxPodEvictionTimeout` and `MaxGracefulTerminationSec` are independent. 
- `maxPodEvictionTimeout` defines how long the cluster autoscaler will retry with a new eviction request, if the eviction request fails.
- `MaxGracefulTerminationSec` is passed to the eviction requests, and defines how long the API server will wait for a pod **which is allowed to be evicted** to get gracefully terminated.

Please correct me if my understanding is distorted.

psharik1:
Thanks for response @gregth ,

Yes i concur that these 2 attributes are indeed independent.

Where for first part of your response `To my understanding, the MaxGracefulTerminationSec is not hard-coded anymore, see:`
"--max-graceful-termination-sec" is configurable to max 10 minutes. (I should not have said hard coded, but meant it can not be extended beyond that). 

Refer below - 

> Does CA respect GracefulTermination in scale-down?

_CA, from version 1.0, gives pods at most 10 minutes graceful termination time by default (configurable via --max-graceful-termination-sec). If the pod is not stopped within these 10 min then the node is deleted anyway. Earlier versions of CA gave 1 minute or didn't respect graceful termination at all._

Let me know if i read it incorrectly. I have tried it setting to 30 mins, but as stated above my node got evicted in 10 minutes.

gregth:
> "--max-graceful-termination-sec" is configurable to max 10 minutes.
This does not hold. Maybe the wording of the text is not quite appropriate, but the graceful termination can be more than 10 minutes. The 10 minutes that the example refers to is just the default value, you can have a larger value X > 10 minutes.

Note however that most pods have their own termination grace period on their spec. The default value of the field is 30 seconds, unless the pod's template specifies otherwise. 

From https://github.com/kubernetes/kubernetes/blob/313d9fb0b12ef627644bf2a7153bd1fd8dc399fd/staging/src/k8s.io/api/core/v1/types_swagger_doc_generated.go#L1643 : 
> 	"terminationGracePeriodSeconds": "Optional duration in seconds the pod needs to terminate gracefully. May be decreased in delete request. Value must be non-negative integer. The value zero indicates stop immediately via the kill signal (no opportunity to shut down). If this value is nil, the default grace period will be used instead. The grace period is the duration in seconds after the processes running in the pod are sent a termination signal and the time when the processes are forcibly halted with a kill signal. Set this value longer than the expected cleanup time for your process. Defaults to 30 seconds."

As exposed in the previous comment, the `maxTermination` time sent in the eviction request is `min(maxGracefulTermination, podToEvict.Spec.TerminationGracePeriodSeconds)`. If the pods specify the default 30s value, then `maxTermination` will be 30 seconds.

PR #4860, opened 2022-05-06
OCI cloud provider : Optional Feature - Skip time-consuming findInstanceByDetails API call by adding non_pool_member annotation to the node

aarontams:
This feature can be turned on by setting environment variable _**OCI_USE_NON_MEMBER_ANNOTATION**_ to true.
    
For example, if a node is not belong to any instance pool,  during the very first GetInstancePoolForInstance call, the node will be annotated with _**oci.oraclecloud.com/instancepool-id: non_pool_member**_.
After new annotation is added to the node, all future GetInstancePoolForInstance calls related to the non-pool-member node will skip the time-consuming findInstanceByDetails API call.
    
    Minor refactoring - when retrieving pool size from instancePoolCache object, use the pool size defined inside the poolCache map instead.


jlamillan:
Thanks for the pull request. Your changes look good to me.

jlamillan (approved):
Looks good to me.

mwielgus (approved):
/lgtm
/approve

PR #4871, opened 2022-05-09
Update vendor to Kubernetes 1.24.0

MaciekPytel:

cluster-autoscaler

#### What type of PR is this?

Update vendor for CA 1.24 release

### DIFF
--- a/cluster-autoscaler/simulator/scheduler_based_predicates_checker_test.go
+++ b/cluster-autoscaler/simulator/scheduler_based_predicates_checker_test.go
@@ -151,6 +151,6 @@ func TestDebugInfo(t *testing.T) {
 
 	predicateErr := predicateChecker.CheckPredicates(clusterSnapshot, p1, "n1")
 	assert.NotNil(t, predicateErr)
-	assert.Equal(t, "node(s) had taint {SomeTaint: WhyNot?}, that the pod didn't tolerate", predicateErr.Message())
+	assert.Equal(t, "node(s) had untolerated taint {SomeTaint: WhyNot?}", predicateErr.Message())
 	assert.Contains(t, predicateErr.VerboseMessage(), "RandomTaint")
 }


cluster-autoscaler/vendor/sigs.k8s.io/cloud-provider-azure/pkg/provider/doc.go
```diff
@@ -16,4 +16,4 @@ limitations under the License.
 
 // Package provider is an implementation of CloudProvider Interface, LoadBalancer
 // and Instances for Azure.
-package provider // import "sigs.k8s.io/cloud-provider-azure/pkg/provider"
+package provider // import "sigs.k8s.io/cloud-provider-azure/pkg/provider
```
gjtempleton, line 19: Seems the test script is choking on this change.

Looks like it was fixed in v1.24.0: https://github.com/kubernetes-sigs/cloud-provider-azure/pull/1158

MaciekPytel: I opened https://github.com/kubernetes/autoscaler/issues/4875 about this. The obvious solution is to carry the suspected fix in vendor, but I'd rather not do that if we can avoid it. Among other problems it may break anyone importing CA code (transitive dependency would be taken from original repo and not the fixed version in CA vendor in that case).

Let's see if Azure maintainers can help us with this one.

MaciekPytel:
/hold 
To allow review without accidentally merging while the discussion is ongoing on https://github.com/kubernetes/autoscaler/issues/4875

cluster-autoscaler/simulator/scheduler_based_predicates_checker_test.go
```diff
@@ -151,6 +151,6 @@ func TestDebugInfo(t *testing.T) {
 
 	predicateErr := predicateChecker.CheckPredicates(clusterSnapshot, p1, "n1")
 	assert.NotNil(t, predicateErr)
-	assert.Equal(t, "node(s) had taint {SomeTaint: WhyNot?}, that the pod didn't tolerate", predicateErr.Message())
+	assert.Equal(t, "node(s) had untolerated taint {SomeTaint: WhyNot?}", predicateErr.Message())
```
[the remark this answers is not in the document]
MaciekPytel: Correct, the test relies on the wording of upstream error message. The message has changed upstream and so I had to update the test.

MaciekPytel:
/hold cancel
Got agreement on the approach in https://github.com/kubernetes/autoscaler/issues/4875#issuecomment-1123927392, no reason to hold any longer.

PR #4873, opened 2022-05-09
Fix fallback for aws cloudprovider

tzneal:

cluster-autoscaler

#### What type of PR is this?

/kind bug

If instance type listing fails, the autoscaler would fail to work.  This change modifies behavior so that it falls back
to the static list of instance types.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider_test.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package aws
 
 import (
+	"os"
 	"testing"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -25,6 +26,7 @@ import (
 	"github.com/stretchr/testify/mock"
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
+	"k8s.io/autoscaler/cluster-autoscaler/config"
 )
 
 var testAwsManager = &AwsManager{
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/cloudprovider/aws/aws_cloud_provider_test.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package aws
 
 import (
+	"k8s.io/autoscaler/cluster-autoscaler/config"
```
jaypipes, line 20: nit: move this import to the block below for 3rd party/local imports...

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider_test.go
```diff
@@ -109,6 +111,23 @@ func TestBuildAwsCloudProvider(t *testing.T) {
...
+	do := cloudprovider.NodeGroupDiscoveryOptions{}
+	opts := config.AutoscalingOptions{}
+
+	os.Setenv("AWS_REGION", "us-west-2")
+	defer os.Unsetenv("AWS_REGION")
+
+	// This test ensures that no klog.Fatalf calls occur when constructing the AWS cloud provider.  Specifically it is
+	// intended to ensure that instance type fallback works correctly in the event of an error enumerating instance
+	// types.
+	_ = BuildAWS(opts, do, resourceLimiter)
```
jaypipes, line 128: AFAICT, this test is only testing the happy path, though, right? If we passed in a non-existent AWS region, like "us-south-42", for instance, would that trigger an error being returned from `GenerateEC2InstanceTypes` and therefore trigger the fallback mechanism?

jaypipes (commented):
Thanks @tzneal! Couple little comments inline but overall, looks good to me.

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider_test.go
tzneal: Good catch, I tested with no credentials set locally, I assumed there were no credentials in the test environment.  I'll fix it.

gjtempleton:
Thanks for this @tzneal !
/lgtm

PR #4898, opened 2022-05-19
Adds experimental feature to delete pods

### LINKED ISSUE #4730: [VPA] Feature Request: Delete OOM Pods

VPA

To apply new resource recommendations the VPA uses the eviction API to evict Pods. This works fine for most cases. The problem we are facing is that when we have multiple replicas and a PodDisruptionBudget (PDB) which allows one disruption, the chance of a successfull eviction shrinks drastically. It might also get to a point where all replicas are CrashLooping and the VPA is unable to evict Pods as this would violate the PDB.

When an OOM Event was detected instead of evicting the Pod, delete it. Since the Pod crashed (got killed), there is no reason to respect a potential PDB. This should of course only be done for the actual CrashLooping Pod, as others might still survive.

n/a

**Additional context.**:

We just wanted to get some feedback from you guys, we would also be happy to implement this, if you think this is a good addition to the VPA.

RuriRyan:

vertical-pod-autoscaler

#### What type of PR is this?

/kind feature

This PR adds a new experimental feature, which can be enabled via
a command line flag which attempts a pod deletion should the eviction
fail for whatever reason.
The deletion is bound to various conditions to not blindly delete a pod
where the eviction failure might be justified.
In general the deletion only occurs if at least one container was
OOMKilled and is currently in CrashLoopBackoff and the number of
restarts exceeeds the configured threshold (also behind a flag).

Fixes #4730

```release-note
New experimental feature to delete pods after eviction fails
```

### DIFF
--- a/vertical-pod-autoscaler/pkg/updater/.gitignore
+++ b/vertical-pod-autoscaler/pkg/updater/.gitignore
@@ -5,3 +5,5 @@ updater-arm64
 updater-arm
 updater-ppc64le
 updater-s390x
+
+.envrc
--- a/vertical-pod-autoscaler/pkg/updater/README.md
+++ b/vertical-pod-autoscaler/pkg/updater/README.md
@@ -26,6 +26,9 @@ Threshold for evicting pods is specified by recommended min/max values from VPA 
 Priority of evictions within a set of replicated pods is proportional to sum of percentages of changes in resources
 (i.e. pod with 15% memory increase 15% cpu decrease recommended will be evicted
 before pod with 20% memory increase and no change in cpu).
+* Deleting pods if the eviction fails and if the corresponding feature flag (`--experimental-deletion`) is enabled.
+Deletion is guarded by a treshold (`--experimental-deletion-threshold`) of how many restarts are required. The pod
+needs to be in `CrashLoopBackOff` and the LastTerminationReason needs to be `OOMKilled`.
 
 # Missing parts
 * Recommendation API for fetching data from Vertical Pod Autoscaler Recommender.
--- a/vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go
+++ b/vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go
@@ -45,6 +45,9 @@ type PodsEvictionRestriction interface {
 	// Evict sends eviction instruction to the api client.
 	// Returns error if pod cannot be evicted or if client returned error.
 	Evict(pod *apiv1.Pod, eventRecorder record.EventRecorder) error
+	// EvictViaDelete sends deletion instruction to the api client.
+	// Returns error if pod cannot be deleted or if client returned error.
+	EvictViaDelete(pod *apiv1.Pod, eventRecorder record.EventRecorder, deletionThreshold int32) error
 	// CanEvict checks if pod can be safely evicted
 	CanEvict(pod *apiv1.Pod) bool
 }
@@ -143,6 +146,63 @@ func (e *podsEvictionRestrictionImpl) Evict(podToEvict *apiv1.Pod, eventRecorder
 	}
 	eventRecorder.Event(podToEvict, apiv1.EventTypeNormal, "EvictedByVPA",
 		"Pod was evicted by VPA Updater to apply resource recommendation.")
+
+	if podToEvict.Status.Phase != apiv1.PodPending {
+		singleGroupStats, present := e.creatorToSingleGroupStatsMap[cr]
+		if !present {
+			return fmt.Errorf("Internal error - cannot find stats for replication group %v", cr)
+		}
+		singleGroupStats.evicted = singleGroupStats.evicted + 1
+		e.creatorToSingleGroupStatsMap[cr] = singleGroupStats
+	}
+
+	return nil
+}
+
+// canDelete checks if the container has been restarted more than the threshold, and if it has been terminated
+// due to OOMKilled and is in CrashLoopBackOff state.
+func (e *podsEvictionRestrictionImpl) canDelete(containerStatus apiv1.ContainerStatus, deletionThreshold int32) bool {
+	return containerStatus.RestartCount > deletionThreshold &&
+		containerStatus.LastTerminationState.Terminated != nil &&
+		containerStatus.LastTerminationState.Terminated.Reason == "OOMKilled" &&
+		containerStatus.State.Waiting != nil &&
+		containerStatus.State.Waiting.Reason == "CrashLoopBackOff"
+}
+
+// EvictViaDelete sends deletion instruction to api client. Returns error if pod cannot be deleted or if client returned error
+// Does not check if pod was actually deleted after termination grace period.
+func (e *podsEvictionRestrictionImpl) EvictViaDelete(podToEvict *apiv1.Pod, eventRecorder record.EventRecorder, deletionThreshold int32) error {
+	cr, present := e.podToReplicaCreatorMap[getPodID(podToEvict)]
+	if !present {
+		return fmt.Errorf("pod not suitable for eviction %v : not in replicated pods map", podToEvict.Name)
+	}
+
+	if !e.CanEvict(podToEvict) {
+		return fmt.Errorf("cannot evict pod %v : eviction budget exceeded", podToEvict.Name)
+	}
+
+	deleted := false
+	// Try to delete the Pod if it's CrashLooping because of OOM.
+	// The Pod is dead anyway so there is no reason to respect any potential PDBs.
+	for _, containerStatus := range podToEvict.Status.ContainerStatuses {
+		if e.canDelete(containerStatus, deletionThreshold) {
+
+			err := e.client.CoreV1().Pods(podToEvict.Namespace).Delete(context.TODO(), podToEvict.Name, metav1.DeleteOptions{})
+			if err != nil {
+				klog.Errorf("failed to delete pod %s/%s, after failed eviction, error: %v", podToEvict.Namespace, podToEvict.Name, err)
+				return err
+			}
+
+			deleted = true
+			eventRecorder.Event(podToEvict, apiv1.EventTypeNormal, "DeletedByVPA",
+				"Pod was deleted by VPA Updater to apply resource recommendation.")
+			break
+		}
+	}
+
+	if !deleted {
+		return fmt.Errorf("cannot delete pod %v : no container matches deletion conditions", podToEvict.Name)
+	}
 
 	if podToEvict.Status.Phase != apiv1.PodPending {
 		singleGroupStats, present := e.creatorToSingleGroupStatsMap[cr]
--- a/vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction_test.go
+++ b/vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction_test.go
@@ -521,6 +521,105 @@ func TestEvictAtLeastOne(t *testing.T) {
 	}
 }
 
+func TestEvictViaDeleteAtLeastOne(t *testing.T) {
+	replicas := int32(5)
+	livePods := 5
+	tolerance := 0.1
+
+	rc := apiv1.ReplicationController{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "rc",
+			Namespace: "default",
+		},
+		TypeMeta: metav1.TypeMeta{
+			Kind: "ReplicationController",
+		},
+		Spec: apiv1.ReplicationControllerSpec{
+			Replicas: &replicas,
+		},
+	}
+
+	cs := &apiv1.ContainerStatus{
+		RestartCount: 5,
+		LastTerminationState: apiv1.ContainerState{
+			Terminated: &apiv1.ContainerStateTerminated{
+				Reason: "OOMKilled",
+			},
+		},
+		State: apiv1.ContainerState{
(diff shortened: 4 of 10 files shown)


jbartosik:
Thanks. I think it would be good to have an enhancement proposal for this.

In the enhancement proposal I'd like to see:
- Short explanation why we want this
- Explanation why do this in VPA instead of in PDB
- Description of the implementation

jbartosik:
LGMT
I think I need someone who knows more about API to take a look too

jbartosik:
I meant to leave those comments on the PR with KEP

vertical-pod-autoscaler/pkg/updater/.gitignore
```diff
@@ -5,3 +5,5 @@ updater-arm64
 updater-arm
 updater-ppc64le
 updater-s390x
+
```
jbartosik, line 8: Please drop the empty line

vertical-pod-autoscaler/pkg/updater/README.md
```diff
@@ -26,6 +26,9 @@ Threshold for evicting pods is specified by recommended min/max values from VPA
 Priority of evictions within a set of replicated pods is proportional to sum of percentages of changes in resources
 (i.e. pod with 15% memory increase 15% cpu decrease recommended will be evicted
 before pod with 20% memory increase and no change in cpu).
+* Deleting pods if the eviction fails and if the corresponding feature flag (`--experimental-deletion`) is enabled.
+Deletion is guarded by a treshold (`--experimental-deletion-threshold`) of how many restarts are required. The pod
+needs to be in `CrashLoopBackOff` and the LastTerminationReason needs to be `OOMKilled`.
```
jbartosik, line 31: `LastTerminationReason` ?

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go
```diff
@@ -156,6 +159,61 @@ func (e *podsEvictionRestrictionImpl) Evict(podToEvict *apiv1.Pod, eventRecorder
...
+			err := e.client.CoreV1().Pods(podToEvict.Namespace).Delete(context.TODO(), podToEvict.Name, metav1.DeleteOptions{})
+			if err != nil {
+				klog.Errorf("failed to delete pod %s/%s, after failed eviction, error: %v", podToEvict.Namespace, podToEvict.Name, err)
+				return err
+			}
+
+			deleted = true
+			eventRecorder.Event(podToEvict, apiv1.EventTypeNormal, "DeletedByVPA",
+				"Pod was deleted by VPA Updater to apply resource recommendation.")
+			break
```
jbartosik, line 197: If I read this loop right it will attempt to delete 0 or 1 containers.

- 0 containers if canDelete doesn't return true for any container,
- 1 container otherwise (returns error if delete fails, breaks if it succeeds)

Is this the intended behavior (as opposed to attempting to delete all containers until we run into an error)?

If so I'd make it into its own function that returns `(bool, error)` to make this easier to read.

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go
```diff
@@ -156,6 +159,61 @@ func (e *podsEvictionRestrictionImpl) Evict(podToEvict *apiv1.Pod, eventRecorder
...
+	return containerStatus.LastTerminationState.Terminated != nil &&
+		containerStatus.LastTerminationState.Terminated.Reason == "OOMKilled" &&
+		containerStatus.State.Waiting != nil &&
+		containerStatus.State.Waiting.Reason == "CrashLoopBackOff"
+}
+
+// EvictViaDelete sends deletion instruction to api client. Returns error if pod cannot be deleted or if client returned error
+// Does not check if pod was actually deleted after termination grace period.
+func (e *podsEvictionRestrictionImpl) EvictViaDelete(podToEvict *apiv1.Pod, eventRecorder record.EventRecorder) error {
+	cr, present := e.podToReplicaCreatorMap[getPodID(podToEvict)]
```
jbartosik, line 173: At first I was confused why we set `cr` here but use it much later. Please add a comment like

// Make sure that we can map pod to replica so any problems prevent deleting the pod.

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction_test.go
```diff
@@ -521,6 +521,105 @@ func TestEvictAtLeastOne(t *testing.T) {
...
+	factory, _ := getEvictionRestrictionFactory(&rc, nil, nil, 2, tolerance)
+	eviction := factory.NewPodsEvictionRestriction(pods, getBasicVpa())
+
+	for _, pod := range pods {
+		assert.True(t, eviction.CanEvict(pod))
+	}
+
+	for _, pod := range pods[:1] {
+		err := eviction.EvictViaDelete(pod, test.FakeEventRecorder())
+		assert.Nil(t, err, "Should evict with no error")
```
jbartosik, line 570: Please use assert.NoError

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction_test.go
```diff
@@ -521,6 +521,105 @@ func TestEvictAtLeastOne(t *testing.T) {
...
+	}
+	for _, pod := range pods[1:] {
+		err := eviction.EvictViaDelete(pod, test.FakeEventRecorder())
+		assert.Error(t, err, "Error expected")
```
jbartosik, line 574: Why do we expect error here?

vertical-pod-autoscaler/pkg/updater/logic/updater.go
```diff
@@ -216,6 +225,25 @@ func (u *updater) RunOnce(ctx context.Context) {
...
+				var DeleteOomingOnEvictionError bool
+				if vpa.Spec.UpdatePolicy != nil && vpa.Spec.UpdatePolicy.DeleteOomingOnEvictionError != nil {
+					DeleteOomingOnEvictionError = *vpa.Spec.UpdatePolicy.DeleteOomingOnEvictionError
+				} else {
+					DeleteOomingOnEvictionError = u.deleteOomingOnEvictionError
+				}
+
+				// only try to delete the pod if the feature is enabled and if we would increase
+				// the resource requests or limits
+				if priority.ScaleUp && DeleteOomingOnEvictionError {
```
jbartosik, line 238: I'd add a function that returns bool and rewrite the condition as

```
if attemptEvictViaDelete(&priority, &vpa.Spec) {
  //...
```

vertical-pod-autoscaler/pkg/updater/main.go
```diff
@@ -65,6 +65,8 @@ var (
 
 	namespace          = os.Getenv("NAMESPACE")
 	vpaObjectNamespace = flag.String("vpa-object-namespace", apiv1.NamespaceAll, "Namespace to search for VPA objects. Empty means all namespaces will be used.")
+
+	deleteOomingOnEvictionError = flag.Bool("delete-ooming-on-eviction-error", false, "If true, updater will try to delete ooming pods when the eviction fails.")
```
jbartosik, line 69: I think there should be "by default" somewhere in the flag description

vertical-pod-autoscaler/pkg/utils/metrics/updater/updater.go
```diff
@@ -59,6 +59,14 @@ var (
 		}, []string{"vpa_size_log2"},
 	)
 
+	deletedCount = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: metricsNamespace,
+			Name:      "deleted_pods_total",
+			Help:      "Number of Pods delete by Updater to apply a new recommendation.",
```
jbartosik, line 66: Number of Pods delete--d--

jbartosik (changes requested):
Thanks, I did first pass, mostly focusing on what was hard for me to read.

Did you test if this works as intended end - to - end?

jbartosik:
/hold

While I'm working on VPA 0.12.0 I don't want to make any changes to VPA that are not related to the release (I think code is good to go but E2E tests are timing out).

I'll remove the hold after I release VPA 0.12.0.

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction_test.go
RuriRyan: This is effectively just a copy & paste of the unit test for the "normal" eviction. I found out that only one eviction is allowed per "replica controller" for every "tick" of the updater. So when multiple pods of the same replica set should get evicted it throws an error until the next tick.

vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go
RuriRyan: It is the intended behaviour, because we cannot delete individual containers, but only the whole pod. So as soon as the pod is deleted, there is no reason to continue for the rest of the containers in the pod.

jbartosik:
@RuriRyan I see this has a few open comments, mostly small things.

RuriRyan:
> @RuriRyan I see this has a few open comments, mostly small things.

@avorima Can you take over? :D

Fyi: I left ionos and no longer have access to this Code.

avorima:
@jbartosik AFAICT all review comments have either been fixed or answered

timoreimann:
We'd also love to see this PR move forward and eventually get merged.

/cc @jbartosik

jbartosik:
@timoreimann looks good to me

There's one thing remaining - checking if this works end to end.

Did you run any (manual?) tests to verify this works as intended?

Can you provide an e2e test that will ensure this keeps working?

I thought that I could take [OOM](https://github.com/kubernetes/autoscaler/blob/1e40dda819249c240b623157f1701f063a893925/vertical-pod-autoscaler/e2e/v1/full_vpa.go#L288) test scenario and modify it:

1. Same scenario but with PDB preventing VPA from evicting any pods
2. Same scenario with PDB and VPA allowed to delete, verify that recommendation is applied

I'm not sure when I'll be able to do that. I'm worried about merging this without any verification that this works as intended end to end

jbartosik:
@timoreimann I want to merge this but  need some verification that this works e2e. I see the following options:

1. Someone writes e2e sends PR so that we can merge them at the same time
2. Someone verifies this works e2e manually, we merge then add e2e test later 

I can contribute e2e tests but I'm not sure when. I don't think 2. is significantly faster than 1.

timoreimann:
Writing an e2e tests sounds like the right thing to do. I'm happy to look into that when I have some spare time unless someone beats me to it.

voelzmo:
I just saw `unhealthyPodEvictionPolicy` was added for PDBs (see [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-apps/3017-pod-healthy-policy-for-pdb/README.md) and [API change in PDB](https://github.com/kubernetes/kubernetes/pull/113375)).

We also saw the original issue happen quite a few times and would want to get this fixed, but the change to PDB seems to solve the problem in a more "k8s native" way instead of work-arounding with the delete call.

@jbartosik @avorima @timoreimann wdyt?

timoreimann:
@voelzmo thanks for sharing the KEP. I wasn't aware of it, this does sound like the right approach going forward.

That said, I think there's value in driving this PR to completion regardless: the solution in the KEP will presumably not become available before Kubernetes 1.27, and it likely won't graduate to at least beta even later. Users on older versions will still want to have some kind of solution (at least we do given a large portion of our fleet is affected by the problem regularly).
It also seems possible to transition from what this PR proposes to what the KEP outlines by checking for the new field in the PDB once available and honor its value with higher precedence. At least, that's how I understand it should work.

Just my 2 cents of course, final call is to be made by @jbartosik. Coincidentally, I have started working on extending the end-to-end tests as proposed just earlier today, so I could probably add the last missing piece fairly soon.

timoreimann:
On second thought, the feature from this PR will likely only be backported so far (if at all), and / or more recent versions of VPA _may_ be incompatible with older Kubernetes versions? So maybe there is only so much value users of older Kubernetes releases could get out of this PR?

I'm not super familiar with VPA's / CA's release policy and compatibility guarantees, so throwing out additional thoughts mostly. @jbartosik still to the rescue for more clarity and preferences. 🙂

avorima:
Thanks for sharing @voelzmo, that looks promising.

This PR and the VPA KEP were just created because, at the time, it seemed like this was never going to make it into the PDB spec, so I would actually be fine with closing it now. The short time that it would be useful would probably be outweighed by the effort it takes to maintain it. Only speaking for myself of course.

jbartosik:
I'm happy to use PDB features instead of working around PDB in VPA. `AlwaysAllow` policy sounds like it does what this was intended to do (from https://github.com/kubernetes/kubernetes/pull/113375/):

> // AlwaysAllow policy means that all running pods (status.phase="Running"),
> // but not yet healthy are considered disrupted and can be evicted regardless
> // of whether the criteria in a PDB is met. This means perspective running
> // pods of a disrupted application might not get a chance to become healthy.
> // Healthy pods will be subject to the PDB for eviction.

avorima:
Ok, what about the KEP? I'm not familiar with the process, but I suppose it needs to be removed or updated to document this decision.

jbartosik:
> Ok, what about the KEP? I'm not familiar with the process, but I suppose it needs to be removed or updated to document this decision.

I added an item about this to the next SIG meeting

PR #4899, opened 2022-05-19
Offset Reserved memory and ephemeral storage for ARM arch in GCE

jayantjain93:

cluster-autoscaler
#### What type of PR is this?
/kind feature

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/gce_reserved.go
+++ b/cluster-autoscaler/cloudprovider/gce/gce_reserved.go
@@ -72,6 +72,8 @@ const (
 	lowMemoryOffset = 8 * MiB
 	// lowMemoryThreshold is the threshold to apply lowMemoryOffset
 	lowMemoryThreshold = 8 * GiB
+	// architectureOffsetRatio is the offset value used for arm architecture
+	architectureOffsetRatio = 0.9971
 )
 
 // EvictionHard is the struct used to keep parsed values for eviction
@@ -95,6 +97,9 @@ func (r *GceReserved) CalculateKernelReserved(physicalMemory int64, os Operating
 		if physicalMemory > swiotlbThresholdMemory {
 			reserved += swiotlbReservedMemory
 		}
+		if physicalMemory <= lowMemoryThreshold {
+			reserved += lowMemoryOffset
+		}
 
 		// Additional reserved memory to correct estimation
 		// The reason for this value is we detected additional reservation, but we were
@@ -102,12 +107,12 @@ func (r *GceReserved) CalculateKernelReserved(physicalMemory int64, os Operating
 		// statistically developed.
 		if osDistribution == OperatingSystemDistributionCOS {
 			reserved += int64(math.Min(correctionConstant*float64(physicalMemory), maximumCorrectionValue))
+			if arch == Arm64 {
+				reserved = int64(math.Round(float64(reserved) * architectureOffsetRatio))
+			}
 		} else if osDistribution == OperatingSystemDistributionUbuntu {
 			reserved += int64(math.Min(correctionConstant*float64(physicalMemory), maximumCorrectionValue))
 			reserved += ubuntuSpecificOffset
-		}
-		if physicalMemory <= lowMemoryThreshold {
-			reserved += lowMemoryOffset
 		}
 
 		return reserved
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/cloudprovider/gce/gce_reserved.go
```diff
@@ -95,20 +97,23 @@ func (r *GceReserved) CalculateKernelReserved(physicalMemory int64, os Operating
...
+			reserved += lowMemoryOffset
+		}
 
 		// Additional reserved memory to correct estimation
 		// The reason for this value is we detected additional reservation, but we were
 		// unable to find the root cause. Hence, we added a best estimated formula that was
 		// statistically developed.
 		if osDistribution == OperatingSystemDistributionCOS {
 			reserved += int64(math.Min(correctionConstant*float64(physicalMemory), maximumCorrectionValue))
+			if arch == Arm64 {
```
towca, line 110: Is ARM only compatible with COS?

cluster-autoscaler/cloudprovider/gce/gce_reserved_test.go
```diff
@@ -91,15 +92,31 @@ func TestCalculateKernelReservedLinux(t *testing.T) {
...
 	for idx, tc := range testCases {
 		r := &GceReserved{}
 		t.Run(fmt.Sprintf("%v", idx), func(t *testing.T) {
-			reserved := r.CalculateKernelReserved(tc.physicalMemory, OperatingSystemLinux, tc.osDistribution, "", "")
+			reserved := r.CalculateKernelReserved(tc.physicalMemory, OperatingSystemLinux, tc.osDistribution, tc.arch, "")
 			if tc.osDistribution == OperatingSystemDistributionUbuntu {
 				assert.Equal(t, tc.reservedMemory+int64(math.Min(correctionConstant*float64(tc.physicalMemory), maximumCorrectionValue)+ubuntuSpecificOffset), reserved)
 			} else if tc.osDistribution == OperatingSystemDistributionCOS {
-				assert.Equal(t, tc.reservedMemory+int64(math.Min(correctionConstant*float64(tc.physicalMemory), maximumCorrectionValue)), reserved)
+				val := tc.reservedMemory + int64(math.Min(correctionConstant*float64(tc.physicalMemory), maximumCorrectionValue))
```
towca, line 115: I'm not sure what this is testing exactly. It looks like we're just repeating the actual calculation in the test code? If there's a bug in the calculation itself, how would we catch it? And what do these tests even give us? Normally I'd expect that `tc.reservedMemory` would be the expected end value that `reserved` would be compared against (possibly with some tolerations).

cluster-autoscaler/cloudprovider/gce/gce_reserved.go
```diff
@@ -95,20 +97,23 @@ func (r *GceReserved) CalculateKernelReserved(physicalMemory int64, os Operating
 		if physicalMemory > swiotlbThresholdMemory {
 			reserved += swiotlbReservedMemory
 		}
+		if physicalMemory <= lowMemoryThreshold {
```
towca, line 100: What's the reasoning for moving this here? Have you verified that it won't mess up cases other than ARM? It seems we could be adding `lowMemoryThreshold` in more cases than before.

cluster-autoscaler/cloudprovider/gce/gce_reserved.go
jayantjain93: Yes

cluster-autoscaler/cloudprovider/gce/gce_reserved.go
jayantjain93: The new logic is taken as an offset of the "final" previous reserved values. To replicate that the reserved logic, we essentially made it the last code in COS. `lowMemoryThreshold` is moved earlier in code as the rest of the code after this point doesn't rely on `var reserved` (except new logic).

I didn't think this would impact addition cases by adding `lowMemoryThreshold` earlier, as the hierarchy of the if statement is the same as before and the var reserved isn't directly used in any existing code. Also, this impact should have directly affect the unit tests, which are not affected.

cluster-autoscaler/cloudprovider/gce/gce_reserved.go
jayantjain93: Yes

cluster-autoscaler/cloudprovider/gce/gce_reserved_test.go
jayantjain93: I agree. The test scenario for reserved tests don't provide a lot of signal.
However, there isn't much "logic" per-say in reserved.go, and it is essentially hardcoded values with some conditions. So extracting signals from this isn't particularly useful. We can only test if the values / conditions in reserved.go don't change or flag unintended changes.

IIUC is the idea of having a "static" value for `reservedMemory`? essentially we would just be compressing the manual conditions and live calculations (`52*MiB + kernelReservedMemory + swiotlbReservedMemory + lowMemoryOffset`) to pre-calculated in the value. Not sure if there is value in that change?

cluster-autoscaler/cloudprovider/gce/gce_reserved.go
towca: Oof, I've misread the if condition - I thought it was `reserved <= lowMemoryThreshold`, like a final bump if the `reserved` value was too low, in which case moving it would change things.

Please disregard :stuck_out_tongue:

cluster-autoscaler/cloudprovider/gce/gce_reserved_test.go
towca: Yeah I guess it doesn't change much, and the current solution is in line with the existing tests.

PR #4902, opened 2022-05-20
[VPA] KEP-4902: Delete OOM Pods

RuriRyan:

vertical-pod-autoscaler

#### What type of PR is this?

/kind documentation

KEP for #4730 and  #4898

### DIFF
--- a/vertical-pod-autoscaler/enhancements/4902-delete-oom-pods/README.md
+++ b/vertical-pod-autoscaler/enhancements/4902-delete-oom-pods/README.md
@@ -44,7 +44,7 @@ failing deployment, if updated/increased limits would solve the problem.
 
 The proposal is to add `--experimental-deletion` to the VPA to enable deletion
 of pods. Currently only as an experimental, or beta feature.
-To add a bit of configuration an additional an additional flag,
+To add a bit of configuration an additional flag,
 `--experimental-deletion-threshold`, should be addedd.
 
 ## Design Details
--- a/vertical-pod-autoscaler/enhancements/4902-delete-oom-pods/README.md
+++ b/vertical-pod-autoscaler/enhancements/4902-delete-oom-pods/README.md
@@ -42,10 +42,14 @@ failing deployment, if updated/increased limits would solve the problem.
 
 ## Proposal
 
-The proposal is to add `--experimental-deletion` to the VPA to enable deletion
-of pods. Currently only as an experimental, or beta feature.
+The proposal is to add `--delete-on-eviction-error` to the VPA to enable
+deletion of pods.
+
 To add a bit of configuration an additional flag,
-`--experimental-deletion-threshold`, should be addedd.
+`--delete-on-eviction-error-threshold`, should be addedd. This value compared
+to the amount of restarts a pod has gone through. The deletion wil only be
+allowed if the amount of restarts exceeds the threshold. This is to further
+ensure that only pods get deleted that are consistently crashing.
 
 ## Design Details
 
--- a/vertical-pod-autoscaler/enhancements/4902-delete-oom-pods/README.md
+++ b/vertical-pod-autoscaler/enhancements/4902-delete-oom-pods/README.md
@@ -13,6 +13,7 @@
    - [Update the eviction API](#update-the-eviction-api)
 <!-- /toc -->
 
+
 ## Summary
 
 The default behaviour of VPA Updater is to evict Pods when new resource
@@ -26,6 +27,7 @@ any further disruptions.
 This proposal addresses the problem by allowing users to enable the deletion of
 pods as a backup if the eviction fails.
 
+
 ## Motivation
 
 The motivation behind the change is to give VPA users a way to recover a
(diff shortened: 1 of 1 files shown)


voelzmo (commented):
Thanks for the KEP PR, we've also run into `CrashLoopBackoff` situations before where an existing PDB prevented the VPA from evicting and applying the new recommendations, so I'm really looking forward to this as a potential solution!

mwielgus:
cc: @jbartosik

RuriRyan:
bump

vertical-pod-autoscaler/enhancements/4902-delete-oom-pods/README.md
```diff
@@ -0,0 +1,79 @@
...
+- Main: allow cluster administrators to enable deletion of pods
+- Secondary: configurable deletion threshold to tune the deletion behaviour
+
+### Non-Goals
+
+- Get rid of or work around the existing eviction behaviour
+
+## Proposal
+
+The proposal is to add `--delete-on-eviction-error` to the VPA to enable
```
jbartosik, line 45: Please describe behavior of the flag in some more detail. Something like:

- when the flag is true and
- pod eviction fails and
- pod has at least one pod which:
  - is waiting with reason `CrashLoopBackOff` and
  - its last reason for termination is `OOMKilled` and 
  - its `RestartCount` is greater than ${threshold}

VPA will delete the pod.

If I get this right then please clarify why one pod meeting those conditions is enough to trigger eviction?

Also I'm not sure why you want to set a threshold on RestartCount? Could you explain to me why I might want to use value >1?

jbartosik, line 45: Why flags not API configuration? To make experiment easier to implement? Do you think there is little benefit to allowing to configure this on VPA-object level instead on a cluster level? Something else?

RuriRyan: The idea is to only use deletion as a very last resort and still mostly rely on eviction to fix problems. I still wanted to honor PDBs as much as possible, so that deletion is only used when there is no way in sight where eviciton would work.
But the more I think about the threshold flag the more unlikely the scenarios get that I initially thought about, because the other  checks will most likely be the first to not succeed.

voelzmo: > Do you think there is little benefit to allowing to configure this on VPA-object level instead on a cluster level

For us it would also be the case that we'd enable this for specific clusters entirely, not for single VPA objects. Even in the case we would make this configurable per VPA, having a global option (that people could override per VPA) would be preferred – having to touch each and every VPA is kind of cumbersome.

voelzmo: > On the other hand there are people who don't manage their clusters (for example on GKE but also when many teams share a cluster) and then they can't change value of the flag.
> 
> While I understand it might be less convenient in some cases I think it's more important to make it possible at all.

That makes perfect sense. Are these two options mutually exclusive or would we want to have a global switch _and_ the option to configure this per VPA? There are already multiple cases where this pattern exists, right?

RuriRyan: > That makes perfect sense. Are these two options mutually exclusive or would we want to have a global switch _and_ the option to configure this per VPA? There are already multiple cases where this pattern exists, right?

have you seen the update I just pushed yesterday? I also added a field for the VPA resource in addition to the flag to the proposal. It's very similar to the minReplicas setting. I think this will be the best solution everyone.

vertical-pod-autoscaler/enhancements/4902-delete-oom-pods/README.md
```diff
@@ -0,0 +1,100 @@
...
+   - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Design Details](#design-details)
+   - [Test Plan](#test-plan)
+- [Implementation History](#implementation-history)
+- [Alternatives](#alternatives)
+   - [Update the eviction API](#update-the-eviction-api)
+<!-- /toc -->
+
+
```
jbartosik, line 16: Nit: extra empty line (also a few times later)

vertical-pod-autoscaler/enhancements/4902-delete-oom-pods/README.md
```diff
@@ -0,0 +1,100 @@
...
+The default behaviour of VPA Updater is to evict Pods when new resource
+recommendations are available. This works fine for most cases.
+A problem that can arise is when there are multiple replicas and a
+PodDisruptionBudget (PDB) which allows one disruption. Now the eviction is at
+risk, because as soon as multiple replicas run into for example the memory
+limit and get killed they will never recover as the eviciton API will not allow
+any further disruptions.
+
+This proposal addresses the problem by allowing users to enable the deletion of
+pods as a backup if the eviction fails.
```
jbartosik, line 28: allowing users to enable the deletion of
*OOMing*
pods as a backup if the eviction fails.

vertical-pod-autoscaler/enhancements/4902-delete-oom-pods/README.md
```diff
@@ -0,0 +1,100 @@
...
+- 2022-05-19: initial version
+
+
+## Alternatives
+
+### Update the eviction API
+
+Instead of implementing this change on the client side, the VPA in this case,
+it could be implemented on the API side. This would have the advantage that it
+would work for all clients. On the other hand this would introduce breaking
```
jbartosik, line 94: Why doing this change in PDB would be a breaking change? I think you could just add a new filed, default being the current behavior and it would be fine.

vertical-pod-autoscaler/enhancements/4902-delete-oom-pods/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Proposal
+
+The proposal is to add a new field to the VPA resource and a global flag.
+
+A new global flag (`--delete-on-eviction-error`) shall be added to the VPA
+updater to enable the new feature globally.
+
+Additionally a new field in the VPA resource
+(`Spec.UpdatePolicy.DeleteOnEvictionError`) which takes precedence to the
```
jbartosik, line 53: I don't think `DeleteOnEvictionError` is a good name - we won't simply delete on error. `DeleteOomingOnEvictionError` ?

vertical-pod-autoscaler/enhancements/4902-delete-oom-pods/README.md
jbartosik: I think it's better to just do the API field. With both API field and flag we get the following semantics:

 - `DeleteOomingOnEvictionError = true` - delete pod if eviction failed and it's OOMing,
 - `DeleteOomingOnEvictionError = false` - don't delete pod even if it's OOMing and eviction fails (current behavior),
 - `DeleteOomingOnEvictionError` not set  - talk to your administrator or run a test to see what happens. But it can change.

I think having clean semantics on what happens by default is better than making (one time!) migration easier.

jbartosik:
One more thing: we probably should only delete OOMing pods if we plan to increase their memory request.

We have information about current request and about recommendation target in updater so it's doable. I'm not sure how much work it will be. But I think we should do it (or at least note to do it later).

RuriRyan:
> One more thing: we probably should only delete OOMing pods if we plan to increase their memory request.
> 
> We have information about current request and about recommendation target in updater so it's doable. I'm not sure how much work it will be. But I think we should do it (or at least note to do it later).

line https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/updater/logic/updater.go#L260 returns a list of pods and could potenially be changed to return more information. But this is only a surface level view and more or less the first thing I found.

I'll add a section describing this and try out a couple implementations to see how easy it is, if it gets too complicated i'll leave a note.

vertical-pod-autoscaler/enhancements/4902-delete-oom-pods/README.md
RuriRyan: If it would only be a one time migration it would be ok, but if we think of other open-source projects who have to potentially adjust their deployment mechanism to support such a setting it might take a very long time to get this done with. So I would still vote for having both.
Also asking your cluster administrator or looking into some documentation for your clusters doesn't sound that bad to me.

jbartosik:
Hi,
I'm back. It looks like we're stuck a bit on whether we should have just API change or API change and flag. I'd like to ask to help.

@RuriRyan @voelzmo  Can you make it to the SIG meeting today? If you can maybe we can resolve it over there? Sorry about asking about it at the last moment

jbartosik (approved):
/kind api-change

Looks good to me.

I think we need someone to check API.

mwielgus (approved):
/lgtm
/approve

PR #4700, opened 2022-02-21
[cluster-autoscaler] backport #4022 Publish node group min/max metrics into 1.21

richerve:
#### Which component this PR applies to?

cluster-autoscaler

#### What type of PR is this?

/kind feature

This is actually a backport

#### Which issue(s) this PR fixes:

NONE

#### Does this PR introduce a user-facing change?

```release-note
Backport #4022 Publish node group min/max metrics
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

NONE

PR #4916, opened 2022-05-25
Backport #4162 and #4172 [cluster-autoscaler] "Add a flag to control DaemonSet eviction on non-empty nodes and Allow DaemonSet pods to opt in/out from eviction" into 1.21

### LINKED ISSUE #4830: backport "daemonset-eviction-for-empty-nodes" and "daemonset-eviction-for-occupied-nodes" from 1.22 to 1.21

cluster-autoscaler

With version 1.22 cluster-autoscaler introduces the possibility to manage the eviction of daemonsets differently :
- Add a flag to control DaemonSet eviction on non-empty nodes (https://github.com/kubernetes/autoscaler/pull/4162).
- Allow DaemonSet pods to opt in/out from eviction (https://github.com/kubernetes/autoscaler/pull/4172).

But in previous versions this is not possible and may cause bugs. 
One of these bugs is perfectly described in this issue: https://github.com/kubernetes/autoscaler/issues/4337

backport the options `daemonset-eviction-for-empty-nodes` and `daemonset-eviction-for-occupied-nodes` to previous version of cluster-autoscaler ( ≤1.21)

**Additional context.**:

In our case, this causes a bug during a downscale if there is a pod on the node that uses a PVC.
because we use an "aws-ebs-csi-driver" daemonset to manage our PVCs and in this case the "aws-ebs-csi-driver" pod is killed before our application pod and so it loses communication with its PVC

Shubham82:

cluster-autoscaler

#### What type of PR is this?

/kind feature

This PR is actually a backport which backport  #4162 and #4172 in cluster-autoscaler-1.21

Fixes #4830

```
Backport "Add a flag to control DaemonSet eviction on non-empty nodes and Allow DaemonSet pods to opt in/out from eviction". 

```

```
None
```

### DIFF
--- a/cluster-autoscaler/config/autoscaling_options.go
+++ b/cluster-autoscaler/config/autoscaling_options.go
@@ -161,6 +161,8 @@ type AutoscalingOptions struct {
 	CordonNodeBeforeTerminate bool
 	// DaemonSetEvictionForEmptyNodes is whether CA will gracefully terminate DaemonSet pods from empty nodes.
 	DaemonSetEvictionForEmptyNodes bool
+	// DaemonSetEvictionForOccupiedNodes is whether CA will gracefully terminate DaemonSet pods from non-empty nodes.
+	DaemonSetEvictionForOccupiedNodes bool
 	// User agent to use for HTTP calls.
 	UserAgent string
 }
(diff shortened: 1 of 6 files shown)


Shubham82:
Hi @feiskyer @towca
Could you please take a look!

Shubham82:
@MaciekPytel  @mwielgus
PTAL!

Shubham82:
Hi @gjtempleton 
PTAL!

gjtempleton:
Hey, thanks for raising this. 

I'm a bit wary on merging this as it appears to be a new feature rather than a bug-fix, and our preference is to backport only bug-fixes. Thoughts @MaciekPytel ?

Shubham82:
Hi @gjtempleton, thanks for looking at this
something like this(feature) also backport previously. For ref #4700

PR #4921, opened 2022-05-26
add a flag to allow event duplication

### LINKED ISSUE #2495: Missing events on status config map
Even though `Scale-up: group` events (https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/scale_up.go#L715) should have been triggered as well for the status configMap as shows the log:

```
Event(v1.ObjectReference{Kind:"ConfigMap", Namespace:"test", Name:"cluster-autoscaler-status", UID:"6f9ba031-9377-4984-8a91-a8f6ef363f63", APIVersion:"v1", ResourceVersion:"225776", FieldPath:""}): type: 'Normal' reason: 'ScaledUpGroup' Scale-up: group test/e2e-d6645-w-2 size set to 2
```
I can always only see/receive one single `Scale-up: setting` (https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/scale_up.go#L701) event:

```
kubectl describe configmap cluster-autoscaler-status

Events:
  Type    Reason         Age   From                Message
  ----    ------         ----  ----                -------
  Normal  ScaledUpGroup  50s   cluster-autoscaler  Scale-up: setting group test/e2e-d6645-w-1 size to 2
```

Not sure what I'm missing.

elmiko:

cluster-autoscaler

#### What type of PR is this?

/kind feature

this change brings in a new command line flag,
`--enable-event-duplication`, which allows a user to enable the
duplication of events bypassing the 5 minute de-duplication window.

Fixes #2495

```release-note
Add enable-event-duplication flag to allow the cluster autoscaler to log duplicate events within a 5 minute window.
```

```docs

```

### DIFF
--- a/cluster-autoscaler/FAQ.md
+++ b/cluster-autoscaler/FAQ.md
@@ -28,6 +28,7 @@ this document:
 * [How to?](#how-to)
   * [I'm running cluster with nodes in multiple zones for HA purposes. Is that supported by Cluster Autoscaler?](#im-running-cluster-with-nodes-in-multiple-zones-for-ha-purposes-is-that-supported-by-cluster-autoscaler)
   * [How can I monitor Cluster Autoscaler?](#how-can-i-monitor-cluster-autoscaler)
+  * [How can I see all the events from Cluster Autoscaler?](#how-can-i-see-all-events-from-cluster-autoscaler)
   * [How can I scale my cluster to just 1 node?](#how-can-i-scale-my-cluster-to-just-1-node)
   * [How can I scale a node group to 0?](#how-can-i-scale-a-node-group-to-0)
   * [How can I prevent Cluster Autoscaler from scaling down a particular node?](#how-can-i-prevent-cluster-autoscaler-from-scaling-down-a-particular-node)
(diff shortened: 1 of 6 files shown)


elmiko:
@enxebre ptal

cluster-autoscaler/FAQ.md
```diff
@@ -759,6 +759,7 @@ The following startup parameters are supported for cluster autoscaler:
 | `daemonset-eviction-for-empty-nodes` | Whether DaemonSet pods will be gracefully terminated from empty nodes | false
 | `daemonset-eviction-for-occupied-nodes` | Whether DaemonSet pods will be gracefully terminated from non-empty nodes | true
 | `feature-gates` | A set of key=value pairs that describe feature gates for alpha/experimental features. | ""
+| `enable-event-duplication` | Enable the autoscaler to print duplicated events within a 5 minute window. | false
```
mwielgus, line 762: The name of the flag is weird. What do you think about "enable-event-deduplicaiton" (with default set to true) or "record-duplicated-events" (with default set to false).

enxebre:
Thanks! Is this to mitigate https://github.com/kubernetes/autoscaler/issues/2495? If so could you please reference in the PR?

cluster-autoscaler/FAQ.md
enxebre: It'd be also good to provide some context on when enable/disable is recommended https://github.com/kubernetes/autoscaler/issues/2495#issuecomment-547981551 cc @vivekbagade @losipiuk

elmiko: yeah, the name is a bit odd. i wanted to preserve the default behavior we have now and allow the flag to enable the old behavior. i think `record-duplicated-events` sounds like a good idea to me.

elmiko:
> Thanks! Is this to mitigate #2495? If so could you please reference in the PR?

yes it is, and yes i will!

elmiko:
updated
* changed flag name to `--record-duplicated-events`
* added FAQ entry about how to see duplicated events

PR #4926, opened 2022-05-30
Don't deref nil nodegroup in deleteCreatedNodesWithErrors

bpineau:
Various cloudproviders' `NodeGroupForNode()` implementations (including [aws](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go#L114), [azure](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/azure/azure_cloud_provider.go#L98), and [gce](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/gce/mig_info_provider.go#L99)) can returns a `nil` error _and_ a `nil` nodegroup. For instance we're seeing AWS returning that on failed upscales on live clusters, with a recent cluster-autoscaler build.

So checking that `deleteCreatedNodesWithErrors` doesn't return an error is not enough to safely dereference the nodegroup (as returned by `NodeGroupForNode()`) by calling nodegroup.Id().

In that situation, logging and returning early seems the safest option, to give various caches (eg. clusterstateregistry's and cloud provider's) the opportunity to eventually converge, rather than resuming with an inconsistent internal state.

===

With regards to the AWS cloudprovider triggering that issue with recent CA builds:

What we're seeing is:
```
I0523 16:37:53.847080     228 aws_manager.go:262] Refreshed ASG list, next refresh after 2022-05-23 16:38:53.847076185 +0000 UTC m=+559.468661724
W0523 16:37:53.905286     228 clusterstate.go:594] Nodegroup is nil for aws:///us-east-1b/i-placeholder-redacted-766c0982e71f-2
I0523 16:37:53.905533     228 clusterstate.go:1008] Found 1 instances with errorCode OutOfResource.placeholder-cannot-be-fulfilled in nodeGroup redacted-d63729665fff
I0523 16:37:53.905546     228 clusterstate.go:1026] Failed adding 1 nodes (0 unseen previously) to group redacted-d63729665fff due to OutOfResource.placeholder-cannot-be-fulfilled; errorMessages=[]string{"AWS cannot provision any more instances for this node group"}
[...]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0x3639435]
goroutine 63 [running]:
k8s.io/autoscaler/cluster-autoscaler/core.(*StaticAutoscaler).deleteCreatedNodesWithErrors(0xc03e307d00)
        /go/src/k8s.io/autoscaler/cluster-autoscaler/core/static_autoscaler.go:677 +0x275
k8s.io/autoscaler/cluster-autoscaler/core.(*StaticAutoscaler).RunOnce(0xc03e307d00, {0x4, 0x0, 0x7b9b840})
        /go/src/k8s.io/autoscaler/cluster-autoscaler/core/static_autoscaler.go:360 +0x12a5
main.run(0xc00088ae00, {0x4dc8998, 0xc000855230})
        /go/src/k8s.io/autoscaler/cluster-autoscaler/main.go:392 +0x2ad
main.main.func2({0x0, 0x0})
        /go/src/k8s.io/autoscaler/cluster-autoscaler/main.go:479 +0x25
created by k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/client-go/tools/leaderelection.(*LeaderElector).Run
        /go/src/k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/client-go/tools/leaderelection/leaderelection.go:211 +0x154
```

I can't easily reproduce the issue to verify that hypothesis, but it seems likely that this is due to the clusterstateregistry and the AWS caches being out-of-sync. That failure path is new, because flagging "failed at creation" instances was recently introduced for AWS provider (8ac87b3f34a43e3af9f7011531870ef5dc724bed).

We might get the following sequence:
1. RunOnce 1 : an AWS upscale is attempted (but the AWS ASG fails to create an instance)
2. RunOnce 2 : a Refresh() call prompts the AWS cloudprovider to regenerate the instances lists and ASG mappings, including by generating a new fake/placeholder for that failed instance
3. RunOnce 2 : clusterstateregistry gather and caches (for 2mn) the nodes list (which includes the placeholder instance)
4. RunOnce 2 : deleteCreatedNodesWithErrors() is called to gc the failed instance
5. RunOnce 3 : a Refresh() call regenerates the nodes/asg mapping, now there's no dangling placeholder nodes anymore
6. RunOnce 3 : deleteCreatedNodesWithErrors() is called and uses the outdated clusterstateregistry cache, tries to deref a nil nodegroup for a deposed placeholder instance, causes a segfault

/kind bug
/kind regression

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -365,7 +365,12 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 		return nil
 	}
 
-	if a.deleteCreatedNodesWithErrors() {
+	danglingNodes, err := a.deleteCreatedNodesWithErrors()
+	if err != nil {
+		klog.Warningf("Failed to remove nodes that were created with errors, skipping iteration: %v", err)
+		return nil
+	}
+	if danglingNodes {
 		klog.V(0).Infof("Some nodes that failed to create were removed, skipping iteration")
 		return nil
 	}
@@ -658,7 +663,7 @@ func removeOldUnregisteredNodes(unregisteredNodes []clusterstate.UnregisteredNod
 	return removedAny, nil
 }
 
-func (a *StaticAutoscaler) deleteCreatedNodesWithErrors() bool {
+func (a *StaticAutoscaler) deleteCreatedNodesWithErrors() (bool, error) {
 	// We always schedule deleting of incoming errornous nodes
 	// TODO[lukaszos] Consider adding logic to not retry delete every loop iteration
 	nodes := a.clusterStateRegistry.GetCreatedNodesWithErrors()
@@ -675,6 +680,9 @@ func (a *StaticAutoscaler) deleteCreatedNodesWithErrors() bool {
 			}
 			klog.Warningf("Cannot determine nodeGroup for node %v; %v", id, err)
 			continue
+		}
+		if nodeGroup == nil || reflect.ValueOf(nodeGroup).IsNil() {
+			return false, fmt.Errorf("node %s has no known nodegroup", node.GetName())
 		}
 		nodesToBeDeletedByNodeGroupId[nodeGroup.Id()] = append(nodesToBeDeletedByNodeGroupId[nodeGroup.Id()], node)
 	}
@@ -700,7 +708,7 @@ func (a *StaticAutoscaler) deleteCreatedNodesWithErrors() bool {
 		a.clusterStateRegistry.InvalidateNodeInstancesCacheEntry(nodeGroup)
 	}
 
-	return deletedAny
+	return deletedAny, nil
 }
 
 func (a *StaticAutoscaler) nodeGroupsById() map[string]cloudprovider.NodeGroup {
--- a/cluster-autoscaler/core/static_autoscaler_test.go
+++ b/cluster-autoscaler/core/static_autoscaler_test.go
@@ -951,7 +951,7 @@ func TestStaticAutoscalerRunOnceWithFilteringOnUpcomingNodesEnabledNoScaleUp(t *
 		podDisruptionBudgetListerMock, daemonSetListerMock, onScaleUpMock, onScaleDownMock)
 }
 
-func TestStaticAutoscalerInstaceCreationErrors(t *testing.T) {
+func TestStaticAutoscalerInstanceCreationErrors(t *testing.T) {
 
 	// setup
 	provider := &mockprovider.CloudProvider{}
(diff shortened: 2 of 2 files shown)


mwielgus (approved):
/lgtm
/approve

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -676,6 +681,9 @@ func (a *StaticAutoscaler) deleteCreatedNodesWithErrors() bool {
 			klog.Warningf("Cannot determine nodeGroup for node %v; %v", id, err)
 			continue
 		}
+		if nodeGroup == nil || reflect.ValueOf(nodeGroup).IsNil() {
+			return false, fmt.Errorf("node %s has no known nodegroup", node.GetName())
+		}
```
qianlei90, line 686: static_autoscaler will always returned when nodegroup is nil and CA can not go on until CloudProviderNodeInstancesCache is refresh, which may up to 2 min.
can we refresh CloudProviderNodeInstancesCache here to sync up with cloud provider?

@bpineau @mwielgus

qianlei90:
> We might get the following sequence:
> 
> 1. RunOnce 1 : an AWS upscale is attempted (but the AWS ASG fails to create an instance)
> 2. RunOnce 2 : a Refresh() call prompts the AWS cloudprovider to regenerate the instances lists and ASG mappings, including by generating a new fake/placeholder for that failed instance
> 3. RunOnce 2 : clusterstateregistry gather and caches (for 2mn) the nodes list (which includes the placeholder instance)
> 4. RunOnce 2 : deleteCreatedNodesWithErrors() is called to gc the failed instance
> 5. RunOnce 3 : a Refresh() call regenerates the nodes/asg mapping, now there's no dangling placeholder nodes anymore
> 6. RunOnce 3 : deleteCreatedNodesWithErrors() is called and uses the outdated clusterstateregistry cache, tries to deref a nil nodegroup for a deposed placeholder instance, causes a segfault

I think the correct sequence is:

1. RunOnce 1: scale up and failed
2. RunOnce 2: cloudprovider `Refresh()`, generate a new fake/placeholder for that failed instance
3. RunOnce 2: `deleteCreatedNodesWithErrors()` is called to gc the failed instance, and the instance cache in clusterstateregistry is removed. 
4. Between RunOnce 2 and RunOnce 3: clusterstateregistry refresh instance cache at this time and get the out-of-sync instances from provider. Notice that `CloudProviderNodeInstancesCache` runs every 2 min to refresh cache.
5. RunOnce 3: a Refresh() call regenerates the nodes/asg mapping, now there's no dangling placeholder nodes anymore
6. RunOnce 3: the outdated clusterstateregistry cache is used and we get a failed instance, then deleteCreatedNodesWithErrors() is called again, tries to deref a nil nodegroup for a deposed placeholder instance, causes a segfault

Issue #4889, opened 2022-05-16
VPA 0.11.0 release


jbartosik:
PR to increasing VPA version number merged: #4881 
I built and tested images, PR promoting them is ready: https://github.com/kubernetes/k8s.io/pull/3730

But e2e tests are failing: #4886
I think that's a problem with VPA only on K8s 1.25 and newer and I think I have a PR fixing this: #4887

To be sure I'm not releasing bad images I'll wait for the fix to merge and e2e to pass.

I also consider cherrypicking those changes to 0.11.0, rebuilding and retesting images.

jbartosik:
E2e tests are passing now.  Since the only problem was with changes in 1.25 I'll release images I've built.

PR #4929, opened 2022-05-31
Increase default VPA version to 0.11.0

jbartosik:
Issue tracking the release: #4889


jbartosik:

PR #4950, opened 2022-06-06
Vendor aws sdk in cloudprovider

AustinSiu:

Cluster Autoscaler

#### What type of PR is this?

/kind feature
/kind api-change

To get around issues with updating the version of Cluster Autoscaler's direct dependency on aws-sdk-go, this PR adds a copy of the aws-sdk-go at a specific newer version within the AWS cloudprovider.

Unblocks several PRs: https://github.com/kubernetes/autoscaler/pull/4424#discussion_r768194678 and https://github.com/kubernetes/autoscaler/pull/4588

```release-note
NONE
```

```docs

```


gjtempleton:
@AustinSiu Sorry for the delay, now getting round to this.

You'll need to exclude the newly vendored package from the verification scripts to pass the failing test. You can follow the approach taken by [this commit](https://github.com/kubernetes/autoscaler/commit/e81c4e96a2a38ba24497d4dae4d536b5edfc4c72).

gjtempleton:
A couple of notes/queries/requests before this looks good to merge:

* Do we need some instructions for those who might want/need to update the vendored code in the future given it's not as simple as just copying the SDK code in?
* Could you squash the commits down a bit before merge? I'm thinking something around the 4 commits mark, though if you have. an alternative suggestion, more than happy to listen:
  * Initial vendoring
  * Changes made to the vendored code to make it work
  * Changes to the core go.mod etc.
  * Changes to the verify scripts

AustinSiu:
> 



> A couple of notes/queries/requests before this looks good to merge:
> 
> * Do we need some instructions for those who might want/need to update the vendored code in the future given it's not as simple as just copying the SDK code in?
> * Could you squash the commits down a bit before merge? I'm thinking something around the 4 commits mark, though if you have. an alternative suggestion, more than happy to listen:
>   
>   * Initial vendoring
>   * Changes made to the vendored code to make it work
>   * Changes to the core go.mod etc.
>   * Changes to the verify scripts

Makes sense, I've cleaned up the commits and added some instructions around the vendored SDK. Please let me know if there are any other instructions I should add, etc.

gjtempleton:
Nope, that looks good to go, thanks for your persistence with this!

/lgtm
/approve

PR #4956, opened 2022-06-09
Add podScaleUpDelay annotation support

damirda:

cluster-autoscaler

#### What type of PR is this?

/kind feature

Extends existing option/feature `new-pod-scale-up-delay` that is global for all the pods and adds ability to specify custom delay per pod through annotation `cluster-autoscaler.kubernetes.io/pod-scale-up-delay`. This allows better control over scale up process in the situations when we are running a lot of batch jobs at the regular intervals, like _run this job every hour, but it is ok to wait 5 minutes to start_. 

There was a discussion on this issue and someone suggested this solution. I liked it, implemented it, and it worked well for what I needed.
https://github.com/kubernetes/autoscaler/issues/3916

```release-note
A new annotation `cluster-autoscaler.kubernetes.io/pod-scale-up-delay` has been added, which allows user to set `new-pod-scale-up-delay` per pod, instead of autoscaler wide configuration. Annotation value will be used only if it is larger then global `new-pod-scale-up-delay`.
```

```docs

```

### DIFF
--- a/cluster-autoscaler/FAQ.md
+++ b/cluster-autoscaler/FAQ.md
@@ -33,6 +33,7 @@ this document:
   * [How can I scale a node group to 0?](#how-can-i-scale-a-node-group-to-0)
   * [How can I prevent Cluster Autoscaler from scaling down a particular node?](#how-can-i-prevent-cluster-autoscaler-from-scaling-down-a-particular-node)
   * [How can I prevent Cluster Autoscaler from scaling down non-empty nodes?](#how-can-i-prevent-cluster-autoscaler-from-scaling-down-non-empty-nodes)
+  * [How can I modify Cluster Autoscaler reaction time?](#how-can-i-modify-cluster-autoscaler-reaction-time)
   * [How can I configure overprovisioning with Cluster Autoscaler?](#how-can-i-configure-overprovisioning-with-cluster-autoscaler)
   * [How can I enable/disable eviction for a specific DaemonSet](#how-can-i-enabledisable-eviction-for-a-specific-daemonset)
   * [How can I enable Cluster Autoscaler to scale up when Node's max volume count is exceeded (CSI migration enabled)?](#how-can-i-enable-cluster-autoscaler-to-scale-up-when-nodes-max-volume-count-is-exceeded-csi-migration-enabled)
@@ -331,6 +332,23 @@ CA might scale down non-empty nodes with utilization below a threshold
 
 To prevent this behavior, set the utilization threshold to `0`.
 
+### How can I modify Cluster Autoscaler reaction time?
+
+There are multiple flags which can be used to configure scale up and scale down delays.
+
+In some environments, you may wish to give the k8s scheduler a bit more time to schedule a pod than the CA's scan-interval.
+One way to do this is by setting `--new-pod-scale-up-delay`, which causes the CA to ignore unschedulable pods until they are
+a certain "age", regardless of the scan-interval. This setting can be overridden per pod through
+`cluster-autoscaler.kubernetes.io/pod-scale-up-delay` annotation. If k8s has not scheduled them by the end of that delay,
+then they may be considered by the CA for a possible scale-up.
+
+```
+"cluster-autoscaler.kubernetes.io/pod-scale-up-delay": "600s"
+```
+
+Scaling down of unneeded nodes can be configured by setting `--scale-down-unneeded-time`. Increasing value will make nodes stay
+up longer, waiting for pods to be scheduled while decreasing value will make nodes be deleted sooner.
+
 ### How can I configure overprovisioning with Cluster Autoscaler?
 
 Below solution works since version 1.1 (to be shipped with Kubernetes 1.9).
@@ -620,10 +638,7 @@ then this node group may be excluded from future scale-ups.
 ### How fast is Cluster Autoscaler?
 
 By default, scale-up is considered up to 10 seconds after pod is marked as unschedulable, and scale-down 10 minutes after a node becomes unneeded.
-There are multiple flags which can be used to configure these thresholds. For example, in some environments, you may wish to give the k8s scheduler
-a bit more time to schedule a pod than the CA's scan-interval. One way to do this is by setting `--new-pod-scale-up-delay`, which causes the CA to
-ignore unschedulable pods until they are a certain "age", regardless of the scan-interval. If k8s has not scheduled them by the end of that delay,
-then they may be considered by the CA for a possible scale-up.
+Read [this section](#how-can-i-modify-cluster-autoscaler-reaction-time) to see how you can modify this behaviour.
 
 Assuming default settings, [SLOs described here apply](#what-are-the-service-level-objectives-for-cluster-autoscaler).
 
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -62,6 +62,9 @@ const (
 
 	// NodeUpcomingAnnotation is an annotation CA adds to nodes which are upcoming.
 	NodeUpcomingAnnotation = "cluster-autoscaler.k8s.io/upcoming-node"
+
+	// podScaleUpDelayAnnotationKey is an annotation how long pod can wait to be scaled up.
+	podScaleUpDelayAnnotationKey = "cluster-autoscaler.kubernetes.io/pod-scale-up-delay"
 )
 
 // StaticAutoscaler is an autoscaler which has all the core functionality of a CA but without the reconfiguration feature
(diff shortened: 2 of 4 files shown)


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -719,13 +722,27 @@ func (a *StaticAutoscaler) nodeGroupsById() map[string]cloudprovider.NodeGroup {
...
 	for _, pod := range allUnschedulablePods {
 		podAge := currentTime.Sub(pod.CreationTimestamp.Time)
-		if podAge > newPodScaleUpDelay {
+		// read podScaleUpDelay from annotations
+		// and use that if bigger than newPodScaleUpDelay
+		podScaleUpDelay := newPodScaleUpDelay
+		if podScaleUpDelayAnnotationStr, ok := pod.Annotations[podScaleUpDelayAnnotationKey]; ok {
+			podScaleUpDelayAnnotation, err := time.ParseDuration(podScaleUpDelayAnnotationStr)
+			if err != nil {
+				klog.Errorf("Failed to parse pod annotation: %v", err)
```
x13n, line 738: nit: Can you identify the pod and annotation in the logs? Maybe `klog.Errorf("Failed to parse pod %q annotation %s: %v", pod.Name, podScaleUpDelayAnnotationKey, err)`

x13n (commented):
The code change LGTM, but I think we also need:
- a unit test confirming this works 
- FAQ entry about the new annotation

Can you add it?

damirda:
@x13n Thank you for your help with this PR so far!
I added FAQ entry that basically covers my use case. I also added unit test for the existing function that I extended.
Let me know if I should add more tests or if you think FAQ update should be more generalized.

cluster-autoscaler/FAQ.md
```diff
@@ -331,6 +332,23 @@ CA might scale down non-empty nodes with utilization below a threshold
 
 To prevent this behavior, set the utilization threshold to `0`.
 
+### How can I prevent Cluster Autoscaler from scaling up too quickly?
+
+Running many shortlived scheduled jobs in a regular intervals could lead to cluster running in
+cycle of scaling up, executing jobs quickly and then waiting for `scale-down-unneeded-time`
```
x13n, line 338: Increasing `--scale-down-unneeded-time` is also a way of improving the situation. In a way the two flags are complementary: `--scale-down-unneeded-time` controls how long a node will wait for pods, `--new-pod-scale-up-delay` controls how long a pod will wait for a node.

cluster-autoscaler/FAQ.md
```diff
@@ -621,7 +639,8 @@ then this node group may be excluded from future scale-ups.
 By default, scale-up is considered up to 10 seconds after pod is marked as unschedulable, and scale-down 10 minutes after a node becomes unneeded.
```
x13n, line 639: Can we avoid duplicating content? I'd drop most of this paragraph and leave just:

> By default, scale-up is considered up to 10 seconds after pod is marked as unschedulable, and scale-down 10 minutes after a node becomes unneeded. Read [this section](...) to see how you can modify this behavior.

x13n (changes requested):

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -719,17 +722,32 @@ func (a *StaticAutoscaler) nodeGroupsById() map[string]cloudprovider.NodeGroup {
...
-// don't consider pods newer than newPodScaleUpDelay seconds old as unschedulable
+// don't consider pods newer than newPodScaleUpDelay or annotated podScaleUpDelay
+// seconds old as unschedulable
 func (a *StaticAutoscaler) filterOutYoungPods(allUnschedulablePods []*apiv1.Pod, currentTime time.Time) []*apiv1.Pod {
 	var oldUnschedulablePods []*apiv1.Pod
 	newPodScaleUpDelay := a.AutoscalingOptions.NewPodScaleUpDelay
 	for _, pod := range allUnschedulablePods {
 		podAge := currentTime.Sub(pod.CreationTimestamp.Time)
-		if podAge > newPodScaleUpDelay {
+		// read podScaleUpDelay from annotations
```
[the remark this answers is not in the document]
damirda: Yes, I agree that comment is unnecessary.

cluster-autoscaler/FAQ.md
damirda: How would increasing `--scale-down-unneeded-time` improve the situation? You will just get nodes that are started running even longer unneeded?

x13n: The way I understand the problem here is that there is too high node churn - too many nodes coming down and up again, wasting compute resources on startup/shutdown. Keeping nodes for a longer time after they become empty is a way of addressing this, so is introducing a delay before triggering a scale up.

If the problem is phrased as "scale up happens too quickly" then it can also help: nodes are not deleted, giving future pods a chance to get scheduled on them instead of triggering a scale up.

damirda: My problem was that nodes are coming up too fast and then staying up too long. :smile: 
We did decrease `--scale-down-unneeded-time` to 5 minutes and it did help, but anyway, I understand your point. I will write in a way to cover different cases, depending on what is that you want to achieve.

damirda:
This failing test seem to be flaky. Works fine locally and also `master` is not broken.
Plus completely unrelated to my changes. What is the best way to handle that?

damirda:
Never mind, restarting the build helped...

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -1377,6 +1380,103 @@ func TestSubtractNodes(t *testing.T) {
...
+			pods:               []*apiv1.Pod{p1, p3},
+			expectedPods:       []*apiv1.Pod{p1},
+			expectedError:      "Failed to set pod scale up delay for",
+		},
+		{
+			name:               "annotation delay with error",
+			newPodScaleUpDelay: 0,
+			runTime:            now,
+			pods:               []*apiv1.Pod{p1, p4},
+			expectedPods:       []*apiv1.Pod{p1},
```
x13n, line 1448: I don't think pods with malformed annotation should be ignored. I think logging an error should be enough - if annotation cannot be parsed, CA behavior should fall back to global setting instead of banning the pod forever.

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -1377,6 +1380,103 @@ func TestSubtractNodes(t *testing.T) {
...
+			var buf bytes.Buffer
+			klog.SetOutput(&buf)
+			defer func() {
+				klog.SetOutput(os.Stderr)
+			}()
+
+			actual := autoscaler.filterOutYoungPods(tt.pods, tt.runTime)
+
+			assert.ObjectsAreEqual(tt.expectedPods, actual)
+			if tt.expectedError != "" {
```
x13n, line 1473: nit: Is this condition changing anything? `Contains(t, buf.String(), "")` is always true, isn't it?

x13n (changes requested):

cluster-autoscaler/core/static_autoscaler_test.go
damirda: I just think it makes test more readable - if there is an error expected, check if we got it. But it would work wothout `if` as well.

cluster-autoscaler/core/static_autoscaler_test.go
damirda: There was a bug in my tests, so it didn't catch that my code reformatting commit actually broke this behavior. Now I reverted code reformatting (pretty much completely) and fixed the tests.

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -1430,22 +1430,22 @@ func TestFilterOutYoungPods(t *testing.T) {
...
 			runTime:            now,
 			pods:               []*apiv1.Pod{p1, p2},
-			expectedPods:       []*apiv1.Pod{},
+			expectedPods:       []*apiv1.Pod(nil),
 		},
 		{
 			name:               "annotation delay smaller than global",
 			newPodScaleUpDelay: 5 * time.Minute,
 			runTime:            now,
 			pods:               []*apiv1.Pod{p1, p3},
```
x13n, line 1439: `p3` has annotation of 2m and is created 1m before `runTime`, so it wouldn't be included even if the implementation allowed for annotation-based delay to be smaller than the global one. I think this test should use `runTime: now.Add(2 * time.Minute)` to actually show the difference.

x13n (commented):
LGTM, just one suggested improvement in tests.

cluster-autoscaler/core/static_autoscaler_test.go
damirda: Yes, that is actually very good point. Thank you!

damirda:
@x13n Thank you for all your help on this PR.
I will squash all the commits in the end and prepare this for a merge.

x13n:
Merge will happen automatically once this has both lgtm and approved labels, so I'd suggest doing it first (it will also reset the lgtm tag, so I'll need to reapply it).

damirda:
@towca Could we get a review here?

towca:
Looks good to me, thanks for the contribution!

/approve

PR #4959, opened 2022-06-10
Correct invalid GCE instances pricing

BigDarkClown:

cluster-autoscaler

#### What type of PR is this?

/kind bug

This PR fixes the internal GCE instance pricing config:
- `n2-highmem-128` was invalid for both standard and preemptible instances
- `n2d` family cost was invalid for preemptible instances


mwielgus (approved):
/lgtm
/approve

PR #4971, opened 2022-06-14
CA - Push Image script - Support newer docker versions

gjtempleton:

Cluster-autoscaler

#### What type of PR is this?

/kind feature

Adds branching logic for the image push command to run based on the version of Docker being used to push the images as per: https://cloud.google.com/container-registry/docs/support/deprecation-notices#gcloud-docker

N/A

Open question: Do we even need to continue support for the `gcloud docker --` command given the age of the docker version this requires?

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/push_image.sh
+++ b/cluster-autoscaler/push_image.sh
@@ -14,15 +14,28 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


MaciekPytel:
/lgtm
On the question of supporting gcloud docker: I don't have particularly strong feelings either way - I don't think we have to keep supporting it, but also it costs us very little given that you already wrote the code. If you feel that we should drop it, I'm fine with that.

PR #4973, opened 2022-06-15
Handle gracefully empty arch in kube env.

olagacek:

cluster-autoscaler

#### What type of PR is this?
/kind bug

It prevents error in cluster-autoscaler in case empty arch is provided in kube env.

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/templates.go
+++ b/cluster-autoscaler/cloudprovider/gce/templates.go
@@ -599,7 +599,7 @@ func ToSystemArchitecture(arch string) SystemArchitecture {
 	case string(Amd64):
 		return Amd64
 	default:
-		return UnknownArch
+		return DefaultArch
 	}
 }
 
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -598,6 +598,8 @@ func ToSystemArchitecture(arch string) SystemArchitecture {
 		return Arm64
 	case string(Amd64):
 		return Amd64
+	case "":
```
MaciekPytel, line 601: Shouldn't we just change default: to return DefaultArch? It seems surprising to differentiate.

MaciekPytel:
/approve
/hold
Approve, because the fix seems clearly like the right thing to fix. Hold to give you a chance to reply to my comment.
Please feel free to cancel hold without my explicit lgtm.

cluster-autoscaler/cloudprovider/gce/templates.go
olagacek: I decided to go with using DefaultArch as default:, given that incorrect value in this place will lead to cluster autoscaler crash in main loop.

PR #4041, opened 2021-04-27
Add Rancher as a cloud provider

douglasmakey:
Hello Kubernetes Community,

We are the Ubisoft Kubernetes Team:  @douglasmakey, @patrickdappollonio, and @promagne. We worked during the past couple of months alongside Rancher (and now Suse) to implement a Node Autoscaler [based on the Rancher software](https://github.com/rancher/rancher) as a cloud provider, and we're sending it here for broader community use. 

It has the basic support you would expect, like increasing and decreasing node groups and configuring using a `ConfigMap`.

We're here if you have any questions!

PR #4975, opened 2022-06-15
Add cloud provider for Rancher with RKE2

ctrox:

cluster-autoscaler

#### What type of PR is this?

/kind feature

Adds cluster-autoscaler provider for [Rancher](https://github.com/rancher/rancher/) with RKE2. This has been tried [before](https://github.com/kubernetes/autoscaler/pull/4041) but as this was for RKE (1) clusters, it works in a completely different way as Rancher fully changed their APIs and provisioning for RKE2. This provider only supports RKE2, which is now GA as of Rancher v2.6.5.

RKE2 provisioning uses cluster-api under the hood, so there are some similarities to the cluster-api provider but not enough to be able to share code. Plus I assume that is probably not really desired anyway. Since the Rancher server manages the cluster-api objects, it's not possible to simply run the cluster-api autoscaler as these two controllers would fight over control, plus changes would not be reflected in the corresponding Rancher API objects.

Not super sure about the proper process of the `OWNERS` file, but I saw in a recent PR that added a cloud provider that it was requested that there's a `OWNERS` file with the gitub handles commented out, so I added that. All the owners I added are members of @ninech and we are committed to maintaining this but of course would be more than happy if anybody else wants to help maintain this (e.g Rancher themselves or Ubisoft which was working on the old PR).

yes

```release-note
Adds support for Rancher with RKE2 in cluster-autoscaler
```

```docs

```

### DIFF
--- a/cluster-autoscaler/README.md
+++ b/cluster-autoscaler/README.md
@@ -33,6 +33,7 @@ You should also take a look at the notes and "gotchas" for your specific cloud p
 * [Vultr](./cloudprovider/vultr/README.md)
 * [TencentCloud](./cloudprovider/tencentcloud/README.md)
 * [Scaleway](./cloudprovider/scaleway/README.md)
+* [Rancher](./cloudprovider/rancher/README.md)
 
 # Releases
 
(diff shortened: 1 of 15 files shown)


mwielgus:
Please rebase and we are good to go.

ctrox:
> Please rebase and we are good to go.

Done, thank you!

mwielgus (approved):
/lgtm
/approve

PR #4983, opened 2022-06-21
Reenable Service Account Check in VPA E2E tests

### LINKED ISSUE #4822: Update VPA E2E K8s dependencies and disable workaround
Clean up workaround for #4764:

- Update K8s deps of E2E tests to [v1.24.0-beta.0](https://github.com/kubernetes/kubernetes/blob/v1.24.0-beta.0/test/e2e/framework/util.go#L311) or newer,
- Clean up workaround,  rollback #4821.

jbartosik:

Vertical Pod Autoscaler

#### What type of PR is this?

/kind cleanup

Sync VPA E2E K8s dependencies to 1.24, which enables us to re-enable service account verification in E2E tests, do that (by reverting #4821).

Fixes #4822

No.

```release-note
NONE
```


jbartosik:

jbartosik:
/hold

The check passed but tests failed for other reasons. I want to make sure that dependency update doesn't break tests

jbartosik:
/hold cancel

Tests pass now.

jbartosik:
I opened #4991 to follow up with reducing pod privileges

vertical-pod-autoscaler/e2e/v1/actuation.go
```diff
@@ -41,13 +41,15 @@ import (
 	framework_rs "k8s.io/kubernetes/test/e2e/framework/replicaset"
 	framework_ss "k8s.io/kubernetes/test/e2e/framework/statefulset"
 	testutils "k8s.io/kubernetes/test/utils"
+	admissionapi "k8s.io/pod-security-admission/api"
```
kgolab, line 44: Would it make sense to use a different package shorthand, to avoid confusion with admission.k8s.io API? How about podsecurity ?

vertical-pod-autoscaler/e2e/v1/recommender.go
```diff
@@ -22,6 +22,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/onsi/ginkgo"
```
kgolab, line 25: Why move these two imports, we usually keep them after k8s.io (as a separate section)?

kgolab:
Thanks, generally LGTM, with only nit comments left.

vertical-pod-autoscaler/e2e/v1/recommender.go
jbartosik: No reason to do that.

vertical-pod-autoscaler/e2e/v1/actuation.go
jbartosik: Done.

vertical-pod-autoscaler/e2e/go.mod
```diff
@@ -8,14 +8,14 @@ go 1.17
...
+	k8s.io/apimachinery v0.24.3-rc.0
 	k8s.io/autoscaler/vertical-pod-autoscaler v0.0.0-20200605154545-936eea18fb1d
 	k8s.io/client-go v11.0.0+incompatible
-	k8s.io/component-base v0.23.5
-	k8s.io/klog/v2 v2.30.0
-	k8s.io/kubernetes v1.23.5
-	k8s.io/utils v0.0.0-20211116205334-6203023598ed
+	k8s.io/component-base v0.24.3-rc.0
+	k8s.io/klog/v2 v2.60.1
+	k8s.io/kubernetes v1.24.3-rc.0
```
MaciekPytel, line 17: Why use rc instead of actual 1.24.3?

MaciekPytel (commented):
Shouldn't this also include removing the flag in vertical-pod-autoscaler/hack/run-e2e-tests.sh? Or is that coming in a follow-up PR?

vertical-pod-autoscaler/e2e/go.mod
jbartosik: Changed to 1.24.3

jbartosik:
> 

https://github.com/kubernetes/autoscaler/pull/4983/commits/82ed501f168583c40109e1af75d52f0bd1f4add7 does that

MaciekPytel:
> https://github.com/kubernetes/autoscaler/commit/82ed501f168583c40109e1af75d52f0bd1f4add7 does that

You're right, it looks like the issue was github diff not showing complete list of affected files. Sorry for false alert.

/lgtm
/approve

PR #4999, opened 2022-06-29
Add note in Azure autoscaler documentation about adding accelerator node label on VMSS when GPU nodes are involved.

Weihrauch:

cluster-autoscaler

#### What type of PR is this?

/kind documentation

I struggled for a long running issue arround GPU node scaling in a Rancher RKE Kubernetes Cluster using Azure VM Scale Set. The scale up process freeze after node count back to zero. After digging the source code I discover that label `accelerator` (hardcoded in azure cloud provider) should be setted on kubelet node to make cluster-autoscaler handle correctly GPU ressources.
I guess it would help further user to deal with it.

NONE

For more information on release notes see: https://git.k8s.io/community/contributors/guide/release-notes.md
-->
```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/README.md
+++ b/cluster-autoscaler/cloudprovider/azure/README.md
@@ -54,6 +54,8 @@ k8s.io_cluster-autoscaler_node-template_resources_cpu: 3800m
 k8s.io_cluster-autoscaler_node-template_resources_memory: 11Gi
 ```
 
+> **_NOTE_**: GPU autoscaling consideration on VMSS : In case of scale set of GPU nodes, kubelet node label `accelerator` have to be added to node provisionned to make GPU scaling works.
+
 #### Autoscaling options
 
 Some autoscaling options can be defined per VM Scale Set, with tags.


marwanad:
The GPU label isn't really an Azure-specific thing (nor VMSS for that matter) and it is part of the cloudprovider API contract so every cloudprovider has its notion of defining what the label on GPU nodes is so that CA can identify those.

https://github.com/kubernetes/autoscaler/blob/a13c59c2430e5fe0e07d8233a536326394e0c925/cluster-autoscaler/cloudprovider/cloud_provider.go#L112-L113

I wonder if we should put a note in the main FAQ instead.

https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md

mwielgus (approved):
/lgtm
/approve

mweibel:
thanks for this change @Weihrauch 

is it correct that you'd need to set the label `accelerator` with value `nvidia-tesla-v100`?

Weihrauch:
Hello mweibel,
On my side I only had labl accelerator. This requirement trigger GPU instance handling in autoscaler algorithm as you can see in NodeHasGpu function. Setting value to this label may configure other things I have no ideas by now.
`func NodeHasGpu(GPULabel string, node *apiv1.Node) bool {
	_, hasGpuLabel := node.Labels[GPULabel]
	gpuAllocatable, hasGpuAllocatable := node.Status.Allocatable[ResourceNvidiaGPU]
	return hasGpuLabel || (hasGpuAllocatable && !gpuAllocatable.IsZero())
}`

PR #5008, opened 2022-07-05
Replacing DeletedNode struct usage with *apiv1.Node directly

fookenc:
cluster-autoscaler

#### What type of PR is this?
/kind cleanup

This PR is a follow-up to the approved PR #4896.  The struct usage (and timestamp field) were unnecessary and didn't serve a purpose in this use-case.  This code change is meant to replace the usage of the newly defined struct in favor of using apiv1.Node directly instead.  A few function definitions have been modified to reflect the new return type and to remove unnecessary parameters for timestamps, which were only utilized in the struct.

NONE

NONE

NONE

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -99,15 +99,6 @@ type UnregisteredNode struct {
 	UnregisteredSince time.Time
 }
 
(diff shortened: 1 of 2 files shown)


fookenc:
Closing PR due to reversion of the code this one is based upon.  Further review and discussion is needed to fully address this without intended side-effects.  Bug #5022 will address the needs of the PRs described here and previously.

PR #5013, opened 2022-07-07
Update Vendor v1.25.0-alpha.2 , bump Go to 1.18.1 and additional support for StorageInfoListers

### LINKED ISSUE #4904: Extend ClusterSnapshot implementations to include StorageInfos lister

cluster autoscaler

Recent scheduler framework changes will cause Cluster Autoscaler code to stop compiling once we update the dependency on scheduler.

There are 3 objects implementing `SharedFramework` interface:
- BasicClusterSnapshot (https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/simulator/basic_cluster_snapshot.go)
- DeltaClusterSnapshot (https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/simulator/delta_cluster_snapshot.go)
- Delegating shared lister: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/simulator/delegating_shared_lister.go (fix is trivial, just delegate the new extra functions).

Both BasicClusterSnapshot and DeltaClusterSnapshot can maintain a mapping from namespaced PVC names to sets of namespaced pod names. `IsPVCUsedByPods` can then just check if size of the set is non-0. This will have O(1) complexity on adding/removing pods. In delta snapshot case it also needs to track deleted pods similarly to how node infos are tracked today, to avoid O(n) computation when forking.

Doing no-op implementation - would be simpler, but would incorrectly handle `ReadWriteOncePod` PVCs, leading to scale ups that wouldn't be able to help some pending pods.

**Additional context.**:

Scheduler PR changing SharedLister interface: https://github.com/kubernetes/kubernetes/pull/109715

jayantjain93:

cluster-autoscaler
#### What type of PR is this?
/kind feature

Fixes #4904

```release-note

```

```docs

```


cluster-autoscaler/simulator/delta_cluster_snapshot.go
```diff
@@ -353,6 +355,27 @@ func (snapshot *deltaSnapshotNodeLister) Get(nodeName string) (*schedulerframewo
...
+	for _, v := range nodeInfos {
+		for k, i := range v.PVCRefCounts {
+			if _, found := pvcNamespaceMap[k]; !found {
+				pvcNamespaceMap[k] = 0
+			}
+			pvcNamespaceMap[k] += i
+		}
+	}
+	data.pvcNamespaceMap = pvcNamespaceMap
+	if v, found := data.pvcNamespaceMap[key]; found && v > 0 {
```
x13n, line 376: nit: Since zero value for int is 0, you can just do `return data.pvcNamespaceMap[key] > 0` here: it doesn't matter if map key was found or not. Same above, since these lines are duplicated.

cluster-autoscaler/simulator/basic_cluster_snapshot.go
```diff
@@ -71,9 +72,43 @@ func (data *internalBasicSnapshotData) getNodeInfo(nodeName string) (*schedulerf
 	return nil, errNodeNotFound
 }
 
+func (data *internalBasicSnapshotData) isPVCUsedByPods(key string) bool {
+	if v, found := data.pvcNamespaceNodeMap[key]; found && v != nil && len(v) > 0 {
```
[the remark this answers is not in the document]
jayantjain93: I had similar thoughts, but it think it is possible to set map[k] = nil. So found == true may not be an initialised map.

cluster-autoscaler/simulator/basic_cluster_snapshot.go
```diff
@@ -31,7 +31,8 @@ type BasicClusterSnapshot struct {
 }
 
 type internalBasicSnapshotData struct {
-	nodeInfoMap map[string]*schedulerframework.NodeInfo
+	nodeInfoMap         map[string]*schedulerframework.NodeInfo
+	pvcNamespaceNodeMap map[string]map[*schedulerframework.NodeInfo]bool
```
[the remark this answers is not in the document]
jayantjain93: Done.

cluster-autoscaler/simulator/delta_cluster_snapshot.go
jayantjain93: Done.

cluster-autoscaler/simulator/basic_cluster_snapshot.go
```diff
@@ -71,9 +72,56 @@ func (data *internalBasicSnapshotData) getNodeInfo(nodeName string) (*schedulerf
...
+	for ref := range nodeInfo.PVCRefCounts {
+		if _, found := data.pvcNamespaceNodeMap[ref]; found {
+			if _, found := data.pvcNamespaceNodeMap[ref]; found {
+				delete(data.pvcNamespaceNodeMap[ref], nodeInfo.Node().GetName())
+			}
+		}
+	}
+}
+
+func (data *internalBasicSnapshotData) removePvcUsedByNodes(nodeInfo *schedulerframework.NodeInfo) {
```
x13n, line 109: Shouldn't this function be removed now?

cluster-autoscaler/simulator/basic_cluster_snapshot.go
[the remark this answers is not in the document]
x13n: It is possible in general, but here you're always initializing the map when creating new entries.

cluster-autoscaler/simulator/delta_cluster_snapshot.go
```diff
@@ -353,6 +355,27 @@ func (snapshot *deltaSnapshotNodeLister) Get(nodeName string) (*schedulerframewo
...
+	for _, v := range nodeInfos {
+		for k, i := range v.PVCRefCounts {
+			if _, found := pvcNamespaceMap[k]; !found {
+				pvcNamespaceMap[k] = 0
+			}
+			pvcNamespaceMap[k] += i
+		}
+	}
+	data.pvcNamespaceMap = pvcNamespaceMap
+	if data.pvcNamespaceMap[key] > 0 {
```
x13n, line 376: nit: Out of curiosity, why do `if cond { return true } return false` instead of just `return cond`?

cluster-autoscaler/simulator/basic_cluster_snapshot.go
jayantjain93: Done.

cluster-autoscaler/simulator/delta_cluster_snapshot.go
```diff
@@ -353,7 +355,25 @@ func (snapshot *deltaSnapshotNodeLister) Get(nodeName string) (*schedulerframewo
 
 // IsPVCUsedByPods returns if PVC is used by pods
 func (snapshot *deltaSnapshotStorageLister) IsPVCUsedByPods(key string) bool {
-	return false
+	data := snapshot.data
+	if data.pvcNamespaceMap != nil {
+		if data.pvcNamespaceMap[key] > 0 {
```
x13n, line 360: Same as below, can be simplified from `if cond { return true } return false` to just `return cond`.

jayantjain93:
> 

Also added tests for the change

cluster-autoscaler/simulator/cluster_snapshot.go
```diff
@@ -39,6 +39,8 @@ type ClusterSnapshot interface {
 	RemovePod(namespace string, podName string, nodeName string) error
 	// AddNodeWithPods adds a node and set of pods to be scheduled to this node to the snapshot.
 	AddNodeWithPods(node *apiv1.Node, pods []*apiv1.Pod) error
+	// IsPVCUsedByPods returns if the pvc is used by any pod
```
x13n, line 42: nit: maybe worth noting here that `key` is supposed to be in `<namespace>/<pvc_name>` format.

cluster-autoscaler/simulator/cluster_snapshot_test.go
```diff
@@ -464,3 +463,102 @@ func TestNodeAlreadyExists(t *testing.T) {
 		}
 	}
 }
+
+func TestPVCUsedByPods(t *testing.T) {
```
x13n, line 467: Please make this test table-based and include a test case with a non-PVC volume.

cluster-autoscaler/simulator/delta_cluster_snapshot.go
```diff
@@ -180,6 +181,7 @@ func (data *internalDeltaSnapshotData) clearCaches() {
 func (data *internalDeltaSnapshotData) clearPodCaches() {
 	data.havePodsWithAffinity = nil
 	data.havePodsWithRequiredAntiAffinity = nil
+	data.pvcNamespaceMap = nil
```
x13n, line 184: Can you add a TODO here to update the cache when adding/removing pods instead of invalidating the whole cache?

jayantjain93: done.

cluster-autoscaler/simulator/cluster_snapshot_test.go
jayantjain93: Included the test.
I don't think we should try to make this a table format. Since each case is testing using different setup. Some require add pod/ some delete pod.

cluster-autoscaler/simulator/cluster_snapshot.go
jayantjain93: done

cluster-autoscaler/simulator/basic_cluster_snapshot.go
```diff
@@ -71,9 +72,46 @@ func (data *internalBasicSnapshotData) getNodeInfo(nodeName string) (*schedulerf
...
+}
+
+func (data *internalBasicSnapshotData) removePvcUsedByPod(pod *apiv1.Pod) {
+	if pod == nil {
+		return
+	}
+
+	nameSpace := pod.GetNamespace()
+	for _, volume := range pod.Spec.Volumes {
+		k := schedulerframework.GetNamespacedName(nameSpace, volume.PersistentVolumeClaim.ClaimName)
```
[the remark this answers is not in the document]
jayantjain93: done.

cluster-autoscaler/simulator/basic_cluster_snapshot.go
```diff
@@ -71,9 +72,46 @@ func (data *internalBasicSnapshotData) getNodeInfo(nodeName string) (*schedulerf
...
+	return false
+}
+
+func (data *internalBasicSnapshotData) addPvcUsedByPod(pod *apiv1.Pod) {
+	if pod == nil {
+		return
+	}
+	nameSpace := pod.GetNamespace()
+	for _, volume := range pod.Spec.Volumes {
+		k := schedulerframework.GetNamespacedName(nameSpace, volume.PersistentVolumeClaim.ClaimName)
```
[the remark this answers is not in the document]
jayantjain93: done

cluster-autoscaler/simulator/cluster_snapshot_test.go
x13n: Each test case is following rougly the same pattern: 
- Add a node
- Add pods to the node
- Do a check
- (optionally) Remove some pods and do a second check

This can be encoded into a table format quite easily.

jayantjain93: done

cluster-autoscaler/simulator/cluster_snapshot_test.go
```diff
@@ -464,3 +463,132 @@ func TestNodeAlreadyExists(t *testing.T) {
...
+			},
+		},
+	}
+	testcase := []struct {
+		desc                 string
+		node                 *apiv1.Node
+		pods                 []*apiv1.Pod
+		claimName            string
+		exists               bool
+		isRemoveCheck        bool
```
x13n, line 521: nit: `isRemoveCheck` is redundant with `removePod != ""`.

cluster-autoscaler/simulator/cluster_snapshot_test.go
```diff
@@ -464,3 +463,132 @@ func TestNodeAlreadyExists(t *testing.T) {
...
+		removePod            string
+		claimNameAfterRemove string
```
x13n, line 523: Why not just reuse `claimName`? It doesn't make sense to check different claims before and after pod removal.

jayantjain93: Done.

mwielgus:
@jayantjain93 Where are we with this PR? It is LGTM'ed and the only missing thing seems to be the rebase.

cluster-autoscaler/simulator/cluster_snapshot_test.go
```diff
@@ -464,3 +463,126 @@ func TestNodeAlreadyExists(t *testing.T) {
 		}
 	}
 }
+
+func TestPVCUsedByPods(t *testing.T) {
```
MaciekPytel, line 467: I'm not going to push super hard, but I feel this is a pretty minimal test coverage for a significant change in such a critical part of CA. For example forking or clearing are pretty big and critical parts of snapshot and I'd like to see tests that validate, for example, that Clear() also clears PVC data.

cluster-autoscaler/simulator/cluster_snapshot_test.go
```diff
@@ -464,3 +463,126 @@ func TestNodeAlreadyExists(t *testing.T) {
...
+			desc:              "pvc new pod with delete volume fetch",
+			node:              node,
+			pods:              []*apiv1.Pod{pod1},
+			claimName:         "claim1",
+			exists:            true,
+			removePod:         "pod1",
+			existsAfterRemove: false,
+		},
+		{
+			desc:              "pvc two pods with duplicated volume, delete one pod, fetch",
```
MaciekPytel, line 558: I feel like some test cases with multiple insertions and deletions, including both pods using and not using a pvc would also be good.

MaciekPytel:
@mwielgus That's on me, I wanted to review this one and haven't manage to do it for a long time.

MaciekPytel:
/hold
/approve

@jayantjain93 Sorry it took me so long to get to this.
Overall looks good, but I left a few minor comments. Leaving a hold to avoid accidentally merging before you have a chance to look at those. Please feel free to "/hold cancel" without blocking on my review.

cluster-autoscaler/simulator/cluster_snapshot_test.go
jayantjain93: Done.

cluster-autoscaler/simulator/cluster_snapshot_test.go
jayantjain93: Done.

cluster-autoscaler/simulator/cluster_snapshot_test.go
```diff
@@ -464,3 +463,126 @@ func TestNodeAlreadyExists(t *testing.T) {
...
+			claimName:         "claim1",
+			exists:            true,
+			removePod:         "pod1",
+			existsAfterRemove: true,
+		},
+	}
+
+	for _, snapshotFactory := range snapshots {
+		for _, tc := range testcase {
+			t.Run(tc.desc, func(t *testing.T) {
```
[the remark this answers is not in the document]
jayantjain93: done.

cluster-autoscaler/simulator/basic_cluster_snapshot.go
```diff
@@ -71,9 +72,52 @@ func (data *internalBasicSnapshotData) getNodeInfo(nodeName string) (*schedulerf
...
+	}
+
+	nameSpace := pod.GetNamespace()
+	for _, volume := range pod.Spec.Volumes {
+		if volume.PersistentVolumeClaim == nil {
+			continue
+		}
+		k := schedulerframework.GetNamespacedName(nameSpace, volume.PersistentVolumeClaim.ClaimName)
+		if _, found := data.pvcNamespacePodMap[k]; found {
+			delete(data.pvcNamespacePodMap[k], pod.GetName())
```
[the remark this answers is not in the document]
jayantjain93: done.

x13n:
I think delta snapshot would be more efficient if it stored actual deltas (possibly negative) of PVC usage counters. Now we'll iterate over all nodeinfos when trying to schedule a pod using PVC on a forked snapshot. This is a huge and 1.25-blocking change though, so let's merge this, we can always optimize further if there's a scalability regression.

/lgtm

PR #5015, opened 2022-07-11
Set different hostname label for upcoming nodes

liorfranko:
Function copying template node to use for upcoming nodes was
not chaning hostname label, meaning that features relying on
this label (ex. pod antiaffinity on hostname topology) would
treat all upcoming nodes as a single node.
This resulted in triggering too many scale-ups for pods
using such features. Analogous function in binpacking didn't
have the same bug (but it didn't set unique UID or pod names).
I extracted the functionality to a util function used in both
places to avoid the two functions getting out of sync again.

Cluster Autoscaler

#### What type of PR is this?
/kind bug

Fixes #
Charry pick https://github.com/kubernetes/autoscaler/pull/3883

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -23,7 +23,6 @@ import (
 
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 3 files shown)


x13n:
We're not releasing to 1.20 branch anymore.

/close

PR #5025, opened 2022-07-13
Get capi targetsize from cache

enxebre:

Cluster autoscaler - cluster api.

#### What type of PR is this?
This ensured that access to replicas during scale down operations were never stale by accessing the API server https://github.com/kubernetes/autoscaler/issues/3104.

This honoured that behaviour while moving to unstructured client https://github.com/kubernetes/autoscaler/pull/3312.

This regressed that behaviour while trying to reduce the API server load https://github.com/kubernetes/autoscaler/pull/4443.

This put back the never stale replicas behaviour at the cost of loading back the API server https://github.com/kubernetes/autoscaler/pull/4634.

Currently on e.g a 48 minutes cluster with no scaling activity it does 1.4k get request to the scale subresource.
This PR tries to satisfy both non stale replicas during scale down and prevent the API server from being overloaded. To achieve that it lets TargetSize which is called on every autoscaling cluster state loop from come from cache while getting fresh replicas at the time of perform the operations.

Also note that the scale down implementation has changed https://github.com/kubernetes/autoscaler/commits/master/cluster-autoscaler/core/scaledown.

/area provider/cluster-api

/kind bug
/kind cleanup

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_nodegroup.go
+++ b/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_nodegroup.go
@@ -18,6 +18,7 @@ package clusterapi
 
 import (
 	"fmt"
+	"github.com/pkg/errors"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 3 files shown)


elmiko (commented):
this makes sense to me, thanks for the update @enxebre 

i just have a quick question about

> Also note that the scale down implementation has changed https://github.com/kubernetes/autoscaler/commits/master/cluster-autoscaler/core/scaledown.

did you have something specific in mind?

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_nodegroup.go
```diff
@@ -52,7 +53,14 @@ func (ng *nodegroup) MaxSize() int {
 // (new nodes finish startup and registration or removed nodes are
 // deleted completely). Implementation required.
 func (ng *nodegroup) TargetSize() (int, error) {
-	return ng.scalableResource.Replicas()
+	replicas, found, err := unstructured.NestedInt64(ng.scalableResource.unstructured.Object, "spec", "replicas")
+	if err != nil {
+		return 0, errors.Wrap(err, "error getting replica count")
+	}
+	if !found {
+		replicas = 0
```
arunmk, line 61: should we set `replicas = ng.scalableResource.MinSize()` here?

arunmk (approved):
LGTM with a minor comment

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_nodegroup.go
elmiko: that's probably better

elmiko:
LGTM, i'll let @arunmk add the label

PR #5036, opened 2022-07-20
use strings.Join to build list of names

grosser:
random little cleanup to make the code more readable / ideomatic
... what we are doing is joining string together, so use `strings.Join` instead of buffer math


mwielgus:
It seems that there are compilation problems there. Please fix.

grosser:
thx, fixed!

mwielgus (approved):
/lgtm
/approve

PR #5047, opened 2022-07-25
Azure: effectively cache instance-types SKUs

bpineau:
The skewer library's cache is re-created at every call, which causes pressure on Azure API, and slows down the cluster-autoscaler startup time by two minutes on a relatively small test cluster (120 nodes, 300 VMSS).

This was also hitting the API twice on cache miss to look for non-promo instance types, even when the instance name doesn't ends with "_Promo".

SKU API call rate before/after: 
image


feiskyer (approved):
/lgtm

Issue #5022, opened 2022-07-13
Not autoscaled node groups are treated as deleted


x13n:
**Which component are you using?**:

cluster-autoscaler

**What version of the component are you using?**:

Component version: running from current HEAD

**What k8s version are you using (`kubectl version`)?**: 1.24

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version

</pre></details>

**What environment is this in?**: GKE

**What did you expect to happen?**: CA should treat upcoming nodes in non-autoscaled node groups as upcoming, not deleted, which would prevent unnecessary scale up.

**What happened instead?**: CA triggers a scale up even though there are nodes in non-autoscaled node groups that could run the pods.

**How to reproduce it (as minimally and precisely as possible)**: In GKE, create a cluster with default nodepool that is not autoscaled and enable NAP. Observe NAP create a new nodepool. Sometimes it won't happen, if scheduler manages to schedule all the pods before CA kicks in.

**Anything else we need to know?**: The way deleted nodes are detected changed in https://github.com/kubernetes/autoscaler/pull/4896, we should probably roll it back and figure this problem out before reapplying the change.

/cc @fookenc @MaciekPytel

x13n:
Chatted with @MaciekPytel about this. It should be sufficient to detect not autoscaled node groups (and stop marking them as deleted) in the same way scale down is doing it: by checking if `NodeGroupForNode(node)` is nil: https://github.com/kubernetes/autoscaler/blob/5745044ddf4378c1aae16c47fb173b1abf709c8e/cluster-autoscaler/core/scaledown/legacy/legacy.go#L313-L322

Perhaps worth wrapping this check into a function and using it in both places.

fookenc:
Hi @x13n & @MaciekPytel,

From my local testing, I'm not sure that NodeGroupForNode will solve the issue.  I found that after a node is deleted from the cloud provider, the NodeGroupForNode also returns nil.  Unfortunately, this makes deleted nodes and not autoscaled nodes appear the same.

Is there another way to determine not autoscaled nodes?

fookenc:
I've submitted a new PR #5054 to reintroduce the code changes that were reverted and address the issue detailed here.  The changes also include new scenarios in the test case for this functionality.  It includes not autoscaled nodes (nodes without a node group) to ensure that they are no longer incorrectly flagged as deleted.

Please review the new changes and notify if there are areas of concern or improvement.

x13n:
[...]

Issue #4456, opened 2021-11-10
CA fails to scale-up or cancel in-progress scale down when there are un-schedulable pods 


varkey:
**Which component are you using?**: `cluster-autoscaler`

**What version of the component are you using?**: `cluster-autoscaler`

Component version: `k8s.gcr.io/autoscaling/cluster-autoscaler:v1.20.0`

**What k8s version are you using (`kubectl version`)?**: `v1.20.10`

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.10", GitCommit:"8152330a2b6ca3621196e62966ef761b8f5a61bb", GitTreeState:"clean", BuildDate:"2021-08-11T18:00:37Z", GoVersion:"go1.15.15", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What environment is this in?**:

KOPS (v1.20.2) based cluster deployed on AWS

CA is run with the following arguments

```
- ./cluster-autoscaler
- --v=2
- --stderrthreshold=info
- --cloud-provider=aws
- --skip-nodes-with-local-storage=false
- --skip-nodes-with-system-pods=false
- --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,k8s.io/cluster-autoscaler/env-name
- --scale-down-utilization-threshold=0.6
- --balance-similar-node-groups
- --ignore-taint=node-upgrading
- --max-graceful-termination-sec=10800
- --scale-down-delay-after-delete=5m
- --max-empty-bulk-delete=2
- --scale-down-delay-after-add=5m
```

The nodes span multiple AZs, but we use separate ASGs for each AZ. Also there are several node groups with different taints so that we can place applications in dedicated node groups. 
```
node-group-1-az1
node-group-1-az2
node-group-1-az3
node-group-2-az1
node-group-2-az2
node-group-2-az3
...
...
...
```

In the above case, all `node-group-1-xx` node groups have the same taints, instance types etc. 

We use HPA for many of our applications, and can dynamically scale up and down the number of replicas. As can be seen below, the scale up happen every hour and then the load subsides, and the HPA scales down the number of replicas. 

<img width="1611" alt="Screenshot 2021-11-10 at 12 16 38 PM" src="https://user-images.githubusercontent.com/664529/141063657-a80fa7ef-0eb5-46f2-834b-989699cc2642.png">

[...]

PR #5054, opened 2022-07-27
Identifying cloud provider deleted nodes

### LINKED ISSUE #3949: Node groups get "stuck" during node deletion

fookenc:
cluster-autoscaler

#### What type of PR is this?
/kind bug

Attempts to address the issue raised in #5022 from previously approved PR.  Original implementation (previous work was reverted #4896 & #5023) included that changes how Deleted nodes are determined in the cluster state.  New iteration utilizes the previously cached Cloud Provider nodes in the ClusterStateRegistry to perform a diff.  Cloud Provider node instances are retrieved by Node Group, which should avoid not autoscaled nodes from being flagged erroneously.  Test case has been modified to include this scenario.  An additional scenario has been included to ensure that previously identified Cloud Provider nodes will still be tracked until they are no longer registered in Kubernetes.

This PR includes the initial work first introduced in https://github.com/kubernetes/autoscaler/pull/4211. Feedback from that PR indicated that the original intent of determining the deleted nodes was incorrect, which led to the issues reported by other users. The nodes tainted with ToBeDeleted were misidentified as Deleted instead of Ready/Unready, which caused a miscalculation of the node being included as Upcoming. This caused problems described in https://github.com/kubernetes/autoscaler/issues/3949 and https://github.com/kubernetes/autoscaler/issues/4456.

Fixes #3949, #4456, #5022

NONE

### DIFF
--- a/cluster-autoscaler/cloudprovider/cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/cloud_provider.go
@@ -100,6 +100,10 @@ type CloudProvider interface {
 	// occurred. Must be implemented.
 	NodeGroupForNode(*apiv1.Node) (NodeGroup, error)
 
+	// NodeExists returns whether the node exists in cloud provider,
+	// true if the node is available, false if it has been deleted
+	NodeExists(*apiv1.Node) (bool, error)
+
 	// Pricing returns pricing model for this cloud provider or error if not available.
 	// Implementation optional.
 	Pricing() (PricingModel, errors.AutoscalerError)
--- a/cluster-autoscaler/cloudprovider/cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/cloud_provider.go
@@ -100,9 +100,9 @@ type CloudProvider interface {
 	// occurred. Must be implemented.
 	NodeGroupForNode(*apiv1.Node) (NodeGroup, error)
 
-	// NodeExists returns whether the node exists in cloud provider,
-	// true if the node is available, false if it has been deleted
-	NodeExists(*apiv1.Node) (bool, error)
+	// HasInstance returns whether the node has corresponding instance in cloud provider,
+	// true if the node has an instance, false if it no longer exists
+	HasInstance(*apiv1.Node) (bool, error)
 
 	// Pricing returns pricing model for this cloud provider or error if not available.
 	// Implementation optional.
--- a/cluster-autoscaler/cloudprovider/test/test_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/test/test_cloud_provider.go
@@ -253,6 +253,14 @@ func (tcp *TestCloudProvider) AddNode(nodeGroupId string, node *apiv1.Node) {
 	tcp.nodes[node.Name] = nodeGroupId
 }
 
+// DeleteNode delete the given node from the provider.
+func (tcp *TestCloudProvider) DeleteNode(node *apiv1.Node) {
+	tcp.Lock()
+	defer tcp.Unlock()
+
+	delete(tcp.nodes, node.Name)
+}
+
 // GetResourceLimiter returns struct containing limits (max, min) for resources (cores, memory etc.).
 func (tcp *TestCloudProvider) GetResourceLimiter() (*cloudprovider.ResourceLimiter, error) {
 	return tcp.resourceLimiter, nil
--- a/cluster-autoscaler/cloudprovider/test/test_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/test/test_cloud_provider.go
@@ -41,8 +41,8 @@ type OnNodeGroupCreateFunc func(string) error
 // OnNodeGroupDeleteFunc is a function called when a node group is deleted.
 type OnNodeGroupDeleteFunc func(string) error
 
-// NodeExists is a function called to determine if a node has been removed from the cloud provider.
-type NodeExists func(string) (bool, error)
+// HasInstance is a function called to determine if a node has been removed from the cloud provider.
+type HasInstance func(string) (bool, error)
 
 // TestCloudProvider is a dummy cloud provider to be used in tests.
 type TestCloudProvider struct {
@@ -53,7 +53,7 @@ type TestCloudProvider struct {
 	onScaleDown       func(string, string) error
 	onNodeGroupCreate func(string) error
 	onNodeGroupDelete func(string) error
-	nodeExists        func(string) (bool, error)
+	hasInstance       func(string) (bool, error)
 	machineTypes      []string
 	machineTemplates  map[string]*schedulerframework.NodeInfo
 	priceModel        cloudprovider.PricingModel
@@ -90,13 +90,13 @@ func NewTestAutoprovisioningCloudProvider(onScaleUp OnScaleUpFunc, onScaleDown O
 
 // NewTestNodeDeletionDetectionCloudProvider builds new TestCloudProvider with deletion detection support
 func NewTestNodeDeletionDetectionCloudProvider(onScaleUp OnScaleUpFunc, onScaleDown OnScaleDownFunc,
-	deleted NodeExists) *TestCloudProvider {
+	hasInstance HasInstance) *TestCloudProvider {
 	return &TestCloudProvider{
 		nodes:           make(map[string]string),
 		groups:          make(map[string]cloudprovider.NodeGroup),
 		onScaleUp:       onScaleUp,
 		onScaleDown:     onScaleDown,
-		nodeExists:      deleted,
+		hasInstance:     hasInstance,
 		resourceLimiter: cloudprovider.NewResourceLimiter(make(map[string]int64), make(map[string]int64)),
 	}
 }
@@ -157,14 +157,14 @@ func (tcp *TestCloudProvider) NodeGroupForNode(node *apiv1.Node) (cloudprovider.
 	return group, nil
 }
 
-// NodeExists returns true if the node is available in cloud provider,
+// HasInstance returns true if the node has corresponding instance in cloud provider,
 // or ErrNotImplemented to fall back to taint-based node deletion in clusterstate
 // readiness calculation.
-func (tcp *TestCloudProvider) NodeExists(node *apiv1.Node) (bool, error) {
+func (tcp *TestCloudProvider) HasInstance(node *apiv1.Node) (bool, error) {
 	tcp.Lock()
 	defer tcp.Unlock()
-	if tcp.nodeExists != nil {
-		return tcp.nodeExists(node.Name)
+	if tcp.hasInstance != nil {
+		return tcp.hasInstance(node.Name)
 	}
 	_, found := tcp.nodes[node.Name]
 	return found, nil
--- a/cluster-autoscaler/cloudprovider/test/test_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/test/test_cloud_provider.go
@@ -41,6 +41,9 @@ type OnNodeGroupCreateFunc func(string) error
 // OnNodeGroupDeleteFunc is a function called when a node group is deleted.
 type OnNodeGroupDeleteFunc func(string) error
 
+// NodeExists is a function called to determine if a node has been removed from the cloud provider.
+type NodeExists func(string) (bool, error)
+
 // TestCloudProvider is a dummy cloud provider to be used in tests.
 type TestCloudProvider struct {
 	sync.Mutex
@@ -50,6 +53,7 @@ type TestCloudProvider struct {
 	onScaleDown       func(string, string) error
 	onNodeGroupCreate func(string) error
 	onNodeGroupDelete func(string) error
+	nodeExists        func(string) (bool, error)
 	machineTypes      []string
 	machineTemplates  map[string]*schedulerframework.NodeInfo
 	priceModel        cloudprovider.PricingModel
@@ -81,6 +85,19 @@ func NewTestAutoprovisioningCloudProvider(onScaleUp OnScaleUpFunc, onScaleDown O
 		machineTypes:      machineTypes,
 		machineTemplates:  machineTemplates,
 		resourceLimiter:   cloudprovider.NewResourceLimiter(make(map[string]int64), make(map[string]int64)),
+	}
+}
+
+// NewTestNodeDeletionDetectionCloudProvider builds new TestCloudProvider with deletion detection support
+func NewTestNodeDeletionDetectionCloudProvider(onScaleUp OnScaleUpFunc, onScaleDown OnScaleDownFunc,
+	deleted NodeExists) *TestCloudProvider {
+	return &TestCloudProvider{
+		nodes:           make(map[string]string),
+		groups:          make(map[string]cloudprovider.NodeGroup),
+		onScaleUp:       onScaleUp,
+		onScaleDown:     onScaleDown,
+		nodeExists:      deleted,
+		resourceLimiter: cloudprovider.NewResourceLimiter(make(map[string]int64), make(map[string]int64)),
 	}
 }
 
@@ -138,6 +155,19 @@ func (tcp *TestCloudProvider) NodeGroupForNode(node *apiv1.Node) (cloudprovider.
 		return nil, nil
 	}
 	return group, nil
+}
+
+// NodeExists returns true if the node is available in cloud provider,
+// or ErrNotImplemented to fall back to taint-based node deletion in clusterstate
+// readiness calculation.
+func (tcp *TestCloudProvider) NodeExists(node *apiv1.Node) (bool, error) {
+	tcp.Lock()
+	defer tcp.Unlock()
+	if tcp.nodeExists != nil {
+		return tcp.nodeExists(node.Name)
+	}
+	_, found := tcp.nodes[node.Name]
+	return found, nil
 }
 
 // Pricing returns pricing model for this cloud provider or error if not available.
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -17,6 +17,7 @@ limitations under the License.
 package clusterstate
 
 import (
+	"errors"
 	"fmt"
 	"reflect"
 	"strings"
@@ -297,6 +298,7 @@ func (csr *ClusterStateRegistry) UpdateNodes(nodes []*apiv1.Node, nodeInfosForGr
 	if err != nil {
 		return err
 	}
+	cloudProviderNodesRemoved := csr.getCloudProviderDeletedNodes(nodes, cloudProviderNodeInstances)
 	notRegistered := getNotRegisteredNodes(nodes, cloudProviderNodeInstances, currentTime)
 
 	csr.Lock()
@@ -544,7 +546,7 @@ func (csr *ClusterStateRegistry) updateReadinessStats(currentTime time.Time) {
 
 	update := func(current Readiness, node *apiv1.Node, nr kube_util.NodeReadiness) Readiness {
 		current.Registered++
-		if _, exists := csr.deletedNodes[node.Name]; exists {
+		if _, isDeleted := csr.deletedNodes[node.Name]; isDeleted {
 			current.Deleted++
 		} else if nr.Ready {
 			current.Ready++
@@ -684,7 +686,7 @@ func (csr *ClusterStateRegistry) updateCloudProviderDeletedNodes(deletedNodes []
 	csr.deletedNodes = result
 }
 
-//GetCloudProviderDeletedNodes returns a list of all nodes removed from cloud provider but registered in Kubernetes.
+// GetCloudProviderDeletedNodes returns a list of all nodes removed from cloud provider but registered in Kubernetes.
 func (csr *ClusterStateRegistry) GetCloudProviderDeletedNodes() []*apiv1.Node {
 	csr.Lock()
 	defer csr.Unlock()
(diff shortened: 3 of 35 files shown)


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -958,6 +984,44 @@ func getNotRegisteredNodes(allNodes []*apiv1.Node, cloudProviderNodeInstances ma
...
+	for nodeName, node := range csr.deletedNodes {
+		if _, found := registeredNodes[nodeName]; found {
+			nodesRemoved = append(nodesRemoved, node)
+		}
+	}
+
+	// Seek nodes that may have been deleted since last update
+	// cloudProviderNodeInstances are retrieved by nodeGroup,
+	// not autoscaled nodes will not be included
+	for _, instances := range csr.cloudProviderNodeInstances {
```
x13n, line 1011: Won't relying on previous state cause issues in case of CA restart? IIUC some nodes will be incorrectly considered as `Ready`/`Unready`/`NotStarted`, potentially leading to bad scaling decisions.

fookenc: That would definitely occur on a restart.  Would it make sense to create a new taint and add it onto the nodes?  There might be some additional overhead for the implementation, but should prevent that scenario.

I also thought about adding a check for each node in the `deletedNodes` (lines 1001 - 1006) to confirm it doesn't appear in the cloud provider nodes.  I don't have good insight to know if that is a case that could happen though.  I wouldn't want to keep flagging a node as deleted if the cloud provider node exists.

fookenc: That makes sense to me.  I agree that the API should eventually correct itself, and the scaling should only momentarily be impacted.  I've created a new commit that includes the backfill safety check mentioned and some other minor cosmetic code changes.

x13n (commented):
Thanks for the changes!

I just realized one more problem with this approach though. Since we cannot tell the difference between nodes that were deleted and the ones that are not autoscaled, this implementation is going to treat nodes as deleted whenever someone opts out a node group from autoscaling. As a result, CA may trigger large scale ups without a good reason. One way to handle this would be to add a TTL to `deletedNodes` to avoid keeping nodes there forever in this scenario. (Perhaps an even better option would be some extension of cloud provider API to distinguish deleted from not autoscaled, but I'm not sure that's feasible.) WDYT?

fookenc:
That is definitely a concern.  I can add a TTL to the deletedNodes entries.  Should I add a configurable value for better control or just a common default value?

I had a similar thought about extending the cloud provider implementation.  After a brief investigation, it looks like most of the interactions are handled by using information in the node (the `node.Spec.ProviderID` specifically).  It doesn't appear to validate whether the node exists in the cloud.  A possible change could be to provide more context of deleted nodes versus not-autoscaled nodes within the `NodeGroupForNode` function.

x13n:
I think extending the API would be much cleaner, but the need to implement it for all cloud providers calls for a broader discussion. I added this topic to SIG meeting agenda to discuss it on Monday.

x13n:
So I guess the conclusion after yesterday's SIG meeting is that we should extend cloud provider interface and preserve the existing (taint-based) behavior in case a specific cloud provider doesn't implement the new function. That way each cloud provider will be able to fix the bug by implementing a function distinguishing deleted from non-autoscaled nodes.

fookenc:
Should this PR be closed until the cloud provider interface can be extended?  What would be the next steps on fixing the bug in the future?  Please let me know if there's anything I can offer for assistance.  Thanks!

x13n:
The cloud provider interface can be extended right now, it's just that all implementations would have to get default implementation that returns NotImplemented error. I thought this PR could be re-purposed into extending the interface and using it to detect deleted nodes. Each cloud provider would then need to implement their part in a separate PR. Are you willing to make that first, cloud provider agnostic, change?

fookenc:
Got it!  I'll start working on a pass at the implementation and submit it as soon as I can.  Thanks again!

x13n:
If this requires more time, maybe it would make sense to submit change to taint cleanup (ready -> all nodes) on startup separately first?

fookenc:
Sorry for the delay here.  I've split the taint removal into another PR #5200, and will continue working on the effort here.  Please let me know if there's anything else needed.  Thanks!

x13n:
Thanks! The other PR already got merged.

fookenc:
I've added some new changes to resolve the issues raised.  An initial implementation for the NodeExists method has been introduced to each cloud provider.  Currently it returns an ErrNotImplemented error to signal the cluster state to fall back to taint based node deletion logic.  The TestCloudProvider has been updated to utilize the new node deletion detection logic.  A new test has been added to verify that the existing taint-based node deletion works as expected when the NodeExists function has not been implemented (i.e. returns the ErrNotImplemented error).

Please review when you have a moment, and let me know if there are any questions or concerns.  Thanks!

cc: @x13n

x13n (changes requested):

x13n:
Thanks for the changes! My main concern is whether simplifying the logic is feasible - I think some of the previous approach got carried over, but isn't really necessary with the new interface.

cluster-autoscaler/cloudprovider/test/test_cloud_provider.go
```diff
@@ -84,6 +88,19 @@ func NewTestAutoprovisioningCloudProvider(onScaleUp OnScaleUpFunc, onScaleDown O
 	}
 }
 
+// NewTestNodeDeletionDetectionCloudProvider builds new TestCloudProvider with deletion detection support
+func NewTestNodeDeletionDetectionCloudProvider(onScaleUp OnScaleUpFunc, onScaleDown OnScaleDownFunc,
+	deleted NodeExists) *TestCloudProvider {
```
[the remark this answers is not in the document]
fookenc: Sorry! This was a missed replacement. An earlier implementation had the function named `IsNodeDeleted`. I'll update this to better reflect the new function name.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -669,6 +674,30 @@ func (csr *ClusterStateRegistry) GetUnregisteredNodes() []UnregisteredNode {
...
+		if prev, found := csr.deletedNodes[deleted.Name]; found {
+			result[deleted.Name] = prev
+		} else {
+			result[deleted.Name] = deleted
+		}
+	}
+	csr.deletedNodes = result
+}
+
+// GetCloudProviderDeletedNodes returns a list of all nodes removed from cloud provider but registered in Kubernetes.
```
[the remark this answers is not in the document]
fookenc: I think this is a carryover from a similar use-case.  I'll remove this function since it isn't used.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -295,6 +298,7 @@ func (csr *ClusterStateRegistry) UpdateNodes(nodes []*apiv1.Node, nodeInfosForGr
 	if err != nil {
 		return err
 	}
+	cloudProviderNodesRemoved := csr.getCloudProviderDeletedNodes(nodes, cloudProviderNodeInstances)
```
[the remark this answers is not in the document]
fookenc: This was also attempting to match with an existing interface design, but could be consolidated for simplicity.  The other consideration used to be to avoid locking the ClusterStateRegistry for too long, which might be the reason for the separation of a discovery phase and updating the cluster state.  It shouldn't be an issue to combine these for this use case.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -958,6 +987,71 @@ func getNotRegisteredNodes(allNodes []*apiv1.Node, cloudProviderNodeInstances ma
...
+			for nodeGroupName, instances := range cloudProviderNodeInstances {
+				for _, instance := range instances {
+					currentCloudInstances[instance.Id] = nodeGroupName
+				}
+			}
+			for _, node := range allNodes {
+				registeredNodes[node.Name] = node
+			}
+
+			// Fill previously deleted nodes, if they are still registered in Kubernetes
```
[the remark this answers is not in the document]
fookenc: One concern I have with the new interface using it to discover deleted nodes.  At the moment, I don't have a clear understanding of the level of effort involved with verifying the instance's existence with each cloud provider.  Since the ClusterStateRegistry stores the previousCloudProviderNodeInstances and cloudProviderNodeInstances, I would like to leverage them to identify nodes that may be removed/terminated on the cloud provider.  Afterwards, the new interface method can be used to verify with the cloud provider that the node exists or not.  This will prevent marking a node as deleted if it was not-autoscaled or removed from a node group.

Unfortunately, I'm not sure how to support a case when a cloud provider might only implement the new interface for specific nodes.  That would require that the check is performed on each node to ensure taint-based deletion is checked for each node.

Are there concerns about looping over `allNodes` and checking each with the new interface method?

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -958,6 +987,71 @@ func getNotRegisteredNodes(allNodes []*apiv1.Node, cloudProviderNodeInstances ma
 	return notRegistered
 }
 
+// Calculates which of the registered nodes in Kubernetes that do not exist in cloud provider.
+func (csr *ClusterStateRegistry) getCloudProviderDeletedNodes(allNodes []*apiv1.Node, cloudProviderNodeInstances map[string][]cloudprovider.Instance) []*apiv1.Node {
+	nodesRemoved := make([]*apiv1.Node, 0)
+	currentCloudInstances := make(map[string]string, 0)
+	registeredNodes := make(map[string]*apiv1.Node, 0)
+	if len(allNodes) > 0 {
+		_, err := csr.cloudProvider.NodeExists(allNodes[0])
```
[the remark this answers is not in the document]
fookenc: That could be the case for some cloud providers, and may need to be addressed.  I have concerns about the runtime efficiency of the detection of deleted nodes by checking each node's status on each run.  Please see my comment below for more details.  If it isn't too large of a concern or polling the cloud provider is a lower cost than I'm expecting, then I absolutely agree the new interface method could be used as you've indicated.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -669,6 +674,30 @@ func (csr *ClusterStateRegistry) GetUnregisteredNodes() []UnregisteredNode {
 	return result
 }
 
+func (csr *ClusterStateRegistry) updateCloudProviderDeletedNodes(deletedNodes []*apiv1.Node) {
+	result := make(map[string]*apiv1.Node)
```
x13n, line 678: nit: maybe set map capacity to `len(deletedNodes)+len(csr.deletedNodes)` to avoid multiple allocs?

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -958,6 +987,71 @@ func getNotRegisteredNodes(allNodes []*apiv1.Node, cloudProviderNodeInstances ma
...
+					nodeExists, existsErr := csr.cloudProvider.NodeExists(node)
+					if existsErr == nil && !nodeExists {
+						nodesRemoved = append(nodesRemoved, node)
+					}
+				}
+			}
+
+			// Seek nodes that may have been deleted since last update
+			// cloudProviderNodeInstances are retrieved by nodeGroup,
+			// not autoscaled nodes will be excluded
```
x13n, line 1033: How can we detect not autoscaled nodes that got deleted?

x13n (changes requested):

cluster-autoscaler/cloudprovider/cloud_provider.go
```diff
@@ -100,6 +100,10 @@ type CloudProvider interface {
 	// occurred. Must be implemented.
 	NodeGroupForNode(*apiv1.Node) (NodeGroup, error)
 
+	// NodeExists returns whether the node exists in cloud provider,
+	// true if the node is available, false if it has been deleted
+	NodeExists(*apiv1.Node) (bool, error)
```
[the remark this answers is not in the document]
fookenc: I wasn't sure about the wording myself, but `HasInstance` sounds great to me.  I'll adjust the naming with my next update.

cluster-autoscaler/clusterstate/clusterstate.go
fookenc: Agreed, I will include this in my next changes.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -958,6 +987,71 @@ func getNotRegisteredNodes(allNodes []*apiv1.Node, cloudProviderNodeInstances ma
 	return notRegistered
 }
 
+// Calculates which of the registered nodes in Kubernetes that do not exist in cloud provider.
+func (csr *ClusterStateRegistry) getCloudProviderDeletedNodes(allNodes []*apiv1.Node, cloudProviderNodeInstances map[string][]cloudprovider.Instance) []*apiv1.Node {
+	nodesRemoved := make([]*apiv1.Node, 0)
+	currentCloudInstances := make(map[string]string, 0)
```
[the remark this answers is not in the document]
fookenc: Similar to another request earlier, I think it would be better to use the lengths that you've suggested.  I will include this in my next change.

cluster-autoscaler/clusterstate/clusterstate.go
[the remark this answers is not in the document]
fookenc: Excellent!  I will get this added in my changes.

cluster-autoscaler/clusterstate/clusterstate.go
fookenc: I think the comment is confusing here, and apologize for that.  The `NodeGroupForNode` could be used to check whether the node is not-autoscaled.  Though I'm not sure that it is required to know if the node is not-autoscaled or not.  Should all nodes that no longer have a cloud instance be included in the deletion detection?

cluster-autoscaler/clusterstate/clusterstate.go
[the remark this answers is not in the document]
fookenc: Unfortunately, I will need to remove this logic since it can't support both the taint-based deletion and new method simultaneously.  The trouble is that the new technique will discover potential nodes that no longer exist between the two caches (`cloudProviderInstances` recently retrieved and the cached data in the clusterStateRegistry).  This would make it tough to also identify the nodes that have the taint, because those instances would not be otherwise known.

As you mentioned, it will be up to each cloud provider to implement an efficient implementation of this `HasInstance` method.  I will modify the behavior here to iterate over the list of `allNodes` to determine whether the cloud instance exists.

cluster-autoscaler/clusterstate/clusterstate.go
x13n: Hm... actually considering only autoscaled nodes here is fine, I think counting upcoming nodes will work properly for autoscaled nodegroups only anyway.

fookenc: I agree, it should not impact the count since the readiness calculation considers only those nodes in autoscaled nodegroups.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -988,70 +988,28 @@ func getNotRegisteredNodes(allNodes []*apiv1.Node, cloudProviderNodeInstances ma
...
 }
 
+func (csr *ClusterStateRegistry) hasCloudProviderInstance(node *apiv1.Node) bool {
+	exists, err := csr.cloudProvider.HasInstance(node)
+	if err == nil {
+		return exists
+	}
+	if !errors.Is(err, cloudprovider.ErrNotImplemented) {
+		klog.Warningf("Failed to check whether node has cloud instance for %s: %v", node.Name, err)
+		return exists
```
x13n, line 1008: I don't think depending on the value of `exists` makes sense here. If the function returned an error, this is likely going to be some const. Maybe just fall through here and do the taint check instead?

cluster-autoscaler/clusterstate/clusterstate.go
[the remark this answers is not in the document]
x13n: Thanks. I see you added a test helper, but this one is still here.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -669,6 +674,30 @@ func (csr *ClusterStateRegistry) GetUnregisteredNodes() []UnregisteredNode {
 	return result
 }
 
+func (csr *ClusterStateRegistry) updateCloudProviderDeletedNodes(deletedNodes []*apiv1.Node) {
+	result := make(map[string]*apiv1.Node, len(deletedNodes)+len(csr.deletedNodes))
+	for _, deleted := range deletedNodes {
+		if prev, found := csr.deletedNodes[deleted.Name]; found {
+			result[deleted.Name] = prev
```
x13n, line 681: Why does this carry over stale `Node` objects from previous iterations? Come to think of it, do we actually need to store anything beyond deleted node names? (We do for taint-based check, but this code is invoked later.)

cluster-autoscaler/clusterstate/clusterstate.go
fookenc: I think falling back to the taint check would be against expectations.  I believe if the cloud provider only implements the check for certain nodes, then falling back to taint check would make sense.  However, if there is an error that is something other than `ErrNotImplemented`, then I believe the expectation would be to return the default value of `false` and log the error.  If taint based detection is used as a fallback, then it could cause conflicting issues in the event that errors are transient.  I feel that the goal is to migrate the cloud providers to the new detection methodology, and the taint-based version is there for backwards compatibility.

Please let me know your thoughts.

cluster-autoscaler/clusterstate/clusterstate.go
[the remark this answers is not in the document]
fookenc: Got it!  Total mistake on my part.  This will be removed on my next update.

cluster-autoscaler/clusterstate/clusterstate.go
fookenc: The logic here is mostly copied from a similar implementation.  It isn't necessary to store more than the node name.  I can update everything to use a map of the node name.  Is there a value that makes sense to store?  My initial instinct is to store an epoch for the value of the map.

cluster-autoscaler/clusterstate/clusterstate.go
x13n: I think if cloudprovider isn't able to tell (for whatever reason: lack of implementation or some runtime error) whether the instance exists, clusterstate can fall back to the next best method of checking this, which is checking taints. In what scenario do you think it would be better to blindly use `false` instead of inspecting the taint?

cluster-autoscaler/clusterstate/clusterstate.go
x13n: I'd have a value only if it is going to be used for something. Otherwise `struct{}` or `bool` should suffice.

fookenc: I will update the usage to match that suggestion of using the empty struct.

cluster-autoscaler/clusterstate/clusterstate.go
fookenc: After thinking it over for a bit, I believe you're correct.  Initially I wanted to avoid confusing the check between the new method and old taint-based version.  However, referring back to an earlier point you raised about allowing cloud providers to implement the new check selectively by node/instance type/etc., it makes sense to fallback to the taint-based when the other option fails.  My previous understanding was that the taint-based check would be replaced entirely, but that might not work for all cloud providers.  I'll work on an update that includes falling back to the taint-based check.

x13n:
Thanks for following up on this!

/lgtm

x13n:
Hm, looks like I should be able to approve this now as well, let's see:

/approve

ohnoitsjmo:
@fookenc Will this release be backported to previous CA versions (including 1.22)?

fookenc:
Unfortunately, this release alone won't fix issues until the cloud providers have implemented the new interface method.  Once the cloud providers have been implemented it could be backported, but I don't know how soon that will occur.  @x13n do you know if this could be backported to earlier CA versions?

x13n:
1.22 is past k8s support window at this point, so unless there's a critical bug to fix I don't think any backport is going to be released.

PR #5056, opened 2022-07-28
update autoscaler-autodiscover.yaml file

parulth91:

Fixes #

```release-note

```

```docs

```


gjtempleton:
Hi @parulth91, is this PR deliberate?

The example files are intended to be generic and modified by users as needed, and hardcoding a cluster name isn't going to help that. If we need to make the docs clearer that this needs to be modified, happy to discuss that.

mwielgus:
We should not put project/cluster names into examples. If someone doesn't change the name, they might deploy/take information from some unwanted, potentially rogue projects.

PR #5057, opened 2022-07-28
Add a new error code to GCE autoscaling client

NastassiaM:

cluster-autoscaler

#### What type of PR is this?

/kind feature

Adds a new error code `VM_EXTERNAL_IP_ACCESS_POLICY_CONSTRAINT` to GCE autoscaling client to provide better visibility into errors caused by the org policy misconfiguration.

N/A

-

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client_test.go
+++ b/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client_test.go
@@ -143,6 +143,7 @@ func TestErrors(t *testing.T) {
 
 	testCases := []struct {
 		errorCodes         []string
+		errorMessage       string
 		expectedErrorCode  string
 		expectedErrorClass cloudprovider.InstanceErrorClass
 	}{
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client_test.go
```diff
@@ -203,6 +203,55 @@ func TestErrors(t *testing.T) {
 	mock.AssertExpectationsForObjects(t, server)
 }
 
+func TestMessageBasedErrors(t *testing.T) {
```
MaciekPytel, line 206: That feels like it's duplicating a lot of TestErrors above. Wouldn't it be better to just errorMessage parameter to testCases config in TestErrors (and leave it empty for all other tests)?

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client_test.go
```diff
@@ -143,31 +143,43 @@ func TestErrors(t *testing.T) {
 
 	testCases := []struct {
 		errorCodes         []string
+		errorMessage       string
 		expectedErrorCode  string
 		expectedErrorClass cloudprovider.InstanceErrorClass
 	}{
 		{
 			errorCodes:         []string{"IP_SPACE_EXHAUSTED"},
+			errorMessage:       "",
```
MaciekPytel, line 152: nit: empty string is the zero value for string, so it will be the same if you don't set it. I think the convention is to skip irrelevant parameters in test cases (it's pretty normal for some parameters to only be relevant for some cases).

MaciekPytel:
/approve
/lgtm
/hold 
Hold for a minor comment, not sure if you want to address it. Feel free to just /hold cancel to merge the PR.

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client_test.go
NastassiaM: Updated.

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client_test.go
NastassiaM: Updated.

PR #5058, opened 2022-07-28
fix FAQ entry for cores-total flag

elmiko:

 cluster-autoscaler

#### What type of PR is this?

/kind cleanup
/kind documentation

this change adds proper escaping to the min/max tags for the cores-total
flag entry in the FAQ, which makes it so that the markdown render will
print the proper formating for this flag.


feiskyer (approved):
/lgtm

PR #5060, opened 2022-07-28
Introduce NodeDeleterBatcher to ScaleDown actuator

yaroslava-serdiuk:

cluster-autoscaler

#### What type of PR is this?

/kind feature

Optimize Scale down by deleting nodes in batch. The --node-deletion-in-batch-interval represent the time that NodeDeleterBatcher will gather nodes for one node group and after that time will call nodeGroup.DeleteNodes(nodes). If the flag is unset, the NodeDeleterBatcher won't wait for other nodes and immediately start node deletion.

```release-note
ScaleDown in batch: --node-deletion-in-batch-interval flag represent how long CA ScaleDown gather nodes to delete them in batch.
```

```docs

```


cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -251,41 +246,50 @@ func (a *Actuator) taintNode(node *apiv1.Node) error {
...
 	}
+	return status.NodeDeleteResult{ResultType: status.NodeDeleteInPogress}
+}
 
-	if err := DeleteNodeFromCloudProvider(a.ctx, node, a.clusterState); err != nil {
-		return status.NodeDeleteResult{ResultType: status.NodeDeleteErrorFailedToDelete, Err: err}
+// scheduleDeletion schedule the deletion on of the provided node by adding a node to NodeDeletionBatcher. If drain is true, the node is drained before being deleted.
+func (a *Actuator) scheduleDeletion(node *apiv1.Node, drain bool) {
+	status := a.cleanUpNode(node, drain)
+	if status.Err != nil {
```
towca, line 273: The exact same error handling repeats in `delete_in_batch.go`, but split into two functions (`process_*`). First off I think this version of having 1 function with `if drain` inside is more readable than having 2 functions. But more importantly, since the 2 places are supposed to do the same thing - can we extract this functionality somehow and call it from both places? Then we're sure the handling actually is the same in both places, and there's less code duplication.

cluster-autoscaler/processors/status/scale_down_status_processor.go
```diff
@@ -119,6 +119,8 @@ const (
 	NodeDeleteErrorFailedToDelete
 	// NodeDeleteErrorInternal - failed to delete the node because of an unexpected error.
 	NodeDeleteErrorInternal
+	// NodeDeleteInPogress - node deletion is currently in progress.
```
towca, line 122: Hmm not sure if we should introduce a new type result for in progress - these values are supposed to be consumed by the scale down status processor, but only for nodes that are actually deleted successfully, or if the deletion fails. Introducing an "in progress" result could be confusing for processor implementations - how should they handle it, will the node be included in another event with a different result? I know it won't happen with the current Actuator implementation, but processor implementations shouldn't have to care about the implementation details of Actuator, so they're not necessarily aware of that.

So I'd either just use `NodeDeleteOk` for the intermediate result, or introduce this as a _private_ const in the actuation package, as it's an implementation detail of that package.

cluster-autoscaler/core/scaledown/actuation/delete_in_batch.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	for _, node := range nodes {
+		d.ctx.Recorder.Eventf(node, apiv1.EventTypeNormal, "ScaleDown", "nodes removed by cluster autoscaler")
+		d.clusterState.RegisterScaleDown(&clusterstate.ScaleDownRequest{
+			NodeGroup:          nodeGroup,
+			NodeName:           node.Name,
+			Time:               time.Now(),
+			ExpectedDeleteTime: time.Now().Add(MaxCloudProviderNodeDeletionTime),
+		})
+		drain := d.nodeDeletionTracker.IsNodeDrained(node.Name)
```
towca, line 137: Hmm, do you think it's worth it to jump through the deletion tracker for this? We could just pass and store the drain variable along with the node in this structure.

cluster-autoscaler/core/scaledown/actuation/delete_in_batch.go
```diff
@@ -0,0 +1,173 @@
...
+			}
+		}
+	}(nodes, nodeGroupId)
+	return nil
+}
+
+// deleteNodeFromCloudProvider removes the given nodes from cloud provider. No extra pre-deletion actions are executed on
+// the Kubernetes side. If successful, the deletion is recorded in CSR, and an event is emitted on the node.
+func (d *NodeBatcherDeleterImpl) deleteNodesFromCloudProvider(nodes []*apiv1.Node) error {
+	d.Lock()
```
towca, line 117: This method doesn't modify `d.deletionsPerNodeGroup`, so it shouldn't need a lock, right?

cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -174,7 +174,6 @@ func (a *Actuator) taintNodesSync(nodes []*apiv1.Node) errors.AutoscalerError {
 	for _, node := range nodes {
 		err := a.taintNode(node)
 		if err != nil {
-			a.ctx.Recorder.Eventf(node, apiv1.EventTypeWarning, "ScaleDownFailed", "failed to mark the node as toBeDeleted/unschedulable: %v", err)
```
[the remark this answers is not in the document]
yaroslava-serdiuk: I think it was done by accident, thanks for catching this!

cluster-autoscaler/core/scaledown/deletionbatcher/node_deletion_batcher.go
```diff
@@ -0,0 +1,29 @@
...
+package deletionbatcher
+
+import (
+	apiv1 "k8s.io/api/core/v1"
+)
+
+// NodeBatcherDeleter is gathering nodes and deleting them in bathch.
+type NodeBatcherDeleter interface {
+	// AddToBucket adds node to delete candidates.
+	AddToBucket(node *apiv1.Node) (string, error)
```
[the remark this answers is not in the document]
yaroslava-serdiuk: I see your point. 
In case we delete nodes in batch we don't delete individual nodes, we delete the whole batch. I initially thought that NodeBatcherDeleter is an interpretation of Bucket, where we can add nodes and call cleanup, but the logic what to add and when is on actuator side. 
However the idea to move the deletion logic to Batcher is good. I wouldn't name function as ScheduleNodeDeletion because we already have scheduleDeletion() method in actuator. I'd name it just AddNode().

cluster-autoscaler/core/scaledown/actuation/delete_in_batch.go
yaroslava-serdiuk: We already store the data if node was drain or not in tracker, why store it in another place?

cluster-autoscaler/processors/status/scale_down_status_processor.go
yaroslava-serdiuk: Ack

cluster-autoscaler/core/scaledown/actuation/actuator.go
yaroslava-serdiuk: Yes, this have similarity since it record events about failed ScaleDown, but scale down failure is different in those cases, so I'm not sure that it worth to create a generalised function. 
Also I don't agree that two functions (process_*) are less readable.

cluster-autoscaler/core/scaledown/deletionbatcher/node_deletion_batcher.go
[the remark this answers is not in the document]
towca: Yeah good idea, the type name already contains "delete" and "batch", so an `AddNode` method SGTM.

cluster-autoscaler/core/scaledown/actuation/delete_in_batch.go
towca: Mostly because it's a simple property that we have available when we add node to the batcher anyway, so we could avoid changing the tracker and introducing a function call. But that's not a very meaningful difference, so probably down to personal preferences. An argument could also be made for reducing lock contention in the tracker, but it'll probably be fine in practice (and we can always change it if we see a scalability regression). Just wanted to flag this as an option, the current approach also LGTM.

cluster-autoscaler/core/scaledown/actuation/actuator.go
towca: Sure, readability is frequently very subjective. However, the two function approach does have more duplication, and worse locality. I know there's not much in these functions anyway, I'm just worried about somebody adding/changing something in the future in one of the 3 places, and forgetting about one of the other ones. I think having all places call 1 function with empty/drain argument (and any other potential differentiators) would make that scenario less likely.

yaroslava-serdiuk: Agreed, updated the code.

cluster-autoscaler/core/scaledown/actuation/delete_in_batch.go
yaroslava-serdiuk: You're right. I added a map with drained nodes to the batcher directly and tested both versions, the versions with drained nodes map perform better (for 1000 empty nodes cluster, scale down was faster for 1 minute).

yaroslava-serdiuk:
/unhold

I updated the code a bit:
- added drained nodes map to the batcher directly and so I removed methods that I added to nodeDeleterTracker
- I also modified AddNode() method of the batcher. Previously it returned nodeGroupId only if the node was the first for current nodeGroup, now I return nodeGroupId always and add additional bool parameter which indicate if the node is the first for current batch.

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
```diff
@@ -937,6 +961,170 @@ func TestStartDeletion(t *testing.T) {
...
+						gotDeletedNodes[ngId] = value + 1
+					} else {
+						gotDeletedNodes[ngId] = 1
+					}
+				case <-time.After(1 * time.Second):
+					t.Errorf("Timeout while waiting for deleted nodes.")
+					break
+				}
+			}
+			if diff := cmp.Diff(test.wantSuccessfulDeletion, gotDeletedNodes); diff != "" {
```
towca, line 1121: What does this test verify that the test above doesn't? I'd imagine we want to test if the nodes actually get batched, but this test doesn't verify that at all, since the cloud provider hook processes nodes one-by-one. I'd try to assert if all nodes from one wave get deleted in the same API call. That will probably require some fiddling around with the cloud provider, but IMO it's needed.

cluster-autoscaler/core/scaledown/actuation/delete_in_batch_test.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	for _, node := range nodesEmpty {
+		_, _, err := d.addNodeToBucket(node, false)
+		if err != nil {
+			t.Errorf("addNodeToBucket return error %q when addidng node %v", err, node)
+		}
+	}
+	batches := len(d.deletionsPerNodeGroup)
+
+	// CleanUp NodeGroup that is not present in bucket.
```
towca, line 150: Why not make this a table-style test as well? I'd like to see a test case where we do `addNode, addNode, remove, addNode, remove for the same node group in quick succession (or maybe all but the last remove started at the same time as goroutines, and then a synchronous remove)`. Then assert that all nodes get deleted, and that the state is as expected at the end (empty `deletionsPerNodeGroup`, empty `drainedNodeDeletions`). I know it should work with the current implementation, but I can see somebody changing something in the future that would make us skip some node in this scenario, or leak memory in the structures.

cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -27,30 +27,32 @@ import (
...
 	ctx                 *context.AutoscalingContext
 	clusterState        *clusterstate.ClusterStateRegistry
 	nodeDeletionTracker *deletiontracker.NodeDeletionTracker
+	nodeDeletionBatcher deletionbatcher.NodeBatcherDeleter
 	evictor             Evictor
 }
 
 // NewActuator returns a new instance of Actuator.
-func NewActuator(ctx *context.AutoscalingContext, csr *clusterstate.ClusterStateRegistry, ndr *deletiontracker.NodeDeletionTracker) *Actuator {
+func NewActuator(ctx *context.AutoscalingContext, csr *clusterstate.ClusterStateRegistry, ndr *deletiontracker.NodeDeletionTracker, nbd deletionbatcher.NodeBatcherDeleter) *Actuator {
```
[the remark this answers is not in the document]
yaroslava-serdiuk: That's depends on batchDeleter implementation. When I added it at the beginning I had two implementation in mind, the first one is what I implemented, and another one is to remove nodes in separate go routine each k seconds. In second case we would like to have batcher outside of actuator.
Now it seems we are going to have one implementation. In this case we actually don't need to define an interface and it makes sense to have only a struct.

cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -251,41 +245,47 @@ func (a *Actuator) taintNode(node *apiv1.Node) error {
...
-	if err := DeleteNodeFromCloudProvider(a.ctx, node, a.clusterState); err != nil {
-		return status.NodeDeleteResult{ResultType: status.NodeDeleteErrorFailedToDelete, Err: err}
+// scheduleDeletion schedule the deletion on of the provided node by adding a node to NodeDeletionBatcher. If drain is true, the node is drained before being deleted.
+func (a *Actuator) scheduleDeletion(node *apiv1.Node, drain bool) {
+	status := a.prepareNodeForDeletion(node, drain)
+	if status.Err != nil {
+		RecordFailedScaleDownEvent(node, drain, a.ctx.Recorder, "prepareNodeForDelition failed", status.Err)
+		_, _ = deletetaint.CleanToBeDeleted(node, a.ctx.ClientSet, a.ctx.CordonNodeBeforeTerminate)
+		nodeGroup, err := a.ctx.CloudProvider.NodeGroupForNode(node)
+		if err != nil {
```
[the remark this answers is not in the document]
yaroslava-serdiuk: Actually we will call NodeGroupForNode() when actual node deletion happens. So if NodeGroupForNode() return nil I think we may have an error during deletion and probably inside batch as well. 
For scheduleDeletion() I would pass nodeGroupId, but for prepareNodeForDeletion we don't need nodeGroupId, so I would leave it as it is to have a middle check for NodeGroupForNode().

cluster-autoscaler/core/scaledown/actuation/delete_in_batch.go
```diff
@@ -0,0 +1,187 @@
...
+	defer d.Unlock()
+	nodes, ok := d.deletionsPerNodeGroup[nodeGroupId]
+	if !ok {
+		return nil
+	}
+	delete(d.deletionsPerNodeGroup, nodeGroupId)
+
+	go func(nodes []*apiv1.Node, nodeGroupId string) {
+		var result status.NodeDeleteResult
+		nodeGroup, err := deleteNodesFromCloudProvider(d.ctx, nodes)
```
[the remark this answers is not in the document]
yaroslava-serdiuk: I return nodeGroup, since recording ScaleDown events is happening outside of deleteNodesFromCloudProvider() function, however the function is exactly deleting nodes and nothing more. 
The name itself is quite long to add "andReturnNodeGroup", alternatively I can extract nodeGroup before calling deleteNodesFromCloudProvider() function and actually can pass nodeGroup to the function. Will it be better?

cluster-autoscaler/core/scaledown/deletionbatcher/node_deletion_batcher.go
```diff
@@ -0,0 +1,27 @@
...
+*/
+
+package deletionbatcher
+
+import (
+	apiv1 "k8s.io/api/core/v1"
+)
+
+// NodeBatcherDeleter is gathering nodes and deleting them in bathch.
+type NodeBatcherDeleter interface {
```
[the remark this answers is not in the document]
yaroslava-serdiuk: I removed interface, since it seems we are going with one implementation of batch

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
```diff
@@ -243,8 +244,10 @@ func TestCropNodesToBudgets(t *testing.T) {
 }
 
 func TestStartDeletion(t *testing.T) {
-	testNg := testprovider.NewTestNodeGroup("test-ng", 0, 100, 3, true, false, "n1-standard-2", nil, nil)
 	toBeDeletedTaint := apiv1.Taint{Key: deletetaint.ToBeDeletedTaint, Effect: apiv1.TaintEffectNoSchedule}
+	testNg := testprovider.NewTestNodeGroup("test-ng", 0, 100, 3, true, false, "n1-standard-2", nil, nil)
+	emptyNodeNodeGroup := generateNodesAndNodeGroupMap(4, "empty")
```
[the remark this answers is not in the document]
yaroslava-serdiuk: We have one error message for all nodes from node group when there is failed deletion scenario for scale down with batcher. Also, if one deletion failed in one node group it may result of failed deletion for another node group. 
However, here we have batch with 0 second batching interval, so the nodes should be deleted one by one. But this is not always true, since we lock batcher in addNodeToBucket() and remove(), so actually we may add few nodes before remove() call. 
And this is what happening time to time, the current implementation make the test flaky. I fixed the test, to not be flaky, but now I'm not sure if it's the right approach. I would like to have an old behaviour (i.e. no batching) when the batch interval is 0 seconds.

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
yaroslava-serdiuk: The waves is show the set of nodes for which one we call actuator.StartDeletion, but only nodes that belong to one nodeGroup are deleted in one API call. 
The previous test is testing batcher with 0 seconds interval and this test is testing batcher with >0 seconds interval. So this test is testing the e2e deletion. The tests in Batches are aimed to test the correctness of batching behaviour.

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
```diff
@@ -937,6 +961,170 @@ func TestStartDeletion(t *testing.T) {
...
+			provider := testprovider.NewTestCloudProvider(nil, func(nodeGroupId string, node string) error {
+				if gotFailedRequest(nodeGroupId) {
+					return fmt.Errorf("SIMULATED ERROR: won't remove node")
+				}
+				deletedResult <- nodeGroupId
+				return nil
+			})
+			// 2d array represent the waves of pushing nodes to delete.
+			deleteNodes := [][]*apiv1.Node{}
+			for ng, numNodes := range test.numNodesToDelete {
```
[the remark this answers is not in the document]
yaroslava-serdiuk: I added a test variable deleteCalls to improve readability.

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
```diff
@@ -274,13 +277,13 @@ func TestStartDeletion(t *testing.T) {
 				ScaledDownNodes: []*status.ScaleDownNode{
 					{
 						Node:        generateNode("empty-node-0"),
-						NodeGroup:   testNg,
+						NodeGroup:   emptyNodeNodeGroup["empty-node-0"],
```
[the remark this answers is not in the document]
yaroslava-serdiuk: The test is actually is back to previous state. Mark it as resolved.

cluster-autoscaler/core/scaledown/actuation/delete_in_batch_test.go
```diff
@@ -0,0 +1,169 @@
...
+		if batchNames != test.wantBatches {
+			t.Errorf("Want %d batches, got %d batches", test.wantBatches, batchNames)
+		}
+
+	}
+}
+
+func TestCleanUp(t *testing.T) {
+	provider := testprovider.NewTestCloudProvider(nil, func(id, nodeName string) error {
+		return nil
```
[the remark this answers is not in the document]
yaroslava-serdiuk: I added gotDeletedNodes slice

cluster-autoscaler/core/scaledown/actuation/delete_in_batch_test.go
yaroslava-serdiuk: The test now is table-style. This is unit test, that verify remove() call only. The whole batch, i.e AddNode() should be tested in actuator_test, that test the whole functionality.

cluster-autoscaler/core/scaledown/actuation/delete_in_batch_test.go
```diff
@@ -0,0 +1,216 @@
...
+			if err == nil {
+				t.Errorf("%s: remove() should return error, but return nil", test.name)
+			}
+			return
+		}
+		if err != nil {
+			t.Errorf("%s: remove() return error, but shouldn't", test.name)
+		}
+		gotDeletedNodes := []string{}
+		for i := 0; i < test.numNodes; i++ {
```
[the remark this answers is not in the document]
yaroslava-serdiuk: Right, the Unsuccessful remove case hasn't run because of return statement, now it fixed.
If we have failed deletion the following nodes are not deleted, so I check this deleted nodes only in success case and otherwise I check at least one failure.

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
```diff
@@ -937,6 +939,170 @@ func TestStartDeletion(t *testing.T) {
...
+			}
+			gotDeletedNodes := map[string]int{
+				"test-ng-1": 0,
+				"test-ng-2": 0,
+				"test-ng-3": 0,
+			}
+			for i := 0; i < wantDeletedNodes; i++ {
+				select {
+				case ngId := <-deletedResult:
+					if value, ok := gotDeletedNodes[ngId]; ok {
```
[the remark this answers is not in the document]
yaroslava-serdiuk: Right, thanks

cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -197,17 +197,13 @@ func (a *Actuator) deleteAsyncDrain(drain []*apiv1.Node) (scaledDownNodes []*sta
...
-			if result.Err == nil {
-				a.ctx.LogRecorder.Eventf(apiv1.EventTypeNormal, "ScaleDown", "Scale-down: node %s removed with drain", node.Name)
-			} else {
-				klog.Errorf("Scale-down: couldn't delete node %q with drain, err: %v", node.Name, result.Err)
-				a.ctx.Recorder.Eventf(node, apiv1.EventTypeWarning, "ScaleDownFailed", "failed to drain and delete node: %v", result.Err)
-				_, _ = deletetaint.CleanToBeDeleted(node, a.ctx.ClientSet, a.ctx.CordonNodeBeforeTerminate)
-			}
-		}(drainNode)
+		nodeGroup, err := a.ctx.CloudProvider.NodeGroupForNode(drainNode)
+		if err != nil {
```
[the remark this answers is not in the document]
yaroslava-serdiuk: Right

cluster-autoscaler/core/scaledown/actuation/delete_in_batch_test.go
```diff
@@ -0,0 +1,236 @@
...
+					t.Errorf("remove() should return error, but return nil")
+				}
+				return
+			}
+			if err != nil {
+				t.Errorf("remove() return error, but shouldn't")
+			}
+			if test.failedDeletion == 0 {
+				gotDeletedNodes := []string{}
+				for i := 0; i < test.numNodes; i++ {
```
towca, line 213: nit: The current implementation only checks the deleted nodes if `test.failedDeletion != 0`, which is not ideal. I think this could all be simplified to:
```

for i := 0; i < test.numNodes; i++ {
  select {
    case deletedNode := <-deletedNodes:
	gotDeletedNodes = append(gotDeletedNodes, deletedNode)
    case notDeletedNode := <-notDeletedNodes:
	notDeletedNodes = append(notDeletedNodes, notDeletedNode)
    case <-time.After(4 * time.Second):
	t.Errorf("Timeout while waiting for deleted nodes.")
}
```

Also, why aren't we asserting `gotDeletedNodes`? It's essentialy not used at all, right?

towca:
/lgtm
/approve
/hold

Thanks for all the changes!! Just one more nit, feel free to unhold if you don't agree or prefer to address it in a follow-up.

cluster-autoscaler/core/scaledown/actuation/delete_in_batch_test.go
yaroslava-serdiuk: Because when we have failed request cloud provider stop delete process and return error. It may happen that the first node in node delete loop is failed and so all other nodes won't appear in deletedNode

towca: Ahh got it, we need to have the different test cloud provider implementation I mentioned in another comment. Let's do that in a follow-up.

Issue #5072, opened 2022-08-04
cluster-autoscaler: Balancing scale up between different node groups does not account for node size difference


grosser:
for BalancingNodeGroupSetProcessor BalanceScaleUpBetweenGroups in cluster-autoscaler/processors/nodegroupset/balancing_processor.go

Given 3 azs with the nodes [[1x, 1x], [4x], [4x]] it will always chose the last 2 groups even though they have the highest capacity.
It should instead take the capacity into account.

I think a workable approach would be:
pass in nodeInfos, map nodes to asg via either instance ids or Comparator + templatenode (like FindSimilarNodeGroups does) and then convert to sum of capacity and use that for deciding where new nodes go

jaypipes:
@grosser apologies for the long delay in response. Are you suggesting that the node group selection algorithm choose the node group with the *least* capacity (in aggregate across all nodes in the group)?

grosser:
yes, it should balance by capacity
so for example `10,5,2 + 16 new nodes` -> 11,11,11

PR #5073, opened 2022-08-05
balance nodegroups by capacity and not node count

grosser:

cluster-autoscaler

#### What type of PR is this?

/kind feature

Scale nodegroups by their resource capacity (cpu or memory) and not by the count of their nodes.

Given a cluster with nodegroups with different node sizes (a=1 x 4 cpu, b=1 x 4 cpu, c=1 x 12 cpu)
and a need for 3 new nodes
It should scale to "a:3,b:2,c:1" to have a proper resource balance
and NOT "a:2,b:2,c:2" which it does atm.

verified the patch with a 3 aws asgs where I used 4xl,4xl,12xl and got this:
```
{"level":"INFO","message":"Final scale-up plan: [{usw2c 1->7 (max: 10)} {usw2a 1->6 (max: 10)} {usw2b 1-2 (max: 10)}]"}
```

Fixes https://github.com/kubernetes/autoscaler/issues/5072

```release-note
when balancing nodegroups to add new nodes, allow using their allocatable resources (cpu or memory) instead of node count with `--balance-similar-node-groups-by=cpu|memory`
```

### DIFF
--- a/cluster-autoscaler/config/autoscaling_options.go
+++ b/cluster-autoscaler/config/autoscaling_options.go
@@ -158,6 +158,8 @@ type AutoscalingOptions struct {
 	StatusConfigMapName string
 	// BalanceSimilarNodeGroups enables logic that identifies node groups with similar machines and tries to balance node count between them.
 	BalanceSimilarNodeGroups bool
+	// BalanceSimilarNodeGroupsBy decides what to balance nodegroups by default is node "count", other options are "cpu" and "memory"
+	BalanceSimilarNodeGroupsBy string
 	// ConfigNamespace is the namespace cluster-autoscaler is running in and all related configmaps live in
 	ConfigNamespace string
 	// ClusterName if available
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -175,6 +175,7 @@ var (
 	maxInactivityTimeFlag            = flag.Duration("max-inactivity", 10*time.Minute, "Maximum time from last recorded autoscaler activity before automatic restart")
 	maxFailingTimeFlag               = flag.Duration("max-failing-time", 15*time.Minute, "Maximum time from last recorded successful autoscaler run before automatic restart")
 	balanceSimilarNodeGroupsFlag     = flag.Bool("balance-similar-node-groups", false, "Detect similar node groups and balance the number of nodes between them")
+	balanceSimilarNodeGroupsByFlag   = flag.String("balance-similar-node-groups-by", "count", "Decides what to balance nodegroups by: \"count\", \"cpu\", or \"memory\"")
 	nodeAutoprovisioningEnabled      = flag.Bool("node-autoprovisioning-enabled", false, "Should CA autoprovision node groups when needed")
 	maxAutoprovisionedNodeGroupCount = flag.Int("max-autoprovisioned-node-group-count", 15, "The maximum number of autoprovisioned groups in the cluster.")
 
@@ -226,10 +227,12 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 	if err != nil {
 		klog.Fatalf("Failed to parse flags: %v", err)
 	}
+
 	minMemoryTotal, maxMemoryTotal, err := parseMinMaxFlag(*memoryTotal)
 	if err != nil {
 		klog.Fatalf("Failed to parse flags: %v", err)
 	}
+
 	// Convert memory limits to bytes.
 	minMemoryTotal = minMemoryTotal * units.GiB
 	maxMemoryTotal = maxMemoryTotal * units.GiB
@@ -238,9 +241,15 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 	if err != nil {
 		klog.Fatalf("Failed to parse flags: %v", err)
 	}
+
+	if !nodegroupset.BalanceBySupported[*balanceSimilarNodeGroupsByFlag] {
+		klog.Fatalf("Unsupported value for --balance-similar-node-groups-by: %v", balanceSimilarNodeGroupsByFlag)
+	}
+
 	if *maxDrainParallelismFlag > 1 && !*parallelDrain {
 		klog.Fatalf("Invalid configuration, could not use --max-drain-parallelism > 1 if --parallel-drain is false")
 	}
+
 	return config.AutoscalingOptions{
 		NodeGroupDefaults: config.NodeGroupAutoscalingOptions{
 			ScaleDownUtilizationThreshold:    *scaleDownUtilizationThreshold,
@@ -284,6 +293,7 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 		WriteStatusConfigMap:               *writeStatusConfigMapFlag,
 		StatusConfigMapName:                *statusConfigMapName,
 		BalanceSimilarNodeGroups:           *balanceSimilarNodeGroupsFlag,
+		BalanceSimilarNodeGroupsBy:         *balanceSimilarNodeGroupsByFlag,
 		ConfigNamespace:                    *namespace,
 		ClusterName:                        *clusterName,
 		NodeAutoprovisioningEnabled:        *nodeAutoprovisioningEnabled,
--- a/cluster-autoscaler/processors/nodegroupset/balancing_processor.go
+++ b/cluster-autoscaler/processors/nodegroupset/balancing_processor.go
@@ -17,15 +17,33 @@ limitations under the License.
 package nodegroupset
 
 import (
+	"fmt"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"sort"
 
+	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
 	"k8s.io/autoscaler/cluster-autoscaler/context"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
-	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
-
 	klog "k8s.io/klog/v2"
+	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
 )
+
+const (
+	// BalanceByCpu is the name for balancing by cpu
+	BalanceByCpu string = "cpu"
+	// BalanceByMemory is the name for balancing by memory
+	BalanceByMemory string = "memory"
+	// BalanceByCount is the name for balancing by count
+	BalanceByCount string = "count"
+)
+
+// BalanceBySupported is a map of supported balancing options for user input verification
+var BalanceBySupported = map[string]bool{
+	BalanceByCpu:    true,
+	BalanceByMemory: true,
+	BalanceByCount:  true,
+}
 
 // BalancingNodeGroupSetProcessor tries to keep similar node groups balanced on scale-up.
 type BalancingNodeGroupSetProcessor struct {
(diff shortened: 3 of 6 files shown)


cluster-autoscaler/main.go
```diff
@@ -265,6 +266,7 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 		WriteStatusConfigMap:               *writeStatusConfigMapFlag,
 		StatusConfigMapName:                *statusConfigMapName,
 		BalanceSimilarNodeGroups:           *balanceSimilarNodeGroupsFlag,
+		BalanceSimilarNodeGroupsBy:         *balanceSimilarNodeGroupsByFlag,
```
x13n, line 269: Can you check somewhere that the flag value is actually valid?

x13n (changes requested):

cluster-autoscaler/processors/nodegroupset/balancing_processor.go
```diff
@@ -179,5 +186,64 @@ func (b *BalancingNodeGroupSetProcessor) BalanceScaleUpBetweenGroups(context *co
 	return result, nil
 }
 
+// averageNodegroupAllocatableResources collects the average capacity by nodegroup id,
+// so we can add new nodes to the one with the lowest capacity instead of the one with the lowest node count
+func (b *BalancingNodeGroupSetProcessor) averageNodegroupAllocatableResources(context *context.AutoscalingContext, groups []cloudprovider.NodeGroup, nodes []*apiv1.Node) map[string]int {
```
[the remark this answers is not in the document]
grosser: template node would not work since they'd all have the same cpu, we need to inspect what is actually running in the cluster
it would work just fine on scale from 0 since it would scale it evenly (fallback path)

cluster-autoscaler/processors/nodegroupset/nodegroup_set_processor.go
```diff
@@ -35,19 +36,26 @@ type ScaleUpInfo struct {
 	NewSize int
 	// MaxSize is the maximum allowed size of the Group
 	MaxSize int
+	// AverageAllocatableResources is the average allocatable resource of the Group (either cpu or memory depending on flag)
```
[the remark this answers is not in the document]
grosser: yeah good idea!

cluster-autoscaler/processors/nodegroupset/balancing_processor.go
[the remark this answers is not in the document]
x13n: Oh, interesting. So how does CA know how many nodes to spin up if the number of cores in actual nodes can vary?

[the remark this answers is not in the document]
grosser: it's a gamble just like when spinning up new nodes to an existing cluster 😞 
... basically we get whatever aws has available since we allow multiple different sizes

cluster-autoscaler/main.go
grosser: added code for that

cluster-autoscaler/processors/nodegroupset/nodegroup_set_processor.go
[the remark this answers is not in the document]
grosser: updated

grosser:
updated as per suggestions, ptal
did not use enums because I saw no examples for that or support in flags

cluster-autoscaler/processors/nodegroupset/balancing_processor.go
```diff
@@ -179,5 +193,66 @@ func (b *BalancingNodeGroupSetProcessor) BalanceScaleUpBetweenGroups(context *co
...
+	nodegroupCounts := map[string]int{}
+	fallback := 0
+
+	if context.BalanceSimilarNodeGroupsBy == BalanceByCpu || context.BalanceSimilarNodeGroupsBy == BalanceByMemory {
+		for _, ng := range groups {
+			averageNodegroupCapacities[ng.Id()] = 0
+			nodegroupCounts[ng.Id()] = 0
+			ngNodes, err := ng.Nodes()
+			if err != nil {
+				klog.V(1).Infof("Failed to get nodes from ASG %v: %v", ng.Id(), err)
```
x13n, line 210: nit: s/ASG/NodeGroup/, since ASG is a AWS-specific context and this code is shared between different cloud providers.

cluster-autoscaler/processors/nodegroupset/balancing_processor.go
```diff
@@ -179,5 +193,66 @@ func (b *BalancingNodeGroupSetProcessor) BalanceScaleUpBetweenGroups(context *co
...
+				klog.V(3).Infof("Failed to find nodes for ASG %v: %v", ng.Id())
+			} else {
+				averageNodegroupCapacities[ng.Id()] /= nodegroupCounts[ng.Id()]
+				fallback = averageNodegroupCapacities[ng.Id()]
+			}
+		}
+	} else if context.BalanceSimilarNodeGroupsBy == BalanceByCount || context.BalanceSimilarNodeGroupsBy == "" {
+		// do nothing: fallback will set everything to 1
+	} else {
+		panic(fmt.Sprintf("Unknown balance option %v", context.BalanceSimilarNodeGroupsBy))
```
x13n, line 242: The flag value should be validated when CA starts, not when it tries to scale up.

cluster-autoscaler/processors/nodegroupset/balancing_processor.go
```diff
@@ -179,5 +193,66 @@ func (b *BalancingNodeGroupSetProcessor) BalanceScaleUpBetweenGroups(context *co
...
+			}
+		}
+
+		// calculate average and collect fallback
+		for _, ng := range groups {
+			if nodegroupCounts[ng.Id()] == 0 {
+				klog.V(3).Infof("Failed to find nodes for ASG %v: %v", ng.Id())
+			} else {
+				averageNodegroupCapacities[ng.Id()] /= nodegroupCounts[ng.Id()]
+				fallback = averageNodegroupCapacities[ng.Id()]
```
x13n, line 236: So fallback is the average capacity in a first (== random) node group? Shouldn't it be average across known node groups?

cluster-autoscaler/processors/nodegroupset/balancing_processor_test.go
```diff
@@ -213,50 +214,141 @@ func TestBalanceHittingMaxSize(t *testing.T) {
...
 	assert.Equal(t, 7, scaleUpMap["ng4"].NewSize)
 
 	// One node group exceeds max.
-	scaleUpInfo, _ = processor.BalanceScaleUpBetweenGroups(context, getGroups("ng2", "ng5"), 1)
+	scaleUpInfo, _ = processor.BalanceScaleUpBetweenGroups(context, getGroups("ng2", "ng5"), 1, []*apiv1.Node{})
 	assert.Equal(t, 1, len(scaleUpInfo))
 	scaleUpMap = toMap(scaleUpInfo)
 	assert.Equal(t, 2, scaleUpMap["ng2"].NewSize)
 }
+
```
x13n, line 264: All the new tests have a very similar structure, can you covert them into a single table-based test instead?

cluster-autoscaler/processors/nodegroupset/nodegroup_set_processor.go
```diff
@@ -35,19 +36,26 @@ type ScaleUpInfo struct {
 	NewSize int
 	// MaxSize is the maximum allowed size of the Group
 	MaxSize int
+	// Weight is the average allocatable resource of the Group (either cpu or memory depending on flag)
```
x13n, line 39: This comment doesn't work for "count".

x13n (changes requested):
I think it might be worth implementing allocatable-based balancing in a separate processor. This would have a few benefits:
- No need to change the interface
- No need to check "mode of operation" at runtime every time balancing happens
- No need to keep the mode of operation in the context, just pass it as param when the object is first created
- Smaller, better focused objects are easier to read, understand & debug

WDYT?

cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -125,6 +125,8 @@ type AutoscalingOptions struct {
 	StatusConfigMapName string
 	// BalanceSimilarNodeGroups enables logic that identifies node groups with similar machines and tries to balance node count between them.
 	BalanceSimilarNodeGroups bool
+	// BalanceSimilarNodeGroupsBy decides what to balance nodegroups by default is node "count", other options are "cpu" and "memory"
+	BalanceSimilarNodeGroupsBy string
```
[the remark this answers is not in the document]
grosser: I stayed with the string but used consts to compare, since I did not see any example of enums being used for cli flags that I could use

cluster-autoscaler/processors/nodegroupset/balancing_processor.go
grosser: yeah, good catch!

cluster-autoscaler/processors/nodegroupset/balancing_processor_test.go
grosser: I want the tests to tell a story, be easy to extend with 1-off cases, and match the existing test structure.

cluster-autoscaler/processors/nodegroupset/balancing_processor.go
grosser: done

grosser:
Re extracting processor interface/creating multiple processors:
I like how the allocatable logic fits into the existing processor (barely changes it and it is easy to add new supported types).
- `check "mode of operation"` seems very cheap to me
- context is in the method already and would have to be replaced by passing the processor
- this PR already touches lots of things so I'd prefer to keep it small
- it sounds like a refactoring that can easily be done after this initial implementation since not interface needs to change

x13n:
Yikes, I surely forgot to follow up here, sorry about that. Assigning so it won't slip again:

/assign

Re: splitting the object into smaller ones - I think the current implementation goes against single responsibility principle and will be harder to maintain down the road. OTOH, as you say, it should be relatively easy to refactor as a follow-up if needed.
Re: interface change - I don't think current set of nodes belongs in the interface. It just so happens that the current implementation of the processor interface requires nodes to make a decision, but another implementation might need another source of information (e.g. user-defined preference for node group size ratio) and skip nodes entirely - no such source of information belongs in the interface. If a specific implementation requires a list of nodes, then some node list fetching object/function should be passed to the object when it is created.

grosser:
So you are suggesting:
- build a new `BalancingNodeGroupByResourceSetProcessor ` or something like that ?
(and then copying 200 existing LOC + adding the new logic)
- keeping BalanceScaleUpBetweenGroups interface the same as before and then passing in the k8s client to then get the nodes from the cluster again ?

correct ?

x13n:
> build a new BalancingNodeGroupByResourceSetProcessor  or something like that ?
(and then copying 200 existing LOC + adding the new logic)

My initial suggestion was indeed to create a dedicated processor, possibly sharing rather than copying the existing code. Pluggable processor with 3 possible balancers might actually be nicer though, making it easier to share the code.

> keeping BalanceScaleUpBetweenGroups interface the same as before and then passing in the k8s client to then get the nodes from the cluster again ?

Not the client directly - that'd cause an extra increase in QPS. I was thinking about e.g. `AllNodeLister` instead: https://github.com/kubernetes/autoscaler/blob/4d9412022398bc4f10209c0a30bb2ff77dad840c/cluster-autoscaler/utils/kubernetes/listers.go#L39

grosser:
FYI rebased commit in case anyone needs this https://github.com/kubernetes/autoscaler/commit/f3f0bb6b65b7acb8b6cd506385017ee74f3fb775

PR #5078, opened 2022-08-09
CA: Iterate through existed node groups in AnnotationNodeInfoProvider

yaroslava-serdiuk:

/cluster-autoscaler

#### What type of PR is this?

/kind bug

In case node group scaled down to 0, the NodeGroupForNode is returning nil NodeGroup and nil error.


mwielgus (approved):
/lgtm
/approve

PR #5083, opened 2022-08-11
cluster-autoscaler balanced nodegroups: allow ignoring missing expansion during pod filtering

grosser:

cluster-autoscaler

#### What type of PR is this?

/kind feature

We saw issues with betting a good balance from this error:
```
No info about pods passing predicates found for group euc1c-1234, skipping it from scale-up consideration
```

Since all our nodegroups are the same, all pods will fit them, so we do not need this check.
Add a new flag to disable the check so we have 1 less things to worry about / debug.

```release-note
Add balance-ignore-missing-expansion flag to allow skipping pod filtering when expansion option failed to be computed
```

### DIFF
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -176,6 +176,7 @@ var (
 	maxInactivityTimeFlag            = flag.Duration("max-inactivity", 10*time.Minute, "Maximum time from last recorded autoscaler activity before automatic restart")
 	maxFailingTimeFlag               = flag.Duration("max-failing-time", 15*time.Minute, "Maximum time from last recorded successful autoscaler run before automatic restart")
 	balanceSimilarNodeGroupsFlag     = flag.Bool("balance-similar-node-groups", false, "Detect similar node groups and balance the number of nodes between them")
+	balanceIgnoreMissingExpansion    = flag.Bool("balance-ignore-missing-expansion", false, "When balancing nodegroups, sometimes expansion options are unknown, in that case proceed with scheduling pods")
 	nodeAutoprovisioningEnabled      = flag.Bool("node-autoprovisioning-enabled", false, "Should CA autoprovision node groups when needed")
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/main.go
```diff
@@ -170,6 +170,7 @@ var (
 	maxInactivityTimeFlag            = flag.Duration("max-inactivity", 10*time.Minute, "Maximum time from last recorded autoscaler activity before automatic restart")
 	maxFailingTimeFlag               = flag.Duration("max-failing-time", 15*time.Minute, "Maximum time from last recorded successful autoscaler run before automatic restart")
 	balanceSimilarNodeGroupsFlag     = flag.Bool("balance-similar-node-groups", false, "Detect similar node groups and balance the number of nodes between them")
+	balanceFilterByPodsFlag          = flag.Bool("balance-filter-by-pods", true, "Enables filtering groups with similar machines by the pods that they need to fit")
```
mwielgus, line 173: Please also update FAQ.md, preferably with more description of what this flag exactly does (hard to tell by just reading this description).

grosser: reworked the PR to keep the error message but allow proceeding so we can still see that the issue exists (and remove the flag once it's fixed)

grosser:
"PR needs rebase" bot is kinda useless if there is not marked as "ready to be merged" 🤷

x13n:
/assign

This looks like a workaround for a bug, but it isn't clear to me what the bug is in the first place. Wouldn't it be more useful to debug & fix the underlying issue instead of adding yet another flag for modifying Cluster Autoscaler behavior? This increases CA API surface, making it harder to maintain.

grosser:
yes
I still don't know how that bug happens

mwielgus:

grosser:
FYI rebased commit in case anyone needs this https://github.com/kubernetes/autoscaler/commit/c2b06feeb68bee2531548002c5e61c60ad40728f

PR #5088, opened 2022-08-12
test: use `T.Setenv` to set env vars in tests

Juneezee:

cluster-autoscaler

#### What type of PR is this?

/kind cleanup

This PR replaces `os.Setenv` with `t.Setenv`. Starting from Go 1.17, we can use `t.Setenv` to set environment variable in test. The environment variable is automatically restored to its original value when the test and all its subtests complete. This ensures that each test does not start with leftover environment variables from previous completed tests.

This saves us at least 2 lines (error check, and unsetting the env var) on every instance.

Reference: https://pkg.go.dev/testing#T.Setenv

```release-note
NONE
```

```docs

```


mwielgus (approved):
/lgtm
/approve

PR #5097, opened 2022-08-17
Improve overprovisioning pause pod termination time in FAQ

jahands:

cluster-autoscaler

#### What type of PR is this?

/kind documentation

Speeds up the time it takes to replace an overprovisioning pause pod with an application pod by 3-11 seconds

Fixes #

I have noticed when deploying pods that replace overprovisioning pause pods, adding this termination grace period will speed it up by 3-11 seconds, without any downsides that I know of. I suspect this is because k8s will send a SIGKILL instead of SIGTERM + poll for completion when `terminationGracePeriodSeconds` is set to `0`

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/FAQ.md
+++ b/cluster-autoscaler/FAQ.md
@@ -419,6 +419,7 @@ spec:
         run: overprovisioning
     spec:
       priorityClassName: overprovisioning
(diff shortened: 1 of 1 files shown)


mwielgus (approved):
/lgtm
/approve

PR #5099, opened 2022-08-17
feat: add support for service monitor annotations to helm chart

pierluigilenoci:

helm chart

#### What type of PR is this?

/kind feature

When creating clusters with the GitOps approach, `autoscaler` is sometimes installed before `Prometheus` so it is possible that the CRD is not present and the installation fails. Through some annotations, it is possible to change the installation order of the manifests and thus solve this problem.

No issue

NONE


pierluigilenoci:
@bskiba can you please take a look?

mwielgus (approved):
/lgtm
/approve

pierluigilenoci:
@mwielgus thank you a lot.

PR #5100, opened 2022-08-17
Export metric about result of queried VPA makes to metrics server

jbartosik:

VPA recommender

#### What type of PR is this?

/kind feature

Make it easier to detect when VPA is not working because it can't get
metrics. Currently one can check logs to do that but having a metric makes the
information easier to use.

Metric reports only error / no error because parsing error messages for more
detailed info would easily break / be more work to maintain than I think it's
worth and we don't have access to structured details.

<!--
```release-note
Added metric_server_responses metric
```

```docs

```


jbartosik:

mwielgus (approved):
/lgtm
/approve

PR #5108, opened 2022-08-18
cluster-autoscaler aws: remove duplicated permissions

grosser:

cluster-autoscaler

#### What type of PR is this?

/kind documentation

remove a duplicated permission in the docs (already in the first block, no need in the second block)

```release-note
NONE
```


gjtempleton:
Thanks!
/lgtm
/approve

PR #5111, opened 2022-08-19
Export metric_server_response like other VPA recommender metrics

jbartosik:
There is no reason for the metric to be exported differently than other metrics

Hyphens in name of the metric caused prometheus errors

VPA recommmender

#### What type of PR is this?

/kind bug

VPA recommender fails to initialize because we call MustRegister on a metric with an invalid name and the registration fails.

```release-note
NONE
```


jbartosik:
/hold
I want to check some more that I'm doing the correct thing here

jbartosik:
@krzysied @mwielgus PTAL. I built recommender & run it in my cluster to make sure that it works

PR #5112, opened 2022-08-19
Export validateVPA to use in other places

pbeschetnov:

VPA admission-controller

#### What type of PR is this?

/kind feature

`validateVPA` is a common function that can be used outside of the `admission-controller/resource/vpa/handler.go`

NONE


jbartosik (approved):
/lgtm
/approve

PR #5115, opened 2022-08-22
Add PdbRemainingDisruptions struct

yaroslava-serdiuk:
cluster-autoscaler

#### What type of PR is this?
/kind feature

A part of Scale Down in Parallel implementation

### DIFF
--- a/cluster-autoscaler/core/scaledown/pdb/pdb.go
+++ b/cluster-autoscaler/core/scaledown/pdb/pdb.go
@@ -1,0 +1,82 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package pdb
+
+import (
+	"fmt"
+
+	apiv1 "k8s.io/api/core/v1"
+	policyv1 "k8s.io/api/policy/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/drain"
+	"k8s.io/klog/v2"
+)
+
+// PdbRemainingDisruptions stores how many discuptiption is left for pdb.
+type PdbRemainingDisruptions struct {
+	pdbs []*policyv1.PodDisruptionBudget
+}
+
+// NewPdbRemainingDisruptions initialize PdbRemainingDisruptions.
+func NewPdbRemainingDisruptions(pdbs []*policyv1.PodDisruptionBudget) *PdbRemainingDisruptions {
+	pdbsCopy := make([]*policyv1.PodDisruptionBudget, len(pdbs))
+	for i, pdb := range pdbs {
+		pdbsCopy[i] = pdb.DeepCopy()
+	}
+	return &PdbRemainingDisruptions{pdbsCopy}
+}
+
+// CanDisrupt return if the pod can be removed.
+func (p *PdbRemainingDisruptions) CanDisrupt(pods []*apiv1.Pod) (bool, *drain.BlockingPod) {
+	for _, pdb := range p.pdbs {
+		selector, err := metav1.LabelSelectorAsSelector(pdb.Spec.Selector)
+		if err != nil {
+			klog.Errorf("Can't get selector for pdb %s", pdb.GetNamespace()+" "+pdb.GetName())
+			return false, nil
+		}
+		count := int32(0)
+		for _, pod := range pods {
+			if pod.Namespace == pdb.Namespace && selector.Matches(labels.Set(pod.Labels)) {
+				count += 1
+				if pdb.Status.DisruptionsAllowed < count {
+					return false, &drain.BlockingPod{Pod: pod, Reason: drain.NotEnoughPdb}
+				}
+			}
+		}
+	}
+	return true, nil
+}
+
+// Update make updates the remaining disruptions for pdb.
+func (p *PdbRemainingDisruptions) Update(pods []*apiv1.Pod) error {
+	for _, pdb := range p.pdbs {
+		selector, err := metav1.LabelSelectorAsSelector(pdb.Spec.Selector)
+		if err != nil {
+			return err
+		}
+		for _, pod := range pods {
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/core/scaledown/pdb/pdb.go
```diff
@@ -0,0 +1,53 @@
...
+	pdbDisruptions, ok := p.remainingDisruptions[pdbName]
+	if !ok {
+		return 0, fmt.Errorf("There is no pdb %s in list", pdbName)
+	}
+	return pdbDisruptions, nil
+}
+
+// UpdatePdb updates the remaining disrubtions for pdb.
+func (p PdbRemainingDisruptions) UpdatePdb(pdb policyv1.PodDisruptionBudget, disrubtions int) error {
+	pdbName := pdb.GetObjectMeta().GetNamespace() + pdb.GetObjectMeta().GetName()
```
x13n, line 43: We should have some kind of separator here, otherwise pdb `foo` in namespace `bar` is the same as pdb `oo` in namespace `barf`.

cluster-autoscaler/core/scaledown/pdb/pdb.go
```diff
@@ -0,0 +1,53 @@
...
+	pdbDisruptions, ok := p.remainingDisruptions[pdbName]
+	if !ok {
+		return 0, fmt.Errorf("There is no pdb %s in list", pdbName)
+	}
+	return pdbDisruptions, nil
+}
+
+// UpdatePdb updates the remaining disrubtions for pdb.
+func (p PdbRemainingDisruptions) UpdatePdb(pdb policyv1.PodDisruptionBudget, disrubtions int) error {
```
x13n, line 42: nit: s/disrubtions/disruptions/

cluster-autoscaler/core/scaledown/pdb/pdb.go
```diff
@@ -0,0 +1,53 @@
...
+package pdb
+
+import (
+	"fmt"
+
+	policyv1 "k8s.io/api/policy/v1"
+)
+
+// PdbRemainingDisruptions stores how many discuptiption is left for pdb.
+type PdbRemainingDisruptions struct {
```
x13n, line 26: I think it would be much more useful to have pod<->pdb matching logic here, and the following API in this struct:

```
// Init the struct with known PDBs
func NewPdbRemainingDisruptions([]*policyv1.PodDisruptionBudget)
// Check all matching disruptions and return true iff disruption is possible
func CanDisrupt([]*apiv1.Pod) bool
// Update all matching disruptions, decrementing remaining disruption count for each pdb
func Update([]*apiv1.Pod)
```

So, scale down logic would create the object before the simulation, use `CanDisrupt` to check if all the pods on a given node are possible to evict, and then finally `Update` if the node satisfies other criteria for scale down.

x13n (changes requested):
I think the API should be slightly different so the struct is more useful. It would also allow for a bit more interesting test cases.

cluster-autoscaler/core/scaledown/pdb/pdb.go
```diff
@@ -0,0 +1,53 @@
+/*
+Copyright 2016 The Kubernetes Authors.
```
mwielgus, line 2: 2022

cluster-autoscaler/core/scaledown/pdb/pdb.go
```diff
@@ -0,0 +1,84 @@
...
+		selector, err := metav1.LabelSelectorAsSelector(pdb.Spec.Selector)
+		if err != nil {
+			return err
+		}
+		for _, pod := range pods {
+			if pod.Namespace == pdb.Namespace && selector.Matches(labels.Set(pod.Labels)) {
+				if pdb.Status.DisruptionsAllowed < 1 {
+					return fmt.Errorf("Pod can't be removed, pdb is blocking by pdb %s, disruptionsAllowed: %v", pdb.GetNamespace()+"/"+pdb.GetName(), pdb.Status.DisruptionsAllowed)
+				}
+				pdb.Status.DisruptionsAllowed -= 1
```
x13n, line 79: I think it would be better if PDBs weren't modified directly. You can either do a deep copy in `NewPdbRemainingDisruptions`, or just keep track of the fields we care about (selector, namespace, disruptions allowed). This will prevent surprising bugs if any other part of CA code tries to use `DisruptionsAllowed` on the same object.

cluster-autoscaler/core/scaledown/pdb/pdb_test.go
```diff
@@ -0,0 +1,138 @@
...
+			pdbsDisruptions: [2]int32{2, 4},
+			canDisrupt:      true,
+		},
+	}
+	for _, test := range testCases {
+		pdb1.Status.DisruptionsAllowed = test.pdbsDisruptions[0]
+		pdb2.Status.DisruptionsAllowed = test.pdbsDisruptions[1]
+		pdbRemainingDisruptions := NewPdbRemainingDisruptions(test.pdbs)
+		pods := []*apiv1.Pod{}
+		for i := 0; i < test.podsLabel1; i++ {
```
x13n, line 104: I'd extract this to some `makePodsWithLabel(string, int)` helper func to avoid having two essentially identical loops.

cluster-autoscaler/core/scaledown/pdb/pdb_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"testing"
+
+	apiv1 "k8s.io/api/core/v1"
+	policyv1 "k8s.io/api/policy/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	. "k8s.io/autoscaler/cluster-autoscaler/utils/test"
+)
+
+func TestCanDisrupt(t *testing.T) {
```
x13n, line 30: Can you also add some test cases for updating? Also, maybe worth testing one pod covered by mutliple PDBs.

x13n (changes requested):

cluster-autoscaler/core/scaledown/pdb/pdb.go
```diff
@@ -0,0 +1,80 @@
...
+			}
+		}
+	}
+	return true, nil
+}
+
+// Update make updates the remaining disruptions for pdb.
+func (p *PdbRemainingDisruptions) Update(pods []*apiv1.Pod) (*PdbRemainingDisruptions, error) {
+	pdbsCopy := make([]*policyv1.PodDisruptionBudget, len(p.pdbs))
+	copy(pdbsCopy, p.pdbs)
```
x13n, line 64: Why copy this with every `Update` call and not just once when creating the object?

Also: will `copy` be sufficient to copy all the nested fields? Shouldn't this be using `DeepCopy` method of PDB object?

x13n (changes requested):

cluster-autoscaler/core/scaledown/pdb/pdb.go
yaroslava-serdiuk: I misunderstood the purpose of a copy, updated.

PR #5116, opened 2022-08-22
Azure: support node label keys having underscores

bpineau:
Azure VMSS tags allowed character set isn't the exact same as the Kubernetes labels' character set. In particular "/" aren't supported as VMSS tags, but are allowed for Kubernetes labels. To support reflecting label values containing "/" to VMSS tags, the cluster-autoscaler evaluates underscores ("_") in VMSS tags keys as slashes ("/") in nodes labels keys.

But due to that substitution, a node label containing "_" (also a valid character for a k8s label) can't be reflected to VMSS tags (as it will be mis-read as a "/").

This PR takes inspiration from rfc6901's JSON escaping (rather hideous, but possibly familiar as it's used by kubectl-patch), by using ("~2") as a replacement for "_".

The tilde (`~`) character isn't allowed in k8s labels (and taints) keys so there's no possible ambiguities or regressions (a VMSS tag key having a "`~`" is currently evaluated to an impossible k8s label).

Labels and taints supported character sets:
https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#taint

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/README.md
+++ b/cluster-autoscaler/cloudprovider/azure/README.md
@@ -37,13 +37,13 @@ If you are using `nodeSelector`, you need to tag the VMSS  with a node-template 
 
 To add the label of `foo=bar` to a node from a VMSS pool, you would add the following tag to the VMSS `k8s.io_cluster-autoscaler_node-template_label_foo: bar`.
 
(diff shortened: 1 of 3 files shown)


bpineau:
Hi @feiskyer @nilo19 @marwanad , could you have a look please?

marwanad:
The change looks good to me.

/cc @gandhipr for the final review/OK as I moved on from Azure.

PR #5120, opened 2022-08-23
aws cloudprovider - unify and consolidate aws session creation

khizunov:
cluster-autoscaler

#### What type of PR is this?
/kind bug

Unify and consolidate aws session creation (currently created in multiple places and with different configuration).
AWS SDK provider will configure and create the session according to provided configuration (e.g cloud config file)

Non standard AWS regions (GovCloud, etc.) which uses different AWS api endpoints should use unified session configuration for all requests - currently GenerateEC2InstanceTypes doesn't take that into consideration.

Fixes #

```release-note
NONE
```

```docs

```


mwielgus:
Can you provide more information (preferably in the PR description) about what exactly has been done there and why do we need this?

khizunov:
pinging @gjtempleton @feiskyer for approval

cluster-autoscaler/cloudprovider/aws/ec2_instance_types/gen.go
```diff
@@ -83,7 +83,14 @@ func main() {
 	flag.Parse()
 	defer klog.Flush()
 
-	instanceTypes, err := aws.GenerateEC2InstanceTypes(*region)
+	sess, err := session.NewSession(&aws.Config{
```
gjtempleton, line 86: This file doesn't import session so `make generate` now fails with:

```
go generate ./cloudprovider/aws
# command-line-arguments
ec2_instance_types/gen.go:86:15: undefined: session
ec2_instance_types/gen.go:86:39: undefined: aws.Config
ec2_instance_types/gen.go:87:15: undefined: aws.String
cloudprovider/aws/aws_manager.go:17: running "go": exit status 2
```

gjtempleton (changes requested):
Thanks for this, there's one thing needed to fix up with the instance generation functionality, but other than that, this looks good to go to me.

cluster-autoscaler/cloudprovider/aws/ec2_instance_types/gen.go
khizunov: @gjtempleton The AWS Session doesn't work without region (maybe it worked some day I am not sure), I did some rework on gen.go if you think I should put it in different PR I can do that?

gjtempleton:
Thanks for addressing. 

Looks good to me now.

/lgtm
/approve

PR #5124, opened 2022-08-25
feat: add support for metricRelabelings

omerlh:

Helm chart for cluster autoscaler

#### What type of PR is this?

/kind feature

Add support for metricRelabelings in the service monitor in the helm chart, so we could drop metrics we don't care about - to reduce costs

Fixes #

```release-note
Added: support for metricRelabelings in the cluster autoscaler helm chart
```

```docs

```

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -11,4 +11,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.20.1
+version: 9.21.0
--- a/charts/cluster-autoscaler/templates/servicemonitor.yaml
+++ b/charts/cluster-autoscaler/templates/servicemonitor.yaml
@@ -20,6 +20,10 @@ spec:
   - port: {{ .Values.service.portName }}
     interval: {{ .Values.serviceMonitor.interval }}
     path: {{ .Values.serviceMonitor.path }}
+    {{- if .Values.serviceMonitor.metricRelabelings }}
+    metricRelabelings:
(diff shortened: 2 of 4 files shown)


charts/cluster-autoscaler/Chart.yaml
```diff
@@ -11,4 +11,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.20.0
+version: 9.20.1
```
gjtempleton, line 14: This is a new feature, please make this a minor version bump.

charts/cluster-autoscaler/templates/servicemonitor.yaml
```diff
@@ -20,6 +20,10 @@ spec:
   - port: {{ .Values.service.portName }}
     interval: {{ .Values.serviceMonitor.interval }}
     path: {{ .Values.serviceMonitor.path }}
+    {{- if .Values.serviceMonitor.metricRelabelings }}
+    metricRelabelings:
+{{ tpl (toYaml .Values.serviceMonitor.metricRelabelings | indent 6) . }}
```
gjtempleton, line 25: Please add this new field to the [values.yml](https://github.com/kubernetes/autoscaler/blob/035e762e89dade935f7954e64d78ef54512946cf/charts/cluster-autoscaler/values.yaml#L331)

gjtempleton (changes requested):
Couple of changes required.

charts/cluster-autoscaler/Chart.yaml
omerlh: fixed

charts/cluster-autoscaler/templates/servicemonitor.yaml
omerlh: fixed

PR #5127, opened 2022-08-25
Update go version to 1.18.5

voelzmo:
vertical-pod-autoscaler

#### What type of PR is this?
/kind cleanup

* Update go version to 1.18.5 to stay in sync with k8s 1.25: https://github.com/kubernetes/release/issues/2625
* Switch from `go get` to `go install`, as `go get` no longer installs binaries

```release-note
Bump go version to 1.18.5
```

### DIFF
--- a/vertical-pod-autoscaler/builder/Dockerfile
+++ b/vertical-pod-autoscaler/builder/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


jbartosik:
Do we also update go version in https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/go.mod ?

mwielgus (approved):
/lgtm
/approve

PR #5130, opened 2022-08-26
Update go.mod go version to 1.18

olagacek:
Update go.mod go version to 1.18 in cluster autoscaler to be in line with version used by k/k repo in v1.25.0-alpha.2.

cluster-autoscaler

#### What type of PR is this?

/kind cleanup

Updates k8s mod version to be in line with k/k repo


olemarkus:
Is it worth just bumping the alpha deps to stable as well now that it has been released?

/lgtm

olagacek:
> 

Seems to be a better idea, let me create another PR

PR #5131, opened 2022-08-26
Allow simulator to persist changes in cluster snapshot

x13n:
#### Which component this PR applies to?

cluster-autoscaler

#### What type of PR is this?

/kind cleanup
/kind feature

#### What this PR does / why we need it:

One more refactor of existing scale down code in order to make parallel scale down (https://github.com/kubernetes/autoscaler/issues/5079) possible.

#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for your reviewer:

This is a follow-up to https://github.com/kubernetes/autoscaler/pull/5118. Only last 2 commits should be reviewed as a part of this PR. It will be on hold until the previous one is merged.

/hold

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

/assign @MaciekPytel 
/assign @yaroslava-serdiuk

PR #331, opened 2017-09-12
Respect minimum cores/memory limit during scale down

aleksandra-malinowska:
Follow up to #295, adding support for lower limits of total cluster cores/memory.

PR #5133, opened 2022-08-26
Stop treating masters differently in scale down

x13n:

cluster-autoscaler

#### What type of PR is this?

/kind bug
/kind cleanup

This filtering was used for two purposes:
- Excluding masters from destination candidates
- Excluding masters from calculating cluster resources

Excluding from destination candidates isn't useful: if pods can schedule
there, they will, so removing them from CA simulation doesn't change
anything.
Excluding from calculating cluster resources actually matches scale up
behavior, where master nodes are treated the same way as regular nodes.

Fixes #

This is a follow-up to https://github.com/kubernetes/autoscaler/pull/5131. Only the last commit should be reviewed as a part of this PR. It will be on hold until the previous one is merged.

/hold

```release-note
Nodes hosting kube-apiserver are no longer treated differently in scale-down logic. The special handling was removed after it was discovered scale down resource limiting was inconsistent with scale-up and haven't actually worked correctly since CA 1.18.
```

```docs

```

/assign @MaciekPytel
/assign @yaroslava-serdiuk

### DIFF
--- a/cluster-autoscaler/core/scaledown/legacy/legacy.go
+++ b/cluster-autoscaler/core/scaledown/legacy/legacy.go
@@ -39,10 +39,8 @@ import (
 
 	apiv1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	klog "k8s.io/klog/v2"
-	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
 )
 
 // ScaleDown is responsible for maintaining the state needed to perform unneeded node removals.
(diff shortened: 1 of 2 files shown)


yaroslava-serdiuk:
> Excluding from calculating cluster resources actually matches scale up
behavior, where master nodes are treated the same way as regular nodes.

CA don't want to scale down a master node, why do we want to include a master node in scale down candidates?

x13n:
The thing is, master nodes were always included in scale down candidates: the filtering happened for destination candidates, which made little sense.

MaciekPytel:
I agree filtering masters from candidate nodes makes no sense, but I think the fact that they're only filtered out from candidate list is a bug. As far as I can tell it was introduced in #2890 (specifically [here](https://github.com/kubernetes/autoscaler/pull/2890/files#diff-a8bcd21686c03aba976761ae055fd3b70e6e2bf5ee77d15f66946f3b0f1d7c30L781) - before that change the loop selecting node to delete would skip master nodes).

With above in mind - I fully agree with not filtering masters from destinations and I also agree resource limits should be treated consistently across scale-up and scale-down. The question is whether we should start filtering out the masters from candidate lists like we did before #2890? The fact that it has been broken since 1.18 and no-one noticed, suggests that we may be fine getting rid of this logic.

I'd consider updating the release note though to give slightly more context on this issue - it wouldn't be immediately obvious to me what you mean by cluster resources there and I'm probably somewhat more familiar with CA than the target audience of the release notes. Maybe something along the lines of `Nodes hosting kube-apiserver are no longer treated differently in scale-down logic. The special handling was removed after it was discovered it haven't worked correctly since CA 1.18.`? This doesn't require much context and anyone interested in details can just follow the link to this PR.

x13n:
The logic to filter out masters was introduced long before that, in https://github.com/kubernetes/autoscaler/pull/331. It was meant to exclude masters from calculating resource limits. Before creating this PR I chatted with @aleksandra-malinowska who authored https://github.com/kubernetes/autoscaler/pull/331 all these years ago and we concluded it is safe to remove the filtering as it is not consistent between scaling up and down. I agree the release note could be improved though - updated.

/hold cancel

PR #5138, opened 2022-08-28
Use cloud-provider-aws instead of legacy-cloud-providers/aws

### LINKED ISSUE #5625: Cluster Autoscaler 1.27 blocked on missing AWS cloud provider dependencies

Cluster Autoscaler

AWS Cloud provider is using `k8s.io/legacy-cloud-providers/aws` package as a dependancy. This package was removed in Kubernetes release `v1.27.0-beta.0`, so migrating from it is a blocker for Cluster Autoscaler release 1.27.

We should migrate from `k8s.io/legacy-cloud-providers/aws`.

**Reproduction steps.**:
1. Change the go version in `cluster-autoscaler/go.mod` from `go 1.19` to `go 1.20`.
2. Run update vendor script `bash hack/update-vendor.sh v1.27.0-beta.0`.

olemarkus:

cluster autoscaler

#### What type of PR is this?

/kind bug

The AWS cloud provider has been removed from k/k and subsequently from the legacy cloud provider repo.

Also, the AWS cloud provider's cloud config has diverged from legacy-cloud-provider. The former has more configuration options that legacy cloud provider does not know about. Parsing the cloud config will fail of any of these additional fields are used.

Switching to using the AWS provider from cloud-provider-aws solves this issue.

/fixes #5625

```release-note
NONE
```


feiskyer:
+AWS owners @jaypipes @gjtempleton @drmorr0

olemarkus:
Could I get some feedback on this one please?

jaypipes (approved):
I think this is great, thank you @olemarkus!

@mwielgus do you have any concerns with this? If not, I'd love to merge it.

gjtempleton:
Hey @olemarkus, sorry, previously missed this.

As we can see with this PR suddenly having conflicts due to the updated vendoring of the k/k 1.26-rc.1 dependencies (#5336) we have a strong preference for sticking to the upstream dependencies at the top level of the repo, and vendoring cloud provider specific dependencies/drifts from these dependencies under the relevant cloud-provider.

Barring a strong reason to deviate from that I'd suggest that's the pattern we should follow here.

olemarkus:
Hey. I am a bit unsure what you mean here. cloud-provider-legacy is ... legacy. And the specific CCMs _should_ be considered _the_ upstream. The OP already describes a bug that comes from using the unmaintained legacy CCM implementations vs the actually maintained ones.

gjtempleton:
Sorry, to add a bit more context, the CA includes the entire scheduler code to perform scheduling simulation inside the CA, this means we currently include its entire dependencies, and update the repo's go.mod to the upstream using [this script](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/hack/update-vendor.sh) when updating the CA to support new versions of k8s.

This means pulling in new dependencies which aren't part of k/k's dependencies can result in significant version conflicts the next time we have to update the vendored deps. Therefore, our approach recently has been that any dependencies which aren't brought in with k/k should be vendored under the cloud provider making use of them. See the [AWS providers' vendoring of the go-aws-sdk](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/aws#using-the-aws-sdk-vendored-in-the-aws-cloudprovider) to allow use of a newer version than that used by k/k for an example.

It does look like only the AWS and GCE providers make use of the legacy-code-providers, so if we can move both to the new ones I'd think we should be able to shrink the dependencies we pull in at the top level might be able to shrink even further.

olemarkus:
Ah. That makes a lot more sense. But I have concerns:

This _can_ get interesting, with the CCMs having its own set of interesting dependencies: CCM depends on specific AWS SDK versions and a host of k8s.io libs on its own. So I am not sure this necessarily improves the situation. The CCM code itself probably has to be modified to use the cloud provider's AWS SDK version too. It may or may not work with the top-level/cloud provider-specific version.

I haven't looked too closely at what exactly CA need from CCM, but it _might_ be that it is simpler to copy over the functionality instead of vendoring (i.e copy functions/structs, not entire files).

gjtempleton:
Aha, yeah, that does sound like a real pain to potentially untangle.

From a quick look through, AWS and GCE are the only two cloudproviders importing the legacy-cloud-providers, though I've not dug deeply into whether they're using the same areas of functionality yet, so agreed that just copying over the required functionality may be a simpler solution.

@MaciekPytel do you have any thoughts?

x13n:
/assign @gjtempleton 

(Since you're already reviewing this.)

FWIW, I think the right way to handle deps in the long term is https://github.com/kubernetes/autoscaler/issues/5394

jayantjain93:
@olemarkus do you think we can expedite making this PR mergeable. This is currently blocking release of CA v1.27 https://github.com/kubernetes/autoscaler/issues/5625

gjtempleton:
Thanks.
/lgtm
/approve

PR #5148, opened 2022-08-30
update clusterapi nodegroupset processor

### LINKED ISSUE #5158: Cluster API provider should ignore labels for Alibaba Cloud, IBM Cloud, GKE, and Azure

cluster-autoscaler

the experience for balancing similar nodes on the various infrastructures requires the user to know about the labels and invoke the ignore label flag for each.

adding known labels to the cluster-api nodegroupset processor would make this experience much better for users.

we could create more documentation to expose these labels, and update the helm charts as well.

**Additional context.**:

[cluster-api](https://cluster-api.sigs.k8s.io/) is a project that creates an abstraction around the infrastructure provider resources. as such, it can provide access to many different clouds. this makes the cluster-api provider a little unique in that it will need to ignore more labels in its nodegroupset processor if we want to make a seamless experience for users.

elmiko:

cluster-autoscaler

#### What type of PR is this?

/kind bug

this change adds a few more known labels to the processor for
cluster-api, in specific it adds more support for IBM Cloud,
Alibaba Cloud, GCP, and Azure.

Fixes #5158

```release-note

```

```docs

```


enxebre:
Hey Mike, could you create an issue providing context on the problem the PR solves?

cluster-autoscaler/processors/nodegroupset/clusterapi_nodegroups.go
```diff
@@ -25,11 +25,13 @@ import (
 // even if they have different infrastructure provider-specific labels.
 func CreateClusterAPINodeInfoComparator(extraIgnoredLabels []string) NodeInfoComparator {
 	capiIgnoredLabels := map[string]bool{
-		"topology.ebs.csi.aws.com/zone":                 true, // this is a label used by the AWS EBS CSI driver as a target for Persistent Volume Node Affinity
-		"topology.diskplugin.csi.alibabacloud.com/zone": true, // this is a label used by the Alibaba Cloud CSI driver as a target for Persistent Volume Node Affinity
-		"ibm-cloud.kubernetes.io/worker-id":             true, // this is a label used by the IBM Cloud Cloud Controler Manager
+		"ibm-cloud.kubernetes.io/vpc-instance-id":       true, // used on IBM Cloud when a VPC is in use
```
enxebre, line 28: can we may be expose them from each providers and reference them here so we don't need to manually keep the list?

sbueringer:
@mdbooth I think there are similar labels for OpenStack (more details in the recording of the CAPI Office hours 31th August)

elmiko:
> Hey Mike, could you create an issue / update the PR desc providing context on the problem the PR solves?

absolutely, i will create one today

cluster-autoscaler/processors/nodegroupset/clusterapi_nodegroups.go
elmiko: i would be ok with modifying the other nodegroupset processors to expose the variables as constants there, would that work?

elmiko: created #5158

elmiko:
created #5158

cluster-autoscaler/processors/nodegroupset/clusterapi_nodegroups.go
elmiko: changed to use consts from other providers where applicable

elmiko:
update
* created const values for other providers ignored labels
* change clusterapi implementation to use consts
* added more labels to clusterapi processor

this update changed the structure here a little, but the core is still the same.

elmiko:
updated to fix linting errors

fabriziopandini:
@elmiko now that I better understand this better (thanks for the overview) I think we should lead cluster API users toward a set of recommendations that are driven by the less surprise principle, that IMO is "when using node-groups load balancing, please explicitly set the balancing-labels you want to use" because this avoids node-groups load balancing to be compromised if someone for any reason add a new label to nodes.
I this idea makes sense and gains consensus, then we can eventually stop maintaining this list

elmiko:
+1 @fabriziopandini 

for reference, we discussed this topic at a special meeting today, https://www.youtube.com/watch?v=jbhca_9oPuQ , and it seems like the community would like to go in the direction that @fabriziopandini is mentioning.

i am going to re-evaluate our work on the nodegroupset processor, most likely remove the cluster-api custom bits, and then propose improved documentation instead.

/hold

elmiko:
i am going to close this in favor of the option we discussed previously (removing the processor for capi)

PR #5165, opened 2022-09-05
correct label query

zouyee:
Signed-off-by: zounengren <zouyee1989@gmail.com>

vpa

#### What type of PR is this?
/kind bug

If we use query interface with `up{job="kubernetes-pods"}`, the returned result would be vector, but expected to get matric type, and we need the label of the pod corresponding to the ResourceHistory query, not the most recent label of the pod, so that we can get the label of the extinct pod.

Fixes #

```release-note
none
```

```docs
none
```

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
@@ -23,8 +23,6 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/klog/v2"
-
 	promapi "github.com/prometheus/client_golang/api"
 	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	prommodel "github.com/prometheus/common/model"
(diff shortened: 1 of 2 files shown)


vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
```diff
@@ -24,7 +24,7 @@ import (
 	"time"
 
 	"k8s.io/klog/v2"
-
+
```
jbartosik, line 27: Why remove this line?

vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
```diff
@@ -223,10 +223,18 @@ func (p *prometheusHistoryProvider) readResourceHistory(res map[model.PodID]*Pod
...
 
 func (p *prometheusHistoryProvider) readLastLabels(res map[model.PodID]*PodHistory, query string) error {
+	end := time.Now()
+	start := end.Add(-time.Duration(p.historyDuration))
+
 	ctx, cancel := context.WithTimeout(context.Background(), p.queryTimeout)
 	defer cancel()
 
-	result, _, err := p.prometheusClient.Query(ctx, query, time.Now())
+	result, _, err := p.prometheusClient.QueryRange(ctx, query, prometheusv1.Range{
```
jbartosik, line 232: I don't understand why changing from `Query` to `QueryRange` is correct

jbartosik (changes requested):
Please fix the failing test-and-veryfy step:

> Please run hack/update-gofmt.sh to fix the following files:
> ./vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go

Also

/hold

while I'm working on releasing VPA 0.12 (I'll remove the hold when VPA 0.12 is out)

PR #5166, opened 2022-09-05
Option to ignore pod extended resources

tombokombo:
autoscaler

#### What type of PR is this?
/kind feature

/kind api-change
If you try to use extended/custom resources as described in k8s https://kubernetes.io/docs/tasks/configure-pod-container/extended-resource/,  CA will not trigger scale up for pods with custom/extended resource because predicates will fail with Insufficient resource. 
This option introduce possibility to ignore extended/custom resource and let CA make calculation based on real resources like cpu, memory etc.

Fixes #

```release-note
A new option has been added to cluster-autoscaler. `ignore-pod-resources` allows autoscaler to ignore extended/custom resources defined on pod and let CA do scheduling calculations as there is no extended/custom resource defined.
```

```docs

```


x13n:
/assign

Extended resources aren't necessarily infinite. Ignoring them can therefore have an opposite effect: CA not scaling up, thinking the pod would fit (but it can't due to lack of extended resources). I think we need a way of injecting extended resources into in-memory template nodes CA is using for simulations. This is already something we have for GPU.

tombokombo:
@x13n this was kind of workaround, please take a look at proper solution https://github.com/kubernetes/autoscaler/pull/5214 
This PR could be closed in favor of https://github.com/kubernetes/autoscaler/pull/5214
thx

x13n:
Got it, thanks! Will take a look at the other one then.

/close

PR #5175, opened 2022-09-08
`skipped_scale_events_count` cluster-autoscaler metric add new reason and node_group label

### LINKED ISSUE #5122: `skipped_scale_events_count` not working as expected (metric never emits)
cluster-autoscaler

app version: 9.19.4
image version: 1.23.1

Component version:

$ kubectl version

</pre></details>
```
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.0", GitCommit:"c2b5237ccd9c0f1d600d3072634ca66cefdf272f", GitTreeState:"clean", BuildDate:"2021-08-04T18:03:20Z", GoVersion:"go1.16.6", Compiler:"gc", Platform:"darwin/arm64"}
Server Version: version.Info{Major:"1", Minor:"20+", GitVersion:"v1.20.15-eks-84b4fe6", GitCommit:"c813d094d1e9fa409079753049cfaa9a67524ed2", GitTreeState:"clean", BuildDate:"2022-06-09T18:20:31Z", GoVersion:"go1.15.15", Compiler:"gc", Platform:"linux/amd64"}
```

AWS + EKS (v1.20)

I expected to see the newly-added metric, `skipped_scale_events_count`, emitted @ the `/metrics` endpoint

**What happened instead?**:
`cluster-autoscaler` did not emit the metric, `skipped_scale_events_count`. 
The logs did not emit what I expected from [scale_up.go](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/scale_up.go#L431). 

1. Ensure `cluster-autoscaler` emits a log @ startup indicating it is running v1.23.1. 
2. Modify replicas for a pod that requests 1 CPU, 2Gi (request and limits equal) from 1 -> N (where N is very large)
  * This causes the `clutser-autoscaler` to hit the max # of nodes. As a result, many pods will remain unschedulable.
3. Observe logs for `cluster-autoscaler` 
4. Hit the `/metrics` endpoint of the `cluster-autoscaler` and observe that the expected metric, `skipped_scale_events_count`, is not present.

### Logs from `scale_up.go`: 
```
I0824 14:14:45.581353       1 scale_up.go:376] Upcoming 0 nodes
I0824 14:14:45.581636       1 scale_up.go:412] Skipping node group platform-dev-eks - max size reached
I0824 14:14:45.581650       1 scale_up.go:412] Skipping node group platform-dev-eks-ephemeral - max size reached
I0824 14:14:45.581655       1 scale_up.go:453] No expansion options
```
Seeing these logs indicates to me that `if currentTargetSize >= nodeGroup.MaxSize()` is the condition my `cluster-autoscaler` is hitting. 
Based on the description of `skipped_scale_events_count`, either: 
* I do not understand scale events or
* Emitting a different, new metric may be useful for when `currentTargetSize >= nodeGroup.MaxSize()`

Thanks in advance

sebradloff:

cluster-autoscaler

#### What type of PR is this?

/kind feature

This PR addresses issue #5122. It adds an additional reason `NodeGroupLimit` to the metric `skipped_scale_events_count`. This is helpful for creating alerts as a k8s cluster admin/operator that new nodes in a node group can't be scaled in because the max node group size has been reached or that the node group can't be further scaled down because the min node group size has been reached. To get this information today, you must monitor the node group metrics from the cloud provider itself. Simply monitoring the `unschedulable_pods_count` is not enough because there are several reasons that a pod could be unschedulable outside of max node size reached (wrong toleration or nonexistent nodeSelector).

This PR also adds the `node_group` label to the `skipped_scale_events_count` since the metric is closely tied to a specific node group(s), I figured it would be useful to know which it applies to.

Fixes #5122.

I'm not sure if we want to drop the `node_group` label from the metric or add any special logic to only emit that label when the `emitPerNodeGroupMetrics` flag is true.

```release-note
Adds `NodeGroupLimit` as a reason to the metric `skipped_scale_events_count`. This allows you to track if the cluster-autoscaler skip a scale up or down event due to the node group already having reached its maximum or minimum size. 
```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/scale_up.go
+++ b/cluster-autoscaler/core/scale_up.go
@@ -410,6 +410,7 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 		if currentTargetSize >= nodeGroup.MaxSize() {
 			klog.V(4).Infof("Skipping node group %s - max size reached", nodeGroup.Id())
 			skippedNodeGroups[nodeGroup.Id()] = maxLimitReachedReason
+			metrics.RegisterSkippedScaleUpNodeGroupMaxSize()
 			continue
 		}
 
--- a/cluster-autoscaler/core/scale_up.go
+++ b/cluster-autoscaler/core/scale_up.go
@@ -410,7 +410,7 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 		if currentTargetSize >= nodeGroup.MaxSize() {
 			klog.V(4).Infof("Skipping node group %s - max size reached", nodeGroup.Id())
 			skippedNodeGroups[nodeGroup.Id()] = maxLimitReachedReason
-			metrics.RegisterSkippedScaleUpNodeGroupMaxSize()
+			metrics.RegisterSkippedScaleUpNodeGroupMaxSize(nodeGroup.Id())
 			continue
 		}
 
@@ -434,9 +434,9 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 			for _, resource := range checkResult.exceededResources {
 				switch resource {
 				case cloudprovider.ResourceNameCores:
-					metrics.RegisterSkippedScaleUpCPU()
+					metrics.RegisterSkippedScaleUpCPU(nodeGroup.Id())
 				case cloudprovider.ResourceNameMemory:
-					metrics.RegisterSkippedScaleUpMemory()
+					metrics.RegisterSkippedScaleUpMemory(nodeGroup.Id())
 				default:
 					continue
 				}
--- a/cluster-autoscaler/core/scaledown/legacy/legacy.go
+++ b/cluster-autoscaler/core/scaledown/legacy/legacy.go
@@ -692,6 +692,7 @@ func (sd *ScaleDown) NodesToDelete(currentTime time.Time, pdbs []*policyv1.PodDi
 		if size-deletionsInProgress <= nodeGroup.MinSize() {
 			klog.V(1).Infof("Skipping %s - node group min size reached", node.Name)
 			sd.unremovableNodes.AddReason(node, simulator.NodeGroupMinSizeReached)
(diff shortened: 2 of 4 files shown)


elmiko (commented):
this generally looks great to me, the one issue is that we have traditionally avoided putting the node group id into metrics as it cause the cardinality of those metrics series to become unbounded.

if we start to add this information, there are many other places we could also add it. i would love to see us add some sort of flag to make these metrics even deeper, but the cost is that the user must be aware to increase the storage for their monitoring infrastructure.

@mwielgus or @MaciekPytel  happy to hear your input about this.

MaciekPytel:
@elmiko You're right, we'we had this discussion many times in the past. The problem with nodegroup ids in labels is that while it's great for monitoring single-clusters, it introduces major scalability issues once you have a fleet of hundreds or thousands of clusters and you want to monitor them in aggregate (both due to storage cost and the additional aggregation needed when querying). And because we're talking about an existing metric, this becomes a backward compatibility issue.

The compromise we've reached in the past was to avoid adding nodegroup id (or any other label that involves unique ids) to any existing metrics and instead add separate per-nodegroup metrics behind --emit-per-nodegroup-metrics flag. The obvious downside is some code duplication (though it probably could be handled entirely in metrics.go?) and the proliferation of seemingly similar metrics, but that seems like a reasonable price to pay for supporting both use-cases of single cluster and large fleet of cluster and maintaining backward compatibility.

elmiko:
> The compromise we've reached in the past was to avoid adding nodegroup id (or any other label that involves unique ids) to any existing metrics and instead add separate per-nodegroup metrics behind --emit-per-nodegroup-metrics flag. 

ooh, good reminder! i forgot we merged that XD

x13n:
Assigning, since looks like you are already reviewing this.

/assign @elmiko 

@sebradloff Are you planning to follow up on this?

sebradloff:
@x13n @elmiko Thanks for the comments. Sorry I've been busy and unfortunately don't really have the time right now to make the required changes.

PR #5176, opened 2022-09-08
API for KEP-4831: Control VPA eviction behavior based on scaling direction and resource

voelzmo:

vertical-pod-autoscaler

#### What type of PR is this?

/kind feature
/kind api-change

#### What this PR does / why we need it
Implements KEP-4831: Control VPA eviction behavior based on scaling direction and resource

API for #4831
Implementation is done in #5599

```release-note

```

```docs

```

### DIFF
--- a/vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml
+++ b/vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml
@@ -382,8 +382,8 @@ spec:
                   evictionRequirements:
                     description: EvictionRequirements is a list of EvictionRequirements
                       that need to evaluate to true in order for a Pod to be evicted.
-                      If more than one EvictionRequirement is specified, they are
-                      combined with AND.
+                      If more than one EvictionRequirement is specified, all of them
+                      need to be fulfilled to allow eviction.
                     items:
                       description: EvictionRequirement defines a single condition
                         which needs to be true in order to evict a Pod
@@ -400,7 +400,8 @@ spec:
                         resource:
                           description: Resources is a list of one or more resources
                             that the condition applies to. If more than one resource
-                            is given, they are combined with OR, not with AND.
+                            is given, the EvictionRequirement is fulfilled if at least
+                            one resource meets `changeRequirement`.
                           items:
                             description: ResourceName is the name identifying various
                               resources in a ResourceList.
--- a/vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml
+++ b/vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml
@@ -364,7 +364,7 @@ spec:
                     description: API version of the referent
                     type: string
                   kind:
-                    description: 'Kind of the referent; More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds"'
+                    description: 'Kind of the referent; More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
                     type: string
                   name:
                     description: 'Name of the referent; More info: http://kubernetes.io/docs/user-guide/identifiers#names'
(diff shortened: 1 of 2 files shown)


voelzmo:
While looking for the best place to put this into the Updater, I found https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/updater/priority/pod_eviction_admission.go which was pulled in years ago with https://github.com/kubernetes/autoscaler/pull/1002 but apparently never used afterwards. Beata (after merging) even commented that the default implementation isn't even used: https://github.com/kubernetes/autoscaler/pull/1002#pullrequestreview-130869322

I guess this is probably a good place to implement the filtering we want to do based on the comparison of `target` and `requests` and if no restrictions are given, just use the already implemented `noopPodEvictionAdmission`?

jbartosik:
> While looking for the best place to put this into the Updater, I found https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/updater/priority/pod_eviction_admission.go which was pulled in years ago with #1002 but apparently never used afterwards. Beata (after merging) even commented that the default implementation isn't even used: [#1002 (review)](https://github.com/kubernetes/autoscaler/pull/1002#pullrequestreview-130869322)
> 
> I guess this is probably a good place to implement the filtering we want to do based on the comparison of `target` and `requests` and if no restrictions are given, just use the already implemented `noopPodEvictionAdmission`?

Yes, adding a new type that implements `PodEvictionAdmission` and passing it to `Updater` seems to be the best way to implement this KEP.

I'd implement it so when there are no restrictions it just allows the pod

> 
> There's also https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go but that seems to be concerned with limiting eviction based on minimum available replicas and ensuring this also works for Jobs and other Pod owners in the same way.
> 
> Joachim WDYT, should I just make use of the PodEvictionAdmission linked above? From a flow perspective, this seems to be a good fit, [as it is used in the priority calculator](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go#L156), where Pod requests and target recommendation are compared to figure out if a Pod should be evicted and in which order.

Yes,

voelzmo:
@jbartosik Can you please take a look? Thanks!

voelzmo:
@jbartosik Can you try to give this a review, please? Thanks!

jbartosik (commented):
Since in-place pod updates are now finally a thing (alpha but I hope for beta soon) I'd like to make sure this works well with support for in place updates.

I can see two approaches:

- have one "actuationRequirements" that affect both in-place and eviction actuation
- have separate controls for "evictionRequirements" and "inPlaceUpdateRequirements"

I lean towards single "actuationRequirements" because it's simpler and I don't see how I'd use separate controls for eviction and in place update. What do you think?

vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml
```diff
@@ -3,7 +3,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     api-approved.kubernetes.io: https://github.com/kubernetes/kubernetes/pull/63797
-    controller-gen.kubebuilder.io/version: v0.4.0
```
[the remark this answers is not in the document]
jbartosik: > 

I sent https://github.com/kubernetes/autoscaler/pull/5754

jbartosik:
Let's decide if we want to change the API for in-place updates before merging this (https://github.com/kubernetes/autoscaler/pull/5754)

voelzmo:
Force-pushed an update:
* rebased on current master
* removed the `OrEqual` ChangeRequirements as discussed in https://github.com/kubernetes/autoscaler/pull/5754#issuecomment-1647577540

vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml
```diff
@@ -379,6 +379,38 @@ spec:
...
+                            current requests, what kind of change is necessary for
+                            the Pod to be evicted
+                          enum:
+                          - TargetHigherThanRequests
+                          - TargetLowerThanRequests
+                          type: string
+                        resource:
+                          description: Resources is a list of one or more resources
+                            that the condition applies to. If more than one resource
+                            is given, they are combined with OR, not with AND.
```
jbartosik, line 403:
```suggestion
                            is given, requirement is fulfilled if at least one resource meets `changeRequirement`
```

jbartosik (commented):
Looks good to me

@mwielgus @gjtempleton could you take a look please

vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml
voelzmo: done, thanks!

mwielgus (approved):
/lgtm
/approve

PR #3227, opened 2020-06-17
VPA: Allow limiting VPA deployment to a namespace

povilasv:
Allows running multiple VPA deployments, each deployment looking at only it's namespace, update it's lease in that namespace, etc.

Also for Prometheus it would limit queries to that namespace.

Fixes #3010

We are currently running this and it works great!

docker images available:
```
# Image runs https://github.com/kubernetes/autoscaler/pull/3227/files
  - name: vpa-updater
    newName: quay.io/utilitywarehouse/vpa-updater
    newTag: dev
  - name: vpa-admission-controller
    newName: quay.io/utilitywarehouse/vpa-admission-controller
    newTag: dev
  - name: vpa-recommender
    newName: quay.io/utilitywarehouse/vpa-recommender
    newTag: dev
```

Role bindings:
<details>
<pre>
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: vpa-actor
  namespace: telecom
rules:
  - apiGroups:
      - ""
    resources:
      - pods
      - limitranges
    verbs:
      - get
      - list
      - watch
  - apiGroups:
      - ""
    resources:
      - events
    verbs:
      - get
      - list
      - watch
      - create
  - apiGroups:
      - "poc.autoscaling.k8s.io"
    resources:
      - verticalpodautoscalers
    verbs:
      - get
      - list
      - watch
      - patch
  - apiGroups:
      - "autoscaling.k8s.io"
    resources:
      - verticalpodautoscalers
    verbs:
      - get
      - list
      - watch
      - patch
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: evictioner
  namespace: telecom
rules:
  - apiGroups:
      - "apps"
      - "extensions"
    resources:
      - replicasets
    verbs:
      - get
  - apiGroups:
      - ""
    resources:
      - pods/eviction
    verbs:
      - create
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: vpa-actor
  namespace: telecom
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: vpa-actor
subjects:
  - kind: ServiceAccount
    name: vpa-recommender
    namespace: telecom
  - kind: ServiceAccount
    name: vpa-updater
    namespace: telecom
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: vpa-target-reader
  namespace: telecom
rules:
[...]

PR #5177, opened 2022-09-08
Restrict Updater PodLister to namespace

voelzmo:

vertical-pod-autoscaler

#### What type of PR is this?

/kind bug
/kind cleanup

The Updater still lists Pods in all namespaces, even when VPA is configured to deal with a single namespace only. Seems like this was missed in https://github.com/kubernetes/autoscaler/pull/3227?

```release-note
Fixed a bug which caused VPA Updater to list Pods in all namespaces, even when VPA was configured to watch a single namespace only.
```


jbartosik (approved):
/lgtm
/approve

PR #5178, opened 2022-09-08
Update controller-gen to 0.9.2

voelzmo:

vertical-pod-autoscaler

#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it
Update controller-gen to 0.9.2 and install with `go install`, as `go get` doesn't install binaries anymore with go1.18

Fixes #

```release-note
NONE
```

### DIFF
--- a/vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml
+++ b/vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml
@@ -3,7 +3,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     api-approved.kubernetes.io: https://github.com/kubernetes/kubernetes/pull/63797
-    controller-gen.kubebuilder.io/version: v0.4.0
+    controller-gen.kubebuilder.io/version: v0.9.2
   creationTimestamp: null
   name: verticalpodautoscalercheckpoints.autoscaling.k8s.io
 spec:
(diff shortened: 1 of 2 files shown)


vertical-pod-autoscaler/hack/generate-crd-yaml.sh
```diff
@@ -32,8 +32,7 @@ trap cleanup EXIT
 if [[ -z $(which controller-gen) ]]; then
     (
         cd $WORKSPACE
-	      go mod init tmp
-	      go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.0
+	      go install sigs.k8s.io/controller-tools/cmd/controller-gen@latest
```
mwielgus, line 35: Can we get an explicit version here? Using latest is usually against best practices...

voelzmo: Absolutely correct, thanks! Done with the newest commit.

vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml
```diff
@@ -203,19 +203,13 @@ spec:
         type: object
     served: true
     storage: false
-status:
```
jbartosik, line 206: Do you know why regenerating deletes this?

voelzmo: I think this was accepted as a bug some time ago: https://github.com/kubernetes-sigs/controller-tools/issues/456 and fixed from [version 0.9.0](https://github.com/kubernetes-sigs/controller-tools/releases/tag/v0.9.0) with https://github.com/kubernetes-sigs/controller-tools/pull/630

PR #5209, opened 2022-09-23
Support for DirectX Devices

Flask:

Under Windows the handling for GPU does currently not support ScaleDown since the Node is not treated as Ready.
Other than on Linux the Resource is named `microsoft.com/directx`

#### What type of PR is this?
/kind feature

Allows GPU accelerated windows nodes to scale down


mwielgus (approved):
/lgtm
/approve

PR #5210, opened 2022-09-25
magnum: add an option to create insecure TLS connections

antonkurbatov:
We use self-signed certificates in the openstack for test purposes. It is not always easy to bring a CA certificate. And so we ran into the problem that there is no option to not check the validity of the certificate in the autoscaler.

This patch adds a new option for the magnum plugin: tls-insecure

Signed-off-by: Anton Kurbatov <Anton.Kurbatov@acronis.com>

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/magnum/magnum_openstack_clients.go
+++ b/cluster-autoscaler/cloudprovider/magnum/magnum_openstack_clients.go
@@ -108,6 +108,7 @@ type Config struct {
 		DomainName      string `gcfg:"domain-name"`
 		Region          string
 		CAFile          string `gcfg:"ca-file"`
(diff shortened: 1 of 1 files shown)


tghartland:
Hi @mwielgus 

On reading the code diff, this looks fine to me.

I'm not working at CERN any more and while I do have access to another openstack cloud, magnum isn't a major service in it (kubernetes 1.15 is the highest version available).

I'm happy to keep reviewing for small changes like this, but if there were any big magnum API changes that needed autoscaler work then at that point I'd ask someone at CERN to step up.

I hope that arrangement is acceptable.

mwielgus (approved):
/lgtm
/approve

PR #5192, opened 2022-09-15
[gce]: DeleteInstances 409 case

Freyert:
#### Which component this PR applies to?

cluster-autoscaler `gce` provider.

#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:
If the cluster autoscaler for some reason re-enters the `DeleteInstances` function after an operation to delete nodes has started the autoscaler will not correctly wait on the operation.

Instead, the autoscaler will loop over this error (from what I've seen) putting the cluster in a state where it can't be modified again until the delete operation finishes.

#### Special notes for your reviewer:

If this PR is merged it would close https://github.com/kubernetes/autoscaler/pull/5213

I expect https://github.com/kubernetes/autoscaler/pull/5213 is potentially the more correct PR so I don't imagine this one will be merged.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #5213, opened 2022-09-26
[gce]: skip instances on validation error

Freyert:

cluster-autoscaler `gce` provider.

#### What type of PR is this?
/kind bug

If the cluster autoscaler for some reason attempts to delete a cluster that is being deleted/abandoned/etc. the operation will fail. There may be a re-entry bug or maybe some external source has started deleting an instance.

The API will ignore these validation errors and allow the autoscaler to continue processing the other instances that need deletion/removal from the cluster.

Adds the following parameter to the GCP DeleteInstances call:

> skipInstancesOnValidationError
> Specifies whether the request should proceed despite the inclusion of instances that are not members of the group or that are already in the process of being deleted or abandoned. If this field is set to false and such an instance is specified in the request, the operation fails. The operation always fails if the request contains a malformed instance URL or a reference to an instance that exists in a zone or region other than the group's zone or region.
> https://cloud.google.com/compute/docs/reference/rest/v1/instanceGroupManagers/deleteInstances

This PR if chosen would supercede https://github.com/kubernetes/autoscaler/pull/5192 and close https://github.com/kubernetes/autoscaler/pull/5192.

```release-note
NONE
```

```docs

```


mwielgus (approved):
/lgtm
/approve

Issue #5006, opened 2022-07-05
Cluster Autoscaler doesn't scale ASG from 0 unless manually scaled first (and afterward works as expected).


ZTGallagher:
If we have at least once scaled up the ASG while the cluster autoscaler is installed, it can from then one scale the deployment to and from 0.

But if the ASG has never once been scaled up, it won't work.

Here is the IAM role permissions set for the cluster autoscaler.

```
{
    "Statement": [
        {
            "Action": [
                "ec2:DescribeLaunchTemplateVersions",
                "ec2:DescribeInstanceTypes",
                "autoscaling:DescribeTags",
                "autoscaling:DescribeLaunchConfigurations",
                "autoscaling:DescribeAutoScalingInstances",
                "autoscaling:DescribeAutoScalingGroups"
            ],
            "Effect": "Allow",
            "Resource": "*",
            "Sid": "eksAutoscalingAll"
        },
        {
            "Action": [
                "eks:DescribeNodegroup",
                "ec2:DescribeInstanceTypes",
                "autoscaling:TerminateInstanceInAutoScalingGroup",
                "autoscaling:SetDesiredCapacity"
            ],
            "Condition": {
                "StringEquals": {
                    "autoscaling:ResourceTag/k8s.io/cluster-autoscaler/enabled": "true",
                    "autoscaling:ResourceTag/kubernetes.io/cluster/use1-cluster": "owned"
                }
            },
            "Effect": "Allow",
            "Resource": "*",
            "Sid": "eksAutoscalingOwn"
        }
    ],
    "Version": "2012-10-17"
}
```

Tags on the ASG.

```
"k8s.io/cluster-autoscaler/node-template/label/nvidia.com/gpu": "true"
"k8s.io/cluster-autoscaler/node-template/resources/nvidia.com/gpu": "1"
```

The tags match the label and gpu the deployment looks for. 

I don't know how Cluster Autoscaler caches which ASG applies to which labels. But it works after at least one manual scale-up. That's not really a viable solution though. The whole deployment would be essentially hands-off without this issue.

dev-rowbot:
@ZTGallagher - have a look at my comment on #4998 - you are probably experiencing the same issue.

WillerWasTaken:
We're using eks_managed_node_groups and we got it working by setting the scaling options like this

```
scaling_config {
  desired_size = 1
  max_size     = 5
  min_size     = 0
}
```

[...]

PR #5214, opened 2022-09-26
Fix/asg resource tags

### LINKED ISSUE #5164: Extended resources provided by ASG via tags is not working
autoscaler

 1.25.0-alpha.0 AND 1.23.1

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.9", GitCommit:"6df4433e288edc9c40c2e344eb336f63fad45cd2", GitTreeState:"clean", BuildDate:"2022-04-13T19:57:43Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.9", GitCommit:"c1de2d70269039fe55efb98e737d9a29f9155246", GitTreeState:"clean", BuildDate:"2022-07-13T14:19:57Z", GoVersion:"go1.17.11", Compiler:"gc", Platform:"linux/amd64"}

</pre></details>

sanbox ( AWS )

I tried to use extended resource defined as tag in ASG according documentation for AWS it should be `k8s.io/cluster-autoscaler/node-template/resources/<resource-name>` https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-1.23.1/cluster-autoscaler/cloudprovider/aws/README.md#auto-discovery-setup . This sould work at least for node-group-auto-discovery mode. Is anybody successfully using it?

**What happened instead?**:
This tag is never read or ignored. CA is still complaining about insufficient resource and not scaling up.

Add extended resource as descirbe in https://kubernetes.io/docs/tasks/configure-pod-container/extended-resource/
Add  `k8s.io/cluster-autoscaler/node-template/resources/<resource-name>` tag with same reasonable value to ASG

I did some tests, it looks like this is never executed https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/aws_manager.go#L412

tombokombo:
autoscaler

#### What type of PR is this?
/kind bug

Problem is described in https://github.com/kubernetes/autoscaler/issues/5164
TLDR 
In [autoscaler docs](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md#auto-discovery-setup) is written
```From version 1.14, Cluster Autoscaler can also determine the resources provided by each Auto Scaling Group via tags. The tag is of the format k8s.io/cluster-autoscaler/node-template/resources/<resource-name>. <resource-name> is the name of the resource, such as ephemeral-storage. The value of each tag specifies the amount of resource provided. The units are identical to the units used in the resources field of a Pod specification.```

But I found no way how to make this work. There is long time relevant part of code but never executed. I'm using kops AWS cluster. So this PR reintroduce this functionality.

Fixes #5164

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_manager.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_manager.go
@@ -410,6 +410,7 @@ func (m *AwsManager) buildNodeFromTemplate(asg *asg, template *asgTemplate) (*ap
 	}
 
 	resourcesFromTags := extractAllocatableResourcesFromAsg(template.Tags)
+	klog.V(5).Infof("Extracted resources from ASG tags %v", resourcesFromTags)
 	for resourceName, val := range resourcesFromTags {
 		node.Status.Capacity[apiv1.ResourceName(resourceName)] = *val
 	}
--- a/cluster-autoscaler/cloudprovider/aws/aws_manager_test.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_manager_test.go
@@ -395,6 +395,8 @@ func TestBuildNodeFromTemplate(t *testing.T) {
 	// Node with custom resource
 	ephemeralStorageKey := "ephemeral-storage"
 	ephemeralStorageValue := int64(20)
+	customResourceKey := "custom-resource"
+	customResourceValue := int64(5)
 	vpcIPKey := "vpc.amazonaws.com/PrivateIPv4Address"
 	observedNode, observedErr := awsManager.buildNodeFromTemplate(asg, &asgTemplate{
 		InstanceType: c5Instance,
@@ -403,12 +405,19 @@ func TestBuildNodeFromTemplate(t *testing.T) {
 				Key:   aws.String(fmt.Sprintf("k8s.io/cluster-autoscaler/node-template/resources/%s", ephemeralStorageKey)),
 				Value: aws.String(strconv.FormatInt(ephemeralStorageValue, 10)),
 			},
+			{
+				Key:   aws.String(fmt.Sprintf("k8s.io/cluster-autoscaler/node-template/resources/%s", customResourceKey)),
+				Value: aws.String(strconv.FormatInt(customResourceValue, 10)),
+			},
 		},
 	})
 	assert.NoError(t, observedErr)
 	esValue, esExist := observedNode.Status.Capacity[apiv1.ResourceName(ephemeralStorageKey)]
 	assert.True(t, esExist)
 	assert.Equal(t, int64(20), esValue.Value())
+	crValue, crExist := observedNode.Status.Capacity[apiv1.ResourceName(customResourceKey)]
+	assert.True(t, crExist)
+	assert.Equal(t, int64(5), crValue.Value())
 	_, ipExist := observedNode.Status.Capacity[apiv1.ResourceName(vpcIPKey)]
 	assert.False(t, ipExist)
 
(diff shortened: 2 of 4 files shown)


tombokombo:
@gjtempleton are you planning to review this PR?

gjtempleton:
Hey @tombokombo, apologies, I largely use assignment to PRs as my filter for which PRs to review hence me missing it in the GH notifications noise before.

/assign @gjtempleton

tombokombo:
@gjtempleton still nothing?

gjtempleton:
Hey, I've tried to spend some time reproducing the issue this is meant to resolve this evening and haven't been able to reproduce this as long as I updated the existing nodes in the ASG to present the custom resource. Are you able to provide further instructions for reproducing the issue you're seeing?

For reference, I was using a custom resource under the name `example.com/customresource`.

paleozogt:
@gjtempleton If you have nodes running then the problem won't happen.  The issue is what autoscaler does when the ASG is at zero and has never been scaled up before.

gjtempleton:
OK, understood and now reproduced.

I'd like to get @x13n's take here, as if we merged this, we'd be at the point of having two cloud provider specific processors in core code, which I'm pretty wary about. I'd also like to ensure we have some tests for this functionality as we already have around the processing of other resources.

I'm wondering whether this also might be affecting any other cloud providers supporting specifying information like this for scale from zero?

paleozogt:
@gjtempleton I can't comment on this MR, but I'd like to note that this problem has been reported over and over (e.g., #3802, #5006, #5164, #5278, etc).  Its like the feature just wasn't implemented?

tombokombo:
@gjtempleton I've added some tests, but functionality was already there and tested with ephemeral-storage. Regarding implementation, auto-scaling group tags are aws specific, there is autoscaler aws specific documentation that custom-resources should work via tags. That is the reason why I made aws specific provider modifications. I've could put it into generic template-info-provider, but it could brake half of other cloudproviders that I'm not able to test. There is already specific processor for gcp, so I'm not breaking any pattern. This patch could fix aws specific issues mention by @paleozogt

x13n:
I think the way this is introduced into core is fine - we already are doing roughly the same thing with other cloud providers. I don't like this, but wouldn't block this PR on following an already established pattern. That being said, we should really look into better isolation between different cloud providers. https://github.com/kubernetes/autoscaler/issues/5394 I proposed some time ago was - in retrospect - too radical, but I think separating each cloud provider to have its own container image (and hence - own main.go) is the direction into which I'd go. I'm planning to update that issue with more details when I find a little bit of extra time.

So - no blockers from me, but leaving the lgtm and approval to people more familiar with aws.

tombokombo:
@x13n thanks for review. I thought the same when I first dig into the code, that it needs core library and separate cloudproviders code...anyway, @gjtempleton  its your turn now.

gjtempleton:
If you're happy with it as is (and addressing for the long term in the linked issue) then I'm happy to go with this approach.

/approve

tombokombo:
@gjtempleton ok, thx and who can put here lgtm?

gjtempleton:
Ah, sorry, my bad...
/lgtm

tombokombo:
@jbartosik i'm going to provide fix.

tombokombo:
@jbartosik fix here https://github.com/kubernetes/autoscaler/pull/5485

PR #5223, opened 2022-09-29
cluster-autoscaler: allow setting kuberentes client burst and qps to avoid rate limiting

grosser:

cluster-autoscaler

#### What type of PR is this?

/kind feature

```
CA Waited for 191.285161ms due to client-side throttling, not priority and fairness, request: POST:[https://172.29.4.1:443/api/v1/namespaces/kube-system/events
```

- we currently see lots of rate limiting from the kubernetes client of CA, we'd like to give it more QPS to avoid that
- users might want to reduce the number of queries that CA does against their cluster

```release-note
- add `--kube-client-burst` and `--kube-client-qps` to control how many requests CA can do against kubernetes
```

### DIFF
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -90,6 +90,8 @@ var (
 	address                 = flag.String("address", ":8085", "The address to expose prometheus metrics.")
 	kubernetes              = flag.String("kubernetes", "", "Kubernetes master location. Leave blank for default")
 	kubeConfigFile          = flag.String("kubeconfig", "", "Path to kubeconfig file with authorization and master location information.")
+	kubeClientBurst         = flag.Int("kube-client-burst", rest.DefaultBurst, "Burst value for kubernetes client.")
+	kubeClientQPS           = flag.Float64("kube-client-qps", float64(rest.DefaultQPS), "QPS value for kubernetes client.")
 	cloudConfig             = flag.String("cloud-config", "", "The path to the cloud provider configuration file.  Empty string for no configuration file.")
 	namespace               = flag.String("namespace", "kube-system", "Namespace in which cluster-autoscaler run.")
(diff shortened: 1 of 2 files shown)


x13n:
Should there be separate flags for events client and everything-else client? With this change as-is, the rate limiting will actually be 2x than whatever is configured via flags (assuming there's a lot of load on both events and non-events and rate limiting is the bottleneck). Maybe it would be cleaner to use separate flags for events client?

grosser:
ideally we'd use a single client, since atm it already uses 2x the configured rate, if we hardcode the "there are 2 different clients" into the cli that could be an issue later too

idk if the event or the regular client produces our throttling issues
also if a user sees throttling it would be nice if they only have to turn a single knob and not try 2 different ones

grosser:
ah the throttling is from `/eviction` so I could only apply it to non-events if that makes this PR merge-able

x13n:
Sorry, I just noticed I missed this one. Assigning so I won't miss it again:

/assign

I think setting the rate limits on a single (non-event) client makes sense. The clients were split on purpose so I think merging them back might be problematic. Perhaps nowadays with API priority and fairness it is doable, but I'd need to better understand the implications first. I'll look into this. Meanwhile, let's apply the limits to just one client.

grosser:
updated it as requested
... not a big fan of each having their own rate limit, but should be fine

cluster-autoscaler/main.go
```diff
@@ -317,7 +321,7 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 	}
 }
 
-func getKubeConfig() *rest.Config {
+func getKubeConfig(context config.AutoscalingOptions) *rest.Config {
```
x13n, line 324: `context` param is not used after the changes, so let's drop it.

grosser: done @x13n

Issue #3966, opened 2021-03-22
[BUG] VPA updater: index out of range durring vpa.applyPodLimitRange()


hjkatz:
**Which component are you using?**: vpa

**What version of the component are you using?**: v0.9.2

**What k8s version are you using (`kubectl version`)?**: 1.17.9

**What environment is this in?**: on-prem

**What did you expect to happen?**: No panic

**What happened instead?**:

```
I0322 16:13:35.702335       1 update_priority_calculator.go:133] not updating a short-lived pod customer-communication-hub/customer-communication-hub-5788cbbf6b-8lx8k, request within recommended range
panic: runtime error: index out of range [3] with length 3

goroutine 1 [running]:
k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/vpa.applyPodLimitRange(0xc01ba96fc0, 0x3, 0x4, 0xc01ca86040, 0x15e9123, 0x3, 0xc01fd69bc0, 0xc01fd69bf0, 0x0, 0x0, ...)
        /gopath/src/k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/vpa/capping.go:397 +0x111a
k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/vpa.(*cappingRecommendationProcessor).capProportionallyToPodLimitRange(0xc022e674a0, 0xc01ba96fc0, 0x3, 0x4, 0xc01ca86040, 0x451c41, 0x2139c6b, 0xfafdd, 0xfafdd, 0xc01edd2fa0)
        /gopath/src/k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/vpa/capping.go:423 +0x2ae
k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/vpa.(*cappingRecommendationProcessor).Apply(0xc022e674a0, 0xc01ee1b3e0, 0xc01ee1b3c0, 0xc01b55fd40, 0x1, 0x1, 0xc01ca86040, 0xc01fd44e80, 0xc01fd690b0, 0x0, ...)
        /gopath/src/k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/vpa/capping.go:71 +0xa2
[...]

PR #5232, opened 2022-10-04
E2e test admission pod recommendation mismatch

jbartosik:

VPA admission contoroller

#### What type of PR is this?

/kind bug

Add E2E tests for behavior described in https://github.com/kubernetes/autoscaler/issues/3966. Fix should make  test cases with limit range behave the way tests without limit range do now.

```release-note
NONE
```


jbartosik:
I ran tests in my cluster (whole admission-controller suite, including those I add in this PR) and they passed.

mwielgus (approved):
/lgtm
/approve

PR #5234, opened 2022-10-05
Add HTTP health-check to Addon Resizer 1.8

kgolab:

addon-resizer

#### What type of PR is this?

/kind feature

Adds a handler for HTTP-based health-check in Addon Resizer 1.8.

NOTE: The healthcheck.go is based on VPA code and simplified to fit Addon Resizer better.

By default the handler listens on port 8080, this can be changed with `healthcheck-address` flag (e.g. be setting it to `:8443`).

This allows to set up a liveness probe for the addon-resizer container, for example by adding following section to its yaml:
```
          livenessProbe:
            httpGet:
              path: /health-check
              port: 8080
            initialDelaySeconds: 10
            periodSeconds: 10
```

Should the binary get stuck (or at least 5 times slower than configured), it would be restarted by kubelet.

I'm not sure if this change should also be introduced on master - README says the recommended version is 1.8 but it seems a waste not to have this on master as well.
If yes - I will create a separate PR since some files moved between master and 1.8 branch.

```release-note
Added health check handler to allow configuring liveness Probe.

Added a new flag `healthcheck-address`, defaulting to ":8080".

No user action required.
```

```docs
Addon Resizer provides an HTTP-based health-check, by default on port 8080, under `/health-check` path.

The port (and IP the health-check handler binds to) can be configured with `healthcheck-address`. If not set it defaults to ":8080".

The check fails if the main reconciliation loop lasts 5 times the configured period (`poll-period` flag).

Using the health-check handler, a liveness probe can be configured for Addon Resizer, for example by adding the following section to container definition:
-- cut here --
          livenessProbe:
            httpGet:
              path: /health-check
              port: 8080
            initialDelaySeconds: 10
            periodSeconds: 10
-- cut here --
```

### DIFF
--- a/addon-resizer/healthcheck/healthcheck.go
+++ b/addon-resizer/healthcheck/healthcheck.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package healthcheck
+
+import (
+	"fmt"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+)
+
+// HealthCheck contains information about last activity time of the monitored component.
+// NOTE: This started as a simplified version of VPA's HealthCheck.
+type HealthCheck struct {
+	// configuration
+	address         string
+	activityTimeout time.Duration
+	// current state
+	mutex        sync.Mutex
+	lastActivity time.Time
+}
+
+// NewHealthCheck builds new HealthCheck object with given timeout.
+func NewHealthCheck(address string, activityTimeout time.Duration) *HealthCheck {
+	return &HealthCheck{
+		address:         address,
+		activityTimeout: activityTimeout,
+		mutex:           sync.Mutex{},
+		lastActivity:    time.Now(),
+	}
+}
+
+// Serve sets up healthCheck handler on the given address
+func (hc *HealthCheck) Serve() {
+	go func() {
+		http.Handle("/health-check", hc)
+		err := http.ListenAndServe(hc.address, nil)
+		glog.Fatalf("Failed to start health check: %v", err)
+	}()
+}
+
+// checkLastActivity returns true if the last activity was too long ago, with duration from it.
+func (hc *HealthCheck) checkLastActivity() (bool, time.Duration) {
+	hc.mutex.Lock()
+	defer hc.mutex.Unlock()
+
+	now := time.Now()
+	lastActivity := hc.lastActivity
+	timedOut := now.After(lastActivity.Add(hc.activityTimeout))
+
+	return timedOut, now.Sub(lastActivity)
+}
+
+// ServeHTTP implements http.Handler interface to provide a health-check endpoint.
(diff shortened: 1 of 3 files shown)


kgolab:
@jbartosik or @krzysied , could you please have a look?

addon-resizer/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,90 @@
...
+
+// HealthCheck contains information about last activity time of the monitored component.
+// NOTE: This started as a simplified version of VPA's HealthCheck.
+type HealthCheck struct {
+	// configuration
+	address         string
+	activityTimeout time.Duration
+	// current state
+	lastActivity time.Time
+	mutex        *sync.Mutex
```
krzysied, line 36: nit: I would switch mutex with lastActivity.
In k8s there is often added a mutex and all variables guarded by it are below it.

krzysied, line 36: This mutex will not be shared. Imho we can sore it by value.

addon-resizer/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,90 @@
...
+	now := time.Now()
+	lastActivity := hc.lastActivity
+	timedOut := now.After(lastActivity.Add(hc.activityTimeout))
+
+	return timedOut, now.Sub(lastActivity)
+}
+
+// ServeHTTP implements http.Handler interface to provide a health-check endpoint.
+func (hc *HealthCheck) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	timedOut, ago := hc.checkLastActivity()
```
krzysied, line 72: I think that golangish way would be:
```
if timedOut, ago := ... ; timedOut {
  http.Error(...)
  return
}

w.WriteHeader...
```

addon-resizer/nanny/nanny_lib.go
```diff
@@ -99,10 +100,11 @@ type ResourceEstimator interface {
 // PollAPIServer periodically counts the number of nodes, estimates the expected
 // ResourceRequirements, compares them to the actual ResourceRequirements, and
 // updates the deployment with the expected ResourceRequirements if necessary.
-func PollAPIServer(k8s KubernetesClient, est ResourceEstimator, pollPeriod, scaleDownDelay, scaleUpDelay time.Duration, threshold uint64) {
+func PollAPIServer(k8s KubernetesClient, est ResourceEstimator, hc *healthcheck.HealthCheck, pollPeriod, scaleDownDelay, scaleUpDelay time.Duration, threshold uint64) {
 	lastChange := time.Now()
 	lastResult := noChange
 
+	hc.Serve()
```
krzysied, line 107: I would move it to main.go

krzysied (commented):
All in all it looks good. There are a few small comments.

addon-resizer/healthcheck/healthcheck.go
kgolab: Switched (here & below).

addon-resizer/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,89 @@
...
+	timedOut := now.After(lastActivity.Add(hc.activityTimeout))
+
+	return timedOut, now.Sub(lastActivity)
+}
+
+// ServeHTTP implements http.Handler interface to provide a health-check endpoint.
+func (hc *HealthCheck) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if timedOut, ago := hc.checkLastActivity() ; timedOut {
+		http.Error(w, fmt.Sprintf("Error: last activity more than %v ago (threshold is %v)", ago, hc.activityTimeout), http.StatusInternalServerError)
+	} else {
```
krzysied, line 74: Do we need this else instead of having a return in the first section?

diranged:
@kgolab,
  This is great .. but just for my own clarity, this does not fail out if the Pod Nanny is failing to do its job right? For example, if a permissions issue prevents it from making changes, this will not help us catch that? What do you think about a ReadinessProbe that would help us catch a failure like that?

PR #5235, opened 2022-10-07
Add option to wait for a period of time after node tainting/cordoning

### LINKED ISSUE #3886: Race condition leading to undesired scale nodes of non-empty nodes (without pod eviction)
**Which component are you using?**: Cluster-autoscaler

**What version of the component are you using?**: 1.18.1

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.11-dispatcher", GitCommit:"2e298c7e992f83f47af60cf4830b11c7370f6668", GitTreeState:"clean", BuildDate:"2019-09-19T22:26:40Z", GoVersion:"go1.11.13", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.15-eks-ad4801", GitCommit:"ad4801fd44fe0f125c8d13f1b1d4827e8884476d", GitTreeState:"clean", BuildDate:"2020-10-20T23:27:12Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What environment is this in?**: AWS

**What happened?**:
CA lists the nodes, finds an empty node, and decides to scale it down.

The race condition occurs when Kubernetes schedules a new pod on the node after CA made the API call to list all the nodes, but before the node is tainted as Unschedulable. 

The consequence is that CA removes the node even though the condition for scale down isn't met. The newly placed pod isn't evicted, because CA only evicts pod it knows about (and here it thinks the node is empty).

I expect CA wouldn't remove the node in this scenario. If it did remove it, I would expect that it evicts the newly placed pod.

**Potential fixes**:
1. After a node is tainted as Unschedulable and ready for removal, make a new API call to list the node and verify that the condition for scale down is met.
2. Or if we can't do this, it would be nice if all pods were evicted, which we could do by making a new API call after the node is tainted as Unschedulable, to retrieve the freshest list of pods to evict.
3. If we don't want to add a new API call, then another option could be to delete a node over 2 executions of the main loop. During the first execution, we tain the node as unschedulable / mark it for deletion. But it's only during the second execution (if the conditions are still met), that we trigger its effective deletion.

**How to reproduce**
In our setting, there's about a 1-second time window where this race condition is possible. This race conditions happens a few times a day (we schedule a lot of short-lived etl applications, order of 10k pods a day). I'm not sure how to reproduce this in a test environment.

alex-matei:

cluster-autoscaler

#### What type of PR is this?

/kind bug

This PR addresses a race condition between kube-scheduler and cluster-autoscaler that happens when scaling down nodes. The solution is to taint all the nodes, wait for a period of time and recheck that nodes can safely be drained and deleted. 

Fixes #3886

```release-note
Add `--node-delete-delay-after-taint` flag that controls how long it waits for after tainting a node, before rechecking and deleting the node.
```

### DIFF
--- a/cluster-autoscaler/core/scaledown/actuation/actuator.go
+++ b/cluster-autoscaler/core/scaledown/actuation/actuator.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	apiv1 "k8s.io/api/core/v1"
+	policyv1 "k8s.io/api/policy/v1"
 	"k8s.io/klog/v2"
 
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
@@ -29,12 +30,16 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/context"
 	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown"
 	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown/deletiontracker"
+	"k8s.io/autoscaler/cluster-autoscaler/core/utils"
 	"k8s.io/autoscaler/cluster-autoscaler/metrics"
 	"k8s.io/autoscaler/cluster-autoscaler/processors/status"
+	"k8s.io/autoscaler/cluster-autoscaler/simulator"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator/utilization"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/deletetaint"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/gpu"
+	kube_util "k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
+	"k8s.io/kubernetes/pkg/scheduler/framework"
 )
 
 // Actuator is responsible for draining and deleting nodes.
@@ -44,17 +49,19 @@ type Actuator struct {
 	nodeDeletionTracker *deletiontracker.NodeDeletionTracker
 	nodeDeletionBatcher *NodeDeletionBatcher
 	evictor             Evictor
+	deleteOptions       simulator.NodeDeleteOptions
 }
 
 // NewActuator returns a new instance of Actuator.
-func NewActuator(ctx *context.AutoscalingContext, csr *clusterstate.ClusterStateRegistry, ndr *deletiontracker.NodeDeletionTracker, batchInterval time.Duration) *Actuator {
-	nbd := NewNodeDeletionBatcher(ctx, csr, ndr, batchInterval)
+func NewActuator(ctx *context.AutoscalingContext, csr *clusterstate.ClusterStateRegistry, ndr *deletiontracker.NodeDeletionTracker, deleteOptions simulator.NodeDeleteOptions) *Actuator {
+	nbd := NewNodeDeletionBatcher(ctx, csr, ndr, ctx.NodeDeletionBatcherInterval)
 	return &Actuator{
 		ctx:                 ctx,
 		clusterState:        csr,
 		nodeDeletionTracker: ndr,
 		nodeDeletionBatcher: nbd,
-		evictor:             NewDefaultEvictor(),
+		evictor:             NewDefaultEvictor(deleteOptions),
+		deleteOptions:       deleteOptions,
 	}
 }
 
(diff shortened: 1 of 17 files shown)


cluster-autoscaler/core/scaledown/actuation/drain.go
```diff
@@ -245,11 +243,7 @@ func evictPod(ctx *acontext.AutoscalingContext, podToEvict *apiv1.Pod, isDaemonS
 	return status.PodEvictionResult{Pod: podToEvict, TimedOut: true, Err: fmt.Errorf("failed to evict pod %s/%s within allowed timeout (last error: %v)", podToEvict.Namespace, podToEvict.Name, lastError)}
 }
 
-func podsToEvict(ctx *acontext.AutoscalingContext, nodeName string) (dsPods, nonDsPods []*apiv1.Pod, err error) {
-	nodeInfo, err := ctx.ClusterSnapshot.NodeInfos().Get(nodeName)
-	if err != nil {
-		return nil, nil, err
-	}
+func podsToEvict(ctx *acontext.AutoscalingContext, nodeInfo *framework.NodeInfo) (dsPods, nonDsPods []*apiv1.Pod, err error) {
```
x13n, line 246: Looks like this cannot really return an error anymore, can you update the function signature?

cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -197,17 +217,90 @@ func (a *Actuator) deleteAsyncDrain(drain []*apiv1.Node) (scaledDownNodes []*sta
...
+		}
+
+		if drain {
+			if _, _, _, err := simulator.GetPodsToMove(nodeInfo, a.ctx.ListerRegistry, pdbs, time.Now()); err != nil {
+				nodeDeleteResult := status.NodeDeleteResult{ResultType: status.NodeDeleteErrorInternal, Err: errors.NewAutoscalerError(errors.InternalError, "GetPodsToMove for %q returned error: %v", node.Name, err)}
+				CleanUpAndRecordFailedScaleDownEvent(a.ctx, node, groupIds[i], drain, a.nodeDeletionTracker, "failed to get pods to move on node", nodeDeleteResult)
+				continue
+			}
+		} else {
+			podsToRemove, _, _, err := simulator.FastGetPodsToMove(nodeInfo, true, true, nil, time.Now())
```
x13n, line 284: `FastGetPodsToMove` and `DetailedGetPodsToMove` aren't really different: they used to be different long time ago before informers were used to keep cached lists of kubernetes objects up to date (the detailed version used to make API calls back then). I'd just use a single call for both empty and drained here to simplify the code a little.

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
```diff
@@ -1150,15 +1144,21 @@ func generatePods(count int, prefix string) []*apiv1.Pod {
...
 }
 
-func generatePod(name string) *apiv1.Pod {
+func generatePod(name string, node string) *apiv1.Pod {
 	return &apiv1.Pod{
-		ObjectMeta: metav1.ObjectMeta{Name: name},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+			Annotations: map[string]string{
+				"cluster-autoscaler.kubernetes.io/safe-to-evict": "true",
```
x13n, line 1157: Why add  this annotation on all test pods? Doesn't think reduce scope of the testing?

cluster-autoscaler/core/scaledown/actuation/drain_test.go
```diff
@@ -63,24 +63,13 @@ func TestDaemonSetEvictionForEmptyNodes(t *testing.T) {
 		{
 			name:              "Successful attempt to evict DaemonSet pods",
 			dsPods:            []string{"d1", "d2"},
-			nodeInfoSuccess:   true,
```
x13n, line 66: nit: `nodeInfoSuccess` shold also be removed from the struct above.

x13n (changes requested):
Thanks for taking the time to prepare this PR!

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
alex-matei: A pod that is not replicated and doesn't have a safe-to-evict annotation will block deletion when checking in GetPodsForDeletionOnNodeDrain. https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/utils/drain/drain.go#L211

cluster-autoscaler/simulator/drain.go
```diff
@@ -107,3 +119,16 @@ func checkPdbs(pods []*apiv1.Pod, pdbs []*policyv1.PodDisruptionBudget) (*drain.
...
 }
+
+// GetPodsToMove returns a list of pods that should be moved elsewhere
+// and a list of DaemonSet pods that should be evicted if the node
+// is drained. Raises error if there is an unreplicated pod.
+// Based on kubectl drain code. It checks whether RC, DS, Jobs and RS that created these pods
+// still exist. It takes into account the flags passed in command line
+// for skip-nodes-with-local-storage, skip-nodes-with-system-pods
+// and min-replica-count
+func GetPodsToMove(nodeInfo *schedulerframework.NodeInfo, listers kube_util.ListerRegistry,
```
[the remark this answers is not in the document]
alex-matei: I moved all the flags to main.go and passed them via a new struct NodeDrainOptions.

cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -17,24 +17,34 @@ limitations under the License.
...
+	"k8s.io/autoscaler/cluster-autoscaler/simulator"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator/utilization"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/deletetaint"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/gpu"
+	"k8s.io/kubernetes/pkg/scheduler/framework"
+)
+
+var (
+	nodeDeleteDelayAfterTaint = flag.Duration("node-delete-delay-after-taint", time.Duration(0),
```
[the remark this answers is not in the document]
alex-matei: Done.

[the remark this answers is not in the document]
alex-matei: Correct, i was thinking initially to keep existing behavior. I set it to 5 seconds, seems reasonable for most use cases.

cluster-autoscaler/core/scaledown/actuation/drain.go
alex-matei: Fixed.

cluster-autoscaler/core/scaledown/actuation/actuator.go
alex-matei: Fixed as well, changed to call only DetailedGetPodsToMove.

cluster-autoscaler/core/scaledown/actuation/drain_test.go
alex-matei: Fixed.

cluster-autoscaler/core/utils/test_utils.go
```diff
@@ -32,7 +32,7 @@ func GetStringFromChan(c chan string) string {
 	select {
 	case val := <-c:
 		return val
-	case <-time.After(100 * time.Millisecond):
+	case <-time.After(2 * time.Second):
```
x13n, line 35: nit: Why bump this? Won't it increase the time it takes to run tests?

cluster-autoscaler/simulator/drain.go
```diff
@@ -29,56 +29,36 @@ import (
...
 // and a list of DaemonSet pods that should be evicted if the node
 // is drained. Raises error if there is an unreplicated pod.
-// Based on kubectl drain code. It checks whether RC, DS, Jobs and RS that created these pods
+// Based on kubectl drain code. If listers is nil it makes an assumption that RC, DS, Jobs and RS were deleted
+// along with their pods (no abandoned pods with dangling created-by annotation).
+// If listers is not nil it checks whether RC, DS, Jobs and RS that created these pods
 // still exist.
-func DetailedGetPodsForMove(nodeInfo *schedulerframework.NodeInfo, skipNodesWithSystemPods bool,
-	skipNodesWithLocalStorage bool, listers kube_util.ListerRegistry, minReplicaCount int32,
+func GetPodsForMove(nodeInfo *schedulerframework.NodeInfo, deleteOptions NodeDeleteOptions, listers kube_util.ListerRegistry,
```
x13n, line 50: nit: I just realized these functions were called DetailedGetPods**For**Move and FastGetPods**To**Move, didn't even notice that before :) I'm not an English native speaker, but GetPods**To**Move you initially used sounds more natural to me, WDYT about renaming?

cluster-autoscaler/simulator/cluster_test.go
```diff
@@ -51,14 +51,19 @@ func TestFindPlaceAllOk(t *testing.T) {
 		"n1": true,
 		"n2": true,
 	}
+	deleteOptions := NodeDeleteOptions{
```
x13n, line 54: nit: since all test cases here use identical `NodeDeleteOptions`, it might make sense to just wrap them into some `testDeleteOptions()` func instead of defining separately everywhere.

x13n (commented):
Thanks! Overall LGTM, just a couple of minor comments.

cluster-autoscaler/core/utils/test_utils.go
alex-matei: I set NodeDeleteDelayAfterTaint to 1second in newWrapperForTesting and some tests time out with the previous wait time. You are correct, it will increase the time needed for tests so I changed it to 0 seconds.

cluster-autoscaler/simulator/drain.go
alex-matei: GetPodsToMove sounds better to me(although i'm not a native speaker). I changed it.

alex-matei:
@x13n I forgot to change TestGetPodsForMove to TestGetPodsToMove. I fixed it.

mwielgus (approved):
/lgtm
/approve

Issue #5237, opened 2022-10-07
VPA - Ensure that CPU recommendation is a multiple of N


dbenque:
**Which component are you using?**:

Vertical-Pod-Autoscaler/recommender

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:

Some of our applications size their internal worker pools as a factor of the number of vCPUs, and require these numbers of workers to be a multiple of 2/4/8 for easier load-balancing / re-sharding

So it is important for us that the VPA apply a recommendation for CPU that respects that constraint "being a multiple of N"

**Describe the solution you'd like.**:

Just like for #5212 and #5236 , before having user feedback on the feature and structuring something in the vpa.spec.resourcePolicy, we can use an annotation on the VPA object that would apply per container, something like:

`vpa-post-processor.kubernetes.io/{containerName}_CPU-multiple-of={value}`
for example:
`vpa-post-processor.kubernetes.io/kafka_CPU-multiple-of=2`

Implementation:
Today the VPA recommendation is amended to stay within the minAllowed/maxAllowed
We should be able to introduce more post-processing capabilities in the recommender.

Example:
The recommender internally delivers a recommendation of 3.85 CPU for container kafka, the recommendation that is written in the VPA object should be 4 CPU if the user has specified vpa-post-processor.kubernetes.io/kafka_CPU-multiple-of=2

**Describe any alternative solutions you've considered.**:

User has to manually compute the target value based on the recommendation and apply it in its yaml :(

Issue #5236, opened 2022-10-07
VPA - Support of Static CPU Allocation


dbenque:
**Which component are you using?**:

Vertical-Pod-Autoscaler

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:

Feature: as a user of the [Static CPU allocation](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy) I would like VPA to ensure that CPU value is set as an integer. As a user I should be able to specify if the recommendation should be round/ceil/floor to become an integer and respect static CPU allocation requirement, from the doc:
```
Only containers that are both part of a Guaranteed pod and have integer CPU requests are assigned exclusive CPUs.
```

**Describe the solution you'd like.**:

The user should be able to define which container should have its CPU recommendation amended to be an integer.
To start the API can simply be an annotation on the VPA object:
`vpa-post-processor.kubernetes.io/kafka_integerCPU=“round”`
Later when we have more user feedback this annotation can become a structured field in the `vpa.Spec.ResourcePolicy`

Implementation:
Today the VPA recommendation is amended to stay within the `minAllowed/maxAllowed`
We should be able to introduce more post-processing capabilities in the recommender.

Example:
The recommender internally delivers a recommendation of 3.85 CPU for container `kafka`, the recommendation that is written in the VPA object should be 4 CPU if the user has specified `vpa-post-processor.kubernetes.io/kafka_integerCPU=“round”`

**Describe any alternative solutions you've considered.**:

Write an entire dedicated custom recommender. I started to do that and realised that I was duplicating 90% of the default recommender code. Plus it will really be an heavy task to operate 2 different recommenders for a small difference in behavior.

**Additional context.**:

The issue #5212 can be handled in the exact same way from an implementation point of view.

jbartosik:
If we implemented #5237 then you could set N=1 and you would get this behavior. Or is there a difference I missed?

Issue #5212, opened 2022-09-26
VPA - The ratio between CPU and memory should be maintained


dbenque:
**Which component are you using?**: vertical-pod-autoscaler

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:
On applications using garbage collected memory, the increase of CPU could be the consequence of an intensive activity of the GC. This happens in Go and Java when using some memory primitive and settings like [SetMemoryLimit](https://pkg.go.dev/runtime/debug#SetMemoryLimit) or -Xmx. The CPU would increase but the memory will stay close to the limit. The value given to limit the memory usage is usually a significant part of what the application can read from the CGroup. 
For this reason it is interesting to grow the Memory in the same proportion than the CPU.

**Describe the solution you'd like.**:

Be able to constrain the VPA to maintain the ratio between the CPU and the Memory. In a more generic and general way allow the user to define a constraint so that the ratio between resource Type A and resource Type B is maintained (same proportion than in the initial user request and limit).

API Proposition: Add one more field in the `ContainerResourcePolicy` that would contain the definition for the ratio to be maintained: 
`maintainRatios *[][2]v1.ResourceName`

For example a user will be able to define:
```
maintainRatios: {{"cpu","memory"}}
```
In that case the `memory` is calculated based on the `cpu` recommendation, by applying the original ratio on the pod spec.

Since the user we be able to add multiple contrains we will have to ensure that the set of constrains can be represented as a set of direct acyclic graph. For example, the validation would reject because it introduce a cycle:
```
maintainRatios: {{"cpu","memory"},{"memory","storage"},{"storage","cpu"}}
```

If this feature is used, the resources measurements will be ignored for all the resources that are not root in the graph. For example with:
```
maintainRatios: {{"cpu","memory"}}
```
the measurements of `memory` will be ignored, the value will be calculated.

**Describe any alternative solutions you've considered.**:

If the memory is set as described in the feature description, the VPA is not giving any acceptable result with the application. So the application owner has to detect and understand the case and manually bump the memory request.

**Additional context.**:

We would like to use this feature at Datadog on top of several applications. We are happy to come and contribute to the project to integrate that feature if it makes sense for the community.

jbartosik:
I'd like to make sure if I understand the issue correctly.

The goal here is to support Java / Golang applications (or at least some of their configurations). And keeping constant memory / CPU ratio is idea for how to reach that goal?

[...]

PR #5239, opened 2022-10-07
[vpa] introduce recommendation post processor

dbenque:

vertical-pod-autoscaler/recommender

#### What type of PR is this?

/kind feature

Not an end user feature, but more a code feature: allow developper to plug recommendation post-processors in any recommender.

Today the recommender computes a recommendation and just before writing the recommendation to the VPA object, it does a small post-processing: Capping (apply policy for minAllowed/maxAllowed).

There are many other post processing cases that we can think about. Look at the following feature request should give good ideas of want we could do:
#5212 #5236 #5237 

During discussions with the SIG for #5212 I was proposed to implement it using a Custom Recommender. Starting the coding exercise in this direction I noticed that I was duplicating the entire default recommender and just changing 2 lines to introduce a call to another post processor, just before calling the existing Capping post processor.
On top of that the solution with the custom recommender would lead me to operate 2 recommenders. Will I do a 3rd one for a 3rd post processor ? ... obviously that does not sound like a good approach: difficult to maintain, difficult to operate, code duplication, non reusable code

Defining a clear interface for post-processor also allows us to share post-processors between multiple recommenders in the future. (See custom recommender KEP)

This PR does not fix any issue, it introduces a software architecture that would make it really easy to add features like the one requested in #5212 #5236 #5237 . Basically any feature that looks like recommendation post processing will be easy to introduce with this new code interface. Also it makes easy to test every post-processors independently of the recommender.

The order of post-processors is predefined (command-line param) and stable. In the future if users express the need it will completely be possible to expose the post-processors order to end users via a specific field in the VPA object. For the moment that sounds like a premature optimization.

When we will introduce new post processors, their behaviours will be piloted by the content of the VPA object. Just like the Capping post-processor is driven by the fields minAllowed/maxAllowed, other post processors will be driven by dedicated fields in the VPA object. When a new post-processor is introduced these fields can be simple annotations. With time and gain in maturity for the feature, the fields can be moved to the spec.

No

- [Other doc]: [detailed design doc](https://docs.google.com/document/d/1PtqdHL8qzuA6swXT3sHwDsEMaTfgvd4CnKV0BpHTqCg/edit#heading=h.y2hjd0r60y1p)


jbartosik (commented):
I left a few small comments on code.

But I think more important is that I think that a different approach would easier (both to use and to implement).

Instead of having one flag that specifies list of post-processors to use (what this PR does) we could have bool flag for each recommendation post-processor.

Then:
- each flag has it's own documentation so it's easy to get list of all post-processors and what each post-processor does,
- it's easier to set correct flag values (you don't have to worry if post processors are in a good order, typos are easier to find),
- it's easier to implement (when adding a post processor add a flag and an `if`).

What do you think?

dbenque:
> I left a few small comments on code.
> 
> But I think more important is that I think that a different approach would easier (both to use and to implement).
> 
> Instead of having one flag that specifies list of post-processors to use (what this PR does) we could have bool flag for each recommendation post-processor.
> 
> Then:
> 
> * each flag has it's own documentation so it's easy to get list of all post-processors and what each post-processor does,
> * it's easier to set correct flag values (you don't have to worry if post processors are in a good order, typos are easier to find),
> * it's easier to implement (when adding a post processor add a flag and an `if`).
> 
> What do you think?

I agree that would be easier to maintain. I am making the change, with "--post-processor-capping=true" by default to have same behaviour than today by default.

vertical-pod-autoscaler/pkg/recommender/logic/recommender.go
```diff
@@ -147,3 +149,31 @@ func CreatePodResourceRecommender() PodResourceRecommender {
 		lowerBoundEstimator,
 		upperBoundEstimator}
 }
+
+// MapToListOfRecommendedContainerResources converts the map of RecommendedContainerResources into a stable sorted list
+// This can be used to get a stable sequence while ranging on the data
+func MapToListOfRecommendedContainerResources(resources RecommendedPodResources) *vpa_types.RecommendedPodResources {
```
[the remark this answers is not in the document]
dbenque: I initially placed it here because it is taking parameter of that same package.
It is a conversion function between a type of `logic` and a type of our api `v1` aka `vpa_types`, so placing it in yet another package `routines` looks strange to me. Later as a developper, if I am looking for a conversion function, this is here in `logic` I would search in a first place.
To be consistent, what I would propose is to move the test function to the `logic` package. In fact the test was initially written in the `routines` package because that function was test tested through the calling function `getCappedRecommendation` and the test is really only about this conversion function.
WDYT?

vertical-pod-autoscaler/pkg/recommender/routines/recommendation_post_processor.go
```diff
@@ -0,0 +1,83 @@
...
+		case Capping:
+			processors = append(processors, &cappingPostProcessor{})
+		default:
+			return nil, fmt.Errorf("unknown Post Processor: %s", name)
+		}
+	}
+	return processors, nil
+}
+
+// ParsePostProcessors parses a comma separated list of post processor names, validate it,
```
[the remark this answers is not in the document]
dbenque: code was removed following the changes on the flag.

vertical-pod-autoscaler/pkg/recommender/main.go
```diff
@@ -82,7 +85,17 @@ func main() {
 	metrics_quality.Register()
 
 	useCheckpoints := *storage != "prometheus"
-	recommender := routines.NewRecommender(config, *checkpointsGCInterval, useCheckpoints, *vpaObjectNamespace, *recommenderName)
+
+	var postProcessorsNames []routines.KnownPostProcessors
+	if *postProcessorCapping {
+		postProcessorsNames = append(postProcessorsNames, routines.Capping)
```
jbartosik, line 91: Since we will have a bool flag for each post processor I think we can have a function that returns a list of post processors based on flags instead of the factory

vertical-pod-autoscaler/pkg/recommender/main.go
```diff
@@ -57,6 +57,9 @@ var (
 	ctrPodNameLabel     = flag.String("container-pod-name-label", "pod_name", `Label name to look for container pod names`)
 	ctrNameLabel        = flag.String("container-name-label", "name", `Label name to look for container names`)
 	vpaObjectNamespace  = flag.String("vpa-object-namespace", apiv1.NamespaceAll, "Namespace to search for VPA objects and pod stats. Empty means all namespaces will be used.")
+
+	// list of post processors flag
```
jbartosik, line 61: this comment is out of date

vertical-pod-autoscaler/pkg/recommender/main.go
```diff
@@ -57,6 +57,9 @@ var (
...
+	postProcessorCapping = flag.Bool("post-processor-capping", true, "Enable 'capping' post processor: apply minAllowed/maxAllowed range for the recommendation")
```
jbartosik, line 62: Do we expect a situation where one would tun off capping? If no then I think we shouldn't allow doing something that we expect might only give bad results

jbartosik (changes requested):

vertical-pod-autoscaler/pkg/recommender/main.go
dbenque: I would say that for the moment, there is no valid reason for unplugging capping.
With post-processors being introduced, we can have at some point use another post processor that would do capping with a different algorithm... at that time only we will have a valid reason for activating the other post-processor and removing that one.

I am removing the flag. We will re-introduce it later if it makes sense.

vertical-pod-autoscaler/pkg/recommender/main.go
dbenque: removed with the flag

vertical-pod-autoscaler/pkg/recommender/logic/recommender_test.go
```diff
@@ -115,3 +117,53 @@ func TestControlledResourcesFilteredDefault(t *testing.T) {
...
+		},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			namespace := "test-namespace"
+			vpa := model.NewVpa(model.VpaID{Namespace: namespace, VpaName: "my-vpa"}, labels.Nothing(), time.Unix(0, 0))
+			vpa.UpdateRecommendation(MapToListOfRecommendedContainerResources(tc.resources))
+			// Check that the slice is in the correct order.
+			for i := range vpa.Recommendation.ContainerRecommendations {
+				assert.Equal(t, tc.expectedLast[i], vpa.Recommendation.ContainerRecommendations[i].ContainerName)
```
jbartosik, line 165: We could iterate over both index and item

```
for i, outRecommendation := range vpa.Recommendation.ContainerRecommendations {
```

and use the item in the assertion

```
assert.Equal(t, tc.expectedLast[i], outRecommendation)
```

jbartosik, line 165: Please add assertion checking that content of the recommendation didn't change

vertical-pod-autoscaler/pkg/recommender/logic/recommender_test.go
```diff
@@ -115,3 +117,53 @@ func TestControlledResourcesFilteredDefault(t *testing.T) {
...
+func TestMapToListOfRecommendedContainerResources(t *testing.T) {
+	cases := []struct {
+		name         string
+		resources    RecommendedPodResources
+		expectedLast []string
+	}{
+		{
+			name: "All recommendations sorted",
+			resources: RecommendedPodResources{
+				"a-container": RecommendedContainerResources{Target: model.Resources{model.ResourceCPU: model.CPUAmountFromCores(1), model.ResourceMemory: model.MemoryAmountFromBytes(1000)}},
```
jbartosik, line 130: Please use different recommendations for different containers so that the test could tell if we were changing which recommendation is for which container

vertical-pod-autoscaler/pkg/recommender/logic/recommender_test.go
```diff
@@ -115,3 +117,53 @@ func TestControlledResourcesFilteredDefault(t *testing.T) {
...
+				"b-container": RecommendedContainerResources{Target: model.Resources{model.ResourceCPU: model.CPUAmountFromCores(1), model.ResourceMemory: model.MemoryAmountFromBytes(1000)}},
```
jbartosik, line 131: Please use realistic resource values (mostly memory >100MiB); I've missed some bugs because my tests were using unrealistically small values (and at realistic values things happened a bit differently) (since here we're just copying values I don't think it's likely to happen but code gets copied)

jbartosik (changes requested):
I left a few small comments.

One bigger comment remains open.

Since we're going with a set of bool flags (instead of the initial approach of a  flag with list of post processors) I think we shouldn't build a string and pass it to the  factory which will build a list of postprocessors. Instead we can do something significantly simpler and build list of postprocessors directly from bool flags.

vertical-pod-autoscaler/pkg/recommender/main.go
voelzmo: Does that mean that we're fixing the order in which the post-processors are applied in this place? This can make a difference, something like the capping post-processor should always be applied last, I guess?

vertical-pod-autoscaler/pkg/recommender/routines/recommender.go
```diff
@@ -80,6 +78,7 @@ type recommender struct {
 	podResourceRecommender        logic.PodResourceRecommender
 	useCheckpoints                bool
 	lastAggregateContainerStateGC time.Time
+	recommendationPostProcessor   []RecommendationPostProcessor
```
voelzmo, line 81:
```suggestion
	recommendationPostProcessors   []RecommendationPostProcessor
```
This should probably be plural?

vertical-pod-autoscaler/pkg/recommender/routines/recommendation_post_processor.go
```diff
@@ -0,0 +1,66 @@
...
+	var processors []RecommendationPostProcessor
+	for _, name := range f.PostProcessorsNames {
+		switch name {
+		case Capping:
+			processors = append(processors, &cappingPostProcessor{})
+		default:
+			return nil, fmt.Errorf("unknown Post Processor: %s", name)
+		}
+	}
+	klog.Infof("List of recommendation post-processors: %v", processors)
```
voelzmo, line 64:
```suggestion
	klog.V(4).Infof("List of configured recommendation post-processors: %v", processors)
```

I think we should guard this with a higher verbosity?

vertical-pod-autoscaler/pkg/recommender/routines/recommendation_post_processor.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// RecommendationPostProcessorsBuilder helps for the creation of the pre-processors list
+type RecommendationPostProcessorsBuilder interface {
+	// Build returns the list of post processors or an error
+	Build() ([]RecommendationPostProcessor, error)
+}
+
+// RecommendationPostProcessorFactory helps to build processors by their name
+// using a struct and not a simple function to hold parameters that might be necessary to create some post processors in the future
```
voelzmo, line 47: Do you have a specific example in mind which parameters could be needed and how they would be passed into this struct? Probably this whole struct can go away, see also @jbartosik's comment
> Instead we can do something significantly simpler and build list of postprocessors directly from bool flags.

dbenque:
> I left a few small comments.
> 
> One bigger comment remains open.
> 
> Since we're going with a set of bool flags (instead of the initial approach of a flag with list of post processors) I think we shouldn't build a string and pass it to the factory which will build a list of postprocessors. Instead we can do something significantly simpler and build list of postprocessors directly from bool flags.

I have remove the postprocessors factory. I initially introduced it because it could be useful to allow for user dynamic configuration. Maybe that was a premature optimisation, we can reintroduce it later if needed.

vertical-pod-autoscaler/pkg/recommender/routines/recommendation_post_processor.go
dbenque: removed with removal of factory requested by Joachim

vertical-pod-autoscaler/pkg/recommender/main.go
jbartosik: Yes. I think that we should fix order of postprocessors unless we have a good reason to allow changing it. I expect it would be easy to make mistake in order of postprocessors but hard to make things better by changing order from what we come up with.

jbartosik (approved):
/lgtm

/aprrove

PR #5244, opened 2022-10-10
add a log message for cluster unreadiness

### LINKED ISSUE #5172: cluster-autoscaler: "Cluster is not ready for autoscaling" needs to print reasons

cluster-autoscaler

Atm logs only show "Cluster is not ready for autoscaling" and nothing else.

Say what is going on / why the cluster is not ready.
"3 out of 10 nodes had a taint" or something like that

drmorr0:
cluster autoscaler

#### What type of PR is this?
/kind feature

Prints a more detailed log message when the cluster is unhealthy

Fixes #5172 

```release-note
NONE
```

n/a

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -353,6 +353,7 @@ func (csr *ClusterStateRegistry) IsClusterHealthy() bool {
 
 	if totalUnready > csr.config.OkTotalUnreadyCount &&
 		float64(totalUnready) > csr.config.MaxTotalUnreadyPercentage/100.0*float64(len(csr.nodes)) {
+		klog.Warningf("%d/%d nodes are unready - cluster not healthy", totalUnready, csr.config.OkTotalUnreadyCount)
 		return false
 	}
 


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -353,6 +353,7 @@ func (csr *ClusterStateRegistry) IsClusterHealthy() bool {
 
 	if totalUnready > csr.config.OkTotalUnreadyCount &&
 		float64(totalUnready) > csr.config.MaxTotalUnreadyPercentage/100.0*float64(len(csr.nodes)) {
+		klog.Warningf("%d/%d nodes are unready - cluster not healthy", totalUnready, csr.config.OkTotalUnreadyCount)
```
mwielgus, line 356: This log should not happen here but somewhere in the caller. Function name suggests that it can be called without side effects and we don't want to log warnings on every call to this method (unless the method failed to complete due to incorrect state/missing information).

mwielgus (changes requested):

cluster-autoscaler/clusterstate/clusterstate.go
drmorr0: @mwielgus my concern here is that the caller doesn't have the information needed to log.  Right now, the only reason a cluster can be unhealthy is if too many nodes are unready, but [the caller](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/static_autoscaler.go#L370) doesn't have access to that info.  In theory, perhaps sometime in the future there may be other reasons why a cluster might be unhealthy that we'd want to log, as well.

I can think of a few possible solutions:

1. Change the function name to something like `ValidateClusterHealth` and leave the log line here.
2. Change the return signature to `(bool, error)`, and if the cluster is unhealthy, return `false, fmt.Errorf("some reason")`
3. Do a larger refactor so this function returns an enum with the reason why the cluster is unhealthy.

Do any of those seem reasonable to you?

x13n: I'd go for `(bool, error)` version or even just `error` one, since `bool` will be `true` iff `error` will be `nil`.

x13n:
Closing since it doesn't seem to be worked on. Please reopen if that changes.

/close

PR #5247, opened 2022-10-11
Add support for extended resource definition in GCE MIG template

zaymat:

cluster-autoscaler

#### What type of PR is this?

/kind feature

This PR adds support for extended resources in GCE MIG template. 

Today, the `cluster-autoscaler` on GCE only supports scaling decisions based the following resources: CPU, Memory, EphemeralStorage and GPU. However, Kubernetes allows defining an arbitrary number of resources through the [Extended Resource API](https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/). This can be useful when your instances have special resources that you want to share between your pods. One example that I can think of is `network bandwidth`, which cannot be requested by pods except through extended resources.
Some other implementations of the `CloudProvider` interface, like AWS or Azure, already support scaling decisions based on extended resources.

This PR adds the possibility to define extended resources for a node group on GCE, so that the cluster-autoscaler can account for them when taking scaling decisions. This is done through the `extended_resources` key inside the AUTOSCALER_ENV_VARS variable set on a MIG template. 

Example:
```
AUTOSCALER_ENV_VARS: kube_reserved=<...>;<...>;extended_resources=foo=10,bar=1M,foobar=2G
```

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/templates.go
+++ b/cluster-autoscaler/cloudprovider/gce/templates.go
@@ -72,7 +72,7 @@ func (t *GceTemplateBuilder) getAcceleratorCount(accelerators []*gce.Accelerator
 
 // BuildCapacity builds a list of resource capacities given list of hardware.
 func (t *GceTemplateBuilder) BuildCapacity(cpu int64, mem int64, accelerators []*gce.AcceleratorConfig, os OperatingSystem, osDistribution OperatingSystemDistribution, arch SystemArchitecture,
-	ephemeralStorage int64, ephemeralStorageLocalSSDCount int64, pods *int64, version string, r OsReservedCalculator) (apiv1.ResourceList, error) {
+	ephemeralStorage int64, ephemeralStorageLocalSSDCount int64, pods *int64, version string, r OsReservedCalculator, extendedResources apiv1.ResourceList) (apiv1.ResourceList, error) {
 	capacity := apiv1.ResourceList{}
 	if pods == nil {
 		capacity[apiv1.ResourcePods] = *resource.NewQuantity(110, resource.DecimalSI)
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/cloudprovider/gce/templates_test.go
```diff
@@ -186,6 +187,23 @@ func TestBuildNodeFromTemplateSetsResources(t *testing.T) {
 			attachedLocalSSDCount:         4,
 			expectedErr:                   false,
 		},
+		{
+			scenario:                      "extended_resources present in kube-env",
+			kubeEnv:                       "AUTOSCALER_ENV_VARS: kube_reserved=cpu=0,memory=0,ephemeral-storage=10Gi;os_distribution=cos;os=linux;ephemeral_storage_local_ssd_count=2;extended_resources=someResource=2,anotherResource=1G\n",
```
x13n, line 192: Please also add some tests for handling malformed `extended_resources` (e.g. `extended_resources=foo`, `extended_resources=foo=bar`).

cluster-autoscaler/cloudprovider/gce/templates.go
```diff
@@ -215,10 +221,16 @@ func (t *GceTemplateBuilder) BuildNodeFromTemplate(mig Mig, template *gce.Instan
 		return nil, fmt.Errorf("could not fetch ephemeral storage from instance template: %v", err)
 	}
 
-	capacity, err := t.BuildCapacity(cpu, mem, template.Properties.GuestAccelerators, os, osDistribution, arch, ephemeralStorage, ephemeralStorageLocalSsdCount, pods, mig.Version(), reserved)
+	extendedResources, err := extractExtendedResourcesFromKubeEnv(kubeEnvValue)
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch extended resources from instance template: %v", err)
```
x13n, line 226: I don't think this should break TemplateNode building. Perhaps just logging an error would be sufficient? Having a TemplateNode with just the basic resources sounds better than not having it at all due to a typo on in `extended_resources` variable definition.

x13n (changes requested):

cluster-autoscaler/cloudprovider/gce/templates.go
zaymat: Sounds legit 👍 I'll make the change.

cluster-autoscaler/cloudprovider/gce/templates_test.go
zaymat: Sound good 👍 

I've added a test case to verify that malformed extended resources won't break the template building function, as well as more test case around the extraction function itself.

x13n:
Oh, actually I missed the fact this only touches GCE cloudprovider specific files, so I an approve it myself.

/approve

PR #5254, opened 2022-10-18
add optional key for deployment configMapRef

jebbens:
This PR adds the `optional` property for the configMapRef in the `deployment.yaml` chart so that users can easily specify it in their `values.yaml` file.

Fixes #

```release-note
Added the optional property to the configMapRef in the deployment spec for cluster-autoscaler. The value can be specified using the envFromConfigMapOptional variable in the values file.
```

```docs

```

### DIFF
--- a/charts/cluster-autoscaler/values.yaml
+++ b/charts/cluster-autoscaler/values.yaml
@@ -199,6 +199,7 @@ extraEnvSecrets: {}
 
 # envFromConfigMap -- ConfigMap name to use as envFrom.
 envFromConfigMap: ""
+envFromConfigMapOptional: true
 
 # envFromSecret -- Secret name to use as envFrom.
 envFromSecret: ""
--- a/charts/cluster-autoscaler/templates/deployment.yaml
+++ b/charts/cluster-autoscaler/templates/deployment.yaml
@@ -201,6 +201,7 @@ spec:
           {{- if .Values.envFromConfigMap }}
             - configMapRef:
                 name: {{ .Values.envFromConfigMap }}
+                optional: {{ .Values.envFromConfigMapOptional }}
           {{- end }}
           {{- end }}
           livenessProbe:


mwielgus (approved):
/lgtm
/approve

charts/cluster-autoscaler/README.md
```diff
@@ -316,6 +316,7 @@ Though enough for the majority of installations, the default PodSecurityPolicy _
 | deployment.annotations | object | `{}` | Annotations to add to the Deployment object. |
 | dnsPolicy | string | `"ClusterFirst"` | Defaults to `ClusterFirst`. Valid values are: `ClusterFirstWithHostNet`, `ClusterFirst`, `Default` or `None`. If autoscaler does not depend on cluster DNS, recommended to set this to `Default`. |
 | envFromConfigMap | string | `""` | ConfigMap name to use as envFrom. |
+| envFromConfigMapOptional | boolean | `true` | Whether or not the ConfigMap in envFrom is optional. |
```
gjtempleton, line 319: @jebbens This takes its description from the comment about the variable in values.yaml via `helm-docs` (see our [pre-commit config](https://github.com/kubernetes/autoscaler/tree/master/charts#pre-commit-hooks)), can you update the values.yaml to reflect this explanation? This is the reason the PR check is currently failing.

jebbens: Pushed a new commit - I think this is what you are looking for

charts/cluster-autoscaler/values.yaml
```diff
@@ -200,6 +200,9 @@ extraEnvSecrets: {}
 # envFromConfigMap -- ConfigMap name to use as envFrom.
 envFromConfigMap: ""
 
+# envFromConfigMapOptional -- Whether or not the ConfigMap in envFrom is optional
+envFromConfigMapOptional: true
```
gjtempleton, line 204: Is there a good reason to change the default behaviour here? 

Unless I'm mistaken the current behaviour prior to this is for the envFromConfigMap to be required if referenced.

gjtempleton:
Hey @jebbens, happy to give this another review if you can address the merge conflict and my question.

charts/cluster-autoscaler/values.yaml
jebbens: What I've found handy about using the "optional" property is that it allows you to use the same manifest for different environments where the ConfigMap only comes into play if it's applied to the cluster. For example, where one environment may have a proxy, which requires a ConfigMap with the proxy env variables, but another environment may not have a proxy so the ConfigMap isn't needed or applied.

gjtempleton: Thanks for the extra detail, I'm not against adding the option, my question was more around the defaulting of the new option to true, which changes the existing behaviour. I'd rather preserve the current behaviour but allow those who want to change it to override it when adding new functionality.

gjtempleton:
Closing for now. Please re-open the PR if you have the time to address the comments and conflicts.

PR #5257, opened 2022-10-19
upgrading mount utils

Bryce-Soghigian:

Fixes #

```release-note

```

```docs

```


marwanad:
based

x13n:
Master branch depends on 1.26 k8s libs, so switching back to 1.21 doesn't seem like the right thing to do. Since 1.21 is already well behind k8s support window, I don't think we should be doing any fixes to that release branch either.

/close

PR #5265, opened 2022-10-20
Add scaledown.Actuator to AutoscalingContext

BigDarkClown:

cluster-autoscaler

#### What type of PR is this?

/kind cleanup

This PR adds scaledown.Actuator reference to the AutoscalingContext. This will allow processors to have access to node-deleting capabilities, enabling more advanced implementations.

NONE


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -23,6 +23,7 @@ import (
 
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	scale_down_status "k8s.io/autoscaler/cluster-autoscaler/core/scaledown/status"
```
x13n, line 26: nit: `scaledownstatus`? I know some imports use underscores, even in this file, but I find it inconsistent with camelCase convention for naming variables.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -168,6 +169,8 @@ func NewStaticAutoscaler(
 	ndt := deletiontracker.NewNodeDeletionTracker(0 * time.Second)
 	scaleDown := legacy.NewScaleDown(autoscalingContext, processors, ndt, deleteOptions)
 	actuator := actuation.NewActuator(autoscalingContext, clusterStateRegistry, ndt, deleteOptions)
+	autoscalingContext.ScaleDownActuator = actuator
```
x13n, line 172: It's a bit weird to have the field in `AutoscalingContext`, but populate it when creating `StaticAutoscaler`. Would it be possible to untangle the dependencies somehow? I'm not sure how feasible that is without a bigger refactor though.

cluster-autoscaler/core/static_autoscaler.go
BigDarkClown: Yeah, I was on the fence about this. I'll change it to scaledownstatus for consistency sake.

cluster-autoscaler/core/static_autoscaler.go
BigDarkClown: I don't think it is without a large refactor. To do this we would need to create `Actuator` before creating `AutoscalingContext`. `Actuator` would need to get passed `CloudProvider`, `AutoscalingOptions`, `AutoscalingKubeClients`, `AutoscalingOptions` and `ClusterSnapshot` unpacked from the `AutoscalingContext` wrapper. This looks like a lot of refactor and IMHO would actually make this less readable than it is now.

`AutoscalingContext` is also initialised in this method, so it is not that far from the definition. Also we are already passing `ScaleDownWrapper` in a similar fashion two lines below - it would be nice to refactor both, which require even more changes.

towca: Agreed it's not good now, but at least the confusion is mostly scoped to `NewStaticAutoscaler`. Could you add a TODO for the bigger refactor somewhere in there? And/or somewhere around `AutoscalingContext` definition? Maybe the next person adding something could pick the refactor up.

towca, line 172: Hmm, it's really confusing to have this implementation in `AutoscalingContext`, and a different wrapped one in `StaticAutoscaler`, especially since they have the exact same name. Could we make `StaticAutoscaler` use the direct version from `AutoscalingContext` and wrap it internally/on demand somehow?

@x13n Do I understand correctly that after we switch to the parallel scale-down planner in 1.26 no wrapping would be needed since `StaticAutoscaler` could just use the direct actuator implementation from the context without any wrapping? If so, maybe we could just add a TODO to make this change after we switch the planner?

x13n: Adding TODO makes sense. Yes, we'll be able to use actuator from context eventually, but not so quickly - in 1.26/1.27 both implementations will be available (one will be selected based on a feature flag).

PR #5268, opened 2022-10-24
VPA: Fix typo, move service accounts to RBAC

joelsmith:

vertical-pod-autoscaler

#### What type of PR is this?

/kind cleanup

This moves updater and recommender service accounts to the RBAC manifests where the admission controller's is.  That keeps the deployments separate from the RBAC needed for the deployments to work.  Also fixes a small harmless typo in the name of a role binding.

N/A

```release-note
The VPA ClusterRoleBinding `system:vpa-evictionter-binding` has been renamed `system:vpa-evictioner-binding`. After applying the new RBAC manifest, the old one can be removed with `kubectl delete clusterrolebinding system:vpa-evictionter-binding`
```

### DIFF
--- a/vertical-pod-autoscaler/hack/generate-crd-yaml.sh
+++ b/vertical-pod-autoscaler/hack/generate-crd-yaml.sh
@@ -53,4 +53,5 @@ resources:
 commonAnnotations:
   "api-approved.kubernetes.io": "https://github.com/kubernetes/kubernetes/pull/63797"
 EOF
-kubectl kustomize . > ${OUTPUT}
+echo --- > ${OUTPUT}
+kubectl kustomize . >> ${OUTPUT}
--- a/vertical-pod-autoscaler/deploy/recommender-deployment.yaml
+++ b/vertical-pod-autoscaler/deploy/recommender-deployment.yaml
@@ -1,10 +1,4 @@
 ---
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: vpa-recommender
-  namespace: kube-system
----
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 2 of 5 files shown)


vertical-pod-autoscaler/hack/generate-crd-yaml.sh
```diff
@@ -53,4 +53,5 @@ resources:
 commonAnnotations:
   "api-approved.kubernetes.io": "https://github.com/kubernetes/kubernetes/pull/63797"
 EOF
-kubectl kustomize . > ${OUTPUT}
+echo --- > ${OUTPUT}
+kubectl kustomize . >> ${OUTPUT}
```
jbartosik, line 57: Why those changes?

jbartosik (changes requested):
This needs release notes:

- We're changing name of the ClusterRoleBinding, people updating might want to clean up the old one
- We're moving service accounts it might be more important to apply changes in correct order now

vertical-pod-autoscaler/hack/generate-crd-yaml.sh
joelsmith: Apparently kustomize doesn't put the `---` at the top of the first YAML document when there are more than one. Most of the other YAML files in `deploy` have the leading `---` so I thought to add it to make them more consistent, although it's optional on the first document.  I'm happy to remove it if it seems hacky.

joelsmith:
@jbartosik thanks for the review. I have added a release notes for the rename. It seems unnecessary to add one about the service accounts, at least to me. If the manifests are applied out of order and if the deployment manages to create the pod before the service account is created, the pod will just be in an error state just as it would if the rest of the RBAC hadn't been applied and the recommender or updater lacked permissions to do its job. And as soon as the RBAC is applied, the next pod restart should succeed. That being said, if you think I should add one, I'm happy to do so.

jbartosik (approved):
/lgtm /approve

jbartosik (approved):
/lgtm

PR #5274, opened 2022-10-26
[CA] Register recently evicted pods in NodeDeletionTracker.

olagacek:

cluster-autoscaler
#### What type of PR is this?

/kind feature

Part of parallel scale down implementation.

### DIFF
--- a/cluster-autoscaler/core/scaledown/actuation/drain.go
+++ b/cluster-autoscaler/core/scaledown/actuation/drain.go
@@ -51,22 +51,28 @@ const (
 	DefaultDsEvictionRetryTime = 3 * time.Second
 )
 
+type evictionRegister interface {
+	RegisterEviction(*apiv1.Pod)
+}
+
 // Evictor can be used to evict pods from nodes.
 type Evictor struct {
 	EvictionRetryTime          time.Duration
 	DsEvictionRetryTime        time.Duration
 	DsEvictionEmptyNodeTimeout time.Duration
 	PodEvictionHeadroom        time.Duration
+	evictionRegister           evictionRegister
 	deleteOptions              simulator.NodeDeleteOptions
 }
 
 // NewDefaultEvictor returns an instance of Evictor using the default parameters.
-func NewDefaultEvictor(deleteOptions simulator.NodeDeleteOptions) Evictor {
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/core/scaledown/actuation/drain.go
```diff
@@ -27,6 +27,7 @@ import (
 	policyv1beta1 "k8s.io/api/policy/v1beta1"
 	kube_errors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown/deletiontracker"
```
x13n, line 30: I would consider dropping dependency on deletion tracker and just accepting a simple local interface in `NewDefaultEvictor`.

x13n (commented):
Thanks for the change! Just two minor comments. I don't think they are blocking, so LGTMing the change and putting on hold for you to consider.

/lgtm
/hold

cluster-autoscaler/core/scaledown/actuation/drain.go
olagacek: to double check I understood it correctly: you would like me to introduce a new local interface with RegisterEviction function (implemented by node deletion tracker)?

olagacek: created new interface

cluster-autoscaler/core/scaledown/actuation/drain_test.go
```diff
@@ -441,8 +442,8 @@ func TestDrainNodeWithPodsEvictionFailure(t *testing.T) {
 	}
 	ctx, err := NewScaleTestAutoscalingContext(options, fakeClient, nil, nil, nil, nil)
 	assert.NoError(t, err)
-
-	evictor := Evictor{EvictionRetryTime: 0, PodEvictionHeadroom: DefaultPodEvictionHeadroom}
+	ndt := deletiontracker.NewNodeDeletionTracker(DefaultPodEvictionHeadroom)
```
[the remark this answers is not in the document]
olagacek: removed

PR #5284, opened 2022-10-31
Check owner reference in scale down planner to avoid double-counting

olagacek:
Check owner reference in scale down planner to avoid double-counting already deleted and recreated pods.

cluster-autoscaler
#### What type of PR is this?

/kind feature

This PR makes sure that in scale down planner we do not account twice for the pods that have been evicted and already recreated by its higher level controller.

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/scaledown/planner/controller.go
+++ b/cluster-autoscaler/core/scaledown/planner/controller.go
@@ -1,0 +1,88 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package planner
+
+import (
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
+)
+
+type controllerCalculatorImpl struct {
+	listers kubernetes.ListerRegistry
+}
+
+func newControllerReplicasCalculator(listers kubernetes.ListerRegistry) controllerReplicasCalculator {
+	return &controllerCalculatorImpl{listers: listers}
+}
+
+func (c *controllerCalculatorImpl) getReplicas(ownerRef metav1.OwnerReference, namespace string) (*replicasInfo, error) {
+	result := &replicasInfo{}
+	switch ownerRef.Kind {
+	case "StatefulSet":
+		sSet, err := c.listers.StatefulSetLister().StatefulSets(namespace).Get(ownerRef.Name)
+		if err != nil {
+			return nil, err
+		}
+		result.currentReplicas = sSet.Status.CurrentReplicas
+		if sSet.Spec.Replicas != nil {
+			result.targetReplicas = *sSet.Spec.Replicas
+		} else {
+			result.targetReplicas = 1
+		}
+	case "ReplicaSet":
+		rSet, err := c.listers.ReplicaSetLister().ReplicaSets(namespace).Get(ownerRef.Name)
+		if err != nil {
+			return nil, err
+		}
+		result.currentReplicas = rSet.Status.Replicas
+		if rSet.Spec.Replicas != nil {
+			result.targetReplicas = *rSet.Spec.Replicas
+		} else {
+			result.targetReplicas = 1
+		}
+	case "ReplicationController":
+		rController, err := c.listers.ReplicationControllerLister().ReplicationControllers(namespace).Get(ownerRef.Name)
+		if err != nil {
+			return nil, err
+		}
+		result.currentReplicas = rController.Status.Replicas
+		if rController.Spec.Replicas != nil {
+			result.targetReplicas = *rController.Spec.Replicas
+		} else {
+			result.targetReplicas = 1
+		}
+	case "Job":
+		job, err := c.listers.JobLister().Jobs(namespace).Get(ownerRef.Name)
+		if err != nil {
+			return nil, err
+		}
+		result.currentReplicas = job.Status.Active
+		if job.Spec.Parallelism != nil {
+			result.targetReplicas = *job.Spec.Parallelism
+		} else {
+			result.targetReplicas = 1
+		}
+		if job.Spec.Completions != nil && *job.Spec.Completions-job.Status.Succeeded < result.targetReplicas {
+			result.targetReplicas = *job.Spec.Completions - job.Status.Succeeded
+		}
+	default:
+		return nil, fmt.Errorf("unhandled controller type: %s", ownerRef.Kind)
+	}
+	return result, nil
+}
--- a/cluster-autoscaler/core/scaledown/planner/controller_test.go
+++ b/cluster-autoscaler/core/scaledown/planner/controller_test.go
@@ -1,0 +1,230 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package planner
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/gogo/protobuf/proto"
+	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/assert"
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	kube_util "k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/test"
+)
+
+var podLabels = map[string]string{
+	"app": "test",
+}
+
+func TestReplicasCounter(t *testing.T) {
+	job := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "job",
+			Namespace: "default",
+			UID:       types.UID("batch/v1/namespaces/default/jobs/job"),
+		},
+		Spec: batchv1.JobSpec{
+			Parallelism: proto.Int32(3),
+			Selector:    metav1.SetAsLabelSelector(podLabels),
+		},
+		Status: batchv1.JobStatus{Active: 1},
+	}
+	unsetJob := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "unset_job",
+			Namespace: "default",
+			UID:       types.UID("batch/v1/namespaces/default/jobs/unset_job"),
+		},
+	}
+	jobWithSucceededReplicas := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "succeeded_job",
+			Namespace: "default",
+			UID:       types.UID("batch/v1/namespaces/default/jobs/succeeded_job"),
+		},
+		Spec: batchv1.JobSpec{
+			Parallelism: proto.Int32(3),
+			Completions: proto.Int32(3),
+			Selector:    metav1.SetAsLabelSelector(podLabels),
+		},
+		Status: batchv1.JobStatus{
+			Active:    1,
+			Succeeded: 2,
+		},
+	}
+	rs := &appsv1.ReplicaSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "rs",
+			Namespace: "default",
+			UID:       types.UID("apps/v1/namespaces/default/replicasets/rs"),
+		},
+		Spec: appsv1.ReplicaSetSpec{
+			Replicas: proto.Int32(1),
+			Selector: metav1.SetAsLabelSelector(podLabels),
+		},
+		Status: appsv1.ReplicaSetStatus{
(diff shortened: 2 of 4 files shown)


cluster-autoscaler/core/scaledown/planner/planner_test.go
```diff
@@ -346,6 +463,37 @@ func generateReplicaSets() []*appsv1.ReplicaSet {
...
+			},
+		},
+	}
+}
+
+func rSetUID(name string) types.UID {
+	return types.UID(fmt.Sprintf("api/v1/namespaces/default/replicasets/%s", name))
+}
+
+func scheduledPodWithRSOwner(name string, cpu, memory int64, nodeName string, rsUID types.UID, labels map[string]string) *apiv1.Pod {
```
x13n, line 489: nit: Both `rsUID` and `labels` are always set to the same value in all tests. Why not just extend `scheduledPod` below to use these constants?

x13n (changes requested):

cluster-autoscaler/core/scaledown/planner/planner_test.go
olagacek: Done.

cluster-autoscaler/core/scaledown/planner/controller.go
```diff
@@ -0,0 +1,91 @@
...
+
+type controllerCalculatorImpl struct {
+	listers kubernetes.ListerRegistry
+}
+
+func newControllerCalculator(listers kubernetes.ListerRegistry) ControllerReplicasCalculator {
+	return &controllerCalculatorImpl{listers: listers}
+}
+
+func (c *controllerCalculatorImpl) GetReplicas(ownerRef metav1.OwnerReference, pod *corev1.Pod) (*replicasInfo, error) {
```
x13n, line 49: Does this need to accept a pod? I think `namespace` + `ownerRef` should not only be sufficient, but also less expensive. `GetPodStatefulSets(pod)` and similar functions list all controller objects and iterates through them whereas getting directly (e.g. `c.listers.StatefulSetLister().StatefulSets(namespace).Get(ownerRef.name)`) requires only O(1) lookup.

cluster-autoscaler/core/scaledown/planner/controller.go
```diff
@@ -0,0 +1,91 @@
...
+// ControllerReplicasCalculator calculates a number of target and expected replicas for a given controller.
+type ControllerReplicasCalculator interface {
+	GetReplicas(metav1.OwnerReference, *corev1.Pod) (*replicasInfo, error)
+}
+
+type controllerCalculatorImpl struct {
+	listers kubernetes.ListerRegistry
+}
+
+func newControllerCalculator(listers kubernetes.ListerRegistry) ControllerReplicasCalculator {
```
x13n, line 45: nit: `newControllerReplicasCalculator`?

cluster-autoscaler/core/scaledown/planner/controller.go
```diff
@@ -0,0 +1,91 @@
...
+var (
+	errNotFound = errors.New("owner not found")
+)
+
+type replicasInfo struct {
+	targetReplicas, currentReplicas int32
+}
+
+// ControllerReplicasCalculator calculates a number of target and expected replicas for a given controller.
+type ControllerReplicasCalculator interface {
```
x13n, line 37: I think in general interfaces belong in the place where they are being used, not next to one of implementations. Also, why public? `newControllerCalculator` is private, so it is hard to use this interface outside of `planner` module anyway.

cluster-autoscaler/core/scaledown/planner/controller_test.go
```diff
@@ -0,0 +1,177 @@
...
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			c := newControllerCalculator(listers)
+			res, err := c.GetReplicas(tc.pod.OwnerReferences[0], tc.pod)
+			if tc.expectErr {
+				assert.Error(t, err)
+			} else {
+				if res.targetReplicas != tc.expectedTargetReplicas {
```
x13n, line 142: super nit: you could just compare `replicasInfo` here and have a single check.

cluster-autoscaler/core/scaledown/planner/planner_test.go
```diff
@@ -346,10 +409,34 @@ func generateReplicaSets() []*appsv1.ReplicaSet {
 	}
 }
 
+func generateReplicaSetWithReplicas(name string, specReplicas, statusReplicas int32, labels map[string]string) []*appsv1.ReplicaSet {
```
x13n, line 412: Since you just introduced `ControllerReplicasCalculator` interface, wouldn't it be simpler to just mock it?

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -175,9 +177,33 @@ func (p *Planner) injectOngoingActuation() error {
...
+
+func filterOutRecreatedPods(pods []*apiv1.Pod, cc ControllerReplicasCalculator) []*apiv1.Pod {
+	var podsToInject []*apiv1.Pod
+	addedReplicas := make(map[string]int32)
+	for _, pod := range pods {
+		if len(pod.GetOwnerReferences()) == 0 {
+			podsToInject = append(podsToInject, pod)
+			continue
+		}
+		ownerRef := pod.GetOwnerReferences()[0]
```
x13n, line 191: I wonder if picking the first owner reference is always right. What if the first reference points to some custom resource, but the second one is a regular controller?

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -175,9 +177,33 @@ func (p *Planner) injectOngoingActuation() error {
...
+		if ownerRef.Kind == "DaemonSet" {
```
x13n, line 192: nit: I don't think DaemonSet really requires special handling here, `GetReplicas` can just return current=target. Might feel a bit hacky though.

x13n (changes requested):

cluster-autoscaler/core/scaledown/planner/controller.go
olagacek: switched to version with ownerRef + namespace. Initially thought we also need to check labels (which GetPodStatefulSets does), but it doesn't seem required

cluster-autoscaler/core/scaledown/planner/controller.go
olagacek: done.

cluster-autoscaler/core/scaledown/planner/controller.go
olagacek: moved to planner.go

cluster-autoscaler/core/scaledown/planner/controller_test.go
olagacek: done.

cluster-autoscaler/core/scaledown/planner/planner.go
olagacek: I was under impression that pod will always have one owner, though just in case I added a function `getSupportedOwnerRef`

cluster-autoscaler/core/scaledown/planner/planner.go
olagacek: I feel like it's a little too hacky to not handle DaemonSet here

cluster-autoscaler/core/scaledown/planner/controller.go
```diff
@@ -0,0 +1,72 @@
...
+func (c *controllerCalculatorImpl) getReplicas(ownerRef metav1.OwnerReference, namespace string) (*replicasInfo, error) {
+	result := &replicasInfo{}
+	switch ownerRef.Kind {
+	case "StatefulSet":
+		sSet, err := c.listers.StatefulSetLister().StatefulSets(namespace).Get(ownerRef.Name)
+		if err != nil {
+			return nil, err
+		}
+		result.currentReplicas = sSet.Status.CurrentReplicas
+		result.targetReplicas = *sSet.Spec.Replicas
```
x13n, line 43: This can cause panic when `sSet.Spec.Replicas` is `nil`. We should handle that and treat it as 1. (IIUC `nil` means value unspecified, use default). Same below. We should also have tests covering such cases.

olagacek: Added checks, defaulting & tests

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -175,9 +187,34 @@ func (p *Planner) injectOngoingActuation() error {
...
+	return p.injectPods(filterOutRecreatedPods(p.actuationStatus.RecentEvictions(), p.cc))
+}
+
+func filterOutRecreatedPods(pods []*apiv1.Pod, cc controllerReplicasCalculator) []*apiv1.Pod {
+	var podsToInject []*apiv1.Pod
+	addedReplicas := make(map[string]int32)
+	for _, pod := range pods {
+		ownerRef := getSupportedOwnerRef(pod.GetOwnerReferences())
+		if ownerRef == nil {
+			podsToInject = append(podsToInject, pod)
```
towca, line 199: If it's not one of the supported owner kinds, it probably won't get recreated right (e.g. just a naked pod)? Are we assuming it could get recreated by some custom controller to be on the safe side? If so, could you clarify that in a comment?

towca: And should we distinguish static/mirror pods? They won't have a supported owner ref, but they won't get recreated (just like DS pods).

cluster-autoscaler/core/scaledown/planner/controller_test.go
```diff
@@ -0,0 +1,206 @@
...
+	kube_util "k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/test"
+)
+
+var podLabels = map[string]string{
+	"app": "test",
+}
+
+func TestReplicasCounter(t *testing.T) {
+	job := &batchv1.Job{
```
towca, line 40: I'd also add a test case for a job in which `*job.Spec.Completions-job.Status.Succeeded < result.targetReplicas` is true.

olagacek: Added test case for that.

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -49,6 +50,15 @@ type removalSimulator interface {
 	SimulateNodeRemoval(node string, podDestinations map[string]bool, timestamp time.Time, pdbs []*policyv1.PodDisruptionBudget) (*simulator.NodeToBeRemoved, *simulator.UnremovableNode)
 }
 
+// controllerReplicasCalculator calculates a number of target and expected replicas for a given controller.
+type controllerReplicasCalculator interface {
+	getReplicas(metav1.OwnerReference, string) (*replicasInfo, error)
+}
+
+type replicasInfo struct {
```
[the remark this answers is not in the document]
olagacek: I think it should be in the same file as controllerReplicasCalculator, which with @x13n we agreed that should be in the same file as the functions that use it.

cluster-autoscaler/core/scaledown/planner/planner.go
olagacek: addressed that using filterRecreatable

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -252,6 +289,16 @@ func (p *Planner) categorizeNodes(podDestinations map[string]bool, scaleDownCand
 	}
 }
 
+func getSupportedOwnerRef(ownerRefs []metav1.OwnerReference) *metav1.OwnerReference {
```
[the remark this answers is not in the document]
olagacek: I renamed it to "known" and added a comment. Basically, this should return any ownerRef that CA knows how to handle (i.e. knows how to compute how many replicas are required)

cluster-autoscaler/core/scaledown/planner/planner.go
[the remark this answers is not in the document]
x13n: I meant the interface should be defined by the user, not implementation of a functionality. The struct itself is a different story - contrary to interfaces, there should be one struct definition used in all the places.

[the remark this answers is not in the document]
olagacek: What I mean is, it makes sense for the interface to be here and it makes sense for the struct it returns to be defined in the same file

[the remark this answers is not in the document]
towca: Fair enough, makes sense!

cluster-autoscaler/core/scaledown/planner/planner.go
towca: I have a couple of comments but all tied to this chain so I'll gather them here:

1. IMO it'd still be very much non-obvious to a future reader why we're choosing to inject the pods (and not disregard) in case of uncertainty/errors. A comment clarifying that injecting the pod actually means being on the safe side (and why) would help a lot. Some of it is mentioned in the method comment for `injectOngoingActuation` (which got outdated with this change btw), so maybe we could just update that, or move that comment here.

2. Hmm, but isn't `filterRecreatable` called only for `currentlyDrainedPods` and not `RecentEvictions`? So right now, won't we inject DS/static/mirror pods returned by `RecentEvictions` when we should't?

3. Also, I'd remove DaemonSet as known ref in `getKnownOwnerRef` - then ownerRef will be nil for all DS/static/mirror pods, they will be added here, but then filtered out by `filterRecreatable` down the line. Because right now I think DS pods will have a non-nil ref, but `getReplicas` will return an error - which seems like doing a lot of extra steps, and differs from how static/mirror pods are handled.

4. Do we have a test case for filtering DS/static/mirror pods? Seems like something should've started failing after you forgot the `filterRecreatable` call, but the CI looks like everything passed. Maybe it'd be worth adding a test case for that?

olagacek: 1. Added a comment here in `filterOutRecreatedPods` and updated comment in `injectOngoingActuation`
2. It was actually called for pods in `filterOutRecreatedPods`  at the beginning of function, though I moved it to `injectOngoingActuation` to make it consistent with `currentlyDrainedPods`
3. Removed DaemonSet from `getKnownOwnerRef`
4. I added test case with DS, though as stated in point 2, the `filterRecreatable` was run in previous revision as well

towca: Thanks, I knew I must've missed something :P Everything looks clear now, thanks!

cluster-autoscaler/core/scaledown/planner/controller_test.go
towca: Thx!

cluster-autoscaler/core/scaledown/planner/planner.go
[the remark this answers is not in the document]
towca: Ack, "known" is slightly better and it's clear in context, thanks!

PR #5290, opened 2022-11-03
Allow forking snapshot more than 1 time

yaroslava-serdiuk:

cluster-autoscaler

#### What type of PR is this?

/kind feature

Forking twice is needed for parallel scale down implementation.

### DIFF
--- a/cluster-autoscaler/core/scaledown/planner/planner.go
+++ b/cluster-autoscaler/core/scaledown/planner/planner.go
@@ -84,9 +84,9 @@ func New(context *context.AutoscalingContext, processors *processors.Autoscaling
 func (p *Planner) UpdateClusterState(podDestinations, scaleDownCandidates []*apiv1.Node, as scaledown.ActuationStatus, pdb []*policyv1.PodDisruptionBudget, currentTime time.Time) errors.AutoscalerError {
 	p.latestUpdate = currentTime
 	p.actuationStatus = as
-	// TODO: clone cluster snapshot to avoid persisting changes done by the
-	// simulation. Or - better yet - allow the snapshot to be forked twice
-	// and just fork it here.
+	// Avoid persisting changes done by the simulation.
+	p.context.ClusterSnapshot.Fork()
+	defer p.context.ClusterSnapshot.Revert()
 	err := p.injectOngoingActuation()
 	if err != nil {
 		p.CleanUpUnneededNodes()
--- a/cluster-autoscaler/simulator/clustersnapshot/basic.go
+++ b/cluster-autoscaler/simulator/clustersnapshot/basic.go
@@ -26,8 +26,7 @@ import (
 // BasicClusterSnapshot is simple, reference implementation of ClusterSnapshot.
 // It is inefficient. But hopefully bug-free and good for initial testing.
 type BasicClusterSnapshot struct {
-	baseData   *internalBasicSnapshotData
-	forkedData *internalBasicSnapshotData
+	data []*internalBasicSnapshotData
 }
 
 type internalBasicSnapshotData struct {
@@ -208,10 +207,7 @@ func NewBasicClusterSnapshot() *BasicClusterSnapshot {
 }
 
 func (snapshot *BasicClusterSnapshot) getInternalData() *internalBasicSnapshotData {
-	if snapshot.forkedData != nil {
-		return snapshot.forkedData
-	}
-	return snapshot.baseData
+	return snapshot.data[len(snapshot.data)-1]
 }
 
 // AddNode adds node to the snapshot.
@@ -258,36 +254,33 @@ func (snapshot *BasicClusterSnapshot) IsPVCUsedByPods(key string) bool {
 }
 
 // Fork creates a fork of snapshot state. All modifications can later be reverted to moment of forking via Revert()
-// Forking already forked snapshot is not allowed and will result with an error.
-func (snapshot *BasicClusterSnapshot) Fork() error {
-	if snapshot.forkedData != nil {
-		return fmt.Errorf("snapshot already forked")
-	}
-	snapshot.forkedData = snapshot.baseData.clone()
-	return nil
+func (snapshot *BasicClusterSnapshot) Fork() {
+	forkData := snapshot.getInternalData().clone()
+	snapshot.data = append(snapshot.data, forkData)
 }
 
 // Revert reverts snapshot state to moment of forking.
-func (snapshot *BasicClusterSnapshot) Revert() error {
-	snapshot.forkedData = nil
-	return nil
+func (snapshot *BasicClusterSnapshot) Revert() {
+	if len(snapshot.data) == 1 {
+		return
+	}
+	snapshot.data = snapshot.data[:len(snapshot.data)-1]
 }
 
 // Commit commits changes done after forking.
 func (snapshot *BasicClusterSnapshot) Commit() error {
-	if snapshot.forkedData == nil {
+	if len(snapshot.data) <= 1 {
 		// do nothing
 		return nil
 	}
-	snapshot.baseData = snapshot.forkedData
-	snapshot.forkedData = nil
+	snapshot.data = append(snapshot.data[:len(snapshot.data)-2], snapshot.data[len(snapshot.data)-1])
 	return nil
 }
 
 // Clear reset cluster snapshot to empty, unforked state
 func (snapshot *BasicClusterSnapshot) Clear() {
-	snapshot.baseData = newInternalBasicSnapshotData()
-	snapshot.forkedData = nil
+	baseData := newInternalBasicSnapshotData()
+	snapshot.data = []*internalBasicSnapshotData{baseData}
 }
 
 // implementation of SharedLister interface
--- a/cluster-autoscaler/simulator/clustersnapshot/clustersnapshot.go
+++ b/cluster-autoscaler/simulator/clustersnapshot/clustersnapshot.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 
 	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
 	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
 )
 
@@ -42,15 +43,36 @@ type ClusterSnapshot interface {
 	// IsPVCUsedByPods returns if the pvc is used by any pod, key = <namespace>/<pvc_name>
 	IsPVCUsedByPods(key string) bool
 
-	// Fork creates a fork of snapshot state. All modifications can later be reverted to moment of forking via Revert()
-	// Forking already forked snapshot is not allowed and will result with an error.
-	Fork() error
+	// Fork creates a fork of snapshot state. All modifications can later be reverted to moment of forking via Revert().
+	// Use WithForkedSnapshot() helper function instead if possible.
+	Fork()
 	// Revert reverts snapshot state to moment of forking.
-	Revert() error
+	Revert()
 	// Commit commits changes done after forking.
 	Commit() error
-	// Clear reset cluster snapshot to empty, unforked state
+	// Clear reset cluster snapshot to empty, unforked state.
 	Clear()
 }
 
 var errNodeNotFound = errors.New("node not found")
+
+// WithForkedSnapshot is a helper function for snapshot that makes sure all Fork() calls are closed with Commit() or Revert() calls.
+// The function return (error, error) pair. The first error comes from the passed function, the second error indicate the success of the function itself.
+func WithForkedSnapshot(snapshot ClusterSnapshot, f func() (bool, error)) (error, error) {
+	var commit bool
+	var err, cleanupErr error
+	snapshot.Fork()
+	defer func() {
+		if commit {
+			cleanupErr = snapshot.Commit()
+			if cleanupErr != nil {
+				klog.Errorf("Got error when calling ClusterSnapshot.Commit(), will try to revert; %v", cleanupErr)
+			}
+		}
+		if !commit || cleanupErr != nil {
+			snapshot.Revert()
+		}
+	}()
+	commit, err = f()
+	return err, cleanupErr
+}
--- a/cluster-autoscaler/simulator/clustersnapshot/clustersnapshot_benchmark_test.go
+++ b/cluster-autoscaler/simulator/clustersnapshot/clustersnapshot_benchmark_test.go
@@ -188,22 +188,23 @@ func BenchmarkForkAddRevert(b *testing.B) {
 					err = clusterSnapshot.AddPod(pod, pod.Spec.NodeName)
 					assert.NoError(b, err)
 				}
-				tmpNode := BuildTestNode("tmp", 2000, 2000000)
+				tmpNode1 := BuildTestNode("tmp-1", 2000, 2000000)
+				tmpNode2 := BuildTestNode("tmp-2", 2000, 2000000)
 				b.ResetTimer()
 				b.Run(fmt.Sprintf("%s: ForkAddRevert (%d nodes, %d pods)", snapshotName, ntc, ptc), func(b *testing.B) {
 					for i := 0; i < b.N; i++ {
-						err = clusterSnapshot.Fork()
+						clusterSnapshot.Fork()
+						err = clusterSnapshot.AddNode(tmpNode1)
 						if err != nil {
 							assert.NoError(b, err)
 						}
-						err = clusterSnapshot.AddNode(tmpNode)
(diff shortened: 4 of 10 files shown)


cluster-autoscaler/simulator/clustersnapshot/clustersnapshot.go
```diff
@@ -43,14 +43,16 @@ type ClusterSnapshot interface {
...
 	Commit() error
-	// Clear reset cluster snapshot to empty, unforked state
+	// Clear reset cluster snapshot to empty, unforked state.
 	Clear()
 }
 
 var errNodeNotFound = errors.New("node not found")
+
+// ForkLimit is a defautl fork limit for cluster cnapshot.
+const ForkLimit = 2
```
x13n, line 58: Why impose an arbitrary limit?

cluster-autoscaler/simulator/clustersnapshot/clustersnapshot_test.go
```diff
@@ -197,8 +214,34 @@ func TestForking(t *testing.T) {
 				// Modifications should no longer be applied.
 				compareStates(t, tc.state, getSnapshotState(t, snapshot))
 			})
+			t.Run(fmt.Sprintf("%s: %s fork & fork & revert & revert", name, tc.name), func(t *testing.T) {
```
x13n, line 217: I think there should be more test cases like this: fork & fork & commit & revert, fork & fork & revert & commit, with snapshot changes between these steps.

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -84,10 +85,12 @@ func New(context *context.AutoscalingContext, processors *processors.Autoscaling
 func (p *Planner) UpdateClusterState(podDestinations, scaleDownCandidates []*apiv1.Node, as scaledown.ActuationStatus, pdb []*policyv1.PodDisruptionBudget, currentTime time.Time) errors.AutoscalerError {
 	p.latestUpdate = currentTime
 	p.actuationStatus = as
-	// TODO: clone cluster snapshot to avoid persisting changes done by the
-	// simulation. Or - better yet - allow the snapshot to be forked twice
-	// and just fork it here.
-	err := p.injectOngoingActuation()
+	err := p.rs.ForkClusterSnapshot()
```
x13n, line 88: `RemovalSimulator` operates on ClusterSnapshot passed through context. Since we have access to `p.context` here, I'd just fork it here instead of extending `removalSimulator` interface. Also note that when we fork, we also need to `defer Revert()` afterwards.

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -84,10 +85,12 @@ func New(context *context.AutoscalingContext, processors *processors.Autoscaling
...
+	if err != nil {
```
x13n, line 89: Forking the snapshot shouldn't return an error anymore. If forking can happen multiple times, `Fork()` shouldn't return a value at all. This will simplify error handling in a few places.

x13n (changes requested):

cluster-autoscaler/simulator/clustersnapshot/clustersnapshot.go
yaroslava-serdiuk: The main reason why there was a limitation to fork once previously is to ensure we don't accidentally fork twice.
The main reason why I added a fork limit is to control recursion depth of forking. Currently we limit fork to two time, so I used this value as a new default.

cluster-autoscaler/core/scaledown/planner/planner.go
yaroslava-serdiuk: >Forking the snapshot shouldn't return an error anymore
Why? I added a fork limit, so the Fork() can return error.

cluster-autoscaler/simulator/clustersnapshot/clustersnapshot.go
x13n: I disagree handling this as a runtime error is the right approach. If we are concerned there are parts of the logic which could fork without committing/reverting, there should be unit tests ensuring this doesn't happen.

x13n: Alternatively, we could modify the interface in a way that ensures forking without commiting/reverting is not possible at all. I actually did just that in one place already:
https://github.com/kubernetes/autoscaler/blob/a703c3f6cff1d54dc92378c5baad2c2cccdbbc8f/cluster-autoscaler/simulator/cluster.go#L198-L217

cluster-autoscaler/core/scaledown/planner/planner.go
yaroslava-serdiuk: Done

cluster-autoscaler/core/scaledown/planner/planner.go
yaroslava-serdiuk: Ack, thanks!

cluster-autoscaler/simulator/clustersnapshot/clustersnapshot_test.go
yaroslava-serdiuk: Done

cluster-autoscaler/simulator/clustersnapshot/clustersnapshot.go
yaroslava-serdiuk: Done

cluster-autoscaler/simulator/clustersnapshot/clustersnapshot_test.go
```diff
@@ -96,6 +96,16 @@ func validTestCases(t *testing.T) []modificationTestCase {
 	pod.Spec.NodeName = node.Name
 
 	testCases := []modificationTestCase{
+		{
+			name: "add node",
```
x13n, line 100: Looks like this just duplicates existing test case below, why add a second one?

cluster-autoscaler/simulator/clustersnapshot/clustersnapshot_test.go
```diff
@@ -695,3 +748,34 @@ func TestPVCClearAndFork(t *testing.T) {
 		})
 	}
 }
+
+func TestWithForkedSnapshot(t *testing.T) {
+	testCases := validTestCases(t)
+	for name, snapshotFactory := range snapshots {
+		for _, tc := range testCases[:1] {
```
x13n, line 755: Let's not depend on filtering out specific test cases like this. Why test just one?

x13n (changes requested):

cluster-autoscaler/simulator/clustersnapshot/clustersnapshot_test.go
yaroslava-serdiuk: Oops, sorry it's a leftover from debugging, thanks for catching it!

cluster-autoscaler/simulator/clustersnapshot/basic.go
```diff
@@ -23,11 +23,10 @@ import (
 	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
 )
 
-// BasicClusterSnapshot is simple, reference implementation of ClusterSnapshot.
+// BasicClusterSnapshot is simple, reference implementation of ClusterSnapshot forks.
```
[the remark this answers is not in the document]
yaroslava-serdiuk: Removing it since it sounds confusing. My intend is to describe that now basic snapshot is a list of forks.

cluster-autoscaler/simulator/clustersnapshot/clustersnapshot.go
```diff
@@ -43,14 +44,40 @@ type ClusterSnapshot interface {
...
 	// Commit commits changes done after forking.
 	Commit() error
-	// Clear reset cluster snapshot to empty, unforked state
+	// Clear reset cluster snapshot to empty, unforked state.
 	Clear()
 }
 
 var errNodeNotFound = errors.New("node not found")
+
+// TODO(yaroslava): move WithForkedSnapshot function to ClusterSnapshot interface.
```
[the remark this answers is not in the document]
yaroslava-serdiuk: I see your point. However now we allow fork infinite time, which actually I would like to avoid somehow.

cluster-autoscaler/simulator/clustersnapshot/clustersnapshot_benchmark_test.go
```diff
@@ -188,15 +188,21 @@ func BenchmarkForkAddRevert(b *testing.B) {
...
+						err = clusterSnapshot.AddNode(tmpNode1)
 						if err != nil {
 							assert.NoError(b, err)
 						}
-						err = clusterSnapshot.AddNode(tmpNode)
+						err = clusterSnapshot.AddNode(tmpNode2)
+						if err != nil {
+							assert.NoError(b, err)
+						}
+						err = clusterSnapshot.Revert()
```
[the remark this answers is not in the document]
yaroslava-serdiuk: Seems I wanted to modify the test and fork twice, but I forgot to fork second time :)

cluster-autoscaler/simulator/clustersnapshot/basic.go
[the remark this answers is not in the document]
x13n: Got it. I think this is an implementation detail, which doesn't have to be a part of object doc comment.

cluster-autoscaler/simulator/clustersnapshot/clustersnapshot.go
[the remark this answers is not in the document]
x13n: Sure, we can document that `Fork`, `Commit` and `Revert` are not meant to be invoked directly. It is always up to the person implementing changes to write correct code - giving them the tool to do so easily (the helper function) seems sufficient to me.

x13n:
Thanks for the change, LGTM now. I'd still drop that TODO, but don't want to block on this. Putting on hold to let you make the change or just drop the hold if you still disagree.

/lgtm
/hold

cluster-autoscaler/simulator/clustersnapshot/basic.go
```diff
@@ -258,36 +254,36 @@ func (snapshot *BasicClusterSnapshot) IsPVCUsedByPods(key string) bool {
...
 func (snapshot *BasicClusterSnapshot) Commit() error {
-	if snapshot.forkedData == nil {
+	if len(snapshot.data) <= 1 {
 		// do nothing
 		return nil
 	}
-	snapshot.baseData = snapshot.forkedData
-	snapshot.forkedData = nil
+	lastData := snapshot.data[len(snapshot.data)-1]
+	snapshot.Revert()
```
MaciekPytel, line 278: nit: I think direct pop from the list would be much more readable, I found this very confusing when skimming through the code (I'm not saying this is incorrect, just surprising to me).

cluster-autoscaler/simulator/clustersnapshot/clustersnapshot.go
```diff
@@ -42,15 +43,40 @@ type ClusterSnapshot interface {
...
-	// Clear reset cluster snapshot to empty, unforked state
+	// Clear reset cluster snapshot to empty, unforked state.
 	Clear()
 }
 
 var errNodeNotFound = errors.New("node not found")
+
+// WithForkedSnapshot is a helper function for snapshot that makes sure all Fork() calls are closed with Commit() or Revert() calls.
+// Error for this function means that cluster snapshot is in broken state.
+func WithForkedSnapshot(snapshot ClusterSnapshot, f func() error) error {
```
MaciekPytel, line 61: I don't really like using error returned by passing function as a signal to commit/revert. 

I think the decision to commit or revert depends more on what the forked snapshot is used for (scale-up/scale-down simulation vs updating the expected state of the cluster) and not necessarily on the success / failure of the logic. Or, in other words, just reverting a forked snapshot and proceeding is generally not the error handling strategy for a failed simulation. By using `func() error` we have no convenient way of propagating the error outside of WithForkedSnapshot.

I'd much rather this took `func() (bool, error)` where the boolean return value determines if the forked snapshot should be commited or reverted. This would separate the decision from any other error handling and force the user to make explicit decision on whether the snapshot should be committed or reverted for each possible error condition.

MaciekPytel:
/approve
/hold
Holding to give you a chance to look at my comments before this is merged.

cluster-autoscaler/simulator/clustersnapshot/clustersnapshot.go
yaroslava-serdiuk: Agreed, thanks for pointing this out! I modified the function and changed the error returned from WithForkedSnapshot regarding cluster snapshot operation. I think we should return error that shows if expected operation was completed (Commit or Revert). 
However all of the implementations of Revert() return nil. And I suppose since Revert() is direct opposite to Fork(), so they both should return the same value. So I actually think that we can drop returned value for Revert() as we did it for Fork(). WDYT?

MaciekPytel: I guess if the revert cannot fail in any implementation, I think we can drop the error from signature. I don't really expect implementations of this interface outside the ones we have, so changing interface doesn't seem like an issue.
The only condition I can think of when revert could fail is if the snapshot was not forked, but IIRC we just treat this as no-op and don't return an error?

yaroslava-serdiuk: Yes, if we haven't fork before calling revert we just do nothing and return nil

yaroslava-serdiuk: Dropped error for Revert()

cluster-autoscaler/simulator/clustersnapshot/basic.go
```diff
@@ -258,36 +254,33 @@ func (snapshot *BasicClusterSnapshot) IsPVCUsedByPods(key string) bool {
...
 // Commit commits changes done after forking.
 func (snapshot *BasicClusterSnapshot) Commit() error {
-	if snapshot.forkedData == nil {
+	if len(snapshot.data) <= 1 {
 		// do nothing
 		return nil
 	}
-	snapshot.baseData = snapshot.forkedData
-	snapshot.forkedData = nil
+	snapshot.data = append(snapshot.data[:len(snapshot.data)-2], snapshot.data[len(snapshot.data)-1])
```
MaciekPytel, line 276: Why not just `snapshot.data = snapshot.data[:len(snapshot.data)-1]` like you do in revert? Is there some subtle difference in behavior between the two?

yaroslava-serdiuk: My understanding of commit() is following: if we have a chain of snapshots A -> B -> C -> D (D is the last fork), then after commit we would like to have A -> B -> D. So we want to have everything up to len-2 (excluding) + the last element. 
Reverting is definitely what we want to have as result for commit()

MaciekPytel: Oh, right, we're deleting last-but-one element. Makes perfect sense, sorry, my comment makes no sense :(

cluster-autoscaler/simulator/clustersnapshot/clustersnapshot.go
```diff
@@ -42,15 +43,36 @@ type ClusterSnapshot interface {
...
+	// Clear reset cluster snapshot to empty, unforked state.
 	Clear()
 }
 
 var errNodeNotFound = errors.New("node not found")
+
+// WithForkedSnapshot is a helper function for snapshot that makes sure all Fork() calls are closed with Commit() or Revert() calls.
+// The function return (error, error) pair. The first error comes from the passed function, the second error indicate the success of the function itself.
+func WithForkedSnapshot(snapshot ClusterSnapshot, f func() (bool, error)) (error, error) {
+	var ok bool
```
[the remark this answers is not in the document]
yaroslava-serdiuk: Done

cluster-autoscaler/simulator/clustersnapshot/clustersnapshot.go
```diff
@@ -42,15 +43,36 @@ type ClusterSnapshot interface {
...
+	var err, cleanupErr error
+	snapshot.Fork()
+	defer func() {
+		if ok {
+			cleanupErr = snapshot.Commit()
+			if cleanupErr != nil {
+				klog.Fatalf("Got error when calling ClusterSnapshot.Commit(), will try to revert; %v", cleanupErr)
```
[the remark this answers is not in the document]
yaroslava-serdiuk: Right, change Fatalf to Errorf, since we can revert and nothing bad will happen.

PR #5289, opened 2022-11-03
Don't say that `Recreate` and `Auto` VPA modes are experimental

jbartosik:
#### Which component this PR applies to?

VPA

#### What type of PR is this?

/kind documentation

#### Which issue(s) this PR fixes:

Fixes #5245

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #5294, opened 2022-11-04
Don't say that `Recreate` and `Auto` VPA modes are experimental

### LINKED ISSUE #5245: Consider not marking Recreate / Auto VPA modes as experimental

VPA

VPA `Recreate` and `Auto` modes are marked as experimental in the [documentation](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler#quick-start). They've been around for a while so we should consider not saying they're experimental anymore.

Check if they're ready & either improve those modes or stop saying they're experimental.

-

**Additional context.**:

-

jbartosik:

VPA

#### What type of PR is this?

/kind documentation

Fixes #5245

```release-note
None
```

### DIFF
--- a/vertical-pod-autoscaler/README.md
+++ b/vertical-pod-autoscaler/README.md
@@ -139,15 +139,21 @@ There are four modes in which *VPAs* operate:
   them on existing pods using the preferred update mechanism. Currently, this is
   equivalent to `"Recreate"` (see below). Once restart free ("in-place") update
   of pod requests is available, it may be used as the preferred update mechanism by
-  the `"Auto"` mode. **NOTE:** This feature of VPA is experimental and may cause downtime
-  for your applications.
+  the `"Auto"` mode. **NOTE:** This feature of VPA may cause downtime for your
+  applications if new pods can't start (for example because there is no space for them
+  in the cluster). Using
+  [Cluster Autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler#cluster-autoscaler)
+  should mitigate the problem.
 * `"Recreate"`: VPA assigns resource requests on pod creation as well as updates
   them on existing pods by evicting them when the requested resources differ significantly
   from the new recommendation (respecting the Pod Disruption Budget, if defined).
   This mode should be used rarely, only if you need to ensure that the pods are restarted
   whenever the resource request changes. Otherwise, prefer the `"Auto"` mode which may take
   advantage of restart-free updates once they are available. **NOTE:** This feature of VPA
-  is experimental and may cause downtime for your applications.
+  may cause downtime for your applications if new pods can't start (for example because
+  there is no space for the in the cluster). Using
+  [Cluster Autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler#cluster-autoscaler)
+  should mitigate the problem.
 * `"Initial"`: VPA only assigns resource requests on pod creation and never changes them
   later.
 * `"Off"`: VPA does not automatically change the resource requirements of the pods.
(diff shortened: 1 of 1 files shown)


jbartosik:
@voelzmo please take a look. I addressed your comments from #5289 but I had to do that in a new PR (I created the earlier one editing on github and it created a branch in the kubernetes/autoscaler repo, not in my fork).

voelzmo (approved):

vertical-pod-autoscaler/README.md
```diff
@@ -139,15 +139,21 @@ There are four modes in which *VPAs* operate:
   them on existing pods using the preferred update mechanism. Currently, this is
   equivalent to `"Recreate"` (see below). Once restart free ("in-place") update
   of pod requests is available, it may be used as the preferred update mechanism by
-  the `"Auto"` mode. **NOTE:** This feature of VPA is experimental and may cause downtime
-  for your applications.
+  the `"Auto"` mode. **NOTE:** This feature of VPA may cause downtime for your
+  applications if new pods can't start (for example because there is no space for them
+  in the cluster). Using
+  [Cluster Autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler#cluster-autoscaler)
+  should mitigate the problem.
```
kgolab, line 146: I wouldn't say "mitigate" here - CA might help by creating a new Node but there are also scenarios when CA won't help:
* no more available capacity (e.g. NodePool at max size),
* selectors not allowing to use big enough Nodes (VPA only looks at the largest Node in the cluster),
* running out of resource quota,
* any webhooks affecting recreated Pods.

I don't think we want to go into so much detail here but maybe it's worth to mention the above in Known limitations or Caveats?

vertical-pod-autoscaler/README.md
```diff
@@ -139,15 +139,21 @@ There are four modes in which *VPAs* operate:
...
 * `"Recreate"`: VPA assigns resource requests on pod creation as well as updates
   them on existing pods by evicting them when the requested resources differ significantly
   from the new recommendation (respecting the Pod Disruption Budget, if defined).
   This mode should be used rarely, only if you need to ensure that the pods are restarted
   whenever the resource request changes. Otherwise, prefer the `"Auto"` mode which may take
   advantage of restart-free updates once they are available. **NOTE:** This feature of VPA
```
kgolab, line 152: Maybe instead of repeating the whole note we should have it written once and only reference from both modes?
Or we could leave a longer note under 'Recreate' and reference from 'Auto' to be more future-proof but I think I'd rather see a separate section about the dangers of recreating Pods by VPA.

vertical-pod-autoscaler/README.md
```diff
@@ -298,9 +304,9 @@ You can then choose which recommender to use by setting `recommenders` inside th
 
 # Known limitations
 
-* Updating running pods is an experimental feature of VPA. Whenever VPA updates
-  the pod resources, the pod is recreated, which causes all running containers to
-  be restarted. The pod may be recreated on a different node.
+* Whenever VPA updates the pod resources, the pod is recreated, which causes all
+  running containers to be restarted. The pod may be recreated on a different
```
kgolab, line 308: To be precise the containers are not restarted, they're also created anew. This might be an important distinction if someone attaches any external resources to these containers (e.g. persistent disks).

vertical-pod-autoscaler/README.md
jbartosik: Done

vertical-pod-autoscaler/README.md
jbartosik: Done.

PR #5309, opened 2022-11-10
Make prometheusRule.interval default to 1m

danksim:

helm-charts

#### What type of PR is this?
/kind bug

What: Set default `prometheusRule.interval` to `1m` instead of `global. evaluation_interval` (which has a default value of `1m`)
Why: Helm release was failing since `prometheusRule.interval` was defaulting to `nil` when not explicitly set

Fixes #

Please close if necessary

```release-note

```

```docs

```

### DIFF
--- a/charts/cluster-autoscaler/values.yaml
+++ b/charts/cluster-autoscaler/values.yaml
@@ -358,8 +358,8 @@ prometheusRule:
   additionalLabels: {}
   # prometheusRule.namespace -- Namespace which Prometheus is running in.
   namespace: monitoring
-  # prometheusRule.interval -- How often rules in the group are evaluated (falls back to the default value of `global.evaluation_interval` (which is 1m) if not set).
-  interval: 1m
+  # prometheusRule.interval -- How often rules in the group are evaluated (falls back to the Prometheus level value of `global.evaluation_interval` if not set).
+  interval: nil
   # prometheusRule.rules -- Rules spec template (see https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#rule).
   rules: []
 
--- a/charts/cluster-autoscaler/values.yaml
+++ b/charts/cluster-autoscaler/values.yaml
@@ -358,8 +358,8 @@ prometheusRule:
   additionalLabels: {}
   # prometheusRule.namespace -- Namespace which Prometheus is running in.
   namespace: monitoring
-  # prometheusRule.interval -- How often rules in the group are evaluated (falls back to `global.evaluation_interval` if not set).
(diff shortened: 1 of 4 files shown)


gjtempleton:
Hey, thanks for the PR, we use helm-docs to generate part of the `README.md` from the values file, you'll need to update the default value and comment there. [Pre-commit](https://github.com/kubernetes/autoscaler/tree/master/charts#pre-commit-hooks) should then take care of running the helm-docs command to update the README.

gjtempleton:
Thanks for that, one last request, can you squash your commits down to one please and then we'll be good to merge.

danksim:
Just squashed and pushed. Thank you! @gjtempleton

charts/cluster-autoscaler/values.yaml
```diff
@@ -358,8 +358,8 @@ prometheusRule:
   additionalLabels: {}
   # prometheusRule.namespace -- Namespace which Prometheus is running in.
   namespace: monitoring
-  # prometheusRule.interval -- How often rules in the group are evaluated (falls back to `global.evaluation_interval` if not set).
-  interval: null
+  # prometheusRule.interval -- How often rules in the group are evaluated (falls back to the default value of `global.evaluation_interval` (which is 1m) if not set).
```
gjtempleton, line 361: Related to above comment on emitting the interval if not set.
```suggestion
  # prometheusRule.interval -- How often rules in the group are evaluated (falls back to the Prometheus level value of `global.evaluation_interval` if not set).
```

charts/cluster-autoscaler/values.yaml
```diff
@@ -358,8 +358,8 @@ prometheusRule:
   additionalLabels: {}
   # prometheusRule.namespace -- Namespace which Prometheus is running in.
   namespace: monitoring
-  # prometheusRule.interval -- How often rules in the group are evaluated (falls back to `global.evaluation_interval` if not set).
-  interval: null
+  # prometheusRule.interval -- How often rules in the group are evaluated (falls back to the Prometheus level value of `global.evaluation_interval` if not set).
+  interval: nil
```
gjtempleton, line 362: I don't think this change is necessary, templating with this as the default value:
```yaml
spec:
  groups:
    - name: release-name-aws-cluster-autoscaler
      interval: nil
      rules:
        []
```

Templating with the new logic, but the old `interval: null`:
```yaml
spec:
  groups:
    - name: release-name-aws-cluster-autoscaler
      rules:
        []
```

gjtempleton:
Closing for now, please re-open if you're able to address the feedback and conflicts.
/close

PR #5313, opened 2022-11-14
recommendation post processor for integer CPU

### LINKED ISSUE #5236: VPA - Support of Static CPU Allocation

dbenque:

vertical-pod-autoscaler/recommender

#### What type of PR is this?

/kind feature

It introduces a recommendation post processor that allows users to have a CPU integer as recommendation.
Thanks to that it is now possible to use VPA with applications using [CPU management with static policy](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy)

Fixes #5236

As discussed during the SIG meeting the user interface is based on an annotations. This could be structured in the spec of the VPA once the feature will have gained maturity.

The "cpu integer post-processor" can be enabled using flag `--cpu-integer-post-processor-enabled=true` when starting the VPA recommender.
The post-processor would round-up CPU recommendation for containers referenced by a VPA annotation with this format:
`vpa-post-processor.kubernetes.io/{containerName}_integerCPU=true`

This post processor is optional. To run it in the recommender the command-line flag should be set:
`--cpu-integer-post-processor-enabled=true`, then only the pods/containers that have the relevant annotation are affected.

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/main.go
+++ b/vertical-pod-autoscaler/pkg/recommender/main.go
@@ -89,12 +89,13 @@ func main() {
 
 	useCheckpoints := *storage != "prometheus"
 
-	postProcessors := []routines.RecommendationPostProcessor{
-		&routines.CappingPostProcessor{},
-	}
+	var postProcessors []routines.RecommendationPostProcessor
 	if *postProcessorCPUasInteger {
 		postProcessors = append(postProcessors, &routines.IntegerCPUPostProcessor{})
 	}
+	// CappingPostProcessor, should always come in the last position for post-processing
+	postProcessors = append(postProcessors, &routines.CappingPostProcessor{})
+
 	recommender := routines.NewRecommender(config, *checkpointsGCInterval, useCheckpoints, *vpaObjectNamespace, *recommenderName, postProcessors)
 
 	promQueryTimeout, err := time.ParseDuration(*queryTimeout)
--- a/vertical-pod-autoscaler/pkg/recommender/main.go
+++ b/vertical-pod-autoscaler/pkg/recommender/main.go
@@ -70,7 +70,7 @@ var (
 // Post processors flags
 var (
 	// CPU as integer to benefit for CPU management Static Policy ( https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy )
-	postProcessorCPUasInteger = flag.Bool("cpu-integer-post-processor-enabled", false, "Enable the cpu-integer recommendation post processor. Only containers that specify it can have their CPU recommendation rounded up to an integer value.")
+	postProcessorCPUasInteger = flag.Bool("cpu-integer-post-processor-enabled", false, "Enable the cpu-integer recommendation post processor. The post processor will round up CPU recommendations to a whole CPU for pods which were opted in by setting an appropriate label on VPA object (experimental)")
 )
 
 func main() {
--- a/vertical-pod-autoscaler/pkg/recommender/main.go
+++ b/vertical-pod-autoscaler/pkg/recommender/main.go
@@ -67,6 +67,12 @@ var (
 	cpuHistogramDecayHalfLife      = flag.Duration("cpu-histogram-decay-half-life", model.DefaultCPUHistogramDecayHalfLife, `The amount of time it takes a historical CPU usage sample to lose half of its weight.`)
 )
 
+// Post processors flags
+var (
+	// CPU as integer to benefit for CPU management Static Policy ( https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy )
+	postProcessorCPUasInteger = flag.Bool("cpu-integer-post-processor-enabled", false, "Enable the cpu-integer recommendation post processor. Only containers that specify it can have their CPU recommendation rounded up to an integer value.")
+)
+
 func main() {
 	klog.InitFlags(nil)
 	kube_flag.InitFlags()
@@ -85,6 +91,9 @@ func main() {
 
 	postProcessors := []routines.RecommendationPostProcessor{
 		&routines.CappingPostProcessor{},
+	}
+	if *postProcessorCPUasInteger {
+		postProcessors = append(postProcessors, &routines.IntegerCPUPostProcessor{})
 	}
 	recommender := routines.NewRecommender(config, *checkpointsGCInterval, useCheckpoints, *vpaObjectNamespace, *recommenderName, postProcessors)
 
--- a/vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor.go
+++ b/vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor.go
@@ -18,9 +18,9 @@ package routines
 
 import (
 	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1"
 	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/recommender/model"
-	"math"
 	"strings"
 )
 
@@ -44,18 +44,6 @@ func (p *IntegerCPUPostProcessor) Process(vpa *model.Vpa, recommendation *vpa_ty
 
 	amendedRecommendation := recommendation.DeepCopy()
 
-	process := func(recommendation apiv1.ResourceList) {
-		for resourceName, recommended := range recommendation {
-			if resourceName != apiv1.ResourceCPU {
-				continue
-			}
-			v := float64(recommended.MilliValue()) / 1000
-			r := int64(math.Ceil(v))
-			recommended.Set(r)
-			recommendation[resourceName] = recommended
-		}
-	}
-
 	for key, value := range vpa.Annotations {
 		containerName := extractContainerName(key, vpaPostProcessorPrefix, vpaPostProcessorIntegerCPUSuffix)
 		if containerName == "" || value != vpaPostProcessorIntegerCPUValue {
@@ -66,13 +54,23 @@ func (p *IntegerCPUPostProcessor) Process(vpa *model.Vpa, recommendation *vpa_ty
 			if r.ContainerName != containerName {
 				continue
 			}
-			process(r.Target)
-			process(r.LowerBound)
-			process(r.UpperBound)
-			process(r.UncappedTarget)
+			setIntegerCPURecommendation(r.Target)
+			setIntegerCPURecommendation(r.LowerBound)
+			setIntegerCPURecommendation(r.UpperBound)
+			setIntegerCPURecommendation(r.UncappedTarget)
 		}
 	}
 	return amendedRecommendation
+}
+
+func setIntegerCPURecommendation(recommendation apiv1.ResourceList) {
+	for resourceName, recommended := range recommendation {
+		if resourceName != apiv1.ResourceCPU {
+			continue
+		}
+		recommended.RoundUp(resource.Scale(0))
+		recommendation[resourceName] = recommended
+	}
 }
 
 // extractContainerName return the container name for the feature based on annotation key
--- a/vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor.go
+++ b/vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor.go
@@ -1,0 +1,89 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package routines
+
+import (
+	apiv1 "k8s.io/api/core/v1"
+	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1"
+	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/recommender/model"
+	"math"
+	"strings"
+)
+
+// IntegerCPUPostProcessor ensures that the recommendation delivers an integer value for CPU
+// This is need for users who want to use CPU Management with static policy: https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy
+type IntegerCPUPostProcessor struct{}
+
+const (
+	// The user interface for that post processor is an annotation on the VPA object with the following format:
+	// vpa-post-processor.kubernetes.io/{containerName}_integerCPU=true
+	vpaPostProcessorPrefix           = "vpa-post-processor.kubernetes.io/"
+	vpaPostProcessorIntegerCPUSuffix = "_integerCPU"
+	vpaPostProcessorIntegerCPUValue  = "true"
+)
+
+var _ RecommendationPostProcessor = &IntegerCPUPostProcessor{}
+
+// Process apply the capping post-processing to the recommendation.
+// For this post processor the CPU value is rounded up to an integer
+func (p *IntegerCPUPostProcessor) Process(vpa *model.Vpa, recommendation *vpa_types.RecommendedPodResources, policy *vpa_types.PodResourcePolicy) *vpa_types.RecommendedPodResources {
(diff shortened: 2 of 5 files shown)


vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor.go
```diff
@@ -0,0 +1,89 @@
...
+
+var _ RecommendationPostProcessor = &IntegerCPUPostProcessor{}
+
+// Process apply the capping post-processing to the recommendation.
+// For this post processor the CPU value is rounded up to an integer
+func (p *IntegerCPUPostProcessor) Process(vpa *model.Vpa, recommendation *vpa_types.RecommendedPodResources, policy *vpa_types.PodResourcePolicy) *vpa_types.RecommendedPodResources {
+
+	amendedRecommendation := recommendation.DeepCopy()
+
+	process := func(recommendation apiv1.ResourceList) {
```
jbartosik, line 47: I think it would be a bit easier to read if this function was declared as a regular function (and then we could also test it directly)

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor.go
```diff
@@ -0,0 +1,89 @@
...
+		for resourceName, recommended := range recommendation {
+			if resourceName != apiv1.ResourceCPU {
+				continue
+			}
+			v := float64(recommended.MilliValue()) / 1000
+			r := int64(math.Ceil(v))
```
jbartosik, line 53: Please use [`Quantity.AsScale`](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity.AsScale)

vertical-pod-autoscaler/pkg/recommender/main.go
```diff
@@ -67,6 +67,12 @@ var (
 	cpuHistogramDecayHalfLife      = flag.Duration("cpu-histogram-decay-half-life", model.DefaultCPUHistogramDecayHalfLife, `The amount of time it takes a historical CPU usage sample to lose half of its weight.`)
 )
 
+// Post processors flags
+var (
+	// CPU as integer to benefit for CPU management Static Policy ( https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy )
+	postProcessorCPUasInteger = flag.Bool("cpu-integer-post-processor-enabled", false, "Enable the cpu-integer recommendation post processor. Only containers that specify it can have their CPU recommendation rounded up to an integer value.")
```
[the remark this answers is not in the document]
jbartosik: After reading the PR:

"it" refers to the fact that VPA can specify (with labels) which containers should have their recommendations rounded up to int.

So clearer description might be something like:

Enable the cpu-integer recommendation post processor. The post processor will round up CPU recommendations to a whole CPU for pods which were opted in by setting an appropriate label on VPA object (experimental)

?

jbartosik (changes requested):
I took a first pass. I couldn't look at tests yet.

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor.go
```diff
@@ -0,0 +1,89 @@
...
+			process(r.UpperBound)
+			process(r.UncappedTarget)
+		}
+	}
+	return amendedRecommendation
+}
+
+// extractContainerName return the container name for the feature based on annotation key
+// if the returned value is empty that means that the key does not match
+func extractContainerName(key, prefix, suffix string) string {
```
jbartosik, line 80: Why are prefix & suffix args here (rather than using consts in the function)?

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor_test.go
```diff
@@ -0,0 +1,160 @@
...
+					test.Recommendation().WithContainer("container2").WithTarget("6", "300Mi").GetContainerResources(),
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := IntegerCPUPostProcessor{}
+			got := c.Process(tt.vpa, tt.recommendation, nil)
+			assert.Equalf(t, reScale3(tt.want), reScale3(got), "Process(%v, %v, nil)", tt.vpa, tt.recommendation)
```
jbartosik, line 139: I'd rather do a direct comparison, without  `reScale3` 

It's slightly annoying to adjust all the irrelevant details of output. But I think it's better than processing got & want (future test readers will need to understand what ` reScale3` does, there's some risk that `reScale3` does something unexpected).

(I think best solution is to have matchers but those are not welcome in Golang apparently ;( )

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor.go
```diff
@@ -0,0 +1,89 @@
...
+	vpaPostProcessorPrefix           = "vpa-post-processor.kubernetes.io/"
+	vpaPostProcessorIntegerCPUSuffix = "_integerCPU"
+	vpaPostProcessorIntegerCPUValue  = "true"
+)
+
+var _ RecommendationPostProcessor = &IntegerCPUPostProcessor{}
+
+// Process apply the capping post-processing to the recommendation.
+// For this post processor the CPU value is rounded up to an integer
+func (p *IntegerCPUPostProcessor) Process(vpa *model.Vpa, recommendation *vpa_types.RecommendedPodResources, policy *vpa_types.PodResourcePolicy) *vpa_types.RecommendedPodResources {
```
jbartosik, line 43: Please replace `policy *vpa_types.PodResourcePolicy` with `*vpa_types.PodResourcePolicy` to clearly indicate the argument is not used

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor_test.go
```diff
@@ -0,0 +1,160 @@
...
+	tests := []struct {
+		name           string
+		vpa            *model.Vpa
+		recommendation *vpa_types.RecommendedPodResources
+		want           *vpa_types.RecommendedPodResources
+	}{
+		{
+			name: "No containers match",
+			vpa: &model.Vpa{Annotations: map[string]string{
+				vpaPostProcessorPrefix + "container-other" + vpaPostProcessorIntegerCPUSuffix: "true",
```
jbartosik, line 82: Please replace `"true"` with `vpaPostProcessorIntegerCPUValue` here and below

jbartosik (changes requested):

vertical-pod-autoscaler/pkg/recommender/main.go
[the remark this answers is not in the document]
dbenque: LGTM, I can use your proposal in the description

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor.go
dbenque: Because I have other features in preparation, where I use the same pattern, but the prefix and suffix values are different. The caller to these function should use the const value as param, depending on the feature.

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor_test.go
dbenque: I will give it a second try, but I had to do that because a simple:
```
assert.Equalf(t, reScale3(tt.want), reScale3(got)
```
would fail. The reason is that some other parts of the code that I depend on change the scale value. In the end the absolute values are the same, but their representation are not. Then the test fails because it consider that {value:1,scale:0} is not equal to {value:1000, scale -3}

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor.go
dbenque: While I agree with you I was trying to be consistent with the rest of the code:
https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/utils/vpa/capping.go#L118

I don't mind changing this in my post processor, I will externalise the function and add test for it.

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor_test.go
jbartosik: Yes. I know those fail if value is the same but some other things are different  (for example scale). The best solution IMO seen is to set expectations to match also on things we don't really care about (like scale).

voelzmo (changes requested):

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor.go
dbenque: done

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor.go
dbenque: done

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor_test.go
dbenque: I finally did a comparison (no more processing of the output), opening and comparing the struct and using `Cmp` function that is defined at Quantity level.

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor_test.go
dbenque: done

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+
+		for _, r := range amendedRecommendation.ContainerRecommendations {
+			if r.ContainerName != containerName {
+				continue
+			}
+			process(r.Target)
+			process(r.LowerBound)
+			process(r.UpperBound)
+			process(r.UncappedTarget)
```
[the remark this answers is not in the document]
dbenque: it make sense that the `UncappedTarget` is not touched by the capping post-processor, but I would say that the same does not apply to other post-processors than capping.
I don't think we need to keep a `UnXYZTarget` for each XYZ post-processor

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor_test.go
```diff
@@ -0,0 +1,160 @@
...
+import (
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/api/core/v1"
+	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1"
+	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/recommender/model"
+	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/test"
+	"testing"
+)
+
+func Test_extractContainerName(t *testing.T) {
```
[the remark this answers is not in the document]
dbenque: Exact, I am going to change it for consistency.
This is weird since test generators out there are creating test function name with Test_***

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor_test.go
```diff
@@ -0,0 +1,160 @@
...
+				ContainerRecommendations: []vpa_types.RecommendedContainerResources{
+					test.Recommendation().WithContainer("container1").WithTarget("9", "200Mi").GetContainerResources(),
+					test.Recommendation().WithContainer("container2").WithTarget("8.2", "300Mi").GetContainerResources(),
+				},
+			},
+		},
+		{
+			name: "2 containers, 2 matching",
+			vpa: &model.Vpa{Annotations: map[string]string{
+				vpaPostProcessorPrefix + "container1" + vpaPostProcessorIntegerCPUSuffix: "true",
```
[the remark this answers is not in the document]
dbenque: "false" and annotation not being present are equivalent indeed.

We already have the support for multiple container with the current code (list of annotation, one per container), adding another way of expressing the same thing may introduce confusion and complexity.
In the future we can add the ability to provide a list of containers within one annotation instead of a list of annotations. I am not sure what the good syntax would be. I prefer to defer such request to another PR and discuss the API and the need for it with the group.

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor.go
```diff
@@ -0,0 +1,89 @@
...
+
+const (
+	// The user interface for that post processor is an annotation on the VPA object with the following format:
+	// vpa-post-processor.kubernetes.io/{containerName}_integerCPU=true
+	vpaPostProcessorPrefix           = "vpa-post-processor.kubernetes.io/"
+	vpaPostProcessorIntegerCPUSuffix = "_integerCPU"
+	vpaPostProcessorIntegerCPUValue  = "true"
+)
+
+var _ RecommendationPostProcessor = &IntegerCPUPostProcessor{}
```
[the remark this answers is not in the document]
dbenque: It is a common pattern in Go that express the intention of the developper: the struct `IntegerCPUPostProcessor` must implement `RecommendationPostProcessor` interface.

It does not do anything at runtime, but it is really useful for understanding. Here we have interface and struct in the same program, so that might be obvious, but this turn out to be a very good habit when you create structs that implements should interface of another package (imported from other project). With a good IDE integration with Go you immediately see that you are breaking the contract.

here is an example in the kubernetes repo:
https://github.com/kubernetes/client-go/blob/1ac8d459351e21458fd1041f41e43403eadcbdba/tools/cache/store.go#L147

You can find many other examples with `git grep "var _ .* = &"` in the codebase of kube.

vertical-pod-autoscaler/pkg/recommender/main.go
```diff
@@ -86,6 +92,9 @@ func main() {
 	postProcessors := []routines.RecommendationPostProcessor{
 		&routines.CappingPostProcessor{},
 	}
+	if *postProcessorCPUasInteger {
```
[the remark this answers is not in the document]
dbenque: Good catch! Of course the Capping should be last, fixing it.

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor.go
jbartosik: Ack.

vertical-pod-autoscaler/pkg/recommender/routines/cpu_integer_post_processor.go
jbartosik: Ack.

jbartosik (commented):
I think we also need some documentation:

- Release note
- Some description of how to use this.

dbenque:
> I think we also need some documentation:
> 
> * Release note
> * Some description of how to use this.

I have added documentation in the last commit https://github.com/kubernetes/autoscaler/pull/5313/commits/17ef62801340072d7c551b05a869f3d2bedfd6a0

Where should I put information for the release note? Should I add something in the PR description?

jbartosik:
Mostly looks good, there are some things I want to make sure are consistent (mostly flags, labels, and docs). I'm not sure when I'll be able to finish the review (I might have to take some time off). I want to merge this before I cut release (#5355).

PR #5322, opened 2022-11-21
fix(cluster-autoscaler/hetzner): pre-existing volumes break scheduling

apricote:

cluster-autoscaler

#### What type of PR is this?

/kind bug

The `hcloud-csi-driver` uses the label `csi.hetzner.cloud/location` for topology. This label was not added in the response to `n.TemplateNodeInfo()`, causing cluster-autoscaler to not consider any node group for scaling when a pre-existing volume was attached to the pending pod.

This is fixed by adding the appropriatly named label to the `NodeInfo`. In practice this label is added by the `hcloud-csi-driver`.

Further details on the bug are available in the original issue: https://github.com/hetznercloud/csi-driver/pull/302

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/hetzner/hetzner_node_group.go
+++ b/cluster-autoscaler/cloudprovider/hetzner/hetzner_node_group.go
@@ -315,9 +315,10 @@ func newNodeName(n *hetznerNodeGroup) string {
 
 func buildNodeGroupLabels(n *hetznerNodeGroup) map[string]string {
 	return map[string]string{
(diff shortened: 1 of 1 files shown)


LKaemmerling (approved):

apricote:
Adding lgtm as @LKaemmerling reviewed and approved (through GitHub) but can not use the command as he is not a member of the org.

/lgtm

mwielgus (approved):
/lgtm
/approve

PR #5326, opened 2022-11-23
Stop filtering aggregatedContainerStates without samples

### LINKED ISSUE #4981: VPA doesn't provide any recommendations when Pod is in OOMKill CrashLoopBackoff right after start
<!--
Please answer these questions before submitting your bug report. Thanks!
-->

vertical-pod-autoscaler
<!--
Which autoscaling component hosted in this repository (cluster-autoscaler, vertical-pod-autoscaler, addon-resizer, helm charts) is the bug in?
-->

<!--
What version of the relevant component are you using? Either the image tag or helm chart version.
-->

Component version: 0.10.0

$ kubectl version --short
Client Version: v1.24.2
Kustomize Version: v4.5.4
Server Version: v1.23.4
</pre></details>

<!--
If you're using a cloud provider or hardware configuration as your deployment environment let us know here.
-->

VPA should be able to help with Pods which are in an OOMKill CrashLoopBackOff and raise Limits/Requests until the workload is running.

<!--
What behaviour did you expect to see?
-->

**What happened instead?**:
VPA did not give a single Recommendation for a Pod that right from the start goes into an OOMKill CrashLoopBackOff

<!--
What behaviour did see instead?
-->

Create a deployment that will be OOMKilled right after starting
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: oomkilled
spec:
  replicas: 1
  selector:
    matchLabels:
      app: oomkilled
  template:
    metadata:
      labels:
        app: oomkilled
    spec:
      containers:
      - image: gcr.io/google-containers/stress:v1
        name: stress
        command: [ "/stress"]
        args: 
          - "--mem-total"
          - "104858000"
          - "--logtostderr"
          - "--mem-alloc-size"
          - "10000000"
        resources:
          requests:
            memory: 1Mi
            cpu: 5m
          limits:
            memory: 20Mi
```
Look at the container
```
(...)
    State:          Waiting                                                                                                                                                                                                                                                                    
      Reason:       CrashLoopBackOff
    Last State:     Terminated
      Reason:       OOMKilled
      Exit Code:    137
      Started:      Mon, 20 Jun 2022 16:56:47 +0200
      Finished:     Mon, 20 Jun 2022 16:56:48 +0200
    Ready:          False
    Restart Count:  5
(...)
```

Create a VPA object for this deployment
```
apiVersion: "autoscaling.k8s.io/v1"
kind: VerticalPodAutoscaler
metadata:
  name: oomkilled-vpa
spec:
  targetRef:
    apiVersion: "apps/v1"
    kind: Deployment
    name: oomkilled
  resourcePolicy:
    containerPolicies:
      - containerName: '*'
        minAllowed:
          cpu: 5m
          memory: 10Mi
        maxAllowed:
          cpu: 1
          memory: 5Gi
        controlledResources: ["cpu", "memory"]
```

VPA _does_ observe the corresponding OOMKill events in the Recommender logs
```
I0620 14:55:04.340502       1 cluster_feeder.go:465] OOM detected {Timestamp:2022-06-20 14:53:52 +0000 UTC Memory:1048576 ContainerID:{PodID:{Namespace:default PodName:oomkilled-6868f896d6-6vfqm} ContainerName:stress}}
I0620 14:55:04.340545       1 cluster_feeder.go:465] OOM detected {Timestamp:2022-06-20 14:54:08 +0000 UTC Memory:1048576 ContainerID:{PodID:{Namespace:default PodName:oomkilled-6868f896d6-6vfqm} ContainerName:stress}}
```

VPA Status is empty
```
kind: VerticalPodAutoscaler
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"autoscaling.k8s.io/v1","kind":"VerticalPodAutoscaler","metadata":{"annotations":{},"name":"oomkilled-vpa","namespace":"default"},"spec":{"resourcePolicy":{"containerPolicies":[{"containerName":"*","controlledResources":["cpu","memory"],"maxAllowed":{"cpu":1,
"memory":"5Gi"},"minAllowed":{"cpu":"5m","memory":"10Mi"}}]},"targetRef":{"apiVersion":"apps/v1","kind":"Deployment","name":"oomkilled"}}}
  creationTimestamp: "2022-06-20T14:54:16Z"
  generation: 2
  name: oomkilled-vpa
  namespace: default
  resourceVersion: "299374"
  uid: f47d84a8-aa6e-4042-b0a4-723888720a9d
spec:
  resourcePolicy:
    containerPolicies:
    - containerName: '*'
      controlledResources:
      - cpu
      - memory
      maxAllowed:
        cpu: 1
        memory: 5Gi
      minAllowed:
        cpu: 5m
        memory: 10Mi
  targetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: oomkilled
  updatePolicy:
    updateMode: Auto
status:
  conditions:
  - lastTransitionTime: "2022-06-20T14:55:04Z"
    status: "False"
    type: RecommendationProvided
  recommendation: {}
```

VPACheckpoint doesn't record any measurements
```
kind: VerticalPodAutoscalerCheckpoint
metadata:
  creationTimestamp: "2022-06-20T14:55:04Z"
  generation: 24
  name: oomkilled-vpa-stress
  namespace: default
  resourceVersion: "304997"
  uid: 127a6331-7d1d-4ea6-b56a-63db3ee07a51
spec:
  containerName: stress
  vpaObjectName: oomkilled-vpa
status:
  cpuHistogram:
    referenceTimestamp: null
  firstSampleStart: null
  lastSampleStart: null
  lastUpdateTime: "2022-06-20T15:18:04Z"
  memoryHistogram:
    referenceTimestamp: "2022-06-22T00:00:00Z"
  version: v3
```

The Pod in CrashLoopBackOff doesn't have any PodMetrics, whereas other Pods _do_ have metrics
```
k get --raw "/apis/metrics.k8s.io/v1beta1/namespaces/default/pods/" | jq
{
  "kind": "PodMetricsList",
  "apiVersion": "metrics.k8s.io/v1beta1",
  "metadata": {},
  "items": [
    {
      "metadata": {
        "name": "hamster-96d4585b7-b9tl9",
        "namespace": "default",
        "creationTimestamp": "2022-06-20T15:24:37Z",
        "labels": {
          "app": "hamster",
          "pod-template-hash": "96d4585b7"
        }
      },
      "timestamp": "2022-06-20T15:24:01Z",
      "window": "56s",
      "containers": [
        {
          "name": "hamster",
          "usage": {
            "cpu": "498501465n",
            "memory": "512Ki"
          }
        }
      ]
    },
    {
      "metadata": {
        "name": "hamster-96d4585b7-c44j7",
        "namespace": "default",
        "creationTimestamp": "2022-06-20T15:24:37Z",
        "labels": {
          "app": "hamster",
          "pod-template-hash": "96d4585b7"
        }
      },
      "timestamp": "2022-06-20T15:24:04Z",
      "window": "57s",
      "containers": [
        {
          "name": "hamster",
          "usage": {
            "cpu": "501837091n",
            "memory": "656Ki"
          }
        }
      ]
    }
  ]
}
```

* The above `List` call is [what the VPA Recommender uses to get metrics for all the Pods](https://github.com/kubernetes/autoscaler/blob/vertical-pod-autoscaler-0.10.0/vertical-pod-autoscaler/pkg/recommender/input/metrics/metrics_client.go#L68-L69) and then increases the `TotalSamplesCount` for the individual Containers [for every CPUSample](https://github.com/kubernetes/autoscaler/blob/vertical-pod-autoscaler-0.10.0/vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state.go#L221) in that List of Podmetrics.

* [OOMKill events are recorded as MemorySamples](https://github.com/kubernetes/autoscaler/blob/vertical-pod-autoscaler-0.10.0/vertical-pod-autoscaler/pkg/recommender/model/container.go#L209), therefore, they also don't increase the `TotalSamplesCount`.

* [This container most likely doesn't get any recommendation, because its `TotalSamplesCount` is `0`](https://github.com/kubernetes/autoscaler/blob/vertical-pod-autoscaler-0.10.0/vertical-pod-autoscaler/pkg/recommender/routines/vpa.go#L34-L37)

* Seems like others have seen this as well (and tried to resolve this by switching to a different metrics source): https://github.com/kubernetes-sigs/metrics-server/issues/976#issuecomment-1076102124
* People really don't want metrics for terminated containers, these things were added intentionally:
  * metrics-server intentionally doesn't provide PodMetrics for Pods with `terminated` containers – this resulted in Pods with `init-containers` not having any metrics when a `cAdvisor` refactoring  included init containers in the kubelet summary API again [[1](https://github.com/kubernetes-sigs/metrics-server/issues/792), [2](https://github.com/kubernetes-sigs/metrics-server/issues/796)]
  * kubelet is meant to only provide metrics for [non-terminated Pods](https://github.com/kubernetes/kubernetes/pull/79409) and for [`running` Containers](https://github.com/kubernetes/kubernetes/pull/83504)


<!--
If possible, provide a recipe for reproducing the error.
A detailed sequence of steps describing what to do to observe the issue is good.
A complete runnable bash shell script is best.
-->

On the same cluster, the `hamster` example works perfectly fine and gets recommendations as expected, so this is not a general issue with the VPA.

I just for fun applied this patch which increases the `TotalSamplesCount` when a memory sample (i.e. also an OOMKill sample) is added and afterwards the above Pod gets a recommendation and can run normally – as expected. I understand that the fix cannot be as simple as that, otherwise we would add two samples for every regular PodMetric (which contains both, CPU _and_ memory), and existing implementations assume otherwise, I guess, but this is just to show that `TotalSamplesCount` seems to be the blocker in this situation.
```
diff --git a/vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state.go b/vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state.go
index 3facbe37e..7accd072e 100644
--- a/vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state.go
+++ b/vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state.go
@@ -184,6 +184,7 @@ func (a *AggregateContainerState) AddSample(sample *ContainerUsageSample) {
        case ResourceCPU:
                a.addCPUSample(sample)
        case ResourceMemory:
+               a.TotalSamplesCount++
                a.AggregateMemoryPeaks.AddSample(BytesFromMemoryAmount(sample.Usage), 1.0, sample.MeasureStart)
        default:
                panic(fmt.Sprintf("AddSample doesn't support resource '%s'", sample.Resource))
```


<!--
Is there anything else you think we should know? Configuration of the component (be careful what you post here if so)? Relevant logs?
-->

voelzmo:

vertical-pod-autoscaler

#### What type of PR is this?

/kind bug

Stop filtering `aggregatedContainerState`s with `TotalSampleCount==0`. These `aggregateContainerState`s *can* contain OOMSamples, which don't increase the `TotalSampleCount` but make it possible to give a better memory recommendation. This change also increases the timeout for the OOMKill e2e test as it takes ~10 minutes for the recommendations to reach the expected range.

Fixes #4981 

I'm not entirely sure if there is a good reason why `aggregateContainerStates` without PodMetric samples were ignored besides trying to optimize. Maybe someone else has context if there's an additional reason behind it why we shouldn't be removing this condition?

I looked at other options to solve this, but they seem much more complicated. 

##### Alternative A: Adding separate `SampleCounts` for CPU and memory
The VPA is built with the assumption that a sample contains both, CPU and memory data, therefore only a single `TotalSampleCount` exists in `aggregateContainerState`. We could move this further down and give each resource its own counter, *but* confidence is computed based on `TotalSampleCount`: therefore this would mean that CPU and memory can have different confidence and therefore different upper and lower bounds. This seems way too intrusive.

##### Alternative B: Increase `TotalSampleCount` when an OOMSample is added
Do something similar to the quick hack I discussed in #4981, but only for OOMSamples, not all memory samples.

```
diff --git a/vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state.go b/vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state.go
index 3facbe37e..7accd072e 100644
--- a/vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state.go
+++ b/vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state.go
@@ -184,6 +184,7 @@ func (a *AggregateContainerState) AddSample(sample *ContainerUsageSample) {
        case ResourceCPU:
                a.addCPUSample(sample)
        case ResourceMemory:
+               a.TotalSamplesCount++
                a.AggregateMemoryPeaks.AddSample(BytesFromMemoryAmount(sample.Usage), 1.0, sample.MeasureStart)
        default:
                panic(fmt.Sprintf("AddSample doesn't support resource '%s'", sample.Resource))
```

Downside: When many OOMSamples are recorded in a short period (e.g. when a Pod suddenly spikes in memory or has waaaaayyyyy too low intial resources defined), the confidence for CPU recommendations increases as well – which is the opposite of what's happening: In a situation like this we *don't* get PodMetrics and therefore no CPU metrics, so this also seems incorrect.

```release-note
Fixed a bug which could lead to the recommender not giving any recommendations for Pods in OOMKill CrashLoopBackoff
```

```docs

```

### DIFF
--- a/vertical-pod-autoscaler/e2e/v1/autoscaling_utils.go
+++ b/vertical-pod-autoscaler/e2e/v1/autoscaling_utils.go
@@ -66,7 +66,7 @@ const (
 
 var (
 	resourceConsumerImage = imageutils.GetE2EImage(imageutils.ResourceConsumer)
-	stressCommand         = []string{"/stress", "--mem-total", "10000000000", "--logtostderr", "--mem-alloc-size", "40000"}
+	stressCommand         = []string{"/stress", "--mem-total", "10000000000", "--logtostderr", "--mem-alloc-size", "50000"}
 )
 
 var (
--- a/vertical-pod-autoscaler/e2e/v1/autoscaling_utils.go
+++ b/vertical-pod-autoscaler/e2e/v1/autoscaling_utils.go
@@ -66,7 +66,7 @@ const (
 
 var (
 	resourceConsumerImage = imageutils.GetE2EImage(imageutils.ResourceConsumer)
-	stressCommand         = []string{"/stress", "--mem-total", "10000000000", "--logtostderr", "--mem-alloc-size", "8000"}
+	stressCommand         = []string{"/stress", "--mem-total", "10000000000", "--logtostderr", "--mem-alloc-size", "40000"}
 )
 
 var (
@@ -443,8 +443,8 @@ func runOomingReplicationController(c clientset.Interface, ns, name string, repl
 		Timeout:     timeoutRC,
 		Replicas:    replicas,
 		Annotations: make(map[string]string),
-		MemRequest:  1024 * 1024 * 1024,
-		MemLimit:    1024 * 1024 * 1024,
+		MemRequest:  1024 * 1024 * 300,
+		MemLimit:    1024 * 1024 * 500,
 	}
 
 	dpConfig := testutils.DeploymentConfig{
--- a/vertical-pod-autoscaler/e2e/v1/full_vpa.go
+++ b/vertical-pod-autoscaler/e2e/v1/full_vpa.go
@@ -43,7 +43,7 @@ const (
 	// the initial values should be outside minimal bounds
 	initialCPU     = int64(10) // mCPU
 	initialMemory  = int64(10) // MB
-	oomTestTimeout = 8 * time.Minute
+	oomTestTimeout = 12 * time.Minute
 )
 
 var _ = FullVpaE2eDescribe("Pods under VPA", func() {
--- a/vertical-pod-autoscaler/pkg/recommender/routines/vpa.go
+++ b/vertical-pod-autoscaler/pkg/recommender/routines/vpa.go
@@ -31,7 +31,7 @@ func GetContainerNameToAggregateStateMap(vpa *model.Vpa) model.ContainerNameToAg
 		containerResourcePolicy := api_utils.GetContainerResourcePolicy(containerName, vpa.ResourcePolicy)
 		autoscalingDisabled := containerResourcePolicy != nil && containerResourcePolicy.Mode != nil &&
 			*containerResourcePolicy.Mode == vpa_types.ContainerScalingModeOff
-		if !autoscalingDisabled && aggregatedContainerState.TotalSamplesCount > 0 {
+		if !autoscalingDisabled {
 			aggregatedContainerState.UpdateFromPolicy(containerResourcePolicy)
 			filteredContainerNameToAggregateStateMap[containerName] = aggregatedContainerState
 		}


jbartosik (commented):
Sorry for slow reply, I'm not sure why I missed this PR.

General approach looks good to me, there are some things we need to do to make sure this works properly.

We use `TotalSamplesCount` to decide if [`isEmpty`](https://github.com/kubernetes/autoscaler/blob/33341a1be840d49da48788d488f8a1a118133b67/vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state.go#L274) should return true, which we later use to decide if we should delete the container state ([here](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/recommender/model/cluster.go#L367) directly and inside call to `isExpired`.

I'm worried that this will result in OOMs only briefly having an effect and then being cleaned up.

jbartosik:
Also I wonder why do you increase timeout for the OOM test?

voelzmo:
> Also I wonder why do you increase timeout for the OOM test?

It's buried in the long PR description, sorry: 
> This change also increases the timeout for the OOMKill e2e test as it takes ~10 minutes for the recommendations to reach the expected range.

When I watched the VPA status changes during the test, I could see that the recommendations were increasing, it just took longer than currently configured. Without the change, the recommendations didn't move at all.

voelzmo:
> Sorry for slow reply, I'm not sure why I missed this PR.
> 
> General approach looks good to me, there are some things we need to do to make sure this works properly.
> 
> We use `TotalSamplesCount` to decide if [`isEmpty`](https://github.com/kubernetes/autoscaler/blob/33341a1be840d49da48788d488f8a1a118133b67/vertical-pod-autoscaler/pkg/recommender/model/aggregate_container_state.go#L274) should return true, which we later use to decide if we should delete the container state ([here](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/recommender/model/cluster.go#L367) directly and inside call to `isExpired`.
> 
> I'm worried that this will result in OOMs only briefly having an effect and then being cleaned up.

That's a good point! For the aggregate state to be garbage collected, the Pod needs to be detected a 'non contributive' in addition to the TotalSampleCount being 0 – which shouldn't be the case as long as the Deployment still exists, right?

I wasn't sure how likely this would happen, but it definitely could! Do you have concrete suggestions on how to improve the solution so we don't run into this case?

voelzmo:
ping @jbartosik is there something missing before merging this?

jbartosik (approved):
/approve

Thanks for the reminder.

PR #5328, opened 2022-11-25
Add note regarding GPU label for the CAPI provider

yankcrime:

cluster-autoscaler

#### What type of PR is this?

/kind documentation

cluster-autoscaler takes into consideration the time that a node takes to initialise a GPU resource on a node, as long as a particular label is in place.  This label differs from provider to provider, and is documented in some cases but not for CAPI.

This commit adds a note with the specific label that should be applied when a node is instantiated.

```release-note
NONE
```


mwielgus (approved):
/lgtm
/approve

mwielgus:
cc: @elmiko

elmiko (commented):
/lgtm

thanks @yankcrime !

PR #5330, opened 2022-11-25
Use ScaleDownSetProcessor.GetNodesToRemove in scale down planner to filter  NodesToDelete.

olagacek:

cluster-autoscaler

#### What type of PR is this?

/kind feature

Allows filtering of scale down candidates in scale down planner.

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/scaledown/legacy/legacy.go
+++ b/cluster-autoscaler/core/scaledown/legacy/legacy.go
@@ -254,7 +254,7 @@ func (sd *ScaleDown) mapNodesToStatusScaleDownNodes(nodes []*apiv1.Node, nodeGro
 }
 
 // NodesToDelete selects the nodes to delete for scale down.
-func (sd *ScaleDown) NodesToDelete(currentTime time.Time, pdbs []*policyv1.PodDisruptionBudget) (empty, drain []*apiv1.Node, res status.ScaleDownResult, err errors.AutoscalerError) {
+func (sd *ScaleDown) NodesToDelete(currentTime time.Time, pdbs []*policyv1.PodDisruptionBudget) (_, drain []*apiv1.Node, res status.ScaleDownResult, err errors.AutoscalerError) {
 	_, drained := sd.nodeDeletionTracker.DeletionsInProgress()
 	if len(drained) > 0 {
 		return nil, nil, status.ScaleDownInProgress, nil
@@ -288,10 +288,10 @@ func (sd *ScaleDown) NodesToDelete(currentTime time.Time, pdbs []*policyv1.PodDi
 	}
 	candidateNames := make([]string, 0, len(empty)+len(nonEmpty))
 	for _, n := range empty {
-		candidateNames = append(candidateNames, n.Name)
+		candidateNames = append(candidateNames, n.Node.Name)
 	}
 	for _, n := range nonEmpty {
-		candidateNames = append(candidateNames, n.Name)
+		candidateNames = append(candidateNames, n.Node.Name)
 	}
 
 	if len(candidateNames) == 0 {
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/core/scaledown/unneeded/nodes.go
```diff
@@ -117,21 +117,20 @@ func (n *Nodes) Drop(node string) {
 // RemovableAt returns all nodes that can be removed at a given time, divided
 // into empty and non-empty node lists, as well as a list of nodes that were
 // unneeded, but are not removable, annotated by reason.
-func (n *Nodes) RemovableAt(context *context.AutoscalingContext, ts time.Time, resourcesLeft resource.Limits, resourcesWithLimits []string, as scaledown.ActuationStatus) (empty, needDrain []*apiv1.Node, unremovable []*simulator.UnremovableNode) {
+func (n *Nodes) RemovableAt(context *context.AutoscalingContext, ts time.Time, resourcesLeft resource.Limits, resourcesWithLimits []string, as scaledown.ActuationStatus) (empty, needDrain []simulator.NodeToBeRemoved, unremovable []*simulator.UnremovableNode) {
```
x13n, line 120: Since you're already changing the function signature to return `simulator.NodeToBeRemoved`, I think it makes sense to just put empty & non-empty on a single list and let the caller worry about splitting them.

olagacek: I don't have strong opinion on this, but both callers of this method require distinction between empty and non-empty nodes to remove. Does it make sense then to merge the lists into one?

x13n: I think this results in a slightly cleaner interface. Don't have a too strong opinion on this either though, so up to you.

x13n:
LGTM, putting on hold to give you a chance to consider my comment. Feel free to cancel hold if you disagree.

/lgtm
/hold

cluster-autoscaler/core/scaledown/unneeded/nodes.go
olagacek: If there is no strong opinion here, I would leave it as is, given we need the distinction.

cluster-autoscaler/core/scaledown/legacy/legacy.go
```diff
@@ -282,7 +282,10 @@ func (sd *ScaleDown) NodesToDelete(currentTime time.Time, pdbs []*policyv1.PodDi
 	}
 
 	scaleDownResourcesLeft := sd.resourceLimitsFinder.LimitsLeft(sd.context, allNodes, resourceLimiter, currentTime)
-	empty, nonEmpty, unremovable := sd.unneededNodes.RemovableAt(sd.context, currentTime, scaleDownResourcesLeft, resourceLimiter.GetResources(), sd.nodeDeletionTracker)
+	emptyToRemove, nonEmpty, unremovable := sd.unneededNodes.RemovableAt(sd.context, currentTime, scaleDownResourcesLeft, resourceLimiter.GetResources(), sd.nodeDeletionTracker)
+	for _, emptyNode := range emptyToRemove {
```
towca, line 286: Do we need the `empty` list? Isn't it only used to extract the name in the loop below? Couldn't we just do the same thing as with `nonEmpty`?

olagacek: Actually it is also used as a return type, so we do need that.

towca: I mean we're setting the variable here, but `empty` isn't returned anywhere from this function, right? It's only used to create `candidateNames`. Or am I reading things wrong?

olagacek: discussed offline, removed empty

PR #5333, opened 2022-11-28
Use PdbRemainingDisruptions in Planner

yaroslava-serdiuk:

cluster-autoscaler

#### What type of PR is this?

/kind feature


cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -289,14 +292,15 @@ func (p *Planner) categorizeNodes(podDestinations map[string]bool, scaleDownCand
...
-		removable, unremovable := p.rs.SimulateNodeRemoval(node, podDestinations, p.latestUpdate, pdbs)
+		removable, unremovable := p.rs.SimulateNodeRemoval(node, podDestinations, p.latestUpdate, p.pdbRemainingDisruptions.GetPdbs())
 		if unremovable != nil {
 			unremovableCount += 1
 			p.unremovableNodes.AddTimeout(unremovable, unremovableTimeout)
 		}
 		if removable != nil {
 			delete(podDestinations, removable.Node.Name)
 			removableList = append(removableList, *removable)
+			p.pdbRemainingDisruptions.Update(removable.PodsToReschedule)
```
x13n, line 303: Please don't ignore the error returned by `Update`. This should be logged at error severity.

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -289,14 +292,15 @@ func (p *Planner) categorizeNodes(podDestinations map[string]bool, scaleDownCand
 		// TODO(x13n): break on timeout. Figure out how to handle nodes
 		// identified as unneeded in previous iteration, but now
 		// skipped due to timeout.
-		removable, unremovable := p.rs.SimulateNodeRemoval(node, podDestinations, p.latestUpdate, pdbs)
+		removable, unremovable := p.rs.SimulateNodeRemoval(node, podDestinations, p.latestUpdate, p.pdbRemainingDisruptions.GetPdbs())
```
x13n, line 295: This cannot rely on PDB checking inside `SimulateNodeRemoval`, `CanDisrupt` should be explicitly checked.

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -24,6 +24,7 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/context"
 	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown"
 	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown/eligibility"
+	pdb_disruptions "k8s.io/autoscaler/cluster-autoscaler/core/scaledown/pdb"
```
x13n, line 27: nit: Please avoid using underscores in imports.

x13n (changes requested):

cluster-autoscaler/core/scaledown/pdb/pdb.go
```diff
@@ -41,25 +39,31 @@ func NewPdbRemainingDisruptions(pdbs []*policyv1.PodDisruptionBudget) *PdbRemain
 	return &PdbRemainingDisruptions{pdbsCopy}
 }
 
-// CanDisrupt return if the pod can be removed.
-func (p *PdbRemainingDisruptions) CanDisrupt(pods []*apiv1.Pod) (bool, *drain.BlockingPod) {
+// CanDisrupt return if the set of pods can be removed.
+// isRisky means that the number of pods that belongs to one PDB,
```
x13n, line 43: nit: "isRisky means that not all pods can be removed in parallel."? Since this is exactly what it means. Maybe it should be called `inParallel` (with true/false meaning inverted) or something?

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -148,12 +152,15 @@ func (p *Planner) NodesToDelete() (empty, needDrain []*apiv1.Node) {
 		append(emptyRemovable, needDrainRemovable...),
 		p.context.AutoscalingOptions.MaxScaleDownParallelism)
 	for _, nodeToRemove := range nodesToRemove {
-		if len(nodeToRemove.PodsToReschedule) > 0 {
+		if len(nodeToRemove.PodsToReschedule) > 0 && !nodeToRemove.IsRisky {
 			needDrain = append(needDrain, nodeToRemove.Node)
+		} else if len(nodeToRemove.PodsToReschedule) > 0 && nodeToRemove.IsRisky {
```
x13n, line 157: I think it would be more readable to have nested `if` instead of repeating the condition for whether the node needs drain:

```
if len(nodeToRemove.PodsToReschedule) > 0 {
    if nodeToRemove.IsRisky {
        (...)
    } else {
        (...)
    }
} else {
    (...)
}
```

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -289,15 +296,34 @@ func (p *Planner) categorizeNodes(podDestinations map[string]bool, scaleDownCand
...
+						BlockingPod: blockingPod,
+					}
+				}
+				continue
+			}
+			if isRisky {
+				removable.IsRisky = true
+			}
 			delete(podDestinations, removable.Node.Name)
+			err := p.pdbRemainingDisruptions.Update(removable.PodsToReschedule)
```
x13n, line 316: `Update` shouldn't really be able to return an error now. Building a selector for each pdb can be done once during `pdbRemainingDisruptions` creation, so the error is handled there. We don't need to build the same selector twice per node anyway (once in `CanDisrupt`, then again in `Update`).

cluster-autoscaler/simulator/drain.go
```diff
@@ -71,7 +71,8 @@ func GetPodsToMove(nodeInfo *schedulerframework.NodeInfo, deleteOptions NodeDele
 }
 
 func checkPdbs(pods []*apiv1.Pod, pdbs []*policyv1.PodDisruptionBudget) (*drain.BlockingPod, error) {
-	// TODO: make it more efficient.
+	// TODO: remove it after depricating legacy scale down.
```
x13n, line 74: nit: s/depricating/deprecating/

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -289,15 +296,34 @@ func (p *Planner) categorizeNodes(podDestinations map[string]bool, scaleDownCand
...
-		if unremovable != nil {
-			unremovableCount += 1
-			p.unremovableNodes.AddTimeout(unremovable, unremovableTimeout)
-		}
+		removable, unremovable := p.rs.SimulateNodeRemoval(node, podDestinations, p.latestUpdate, p.pdbRemainingDisruptions.GetPdbs())
 		if removable != nil {
+			ok, isRisky, blockingPod := p.pdbRemainingDisruptions.CanDisrupt(removable.PodsToReschedule)
+			if !ok {
+				if blockingPod != nil {
+					unremovable = &simulator.UnremovableNode{
```
x13n, line 304: There is quite a lot of nesting here, can you move some of it to a helper function?

cluster-autoscaler/core/scaledown/pdb/pdb.go
x13n: Sure, but this is a problem of scale down implementation. `PdbRemainingDisruptions` just tells the caller whether the pods can be removed in parallel and it is up to the caller to decide whether or not to try and start node removal. The caller (i.e. scale down planner) is free to say `risky := !inParallel` in the code.

cluster-autoscaler/core/scaledown/planner/planner.go
yaroslava-serdiuk: I removed `if blockingPod != nil` block, since now for removable pod we need to check only isRisky value, seems it looks better now, WDYT?

cluster-autoscaler/core/scaledown/pdb/pdb.go
yaroslava-serdiuk: I don't think we are going to use PdbRemainingDisruptions somewhere else, so I would like to use the same logic inside caller and implementation. Having risky := !inParallel add unnecessary complexity as for me.

cluster-autoscaler/core/scaledown/planner/planner.go
x13n: I'm a fan of short functions with meaningful names, but won't insist, this looks good enough now.

cluster-autoscaler/core/scaledown/pdb/pdb.go
x13n: `PdbRemainingDisruptions` is generic enough to be useful beyond this class, but I guess we can always extract it later when actually needed.

cluster-autoscaler/core/scaledown/pdb/pdb.go
```diff
@@ -41,42 +39,50 @@ func NewPdbRemainingDisruptions(pdbs []*policyv1.PodDisruptionBudget) *PdbRemain
...
 			if pod.Namespace == pdb.Namespace && selector.Matches(labels.Set(pod.Labels)) {
-				if pdb.Status.DisruptionsAllowed < 1 {
-					return fmt.Errorf("Pod can't be removed, pdb is blocking by pdb %s, disruptionsAllowed: %v", pdb.GetNamespace()+"/"+pdb.GetName(), pdb.Status.DisruptionsAllowed)
-				}
 				pdb.Status.DisruptionsAllowed -= 1
 			}
 		}
 	}
-	return nil
+	return
```
x13n, line 82: nit: `return` here is superfluous.

cluster-autoscaler/core/scaledown/pdb/pdb.go
```diff
@@ -41,42 +39,50 @@ func NewPdbRemainingDisruptions(pdbs []*policyv1.PodDisruptionBudget) *PdbRemain
...
 }
 
 // Update make updates the remaining disruptions for pdb.
-func (p *PdbRemainingDisruptions) Update(pods []*apiv1.Pod) error {
+func (p *PdbRemainingDisruptions) Update(pods []*apiv1.Pod) {
 	for _, pdb := range p.pdbs {
 		selector, err := metav1.LabelSelectorAsSelector(pdb.Spec.Selector)
 		if err != nil {
-			return err
+			return
```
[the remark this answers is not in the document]
yaroslava-serdiuk: Yeah, good idea

cluster-autoscaler/core/scaledown/pdb/pdb.go
```diff
@@ -17,66 +17,71 @@ limitations under the License.
...
 	}
-	return &PdbRemainingDisruptions{pdbsCopy}
+	return &PdbRemainingDisruptions{pdbsCopy, selectors}, nil
 }
 
-// CanDisrupt return if the pod can be removed.
-func (p *PdbRemainingDisruptions) CanDisrupt(pods []*apiv1.Pod) (bool, *drain.BlockingPod) {
+// CanDisrupt return if the set of pods can be removed.
+// inParallel indicates that the pods could not be removed in parallel.
+// If inParalles == false evict this set of pods from node could fail due to drain timeout.
```
x13n, line 50: nit: s/inParalles/inParallel/

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -97,20 +99,26 @@ func New(context *context.AutoscalingContext, processors *processors.Autoscaling
 // up-to-date information about the cluster.
 // Planner will evaluate scaleDownCandidates in the order provided here.
 func (p *Planner) UpdateClusterState(podDestinations, scaleDownCandidates []*apiv1.Node, as scaledown.ActuationStatus, pdb []*policyv1.PodDisruptionBudget, currentTime time.Time) errors.AutoscalerError {
+	var err error
 	p.latestUpdate = currentTime
 	p.actuationStatus = as
+	p.pdbRemainingDisruptions, err = pdbdisruptions.NewPdbRemainingDisruptions(pdb)
+	if err != nil {
+		p.CleanUpUnneededNodes()
+		return errors.ToAutoscalerError(errors.UnexpectedScaleDownStateError, err)
```
x13n, line 108: I think this should just be internal error. Malformed PDB doesn't match the description of `UnexpectedScaleDownStateError`.

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -149,11 +158,16 @@ func (p *Planner) NodesToDelete() (empty, needDrain []*apiv1.Node) {
 		p.context.AutoscalingOptions.MaxScaleDownParallelism)
 	for _, nodeToRemove := range nodesToRemove {
 		if len(nodeToRemove.PodsToReschedule) > 0 {
-			needDrain = append(needDrain, nodeToRemove.Node)
+			if nodeToRemove.IsRisky {
```
x13n, line 161: This check is too late: `GetNodesToRemove` above will already drop some nodes, so putting risky nodes at the end needs to happen before that.

yaroslava-serdiuk: Thanks!

x13n:
Looks good, just a minor comment. Putting on hold, but feel free to disregard if you don't agree.

/lgtm
/hold

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -140,6 +148,17 @@ func (p *Planner) NodesToDelete() (empty, needDrain []*apiv1.Node) {
 	for _, u := range unremovable {
 		p.unremovableNodes.Add(u)
 	}
+	// Sort risky and not risky nodes for drain.
```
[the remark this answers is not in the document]
yaroslava-serdiuk: Done

PR #5336, opened 2022-11-29
update vendor to v1.26.0-rc.1

liggitt:

cluster-autoscaler

#### What type of PR is this?

/kind cleanup

Updates to 1.26 rc.1 library versions, which allows dropping openstack and heketi libraries (xref https://github.com/kubernetes/community/issues/6992#issuecomment-1331101800)

```release-note
NONE
```

```docs

```


liggitt:
cc @jayantjain93, @MaciekPytel

liggitt:
uh.... staging repo v0.26.0-rc.0 tag content doesn't match k/k v1.26.0-rc.0

compare https://github.com/kubernetes/apimachinery/blob/v0.26.0-rc.0/pkg/apis/meta/v1/validation/validation.go and https://github.com/kubernetes/kubernetes/blob/v1.26.0-rc.0/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go

https://github.com/kubernetes/autoscaler/actions/runs/3577027159 is failing with `undefined: validation.LabelSelectorValidationOptions` because the published apimachinery tag isn't right

liggitt:
reported ^^ in https://github.com/kubernetes/kubernetes/issues/114192

Issue #4016, opened 2021-04-15
Restart-free update of pod requests


noBlubb:
Hey everyone,

as I gather the VPA currently cannot update pods without recreating them:
> Once restart free ("in-place") update of pod requests is available
> from README

and neither can the GKE vertical scaler:

> Due to Kubernetes limitations, the only way to modify the resource requests of a running Pod is to recreate the Pod
> from https://cloud.google.com/kubernetes-engine/docs/concepts/verticalpodautoscaler#vertical_pod_autoscaling_in_auto_mode

Unfortunately, I was unable to learn the specific limitation from this (other than the mere absence of any such feature?) nor timeline for this to appear in VPA or how to contribute on this if possible. Could you please outline what is missing in VPA for this to be implemented?

Best regards,
Raffael

morganchristiansson:
Would be nice with more details on the status feature. I would guess it's limitation in Kubernetes or from a lower level like containerd or kernel?

bskiba:
At this moment this is a Kubernetes limitation (kernel and container runtime already supports resizing containers). There is work needed in scheduler, kubelet, core API so a pretty cross-cutting problem. Also a lot of systems assumed pod sized are immutable for a long time so there is need to untangle those as well.

There is ongoing work in Kubernetes to provide in-place pod resizes (Example: https://github.com/kubernetes/enhancements/pull/1883). Once that work completes VPA will be able to take advantage of that.

jmo-qap:
https://github.com/kubernetes/kubernetes/pull/102884

voelzmo:
/remove-lifecycle stale
Feature didn't make it in 1.26, but now targeted for 1.27 ;)

frivoire:
This issue seems to be a duplicate of: https://github.com/kubernetes/autoscaler/issues/5046
Shouldn't we close one of those 2 issues ?

jbartosik:
https://github.com/kubernetes/kubernetes/pull/102884/ merged today. I will resume working on using that in VPA

@wangchen615 @voelzmo FYI

voelzmo:
/retitle "Support in-place Pod vertical scaling in VPA"

voelzmo:
args, don't take the quotes too literally, dear bot! 🙈 

/retitle Support in-place Pod vertical scaling in VPA

sftim:
Also see https://github.com/kubernetes/kubernetes/issues/116214

sftim:
[...]

PR #5342, opened 2022-11-30
Added AEP for multi-dimensional pod autoscaler

James-QiuHaoran:

This PR adds a new component which is a multi-dimensional pod autoscaler.

#### What type of PR is this?

/kind feature

Currently, Horizontal Pod Autoscaler (HPA) and Vertical Pod Autoscaler (VPA) control the scaling actions separately as independent controllers to determine the resource allocation for a containerized application.
Due to the independence of these two controllers, when they are configured to optimize the same target, e.g., CPU usage, they can lead to an awkward situation where HPA tries to spin more pods based on the higher-than-threshold CPU usage while VPA tries to squeeze the size of each pod based on the lower CPU usage (after scaling out by HPA).
The final outcome would be a large number of small pods created for the workloads.
Manual fine-tuning the timing to do vertical/horizontal scaling and prioritization are usually needed for syncronization of the HPA and VPA.

We propose a Multi-dimensional Pod Autoscaling (MPA) framework that combines the actions of vertical and horizontal autoscaling in a single action but separates the actuation completely from the controlling algorithms.
It consists of three controllers (i.e., a recommender, an updater, and an admission controller) and an MPA API (i.e., a CRD object or CR) that connects the autoscaling recommendations to actuation.
The multidimensional scaling algorithm is implemented in the recommender.
The scaling decisions derived from the recommender are stored in the MPA object.
The updater and the admission controller retrieve those decisions from the MPA object and actuate those vertical and horizontal actions.
Our proposed MPA (with the separation of recommendations from actuation) allows developers to replace the default recommender with their alternative customized recommender, so developers can provide their own recommender implementing advanced algorithms that control both scaling actions across different resource dimensions.

N/A

N/A

NONE

- KEP: https://github.com/James-QiuHaoran/autoscaler/blob/master/multidimensional-pod-autoscaler/KEP.md


Shubham82:
Hi @James-QiuHaoran 
Please sign the CLA, PR will not be reviewed and merged without the CLA.

James-QiuHaoran:
> Hi @James-QiuHaoran Please sign the CLA, PR will not be reviewed and merged without the CLA.

Hi @Shubham82 Thank you for the reminder. I signed the CLA just now as an individual, as my university (UIUC) appears to not have a manager managing CLA.

James-QiuHaoran:
Kindly following up with this review process. Are there any comments/questions regarding the KEP? @gjtempleton @mwielgus @MaciekPytel

multidimensional-pod-autoscaler/KEP.md
```diff
@@ -0,0 +1,499 @@
...
+
+[multidimensional Pod scaling service]: https://cloud.google.com/kubernetes-engine/docs/how-to/multidimensional-pod-autoscaling
+
+**Metric Server API.** The Metric Server API serves both default metrics or custom metrics associated with any Kubernetes objects. Custom metrics could be the application latency, throughput, or any other application-specific metrics. HPA already consumes metrics from such [a variety of metric APIs] (e.g., `metrics.k8s.io` API for resource metrics provided by metrics-server, `custom.metrics.k8s.io` API for custom metrics provided by "adapter" API servers provided by metrics solution vendors, and the `external.metrics.k8s.io` API for external metrics provided by the custom metrics adapters as well. A popular choice for the metrics collector is Prometheus. The metrics are then used by the MPA Recommender for making autoscaling decisions.
```
gjtempleton, line 128: I'm not sure about `Metric Server API`, feels like an overloaded term and readers may assume that all these APIs are being served by the [Metrics Server](https://github.com/kubernetes-sigs/metrics-server)? Would `Metric Serving APIs` or `Metrics APIs` maybe be clearer?

multidimensional-pod-autoscaler/KEP.md
```diff
@@ -0,0 +1,499 @@
...
+
+### Action Actuation Implementation
+
+To actuate the decisions without losing availability, we plan to (1) evict pods with min-replicas configured and update Pod sizes with the web-hooked admission controller (for vertical scaling), and (2) add or remove replicas (for horizontal scaling).
+We use a web-hooked admission controller to manage vertical scaling because if the actuator directly updates the vertical scaling configurations through deployment, it will potentially overload etcd (as vertical scaling might be quite frequent).
+MPA Admission Controller intercepts Pod creation requests and rewrites the request by applying recommended resources to the Pod spec.
+We do not use the web-hooked admission controller to manage the horizontal scaling as it could slow down the pod creation process.
+In the future when the in-place vertical resizing is enabled, we can remove the web-hooked admission controller and only have the updater.
```
gjtempleton, line 144: Unclear here, is the dependency on the VPA supporting in-place updates of container resources, or just the underlying kubelet behaviour?

Could we add a diagram of what that future state would look like?

multidimensional-pod-autoscaler/KEP.md
```diff
@@ -0,0 +1,499 @@
...
+Cons:
+- Webhooks introduce extra overhead for vertical scaling operations (can be avoided after in-place resizing of pod is enabled without eviction)
+- Vertical and horizontal scaling executions are separated (can be avoided after in-place resizing of pod is enabled without eviction)
+- State changes in pod sizes are not persisted (too much to keep in etcd, could use Prometheus to store pod state changes)
+
+### Action Recommendation Implementation
+
+To generate the vertical scaling action recommendation, we reuse VPA libraries as much as possible to implement scaling algorithm integrated with the newly generated MPA API code.
+To do that, we need to update accordingly the code which reads and updates the VPA objects to be interacting with the MPA objects.
+To generate the horizontal scaling action recommendation, we reuse HPA libraries, integrating with the MPA API code, to reads and updates the MPA objects.
```
gjtempleton, line 163: Are we certain the existing HPA code can be reused as a library, or will this potentially require changes in k/k to make this code more easily reusable? Is there a risk that we become dependent on guarantees that the HPA k/k code doesn't provide?

multidimensional-pod-autoscaler/KEP.md
```diff
@@ -0,0 +1,499 @@
...
+
+<!--
+This section must be completed when targeting beta to a release.
+-->
+
+#### Does this feature depend on any specific services running in the cluster?
+
+MPA relies on cluster-level service metrics-server and the API server needs to support Mutating Webhooks.
+
+### Scalability
```
gjtempleton, line 367: We call out some concerns around scalability/frequency of updates and potential impacts on etcd higher up, could we make a first pass at addressing some of the scalability questions based on the thinking you've already done here?

jbartosik (commented):
I don't have any specific comments yet but I wanted to let you know that I've read the KEP and I hope I'll be able to come back with some specific comments soon.

multidimensional-pod-autoscaler/KEP.md
```diff
@@ -0,0 +1,499 @@
...
+However, existing VPA/HPA designs cannot accommodate such requirements.
+Manual fine-tuning the timing or frequency to do vertical/horizontal scaling and prioritization are usually needed for synchronization of the HPA and VPA.
+
+[HPA]: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/
+[VPA]: https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler
+[not recommended]: https://cloud.google.com/kubernetes-engine/docs/concepts/horizontalpodautoscaler
+
+### Goals
+
+- Design and implement a holistic framework with a set of controllers to achieve multi-dimensional pod autoscaling (MPA).
```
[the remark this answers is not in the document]
James-QiuHaoran: SGTM! The separation of decision recommendation and execution enables users to replace the default recommender with their customized recommender. I'll add this goal explicitly.

multidimensional-pod-autoscaler/KEP.md
James-QiuHaoran: I agree that the term `Metric Server API` here is inappropriate, and it's used to refer to the specific Metrics Server. Thanks for pointing it out! I'll change the term to `Metrics APIs` for clarity.

multidimensional-pod-autoscaler/KEP.md
James-QiuHaoran: The dependency is on the underlying kubelet behavior, which is described in [this thread](https://github.com/kubernetes/enhancements/issues/1287). I'll add a diagram here and clarify the dependency.

James-QiuHaoran:
Hi @gjtempleton, I've updated the KEP by addressing your questions/comments. Please review the updated KEP and let me know if there is any concern. Thank you!

jbartosik (commented):
I left a few small comments.

But I have one bigger question. I'm not sure what will be implemented as result of this KEP.

- Refactoring VPA and HPA so that they can be used a libraries?
- MPA API?
- MPA recommender / updater / admission controller?

multidimensional-pod-autoscaler/KEP.md
```diff
@@ -0,0 +1,519 @@
...
+
+[<img src="./kep-imgs/mpa-design.png" width="700"/>](./kep-imgs/mpa-design.png "MPA Design Overview")
+
+**MPA API.** Application owners specify the autoscaling configurations which include (1) whether they only want to know the recommendations from MPA or they want MPA to directly actuate the autoscaling decisions; (2) application SLAs (e.g., in terms of latency or throughput); (3) any custom metrics if there are; and (4) other autoscaling configurations that exist in HPA and VPA (e.g., desired resource utilizations, container update policies, min and max number of replicas). MPA API is also responsible for connecting the autoscaling actions generated from the MPA Recommender to MPA Admission Controller and Updater which actually execute the scaling actions. MPA API is created based on the [multidimensional Pod scaling service] (not open-sourced) provided by Google. MPA API is a Custom Resource Definition (CRD) in Kubernetes and each MPA instance is a CR. MPA CR keeps track of recommendations on target requests and target replica numbers.
```
jbartosik, line 124: "(2) application SLAs" seems to be missing from the MPA spec below.

multidimensional-pod-autoscaler/KEP.md
```diff
@@ -0,0 +1,519 @@
...
+  - type: Resource
+    resource:
+    # Define the target CPU utilization request here
+    name: cpu
+    target:
+      type: Utilization
+      averageUtilization: target-cpu-util
+  constraints:
+    minReplicas: min-num-replicas
+    maxReplicas: max-num-replicas
```
sftim, line 206: I'd expect to see the minimum and maximum memory (or other resources) listed here.

multidimensional-pod-autoscaler/KEP.md
```diff
@@ -0,0 +1,519 @@
...
+### MPA API Object
+
+We reuse the CR definitions from the [MultidimPodAutoscaler](https://cloud.google.com/kubernetes-engine/docs/how-to/multidimensional-pod-autoscaling) object developed by Google.
+`MultidimPodAutoscaler` is the configuration for multi-dimensional Pod autoscaling, which automatically manages Pod resources and their count based on historical and real-time resource utilization.
+MultidimPodAutoscaler has two main fields: `spec` and `status`.
+
+#### MPA Spec
+
+```
+# MultidimPodAutoscalerSpec
```
sftim, line 184: This is an object, not a spec

multidimensional-pod-autoscaler/KEP.md
```diff
@@ -0,0 +1,519 @@
...
+
+### Dependencies
+
+<!--
+This section must be completed when targeting beta to a release.
+-->
+
+#### Does this feature depend on any specific services running in the cluster?
+
+MPA relies on cluster-level service metrics-server and the API server needs to support Mutating Webhooks.
```
sftim, line 365:
```suggestion
MPA relies on cluster-level `metrics.k8s.io` API (for example, from [metrics-server](https://github.com/kubernetes-sigs/metrics-server))
For the evict-and-replace mechanism, the API server needs to support the MutatingAdmissionWebhook API.
```

multidimensional-pod-autoscaler/KEP.md
```diff
@@ -0,0 +1,519 @@
+# KEP-5342: Multi-dimensional Pod Autoscaler
```
sftim, line 1: This is a made up KEP ID.

sftim (commented):
Should we make this a KEP against https://github.com/kubernetes/enhancements/?

If we want to assign a real KEP number, that's how we do it: open an issue in the enhancements repo.

There are a bunch of details to consider; this will interact with in-place Pod resizing. I imagine we'd also want to document the new feature and generally track it as part of the Kubernetes project.

multidimensional-pod-autoscaler/KEP.md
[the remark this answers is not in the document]
James-QiuHaoran: Done.

multidimensional-pod-autoscaler/KEP.md
```diff
@@ -0,0 +1,519 @@
...
+
+**MPA Recommender.** MPA Recommender retrieves the time-indexed measurement data from the Metrics APIs and generates the vertical and horizontal scaling actions. The actions from the MPA Recommender are then updated in the MPA API object. The autoscaling behavior is based on user-defined configurations. Users can implement their own recommenders as well.
+
+**MPA Updater.** MPA Updater will update the number of replicas in the deployment and evict the eligible pods for vertical scaling.
+
+**MPA Admission-Controller.** If users intend to directly execute the autoscaling recommendations generated from the MPA Recommender, the MPA Admission-Controller will update the deployment configuration (i.e., the size of each replica) and configure the rolling update to the Application Deployment.
+
+### Action Actuation Implementation
+
+To actuate the decisions without losing availability, we plan to (1) evict pods with min-replicas configured and update Pod sizes with the web-hooked admission controller (for vertical scaling), and (2) add or remove replicas (for horizontal scaling).
```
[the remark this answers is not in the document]
James-QiuHaoran: Done.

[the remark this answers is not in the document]
James-QiuHaoran: It is shown on the arrow from the admission controller to the deployment where it says resizing pods in the left figure. Let me know if it's not concise or needs clarification here.

multidimensional-pod-autoscaler/KEP.md
```diff
@@ -0,0 +1,519 @@
...
+We use a web-hooked admission controller to manage vertical scaling because if the actuator directly updates the vertical scaling configurations through deployment, it will potentially overload etcd (as vertical scaling might be quite frequent).
+MPA Admission Controller intercepts Pod creation requests and rewrites the request by applying recommended resources to the Pod spec.
+We do not use the web-hooked admission controller to manage the horizontal scaling as it could slow down the pod creation process.
+In the future when the [in-place vertical resizing](https://github.com/kubernetes/enhancements/issues/1287) is enabled, we can remove the web-hooked admission controller and only have the updater.
```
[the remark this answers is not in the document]
James-QiuHaoran: That's a good point! I changed it to that we will keep the original webhook-based update as an option in addition to the new in-place update.

multidimensional-pod-autoscaler/KEP.md
```diff
@@ -0,0 +1,519 @@
...
+Cons:
+- Webhooks introduce extra overhead for vertical scaling operations (can be avoided after in-place resizing of pod is enabled without eviction)
+- Vertical and horizontal scaling executions are separated (can be avoided after in-place resizing of pod is enabled without eviction)
+- State changes in pod sizes are not persisted (too much to keep in etcd, could use Prometheus to store pod state changes)
+
+### Action Recommendation Implementation
+
+To generate the vertical scaling action recommendation, we reuse VPA libraries as much as possible to implement scaling algorithm integrated with the newly generated MPA API code.
+To do that, we need to update accordingly the code which reads and updates the VPA objects to be interacting with the MPA objects.
+To generate the horizontal scaling action recommendation, we reuse HPA libraries, integrating with the MPA API code, to reads and updates the MPA objects.
```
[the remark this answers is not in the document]
James-QiuHaoran: Corrected.

multidimensional-pod-autoscaler/KEP.md
James-QiuHaoran: Added to the spec.

multidimensional-pod-autoscaler/KEP.md
James-QiuHaoran: The min/max memory is specified under the resource policy field, following the convention defined in the verticalpodautoscaler spec. Let me know if that is confusing.

multidimensional-pod-autoscaler/KEP.md
James-QiuHaoran: Corrected

James-QiuHaoran:
> I left a few small comments.
> But I have one bigger question. I'm not sure what will be implemented as result of this KEP.
> Refactoring VPA and HPA so that they can be used a libraries?
> MPA API?
> MPA recommender / updater / admission controller?

@jbartosik Thanks for the question! The scope of this KEP is to deliver a functional MPA by implementing MPA API and three controllers reusing existing VPA/HPA packages without refactoring them (as mentioned in the goals and non-goals).

multidimensional-pod-autoscaler/KEP.md
wangchen615: The `in-place pod vertical scaling` feature can be picked up by MPA when the [issue#4016](https://github.com/kubernetes/autoscaler/issues/4016) is solved in VPA.

multidimensional-pod-autoscaler/KEP.md
wangchen615: There is no need to change code k/k for now.

multidimensional-pod-autoscaler/KEP.md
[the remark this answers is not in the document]
wangchen615: The plan is to wait for VPA to adopt the in-place update first, and then we will update the VPA libraries in MPA to adopt it in MPA as well. Namely, it will support in-place update when the [issue#4016](https://github.com/kubernetes/autoscaler/issues/4016) is resolved.

multidimensional-pod-autoscaler/KEP.md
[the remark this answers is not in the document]
wangchen615: > HPA packages without refactoring them (as mention



> I left a few small comments.
> 
> But I have one bigger question. I'm not sure what will be implemented as result of this KEP.
> 
> * Refactoring VPA and HPA so that they can be used a libraries?
> * MPA API?
> * MPA recommender / updater / admission controller?
 
There will be an MPA API, MPA recommender/updater/admission controller. However, inside each controller, it will reuse VPA and HPA libraries as much as possible.

multidimensional-pod-autoscaler/KEP.md
wangchen615: For this question, we are unsure if Kubernetes is the right target. We would like to start with the autoscaler as the code base mostly reuses existing VPA and HPA.

multidimensional-pod-autoscaler/AEP.md
```diff
@@ -0,0 +1,534 @@
...
+
+#### `MultidimPodAutoscalerSpec`
+
+```
+# MultidimPodAutoscalerSpec
+apiVersion: autoscaling.gke.io/v1beta1
+kind: MultidimPodAutoscaler
+metadata:
+  name: my-autoscaler
+spec:
```
jbartosik, line 201: [MultidimPodAutoscalerSpec](https://cloud.google.com/kubernetes-engine/docs/how-to/multidimensional-pod-autoscaling#multidimpodautoscalerspec) has somewhat different fields than what you describe here (for example here `updatePolicy` is directly in `MultidimPodAutoscalerSpec` , the linked doc its in Policy. I'd rather keep differences minimal

multidimensional-pod-autoscaler/AEP.md
```diff
@@ -0,0 +1,534 @@
...
+
+### MPA API Object
+
+We reuse the CR definitions from the [MultidimPodAutoscaler](https://cloud.google.com/kubernetes-engine/docs/how-to/multidimensional-pod-autoscaling) object developed by Google.
+`MultidimPodAutoscaler` is the configuration for multi-dimensional Pod autoscaling, which automatically manages Pod resources and their count based on historical and real-time resource utilization.
+MultidimPodAutoscaler has two main fields: `spec` and `status`.
+
+#### `MultidimPodAutoscalerSpec`
+
+```
```
jbartosik, line 195: Please provide this in the same format as `MultidimPodAutoscalerStatus` below (for consistency and to make API easier to understand).

multidimensional-pod-autoscaler/AEP.md
```diff
@@ -0,0 +1,534 @@
...
+    resource:
+    # Define the target CPU utilization request here
+    name: cpu
+    target:
+      type: Utilization
+      averageUtilization: target-cpu-util
+  constraints:
+    minReplicas: min-num-replicas
+    maxReplicas: max-num-replicas
+    applicationSLA: value  # customizable SLA for application metrics such as latency or throughput
```
jbartosik, line 219: I don't understand what does SLA specification look like.

multidimensional-pod-autoscaler/AEP.md
```diff
@@ -0,0 +1,534 @@
...
+This question should be filled when targeting a release.
+For Alpha, describe what tests will be added to ensure proper quality of the enhancement.
+
+For Beta and GA, add links to added tests together with links to k8s-triage for those tests:
+https://storage.googleapis.com/k8s-triage/index.html
+-->
+
+<!-- - <test>: <link to test coverage> -->
+
+Integration tests are to be added in the beta version.
```
jbartosik, line 304: I think we should have some basic tests to make sure the feature works

James-QiuHaoran:
> > > I left a few small comments.
> > > But I have one bigger question. I'm not sure what will be implemented as result of this KEP.
> > > Refactoring VPA and HPA so that they can be used a libraries?
> > > MPA API?
> > > MPA recommender / updater / admission controller?
> > 
> > 
> > Joachim Thanks for the question! The scope of this KEP is to deliver a functional MPA by implementing MPA API and three controllers reusing existing VPA/HPA packages without refactoring them (as mentioned in the goals and non-goals).
> 
> Please briefly explain that's the plan for calculating recommendations. I think we should at least briefly describe how we're going to calculate recommendations (HPA based on CPU / configured metrics + VPA for memory?)

Added brief explanation. As an initial solution, VPA and HPA are performed alternatively in each iteration: VPA adjusts the CPU/memory allocation based on history usage; HPA then adjusts the number of replicas based on either CPU utilization or a custom metric. In the future, we can consider more complex prioritization and conflict resolution algorithms.

multidimensional-pod-autoscaler/AEP.md
```diff
@@ -0,0 +1,534 @@
...
+### MPA API Object
+
+We reuse the CR definitions from the [MultidimPodAutoscaler](https://cloud.google.com/kubernetes-engine/docs/how-to/multidimensional-pod-autoscaling) object developed by Google.
+`MultidimPodAutoscaler` is the configuration for multi-dimensional Pod autoscaling, which automatically manages Pod resources and their count based on historical and real-time resource utilization.
+MultidimPodAutoscaler has two main fields: `spec` and `status`.
+
+#### `MultidimPodAutoscalerSpec`
+
+```
+# MultidimPodAutoscalerSpec
```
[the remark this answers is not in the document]
James-QiuHaoran: Yes, let me know if there's anything to clarify here.

multidimensional-pod-autoscaler/AEP.md
James-QiuHaoran: It's currently following the [VerticalPodAutoscalerSpec](https://github.com/kubernetes/autoscaler/blob/bd51490533e5fdbe5a7af897020025057f9c291a/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L73) style with `updatePolicy` directly in `VerticalPodAutoscalerSpec`. Let me know if it's necessary to have it in a wrapper `Policy`.

multidimensional-pod-autoscaler/AEP.md
James-QiuHaoran: Done.

multidimensional-pod-autoscaler/AEP.md
James-QiuHaoran: E.g., if the SLA used is latency, then it could be something like 100ms. And the replica-set can be horizontally scaled when the measured latency is greater than 100ms, i.e., the SLA value.

pbetkier: It sounds to me `applicationSLA` is only useful for a particular recommender that is SLA-based. Since we're proposing MPA API supporting various recommenders I believe the `applicationSLA` belongs as config options of a concrete recommender rather than common API.

multidimensional-pod-autoscaler/KEP.md
pbetkier: I'm not sure here. I believe it was possible to import HPA code and reuse it with little modification. But HPA code in k/k doesn't seem to have a well defined API to be used as library at the moment. Let's think about the right API for code reuse in HPA so that we don't end up with having to fix MPA whenever current method signatures change in HPA. I guess we would need concrete PR to see if the current API is fine or what is missing.

Interesting side-note: how will we know when making code changes to k/k HPA that we don't break MPA by mistake?

multidimensional-pod-autoscaler/AEP.md
pbetkier: Is the default recommender implemented as part of this AEP going to use the `applicationSLA` field?

James-QiuHaoran: The default recommender implemented in this AEP will not use the `applicationSLA` field and this field will be used by users who want to implement their own recommender. E.g., an RL/ML-based recommender can have `applicationSLA` as part of the reward/loss function and thus they can have extra constraints in addition to min/max replicas.

The `applicationSLA` field is part of `constraints` for simplicity.

multidimensional-pod-autoscaler/KEP.md
James-QiuHaoran: That's a good point. The current plan is to have a maintainer watch the imported HPA library for incorporating new changes that will necessarily affect MPA.

multidimensional-pod-autoscaler/AEP.md
jbartosik: I don't understand what is type of this field. Is it a string? If so what does the string represent? Does it point to a metric? Or is it a struct? If it's a struct I think we should describe what is in the struct. Or maybe it's something else?

jbartosik (commented):
@mwielgus this mostly looks good to me.

I have one open question that I think is worth answering before merging. [Here](https://github.com/kubernetes/autoscaler/pull/5342#discussion_r1149768931) I'm asking for clarification on what is format of the `applicationSLA` field (not used by default recommender, additional input for reward in ML recommender).

This is an alpha API so as I understand it won't be too much pain to change it if we don't like the format but I think it'd be nice to have it described.

Other than that looks good to me

multidimensional-pod-autoscaler/AEP.md
James-QiuHaoran: It was initially designed simply to be a floating point number (most application metrics like latency and throughput are floating point numbers). Let's say an application has a latency SLA of 100ms, but the actual latency is 120.8ms; then the recommender can suggest a scale-out option. I think it should cover most use cases but I'd like to learn from you about more potential production use cases.

multidimensional-pod-autoscaler/AEP.md
```diff
@@ -0,0 +1,541 @@
...
+    - name: '*' # either a literal name, or "*" to match all containers
+                # this is not a general wildcard match
+    # Define boundaries for the memory request here
+      requests:
+        minAllowed:
+          memory: min-allowed-memory
+        maxAllowed:
+          memory: max-allowed-memory
+  # Define the recommender to use here
+  recommenders:
```
pbetkier, line 233: This is now a list of recommenders, not a single recommender to use as the comment suggests. How does it work to have multiple implementations in use at the same time?

multidimensional-pod-autoscaler/AEP.md
pbetkier: > The applicationSLA field is part of constraints for simplicity.

Well, it’s going to be simple for users of the particular proprietary recommender that supports this field, rather confusing for other users that need to ignore it ;) Worse if someone else comes up with their slo-based autoscaling that requires slightly different set (or format) of constraints. I still think this field belongs to an ‘options’ map under ‘recommender’ field.

pbetkier: Also, to be nit picky, I think this is more of an SLO than SLA.

James-QiuHaoran: Replaced SLO with SLA. I'll leave the field as optional under 'constraints' and see if there are more use cases to support moving it under the 'recommender' field. Thanks!

multidimensional-pod-autoscaler/AEP.md
James-QiuHaoran: So it's actually consistent with the definition in the vertical autoscaler: https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L105

Only the first recommender will be supported initially, but later, when multiples are running, only one recommender’s recommendations are selected, and others run in recommendation-only mode.

wangchen615: @pbetkier, this is consistent with the VPA support on alternative recommenders. The first support of single alternative recommender is already added, but the multiple alternative recommenders' support for A/B testing is still in planning. We may introduce that soon.

multidimensional-pod-autoscaler/AEP.md
```diff
@@ -254,6 +254,11 @@ status:
     value: metric-value
 ```
 
+Note that application SLO field is **optional** and SLO is defined to be the quality of service target that an application must meet (regarding latency, throughput, and so on).
+For example, if the latency SLO is in use, then it could be 99% of the requests finish within 100ms. Accordingly, the replica set can be horizontally scaled when the measured latency is greater than 100ms, i.e., violating the SLO value.
+The default MPA recommender implemented in this AEP will not use the `applicationSLO` field and this field will be used by users who want to implement their own recommender. For example, an RL/ML-based recommender can have `applicationSLO` as part of the reward/loss function and thus they can have extra constraints in addition to min/max replicas.
+The `applicationSLO` field is a floating point number (most application metrics like latency and throughput are floating point numbers).
```
jbartosik, line 260: It seems to me that just the target number is not enough.

Let's use the example from the doc: we're targeting 99% of requests handled in 100 ms or faster.

It seems that we need to provide at least 2 pieces of information:

1) Metric reporting 99th percentile of request handling time and
2) 100ms (in the same units as used by the metric)

The proposed API has only 1 field of type double, so I can provide 2) (target SLO value) but I don't see a way to specify 1) (metric we're targeting).

For beta I think we ill need ability to specify something more complicated than just a metric value (for example percentile of histogram metric, support for incremental metrics). But for alpha I think we need at least ability to specify what metric we're targetting.

Please amend the proposal to allow that or explain how it's possible with the current proposdal

jbartosik (changes requested):
@wangchen615 @James-QiuHaoran than you for answering my question. I left one more comment. 
I don't understand how the API described in the proposal allows configuring SLO target. To me it looks like we're missing a field. Please add an appropriate field or explain what I'm missing

multidimensional-pod-autoscaler/AEP.md
jbartosik: @wangchen615 @James-QiuHaoran 

I still think the proposal doesn't work. Now there are 2 fields:

- `applicationSLO` and
-  `percentageSLO `

This allows us to specify that we want value of 100 or less 99% of the time.

It still doesn't allow us to specify what metric we're targeting. (Which metric should have value 100 or less 99% of the time?)

Maybe it's better to do like @pbetkier suggested and put those parameters in recommender-specific place instead of putting them in the general API?

In addition to what Piotr pointed out (OS recommender is not going to use those fields so it's a bit weird to have them) it looks like this part of the API is not fully ready yet.

jbartosik (changes requested):

James-QiuHaoran:
> It still doesn't allow us to specify what metric we're targeting. (Which metric should have value 100 or less 99% of the time?)
> Maybe it's better to do like @pbetkier suggested and put those parameters in recommender-specific place instead of putting them in the general API?
> In addition to what Piotr pointed out (OS recommender is not going to use those fields so it's a bit weird to have them) it looks like this part of the API is not fully ready yet.

Hi @jbartosik , It makes sense to me to move it under `recommenders` and add a metric name field to suggest the targeted metric. We pushed the changes. Let us know if there are any further questions. Thanks!

pbetkier (commented):
1. **custom options for `recommenders`**: we cannot define optional fields in the open source API for use solely in a custom, non open sourced recommender. With 1 custom recommender it's going to be already confusing for open source users, with potential 5 custom recommenders it's going to become a mess of optional fields where it's not clear which belong to which recommender. I strongly believe the correct way is to either a) define an `options` or `config` field of type `map[string]string` besides `name` in `recommenders` items (the recommender is responsible to parse and validate) or b) have custom CRDs for custom recommenders that require additional config – this way the API can be tailored to every custom recommender's need. Wasn't a) already thought of for [VPA recommenders](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa#implementation-details)?

2. Can't we stick to API stucture that's already in [Google MPA](https://cloud.google.com/kubernetes-engine/docs/how-to/multidimensional-pod-autoscaling)? I believe the changes made in this AEP are rather cosmetic, so perhaps it's going to be clearer for users if we not diverge?

3. Can we explore the alternative `MPA as a Recommender Only` more? E.g. would it be enough for HPA to read the replica count from a custom metric exposed by MPA recommender? This can be configured right now without any changes to HPA code. And VPA already supports custom recommenders. Perhaps we can think how to solve the `hard to coordinate/synchronize when HPA and VPA states kept in different places` and get away without producing a large amount of copied / imported code. The idea of a full MPA with code copied / imported from VPA and HPA is fair enough for `v1alpha1` (no guarantees so it can stop working any time), but we would have to solve the maintenance problem before graduating to beta IMO.

Let's hear what @mwielgus has to say as the approver.

James-QiuHaoran:
@pbetkier, thanks for the comments and suggestions!

> 1. Custom options for recommenders: we cannot define optional fields in the open source API for use solely in a custom, non open sourced recommender. With 1 custom recommender it's going to be already confusing for open source users, with potential 5 custom recommenders it's going to become a mess of optional fields where it's not clear which belong to which recommender. I strongly believe the correct way is to either a) define an options or config field of type map[string]string besides name in recommenders items (the recommender is responsible to parse and validate) or b) have custom CRDs for custom recommenders that require additional config – this way the API can be tailored to every custom recommender's need. Wasn't a) already thought of for [VPA recommenders](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/enhancements/3919-customized-recommender-vpa#implementation-details)?

> 2. This MPA API vs. Google MPA API: Can't we stick to API structure that's already in [Google MPA](https://cloud.google.com/kubernetes-engine/docs/how-to/multidimensional-pod-autoscaling)? I believe the changes made in this AEP are rather cosmetic, so perhaps it's going to be clearer for users if we not diverge?

1. and 2. make sense and it should be feasible to have custom recommenders themselves configure their own SLOs in use. Removed SLO-related field from the spec.

> 3. MPA as recommender only: Can we explore the alternative MPA as a Recommender Only more? E.g. would it be enough for HPA to read the replica count from a custom metric exposed by MPA recommender? This can be configured right now without any changes to HPA code. And VPA already supports custom recommenders. Perhaps we can think how to solve the hard to coordinate/synchronize when HPA and VPA states kept in different places and get away without producing a large amount of copied / imported code. The idea of a full MPA with code copied / imported from VPA and HPA is fair enough for v1alpha1 (no guarantees so it can stop working any time), but we would have to solve the maintenance problem before graduating to beta IMO.

Having MPA as a recommender only indeed reduces the effort of maintenance, but I think it requires more assumptions from the HPA side, e.g., (1) the replica count should be able to be actuated in real-time (2) the original logic of HPA should be paused. It doesn't resolve the conflicts between HPA and VPA if the MPA falls back to simply HPA+VPA. The benefit of the proposed framework is to replace HPA and VPA's independent loops with a coordinated loop of generating horizontal and vertical scaling actions. It would be a better way if both the conflict and the maintenance could be resolved. What do you think?

pbetkier:
@James-QiuHaoran Alright then IMO let's align the API structure to Google MPA structure and we're good to go.

Note that the MPA will be considered an alpha feature, that may break any time, as w discussed in https://github.com/kubernetes/autoscaler/pull/5342#discussion_r1148108124:

> That's a good point. The current plan is to have a maintainer watch the imported HPA library for incorporating new changes that will necessarily affect MPA.

I mean: if we make changes in VPA or HPA code then we may or may not notice in MPA. I suppose kubernetes/autoscaler reviewers/approvers won't be proactively checking every now and then :) I guess that's fine for an alpha feature. If there's a demand for this feature then we need to address this before graduating to beta. If it turns out ~nobody uses the MPA then we can decide to abort at some point.

That's how I see it, I think it's time for approvers @gjtempleton or @mwielgus to approve or request changes.

pbetkier:
@James-QiuHaoran @wangchen615 please align the API structure to Google API as we discussed before and we're good to go. I've talked with @mwielgus and doesn't have more comments.

James-QiuHaoran:
> @James-QiuHaoran @wangchen615 please align the API structure to Google API as we discussed before and we're good to go. I've talked with @mwielgus and doesn't have more comments.

@pbetkier The API structure has already been aligned with Google API. We followed the same naming of each field in the API as well. Thank you for following up!

mwielgus (approved):
/lgtm
/approve

PR #5344, opened 2022-12-02
Break node categorization in scale down planner on timeout.

olagacek:

cluster-autoscaler

#### What type of PR is this?

/kind feature

Stops node categorisation of nodes into needed/unneeded in scale down logic so that autoscaler loop finishes in timely manner (thus avoiding e.g. cluster-autoscaler being killed by a liveness probe)

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/scaledown/planner/planner.go
+++ b/cluster-autoscaler/core/scaledown/planner/planner.go
@@ -285,10 +285,14 @@ func (p *Planner) categorizeNodes(podDestinations map[string]bool, scaleDownCand
 		p.unremovableNodes.Add(n)
 	}
 	p.nodeUtilizationMap = utilizationMap
-	for _, node := range currentlyUnneededNodeNames {
-		// TODO(x13n): break on timeout. Figure out how to handle nodes
-		// identified as unneeded in previous iteration, but now
-		// skipped due to timeout.
+	timer := time.NewTimer(p.context.ScaleDownSimulationTimeout)
+	for i, node := range currentlyUnneededNodeNames {
+		select {
+		case <-timer.C:
+			klog.Warningf("%d out of %d nodes skipped in scale down simulation due to timeout.", len(currentlyUnneededNodeNames)-i, len(currentlyUnneededNodeNames))
+			break
+		default:
+		}
 		removable, unremovable := p.rs.SimulateNodeRemoval(node, podDestinations, p.latestUpdate, pdbs)
 		if unremovable != nil {
 			unremovableCount += 1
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -285,18 +285,21 @@ func (p *Planner) categorizeNodes(podDestinations map[string]bool, scaleDownCand
...
-		removable, unremovable := p.rs.SimulateNodeRemoval(node, podDestinations, p.latestUpdate, pdbs)
-		if unremovable != nil {
-			unremovableCount += 1
-			p.unremovableNodes.AddTimeout(unremovable, unremovableTimeout)
-		}
-		if removable != nil {
-			delete(podDestinations, removable.Node.Name)
-			removableList = append(removableList, *removable)
+		select {
+		case <-timer.C:
```
x13n, line 291: I think it would be useful to log a warning here: `%d out of %d nodes skipped in scale down simulation due to timeout`.

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -285,18 +285,21 @@ func (p *Planner) categorizeNodes(podDestinations map[string]bool, scaleDownCand
...
+			break
+		default:
+			removable, unremovable := p.rs.SimulateNodeRemoval(node, podDestinations, p.latestUpdate, pdbs)
```
x13n, line 294: nit: Can we avoid increasing the indent too much? I think the function would be slightly more concise if you e.g. closed the `select` block right after `default:`.

cluster-autoscaler/main.go
```diff
@@ -211,6 +211,7 @@ var (
 	skipNodesWithLocalStorage          = flag.Bool("skip-nodes-with-local-storage", true, "If true cluster autoscaler will never delete nodes with pods with local storage, e.g. EmptyDir or HostPath")
 	minReplicaCount                    = flag.Int("min-replica-count", 0, "Minimum number or replicas that a replica set or replication controller should have to allow their pods deletion in scale down")
 	nodeDeleteDelayAfterTaint          = flag.Duration("node-delete-delay-after-taint", 5*time.Second, "How long to wait before deleting a node after tainting it")
+	scaleDownNodeCategorizationTimeout = flag.Duration("scale-down-node0categorization-timeout", 5*time.Minute, "How long should we categorize nodes as needed/unneeded in scale down logic.")
```
x13n, line 214: nit: wouldn't "scale down simulation" be slightly more understandable than "scale down node categorization" at the flag level?

Also, there's a typo in the flag name (s/0/-/).

x13n (changes requested):

cluster-autoscaler/core/scaledown/planner/planner.go
olagacek: added a warning

cluster-autoscaler/main.go
olagacek: renamed

cluster-autoscaler/core/scaledown/planner/planner.go
olagacek: refactored.

PR #5198, opened 2022-09-19
Introduce a formal policy for maintaining cloudproviders

MaciekPytel:
This is an initial proposal for a policy on cloudprovider maintenance discussed in a recent sig-meeting.

The policy largely codifies what we've already been doing for years (including the requirements we've already imposed on new providers). It also introduces a new 'cloudprovider maintenance request' mechanism, following the general idea previously discussed in the sig-meeting.

#### Which component this PR applies to?

cluster-autoscaler

#### What type of PR is this?

/kind documentation

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

This policy is only relevant for CA / cloudprovider developers and not its users.

PR #5350, opened 2022-12-05
Introduce Cloudprovider Maintenance Request policy

MaciekPytel:

cluster-autoscaler 

#### What type of PR is this?

This is a follow-up to https://github.com/kubernetes/autoscaler/pull/5198, separating new policy proposal into a separate PR.


x13n:
/hold

This is a follow-up, so we don't want to accidentally merge this before the previous one.

mwielgus (approved):
/lgtm
/approve

PR #5351, opened 2022-12-05
feat(helm): add rancher cloud config support

24601:
Autoscaler 1.25.0 adds "rancher" cloud provider support, it requires setting cloudConfigPath. If the user mounts this as a secret and sets this value appropriately, this change sets the argument required to point to the mounted secret. Previously, this was only set if cloud provider was magnum or aws.

cluster-autoscaler (helm chart)

#### What type of PR is this?

/kind feature

adds support for the rancher cloud provider added in 1.25.0 to pass its config file

Fixes #

Not directly, just supports using an existing Helm value in the right place

```docs

```

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -11,4 +11,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.21.0
+version: 9.21.1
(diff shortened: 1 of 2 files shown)


gjtempleton:
/assign @gjtempleton 

Hey @24601, thanks for the PR. This is currently failing checks as you need to bump the [chart version](https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/Chart.yaml#L14).

24601:
> /assign @gjtempleton
> 
> Hey @24601, thanks for the PR. This is currently failing checks as you need to bump the [chart version](https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/Chart.yaml#L14).

DOH! That would be useful, wouldn't it? Done. 

Thank you, sir!

mwielgus (approved):
/lgtm
/approve

PR #5354, opened 2022-12-06
Add pods from node undergoing scale-down to unschedulable pods

BigDarkClown:

cluster-autoscaler

#### What type of PR is this?

/kind feature

This PR adds pods from nodes undergoing scale-down to unschedulable pods. Thanks to that we can quickly detect if these pods can schedule on exisiting nodes in the cluster and trigger scale-up as needed.

This PR additionally removed similar logic from the scale-down to avoid pod duplication.

```release-note
NONE
```

### DIFF
--- a/cluster-autoscaler/core/podlistprocessor/currently_drained_nodes_test.go
+++ b/cluster-autoscaler/core/podlistprocessor/currently_drained_nodes_test.go
@@ -15,3 +15,288 @@ limitations under the License.
 */
 
 package podlistprocessor
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/autoscaler/cluster-autoscaler/context"
+	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown"
+	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown/status"
+	"k8s.io/autoscaler/cluster-autoscaler/simulator/clustersnapshot"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
+	. "k8s.io/autoscaler/cluster-autoscaler/utils/test"
+)
+
+func TestCurrentlyDrainedNodesPodListProcessor(t *testing.T) {
+	testCases := []struct {
+		name         string
+		drainedNodes []string
+		nodes        []*apiv1.Node
+		pods         []*apiv1.Pod
+
+		unschedulablePods []*apiv1.Pod
+		wantPods          []*apiv1.Pod
+	}{
+		{
+			name: "no nodes, no unschedulable pods",
+		},
+		{
+			name: "no nodes, some unschedulable pods",
+			unschedulablePods: []*apiv1.Pod{
+				BuildTestPod("p1", 100, 1),
+				BuildTestPod("p2", 200, 1),
+			},
+			wantPods: []*apiv1.Pod{
+				BuildTestPod("p1", 100, 1),
+				BuildTestPod("p2", 200, 1),
+			},
+		},
+		{
+			name:         "single node undergoing deletion",
+			drainedNodes: []string{"n"},
+			nodes: []*apiv1.Node{
+				BuildTestNode("n", 1000, 10),
+			},
+			pods: []*apiv1.Pod{
+				BuildScheduledTestPod("p1", 100, 1, "n"),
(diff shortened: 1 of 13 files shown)


cluster-autoscaler/core/podlistprocessor/currently_drained_nodes_test.go
```diff
@@ -0,0 +1,308 @@
...
+			}
+			clustersnapshot.InitializeClusterSnapshotOrDie(t, ctx.ClusterSnapshot, tc.nodes, tc.pods)
+
+			processor := NewCurrentlyDrainedNodesPodListProcessor()
+			pods, err := processor.Process(&ctx, tc.unschedulablePods)
+			var n1 []string
+			for _, p := range pods {
+				n1 = append(n1, p.Name)
+			}
+			fmt.Printf("%v\n", n1)
```
x13n, line 267: Leftover from debugging?

cluster-autoscaler/core/podlistprocessor/pod_list_processor.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+type defaultPodListProcessor struct {
+	currentlyDrainedNodes *currentlyDrainedNodesPodListProcessor
+	filterOutSchedulable  *filterOutSchedulablePodListProcessor
+}
+
+// NewDefaultPodListProcessor returns a default implementation of the pod list
+// processor, which wraps and sequentially runs other sub-processors.
+func NewDefaultPodListProcessor(currentlyDrainedNodes *currentlyDrainedNodesPodListProcessor, filterOutSchedulable *filterOutSchedulablePodListProcessor) *defaultPodListProcessor {
```
x13n, line 31: I'd prefer having a generic pod list processor aggregating other pod list processors, but I remember having a similar discussion in the past with opinions on both sides and I'm not sure if we agreed on anything back then. I guess either one is fine then.

cluster-autoscaler/core/utils/utils.go
```diff
@@ -212,3 +213,26 @@ func GetOldestCreateTimeWithGpu(pods []*apiv1.Pod) (bool, time.Time) {
 	}
 	return gpuFound, oldest
 }
+
+// FilterRecreatablePods filters pods that will be recreated by their controllers
+func FilterRecreatablePods(pods []*apiv1.Pod) []*apiv1.Pod {
```
x13n, line 218: Would it be possible to move these functions to a place with a name that is less generic than "utils"?

cluster-autoscaler/utils/test/test_utils.go
```diff
@@ -103,6 +104,37 @@ func BuildTestPodWithEphemeralStorage(name string, cpu, mem, ephemeralStorage in
...
+
+// SetMirrorPodSpec sets pod spec to make it a mirror pod
+func SetMirrorPodSpec(pod *apiv1.Pod) *apiv1.Pod {
+	pod.ObjectMeta.Annotations[kube_types.ConfigMirrorAnnotationKey] = "mirror"
+	return pod
+}
+
+// SetDSPodSpec sets pod spec to make it a DS pod
+func SetDSPodSpec(pod *apiv1.Pod) *apiv1.Pod {
+	pod.OwnerReferences = GenerateOwnerReferences("ds", "DaemonSet", "extensions/v1beta1", "api/v1/namespaces/default/daemonsets/ds")
```
x13n, line 128: nit: DaemonSet is in apps/v1, not extensions/v1beta1 for quite a while now. This shouldn't (?) affect the tests, but maybe worth using the right API group?

x13n (changes requested):

cluster-autoscaler/core/podlistprocessor/currently_drained_nodes_test.go
BigDarkClown: Done

cluster-autoscaler/utils/test/test_utils.go
BigDarkClown: Done

cluster-autoscaler/core/podlistprocessor/pod_list_processor.go
x13n: I'm not convinced by this - the concept of order is definitely there in a generic aggregator, I don't see why determining it during initialization would be less safe. But then again, not going to block on this. :)

mwielgus (approved):
/lgtm
/approve

PR #5358, opened 2022-12-07
Export execution_latency_seconds metric from VPA admission controller

jbartosik:
Sometimes I see admissions that are slower than the rest. Logs indicate that `AdmissionServer.admit` doesn't get slow (it's only part with logging). I'd like to have a metric which will tell us what's slow so that we can maybe improve that.

I considered adding a label to existing metric but I think this approach is better:

- existing metric stays the same
- we get metric that's the same for all 3 VPA comonents

VPA Admission Controller

```release-note
Export execution_latency_seconds metric from VPA admission controller
```


jbartosik:
@kgolab @mwielgus PTAL

PR #5375, opened 2022-12-19
Remove nodegroupset processor from clusterapi

### LINKED ISSUE #5374: Cluster Autoscaler Cluster API provider should not carry inherent balancing ignore labels
The Cluster API provider for Cluster Autoscaler can be deployed on many different infrastructure providers as defined by the Cluster API interface. Due to this abundance of deployment options, we would like to simplify the nodegroupset processor for Cluster API by removing the inherently compiled ignore labels. This decision has been agreed to by the community in [this deep dive meeting](https://www.youtube.com/watch?v=jbhca_9oPuQ).

We believe that instructing Cluster API users on the proper labels to ignore is a better path forward than carrying this technical debt inside the autoscaler. We are concerned about keeping the list updated and also the possibility of multiple infrastructure providers having conflicting labels. By removing all labels and focusing on the user utilizing the `--balancing-ignore-label` flag will make this feature easier to maintain in the future.

elmiko:

cluster-autoscaler

#### What type of PR is this?

/kind cleanup

As agreed by the Cluster API community ([see this recording for more context](https://www.youtube.com/watch?v=jbhca_9oPuQ)), we are removing the inherently ignored labels from the nodegroupset processor for clusterapi. This is being done because the Cluster API provider can be deployed to many different infrastructure providers, and as such we would prefer to instruct our users on the usage of balancing ignore labels instead of carrying them as part of the autoscaler.

Fixes #5374 

```release-note
Inherently ignored labels for balancing similar node groups with the Cluster API provider has been removed, please use the `--balancing-ignore-label` flag to add any labels which must be ignored.
```

```docs
- [Usage]: See README.md file in the Cluster API provider directory of the Cluster Autoscaler.
```


cluster-autoscaler/cloudprovider/clusterapi/README.md
```diff
@@ -336,3 +360,52 @@ CAPI cloudprovider, the label format is as follows:
...
+Because Cluster API can address many different cloud providers, it is important
+to configure the balancing labels to ignore provider-specific labels which
+are used for carrying zonal information on Kubernetes nodes. The Cluster
+Autoscaler implementation for Cluster API does not assume any labels to be
+ignored when running. Users must configure their Cluster Autoscaler deployment
+to ignore labels which might be different between nodes, but which do not
+otherwise affect node behavior or size (for example when two MachineDeployments
+are the same except for their deployment zones).
+
+The following table shows some of the most common labels used by cloud providers
```
enxebre, line 391: I'm not sure how it hurts to have the most common known labels skipped by default programatically?
Can we clarify why dropping this is a good thing?

elmiko: sure thing, i'll try to add some of the reasoning we discussed in the meeting.

enxebre:
Thanks @elmiko! FWIW one of the main values of CAPI is that is could agnostic for consumers, so moving cloud provider detail up to the autoscaler is unfortunate in that sense. Let's keep that in mind as we keep moving design forward.

elmiko:
> moving cloud provider detail up to the autoscaler is unfortunate in that sense. Let's keep that in mind as we keep moving design forward.

thanks @enxebre , and yes i agree

elmiko:
@arunmk , @enxebre , or @x13n , might i trouble for another review?

we have had enough time for reviews from the capi community, i am removing the hold.

/hold cancel

PR #5380, opened 2022-12-22
Cap logs logged by HintingSimulator.

kisieland:
To avoid excesive logging of the HintingSimulator we should use klogx to cap the number of logged lines depending on the verbosity level.

Cluster Autoscaler

/kind bug

Uses klogx to cap the number of log lines logged by the Hinting Simulator.
This avoids excessive logging.

```release-note
Cap number of log lines by the HintingSimulator.
```

### DIFF
--- a/cluster-autoscaler/simulator/scheduling/hinting_simulator.go
+++ b/cluster-autoscaler/simulator/scheduling/hinting_simulator.go
@@ -21,9 +21,9 @@ import (
 
 	"k8s.io/autoscaler/cluster-autoscaler/simulator/clustersnapshot"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator/predicatechecker"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/klogx"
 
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/simulator/scheduling/hinting_simulator.go
```diff
@@ -83,6 +84,7 @@ func (s *HintingSimulator) TrySchedulePods(clusterSnapshot clustersnapshot.Clust
 			break
 		}
 	}
+	klogx.V(4).Over(loggingQuota).Infof("There were also %v other logs that were capped.", -loggingQuota.Left())
```
towca, line 87: nit: Maybe clarify what kind of logs, or include some other more precise info (e.g. prefix with the name of the function). Log lines from other goroutines could get inserted in between and then it'd be hard to know what this line refers to.

towca:
/assign @towca 
/lgtm
/approve
/hold

Feel free to unhold if you disagree with the nit.

kisieland:
@towca  makes sense added clarification. PTAL

PR #5384, opened 2022-12-23
Set MaxNodesTotal to maximum number of nodes in all node groups

yasinlachiny:
Signed-off-by: yasinlachiny <yourlachiny@gmail.com>

cluster-autoscaler
#### What type of PR is this?

bug
static_autoscaler doesn't show 

>  "Max total nodes in cluster reached"

because this [this condition](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/static_autoscaler.go#L512) check  `a.MaxNodesTotal > 0` and it's zero by default we also have this condition in other place [like](https://github.com/kubernetes/autoscaler/blob/01389b01e66ae2b3ccedc6781ac2411e3e934cea/cluster-autoscaler/core/scale_up.go#L150). By this PR we can show the proper log when the cluster is full when we haven't set `MaxNodesTotal` 

Fixes #

No
```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/context/autoscaling_context.go
+++ b/cluster-autoscaler/context/autoscaling_context.go
@@ -57,6 +57,8 @@ type AutoscalingContext struct {
 	DebuggingSnapshotter debuggingsnapshot.DebuggingSnapshotter
 	// ScaleDownActuator is the interface for draining and deleting nodes
 	ScaleDownActuator scaledown.Actuator
+	// MaxNodes the maximum number of nodes in the whole cluster when MaxNodesTotal is zero
+	MaxNodes int
 }
 
 // AutoscalingKubeClients contains all Kubernetes API clients,
--- a/cluster-autoscaler/core/scale_up.go
+++ b/cluster-autoscaler/core/scale_up.go
@@ -148,10 +148,10 @@ func isNodeGroupResourceExceeded(ctx *context.AutoscalingContext, resourceManage
 }
 
 func getCappedNewNodeCount(context *context.AutoscalingContext, newNodeCount, currentNodeCount int) (int, errors.AutoscalerError) {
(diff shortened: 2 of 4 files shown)


towca:
I don't think overwriting `AutoscalingOptions.MaxNodesTotal` (or `AutoscalingOptions` fields in general) is the way to go, since it's just a thin wrapper over a user-configurable flag. Consider what happens in this scenario:
1. `--max-nodes-total` is set to 0, so the user expects CA not to enforce a cluster-wide node limit, only per-node-group limits. There are 2 node groups with max node limit initially set to 3.
2. After your change, `AutoscalingOptions.MaxNodesTotal` is overwritten to the sum of node group limits - 6.
3. The max limits for the existing node groups are dynamically changed on the cloud provider site, let's say to 10. CA should now be able to scale to 20 nodes, since there was no cluster-wide limit set, and there are 2 node groups with max limit set to 10.
4.  However, `AutoscalingOptions.MaxNodesTotal` is now larger than 0, so it's not overwritten again, just enforced as if it was configured by the user. Which means that CA can only scale to 6 nodes, despite the user's expectations. This will persist until CA is restarted.

The `Max total nodes in cluster reached` log line is really only an indication of reaching the cluster-wide limit, and IMO it's useful to have that distinguished from when per-node-group limits are reached for all node groups. In the latter scenario, `Skipping node group %s - max size reached` should be logged for all node groups. Is that information not enough for your use case?

yasinlachiny:
Hi @towca 
Sorry for responding late. I got your point and changed the code. 
Yeah, you are right and we can find out about max node reached in a node group, but I think having the actual log means `Max total nodes in cluster reached` is more clear. 

Could you please share your opinion about the new code?

PR #5395, opened 2023-01-03
Balancer API

mwielgus:
#### What type of PR is this?

/kind feature

API for a new component - see https://github.com/kubernetes/autoscaler/blob/master/balancer/proposals/balancer.md

```release-note
None
```

SIG-Autoscaling KEP: https://github.com/kubernetes/autoscaler/blob/master/balancer/proposals/balancer.md


balancer/pkg/apis/balancer.x-k8s.io/v1alpha1/types.go
```diff
@@ -0,0 +1,210 @@
...
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	MinReplicas *int32 `json:"minReplicas,omitempty" protobuf:"varint,3,opt,name=minReplicas"`
+
+	// MaxReplicas is the maximum number of replicas inside of this target.
+	// Balancer will set at most this amount on the target, even if the total
+	// desired number of replicas for the Balancer is higher.
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	MaxReplicas *int32 `json:"maxReplicas,omitempty" protobuf:"varint,4,opt,name=maxReplicas"`
```
MaciekPytel, line 113: Should max be optional? What is the semantics of not specifying max? I'm guessing no limitation on size? If so should this be documented in the docstring?

MaciekPytel:
Left a single comment (and the golint also has one :smile:). Otherwise lgtm.

@kgolab Do you also want to review given your role of co-owner?

balancer/pkg/apis/balancer.x-k8s.io/v1alpha1/types.go
mwielgus: Yes, no limit. Added to docstring.

PR #5402, opened 2023-01-08
adding configurable difference ratios

### LINKED ISSUE #5381: Increase or Allow controlling the value of MaxCapacityMemoryDifferenceRatio  when comparing Node Groups
I'm using cluster-autoscaler's `balance-similar-node-groups` feature (on EKS-based clusters)

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**
I am facing the same issue described [here](https://github.com/kubernetes/autoscaler/pull/3124). 
+ We use AWS Spot Node Groups with many instance types.
+  For a cluster, we have 3 node groups (1 per Availability Zone) and I get an uneven balance of nodes across AZ because some nodes are not considered similar.
    + This is happening because the capacity.memory 's difference is too high. I understand the purpose of the `MaxCapacityMemoryDifferenceRatio` not being too high, but in this case, we want to run spot instances with as many instance types in the mix in order to avoid running out of options (and having the cluster shrinking because the cloud provider reclaims the instances)
It seems that in my case, I end up with nodes of relatively similar capacity (m5d.xlarge and t3.xlarge instance types are both supposed to have 4vCPU and 16Gib), having a `capacity.memory` difference ratio higher than the currently implemented `1.5% MaxCapacityMemoryDifferenceRatio` as specified in [this part of the code](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/processors/nodegroupset/compare_nodegroups.go#L37). 

Example with an m5d.xlarge node vs t3.xlarge node
m5d.xlarge: 
```yaml
allocatable:
    attachable-volumes-aws-ebs: "25"
    cpu: 3920m
    ephemeral-storage: "95551679124"
    hugepages-1Gi: "0"
    hugepages-2Mi: "0"
    memory: 14902308Ki
    pods: "58"
capacity:
    attachable-volumes-aws-ebs: "25"
    cpu: "4"
    ephemeral-storage: 104845292Ki
    hugepages-1Gi: "0"
    hugepages-2Mi: "0"
    memory: 15919140Ki
    pods: "58"
```

t3.xlarge
```yaml
allocatable:
  attachable-volumes-aws-ebs: "25"
  cpu: 3920m
  ephemeral-storage: "95551679124"
  hugepages-1Gi: "0"
  hugepages-2Mi: "0"
  memory: 15186976Ki
  pods: "58"
capacity:
  attachable-volumes-aws-ebs: "25"
  cpu: "4"
  ephemeral-storage: 104845292Ki
  hugepages-1Gi: "0"
  hugepages-2Mi: "0"
  memory: 16203808Ki
  pods: "58"
```

Allow changing this maxDifferenceRatio to fit cloud/user-specific use cases in an cloud-independant way.
I suggest allowing controlling this value through an optional config flag to be specified when starting cluster-autoscaler. For example:
```bash
spec:
      containers:
      - command:
        - ./cluster-autoscaler
        - --v=4
        - ...
        - --balance-similar-node-groups
        - --balance-similar-node-groups-max-capacity-memory-difference-ratio=0.020

```

Optionally, create a new `NodeInfoComparator` which would be trusting the end-user's judgment. 
For instance, we could always consider 2 node groups to be similar if the nodes have a specific label in common.
Example: 
```yaml
labels:
  cluster-autoscaler-similar-node-group-class: foo
```

Bryce-Soghigian:
#### What type of PR is this?
/kind feature

Allows us to adjust difference ratios and configure the balance of nodegroupsets.

Fixes #5381

```release-note
Adding 3 Flags to CA. 
	maxCapacityMemoryDifferenceRatio   = flag.Float64("memory-difference-ratio", 0.015, "Maximum difference in memory capacity between two similar node groups to be considered for balancing. Value is a ratio of the smaller node group's memory capacity.")
	maxFreeDifferenceRatio             = flag.Float64("max-free-difference-ratio", 0.05, "Maximum difference in free resources between two similar node groups to be considered for balancing. Value is a ratio of the smaller node group's free resource.")
	maxAllocatableDifferenceRatio      = flag.Float64("max-allocatable-difference-ratio", 0.05, "Maximum difference in allocatable resources between two similar node groups to be considered for balancing. Value is a ratio of the smaller node group's allocatable resource.")
)

```

### DIFF
--- a/cluster-autoscaler/config/autoscaling_options.go
+++ b/cluster-autoscaler/config/autoscaling_options.go
@@ -46,6 +46,33 @@ type NodeGroupAutoscalingOptions struct {
 	ScaleDownUnreadyTime time.Duration
 }
 
+const (
+	// DefaultMaxAllocatableDifferenceRatio describes how Node.Status.Allocatable can differ between groups in the same NodeGroupSet
+	DefaultMaxAllocatableDifferenceRatio = 0.05
+	// DefaultMaxFreeDifferenceRatio describes how free resources (allocatable - daemon and system pods)
+	DefaultMaxFreeDifferenceRatio = 0.05
+	// DefaultMaxCapacityMemoryDifferenceRatio describes how Node.Status.Capacity.Memory
+	DefaultMaxCapacityMemoryDifferenceRatio = 0.015
+)
+
+// NodeGroupDifferenceRatios contains various ratios used to determine if two NodeGroups are similar and makes scaling decisions
+type NodeGroupDifferenceRatios struct {
+	// MaxAllocatableDifferenceRatio describes how Node.Status.Allocatable can differ between groups in the same NodeGroupSet
+	MaxAllocatableDifferenceRatio float64
+	// MaxFreeDifferenceRatio describes how free resources (allocatable - daemon and system pods) can differ between groups in the same NodeGroupSet
+	MaxFreeDifferenceRatio float64
+	// MaxCapacityMemoryDifferenceRatio describes how Node.Status.Capacity.Memory can differ between groups in the same NodeGroupSetAutoscalingOptions
+	MaxCapacityMemoryDifferenceRatio float64
+}
+
+func NewDefaultNodeGroupDifferenceRatios() NodeGroupDifferenceRatios {
+	return NodeGroupDifferenceRatios{
+		MaxAllocatableDifferenceRatio:    DefaultMaxAllocatableDifferenceRatio,
+		MaxFreeDifferenceRatio:           DefaultMaxFreeDifferenceRatio,
+		MaxCapacityMemoryDifferenceRatio: DefaultMaxCapacityMemoryDifferenceRatio,
+	}
+}
+
 // AutoscalingOptions contain various options to customize how autoscaling works
 type AutoscalingOptions struct {
 	// NodeGroupDefaults are default values for per NodeGroup options.
@@ -213,4 +240,6 @@ type AutoscalingOptions struct {
 	NodeDeleteDelayAfterTaint time.Duration
 	// ParallelDrain is whether CA can drain nodes in parallel.
 	ParallelDrain bool
+	// NodeGroupSetRatio is a collection of ratios used by CA used to make scaling decisions.
+	NodeGroupSetRatios NodeGroupDifferenceRatios
 }
--- a/cluster-autoscaler/config/autoscaling_options.go
+++ b/cluster-autoscaler/config/autoscaling_options.go
@@ -65,6 +65,7 @@ type NodeGroupDifferenceRatios struct {
 	MaxCapacityMemoryDifferenceRatio float64
 }
 
+// NewDefaultNodeGroupDifferenceRatios returns default NodeGroupDifferenceRatios values
 func NewDefaultNodeGroupDifferenceRatios() NodeGroupDifferenceRatios {
 	return NodeGroupDifferenceRatios{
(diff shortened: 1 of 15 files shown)


cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -46,6 +46,52 @@ type NodeGroupAutoscalingOptions struct {
 	ScaleDownUnreadyTime time.Duration
 }
 
+const (
+	// MaxAllocatableDifferenceRatio describes how Node.Status.Allocatable can differ between
+	// groups in the same NodeGroupSet
+	MaxAllocatableDifferenceRatio = 0.05
```
x13n, line 52: Can you rename this to `DefaultMaxAllocatableDifferenceRatio`? Similarly for the two consts below.

cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -46,6 +46,52 @@ type NodeGroupAutoscalingOptions struct {
...
+type NodeGroupDifferenceRatios struct {
+	// MaxAllocatableDifferenceRatio describes how Node.Status.Allocatable can differ between groups in the same NodeGroupSet
+	MaxAllocatableDifferenceRatio float64
+	// MaxFreeDifferenceRatio describes how free resources (allocatable - daemon and system pods) can differ between groups in the same NodeGroupSet
+	MaxFreeDifferenceRatio float64
+	// MaxCapacityMemoryDifferenceRatio describes how Node.Status.Capacity.Memory can differ between groups in the same NodeGroupSetAutoscalingOptions
+	MaxCapacityMemoryDifferenceRatio float64
+}
+
+// GetMaxAllocatableDifferenceRatio returns MaxAllocatableDifferenceRatio if it is set, otherwise returns default value
```
x13n, line 71: Why do the defaulting both here and in flag definitions? This may be confusing, setting any flag explicitly to 0 will not work as expected. I'd suggest dropping the getter functions completely and using the consts above to populate flag defaults. For tests, creating struct populated with defaults would just require a test helper then. WDYT?

cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -46,6 +46,52 @@ type NodeGroupAutoscalingOptions struct {
 	ScaleDownUnreadyTime time.Duration
 }
 
+const (
```
x13n, line 49: Since you're moving the consts here, they should be deleted from the file they used to live in.

x13n (changes requested):

cluster-autoscaler/config/autoscaling_options.go
Bryce-Soghigian: I agree we should only have a single default. I guess I initially thought maybe someone could leverage these types and not have access to the defaults set in the flags, but looking at how its used now I think your plan makes much more sense because of the zero case.

cluster-autoscaler/config/autoscaling_options.go
Bryce-Soghigian: works for me.

Bryce-Soghigian: Modified to be Default... ratio

cluster-autoscaler/config/autoscaling_options.go
Bryce-Soghigian: I agree we should have them in one place. I think they should live near ratio options. Deleted them from the other location.

feiskyer (approved):
LGTM

PR #5415, opened 2023-01-12
patch instead of updating to taint nodes

grosser:
#### What type of PR is this?

/kind bug

Currently tainting a node replaces the whole node object with the cache the CA maintains.
This is a race with other cluster components that do node updates too.
What we observed was that a change by one of our tools that uses `patch` to add an annotation got lost, because of CA overwriting then whole node.

```release-note
Use Patch to add node to taints instead of updating the whole node.
```

TODO:
- get consensus on this being the right approach to build a patch (and wether we should replace or add)
- refactor `addTaintToSpec` to `createTaintPatch` to not have to duplicate it's logic (which fields get modified)

/cc @alam0rt

### DIFF
--- a/cluster-autoscaler/utils/deletetaint/delete.go
+++ b/cluster-autoscaler/utils/deletetaint/delete.go
@@ -18,6 +18,7 @@ package deletetaint
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"time"
@@ -25,6 +26,7 @@ import (
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	kube_client "k8s.io/client-go/kubernetes"
 	kube_record "k8s.io/client-go/tools/record"
 
@@ -37,6 +39,12 @@ const (
 	// DeletionCandidateTaint is a taint used to mark unneeded node as preferably unschedulable.
 	DeletionCandidateTaint = "DeletionCandidateOfClusterAutoscaler"
 )
+
+type patchStruct struct {
+	Op    string      `json:"op"`
+	Path  string      `json:"path"`
+	Value interface{} `json:"value"`
+}
 
 // Mutable only in unit tests
 var (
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/utils/deletetaint/delete.go
```diff
@@ -89,7 +91,20 @@ func addTaint(node *apiv1.Node, client kube_client.Interface, taintKey string, e
...
+		if err != nil {
+			klog.Warningf("Error while adding %v taint on node %v: %v", getKeyShortName(taintKey), node.Name, err)
+			return err
+		}
+		patch := []byte(
+			`[{"op":"replace","path":"/spec/taints","value":` + string(taintsJson) + `},` +
+				` {"op":"replace","path":"/spec/unschedulable","value":` + strconv.FormatBool(freshNode.Spec.Unschedulable) + `}]`,
+		)
+
+		_, err = client.CoreV1().Nodes().Patch(context.TODO(), freshNode.Name, types.JSONPatchType, patch, metav1.PatchOptions{})
```
qianlei90, line 107: Should we replace all `client.CoreV1().Nodes().Update()` to `client.CoreV1().Nodes().Patch()` or just here?

grosser: ideally all ... but I'm trying to keep this PR small to get consensus on if this approach is good
and see if there is maybe a reason I'm overlooking for using `Update`
so ideally merge this once everyone agrees that it's good and then rework the others in the same way

cluster-autoscaler/utils/deletetaint/delete.go
```diff
@@ -89,7 +97,20 @@ func addTaint(node *apiv1.Node, client kube_client.Interface, taintKey string, e
...
+			{Op: "replace", Path: "/spec/unschedulable", Value: freshNode.Spec.Unschedulable},
+		}
+		patchJson, err := json.Marshal(patch)
+		if err != nil {
+			klog.Warningf("Error while adding %v taint on node %v: %v", getKeyShortName(taintKey), node.Name, err)
+			return err
+		}
+
+		_, err = client.CoreV1().Nodes().Patch(context.TODO(), freshNode.Name, types.JSONPatchType, patchJson, metav1.PatchOptions{})
 		if err != nil && errors.IsConflict(err) && time.Now().Before(retryDeadline) {
```
grosser, line 114: I think a conflict can not happen when patching so this code might need a rework too

qianlei90:
We can not prevent other component to update the Node object, and replace `Update()` to `Patch()` may introduce too much complexity.  How about use [`RetryOnConflict`](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/util/retry/util.go#L103) on both CA side and your tool？

grosser:
I'm not sure about the "too much complexity", CA is only updating a limited set of fields right ?

You are suggesting that if both tools do full updates and use the resourceVersion to to create conflicts (and then reload and re-apply) then this situation could also be avoided ?

qianlei90:
> CA is only updating a limited set of fields right ?

@grosser You are right. I searched the keyword `Nodes().Update` and found CA only update Node object in function `addTaint`、`cleanTaint`(some cloud providers also use `Update` but i think it's ok to leave it unchanged). So this solution seems simpler than I thought.

Ignore my comment.

grosser:
played around with this more ... updating is actually safe since it results in conflict,
so this cannot cause the issue we saw (since we patch and CA updates)
also patching the whole taints array could lead to a bug where 2 actors override the taints
so closing until I find more evidence or can reproduce reliably

PR #5418, opened 2023-01-12
clusterapi: track upcoming unprovisioned machines with a temporary providerID to enable detection of exhausted nodegroups

### LINKED ISSUE #5417: clusterapi: Cluster Autoscaler fails to backoff exhausted nodegroups
cluster-autoscaler

Component version:
cluster-autoscaler 1.26

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"24+", GitVersion:"v1.24.0-alpha.3.428+425ff1c8e2d298-dirty", GitCommit:"425ff1c8e2d29874926abb7da37641cf785c4ff3", GitTreeState:"dirty", BuildDate:"2022-03-10T21:18:29Z", GoVersion:"go1.17.7", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.7", GitCommit:"e6f35974b08862a23e7f4aad8e5d7f7f2de26c15", GitTreeState:"clean", BuildDate:"2022-10-12T10:50:21Z", GoVersion:"go1.18.7", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

AWS, clusterAPI provider

When the cluster autoscaler is configured with a priority expander for the clusterapi provider, and multiple MachineDeployments of differing priorities is provided, the cluster autoscaler can become stuck attempting to provision nodes in a higher priority nodegroup whose backing instance type is exhausted in the cloud provider. 

**What happened instead?**:
We have CAPI/CAPA configured in a self-managed cluster with the cluster autoscaler, and the highest priority MachineDeployment instance type became unavailable from AWS. 

The exact behavior we saw was that a new pod was created in the cluster which could not fit on any node, so the cluster autoscaler scaled up the highest priority MachineDeployment. CAPI created a new Machine for the upcoming node, and CAPA created an AWSMachine to fulfill the request. However, AWS was out of capacity of the instance type (m6a.32xlarge) in the AZ and the CAPA controller logged a warning. Because the node does not yet exist in AWS, it lacks a provider ID (the aws ID), and because this is not a failure (in theory the instance type might become available at any time), there is no failure message in the Machine or AWSMachine status fields. 

So the cluster autoscaler believes the upcoming node is available to schedule pods onto, so does not scale up any more nodegroups, but can not track the node as being "unregistered" or "long unregistered" in the cluster state registry, as it lacks a provider id, and nodes without an id can not be mapped back to a node group correctly either. As a result, the cluster autoscaler never backs off the unhealthy MachineDeployment picks another MachineDeployment to attempt to scale up, even when lower priority MachineDeployments exist.

The detail that led me to this finding was noticing this message in the cluster autoscaler logs frequently when this occurred:
```
W0111 18:40:44.125721       1 clusterapi_controller.go:476] Machine "audit-compliant-rad-worker-private-m6a-az3-5f7c8d44c6-j854x" has no providerID
```
When pending nodes with no ID exist, that is logged by the cluster api provider when the cluster state registry tries to understand the node and fails. The pending nodes seem to need a temporary "fake" pending ID to allow the cluster state registry to track the node and then utilize it's existing logic to move the node to a "long unregistered" state if the node fails to provision in time. In fact, there is already a mechanism to create a fake ID for failed Machines, but not pending unprovisioned Machines in the cluster api provider.

Configure the cluster autoscaler with a priority expander and multiple MachineDeployments with different priorities. Create some pods that will deploy onto these nodegroups. When the highest priority MachineDeployment's instance type can not be satisfied, the cluster autoscaler will not advance to the lower priority MachineDeployments for scale up. 

The bug is also visible in the configmap cluster-autoscaler-status which has the nodegroup health and state information. Node groups backed by MachineDeployments will _not_ having pending or provisioning Machines counted in the nodegroups count of nodes, not even as unready or notStarted, instead they only count towards the cloudProviderTarget until the AWSMachine has an AWS ID assigned.

cnmcavoy:
#### What type of PR is this?
/kind bug

Returns a temporary fake provider ID for Machines which have not been fulfilled by the backing cloud provider. Without an ID the nodes confuse the cluster autoscaler, it believes they can fit pods, but it does not account for the node as part of the nodegroup. This change allows the cluster state registry in the cluster autoscaler to correctly track these unprovisioned nodes and detect if they become "long unregistered" and back off the nodegroup.

Fixes #5417

N/A

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
+++ b/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go
@@ -53,6 +53,7 @@ const (
 	resourceNameMachineSet        = "machinesets"
 	resourceNameMachineDeployment = "machinedeployments"
 	failedMachinePrefix           = "failed-machine-"
+	pendingMachinePrefix          = "pending-machine-"
 	machineTemplateKind           = "MachineTemplate"
 	machineDeploymentKind         = "MachineDeployment"
 	machineSetKind                = "MachineSet"
@@ -252,6 +253,9 @@ func (c *machineController) findMachineByProviderID(providerID normalizedProvide
 	if isFailedMachineProviderID(providerID) {
 		return c.findMachine(machineKeyFromFailedProviderID(providerID))
 	}
+	if isPendingMachineProviderID(providerID) {
+		return c.findMachine(machineKeyFromPendingMachineProviderID(providerID))
+	}
 
 	// If the machine object has no providerID--maybe actuator
 	// does not set this value (e.g., OpenStack)--then first
@@ -268,6 +272,15 @@ func (c *machineController) findMachineByProviderID(providerID normalizedProvide
 
 	machineID, _ := node.Annotations[machineAnnotationKey]
 	return c.findMachine(machineID)
+}
+
+func isPendingMachineProviderID(providerID normalizedProviderID) bool {
+	return strings.HasPrefix(string(providerID), pendingMachinePrefix)
+}
+
+func machineKeyFromPendingMachineProviderID(providerID normalizedProviderID) string {
+	namespaceName := strings.TrimPrefix(string(providerID), pendingMachinePrefix)
+	return strings.Replace(namespaceName, "_", "/", 1)
 }
 
 func isFailedMachineProviderID(providerID normalizedProviderID) bool {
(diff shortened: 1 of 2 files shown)


elmiko:
thanks @cnmcavoy , i'll give a review this week

elmiko (commented):
makes sense to me, thanks for the fix!

/approve
/lgtm

Issue #5387, opened 2022-12-27
Don't block scale down on pods owned by custom controllers


x13n:
**Which component are you using?**:

/label cluster-autoscaler

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:

As the k8s ecosystem grows, a hardcoded list currently used by CA is not going to work with custom controllers. In https://github.com/kubernetes/autoscaler/blob/ee667550c081d99924e7b1100f0c1596551be3d6/cluster-autoscaler/utils/drain/drain.go#L76, only a few controllers are handled. We could extend CA to treat all ownerReferences with `controller: true` as replicated. In case of controllers that do not recreate evicted pods, existing annotations could be used (`cluster-autoscaler.kubernetes.io/safe-to-evict`: `false` or `cluster-autoscaler.kubernetes.io/daemonset-pod`: `true`) to prevent CA from evicting such pods.

**Describe the solution you'd like.**:

Just look at `controller` field in owner references and treat all controller-owned pods as replicated by default.

**Describe any alternative solutions you've considered.**:

Add some extension mechanism (CRD?) allowing CA to know how to handle specific controllers. This might be useful in the long term, but will require much more work.

**Additional context.**:

Bryce-Soghigian:
Could i pick this up?

Bryce-Soghigian:
@vadasambar Are you still planning an approach for this or can I pick it up?

vadasambar:
> @vadasambar Are you still planning an approach for this or can I pick it up?

Thank you for noticing I was looking at this issue. I am new to cluster-autoscaler code. I have an understanding of the problem. I am trying to solidify my approach which goes something like this:
1. Add a map of `Kind` to `Lister` in [`ListerRegistry`](https://github.com/kubernetes/autoscaler/blob/c127763a4529d7d07961102856afe0ba8781f8ee/cluster-autoscaler/utils/kubernetes/listers.go#L38-L49) called `GenericListerMap` (<- the name can change)
2. Every time we encounter a pod owner reference that doesn't reference in-built types like `Deployment`. `StatefulSet` etc., (for which we already have Listers), we add a new lister entry to `GenericListerMap`. I want to leverage the [`dynamiclister`](https://pkg.go.dev/k8s.io/client-go@v0.26.0/dynamic/dynamiclister) library here

[...]

PR #5419, opened 2023-01-13
feat: wip check only controller ref to decide if a pod is replicated

vadasambar:
**This PR is still WIP. I will mention the reviewers here once I am done. Although it might not be in the best shape as a WIP PR but if the reviewers have any feedback, I would love to have it. :pray:** 

#### What type of PR is this?

/kind feature

- scale down [is blocked](https://github.com/kubernetes/autoscaler/blob/73d02f1d641a38314e93077d375d80e03468ad60/cluster-autoscaler/utils/drain/drain.go#L210-L211) on the pods where the controller reference is not understood by the CA ([CA as of now only understands controller refs for Daemonsets, ReplicaSets, Jobs, ReplicationControllers, Deployments and StatefulSets](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/utils/drain/drain.go#L124-L203)). This is because CA treats pods with non-understood controller references as non-replicated and hence deems scale-down as not-safe. 
- this PR introduces a change where all pods with ownerReference `controller: true` are treated as replicated and scale-down is not blocked

Fixes https://github.com/kubernetes/autoscaler/issues/5387

```release-note
All pods with controller reference of any sort are treated as replicated. This means cluster-autoscaler would not block scale down on these pods unless `cluster-autoscaler.kubernetes.io/safe-to-evict: false` or `cluster-autoscaler.kubernetes.io/daemonset-pod: true` annotation is set on them. 
```

TBD
```docs

```

### DIFF
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -381,7 +381,8 @@ func buildAutoscaler(debuggingSnapshotter debuggingsnapshot.DebuggingSnapshotter
 	kubeClientConfig.Burst = autoscalingOptions.KubeClientBurst
 	kubeClientConfig.QPS = float32(autoscalingOptions.KubeClientQPS)
 	kubeClient := createKubeClient(kubeClientConfig)
-	dynamicClient := dynamic.NewForConfigOrDie(kubeClientConfig)
+	// re-use kubeClient's REST client
+	dynamicClient := dynamic.New(kubeClient.Discovery().RESTClient())
 
 	eventsKubeClient := createKubeClient(getKubeConfig())
 
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -54,6 +54,7 @@ import (
 	kube_util "k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/units"
 	"k8s.io/autoscaler/cluster-autoscaler/version"
+	"k8s.io/client-go/dynamic"
 	kube_client "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
@@ -380,6 +381,7 @@ func buildAutoscaler(debuggingSnapshotter debuggingsnapshot.DebuggingSnapshotter
 	kubeClientConfig.Burst = autoscalingOptions.KubeClientBurst
 	kubeClientConfig.QPS = float32(autoscalingOptions.KubeClientQPS)
 	kubeClient := createKubeClient(kubeClientConfig)
+	dynamicClient := dynamic.NewForConfigOrDie(kubeClientConfig)
 
 	eventsKubeClient := createKubeClient(getKubeConfig())
 
@@ -395,6 +397,7 @@ func buildAutoscaler(debuggingSnapshotter debuggingsnapshot.DebuggingSnapshotter
 		EventsKubeClient:     eventsKubeClient,
 		DebuggingSnapshotter: debuggingSnapshotter,
 		PredicateChecker:     predicateChecker,
+		DynamicClient:        dynamicClient,
 	}
 
 	opts.Processors = ca_processors.DefaultProcessors()
--- a/cluster-autoscaler/utils/drain/drain.go
+++ b/cluster-autoscaler/utils/drain/drain.go
@@ -133,7 +133,7 @@ func GetPodsForDeletionOnNodeDrain(
 			if _, err := l.Get(controllerRef.Name); err == nil {
 				replicated = true
 			} else {
-				return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: ControllerNotFound}, fmt.Errorf("%s controller for %s/%s is not available, err: %v", strings.ToLower(refKind), pod.Namespace, pod.Name, err)
+				return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: ControllerNotFound}, fmt.Errorf("owner %s for %s/%s is not available, err: %v", refKind, pod.Namespace, pod.Name, err)
 			}
 		}
 
--- a/cluster-autoscaler/utils/drain/drain.go
+++ b/cluster-autoscaler/utils/drain/drain.go
@@ -18,7 +18,6 @@ package drain
 
 import (
 	"fmt"
-	"strings"
 	"time"
 
 	apiv1 "k8s.io/api/core/v1"
@@ -120,7 +119,12 @@ func GetPodsForDeletionOnNodeDrain(
 		}
 
 		if refKind != "" {
-			gv := strings.Split(controllerRef.APIVersion, "/")
+
+			gv, err := schema.ParseGroupVersion(controllerRef.APIVersion)
+			if err != nil {
+				return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: UnexpectedError}, fmt.Errorf("%s/%s pod's owner %s/%s has invalid group version: %v", pod.Namespace, pod.Name, refKind, controllerName, err)
+			}
+
 			// controllerRef doesn't have a namespace by design
 			// The controller/owner is either in the same namespace as the pod
 			// or it's assumed to be a cluster-scoped resource
@@ -128,8 +132,8 @@ func GetPodsForDeletionOnNodeDrain(
 			// a pod resource will always be in the same namespace
 			// TODO: find a better way to handle this
 			l := listers.GenericListerFactory().GetLister(schema.GroupVersionResource{
-				Group:    gv[0],
-				Version:  gv[1],
+				Group:    gv.Group,
+				Version:  gv.Version,
 				Resource: controllerRef.Kind,
 			}, pod.GetNamespace())
 
--- a/cluster-autoscaler/utils/drain/drain.go
+++ b/cluster-autoscaler/utils/drain/drain.go
@@ -131,10 +131,10 @@ func GetPodsForDeletionOnNodeDrain(
 			// The assumption in the code below is that the controllerRef/owner of
 			// a pod resource will always be in the same namespace
 			// TODO: find a better way to handle this
-			l := listers.GenericListerFactory().GetLister(schema.GroupVersionResource{
-				Group:    gv.Group,
-				Version:  gv.Version,
-				Resource: controllerRef.Kind,
+			l := listers.GenericListerFactory().GetLister(schema.GroupVersionKind{
+				Group:   gv.Group,
+				Version: gv.Version,
+				Kind:    controllerRef.Kind,
 			}, pod.GetNamespace())
 
 			if _, err := l.Get(controllerRef.Name); err == nil {
--- a/cluster-autoscaler/utils/drain/drain.go
+++ b/cluster-autoscaler/utils/drain/drain.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	kube_util "k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
 	pod_util "k8s.io/autoscaler/cluster-autoscaler/utils/pod"
+	"k8s.io/klog/v2"
 )
 
 const (
@@ -111,9 +112,11 @@ func GetPodsForDeletionOnNodeDrain(
 		terminal := isPodTerminal(pod)
 
 		refKind := ""
+		controllerName := ""
 		controllerRef := ControllerRef(pod)
 		if controllerRef != nil {
 			refKind = controllerRef.Kind
+			controllerName = controllerRef.Name
 		}
 
 		if refKind != "" {
@@ -131,9 +134,12 @@ func GetPodsForDeletionOnNodeDrain(
 			}, pod.GetNamespace())
 
 			if _, err := l.Get(controllerRef.Name); err == nil {
+				klog.V(4).Infof("pod %s/%s is replicated by %s '%s'", pod.Namespace, pod.Name, refKind, controllerName)
 				replicated = true
 			} else {
-				return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: ControllerNotFound}, fmt.Errorf("owner %s for %s/%s is not available, err: %v", refKind, pod.Namespace, pod.Name, err)
+				notAvailableErr := fmt.Errorf("owner %s '%s' for %s/%s is not available, err: %v", refKind, controllerName, pod.Namespace, pod.Name, err)
+				klog.V(4).Info(notAvailableErr)
+				return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: ControllerNotFound}, notAvailableErr
 			}
 		}
 
--- a/cluster-autoscaler/utils/drain/drain.go
+++ b/cluster-autoscaler/utils/drain/drain.go
@@ -83,7 +83,6 @@ func GetPodsForDeletionOnNodeDrain(
 
 	pods = []*apiv1.Pod{}
 	daemonSetPods = []*apiv1.Pod{}
-	checkReferences := listers != nil
 	// filter kube-system PDBs to avoid doing it for every kube-system pod
 	kubeSystemPDBs := make([]*policyv1.PodDisruptionBudget, 0)
 	for _, pdb := range pdbs {
--- a/cluster-autoscaler/utils/drain/drain.go
+++ b/cluster-autoscaler/utils/drain/drain.go
@@ -121,25 +121,7 @@ func GetPodsForDeletionOnNodeDrain(
 		// so OwnerReference doesn't have its own Namespace field
 		controllerNamespace := pod.Namespace
 
-		if refKind == "ReplicationController" {
-			if checkReferences {
-				rc, err := listers.ReplicationControllerLister().ReplicationControllers(controllerNamespace).Get(controllerRef.Name)
-				// Assume a reason for an error is because the RC is either
-				// gone/missing or that the rc has too few replicas configured.
-				// TODO: replace the minReplica check with pod disruption budget.
-				if err == nil && rc != nil {
-					if rc.Spec.Replicas != nil && *rc.Spec.Replicas < minReplica {
-						return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: MinReplicasReached}, fmt.Errorf("replication controller for %s/%s has too few replicas spec: %d min: %d",
-							pod.Namespace, pod.Name, rc.Spec.Replicas, minReplica)
-					}
-					replicated = true
-				} else {
-					return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: ControllerNotFound}, fmt.Errorf("replication controller for %s/%s is not available, err: %v", pod.Namespace, pod.Name, err)
-				}
-			} else {
-				replicated = true
-			}
-		} else if pod_util.IsDaemonSetPod(pod) {
+		if pod_util.IsDaemonSetPod(pod) {
 			isDaemonSetPod = true
 			// don't have listener for other DaemonSet kind
 			// TODO: we should use a generic client for checking the reference.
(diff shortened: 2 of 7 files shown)


vadasambar:
This PR is still WIP. I will mention the reviewers here once I am done. Although it might not be in the best shape as a WIP PR but if the reviewers have any feedback, I would love to have it. :pray:

cluster-autoscaler/utils/drain/drain.go
```diff
@@ -121,25 +121,7 @@ func GetPodsForDeletionOnNodeDrain(
 		// so OwnerReference doesn't have its own Namespace field
 		controllerNamespace := pod.Namespace
 
-		if refKind == "ReplicationController" {
```
vadasambar, line 124: We can't remove the listers for the removed code  from [`ListerRegistry`]() because we use somewhere else ([ref1](https://github.com/kubernetes/autoscaler/blob/6b7c2911c277918881bd1b41de950188c93e29f7/cluster-autoscaler/core/scaledown/planner/controller.go#L34-L88), [ref4](https://github.com/kubernetes/autoscaler/blob/7a1668ef124fe98e8643cb6e8c2d0734fa72914e/cluster-autoscaler/core/static_autoscaler.go#L302))

cluster-autoscaler/utils/drain/drain.go
```diff
@@ -106,102 +105,16 @@ func GetPodsForDeletionOnNodeDrain(
 			continue
 		}
 
-		isDaemonSetPod := false
```
vadasambar, line 109: Not needed anymore.

cluster-autoscaler/utils/drain/drain.go
```diff
@@ -84,7 +83,6 @@ func GetPodsForDeletionOnNodeDrain(
 
 	pods = []*apiv1.Pod{}
 	daemonSetPods = []*apiv1.Pod{}
-	checkReferences := listers != nil
```
vadasambar, line 87: Not needed anymore.

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -59,14 +77,78 @@ type listerRegistryImpl struct {
 	jobLister                   v1batchlister.JobLister
 	replicaSetLister            v1appslister.ReplicaSetLister
 	statefulSetLister           v1appslister.StatefulSetLister
+	genericListerMap            map[string]dynamiclister.Lister
+	stopCh                      <-chan struct{}
+	dynamicClient               *dynamic.DynamicClient
+}
+
+// GetDynamicLister returns the lister for a particular GVR
+func (g *GenericListerFactory) GetLister(gvr schema.GroupVersionResource, namespace string) dynamiclister.Lister {
```
vadasambar, line 86: A PoC of this can be found here: https://github.com/vadasambar/dynamic-lister

cluster-autoscaler/utils/drain/drain.go
```diff
@@ -106,102 +106,38 @@ func GetPodsForDeletionOnNodeDrain(
...
-			} else {
+		if refKind != "" {
+			gv := strings.Split(controllerRef.APIVersion, "/")
+			// controllerRef doesn't have a namespace by design
+			// The controller/owner is either in the same namespace as the pod
+			// or it's assumed to be a cluster-scoped resource
+			// The assumption in the code below is that the controllerRef/owner of
+			// a pod resource will always be in the same namespace
+			// TODO: find a better way to handle this
+			l := listers.GenericListerFactory().GetLister(schema.GroupVersionResource{
```
vadasambar, line 127: I am using generic lister for all resources even when we have a lister for some in-built resources like DaemonSets, StatefulSets etc., to simplify the code at the cost additional overhead of initializing on-demand listers (per namespace). Not sure if this is the best approach to do this. Would love to have any feedback on this :pray:

x13n: How about making it an implementation detail of `GetLister`? If there's a dedicated lister already, it would use that, otherwise fall back to the dynamic one?

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -46,6 +57,19 @@ type ListerRegistry interface {
 	JobLister() v1batchlister.JobLister
 	ReplicaSetLister() v1appslister.ReplicaSetLister
 	StatefulSetLister() v1appslister.StatefulSetLister
+	GenericListerFactory() GenericListerFactory
```
x13n, line 60: What do we need such two-level interface for? Why not just have `GenericLister(gvr schema.GroupVersionResource, namespace string) dynamiclister.Lister` directly here?

cluster-autoscaler/utils/drain/drain.go
```diff
@@ -106,102 +107,43 @@ func GetPodsForDeletionOnNodeDrain(
...
-				}
-			} else {
+		if refKind != "" {
+			gv := strings.Split(controllerRef.APIVersion, "/")
+			// controllerRef doesn't have a namespace by design
+			// The controller/owner is either in the same namespace as the pod
+			// or it's assumed to be a cluster-scoped resource
+			// The assumption in the code below is that the controllerRef/owner of
+			// a pod resource will always be in the same namespace
+			// TODO: find a better way to handle this
```
x13n, line 129: Can we always list controllers from all namespaces? This is equivalent to what existing listers do, right?

x13n (changes requested):

cluster-autoscaler/utils/drain/drain.go
vadasambar: I like the idea. Thank you for the suggestion!

cluster-autoscaler/utils/drain/drain.go
```diff
@@ -106,102 +107,43 @@ func GetPodsForDeletionOnNodeDrain(
...
-						return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: MinReplicasReached}, fmt.Errorf("replication controller for %s/%s has too few replicas spec: %d min: %d",
-							pod.Namespace, pod.Name, rs.Spec.Replicas, minReplica)
-					}
-					replicated = true
-				} else {
-					return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: ControllerNotFound}, fmt.Errorf("replication controller for %s/%s is not available, err: %v", pod.Namespace, pod.Name, err)
-				}
-			} else {
+		if refKind != "" {
+			gv := strings.Split(controllerRef.APIVersion, "/")
```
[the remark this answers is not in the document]
vadasambar: Makes sense. I will use this. :+1:

cluster-autoscaler/utils/kubernetes/listers.go
vadasambar: The main reason is so that tests can implement this interface (like [this](https://github.com/kubernetes/autoscaler/pull/5419/files#diff-35dba75ac3715863c6460ab9d5515b582d1d05d705b5e597bb25a511ed42171fR177-R195) for example). I am still going through the unit tests. This interface might not be necessary but when I look at the tests it seems like having this would be useful. I would be happy to remove it if I am wrong because like you said not having the interface simplifies the code.

cluster-autoscaler/utils/drain/drain.go
vadasambar: Not sure if I understand your question. [The code in `master` branch assumes the controller/owner resource would be in the same namespace as the pod](https://github.com/kubernetes/autoscaler/blob/b3b7382819e24343ffeaf95727c164fe0879fc3f/cluster-autoscaler/utils/drain/drain.go#L126-L190). The code I wrote does the same thing i.e., it assumes the controller/owner resource is in the same namespace as the pod. This approach has a caveat where if the pod owner is a cluster scoped resource, our logic wouldn't work correctly (I haven't seen a case like this yet).

The comment above basically puts this caveat in words and the `TODO` is for handling the case of cluster-scoped resource as the owner. Maybe I should tweak the wording a bit. 

Let me know if that doesn't answer the question :pray:

cluster-autoscaler/utils/kubernetes/listers.go
vadasambar: Also, having a factory like this means, `GenericLister` can be used independently of the `ListerRegistry` (not sure if we have a use-case like that in cluster-autoscaler but I thought having that de-coupling might prove useful in the future.

cluster-autoscaler/utils/drain/drain.go
vadasambar: This idea might not work because there is no common `Lister` interface we can use as a return type. We could use `interface{}` as the return type but then the user would have to typecast the lister i.e., they'd have to know the lister type beforehand. 
e.g., 
 [`Lister` interface for dynamic/unstructured](https://github.com/kubernetes/autoscaler/blob/3955223eeeeea11089cc15779fcf1b4c5d8b54e9/cluster-autoscaler/vendor/k8s.io/client-go/dynamic/dynamiclister/interface.go#L24-L32) type is different from the [`ReplicaSet`'s Lister interface](https://github.com/kubernetes/autoscaler/blob/222efd9a4c4f5a933b79307d030c27d3c9913281/cluster-autoscaler/vendor/k8s.io/client-go/listers/apps/v1/replicaset.go#L28-L37) which is different from [Job's Lister interface](https://github.com/kubernetes/autoscaler/blob/222efd9a4c4f5a933b79307d030c27d3c9913281/cluster-autoscaler/vendor/k8s.io/client-go/listers/batch/v1/job.go#L28-L37) which is different from [StatefulSet's lister](https://github.com/kubernetes/autoscaler/blob/222efd9a4c4f5a933b79307d030c27d3c9913281/cluster-autoscaler/vendor/k8s.io/client-go/listers/apps/v1/statefulset.go#L28-L37) etc., 

I don't think there's a common Lister interface for in-built types but assuming there was, we would still have the same problem since unstructured's Lister interface is different.

vadasambar: I wonder if there's an elegant way to handle this problem.

vadasambar:
The code has a problem where [I pass in `Kind` as `Resource` to the GVR](https://github.com/vadasambar/autoscaler/blob/7408206217a907134dde15470f550a47f09eb5cc/cluster-autoscaler/utils/drain/drain.go#L134-L138) in https://github.com/kubernetes/autoscaler/pull/5419 but [that won't work](https://github.com/vadasambar/dynamic-lister/blob/2cffed3120efbd1144a1205382c2294786fc2f7f/main.go#L64-L67). `Resource` has to be plural (e.g., `Deployment` or `deployment` doesn't work. It has to be `deployments`. You can experiment with this using the [dynamic-lister PoC](https://github.com/vadasambar/dynamic-lister) by changing [this](https://github.com/vadasambar/dynamic-lister/blob/2cffed3120efbd1144a1205382c2294786fc2f7f/main.go#L46) line).

cluster-autoscaler/utils/drain/drain.go
x13n: Right, you'd have to wrap existing listers into the dynamic one and just use that return type.

cluster-autoscaler/main.go
```diff
@@ -380,6 +381,7 @@ func buildAutoscaler(debuggingSnapshotter debuggingsnapshot.DebuggingSnapshotter
 	kubeClientConfig.Burst = autoscalingOptions.KubeClientBurst
 	kubeClientConfig.QPS = float32(autoscalingOptions.KubeClientQPS)
 	kubeClient := createKubeClient(kubeClientConfig)
+	dynamicClient := dynamic.NewForConfigOrDie(kubeClientConfig)
```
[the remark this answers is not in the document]
vadasambar: I am not sure if we can create a `dynamicClient` out of `kubeClient` because [`createKubeClient`](https://github.com/kubernetes/autoscaler/blob/84b995233656c3a71c90f17d6aa2d39ca68fe9dd/cluster-autoscaler/main.go#L357-L359) returns a `kube_client.Interface` which actually is a [`ClientSet`](https://github.com/kubernetes/autoscaler/blob/bd2ff82aa1e369456022935591464d37f63881d6/cluster-autoscaler/vendor/k8s.io/client-go/kubernetes/clientset.go#L691-L699). [I don't think we can use ClientSet for creating a dynamicClient](https://github.com/kubernetes/autoscaler/blob/bd2ff82aa1e369456022935591464d37f63881d6/cluster-autoscaler/vendor/k8s.io/client-go/kubernetes/clientset.go#L134-L187) (would be happy to be corrected here).

Since we pass `kubeClientConfig` to `dynamic.NewForConfigDie`, all of the QPS/Burst settings are applied `dynamicClient` [here](https://github.com/kubernetes/autoscaler/blob/bd2ff82aa1e369456022935591464d37f63881d6/cluster-autoscaler/vendor/k8s.io/client-go/rest/config.go#L351-L364) which can be traced as:
- [`dynamic.NewForConfigOrDie(kubeClientConfig)`](https://github.com/kubernetes/autoscaler/blob/bd2ff82aa1e369456022935591464d37f63881d6/cluster-autoscaler/vendor/k8s.io/client-go/dynamic/simple.go#L61) -> [`NewForConfigAndClient` is called](https://github.com/kubernetes/autoscaler/blob/bd2ff82aa1e369456022935591464d37f63881d6/cluster-autoscaler/vendor/k8s.io/client-go/dynamic/simple.go#L78) -> [`RESTClientForConfigAndClient` is called](https://github.com/kubernetes/autoscaler/blob/bd2ff82aa1e369456022935591464d37f63881d6/cluster-autoscaler/vendor/k8s.io/client-go/dynamic/simple.go#L89) -> [place where `QPS` and `Burst` are set](https://github.com/kubernetes/autoscaler/blob/bd2ff82aa1e369456022935591464d37f63881d6/cluster-autoscaler/vendor/k8s.io/client-go/rest/config.go#L351-L364)

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -59,14 +85,66 @@ type listerRegistryImpl struct {
...
+// GetLister returns the lister for a particular GVR
+func (g listerRegistryImpl) GetLister(gvr schema.GroupVersionKind, namespace string) dynamiclister.Lister {
+	crd, err := g.crdLister.Get(fmt.Sprintf("%s/%s", gvr.Group, gvr.Kind))
+	if err != nil {
+		fmt.Println(fmt.Errorf("crd not found: %v", err))
+	}
+
+	resource, found, err := unstructured.NestedString(crd.Object, "spec", "names", "plural")
+	if !found {
+		fmt.Println(fmt.Errorf("couldn't find the field 'spec.names.plural' on the CRD '%s'", crd.GetName()))
```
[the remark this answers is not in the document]
vadasambar: Makes sense :+1:

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -368,3 +549,75 @@ func NewConfigMapListerForNamespace(kubeClient client.Interface, stopchannel <-c
...
+			klog.Error("err", err)
+		}
+		return len(no) > 0
+	})
+	if !synced {
+		// don't return an error but don't add
+		// this lister to listers map
+		// so that another attempt is made
+		// to create the lister and sync cache
+		klog.Error("couldn't sync cache")
```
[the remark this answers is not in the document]
vadasambar: Let me take another look at this.

cluster-autoscaler/utils/drain/drain.go
vadasambar: Let me think about this and get back. I am imagining something like a `BigLister` interface which would have the in-built listers along with the dynamic lister.

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -193,6 +291,89 @@ func NewUnschedulablePodInNamespaceLister(kubeClient client.Interface, namespace
...
+
+	reflector := cache.NewReflector(lw, unstructured.Unstructured{}, store, time.Hour)
+
+	crdLister := dynamiclister.New(store, gvr)
+
+	// Run reflector in the background so that we get new updates from the api-server
+	go reflector.Run(stopChannel)
+
+	// Wait for reflector to sync the cache for the first time
+	// TODO: check if there's a better way to do this (listing all the nodes seems wasteful)
```
vadasambar, line 359: TODO: remove this line (irrelevant)

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -368,3 +549,75 @@ func NewConfigMapListerForNamespace(kubeClient client.Interface, stopchannel <-c
 	go reflector.Run(stopchannel)
 	return lister
 }
+
+// // NewGenericListerFactory initializes a new generic lister factory
```
vadasambar, line 553: TODO: remove this code block

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -59,14 +85,66 @@ type listerRegistryImpl struct {
...
+// 	}
+
+// 	return NewGenericLister(g.dynamicClient, g.listersMap, g.stopCh, schema.GroupVersionResource{Group: gvr.Group,
+// 		Version:  gvr.Version,
+// 		Resource: resource}, namespace)
+// }
+
+// GetLister returns the lister for a particular GVR
+func (g listerRegistryImpl) GetLister(gvr schema.GroupVersionKind, namespace string) dynamiclister.Lister {
+	crd, err := g.crdLister.Get(fmt.Sprintf("%s/%s", gvr.Group, gvr.Kind))
```
vadasambar, line 119: I am retrieving `Resource` name from the CRD (crdLister is defined [here](https://github.com/kubernetes/autoscaler/pull/5419/files#diff-17617c4ca5760f3b99f687bc12f0b85eeba117803b7f796f1cbf6bf5be067c08R296)). I use `Resource` name to create GVR which is needed to create `List` and `Watch` functions in the `GenericLister` [here](https://github.com/kubernetes/autoscaler/pull/5419/files#diff-17617c4ca5760f3b99f687bc12f0b85eeba117803b7f796f1cbf6bf5be067c08R574-R581).

`Resource` has to be plural. ownerRef on pod doesn't have a `Resource`, it has a `Kind`. We can't convert `Kind` to `Resource` easily because `Kind` is singular and `Resource` is a plural in small case. 
e.g., If `Kind` is `Deployment`, `Resource` would be `deployments` (If you use `deployment` in place of `deployments` `List` and `Watch` functions don't work; to check this for yourself you can change the `nodes` to `node` [in this PoC](https://github.com/vadasambar/dynamic-lister/blob/3286411908eafa859158b7a16f60e1f4f5f19ce7/main.go#L46))
It's not easy to predict what would be the correct plural of the `Kind` because of which we need to look up the CRD to check the plural name. This approach gives us accurate results but also feels like an overkill. 

Would love any suggestion on this.

vadasambar: Ref: https://github.com/kubernetes/autoscaler/pull/5419#issuecomment-1409839966

cluster-autoscaler/main.go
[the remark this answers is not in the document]
x13n: `kube_client.Interface` is a superset of [rest.Interface](https://github.com/kubernetes/client-go/blob/v0.26.1/rest/client.go#L41), which is what [dynamic.New](https://pkg.go.dev/k8s.io/client-go/dynamic#New) expects. I didn't test this, but from docs it seems like what we want?

cluster-autoscaler/utils/kubernetes/listers.go
x13n: That's a good question. I don't think there's a better way though, as the plural string for the resource doesn't appear anywhere in the owner ref.

cluster-autoscaler/main.go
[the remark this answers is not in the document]
vadasambar: Looks like I missed that obvious function. Thanks for the link. I will test it out.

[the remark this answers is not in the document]
vadasambar: It works ([can be tested](https://github.com/vadasambar/dynamic-lister/blob/e4ecdc44c79c48be75b5ff8ff7a5205c768d428a/main.go#L41) by running [this PoC](https://github.com/vadasambar/dynamic-lister)). I have updated the code.

vadasambar:
> This pull-request has been approved by: _[vadasambar](https://github.com/kubernetes/autoscaler/pull/5419#)_

Not sure why it says this. You can't approve your own PR in Github.

cluster-autoscaler/utils/drain/drain.go
vadasambar: > Right, you'd have to wrap existing listers into the dynamic one and just use that return type.

Not sure if I understand this correctly. 

I am guessing the idea is to change the return-type for built-in listers (deployments, statefulsets etc.,) with dynamic lister 

e.g., 
```go
// ListerRegistry is a registry providing various listers to list pods or nodes matching conditions
type ListerRegistry interface {
	AllNodeLister() NodeLister
	ReadyNodeLister() NodeLister
	ScheduledPodLister() PodLister
	UnschedulablePodLister() PodLister
	PodDisruptionBudgetLister() PodDisruptionBudgetLister
	DaemonSetLister() v1appslister.DaemonSetLister
	ReplicationControllerLister() v1lister.ReplicationControllerLister
	JobLister() v1batchlister.JobLister
	ReplicaSetLister() v1appslister.ReplicaSetLister
	StatefulSetLister() v1appslister.StatefulSetLister
	GetLister(gvr schema.GroupVersionKind, namespace string) dynamiclister.Lister
}
```
https://github.com/kubernetes/autoscaler/blob/13a3bf943d456cfdcdd266ffbae0a0231533b8c5/cluster-autoscaler/utils/kubernetes/listers.go#L49-L61

to 
```go
// ListerRegistry is a registry providing various listers to list pods or nodes matching conditions
type ListerRegistry interface {
	AllNodeLister() dynamiclister.Lister
	ReadyNodeLister() dynamiclister.Lister
	ScheduledPodLister() dynamiclister.Lister
	UnschedulablePodLister() dynamiclister.Lister
	PodDisruptionBudgetLister() dynamiclister.Lister
	DaemonSetLister() dynamiclister.Lister
	ReplicationControllerLister() dynamiclister.Lister
	JobLister() dynamiclister.Lister
	ReplicaSetLister() dynamiclister.Lister
	StatefulSetLister() dynamiclister.Lister
	GetLister(gvr schema.GroupVersionKind, namespace string) dynamiclister.Lister
}
```

vadasambar: I like the idea.

One problem I see is, if we use `dynamic.Lister` as a return type (or wrapper type), accessing fields in the struct [won't be as straightforward](https://github.com/kubernetes/autoscaler/blob/13a3bf943d456cfdcdd266ffbae0a0231533b8c5/cluster-autoscaler/utils/kubernetes/listers.go#L267) as just doing something like `ob.Spec.replicas` ([which is possible for in-built types](https://github.com/kubernetes/autoscaler/blob/222efd9a4c4f5a933b79307d030c27d3c9913281/cluster-autoscaler/vendor/k8s.io/client-go/listers/apps/v1/statefulset.go#L33))

vadasambar: I will think a bit more around https://github.com/kubernetes/autoscaler/pull/5419#discussion_r1099676412

x13n:
After thinking more about this change I convinced myself it might be introducing too much complexity for little extra benefit. My reasoning is: 
- There is a way out: if there are workloads affected by this change, users can always configure `safe-to-evict` annotation to indicate the behavior they want.
- Nodes may get restarted for other reasons than just CA - version upgrades & node crashes/repairs will also cause pods owned by custom controllers to disappear sometimes.
- I have a bit of general architectural concern about making CA aware of so many additional controllers in the cluster. Where possible I think we should strive for simplicity.
- Extra QPS to kube-apiserver may throttle other CA behavior (e.g. draining nodes in scale down).
- If, for whatever reason, it turns out we need to have such elaborate mechanism, we can always revisit this in the future.

Given the above, I'd suggest limiting the scope and start with blocking scale down only on bare pods (i.e. having no owner refs) and on owner refs pointing to missing *known* controllers (i.e. the ones we support today, to prevent backwards incompatible changes). WDYT?

vadasambar:
> After thinking more about this change I convinced myself it might be introducing too much complexity for little extra benefit. My reasoning is:
> 
>     * There is a way out: if there are workloads affected by this change, users can always configure `safe-to-evict` annotation to indicate the behavior they want.
> 
>     * Nodes may get restarted for other reasons than just CA - version upgrades & node crashes/repairs will also cause pods owned by custom controllers to disappear sometimes.
> 
>     * I have a bit of general architectural concern about making CA aware of so many additional controllers in the cluster. Where possible I think we should strive for simplicity.
> 
>     * Extra QPS to kube-apiserver may throttle other CA behavior (e.g. draining nodes in scale down).
> 
>     * If, for whatever reason, it turns out we need to have such elaborate mechanism, we can always revisit this in the future.
> 
> 
> Given the above, I'd suggest limiting the scope and start with blocking scale down only on bare pods (i.e. having no owner refs) and on owner refs pointing to missing _known_ controllers (i.e. the ones we support today, to prevent backwards incompatible changes). WDYT?

I kind of agree. I didn't think our idea would turn into something big and complex like this. I like the idea of querying for owner of the pods but I have similar concerns over introducing such a level of complexity (for not that big of a benefit)

I think we can close this PR and keep it for posterity while I can create another PR with the limited scope as the beginning step. If we feel like we need an elaborate solution like the current PR, we can come back to it again.

x13n:
Thanks! If you're planning to follow up with a separate PR, I'm going to close this one.

/close

vadasambar:
> Thanks! If you're planning to follow up with a separate PR, I'm going to close this one.
> 
> /close

Thanks. I will create a follow-up PR

Issue #4095, opened 2021-05-21
[cluster-autoscaler][AWS] Repeated deletion every 10s of same node takes AWS ASG down to min size killing multiple running pods


f-ld:
**Which component are you using?**:
cluster-autoscaler 

**What version of the component are you using?**:
1.14.1 (but investigations show it could be the same with newer versions) 

**What environment is this in?**:
Running on kops managed cluster on AWS

**What did you expect to happen?**:
I would not expect to see all nodes of a node group to be stopped while some pods are to be executed on them.

**What happened instead?**:
See logs below.
We have some nodes that take some time to join the k8s cluster. So AWS creates them (it seems) but they do not join. So after 15 minutes the cluster autoscaler tries to remove the "unregistered node" which usually works well.
But from time to time we can see something like in attached screenshot:
- every 10 secs the node is attempted to remove
- because of [this piece of code](https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-release-1.14/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go#L227-L228), the group size is decreased on every attemtp down to min (0) when it complains about not being able to go below min
- then the group notices it needs some more nodes (because more pods are created) so starts counting from 0.

On AWS side, in CloudTrail we can see for example:
- SetDesiredCapacity to 6
- Multiple calls to TerminateInstanceInAutoScalingGroup for instance "i-0d9e512ddf8a7321b"
- SetDesiredCapacity to 2

**How to reproduce it (as minimally and precisely as possible)**:
This happened twice in months (or we noticed it only twice)
No idea on how this can be reproduced, probably understanding how [involved code can fail without returning an error](https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-release-1.14/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go#L271-L278) would be a good clue.

**Anything else we need to know?**:
We are in the process of upgrading our k8s cluster versions, but opening this ticket since from what I can see there's not been big changes around this.

**Logs**
Here are logs filtered on instance ID that got removed multiple times. Log on `W0520 06:21:48.687757` shows the problem: we reached min size (0) because of that repeated deletion.

Note: I think the access denial that appears after some time are due to the node not existing anymore (so 403 instead of 404 from AWS). We were not changing anything on AWS permissions side at that time that could explain it.
```
[...]

PR #5420, opened 2023-01-13
cluster-autoscaler: terminating an instance does not mean there will be less desired instances

grosser:
#### What type of PR is this?

/kind bug

while working on https://github.com/kubernetes/autoscaler/pull/5411 to fix https://github.com/kubernetes/autoscaler/issues/4095 I realized that the core bug is that we decrement the currentSize (aka asg desired size) when terminating an instance
... but that's now how aws works, terminating an instance does not decreases the desired size, it only makes a new instance come up

(https://github.com/kubernetes/autoscaler/pull/5411 is still useful since it reduces api call overhead and log spam)

partially reverts https://github.com/kubernetes/autoscaler/pull/1378

/cc @alam0rt  @johanneswuerbach @aleksandra-malinowska

Fixes https://github.com/kubernetes/autoscaler/issues/4095

```release-note
NONE
```

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
+++ b/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
@@ -309,9 +309,6 @@ func (m *asgCache) DeleteInstances(instances []*AwsInstanceRef) error {
 				return err
 			}
 			klog.V(4).Infof(*resp.Activity.Description)
-
-			// Proactively decrement the size so autoscaler makes better decisions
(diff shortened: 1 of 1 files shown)


johanneswuerbach:
> terminating an instance does not decreases the desired size, it only makes a new instance come up

The call above is made with `ShouldDecrementDesiredCapacity: aws.Bool(true)`, so I think it actually does both, terminate the instance and decreases the desired size.

grosser:
ah missed that, thx!

PR #5423, opened 2023-01-17
Add scale down candidates observer

yaroslava-serdiuk:
#### What type of PR is this?
Create scale down candidates observer and put cluster registry there

/kind cleanup


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -163,6 +163,10 @@ func NewStaticAutoscaler(
 	}
 
 	clusterStateRegistry := clusterstate.NewClusterStateRegistry(autoscalingContext.CloudProvider, clusterStateConfig, autoscalingContext.LogRecorder, backoff)
+	autoscalingContext.ScaleDownCandidatesObservers = []scaledown.ScaleDownCandidatesObserver{clusterStateRegistry}
+	if p, ok := processors.ScaleDownNodeProcessor.(scaledown.ScaleDownCandidatesObserver); ok {
```
x13n, line 167: This feels a bit hacky. In particular, while in principle we allow multiple observers, in practice they are hardcoded here, making introduction of new observers tricky. How about passing a slice of observers to register in autoscaling opts? Then the code here would register `clusterStateRegistry` and all the extra observers passed through the options.

x13n (changes requested):

cluster-autoscaler/core/static_autoscaler.go
yaroslava-serdiuk: Moving ScaleDownCandidatesObservers to the autoscaling opts creates an import cycle. 

We already have `if opts.ParallelDrain` condition. Maybe I can make processor assignment there and add an observer there.

cluster-autoscaler/context/autoscaling_context.go
```diff
@@ -57,6 +57,8 @@ type AutoscalingContext struct {
 	DebuggingSnapshotter debuggingsnapshot.DebuggingSnapshotter
 	// ScaleDownActuator is the interface for draining and deleting nodes
 	ScaleDownActuator scaledown.Actuator
+	// ScaleDownCandidatesObservers is the list of scale down candidates observers.
+	ScaleDownCandidatesObservers []scaledown.ScaleDownCandidatesObserver
```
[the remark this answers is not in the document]
yaroslava-serdiuk: Updated

cluster-autoscaler/core/static_autoscaler.go
x13n: Hm... if cyclic imports are the problem, maybe bundling the observer interface with other scale down interfaces is not the right place. WDYT about moving all the newly added code to a dedicated package?

cluster-autoscaler/core/scaledown/scaledown.go
```diff
@@ -85,3 +85,34 @@ type ActuationStatus interface {
...
+}
+
+// Register adds new observer to the list.
+func (l *CandidatesObserverList) Register(observer CandidatesObserver) {
+	l.observers = append(l.observers, observer)
+}
+
+// Update updates scale down candidates for each observer.
+func (l *CandidatesObserverList) Update(nodes []*apiv1.Node, now time.Time) {
+	if len(l.observers) == 0 {
```
x13n, line 107: nit: This if is superfluous, I think, so maybe drop it?

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -571,7 +575,7 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 		typedErr := a.scaleDownPlanner.UpdateClusterState(podDestinations, scaleDownCandidates, actuationStatus, pdbs, currentTime)
 		// Update clusterStateRegistry and metrics regardless of whether ScaleDown was successful or not.
 		unneededNodes := a.scaleDownPlanner.UnneededNodes()
-		a.clusterStateRegistry.UpdateScaleDownCandidates(unneededNodes, currentTime)
+		a.ScaleDownCandidatesObservers.Update(unneededNodes, currentTime)
 		metrics.UpdateUnneededNodesCount(len(unneededNodes))
```
x13n, line 579: nit: Metric updating is yet another candidate for an observer of unneeded nodes.

cluster-autoscaler/core/static_autoscaler.go
yaroslava-serdiuk: It seems for me that autoscaling opts is a set of options set via flags +-, so autoscaling context object sounds like a better place for scale down candidates observers. Also, we are now creating the initial list there, instead of static autoscaler, so maybe it solves initial issue?

cluster-autoscaler/core/static_autoscaler.go
yaroslava-serdiuk: Yes, but this requires creating an object instead of just calling function, not sure we need it.

cluster-autoscaler/core/scaledown/scaledown.go
x13n: (re-opening: I'm not sure why this was marked as resolved.)

cluster-autoscaler/core/static_autoscaler.go
x13n: I think separating metrics reporting from the rest of the logic would be beneficial, we'd get a nice separation of concerns, making the code more testable. That'd require much more than just removing a single metric update, so yeah, I guess it's not very beneficial on its own.

cluster-autoscaler/core/static_autoscaler.go
x13n: My concern here is: how can I register a new scaledown candidates observer without modifying code of StaticAutoscaler? Or to put it differently: how can we make static autoscaler unaware of who wants to observe scale down candidates?

yaroslava-serdiuk: Hm, I think the common approach that we used so far is to modify processors. I introduced ScaleDownCandidatesObservers there, please take a look

cluster-autoscaler/processors/processors.go
```diff
@@ -61,6 +62,8 @@ type AutoscalingProcessors struct {
 	CustomResourcesProcessor customresources.CustomResourcesProcessor
 	// ActionableClusterProcessor is interface defining whether the cluster is in an actionable state
 	ActionableClusterProcessor actionablecluster.ActionableClusterProcessor
+	// ScaleDownCandidatesObservers  is used to Update and Register new scale down candidates observer.
+	ScaleDownCandidatesObservers scaledowncandidates.CandidatesObserverProcessor
```
x13n, line 66: nit: Wouldn't `ScaleDownCandidatesNotifier` better reflect the purpose of this object? (Probably worth aliging the type name then too.)

cluster-autoscaler/main.go
```diff
@@ -403,6 +404,10 @@ func buildAutoscaler(debuggingSnapshotter debuggingsnapshot.DebuggingSnapshotter
 		podlistprocessor.NewCurrentlyDrainedNodesPodListProcessor(),
 		podlistprocessor.NewFilterOutSchedulablePodListProcessor(opts.PredicateChecker),
 	)
+	if autoscalingOptions.ParallelDrain {
+		opts.Processors.ScaleDownNodeProcessor = nodes.NewScaleDownCandidatesSortingProcessor()
+		opts.Processors.ScaleDownCandidatesObservers.Register(nodes.NewScaleDownCandidatesSortingProcessor())
```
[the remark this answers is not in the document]
yaroslava-serdiuk: Oh right, thanks

cluster-autoscaler/processors/scaledowncandidates/scale_down_candidates_observer.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+// CandidatesObserver is an observer of scale down candidates
+type CandidatesObserver interface {
+	// UpdateScaleDownCandidates updates scale down candidates.
+	UpdateScaleDownCandidates([]*apiv1.Node, time.Time)
+}
+
+// CandidatesObserverProcessor is a processor that Updates and Register new observer of scale down candidates.
+type CandidatesObserverProcessor interface {
```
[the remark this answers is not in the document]
yaroslava-serdiuk: I added interface for consistency with AuotscalingProcessors struct, but agree it's not needed.

PR #5430, opened 2023-01-19
Rephrase error messages specific for GPUs.

hbostan:
#### What type of PR is this?

/kind cleanup

Some error messages related to custom resources assumes that the custom resource is always GPU. Rephrasing these error messages to clear up any confusion.

```release-note
NONE
```

```docs

```


MaciekPytel:
/lgtm
/approve

One question I have is - given that we're talking about resource limits, I think the only way of specifying those is using --gpu-total flag? I realize you can use the flag for any resource, but isn't it equally confusing (or maybe even more so?). I don't think it blocks this change in any way, but maybe we also want to add a new flag (while keeping the old one for backward compatibility)? Than again - I'm not sure if this works at all for non-gpu resources without a custom implementation of processor?

PR #5439, opened 2023-01-23
GCE: Add os parameters to MigOsInfo interface

yaroslava-serdiuk:
#### What type of PR is this?

/kind cleanup

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/os_reserved.go
+++ b/cluster-autoscaler/cloudprovider/gce/os_reserved.go
@@ -16,13 +16,23 @@ limitations under the License.
 
 package gce
 
+// MigOsInfo store os parameters.
+type MigOsInfo interface {
+	// Os return operating system.
+	Os() OperatingSystem
+	// OsDistribution return operating system distribution.
+	OsDistribution() OperatingSystemDistribution
+	// Arch return system architecture.
+	Arch() SystemArchitecture
+}
+
 // OsReservedCalculator calculates the OS reserved values.
 type OsReservedCalculator interface {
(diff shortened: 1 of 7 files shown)


cluster-autoscaler/cloudprovider/gce/os_reserved.go
```diff
@@ -16,13 +16,23 @@ limitations under the License.
...
+	// Arch return system architecture.
+	Arch() SystemArchitecture
+}
+
 // OsReservedCalculator calculates the OS reserved values.
 type OsReservedCalculator interface {
 	// CalculateKernelReserved computes how much memory OS kernel will reserve.
 	// NodeVersion parameter is optional. If empty string is passed a result calculated using default node version will be returned.
-	CalculateKernelReserved(physicalMemory int64, os OperatingSystem, osDistribution OperatingSystemDistribution, arch SystemArchitecture, nodeVersion string) int64
+	CalculateKernelReserved(m MigOsInfo, physicalMemory int64, nodeVersion string) int64
```
BigDarkClown, line 33: I think now we can remove the nodeVersion as well, both from these functions and from Mig interface. It is not needed in the GCE implementation, and we are now able to add this in internal fork one using MigOsInfo.

BigDarkClown (commented):
That change looks fantastic, it cleans up the interfaces so much! I added a little comment regarding removing the node version, but other than that - great job!

cluster-autoscaler/cloudprovider/gce/os_reserved.go
yaroslava-serdiuk: Right, done

BigDarkClown (approved):

Issue #5372, opened 2022-12-16
Cluster Autoscaler: pick up the fix for k/k#114509


towca:
https://github.com/kubernetes/kubernetes/issues/114509 is a bug in scheduler which affects Cluster Autoscaler at scale, introduced in k8s 1.24. It was already fixed on k/k master (https://github.com/kubernetes/kubernetes/pull/114518), and the fix will be cherry-picked to 1.24 and 1.25. After that happens, and new k8s patch releases are tagged, we need to update the vendor in 1.24 and 1.25 to pick up the fix, and cut new CA patch versions.

Bryce-Soghigian:
Looks like it was cherry-picked here
 https://github.com/kubernetes/kubernetes/pull/114524
 https://github.com/kubernetes/kubernetes/pull/114525
 
 So we are just waiting on the patch release.https://kubernetes.io/releases/patch-releases/#1-25

jayantjain93:
PRs are open for all affected versions.

PR #5443, opened 2023-01-25
update vendor to v1.27.0-alpha.1

### LINKED ISSUE #5397: Adapt to scheduler framework interface changes in RunFilterPlugins
In https://github.com/kubernetes/kubernetes/pull/114629, we changed the return type of `RunFilterPlugins` from a map to `*Status`, this will influence autoscaler when we bump the dependence of k8s.

jayantjain93:
#### What type of PR is this?

/kind bug

Fixes #5397 

[5372](https://github.com/kubernetes/autoscaler/issues/5372)

```release-note

```

```docs

```


cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
```diff
@@ -126,15 +126,8 @@ func (p *SchedulerBasedPredicateChecker) FitsAnyNodeMatching(clusterSnapshot clu
...
-		ok := true
-		for _, filterStatus := range filterStatuses {
-			if !filterStatus.IsSuccess() {
-				ok = false
-				break
-			}
-		}
-		if ok {
+		filterStatus := p.framework.RunFilterPlugins(context.TODO(), state, pod, nodeInfo)
+		if filterStatus == nil || filterStatus.IsSuccess() {
```
towca, line 130: nit: Status.IsSuccess already returns true for nil statuses, so I don't think the nil check is needed.

jayantjain93: It's actually the other way around.
Current implementation returns nil for success. Adding IsSuccess() is "future-proof" incase they return non-nil obj for success=true

towca: As discussed offline, `IsSuccess` already handled status being nil, so we don't have to do it again.

towca:
This fixes #5397.

cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
jayantjain93: offline conversation to remove  nil check. because nil obj is handled by `IsSuccess` func

PR #5444, opened 2023-01-25
Add force Daemon Sets option

BigDarkClown:
#### What type of PR is this?

/kind feature

This PR adds an optional feature that force-schedules all Daemon Set pods on node group templates. That means that node groups which cannot schedule all the expected daemon sets (eg. because of small nodes) will no longer be chosen during scale-up. 

The change is disabled and guarded by a manifest flag, so it should not have any effect on the component yet.

### DIFF
--- a/cluster-autoscaler/simulator/nodes.go
+++ b/cluster-autoscaler/simulator/nodes.go
@@ -17,46 +17,56 @@ limitations under the License.
 package simulator
 
 import (
+	appsv1 "k8s.io/api/apps/v1"
 	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/daemonset"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
 	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
 
-	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
 	pod_util "k8s.io/autoscaler/cluster-autoscaler/utils/pod"
 )
 
-// getRequiredPodsForNode returns a list of pods that would appear on the node if the
-// node was just created (like daemonset and manifest-run pods). It reuses kubectl
-// drain command to get the list.
-func getRequiredPodsForNode(nodename string, podsForNodes map[string][]*apiv1.Pod) ([]*apiv1.Pod, errors.AutoscalerError) {
-	allPods := podsForNodes[nodename]
-
-	return filterRequiredPodsForNode(allPods), nil
-}
-
 // BuildNodeInfoForNode build a NodeInfo structure for the given node as if the node was just created.
-func BuildNodeInfoForNode(node *apiv1.Node, podsForNodes map[string][]*apiv1.Pod) (*schedulerframework.NodeInfo, errors.AutoscalerError) {
-	requiredPods, err := getRequiredPodsForNode(node.Name, podsForNodes)
-	if err != nil {
-		return nil, err
-	}
-	result := schedulerframework.NewNodeInfo(requiredPods...)
-	result.SetNode(node)
-	return result, nil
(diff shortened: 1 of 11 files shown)


cluster-autoscaler/simulator/nodes.go
```diff
@@ -17,46 +17,43 @@ limitations under the License.
...
+		// Ignore scheduled pods in deletion phase
 		if pod.DeletionTimestamp != nil {
 			continue
 		}
-
-		if pod_util.IsMirrorPod(pod) || pod_util.IsDaemonSetPod(pod) {
-			selectedPods = append(selectedPods, allPods[id])
+		// Add scheduled mirror pods and scheduled DS pods if not force scheduling DS
+		if pod_util.IsMirrorPod(pod) || (!forceDaemonSets && pod_util.IsDaemonSetPod(pod)) {
+			// Add scheduled mirror pods
```
x13n, line 44: nit: This comment is true only if `forceDaemonSets` is false, otherwise it will also include DS pods. I'd just drop it.

cluster-autoscaler/simulator/nodes.go
```diff
@@ -17,46 +17,43 @@ limitations under the License.
...
+	for _, pod := range scheduledPods {
+		// Ignore scheduled pods in deletion phase
 		if pod.DeletionTimestamp != nil {
 			continue
 		}
-
-		if pod_util.IsMirrorPod(pod) || pod_util.IsDaemonSetPod(pod) {
-			selectedPods = append(selectedPods, allPods[id])
+		// Add scheduled mirror pods and scheduled DS pods if not force scheduling DS
+		if pod_util.IsMirrorPod(pod) || (!forceDaemonSets && pod_util.IsDaemonSetPod(pod)) {
```
x13n, line 43: This will ignore all DaemonSet-like (but not really DaemonSet) pods for which `IsDaemonSetPod` currently returns true. I think we need to distinguish between "real DS" and "DS-like" pods here to make this function work correctly.

cluster-autoscaler/simulator/nodes_test.go
```diff
@@ -17,175 +17,175 @@ limitations under the License.
...
-			name:      "should return daemonset pod",
-			inputPods: []*apiv1.Pod{pod1, podDaemonset},
-			want:      []*apiv1.Pod{podDaemonset},
+			name: "everything together [forceDS=false]",
+			node: test.BuildTestNode("n", 1000, 10),
+			pods: []*apiv1.Pod{
+				test.BuildScheduledTestPod("p1", 100, 1, "n"),
+				test.BuildScheduledTestPod("p2", 100, 1, "n"),
+				test.SetMirrorPodSpec(test.BuildScheduledTestPod("p3", 100, 1, "n")),
+				func() *apiv1.Pod {
```
x13n, line 136: nit: I'd consider introducing a named helper function instead of in-lining functions in test cases, especially when multiple cases do the same thing (e.g. add deletion timestamp). Ideally, I'd love to have some builders for test pods and nodes, but that's clearly out of scope here.

x13n (changes requested):

cluster-autoscaler/simulator/nodes.go
BigDarkClown: That is a good point. I think we can remedy this by checking which DS pods are already present on the chosen node and inserting new DS pods only for DSs that are not present. That way we will still keep all the "DS-like" pods to the template.

cluster-autoscaler/simulator/nodes.go
BigDarkClown: Done

cluster-autoscaler/simulator/nodes.go
BigDarkClown: Changed, PTAL

cluster-autoscaler/simulator/nodes_test.go
BigDarkClown: Done

cluster-autoscaler/simulator/nodes.go
BigDarkClown: Done

BigDarkClown:
/unhold 
Tested the changes, everything looks

Issue #5448, opened 2023-01-26
Migrate from the notion of "GPU" to a more general notion of "Accelerators"


hbostan:
**Which component are you using?**:
cluster-autoscaler

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:

A lot of the code is specific to GPUs and isn't extensible to other type of accelerators (e.g. TPUs).

**Describe the solution you'd like.**:
Move away from the notion of GPU to more general notion of Accelerator, refactoring/updating the code as necessary.
Planned changes:
* Generalize GPU specific penalty in price expander to use Accelerators.
* Adapt custom resource processor(s) to use Accelerators.
* Eventually deprecate GPU related methods from CloudProvider.

**Describe any alternative solutions you've considered.**:

**Additional context.**:

PR #5459, opened 2023-02-01
Add GpuConfig to cloud provider. Use GpuConfig in utilization calculations.

hbostan:
#### What type of PR is this?

/kind feature

Extends the GPU notion to all GPU-like accelerators. It adds the GpuConfig struct and GetNodeGpuConfig to cloud provider for this. These methods are used in utilization calculations for scale-downs which were previously depending on single a gpu label and resource name. With GpuConfig different labels and resource names are supported.

This PR doesn't completely fix this issue #5448 But it is a starting point.

```release-note
NONE
```

```docs

```


hbostan:
I did messed up with my local branch and commits. I guess everything should be okay now except labels. helm-charts and vertical-pod-autoscaler labels seems wrong and are added while I was trying to fix the commit history. Sorry for the confusion :)

cluster-autoscaler/cloudprovider/cloud_provider.go
```diff
@@ -127,6 +134,11 @@ type CloudProvider interface {
 	// GetAvailableGPUTypes return all available GPU types cloud provider supports.
 	GetAvailableGPUTypes() map[string]struct{}
 
+	// GetNodeAccelerator returns the label and resource name for the accelerator added to node.
+	GetNodeAccelerator(*apiv1.Node) Accelerator
```
BigDarkClown, line 138: nit: I think it's worth to consider for this method to return the Accelerator pointer rather than struct. That way we would be able to better indicate the "no accelerator present" behaviour.

Also, there should be some info in the comment above about what happens if the node has no accelerator.

cluster-autoscaler/cloudprovider/test/test_cloud_provider.go
```diff
@@ -120,6 +121,27 @@ func (tcp *TestCloudProvider) GetAvailableGPUTypes() map[string]struct{} {
...
+	availableAccelerators := []cloudprovider.Accelerator{}
+	gpuLabel := tcp.GPULabel()
+	availableGPUTypes := tcp.GetAvailableGPUTypes()
+	for acceleratorType := range availableGPUTypes {
+		accelerator := cloudprovider.Accelerator{Label: gpuLabel, Type: acceleratorType, ResourceName: gpu.ResourceNvidiaGPU}
+		availableAccelerators = append(availableAccelerators, accelerator)
+	}
+	return availableAccelerators
+}
+
```
BigDarkClown, line 144: I don't like how there is a semi-identical implementation copy for every cloud provider. That way 
1. We are duplicating a lot of code - this commit has 2k lines, it could easily be a fraction of that.
2. I some implementations we have 2 different accelerator info source of truth - accelerator functions you added should always use `GpuLabel()` and `GetAvailableGPUTypes()` methods, otherwise we risk that somebody will change only one of them and the behaviour will diverge.

I think you should add the implementation to a separate method which takes the `CloudProvider` as an argument. Then you can use it in every CloudProvider implementation, this should be much cleaner.

cluster-autoscaler/cloudprovider/externalgrpc/externalgrpc_cloud_provider.go
```diff
@@ -262,6 +264,68 @@ func (e *externalGrpcCloudProvider) GetAvailableGPUTypes() map[string]struct{} {
...
+		accelerator := cloudprovider.Accelerator{
+			Label:        pbAccelerator.GetLabel(),
+			Type:         pbAccelerator.GetType(),
+			ResourceName: apiv1.ResourceName(pbAccelerator.GetResourceName()),
+		}
+		availableAccelerators = append(availableAccelerators, accelerator)
+	}
+	e.availableAcceleratorsCache = availableAccelerators
+	return availableAccelerators
+}
```
BigDarkClown, line 327: This separate cache is really not needed, you could just use the `GpuLabel()` and `GetAvailableGPUTypes()` which already have caching.

cluster-autoscaler/core/scaledown/eligibility/eligibility.go
```diff
@@ -49,6 +48,8 @@ type utilizationThresholdGetter interface {
 	GetScaleDownUtilizationThreshold(context *context.AutoscalingContext, nodeGroup cloudprovider.NodeGroup) (float64, error)
 	// GetScaleDownGpuUtilizationThreshold returns ScaleDownGpuUtilizationThreshold value that should be used for a given NodeGroup.
 	GetScaleDownGpuUtilizationThreshold(context *context.AutoscalingContext, nodeGroup cloudprovider.NodeGroup) (float64, error)
```
BigDarkClown, line 50: Why keep this method? It's redundant now, I think we should remove it.

cluster-autoscaler/utils/test/test_utils.go
```diff
@@ -240,6 +241,24 @@ func GetGPULabel() string {
...
+			Label:        gpuLabel,
+			Type:         gpuType,
+			ResourceName: resourceNvidiaGPU,
+		}
+	}
+	return cloudprovider.Accelerator{
+		Label:        "",
+		Type:         "",
+		ResourceName: "",
+	}
```
BigDarkClown, line 259: This is equivalent to `return cloudprovider.Accelerator{}`, but much less concise.

cluster-autoscaler/cloudprovider/test/test_cloud_provider.go
hbostan: * Created two functions in "cloud_provider.go" called `GetNodeAcceleratorFromCloudProvider()` and `GetAvailableAcceleratorsFromCloudProvider()`. Each cloud provider now calls these functions, and these functions work using `GPULabel()` and `GetAvailableGPUTypes()`.
* However, in order not to create a cyclic dependency between `gpu` and `cloudprovider` packages, each cloud provider has to check whether a node has an accelerator (by calling `gpu.NodeHasGpu()`) before calling  `GetNodeAcceleratorFromCloudProvider()`. This is the best solution I could come up with.

cluster-autoscaler/cloudprovider/externalgrpc/externalgrpc_cloud_provider.go
hbostan: Removed these caches. Now only using `GPULabel()` and `GetAvailableGPUTypes()`

cluster-autoscaler/core/scaledown/eligibility/eligibility.go
hbostan: Removed `GetScaleDownGpuUtilizationThreshold()`.

cluster-autoscaler/utils/test/test_utils.go
hbostan: Now returning nil, instead of an accelerator with empty strings.

cluster-autoscaler/cloudprovider/test/test_cloud_provider.go
BigDarkClown: Hmm, what if you moved these two functions to `gpu` package instead? This way you could use the `NodeHasGpu` function without cyclic imports. I also like this option because:
1. `cloudprovider.go` is mostly interfaces used in a lot of places in the code. Adding implementations there is not that great - if we would want to change the implementation in the future in some way that require imports, then we could potentially create a whole bunch of non trivial cyclic imports.
2. Both `GetNodeAcceleratorFromCloudProvider()` and `GetAvailableAcceleratorsFromCloudProvider()` implementations are GPU specific, so adding them in a `gpu.go` seems more intuitive. I would also change their names to  `GetNodeGPUFromCloudProvider()` and `GetAvailableGPUFromCloudProvider()` for the same reason.

cluster-autoscaler/processors/nodegroupconfig/node_group_config_processor.go
```diff
@@ -33,6 +33,8 @@ type NodeGroupConfigProcessor interface {
 	GetScaleDownUtilizationThreshold(context *context.AutoscalingContext, nodeGroup cloudprovider.NodeGroup) (float64, error)
 	// GetScaleDownGpuUtilizationThreshold returns ScaleDownGpuUtilizationThreshold value that should be used for a given NodeGroup.
 	GetScaleDownGpuUtilizationThreshold(context *context.AutoscalingContext, nodeGroup cloudprovider.NodeGroup) (float64, error)
```
BigDarkClown, line 35: Also this method should no long be necessary, sorry for missing that.

hbostan: Removed `GetScaleDownGpuUtilizationThreshold()`

cluster-autoscaler/cloudprovider/test/test_cloud_provider.go
hbostan: Moved those two functions to the `gpu` package and renamed as suggested.

towca (commented):
I know this PR wasn't aiming at covering everything, but just wanted to make sure the following changes are at least planned:
* The price expander has a GPU-specific penalty in its scoring function - this should probably be adapted to Accelerators now.
* We already have a concept of "CustomResources" that Accelerators fit into, managed by `CustomResourcesProcessor`. The processor is used for 2 purposes - hacking node readiness for when a custom resource is in fact unready, and calculating cluster-wide resource limits. Currently the only such processor in this repo is GPU-related - it hacks the nodes to be unready when the GPU resource is not in allocatable, and handles GPU resource limits by looking at the GPU label. Do we expect all Accelerators to follow these semantics? The processor should probably be adapted to Accelerators.
* In general, do we expect any GPU-specific handling after the whole migration is complete, or is the intention to just replace GPU with Accelerator? Just replacing makes the most sense to me, and in that case we should probably deprecate, and eventually remove the GPU-related methods from CloudProvider.

cluster-autoscaler/simulator/utilization/info.go
```diff
@@ -46,17 +46,16 @@ type Info struct {
...
-	if gpu.NodeHasGpu(gpuLabel, nodeInfo.Node()) {
-		gpuUtil, err := calculateUtilizationOfResource(nodeInfo, gpu.ResourceNvidiaGPU, skipDaemonSetPods, skipMirrorPods, currentTime)
+func Calculate(nodeInfo *schedulerframework.NodeInfo, skipDaemonSetPods, skipMirrorPods bool, accelerator *cloudprovider.Accelerator, currentTime time.Time) (utilInfo Info, err error) {
+	if accelerator != nil {
+		acceleratorUtil, err := calculateUtilizationOfResource(nodeInfo, accelerator.ResourceName, skipDaemonSetPods, skipMirrorPods, currentTime)
 		if err != nil {
-			klog.V(3).Infof("node %s has unready GPU", nodeInfo.Node().Name)
-			// Return 0 if GPU is unready. This will guarantee we can still scale down a node with unready GPU.
-			return Info{GpuUtil: 0, ResourceName: gpu.ResourceNvidiaGPU, Utilization: 0}, nil
+			klog.V(3).Infof("node %s has unready accelerator", nodeInfo.Node().Name)
```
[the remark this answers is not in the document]
hbostan: Also added the resource name to the log message.

cluster-autoscaler/processors/nodegroupconfig/node_group_config_processor.go
```diff
@@ -79,8 +79,8 @@ func (p *DelegatingNodeGroupConfigProcessor) GetScaleDownUtilizationThreshold(co
 	return ngConfig.ScaleDownUtilizationThreshold, nil
 }
 
-// GetScaleDownGpuUtilizationThreshold returns ScaleDownGpuUtilizationThreshold value that should be used for a given NodeGroup.
-func (p *DelegatingNodeGroupConfigProcessor) GetScaleDownGpuUtilizationThreshold(context *context.AutoscalingContext, nodeGroup cloudprovider.NodeGroup) (float64, error) {
+// GetScaleDownAcceleratorUtilizationThreshold returns the accelerator utilization threshold value that should be used for a given NodeGroup
+func (p *DelegatingNodeGroupConfigProcessor) GetScaleDownAcceleratorUtilizationThreshold(context *context.AutoscalingContext, nodeGroup cloudprovider.NodeGroup) (float64, error) {
```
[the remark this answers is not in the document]
hbostan: Renamed the option `ScaleDownGpuUtilizationThreshold` to `ScaleDownAcceleratorUtilizationThreshold` and the flag `scale-down-gpu-utilization-threshold` to `scale-down-accelerator-utilization-threshold`. Also generated proto files with variables using these new names.

hbostan:
I have updated the description of #5448 to reflect the changes you mentioned as planned. I think we don't want to expand the scope of this pr too much and I think making a separate pr for those changes will be clearer about what has changed.

towca:
Sorry to go back on my previous comment, I didn’t think about backwards compatibility properly.

We definitely can’t just get rid of the existing GPU flag, because that would break existing users who set it. At minimum we’d have to introduce a new flag, deprecate the old one, and make them both set the same option. And the situation here is worse, since as you can see some cloud providers have built higher level features tied to the flag (e.g. the AWS annotation) . We’d have to deprecate those as well, probably going through the [CMR mechanism](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/POLICY.md#cloudprovider-maintenance-requirements) - we definitely shouldn’t just change it in a breaking way.

All this work doesn’t seem worth it for just introducing the “Accelerator” name. AFAIU, functionally "Accelerators" are just extending the "GPU" concept to handle different label keys (than just the one provided by `CloudProvider.GPULabel()`), and resource names (than just the hard-coded NVIDIA resource). In all other ways, we’d want them to work just like GPUs work today, right? And AFAIK it’s not a concept that already exists in k8s, right (GPUs seem to fall under “custom resources” in k8s)? If so, maybe we could do the same thing without introducing the “Accelerator” concept at all, just building on “GPU” instead? We could change `Accelerator` to something like `GpuConfig`, and `GetNodeAccelerator` to `GetNodeGpuConfig` (and probably get rid of `GetAvailableAccelerators` - unless you’re planning to extend the validation in `GetGpuTypeForMetrics`, `GetNodeGpuConfig`+`GetAvailableGPUTypes` seem enough). Then Cluster Autoscaler’s GPU handling would cover all GPU-like devices (=”Accelerators”), without getting rid of the “GPU” name and introducing backwards compatibility considerations.

Another solution I see is extending Cluster Autoscaler’s Custom Resource support (which is in the backlog anyway) to the point when we know what CloudProvider methods are needed, and then migrating GPU-related logic to generic Custom Resource logic (that would cover all “Accelerators”). But that would be a much bigger effort, and would require some upfront design work.

cluster-autoscaler/cloudprovider/cloud_provider.go
```diff
@@ -86,6 +86,13 @@ const (
 	RancherProviderName = "rancher"
 )
 
+// GpuConfig contains the label, type and the resource name for an accelerator (e.g. gpu)
```
towca, line 89: nit: I think there's one "accelerator" leftover in the comment here

towca:
/lgtm
/approve
/hold

The changes LGTM, just one more nit - feel free to unhold if you prefer not to fix.

cluster-autoscaler/cloudprovider/cloud_provider.go
hbostan: Done :) Removed accelerator from the comment and replaced it with GPU

PR #5460, opened 2023-02-01
Bump default VPA version to 0.13.0

jbartosik:
#5355

This is for the master branch, #5462 is for the release branch

Issue #5355, opened 2022-12-06
Release VPA 0.13.0


jbartosik:
K8s 1.26 will be released on 2022-12-08, release VPA

jbartosik:
#5313 didn't have approval only lgtm so it didn't merge until today.

I want to wait for e2e to pick it up and plan to cut release tomorrow

Shubham82:
Hi, @jbartosik when will we have VPA 0.13.0?

are we waiting for this PR #5275 to be merged?

jbartosik:
@Shubham82 yes I want to merge #5275. I was also busy with some other things. This week I expect to have time for the release so I expect to finish it (unless something unexpected happens).

Shubham82:
> @Shubham82 yes I want to merge #5275. I was also busy with some other things. This week I expect to have time for the release so I expect to finish it (unless something unexpected happens).

Thanks, @jbartosik for the information.

jbartosik:
I created [vpa-release-0.13](https://github.com/kubernetes/autoscaler/tree/vpa-release-0.13) branch

jbartosik:
I've built and tested images

jbartosik:
Release is available: https://github.com/kubernetes/autoscaler/releases/tag/vertical-pod-autoscaler-0.13.0

PR #5231, opened 2022-10-03
add example for multiple recommenders

matthyx:
#### Which component this PR applies to?
vertical-pod-autoscaler

#### What type of PR is this?
/kind documentation

#### What this PR does / why we need it:
Adds documentation for https://github.com/kubernetes/autoscaler/pull/4799

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note

```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #5462, opened 2023-02-01
Bump default VPA version to 0.13.0

jbartosik:
#5355

This is for release branch, #5460 is for master branch


Shubham82:
thanks @jbartosik for this. 
/lgtm

Shubham82:
Thanks @kgolab for looking at this.
I think yes, it should be part of vpa 0.13.0 release not the vpa 0.12.0. Please see the following link for vpa 0.12 release branch:
https://github.com/kubernetes/autoscaler/tree/vpa-release-0.12/vertical-pod-autoscaler/deploy

Also it will merge after vpa 0.12.0 branch cut. Here is PR for it: #5231 as the part of the PR #4799

PR #5472, opened 2023-02-02
Fix scaledown:nodedeletion metric calculation

domenicbozzuto:
The `scaledown:nodedeletion` is improperly tracking the actual time required for scheduling node deletions. It's using the `currentTime` as a reference point for the duration, which is set at the start of `RunOnce`.  Long duration activities like a long cloud provider refresh result in this metric being skewed and essentially tracking similar quantiles to that of `main`.

This change moves the start time inside of `StartDeletion` (similar to how `TryScaleDown` tracked the metric in 1.24), so the metric now more directly tracks the duration of `StartDeletion`

##### Before
image
The quantiles for `scaledown:nodedeletion` very closely track those of `main`

##### After
image
The quantiles for `scaledown:nodedeletion` are no longer skewed alongside `main` 

#### What type of PR is this?

/kind bug

Fixes an issue where the `scaledown:nodedeletion` metric was using the wrong frame of reference for the starting time.

```release-note
NONE
```

```docs
```

### DIFF
--- a/cluster-autoscaler/core/podlistprocessor/currently_drained_nodes_test.go
+++ b/cluster-autoscaler/core/podlistprocessor/currently_drained_nodes_test.go
@@ -269,7 +269,7 @@ type mockActuator struct {
 	status *mockActuationStatus
 }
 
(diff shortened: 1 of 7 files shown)


x13n:
Thanks for the PR, good catch!

/lgtm
/approve

PR #5475, opened 2023-02-03
Update default versions in recommender high and low deployments

jbartosik:
Missed in #5462
#5355

### DIFF
--- a/vertical-pod-autoscaler/README.md
+++ b/vertical-pod-autoscaler/README.md
@@ -48,12 +48,13 @@ procedure described below.
 
 # Installation
 
-The current default version is Vertical Pod Autoscaler 0.12.0
+The current default version is Vertical Pod Autoscaler 0.13.0
 
 ### Compatibility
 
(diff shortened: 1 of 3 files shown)


vertical-pod-autoscaler/README.md
```diff
@@ -54,6 +54,7 @@ The current default version is Vertical Pod Autoscaler 0.12.0
 
 | VPA version     | Kubernetes version |
 |-----------------|--------------------|
+| 0.13            | 1.25+              |
```
kgolab, line 57: If you're changing documentation as well - shouldn't the text above ("current default version") be also changed?
But maybe doc change should be a separate thing (e.g. part of Release Notes PR)?

kgolab:
Thanks!

/lgtm

vertical-pod-autoscaler/README.md
jbartosik: Done

PR #5480, opened 2023-02-07
fix: issue with apiVersion not being set  for psp

### LINKED ISSUE #5364: Helm Chart - apiVersion not set - k8s version 1.10+

**What version of the component are you using?**: Helm 9.21.1 cluster-autoscaler

Component version:

**What k8s version are you using (`kubectl version`)?**: 1.22

**What did you expect to happen?**: Upgrade 9.21.0 -> 9.21.1 to work

**What happened instead?**:
Helm upgrade fails: 

Error: error validating "": error validating data: apiVersion not set

Helm upgrade from 9.21.0 to 9.21.1 on a k8s version 1.10+

There seems so be a problem in the if/else construct, it seems the 2nd if is nested, so it never checks for >1.21-0 and apiVersion is not set.
**charts/cluster-autoscaler/templates/_helpers.tpl** -> 
```{{/*
Return the appropriate apiVersion for podsecuritypolicy.
*/}}
{{- define "podsecuritypolicy.apiVersion" -}}
{{- $kubeTargetVersion := default .Capabilities.KubeVersion.GitVersion .Values.kubeTargetVersionOverride }}
{{- if semverCompare "<1.10-0" $kubeTargetVersion -}}
{{- print "extensions/v1beta1" -}}
{{- if semverCompare ">1.21-0" $kubeTargetVersion -}}
{{- print "policy/v1" -}}
{{- else -}}
{{- print "policy/v1beta1" -}}
{{- end -}}
{{- end -}}
{{- end -}}```

### PR DESCRIPTION
**Pull request by ViktorCollin**
#### What type of PR is this?
/kind bug

Make sure to use `apiVersion: policy/v1beta1` for PodSecurityPolicy when kubernetes version is above 1.10. This fix will work (with a deprecation warning) for cluster version below 1.25 when psp is removed.

Fixes #5364

NONE

NONE

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -11,4 +11,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.23.1
+version: 9.23.2
--- a/charts/cluster-autoscaler/templates/_helpers.tpl
+++ b/charts/cluster-autoscaler/templates/_helpers.tpl
@@ -70,8 +70,6 @@ Return the appropriate apiVersion for podsecuritypolicy.
 {{- $kubeTargetVersion := default .Capabilities.KubeVersion.GitVersion .Values.kubeTargetVersionOverride }}
 {{- if semverCompare "<1.10-0" $kubeTargetVersion -}}
 {{- print "extensions/v1beta1" -}}
-{{- else if semverCompare ">1.21-0" $kubeTargetVersion -}}
-{{- print "policy/v1" -}}
 {{- else -}}
 {{- print "policy/v1beta1" -}}
 {{- end -}}

### CODE REVIEW

**Comment by ikarlashov** [comment]
Well, there's no need to delete a valid block I believe. There's a simple typo in the construction:
```
{{/*
Return the appropriate apiVersion for podsecuritypolicy.
*/}}
{{- define "podsecuritypolicy.apiVersion" -}}
{{- $kubeTargetVersion := default .Capabilities.KubeVersion.GitVersion .Values.kubeTargetVersionOverride }}
{{- if semverCompare "<1.10-0" $kubeTargetVersion -}}
{{- print "extensions/v1beta1" -}}
{{- **else if** semverCompare ">1.21-0" $kubeTargetVersion -}}
{{- print "policy/v1" -}}
{{- else -}}
{{- print "policy/v1beta1" -}}
{{- end -}}
{{- end -}}
```

ViktorCollin:
> Well, there's no need to delete a valid block I believe. There's a simple typo in the construction, where _else if_ is required instead of extra _if_ block. This works fine:
> 
> ```
> {{/*
> Return the appropriate apiVersion for podsecuritypolicy.
> */}}
> {{- define "podsecuritypolicy.apiVersion" -}}
> {{- $kubeTargetVersion := default .Capabilities.KubeVersion.GitVersion .Values.kubeTargetVersionOverride }}
> {{- if semverCompare "<1.10-0" $kubeTargetVersion -}}
> {{- print "extensions/v1beta1" -}}
> {{- else if semverCompare ">1.21-0" $kubeTargetVersion -}}
> {{- print "policy/v1" -}}
> {{- else -}}
> {{- print "policy/v1beta1" -}}
> {{- end -}}
> {{- end -}}
> ```

It is not only that it should have been and if else statement is. The bigger problem is that psp has never graduated to GA, it went from beta to removed directly. That means that the `policy/v1` is invalid for psp and shall therefore be removed https://v1-24.docs.kubernetes.io/docs/reference/kubernetes-api/policy-resources/pod-security-policy-v1beta1/

gjtempleton:
Thanks for this, can you please squash the commits down to a single commit, happy to get this merged then.

ViktorCollin:
@gjtempleton commits are now squashed

gjtempleton:
Thanks!
/lgtm
/approve

Issue #5455, opened 2023-01-31
Go version in Admission Controller 0.9.2


coltonfreeman26:
Good day all,
I am a member of the Iron Bank team on Platform One and we are consuming the vpa-admission-controller:0.9.2 image from k8s.gcr.io
Our scan tools have picked up quite a few findings in regards to the version of Go currently being used (1.14.3)
We are using the VPA sweet and have these findings for all of them. Recommender, Updated and Admission Controller. 
Do you know if there is any plan to upgrade the Go version for these products? 
I did reach out to the security team via email and they pointed me back to the public forum.
Please let me know if you have any questions.

coltonfreeman26:
So my apologies,
Our Renovate tool picked up a newer version 0.12.0 which resolved quite a few of the findings by updating to go 1.19
Do you know if there is any plan to update to the latest 1.19?

coltonfreeman26:
Good morning @jbartosik 
I can pull the latest 0.13.0 and look at it. Also once our automation tool picks it up and it runs through our scanners I can comment here again with what all they are still catching. Thank you for this.

coltonfreeman26:
Hello @jbartosik 
I just pushed it through our pipelines and the findings are still present. There are a total of 8 (7 of which are related to go 1.19.3 and prior).
Do you work on the autoscaler team as well or should I open a new issue to see about that image getting bumped?

jbartosik:
I can update go and go-restful

Is the list of vulnerabilities available in some public place?

coltonfreeman26:
If this would also update autoscaler that would be awesome. I can post the CVE' here if you would like.

jbartosik:
> If this would also update autoscaler

Do you mean Cluster Autoscaler? I don't maintain CA, only VPA (Vertical Pod Autoscaler).

>  I can post the CVE' here if you would like.

Please do.

coltonfreeman26:
https://nvd.nist.gov/vuln/detail/CVE-2022-27664
https://nvd.nist.gov/vuln/detail/CVE-2022-2879
https://nvd.nist.gov/vuln/detail/CVE-2022-2880
https://nvd.nist.gov/vuln/detail/CVE-2022-32190
https://nvd.nist.gov/vuln/detail/CVE-2022-41715
https://github.com/emicklei/go-restful/issues/497
These are the big ones that we have found.

voelzmo:
This indeed doesn't look like the vpa components were built with the most recent go version:
```
[...]

PR #5484, opened 2023-02-07
Update go used to build VPA to 1.19.5

jbartosik:
/kind cleanup

For #5455

### DIFF
--- a/vertical-pod-autoscaler/builder/Dockerfile
+++ b/vertical-pod-autoscaler/builder/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.19
+FROM golang:1.19.5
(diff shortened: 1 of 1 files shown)


voelzmo:
If we're going to specify the builder images like this, should we include a step to ensure we're using the latest go image version in https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/RELEASE.md, regardless of why we're creating a new release?

jbartosik:
> If we're going to specify the builder images like this, should we include a step to ensure we're using the latest go image version in https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/RELEASE.md, regardless of why we're creating a new release?

Yes, this sounds like a good idea.

mwielgus (approved):
/lgtm
/approve

PR #5488, opened 2023-02-08
Stop scale-down from considering registered upcoming nodes as candidates

towca:
#### What type of PR is this?

/kind bug

Stop scale-down from considering registered upcoming nodes as candidates.

```release-note
Fixed a bug where Cluster Autoscaler would sometimes delete recently created nodes while they were still booting up.
```

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -916,12 +916,14 @@ func (csr *ClusterStateRegistry) GetIncorrectNodeGroupSize(nodeGroupName string)
 }
 
 // GetUpcomingNodes returns how many new nodes will be added shortly to the node groups or should become ready soon.
-// The function may overestimate the number of nodes.
-func (csr *ClusterStateRegistry) GetUpcomingNodes() map[string]int {
+// The function may overestimate the number of nodes. The second return value contains the names of upcoming nodes
+// that are already registered in the cluster.
+func (csr *ClusterStateRegistry) GetUpcomingNodes() (upcomingCounts map[string]int, registeredNodeNames map[string][]string) {
 	csr.Lock()
 	defer csr.Unlock()
 
-	result := make(map[string]int)
+	upcomingCounts = map[string]int{}
+	registeredNodeNames = map[string][]string{}
 	for _, nodeGroup := range csr.cloudProvider.NodeGroups() {
 		id := nodeGroup.Id()
 		readiness := csr.perNodeGroupReadiness[id]
@@ -932,9 +934,14 @@ func (csr *ClusterStateRegistry) GetUpcomingNodes() map[string]int {
 			// Negative value is unlikely but theoretically possible.
 			continue
 		}
-		result[id] = newNodes
+		upcomingCounts[id] = newNodes
+		// newNodes should include instances that have registered with k8s but aren't ready yet, instances that came up on the cloud provider side
+		// but haven't registered with k8s yet, and instances that haven't even come up on the cloud provider side yet (but are reflected in the target
+		// size). The first category is categorized as NotStarted in readiness, the other two aren't registered with k8s, so they shouldn't be
+		// included.
+		registeredNodeNames[id] = readiness.NotStarted
 	}
-	return result
+	return upcomingCounts, registeredNodeNames
 }
 
 // getCloudProviderNodeInstances returns map keyed on node group id where value is list of node instances
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -354,7 +354,7 @@ func (csr *ClusterStateRegistry) IsClusterHealthy() bool {
 	csr.Lock()
 	defer csr.Unlock()
 
-	totalUnready := csr.totalReadiness.Unready
+	totalUnready := len(csr.totalReadiness.Unready)
 
 	if totalUnready > csr.config.OkTotalUnreadyCount &&
 		float64(totalUnready) > csr.config.MaxTotalUnreadyPercentage/100.0*float64(len(csr.nodes)) {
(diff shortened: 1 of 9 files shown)


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -467,6 +467,20 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
 			return errors.ToAutoscalerError(errors.InternalError, err)
 		}
 	}
+	// Nodes that are registered but not yet ready are classified as NotStarted by CSR. Above, we inject "upcoming" replacement
+	// nodes for them that are faked to appear ready, so that we can pack unschedulable pods on them and not trigger another scale-up.
+	// The initial nodes have to be filtered out of the all nodes list so that scale-down can't consider them as candidates. Otherwise, with
+	// aggressive scale-down settings, we could be removing the nodes before they have a chance to first become ready (the duration
+	// of which should be unrelated to the scale-down settings).
+	allNodes = subtractNodesByName(allNodes, a.clusterStateRegistry.GetClusterReadiness().NotStarted)
```
MaciekPytel, line 475: What you want to filter here is in my opinion upcoming nodes (since those are the nodes that would have templateNode added for them). I think that by having 2 different paths for providing the same information we're setting ourselves up for future bugs when those get out of sync.

Also - I don't think we should expose Readiness outside of CSR. I know it's technically public today, but so far it was only for the purpose of pushing it to metrics. My main reason is that readiness is a huge mess, where some values are additive and some are not (ResourceUnready is a subset of Unready, but other stuff can be added etc) and we keep changing semantics completely when we're fixing various bugs.

As an example - we used to have LongNotStarted for a long time, that would change the logic here.

towca: Upcoming as in "fake replacements" are not added to `allNodes` in the first place. Upcoming as in "the nodes that are being replaced" consist of Unregistered (for which there are no names so nothing to filter), and NotStarted - which have names and should be filtered out of `allNodes`. So it's only "registered upcoming" nodes that should be filtered out.

Good points about having one path and not leaking the details of CSR. I changed CSR so that it is the one determining which nodes should be filtered without leaking any details (apart from the lists being public, but we've had that with the ints as well), in the same method that determines upcoming node counts.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -919,25 +916,30 @@ func (csr *ClusterStateRegistry) GetIncorrectNodeGroupSize(nodeGroupName string)
...
+		newNodes := ar.CurrentTarget - (len(readiness.Ready) + len(readiness.Unready) + len(readiness.LongUnregistered))
 		if newNodes <= 0 {
 			// Negative value is unlikely but theoretically possible.
 			continue
 		}
-		result[id] = newNodes
+		upcomingCounts[id] = newNodes
+		// newNodes should be roughly equal to readiness.NotStarted + readiness.Unregistered. NotStarted are nodes that are
+		// already registered in the cluster but not yet ready, Unregistered are nodes that aren't registered in the cluster yet.
+		registeredNodeNames[id] = readiness.NotStarted
```
MaciekPytel, line 940: nit: I would personally return all known instance names (ie. including unregistered) and later filter out from snapshot the ones that are there. Or maybe return both registered and unregistered lists as separate values. 
The interface where we return one, but not the other seems strange to me. But, of course, we don't need unregistered node names for anything in this codepath and my aesthetic feelings don't feel like strong enough justification, so feel free to ignore this comment.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -467,6 +471,22 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) errors.AutoscalerError
...
+	for _, ngRegisteredUpcoming := range registeredUpcoming {
+		allRegisteredUpcoming = append(allRegisteredUpcoming, ngRegisteredUpcoming...)
+	}
+	allNodes = subtractNodesByName(allNodes, allRegisteredUpcoming)
+	// Remove the nodes from the snapshot as well so that the state is consistent.
+	for _, notStartedNodeName := range allRegisteredUpcoming {
+		err := a.ClusterSnapshot.RemoveNode(notStartedNodeName)
+		if err != nil {
+			klog.Errorf("Failed to remove NotStarted node %s from cluster snapshot: %v", notStartedNodeName, err)
+			return errors.ToAutoscalerError(errors.InternalError, err)
```
MaciekPytel, line 487: Should we really abort the loop on error here? Looking at snapshot implementation the only error we can get is NotFound. That should never happen [1], but if it did for some reason is it really fatal? Previously we would just continue the loop.
Maybe we can just log error for notFound and fail on other errors (actually failing to remove the node would sound pretty scary)? Or check that the node is in the snapshot before trying to remove it and log error if it's not?

[1] At least I don't see how it could happen, since we use the same list of nodes to initialize CSR and snapshot and there aren't any processor calls in the meantime where I'd expect anyone to do any shenanigans with snapshot (there is TemplateNodeInfoProvider call, but modifying snapshot there strikes me as particularly evil; CustomResourcesProcessor happens before CSR update or building the snapshot so any changes should be consistent).

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -1273,6 +1276,167 @@ func TestStaticAutoscalerInstanceCreationErrors(t *testing.T) {
...
+	snapshotNodeNames := map[string]bool{}
+	for _, nodeInfo := range nodeInfos {
+		snapshotNodeNames[nodeInfo.Node().Name] = true
+	}
+	for nodeName := range nodeNames {
+		assert.Contains(t, snapshotNodeNames, nodeName)
+	}
+}
+
+func TestStaticAutoscalerUpcomingScaleDownCandidates(t *testing.T) {
```
MaciekPytel, line 1336: nit, no action expected: This is pretty well written for what it is, but I'm not very fond of those big tests of RunOnce. I'll let this one in, but I'd prefer a set of smaller tests that don't actually call the entire RunOnce. In this case snapshot filtering could be a separate function that could be tested in isolation and if we also wanted to test scale-down logic to make sure it doesn't add nodes that are not in snapshot to candidates (how could it though?), we could test this in scale-down.

cluster-autoscaler/clusterstate/clusterstate.go
towca: I agree it's not ideal, but I'm afraid returning them both in the same object could be confusing, since they'd have a different format depending on whether they're registered or not (unregistered nodes get the instance id as a placeholder name, which is probably actually an URL with more info, not just the name) - and this wouldn't be obvious to the caller. Returning both in separate objects would solve that and would be cleaner, I just didn't want to do that just to discard at the call site. We can always add the second object if needed in the future.

cluster-autoscaler/core/static_autoscaler_test.go
towca: Full agreement here, I just tried to match the convention in this file/pkg.

cluster-autoscaler/core/static_autoscaler.go
towca: Agreed, all this part cares about is that the node is not in the snapshot - so it seems reasonable to just log and continue on NotFound. I had to export the error and rename a bunch of imports for this though.

PR #5496, opened 2023-02-10
Fix MEMBER_ALREADY_EXISTS errors in GCE cloudprovider

x13n:
The CreateInstances is passed a list of provider ids, which is not the same as the list of names: provider id includes zone and project information as well. This can cause scale ups in large clusters to occassionally fail.

#### What type of PR is this?

/kind bug

Fixes a bug affecting scale up performance.

Fixes #

```release-note
Fixed a bug in GCE cloud provider causing large scale ups to occasionally fail.
```

```docs

```


MaciekPytel:
/lgtm
/approve
Nice catch

PR #5497, opened 2023-02-10
Track PDBRemainingDisruptions in AutoscalingContext

BigDarkClown:
#### What type of PR is this?

/kind cleanup

This PR adds PDBRemainingDisruptions struct to AutoscalingContext, enabling remaining PDB tracking throughout the CA.

```release-note
NONE
```

### DIFF
--- a/cluster-autoscaler/core/scaledown/planner/planner.go
+++ b/cluster-autoscaler/core/scaledown/planner/planner.go
@@ -26,7 +26,6 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/context"
 	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown"
 	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown/eligibility"
-	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown/pdb"
 	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown/resource"
 	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown/unneeded"
 	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown/unremovable"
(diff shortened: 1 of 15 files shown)


cluster-autoscaler/main.go
```diff
@@ -390,13 +391,16 @@ func buildAutoscaler(debuggingSnapshotter debuggingsnapshot.DebuggingSnapshotter
 		return nil, err
 	}
 
+	remainingPdbTracker := pdb.NewBasicRemainingPdbTracker()
```
towca, line 394: Any reason not to have it initialized by default in `initializeDefaultOptions` instead? Seems like it's needed in any cloud provider/CA implementation.

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -96,19 +94,13 @@ func New(context *context.AutoscalingContext, processors *processors.Autoscaling
...
 // up-to-date information about the cluster.
 // Planner will evaluate scaleDownCandidates in the order provided here.
-func (p *Planner) UpdateClusterState(podDestinations, scaleDownCandidates []*apiv1.Node, as scaledown.ActuationStatus, pdbs []*policyv1.PodDisruptionBudget, currentTime time.Time) errors.AutoscalerError {
+func (p *Planner) UpdateClusterState(podDestinations, scaleDownCandidates []*apiv1.Node, as scaledown.ActuationStatus, currentTime time.Time) errors.AutoscalerError {
 	p.latestUpdate = currentTime
 	p.actuationStatus = as
-	p.remainingPdbTracker = pdb.NewBasicRemainingPdbTracker()
-	err := p.remainingPdbTracker.AddPdbs(pdbs)
-	if err != nil {
-		p.CleanUpUnneededNodes()
```
towca, line 105: Shouldn't we now call this in `static_autoscaler.go` when the same call fails? Or does that already happen because we're failing the whole loop?

cluster-autoscaler/main.go
BigDarkClown: Done

cluster-autoscaler/core/scaledown/planner/planner.go
BigDarkClown: I think this code was actually dead. The only way for this check to trigger would be for the PDB Lister to return PDBs with ill configured selectors, which should not be possible.

PR #5507, opened 2023-02-14
feat: check only controller ref to decide if a pod is replicated

vadasambar:
### This PR is a follow-up to https://github.com/kubernetes/autoscaler/pull/5419#issuecomment-1426195822

~**This PR is still WIP. I will mention the reviewers here once I am done. Although it might not be in the best shape as a WIP PR but if the reviewers have any feedback, I would love to have it. :pray:**~

#### What type of PR is this?

/kind feature

- scale down [is blocked](https://github.com/kubernetes/autoscaler/blob/73d02f1d641a38314e93077d375d80e03468ad60/cluster-autoscaler/utils/drain/drain.go#L210-L211) on the pods where the controller reference is not understood by the CA ([CA as of now only understands controller refs for Daemonsets, ReplicaSets, Jobs, ReplicationControllers, Deployments and StatefulSets](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/utils/drain/drain.go#L124-L203)). This is because CA treats pods with non-understood controller references as non-replicated and hence deems scale-down as not-safe. 
- this PR introduces a change where all pods with ownerReference `controller: true` are treated as replicated and scale-down is not blocked

Fixes https://github.com/kubernetes/autoscaler/issues/5387

```release-note
Added: New flag `--allow-scale-down-on-custom-controller-owned-pods`. If this flag is set to true cluster-autoscaler doesn't block node scale-down if a pod owned by a custom controller is running on the node.
```

TBD
```docs

```

### DIFF
--- a/cluster-autoscaler/FAQ.md
+++ b/cluster-autoscaler/FAQ.md
@@ -785,6 +785,7 @@ The following startup parameters are supported for cluster autoscaler:
 | `aws-use-static-instance-list` | Should CA fetch instance types in runtime or use a static list. AWS only | false
 | `skip-nodes-with-system-pods` | If true cluster autoscaler will never delete nodes with pods from kube-system (except for DaemonSet or mirror pods) | true
 | `skip-nodes-with-local-storage`| If true cluster autoscaler will never delete nodes with pods with local storage, e.g. EmptyDir or HostPath | true
+| `skip-nodes-with-custom-controller-pods` | If true cluster autoscaler will never delete nodes with pods owned by custom controllers | true
 | `min-replica-count` | Minimum number or replicas that a replica set or replication controller should have to allow their pods deletion in scale down | 0
 | `daemonset-eviction-for-empty-nodes` | Whether DaemonSet pods will be gracefully terminated from empty nodes | false
 | `daemonset-eviction-for-occupied-nodes` | Whether DaemonSet pods will be gracefully terminated from non-empty nodes | true
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -216,6 +216,7 @@ var (
 	maxNodeGroupBinpackingDuration     = flag.Duration("max-nodegroup-binpacking-duration", 10*time.Second, "Maximum time that will be spent in binpacking simulation for each NodeGroup.")
 	skipNodesWithSystemPods            = flag.Bool("skip-nodes-with-system-pods", true, "If true cluster autoscaler will never delete nodes with pods from kube-system (except for DaemonSet or mirror pods)")
 	skipNodesWithLocalStorage          = flag.Bool("skip-nodes-with-local-storage", true, "If true cluster autoscaler will never delete nodes with pods with local storage, e.g. EmptyDir or HostPath")
+	skipNodesWithCustomControllerPods  = flag.Bool("skip-nodes-with-custom-controller-pods", true, "If true cluster autoscaler will never delete nodes with pods owned by custom controllers")
 	minReplicaCount                    = flag.Int("min-replica-count", 0, "Minimum number or replicas that a replica set or replication controller should have to allow their pods deletion in scale down")
 	nodeDeleteDelayAfterTaint          = flag.Duration("node-delete-delay-after-taint", 5*time.Second, "How long to wait before deleting a node after tainting it")
 	scaleDownSimulationTimeout         = flag.Duration("scale-down-simulation-timeout", 30*time.Second, "How long should we run scale down simulation.")
@@ -328,6 +329,7 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 		NodeDeleteDelayAfterTaint:          *nodeDeleteDelayAfterTaint,
 		ScaleDownSimulationTimeout:         *scaleDownSimulationTimeout,
 		ParallelDrain:                      *parallelDrain,
+		SkipNodesWithCustomControllerPods:  *skipNodesWithCustomControllerPods,
 		NodeGroupSetRatios: config.NodeGroupDifferenceRatios{
 			MaxCapacityMemoryDifferenceRatio: *maxCapacityMemoryDifferenceRatio,
 			MaxAllocatableDifferenceRatio:    *maxAllocatableDifferenceRatio,
--- a/cluster-autoscaler/simulator/drain_test.go
+++ b/cluster-autoscaler/simulator/drain_test.go
@@ -42,9 +42,10 @@ func TestGetPodsToMove(t *testing.T) {
 		},
 	}
 	deleteOptions := NodeDeleteOptions{
-		SkipNodesWithSystemPods:   true,
-		SkipNodesWithLocalStorage: true,
-		MinReplicaCount:           0,
+		SkipNodesWithSystemPods:           true,
+		SkipNodesWithLocalStorage:         true,
+		MinReplicaCount:                   0,
+		SkipNodesWithCustomControllerPods: true,
 	}
 	_, _, blockingPod, err := GetPodsToMove(schedulerframework.NewNodeInfo(pod1), deleteOptions, nil, nil, testTime)
 	assert.Error(t, err)
--- a/cluster-autoscaler/utils/drain/drain.go
+++ b/cluster-autoscaler/utils/drain/drain.go
@@ -78,13 +78,13 @@ func GetPodsForDeletionOnNodeDrain(
 	pdbs []*policyv1.PodDisruptionBudget,
 	skipNodesWithSystemPods bool,
 	skipNodesWithLocalStorage bool,
+	skipNodesWithCustomControllerPods bool,
 	listers kube_util.ListerRegistry,
 	minReplica int32,
 	currentTime time.Time) (pods []*apiv1.Pod, daemonSetPods []*apiv1.Pod, blockingPod *BlockingPod, err error) {
 
 	pods = []*apiv1.Pod{}
 	daemonSetPods = []*apiv1.Pod{}
-	checkReferences := listers != nil
 	// filter kube-system PDBs to avoid doing it for every kube-system pod
 	kubeSystemPDBs := make([]*policyv1.PodDisruptionBudget, 0)
 	for _, pdb := range pdbs {
@@ -93,6 +93,7 @@ func GetPodsForDeletionOnNodeDrain(
 		}
 	}
 
+	isDaemonSetPod := false
 	for _, pod := range podList {
 		if pod_util.IsMirrorPod(pod) {
 			continue
@@ -106,101 +107,25 @@ func GetPodsForDeletionOnNodeDrain(
 			continue
 		}
 
-		isDaemonSetPod := false
 		replicated := false
 		safeToEvict := hasSafeToEvictAnnotation(pod)
 		terminal := isPodTerminal(pod)
 
-		controllerRef := ControllerRef(pod)
-		refKind := ""
-		if controllerRef != nil {
-			refKind = controllerRef.Kind
+		if skipNodesWithCustomControllerPods {
+			// TODO(vadasambar): remove this when we get rid of skipNodesWithCustomControllerPods
+			replicated, isDaemonSetPod, blockingPod, err = legacyCheckForReplicatedPods(listers, pod, minReplica)
+			if err != nil {
+				return []*apiv1.Pod{}, []*apiv1.Pod{}, blockingPod, err
+			}
+		} else {
+			if ControllerRef(pod) != nil {
+				replicated = true
+			}
+			if pod_util.IsDaemonSetPod(pod) {
+				isDaemonSetPod = true
+			}
 		}
 
-		// For now, owner controller must be in the same namespace as the pod
-		// so OwnerReference doesn't have its own Namespace field
-		controllerNamespace := pod.Namespace
-
-		if refKind == "ReplicationController" {
-			if checkReferences {
-				rc, err := listers.ReplicationControllerLister().ReplicationControllers(controllerNamespace).Get(controllerRef.Name)
-				// Assume a reason for an error is because the RC is either
-				// gone/missing or that the rc has too few replicas configured.
-				// TODO: replace the minReplica check with pod disruption budget.
-				if err == nil && rc != nil {
-					if rc.Spec.Replicas != nil && *rc.Spec.Replicas < minReplica {
-						return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: MinReplicasReached}, fmt.Errorf("replication controller for %s/%s has too few replicas spec: %d min: %d",
-							pod.Namespace, pod.Name, rc.Spec.Replicas, minReplica)
-					}
-					replicated = true
-				} else {
-					return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: ControllerNotFound}, fmt.Errorf("replication controller for %s/%s is not available, err: %v", pod.Namespace, pod.Name, err)
-				}
-			} else {
-				replicated = true
-			}
-		} else if pod_util.IsDaemonSetPod(pod) {
-			isDaemonSetPod = true
-			// don't have listener for other DaemonSet kind
-			// TODO: we should use a generic client for checking the reference.
-			if checkReferences && refKind == "DaemonSet" {
-				_, err := listers.DaemonSetLister().DaemonSets(controllerNamespace).Get(controllerRef.Name)
-				if apierrors.IsNotFound(err) {
-					return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: ControllerNotFound}, fmt.Errorf("daemonset for %s/%s is not present, err: %v", pod.Namespace, pod.Name, err)
-				} else if err != nil {
-					return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: UnexpectedError}, fmt.Errorf("error when trying to get daemonset for %s/%s , err: %v", pod.Namespace, pod.Name, err)
-				}
-			}
-		} else if refKind == "Job" {
-			if checkReferences {
-				job, err := listers.JobLister().Jobs(controllerNamespace).Get(controllerRef.Name)
-
-				// Assume the only reason for an error is because the Job is
-				// gone/missing, not for any other cause.  TODO(mml): something more
-				// sophisticated than this
-				if err == nil && job != nil {
-					replicated = true
-				} else {
-					return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: ControllerNotFound}, fmt.Errorf("job for %s/%s is not available: err: %v", pod.Namespace, pod.Name, err)
-				}
-			} else {
-				replicated = true
-			}
-		} else if refKind == "ReplicaSet" {
-			if checkReferences {
-				rs, err := listers.ReplicaSetLister().ReplicaSets(controllerNamespace).Get(controllerRef.Name)
-
-				// Assume the only reason for an error is because the RS is
-				// gone/missing, not for any other cause.  TODO(mml): something more
-				// sophisticated than this
-				if err == nil && rs != nil {
-					if rs.Spec.Replicas != nil && *rs.Spec.Replicas < minReplica {
-						return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: MinReplicasReached}, fmt.Errorf("replication controller for %s/%s has too few replicas spec: %d min: %d",
-							pod.Namespace, pod.Name, rs.Spec.Replicas, minReplica)
-					}
-					replicated = true
-				} else {
-					return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: ControllerNotFound}, fmt.Errorf("replication controller for %s/%s is not available, err: %v", pod.Namespace, pod.Name, err)
-				}
-			} else {
-				replicated = true
-			}
-		} else if refKind == "StatefulSet" {
-			if checkReferences {
-				ss, err := listers.StatefulSetLister().StatefulSets(controllerNamespace).Get(controllerRef.Name)
-
-				// Assume the only reason for an error is because the StatefulSet is
-				// gone/missing, not for any other cause.  TODO(mml): something more
-				// sophisticated than this
-				if err == nil && ss != nil {
-					replicated = true
-				} else {
-					return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: ControllerNotFound}, fmt.Errorf("statefulset for %s/%s is not available: err: %v", pod.Namespace, pod.Name, err)
-				}
-			} else {
-				replicated = true
-			}
-		}
 		if isDaemonSetPod {
 			daemonSetPods = append(daemonSetPods, pod)
 			continue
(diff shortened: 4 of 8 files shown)


cluster-autoscaler/utils/drain/drain_test.go
```diff
@@ -183,6 +183,20 @@ func TestDrain(t *testing.T) {
...
 	}
 
+	customControllerPod := &apiv1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "bar",
+			Namespace: "default",
+			// Using names like FooController is discouraged
+			// https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions
+			// vadasambar: I am using it here just because `FooController`` is easier to understand than say `FooSet`
+			OwnerReferences: GenerateOwnerReferences("Foo", "FooController", "apps/v1", ""),
```
vadasambar, line 193: Any other name suggestions are welcome.

x13n:
/assign

I would preserve the existing behavior for known controllers - perhaps behind a flag - so that there's no regression. We can eventually delete the flag after a few releases, but I'd like to avoid breaking existing users.

vadasambar:
> /assign
> 
> I would preserve the existing behavior for known controllers - perhaps behind a flag - so that there's no regression. We can eventually delete the flag after a few releases, but I'd like to avoid breaking existing users.

Makes sense to me. Let me think about this. 
Thank you for the feedback (I always get to learn something because of your feedback).

vadasambar:
I had a few hiccups around trying to get a custom CA run properly on GKE (all good now). I want to test this PR manually on a GKE cluster. The rough idea is something like this:
- [ ] add debug log to show pod with an owner reference was skipped during scale down
- [ ] create custom CA in a GKE cluster with PR image tag and `--v` flag to set log level high enough to show the debug log for owner reference
- [ ] schedule a workload with N replicas that needs a new node e.g., set pod anti-affinity so that N-1 nodes scale out to N nodes
- [ ] run a pod with owner reference
- [ ] delete the N replicas workload
- [ ] tail logs for cluster-autoscaler to make sure it skips the pod with owner reference
- [ ] if I see the owner reference log, screenshot it and paste it in the ticket PR
- [ ] change the PR from draft to review
- [ ] ask for review

x13n:
Rather than on logs, I'd just rely on behavior. In this case you could do something like this:

- Scale cluster to N nodes using N replica workload you described.
- Put static pods on N-1 nodes and one custom controlled pod on the last node.
- Delete the workload.
- Observe if the node with custom controlled pod got rescheduled.

Note: N=2 might be problematic due to system workloads, but N=3 should work. Alternatively, you can use a separate nodepool to put all non-DS system workloads there using taints/tolerations, so that they don't interfere with the test.

cluster-autoscaler/main.go
```diff
@@ -205,22 +205,23 @@ var (
...
+	maxFreeDifferenceRatio                    = flag.Float64("max-free-difference-ratio", config.DefaultMaxFreeDifferenceRatio, "Maximum difference in free resources between two similar node groups to be considered for balancing. Value is a ratio of the smaller node group's free resource.")
+	maxAllocatableDifferenceRatio             = flag.Float64("max-allocatable-difference-ratio", config.DefaultMaxAllocatableDifferenceRatio, "Maximum difference in allocatable resources between two similar node groups to be considered for balancing. Value is a ratio of the smaller node group's allocatable resource.")
+	forceDaemonSets                           = flag.Bool("force-ds", false, "Blocks scale-up of node groups too small for all suitable Daemon Sets pods.")
+	allowScaleDownOnCustomControllerOwnedPods = flag.Bool("allow-scale-down-on-custom-controller-owned-pods", false, "Don't block node scale-down if a pod owned by a custom controller is running on the node.")
```
vadasambar, line 224: Better naming suggestions are welcome (the flag name feels a little too long).

vadasambar: The default is set to `false` right now to provide backwards compatibility. This flag would be set to `true` in the future (we might remove it completely and use `true` one as the default behavior).

cluster-autoscaler/utils/drain/drain.go
```diff
@@ -106,104 +106,118 @@ func GetPodsForDeletionOnNodeDrain(
...
-				// sophisticated than this
-				if err == nil && job != nil {
-					replicated = true
-				} else {
-					return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: ControllerNotFound}, fmt.Errorf("job for %s/%s is not available: err: %v", pod.Namespace, pod.Name, err)
-				}
-			} else {
-				replicated = true
+		} else {
+			checkReferences := listers != nil
```
vadasambar, line 127: The code in the `else` part is just copy-paste of our current code. I have moved a couple of variable declarations into the else part so that it's easier to remove in the future.

cluster-autoscaler/main.go
gregth: @vadasambar Just an idea, maybe you could follow the `skip-nodes-with-*` scheme, for instance,
`skip-nodes-with-custom-controller-pods`?

vadasambar: @gregth thank you for the suggestion. I thought the new flag name `skip-nodes-with-custom-controller-pods` wouldn't tell the user about if the skipping happens during scale-up or scale-down but looks like we already use `skip-nodes-with-*` flag naming for skipping things when scaling down nodes:
https://github.com/kubernetes/autoscaler/blob/1acb6b23608c50cdec80c2846c690c82e51377d0/cluster-autoscaler/main.go#L214-L215

Making the naming consistent makes sense to me. I will change it to `skip-nodes-with-custom-controller-pods`. Thank you.

P.S.: If you have any other suggestions, I would love to know.

vadasambar:
> [Keywords](https://help.github.com/articles/closing-issues-using-keywords) which can automatically close issues and at(@) or hashtag(#) mentions are not allowed in commit messages.
> 
> **The list of commits with invalid commit messages**:
> 
>     * [18902e8](https://github.com/kubernetes/autoscaler/commits/18902e8fd48a0f789608b69f42d25d02ab140406) fix: remove `@` in `@vadasambar`

:memo:

vadasambar:
### Testing
Created a new nodepool in GKE with taints (`pool-1`)
image

Set the flag to `true`:
image

Scaled up the `pool-1` nodepool using the following workload:
```yaml
# +kubectl
apiVersion: apps/v1
kind: Deployment
metadata:
  name: node-scale-up-with-pod-anti-affinity
  namespace: default
spec:
  selector:
    matchLabels:
      app: node-scale-up-with-pod-anti-affinity
  replicas: 1
  template:
    metadata:
      labels:
        app: node-scale-up-with-pod-anti-affinity
    spec:
      nodeSelector:
        cloud.google.com/gke-nodepool: pool-1
      tolerations:
        - effect: NoSchedule
          key: test
          operator: Equal
          value: "true"
      affinity:
        podAntiAffinity:
          - labelSelector:
              matchExpressions:
              - key: app
                operator: In
                values:
                - node-scale-up-with-pod-anti-affinity
            topologyKey: "kubernetes.io/hostname"
      containers:
      - name: node-scale-up-with-pod-anti-affinity
        image: registry.k8s.io/pause:2.0
```
Deployed the test workload pod with custom controller owner ref
```
apiVersion: v1
kind: Pod
metadata:
  name: custom-controller-owned-pod
  ownerReferences:
  - apiVersion: foos
    kind: FooSet
    controller: true
    name: custom-controller-owned-pod-56fdfd787b
    uid: 1c6544a7-12e7-426c-bd2d-7ac858d18d7d 
spec:
  nodeName: gke-cluster-1-pool-1-b66d130e-9mrw # name of the node that was created to accomodate scale-up workload
  tolerations:
    - effect: NoSchedule 
      key: test  # pool-1 nodepool is tainted with this taint
      operator: Equal
      value: "true"
  containers:
  - name: nginx
    image: nginx
    imagePullPolicy: IfNotPresent
```

Removed the scale up workload to see how CA behaves. Sure enough, it removes the pod with custom controller owner reference:
image

You can see some logs like `inside new loop` and `allowScaleDownOnCustomController..`. This was to debug a problem where CA didn't skip the custom controller owned pod. Turns out I forgot to add `controller: true` in the owner reference. 

###More info around the `controller: true` problem

vadasambar:
@gregth, @feiskyer, @x13n, @BigDarkClown 
I think the PR is ready for review. Looking for your feedback/review :pray:

cluster-autoscaler/main.go
```diff
@@ -216,6 +216,7 @@ var (
 	maxNodeGroupBinpackingDuration     = flag.Duration("max-nodegroup-binpacking-duration", 10*time.Second, "Maximum time that will be spent in binpacking simulation for each NodeGroup.")
 	skipNodesWithSystemPods            = flag.Bool("skip-nodes-with-system-pods", true, "If true cluster autoscaler will never delete nodes with pods from kube-system (except for DaemonSet or mirror pods)")
 	skipNodesWithLocalStorage          = flag.Bool("skip-nodes-with-local-storage", true, "If true cluster autoscaler will never delete nodes with pods with local storage, e.g. EmptyDir or HostPath")
+	skipNodesWithCustomControllerPods  = flag.Bool("skip-nodes-with-custom-controller-pods", false, "If true cluster autoscaler will delete nodes with pods owned by custom controllers")
```
x13n, line 219: The other `skip-nodes-with-` flags mean skipping during scale down: if "skip XYZ" == true, then don't delete the node. The semantics here should be the same, right now it is inverted according to the flag description.

cluster-autoscaler/FAQ.md
```diff
@@ -792,6 +792,7 @@ The following startup parameters are supported for cluster autoscaler:
 | `cordon-node-before-terminating` | Should CA cordon nodes before terminating during downscale process | false
 | `record-duplicated-events` | Enable the autoscaler to print duplicated events within a 5 minute window. | false
 | `debugging-snapshot-enabled` | Whether the debugging snapshot of cluster autoscaler feature is enabled. | false
+| `allow-scale-down-on-custom-controller-owned-pods` | Don't block node scale-down if a pod owned by a custom controller is running on the node. | false
```
x13n, line 795: This is a different flag name to the on in main. Please update.

cluster-autoscaler/utils/drain/drain.go
```diff
@@ -106,104 +106,118 @@ func GetPodsForDeletionOnNodeDrain(
...
-					if rc.Spec.Replicas != nil && *rc.Spec.Replicas < minReplica {
-						return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: MinReplicasReached}, fmt.Errorf("replication controller for %s/%s has too few replicas spec: %d min: %d",
-							pod.Namespace, pod.Name, rc.Spec.Replicas, minReplica)
-					}
-					replicated = true
-				} else {
-					return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: ControllerNotFound}, fmt.Errorf("replication controller for %s/%s is not available, err: %v", pod.Namespace, pod.Name, err)
-				}
-			} else {
+		// vadasambar: The else part of this code hasn't been abstracted into
```
x13n, line 116: I disagree with this comment. The fact that something is going to be eventually deleted isn't a good argument for making an extra long function even longer. Firstly, "near future" here means at least months, so it will be around for quite some time. Secondly, eventual removal might actually be cleaner if we can delete the whole function at once.

cluster-autoscaler/utils/drain/drain_test.go
```diff
@@ -624,50 +678,78 @@ func TestDrain(t *testing.T) {
 		},
 	}
 
+	// run all tests for skipNodesWithCustomControllerPods=false
+	// vadasambar: TODO: remove this when we get rid of skipNodesWithCustomControllerPods
```
x13n, line 682: nit: Let's format the TODO as `// TODO(vadasambar)` to be consistent with the rest of the codebase. As far as I can tell, we don't use the `// <github handle>: TODO:` pattern anywhere.

cluster-autoscaler/utils/drain/drain_test.go
```diff
@@ -33,6 +34,31 @@ import (
...
+	// returned pods are appended to expectPods above
+	expectPodsFn        func() []*apiv1.Pod
+	expectDaemonSetPods []*apiv1.Pod
+	expectBlockingPod   *BlockingPod
+	// expectBlockingPodFn provides more flexibility than expectBlockingPod
+	// result from this function is used only if expectBlockingPod is nil
+	expectBlockingPodFn func() *BlockingPod
+}
+
+var skipNodesWithCustomControllerPods bool = false
```
x13n, line 60: I get that this helps reduce # of lines of code, but making unit tests rely on injected functions that use a global variable is tricky. I think it would be best to keep unit tests simple (even if overly verbose), to make them easier to reason about. Otherwise we would have to have tests of the tests to check whether they work correctly.

Also, introducing a global prevents us from ever running the test cases in parallel.

x13n (changes requested):

cluster-autoscaler/main.go
vadasambar: Sorry, you are right. I feel stupid for not noticing that. I will change the flag name.

cluster-autoscaler/FAQ.md
vadasambar: Looks like I missed this. Thank you for noticing.

cluster-autoscaler/utils/drain/drain.go
vadasambar: I was hesitant about putting extra time and effort to abstract it out since we are going to delete it anyway but I see your point. Will abstract it out.

cluster-autoscaler/utils/drain/drain_test.go
vadasambar: Makes sense. I will update this.

cluster-autoscaler/utils/drain/drain_test.go
vadasambar: I see your point about not making tests complicated. Let me think about this. 

> Also, introducing a global prevents us from ever running the test cases in parallel.

I am a little on the fence about this but I am not too happy about using global variable either. Let me think more about this.

cluster-autoscaler/main.go
vadasambar: I am thinking of `--scale-down-node-with-custom-controller-pods`. Any other suggestions are welcome!

cluster-autoscaler/main.go
```diff
@@ -208,22 +208,23 @@ var (
...
+	recordDuplicatedEvents                 = flag.Bool("record-duplicated-events", false, "enable duplication of similar events within a 5 minute window.")
+	maxNodesPerScaleUp                     = flag.Int("max-nodes-per-scaleup", 1000, "Max nodes added in a single scale-up. This is intended strictly for optimizing CA algorithm latency and not a tool to rate-limit scale-up throughput.")
+	maxNodeGroupBinpackingDuration         = flag.Duration("max-nodegroup-binpacking-duration", 10*time.Second, "Maximum time that will be spent in binpacking simulation for each NodeGroup.")
+	skipNodesWithSystemPods                = flag.Bool("skip-nodes-with-system-pods", true, "If true cluster autoscaler will never delete nodes with pods from kube-system (except for DaemonSet or mirror pods)")
+	skipNodesWithLocalStorage              = flag.Bool("skip-nodes-with-local-storage", true, "If true cluster autoscaler will never delete nodes with pods with local storage, e.g. EmptyDir or HostPath")
+	scaleDownNodesWithCustomControllerPods = flag.Bool("scale-down-nodes-with-custom-controller-pods", false, "If true cluster autoscaler will delete nodes with pods owned by custom controllers")
```
vadasambar, line 219: This is the new flag. Looks like adding the new flag added extra space in every flag definition.

cluster-autoscaler/main.go
```diff
@@ -253,81 +254,82 @@ func createAutoscalingOptions() config.AutoscalingOptions {
...
+		CordonNodeBeforeTerminate:              *cordonNodeBeforeTerminate,
+		DaemonSetEvictionForEmptyNodes:         *daemonSetEvictionForEmptyNodes,
+		DaemonSetEvictionForOccupiedNodes:      *daemonSetEvictionForOccupiedNodes,
+		UserAgent:                              *userAgent,
+		InitialNodeGroupBackoffDuration:        *initialNodeGroupBackoffDuration,
+		MaxNodeGroupBackoffDuration:            *maxNodeGroupBackoffDuration,
+		NodeGroupBackoffResetTimeout:           *nodeGroupBackoffResetTimeout,
+		MaxScaleDownParallelism:                *maxScaleDownParallelismFlag,
+		MaxDrainParallelism:                    *maxDrainParallelismFlag,
+		GceExpanderEphemeralStorageSupport:     *gceExpanderEphemeralStorageSupport,
```
vadasambar, line 321: Same here. Extra space was added because I added `ScaleDownNodesWithCustomControllerPods` on line 332 below.

cluster-autoscaler/utils/drain/drain.go
```diff
@@ -231,6 +158,104 @@ func GetPodsForDeletionOnNodeDrain(
 	return pods, daemonSetPods, nil, nil
 }
 
+func legacyCheckForReplicatedPods(listers kube_util.ListerRegistry, pod *apiv1.Pod, minReplica int32) (replicated bool, isDaemonSetPod bool, blockingPod *BlockingPod, err error) {
```
vadasambar, line 161: The name `legacyCheckForReplicatedPods` doesn't account for checking for daemon set pods or blocking pods but using `legacyCheckForReplicatedAndDaemonSetAndBlockingPods` feels too long so I have kept it as is for now.

vadasambar: Note that the return type `replicated bool, isDaemonSetPod bool, blockingPod *BlockingPod, err error` has names. I did this because I think it improves readability when you hover over the function call like in IDEs e.g.,
ile
image

This might not be the best approach since it can confuse someone when they look at line 162 and 165 below (because the variables are already defined due to them being named return types)

cluster-autoscaler/utils/drain/drain.go
```diff
@@ -231,6 +158,104 @@ func GetPodsForDeletionOnNodeDrain(
...
+	replicated = false
+	refKind := ""
+	checkReferences := listers != nil
+	isDaemonSetPod = false
+
+	controllerRef := ControllerRef(pod)
```
vadasambar, line 167: This is the same code as our current code [here](https://github.com/kubernetes/autoscaler/pull/5507/files#diff-a22fbdb27ac8bbd8f9c9597042678aa5f6541d207160adc775d56e13f3fe99beL115-L201) abstracted into a function. I have done some minor changes like:
1. Change return values to include `replicated` and `isDaemonSetPod`. Also, remove `[]*apiv1.Pod` since we don't need to return it from this function. Returning pod list is done in the `else` part [here](https://github.com/kubernetes/autoscaler/pull/5507/files#diff-a22fbdb27ac8bbd8f9c9597042678aa5f6541d207160adc775d56e13f3fe99beR126-R127). 
2. Don't do appending to `daemonSetPods` slice here. Instead just return if `isDaemonSetPod` and [let the calling function do the appending](https://github.com/kubernetes/autoscaler/pull/5507/files#diff-a22fbdb27ac8bbd8f9c9597042678aa5f6541d207160adc775d56e13f3fe99beR131-R133).

cluster-autoscaler/utils/drain/drain_test.go
```diff
@@ -624,7 +642,73 @@ func TestDrain(t *testing.T) {
 		},
 	}
 
-	for _, test := range tests {
+	// run all tests for scaleDownNodesWithCustomControllerPods=false
+	// TODO(vadasambar): remove this when we get rid of scaleDownNodesWithCustomControllerPods
+	for _, test := range append(tests, testOpts{
```
vadasambar, line 647: I am doing `append` in the for loop declaration itself. This might not be the best approach. I didn't want to create 2 variables called `testsWithScaleDownNodesWithCustomControllerPodsDisabled` and `testsWithScaleDownNodesWithCustomControllerPodsEnabled`. Any better suggestions are welcome.

cluster-autoscaler/utils/drain/drain_test.go
```diff
@@ -624,7 +642,73 @@ func TestDrain(t *testing.T) {
...
+
+		if len(pods) != len(test.expectPods) {
+			t.Fatalf("Wrong pod list content: %v", test.description)
+		}
+
+		assert.ElementsMatch(t, test.expectDaemonSetPods, daemonSetPods)
+	}
+
+	// run all tests for scaleDownNodesWithCustomControllerPods=true
+	for _, test := range append(tests, testOpts{
```
vadasambar, line 702: Note that I am running the same tests twice. Code [here](https://github.com/kubernetes/autoscaler/pull/5507/files#diff-cd1a98b8de97414dad2b66a3cf3229b1572a169086d9b7141a266f8ddb597f68R656-R699) is practically the same as code beyond line 702 above except. The only difference is `scaleDownNodesWithCustomControllerPods` is set to false [here](https://github.com/kubernetes/autoscaler/pull/5507/files#diff-cd1a98b8de97414dad2b66a3cf3229b1572a169086d9b7141a266f8ddb597f68R678) and it is set to `true` in the test cases beyond the above line. 

I did think of abstracting duplicate code into a separate function but I am not so sure after thinking about the comment [here](https://github.com/kubernetes/autoscaler/pull/5507#discussion_r1135330465) since abstracting the tests into a function would add another level of indirection. I am a little on the fence about doing that now. Any suggestions here are welcome.

cluster-autoscaler/FAQ.md
vadasambar: Updated.

cluster-autoscaler/utils/drain/drain.go
vadasambar: Abstracted the code into a function. Thank you for the comment.

cluster-autoscaler/utils/drain/drain_test.go
vadasambar: Updated.

cluster-autoscaler/utils/drain/drain_test.go
vadasambar: Updated the code (removed global variable and injected functions). Might need a little more feedback.

cluster-autoscaler/utils/drain/drain_test.go
x13n: What is wrong about being explicit? I'd something along the lines of:
1. Make custom controller pods flag a part of the test case struct.
2. Define a list of shared test cases.
3. Define a list of flag-disabled test cases (copy the shared ones and add the extra one)
4. Define a list of flag-enabled test cases (copy the shared ones, flip the flag on all and add the extra one)
5. Define the body of the test once and run it for a combined list (flag-enabled + flag-disabled).

WDYT?

x13n: Alternatively, just copy all test cases and have a single list in the first place :)

cluster-autoscaler/main.go
vadasambar: Let me think about this. :pray:

vadasambar: I will change the name. I hope I don't have to change it again.

cluster-autoscaler/utils/drain/drain_test.go
vadasambar: Thank you for the suggestion. I like the idea of putting the flag on the test struct and combining all tests into one. Let me try it out.

cluster-autoscaler/simulator/drain_test.go
```diff
@@ -42,9 +42,10 @@ func TestGetPodsToMove(t *testing.T) {
 		},
 	}
 	deleteOptions := NodeDeleteOptions{
-		SkipNodesWithSystemPods:   true,
-		SkipNodesWithLocalStorage: true,
-		MinReplicaCount:           0,
+		SkipNodesWithSystemPods:           true,
+		SkipNodesWithLocalStorage:         true,
+		MinReplicaCount:                   0,
+		SkipNodesWithCustomControllerPods: true,
```
vadasambar, line 48: I have set `SkipNodesWithCustomControllerPods: true` to preserve current behavior of these test.

cluster-autoscaler/utils/drain/drain_test.go
vadasambar: Updated the code https://github.com/kubernetes/autoscaler/pull/5507/files#diff-cd1a98b8de97414dad2b66a3cf3229b1572a169086d9b7141a266f8ddb597f68

cluster-autoscaler/main.go
vadasambar: Changed the name.

cluster-autoscaler/utils/drain/drain.go
```diff
@@ -113,19 +113,19 @@ func GetPodsForDeletionOnNodeDrain(
...
 
-		if scaleDownNodesWithCustomControllerPods {
+		if skipNodesWithCustomControllerPods {
+			// TODO(vadasambar): remove this when we get rid of skipNodesWithCustomControllerPods
+			replicated, isDaemonSetPod, blockingPod, err = legacyCheckForReplicatedPods(listers, pod, minReplica)
+			if err != nil {
+				return []*apiv1.Pod{}, []*apiv1.Pod{}, blockingPod, err
+			}
+		} else {
 			if controllerRef != nil {
```
x13n, line 123: Looks like `controllerRef` is not used anywhere else, so maybe just directly call `ControllerRef()` here?

cluster-autoscaler/utils/drain/drain_test.go
```diff
@@ -640,75 +642,36 @@ func TestDrain(t *testing.T) {
...
-			assert.Nil(t, blockingPod)
-			if err != nil {
-				t.Fatalf("%s: error occurred: %v", test.description, err)
-			}
-		}
-
-		if len(pods) != len(test.expectPods) {
-			t.Fatalf("Wrong pod list content: %v", test.description)
-		}
+	for i := range tests {
```
x13n, line 656: My suggestion was to duplicate the test cases and flip the flag on the copy to ensure the new flag doesn't affect the vast majority of test cases.

x13n (changes requested):

cluster-autoscaler/utils/drain/drain_test.go
vadasambar: I think there's another problem with the code where the test cases for flag=false are not there for shared test cases. 

> My suggestion was to duplicate the test cases and flip the flag on the copy to ensure the new flag doesn't affect the vast majority of test cases.

I see. Just to confirm my understanding. You mean something like this right?
```go
tests := []{
		// *flag: true test cases*
		{
			// shared test case 1
                       flag: true
		},
		{
			// shared test case 2
                       flag: true
		},
		{
			// additional test case 1
                       flag: true
		},
		// *flag: false test cases*
		{
			// shared test case 1 
                       flag: false
		},
		{
			// shared test case 2
                       flag: false
		},
		{
			// additional test case 1
                       flag: false
		},

}

for _, test := range tests {
	// execute test case
}
```
`shared test case 1`, `shared test case 2` and `additional test case 1` are duplicated with some minor tweaks (if required) and flipped flags.

cluster-autoscaler/utils/drain/drain.go
vadasambar: Updated. Thank you for pointing this out.

cluster-autoscaler/utils/drain/drain_test.go
vadasambar: I think we might need a more future-proof solution here but it would also increase the code complexity.

vadasambar: Maybe I can duplicate the test cases for now and there can be another issue to work through a better solution for test cases duplication/maintainability problem.

vadasambar: > we would have to change name of the duplicated test cases 

This can be solved by adding the flag in the error log output [here](https://github.com/kubernetes/autoscaler/blob/2f10f76b3f4a974f5285415262ddfaf0d44cfe02/cluster-autoscaler/utils/drain/drain_test.go#L952-L968) so that the error output is something like this:
```
drain_test.go:955: Custom-controller-managed non-blocking pod: unexpected non-error, skipNodesWithCustomControllerPods: false
```

vadasambar: Not sure if this is what you wanted to tell me. This can be a possible solution:
```go
sharedTests := []testOpts{
		{
			// shared test case 1
		},
		{
			// shared test case 2
		},

}

allTests = []testOpts{}
for _, sharedTest := range sharedTests {
	sharedTest.skipNodesWithCustomControllerPods = true
	allTests = append(allTests, sharedTest)
	sharedTest.skipNodesWithCustomControllerPods = false
	allTests = append(allTests, sharedTest)
}

allTests = append(allTests, testOpts{
	// additional test case 1
			   flag: true
	})

allTests = append(allTests, testOpts{
	// additional test case 1
			   flag: false
	})


for _, sharedTest := range sharedTests {
	sharedTest.newFlagInFuture = true
	allTests = append(allTests, sharedTest)
	sharedTest.newFlagInFuture = false
	allTests = append(allTests, sharedTest)
}

allTests = append(allTests, testOpts{
	// additional test case 2 for new flag in future
			   flag: false
	})

allTests = append(allTests, testOpts{
	// additional test case 2 for new flag in future
				flag: true
	})


for _, test := range tests {
	// execute test case
}

```

vadasambar: I have updated this PR with a reference implementation based on the above idea. Happy to change it based on your review comments.

cluster-autoscaler/utils/drain/drain_test.go
```diff
@@ -642,23 +642,30 @@ func TestDrain(t *testing.T) {
...
 	}
 
-	for i := range tests {
-		// this flag is true by default in CA flags
-		tests[i].skipNodesWithCustomControllerPods = true
+	allTests := []testOpts{}
+	for _, sharedTest := range sharedTests {
+		// to execute the same shared tests for when the skipNodesWithCustomControllerPods flag is true
+		// and when the flag is false
+		sharedTest.skipNodesWithCustomControllerPods = true
```
x13n, line 651: nit: You could update `sharedTest.description` to append a flag here, so it only affects the shared tests.

cluster-autoscaler/utils/drain/drain_test.go
x13n: LGTM, thanks!

x13n:
Thanks for the changes! The code looks good to me now, can you just squash the commits before merging?

cluster-autoscaler/utils/drain/drain_test.go
vadasambar: Updated. :+1:

cluster-autoscaler/utils/drain/drain_test.go
```diff
@@ -624,7 +645,52 @@ func TestDrain(t *testing.T) {
...
 
-	for _, test := range tests {
+	allTests := []testOpts{}
+	// Note: be careful about modifying the underlying reference values for sharedTest
+	// since they are shared (changing it once will change it for all shallow copies of sharedTest)
+	for _, sharedTest := range sharedTests {
+		// make sure you shallow copy the test like this
+		// before you modify it
+		// (so that modifying one test doesn't affect another)
+		enabledTest := sharedTest
```
vadasambar, line 655: Note that I am creating a shallow copy of the `sharedTest` I want to modify `sharedTest` once doesn't affect the second modification. e.g.,
```go
for _, sharedTest := range sharedTests {
	sharedTest.skipNodesWithCustomControllerPods = true
	sharedTest.description = fmt.Sprintf("%s with skipNodesWithCustomControllerPods: %v", 
		sharedTest.description, sharedTest.skipNodesWithCustomControllerPods)
	allTests = append(allTests, sharedTest)

	sharedTest.skipNodesWithCustomControllerPods = false
	sharedTest.description = fmt.Sprintf("%s with skipNodesWithCustomControllerPods: %v", 
		sharedTest.description, sharedTest.skipNodesWithCustomControllerPods)
	allTests = append(allTests, sharedTest)
}

fmt.Println("allTests[0]", allTests[0])
fmt.Println("allTests[1]", allTests[1])
```
Prints
```
allTests[0] {RC-managed pod with skipNodesWithCustomControllerPods:true [0xc00018ad80] [] [0xc000498c60] [] false [0xc00018ad80] [] <nil> true}
allTests[1] {RC-managed pod with skipNodesWithCustomControllerPods:true with skipNodesWithCustomControllerPods:false [0xc00018ad80] [] [0xc000498c60] [] false [0xc00018ad80] [] <nil> false}
```
Note the description and the last bool (which is our flag)

To overcome this, I use shallow copy.

vadasambar: Output when we use shallow copy
```
allTests[0] {RC-managed pod with skipNodesWithCustomControllerPods:true [0xc000202d80] [] [0xc0004b54a0] [] false [0xc000202d80] [] <nil> true}
allTests[1] {RC-managed pod with skipNodesWithCustomControllerPods:false [0xc000202d80] [] [0xc0004b54a0] [] false [0xc000202d80] [] <nil> false}
```

vadasambar:
> Thanks for the changes! The code looks good to me now, can you just squash the commits before merging?

Thank you for bearing with me. I have squashed the commits and [added a comment](https://github.com/kubernetes/autoscaler/pull/5507#discussion_r1144238214).

vadasambar:
@x13n unless you have any other comments, I think I'm done from my side.

x13n:
Great, thank you for the changes!

/lgtm
/approve

PR #5518, opened 2023-02-17
Add "resource_name" to scaled_up_gpu_nodes_total and scaled_down_gpu_nodes_total metrics

kawych:
#### What type of PR is this?
/kind feature

Add a resource_name field to scaled_up/down_gpu_nodes_total to differentiate between different types of GPU, which are represented by different custom resource. Credit to @hbostan for the implementation.

```release-note
Add breakdown by custom resource name to GPU-related metrics
```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/scale_up.go
+++ b/cluster-autoscaler/core/scale_up.go
@@ -209,7 +209,6 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 	}
 
 	now := time.Now()
-	gpuLabel := context.CloudProvider.GPULabel()
 	availableGPUTypes := context.CloudProvider.GetAvailableGPUTypes()
 	expansionOptions := make(map[string]expander.Option, 0)
 	skippedNodeGroups := map[string]status.Reasons{}
@@ -405,7 +404,9 @@ func ScaleUp(context *context.AutoscalingContext, processors *ca_processors.Auto
 
 		klog.V(1).Infof("Final scale-up plan: %v", scaleUpInfos)
 		for _, info := range scaleUpInfos {
-			typedErr := executeScaleUp(context, clusterStateRegistry, info, gpu.GetGpuTypeForMetrics(gpuLabel, availableGPUTypes, nodeInfo.Node(), nil), now)
+			gpuConfig := context.CloudProvider.GetNodeGpuConfig(nodeInfo.Node())
+			gpuResourceName, gpuType := gpu.GetGpuInfoForMetrics(gpuConfig, availableGPUTypes, nodeInfo.Node(), nil)
+			typedErr := executeScaleUp(context, clusterStateRegistry, info, gpuResourceName, gpuType, now)
 			if typedErr != nil {
 				return scaleUpError(
 					&status.ScaleUpStatus{
@@ -444,7 +445,6 @@ func ScaleUpToNodeGroupMinSize(context *context.AutoscalingContext, processors *
 	nodes []*apiv1.Node, nodeInfos map[string]*schedulerframework.NodeInfo) (*status.ScaleUpStatus, errors.AutoscalerError) {
 	now := time.Now()
 	nodeGroups := context.CloudProvider.NodeGroups()
-	gpuLabel := context.CloudProvider.GPULabel()
 	availableGPUTypes := context.CloudProvider.GetAvailableGPUTypes()
 	scaleUpInfos := make([]nodegroupset.ScaleUpInfo, 0)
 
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/metrics/metrics.go
```diff
@@ -246,7 +246,7 @@ var (
 			Namespace: caNamespace,
 			Name:      "scaled_up_gpu_nodes_total",
 			Help:      "Number of GPU nodes added by CA, by GPU name.",
-		}, []string{"gpu_name"},
+		}, []string{"resource_name", "gpu_name"},
```
towca, line 249: Since this is specifically for the resource name of a GPU, and will only be set on a GPU scale-up, I'd prefix "resource_name" with "gpu" as well (here and everywhere in this PR)? Otherwise the label name is quite ambiguous - CPU and memory are resources as well and the trigger for most scale-ups, but this won't be set for them.

cluster-autoscaler/metrics/metrics.go
```diff
@@ -270,7 +270,7 @@ var (
 			Namespace: caNamespace,
 			Name:      "scaled_down_gpu_nodes_total",
 			Help:      "Number of GPU nodes removed by CA, by reason and GPU name.",
-		}, []string{"reason", "gpu_name"},
+		}, []string{"reason", "resource_name", "gpu_name"},
```
towca, line 273: Do you know if it's safe to add a new label in the middle of the existing ones (e.g. I could imagine that some metric collector could treat earlier "gpu_name" values as "resource_name" after new 3-value metrics are emitted)?

cluster-autoscaler/utils/gpu/gpu.go
```diff
@@ -47,47 +47,44 @@ const (
...
-
-	if !labelFound {
-		// no label, fallback to generic solution
-		if capacityFound && !capacity.IsZero() {
-			return MetricsGenericGPU
-		}
-
-		// no signs of GPU
-		return MetricsNoGPU
+func GetGpuTypeForMetrics(gpuConfig *cloudprovider.GpuConfig, availableGPUTypes map[string]struct{}, node *apiv1.Node, nodeGroup cloudprovider.NodeGroup) (string, string) {
```
towca, line 53: The name of the function no longer reflects the returned values, and it's hard to figure out what's returned just from reading the signature. Maybe we could make the name more generic (GetGpuInfoForMetrics?), and name the return values instead?

cluster-autoscaler/utils/gpu/gpu.go
```diff
@@ -47,47 +47,44 @@ const (
...
+	// There is no sign of GPU
+	if gpuConfig == nil {
```
towca, line 55: nit: It took me a bit to figure out that the PR doesn't change behavior for the existing GPU logic (there's still 1 difference - this function only looks at capacity, while GetNodeGpuConfig utilizes NodeHasGpu which looks at allocatable - but capacity and allocatable should be in sync for GPUs, and allocatable is arguably more correct - so it looks fine to me). This function could really use a unit test, if you're up for it.

towca (changes requested):

cluster-autoscaler/metrics/metrics.go
kawych: done

cluster-autoscaler/metrics/metrics.go
kawych: it's safe, it should be exposed in prometheus format with label names, so that metric collectors have no trouble identifying the right label

cluster-autoscaler/utils/gpu/gpu.go
kawych: done

cluster-autoscaler/utils/gpu/gpu.go
kawych: This is a little bit tricky, started writing a test but I found out that previous PR #5459 introduced a "hidden" import cycle (so far the cycle happens only if you use the cloudprovider test package). If you're OK with it, I'd prefer to follow up in the next PR

cluster-autoscaler/core/scale_up.go
```diff
@@ -605,7 +605,7 @@ func filterNodeGroupsByPods(
 	return result
 }
 
-func executeScaleUp(context *context.AutoscalingContext, clusterStateRegistry *clusterstate.ClusterStateRegistry, info nodegroupset.ScaleUpInfo, gpuType string, now time.Time) errors.AutoscalerError {
+func executeScaleUp(context *context.AutoscalingContext, clusterStateRegistry *clusterstate.ClusterStateRegistry, info nodegroupset.ScaleUpInfo, resourceName, gpuType string, now time.Time) errors.AutoscalerError {
```
towca, line 608: nit: I'd add the gpu prefix to the parameter name as well, it's ambiguous without it. If you're changing that, `metricResourceName` x3 in this PR would probably be better as `gpuResourceName` as well (but that's much less ambiguous since they're returned from a function with GPU in the name and immediately used).

cluster-autoscaler/metrics/metrics.go
```diff
@@ -490,10 +490,10 @@ func RegisterError(err errors.AutoscalerError) {
 }
 
 // RegisterScaleUp records number of nodes added by scale up
-func RegisterScaleUp(nodesCount int, gpuType string) {
+func RegisterScaleUp(nodesCount int, resourceName, gpuType string) {
```
towca, line 493: nit: Gpu would be nice here and in RegisterScaleDown below as well - seems ambiguous as in executeScaleUp.

cluster-autoscaler/utils/gpu/gpu.go
towca: The cycle will probably have to be solved sooner or later, but a follow-up SGTM.

towca (commented):
/lgtm
/approve
/hold

I'd still like to see more "gpu" prefixes in some places since the names seem ambiguous without them. Feel free to unhold if you disagree.

towca (approved):

cluster-autoscaler/core/scale_up.go
kawych: done

cluster-autoscaler/metrics/metrics.go
kawych: done

PR #5519, opened 2023-02-17
remove dead code in clusterapi provider tests

elmiko:
#### What type of PR is this?

/kind cleanup

This change removes an `if` statement that was left behind after a refactor. The test in question has the same logic embedded into a previous conditional and the removed statement has no effect on the tests.

```release-note
NONE
```


alexander-demicev (approved):
/lgtm

PR #5521, opened 2023-02-20
fix(*): refresh node instance cache when nodegroup not found in deleteCreatedNodesWithErrors

qianlei90:
#### What type of PR is this?

/kind bug

In #4926, @bpineau have fixed the panic issue in `deleteCreatedNodesWithErrors`, but CA is still not working until instance cache in clusterstateregistry is refreshed， which may up to 2min.

This PR sync up the instance cache with cloud provider.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -771,6 +771,7 @@ func (a *StaticAutoscaler) deleteCreatedNodesWithErrors() (bool, error) {
 			continue
 		}
 		if nodeGroup == nil || reflect.ValueOf(nodeGroup).IsNil() {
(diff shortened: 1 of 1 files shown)


bpineau:
Thanks @qianlei90 ; I wondered if we might end up here and force-refreshing in loop, until we get throttled.

Had a look at my fleet's "has no known nodegroup" logs emitted today, they seems to last between a few seconds to a couple minutes (likely, resolved by the scheduled background refreshes) so it seems ok (with that limited dataset).

qianlei90:
@x13n PTAL.

x13n:
I think high qps generated by cache refreshes could be addressed by exponential backoff on actual API calls. The PR as is makes sense to me. Thanks @qianlei90 !

/lgtm
/approve

PR #5537, opened 2023-02-27
cluster-autoscaler: Add option to disable scale down of unready nodes

### LINKED ISSUE #4876: cluster-autoscaler: Add option to disable scale down for unready nodes
Cluster Autoscaler

There are cases where a user may not want the unready nodes to be removed from a cluster. 

As an example, but not limited to, this might me useful in case a node is unreachable for a period of a time and local data live there, the node shall remain in the cluster, and possibly an admin may want to take any actions for recovering it.

Provide the ability to disable scale-down for unready nodes. The Cluster Autoscaler ships with the `--scale-down-unready-time` flag, that configures "How long an unready node should be unneeded before it is eligible for scale down". Currently, if the flag is set to a negative or zero value, the unready nodes will be removed immediately. We could modify the logic of the code and if a negative value is provided, the cluster autoscaler shall disable scale down for unready nodes.

Another alternative solution is to introduce a new flag for disabling scale down for unready nodes, but, we will end up with two flags that will be dependant.

**Additional context.**:

gregth:
Add flag '--scale-down-unready-enabled' to enable or disable scale-down of unready nodes. Default value set to true for backwards compatibility (i.e., allow scale-down of unready nodes).

Cluster Autoscaler

#### What type of PR is this?
/kind feature

This PR adds the option to disable scale down for unready nodes. 

There are cases where a user may not want the unready nodes to be removed from a cluster.
As an example, but not limited to, this might be useful in case a node is unreachable for a period of time and local data live there, the node shall remain in the cluster, and possibly an admin may want to take any actions for recovering it.

Fixes #4876.

```release-note
Set the --scale-down-unready-enabled flag to a false in order to disable scale-down for unready nodes. The default value is true for backwards compatibility.
```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/scaledown/eligibility/eligibility.go
+++ b/cluster-autoscaler/core/scaledown/eligibility/eligibility.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/utils/klogx"
 
 	apiv1 "k8s.io/api/core/v1"
+	kube_util "k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
 	klog "k8s.io/klog/v2"
 	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
 )
@@ -133,6 +134,15 @@ func (c *Checker) unremovableReasonAndNodeUtilization(context *context.Autoscali
 		// (and the default PreFilteringScaleDownNodeProcessor would indeed filter them out).
 		klog.Warningf("Skipped %s from delete consideration - the node is not autoscaled", node.Name)
 		return simulator.NotAutoscaled, nil
+	}
+
+	// If scale down of unready nodes is disabled, skip the node if it is unready
+	if !context.ScaleDownUnreadyEnabled {
+		ready, _, _ := kube_util.GetReadinessState(node)
+		if !ready {
+			klog.V(4).Infof("Skipping unready node %s from delete consideration - scale-down of unready nodes is disabled", node.Name)
+			return simulator.ScaleDownUnreadyDisabled, nil
+		}
 	}
 
 	underutilized, err := c.isNodeBelowUtilizationThreshold(context, node, nodeGroup, utilInfo)
--- a/cluster-autoscaler/core/scaledown/eligibility/eligibility_test.go
+++ b/cluster-autoscaler/core/scaledown/eligibility/eligibility_test.go
@@ -50,6 +50,9 @@ func TestFilterOutUnremovable(t *testing.T) {
 	noScaleDownNode.Annotations = map[string]string{ScaleDownDisabledKey: "true"}
 	SetNodeReadyState(noScaleDownNode, true, time.Time{})
 
+	unreadyNode := BuildTestNode("unready", 1000, 10)
+	SetNodeReadyState(unreadyNode, false, time.Time{})
+
 	bigPod := BuildTestPod("bigPod", 600, 0)
 	bigPod.Spec.NodeName = "regular"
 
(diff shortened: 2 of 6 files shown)


cluster-autoscaler/core/scaledown/eligibility/eligibility_test.go
```diff
@@ -50,18 +50,45 @@ func TestFilterOutUnremovable(t *testing.T) {
...
-	testCases := []struct {
+	type testCase struct {
 		desc  string
 		nodes []*apiv1.Node
 		pods  []*apiv1.Pod
 		want  []string
-	}{
+	}
+
+	runTestCase := func(tc testCase, scaleDownUnready bool) {
```
x13n, line 69: nit: Why isn't `scaleDownUnready` a part of `testCase`?

x13n (commented):
Mostly LGTM, just one minor comment.

cluster-autoscaler/core/scaledown/eligibility/eligibility_test.go
gregth: I created a fixup to make this part of `testCase`. The diff is a bit ugly (due to go indentation). Shall I *force-push*?

x13n: Yeah, please do!

gregth: @x13n done

cluster-autoscaler/core/scaledown/eligibility/eligibility.go
```diff
@@ -135,6 +136,13 @@ func (c *Checker) unremovableReasonAndNodeUtilization(context *context.Autoscali
 		return simulator.NotAutoscaled, nil
 	}
 
+	// In case the node in unready, check if scale down of unready nodes is enabled
+	ready, _, _ := kube_util.GetReadinessState(node)
```
x13n, line 140: nit: `GetNodeReadiness` is a bit more costly than bool flag check, so I'd avoid checking readiness when `ScaleDownUnreadyEnabled` is `true`.

cluster-autoscaler/core/scaledown/eligibility/eligibility_test.go
```diff
@@ -50,44 +50,66 @@ func TestFilterOutUnremovable(t *testing.T) {
...
+			nodes:            []*apiv1.Node{unreadyNode},
+			want:             []string{"unready"},
+			scaleDownUnready: true,
 		},
 		{
-			desc:  "underutilized node stays",
-			nodes: []*apiv1.Node{regularNode},
-			pods:  []*apiv1.Pod{smallPod},
-			want:  []string{"regular"},
+			// Ensure that unready node is filtered out when scale-down for unready is disabled
```
x13n, line 108: Isn't this comment redundant with test case `desc`?

x13n (changes requested):
Again, mostly LGTM, but found two more really minor comments.

cluster-autoscaler/core/scaledown/eligibility/eligibility_test.go
gregth: Totally, I will remove it.

cluster-autoscaler/core/scaledown/eligibility/eligibility.go
gregth: @x13n ack, that makes sense. I pushed a fixup on the branch. I put a separate commit for better visibility of the changes, let's squash them all together before merging.

x13n:
Thanks, looks good to me now, can you squash the commits?

gregth:
@x13n thanks, squashed the commits and force-pushed.

PR #5538, opened 2023-02-27
[WIP] fix: [ISS-5465] add a flag to let CA delegate node deletion to the cloud provider

### LINKED ISSUE #5465: clusterapi: Cluster Autoscaler destroys healthy nodes when processing old unregistered nodes when multiple machinesets exist
cluster-autoscaler

v1.2.4

Component version:

```
Client Version: version.Info{Major:"1", Minor:"24+", GitVersion:"v1.24.0-alpha.3.428+425ff1c8e2d298-dirty", GitCommit:"425ff1c8e2d29874926abb7da37641cf785c4ff3", GitTreeState:"dirty", BuildDate:"2022-03-10T21:18:29Z", GoVersion:"go1.17.7", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.7", GitCommit:"e6f35974b08862a23e7f4aad8e5d7f7f2de26c15", GitTreeState:"clean", BuildDate:"2022-10-12T10:50:21Z", GoVersion:"go1.18.7", Compiler:"gc", Platform:"linux/amd64"}
```

AWS

The cluster autoscaler should manage machine deployments and understand how to manage scale up and scale down behavior when a machine deployment is in the midst of a rollout. When a machine deployment is in a rollout, multiple machinesets exist with non-zero replica counts.

When the cluster autoscaler is managing machine deployments that are in undergoing a rollout (they own multiple machinesets that have non-zero replicas) and the provisioning nodes are not able to join the cluster in a timely manner, the cluster autoscaler detects these as "long unregistered" nodes. I expected the cluster autoscaler to correctly remove these long unregistered nodes and stop considering them for pod placement, and scale up another node group.

**What happened instead?**:
What we observed was that when the cluster autoscaler is managing machine deployments that are in undergoing a rollout (they own multiple machinesets that have non-zero replicas) and the provisioning nodes are not able to join the cluster in a timely manner, the cluster autoscaler detects these as "long unregistered" nodes. When it determines that these nodes have not joined the cluster after the max provision time has been exceeded, the cluster autoscaler marks these machines for deletion by setting an annotation and decreases the replica count of the machine deployment for the nodegroup.

Scaling the machine deployment at this point is tricky - scale up on the replica count causes the newest machineset desired replicas to increase. Scale down on the machine deployment replica count causes the oldest machineset to shrink the desired replicas. I think this is the core confusion for the cluster autoscaler.

So when the replica count for the machine deployment is decreased, the CAPI controller takes over. When the machine deployment replica count is decreased and multiple machinesets exist with non-zero replicas, the oldest machineset has a replica removed. This is not the machine that was stuck provisioning, because it was part of the new machineset that did not successfully finish coming up. So this causes a healthy node in the old machineset to be scaled down, in addition to the old provisioning machine that was annotated.

1. Create a machine deployment with 2 replicas, backed by an AWSMachineTemplate and KubeadmConfigTemplate for the infrastructure and a no-op workload to occupy the each of the 2 nodes resources in a cluster with the cluster autoscaler configured
2. Wait for the node group to come up with 2 healthy nodes and the no-op workload to become ready
3. Configure a new KubeadmConfigTemplate that won't join the cluster (e.g `preKubeadmCommands: ['sleep infinity']`) and update the machine deployment to begin a rollout
4. Observe that the cluster autoscaler marks the stuck "provisioned" machines for deletion with the annotation and scales the node group down _even though there is no place for the pods to go_.

A brief gist which shows the sort of configuration we used to reproduce this: https://gist.github.com/cnmcavoy/364cb3020017c22fc63c75540cf58923 

We first encountered this running a fork of the v1.26 cluster autoscaler, which we were running to use the scale from zero feature set. We re-tested on the official tagged v1.24 release and reproduced it there as well.

wmgroot:
#### What type of PR is this?
/kind bug
/kind api-change

Cluster Autoscaler currently tries to remediate nodes that fail to join a cluster by deleting them. This is helpful to retry nodes sooner, but CA does not have a good way to delete nodes available to it. To accomplish this, it uses the replica count on a node group resource to downsize the number of desired nodes after marking a node with an annotation that the cloud provider should prefer this node for deletion.

During this period of time, the desired count of replicas is incorrect. Furthermore, this behavior can occur multiple times in parallel, resulting in multiple node deletions occurring simultaneously, which can significantly decrease the number of desired replicas in a node group far below what is actually needed to run all the pods that would schedule in that node group.

If a significant number of nodes fail to join the cluster, this can result in node groups operating with desired replica counts far below their actual intended replica count for extended periods of time, resulting in losses of cluster capacity and evicted pods with no node to schedule on.

#### Actual vs Expected Behavior
Example situation.
* 4 pods, which can schedule 2 per node.
* 1 worker node group, configured to replace nodes 1 at a time by adding a new node, then deleting an old node.
* 2 nodes, which hold the 4 pods.

You push an update the specifications of the worker node group, such as by changing the AMI, root volume size, etc.
* The node rollout begins, and a new worker begins to join the cluster.
* The worker node has a problem that prevents it from successfully joining the cluster (instance type shortage, failure to execute kubeadm successfully, etc)
* CA eventually notes the node as an "unregisteredOldNode" after `max-node-provision-time` expires and attempts to delete the node via `DeleteNodes()`.
* CA marks the new node for deletion.
* CA adjusts the node group desired replica count from 2 to 1.
* 1 replica is only enough to schedule 2 of our 4 pods.

The replica count should always reflect the number of replicas needed to schedule the pods in the cluster. CA has no other way to cause nodes to be deleted from the cluster, so it currently misuses the desired replica count in order to achieve this.

Fixes #5465

This change unfortunately is not compatible across all existing cloud providers, since it requires cloud provider specific implementation of marking nodes for deletion instead of downscaling the replica counts in certain situations. Previously, DeleteNodes() handled all cases equally.

The core cluster-api controller also needs to be modified to delete machines that have been annotated by Cluster Autoscaler, rather than only preferring them for deletion during scaling events.
https://github.com/kubernetes-sigs/cluster-api/pull/8205

These changes can be safely introduced independently of each other, but the external-node-deletion flag should not be enabled until the cluster-api core controller is running the updated version.

```release-note
NONE
```
No immediate user-facing changes are required. This change is gated behind a feature gate applied as an arg to CA. By default the behavior is unchanged.

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_utils.go
+++ b/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_utils.go
@@ -63,9 +63,14 @@ var (
 	errInvalidMaxAnnotation = errors.New("invalid max annotation")
 
 	// machineDeleteAnnotationKey is the annotation used by cluster-api to indicate
-	// that a machine should be deleted. Because this key can be affected by the
+	// that a machine should be preferred for deletion. Because this key can be affected by the
 	// CAPI_GROUP env variable, it is initialized here.
 	machineDeleteAnnotationKey = getMachineDeleteAnnotationKey()
+
+	// markMachineForDelete is the annotation used by cluster-api to indicate
+	// that a machine should be deleted ASAP. Because this key can be affected by the
+	// CAPI_GROUP env variable, it is initialized here.
+	markMachineForDeleteAnnotationKey = getMarkMachineForDeleteAnnotationKey()
 
 	// machineAnnotationKey is the annotation used by the cluster-api on Node objects
 	// to specify the name of the related Machine object. Because this can be affected
@@ -265,10 +270,21 @@ func getNodeGroupMaxSizeAnnotationKey() string {
 }
 
 // getMachineDeleteAnnotationKey returns the key that is used by cluster-api for marking
-// machines to be deleted. This function is needed because the user can change the default
+// machines to be preferred for deletion during scaling events.
+// This function is needed because the user can change the default
 // group name by using the CAPI_GROUP environment variable.
 func getMachineDeleteAnnotationKey() string {
 	key := fmt.Sprintf("%s/delete-machine", getCAPIGroup())
+	return key
+}
+
+// getMarkMachineForDeleteAnnotationKey returns the key that is used by cluster-api for marking
+// machines to be deleted as soon as possible by cluster-api without a scaling event occurring.
+// This function is needed because the user can change the default group name by using the CAPI_GROUP
+// environment variable. A new annotation was needed to avoid breaking compatibility with the existing
+// delete-machine annotation that only marks nodes as "preferred" for deletion during scaling events.
+func getMarkMachineForDeleteAnnotationKey() string {
+	key := fmt.Sprintf("%s/machine-marked-for-delete", getCAPIGroup())
 	return key
 }
 
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -748,7 +748,16 @@ func removeOldUnregisteredNodes(unregisteredNodes []clusterstate.UnregisteredNod
 				klog.Warningf("Failed to remove node %s: node group min size reached, skipping unregistered node removal", unregisteredNode.Node.Name)
 				continue
 			}
-			err = nodeGroup.DeleteNodes([]*apiv1.Node{unregisteredNode.Node})
+
+			if context.AutoscalingOptions.ExternalNodeDeletion {
+				// allow the cloud provider to handle node deletion instead of deleting nodes through downscaling replica counts
+				fmt.Printf("MarkNodesForDeletion (removeOldUnregisteredNodes)\n")
+				err = nodeGroup.MarkNodesForDeletion([]*apiv1.Node{unregisteredNode.Node})
+			} else {
+				fmt.Printf("DeleteNodes (removeOldUnregisteredNodes)\n")
+				err = nodeGroup.DeleteNodes([]*apiv1.Node{unregisteredNode.Node})
+			}
+
 			csr.InvalidateNodeInstancesCacheEntry(nodeGroup)
 			if err != nil {
 				klog.Warningf("Failed to remove node %s: %v", unregisteredNode.Node.Name, err)
@@ -800,7 +809,15 @@ func (a *StaticAutoscaler) deleteCreatedNodesWithErrors() (bool, error) {
 		if nodeGroup == nil {
 			err = fmt.Errorf("node group %s not found", nodeGroupId)
 		} else {
-			err = nodeGroup.DeleteNodes(nodesToBeDeleted)
+
+			if a.AutoscalingContext.AutoscalingOptions.ExternalNodeDeletion {
+				// allow the cloud provider to handle node deletion instead of deleting nodes through downscaling replica counts
+				fmt.Printf("MarkNodesForDeletion (deleteCreatedNodesWithErrors)\n")
+				err = nodeGroup.MarkNodesForDeletion(nodesToBeDeleted)
+			} else {
+				fmt.Printf("DeleteNodes (deleteCreatedNodesWithErrors)\n")
+				err = nodeGroup.DeleteNodes(nodesToBeDeleted)
+			}
 		}
 
 		if err != nil {
--- a/cluster-autoscaler/FAQ.md
+++ b/cluster-autoscaler/FAQ.md
@@ -792,6 +792,7 @@ The following startup parameters are supported for cluster autoscaler:
 | `cordon-node-before-terminating` | Should CA cordon nodes before terminating during downscale process | false
 | `record-duplicated-events` | Enable the autoscaler to print duplicated events within a 5 minute window. | false
 | `debugging-snapshot-enabled` | Whether the debugging snapshot of cluster autoscaler feature is enabled. | false
+| `external-node-deletion` | If true, CA will not adjust replica counts to delete failed nodes, deletion is delegated to the cloud provider. | false
 
 # Troubleshooting:
 
--- a/cluster-autoscaler/cloudprovider/alicloud/alicloud_auto_scaling_group.go
+++ b/cluster-autoscaler/cloudprovider/alicloud/alicloud_auto_scaling_group.go
@@ -115,6 +115,13 @@ func (asg *Asg) Belongs(node *apiv1.Node) (bool, error) {
 	return true, nil
 }
 
+// MarkNodesForDeletion notifies the cloud provider that the provided nodes need to be deleted.
+// The cloud provider is expected to handle deletion of the marked nodes gracefully.
+// Error is returned either on failure or if the given node doesn't belong to this node group.
+func (asg *Asg) MarkNodesForDeletion(nodes []*apiv1.Node) error {
+	return cloudprovider.ErrNotImplemented
+}
+
 // DeleteNodes deletes the nodes from the group.
 func (asg *Asg) DeleteNodes(nodes []*apiv1.Node) error {
 	size, err := asg.manager.GetAsgSize(asg)
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
@@ -289,6 +289,13 @@ func (ng *AwsNodeGroup) Belongs(node *apiv1.Node) (bool, error) {
 	return true, nil
 }
 
+// MarkNodesForDeletion notifies the cloud provider that the provided nodes need to be deleted.
+// The cloud provider is expected to handle deletion of the marked nodes gracefully.
+// Error is returned either on failure or if the given node doesn't belong to this node group.
+func (ng *AwsNodeGroup) MarkNodesForDeletion(nodes []*apiv1.Node) error {
+	return cloudprovider.ErrNotImplemented
+}
+
 // DeleteNodes deletes the nodes from the group.
 func (ng *AwsNodeGroup) DeleteNodes(nodes []*apiv1.Node) error {
 	size := ng.asg.curSize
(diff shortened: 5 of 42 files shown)


x13n (changes requested):
I'm not convinced this requires cloud provider interface change. Why can't it be an option on Cluster API cloud provider level to control whether nodes are removed by CA or just marked by CA for removal by cloud provider?

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_utils.go
```diff
@@ -63,10 +63,15 @@ var (
...
 	// machineDeleteAnnotationKey is the annotation used by cluster-api to indicate
-	// that a machine should be deleted. Because this key can be affected by the
+	// that a machine should be preferred for deletion. Because this key can be affected by the
 	// CAPI_GROUP env variable, it is initialized here.
 	machineDeleteAnnotationKey = getMachineDeleteAnnotationKey()
 
+	// markMachineForDelete is the annotation used by cluster-api to indicate
+	// that a machine should be deleted ASAP. Because this key can be affected by the
+	// CAPI_GROUP env variable, it is initialized here.
+	markMachineForDeleteAnnotationKey = getMarkMachineForDeleteAnnotationKey()
```
wmgroot, line 73: I'm adding this at the recommendation of the cluster-api maintainers to avoid breaking the existing behavior of the `delete-machine` annotation.

x13n:
> I don't think the CAPI cloud provider implementation of DeleteNodes() would have a way to tell if DeleteNodes was called as a scale down event vs clean up of nodes that are failing to join the cluster.

Why not? It would be exactly the same - through a commandline flag. The only difference would be that all the other cloud providers wouldn't have to have the boilerplate "not implemented" methods and that the flag would live somewhere under `cloudprovider/clusterapi` dir, rather than in `main.go`.

wmgroot:
> > I don't think the CAPI cloud provider implementation of DeleteNodes() would have a way to tell if DeleteNodes was called as a scale down event vs clean up of nodes that are failing to join the cluster.
> 
> Why not? It would be exactly the same - through a commandline flag. The only difference would be that all the other cloud providers wouldn't have to have the boilerplate "not implemented" methods and that the flag would live somewhere under `cloudprovider/clusterapi` dir, rather than in `main.go`.

I'm not completely following. There are 3 places where DeleteNodes() is called outside of any cloudprovider implementation. The solution requires that we handle 2 of those 3 calls (failed node cases) differently, and leave the existing behavior alone in the last (scaledown case). The cluster-api implementation has no idea which of the 3 potential DeleteNode() paths is being triggered unless we make a change to break it into 2 separate functions, or pass an additional parameter in DeleteNodes() to allow branching functionality in the provider implementations of the function.

Furthermore, I believe this issue is not limited to the cluster-api implementation. Other cloud provider implementations of DeleteNodes() reduce the replica count of their node groups as a way to delete failing nodes. This is flawed behavior, and should be corrected. The replica count should always reflect the desired state of the cluster, even when nodes are failing to join the cluster quickly.

elmiko:
this pr generally makes sense to me, and i get what you are saying @wmgroot about the provider not knowing the difference between the 2 types of deletion. i think it would be good to bring this issue to a SIG if possible to get a few more ideas about if we should expand the interface.

i'd also like to study this a little more to better understand the different deletion methodologies and how those interact with the cluster-api provider. thanks @wmgroot for such a thorough pr.

wmgroot:
Sounds good to me. We've got the Monday SIG Autoscaling meeting on our calendars, and I should be able to bring this up there for further discussion.

We're going to do some more testing around how this shift in behavior plays with fallback scenarios, since CA would still want to reduce replica counts for a node group that failed to have a node join if it will be switching to a new fallback node group.

elmiko:
apologies @wmgroot , i had a conflict today and was not able to make the meeting. i'm happy to help with followups if i can.

x13n:
Apologies on my end as well - I didn't make it to the SIG meeting either. I see from the meeting notes you're following up on this with @MaciekPytel. Is this PR on hold for now?

wmgroot:
@MaciekPytel mentioned that there was some nuance to be understood about DecreaseTargetSize() vs DeleteNodes(). It might be possible for the replica reduction logic present in the cluster-api implementation of DeleteNodes() to be handled by DecreaseTargetSize() instead.

I think we should put review of this PR on hold until that possibility can be explored. I can update here once that's happened.

elmiko:
thanks for the update @wmgroot , i'm fine to put this on hold until you are ready to move forward. feel free to remove the hold when you are ready.

/hold for further investigation

wmgroot:
Alright, sorry for the delay, things got busy at the end of quarter for me.
I've tested trying to avoid the use of any new flags by tweaking the CAPI logic to only decrease replicas of a node group when DecreaseTargetSize() is called.

This did not work. In my test case I have 2 nodes that have nothing scheduled to them (other than daemonsets).
```
$ kubectl get machine -n capi-awscmhdev2 --context awscmhdev2
NAME                                                   CLUSTER      NODENAME                                       PROVIDERID                              PHASE          AGE     VERSION
worker-private-test-ca-az2-744477878-bxpqk             awscmhdev2   ip-10-115-223-57.us-east-2.compute.internal    aws:///us-east-2b/i-0fe56f28fa8b8be30   Running        10d     v1.24.7
worker-private-test-ca-az2-744477878-gvpp5             awscmhdev2   ip-10-115-213-94.us-east-2.compute.internal    aws:///us-east-2b/i-0b31e1f26e69e23e9   Running        10d     v1.24.7
```

CA still identifies that the nodes should be deleted, but does not ever actually delete the unneeded nodes.
```
I0324 18:07:56.806731       1 klogx.go:87] Node ip-10-115-223-57.us-east-2.compute.internal - cpu utilization 0.233571
I0324 18:07:56.807114       1 nodes.go:84] ip-10-115-223-57.us-east-2.compute.internal is unneeded since 2023-03-24 18:05:31.597356663 +0000 UTC m=+3033.013308437 duration 0s
I0324 18:08:06.605511       1 nodes.go:126] ip-10-115-223-57.us-east-2.compute.internal was unneeded for 0s
I0324 18:08:06.626397       1 taints.go:135] Successfully added DeletionCandidateTaint on node ip-10-115-223-57.us-east-2.compute.internal
I0324 18:10:57.034238       1 klogx.go:87] Node ip-10-115-223-57.us-east-2.compute.internal - cpu utilization 0.233571
I0324 18:10:57.038712       1 nodes.go:84] ip-10-115-223-57.us-east-2.compute.internal is unneeded since 2023-03-24 18:05:31.597356663 +0000 UTC m=+3033.013308437 duration 2m54.827681204s
I0324 18:11:11.232802       1 nodes.go:126] ip-10-115-223-57.us-east-2.compute.internal was unneeded for 2m54.827681204s
I0324 18:11:11.244785       1 taints.go:135] Successfully added ToBeDeletedTaint on node ip-10-115-223-57.us-east-2.compute.internal
I0324 18:11:11.244973       1 event_sink_logging_wrapper.go:48] Event(v1.ObjectReference{Kind:"Node", Namespace:"", Name:"ip-10-115-223-57.us-east-2.compute.internal", UID:"e41a2ddf-12ce-4a44-8de6-15292ed65286", APIVersion:"v1", ResourceVersion:"488849061", FieldPath:""}): type: 'Normal' reason: 'ScaleDown' marked the node as toBeDeleted/unschedulable
I0324 18:11:12.015497       1 actuator.go:160] Scale-down: removing empty node "ip-10-115-223-57.us-east-2.compute.internal"
I0324 18:11:12.021442       1 event_sink_logging_wrapper.go:48] Event(v1.ObjectReference{Kind:"ConfigMap", Namespace:"capi-awscmhdev2", Name:"cluster-autoscaler-status", UID:"0f83dc30-13e4-4769-aef5-df069f60cb20", APIVersion:"v1", ResourceVersion:"488849019", FieldPath:""}): type: 'Normal' reason: 'ScaleDownEmpty' Scale-down: removing empty node "ip-10-115-223-57.us-east-2.compute.internal"
I0324 18:14:16.263272       1 eligibility.go:111] Skipping ip-10-115-223-57.us-east-2.compute.internal from delete consideration - the node is currently being deleted
```

This is because DecreaseTargetSize() is only ever called during one code path, in fixNodeGroupSize(), which is not part of the standard scaledown path. Scale down currently relies on DeleteNodes() to perform the actual scale down of a node group. This seems inherently wrong to me if I consider what the intent of each function should be based on its name.

So I think we're back to square one with this; I do not see any way to start untangling these assumptions that can result in incorrect replica count reduction in certain cases without adding a flag. These are the options that I think we have.
* Update DeleteNodes to only delete nodes across all cloud providers and start calling DecreaseTargetSize() as part of the standard scaledown path (too much work for one PR).
* Do the above, but enable a flag to allow users to contribute fixes to each cloud provider one at a time.
* Implement the current PR, which works around the issue by separating the cases where replicas should and should not be scaled down by adding a function that deletes nodes without downscaling the node groups (this PR).

I'll plan to pop into the next CA contributor meeting to discuss further.

x13n:
/assign @MaciekPytel 

Thoughts on this one?

elmiko:
i think the flag addition is probably the simplest way to solve this, at the expense of adding a new interface function. i'd like to talk about this again at an upcoming SIG meeting since this seems like it is going to affect cluster-api users the most.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -748,7 +748,16 @@ func removeOldUnregisteredNodes(unregisteredNodes []clusterstate.UnregisteredNod
 				klog.Warningf("Failed to remove node %s: node group min size reached, skipping unregistered node removal", unregisteredNode.Node.Name)
 				continue
 			}
-			err = nodeGroup.DeleteNodes([]*apiv1.Node{unregisteredNode.Node})
+
+			if context.AutoscalingOptions.ExternalNodeDeletion {
+				// allow the cloud provider to handle node deletion instead of deleting nodes through downscaling replica counts
+				fmt.Printf("MarkNodesForDeletion (removeOldUnregisteredNodes)\n")
```
elmiko, line 754: i think we'll need to upgrade these printf's to log messages, perhaps using `V(2).InfoS` level.

elmiko (commented):
aside from the printf comment, my only other question is should we emit warnings for the providers that do not implement this as it will help debugging for users who might enable this flag without fully understanding the implications?

x13n:
Closing due to lack of activity. Please reopen if still relevant.

/close

PR #5552, opened 2023-02-28
Fix RemovalSimulation for parallel scale down

yaroslava-serdiuk:
#### What type of PR is this?

/kind bug

Added label for outer loop and break outer loop if the timer send signal. Before CA did nothing and didn't have any limit for scale down simulation. 

Also, I modified the default value for scaleDownSimulationTimeout flag, because 30 seconds is enough to process ~1000 non empty nodes and in the same time cluster snapshot is not too old comparing to the snapshot for 5 minutes timeout.

### DIFF
--- a/cluster-autoscaler/core/scaledown/planner/planner.go
+++ b/cluster-autoscaler/core/scaledown/planner/planner.go
@@ -239,7 +239,6 @@ func (p *Planner) injectPods(pods []*apiv1.Pod) error {
 
 // categorizeNodes determines, for each node, whether it can be eventually
 // removed or if there are reasons preventing that.
-// TODO: Track remaining PDB budget.
 func (p *Planner) categorizeNodes(podDestinations map[string]bool, scaleDownCandidates []*apiv1.Node) {
 	unremovableTimeout := p.latestUpdate.Add(p.context.AutoscalingOptions.UnremovableNodeRecheckTimeout)
 	unremovableCount := 0
@@ -251,12 +250,11 @@ func (p *Planner) categorizeNodes(podDestinations map[string]bool, scaleDownCand
 	}
 	p.nodeUtilizationMap = utilizationMap
 	timer := time.NewTimer(p.context.ScaleDownSimulationTimeout)
+
 	for i, node := range currentlyUnneededNodeNames {
-		select {
-		case <-timer.C:
+		if timedOut(timer) {
 			klog.Warningf("%d out of %d nodes skipped in scale down simulation due to timeout.", len(currentlyUnneededNodeNames)-i, len(currentlyUnneededNodeNames))
 			break
-		default:
 		}
 		removable, unremovable := p.rs.SimulateNodeRemoval(node, podDestinations, p.latestUpdate, p.context.RemainingPdbTracker.GetPdbs())
 		if removable != nil {
@@ -332,4 +330,13 @@ func sortByRisk(nodes []simulator.NodeToBeRemoved) []simulator.NodeToBeRemoved {
 		}
 	}
 	return append(okNodes, riskyNodes...)
+}
+
+func timedOut(timer *time.Timer) bool {
+	select {
+	case <-timer.C:
+		return true
+	default:
+		return false
+	}
 }
--- a/cluster-autoscaler/core/scaledown/planner/planner.go
+++ b/cluster-autoscaler/core/scaledown/planner/planner.go
@@ -40,6 +40,8 @@ import (
 	klog "k8s.io/klog/v2"
 )
 
+const unneededNodesLimit = 1000
+
 type eligibilityChecker interface {
 	FilterOutUnremovable(context *context.AutoscalingContext, scaleDownCandidates []*apiv1.Node, timestamp time.Time, unremovableNodes *unremovable.Nodes) ([]string, map[string]utilization.Info, []*simulator.UnremovableNode)
 }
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/core/scaledown/unremovable/nodes.go
```diff
@@ -67,6 +67,12 @@ func (n *Nodes) Update(nodeInfos NodeInfoGetter, timestamp time.Time) {
 	n.ttls = newTTLs
 }
 
+// Contains returns true iff a given node is unremovable.
+func (n *Nodes) Contains(nodeName string) bool {
+	_, found := n.ttls[nodeName]
```
x13n, line 72: This works for nodes added with `AddTimeout`, but not for `Add`/`AddReason`. I think checking `n.reasons` would actually deliver on the promise from the function comment. Also, since you're extending the public interface of this struct, it might be worth to add tests verifying that nodes added  via any of the `Add*` methods can then be checked for presence using `Contains`.

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -251,26 +250,27 @@ func (p *Planner) categorizeNodes(podDestinations map[string]bool, scaleDownCand
 	}
 	p.nodeUtilizationMap = utilizationMap
 	timer := time.NewTimer(p.context.ScaleDownSimulationTimeout)
+RemovalSimulation:
```
x13n, line 253: nit, optional: I'd consider extracting an extra function (e.g. `timedOut(time.Timer) bool`) instead of introducing the loop label. I'm a fan of short functions :)

cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -251,26 +250,27 @@ func (p *Planner) categorizeNodes(podDestinations map[string]bool, scaleDownCand
...
 	for i, node := range currentlyUnneededNodeNames {
 		select {
 		case <-timer.C:
 			klog.Warningf("%d out of %d nodes skipped in scale down simulation due to timeout.", len(currentlyUnneededNodeNames)-i, len(currentlyUnneededNodeNames))
-			break
+			break RemovalSimulation
```
x13n, line 258: WDYT about testing if the timeout is actually honored?

x13n (changes requested):

cluster-autoscaler/core/scaledown/planner/planner.go
yaroslava-serdiuk: Not sure I understand your proposal. Do you mean extract the whole loop to another function or perform a for loop with additional condition `timedOut(time.Timer)`?

yaroslava-serdiuk: Extracting the whole loop into the function will reduce readability as for me, because the function will have a lot of parameters.

cluster-autoscaler/core/scaledown/planner/planner.go
yaroslava-serdiuk: That's require a timer mock, because with ScaleDownSimulationTimeout=0seconds, the channel still sends a signal later then the loop processing for small number of nodes. 
I don't think we want to move timer to categorizeNodes() variables. Do you have other ideas how to mock the timer?

cluster-autoscaler/core/scaledown/planner/planner.go
x13n: Yeah, extracting the loop might not be very readable, I was thinking about replacing the whole select statement with:

```
if timedOut(timer) {
    break
}
```

to avoid the need for a label.

cluster-autoscaler/core/scaledown/planner/planner.go
x13n: I was thinking rather about using real timer, but mocking removalSimulator to make sure it is not called after the timeout.

cluster-autoscaler/core/scaledown/unremovable/nodes_test.go
```diff
@@ -80,6 +81,21 @@ func TestUpdate(t *testing.T) {
 	}
 }
 
+func TestContains(t *testing.T) {
```
x13n, line 84: Thanks for the test! It would be good to have a test case for a node that wasn't added. Right now it seems the test checks whether `Contains` always returns `true` :)

cluster-autoscaler/core/scaledown/planner/planner_test.go
```diff
@@ -441,17 +480,27 @@ func TestUpdateClusterState(t *testing.T) {
...
 			provider := testprovider.NewTestCloudProvider(nil, nil)
-			context, err := NewScaleTestAutoscalingContext(config.AutoscalingOptions{ScaleDownSimulationTimeout: 5 * time.Minute}, &fake.Clientset{}, registry, provider, nil, nil)
+			context, err := NewScaleTestAutoscalingContext(config.AutoscalingOptions{ScaleDownSimulationTimeout: 1 * time.Second}, &fake.Clientset{}, registry, provider, nil, nil)
 			assert.NoError(t, err)
 			clustersnapshot.InitializeClusterSnapshotOrDie(t, context.ClusterSnapshot, tc.nodes, tc.pods)
 			deleteOptions := simulator.NodeDeleteOptions{}
 			p := New(&context, NewTestProcessors(&context), deleteOptions)
 			p.eligibilityChecker = &fakeEligibilityChecker{eligible: asMap(tc.eligible)}
+			if tc.isSimulationTimeout {
+				context.AutoscalingOptions.ScaleDownSimulationTimeout = 1 * time.Nanosecond
```
x13n, line 490: I'm a bit worried ns/ms scale may make this test flaky - in particular 1ns can easily pass before we have a chance to process the first node.

PR #5553, opened 2023-02-28
[cluster-api] Handle ignored errors

mboersma:
#### What type of PR is this?

/kind cleanup

Cleans up code in the cluster-api cluster-autoscaler backend, in particular handling ignored errors. Also makes other small changes suggested by golangci-lint.

N/A

```release-note
NONE
```

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller_test.go
+++ b/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller_test.go
@@ -171,9 +171,7 @@ func mustCreateTestController(t *testing.T, testConfigs ...*testConfig) (*machin
 		if subresource != "scale" {
 			// Handle a bug in the client-go fakeNamespaceScaleClient, where the action namespace and subresource are
 			// switched for update actions
-			if action.GetVerb() == "update" && action.GetNamespace() == "scale" {
-				subresource = "scale"
-			} else {
+			if action.GetVerb() != "update" || action.GetNamespace() != "scale" {
 				// Do not attempt to respond to anything but scale subresource requests
 				return false, nil, nil
 			}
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller_test.go
```diff
@@ -385,7 +383,9 @@ func createTestConfigs(specs ...testSpec) []*testConfig {
 			}
 			config.machineDeployment.SetAnnotations(spec.annotations)
 			config.machineDeployment.SetLabels(machineDeploymentLabels)
-			unstructured.SetNestedStringMap(config.machineDeployment.Object, machineDeploymentLabels, "spec", "selector", "matchLabels")
+			if err := unstructured.SetNestedStringMap(config.machineDeployment.Object, machineDeploymentLabels, "spec", "selector", "matchLabels"); err != nil {
+				panic(err)
```
mboersma, line 387: I wagered that using `panic()` in a unit test might be acceptable. It's either that or change several functions in this unit test to return and expect errors; I'm happy to do that if it's preferable.

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller_test.go
```diff
@@ -1899,11 +1903,6 @@ func Test_machineController_nodeGroupForNode(t *testing.T) {
 	}, nil)
 	mdTestConfigs = append(mdTestConfigs, uniqueMDConfig)
 
-	allMachineDeployments := make([]*unstructured.Unstructured, 0, len(mdTestConfigs))
```
mboersma, line 1902: This and the similar section below look like they were copied from a similar unit test, but the values aren't used here.

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller_test.go
elmiko: i think this is probably ok

elmiko (commented):
this is nice cleanup, thanks @mboersma !

/approve
/lgtm

PR #5554, opened 2023-03-01
Opt-in to not violate pod maxSkew when scaling down ScheduleAnyway pods

grosser:
#### What type of PR is this?

/kind feature

Given a deployment that has a AZ topologySpreadConstraint with ScheduleAnyway
And it's skew is almost violated (for example skewed 1 and has maxSkew 1)
And 1 of the pods is on a node that is underutilized
And that node is in an AZ that otherwise has no room to host the pod
Then CA should not remove that node.

An obvious counter point would be "then just set DoNotSchedule!".
The reason this does not work is that DoNotSchedule adds a pretty big reliability risk to any application since it will refuse to schedule pods even when there is room available in other AZs.

So this is a hybrid approach, where we get the reliability of ScheduleAnyway but also prevent CA from violating maxskew by scaling down.
Obviously this can never be the default for CA since the usecase of "I don't care much about skew" is also valid.

So we could go for either
- **Option A**: add an opt-in flag to CA to treat certain (or all) topology skews as DoNotSchedule
  - Pro: cluster users do not need to be aware of this gotcha and the whole cluster is instantly improved
  - Con: users cannot opt-out/in to this behavior and it might increase cost from empty nodes
- **Option B**: treat certain (or all) topology skews as DoNotSchedule depending on pod annotation
  - Pro: behavior can be slowly rolled out for selected workloads or only applied where needed
  - Con: adoption would be slow or need an admission-webhook that sets it everwhere
- **Option C**: A + B
  - Pro: everyone can use it as they need
  - Con: more complexity + user interface

Doing the change further up the stack would allow using cli options, but also means we modify pods outside of the `fork` of cluster-state. So if we decide on **option A** then we can either move it up the stack or pass the autoscaling-context into the simulator.
Passing options to the simulator will be a bit ugly for legacy, so ideally that would get removed first.

```release-note
- CA will not scale down nodes when it would cause skew violations for pods
```

#### Next steps
- decide which approach to chose
- change code as necessary + add docs

### DIFF
--- a/cluster-autoscaler/simulator/cluster.go
+++ b/cluster-autoscaler/simulator/cluster.go
@@ -167,6 +167,15 @@ func (r *RemovalSimulator) SimulateNodeRemoval(
 	}
 
 	err = r.withForkedSnapshot(func() error {
+		// try hard to not disrupt skew by only removing nodes that have pods that can be rescheduled without skew
+		for podIndex, pod := range podsToRemove {
+			for constraintIndex, c := range pod.Spec.TopologySpreadConstraints {
+				if c.TopologyKey == "topology.kubernetes.io/zone" && c.WhenUnsatisfiable == "ScheduleAnyway" {
+					podsToRemove[podIndex].Spec.TopologySpreadConstraints[constraintIndex].WhenUnsatisfiable = "DoNotSchedule"
+				}
+			}
+		}
+
 		return r.findPlaceFor(nodeName, podsToRemove, destinationMap, timestamp)
 	})
 	if err != nil {


grosser:
marking this as non-draft since it is now confirmed working ... but it's still not ready to merge

grosser:
This seemed like it was working but now I got some random failures ... even when using actual `DoNotSchedule` so maybe this can't work and I misunderstand how CA behaves in this scenario

towca:
Do I understand correctly that you'd like to swap `ScheduleAnyway` to `DoNotSchedule` if a cluster/pod opts in and the following conditions are true?
1. The pod's configured `maxSkew` is 1
2. The pod's current maximum skew is 1
3. The pod's node is in a zone that otherwise has no room to host the pod

If so, how do you plan on checking conditions 2. and 3.? For 2., Cluster Autoscaler doesn't have skew-checking logic itself, it relies on the vendored scheduler plugins. For 3., it looks like you'd have to run the scale-down simulations multiple times - once without the swap, analyze the available pod destinations, then if there are none in a given zone, rerun them again with the swap? It seems pretty complex for handling a very specific case. 

Also, keep in mind that if the pods actually have `ScheduleAnyway` specified, it's unlikely that they'll have their configured `maxSkew` satisfied in an autoscaled cluster (because it's not taken into account while scaling up). So the described scenario doesn't seem likely in practice.

Unless the plan is to always swap `ScheduleAnyway` to `DoNotSchedule` in scale-down simulations for opted in clusters/pods (=what's currently in the PR, minus the opting in)? While this approach would indeed help the pod spread in some cases where `maxSkew` is already satisfied (like the example you describe), there are also cases where it would make the spread worse if `maxSkew` is not already satisfied e.g.:
* 3 zones (A,B, C); 52 pods; A: 50 pods, B: 2, C:0 pods; `maxSkew=1`
* CA could scale-down nodes in zone A, and reschedule some of the pods to B, resulting in a better spread (e.g. `A: 25, B:25, C:0`), but with this approach it won't be able to, since the pods would only be schedulable in C.
* Scaling down within A to improve utilization (while not changing the skew at all) would also be blocked.

Looking from a more general perspective, using `topologySpreadConstraint` for high-availability pod spreading with Autoscaling is IMO fundamentally broken. DoNotSchedule doesn't handle zone failures well as you're well aware, and ScheduleAnyway is not taken into account while scaling up - which I think is a much bigger problem than it not being taken into account while scaling down. While we can technically try to make it better in specific scenarios (like your PR is aiming to), I don't think it's the way to go just because `topologySpreadConstraint` is not the right tool for the job. 

There is a new component being developed that allows for balancing and autoscaling at the same time: https://github.com/kubernetes/autoscaler/blob/master/balancer/proposals/balancer.md. I'm not sure how far along it is, but maybe that'd work for you use-case?

/assign @towca

grosser:
Updated the description to clarify that this is not about `maxSkew 1` but "do not violate the skew, no matter what the maxSkew is".
Yes, the plan is to "plan is to always swap ScheduleAnyway to DoNotSchedule in scale-down simulations".

Skew is usually not a problem during scale-up since we use `--balance-similar-node-groups=true`.

Re A:52/B:50/C:2 scenario: yeah it would be worse in that case, but I'm trying to prevent ever getting into that state :D

Re high-availability pod spreading: yes agree, none of these tools are great 😞 
... our next step was to use per-az deployments as a kind of middle-ground.
(lmk if you know any gotchas for that too before we invest more time there :D)

I'm in general hoping this feature makes things a bit better without making it worse.
For the opt-in I can go either way, but wanted to ask for direction first before I add more complexity to this PR.

grosser:
That Balancer KEP looks good ... yeah basically our conclusion too, but we'll use 1 HPA per AZ to embrace the uneven load.

PR #5556, opened 2023-03-02
Limit the rate of adding new uneeded nodes

x13n:
This is an optimization to avoid spending too much time in scale down simulation. The main idea is that it doesn't make sense to add new unneeded nodes at a rate that is higher than the rate at which we can delete them afterwards.

Faster scale down in large clusters.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/scaledown/planner/planner.go
+++ b/cluster-autoscaler/core/scaledown/planner/planner.go
@@ -40,8 +40,6 @@ import (
 	klog "k8s.io/klog/v2"
 )
 
-const unneededNodesLimit = 1000
-
 type eligibilityChecker interface {
 	FilterOutUnremovable(context *context.AutoscalingContext, scaleDownCandidates []*apiv1.Node, timestamp time.Time, unremovableNodes *unremovable.Nodes) ([]string, map[string]utilization.Info, []*simulator.UnremovableNode)
 }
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -279,6 +291,48 @@ func (p *Planner) categorizeNodes(podDestinations map[string]bool, scaleDownCand
...
+	n := p.context.AutoscalingOptions.MaxScaleDownParallelism
+	extraBuffer := n
+	limit := len(p.unneededNodes.AsList()) + n + extraBuffer
+	// TODO(x13n): Use moving average instead of min.
+	loopInterval := int64(p.minUpdateInterval)
+	u := int64(p.context.AutoscalingOptions.NodeGroupDefaults.ScaleDownUnneededTime)
+	if u < loopInterval {
+		u = loopInterval
+	}
+	upperBound := n * int(u/loopInterval)
```
yaroslava-serdiuk, line 329: Let's have higher limit. 

If ScaleDownUnneededTime=60s, and loopInterval=10s we will have upperBound n*6. Imagine it's smaller than limit. 
Each loop we will remove n nodes, so n*6 nodes we will remove in exactly 1minute. However if CA have filtered some nodes in NodesToDelete (because min size would be violate or resource limits) we would have <n*6 nodes, let's say n*3 which we will be able to remove in 30 seconds and so other empty nodes won't start removal because unneeded time haven't pass yet.

x13n: Right. This is exactly what the extra buffer is for, so adding it here so that it is consistent between `limit` and `upperBound`.

yaroslava-serdiuk: Thanks!

PR #5558, opened 2023-03-02
chart: Add option to install VPA

avorima:
#### What type of PR is this?

/kind feature

I currently install a VPA for the cluster-autoscaler alongside the chart, but it would be nice to get it all from the same source.

```release-note
Add chart option to install VPA
```

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -11,4 +11,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.26.0
+version: 9.27.0
--- a/charts/cluster-autoscaler/README.md
+++ b/charts/cluster-autoscaler/README.md
@@ -318,6 +318,26 @@ Containers:
 
 Though enough for the majority of installations, the default PodSecurityPolicy _could_ be too restrictive depending on the specifics of your release. Please make sure to check that the template fits with any customizations made or disable it by setting `rbac.pspEnabled` to `false`.
 
+### VerticalPodAutoscaler
+
+The chart can install a [`VerticalPodAutoscaler`](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/README.md) for the Deployment if needed. A VPA can help minimize wasted resources when usage spikes periodically.
+The following example snippet can be used to install a [`autoscaling.k8s.io/v1` VPA](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go).
+
+```yaml
+vpa:
+  enabled: true
+  apiVersion: v1
+  containerPolicy:
+    minAllowed:
+      cpu: 50m
+      memory: 150Mi
+    maxAllowed:
+      cpu: 100m
+      memory: 300Mi
+```
+
+Supported VPA versions are `v1` and `v1beta2`.
+
 ## Values
 
 | Key | Type | Default | Description |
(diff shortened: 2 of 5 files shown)


charts/cluster-autoscaler/Chart.yaml
```diff
@@ -11,4 +11,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.25.0
+version: 9.25.1
```
gjtempleton, line 14: We're adding new functionality here, so can we make this a minor version bump please?

charts/cluster-autoscaler/README.md
```diff
@@ -309,6 +309,25 @@ Containers:
 
 Though enough for the majority of installations, the default PodSecurityPolicy _could_ be too restrictive depending on the specifics of your release. Please make sure to check that the template fits with any customizations made or disable it by setting `rbac.pspEnabled` to `false`.
 
+### VerticalPodAutoscaler
+
+The chart can install a VPA for the Deployment if needed. A VPA can help minimize wasted resources when usage spikes periodically.
+The following example snippet can be used to install a `autoscaling.k8s.io/v1` VPA.
```
gjtempleton, line 315: Can we link out to the VPA docs please, there's a reasonable chance users will be reading this Readme disconnected from this repo, and may want to find out more about the VPA.

gjtempleton (commented):
/assign @gjtempleton 

A couple of small requests.

charts/cluster-autoscaler/templates/vpa.yaml
```diff
@@ -0,0 +1,20 @@
...
+{{ include "cluster-autoscaler.labels" . | indent 4 }}
+  name: {{ template "cluster-autoscaler.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  targetRef:
+    apiVersion: {{ template "deployment.apiVersion" . }}
+    kind: Deployment
+    name: {{ template "cluster-autoscaler.fullname" . }}
+  updatePolicy:
+    updateMode: "Auto"
```
voelzmo, line 15: I see people run their VPA in `updateMode: off` for a while to observe if anything weird would happen when they switched on `updateMode: auto` – so maybe this is an option you'd want to make configurable?

voelzmo (changes requested):
Thanks!

Some small comments about VPA configuration settings, the API version is the only thing that I'd really recommend to change – the other two remarks are just to start a discussion. I wouldn't argue against merging this, even if you were to decide to keep these as-is.

charts/cluster-autoscaler/templates/vpa.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Values.vpa.enabled -}}
+apiVersion: "autoscaling.k8s.io/{{ .Values.vpa.apiVersion }}"
```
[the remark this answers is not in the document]
avorima: Oh sure. I was actually trying to find out when it got released, but I must've missed those release notes.

charts/cluster-autoscaler/values.yaml
```diff
@@ -384,3 +384,22 @@ updateStrategy: {}
...
+  # vpa.apiVersion -- APIVersion of the VerticalPodAutoscaler. The template only supports v1 and v1beta2.
+  apiVersion: v1
+  # vpa.containerPolicy -- Configure the containerPolicy of the cluster-autoscaler container. Content is rendered as YAML.
+  containerPolicy: {}
+    # minAllowed:
+    #   cpu: 50m
+    #   memory: 150Mi
+    # maxAllowed:
+    #   cpu: 100m
+    #   memory: 300Mi
```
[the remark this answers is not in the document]
avorima: The values here were meant to illustrate the possible contents of a containerPolicy. I'll add a link to the spec instead.

I just checked some deployments and the lowest usage I saw was `2m/22Mi`. I'd say `20m/50Mi` is a good compromise that still allows some headroom. I'll add these to the example in the README and leave out the `maxAllowed`.

charts/cluster-autoscaler/templates/vpa.yaml
avorima: I didn't think about that. I'll add it

voelzmo (approved):
/lgtm thanks!

gjtempleton:
Thanks for addressing the feedback.

/lgtm
/approve

PR #5562, opened 2023-03-03
Use snapshot of NodeDeletionTracker for ActuationStatus

yaroslava-serdiuk:
#### What type of PR is this?

/kind bug

Node deletions happens in separate go routine, so it's possible that node was removed during unneeded node verification. In such case we have updated number of ongoing deletions but outdated size of the node group, so CA will have wrong verification of min size of the node group.

### DIFF
--- a/cluster-autoscaler/core/scaledown/actuation/actuator.go
+++ b/cluster-autoscaler/core/scaledown/actuation/actuator.go
@@ -66,12 +66,9 @@ func NewActuator(ctx *context.AutoscalingContext, csr *clusterstate.ClusterState
 	}
 }
 
-// CheckStatus should return an immutable snapshot of ongoing deletions. Before the TODO is addressed, a live object
-// is returned instead of an immutable snapshot.
+// CheckStatus should returns an immutable snapshot of ongoing deletions.
 func (a *Actuator) CheckStatus() scaledown.ActuationStatus {
-	// TODO: snapshot information from the tracker instead of keeping live
-	// updated object.
-	return a.nodeDeletionTracker
+	return a.nodeDeletionTracker.Snapshot()
 }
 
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/core/scaledown/deletiontracker/nodedeletiontracker.go
```diff
@@ -168,3 +168,23 @@ func (n *NodeDeletionTracker) ClearResultsNotNewerThan(t time.Time) {
 	defer n.Unlock()
 	n.deletionResults.DropNotNewerThan(t)
 }
+
+func (n *NodeDeletionTracker) Snapshot() *NodeDeletionTracker {
+	snapshot := NewNodeDeletionTracker(n.evictionsTTL)
+	for k, val := range n.emptyNodeDeletions {
```
x13n, line 174: Accessing fields in `n` should be done with a mutex.

cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -69,9 +69,7 @@ func NewActuator(ctx *context.AutoscalingContext, csr *clusterstate.ClusterState
 // CheckStatus should return an immutable snapshot of ongoing deletions. Before the TODO is addressed, a live object
 // is returned instead of an immutable snapshot.
 func (a *Actuator) CheckStatus() scaledown.ActuationStatus {
-	// TODO: snapshot information from the tracker instead of keeping live
-	// updated object.
-	return a.nodeDeletionTracker
+	return a.nodeDeletionTracker.Snapshot()
```
x13n, line 72: I think ideally we'd just replace `ActuationStatus` interface with a simple struct that doesn't have the mutex overhead. It would also be simpler to mock in tests. Can you add a TODO for this somewhere? Perhaps next to the interface definition.

x13n (changes requested):

PR #5565, opened 2023-03-06
docs: add info about mirror pods

vadasambar:
#### What type of PR is this?

/kind documentation

We use the wording `mirror pods` in our flags but don't really talk about what they mean anywhere.  e.g.,
image
More can be found in the FAQ [here](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca)
This can be confusing for end users. 

Fixes https://github.com/kubernetes/autoscaler/issues/5566

NONE

### DIFF
--- a/cluster-autoscaler/FAQ.md
+++ b/cluster-autoscaler/FAQ.md
@@ -722,6 +722,9 @@ CA respects `nodeSelector` and `requiredDuringSchedulingIgnoredDuringExecution` 
 
 However, CA does not consider "soft" constraints like `preferredDuringSchedulingIgnoredDuringExecution` when selecting node groups. That means that if CA has two or more node groups available for expansion, it will not use soft constraints to pick one node group over another.
 
+### What are mirror pods?
+Mirror pods are pods which are considered `replicated` i.e., if the node on which the pod is running goes down, the pod can get scheduled on another node.  
+
 ****************
 
 ### What are the parameters to CA?


cluster-autoscaler/FAQ.md
```diff
@@ -722,6 +722,9 @@ CA respects `nodeSelector` and `requiredDuringSchedulingIgnoredDuringExecution`
 
 However, CA does not consider "soft" constraints like `preferredDuringSchedulingIgnoredDuringExecution` when selecting node groups. That means that if CA has two or more node groups available for expansion, it will not use soft constraints to pick one node group over another.
 
+### What are mirror pods?
+Mirror pods are pods which are considered `replicated` i.e., if the node on which the pod is running goes down, the pod can get scheduled on another node.
```
vadasambar, line 726: I have deliberately kept the description short here since our internal mechanism for identifying mirror pods is going to change.
Current way of finding mirror pods: https://github.com/kubernetes/autoscaler/blob/e551e1f82e33c044a01daa0e91960d334ad5db7d/cluster-autoscaler/utils/drain/drain.go#L96-L231
New way of finding mirror pods we are working on: https://github.com/kubernetes/autoscaler/pull/5507

gregth: Hey @vadasambar, just took a look at the PR. Thanks for exposing that context! My 2 cents:
1. According to my understanding, a `mirror Pod` is a Kubernetes-wide term, it is not autoscaler specific; as explained in the guide https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/. Essentially, a mirror Pod is a Pod on the API server to represent a static Pod configured with the kubelet. 
2. A mirror Pod is not controlled by the control plane, thus it is not scheduled (please correct me if I am wrong).
3. If the node of the Pod goes down, no control-plane controller will replicate it on another node.

IMHO the proposed definition is not precise and we should rather stick to the Kubernetes website definition. We may point the users to the Kubernetes guide to better understand what a mirror Pod is, and the section you propose to add could be reformed to a "How the CA treats mirror Pods". In particular, we may explain wrt to mirror Pods:
1. The cluster autoscaler can be configured to take mirror Pods into consideration or ignore them when calculating the utilization of a node
2.  The mirror Pods of a node will not prevent the autoscaler from removing a node.

WDYT?

vadasambar: @gregth thanks for the comment and explaining everything to me. I feel quite stupid after creating this PR. I thought mirror pod could be used interchangeably with replicated pod (I thought mirror pod is a lingo specific to cluster-autoscaler). I couldn't be more wrong. 

> According to my understanding, a mirror Pod is a Kubernetes-wide term, it is not autoscaler specific; as explained in the guide https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/. Essentially, a mirror Pod is a Pod on the API server to represent a static Pod configured with the kubelet.

You are right. 

> If the node of the Pod goes down, no control-plane controller will replicate it on another node.

You are right. 

> IMHO the proposed definition is not precise and we should rather stick to the Kubernetes website definition. We may point the users to the Kubernetes guide to better understand what a mirror Pod is, and the section you propose to add could be reformed to a "How the CA treats mirror Pods". In particular, we may explain wrt to mirror Pods:

I agree with linking to the official definition. I don't think the section I wrote in this PR is of much use now. I think having a section around `"How the CA treats mirror Pods"` seems valuable to me. Looks like I can change this PR into:
1. Removing the section I added
2. Linking official docs for mirror pod wherever we have used the term `mirror pod` in the FAQ
3. Adding a section around  `"How the CA treats mirror Pods"`

vadasambar: P.S.
> New way of finding mirror pods we are working on: https://github.com/kubernetes/autoscaler/pull/5507

This ^ is wrong too. We are working on a new way to find pods which should be deleted on node drain.

vadasambar: I think that would reduce confusion.
It seems like we are using the term `mirror pods` in 27 different places
```
suraj@suraj:~/sandbox/autoscaler/cluster-autoscaler$ grep -R "mirror" . --exclude-dir=vendor | grep "pod"
./main.go:      ignoreMirrorPodsUtilization = flag.Bool("ignore-mirror-pods-utilization", false,
./main.go:      skipNodesWithSystemPods            = flag.Bool("skip-nodes-with-system-pods", true, "If true cluster autoscaler will never delete nodes with pods from kube-system (except for DaemonSet or mirror pods)")
./FAQ.md:* The sum of cpu and memory requests of all pods running on this node (DaemonSet pods and Mirror pods are included by default but this is configurable with `--ignore-daemonsets-utilization` and `--ignore-mirror-pods-utilization` flags) is smaller
./FAQ.md:| `ignore-mirror-pods-utilization` | Whether Mirror pods will be ignored when calculating resource utilization for scaling down | false
./FAQ.md:| `skip-nodes-with-system-pods` | If true cluster autoscaler will never delete nodes with pods from kube-system (except for DaemonSet or mirror pods) | true
./utils/test/test_utils.go:// SetMirrorPodSpec sets pod spec to make it a mirror pod
./utils/test/test_utils.go:     pod.ObjectMeta.Annotations[kube_types.ConfigMirrorAnnotationKey] = "mirror"
./utils/pod/pod.go:// IsMirrorPod checks whether the pod is a mirror pod.
./utils/pod/pod_test.go:                        name: "filter-out single mirror pod",
./utils/drain/drain.go: // UnmovableKubeSystemPod - pod is blocking scale down because it's a non-daemonset, non-mirrored, non-pdb-assigned kube-system pod.
./utils/drain/drain.go:                                 return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: UnmovableKubeSystemPod}, fmt.Errorf("non-daemonset, non-mirrored, non-pdb-assigned kube-system pod present: %s", pod.Name)
./config/autoscaling_options.go:        // SkipNodesWithSystemPods tells if nodes with pods from kube-system should be deleted (except for DaemonSet or mirror pods)
./simulator/nodes_test.go:                      name: "node with non-DS/mirror pods",
./simulator/nodes_test.go:                      name: "node with a mirror pod",
./simulator/nodes_test.go:                      name: "node with a deleted mirror pod",
./simulator/utilization/info.go:                // factor mirror pods out of the utilization calculations
./simulator/utilization/info_test.go:   nodeInfo = newNodeInfo(node, pod, pod, pod2, mirrorPod)
./simulator/utilization/info_test.go:   nodeInfo = newNodeInfo(node, pod, pod2, mirrorPod)
./simulator/utilization/info_test.go:   nodeInfo = newNodeInfo(node, pod, mirrorPod, daemonSetPod3)
./simulator/drain.go:   // SkipNodesWithSystemPods tells if nodes with pods from kube-system should be deleted (except for DaemonSet or mirror pods)
./simulator/nodes.go:           // Add scheduled mirror and DS pods
./core/scaledown/planner/planner_test.go:                       name: "recently evicted mirror pod, eligible",
./core/scaledown/actuation/drain.go:// DrainNode works like DrainNodeWithPods, but lists of pods to evict don't have to be provided. All non-mirror, non-DS pods on the
./core/scaledown/actuation/drain_test.go:               "mirror pods are never returned": {
./core/scaledown/actuation/drain_test.go:                       pods:          []*apiv1.Pod{mirrorPod("pod-1"), mirrorPod("pod-2")},
./core/scaledown/actuation/drain_test.go:                               mirrorPod("mirror-pod-1"), mirrorPod("mirror-pod-2"),
./README.md:namespace (Cluster Autoscaler doesn't scale down node with non-mirrored kube-system pods running
```
```
suraj@suraj:~/sandbox/autoscaler/cluster-autoscaler$ grep -R "mirror" . --exclude-dir=vendor | grep "pod" | wc -l
27
```
Based on the latest master commit
```
suraj@suraj:~/sandbox/autoscaler/cluster-autoscaler$ git log
commit 63b334f81a8820511c835401e0133d242d84da08 (HEAD -> master, upstream/master)
```

vadasambar: Based on the discussion, it seems like we can do:
1. add link to official definition of mirror pod OR static pod docs in FAQ wherever mirror pod is mentioned
2. add link to static pod docs wherever mirror pod is mentioned in the code

vadasambar: I will close this PR and create a separate issue for this. Thank you for the feedback.

PR #5577, opened 2023-03-07
OCI: Enhance oci implementation to include oci node pools

trungng92:
#### What type of PR is this?

/kind feature

This pull request enhances the cluster autoscaler's current oci (Oracle Cloud Infrastructure) cloud provider implementation to work with oci node pools in addition to working with oci instance pools.

None. This is a new feature. OCI currently provides a close source implementation for node pools, but this pull requests open sources this implementation.

In OCI, there are two concepts, instance pools and node pools. Instance pools are generic instances. Node pools are specifically a set of instances in a pool used as Kubernetes nodes.

In this implementation, only one type of pool is accepted. This means that only instance pools **xor** node pools will be accepted. If the user provides multiple types of pools, this will exit out with a relevant error statement. In the future, we will support mixed instance pools and node pools.

Users will now be able to use the open source version of the cluster autoscaler with Oracle Cloud Infrastructure's Container Engine for Kubernetes Node Pools.


cluster-autoscaler/cloudprovider/oci/OWNERS
```diff
@@ -1,5 +1,15 @@
 approvers:
 - jlamillan
+- trungng92
```
jlamillan, line 3: I think we should be more restrictive with the number of approvers. Also, there is a _requirement_ that approvers (and reviewers) be [Kubernetes org members](https://github.com/kubernetes/community/blob/master/community-membership.md). There is also an expectation that approvers be involved with the autoscaling project (e.g. monitoring for provider issues, regularly attending Kubernetes SIG-Autoscaling, etc.).

cluster-autoscaler/cloudprovider/oci/oke/oci/cache.go
```diff
@@ -0,0 +1,157 @@
+/*
+Copyright 2020-2023 Oracle and/or its affiliates.
+*/
+
+package oci
```
jlamillan, line 5: I see the `oci` package name is duplicated. For maintainability, we should organize the implementation specific code in separate packages (`oke` & `instancepools`). 

There is also common code between the two implementations that could live an a common package (`common`) so that we could re-use certain functions, constants, etc. rather than having duplicate code (`oci_util.go` is a good start).

cluster-autoscaler/cloudprovider/oci/oke/oci/oci_ref.go
```diff
@@ -0,0 +1,34 @@
...
+type OciRef struct {
+	CompartmentID string
+	NodePoolID    string
+	InstanceID    string
+}
+
+func nodeToOciRef(n *apiv1.Node) (OciRef, error) {
+	return OciRef{
+		// This field comes from the CCM.
+		CompartmentID: n.Annotations["oci.oraclecloud.com/compartment-id"],
```
jlamillan, line 23: This function and even this annotation string are both examples of code-reuse opportunities. 

`const ociAnnotationCompartmentID string = "oci.oraclecloud.com/compartment-id"` is is defined in `oci_cloud_provider.go` and the now two separate implementations of `nodeToOciRef` should be combined into one shared function.

cluster-autoscaler/cloudprovider/oci/oke/oci/oci_cloud_provider.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// Refresh is called before every main loop and can be used to dynamically update cloud provider state.
+// In particular the list of node groups returned by NodeGroups can change as a result of CloudProvider.Refresh().
+func (ocp *OciCloudProvider) Refresh() error {
+	return ocp.manager.Refresh()
+}
+
+func BuildOci(opts config.AutoscalingOptions, do cloudprovider.NodeGroupDiscoveryOptions, rl *cloudprovider.ResourceLimiter) *OciCloudProvider {
+	manager, err := CreateOciManager(nil, do, createKubeClient(opts))
```
jlamillan, line 135: Configuration parameters that make sense for both implementations should be honored. Our [README](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/oci/README.md#optional-cloud-config-file) documents the parameters that can be set via `--cloud-config`. However, the OKE implementation is dropping the config here: `CreateOciManager(nil, do, createKubeClient(opts))`.

cluster-autoscaler/cloudprovider/oci/oke/oci/oci_manager.go
```diff
@@ -0,0 +1,477 @@
...
+
+// SetNodePoolSize sets NodePool size.
+func (m *ociManagerImpl) SetNodePoolSize(np NodePool, size int) error {
+
+	err := m.nodePoolCache.setSize(np.Id(), size)
+	if err != nil {
+		return err
+	}
+
+	// We do not wait for the work request to finish or nodes become active on purpose. This allows
```
jlamillan, line 335: However, we _do_ wait in the instancepool implementation - at least for the instances to be provisioned. 

We also check for capacity and quota errors so that the autoscaler to try to scale a different pool (rather than waiting for `max-node-provision-time` and then trying to scale the same pool again).

We may decide that certain functional behaviors are acceptable between the two implementations. However, they should probably be considered on a case by case basis.

jlamillan (changes requested):
Looks like there is still some work to do with this PR to ensure the behaviors & configurations of the two implementations are consistent. Also, we probably have some more initial refactoring to do in terms of code-organization and reuse to do.

cluster-autoscaler/cloudprovider/oci/oke/oci/cache.go
trungng92: Yes, I agree that there is a lot of common code between the two implementations. There is a plan afterwards to integrate the two implementations more tightly. This is an initial pass to open source our internal code that we know is working.

cluster-autoscaler/cloudprovider/oci/oke/oci/oci_manager.go
trungng92: > However, we do wait in the instancepool implementation - at least for the instances to be provisioned.

I'll set up a meeting with relevant people to discuss how we want to handle this.

> We also check for capacity and quota errors so that the autoscaler to try to scale a different pool (rather than waiting for max-node-provision-time and then trying to scale the same pool again). We added this stuff based on customer feedback.

I do not know if Node Pools provide an easy way to know whether it is failing based on these criteria. But these are a good ideas to add as a future enhancement.

cluster-autoscaler/cloudprovider/oci/OWNERS
mwielgus: We (SIG-Autoscaling) don't apply any necessary contribution count threshold to become a reviewer/approver of company-specific code. As long as there is some indication that a person is employed at the company, we are fine with adding them, trusting that the company internal org structure will handle the rest. So we are fine with both long and short list here. Anyway, as @jlamillan is the current approver, he has the final say in this topic.

trungng92:
> Looks like there is still some work to do with this PR to ensure the behaviors & configurations of the two implementations are consistent. Also, we probably have some more initial refactoring to do in terms of code-organization and reuse to do.

I've made changes relevant to the previous comments. Mainly, I've created 3 different packages:

1. package to contain instance pool code
2. package to contain node pool code
3. shared common package

cluster-autoscaler/cloudprovider/oci/oke/oci/oci_manager.go
trungng92: I had a discussion with our team regarding adding a wait to the node pool implementation. We have decided not to add a wait to the node pool implementation as the current implementation does not do this, and we did not want to introduce any kind of potential behavior change.

In particular, adding a wait for a scale up that is quickly followed by a scale down should not cause any issues because the cluster autoscaler already has logic that puts the scale down into a backoff for X minutes after a scale up is performed.

However, in the case where a scale up is fast followed by another scale up, introducing a wait may mean that the autoscaler will have to wait for the previous scale to complete before it can make an additional decision to scale up again.

cluster-autoscaler/cloudprovider/oci/oke/oci/oci_cloud_provider.go
trungng92: completed with current refactor

cluster-autoscaler/cloudprovider/oci/oke/oci/oci_ref.go
trungng92: completed with current refactor

cluster-autoscaler/cloudprovider/oci/oke/oci/cache.go
trungng92: completed with current refactor. See latest comment. I have split things into three compartments.

1. instance pool
2. node pool
3. common

cluster-autoscaler/cloudprovider/builder/builder_all.go
```diff
@@ -130,7 +130,7 @@ func buildCloudProvider(opts config.AutoscalingOptions, do cloudprovider.NodeGro
 	case cloudprovider.LinodeProviderName:
 		return linode.BuildLinode(opts, do, rl)
 	case cloudprovider.OracleCloudProviderName:
-		return oci.BuildOCI(opts, do, rl)
+		return instancepools.BuildOCI(opts, do, rl)
```
jlamillan, line 133: It _might_ be better to import `instancepools` aliased as "oci" so that we can keep the keep the return `cloudprovider.Build` format. 

i.e. 

```
import (
	oci "k8s.io/autoscaler/cluster-autoscaler/cloudprovider/oci/instancepools"
)
```
...
```
	case cloudprovider.OracleCloudProviderName:
		return oci.BuildOCI(opts, do, rl)
```

cluster-autoscaler/cloudprovider/oci/README.md
```diff
@@ -122,6 +140,10 @@ Configuration via environment-variables:
 - `OCI_REFRESH_INTERVAL` - Optional refresh interval to sync internal cache with OCI API defaults to `2m`.
 - `OCI_USE_NON_POOL_MEMBER_ANNOTATION` - Optional. If true, the node will be annotated as non-pool-member if it doesn't belong to any instance pool and the time-consuming instance pool lookup will be skipped.
```
jlamillan, line 141: Probably want to note that `OCI_USE_NON_POOL_MEMBER_ANNOTATION` applies to instance pools only.

cluster-autoscaler/cloudprovider/oci/nodepools/oci_manager.go
```diff
@@ -0,0 +1,678 @@
...
+
+// CreateNodePoolManager creates an NodePoolManager that can manage autoscaling node pools
+func CreateNodePoolManager(cloudConfigPath string, discoveryOpts cloudprovider.NodeGroupDiscoveryOptions, kubeClient kubernetes.Interface) (NodePoolManager, error) {
+
+	var err error
+	var configProvider common.ConfigurationProvider
+
+	if os.Getenv(ipconsts.OciUseInstancePrincipalEnvVar) == "true" || os.Getenv(npconsts.OkeUseInstancePrincipalEnvVar) == "true" {
+		klog.Info("using instance principal provider")
+		configProvider, err = auth.InstancePrincipalConfigurationProvider()
```
jlamillan, line 90: I'm using `InstancePrincipalConfigurationProviderForRegion` in the instance pool implementation and therefore _require_ `OCI_REGION` be set when `OCI_USE_INSTANCE_PRINCIPAL` is set to `true` (see [note in README](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/oci/README.md#environment-variables)).

If both implementations can use `InstancePrincipalConfigurationProvider` we should update the instance pool implementation so the implementations are consistent wrt configuring Instance Principals (and remove the note in the `README`).

cluster-autoscaler/cloudprovider/builder/builder_oci.go
```diff
@@ -36,7 +36,7 @@ const DefaultCloudProvider = cloudprovider.OracleCloudProviderName
 func buildCloudProvider(opts config.AutoscalingOptions, do cloudprovider.NodeGroupDiscoveryOptions, rl *cloudprovider.ResourceLimiter) cloudprovider.CloudProvider {
 	switch opts.CloudProviderName {
 	case cloudprovider.OracleCloudProviderName:
-		return oci.BuildOCI(opts, do, rl)
+		return instancepools.BuildOCI(opts, do, rl)
```
jlamillan, line 39: It _might_ be better to import `instancepools` aliased as "oci" so that we can keep the keep the return `cloudprovider.Build` format. 

i.e. 

```
import (
	oci "k8s.io/autoscaler/cluster-autoscaler/cloudprovider/oci/instancepools"
)
```
...
```
	case cloudprovider.OracleCloudProviderName:
		return oci.BuildOCI(opts, do, rl)
```

cluster-autoscaler/cloudprovider/oci/README.md
```diff
@@ -122,6 +140,10 @@ Configuration via environment-variables:
...
 
+#### Node Pool environment-variables
+
+XXX
```
jlamillan, line 145: Is `XXX` intended to mean that the node pool does not support any environment variables? Or was this a TODO that didn't get updated?

As far as I can tell, _both_ implementations support `OCI_USE_INSTANCE_PRINCIPAL` and `OCI_REFRESH_INTERVAL` environment variables. I added another comment asking whether if the call to `InstancePrincipalConfigurationProviderForRegion` in `oci_instance_pool_manager.go` could be replaced with `InstancePrincipalConfigurationProvider`. If so, the two implementations would behave similarly and the need for the `OCI_REGION` environment variable could go away. That would potentially leave only `OCI_USE_NON_POOL_MEMBER_ANNOTATION` as the only environment variable that only had use in one implementation. The others would be common to both.

cluster-autoscaler/cloudprovider/oci/nodepools/oci_manager.go
```diff
@@ -0,0 +1,678 @@
...
+	availabilityDomain := strings.Split(*np.NodeConfigDetails.PlacementConfigs[0].AvailabilityDomain, ":")[1]
+	return availabilityDomain, nil
+}
+
+func buildGenericLabels(nodePool *oke.NodePool, nodeName string, availabilityDomain string) map[string]string {
+	result := make(map[string]string)
+
+	// Check if this is an ARM node pool and label the nodes appropriately
+	if compiledArmRegexp.MatchString(*nodePool.NodeShape) {
+		result[kubeletapis.LabelArch] = npconsts.ArmArch
```
jlamillan, line 580: This function is _almost_ identical to `BuildGenericLabelsForInstancePool` in `oci_instance_pool_manager.go` except that the instance pool implementation appears to be missing support for checking for and adding the Ampere / ARM labels if the shape matches. 

Could you add the simple ARM shape check to the instance pool implementation and possibly move the `ArmArch` constant and regular expression into the common package?

cluster-autoscaler/cloudprovider/oci/nodepools/oci_manager.go
```diff
@@ -0,0 +1,678 @@
...
+						ErrorCode:    *node.NodeError.Code,
+						ErrorMessage: *node.NodeError.Message,
+					},
+				},
+			})
+
+			continue
+		}
+
+		switch node.LifecycleState {
```
jlamillan, line 393: If you could add the following line to the instance-pool implementation (`oci_instance_pool_manager.go` `GetInstancePoolNodes`) the behaviors would be consistent: 

```
@@ -261,6 +264,8 @@ func (m *InstancePoolManagerImpl) GetInstancePoolNodes(ip InstancePoolNodeGroup)
        for _, instance := range *instanceSummaries {
                status := &cloudprovider.InstanceStatus{}
                switch *instance.State {
+               case string(core.InstanceLifecycleStateStopped), string(core.InstanceLifecycleStateTerminated):
+                       klog.V(4).Infof("skipping instance is in stopped/terminated state: %q", *instance.Id)
                case string(core.InstanceLifecycleStateRunning):
                        status.State = cloudprovider.InstanceRunning
                case string(core.InstanceLifecycleStateCreatingImage):
```

cluster-autoscaler/cloudprovider/oci/nodepools/oci_manager.go
trungng92: I see. I was wondering why the OCI_REGION env var was required for instance principals and it's because of the difference in how the configuration provider is initialized. I'll update the instance principal implementation to also use this provider so that OCI_REGION is not needed.

jlamillan: Updated the `InstancePrincipalConfigurationProvider` in the Instance Pool implementation to be non region specific

cluster-autoscaler/cloudprovider/builder/builder_oci.go
jlamillan: Added 'oci' alias to Oracle Cloud import to preserve provider naming pattern.

cluster-autoscaler/cloudprovider/oci/OWNERS
trungng92: Discussed with Jesse Millan, I've added me and sohan-oracle to the list of reviewers/approvers (commented out)

cluster-autoscaler/cloudprovider/oci/README.md
trungng92: We've since updated this to have

```
#### Environment-variables applicable _only_ for Node Pools
n/a
```

jlamillan (approved):
Thanks Trung. You and I have discussed some followup work offline, which we can continue to iterate on (including looking into existing whether the implementations should leverage the External gRPC provider mechanism).

For now, this pull-request LGTM.
 

/lgtm

towca:
The OCI part was LGTMd by an OCI owner, the rest LGTM.

/approve

PR #5578, opened 2023-03-08
Add empty nodes sorting for scale down candidates

yaroslava-serdiuk:
#### What type of PR is this?
This is an optimisation for parallel scale down to check empty nodes first. Currently CA can simulate drain node by scheduling pods on empty nodes and so CA may prefer draining node over removing the empty one.

/kind feature

### DIFF
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -49,7 +49,9 @@ import (
 	ca_processors "k8s.io/autoscaler/cluster-autoscaler/processors"
 	"k8s.io/autoscaler/cluster-autoscaler/processors/nodegroupset"
 	"k8s.io/autoscaler/cluster-autoscaler/processors/nodeinfosprovider"
-	"k8s.io/autoscaler/cluster-autoscaler/processors/nodes"
+	"k8s.io/autoscaler/cluster-autoscaler/processors/scaledowncandidates"
+	"k8s.io/autoscaler/cluster-autoscaler/processors/scaledowncandidates/emptycandidates"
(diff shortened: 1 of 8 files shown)


cluster-autoscaler/main.go
```diff
@@ -402,11 +404,17 @@ func buildAutoscaler(debuggingSnapshotter debuggingsnapshot.DebuggingSnapshotter
...
-		opts.Processors.ScaleDownCandidatesNotifier.Register(sdProcessor)
+		sdCandidatesSorting := previouscandidates.NewPreviousCandidates()
+		scaleDownCandidatesComparers = []scaledowncandidates.CandidatesComparer{
+			emptycandidates.NewEmptySortingProcessor(&autoscalingOptions, emptycandidates.NewNodeInfoGetter(opts.ClusterSnapshot)),
+			sdCandidatesSorting,
+		}
+		opts.Processors.ScaleDownCandidatesNotifier.Register(sdCandidatesSorting)
 	}
+	sdProcessor := scaledowncandidates.NewScaleDownCandidatesSortingProcessor(scaleDownCandidatesComparers)
+	opts.Processors.ScaleDownNodeProcessor = sdProcessor
```
x13n, line 417: Shouldn't this happen only when parallel drain is enabled?

yaroslava-serdiuk: If parallel drain is false, we will have only PreFilteringScaleDownNodeProcessor (scaleDownCandidatesComparers will be empty list). 
I think we want to have PreFilteringScaleDownNodeProcessor for any case

PR #5583, opened 2023-03-10
ovhcloud: For any call to api.ovh.com returning a 500 error, retry call on ca.api.ovh.com

XavierDuthil:
#### What type of PR is this?

/kind bug

Should cover the cases where a Canadian project has been freshly created and OVHCloud Canadian authentication API has not yet synced with EU authentication API.

Retrying the same call on the Canadian API should successfully authenticate and succeed. This is a temporary fix until the issue is correctly handled.

NONE


mwielgus (approved):
/lgtm
/approve

PR #5584, opened 2023-03-10
ovhcloud: Fix deletion of multiple nodes at the same time

XavierDuthil:
#### What type of PR is this?

/kind bug

Adding a mutex on NodeGroup.DeleteNodes will prevent outdated/conflicting scale down payloads, which result in rejected calls.

```release-note
## OVHcloud
- Fix deletion of multiple nodes at the same time
```


mwielgus (approved):
/lgtm
/approve

PR #5585, opened 2023-03-10
ovhcloud: Fix panic on concurrent map read/write

XavierDuthil:
#### What type of PR is this?

/kind bug

Fixes a crash occurring in the ovhcloud provider code, due to a concurrent read/write access to a map.


mwielgus (approved):
/lgtm
/approve

PR #5589, opened 2023-03-10
Introduce a regular patch release schedule for CA

MaciekPytel:


MaciekPytel:
@gjtempleton @towca @x13n I wrote down the proposal I discussed with you earlier on. I assigned release dates randomly, please suggest changes if you plan to be unavailable during your shift.

/hold 
To get all the lgtms.

x13n:
/lgtm

That's a bit outside of the scope of what you're adding, but I think that the new section should also include info about which minors actually are in scope for getting patched at all.

MaciekPytel:
@x13n I think it already does - "Cluster Autoscaler releases patches for versions corresponding to currently
supported [Kubernetes versions](https://kubernetes.io/releases/)". The link is updated and contains a list of versions.

aaron-trout:
Can we describe in the process what is in-scope / out-of-scope for patch releases? 

I joined the sig-autoscaling channel/call recently to look at regular rebuilds/release for security reasons (vulnerabilities in the existing images) and this seems to go a long way towards fixing that already. I'm wondering if it's reasonable to also update go dependencies on the release branches when creating these patch releases (at least for those with HIGH+ CVEs maybe?)

MaciekPytel:
@aaron-trout That's a good point. A patch release would always be built using the newest base image, so it should address any vulnerabilities in the images. We could argue if 2-month latency is good enough, but t
Historically we haven't updated dependencies in patch releases, but we explicitly say additional releases may happen in response to vulnerabilities and if a serious vulnerability is escalated I'd imagine we'd be open to building a release in response (best escalation channel is probably via sig-meeting, either by participating or just adding to [agenda](https://docs.google.com/document/d/1RvhQAEIrVLHbyNnuaT99-6u9ZUMp7BfkPupT2LAZK7w/edit). 

Dependencies are more tricky - historically we haven't bumped them on patch releases, unless there was some specific bug that required us to do so. And vast majority of our dependencies are transitive dependencies of kubernetes, which we don't monitor directly. I don't think we want to get into the business of checking all those deps we're not even familiar with for vulns and even less so into solving any conflicts between kubernetes transitive deps and kubernetes itself. 
That being said I'd imagine kubernetes itself is much better at managing this than we're (they have a sig dedicated to security after all), so maybe committing to bumping k8s to the latest patch version (which in turn will bump any transitive deps) on any patch release would be a reasonable middle ground here?

@x13n @towca @gjtempleton thoughs?

x13n:
> @x13n I think it already does - "Cluster Autoscaler releases patches for versions corresponding to currently supported [Kubernetes versions](https://kubernetes.io/releases/)". The link is updated and contains a list of versions.

Of course I missed the first sentence of the diff :facepalm: Thanks, just lgtm from me then!

aaron-trout:
Sounds good. It may be worth doing this (checking for vulns and updating transitive deps) as a one off when implementing this new release policy since many of the images have HIGH / CRITICAL CVEs currently. For example here is a scan of the 1.24.0 image using trivy:

Screenshot 2023-03-06 at 18 02 24

I'm also happy to do some local builds off the release branches and scan the resulting images for comparison if that is helpful? I can raise some PRs to the release branches to attempt to fix any found issues.

mwielgus (approved):
/lgtm
/approve

PR #5590, opened 2023-03-10
[WIP] Improve ignore-taint naming

azylinski:
#### What type of PR is this?

/kind cleanup

Rename:
 * "ignore-taint" flag to "initialization-taint"
 * "ignore-taint.cluster-autoscaler.kubernetes.io/" special, hardcoded prefix into "initialization-resource-unready.cluster-autoscaler.kubernetes.io/"

The “ignore-taint” naming is confusing and even inaccurate. 
It's initial purpose was to mark those node that needs additional startup/initialization/installation before they are ready to run workloads (example: GPU Nvidia driver installation).

Current flag description: "Specifies a taint to ignore in node templates when considering to scale a node group".

The more precise description, would be something like: "Specify a list of special* initialization taints, would you be used to mark nodes which are not ready yet. Those nodes would be counted as `NotStarted` in CA clasterstate (to avoid unnecessary scale-ups). If node doesn't start in 15min - they are considered an unexpectedly Unready. Too many of those (max-total-unready-percentage flag, default to 45%) would stop the CA all together.

Those taints are skipped when creating/cloning a new node (on scale-up).

```release-note
A new "initialization-taint" flag, which would be 1-1 replacement for the "ignore-taint", which is meant to be Deprecated.
```

/assign MaciekPytel

### DIFF
--- a/cluster-autoscaler/config/autoscaling_options.go
+++ b/cluster-autoscaler/config/autoscaling_options.go
@@ -184,8 +184,8 @@ type AutoscalingOptions struct {
 	MaxBulkSoftTaintTime time.Duration
 	// MaxPodEvictionTime sets the maximum time CA tries to evict a pod before giving up.
 	MaxPodEvictionTime time.Duration
-	// IgnoredTaints is a list of taints to ignore when considering a node template for scheduling.
-	IgnoredTaints []string
+	// InitializationTaints is a list of taints to ignore when considering a node template for scheduling.
+	InitializationTaints []string
 	// BalancingExtraIgnoredLabels is a list of labels to additionally ignore when comparing if two node groups are similar.
 	// Labels in BasicIgnoredLabels and the cloud provider-specific ignored labels are always ignored.
 	BalancingExtraIgnoredLabels []string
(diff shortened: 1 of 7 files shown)


hterik:
> The more precise description, would be something like: "Specify a list of special* initialization taints, would you be used to mark nodes which are not ready yet. Those nodes would be counted as `NotStarted` in CA clasterstate (to avoid unnecessary scale-ups). If node doesn't start in 15min - they are considered an unexpectedly Unready. Too many of those (max-total-unready-percentage flag, default to 45%) would stop the CA all together.
> 
> Those taints are skipped when creating/cloning a new node (on scale-up).

Thanks for the explanation. Is this documented somewhere, without having to read the code? Even if the new name is a lot more self-explanatory, there are many important factors you just described, that are now only hidden in the code, that affect how a user of autoscaler would design their system, which should be more prominently described somewhere imo.

Any chance to change the hardcoded 15 mins to something configurable btw? 


Reviewing the full code of this change is out of my domain sorry.

sftim:
Suggestion for the changelog entry:
```diff
-A new "initialization-taint" flag, which would be 1-1 replacement for the "ignore-taint", which is meant to be Deprecated.
+Added a new `--initialization-taint` command line argument.
+The `--initialization-taint` argument is a direct replacement for `--ignore-taint`, which is now deprecated.
```

I'm not sure if `ignore-taint.cluster-autoscaler.kubernetes.io` is an annotation subdomain or similar? If it is part of a label key, annotation key, or taint, it should be registered in https://kubernetes.io/docs/reference/labels-annotations-taints/

PR #5593, opened 2023-03-14
docs: fix faq around expendable-pods-priority-cutoff

grosser:
#### What type of PR is this?

/kind documentation

https://github.com/kubernetes/autoscaler/blame/master/cluster-autoscaler/FAQ.md#L412 is wrong and confusing the default is -10 https://github.com/kubernetes/autoscaler/blob/63b334f81a8820511c835401e0133d242d84da08/cluster-autoscaler/main.go#L186 so no need for an extra step

```release-note
NONE
```


mwielgus (approved):
/lgtm
/approve

PR #5596, opened 2023-03-15
Add support for tags on AWS managed nodegroups to indicate resources

tbalzer:
#### What type of PR is this?

/kind feature

This PR introduces a feature that allows handling of ephemeral storage and custom resources in a scale-from-zero scenario when using EKS nodegroups.

It does this via allowing tags in the form of `k8s.io/cluster-autoscaler/resources/<resource>: <quantity>` analogous to how they are allowed directly on an ASG.
As the EKS-managed ASGs in case of managed nodegroups don't support custom tags for those resources they will be read from the managed nodegroup and processed accordingly.

NONE

I have modeled this after the implementation for ASGs and tried to be as clean as possible, passing through the Tags in the managed node group cache. I don't expect anybody will find it unexpected that these specific annotations influence the node templates.

As for documentation regarding this I didn't immediately find a place where one could mention that the ASG annotations also work for managed nodegroups, however I initially just assumed they would until I looked at the code.

```release-note
Allow k8s.io/cluster-autoscaler/resources/<resource>:<quantity> tags on AWS managed nodegroups (in addition to ASGs) to indicate availability of custom resources or ephemeral storage in a scale-from-zero scenario
```

NONE

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_manager.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_manager.go
@@ -308,6 +308,18 @@ func (m *AwsManager) buildNodeFromTemplate(asg *asg, template *asgTemplate) (*ap
 			node.Spec.Taints = append(node.Spec.Taints, mngTaints...)
 			klog.V(5).Infof("node.Spec.Taints : %+v\n", node.Spec.Taints)
 		}
+
+		mngTags, err := m.managedNodegroupCache.getManagedNodegroupTags(nodegroupName, clusterName)
+		if err != nil {
+			klog.Errorf("Failed to get tags from EKS DescribeNodegroup API for nodegroup %s in cluster %s because %s.", nodegroupName, clusterName, err)
+		} else if mngTags != nil && len(mngTags) > 0 {
+			resourcesFromMngTags := extractAllocatableResourcesFromTags(mngTags)
+			klog.V(5).Infof("Extracted resources from EKS nodegroup tags %v", resourcesFromTags)
+			// ManagedNodeGroup resource-indicating tags override conflicting tags on the ASG if they exist
+			for resourceName, val := range resourcesFromMngTags {
+				node.Status.Capacity[apiv1.ResourceName(resourceName)] = *val
+			}
+		}
 	}
 
 	node.Status.Conditions = cloudprovider.BuildReadyConditions()
@@ -448,6 +460,27 @@ func extractAllocatableResourcesFromAsg(tags []*autoscaling.TagDescription) map[
 			if label != "" {
 				quantity, err := resource.ParseQuantity(v)
 				if err != nil {
+					continue
+				}
+				result[label] = &quantity
+			}
+		}
+	}
+
+	return result
+}
+
+func extractAllocatableResourcesFromTags(tags map[string]string) map[string]*resource.Quantity {
+	result := make(map[string]*resource.Quantity)
+
+	for k, v := range tags {
+		splits := strings.Split(k, "k8s.io/cluster-autoscaler/node-template/resources/")
+		if len(splits) > 1 {
+			label := splits[1]
+			if label != "" {
+				quantity, err := resource.ParseQuantity(v)
+				if err != nil {
+					klog.Warningf("Failed to parse resource quanitity '%s' for resource '%s'", v, label)
 					continue
 				}
 				result[label] = &quantity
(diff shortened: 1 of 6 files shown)


tbalzer:
Hi @jaypipes @drmorr0,

Any change to get a review on this? We're willing to contribute more time into this as needed, as this missing feature prevented us from using scale-from-zero when having pending pods with ephemeral-storage requests on AWS with managed nodegroups and we think that's a perfectly valid usecase that's missing after the support for Labels/Taints being extracted from managed nodegroups in a scale-from-zero scenario.

Currently we're running our clusters with our self-hosted cluster-autoscaler image that has this included, but would obviously prefer to be able to use upstream again. So far I can say it's worked flawlessly.

cluster-autoscaler/cloudprovider/aws/aws_manager.go
```diff
@@ -458,6 +469,26 @@ func extractAllocatableResourcesFromAsg(tags []*autoscaling.TagDescription) map[
...
+	result := make(map[string]*resource.Quantity)
+
+	for k, v := range tags {
+		splits := strings.Split(k, "k8s.io/cluster-autoscaler/node-template/resources/")
+		if len(splits) > 1 {
+			label := splits[1]
+			if label != "" {
+				quantity, err := resource.ParseQuantity(v)
+				if err != nil {
+					continue
```
jaypipes, line 482: Did we want to log something here about an unparseable quantity string? I would think this might be something difficult to diagnose if there is no signal to the end user anywhere that their tag is malformed :)

cluster-autoscaler/cloudprovider/aws/aws_manager_test.go
```diff
@@ -103,6 +103,24 @@ func TestExtractAllocatableResourcesFromAsg(t *testing.T) {
 	assert.Equal(t, resource.NewQuantity(5, resource.DecimalSI).String(), labels["custom-resource"].String())
 }
 
+func TestExtractAllocatableResourcesFromTags(t *testing.T) {
```
jaypipes, line 106: perhaps worth adding a test case that stresses the error branch for ParseQuantity?

cluster-autoscaler/cloudprovider/aws/aws_manager.go
```diff
@@ -308,6 +308,17 @@ func (m *AwsManager) buildNodeFromTemplate(asg *asg, template *asgTemplate) (*ap
 			node.Spec.Taints = append(node.Spec.Taints, mngTaints...)
 			klog.V(5).Infof("node.Spec.Taints : %+v\n", node.Spec.Taints)
 		}
+
+		mngTags, err := m.managedNodegroupCache.getManagedNodegroupTags(nodegroupName, clusterName)
+		if err != nil {
+			klog.Errorf("Failed to get tags from EKS DescribeNodegroup API for nodegroup %s in cluster %s because %s.", nodegroupName, clusterName, err)
+		} else if mngTags != nil && len(mngTags) > 0 {
+			resourcesFromTags := extractAllocatableResourcesFromTags(mngTags)
```
jaypipes, line 316: This is shadowing the variable `resourcesFromTags` defined on line 275. That variable is used to populate resource capacities on lines 277-279 from the ASG template. I gather from your PR summary that you intend specifically for the values from the NodeGroupTags to override any values previously set by that code. Perhaps worth adding a code comment here indicating that if there is a Managed Node Group then its tags will override anything from an ASG? I think? :) Honestly, it's really not clear to me how Managed Node Groups are different from "normal" EC2 ASGs...

jaypipes (commented):
@tbalzer thanks for this :) Left a couple questions/suggestions inline for you but I think the code makes sense.

cluster-autoscaler/cloudprovider/aws/aws_manager_test.go
tbalzer: I have added a test case here that ensures an unparseable resource will be ignored in the output.

cluster-autoscaler/cloudprovider/aws/aws_manager.go
tbalzer: You are right, `resourcesFromTags` probably shouldn't be shadows so I renamed this variable to `resourcesFromMngTags`.
I also added a comment making it clear that if the same tag with different values were present on the ASG and the MNG the value on the MNG will take precedence.

tbalzer:
@jaypipes : Thanks for the feedback and attention. I've updated the code according to your suggestions and replaced the commit. PTAL again.

jaypipes (approved):
Yes, I'm happy with this. Thank you @tbalzer for your work! :)

tbalzer:
@jaypipes : Can you also add an lgtm here or should I ping someone else for this?

jaypipes:
> @jaypipes : Can you also add an lgtm here or should I ping someone else for this?

@tbalzer I'd prefer another approver have a quick look. @gjtempleton can you have a quick looksie? Thanks!

tbalzer:
@gjtempleton : Any chance to get a review? This is getting more and more stale.

gjtempleton:
@tbalzer apologies, I'll give this a review this week.

gjtempleton:
Thanks for this.

/lgtm

PR #5599, opened 2023-03-15
Implementation for KEP-4831: Control VPA eviction behavior based on scaling direction and resource

voelzmo:
#### What type of PR is this?
/kind feature

* Implements `ScalingDirectionPodEvictionAdmission`, which controls Pod admission to eviction based on `EvictionRequirements` defined in a VPA's `.spec.updatePolicy.evictionRequirements`
* Makes Updater use the new `ScalingDirectionPodEvictionAdmission` instead of `DefaultPodEvictionAdmission`
* Change e2e tests to use VPA builder that's already used in the unit tests
  * Changes to VPA builder to allow for multiple containers and more detailed settings per container 

Implements #4831
API PR done in #5176 

Depends on the API changes in #5176

```release-note
Added `.updatePolicy.evictionRequirements` to the VPA spec allowing to constrain Pod eviction based on scaling direction (scaling up or scaling down) and resource (cpu or memory).
```

```docs
- [AEP]: https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/enhancements/4831-control-eviction-behavior
```


voelzmo:
@jbartosik Can you please take a look? Thanks!

voelzmo:
@jbartosik Can you try to give this a review, please? Thanks!

vertical-pod-autoscaler/pkg/updater/priority/scaling_direction_pod_eviction_admission.go
```diff
@@ -0,0 +1,104 @@
...
+// Admit denies Pod eviction, if an EvictionRequirement exists for at least one Container which evaluates to 'false'.
+// If no EvictionRequirement exists or no Recommendation is present yet, the Pod is admitted for eviction.
+func (s *scalingDirectionPodEvictionAdmission) Admit(pod *apiv1.Pod, resources *vpa_types.RecommendedPodResources) bool {
+	podEvictionRequirements, found := s.EvictionRequirements[pod]
+	if !found {
+		return true
+	}
+	for _, container := range pod.Spec.Containers {
+		recommendedResources := vpa_utils.GetRecommendationForContainer(container.Name, resources)
+		if recommendedResources == nil {
```
jbartosik, line 45: This seems incorrect. This will admit pod if at least one container:
- doesn't have a recommendation or
- has a recommendation in the "correct" direction

I think this should return false if it finds at leas 1 container that has recommendation going in the wrong direction, default to true
?

vertical-pod-autoscaler/pkg/updater/priority/scaling_direction_pod_eviction_admission.go
```diff
@@ -0,0 +1,104 @@
...
+	if foundMemoryRequests && !s.allowEvictionForResourceAndScalingDirection(apiv1.ResourceMemory, requestedMemory.MilliValue(), recommendedMemory.MilliValue(), podEvictionRequirements) {
+		return false
+	}
+	return true
+}
+
+func (s *scalingDirectionPodEvictionAdmission) allowEvictionForResourceAndScalingDirection(resource apiv1.ResourceName, currentRequests int64, recommendation int64, podEvictionRequirements []*vpa_types.EvictionRequirement) bool {
+	for _, evictionRequirement := range podEvictionRequirements {
+		for _, resourceName := range evictionRequirement.Resources {
+			if resourceName == resource {
```
jbartosik, line 73: if resourceName != resource {
  continue
}

then we have 1 less indent level

jbartosik (changes requested):

vertical-pod-autoscaler/pkg/updater/priority/scaling_direction_pod_eviction_admission.go
jbartosik: Description of the function[1] and AEP[2] say that all `EvictionRequirement`s must be met to allow eviction. 

If that's not what you need then we can revisit the KEP (at least rename and have the description match implementation; I didn't think to much about if allowing eviction if at least one container matches is better than requiring all of them match restriction).

[1]
> Admit denies Pod eviction, if an EvictionRequirement exists for at least one Container which evaluates to 'false'.

[2]
[permalink](https://github.com/kubernetes/autoscaler/blob/d3ec0c4d049084bc5bb266a19448df0e82e3056a/vertical-pod-autoscaler/enhancements/4831-control-eviction-behavior/README.md?plain=1#L24):

> A single `EvictionRequirement` defines a condition which must be `true` to allow eviction for the corresponding `Pod`. When multiple `EvictionRequirements` are specified for a `Pod`, all of them must evaluate to `true` to allow eviction.

jbartosik (changes requested):

vertical-pod-autoscaler/pkg/updater/priority/scaling_direction_pod_eviction_admission.go
voelzmo: I think what the AEP mentions still holds true:
> A single EvictionRequirement defines a condition which must be true to allow eviction for the corresponding Pod. When multiple EvictionRequirements are specified for a Pod, all of them must evaluate to true to allow eviction.

In [`allowEvictionForResourceAndScalingDirection`](https://github.com/kubernetes/autoscaler/blob/dcec527d41e1604c64609ea6640833242db9f85a/vertical-pod-autoscaler/pkg/updater/priority/scaling_direction_pod_eviction_admission.go#L71) we iterate over all `EvictionRequirements` and if _one_ of the Requirements returns `false`, we return `false` for the entire Pod.

The critical point is _when_ an `EvictionRequirement` is `true` for a `Pod` based on the individual evaluations for the `Containers`.

I realize that the current comment is incorrect, I think I moved this one back and forth a few times
> Admit denies Pod eviction, if an EvictionRequirement exists for at least one Container which evaluates to 'false'.

This probably should be 
> Admit admits a Pod for eviction in one of three cases 
>  * an EvictionRequirement exists which is evaluated to 'true' for at least one Container in this Pod
>  * no Recommendation exists for at least one Container in this Pod
>  * no EvictionRequirement exists for this Pod

Does this make sense? Do we need a clarification in the AEP as well regarding what it means that an `EvictionRequirement` is true for a `Pod`?

jbartosik: Thanks for the explanation. I'll think about this a bit, I hope I'll be back with a proper reply on Monday

jbartosik: My understanding of the AEP:

- If there is only 1 eviction requirement, on 1 resource we don't evict unless one resource changes in the specified direction
- If there is only 1 eviction requirement, on both resources we don't evict unless at least one resource changed 
- If there are 2 eviction requirements, on both resources separately we don't evict unless both resources change in their specified directions

This code treats separate `EvictionRequirement`s  for CPU the way KEP describes putting them in a single `EvictionRequirement` (one of them changes in the specified direction, not necessarily both). This seems wrong to me

Also I think we should clarify in the AEP that "A single EvictionRequirement defines a condition which must be true to allow eviction for the corresponding Pod" means that at least one container has a change in the specified direction. This makes sense but got me confused for a while

jbartosik (changes requested):

vertical-pod-autoscaler/pkg/updater/priority/scaling_direction_pod_eviction_admission.go
voelzmo: Good catch, thanks! I think there was a test case missing that covered the scenario of a single `EvictionRequirement` with multiple resources, where only one of them was matched and the Pod was still admitted.
I added the case and fixed the code to make it green.

The other scenario ensuring that multiple `EvictionRequirements` are connected with `AND` was already in the test cases and still works.

I've also restructured the code a bit to match the comments and early exists for the cases in which we admit a Pod, hope this makes more sense now?

voelzmo:
Force-pushed an update:
* rebasing on most recent state of #5176 and adapting the test cases to https://github.com/kubernetes/autoscaler/pull/5896
* removing the `OrEqual` ChangeRequirements and test cases as discussed in https://github.com/kubernetes/autoscaler/pull/5754#issuecomment-1647577540

vertical-pod-autoscaler/pkg/updater/priority/scaling_direction_pod_eviction_admission.go
voelzmo: Opened PR https://github.com/kubernetes/autoscaler/pull/5989 to address this part of the comment 

> Also I think we should clarify in the AEP that "A single EvictionRequirement defines a condition which must be true to allow eviction for the corresponding Pod" means that at least one container has a change in the specified direction.

vertical-pod-autoscaler/e2e/v1/common.go
```diff
@@ -297,14 +297,15 @@ func SetupVPA(f *framework.Framework, cpu string, mode vpa_types.UpdateMode, tar
 
 // SetupVPAForNHamsters creates and installs a simple hamster VPA for a pod with n containers, for e2e test purposes.
 func SetupVPAForNHamsters(f *framework.Framework, n int, cpu string, mode vpa_types.UpdateMode, targetRef *autoscaling.CrossVersionObjectReference) *vpa_types.VerticalPodAutoscaler {
-	return SetupVPAForNHamstersWithMinReplicas(f, n, cpu, mode, targetRef, nil)
+	return SetupVPAForNHamstersWithMinReplicas(f, n, cpu, mode, targetRef, nil, nil)
 }
 
 // SetupVPAForNHamstersWithMinReplicas creates and installs a simple hamster VPA for a pod with n containers, setting MinReplicas. To be used for e2e test purposes.
-func SetupVPAForNHamstersWithMinReplicas(f *framework.Framework, n int, cpu string, mode vpa_types.UpdateMode, targetRef *autoscaling.CrossVersionObjectReference, minReplicas *int32) *vpa_types.VerticalPodAutoscaler {
+func SetupVPAForNHamstersWithMinReplicas(f *framework.Framework, n int, cpu string, mode vpa_types.UpdateMode, targetRef *autoscaling.CrossVersionObjectReference, minReplicas *int32, er []*vpa_types.EvictionRequirement) *vpa_types.VerticalPodAutoscaler {
```
jbartosik, line 304: Seems like we're never setting both minReplicas and er; Please:

- Add a new helper

or even better

- Add a builder (which will be cleaner to use)

vertical-pod-autoscaler/e2e/v1/updater.go
```diff
@@ -60,34 +60,104 @@ var _ = UpdaterE2eDescribe("Updater", func() {
...
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		}()
+		statusUpdater.Run(stopCh)
+
+		podList := setupPodsForDownscalingEviction(f, nil)
+
+		ginkgo.By("Waiting for pods to be evicted")
+		err := WaitForPodsEvicted(f, podList)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	})
```
jbartosik, line 97:
```suggestion
	})

```

vertical-pod-autoscaler/e2e/v1/updater.go
```diff
@@ -60,34 +60,104 @@ var _ = UpdaterE2eDescribe("Updater", func() {
...
 		statusUpdater.Run(stopCh)
 
-		podList := setupPodsForEviction(f)
+		podList := setupPodsForUpscalingEviction(f)
 
 		ginkgo.By("Waiting for pods to be evicted")
 		err := WaitForPodsEvicted(f, podList)
 		gomega.Expect(err).NotTo(gomega.HaveOccurred())
 	})
+	ginkgo.It("evicts pods for downscaling", func() {
```
jbartosik, line 69:
```suggestion

	ginkgo.It("evicts pods for downscaling", func() {
```

vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go
jbartosik, line 1: I'm reviewing this in #5176 and presume we'll not change this file in this PR

vertical-pod-autoscaler/e2e/v1/common.go
voelzmo: Thanks, that makes sense! I'll make the builders available in the e2e tests and push an update soon.

voelzmo:
force-pushed an update:
* rebased on current master that includes the merged API PR
* refactorings to expose VPA builder in e2e tests
* modifications to VPA builder such that it can replace the previous `SetupVPAForNHamstersWithMinReplicas` functions

Open points:
* one TODO in the VPA builder
* Simplifications/inlining for the `SetupVPAForNHamstersWithMinReplicas` functions

I can get to that next week

voelzmo:
@jbartosik This is good for another round of review from my side. I've addressed the last remaining comments and switched to using the VPABuilder in the e2e tests.

vertical-pod-autoscaler/e2e/v1/common.go
voelzmo: done.

vertical-pod-autoscaler/pkg/updater/priority/scaling_direction_pod_eviction_admission_test.go
```diff
@@ -0,0 +1,390 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package priority
+
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
```
jbartosik, line 20: This should be with other k8s imports

jbartosik (commented):
Mostly looks good, I'm leaving some small comments; I want to take another look at vertical-pod-autoscaler/pkg/updater/priority/scaling_direction_pod_eviction_admission_test.go before merging

vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go
```diff
@@ -153,7 +153,7 @@ func (calc *UpdatePriorityCalculator) GetSortedPods(admission PodEvictionAdmissi
 
 	result := []*apiv1.Pod{}
 	for _, podPrio := range calc.pods {
-		if admission == nil || admission.Admit(podPrio.pod, podPrio.recommendation) {
```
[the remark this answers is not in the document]
voelzmo: In the [same changeset](https://github.com/kubernetes/autoscaler/pull/5599/commits/4497aea552cc338e832dd5f389aae625eb545ac9) [there's a change to ensure that `admission` is always initialized and never `nil`](https://github.com/kubernetes/autoscaler/pull/5599/commits/4497aea552cc338e832dd5f389aae625eb545ac9).

[It is this `Updater` property that's passed into `GetSortedPods`](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/updater/logic/updater.go#L260), so I thought it should be safe to remove the nilcheck.

vertical-pod-autoscaler/e2e/v1/admission_controller.go
```diff
@@ -19,6 +19,7 @@ package autoscaling
 import (
 	"context"
 	"fmt"
+	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/test"
```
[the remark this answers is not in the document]
voelzmo: done.

vertical-pod-autoscaler/e2e/v1/recommender.go
```diff
@@ -19,6 +19,7 @@ package autoscaling
 import (
 	"context"
 	"fmt"
+	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/test"
```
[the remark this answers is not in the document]
voelzmo: done.

vertical-pod-autoscaler/e2e/v1/updater.go
```diff
@@ -19,6 +19,7 @@ package autoscaling
 import (
 	"context"
 	"fmt"
+	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/test"
```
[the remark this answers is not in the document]
voelzmo: done.

vertical-pod-autoscaler/pkg/updater/priority/scaling_direction_pod_eviction_admission_test.go
voelzmo: done.

jbartosik:
LGTM but needs rebase

voelzmo:
@jbartosik rebased on master, please take another look. Unfortunately, the bot removes LGTM labels on new changes ;(

voelzmo:
@jbartosik can we merge this before the next rebase is necessary? In kind of afraid having to go through a lot of files again when other changes get merged 😱
Thanks!

jbartosik (approved):
/lgtm
/approve

PR #5601, opened 2023-03-15
Sets max/min size for ASG to desired capacity when processes suspended

### LINKED ISSUE #5591: Cluster Autoscaler gracefully handles ASG launches being suspended
Cluster Autoscaler

Yes.

Scenario: Set up multiple nodegroups on EKS. A common scenario is creating one ASG per AZ, so let's say we have 3 nodegroups all in different AZs. If I'm seeing workloads having issue in one of the AZs, I could call [suspend-processes](https://docs.aws.amazon.com/cli/latest/reference/autoscaling/suspend-processes.html) on that ASG to avoid scaling up in the impacted AZ. As pods become unschedulable, CAS will scale up one of my ASGs. However, it could pick the ASG with processes suspended to scale, and those instances won't come up until processes are resumed, and it could pick it multiple times. I don't know if CAS will eventually pick another ASG to scale, but in my testing, it definitely doesn't happen within a short amount of time. While you could remove the ASG from CAS (removing ASG tags or whatever), it would be ideal for CAS to just handle the scenario gracefully.

When CAS chooses an ASG to scale, it should discard ASGs with launches suspended. `SuspendedProcesses` processes is returned in the `DescribeAutoScalingGroups` call, so we can tell if it's able to launch instances or not.

The alternative would be the current state: expect users to remove the ASG from CAS in this scenario.

**Additional context.**:

Currently looking at the code base and may update as my understanding of CAS improves.

mmerkes:
#### What type of PR is this?

/kind bug

When using the AWS cloud provider, CAS does not take into account whether or not the ASG can actually scale up or down. If you call [suspend-processes](https://docs.aws.amazon.com/cli/latest/reference/autoscaling/suspend-processes.html) on the ASG, you can disable launching or terminating instances, and when that happens, CAS may trigger a scaling action (and do some continually) when ASG will never actually launch or terminate instances.

In this PR, I made a change to set the `MinSize` to the desired size if terminate is suspended for an ASG and `MaxSize` to the desired size if launch is suspended for an ASG. This results in CAS ignoring the ASG when choosing who to scale. While this is a little indirect, it provides a simple solution to the problem.

Fixes #5591

I did deploy this change to an EKS 1.24 cluster and verified that it worked as expected.

I started with 2 ASGs, one with all processes suspended and the other one untouched and verified that only the one ASG overrode the min/max sizes:

```
I0315 18:28:10.064281       1 auto_scaling_groups.go:514] Setting min size to desired size for eks-ng-ee55ae76-1cc36661-4ed9-07f9-f199-699fc0224ce0 due to suspended Terminate process
I0315 18:28:10.064300       1 auto_scaling_groups.go:525] Setting max size to desired size for eks-ng-ee55ae76-1cc36661-4ed9-07f9-f199-699fc0224ce0 due to suspended Launch process
I0315 18:28:10.064314       1 auto_scaling_groups.go:161] Registering ASG eks-ng-ee55ae76-1cc36661-4ed9-07f9-f199-699fc0224ce0
I0315 18:28:10.064324       1 auto_scaling_groups.go:161] Registering ASG eks-second-58c36ddd-ab39-ea43-4f02-7ed9233931c9
```

Also, verified that the ASG cache gets regenerated regularly, as expected.

Then, I scaled pods and verified that it results in the suspended processes ASG to get ignored:

```
I0315 18:32:52.774186       1 scale_up.go:231] Skipping node group eks-ng-ee55ae76-1cc36661-4ed9-07f9-f199-699fc0224ce0 - max size reached
```

Then, I cleared the suspended processes, scaled up pods again and verified that the ASG goes back into circulation and was eventually picked for scaling.

```release-note
Fixed AWS cloudprovider to treat nodegroups as unscalable if relevant processes are suspended in the underlying Auto Scaling Group.
```

None

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
+++ b/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
@@ -495,11 +495,45 @@ func (m *asgCache) isNodeGroupAvailable(group *autoscaling.Group) (bool, error) 
 	return true, nil
 }
 
+func isProcessSuspended(g *autoscaling.Group, name string) bool {
+	if g.SuspendedProcesses != nil && len(g.SuspendedProcesses) > 0 {
+		for _, p := range g.SuspendedProcesses {
+			if *p.ProcessName == name {
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
+func getMinSize(g *autoscaling.Group) int {
+	// If the Terminate process is suspended, we set min size to the current
+	// size to avoid trying to scale it down.
+	if isProcessSuspended(g, "Terminate") {
+		klog.V(4).Infof("Setting min size to desired size for %s due to suspended Terminate process", *g.AutoScalingGroupName)
+		return int(aws.Int64Value(g.DesiredCapacity))
+	}
+
+	return int(aws.Int64Value(g.MinSize))
+}
+
+func getMaxSize(g *autoscaling.Group) int {
+	// If the Launch process is suspended, we set max size to the current
+	// size to avoid trying to scale it up.
+	if isProcessSuspended(g, "Launch") {
+		klog.V(4).Infof("Setting max size to desired size for %s due to suspended Launch process", *g.AutoScalingGroupName)
+		return int(aws.Int64Value(g.DesiredCapacity))
+	}
+
+	return int(aws.Int64Value(g.MaxSize))
+}
+
 func (m *asgCache) buildAsgFromAWS(g *autoscaling.Group) (*asg, error) {
 	spec := dynamic.NodeGroupSpec{
 		Name:               aws.StringValue(g.AutoScalingGroupName),
-		MinSize:            int(aws.Int64Value(g.MinSize)),
-		MaxSize:            int(aws.Int64Value(g.MaxSize)),
+		MinSize:            getMinSize(g),
+		MaxSize:            getMaxSize(g),
 		SupportScaleToZero: scaleToZeroSupported,
 	}
 
--- a/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups_test.go
+++ b/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups_test.go
@@ -45,6 +45,88 @@ func TestBuildAsg(t *testing.T) {
 	assert.Error(t, err)
 }
 
+func TestBuildAsgFromAWSSizing(t *testing.T) {
+	defaultMin, defaultDesired, defaultMax := 1, 3, 5
+	cases := []struct {
+		name               string
+		expectedMin        int
+		expectedMax        int
+		suspendedProcesses []*autoscaling.SuspendedProcess
+	}{
+		{
+			name:        "No suspended processes should set match ASG settings",
+			expectedMin: defaultMin,
+			expectedMax: defaultMax,
+		},
+		{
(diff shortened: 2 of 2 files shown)


x13n:
/assign

I think using min/max for this is not the best approach, because:
- These nodes will still be used by CA simulations and only then filtered out as infeasible for scaling, which wastes resources.
- If a node crashes, CA may still trigger a scale up when there are pending pods, since the current size will suddenly become lower than the desired size.

Alternatively, returning `nil` from `NodeGroupForNode` ([here](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go#LL109C37-L109C37)) will prevent CA from considering the node to be a part of an autoscaled node group at all, which I think is what you want.

x13n:
Btw, this is why it works: https://github.com/kubernetes/autoscaler/blob/57bd2bca97ac80ceb0fe92f604f01a7b0afa4f9c/cluster-autoscaler/core/utils/utils.go#L71

mmerkes:
@x13n Tx for taking a look and providing more context.

> These nodes will still be used by CA simulations and only then filtered out as infeasible for scaling, which wastes resources.

While that's true, it's an improvement over the current behavior as the result will be better. I didn't filter out the impacted ASGs because you could suspend `Launch` and not `Terminate`, or visa-versa, though if you suspend both, it would probably make sense to remove.

> If a node crashes, CA may still trigger a scale up when there are pending pods, since the current size will suddenly become lower than the desired size.

That's true, though it's a race condition of a race condition, and given that the cache refreshes every 1 minute, CA won't ever guarantee that it's making the right decision 100% of the time.

> Alternatively, returning nil from NodeGroupForNode ([here](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go#LL109C37-L109C37)) will prevent CA from considering the node to be a part of an autoscaled node group at all, which I think is what you want.

Tx for the suggestion. I'll check this out. That will handle the node crashing case, but will that stop it from scaling the ASG?

x13n:
> While that's true, it's an improvement over the current behavior as the result will be better. I didn't filter out the impacted ASGs because you could suspend `Launch` and not `Terminate`, or visa-versa, though if you suspend both, it would probably make sense to remove.
 
Ah, good point, I actually didn't understand what suspend-process in AWS really does. :sweat_smile:  If you can disable VM creation/termination separately, then overriding min/max separately makes more sense. It still feels a bit hacky - ideally, suspending `Launch` would remove the node group from scale up logic completely, while suspending `Terminate` would remove the node group from scale down. Both could be alternatively achieved by implementing `NodeGroupListProcessor` and `ScaleDownNodeProcessor`, respectively. Cloud provider specific processors are not easily introduce-able when we have a shared binary for all cloud providers. I'd be interesting on aws owners' thoughts - @gjtempleton @drmorr0 
 
> Tx for the suggestion. I'll check this out. That will handle the node crashing case, but will that stop it from scaling the ASG based on pods scaling up or scaling down?

It would work well if one wanted to disable scaling completely - CA would consider the entire node group as not autoscaled. (You'd also need to filter it out from `NodeGroups()` call too.) Since you want to separately control `Launch` and `Terminate`, I admit it is not the way to go.

x13n:
This doesn't seem to be worked on, please reopen if needed.

/close

PR #5603, opened 2023-03-17
baseURI for skuClient based on cloud

gandhipr:
#### What type of PR is this?
/kind bug

This change provides correct endpoint/baseURI to the skuClient based on the cloud environment.
By default, endpoint always pointed to public cloud. If the environment was not cloud, sku Cache creation fails and CA pod crashes.

Fixes #

yes, CA pod will no longer crash.

```release-note
Cluster autoscaler will work correctly without crash in non-public clouds.
```

```docs

```


tallaxes (approved):
LGTM

Issue #5240, opened 2022-10-07
Behavior change in daemonset pod eviction


jdn5126:
**Which component are you using?**:
cluster-autoscaler

**What version of the component are you using?**:
v1.21.3

Component version: v1.21.3

**What k8s version are you using (`kubectl version`)?**:

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"23+", GitVersion:"v1.23.7-eks-4721010", GitCommit:"b77d9473a02fbfa834afa67d677fd12d690b195f", GitTreeState:"clean", BuildDate:"2022-06-27T22:22:16Z", GoVersion:"go1.17.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"22+", GitVersion:"v1.22.13-eks-15b7512", GitCommit:"94138dfbea757d7aaf3b205419578ef186dd5efb", GitTreeState:"clean", BuildDate:"2022-08-31T19:15:48Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What environment is this in?**:
AWS EKS

**What did you expect to happen?**:
Daemonset pods, especially those marked `system-node-critical`, to either not be evicted by default or to be evicted only after all other pods are evicted.

**What happened instead?**:
Daemonset pods were evicted before other non-critical and non-daemonset pods.

**How to reproduce it (as minimally and precisely as possible)**:
1. Deploy daemonset pod without the `cluster-autoscaler.kubernetes.io/enable-ds-eviction` annotation and without `--daemonset-eviction-for-occupied-nodes=false`.
2. Deploy other pods on node with high termination grace period
4. Schedule node eviction
5. See that daemonset pod is evicted before other pods

**Anything else we need to know?**:
I understand that `cluster-autoscaler.kubernetes.io/enable-ds-eviction` and `--daemonset-eviction-for-occupied-nodes` were added to control daemonset pod eviction, but the default behavior on occupied nodes changed from "do not evict daemonset pods" to "evict daemonset pods unless annotation is present". This behavior change can lead to regressions and unexpected changes.

If the default behavior cannot be changed back, it seems like daemonset pods, or even more specifically `system-node-critical` pods, should not be evicted until all other pods are evicted. Adding issue #4337 for referencing similar discussion

jdn5126:
Any thoughts on this?

[...]

PR #5607, opened 2023-03-20
Only evict ds pods after other pods are evicted

jacobwolfaws:
#### What type of PR is this?
/kind bug

Optionally add one or more of the following kinds if applicable:
/kind regression
-->

Daemonset pods are sometimes being evicted before non-ds pods, which in turn can prevent the other pods from terminating properly: https://github.com/kubernetes/autoscaler/issues/5240 . This would change the eviction behavior to evict/delete all non-DS pods before evicting DS pods. 

Note: didn't add issue to fixes because this doesn't change default behavior to non-evict

Fixes # 

open questions:
1. In the original code the waiter doesn't watch for successful daemonset pod deletion, only the non-daemonset pods. Do we want to add one just to prevent some other race condition from potentially emerging?
2. This behavior will differ slightly from the original code in the sense that daemonset pods are only evicted if non-ds pods are evicted successfully. This makes sense to me, do others agree?
3. It seems like it would be preferable for the default daemonset eviction to be false, but that’s not included in this PR (I'd be happy to force push a change to include this behavior). Would love to hear people’s thoughts on this.
4. Can I get some help in testing this to confirm satisfactory performance?

```release-note
When nodes are drained on scale down, non-Daemonset pods are evicted prior to Daemonset pods.
```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go
+++ b/cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go
@@ -30,10 +30,10 @@ import (
 	apiv1 "k8s.io/api/core/v1"
 
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider/hetzner/hcloud-go/hcloud"
+	"k8s.io/autoscaler/cluster-autoscaler/version"
 )
 
 var (
-	version    = "dev"
 	httpClient = &http.Client{
 		Transport: instrumentedRoundTripper(),
 	}
@@ -71,6 +71,7 @@ func newManager() (*hetznerManager, error) {
 	client := hcloud.NewClient(
 		hcloud.WithToken(token),
 		hcloud.WithHTTPClient(httpClient),
+		hcloud.WithApplication("cluster-autoscaler", version.ClusterAutoscalerVersion),
 	)
 
 	ctx := context.Background()


jacobwolfaws:
```
 --- FAIL: TestDrainNodeWithPodsDaemonSetEvictionFailure (0.00s)
    drain_test.go:396: 
        	Error Trace:	/home/runner/work/autoscaler/autoscaler/go/src/k8s.io/autoscaler/cluster-autoscaler/core/scaledown/actuation/drain_test.go:396
        	Error:      	Not equal: 
        	            	expected: 2
        	            	actual  : 4
        	Test:       	TestDrainNodeWithPodsDaemonSetEvictionFailure
```
I think this is because I changed the behavior from `always evict daemonset pods when eviction is true` to `evict daemonset pods only when non-ds pod eviction was successful`. Would appreciate confirmation on that + highlights open question 2.

MaciekPytel:
I'm worried that this may fix one issue, but create a different one - I don't think there is any delay between daemonset eviction and deleting VM after your changes. I'm worried that will break anyone who has daemonsets that need some graceful termination, by effectively reducing the time ds have available to cleanup.

An obvious solution would be to wait for ds evictions, but that has its own problems:
 * Daemonsets often come with wildcard tolerations, so a ds may be rescheduled during CA drain effectively getting the drain stuck.
 * Sequential, blocking evictions will decrease scale-down speed.

I'm not sure if there is any solution here that just works for everyone and IIRC we've caused regressions for a lot of people last time we changed daemonset drain handling. So I'd be mindful of backward compatibility and consider if the changes are obviously positive in all use-cases; otherwise we may want to flag guard defaulting to old behavior at least for a few versions.

jacobwolfaws:
> I'm worried that will break anyone who has daemonsets that need some graceful termination, by effectively reducing the time ds have available to cleanup.

I think this is a valid concern, but this 'race condition' for lack of a better term, already technically exists within the existing code, since the ds pod eviction isn't being waited on, so this change would just exacerbate its occurence rather than introduce an entirely new behavior.

> Daemonsets often come with wildcard tolerations, so a ds may be rescheduled during CA drain effectively getting the drain stuck.

Can you elaborate on this/why it wouldn't occur with the current code/implementation

> Sequential, blocking evictions will decrease scale-down speed.
We could add a flag for non-sequential evictions? It seems like defaulting for no race conditions would be better than defaulting for faster evictions

Issue #3879, opened 2021-02-10
cluster-autoscaler : KubeSchedulerConfiguration plugin configuration PodTopologySpread


Ramyak:
**Which component are you using?**:

component: cluster-autoscaler

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:

Scheduler now supports [PodTopologySpread - Cluster-level default constraints](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/#cluster-level-default-constraints) from kubernetes release 1.18 - [commit](https://github.com/kubernetes/kubernetes/commit/1b4b1553328892e1e5486b5151f423ef78668b52).

**1. PodTopologySpread - `defaultConstraints` at the cluster level**: Cluster-autoscaler does not consider PodTopologySpread `defaultConstraints` at the cluster level. 

```
apiVersion: kubescheduler.config.k8s.io/v1alpha2
kind: KubeSchedulerConfiguration
leaderElection:
  leaderElect: true
profiles:
  - pluginConfig:
      - name: PodTopologySpread
        args:
          defaultConstraints:
            - maxSkew: 1
              topologyKey: topology.kubernetes.io/zone
              whenUnsatisfiable: ScheduleAnyway
          defaultingType: List
```
Pods remain unscheduled. You get the error.
Note: Pod specs do not have `topologySpreadConstraints` in this case.
```
I0210 18:03:11.409934       1 filter_out_schedulable.go:118] Pod test-app-5b75d455c9-7gpf5 marked as unschedulable can be scheduled on node ip-172-21-145-192.ec2.internal (based on hinting). Ignoring in scale up.
```

**2. PodTopologySpread when set at deployment**: works since pod spec starts having topologySpreadConstraints.
```
  topologySpreadConstraints:
  - labelSelector:
      matchLabels:
        app: some-app
        release-unixtime: "1611668612"
    maxSkew: 1
    topologyKey: topology.kubernetes.io/zone
    whenUnsatisfiable: DoNotSchedule
```

**Describe the solution you'd like.**:

Cluster-autoscaler consider [PodTopologySpread - Cluster-level default constraints](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/#cluster-level-default-constraints) during attempts to schedule pods

**Describe any alternative solutions you've considered.**:

**Additional context.**:

vadasambar:
[...]

Issue #1406, opened 2018-11-15
Getting the CA to play well with a custom scheduler


consideRatio:
I opened this issue: https://github.com/kubernetes/kubernetes/issues/71070

I understand it as that the Cluster Autoscaler is utilizing hardcoded logic from the default scheduler. Perhaps it would it be possible in the future to avoid hardcoding this but instead cooperating with the scheduler for the pod?

MaciekPytel:
There were many discussions of this in the past. The problem is CA doesn't actually know anything about scheduling. It's built around importing scheduler code and using it as a sort of black box oracle. All CA decisions are based on simulations, performed by creating in memory node objects and feeding them to scheduler to see if currently pending pods would be able to schedule if a new node was added. 

There were some discussions about exposing a 'dry run' API in scheduler. The problem is we need to run a ton of those scenarios, using some imaginary nodes, pretending some nodes don't exist or pretending some pods are running on different pods than they really are. Scheduler doesn't support any of that and even if it did we believe the performance impact of serializing all those objects and sending them to scheduler is prohibitive (we would need to make thousands of such requests per loop). 

To sum up full support for custom scheduling would require massive changes in both scheduler and CA and we don't think it would work anyway. I think it's safe to say it's not likely to happen for general case.

Now, some forms of custom scheduling are fine - if you only touch priority functions, not predicates, than there is no problem, as CA ignores those anyway. If all you do is change scheduler config it should be possible to expose analogous config options in CA that would be forwarded to imported scheduler code. Finally if you add a new predicate function, you should be able to import it into CA, build your own image and use that.

consideRatio:
@MaciekPytel this was an excellent write up! Thank you!!

PR #5611, opened 2023-03-24
leave a buffer of underutilized nodes when scaling down

### LINKED ISSUE #5377: support overprovsioning without pending pods

cluster-autoscaler

we use pending pods for overprovisioning atm, but that results in there not being "real" gaps,
so when an az preferred workload wants to get scheduled it always goes to the open az instead of kicking out an overprovisioned pod

make CA pretend there are pending pods when doing the math and scale based on that

current [recommended approahc](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-configure-overprovisioning-with-cluster-autoscaler)

**Additional context.**:

https://github.com/kubernetes/autoscaler/issues/4384 describes the same issue but lacks context

/cc @mattyev87 @MaciekPytel

grosser:
#### What type of PR is this?

/kind feature

- When scheduling nodes with topologySpreadConstraints ScheduleAnyway the scheduler does not evict capacity buffers and creates skew
- When trying to de-schedule skewed pods descheduler noops when there is no empty space
... so allow users to opt-in to empty space

This is not perfect since it will not create "new empty space" by scaling up, but I think it's a good step forward and allows us to fix some edge-cases that capacity buffer does not solve.

Fixes #5377

```release-note
New --scale-down-buffer-ratio flag for ratio of empty or underutilized nodes to leave as capacity buffer per nodegroup
```

### DIFF
--- a/cluster-autoscaler/config/autoscaling_options.go
+++ b/cluster-autoscaler/config/autoscaling_options.go
@@ -149,6 +149,8 @@ type AutoscalingOptions struct {
 	// The formula to calculate additional candidates number is following:
 	// max(#nodes * ScaleDownCandidatesPoolRatio, ScaleDownCandidatesPoolMinCount)
 	ScaleDownCandidatesPoolMinCount int
+	// ScaleDownBufferRatio Ratio of empty or underutilized nodes to leave as capacity buffer per nodegroup
+	ScaleDownBufferRatio float64
 	// ScaleDownSimulationTimeout defines the maximum time that can be
 	// spent on scale down simulation.
 	ScaleDownSimulationTimeout time.Duration
--- a/cluster-autoscaler/core/scaledown/unneeded/nodes.go
+++ b/cluster-autoscaler/core/scaledown/unneeded/nodes.go
@@ -17,6 +17,7 @@ limitations under the License.
 package unneeded
 
 import (
+	"math"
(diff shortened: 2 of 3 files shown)


vadasambar:
Appreciate the PR (though I am not sure I fully understand the problem yet) :+1: 

> When scheduling nodes with topologySpreadConstraints ScheduleAnyway the scheduler does not evict capacity buffers and creates skew

This is expected. CA doesn't support `ScheduleAnyway` because it is a part of Scoring phase of scheduler. CA only simulates Filter phase (`PreFilter` and `Filter` extension points to be precise) and that is so by design. 
You might be interested in
- https://github.com/kubernetes/autoscaler/issues/1406#issuecomment-439066582
- https://github.com/kubernetes/autoscaler/issues/3879#issuecomment-1480661410

grosser:
I know that it's not supported, that's why I made this PR to make it somewhat supported, it's not perfect but might be good enough ... I'll have to test it in our clusters to know more, but wanted to share the approach in case anyone else finds it useful or has inout on how to make it better.

grosser:
this worked but was not reliable enough (sometimes leaves half a node empty, sometimes full nodes) and cost accounting was not great either (cannot differentiate between bad binpacking and intentional gaps)

PR #5621, opened 2023-03-27
Make the "OOMing pods under VPA" test more robust.

kgolab:
#### What type of PR is this?

/kind flake

After #5326 the VPA test in "OOMing pods under VPA" in full_vpa.go became unstable when ran in GKE.

I debugged it a bit and I can see that in the set-up used by the test VPA is reacting to both OOMs and regular memory samples which kind of defeats the purpose of the test.
IIUC the whole idea of the test is based on how fast VPA would scale on average - multiple scale ups are required to reach 1.4 GB. This makes the test brittle as the VPA reaction time is not specified and could vary.

This PR attempts to create a more deterministic test:
* set memory limit to 1024 MiB,
* allocate 1025 MiB worth of memory in one go to significantly decrease a chance of a regular sample,
* expect a recommendation taking into account both the OOM bump (+20% by default) and the usual 15% margin, to make sure at least one OOM was taken into account,
* never allocate more than 1025 MiB worth of memory to decrease a chance of getting to the expected recommendation w/o OOMs.

It's definitely not perfect since I don't see a way to make 100% sure that there are no regular memory samples but I think that decreasing the number of moving parts will make the test more robust.

However I ran the stress command from the test for over 1 hour in my cluster and I haven't seen any regular memory sample for these containers.

Related to #5326 & thus also #4981

No.

### DIFF
--- a/vertical-pod-autoscaler/e2e/v1/autoscaling_utils.go
+++ b/vertical-pod-autoscaler/e2e/v1/autoscaling_utils.go
@@ -67,7 +67,6 @@ const (
 
 var (
 	resourceConsumerImage = imageutils.GetE2EImage(imageutils.ResourceConsumer)
-	stressCommand         = []string{"/stress", "--mem-total", "10000000000", "--logtostderr", "--mem-alloc-size", "50000"}
 )
 
 var (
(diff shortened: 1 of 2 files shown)


jbartosik (changes requested):
Mostly looks god to me but there is a formatting error, please fix:

```
Please run hack/update-gofmt.sh to fix the following files:
./vertical-pod-autoscaler/e2e/v1/autoscaling_utils.go
FAILED
```

kgolab:
Let me attempt to fix this today, if not successful let's just revert it.

PR #5632, opened 2023-03-29
fix: implement function to identify if node is present in aws

### LINKED ISSUE #4456: CA fails to scale-up or cancel in-progress scale down when there are un-schedulable pods

vadasambar:
- this is a follow-up to https://github.com/kubernetes/autoscaler/pull/5054
- this might fix https://github.com/kubernetes/autoscaler/issues/4456

#### What type of PR is this?

/kind bug

I think description of https://github.com/kubernetes/autoscaler/pull/5054#issue-1319989728 explains it well:

> ...original intent of determining the deleted nodes was incorrect, which led to the issues reported by other users. The nodes tainted with ToBeDeleted were misidentified as Deleted instead of Ready/Unready, which caused a miscalculation of the node being included as Upcoming. This caused problems described in https://github.com/kubernetes/autoscaler/issues/3949 and https://github.com/kubernetes/autoscaler/issues/4456.

https://github.com/kubernetes/autoscaler/pull/5054 exposed interface function `HasInstance` for cloud providers to implement. This function is used to determine if a node is deleted by querying the cloud provider. It falls back to the `ToBeDeleted` taint if the function is not implemented. 

Fixes #4456

Nothing in particular.

```release-note
NONE
```

```docs
Nothing in particular.
```

cc: @fookenc

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
@@ -128,8 +128,19 @@ func (aws *awsCloudProvider) NodeGroupForNode(node *apiv1.Node) (cloudprovider.N
 }
 
 // HasInstance returns whether a given node has a corresponding instance in this cloud provider
-func (aws *awsCloudProvider) HasInstance(*apiv1.Node) (bool, error) {
-	return true, cloudprovider.ErrNotImplemented
+func (aws *awsCloudProvider) HasInstance(node *apiv1.Node) (bool, error) {
+	awsRef, err := AwsRefFromProviderId(node.Spec.ProviderID)
+	if err != nil {
+		return false, err
+	}
+
+	// we don't care about the status
+	status, err := aws.awsManager.asgCache.InstanceStatus(*awsRef)
+	if status != nil {
+		return true, nil
+	}
+
+	return false, fmt.Errorf("node is not present in aws: %v", err)
 }
 
 // Pricing returns pricing model for this cloud provider or error if not available.


cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -128,8 +128,18 @@ func (aws *awsCloudProvider) NodeGroupForNode(node *apiv1.Node) (cloudprovider.N
...
-func (aws *awsCloudProvider) HasInstance(*apiv1.Node) (bool, error) {
-	return true, cloudprovider.ErrNotImplemented
+func (aws *awsCloudProvider) HasInstance(node *apiv1.Node) (bool, error) {
+	awsRef, err := AwsRefFromProviderId(node.Spec.ProviderID)
+	if err != nil {
+		return false, err
+	}
+
+	// we don't care about the instance health
+	if _, ok := aws.awsManager.asgCache.instanceStatus[*awsRef]; !ok {
```
vadasambar, line 138: We already have `instanceStatus` field on `asgCache` to track health status by EC2 instance. I am re-using the `instantStatus` map to check if the node is present without caring for its health since we just want to check if the instance is present in the cloud provider. 

`instanceStatus` is updated at https://github.com/kubernetes/autoscaler/blob/1d56b81b578983b27dc8024defe9561605bd3273/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go#L401-L404

jaypipes, line 138: @vadasambar please use the [`asgCache.InstanceStatus` function](https://github.com/kubernetes/autoscaler/blob/e29dff136fabc9ecf8b3efb03cd7fb847b7686d4/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go#L231-L240) instead of directly accessing (in a non-thread-safe fashion) the underlying map :)

jaypipes (changes requested):

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
vadasambar: Thank you for noticing it! I have updated the code.

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -128,8 +128,19 @@ func (aws *awsCloudProvider) NodeGroupForNode(node *apiv1.Node) (cloudprovider.N
...
-	return true, cloudprovider.ErrNotImplemented
+func (aws *awsCloudProvider) HasInstance(node *apiv1.Node) (bool, error) {
+	awsRef, err := AwsRefFromProviderId(node.Spec.ProviderID)
+	if err != nil {
+		return false, err
+	}
+
+	// we don't care about the status
+	status, err := aws.awsManager.asgCache.InstanceStatus(*awsRef)
+	if status != nil {
```
vadasambar, line 139: I am checking `status` for nil because I think this is a more reliable way (as opposed to checking there was no error) to check if the node is present since `InstanceStatus` wouldn't return nil for a `status` unless some error occurs.

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -128,8 +128,19 @@ func (aws *awsCloudProvider) NodeGroupForNode(node *apiv1.Node) (cloudprovider.N
...
+		return true, nil
+	}
+
+	return false, fmt.Errorf("node is not present in aws: %v", err)
```
vadasambar, line 143: I believe this CA should print the line number of the error but I added an extra wrapper around the error just in case to identify the error came from `HasInstance`. Happy to change it based on feedback.

vadasambar:
> +1 thank you @vadasambar!

Thank you for the review. I have squashed the commits into a single commit. I think I can't merge the PR by myself.

This is what I see:
image

zaafar:
Can this be back ported to 1.24 and 1.25?

Issue #5324, opened 2022-11-22
Make use of Capabilities.KubeVersion so we do not need to override image.tag in helm


miroljub1995:
**Which component are you using?**:
cluster-autoscaler (helm)

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:
no

**Describe the solution you'd like.**:
It would be great if image tag could be inferred based on  Capabilities.KubeVersion so that we do not need to override image.tag in Chart Values. Right now default tag is v1.23.0, but if we run on k8s v1.24, we need to override it to v1.24.0.

**Describe any alternative solutions you've considered.**:
The only alternative way is to set image.tag explicitly.

khizunov:
@miroljub1995 Looks like cool feature, but the KubeVersion will not necessary match the autoscaler image tag.

Shubham82:
@gjtempleton your thoughts on this?

Shubham82:
if this seems valid, then I will give it a try.
/assign

gjtempleton:
Hey, thanks for the suggestion, as @khizunov says, the KubeVersion won't necessarily match the image tag, particularly the patch version, as well as the lag between a new minor release of k8s being cut and the relevant CA image being released/promoted.

If we were to support this, I certainly wouldn't want it being the default behaviour unless we had a reliable way of fetching the latest patch release for the detected minor version and using that as an input.

Shubham82:
Hi @gjtempleton

> If we were to support this, I certainly wouldn't want it being the default behaviour unless we had a reliable way of fetching the latest patch release for the detected minor version and using that as an input.

Could please give any pointer or lead how we can do this?

Shubham82:
Hi @gjtempleton 
I saw your comment on the issue: #5640. I agree with your point.

Is there any way or pointer to start with this?

Also, we need the same minor version of k8s for CA, but there is always a delay in the release of the CA corresponding to the k8s release and it affects the chart version. we have to consider this thing also. suppose a user uses the new release of k8s but we would not have the corresponding release of CA at that time so what version it would consider for CA?

Shubham82:
> Hi @gjtempleton I saw your comment on the issue: #5640. I agree with your point.
> 
> Is there any way or pointer to start with this?
> 
[...]

PR #5640, opened 2023-03-30
[helm chart] Update to use cluster-autoscaler 1.26.2 + default tag value from the Chart's AppVersion

asherf:
This is similar to how it is done in other chats:
https://github.com/kubernetes/dashboard/blob/313af7c8e1922fb4d7c6be4dc3fc006e286e759a/charts/helm-chart/kubernetes-dashboard/templates/deployment.yaml#L66 https://github.com/kubernetes/dashboard/blob/313af7c8e1922fb4d7c6be4dc3fc006e286e759a/charts/helm-chart/kubernetes-dashboard/values.yaml#L23

/kind feature

### DIFF
--- a/charts/cluster-autoscaler/values.yaml
+++ b/charts/cluster-autoscaler/values.yaml
@@ -229,8 +229,7 @@ hostNetwork: false
 image:
   # image.repository -- Image repository
   repository: registry.k8s.io/autoscaling/cluster-autoscaler
-  # image.tag -- Image tag
-  tag: v1.24.0
+  tag: ""  # If not defined, uses appVersion of Chart.yaml
   # image.pullPolicy -- Image pull policy
   pullPolicy: IfNotPresent
   ## Optionally specify an array of imagePullSecrets.
--- a/charts/cluster-autoscaler/values.yaml
+++ b/charts/cluster-autoscaler/values.yaml
@@ -229,6 +229,7 @@ hostNetwork: false
 image:
   # image.repository -- Image repository
   repository: registry.k8s.io/autoscaling/cluster-autoscaler
+  # image.tag - Image tag - defaults to the version defined in Chart.AppVersion
   tag: ""  # If not defined, uses appVersion of Chart.yaml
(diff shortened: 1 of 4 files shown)


asherf:
not sure about the docs lint fail.... any pointers ?

charts/cluster-autoscaler/values.yaml
```diff
@@ -229,8 +229,7 @@ hostNetwork: false
 image:
   # image.repository -- Image repository
   repository: registry.k8s.io/autoscaling/cluster-autoscaler
-  # image.tag -- Image tag
-  tag: v1.24.0
+  tag: ""  # If not defined, uses appVersion of Chart.yaml
```
gjtempleton, line 232:
```suggestion
  # image.tag - Image tag - defaults to the version defined in Chart.AppVersion
  tag: ""  # If not defined, uses appVersion of Chart.yaml
```

gjtempleton (changes requested):
Hey, thanks for the PR.

We use and enforce helm-docs to generate our chart README, which is why the CI is currently failing, you can either set it up as a [pre-commit hook](https://github.com/kubernetes/autoscaler/tree/master/charts#pre-commit-hooks) or run it manually.

gjtempleton:
On balance I'm not sure this is a path we want to take with the CA given versions of the CA are so closely tied to the version of kubernetes and mismatching minor versions between these can result in unexpected behaviour.

There's been a previous request to look at automatically picking an image to match the minor version of the underlying cluster as per #5324.

asherf:
> On balance I'm not sure this is a path we want to take with the CA given versions of the CA are so closely tied to the version of kubernetes and mismatching minor versions between these can result in unexpected behaviour.
> 
> There's been a previous request to look at automatically picking an image to match the minor version of the underlying cluster as per #5324.

good point.
closing in favor of https://github.com/kubernetes/autoscaler/pull/5640#issuecomment-1493723676
I appreciate the feedback and the guidance on using pre-commit, I somehow missed that.

PR #5649, opened 2023-04-03
Move MaxNodeProvisionTime to NodeGroupAutoscalingOptions

morshielt:
####  Which component does this PR applies to?
cluster-autoscaler

#### What type of PR is this?
/kind cleanup

This PR enables setting `MaxNodeProvisionTime` per NodeGroup (it was previously set per cluster).

Fixes #

Please take a look.

```release-note
MaxNodeProvisionTime can now be defined per node group using 'maxnodeprovisiontime' key. The value of 'max-node-provision-time flag' is still used as the default for all node groups, so this change should have no visible impact on the behavior of Cluster Autoscaler.
```
NA

/assign @kisieland

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -45,6 +45,11 @@ const (
 	MaxNodeStartupTime = 15 * time.Minute
 )
 
+type maxNodeProvisionTimeProvider interface {
+	// GetMaxNodeProvisionTime returns MaxNodeProvisionTime value that should be used for the given NodeGroup.
+	GetMaxNodeProvisionTime(nodeGroup cloudprovider.NodeGroup) (time.Duration, error)
+}
+
 // ScaleUpRequest contains information about the requested node group scale up.
 type ScaleUpRequest struct {
 	// NodeGroup is the node group to be scaled up.
@@ -76,8 +81,6 @@ type ClusterStateRegistryConfig struct {
 	// Minimum number of nodes that must be unready for MaxTotalUnreadyPercentage to apply.
 	// This is to ensure that in very small clusters (e.g. 2 nodes) a single node's failure doesn't disable autoscaling.
 	OkTotalUnreadyCount int
-	//  Maximum time CA waits for node to be provisioned
-	MaxNodeProvisionTime time.Duration
 }
 
 // IncorrectNodeGroupSize contains information about how much the current size of the node group
@@ -132,6 +135,7 @@ type ClusterStateRegistry struct {
 	previousCloudProviderNodeInstances map[string][]cloudprovider.Instance
 	cloudProviderNodeInstancesCache    *utils.CloudProviderNodeInstancesCache
 	interrupt                          chan struct{}
+	maxNodeProvisionTimeProvider       maxNodeProvisionTimeProvider
 
 	// scaleUpFailures contains information about scale-up failures for each node group. It should be
 	// cleared periodically to avoid unnecessary accumulation.
@@ -139,7 +143,7 @@ type ClusterStateRegistry struct {
 }
 
 // NewClusterStateRegistry creates new ClusterStateRegistry.
-func NewClusterStateRegistry(cloudProvider cloudprovider.CloudProvider, config ClusterStateRegistryConfig, logRecorder *utils.LogEventRecorder, backoff backoff.Backoff) *ClusterStateRegistry {
+func NewClusterStateRegistry(cloudProvider cloudprovider.CloudProvider, config ClusterStateRegistryConfig, logRecorder *utils.LogEventRecorder, backoff backoff.Backoff, maxNodeProvisionTimeProvider maxNodeProvisionTimeProvider) *ClusterStateRegistry {
 	emptyStatus := &api.ClusterAutoscalerStatus{
 		ClusterwideConditions: make([]api.ClusterAutoscalerCondition, 0),
 		NodeGroupStatuses:     make([]api.NodeGroupStatus, 0),
@@ -163,6 +167,7 @@ func NewClusterStateRegistry(cloudProvider cloudprovider.CloudProvider, config C
 		cloudProviderNodeInstancesCache: utils.NewCloudProviderNodeInstancesCache(cloudProvider),
 		interrupt:                       make(chan struct{}),
 		scaleUpFailures:                 make(map[string][]ScaleUpFailure),
+		maxNodeProvisionTimeProvider:    maxNodeProvisionTimeProvider,
 	}
 }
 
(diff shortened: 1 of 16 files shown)


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -734,17 +737,23 @@ func removeOldUnregisteredNodes(unregisteredNodes []clusterstate.UnregisteredNod
...
+			return removedAny, err
+		}
+		if nodeGroup == nil || reflect.ValueOf(nodeGroup).IsNil() {
+			klog.Warningf("No node group for node %s, skipping", unregisteredNode.Node.Name)
+			continue
+		}
+
+		maxNodeProvisionTime, err := csr.MaxNodeProvisionTime(context, nodeGroup)
+		if err != nil {
+			return false, fmt.Errorf("failed to retrieve maxNodeProvisionTime for node %s in nodeGroup %s", unregisteredNode.Node.Name, nodeGroup.Id())
```
BigDarkClown, line 752: For consistency this should return `removedAny` instead of `false`.

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -1280,8 +1281,8 @@ func TestStaticAutoscalerInstanceCreationErrors(t *testing.T) {
 		func(node *apiv1.Node) bool {
 			return false
 		}, nil)
-
-	clusterState = clusterstate.NewClusterStateRegistry(provider, clusterStateConfig, context.LogRecorder, NewBackoff())
+	context.CloudProvider = provider
```
BigDarkClown, line 1284: This test is confusing as hell, but from what I was able to understand it uses two different CloudProvider mocks, each with different mocks, and then passes first one to the StaticAutoscaler and the second one two the ClusterStateRegistry. Switching the CloudProvider in this context would actually make both StaticAutoscaler and ClusterStateRegistry work on the same copy, which I don't think is intended.

I don't think that unraveling this bundle of joy is in the scope here, but just to be safe I would create a new context and pass it to `NewClusterStateRegistry` instead. This way we keep the original functionality of this test without going too deep into the specifics.

BigDarkClown (changes requested):

cluster-autoscaler/core/static_autoscaler_test.go
morshielt: Done, thanks!

cluster-autoscaler/core/static_autoscaler.go
morshielt: Done, thanks!

cluster-autoscaler/main.go
```diff
@@ -148,7 +148,7 @@ var (
 	maxTotalUnreadyPercentage  = flag.Float64("max-total-unready-percentage", 45, "Maximum percentage of unready nodes in the cluster.  After this is exceeded, CA halts operations")
 	okTotalUnreadyCount        = flag.Int("ok-total-unready-count", 3, "Number of allowed unready nodes, irrespective of max-total-unready-percentage")
 	scaleUpFromZero            = flag.Bool("scale-up-from-zero", true, "Should CA scale up when there 0 ready nodes.")
-	maxNodeProvisionTime       = flag.Duration("max-node-provision-time", 15*time.Minute, "Maximum time CA waits for node to be provisioned")
+	maxNodeProvisionTime       = flag.Duration("max-node-provision-time", 15*time.Minute, "The default maximum time CA waits for node to be provisioned")
```
x13n, line 151: nit: "default maximum" sounds a bit confusing. I'd explain here that it can be overridden per node group.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -111,6 +115,7 @@ type ScaleUpFailure struct {
 // ClusterStateRegistry is a structure to keep track the current state of the cluster.
 type ClusterStateRegistry struct {
 	sync.Mutex
+	context                            *context.AutoscalingContext
```
x13n, line 118: I'm a bit sad we need to keep the entire context as a field here. Could we have an object that is just concerned with saying what max node provision time is for a given node (or nodegroup)? The context would then be kept there, making `ClusterStateRegistry` easier (well, at least not harder) to reason about.

x13n (commented):
I'm not going to block this PR on this, but leaving 2 comments to think about. If you decide to ignore them, just drop the hold.

/approve
/hold

cluster-autoscaler/main.go
morshielt: Changed to `The default maximum time CA waits for node to be provisioned - the value can be overriden per node group`

cluster-autoscaler/clusterstate/clusterstate.go
morshielt: Renamed the provider interface to `maxNodeProvisionTimeProvider` and introduced `nodeRegistrationTimeLimitProvider` struct containing the provider and context. PTAL :)

x13n: Thanks, although my point was really that whether or not the context is used is an implementation detail and doesn't have to be a part of the interface. So, one could imagine a `staticMaxNodeProvisionTimeProvider` which just returns a constant duration, as well as more advanced implementations depending on a specific use case. Regardless of what implementation of `maxNodeProvisionTimeProvider` interface is used, CSR should not require context passed to the constructor (i.e. `New...` function).

morshielt: Thank you for the explanation -  I've introduced `NewDefaultMaxNodeProvisionTimeProvider` returning an instance of `maxNodeProvisionTimeProvider`. It's kept as a field in `ClusterStateRegistry` and it already contains the context and `nodeGroupConfigProcessor` when it's assigned, so the `NewClusterStateRegistry` function signature is mostly back to as it was, it only takes the `maxNodeProvisionTimeProvider` as a new argument. Please take a look :)

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -45,6 +47,26 @@ const (
 	MaxNodeStartupTime = 15 * time.Minute
 )
 
+type maxNodeProvisionTimeProvider interface {
+	// GetMaxNodeProvisionTime returns MaxNodeProvisionTime value that should be used for the given NodeGroup.
+	GetMaxNodeProvisionTime(nodeGroup cloudprovider.NodeGroup) (time.Duration, error)
+}
+
+// NewDefaultMaxNodeProvisionTimeProvider returns the default maxNodeProvisionTimeProvider which uses the NodeGroupConfigProcessor.
+func NewDefaultMaxNodeProvisionTimeProvider(context *context.AutoscalingContext, nodeGroupConfigProcessor nodegroupconfig.NodeGroupConfigProcessor) maxNodeProvisionTimeProvider {
```
x13n, line 56: This doesn't really have much to do with clusterstate, so I'd move it out. Separate package might be an overkill, but perhaps a separate file?

cluster-autoscaler/core/scaleup/orchestrator/orchestrator_test.go
```diff
@@ -696,11 +696,11 @@ func TestScaleUpUnhealthy(t *testing.T) {
 
 	nodes := []*apiv1.Node{n1, n2}
 	nodeInfos, _ := nodeinfosprovider.NewDefaultTemplateNodeInfoProvider(nil, false).Process(&context, nodes, []*appsv1.DaemonSet{}, nil, now)
-	clusterState := clusterstate.NewClusterStateRegistry(provider, clusterstate.ClusterStateRegistryConfig{}, context.LogRecorder, NewBackoff())
+	processors := NewTestProcessors(&context)
+	clusterState := clusterstate.NewClusterStateRegistry(provider, clusterstate.ClusterStateRegistryConfig{}, context.LogRecorder, NewBackoff(), clusterstate.NewDefaultMaxNodeProvisionTimeProvider(&context, processors.NodeGroupConfigProcessor))
```
x13n, line 700: nit: Testing would perhaps be simpler with a test/static maxnodeProvisionTimeProvider - no need to pass `context` or `NodeGroupConfigProcessor`.

x13n:
Looks good now, thanks!

/lgtm
/approve

PR #5658, opened 2023-04-05
Add unit tests for gpu utils

kawych:
This requires moving parts of gpu util to cloudprovider package to get rid of a dependency cycle.

#### What type of PR is this?

/kind cleanup

Adds an unit test

```release-note
```

### DIFF
--- a/cluster-autoscaler/cloudprovider/util/gpu_test.go
+++ b/cluster-autoscaler/cloudprovider/util/gpu_test.go
@@ -1,0 +1,84 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"testing"
+
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/test"
+
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	GPULabel = "TestGPULabel/accelerator"
+)
+
+func TestNodeHasGpu(t *testing.T) {
+	gpuLabels := map[string]string{
+		GPULabel: "nvidia-tesla-k80",
+	}
+	nodeGpuReady := &apiv1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   "nodeGpuReady",
+			Labels: gpuLabels,
+		},
+		Status: apiv1.NodeStatus{
+			Capacity:    apiv1.ResourceList{},
+			Allocatable: apiv1.ResourceList{},
+		},
+	}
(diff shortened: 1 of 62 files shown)


cluster-autoscaler/utils/gpu/gpu_test.go
```diff
@@ -31,54 +34,191 @@ const (
...
+			},
+			node:         newNode("label/gpu-type", "gpu-type-value", nil),
+			wantResource: util.ResourceNvidiaGPU,
+			wantLabel:    "unexpected-label",
 		},
 	}
-	assert.False(t, NodeHasGpu(GPULabel, nodeNoGpu))
-}
+	for _, tc := range tests {
+		func(tc testCase) {
```
BigDarkClown, line 174: Why is this additional func wrapper here? I don't really see why this is needed.

cluster-autoscaler/utils/gpu/gpu_test.go
```diff
@@ -31,54 +34,191 @@ const (
...
+			gpus: map[string]struct{}{
+				"gpu-type-value": {},
+			},
+			node:         newNode("label/gpu-type", "gpu-type-value", nil),
+			templateNode: newNode("label/gpu-type", "gpu-type-value", nil),
+			wantResource: util.ResourceNvidiaGPU,
+			wantLabel:    "unexpected-label",
+		},
+		{
+			desc: "no capacity and node group",
```
BigDarkClown, line 159: Should be `"no capacity and no node group"`

cluster-autoscaler/utils/gpu/gpu_test.go
```diff
@@ -31,54 +34,191 @@ const (
...
+				gotResource, gotLabel := GetGpuInfoForMetrics(tc.config, tc.gpus, tc.node, nodeGroup)
+
+				assert.Equal(t, tc.wantResource, gotResource)
+				assert.Equal(t, tc.wantLabel, gotLabel)
+			})
+		}(tc)
+	}
+}
+
+func newNode(gpuLabel, gpuType string, gpuCapacity *resource.Quantity) *apiv1.Node {
```
BigDarkClown, line 204: The current `GetGpuInfoForMetrics` implementation uses only the node capacity, not the labels. The gpu label/value is extracted directly from the `GpuConfig`. I think it would be better to change this method to `newNode(gpuCapacity *resource.Quantity) *apiv1.Node` and not set the labels at all. This way we also document that labels/values are taken from `GpuConfig`

cluster-autoscaler/cloudprovider/util/gpu_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"k8s.io/autoscaler/cluster-autoscaler/utils/test"
+
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	GPULabel = "TestGPULabel/accelerator"
+)
+
+func TestNodeHasGpu(t *testing.T) {
```
BigDarkClown, line 34: Maybe also add a `no gpu labels but allocatable defined` test case. Also this would make a good table unit test.

BigDarkClown (changes requested):

cluster-autoscaler/utils/gpu/gpu_test.go
kawych: Replaced with a simpler pattern of tc := tc

cluster-autoscaler/utils/gpu/gpu_test.go
kawych: done

cluster-autoscaler/utils/gpu/gpu_test.go
kawych: Done

cluster-autoscaler/cloudprovider/util/gpu_test.go
kawych: Added a test case

kawych:
@BigDarkClown I rebased it, please take another look. If we still want this cleanup I'd like to finalize this change.

kawych:
Looks like the tests are still failing, I don't really have much time to spend on it right now - closing for now, I think we can resume later if needed.

PR #5659, opened 2023-04-05
Sanitize taints before scheduling DSs on template node infos

BigDarkClown:
#### What type of PR is this?

/kind bug
/kind cleanup

NodeInfoProvider returns node infos representing each node group for scale-up. If there is at least one node in the node group, then it is used as a template. If there are none, the node info is based on the node group instance template.

In both these cases we want to pack the node infos with all the expected DaemonSet pods. That way CA won't perform a scale-up which leaves some DSs pending. All DS pods should be added to the node info if it is generated from instance template or if it is generated from existing node and the `force-ds` flag is enabled. 

The issue is that the node sanitisation, which includes cleaning unwanted taints, is performed before the DS are added. That means that it is possible to not include all of the DS pods in the node info if for example:
- There are no nodes in the node group, node group has `ignore taint`.
- The node picked for representation is currently being scaled-down and has the `deletion taint`.

This change fixes it by performing additional sanitisation before trying to schedule DS pods. It also add additional `unwantedTaints` option, which enables defining taints which are removed from templates but not treated as ignore taints.

### DIFF
--- a/cluster-autoscaler/core/utils/utils.go
+++ b/cluster-autoscaler/core/utils/utils.go
@@ -45,6 +45,12 @@ func GetNodeInfoFromTemplate(nodeGroup cloudprovider.NodeGroup, daemonsets []*ap
 
 	labels.UpdateDeprecatedLabels(baseNodeInfo.Node().ObjectMeta.Labels)
 
+	sanitizedNode, typedErr := SanitizeNode(baseNodeInfo.Node(), id, taintConfig)
+	if err != nil {
+		return nil, typedErr
+	}
+	baseNodeInfo.SetNode(sanitizedNode)
+
 	pods, err := daemonset.GetDaemonSetPodsForNode(baseNodeInfo, daemonsets)
 	if err != nil {
 		return nil, errors.ToAutoscalerError(errors.InternalError, err)
@@ -52,12 +58,9 @@ func GetNodeInfoFromTemplate(nodeGroup cloudprovider.NodeGroup, daemonsets []*ap
 	for _, podInfo := range baseNodeInfo.Pods {
 		pods = append(pods, podInfo.Pod)
 	}
-	fullNodeInfo := schedulerframework.NewNodeInfo(pods...)
-	fullNodeInfo.SetNode(baseNodeInfo.Node())
-	sanitizedNodeInfo, typedErr := SanitizeNodeInfo(fullNodeInfo, id, taintConfig)
-	if typedErr != nil {
-		return nil, typedErr
-	}
+
+	sanitizedNodeInfo := schedulerframework.NewNodeInfo(SanitizePods(pods, sanitizedNode)...)
+	sanitizedNodeInfo.SetNode(sanitizedNode)
 	return sanitizedNodeInfo, nil
 }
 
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -186,6 +186,8 @@ type AutoscalingOptions struct {
 	MaxPodEvictionTime time.Duration
 	// IgnoredTaints is a list of taints to ignore when considering a node template for scheduling.
 	IgnoredTaints []string
+	// UnwantedTaints is a list of taints to remove when creating a node template for scheduling.
+	UnwantedTaints []string
```
MaciekPytel, line 190: I really don't like this name, it only make sense in context of how the variable is used in the code, not what the variable wants to achieve (ie. name is driven by implementation not use-case).
I'd consider calling those StatusTaints, StateTaints or ConditionTaints or similar. I think that's a good name for non-permanent taint that can be added during lifetime of the node, because any such taint would presumably represent some sort of node state. And the name doesn't say how we're treating the taint in CA logic, instead focusing on what the taint is meant to represent.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -167,6 +168,12 @@ func NewStaticAutoscaler(
 		ignoredTaints[taintKey] = true
 	}
 
+	unwantedTaints := make(taints.TaintKeySet)
```
MaciekPytel, line 171: Can we maybe move this + ignored taints above to a helper function? static_autoscaler.go is already huge.

Also, I'm wondering - if we move it to taints package, maybe we could just have a single variable that represents all taint-related config and pass that around rather than adding more parameters all over the place? If we ever expected 3 TaintKeySets that would be a no-brainer to me, with just 2 I'm not sure if it's worth the effort. I don't really have strong feelings either way so take this part just as a suggestion and feel free to ignore it if you disagree.

cluster-autoscaler/core/utils/utils.go
```diff
@@ -36,25 +36,30 @@ import (
...
-func GetNodeInfoFromTemplate(nodeGroup cloudprovider.NodeGroup, daemonsets []*appsv1.DaemonSet, ignoredTaints taints.TaintKeySet) (*schedulerframework.NodeInfo, errors.AutoscalerError) {
+func GetNodeInfoFromTemplate(nodeGroup cloudprovider.NodeGroup, daemonsets []*appsv1.DaemonSet, ignoredTaints, unwantedTaints taints.TaintKeySet) (*schedulerframework.NodeInfo, errors.AutoscalerError) {
 	id := nodeGroup.Id()
 	baseNodeInfo, err := nodeGroup.TemplateNodeInfo()
 	if err != nil {
 		return nil, errors.ToAutoscalerError(errors.CloudProviderError, err)
 	}
 
-	labels.UpdateDeprecatedLabels(baseNodeInfo.Node().ObjectMeta.Labels)
+	sanitizedBaseNodeInfo, typedErr := SanitizeNodeInfo(baseNodeInfo, id, ignoredTaints, unwantedTaints)
```
MaciekPytel, line 46: Should we really SanitizeNodeInfo twice? Why would we need to go through deep copying the node twice, generate 2 random names for it, etc? The performance impact is negligible, but it seems confusing and potentially error prone with extra temporary deep copies.

I think here what we really need is just SanitizeNode()? And than we can decouple it from sanitizing pods that we would need to do at the end of this function?

cluster-autoscaler/processors/nodeinfosprovider/mixed_nodeinfos_processor.go
```diff
@@ -95,8 +95,15 @@ func (p *MixedTemplateNodeInfoProvider) Process(ctx *context.AutoscalingContext,
...
-			sanitizedNodeInfo, err := utils.SanitizeNodeInfo(nodeInfo, id, ignoredTaints)
+			sanitizedNode, err := utils.SanitizeNode(node, id, ignoredTaints, unwantedTaints)
+			if err != nil {
+				return false, "", err
+			}
+			nodeInfo, err := simulator.BuildNodeInfoForNode(sanitizedNode, podsForNodes[node.Name], daemonsets, p.forceDaemonSets)
+			if err != nil {
+				return false, "", err
+			}
+			sanitizedNodeInfo, err := utils.SanitizeNodeInfo(nodeInfo, id, ignoredTaints, unwantedTaints)
```
MaciekPytel, line 106: This isn't really a part of your changes, but with the extra SanitizeNode call it's just more visible - why is the above logic different from the GetNodeInfoFromTemplate()? Doesn't it do the exact same thing just with different nodeInfo as a source?

I understand this may not be in the scope of this PR, so feel free not to do it, but maybe it would be worth a small refactor to unify those codepaths?

cluster-autoscaler/core/static_autoscaler.go
BigDarkClown: Added TaintConfig to encapsulate both ignore and status taints.

cluster-autoscaler/core/utils/utils.go
BigDarkClown: Nice catch, changed to SanitizeNode and setting it on the baseNodeInfo.

cluster-autoscaler/processors/nodeinfosprovider/mixed_nodeinfos_processor.go
BigDarkClown: That's a valid point, but this is not 100% true. We treat DS pods slightly differently depending on where we get template from. For existing nodes we take DS pods scheduled them and fill with missing DS pods if the option `--force-ds` is enabled. For instance template ones we always pack all the DS pods.

I think long term solution here is to set `--force-ds` to true always and unify both these logic pieces. But for now I'm not sure if it won't obfuscate the different underlying behaviour.

PR #5660, opened 2023-04-05
[vpa] [updater] new flag to have labelSelector on pods

dbenque:
#### What type of PR is this?

/kind feature

While it is technically possible to define an admission webhook that will act on a subset of pods thanks to `MutatingWebhookConfiguration.Webhooks.ObjectSelector`, it was impossible so far to limit the scope of the Updater to this same subset of pods.

Thanks to this PR the user operating the solution can defined what are the pods that are eligible for the Updater. This way the operator is able to set same type of constraint (labelSelector) on both admission and updater.

No issue opened, but I am facing the problem while operating the solution:

I am using VPA to deliver recommendations everywhere but only some pods are eligible for the admission. If the VPA is set to is different than "Off" but the pod is not eligible for admission (constraint enforced thanks to `Webhooks.ObjectSelector`) , then the Updater is evicting the pod but the pod is not going through admission. In that case I would like the Updater to skip that pod. In other words we ne to give a mean to align Updater pod scope and Admission pod scope.

Yes a new flag is introduced. The default value ensure that the program behaviour is backward compatible.
```release-note
A new flag --pod-label-selector was introduced. Only pods that match that label selector can be evicted by the Updater. By default it is empty and match all the pods. The syntax is defined in the labels package here: https://github.com/kubernetes/apimachinery/blob/8d1258da8f386b809d312cdda316366d5612f54e/pkg/labels/selector.go#L843
```

```docs

```

### DIFF
--- a/vertical-pod-autoscaler/pkg/updater/logic/updater.go
+++ b/vertical-pod-autoscaler/pkg/updater/logic/updater.go
@@ -22,10 +22,18 @@ import (
 	"time"
 
 	"golang.org/x/time/rate"
-
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
+	kube_client "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/kubernetes/scheme"
+	clientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	v1lister "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/klog/v2"
+
 	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1"
 	vpa_clientset "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/client/clientset/versioned"
 	vpa_lister "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/client/listers/autoscaling.k8s.io/v1"
@@ -35,14 +43,6 @@ import (
 	metrics_updater "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/metrics/updater"
 	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/status"
 	vpa_api_util "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/vpa"
-	kube_client "k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/kubernetes/fake"
-	"k8s.io/client-go/kubernetes/scheme"
-	clientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
-	v1lister "k8s.io/client-go/listers/core/v1"
-	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/tools/record"
-	"k8s.io/klog/v2"
 )
 
 // Updater performs updates on pods if recommended by Vertical Pod Autoscaler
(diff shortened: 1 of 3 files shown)


dbenque:
The problem is not inside a given deployment.
It is a matter of governance and responsibility here:
- recommenders, updater, admission controller and where the admission can run is defined by the platform owner my team.
- VPA are under control of the application owner other team.

My team runs and define where the admission is eligible/acceptable/allowed (governance). This is done thanks to the webhook configuration. For example I only select some namespace that are eligible via a namespace selector or a set of object with object selector. This capability is already exposed in the kubernetes API.
Other teams deploy their applications and VPA. They deploy with their "wish" in terms of VPA (Recreate/Auto/Off ... and even we have internally another mode that is OnDeployOnly). But this is a wish... if for governance reason VPA is not activated in this namespace (or cluster) then it is important that the updater do not try to evict the pods, because they won't go in the admission.

In other words, the admission webhook configuration allows the platform owner to define a scope of where the webhook is applicable. It is important to give to that same actor the ability to declare that same scope at updater level.

dbenque:
As per the discussion I had with @voelzmo , this PR is missing a parameter for Namespace selector.

In my intention to deliver to the platform operator (the person operation VPA in the cluster) the ability to align scope for Admission and Updater I for got one part. The PR only cover the [ObjectSelector](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-namespaceselector) for the moment (how to select pods), but I am missing [NamespaceSelector](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-namespaceselector)!

jbartosik:
> Other teams deploy their applications and VPA. They deploy with their "wish" in terms of VPA (Recreate/Auto/Off ... and even we have internally another mode that is OnDeployOnly). But this is a wish... if for governance reason VPA is not activated in this namespace (or cluster) then it is important that the updater do not try to evict the pods, because they won't go in the admission.

Why not have a validation and allow only VPAs with mode `Off` for cases when you don't want to allow VPA to actuate[1]? That way it's clear what's happening. With approach in this change there will be VPA objects that look like they're actuating their recommendations (they have mode `Auto`) but don't really do that.

I think rejecting requests to create/ modify VPAs to have actuation enabled results in clearer situation.

What do you think?

[1] Using a webhook: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/

dbenque:
> for cases when you don't want to allow VPA to actuate[1]?

@jbartosik , for the moment I am still evaluating your previous proposition based on validation. But I have a problem with it:
If the scope of admission/updater is changing, then people owning the VPA have to got an update their chart. It is potentially ok in that way, but won't work way the other way around. I mean if ever we reduce the scope at admission/updater level what do we do with the VPA that have already been approved by the validation? They are now mode!=Off and they are not in scope of the admission/updater

sophieliu15:
Hi David,

Sorry for the late response. Workload autoscaling work is transitioned from [Joachim](https://github.com/jbartosik)'s team to our team. We are in the process of getting familiar with the scope and working towards becoming reviewers of this domain under the guidance of Joachim. I notice this PR has been opened for quite some time. If this feature is no longer needed, please let us know.

In terms of the option you proposed, I noticed that you added the selector in the Updater. How do we address inconsistency between selector in Admission Controller and the selector you defined in the Updater? Does it make sense (and if it is feasible) to have a centralized selector that will be respected by both Admission Controller and Updater? If such centralized selector is defined, the `Webhooks.ObjectSelector` on Admission Controller will be ignored. VPAs who are responsible for objects not on the selector will either issue an error status and/or automatically switch to "off" mode. 

On the other hand, one could argue current behaviour is intended even though it is not user friendly. Platform admin has the right to decide which objects get admitted and it is application developers' responsibility to make sure they configure the VPA accordingly. If a pod gets evicted by the Updater and is not going to be accepted by Admission Controller, the platform can issue a log or other kinds of warning telling the application developers what is going on so the developers can adjust their expectations of the behaviour of VPA accordingly. It might also be useful to provide developers a way to check what objects are currently accepted by Admission Controller so they do not define VPAs in vain. 

Please let me know your thoughts on this topic!

cc: @jbartosik @laoj2 @voelzmo

mwielgus:
Closing due to inactivity. Please reopen once needed.

PR #5663, opened 2023-04-06
Fix the resource cap when scaling up multiple node groups to min size at a time

### LINKED ISSUE #5624: Cluster autoscaler scale to min may breach cluster wide limits

cluster-autoscaler

Most recent from the HEAD.

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.3", GitCommit:"9e644106593f3f4aa98f8a84b23db5fa378900bd", GitTreeState:"clean", BuildDate:"2023-03-15T13:40:17Z", GoVersion:"go1.19.7", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.7
Server Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.1-gke.1500", GitCommit:"633b8d38c09de2c1acd636c37f1f555bef6f905f", GitTreeState:"clean", BuildDate:"2023-02-07T09:38:46Z", GoVersion:"go1.19.5 X:boringcrypto", Compiler:"gc", Platform:"linux/amd64"}
```

</pre></details>

GCE running a custom Cluster autoscaler with `enforce-node-group-min-size` enabled maximal number of cores set to 5. 
There were 6 node groups present in the cluster, each running node with 1 core. 
In one there were 2 VMs, in another 1 and all others were empty.
When the 3 of empty nodegroups got updated to set min-nodes to 1, CA created 1 VM in each of them, resulting in 6 cores used.

I would expect CA to respect the total resource limits and only create 2 new nodes.

**What happened instead?**:

CA created 3 nodes.

Have a cluster close to core/node limit and atomically increase the min-size of two or more node groups.

The code is in [here](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go#L398-L418), I've noticed the bug during refactor of the scale-up logic.

CC: @liuxintong as author of #5195

xint6:
#### What type of PR is this?

/kind bug

Fix the resource cap when scaling up multiple node groups to min size at a time.

Fixes #5624

A new function `AllocateResources` is introduced in the resource manager, which allocates the resources of required number of nodes from resources left and return true if resources are allocated successfully. `ScaleUpToNodeGroupMinSize` calls `AllocateResources` to deduct resources for each node group.

NONE

```docs
- [FAQ]: https://github.com/kubernetes/autoscaler/blob/ad86bdd64190baa3a685ec4816a09372453b67b3/cluster-autoscaler/FAQ.md#my-cluster-is-below-minimum--above-maximum-number-of-nodes-but-ca-did-not-fix-that-why
```

### DIFF
--- a/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
+++ b/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
@@ -344,6 +344,18 @@ func (o *ScaleUpOrchestrator) ScaleUpToNodeGroupMinSize(
 		return scaleUpError(&status.ScaleUpStatus{}, aErr.AddPrefix("could not compute total resources: "))
 	}
 
+	nodeCountInCluster := 0
+	for _, ng := range nodeGroups {
+		if !ng.Exist() {
+			continue
+		}
+		targetSize, err := ng.TargetSize()
+		if err != nil {
+			continue
+		}
+		nodeCountInCluster += targetSize
+	}
+
 	for _, ng := range nodeGroups {
 		if !ng.Exist() {
 			klog.Warningf("ScaleUpToNodeGroupMinSize: NodeGroup %s does not exist", ng.Id())
@@ -373,23 +385,35 @@ func (o *ScaleUpOrchestrator) ScaleUpToNodeGroupMinSize(
 		}
 
 		if skipReason := o.IsNodeGroupResourceExceeded(resourcesLeft, ng, nodeInfo); skipReason != nil {
-			klog.Warning("ScaleUpToNodeGroupMinSize: node group resource excceded: %v", skipReason)
+			klog.Warningf("ScaleUpToNodeGroupMinSize: node group resource excceded: %v", skipReason.Reasons())
 			continue
 		}
 
 		newNodeCount := ng.MinSize() - targetSize
 		newNodeCount, err = o.resourceManager.ApplyLimits(o.autoscalingContext, newNodeCount, resourcesLeft, nodeInfo, ng)
 		if err != nil {
-			klog.Warning("ScaleUpToNodeGroupMinSize: failed to apply resource limits: %v", err)
+			klog.Warningf("ScaleUpToNodeGroupMinSize: failed to apply resource limits: %v", err)
 			continue
 		}
 
-		newNodeCount, err = o.GetCappedNewNodeCount(newNodeCount, targetSize)
+		newNodeCount, err = o.GetCappedNewNodeCount(newNodeCount, nodeCountInCluster)
 		if err != nil {
-			klog.Warning("ScaleUpToNodeGroupMinSize: failed to get capped node count: %v", err)
+			klog.Warningf("ScaleUpToNodeGroupMinSize: failed to get capped node count: %v", err)
 			continue
 		}
 
+		delta, err := o.resourceManager.DeltaForNode(o.autoscalingContext, nodeInfo, ng)
+		if err != nil {
+			klog.Warningf("ScaleUpToNodeGroupMinSize: failed to get delta for node: %v", err)
+			continue
+		}
+		resourcesLeft, err = resource.UpdateLimits(resourcesLeft, delta, newNodeCount)
+		if err != nil {
+			klog.Warningf("ScaleUpToNodeGroupMinSize: failed to allocate %d nodes from resources left: %v", newNodeCount, err)
+			continue
+		}
+
+		nodeCountInCluster += newNodeCount
 		info := nodegroupset.ScaleUpInfo{
 			Group:       ng,
 			CurrentSize: targetSize,
--- a/cluster-autoscaler/core/scaleup/resource/manager.go
+++ b/cluster-autoscaler/core/scaleup/resource/manager.go
@@ -180,6 +180,32 @@ func (m *Manager) ApplyLimits(ctx *context.AutoscalingContext, newCount int, res
 	return newCount, nil
 }
 
+// UpdateLimits updates the resources of required number of nodes from resources left and return the updated resources left and error if any.
+func UpdateLimits(left Limits, delta Delta, count int) (Limits, error) {
+	newLeft := Limits{}
+	for resourceType, resourceLeft := range left {
+		newLeft[resourceType] = resourceLeft
+	}
+	for resourceType, resourceDelta := range delta {
+		if resourceDelta == 0 {
+			continue
+		}
+		resourceLeft, found := left[resourceType]
+		if !found {
+			continue
+		}
+		if resourceLeft == LimitUnknown {
+			return left, fmt.Errorf("limit unknown for resource %s", resourceType)
+		}
+		resourceRequired := resourceDelta * int64(count)
+		if resourceLeft < resourceRequired {
+			return left, fmt.Errorf("cannot create %d nodes: %s resource left (%d) is less than resource required (%d)", count, resourceType, resourceLeft, resourceRequired)
+		}
+		newLeft[resourceType] -= resourceRequired
+	}
+	return newLeft, nil
+}
+
 // CheckDeltaWithinLimits compares the resource limit and resource delta, and returns the limit check result.
 func CheckDeltaWithinLimits(left Limits, delta Delta) LimitsCheckResult {
 	exceededResources := sets.NewString()
--- a/cluster-autoscaler/core/scaleup/resource/manager_test.go
+++ b/cluster-autoscaler/core/scaleup/resource/manager_test.go
@@ -18,6 +18,7 @@ package resource
 
 import (
 	"fmt"
+	"reflect"
 	"testing"
 	"time"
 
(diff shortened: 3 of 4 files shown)


cluster-autoscaler/core/scaleup/resource/manager.go
```diff
@@ -180,6 +180,36 @@ func (m *Manager) ApplyLimits(ctx *context.AutoscalingContext, newCount int, res
 	return newCount, nil
 }
 
+// AllocateResources allocates the resources of required number of nodes from resources left and return true if resources are allocated successfully.
+func AllocateResources(left Limits, delta Delta, count int) bool {
+	allocate := func(dryRun bool) bool {
```
kisieland, line 185: My personal preference would be to make it a private function (and explicitly pass all input), rather than have it as func object.

kisieland (commented):
Thanks, this should fix the issue, leaving one comment.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -391,20 +391,30 @@ func (o *ScaleUpOrchestrator) ScaleUpToNodeGroupMinSize(
...
 
 		newNodeCount := ng.MinSize() - targetSize
 		newNodeCount, err = o.resourceManager.ApplyLimits(o.autoscalingContext, newNodeCount, resourcesLeft, nodeInfo, ng)
 		if err != nil {
-			klog.Warning("ScaleUpToNodeGroupMinSize: failed to apply resource limits: %v", err)
+			klog.Warningf("ScaleUpToNodeGroupMinSize: failed to apply resource limits: %v", err)
 			continue
 		}
 
 		newNodeCount, err = o.GetCappedNewNodeCount(newNodeCount, targetSize)
```
kisieland, line 405: One thing that I just noticed (and is a separate bug):
`IsNodeGroupResourceExceeded` is expecting the **total** size of the **cluster** as the second parameter. 
As it checks the total cluster size.
See the line in the standard scaleup: `GetCappedNewNodeCount(bestOption.NodeCount, len(nodes)+len(upcomingNodes))`

Once this is addressed it should also use an approach similar to `AllocateResources`

cluster-autoscaler/core/scaleup/resource/manager.go
xint6: Updated, either looks good to me.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
xint6: I'm assuming you were talking about `GetCappedNewNodeCount`. Yes, you are right, I just pushed the fix and added test cases to cover this. Thanks for pointing this out.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -362,6 +362,18 @@ func (o *ScaleUpOrchestrator) ScaleUpToNodeGroupMinSize(
 		return scaleUpError(&status.ScaleUpStatus{}, aErr.AddPrefix("could not compute total resources: "))
 	}
 
+	nodeCountInCluster := 0
```
kisieland, line 365: I think this variable should also contain nodes that are upcoming, i.e. are being created by the cloud provider (and didn't yet register in K8s).
The standard ScaleUp function uses a following call to fetch this information: 
```
upcomingCounts, _ := o.clusterStateRegistry.GetUpcomingNodes()
```
upcomingCounts contains a map from the nodegroup name to a number of VM so something as follows should add this information:
```
for _, numberOfNodes := range upcomingCounts {
	nodeCountInCluster += numberOfNodes
}
```

kisieland (commented):
one more comment to the `nodeCountInCluster` part

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
xint6: I believe `ng.TargetSize()` already includes the latest node count including upcoming nodes. This is a direct call to cloud provider, and cloud provider will return the real-time desired node count of the node group at that moment. Your point would be valid if I used `len(nodes)` here. Please let me know if you have different ideas.

kisieland: Yup, you are right.

kisieland (approved):

cluster-autoscaler/core/scaleup/resource/manager.go
```diff
@@ -180,6 +180,37 @@ func (m *Manager) ApplyLimits(ctx *context.AutoscalingContext, newCount int, res
...
+}
+
+func allocateResourcesWithDryRun(left Limits, delta Delta, count int, dryRun bool) bool {
+	for resourceType, resourceDelta := range delta {
+		if resourceDelta == 0 {
+			continue
+		}
+		resourceLeft, found := left[resourceType]
+		if !found {
+			continue
```
BigDarkClown, line 198: Shouldn't this return false in this case as well? And if so, could you add unit test case for that?

cluster-autoscaler/core/scaleup/resource/manager.go
```diff
@@ -180,6 +180,37 @@ func (m *Manager) ApplyLimits(ctx *context.AutoscalingContext, newCount int, res
 	return newCount, nil
 }
 
+// AllocateResources allocates the resources of required number of nodes from resources left and return true if resources are allocated successfully.
+func AllocateResources(left Limits, delta Delta, count int) bool {
```
BigDarkClown, line 184: The signature is IMHO a bit misleading - it is not immediately obvious that this method modifies the limits or what is the boolean value representing.

I think something like `func UpdateLimits(left Limits, delta Delta, count int) (Limits, error)` would be better. It also allows to pass the exact error to the call site.

cluster-autoscaler/core/scaleup/resource/manager_test.go
```diff
@@ -169,6 +170,38 @@ func TestApplyLimits(t *testing.T) {
...
+type allocateResourcesTestCase struct {
+	resourcesLeft  Limits
+	resourcesDelta Delta
+	nodeCount      int
+	expectedOutput bool
+	expectedLeft   Limits
+}
+
+func TestAllocateResources(t *testing.T) {
+	testCases := []allocateResourcesTestCase{
```
BigDarkClown, line 182: I would add a test case in which only one of the resources is exceeded, to emphasise that the method does not modify the resources left if it would result in exceeding the limit for one of them.

cluster-autoscaler/core/scaleup/resource/manager.go
```diff
@@ -180,6 +180,37 @@ func (m *Manager) ApplyLimits(ctx *context.AutoscalingContext, newCount int, res
...
+func allocateResourcesWithDryRun(left Limits, delta Delta, count int, dryRun bool) bool {
+	for resourceType, resourceDelta := range delta {
+		if resourceDelta == 0 {
+			continue
+		}
+		resourceLeft, found := left[resourceType]
+		if !found {
+			continue
+		}
+		if resourceLeft == LimitUnknown {
```
BigDarkClown, line 200: Could you add a unit test case for that?

BigDarkClown (changes requested):

cluster-autoscaler/core/scaleup/resource/manager.go
xint6: My understanding is that we could skip the resource check if the resource has no limits. `ApplyLimits` (line 155) follows the same logic.

cluster-autoscaler/core/scaleup/resource/manager.go
xint6: Added.

cluster-autoscaler/core/scaleup/resource/manager.go
xint6: Unit test has been added for this scenario.

cluster-autoscaler/core/scaleup/resource/manager_test.go
xint6: Added.

cluster-autoscaler/core/scaleup/resource/manager.go
BigDarkClown: SGTM, thanks for pointing that out.

cluster-autoscaler/core/scaleup/resource/manager_test.go
```diff
@@ -170,6 +171,63 @@ func TestApplyLimits(t *testing.T) {
...
+			nodeCount:      1,
+			expectedLeft:   Limits{"cpu": 2, "memory": 4},
+			expectedErr:    nil,
+		},
+		{
+			resourcesLeft:  Limits{"cpu": 10, "memory": 20},
+			resourcesDelta: Delta{"cpu": 8, "memory": 16},
+			nodeCount:      2,
+			expectedLeft:   Limits{"cpu": 10, "memory": 20},
+			expectedErr:    fmt.Errorf("cannot create 2 nodes: cpu resource left (10) is less than resource required (16)"),
```
BigDarkClown, line 196: I think this test case indeterministic. The Limits are just maps, and in `golang` maps are not ordered. Because of that it's possible that the error here would be the memory related one. I think the cpu one is more likely based on the map internal implementation, but in ~1/8 of runs the memory limit will be checked first. You can check it by running this using `go test` with `-count 1000` flag - it should fail fairly quickly.

xint6: Good catch! Just updated this test case to a deterministic one, thanks.

mwielgus:
Is it still needed?

xint6:
> Is it still needed?

Yes, it didn't get merged due to pending approval and then git conflict. I'll fix it in an upcoming PR, hopefully this week.

PR #5665, opened 2023-04-06
Limit refresh rate of GCE MIG instances.

olagacek:
#### What type of PR is this?

/kind feature

This PR limits the number of GCE calls to fetch mig instances. After large scale down, when a lot of instances got removed, we perform a lot of redundant calls to GCE - we try to fetch already deleted instances, even though it was done moments ago for a different, already deleted instance. To avoid that this PR introduces rate (by default 5s) at which mig instances can be refetched.

NONE

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
+++ b/cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
@@ -23,6 +23,7 @@ import (
 	"path"
 	"strings"
 	"sync"
+	"time"
 
 	gce "google.golang.org/api/compute/v1"
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
@@ -52,24 +53,40 @@ type MigInfoProvider interface {
 	GetMigMachineType(migRef GceRef) (MachineType, error)
 }
 
+type timeProvider interface {
+	Now() time.Time
+}
+
 type cachingMigInfoProvider struct {
-	migInfoMutex           sync.Mutex
-	cache                  *GceCache
-	migLister              MigLister
-	gceClient              AutoscalingGceClient
-	projectId              string
-	concurrentGceRefreshes int
-	migInstanceMutex       sync.Mutex
+	migInfoMutex                   sync.Mutex
+	cache                          *GceCache
+	migLister                      MigLister
+	gceClient                      AutoscalingGceClient
+	projectId                      string
+	concurrentGceRefreshes         int
+	migInstanceMutex               sync.Mutex
+	migInstancesMinRefreshWaitTime time.Duration
+	migInstancesLastRefreshedInfo  map[string]time.Time
+	timeProvider                   timeProvider
+}
+
+type realTime struct{}
+
+func (r *realTime) Now() time.Time {
+	return time.Now()
 }
 
 // NewCachingMigInfoProvider creates an instance of caching MigInfoProvider
-func NewCachingMigInfoProvider(cache *GceCache, migLister MigLister, gceClient AutoscalingGceClient, projectId string, concurrentGceRefreshes int) MigInfoProvider {
+func NewCachingMigInfoProvider(cache *GceCache, migLister MigLister, gceClient AutoscalingGceClient, projectId string, concurrentGceRefreshes int, migInstancesMinRefreshWaitTime time.Duration) MigInfoProvider {
 	return &cachingMigInfoProvider{
-		cache:                  cache,
-		migLister:              migLister,
-		gceClient:              gceClient,
-		projectId:              projectId,
-		concurrentGceRefreshes: concurrentGceRefreshes,
+		cache:                          cache,
+		migLister:                      migLister,
+		gceClient:                      gceClient,
+		projectId:                      projectId,
+		concurrentGceRefreshes:         concurrentGceRefreshes,
+		migInstancesMinRefreshWaitTime: migInstancesMinRefreshWaitTime,
+		migInstancesLastRefreshedInfo:  make(map[string]time.Time),
+		timeProvider:                   &realTime{},
 	}
 }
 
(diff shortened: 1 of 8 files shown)


cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
```diff
@@ -53,23 +54,27 @@ type MigInfoProvider interface {
...
-		migLister:              migLister,
-		gceClient:              gceClient,
-		projectId:              projectId,
-		concurrentGceRefreshes: concurrentGceRefreshes,
+		cache:                         cache,
+		migLister:                     migLister,
+		gceClient:                     gceClient,
+		projectId:                     projectId,
+		concurrentGceRefreshes:        concurrentGceRefreshes,
+		migInstancesMaxRefreshRate:    migInstancesMaxRefreshRate,
```
jayantjain93, line 76: I would rename this variable (incl. in different places) to something on this lines of `migInstancesMinRefreshWaitTime`. 
We want to signify this is the amount of time you have to wait before refetching the mig instances.

cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
```diff
@@ -158,12 +163,19 @@ func (c *cachingMigInfoProvider) findMigWithMatchingBasename(instanceRef GceRef)
...
+			return nil
+		}
+	}
 	klog.V(4).Infof("Regenerating MIG instances cache for %s", migRef.String())
 	instances, err := c.gceClient.FetchMigInstances(migRef)
 	if err != nil {
 		c.migLister.HandleMigIssue(migRef, err)
 		return err
 	}
+	c.migInstancesLastRefreshedInfo[migRef.String()] = time.Now()
```
jayantjain93, line 178: this only sets the refresh value for a successful fetch which return instances.
Don't we also want to wait for the same timeout for unsuccessful fetches?

Maybe we don't, can you drop a small rationale for it tho.

olagacek: added a comment, I don't think we should cache unsuccessful results, as the issues can be transient with those so another call might make sense

cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
olagacek: makes sense, changed

cluster-autoscaler/main.go
```diff
@@ -192,6 +192,7 @@ var (
 	balancingLabelsFlag                = multiStringFlag("balancing-label", "Specifies a label to use for comparing if two node groups are similar, rather than the built in heuristics. Setting this flag disables all other comparison logic, and cannot be combined with --balancing-ignore-label.")
 	awsUseStaticInstanceList           = flag.Bool("aws-use-static-instance-list", false, "Should CA fetch instance types in runtime or use a static list. AWS only")
 	concurrentGceRefreshes             = flag.Int("gce-concurrent-refreshes", 1, "Maximum number of concurrent refreshes per cloud object type.")
+	gceMigInstancesMaxRefreshRate      = flag.Duration("gce-mig-instances-max-refresh-rate", 5*time.Second, "Maximum rate at which GCE MIG instances from a given MIG can be refreshed.")
```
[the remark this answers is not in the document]
olagacek: done

cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -196,6 +196,8 @@ type AutoscalingOptions struct {
 	AWSUseStaticInstanceList bool
 	// ConcurrentGceRefreshes is the maximum number of concurrently refreshed instance groups or instance templates.
 	ConcurrentGceRefreshes int
+	// GCEMigInstancesMaxRefreshRate is the maximum rate at which GCE MIG instances from a given MIG can be refreshed.
+	GCEMigInstancesMaxRefreshRate time.Duration
```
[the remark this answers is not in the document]
olagacek: makes sense, created a new struct

cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -46,6 +46,16 @@ type NodeGroupAutoscalingOptions struct {
 	ScaleDownUnreadyTime time.Duration
 }
 
+// GCESpecificOptions contain autoscaling options specific to GCE cloud provider.
+type GCESpecificOptions struct {
+	// ConcurrentGceRefreshes is the maximum number of concurrently refreshed instance groups or instance templates.
+	ConcurrentGceRefreshes int
```
BigDarkClown, line 52: All these variables should not have `Gce` in them, as they are already defined in `GCESpecificOptions`.

cluster-autoscaler/cloudprovider/gce/mig_info_provider_test.go
```diff
@@ -928,6 +929,68 @@ func TestGetMigMachineType(t *testing.T) {
...
+			cache.migBaseNameCache = map[GceRef]string{mig.GceRef(): "base-instance-name"}
+			callCounter := make(map[GceRef]int)
+			client := &mockAutoscalingGceClient{
+				fetchMigInstances: fetchMigInstancesWithCounter(nil, callCounter),
+			}
+			migLister := NewMigLister(cache)
+			provider := NewCachingMigInfoProvider(cache, migLister, client, mig.GceRef().Project, 1, tc.refreshRateDuration)
+			_, err = provider.GetMigForInstance(instanceRef)
+			assert.NoError(t, err)
+			time.Sleep(tc.delayBetweenCalls)
```
BigDarkClown, line 986: Sleeps in the unit test are a bit fragile. Could you add a clock to the `cachingMigInfoProvider` and mock it in unit tests instead? That way we could avoid using sleep.

BigDarkClown (changes requested):

cluster-autoscaler/config/autoscaling_options.go
olagacek: makes sense, done.

cluster-autoscaler/cloudprovider/gce/mig_info_provider_test.go
olagacek: done

cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
```diff
@@ -158,12 +175,20 @@ func (c *cachingMigInfoProvider) findMigWithMatchingBasename(instanceRef GceRef)
 }
 
 func (c *cachingMigInfoProvider) fillMigInstances(migRef GceRef) error {
+	if val, ok := c.migInstancesLastRefreshedInfo[migRef.String()]; ok {
+		// do not regenerate MIG instances cache if last refresh happened recently.
```
x13n, line 179: Maybe log info that we're serving stale data from cache?

olagacek: done

PR #5671, opened 2023-04-07
Allow custom checks on whether or not a pod can be drained.

x13n:
#### What type of PR is this?

/kind cleanup

This makes the checks for whether a pod can be drained or not pluggable. It also rewrites existing unit tests for this logic to be table-based. I intend to rewrite the existing logic into new plugins as a follow-up. This change should not result in any behavior changes on its own.

Fixes #

```release-note
NONE
```

```docs

```


cluster-autoscaler/simulator/drain_test.go
```diff
@@ -249,8 +191,153 @@ func TestGetPodsToMove(t *testing.T) {
...
+		wantDs       []*apiv1.Pod
+		wantBlocking *drain.BlockingPod
+		wantErr      bool
+	}{
+		{
+			desc:    "Unreplicated pod",
+			pods:    []*apiv1.Pod{pod1},
+			wantErr: true,
+			wantBlocking: &drain.BlockingPod{
+				Pod:    pod1,
```
MaciekPytel, line 209: First of all, thanks for refactoring the unittest, much appreciated :)

Know putting my nitpicky hat on - maybe the test would be more readable if we renamed the pods to have more descriptive names? ex. rsPod, dsPod, localStoragePod?

cluster-autoscaler/simulator/drainability/filter.go
```diff
@@ -0,0 +1,65 @@
...
+	apiv1 "k8s.io/api/core/v1"
+)
+
+// Status indicates whether a pod can be drained, with an optional error message when not.
+// TODO(x13n): Move values from drain.BlockingPodReason to some typed string.
+type Status struct {
+	// Decided indicates whether a Filter has made a decision. `false` indicates that
+	// the Filter couldn't make a decision and that another one has to
+	// decide.
+	Decided bool
```
MaciekPytel, line 31: That's somewhat surprising - I'd expect most Filter to be able to give confident 'blocking', but not confident 'non-blocking'. In most cases all you know is that a pod can be drained as far as your criteria are considered, but how would you know what other checks are there?
I think your semantics actually make a lot of sense - "safe-to-evict: true" is a use-case for it. It's just really unintuitive and the naming doesn't help. "Undecided status" really isn't undecided, it's not blocking, just without overriding every other Filter. I think renaming things could make it more clear.

Of course I don't really have good naming suggestions... Maybe NewDrainableStatus() for current undecided and NewOverridingDrainableStatus()? But that's a mouthful and it doesn't really seem clear either

MaciekPytel: Hmm, that comment will only really make sense once you do the TODO above though, in the current form the naming actually matches reality pretty well. So I guess the question is should we optimize for end state already?

MaciekPytel:
/approve
Overall looks good, I left a few minor comments. The only one that I consider merge-blocking is the one about missing test case, if you disagree with the other ones just address this and I'll lgtm

cluster-autoscaler/simulator/drainability/filter.go
x13n: Yeah, I tried to model this after ACLs: each rule either matches (and yields ALLOW/DENY decision) or it doesn't match in which case it is up to the next rule to decide. Would renaming `Decided` to `Matched` (and perhaps - renaming `Filter` to `Rule`) address your concern?

cluster-autoscaler/simulator/drain_test.go
x13n: Yeah, I definitely should've done that while refactoring, but was just too lazy :) Will update in a moment.

cluster-autoscaler/simulator/drain_test.go
```diff
@@ -249,8 +191,153 @@ func TestGetPodsToMove(t *testing.T) {
...
+		{
+			desc:    "Second filter blocks",
+			pods:    []*apiv1.Pod{pod2},
+			filters: []drainability.Filter{cantDecide{}, neverDrain{}},
+			wantErr: true,
+			wantBlocking: &drain.BlockingPod{
+				Pod:    pod2,
+				Reason: drain.UnexpectedError,
+			},
+		},
```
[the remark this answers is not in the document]
x13n: Thanks, good catch! Added two more test cases.

PR #5672, opened 2023-04-10
feat: set `IgnoreDaemonSetsUtilization` per nodegroup for AWS

### LINKED ISSUE #5399: Override --ignore-daemonsets-utilization for Cluster Autoscaler on a per AWS ASG basis using ASG label
Cluster Autoscaler on AWS

Especially for smaller node types DaemonSet CPU/memory requests contribute a lot to node CPU/memory requests and therefore sometimes blocking node termination for scale-down. --ignore-daemonsets-utilization is already available as a global CAS option to solve this.

Similar to the already available approach of AWS ASG labels which override the global cluster-autoscaler settings for the labeled ASGs , see [https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md](url). I 'd like to introduce:

`k8s.io/cluster-autoscaler/node-template/autoscaling-options/ignoredaemonsetsutilization: true (overrides ignore-daemonsets-utilization value for that specific ASG)`

to override global setting per ASG using this label

vadasambar:
#### What type of PR is this?
/kind feature

We want to support the ability to specify `IgnoreDaemonSetsUtilization` per nodegroup through ASG tags. Why? It is to cater for the following case (among other possible cases):
1. `IgnoreDaemonSetsUtilization` is set `false` globally because we want to consider daemonsets when calculating node resource utilization. This works well for ASGs/nodegroups which have large nodes and daemonsets utilization doesn't contribute much to node resource utilization. Due to this, we have no problem during node scale down. 
2. But, we also have some ASGs/nodegroups with small nodes where daemonsets utilization contributes a lot to the node resource utilization. Due to this, the nodes might not get scaled down because node resource utilization doesn't fall below `ScaleDownUtilizationThreshold`. 

To solve this problem, we want to support setting `IgnoreDaemonSetsUtilization` per ASG/nodegroup through tag like `k8s.io/cluster-autoscaler/node-template/autoscaling-options/ignoredaemonsetsutilization: true` which would override the global value of `IgnoreDaemonSetsUtilization` (which is false)

More details: https://github.com/kubernetes/autoscaler/issues/5399

Fixes #5399

Nothing in particular. 

```release-note
Added: support to ignore daemonset resource utilization per nodegroup for AWS using the tag `k8s.io/cluster-autoscaler/node-template/autoscaling-options/ignoredaemonsetsutilization`
```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_manager_test.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_manager_test.go
@@ -130,6 +130,7 @@ func TestGetAsgOptions(t *testing.T) {
 		ScaleDownGpuUtilizationThreshold: 0.2,
 		ScaleDownUnneededTime:            time.Second,
 		ScaleDownUnreadyTime:             time.Minute,
+		IgnoreDaemonSetsUtilization:      false,
 	}
 
 	tests := []struct {
@@ -145,39 +146,60 @@ func TestGetAsgOptions(t *testing.T) {
 		{
 			description: "keep defaults on invalid tags values",
 			tags: map[string]string{
-				"scaledownutilizationthreshold": "not-a-float",
-				"scaledownunneededtime":         "not-a-duration",
-				"ScaleDownUnreadyTime":          "",
+				config.DefaultScaleDownUtilizationThresholdKey: "not-a-float",
+				config.DefaultScaleDownUnneededTimeKey:         "not-a-duration",
+				"ScaleDownUnreadyTime":                         "",
+				config.DefaultIgnoreDaemonSetsUtilizationKey:   "not-a-bool",
 			},
 			expected: &defaultOptions,
 		},
 		{
 			description: "use provided tags and fill missing with defaults",
 			tags: map[string]string{
-				"scaledownutilizationthreshold": "0.42",
-				"scaledownunneededtime":         "1h",
+				config.DefaultScaleDownUtilizationThresholdKey: "0.42",
+				config.DefaultScaleDownUnneededTimeKey:         "1h",
+				config.DefaultIgnoreDaemonSetsUtilizationKey:   "true",
 			},
 			expected: &config.NodeGroupAutoscalingOptions{
 				ScaleDownUtilizationThreshold:    0.42,
 				ScaleDownGpuUtilizationThreshold: defaultOptions.ScaleDownGpuUtilizationThreshold,
 				ScaleDownUnneededTime:            time.Hour,
 				ScaleDownUnreadyTime:             defaultOptions.ScaleDownUnreadyTime,
+				IgnoreDaemonSetsUtilization:      true,
+			},
+		},
+		{
+			description: "use provided tags (happy path)",
+			tags: map[string]string{
+				config.DefaultScaleDownUtilizationThresholdKey:    "0.42",
+				config.DefaultScaleDownUnneededTimeKey:            "1h",
+				config.DefaultScaleDownGpuUtilizationThresholdKey: "0.7",
+				config.DefaultScaleDownUnreadyTimeKey:             "25m",
+				config.DefaultIgnoreDaemonSetsUtilizationKey:      "true",
+			},
+			expected: &config.NodeGroupAutoscalingOptions{
+				ScaleDownUtilizationThreshold:    0.42,
+				ScaleDownGpuUtilizationThreshold: 0.7,
+				ScaleDownUnneededTime:            time.Hour,
+				ScaleDownUnreadyTime:             25 * time.Minute,
+				IgnoreDaemonSetsUtilization:      true,
 			},
 		},
 		{
 			description: "ignore unknown tags",
 			tags: map[string]string{
-				"scaledownutilizationthreshold":    "0.6",
-				"scaledowngpuutilizationthreshold": "0.7",
-				"scaledownunneededtime":            "1m",
-				"scaledownunreadytime":             "1h",
-				"notyetspecified":                  "42",
+				config.DefaultScaleDownUtilizationThresholdKey:    "0.6",
+				config.DefaultScaleDownGpuUtilizationThresholdKey: "0.7",
+				config.DefaultScaleDownUnneededTimeKey:            "1m",
+				config.DefaultScaleDownUnreadyTimeKey:             "1h",
+				"notyetspecified":                                 "42",
 			},
 			expected: &config.NodeGroupAutoscalingOptions{
 				ScaleDownUtilizationThreshold:    0.6,
 				ScaleDownGpuUtilizationThreshold: 0.7,
 				ScaleDownUnneededTime:            time.Minute,
 				ScaleDownUnreadyTime:             time.Hour,
+				IgnoreDaemonSetsUtilization:      false,
 			},
 		},
 	}
--- a/cluster-autoscaler/config/autoscaling_options.go
+++ b/cluster-autoscaler/config/autoscaling_options.go
@@ -48,6 +48,8 @@ type NodeGroupAutoscalingOptions struct {
 	MaxNodeProvisionTime time.Duration
 	// ZeroOrMaxNodeScaling means that a node group should be scaled up to maximum size or down to zero nodes all at once instead of one-by-one.
 	ZeroOrMaxNodeScaling bool
+	// IgnoreDaemonSetsUtilization sets if daemonsets utilization should be considered during node scale-down
+	IgnoreDaemonSetsUtilization bool
 }
 
 // GCEOptions contain autoscaling options specific to GCE cloud provider.
@@ -117,8 +119,6 @@ type AutoscalingOptions struct {
 	GRPCExpanderCert string
 	// GRPCExpanderURL is the url of the gRPC server when using the gRPC expander
 	GRPCExpanderURL string
-	// IgnoreDaemonSetsUtilization is whether CA will ignore DaemonSet pods when calculating resource utilization for scaling down
-	IgnoreDaemonSetsUtilization bool
 	// IgnoreMirrorPodsUtilization is whether CA will ignore Mirror pods when calculating resource utilization for scaling down
 	IgnoreMirrorPodsUtilization bool
 	// MaxGracefulTerminationSec is maximum number of seconds scale down waits for pods to terminate before
--- a/cluster-autoscaler/config/const.go
+++ b/cluster-autoscaler/config/const.go
@@ -16,6 +16,8 @@ limitations under the License.
 
 package config
 
+import "time"
+
 const (
 	// DefaultMaxClusterCores is the default maximum number of cores in the cluster.
 	DefaultMaxClusterCores = 5000 * 64
@@ -32,4 +34,14 @@ const (
 	DefaultScaleDownUnreadyTimeKey = "scaledownunreadytime"
 	// DefaultMaxNodeProvisionTimeKey identifies MaxNodeProvisionTime autoscaling option
 	DefaultMaxNodeProvisionTimeKey = "maxnodeprovisiontime"
+	// DefaultIgnoreDaemonSetsUtilizationKey identifies IgnoreDaemonSetsUtilization autoscaling option
+	DefaultIgnoreDaemonSetsUtilizationKey = "ignoredaemonsetsutilization"
+	// DefaultScaleDownUnneededTime identifies ScaleDownUnneededTime autoscaling option
+	DefaultScaleDownUnneededTime = 10 * time.Minute
+	// DefaultScaleDownUnreadyTime identifies ScaleDownUnreadyTime autoscaling option
+	DefaultScaleDownUnreadyTime = 20 * time.Minute
+	// DefaultScaleDownUtilizationThreshold identifies ScaleDownUtilizationThreshold autoscaling option
+	DefaultScaleDownUtilizationThreshold = 0.5
+	// DefaultScaleDownGpuUtilizationThreshold identifies ScaleDownGpuUtilizationThreshold autoscaling option
+	DefaultScaleDownGpuUtilizationThreshold = 0.5
 )
--- a/cluster-autoscaler/core/scaledown/actuation/actuator.go
+++ b/cluster-autoscaler/core/scaledown/actuation/actuator.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown/status"
 	"k8s.io/autoscaler/cluster-autoscaler/core/utils"
 	"k8s.io/autoscaler/cluster-autoscaler/metrics"
+	"k8s.io/autoscaler/cluster-autoscaler/processors"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator/clustersnapshot"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator/utilization"
@@ -52,10 +53,18 @@ type Actuator struct {
 	// This is a larger change to the code structure which impacts some existing actuator unit tests
 	// as well as Cluster Autoscaler implementations that may override ScaleDownSetProcessor
 	budgetProcessor *budgets.ScaleDownBudgetProcessor
+	configGetter    actuatorNodeGroupConfigGetter
+}
+
+// actuatorNodeGroupConfigGetter is an interface to limit the functions that can be used
+// from NodeGroupConfigProcessor interface
+type actuatorNodeGroupConfigGetter interface {
+	// GetIgnoreDaemonSetsUtilization returns IgnoreDaemonSetsUtilization value that should be used for a given NodeGroup.
+	GetIgnoreDaemonSetsUtilization(context *context.AutoscalingContext, nodeGroup cloudprovider.NodeGroup) (bool, error)
 }
 
 // NewActuator returns a new instance of Actuator.
-func NewActuator(ctx *context.AutoscalingContext, csr *clusterstate.ClusterStateRegistry, ndt *deletiontracker.NodeDeletionTracker, deleteOptions simulator.NodeDeleteOptions) *Actuator {
+func NewActuator(ctx *context.AutoscalingContext, csr *clusterstate.ClusterStateRegistry, ndt *deletiontracker.NodeDeletionTracker, deleteOptions simulator.NodeDeleteOptions, processors *processors.AutoscalingProcessors) *Actuator {
 	ndb := NewNodeDeletionBatcher(ctx, csr, ndt, ctx.NodeDeletionBatcherInterval)
 	return &Actuator{
 		ctx:                   ctx,
@@ -64,6 +73,7 @@ func NewActuator(ctx *context.AutoscalingContext, csr *clusterstate.ClusterState
 		nodeDeletionScheduler: NewGroupDeletionScheduler(ctx, ndt, ndb, NewDefaultEvictor(deleteOptions, ndt)),
 		budgetProcessor:       budgets.NewScaleDownBudgetProcessor(ctx),
 		deleteOptions:         deleteOptions,
+		configGetter:          processors.NodeGroupConfigProcessor,
 	}
 }
 
@@ -263,8 +273,14 @@ func (a *Actuator) scaleDownNodeToReport(node *apiv1.Node, drain bool) (*status.
 	if err != nil {
 		return nil, err
 	}
+
+	ignoreDaemonSetsUtilization, err := a.configGetter.GetIgnoreDaemonSetsUtilization(a.ctx, nodeGroup)
+	if err != nil {
+		return nil, err
+	}
+
 	gpuConfig := a.ctx.CloudProvider.GetNodeGpuConfig(node)
-	utilInfo, err := utilization.Calculate(nodeInfo, a.ctx.IgnoreDaemonSetsUtilization, a.ctx.IgnoreMirrorPodsUtilization, gpuConfig, time.Now())
+	utilInfo, err := utilization.Calculate(nodeInfo, ignoreDaemonSetsUtilization, a.ctx.IgnoreMirrorPodsUtilization, gpuConfig, time.Now())
 	if err != nil {
 		return nil, err
 	}
--- a/cluster-autoscaler/core/scaledown/actuation/actuator.go
+++ b/cluster-autoscaler/core/scaledown/actuation/actuator.go
@@ -33,7 +33,6 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown/status"
 	"k8s.io/autoscaler/cluster-autoscaler/core/utils"
 	"k8s.io/autoscaler/cluster-autoscaler/metrics"
-	"k8s.io/autoscaler/cluster-autoscaler/processors"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator/clustersnapshot"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator/utilization"
@@ -64,7 +63,7 @@ type actuatorNodeGroupConfigGetter interface {
 }
 
 // NewActuator returns a new instance of Actuator.
-func NewActuator(ctx *context.AutoscalingContext, csr *clusterstate.ClusterStateRegistry, ndt *deletiontracker.NodeDeletionTracker, deleteOptions simulator.NodeDeleteOptions, processors *processors.AutoscalingProcessors) *Actuator {
+func NewActuator(ctx *context.AutoscalingContext, csr *clusterstate.ClusterStateRegistry, ndt *deletiontracker.NodeDeletionTracker, deleteOptions simulator.NodeDeleteOptions, configGetter actuatorNodeGroupConfigGetter) *Actuator {
 	ndb := NewNodeDeletionBatcher(ctx, csr, ndt, ctx.NodeDeletionBatcherInterval)
 	return &Actuator{
 		ctx:                   ctx,
@@ -73,7 +72,7 @@ func NewActuator(ctx *context.AutoscalingContext, csr *clusterstate.ClusterState
 		nodeDeletionScheduler: NewGroupDeletionScheduler(ctx, ndt, ndb, NewDefaultEvictor(deleteOptions, ndt)),
 		budgetProcessor:       budgets.NewScaleDownBudgetProcessor(ctx),
 		deleteOptions:         deleteOptions,
-		configGetter:          processors.NodeGroupConfigProcessor,
+		configGetter:          configGetter,
 	}
 }
 
--- a/cluster-autoscaler/core/scaledown/actuation/actuator_test.go
+++ b/cluster-autoscaler/core/scaledown/actuation/actuator_test.go
@@ -43,32 +43,40 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown/deletiontracker"
 	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown/status"
 	. "k8s.io/autoscaler/cluster-autoscaler/core/test"
+	"k8s.io/autoscaler/cluster-autoscaler/processors/nodegroupconfig"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator/utilization"
 	kube_util "k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/taints"
 	. "k8s.io/autoscaler/cluster-autoscaler/utils/test"
 )
 
-func TestStartDeletion(t *testing.T) {
-	testNg := testprovider.NewTestNodeGroup("test", 0, 100, 3, true, false, "n1-standard-2", nil, nil)
+type startDeletionTestCase struct {
+	emptyNodes            []*budgets.NodeGroupView
+	drainNodes            []*budgets.NodeGroupView
+	pods                  map[string][]*apiv1.Pod
+	failedPodDrain        map[string]bool
+	failedNodeDeletion    map[string]bool
+	failedNodeTaint       map[string]bool
+	wantStatus            *status.ScaleDownStatus
+	wantErr               error
+	wantDeletedPods       []string
+	wantDeletedNodes      []string
+	wantTaintUpdates      map[string][][]apiv1.Taint
+	wantNodeDeleteResults map[string]status.NodeDeleteResult
+}
+
+func getStartDeletionTestCases(testNg *testprovider.TestNodeGroup, ignoreDaemonSetsUtilization bool, suffix string) map[string]startDeletionTestCase {
+	toBeDeletedTaint := apiv1.Taint{Key: taints.ToBeDeletedTaint, Effect: apiv1.TaintEffectNoSchedule}
+	dsUtilInfo := generateUtilInfo(2./8., 2./8.)
+
+	if ignoreDaemonSetsUtilization {
+		dsUtilInfo = generateUtilInfo(0./8., 0./8.)
+	}
+
 	atomic2 := sizedNodeGroup("atomic-2", 2, true)
 	atomic4 := sizedNodeGroup("atomic-4", 4, true)
-	toBeDeletedTaint := apiv1.Taint{Key: taints.ToBeDeletedTaint, Effect: apiv1.TaintEffectNoSchedule}
 
-	for tn, tc := range map[string]struct {
-		emptyNodes            []*budgets.NodeGroupView
-		drainNodes            []*budgets.NodeGroupView
-		pods                  map[string][]*apiv1.Pod
-		failedPodDrain        map[string]bool
-		failedNodeDeletion    map[string]bool
-		failedNodeTaint       map[string]bool
-		wantStatus            *status.ScaleDownStatus
-		wantErr               error
-		wantDeletedPods       []string
-		wantDeletedNodes      []string
-		wantTaintUpdates      map[string][][]apiv1.Taint
-		wantNodeDeleteResults map[string]status.NodeDeleteResult
-	}{
+	testCases := map[string]startDeletionTestCase{
 		"nothing to delete": {
 			emptyNodes: nil,
 			drainNodes: nil,
@@ -528,8 +536,8 @@ func TestStartDeletion(t *testing.T) {
 		"DS pods are evicted from empty nodes, but don't block deletion on error": {
 			emptyNodes: generateNodeGroupViewList(testNg, 0, 2),
 			pods: map[string][]*apiv1.Pod{
-				"test-node-0": {generateDsPod("test-node-0-ds-pod-0", "test-node-0"), generateDsPod("test-node-0-ds-pod-1", "test-node-0")},
-				"test-node-1": {generateDsPod("test-node-1-ds-pod-0", "test-node-1"), generateDsPod("test-node-1-ds-pod-1", "test-node-1")},
+				"test-node-0": generateDsPods(2, "test-node-0"),
+				"test-node-1": generateDsPods(2, "test-node-1"),
 			},
 			failedPodDrain: map[string]bool{"test-node-1-ds-pod-0": true},
 			wantStatus: &status.ScaleDownStatus{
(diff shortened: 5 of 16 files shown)


vadasambar:
@drmorr0 , @feiskyer 
This is a WIP implementation. I will mention the reviewers here once I am done. Meanwhile, if you have any feedback on the PR, I would love to have it. Please note that at the time of writing this PR, it's a quick, dirty and I think an incomplete implementation of the solution. It needs more polishing and thinking (I have outright replaced an interface).

vadasambar:
There are many cloud providers which have `GetOptions` implemented and many which don't. We might want to support it for all cloud providers which have implemented `GetOptions` and support setting tags on their specific nodegroup implementations. 
```
suraj@suraj:~/sandbox/autoscaler/cluster-autoscaler$ grep -R ") GetOptions" ./cloudprovider -A 2 --exclude-dir=vendor
./cloudprovider/rancher/rancher_nodegroup.go:func (ng *nodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/rancher/rancher_nodegroup.go-   return nil, cloudprovider.ErrNotImplemented
./cloudprovider/rancher/rancher_nodegroup.go-}
--
./cloudprovider/hetzner/hetzner_node_group.go:func (n *hetznerNodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/hetzner/hetzner_node_group.go-  return nil, cloudprovider.ErrNotImplemented
./cloudprovider/hetzner/hetzner_node_group.go-}
--
./cloudprovider/tencentcloud/tencentcloud_auto_scaling_group.go:func (asg *tcAsg) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/tencentcloud/tencentcloud_auto_scaling_group.go-        return nil, nil
./cloudprovider/tencentcloud/tencentcloud_auto_scaling_group.go-}
--
./cloudprovider/aws/aws_cloud_provider.go:func (ng *AwsNodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/aws/aws_cloud_provider.go-      if ng.asg == nil || ng.asg.Tags == nil || len(ng.asg.Tags) == 0 {
./cloudprovider/aws/aws_cloud_provider.go-              return &defaults, nil
--
./cloudprovider/test/test_cloud_provider.go:func (tng *TestNodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/test/test_cloud_provider.go-    return tng.opts, nil
./cloudprovider/test/test_cloud_provider.go-}
--
./cloudprovider/magnum/magnum_nodegroup.go:func (ng *magnumNodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/magnum/magnum_nodegroup.go-     return nil, cloudprovider.ErrNotImplemented
./cloudprovider/magnum/magnum_nodegroup.go-}
--
./cloudprovider/digitalocean/digitalocean_node_group.go:func (n *NodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/digitalocean/digitalocean_node_group.go-        return nil, cloudprovider.ErrNotImplemented
./cloudprovider/digitalocean/digitalocean_node_group.go-}
--
./cloudprovider/scaleway/scaleway_node_group.go:func (ng *NodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/scaleway/scaleway_node_group.go-        return nil, cloudprovider.ErrNotImplemented
./cloudprovider/scaleway/scaleway_node_group.go-}
--
./cloudprovider/cherryservers/cherry_node_group.go:func (ng *cherryNodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/cherryservers/cherry_node_group.go-     return nil, cloudprovider.ErrNotImplemented
./cloudprovider/cherryservers/cherry_node_group.go-}
--
./cloudprovider/mocks/NodeGroup.go:func (_m *NodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/mocks/NodeGroup.go-     ret := _m.Called(defaults)
./cloudprovider/mocks/NodeGroup.go-
--
./cloudprovider/bizflycloud/bizflycloud_node_group.go:func (n *NodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/bizflycloud/bizflycloud_node_group.go-  return nil, cloudprovider.ErrNotImplemented
./cloudprovider/bizflycloud/bizflycloud_node_group.go-}
--
./cloudprovider/kamatera/kamatera_node_group.go:func (n *NodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/kamatera/kamatera_node_group.go-        return nil, cloudprovider.ErrNotImplemented
./cloudprovider/kamatera/kamatera_node_group.go-}
--
./cloudprovider/brightbox/brightbox_node_group.go:func (ng *brightboxNodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/brightbox/brightbox_node_group.go-      return nil, cloudprovider.ErrNotImplemented
./cloudprovider/brightbox/brightbox_node_group.go-}
--
./cloudprovider/cloudstack/cloudstack_node_group.go:func (asg *asg) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/cloudstack/cloudstack_node_group.go-    return nil, cloudprovider.ErrNotImplemented
./cloudprovider/cloudstack/cloudstack_node_group.go-}
--
./cloudprovider/azure/azure_scale_set.go:func (scaleSet *ScaleSet) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/azure/azure_scale_set.go-       template, err := scaleSet.getVMSSFromCache()
./cloudprovider/azure/azure_scale_set.go-       if err != nil {
--
./cloudprovider/azure/azure_kubernetes_service_pool.go:func (agentPool *AKSAgentPool) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/azure/azure_kubernetes_service_pool.go- return nil, cloudprovider.ErrNotImplemented
./cloudprovider/azure/azure_kubernetes_service_pool.go-}
--
./cloudprovider/azure/azure_agent_pool.go:func (as *AgentPool) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/azure/azure_agent_pool.go-      return nil, cloudprovider.ErrNotImplemented
./cloudprovider/azure/azure_agent_pool.go-}
--
./cloudprovider/externalgrpc/externalgrpc_node_group.go:func (n *NodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/externalgrpc/externalgrpc_node_group.go-        ctx, cancel := context.WithTimeout(context.Background(), grpcTimeout)
./cloudprovider/externalgrpc/externalgrpc_node_group.go-        defer cancel()
--
./cloudprovider/packet/packet_node_group.go:func (ng *packetNodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/packet/packet_node_group.go-    return nil, cloudprovider.ErrNotImplemented
./cloudprovider/packet/packet_node_group.go-}
--
./cloudprovider/ionoscloud/ionoscloud_cloud_provider.go:func (n *nodePool) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/ionoscloud/ionoscloud_cloud_provider.go-        return nil, cloudprovider.ErrNotImplemented
./cloudprovider/ionoscloud/ionoscloud_cloud_provider.go-}
--
./cloudprovider/baiducloud/baiducloud_cloud_provider.go:func (asg *Asg) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/baiducloud/baiducloud_cloud_provider.go-        return nil, cloudprovider.ErrNotImplemented
./cloudprovider/baiducloud/baiducloud_cloud_provider.go-}
--
./cloudprovider/gce/gce_cloud_provider.go:func (mig *gceMig) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/gce/gce_cloud_provider.go-      return mig.gceManager.GetMigOptions(mig, defaults), nil
./cloudprovider/gce/gce_cloud_provider.go-}
--
./cloudprovider/civo/civo_node_group.go:func (n *NodeGroup) GetOptions(autoscaler.NodeGroupAutoscalingOptions) (*autoscaler.NodeGroupAutoscalingOptions, error) {
./cloudprovider/civo/civo_node_group.go-        return n.getOptions, nil
./cloudprovider/civo/civo_node_group.go-}
--
./cloudprovider/huaweicloud/huaweicloud_auto_scaling_group.go:func (asg *AutoScalingGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/huaweicloud/huaweicloud_auto_scaling_group.go-  return nil, cloudprovider.ErrNotImplemented
./cloudprovider/huaweicloud/huaweicloud_auto_scaling_group.go-}
--
./cloudprovider/vultr/vultr_node_group.go:func (n *NodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/vultr/vultr_node_group.go-      return nil, cloudprovider.ErrNotImplemented
./cloudprovider/vultr/vultr_node_group.go-}
--
./cloudprovider/exoscale/exoscale_node_group_sks_nodepool.go:func (n *sksNodepoolNodeGroup) GetOptions(_ config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/exoscale/exoscale_node_group_sks_nodepool.go-   return nil, cloudprovider.ErrNotImplemented
./cloudprovider/exoscale/exoscale_node_group_sks_nodepool.go-}
--
./cloudprovider/exoscale/exoscale_node_group_instance_pool.go:func (n *instancePoolNodeGroup) GetOptions(_ config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/exoscale/exoscale_node_group_instance_pool.go-  return nil, cloudprovider.ErrNotImplemented
./cloudprovider/exoscale/exoscale_node_group_instance_pool.go-}
--
./cloudprovider/linode/linode_node_group.go:func (n *NodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/linode/linode_node_group.go-    return nil, cloudprovider.ErrNotImplemented
./cloudprovider/linode/linode_node_group.go-}
--
./cloudprovider/clusterapi/clusterapi_nodegroup.go:func (ng *nodegroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/clusterapi/clusterapi_nodegroup.go-     return nil, cloudprovider.ErrNotImplemented
./cloudprovider/clusterapi/clusterapi_nodegroup.go-}
--
./cloudprovider/oci/oci_instance_pool.go:func (ip *InstancePoolNodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/oci/oci_instance_pool.go-       return nil, cloudprovider.ErrNotImplemented
./cloudprovider/oci/oci_instance_pool.go-}
--
./cloudprovider/alicloud/alicloud_auto_scaling_group.go:func (asg *Asg) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/alicloud/alicloud_auto_scaling_group.go-        return nil, cloudprovider.ErrNotImplemented
./cloudprovider/alicloud/alicloud_auto_scaling_group.go-}
--
./cloudprovider/kubemark/kubemark_linux.go:func (nodeGroup *NodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/kubemark/kubemark_linux.go-     return nil, cloudprovider.ErrNotImplemented
./cloudprovider/kubemark/kubemark_linux.go-}
--
./cloudprovider/ovhcloud/ovh_cloud_node_group.go:func (ng *NodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
./cloudprovider/ovhcloud/ovh_cloud_node_group.go-       // If node group autoscaling options nil, return defaults
./cloudprovider/ovhcloud/ovh_cloud_node_group.go-       if ng.Autoscaling == nil {
```

cluster-autoscaler/main.go
```diff
@@ -169,7 +169,7 @@ var (
 	grpcExpanderCert = flag.String("grpc-expander-cert", "", "Path to cert used by gRPC server over TLS")
 	grpcExpanderURL  = flag.String("grpc-expander-url", "", "URL to reach gRPC expander server.")
 
-	ignoreDaemonSetsUtilization = flag.Bool("ignore-daemonsets-utilization", false,
+	globaIgnoreDaemonSetsUtilization = flag.Bool("ignore-daemonsets-utilization", false,
```
vadasambar, line 172: Renamed `ignoreDaemonSetsUtilization` ->  `globaIgnoreDaemonSetsUtilization` in the code to avoid confusion with per nodegroup/ASG `ignoreDaemonSetsUtilization`

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
```diff
@@ -708,239 +714,268 @@ func TestStartDeletion(t *testing.T) {
...
-				if !found {
-					return true, nil, fmt.Errorf("node %q not found", getAction.GetName())
+	testSets := []testSet{
+		{
+			nodeGroup: testNg1,
+			testCases: getStartDeletionTestCases(testNg1),
+		},
+		{
+			nodeGroup: testNg2,
+			testCases: getStartDeletionTestCases(testNg2),
```
vadasambar, line 739: I am running the test cases twice one for each nodegroup defined above where in the first nodegroup, `IgnoreDaemonSetsUtilization: false` and for the second nodegroup `IgnoreDaemonSetsUtilization: true`. This is to test `IgnoreDaemonSetsUtilization` option on the nodegroup which is called [here](https://github.com/kubernetes/autoscaler/pull/5672/files#diff-73af8b1e646ba0094f2cc947fc77c027e52714908ed9b6452c6b9cd577229beeR320).

cluster-autoscaler/cloudprovider/aws/aws_manager_test.go
```diff
@@ -124,39 +125,60 @@ func TestGetAsgOptions(t *testing.T) {
 		{
 			description: "keep defaults on invalid tags values",
 			tags: map[string]string{
-				"scaledownutilizationthreshold": "not-a-float",
-				"scaledownunneededtime":         "not-a-duration",
-				"ScaleDownUnreadyTime":          "",
+				config.DefaultScaleDownUtilizationThresholdKey: "not-a-float",
+				config.DefaultScaleDownUnneededTimeKey:         "not-a-duration",
+				"ScaleDownUnreadyTime":                         "",
```
vadasambar, line 130: `ScaleDownUnreadyTime` here is a wrong key (for testing).

cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -115,8 +117,6 @@ type AutoscalingOptions struct {
 	GRPCExpanderCert string
 	// GRPCExpanderURL is the url of the gRPC server when using the gRPC expander
 	GRPCExpanderURL string
-	// IgnoreDaemonSetsUtilization is whether CA will ignore DaemonSet pods when calculating resource utilization for scaling down
-	IgnoreDaemonSetsUtilization bool
```
vadasambar, line 119: Moved this to `NodeGroupAutoscalingOptions`

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
```diff
@@ -253,24 +254,35 @@ func TestCropNodesToBudgets(t *testing.T) {
...
+
+type testSet struct {
+	nodeGroup *testprovider.TestNodeGroup
+	testCases map[string]startDeletionTestCase
+}
+
+func getStartDeletionTestCases(testNg *testprovider.TestNodeGroup, ignoreDaemonSetsUtilization bool) map[string]startDeletionTestCase {
 	toBeDeletedTaint := apiv1.Taint{Key: taints.ToBeDeletedTaint, Effect: apiv1.TaintEffectNoSchedule}
 
-	for tn, tc := range map[string]struct {
```
vadasambar, line 260: Abstracted the struct here into a separate definition.

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
```diff
@@ -632,29 +749,34 @@ func TestStartDeletion(t *testing.T) {
...
+						// this is nil because DaemonSetEvictionForOccupiedNodes is
+						// not enabled for drained nodes in this test suite
 						EvictedPods: nil,
-						UtilInfo:    generateUtilInfo(2./8., 2./8.),
+						UtilInfo:    dsUtilInfo,
 					},
 				},
 			},
 			wantDeletedNodes: []string{"empty-node-0", "empty-node-1"},
-			wantDeletedPods:  []string{"empty-node-0-ds-pod-0", "empty-node-0-ds-pod-1", "empty-node-1-ds-pod-1"},
```
vadasambar, line 657: This is not correct because no DS pods should be evicted since `DaemonSetEvictionForOccupiedNodes` is not enabled. Changed `wantDeletedPods` here to `nil`.

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
```diff
@@ -349,6 +361,49 @@ func TestStartDeletion(t *testing.T) {
 				"drain-node-1": {ResultType: status.NodeDeleteOk},
 			},
 		},
+		"DS pods and deletion with drain": {
```
vadasambar, line 364: Test case here and below are for testing code paths for when `IgnoreDaemonSetUtilization` is set to to `true` and `false`.

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
vadasambar: This can be verified by checking
For drain nodes: [`prepareNodeForDeletion`](https://github.com/kubernetes/autoscaler/blob/4400d80ddb81fb570a5ef608c1bb4a8260b34d83/cluster-autoscaler/core/scaledown/actuation/actuator.go#L356) -> [`evictor.DrainNode`](https://github.com/kubernetes/autoscaler/blob/4373c467fe6f0a3b1a8beb9bc79e0ae1ee560445/cluster-autoscaler/core/scaledown/actuation/drain.go#L84) -> [`podsToEvict`-(https://github.com/kubernetes/autoscaler/blob/4373c467fe6f0a3b1a8beb9bc79e0ae1ee560445/cluster-autoscaler/core/scaledown/actuation/drain.go#L264) -> [`daemonset.PodsToEvict`](https://github.com/kubernetes/autoscaler/blob/0470fdfc354f72d51a4b5c382f552c0dff3fac4a/cluster-autoscaler/utils/daemonset/daemonset.go#L56)
For empty nodes: [`prepareNodeForDeletion`](https://github.com/kubernetes/autoscaler/blob/4400d80ddb81fb570a5ef608c1bb4a8260b34d83/cluster-autoscaler/core/scaledown/actuation/actuator.go#L360) -> [`evictor.EvictDaemonSetPods`](https://github.com/kubernetes/autoscaler/blob/4373c467fe6f0a3b1a8beb9bc79e0ae1ee560445/cluster-autoscaler/core/scaledown/actuation/drain.go#L185) -> [`daemonset.PodsToEvict`](https://github.com/kubernetes/autoscaler/blob/0470fdfc354f72d51a4b5c382f552c0dff3fac4a/cluster-autoscaler/utils/daemonset/daemonset.go#L56)

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
```diff
@@ -708,239 +830,275 @@ func TestStartDeletion(t *testing.T) {
...
+		},
+		{
+			nodeGroup: testNg2,
+			testCases: getStartDeletionTestCases(testNg2, opts2.IgnoreDaemonSetsUtilization),
+		},
+	}
 
-			ds := generateDaemonSet()
+	for _, testSet := range testSets {
+		for tn, tc := range testSet.testCases {
```
vadasambar, line 867: I have added a comment for the code I have changed below this line (in this file). The changes are small. Most of the code below this line is same as before, just shifted by some lines.

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
```diff
@@ -708,239 +830,275 @@ func TestStartDeletion(t *testing.T) {
...
-
-			// Hook node deletion at the level of cloud provider, to gather which nodes were deleted, and to fail the deletion for
-			// certain nodes to simulate errors.
-			provider := testprovider.NewTestCloudProvider(nil, func(nodeGroup string, node string) error {
-				if tc.failedNodeDeletion[node] {
-					return fmt.Errorf("SIMULATED ERROR: won't remove node")
+				testSet.nodeGroup.SetCloudProvider(provider)
+				provider.InsertNodeGroup(testSet.nodeGroup)
+				for _, node := range nodesByName {
+					provider.AddNode(testSet.nodeGroup.Id(), node)
```
vadasambar, line 959: I have changed line 956 to 959 to use nodegroup from `testSet` instead of hardcoding `testNg1`

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
```diff
@@ -1191,8 +1349,18 @@ func removablePod(name string, node string) *apiv1.Pod {
 	}
 }
 
+func generateDsPods(count int, node string) []*apiv1.Pod {
```
vadasambar, line 1352: Added an extra helper to generate multiple DS pods.

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
```diff
@@ -1191,8 +1349,18 @@ func removablePod(name string, node string) *apiv1.Pod {
...
+	for i := 0; i < count; i++ {
+		name := fmt.Sprintf("ds-pod-%d", i)
+		result = append(result, generateDsPod(name, node))
+	}
+	return result
+}
+
 func generateDsPod(name string, node string) *apiv1.Pod {
-	pod := removablePod(name, node)
+	pod := removablePod(fmt.Sprintf("%s-%s", node, name), node)
```
vadasambar, line 1363: Earlier, we were using this function like this:
```
generateDsPod("drain-node-0-ds-pod-0", "drain-node-0")
```
And it would give us a DS pod with the name `drain-node-0-ds-pod-0`
Now, we use `generateDsPods` which calls `generateDsPod` like this:
```
generateDsPod("ds-pod-0", "drain-node-0")
```
This would give us the same result i.e., `drain-node-0-ds-pod-0`

cluster-autoscaler/utils/test/test_utils.go
```diff
@@ -67,6 +67,15 @@ func BuildTestPod(name string, cpu int64, mem int64) *apiv1.Pod {
 	return pod
 }
 
+// BuildDSTestPod creates a DaemonSet pod with cpu and memory.
+func BuildDSTestPod(name string, cpu int64, mem int64) *apiv1.Pod {
```
vadasambar, line 71: Helpers function in this file and actuator_test are similar. I am not sure why we didn't modify the functions here to use in actuator_test.go (maybe we didn't want to introduce dependency?). I have kept them separate for now i.e., I have added a new helper function for DS in actuator_test.go and introduced a new helper function for DS here.

vadasambar: I am not sure if merging the helper functions should be done in this PR but any feedback is welcome :pray:

gjtempleton:
/lgtm
LGTM from an AWS provider viewpoint. 

/assign @MaciekPytel 
Maciek, can you or someone else give the changes to the core code a review please?

x13n:
Maciek is out for few days, I'll try to take a look this week.

/assign

cluster-autoscaler/utils/test/test_utils.go
x13n: I think it would be nice to leverage [option pattern](https://golang.cafe/blog/golang-functional-options-pattern.html) in this file, so that tests can only set pod fields they care about:

```
BuildTestPod("pod", withCpu(2), withMemoryMb(123), withDaemonSetOwnerRef("some-uid"))
```

That would make this more readable and maintainable vs current approach, which duplicates existing code. This file should only really depend on k8s and be importable from other places like actuator tests.

cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -51,10 +52,18 @@ type Actuator struct {
...
+// actuatorNodeGroupConfigGetter is an interface to limit the functions that can be used
+// from NodeGroupConfigProcessor interface
+type actuatorNodeGroupConfigGetter interface {
+	// GetIgnoreDaemonSetsUtilization returns IgnoreDaemonSetsUtilization value that should be used for a given NodeGroup.
+	GetIgnoreDaemonSetsUtilization(context *context.AutoscalingContext, nodeGroup cloudprovider.NodeGroup) (bool, error)
 }
 
 // NewActuator returns a new instance of Actuator.
-func NewActuator(ctx *context.AutoscalingContext, csr *clusterstate.ClusterStateRegistry, ndt *deletiontracker.NodeDeletionTracker, deleteOptions simulator.NodeDeleteOptions) *Actuator {
+func NewActuator(ctx *context.AutoscalingContext, csr *clusterstate.ClusterStateRegistry, ndt *deletiontracker.NodeDeletionTracker, deleteOptions simulator.NodeDeleteOptions, processors *processors.AutoscalingProcessors) *Actuator {
```
x13n, line 66: nit: Do we need to have all processors passed to actuator? It should suffice to just have a param of type `actuatorNodeGroupConfigGetter` here.

cluster-autoscaler/config/const.go
```diff
@@ -32,4 +34,14 @@ const (
 	DefaultScaleDownUnreadyTimeKey = "scaledownunreadytime"
 	// DefaultMaxNodeProvisionTimeKey identifies MaxNodeProvisionTime autoscaling option
 	DefaultMaxNodeProvisionTimeKey = "maxnodeprovisiontime"
+	// DefaultIgnoreDaemonSetsUtilizationKey identifies IgnoreDaemonSetsUtilization autoscaling option
+	DefaultIgnoreDaemonSetsUtilizationKey = "ignoredaemonsetsutilization"
```
x13n, line 38: This is only used in `aws` cloudprovider, shouldn't this const be moved there?

cluster-autoscaler/core/scaledown/eligibility/eligibility_test.go
```diff
@@ -111,14 +114,32 @@ func TestFilterOutUnremovable(t *testing.T) {
 			scaleDownUnready: false,
 		},
 	}
+
+	allTestCases := testCases
+
+	// run all test cases again with `IgnoreDaemonSetsUtilization` set to true
```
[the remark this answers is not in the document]
x13n: This will cause different test cases to have the same `desc`, making them harder to debug. For the sake of readability, I'd just have one long list of test cases to check, but if you have to generate them programatically like this, please update `desc` as well.

cluster-autoscaler/config/const.go
MaciekPytel: +1

cluster-autoscaler/main.go
```diff
@@ -257,7 +257,9 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 			ScaleDownGpuUtilizationThreshold: *scaleDownGpuUtilizationThreshold,
 			ScaleDownUnneededTime:            *scaleDownUnneededTime,
 			ScaleDownUnreadyTime:             *scaleDownUnreadyTime,
-			MaxNodeProvisionTime:             *maxNodeProvisionTime,
+			// per nodegroup setting, defaulted to global setting
```
MaciekPytel, line 260: nit: Is this comment needed? Or at least - is it needed here? It's just as true for every other setting in this struct

MaciekPytel:
/hold
/approve
Sorry for not replying earlier :( Overall looks good, leaving the hold to address comments by @x13n

cluster-autoscaler/config/const.go
vadasambar: @x13n as of now, yes but this is a generic option like [`DefaultScaleDownUnneededTimeKey`](https://github.com/search?q=repo%3Akubernetes%2Fautoscaler%20DefaultScaleDownUnneededTimeKey&type=code) which can be supported by multiple cloud providers in the future (not limited to AWS provider). 

I am fine with moving it to AWS provider since no other cloud provider is using it now but say azure provider wants to support this key tomorrow, they'd have to define their own constant with the same name (or maybe worse, a different name like `ignore-daemonset-utilization` thereby increasing the confusion for users) in their own cloud provider. 

So far it seems like we have kept such keys common across providers. If we want to keep them separate, I think that would be going against what we have been doing so far. It can be argued that every cloud provider should have freedom to define this key as they like but that means it opens up possibility for different cloudproviders naming this key differently. I am not sure if that's a good idea. 

Any feedback on this is welcome. If both of you think it's better to move this to AWS provider, I am happy to move the key to AWS provider.

cluster-autoscaler/core/scaledown/actuation/actuator.go
vadasambar: Thank you for the comment. Makes sense. I have updated the code.

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
```diff
@@ -668,262 +781,294 @@ func TestStartDeletion(t *testing.T) {
...
+	testSets := []map[string]startDeletionTestCase{
+		// IgnoreDaemonSetsUtilization is false
+		getStartDeletionTestCases(testNg1, opts1.IgnoreDaemonSetsUtilization, "testNg1"),
+		// IgnoreDaemonSetsUtilization is true
+		getStartDeletionTestCases(testNg2, opts2.IgnoreDaemonSetsUtilization, "testNg2"),
+	}
+
+	for _, testSet := range testSets {
+		for tn, tc := range testSet {
+			t.Run(tn, func(t *testing.T) {
```
vadasambar, line 815: Code from this line and below is just a copy and paste of the older code (shifted by some lines). I haven't changed anything in the code until https://github.com/kubernetes/autoscaler/pull/5672/files#r1174841347

cluster-autoscaler/core/scaledown/eligibility/eligibility_test.go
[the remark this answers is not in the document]
vadasambar: Thank you for the suggestion. 

I have [started adding a suffix now](https://github.com/kubernetes/autoscaler/pull/5672/files#diff-e7f41c366e8f9e299ef7f726cb60c9fe2e6943fcdb6f2945a0df18d27f6bd015R152) like `<test-case-desc> IgnoreDaemonSetsUtilization=true` and `<test-case-desc> IgnoreDaemonSetsUtilization=false` and

cluster-autoscaler/utils/test/test_utils.go
vadasambar: @x13n option pattern is very interesting. Thank you for telling me about it. If you have any suggestions like these again, please tell me all about it :D  

As for implementing it, I think it might be better to do it in a separate PR. This is because `BuildTestPod` [is used in a lot of places](https://github.com/search?q=repo%3Akubernetes%2Fautoscaler%20BuildTestPod&type=code). 
Current PR is big as is. Making changes in the test utils would increase the PR size further (which might make it harder to review). Also, I think we could replace many functions in [test_utils.go](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/utils/test/test_utils.go) for both pods and nodes. I have started a separate PR to refactor test utils using options pattern: https://github.com/kubernetes/autoscaler/pull/5932

Another option is to create a separate function which wraps BuildTestPod and uses options pattern but I think that would mean we end up creating a function very similar to `BuildTestPod` (which can be a potential source of confusion). 

Let me know what you think.

cluster-autoscaler/main.go
vadasambar: Thank you for the comment. I don't think this comment is needed anymore (leftover from initial implementation of this PR). I have removed it.

vadasambar:
> @x13n I've addressed all the review comments. Can you please review this PR again pray

@x13n wanted to remind you about this ^

cluster-autoscaler/utils/test/test_utils.go
x13n: Separate PR makes a lot of sense, thanks!

x13n (approved):
Thanks for the changes, looks good to me!

/lgtm

x13n:
Oh, and the hold was meant for me, so removing:

/hold cancel

seh:
I read #5399 first, and then the discussion here, and it's still not clear to me what problem you encounter by ignoring _DaemonSet_ utilization globally. [The description](https://github.com/kubernetes/autoscaler/pull/5672#issue-1660293266) notes:
> `IgnoreDaemonSetsUtilization` is set false globally because we want to consider daemonsets when calculating node resource utilization. This works well for ASGs/nodegroups which have large nodes and daemonsets utilization doesn't contribute much to node resource utilization. Due to this, we have no problem during node scale down.

If the nodes are large enough and the contribution from _DaemonSets_ is proportionally small, what harm comes from ignoring their contribution on these large nodes too? What is the benefit of noting their contribution there?

vadasambar:
@seh 

> One problem with this (I think) is scale down wouldn't consider node resource utilization accurately because it would always ignore daemonset resource utilization. 

https://github.com/kubernetes/autoscaler/issues/5399#issuecomment-1501436228

> If the nodes are large enough and the contribution from DaemonSets is proportionally small, what harm comes from ignoring their contribution on these large nodes too? What is the benefit of noting their contribution there?

If you look at really small consumption in daemonsets, I think ignoring daemonset utilization makes sense. e.g., only 1 CPU is consumed by daemonsets on a 32 CPU node. It might not matter as much. 

But if you look at relatively larger consumption in daemonsets, say 1CPU in a 10 CPU node. For example imagine a scenario like this:
* `IgnoreDaemonSetsUtilization` is set to `true` (globally)
* Cluster has 2 nodegroups
   * nodegroup 1:  10 CPUs
   * nodegroup 2: 2 CPU
 * Daemonsets running per node use 1 CPU
 * `--scale-down-utilization-threshold` is set to 0.7 (i.e., scale down nodes with total cpu requests less than 70% of node's cpu capacity)
 * nodegroup 1 runs application workload which in total uses 6 CPUs. 
 * nodegroup 2 runs short-lived workloads which in total use 0.5 CPU

1. CPU utilization in nodegroup1 according to CA (cluster-autoscaler): 6/10 = > 60% (we are ignoring daemonsets)
2. CPU utilization in nodegroup2 according to CA: 0.5/2 => 25 % 

In case of 2, we want the node to get scaled down. But in case of 1, we might not want the node to get scaled down. Ignoring daemonset utilization for nodegroup1 means CA would scale down nodes in nodegroup 1 even though the node is actually using 70% of cpu (if you include daemonset utilization). This means, the pods would need to find a new home now. This can create potential disruption (application downtime until new new node is brought in by CA)

seh:
Thank you for the explanation. Regarding case one, the target utilization of 0.7 (70%) is higher than I've ever used, and I still don't see why you want to count the _DaemonSet_'s contribution there. If you want the node to survive your 70% threshold, it either needs 10% more "real" workload, or your actual threshold should be closer to 60% to get what you want. Counting the _DaemonSet_ as part of that utilization is falsifying the baseline atop which you sum the rest of the non-daemon workload's utilization. You could adjust the threshold instead.

PR #5674, opened 2023-04-11
Only evict ds pods after other pods are evicted

jacobwolfaws:

Daemonset pods are sometimes being evicted before non-ds pods, which in turn can prevent the other pods from terminating properly: https://github.com/kubernetes/autoscaler/issues/5240 . This would change the eviction behavior to evict/delete all non-DS pods before evicting DS pods. Previously opened (and accidentally closed) PR here: https://github.com/kubernetes/autoscaler/pull/5607 

Note: didn't add issue to fixes because this doesn't change default behavior to non-evict

Fixes #

open questions:

1. In the original code the waiter doesn't watch for successful daemonset pod deletion, only the non-daemonset pods. Do we want to add one just to prevent some other race condition from potentially emerging? Based on the `EvictDaemonSetPods` function, it seems like there's precedent for not watching for a successful daemonset pod deletion.
2. This behavior will differ slightly from the original code in the sense that daemonset pods are only evicted if non-ds pods are evicted successfully. This makes sense to me, do others agree?
3. It seems like it would be preferable for the default daemonset eviction to be false, but that’s not included in this PR (I'd be happy to force push a change to include this behavior). Would love to hear people’s thoughts on this.

```release-note
When nodes are drained on scale down, non-Daemonset pods are evicted prior to Daemonset pods.
```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/scaledown/actuation/drain.go
+++ b/cluster-autoscaler/core/scaledown/actuation/drain.go
@@ -92,6 +92,7 @@ func (e Evictor) DrainNodeWithPods(ctx *acontext.AutoscalingContext, node *apiv1
 	retryUntil := time.Now().Add(ctx.MaxPodEvictionTime)
 	confirmations := make(chan status.PodEvictionResult, len(pods))
 	daemonSetConfirmations := make(chan status.PodEvictionResult, len(daemonSetPods))
+
 	for _, pod := range pods {
 		evictionResults[pod.Name] = status.PodEvictionResult{Pod: pod, TimedOut: true, Err: nil}
 		go func(podToEvict *apiv1.Pod) {
@@ -99,16 +100,8 @@ func (e Evictor) DrainNodeWithPods(ctx *acontext.AutoscalingContext, node *apiv1
 		}(pod)
 	}
 
-	// Perform eviction of daemonset. We don't want to raise an error if daemonsetPod wasn't evict properly
-	for _, daemonSetPod := range daemonSetPods {
-		go func(podToEvict *apiv1.Pod) {
-			daemonSetConfirmations <- evictPod(ctx, podToEvict, true, retryUntil, e.EvictionRetryTime, e.evictionRegister)
-		}(daemonSetPod)
-
-	}
-
 	podsEvictionCounter := 0
-	for i := 0; i < len(pods)+len(daemonSetPods); i++ {
+	for i := 0; i < len(pods); i++ {
 		select {
 		case evictionResult := <-confirmations:
 			podsEvictionCounter++
@@ -116,13 +109,11 @@ func (e Evictor) DrainNodeWithPods(ctx *acontext.AutoscalingContext, node *apiv1
 			if evictionResult.WasEvictionSuccessful() {
 				metrics.RegisterEvictions(1)
 			}
-		case <-daemonSetConfirmations:
 		case <-time.After(retryUntil.Sub(time.Now()) + 5*time.Second):
 			if podsEvictionCounter < len(pods) {
 				// All pods initially had results with TimedOut set to true, so the ones that didn't receive an actual result are correctly marked as timed out.
 				return evictionResults, errors.NewAutoscalerError(errors.ApiCallError, "Failed to drain node %s/%s: timeout when waiting for creating evictions", node.Namespace, node.Name)
 			}
-			klog.Infof("Timeout when waiting for creating daemonSetPods eviction")
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


x13n:
Two high level comments:
- This behavior should be opt-in. As far as I can see from  the other PR & Slack, this was already discussed, but the code here will affect all users with no way out for them. Note: this doesn't have to be a command line flag. For some DaemonSets early eviction might be perfectly fine, so an alternative solution would be to allow annotating workloads that cannot be evicted early. I'd actually consider building on top of the existing `cluster-autoscaler.kubernetes.io/enable-ds-eviction` annotation and have `cluster-autoscaler.kubernetes.io/enable-ds-eviction`: `delayed` mean it will be evicted, but after everything that is not similarly `delayed`. Using an annotation has the advantage of being configurable even in managed environments like GKE.
- There are no tests for the code change, please add some to ensure the new code behaves as expected.

jacobwolfaws:
@x13n Thanks for the feedback! Just to clarify some things:
to point 1 from my original post) Do you think there should be a wait-for-eviction mechanism with daemonset pods like there is for non-ds pods? @MaciekPytel raised the concern of there being a race condition without such a wait, where there wouldn't be enough time between ds pod eviction and deleting the VM. The argument for no-wait is that there's already precedent here: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/scaledown/actuation/actuator.go#L344?

to point 3 from my original post) 
> As far as I can see from the other PR & Slack, this was already discussed, but the code here will affect all users with no way out for them.

Since changing default behavior was discussed here: https://github.com/kubernetes/autoscaler/issues/5240, I'd just like to confirm why default isn't being changed here so that we can close the above issue once this PR is submitted. There could be an `evict-pods-synchronously` annotation that allows for users to continue receiving this behavior, so it seems like there's a clear path forward if we wanted to change behavior. Is the argument for not changing default behavior that slower scale down causes more pain for users than this race condition that affects some users or is it that the default behavior was already changed one, which caused pain for users, and swapping back will come at the cost of introducing new pain and a "new" regression?

x13n:
Actually before writing code maybe it would be better to align on the desired behavior, otherwise there's a lot of risk for wasted effort (as the opinions change). Pulling in people from https://github.com/kubernetes/autoscaler/pull/5607:

/cc @MaciekPytel @elmiko 

I see two concerns @MaciekPytel had for delaying DS eviction:
> - Daemonsets often come with wildcard tolerations, so a ds may be rescheduled during CA drain effectively getting the drain stuck.

Current implementation lists DS pods once and triggers eviction for all of them, without waiting for results. Even if DS schedules a pod again, CA will not evict the new one. But I see no reason why this couldn't work the same way with delayed DS eviction.

> - Sequential, blocking evictions will decrease scale-down speed.

I think this one is a real concern for large clusters. However, with parallel drain now available, maybe it would suffice to make `--parallel-drain=true` the default? (And set `--max-drain-parallelism` to something bigger than 1.)

Also, to the question above:

> Do you think there should be a wait-for-eviction mechanism with daemonset pods like there is for non-ds pods?

I think yes, otherwise the eviction is really a no-op. Perhaps the DS eviction timeout should be configurable though (and shorter by default).

elmiko:
@x13n +1 to agreeing on desired behavior. i will have very high latency this week as i am at kubecon, and i think @MaciekPytel is as well. just fyi

elmiko:
ok, partially recovered from kubecon, leaving my thoughts

* this behavior should definitely be behind a flag of some sort, eg `--daemonset-eviction-delayed`
* given that we might want to group together several configurations (eg drain flags), i wonder if we want to start thinking about configuration profiles for the autoscaler?
* although i think the annotation is very convenient, and aligns with the current work, i am concerned about extending an API through annotations. that said, i don't have a better alternative, so...
* whatever we decide here, we need to make sure to add documentation, either as part of the FAQ or through an enhancement of its own. if we decide to add new flags and new annotation values then i think we should look at creating an enhancement to fully document how this will work, what users expectations there are, and what failure modes might be encountered.

jacobwolfaws:
> I wonder why CA evicts DS pods in the same time as application pods. With Kubernetes Cordon + Drain I have never seen an issue of DS pod such as CNI being removed before deployment (application) pod. Is there any reason for different behavior?

That was part of the argument for changing the default behavior, I don't think the conversation ever wrapped up with this.

SergeyKanzhelev:
Would be great to align CA behavior with graceful Node Shutdown behavior. Two relevant KEPs:

- https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2000-graceful-node-shutdown
- https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2712-pod-priority-based-graceful-node-shutdown

For end users and daemonset authors it will be easier to understand if all node termination options are following the same pattern

x13n:
+1 to making this consistent with node shutdown.

x13n:
https://github.com/kubernetes/autoscaler/pull/6139 implements priority based approach.

/close

PR #5679, opened 2023-04-12
[vpa] [recommender] post processor: maintain ratio between resources

### LINKED ISSUE #5212: VPA - The ratio between CPU and memory should be maintained

dbenque:
#### What type of PR is this?

/kind feature

Under some circumstances, it is preferable to maintain the resource ratio (CPU/Memory) as defined in the podTemplate. Situations are describe in https://github.com/kubernetes/autoscaler/issues/5212 . Thanks to this PR it is possible to define which resource should be calculated based on another resource recommendation.

For example one can say that CPU recommendation should be deliver by the recommender (as usual) but the memory has to be computed by maintaining the ratio initially defined in the podTemplateSpec. The use is able to describe this situation by setting an annotation: `vpa-post-processor.kubernetes.io/{container-name}_resourceRatios = {"cpu":"memory"}`

Fixes #5212 

This PR is based on https://github.com/kubernetes/autoscaler/pull/5676 . Thanks to that PR the ControllerFetcher can be passed to the post processing layer. I keep this one as `draft` till we conclude on #5676, then I will rebase.

Yes

```release-note
The "resource ratio post-processor" can be enabled using flag --resource-ratio-post-processor-enabled=true when starting the VPA recommender.
The post-processor would maintain resource ratio for containers referenced by a VPA annotation with this format:
vpa-post-processor.kubernetes.io/{containerName}_resourceRatios={resourceName:resourceName} where the value is a json representation of a map of resourceName to resourceName. For example:

vpa-post-processor.kubernetes.io/server_resourceRatios = {"cpu":"memory"}
```

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
@@ -76,6 +76,9 @@ type ControllerKeyWithAPIVersion struct {
 type ControllerFetcher interface {
 	// FindTopMostWellKnownOrScalable returns topmost well-known or scalable controller. Error is returned if controller cannot be found.
 	FindTopMostWellKnownOrScalable(controller *ControllerKeyWithAPIVersion) (*ControllerKeyWithAPIVersion, error)
+
+	// Get Pod Template
+	GetPodTemplateFromTopMostWellKnown(controller *ControllerKeyWithAPIVersion) (*corev1.PodTemplateSpec, error)
 }
 
 type controllerFetcher struct {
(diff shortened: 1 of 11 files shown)


vertical-pod-autoscaler/pkg/recommender/input/controller_fetcher/controller_fetcher.go
```diff
@@ -309,6 +312,43 @@ func (f *controllerFetcher) getOwnerForScaleResource(groupKind schema.GroupKind,
...
+	name := key.Name
+
+	obj, exists, err := informer.GetStore().GetByKey(namespace + "/" + name)
+	if err != nil {
+		return nil, err
+	}
+	if !exists {
+		return nil, fmt.Errorf("%s %s/%s does not exist", kind, namespace, name)
+	}
+	switch apiObj := obj.(type) {
```
iksaif, line 332: this part could be in its own function as well

dbenque: Agree, but I am trying to be consistent with the style, like in `getParentOfWellKnownController`

sophieliu15:
Hi @dbenque ! Now that https://github.com/kubernetes/autoscaler/pull/5676 is merged, do you mind rebasing this PR so we can give it another round of review? If you no longer work on this PR, please let us know and close it off. Thank you!

Issue #5675, opened 2023-04-12
vpa CRD not contain status subresource


wu0407:
**Which component are you using?**:

vertical-pod-autoscaler

**What version of the component are you using?**:

Component version:  0.13.0

**What k8s version are you using (`kubectl version`)?**:

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.10", GitCommit:"7e54d50d3012cf3389e43b096ba35300f36e0817", GitTreeState:"clean", BuildDate:"2022-08-17T18:32:54Z", GoVersion:"go1.17.13", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.10", GitCommit:"7e54d50d3012cf3389e43b096ba35300f36e0817", GitTreeState:"clean", BuildDate:"2022-08-17T18:26:59Z", GoVersion:"go1.17.13", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What environment is this in?**:
kubernetes 1.23

**What did you expect to happen?**:

vpa crd has stataus field in subresources, that `metadata.generation` only increase in spec change
relate pr #3606 

**What happened instead?**:

The vpa crd has empty subresources that leads to `metadata.generation` increase on vpa status update
https://github.com/kubernetes/autoscaler/blob/f173888d38b57f38c4328146c51284ebc62bfef2/vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml#L516

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

relate code: https://github.com/kubernetes/autoscaler/blob/f173888d38b57f38c4328146c51284ebc62bfef2/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L39-L52

PR #5680, opened 2023-04-13
add subresource status for vpa

wu0407:

Fixes: #5675

```release-note
NONE
```

```docs

```


vertical-pod-autoscaler/deploy/vpa-rbac.yaml
```diff
@@ -82,6 +82,13 @@ rules:
       - create
       - patch
       - delete
+  - apiGroups:
```
voelzmo, line 85: I don't think this belongs into the `system:vpa-checkpoint-actor` `ClusterRole` – this role only deals with access to `VPACheckpoints`. The role `system:vpa-actor` seems like a better semantical fit.
The only issue is that this role is also given to the `vpa-updater`, which doesn't need access to the `/status` subresource.

So probably we should introduce a new ClusterRole with name `system:vpa-status-actor` here, to keep things minimal.

voelzmo (changes requested):
Good catch, thanks! Just a few comments from my side, mostly dealing with semantics.

vertical-pod-autoscaler/deploy/vpa-rbac.yaml
wu0407: done

voelzmo:
Thanks for the changes! The only thing missing from my side is a release note pointing people to the new, required permissions. There are lots of people using different means to deploy the VPA (e.g. Helm charts or home-grown scripts/code), we need to point them and the maintainers of these solutions to the changes they need to make to their solutions.
What about something like this
```
Added the /status subresource to VPA
action required: If you're installing VPA with different means than the `hack/vpa-up.sh` script, you need to add the ClusterRole `system:vpa-status-actor` and corresponding ClusterRoleBinding yourself. Otherwise VPA will no longer be able to update the recommendations!
```

@jbartosik: I don't think we're adding these release notes automatically to the github releases yet, but I guess this would be helpful when compiling the release notes? WDYT?

jbartosik:
> Thanks for the changes! The only thing missing from my side is a release note pointing people to the new, required permissions. There are lots of people using different means to deploy the VPA (e.g. Helm charts or home-grown scripts/code), we need to point them and the maintainers of these solutions to the changes they need to make to their solutions. What about something like this
> 
> ```
> Added the /status subresource to VPA
> action required: If you're installing VPA with different means than the `hack/vpa-up.sh` script, you need to add the ClusterRole `system:vpa-status-actor` and corresponding ClusterRoleBinding yourself. Otherwise VPA will no longer be able to update the recommendations!
> ```
> 
> Joachim: I don't think we're adding these release notes automatically to the github releases yet, but I guess this would be helpful when compiling the release notes? WDYT?

Release notes from PR descriptions aren't automatically added to Github releases. But they make writing release notes easier. Please add

jbartosik (changes requested):
Also if I understand correctly the point here is that there will be user-visible API change.

Could you explain what will change? It's not clear to me from reading the isse

jbartosik:
Please fix the commit message. I think you can't reference issues / PRs in commit messages. You can (and should and do) do that in PR mesages

vertical-pod-autoscaler/deploy/vpa-rbac.yaml
```diff
@@ -57,6 +57,19 @@ rules:
...
+metadata:
+  name: system:vpa-status-actor
+rules:
+  - apiGroups:
+      - "autoscaling.k8s.io"
+    resources:
+      - verticalpodautoscalers/status
+    verbs:
+      - get
+      - patch
```
voelzmo, line 69: One thing I missed in the first review: Now that we have a `/status` subresource, the `system:vpa-actor` no longer needs permission to `patch` the VPA objects, so we can also remove that as part of this change.

voelzmo (changes requested):

voelzmo:
> Also if I understand correctly the point here is that there will be user-visible API change.

I don't think anything will change for users, except that currently `metadata.generation` is increased on every new recommendation, which isn't necessary. Another aspect is that the `recommender` has permissions to change all VPA `.spec`s, which it doesn't need – permissions for changing `/status` is sufficient.

For _operators_, though, we need to ensure that they update CRD, RBAC rules and VPA version at the same time, otherwise this will just stop working.

wu0407:
> > Thanks for the changes! The only thing missing from my side is a release note pointing people to the new, required permissions. There are lots of people using different means to deploy the VPA (e.g. Helm charts or home-grown scripts/code), we need to point them and the maintainers of these solutions to the changes they need to make to their solutions. What about something like this
> > ```
> > Added the /status subresource to VPA
> > action required: If you're installing VPA with different means than the `hack/vpa-up.sh` script, you need to add the ClusterRole `system:vpa-status-actor` and corresponding ClusterRoleBinding yourself. Otherwise VPA will no longer be able to update the recommendations!
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Joachim: I don't think we're adding these release notes automatically to the github releases yet, but I guess this would be helpful when compiling the release notes? WDYT?
> 
> Release notes from PR descriptions aren't automatically added to Github releases. But they make writing release notes easier. Please add

Added detail what to do in this pr and commit

vertical-pod-autoscaler/deploy/vpa-rbac.yaml
jbartosik: Line 47 still gives `patch` permission to role `system:vpa-actor`

wu0407: done

voelzmo:
The release notes look good so far, they just need to mention that operators also need to update their CRD. Sorry I missed that in the first suggestion I had about the notes.

mwielgus (approved):
/lgtm
/approve

PR #5689, opened 2023-04-17
Move mirror pods check to a dedicated rule

x13n:
#### What type of PR is this?

/kind cleanup

Part of the refactor to make drain rules more pluggable.

Fixes #

I had to slightly change the `Rule` interface and naming to allow skipping pods - initial implementation only allowed for draining or blocking, which left out skipping (applicable e.g. to mirror pods).

```release-note
NONE
```

```docs

```

/assign @MaciekPytel


cluster-autoscaler/simulator/drain.go
```diff
@@ -130,7 +134,7 @@ func checkPdbs(pods []*apiv1.Pod, pdbs []*policyv1.PodDisruptionBudget) (*drain.
 
 func drainabilityStatus(pod *apiv1.Pod, dr []drainability.Rule) drainability.Status {
```
BigDarkClown, line 135: I think this function would be more readable if as a default you return the `drainability.Status{Outcome: drainability.UndefinedOutcome}` instead of just `drainability.Status{}`. They're equivalent now, but this will both make this more readable and ensure that we won't break if someone accidentally changes the 0 value of `OutcomeType`

cluster-autoscaler/simulator/drain.go
```diff
@@ -68,25 +69,28 @@ func NewNodeDeleteOptions(opts config.AutoscalingOptions) NodeDeleteOptions {
 func GetPodsToMove(nodeInfo *schedulerframework.NodeInfo, deleteOptions NodeDeleteOptions, listers kube_util.ListerRegistry,
 	pdbs []*policyv1.PodDisruptionBudget, timestamp time.Time) (pods []*apiv1.Pod, daemonSetPods []*apiv1.Pod, blockingPod *drain.BlockingPod, err error) {
 	var drainPods, drainDs []*apiv1.Pod
+	drainabilityRules := deleteOptions.DrainabilityRules
+	if drainabilityRules == nil {
```
BigDarkClown, line 73: I'm not a fan of this defaulting, seems a bit like pushing the initialisation into the implementation. We already have this set in `NewNodeDeleteOptions`, which should be enough to make it always used. Additionally this can make testing more complex.

I get that we are now putting the core functionality into a parameter, which is a bit scary. But maybe we can do this better by splitting `DrainabilityRules` into something like `CoreDrainabilityRules` and `CustomDrainabilityRules`? This way we will ensure that the core rules are always executed, while also giving a way for the user to define their own rules if they so desire. This does not allow to override the core rules, but maybe that's for the better? User can still go around it if we execute the custom rules before core ones.

BigDarkClown (commented):
/assign @BigDarkClown

cluster-autoscaler/simulator/drain.go
x13n: `UndefinedOutcome` is a zero value by design, but I guess a little extra verbosity won't hurt. Updated.

cluster-autoscaler/simulator/drain.go
x13n: The thing is that sometimes we may want a custom rule that is evaluated before the core rules, sometimes after them. And sometimes - instead of them. For example, today we have a bunch of flags that control drainability behavior. Rather than having rules that check the flag value every time, I was envisioning this as picking the set of rules once based on the flags. If there's immutable "core" part, we won't be able to do this.

> Additionally this can make testing more complex.

Actually, this defaulting here is meant to make tests work. There's a [bunch of them](https://github.com/search?q=repo%3Akubernetes%2Fautoscaler+NodeDeleteOptions+path%3A**_test.go&type=code) that don't use `NewNodeDeleteOptions` to create the `NodeDeleteOptions`:  Alternatively, I can go and fix all these tests instead, perhaps this is going to be cleaner than handling `nil` here that way?

cluster-autoscaler/simulator/drain_test.go
```diff
@@ -49,15 +48,6 @@ func TestGetPodsToMove(t *testing.T) {
 			OwnerReferences: GenerateOwnerReferences("rs", "ReplicaSet", "extensions/v1beta1", ""),
 		},
 	}
-	manifestPod := &apiv1.Pod{
```
BigDarkClown, line 52: Why remove the `manifestPod` test cases? They should be still passing and they indirectly test the default rules set - seems useful to keep them.

cluster-autoscaler/simulator/drain.go
BigDarkClown: Ok, I see. Initially I thought that it would be still doable with just the "prefix rules", like youwould be able to override any default behaviour if you defined rules before. But the "UndefinedOutcome" cannot be passed to force end iteration over rules, so that wouldn't work :/

As a side note - maybe a drainabilityStatus which works as `UndefinedOutcome` (passes the pod to `GetPodsForDeletionOnNodeDrain`) *but* force ends iteration over rules would be useful?

cluster-autoscaler/simulator/drain.go
BigDarkClown: Thanks!

cluster-autoscaler/simulator/drain.go
x13n: It would make sense if `GetPodsForDeletionOnNodeDrain` was here to stay, but the idea was to migrate all of it into drainability rules.

cluster-autoscaler/simulator/drain_test.go
x13n: Ok, let me bring that back.

PR #5696, opened 2023-04-18
Extract 'SchedulablePods' method in orchestrator.

kisieland:
This simplifies the 'ComputeExpansionOption' and allows us to use defer
to revert the ClusterSnapshot rather than calling it explicitly on each exit from the method.

#### What type of PR is this?
/kind cleanup

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
+++ b/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
@@ -156,11 +156,7 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 			continue
 		}
 
-		option, err := o.ComputeExpansionOption(podEquivalenceGroups, nodeGroup, nodeInfo, upcomingNodes)
-		if err != nil {
-			return scaleUpError(&status.ScaleUpStatus{}, errors.ToAutoscalerError(errors.InternalError, err))
-		}
-
+		option := o.ComputeExpansionOption(podEquivalenceGroups, nodeGroup, nodeInfo, upcomingNodes)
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -436,30 +428,47 @@ func (o *ScaleUpOrchestrator) ComputeExpansionOption(
...
 	}
-	if err := o.autoscalingContext.ClusterSnapshot.AddNodeWithPods(nodeInfo.Node(), pods); err != nil {
+	if err := o.autoscalingContext.ClusterSnapshot.AddNodeWithPods(nodeInfo.Node(), allPods); err != nil {
 		klog.Errorf("Error while adding test Node: %v", err)
-		o.autoscalingContext.ClusterSnapshot.Revert()
-		return expander.Option{}, nil
+		return []*apiv1.Pod{}
 	}
 
+	schedulablePods := make([]*apiv1.Pod, 0)
```
BigDarkClown, line 466: micro nit: Why not just `var schedulablePods []*apiv1.Pod`?

BigDarkClown (commented):
/assign @BigDarkClown

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
kisieland: This code is just moved to a dedicated function (so I do not know)
Either way I applied the suggestion :)

PR #5697, opened 2023-04-18
Fix the buildTemplateLabels method for the ClusterApi provider

aleskandro:
The joinStringMaps call in the `buildTemplateLabels` method of the clusterApi provider should not overwrite any custom labels with the generic ones returned by `buildGenericLabels()`

/kind bug

Currently, when the joinStringMaps is called and the `unstructuredScalableResource` already defines one of the labels also returned by the `buildGenericLabels` method (e.g., `kubernetes.io/arch: arm64`), the generic value returned by `buildGenericLabels` gets precedence over the user-defined one. This should be considered an unwanted behavior.

For example, the `buildGenericLabels` method sets the `kubernetes.io/arch` value to amd64. 

If the nodeGroup is for arm64 machines and the label `kubernetes.io/arch: arm64` is set in the annotation   `capacity.cluster-autoscaler.kubernetes.io/labels: "kubernetes.io/arch=arm64"`, the autoscaler will finally consider the value for that label set to amd64, not arm64, leading it to be unable to choose the proper nodeGroup to scale when a workload with nodeSelector/Affinity requires a node having the label `kubernetes.io/arch: arm64`.

```release-note

```

```docs

```

cc @Prashanth684 @elmiko

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_nodegroup_test.go
+++ b/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_nodegroup_test.go
@@ -1344,6 +1344,31 @@ func TestNodeGroupTemplateNodeInfo(t *testing.T) {
 			},
 		},
 		{
+			name: "When the NodeGroup can scale from zero, the label capacity annotations merge with the pre-built node labels and take precedence if the same key is defined in both",
+			nodeGroupAnnotations: map[string]string{
+				memoryKey:   "2048Mi",
+				cpuKey:      "2",
+				gpuTypeKey:  gpuapis.ResourceNvidiaGPU,
+				gpuCountKey: "1",
+				labelsKey:   "kubernetes.io/arch=arm64,my-custom-label=custom-value",
+			},
+			config: testCaseConfig{
+				expectedErr: nil,
+				expectedCapacity: map[corev1.ResourceName]int64{
+					corev1.ResourceCPU:        2,
+					corev1.ResourceMemory:     2048 * 1024 * 1024,
+					corev1.ResourcePods:       110,
+					gpuapis.ResourceNvidiaGPU: 1,
+				},
+				expectedNodeLabels: map[string]string{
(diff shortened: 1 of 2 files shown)


elmiko (commented):
this looks like a good ugprade, thanks for the merge request @aleskandro . i have one suggestion though, could we add some unit tests to ensure that this works the way we want and so that we don't regress in the future.

elmiko:
we could add something in here, https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_nodegroup_test.go#L1294

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_nodegroup_test.go
```diff
@@ -1323,13 +1323,10 @@ func TestNodeGroupTemplateNodeInfo(t *testing.T) {
 				cpuKey:      "2",
 				gpuTypeKey:  gpuapis.ResourceNvidiaGPU,
 				gpuCountKey: "1",
+				labelsKey:   "kubernetes.io/arch=arm64,my-custom-label=custom-value",
```
aleskandro, line 1326: > this looks like a good ugprade, thanks for the merge request @aleskandro . i have one suggestion though, could we add some unit tests to ensure that this works the way we want and so that we don't regress in the future.

Hi @elmiko, thanks for the review. Do you think it is good to patch one of the current cases to also consider the labels set via the capacity annotations or were you meant to have a dedicated new test case?

(The `nodeLabels` map I deleted seems not needed when `includeNodes` is false, is it?)

elmiko: i'm ok with either as long as we have all the conditions of the new logic tested for both positive and negative assertions.

elmiko (commented):
thanks for the updates @aleskandro !

/approve
/lgtm

PR #5705, opened 2023-04-24
bugfix: fix race condition between CA fetching list of scheduled pods…

damikag:
There is a possibility of a race condition because the scheduled and unscheduled pods are fetched in two subsequent calls.
Between those two operations, it gets a lot of data, for example node info, information about upcoming nodes etc. In case of large cluster it can take up to a few seconds. It is possible that during that time scheduler will schedule some pods onto nodes and it won't have information about those in neither scheduled nor unscheduled pod list. In result, CA might incorrectly remove a node with pods running on it, unaware of the fact that something got scheduled there.

#### What type of PR is this?

/kind bug

This PR introduces a new method `ScheduledAndUnscheduledPodLister() PodLister `to the `ListerRegistry interface` in `cluster-autoscaler/utils/kubernetes/listers.go`

`RunOnce(currentTime time.Time) caerrors.AutoscalerError` method of `cluster-autoscaler/core/static_autoscaler.go` is updated to use `ScheduledAndUnscheduledPodLister() `and fetch all scheduled and unscheduled pods in one call and separate scheduled and unscheduled pods after that.

```NONE
```


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -288,8 +288,21 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
...
-	unschedulablePodLister := a.UnschedulablePodLister()
-	scheduledPodLister := a.ScheduledPodLister()
+	scheduledAndUnscheduledPodLister := a.ScheduledAndUnscheduledPodLister()
+	scheduledAndUnscheduledPods, err := scheduledAndUnscheduledPodLister.List()
+	if err != nil {
+		klog.Errorf("Failed to list scheduled and unscheduled pods: %v", err)
+		return caerrors.ToAutoscalerError(caerrors.ApiCallError, err)
+	}
+
+	var originalScheduledPods, unschedulablePods []*apiv1.Pod
```
olagacek, line 298: nit: let's extract this logic to a separate funcion

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -218,6 +218,30 @@ func NewScheduledPodLister(kubeClient client.Interface, stopchannel <-chan struc
 	}
 }
 
+// ScheduledAndUnscheduledPodLister lists scheduled and unscheduled pods.
+type ScheduledAndUnscheduledPodLister struct {
```
olagacek, line 222: To be precise it should be ScheduledAndUnschedulablePodLister (unschedulable instead of unscheduled :) ). Fix in other places as well.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -288,8 +288,21 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
 	a.DebuggingSnapshotter.StartDataCollection()
 	defer a.DebuggingSnapshotter.Flush()
 
-	unschedulablePodLister := a.UnschedulablePodLister()
-	scheduledPodLister := a.ScheduledPodLister()
+	scheduledAndUnscheduledPodLister := a.ScheduledAndUnscheduledPodLister()
```
olagacek, line 291: nit: let's insert this whole logic after call to obtain node list, as it was before

olagacek (changes requested):

cluster-autoscaler/core/static_autoscaler.go
damikag: Done

cluster-autoscaler/core/static_autoscaler.go
damikag: Done

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -902,23 +902,23 @@ func TestStaticAutoscalerRunOnceWithFilteringOnBinPackingEstimator(t *testing.T)
 	// Scale up
 	readyNodeLister.SetNodes([]*apiv1.Node{n1, n2})
 	allNodeLister.SetNodes([]*apiv1.Node{n1, n2})
-	scheduledPodMock.On("List").Return([]*apiv1.Pod{p3, p4}, nil).Times(2) // 1 to get pods + 1 when building nodeInfo map
-	unschedulablePodMock.On("List").Return([]*apiv1.Pod{p1}, nil).Once()
+	scheduledPodMock.On("List").Return([]*apiv1.Pod{p3, p4}, nil).Once()
+	scheduledAndUnschedulablePodMock.On("List").Return([]*apiv1.Pod{p3, p4, p1}, nil).Once()
```
BigDarkClown, line 906: micro nit: This may be just my OCD, but these pods not being sorted by number gives my chills.

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -1002,15 +1002,15 @@ func TestStaticAutoscalerRunOnceWithFilteringOnUpcomingNodesEnabledNoScaleUp(t *
 	// Scale up
 	readyNodeLister.SetNodes([]*apiv1.Node{n2, n3})
 	allNodeLister.SetNodes([]*apiv1.Node{n2, n3})
-	scheduledPodMock.On("List").Return([]*apiv1.Pod{p2, p3}, nil).Times(2) // 1 to get pods + 1 when building nodeInfo map
-	unschedulablePodMock.On("List").Return([]*apiv1.Pod{p1}, nil).Once()
+	scheduledPodMock.On("List").Return([]*apiv1.Pod{p2, p3}, nil).Once()
+	scheduledAndUnschedulablePodMock.On("List").Return([]*apiv1.Pod{p2, p3, p1}, nil).Once()
```
BigDarkClown, line 1006: Same micro nit about sorting

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -49,41 +49,41 @@ type ListerRegistry interface {
...
+		replicationControllerLister:        replicationControllerLister,
+		jobLister:                          jobLister,
+		replicaSetLister:                   replicaSetLister,
+		statefulSetLister:                  statefulSetLister,
 	}
 }
 
 // NewListerRegistryWithDefaultListers returns a registry filled with listers of the default implementations
 func NewListerRegistryWithDefaultListers(kubeClient client.Interface, stopChannel <-chan struct{}) ListerRegistry {
-	unschedulablePodLister := NewUnschedulablePodLister(kubeClient, stopChannel)
```
BigDarkClown, line 86: `NewUnschedulablePodLister`, `NewUnschedulablePodInNamespaceLister` and `UnschedulablePodLister` are no longer used anywhere. Could you remove them? We don't want to keep dead code.

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -174,46 +174,51 @@ func (unschedulablePodLister *UnschedulablePodLister) List() ([]*apiv1.Pod, erro
...
-	return &UnschedulablePodLister{
+
+	return &ScheduledPodLister{
 		podLister: podLister,
 	}
 }
 
-// ScheduledPodLister lists scheduled pods.
-type ScheduledPodLister struct {
+// ScheduledAndUnschedulablePodLister lists scheduled and unschedulabke pods.
```
BigDarkClown, line 202: nit: type in `unschedulable`

cluster-autoscaler/core/static_autoscaler_test.go
damikag: Done

cluster-autoscaler/core/static_autoscaler_test.go
damikag: Done

cluster-autoscaler/utils/kubernetes/listers.go
damikag: Done

cluster-autoscaler/utils/kubernetes/listers.go
damikag: Done

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -302,17 +301,19 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
...
 		a.AutoscalingContext, allNodes, readyNodes, currentTime); abortLoop {
 		return err
 	}
 
+	scheduledAndUnschedulablePods, err := scheduledAndUnschedulablePodLister.List()
+	if err != nil {
+		klog.Errorf("Failed to list scheduled and unschedulable pods: %v", err)
+		return caerrors.ToAutoscalerError(caerrors.ApiCallError, err)
+	}
+	originalScheduledPods, unschedulablePods := seperateScheduledAndUnschedulablePods(scheduledAndUnschedulablePods)
```
towca, line 315: If the object's responsibility is to list scheduled and unschedulable pods, I don't think it should be the caller's responsibility to distinguish between them, the object should handle that. I'd change the `List` method to return scheduled and unschedulable pods in separate values in the first place.

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -153,67 +152,51 @@ type PodLister interface {
...
-type ScheduledPodLister struct {
+// ScheduledAndUnschedulablePodLister lists scheduled and unschedulable pods.
+type ScheduledAndUnschedulablePodLister struct {
 	podLister v1lister.PodLister
 }
 
-// List returns all scheduled pods.
-func (lister *ScheduledPodLister) List() ([]*apiv1.Pod, error) {
+// List returns all scheduled and unschedulable pods.
+func (lister *ScheduledAndUnschedulablePodLister) List() ([]*apiv1.Pod, error) {
```
towca, line 186: This is a major change in the logic. "Unschedulable" pods are pods that the scheduler explicitly marks as such using the `PodScheduled` condition with `PodReasonUnschedulable`, not just pods that don't have a `nodeName` assigned. I can't recall particular examples, but I think the scheduler intentionally doesn't mark some pods as unschedulable (while leaving `nodeName` empty). Currently CA would not react to such pods, after these changes it would start to.

In particular in certain scaling scenarios we rely on scheduler's marking of pods as unschedulable as a rate limiting mechanism (if a lot of new pods are created at once, scheduler effectively rate-limits how many CA has to handle at once because it has to mark the pods first). We'd lose that ability after these changes.

IMO this method should list the pods from `podLister`, filter them into two lists:
* Pods with `nodeName != ""` -> scheduled pods
* Pods with `nodeName == ""` AND the Unschedulable condition present (you can use the old logic for that check) -> unschedulable pods

And then return both lists separately. This should keep the logic intact and improve responsibility separation (it should be the listers job to separate).

towca (changes requested):

cluster-autoscaler/utils/kubernetes/listers.go
damikag: Done

cluster-autoscaler/core/static_autoscaler.go
damikag: Done

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -153,67 +153,73 @@ type PodLister interface {
...
+	return &ScheduledPodLister{
 		podLister: podLister,
 	}
 }
 
-// ScheduledPodLister lists scheduled pods.
-type ScheduledPodLister struct {
+// ScheduledAndUnschedulablePodLister lists scheduled and unschedulable pods separately.
+type ScheduledAndUnschedulablePodLister interface {
+	List() ([]*apiv1.Pod, []*apiv1.Pod, error)
```
towca, line 183: Could you specify which return value is supposed to be which - here and in the List implementation below? For the interface, it's critical so that the implementations know which list to return in which value. For the implementation, it improves readability to always have the return values named if there are multiple values of the same type.

I'd also mention something about the two lists being obtained at the same point in time in the comment - the current wording actually sort of suggests something else (the "separately" part). Replacing "separately" with "obtained at the same point in time" should be enough.

towca:
Just one minor readability comment, other than that LGTM

cluster-autoscaler/utils/kubernetes/listers.go
damikag: Done

towca:
/lgtm
/approve

Looks good, thanks for the changes!

towca (approved):

PR #5708, opened 2023-04-25
feat: support custom scheduler config (without extenders)

### LINKED ISSUE #5106: Cluster autoscaler to support custom schedulers' Filter logic
Cluster autoscaler (CA)

When a Pod is being scheduled with a custom scheduler (i.e. non default-scheduler) that uses the Filter API, it can be marked as Unschedulable. When CA polls for unschedulable pods, it uses a different Filter API (I believe the non default-scheduler) to determine whether it can be scheduled on existing nodes ([from this block](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/filteroutschedulable/filter_out_schedulable.go#L49-L63)). This difference in logic can cause a pod to be stuck in Pending given that differing logic is used to determine its placement.  

To allow CA to be extended to use the same Filter API as the scheduler used for the Pod that is currently in Unscheduled state, so that it can scale up the node that is required.

N/A

**Additional context.**:

I am doing testing with the placement-policy scheduler plugin for this.

- https://github.com/Azure/placement-policy-scheduler-plugins/

I am also open to any other workaround to allow CA to play nicely with other schedulers, of which I am probably unaware of.

Thanks for taking the time to read this!

vadasambar:

#### What type of PR is this?

/kind feature

**Problem**
1. We don't have a mechanism to to tweak configuration for default plugins acting on `PreFilter` and `Filter` extension points (which is used in predicate checker).
2. We don't have a way to support simulating custom scheduler logic without re-compiling the CA

This PR introduces `--scheduler-config-file` flag to 
1. tweak configuration for default plugins
~2. add support for [scheduler extenders](https://github.com/kubernetes/design-proposals-archive/blob/main/scheduling/scheduler_extender.md) so that the predicate checker can call an external filter URL.~ DECIDED AGAINST GOING FORWARD WITH EXTENDERS. Check https://github.com/kubernetes/autoscaler/pull/5708#issuecomment-1538647324 and https://github.com/kubernetes/autoscaler/pull/5708#issuecomment-1539546973 

Fixes #5106

```release-note
Added: `--scheduler-config-file` flag which supports configuring `PreFilter` and `Filter` extension points for in-tree scheduler plugins used to run scheduler simulations in CA 
```

```docs

```

### DIFF
--- a/cluster-autoscaler/config/autoscaling_options.go
+++ b/cluster-autoscaler/config/autoscaling_options.go
@@ -18,6 +18,8 @@ package config
 
 import (
 	"time"
+
+	scheduler_config "k8s.io/kubernetes/pkg/scheduler/apis/config"
 )
 
 // GpuLimits define lower and upper bound on GPU instances of given type in cluster
@@ -166,6 +168,9 @@ type AutoscalingOptions struct {
 	// ScaleDownSimulationTimeout defines the maximum time that can be
 	// spent on scale down simulation.
 	ScaleDownSimulationTimeout time.Duration
+	// SchedulerConfig allows changing configuration of in-tree
+	// scheduler plugins acting on PreFilter and Filter extension points
+	SchedulerConfig *scheduler_config.KubeSchedulerConfiguration
 	// NodeDeletionDelayTimeout is maximum time CA waits for removing delay-deletion.cluster-autoscaler.kubernetes.io/ annotations before deleting the node.
 	NodeDeletionDelayTimeout time.Duration
 	// WriteStatusConfigMap tells if the status information should be written to a ConfigMap
--- a/cluster-autoscaler/config/const.go
+++ b/cluster-autoscaler/config/const.go
@@ -19,6 +19,10 @@ package config
 import "time"
 
 const (
+	// SchedulerConfigFileFlag is the name of the flag
+	// for passing in custom scheduler config for in-tree scheduelr plugins
+	SchedulerConfigFileFlag = "scheduler-config-file"
+
 	// DefaultMaxClusterCores is the default maximum number of cores in the cluster.
 	DefaultMaxClusterCores = 5000 * 64
 	// DefaultMaxClusterMemory is the default maximum number of gigabytes of memory in cluster.
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -56,6 +56,7 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/simulator/clustersnapshot"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/errors"
 	kube_util "k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
+	scheduler_util "k8s.io/autoscaler/cluster-autoscaler/utils/scheduler"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/units"
 	"k8s.io/autoscaler/cluster-autoscaler/version"
 	kube_client "k8s.io/client-go/kubernetes"
@@ -68,6 +69,7 @@ import (
 	"k8s.io/component-base/config/options"
 	"k8s.io/component-base/metrics/legacyregistry"
 	"k8s.io/klog/v2"
+	scheduler_config "k8s.io/kubernetes/pkg/scheduler/apis/config"
 )
 
 // MultiStringFlag is a flag for passing multiple parameters using same flag
@@ -133,6 +135,7 @@ var (
 			"for scale down when some candidates from previous iteration are no longer valid."+
 			"When calculating the pool size for additional candidates we take"+
 			"max(#nodes * scale-down-candidates-pool-ratio, scale-down-candidates-pool-min-count).")
+	schedulerConfigFile         = flag.String(config.SchedulerConfigFileFlag, "", "scheduler-config allows changing configuration of in-tree scheduler plugins acting on PreFilter and Filter extension points")
 	nodeDeletionDelayTimeout    = flag.Duration("node-deletion-delay-timeout", 2*time.Minute, "Maximum time CA waits for removing delay-deletion.cluster-autoscaler.kubernetes.io/ annotations before deleting the node.")
 	nodeDeletionBatcherInterval = flag.Duration("node-deletion-batcher-interval", 0*time.Second, "How long CA ScaleDown gather nodes to delete them in batch.")
 	scanInterval                = flag.Duration("scan-interval", 10*time.Second, "How often cluster is reevaluated for scale up or down")
@@ -274,6 +277,15 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 		*maxEmptyBulkDeleteFlag = *maxScaleDownParallelismFlag
 	}
 
+	var parsedSchedConfig *scheduler_config.KubeSchedulerConfiguration
+	// if scheduler config flag was set by the user
+	if pflag.CommandLine.Changed(config.SchedulerConfigFileFlag) {
+		parsedSchedConfig, err = scheduler_util.ConfigFromPath(*schedulerConfigFile)
+	}
+	if err != nil {
+		klog.Fatalf("Failed to get scheduler config: %v", err)
+	}
+
 	return config.AutoscalingOptions{
 		NodeGroupDefaults: config.NodeGroupAutoscalingOptions{
 			ScaleDownUtilizationThreshold:    *scaleDownUtilizationThreshold,
@@ -316,6 +328,7 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 		ScaleDownNonEmptyCandidatesCount: *scaleDownNonEmptyCandidatesCount,
 		ScaleDownCandidatesPoolRatio:     *scaleDownCandidatesPoolRatio,
 		ScaleDownCandidatesPoolMinCount:  *scaleDownCandidatesPoolMinCount,
+		SchedulerConfig:                  parsedSchedConfig,
 		WriteStatusConfigMap:             *writeStatusConfigMapFlag,
 		StatusConfigMapName:              *statusConfigMapName,
 		BalanceSimilarNodeGroups:         *balanceSimilarNodeGroupsFlag,
@@ -422,7 +435,8 @@ func buildAutoscaler(debuggingSnapshotter debuggingsnapshot.DebuggingSnapshotter
 
 	eventsKubeClient := createKubeClient(getKubeConfig())
 
-	predicateChecker, err := predicatechecker.NewSchedulerBasedPredicateChecker(kubeClient, make(chan struct{}))
+	predicateChecker, err := predicatechecker.NewSchedulerBasedPredicateChecker(kubeClient,
+		autoscalingOptions.SchedulerConfig, make(chan struct{}))
 	if err != nil {
 		return nil, err
 	}
--- a/cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
+++ b/cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
@@ -27,6 +27,7 @@ import (
 	kube_client "k8s.io/client-go/kubernetes"
 	v1listers "k8s.io/client-go/listers/core/v1"
 	klog "k8s.io/klog/v2"
+	"k8s.io/kubernetes/pkg/scheduler/apis/config"
 	scheduler_config "k8s.io/kubernetes/pkg/scheduler/apis/config/latest"
 	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
 	scheduler_plugins "k8s.io/kubernetes/pkg/scheduler/framework/plugins"
@@ -44,21 +45,26 @@ type SchedulerBasedPredicateChecker struct {
 }
 
 // NewSchedulerBasedPredicateChecker builds scheduler based PredicateChecker.
-func NewSchedulerBasedPredicateChecker(kubeClient kube_client.Interface, stop <-chan struct{}) (*SchedulerBasedPredicateChecker, error) {
+func NewSchedulerBasedPredicateChecker(kubeClient kube_client.Interface, schedConfig *config.KubeSchedulerConfiguration, stop <-chan struct{}) (*SchedulerBasedPredicateChecker, error) {
 	informerFactory := informers.NewSharedInformerFactory(kubeClient, 0)
-	config, err := scheduler_config.Default()
-	if err != nil {
-		return nil, fmt.Errorf("couldn't create scheduler config: %v", err)
+
+	if schedConfig == nil {
+		var err error
+		schedConfig, err = scheduler_config.Default()
+		if err != nil {
+			return nil, fmt.Errorf("couldn't create scheduler config: %v", err)
+		}
 	}
-	if len(config.Profiles) != 1 || config.Profiles[0].SchedulerName != apiv1.DefaultSchedulerName {
-		return nil, fmt.Errorf("unexpected scheduler config: expected default scheduler profile only (found %d profiles)", len(config.Profiles))
+
+	if len(schedConfig.Profiles) != 1 {
+		return nil, fmt.Errorf("unexpected scheduler config: expected one scheduler profile only (found %d profiles)", len(schedConfig.Profiles))
 	}
 	sharedLister := NewDelegatingSchedulerSharedLister()
 
 	framework, err := schedulerframeworkruntime.NewFramework(
 		context.TODO(),
 		scheduler_plugins.NewInTreeRegistry(),
-		&config.Profiles[0],
+		&schedConfig.Profiles[0],
 		schedulerframeworkruntime.WithInformerFactory(informerFactory),
 		schedulerframeworkruntime.WithSnapshotSharedLister(sharedLister),
 	)
@@ -181,6 +187,7 @@ func (p *SchedulerBasedPredicateChecker) CheckPredicates(clusterSnapshot cluster
 			filterReasons,
 			p.buildDebugInfo(filterName, nodeInfo))
 	}
+
 	return nil
 }
 
--- a/cluster-autoscaler/simulator/predicatechecker/schedulerbased_test.go
+++ b/cluster-autoscaler/simulator/predicatechecker/schedulerbased_test.go
@@ -17,10 +17,14 @@ limitations under the License.
 package predicatechecker
 
 import (
+	"os"
+	"path/filepath"
 	"testing"
 	"time"
 
+	testconfig "k8s.io/autoscaler/cluster-autoscaler/config/test"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator/clustersnapshot"
+	scheduler "k8s.io/autoscaler/cluster-autoscaler/utils/scheduler"
 	. "k8s.io/autoscaler/cluster-autoscaler/utils/test"
 
 	"github.com/stretchr/testify/assert"
@@ -36,52 +40,114 @@ func TestCheckPredicate(t *testing.T) {
 
 	n1000 := BuildTestNode("n1000", 1000, 2000000)
 	SetNodeReadyState(n1000, true, time.Time{})
+	n1000Unschedulable := BuildTestNode("n1000", 1000, 2000000)
+	SetNodeReadyState(n1000Unschedulable, true, time.Time{})
+
+	defaultPredicateChecker, err := NewTestPredicateChecker()
+	assert.NoError(t, err)
+
+	// temp dir
+	tmpDir, err := os.MkdirTemp("", "scheduler-configs")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmpDir)
+
+	customConfigFile := filepath.Join(tmpDir, "custom_config.yaml")
+	if err := os.WriteFile(customConfigFile,
+		[]byte(testconfig.SchedulerConfigNodeResourcesFitDisabled),
+		os.FileMode(0600)); err != nil {
+		t.Fatal(err)
+	}
+
+	customConfig, err := scheduler.ConfigFromPath(customConfigFile)
+	assert.NoError(t, err)
+	customPredicateChecker, err := NewTestPredicateCheckerWithCustomConfig(customConfig)
+	assert.NoError(t, err)
 
 	tests := []struct {
-		name          string
-		node          *apiv1.Node
-		scheduledPods []*apiv1.Pod
-		testPod       *apiv1.Pod
-		expectError   bool
+		name             string
+		node             *apiv1.Node
+		scheduledPods    []*apiv1.Pod
+		testPod          *apiv1.Pod
+		predicateChecker PredicateChecker
+		expectError      bool
 	}{
+		// default predicate checker test cases
 		{
-			name:          "other pod - insuficient cpu",
-			node:          n1000,
-			scheduledPods: []*apiv1.Pod{p450},
-			testPod:       p600,
-			expectError:   true,
+			name:             "default - other pod - insuficient cpu",
+			node:             n1000,
+			scheduledPods:    []*apiv1.Pod{p450},
+			testPod:          p600,
+			expectError:      true,
+			predicateChecker: defaultPredicateChecker,
 		},
 		{
-			name:          "other pod - ok",
-			node:          n1000,
-			scheduledPods: []*apiv1.Pod{p450},
-			testPod:       p500,
-			expectError:   false,
+			name:             "default - other pod - ok",
+			node:             n1000,
+			scheduledPods:    []*apiv1.Pod{p450},
+			testPod:          p500,
+			expectError:      false,
+			predicateChecker: defaultPredicateChecker,
 		},
 		{
-			name:          "empty - insuficient cpu",
-			node:          n1000,
-			scheduledPods: []*apiv1.Pod{},
-			testPod:       p8000,
-			expectError:   true,
+			name:             "default - empty - insuficient cpu",
+			node:             n1000,
+			scheduledPods:    []*apiv1.Pod{},
+			testPod:          p8000,
+			expectError:      true,
+			predicateChecker: defaultPredicateChecker,
 		},
 		{
-			name:          "empty - ok",
-			node:          n1000,
-			scheduledPods: []*apiv1.Pod{},
-			testPod:       p600,
-			expectError:   false,
+			name:             "default - empty - ok",
+			node:             n1000,
+			scheduledPods:    []*apiv1.Pod{},
+			testPod:          p600,
+			expectError:      false,
+			predicateChecker: defaultPredicateChecker,
+		},
+		// custom predicate checker test cases
+		{
+			name:             "custom - other pod - ok",
+			node:             n1000,
+			scheduledPods:    []*apiv1.Pod{p450},
+			testPod:          p600,
+			expectError:      false,
+			predicateChecker: customPredicateChecker,
+		},
+		{
+			name:             "custom -other pod - ok",
+			node:             n1000,
+			scheduledPods:    []*apiv1.Pod{p450},
+			testPod:          p500,
+			expectError:      false,
+			predicateChecker: customPredicateChecker,
+		},
+		{
+			name:             "custom -empty - ok",
+			node:             n1000,
+			scheduledPods:    []*apiv1.Pod{},
+			testPod:          p8000,
+			expectError:      false,
+			predicateChecker: customPredicateChecker,
+		},
+		{
+			name:             "custom -empty - ok",
+			node:             n1000,
+			scheduledPods:    []*apiv1.Pod{},
+			testPod:          p600,
+			expectError:      false,
+			predicateChecker: customPredicateChecker,
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			var err error
-			predicateChecker, err := NewTestPredicateChecker()
 			clusterSnapshot := clustersnapshot.NewBasicClusterSnapshot()
 			err = clusterSnapshot.AddNodeWithPods(tt.node, tt.scheduledPods)
 			assert.NoError(t, err)
 
-			predicateError := predicateChecker.CheckPredicates(clusterSnapshot, tt.testPod, tt.node.Name)
+			predicateError := tt.predicateChecker.CheckPredicates(clusterSnapshot, tt.testPod, tt.node.Name)
 			if tt.expectError {
 				assert.NotNil(t, predicateError)
 				assert.Equal(t, NotSchedulablePredicateError, predicateError.ErrorType())
@@ -102,7 +168,80 @@ func TestFitsAnyNode(t *testing.T) {
 	n1000 := BuildTestNode("n1000", 1000, 2000000)
 	n2000 := BuildTestNode("n2000", 2000, 2000000)
 
-	var err error
+	defaultPredicateChecker, err := NewTestPredicateChecker()
+	assert.NoError(t, err)
+
+	// temp dir
+	tmpDir, err := os.MkdirTemp("", "scheduler-configs")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmpDir)
+
+	customConfigFile := filepath.Join(tmpDir, "custom_config.yaml")
+	if err := os.WriteFile(customConfigFile,
+		[]byte(testconfig.SchedulerConfigNodeResourcesFitDisabled),
+		os.FileMode(0600)); err != nil {
+		t.Fatal(err)
+	}
+
+	customConfig, err := scheduler.ConfigFromPath(customConfigFile)
+	assert.NoError(t, err)
+	customPredicateChecker, err := NewTestPredicateCheckerWithCustomConfig(customConfig)
+	assert.NoError(t, err)
+
+	testCases := []struct {
+		name             string
+		predicateChecker PredicateChecker
+		pod              *apiv1.Pod
+		expectedNodes    []string
+		expectError      bool
+	}{
+		// default predicate checker test cases
+		{
+			name:             "default - small pod - no error",
+			predicateChecker: defaultPredicateChecker,
+			pod:              p900,
+			expectedNodes:    []string{"n1000", "n2000"},
+			expectError:      false,
+		},
+		{
+			name:             "default - medium pod - no error",
+			predicateChecker: defaultPredicateChecker,
+			pod:              p1900,
+			expectedNodes:    []string{"n2000"},
+			expectError:      false,
+		},
+		{
+			name:             "default - large pod - insufficient cpu",
+			predicateChecker: defaultPredicateChecker,
+			pod:              p2100,
+			expectError:      true,
+		},
+
+		// custom predicate checker test cases
+		{
+			name:             "custom - small pod - no error",
+			predicateChecker: customPredicateChecker,
+			pod:              p900,
+			expectedNodes:    []string{"n1000", "n2000"},
+			expectError:      false,
+		},
+		{
+			name:             "custom - medium pod - no error",
+			predicateChecker: customPredicateChecker,
+			pod:              p1900,
+			expectedNodes:    []string{"n1000", "n2000"},
+			expectError:      false,
+		},
+		{
+			name:             "custom - large pod - insufficient cpu",
+			predicateChecker: customPredicateChecker,
+			pod:              p2100,
+			expectedNodes:    []string{"n1000", "n2000"},
+			expectError:      false,
+		},
+	}
 
 	clusterSnapshot := clustersnapshot.NewBasicClusterSnapshot()
 	err = clusterSnapshot.AddNode(n1000)
@@ -110,19 +249,18 @@ func TestFitsAnyNode(t *testing.T) {
 	err = clusterSnapshot.AddNode(n2000)
 	assert.NoError(t, err)
 
-	predicateChecker, err := NewTestPredicateChecker()
-	assert.NoError(t, err)
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			nodeName, err := tc.predicateChecker.FitsAnyNode(clusterSnapshot, tc.pod)
+			if tc.expectError {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+				assert.Contains(t, tc.expectedNodes, nodeName)
+			}
+		})
+	}
 
-	nodeName, err := predicateChecker.FitsAnyNode(clusterSnapshot, p900)
-	assert.NoError(t, err)
-	assert.True(t, nodeName == "n1000" || nodeName == "n2000")
-
-	nodeName, err = predicateChecker.FitsAnyNode(clusterSnapshot, p1900)
-	assert.NoError(t, err)
-	assert.Equal(t, "n2000", nodeName)
-
-	nodeName, err = predicateChecker.FitsAnyNode(clusterSnapshot, p2100)
-	assert.Error(t, err)
 }
 
 func TestDebugInfo(t *testing.T) {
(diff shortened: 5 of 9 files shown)


cluster-autoscaler/main.go
```diff
@@ -368,6 +376,36 @@ func getKubeConfig() *rest.Config {
...
+	data, err := os.ReadFile(path)
+
+	defaultFallback := "(falling back to default scheduler config)"
+
+	if err != nil {
+		klog.Errorf("couldn't load scheduler config %s: %v", defaultFallback, err)
+		return &schedconfig.KubeSchedulerConfiguration{}
+	}
+
+	obj, gvk, err := schedscheme.Codecs.UniversalDecoder().Decode(data, nil, nil)
```
vadasambar, line 389: This line creates a default scheduler config if none is specified. 
TODO: return output of this call as default. If it errors, fall back to the old behavior of getting a default config. 

Not doing the TODO above leads to a problem where the scheduler config ends up as empty and it shows up in the error log like this:
```
suraj@suraj:~/sandbox/autoscaler/cluster-autoscaler$ ./cluster-autoscaler --kubeconfig=/home/suraj/.kube/config --namespace=default
I0428 19:48:18.545718 1865631 leaderelection.go:245] attempting to acquire leader lease default/cluster-autoscaler...
I0428 19:48:40.443736 1865631 leaderelection.go:255] successfully acquired lease default/cluster-autoscaler
E0428 19:48:40.444001 1865631 main.go:385] couldn't load scheduler config (falling back to default scheduler config): open : no such file or directory
F0428 19:48:40.446465 1865631 main.go:504] Failed to create autoscaler: unexpected scheduler config: expected one scheduler profile only (found 0 profiles)
```

vadasambar:
Note that Go tests are probably failing because of https://github.com/kubernetes/autoscaler/pull/5708#discussion_r1180477801
### Testing for scheduler extenders
1. Created a GKE cluster
2. Deployed a custom scheduler called `custom-scheduler` in `kube-system` namespace with custom scheduler config.
<details>
<summary>Custom scheduler manifests</summary>

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: custom-scheduler
  namespace: kube-system
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: custom-scheduler-as-kube-scheduler
subjects:
- kind: ServiceAccount
  name: custom-scheduler
  namespace: kube-system
roleRef:
  kind: ClusterRole
  name: cluster-admin # only for testing
  apiGroup: rbac.authorization.k8s.io
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: custom-scheduler-config
  namespace: kube-system
data:
  custom-scheduler-config.yaml: |
    apiVersion: kubescheduler.config.k8s.io/v1beta2
    kind: KubeSchedulerConfiguration
    profiles:
      - schedulerName: custom-scheduler
    leaderElection:
      leaderElect: false
    # I want scheduler to call an external http extender to 
    # filter schedulable nodes
    extenders:
    - urlPrefix: http://sample-scheduler-extender-service.default
      filterVerb: apiv1/filter
      nodeCacheCapable: false
      ignorable: false
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    component: scheduler
    tier: control-plane
  name: custom-scheduler
  namespace: kube-system
spec:
  selector:
    matchLabels:
      component: scheduler
      tier: control-plane
  replicas: 1
  template:
    metadata:
      labels:
        component: scheduler
        tier: control-plane
        version: second
    spec:
      serviceAccountName: custom-scheduler
      volumes:
        - name: config-volume
          configMap:
            name: custom-scheduler-config
      containers:
      - command:
        - /usr/local/bin/kube-scheduler
        - --config=/etc/kubernetes/custom-scheduler/custom-scheduler-config.yaml
        - --v=5
        image: k8s.gcr.io/kube-scheduler:v1.25.3
        volumeMounts:
          - name: config-volume
            mountPath: /etc/kubernetes/custom-scheduler
        livenessProbe:
          httpGet:
            path: /healthz
            port: 10259
            scheme: HTTPS
          initialDelaySeconds: 15
        name: kube-second-scheduler
        readinessProbe:
          httpGet:
            path: /healthz
            port: 10259
            scheme: HTTPS
        resources:
          requests:
            cpu: '0.1'
        securityContext:
          privileged: false

      hostNetwork: false
      hostPID: false


```

Based on https://kubernetes.io/docs/tasks/extend-kubernetes/configure-multiple-schedulers/#define-a-kubernetes-deployment-for-the-scheduler
</details> 

3. Deployed a simple scheduler extender called [`sample-scheduler-extender`](https://github.com/vadasambar/sample-scheduler-extender)
4. Deployed a custom CA (built a custom image for this PR)
image


##### When nodes don't have `extender='true'` label
Pod (which has `spec.schedulerName: custom-scheduler`) gets stuck in `Pending` and CA is not able to scale-up as expected.
<details>
<summary>Pod manifest</summary>

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: annotation-second-scheduler
  labels:
    name: multischeduler-example
spec:
  schedulerName: custom-scheduler
  containers:
  - name: pod-with-second-annotation-container
    image: registry.k8s.io/pause:2.0
```

Based on https://kubernetes.io/docs/tasks/extend-kubernetes/configure-multiple-schedulers/#specify-schedulers-for-pods
</details>

image

As expected, CA throws error when it sees that the pod cannot be scheduled on the default nodepool (whose nodes don't have `extender: 'true'` label)

##### When nodes have `extender: 'true'` label
I created a new nodepool in GKE which brings a node with `extender: 'true'` label. 
image

CA correctly thinks that the pod can be scheduled on the upcoming node.
image
image

Node has the label
image
Pod gets scheduled
image

vadasambar:
Deleted my comment around test results because the screenshots have sensitive info. Will post another comment with sensitive details redacted.

vadasambar:
Note that Go tests are probably failing because of https://github.com/kubernetes/autoscaler/pull/5708#discussion_r1180477801
### Testing for scheduler extenders
1. Created a GKE cluster
2. Deployed a custom scheduler called `custom-scheduler` in `kube-system` namespace with custom scheduler config.
<details>
<summary>Custom scheduler manifests</summary>

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: custom-scheduler
  namespace: kube-system
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: custom-scheduler-as-kube-scheduler
subjects:
- kind: ServiceAccount
  name: custom-scheduler
  namespace: kube-system
roleRef:
  kind: ClusterRole
  name: cluster-admin # only for testing
  apiGroup: rbac.authorization.k8s.io
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: custom-scheduler-config
  namespace: kube-system
data:
  custom-scheduler-config.yaml: |
    apiVersion: kubescheduler.config.k8s.io/v1beta2
    kind: KubeSchedulerConfiguration
    profiles:
      - schedulerName: custom-scheduler
    leaderElection:
      leaderElect: false
    # I want scheduler to call an external http extender to 
    # filter schedulable nodes
    extenders:
    - urlPrefix: http://sample-scheduler-extender-service.default
      filterVerb: apiv1/filter
      nodeCacheCapable: false
      ignorable: false
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    component: scheduler
    tier: control-plane
  name: custom-scheduler
  namespace: kube-system
spec:
  selector:
    matchLabels:
      component: scheduler
      tier: control-plane
  replicas: 1
  template:
    metadata:
      labels:
        component: scheduler
        tier: control-plane
        version: second
    spec:
      serviceAccountName: custom-scheduler
      volumes:
        - name: config-volume
          configMap:
            name: custom-scheduler-config
      containers:
      - command:
        - /usr/local/bin/kube-scheduler
        - --config=/etc/kubernetes/custom-scheduler/custom-scheduler-config.yaml
        - --v=5
        image: k8s.gcr.io/kube-scheduler:v1.25.3
        volumeMounts:
          - name: config-volume
            mountPath: /etc/kubernetes/custom-scheduler
        livenessProbe:
          httpGet:
            path: /healthz
            port: 10259
            scheme: HTTPS
          initialDelaySeconds: 15
        name: kube-second-scheduler
        readinessProbe:
          httpGet:
            path: /healthz
            port: 10259
            scheme: HTTPS
        resources:
          requests:
            cpu: '0.1'
        securityContext:
          privileged: false

      hostNetwork: false
      hostPID: false


```

Based on https://kubernetes.io/docs/tasks/extend-kubernetes/configure-multiple-schedulers/#define-a-kubernetes-deployment-for-the-scheduler
</details> 

3. Deployed a simple scheduler extender called [`sample-scheduler-extender`](https://github.com/vadasambar/sample-scheduler-extender)
4. Deployed a custom CA (built a custom image for this PR)
image


##### When nodes don't have `extender='true'` label
Pod (which has `spec.schedulerName: custom-scheduler`) gets stuck in `Pending` and CA is not able to scale-up as expected.
<details>
<summary>Pod manifest</summary>

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: annotation-second-scheduler
  labels:
    name: multischeduler-example
spec:
  schedulerName: custom-scheduler
  containers:
  - name: pod-with-second-annotation-container
    image: registry.k8s.io/pause:2.0
```

Based on https://kubernetes.io/docs/tasks/extend-kubernetes/configure-multiple-schedulers/#specify-schedulers-for-pods
</details>

redacted1



As expected, CA throws error when it sees that the pod cannot be scheduled on the default nodepool (whose nodes don't have `extender: 'true'` label)

##### When nodes have `extender: 'true'` label
I created a new nodepool in GKE which brings a node with `extender: 'true'` label. 
image

CA correctly thinks that the pod can be scheduled on the upcoming node.
redacted2

image

Node has the label
redacted3

Pod gets scheduled
image

cluster-autoscaler/main.go
```diff
@@ -368,6 +376,36 @@ func getKubeConfig() *rest.Config {
 	return kubeConfig
 }
 
+func getSchedulerConfig(path string) *schedconfig.KubeSchedulerConfiguration {
```
x13n, line 379: This function represents non-trivial logic which I think should live in a separate package, not main. (Same goes for a lot of the existing code, but I think least we can do is not making things worse.)

cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
```diff
@@ -181,6 +212,41 @@ func (p *SchedulerBasedPredicateChecker) CheckPredicates(clusterSnapshot cluster
...
 	}
+
+	return p.runExtendersForNode(pod, nodeInfo)
+}
+
+// runExtendersForNode runs `Filter` function of all extenders for a particular pod on a single node
+func (p *SchedulerBasedPredicateChecker) runExtendersForNode(pod *apiv1.Pod, nodeInfo *schedulerframework.NodeInfo) *PredicateError {
+	for _, extender := range p.framework.Extenders() {
+		if extender.IsInterested(pod) {
+			filteredNodes, failedNodesMap, unresolvableNodesMap, err := extender.Filter(pod, []*apiv1.Node{nodeInfo.Node()})
```
x13n, line 223: nit: Looks like `extender.Filter` accepts a list of nodes - wouldn't it make more sense to run this once after `PreFilter`, but before `Filter` and pass all the nodes that came out of `PreFilter`?

cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
```diff
@@ -127,10 +152,16 @@ func (p *SchedulerBasedPredicateChecker) FitsAnyNodeMatching(clusterSnapshot clu
...
 
 		filterStatus := p.framework.RunFilterPlugins(context.TODO(), state, pod, nodeInfo)
-		if filterStatus.IsSuccess() {
-			p.lastIndex = (p.lastIndex + i + 1) % len(nodeInfosList)
-			return nodeInfo.Node().Name, nil
+		if !filterStatus.IsSuccess() {
+			continue
 		}
+
+		if err := p.runExtendersForNode(pod, nodeInfo); err != nil {
```
x13n, line 159: Can we not ignore `err` here? I think it has some useful information when not `nil`.

x13n (changes requested):
/assign

cluster-autoscaler/main.go
vadasambar: I like the idea. Thanks for the feedback. 

P.S.:
I created this PR as a PoC to evaluate the idea and get an OK from the community before proceeding ahead (one of the items on my agenda for today i.e., 8th May 2023's SIG meeting since we didn't have 2 meetings before). I had planned on refactoring this function and the PR as a whole (+ tests) but stopped myself so that I don't end up spending too much time in case this feature doesn't sound like a good idea to the community.

cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
vadasambar: Passing all nodes `extender.Filter` at once so that we don't call it multiple times makes sense to me but same could be said of running filter plugins (even though cost of calling in-tree filter plugins per node won't be as high). I just took a safe bet here and went with the same approach as running filter plugins (might not be the best approach as you rightfully pointed out). I guess the question here would be, why didn't we do the same for running filter plugins?

`extender.Filter` is called in the upstream scheduler code after all `PreFilter` and `Filter` plugins are called. I don't see much benefit in changing the order. 
```go
File: pkg/scheduler/schedule_one.go
385: // Filters the nodes to find the ones that fit the pod based on the framework
386: // filter plugins and filter extenders.
387: func (sched *Scheduler) findNodesThatFitPod(ctx context.Context, fwk framework.Framework, state *framework.CycleState, pod *v1.Pod) ([]*v1.Node, framework.Diagnosis, error) {
... 
393: 	allNodes, err := sched.nodeInfoSnapshot.NodeInfos().List()
394: 	if err != nil {
395: 		return nil, diagnosis, err
396: 	}
397: 	// Run "prefilter" plugins.
398: 	preRes, s := fwk.RunPreFilterPlugins(ctx, state, pod)
399: 	...
413: 
414: 	...
438: 	feasibleNodes, err := sched.findNodesThatPassFilters(ctx, fwk, state, pod, diagnosis, nodes)
...
446: 
447: 	feasibleNodes, err = findNodesThatPassExtenders(sched.Extenders, pod, feasibleNodes, diagnosis.NodeToStatusMap)
448: 	if err != nil {
449: 		return nil, diagnosis, err
450: 	}

```
https://github.com/kubernetes/kubernetes/blob/8d18ae6fc245deb68534cab608056c5aae9160bd/pkg/scheduler/schedule_one.go#L385-L452

```go
File: pkg/scheduler/schedule_one.go
576: func findNodesThatPassExtenders(extenders []framework.Extender, pod *v1.Pod, feasibleNodes []*v1.Node, statuses framework.NodeToStatusMap) ([]*v1.Node, error) {
577: 	// Extenders are called sequentially.
578: 	// Nodes in original feasibleNodes can be excluded in one extender, and pass on to the next
579: 	// extender in a decreasing manner.
...
587: 
588: 		// Status of failed nodes in failedAndUnresolvableMap will be added or overwritten in <statuses>,
589: 		// so that the scheduler framework can respect the UnschedulableAndUnresolvable status for
590: 		// particular nodes, and this may eventually improve preemption efficiency.
591: 		// Note: users are recommended to configure the extenders that may return UnschedulableAndUnresolvable
592: 		// status ahead of others.
593: 		feasibleList, failedMap, failedAndUnresolvableMap, err := extender.Filter(pod, feasibleNodes)
...

```
https://github.com/kubernetes/kubernetes/blob/8d18ae6fc245deb68534cab608056c5aae9160bd/pkg/scheduler/schedule_one.go#L576-L627

cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
vadasambar: Fair enough. I will think something around this. Thanks for the comment.

vadasambar:
> Supporting configuration of existing PreFilters/Filters seems non-controversial to me. I'm worried about extenders though. The 2 specific problems I see are:
> 
>  1. What exactly is being sent to HTTP extender? I suspect it's just node names, with extender taking the actual Node and Pod objects from informer / api-server? If so, that doesn't really work with CA simulation that operate on hypothetical cluster state and not real one (CA simulation has in-memory template nodes and even existing nodes can run an entirely different set of pods in CA simulation than what is really scheduled on them). Unless we serialize the entire ClusterSnapshot and send it over to extender, the extender is operating on wrong data and CA will make incorrect decisions as a result.
> 
>  2. Is there a separate HTTP request for each Filter run? If so, I don't think it will work on scales supported by CA (ex. simulating a scale-up of 100 nodes on 100 different nodegroups to make a scale-up decision).
> 
> 
> The problem with those 2 issues is that they are subtle and a simple enough test (an extender that only looks at the pending pod and is independent from any other pods in a small cluster) will work just fine. Even a more complex extender will not fail every time and the probability of problems likely depends on scale and complexity of the cluster (ie. my tests may pass just fine and it will only blow up once I hit prod). My personal opinion is that it's better to fail early and explicitly, than to have features that work in some circumstances, but fail unexpectedly. So unless we have a way of addressing my 2 concerns above, I'd be against including the extender support.

Thank you for the comment. 

1. This is configurable. 
    * If [`nodeCacheCapable`](https://kubernetes.io/docs/reference/config-api/kube-scheduler-config.v1beta2/#kubescheduler-config-k8s-io-v1beta2-Extender) field in `extenders[].nodeCacheCapable` in scheduler config is set to `true`, scheduler sends node names to HTTP extender
    * If the field is set to `false` scheduler sends nodes to the HTTP extender
    * References: 
         * Code ref in scheduler: https://github.com/kubernetes/kubernetes/blob/8d18ae6fc245deb68534cab608056c5aae9160bd/pkg/scheduler/extender.go#L267-L278
         * HTTP extender I used to test: https://github.com/vadasambar/sample-scheduler-extender
2. If you mean is there a separate HTTP request for every time we want to check a pod against `Filter` predicates in our CA code, then yes. 

I had a discussion around 2 with @mwielgus in yesterday's SIG meeting (8th May 2023) and he had similar concerns ([summary of the discussion](https://docs.google.com/document/d/1RvhQAEIrVLHbyNnuaT99-6u9ZUMp7BfkPupT2LAZK7w/edit#bookmark=id.heypmz2zh4hz)). There are concerns around performance of HTTP extender in scheduler as well ([ref](https://github.com/kubernetes/kubernetes/issues/72634)). 
I was hoping we could provide this as an optional feature for small to medium clusters since we don't support any kind of customization for scheduler simulation at the moment (and maybe ask for GRPC support for extenders?). 

That being said,

> The problem with those 2 issues is that they are subtle and a simple enough test (an extender that only looks at the pending pod and is independent from any other pods in a small cluster) will work just fine. Even a more complex extender will not fail every time and the probability of problems likely depends on scale and complexity of the cluster (ie. my tests may pass just fine and it will only blow up once I hit prod).
> My personal opinion is that it's better to fail early and explicitly, than to have features that work in some circumstances, but fail unexpectedly. So unless we have a way of addressing my 2 concerns above, I'd be against including the extender support.
> 

I understand your concern. Personally, I think we can add some checks to disable the extender when the cluster size becomes too big but there are problems with that approach too since it would surprise our users (e.g., `I expect the extender to be working but it has suddenly stopped working`) and the size would depend on the HTTP extender being used (some might degrade in performance at 20 nodes while others might degrade at 80 nodes). 

I will drop the idea of supporting external extenders and proceed ahead with supporting custom scheduler config for in-tree scheduler plugins. 

Thanks for taking the time to write all of this. Really helps understand the `why` of things.

cluster-autoscaler/utils/scheduler/scheduler.go
```diff
@@ -106,3 +112,38 @@ func ResourceToResourceList(r *schedulerframework.Resource) apiv1.ResourceList {
 	}
 	return result
 }
+
+// GetSchedulerConfig loads scheduler config from a path
+// (falls back to default config in case of error)
+func GetSchedulerConfig(path string) (*schedconfig.KubeSchedulerConfiguration, error) {
```
x13n, line 118: This function is actually the main code change in this PR now. Maybe worth adding some unit tests?

cluster-autoscaler/simulator/predicatechecker/testchecker.go
```diff
@@ -17,11 +17,18 @@ limitations under the License.
...
 
 import (
+	"k8s.io/autoscaler/cluster-autoscaler/utils/scheduler"
 	clientsetfake "k8s.io/client-go/kubernetes/fake"
 )
 
 // NewTestPredicateChecker builds test version of PredicateChecker.
 func NewTestPredicateChecker() (PredicateChecker, error) {
+
+	cfg, err := scheduler.GetSchedulerConfig("")
```
x13n, line 27: nit: Why not use the default config directly?

x13n (changes requested):

cluster-autoscaler/utils/scheduler/scheduler.go
```diff
@@ -106,3 +112,38 @@ func ResourceToResourceList(r *schedulerframework.Resource) apiv1.ResourceList {
...
+	defaultFallbackMsg := "(falling back to default scheduler config)"
```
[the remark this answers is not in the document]
vadasambar: Using default config instead of failing fast might surprise our users. I will change this. Thanks for the suggestion.

cluster-autoscaler/simulator/predicatechecker/testchecker.go
vadasambar: Let me think about this :pray:

cluster-autoscaler/utils/scheduler/scheduler.go
vadasambar: Sorry, I should have added a comment after https://github.com/kubernetes/autoscaler/pull/5708#issuecomment-1539546973. Since we have moved on from the PoC phase, I was planning to do another pass on PR (cleanup, add unit tests etc.,) and the change the PR from draft to Ready for Review. Maybe I can mention you again here once I am done?

x13n: Yup, sounds good!

cluster-autoscaler/utils/scheduler/scheduler.go
```diff
@@ -106,3 +117,31 @@ func ResourceToResourceList(r *schedulerframework.Resource) apiv1.ResourceList {
...
+	}
+
+	cfgObj, ok := obj.(*scheduler_config.KubeSchedulerConfiguration)
+	if !ok {
+		return nil, fmt.Errorf("%s, gvk: %s", SchedulerConfigTypeCastErr, gvk)
+	}
+
+	// this needs to be set explicitly because config's api version is empty after decoding
+	// check kubernetes/cmd/kube-scheduler/app/options/configfile.go for more info
+	cfgObj.TypeMeta.APIVersion = gvk.GroupVersion().String()
```
vadasambar, line 140: This function is based on upstream scheduler's private function called `loadConfigFromFile`

```go
File: cmd/kube-scheduler/app/options/configfile.go
34: func loadConfigFromFile(logger klog.Logger, file string) (*config.KubeSchedulerConfiguration, error) {
35: 	data, err := os.ReadFile(file)
36: 	if err != nil {
37: 		return nil, err
38: 	}
39: 
40: 	return loadConfig(logger, data)
41: }
42: 
43: func loadConfig(logger klog.Logger, data []byte) (*config.KubeSchedulerConfiguration, error) {
44: 	// The UniversalDecoder runs defaulting and returns the internal type by default.
45: 	obj, gvk, err := scheme.Codecs.UniversalDecoder().Decode(data, nil, nil)
46: 	if err != nil {
47: 		return nil, err
48: 	}
49: 	if cfgObj, ok := obj.(*config.KubeSchedulerConfiguration); ok {
50: 		// We don't set this field in pkg/scheduler/apis/config/{version}/conversion.go
51: 		// because the field will be cleared later by API machinery during
52: 		// conversion. See KubeSchedulerConfiguration internal type definition for
53: 		// more details.
54: 		cfgObj.TypeMeta.APIVersion = gvk.GroupVersion().String()
55: 		switch cfgObj.TypeMeta.APIVersion {
56: 		case configv1beta2.SchemeGroupVersion.String():
57: 			logger.Info("KubeSchedulerConfiguration v1beta2 is deprecated in v1.25, will be removed in v1.28")
58: 		case configv1beta3.SchemeGroupVersion.String():
59: 			logger.Info("KubeSchedulerConfiguration v1beta3 is deprecated in v1.26, will be removed in v1.29")
60: 		}
61: 		return cfgObj, nil
62: 	}
63: 	return nil, fmt.Errorf("couldn't decode as KubeSchedulerConfiguration, got %s: ", gvk)
64: }


```
https://github.com/kubernetes/kubernetes/blob/6b34fafdaf5998039c7e01fa33920a641b216d3e/cmd/kube-scheduler/app/options/configfile.go#L34-L64

Note that I haven't included the deprecated log messages because they might change and we'd have to remember updating them every time. I don't see a lot of value in printing these messages because we'd have to support the scheduler config even if it is deprecated anyway (open to other opinions here).

Note that if the user uses an old scheduler api version which isn't supported anymore like say `v1alpha1`, `UniversalDecoder().Decode` throws an error like this:
```
panic: no kind "KubeSchedulerConfiguration" is registered for version "kubescheduler.config.k8s.io/v1alpha1" in scheme "pkg/runtime/scheme.go:100"
```

cluster-autoscaler/utils/scheduler/scheduler_test.go
```diff
@@ -102,3 +105,87 @@ func TestResourceList(t *testing.T) {
...
+	tmpDir, err := os.MkdirTemp("", "scheduler-configs")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmpDir)
+
+	// Note that even if we are passing minimal config like below
+	// `GetSchedulerConfig` will set the rest of the default fields
+	// on its own (including default profile and default plugins)
+	correctConfigFile := filepath.Join(tmpDir, "correct_config.yaml")
```
vadasambar, line 120: Here's a simple program to test this. Just put the scheduler config `test-config.yaml`
```go
import (
	"fmt"
	"os"

	config "k8s.io/kubernetes/pkg/scheduler/apis/config"
	"k8s.io/kubernetes/pkg/scheduler/apis/config/scheme"
	"k8s.io/kubernetes/pkg/scheduler/apis/config/validation"
)

func main() {
	file := "test-config.yaml"
	data, err := os.ReadFile(file)
	if err != nil {
		panic(err)
	}

	obj, gvk, err := scheme.Codecs.UniversalDecoder().Decode(data, nil, nil)
	if err != nil {
		panic(err)
	}
	cfgObj, ok := obj.(*config.KubeSchedulerConfiguration)
	if !ok {
		panic(fmt.Errorf("couldn't decode as KubeSchedulerConfiguration, got %s: ", gvk))
	}

	cfgObj.TypeMeta.APIVersion = gvk.GroupVersion().String()

	if err := validation.ValidateKubeSchedulerConfiguration(cfgObj); err != nil {
		panic(err)
	}

	fmt.Println("cfgObj", *cfgObj)

}
```
test-config.yaml
```yaml
apiVersion: kubescheduler.config.k8s.io/v1
kind: KubeSchedulerConfiguration
```

When you run the program:
```
$ go run main.go
cfgObj {{ kubescheduler.config.k8s.io/v1} 16 {true {15s} {10s} {2s} leases kube-scheduler kube-system} {  application/vnd.kubernetes.protobuf 50 100}   {true true} 0xc00003b818 1 10 [{default-scheduler <nil> 0xc000324780 [{DefaultPreemption &TypeMeta{Kind:,APIVersion:,}} {InterPodAffinity &TypeMeta{Kind:,APIVersion:,}} {NodeAffinity &TypeMeta{Kind:,APIVersion:,}} {NodeResourcesBalancedAllocation &TypeMeta{Kind:,APIVersion:,}} {NodeResourcesFit &TypeMeta{Kind:,APIVersion:,}} {PodTopologySpread &TypeMeta{Kind:,APIVersion:,}} {VolumeBinding &TypeMeta{Kind:,APIVersion:,}}]}] []}
```

cluster-autoscaler/main.go
```diff
@@ -252,6 +255,16 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 	if *maxDrainParallelismFlag > 1 && !*parallelDrain {
 		klog.Fatalf("Invalid configuration, could not use --max-drain-parallelism > 1 if --parallel-drain is false")
 	}
+
+	var parsedSchedConfig *scheduler_config.KubeSchedulerConfiguration
+	// if scheduler config flag was set by the user
+	if pflag.CommandLine.Changed(config.SchedulerConfigFileFlag) {
```
vadasambar, line 261: This is so that if the `--scheduler-config-file` flag is not used, we don't try to look for the scheduler config in the default path set by the flag i.e., `""`. We want to use the default scheduler config. Setting the scheduler config to `nil` [ensures that the default scheduler config is used](https://github.com/kubernetes/autoscaler/pull/5708/files#diff-5818325464670a5cd54280fe253b0841a163da5de2764491a3065805785ffb1bR51-R56).

cluster-autoscaler/simulator/predicatechecker/schedulerbased_test.go
```diff
@@ -36,52 +39,122 @@ func TestCheckPredicate(t *testing.T) {
...
+			predicateChecker: defaultPredicateChecker,
 		},
+		// custom predicate checker test cases
 		{
-			name:          "other pod - ok",
-			node:          n1000,
-			scheduledPods: []*apiv1.Pod{p450},
-			testPod:       p500,
-			expectError:   false,
+			name:             "custom - other pod - ok",
```
vadasambar, line 118: Test cases for custom predicate checker are a copy of the default predicate checker except we don't expect an error for any of the test cases.

cluster-autoscaler/simulator/predicatechecker/schedulerbased_test.go
```diff
@@ -102,27 +175,107 @@ func TestFitsAnyNode(t *testing.T) {
...
+			pod:              p2100,
+			expectError:      true,
+		},
 
-	nodeName, err := predicateChecker.FitsAnyNode(clusterSnapshot, p900)
-	assert.NoError(t, err)
-	assert.True(t, nodeName == "n1000" || nodeName == "n2000")
+		// custom predicate checker test cases
+		{
+			name:             "custom - small pod - no error",
```
vadasambar, line 239: Test cases for custom predicate checker are a copy of the default predicate checker's test cases except we don't expect an error for any of the custom predicate checker test cases.

vadasambar:
@x13n I think the PR is ready for review. :pray:

x13n (changes requested):
Thanks for the changes! Left a couple of comments. Could you also update the PR title so it doesn't mention extenders?

vadasambar:
> Thanks for the changes! Left a couple of comments. Could you also update the PR title so it doesn't mention extenders?

I wanted to keep it as is so that it is easily searchable in the future in case anyone wants to know why we didn't support extenders but it can also be confusing. I will update the title to: `feat: support custom scheduler config without extenders`. Let me know what you think.

cluster-autoscaler/simulator/predicatechecker/testchecker.go
vadasambar: This has been updated to use the default scheduler config in case no scheduler config is passed through the function parameter: https://github.com/kubernetes/autoscaler/pull/5708/files#diff-5818325464670a5cd54280fe253b0841a163da5de2764491a3065805785ffb1bR51-R56

cluster-autoscaler/utils/scheduler/scheduler.go
vadasambar: Thanks for the comment. I thought about this but I was also on the fence because scheduler has a similar function to write the config to a file
```go
File: cmd/kube-scheduler/app/options/configfile.go
091: // LogOrWriteConfig logs the completed component config and writes it into the given file name as YAML, if either is enabled
092: func LogOrWriteConfig(logger klog.Logger, fileName string, cfg *config.KubeSchedulerConfiguration, completedProfiles []config.KubeSchedulerProfile) error {
093: 	loggerV := logger.V(2)
094: 	if !loggerV.Enabled() && len(fileName) == 0 {
095: 		return nil
096: 	}
097: 	cfg.Profiles = completedProfiles
098: 
099: 	buf, err := encodeConfig(cfg)
100: 	if err != nil {
101: 		return err
102: 	}
103: 
104: 	if loggerV.Enabled() {
105: 		loggerV.Info("Using component config", "config", buf.String())
106: 	}
107: 
108: 	if len(fileName) > 0 {
109: 		configFile, err := os.Create(fileName)
110: 		if err != nil {
111: 			return err
112: 		}
113: 		defer configFile.Close()
114: 		if _, err := io.Copy(configFile, buf); err != nil {
115: 			return err
116: 		}
117: 		logger.Info("Wrote configuration", "file", fileName)
118: 		os.Exit(0)
119: 	}
120: 	return nil
121: }

```
https://github.com/kubernetes/kubernetes/blob/6b34fafdaf5998039c7e01fa33920a641b216d3e/cmd/kube-scheduler/app/options/configfile.go#L91-L121

Notice how it is calling `os.Exit(0)` on line 118. If you import and use this function in your code, the program is going to exit. My concern is, if we don't implement the function to read config file on our own we might make ourselves dependent on the implementation of `LoadConfigFromFile` (if `loadConfigFromFile` is converted to a public function) whose implementation is dependent on use-case of scheduler and not CA. If there's a change in the `LoadConfigFromFile` in the future, it will break things for us. That being said, same thing can be argued for `Filter` and `PostFilter` extension points but I think there's an important difference between `Filter`/`PostFilter` extension points and `LoadConfigFromFile` or `LogOrWriteConfig`. Former is similar to a public API and it is unlikely to change without a strong deprecation policy or notice to users but that might not be the case with latter. 

There's other options here as well:
Option 1:
* Get this PR in to CA codebase
* Talk with sig-scheduling about setting expectation about `LoadConfigFromFile` 
* Create another PR to replace our implementation with sig-scheduling's implementation (might have to wait until they release 1.28 so that we can import their code or ask for a patch to 1.27)

Option 2:
* Put this PR on hold
* Talk with sig-scheduling about setting expectation about `LoadConfigFromFile` 
* Create another PR to replace our implementation with sig-scheduling's implementation (might have to wait until they release 1.28 so that we can import their code or ask for a patch to 1.27)

If we decide to go with asking for a public implementation of `LoadConfigFromFile` and we want to go with Option 1, I will create a separate issue to co-ordinate with sig-scheduling and use the public implementation. If we go with Option 2, we will have to put this PR on hold until we get `LoadConfigFromFile`. 

Let me know what you think.

cluster-autoscaler/main.go
vadasambar: I think there are two lines of thought here:
1. User has set `--scheduler-config-file=""` (note that this is a file path) which means they have made a mistake. Hence, fail. 
2. User has set `--scheduler-config-file=""` which means they don't want to set a scheduler config by themselves. Fall back to using a default one. 

Falling back to default in case of 2 doesn't seem as valuable to me because if the user wants to use the default config, they can just skip the flag altogether because the flag specifies path to a config file (and not scheduler config itself) and if they have set it to `""`, I think it's more likely that they have made a mistake (which can happen if say this value comes from a helm chart file and it's set to empty there by mistake) because they can just skip specifying the flag if they want to use the default config. 

A counter-argument could be, if the user sets only one or two fields in the custom scheduler config we anyway fill rest of the required fields by default (in this PR). It can be argued that 2 is inline with this behavior. 

I think failing here makes more sense than using default config but I am open to change it if you think there's a strong reason for not doing so.

cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -164,6 +166,9 @@ type AutoscalingOptions struct {
 	// ScaleDownSimulationTimeout defines the maximum time that can be
 	// spent on scale down simulation.
 	ScaleDownSimulationTimeout time.Duration
+	// SchedulerConfig allows changing configuration of in-tree
+	// scheduler plugins acting on PreFilter and Filter extension points
+	SchedulerConfig *scheduler_config.KubeSchedulerConfiguration
```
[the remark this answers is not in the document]
vadasambar: Let me think about this :pray:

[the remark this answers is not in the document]
vadasambar: I guess the reason we want to keep it simple is because these are `AutoscalingOptions` and they should represent options instead of result of options.

cluster-autoscaler/simulator/predicatechecker/schedulerbased_test.go
```diff
@@ -102,27 +175,107 @@ func TestFitsAnyNode(t *testing.T) {
...
+	// temp dir
+	tmpDir, err := os.MkdirTemp("", "scheduler-configs")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmpDir)
+
+	customConfigFile := filepath.Join(tmpDir, "custom_config.yaml")
+	if err := os.WriteFile(customConfigFile, []byte(`
+apiVersion: kubescheduler.config.k8s.io/v1
```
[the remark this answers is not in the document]
vadasambar: I will do that. Thanks for the suggestion.

cluster-autoscaler/core/test/common.go
```diff
@@ -199,7 +199,7 @@ func NewScaleTestAutoscalingContext(
 	// Ignoring error here is safe - if a test doesn't specify valid estimatorName,
 	// it either doesn't need one, or should fail when it turns out to be nil.
 	estimatorBuilder, _ := estimator.NewEstimatorBuilder(options.EstimatorName, estimator.NewThresholdBasedEstimationLimiter(0, 0), estimator.NewDecreasingPodOrderer())
-	predicateChecker, err := predicatechecker.NewTestPredicateChecker()
+	predicateChecker, err := predicatechecker.NewTestPredicateChecker(nil)
```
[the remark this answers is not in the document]
vadasambar: Makes sense. I like this more than passing `nil`. Thanks for the suggestion.

cluster-autoscaler/config/const.go
```diff
@@ -17,6 +17,10 @@ limitations under the License.
 package config
 
 const (
+	// SchedulerConfigFileFlag is the name of the flag
+	// for passing in custom scheduler config for in-tree scheduelr plugins
+	SchedulerConfigFileFlag = "scheduler-config-file"
```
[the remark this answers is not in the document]
vadasambar: So that it is easier to refer again [here](https://github.com/kubernetes/autoscaler/pull/5708/files/7ba7fdc1009138ac29d8d894015a8c0e2ff17ca1#diff-76f2867167fb87a2c61e82ea733b893d48724ca43e861b86a96f6dc582805ad4R261).

vadasambar:
Looks like there are duplicate review comments.

cluster-autoscaler/main.go
x13n: I was thinking about someone templating the manifest and using empty string to mean "default", but I guess such templating can also just skip the flag in such cases, so maybe it is fine to just leave as is.

x13n:
Super nit: I'd put "without extenders" inside parentheses. Otherwise looks good, thanks!

cluster-autoscaler/simulator/predicatechecker/schedulerbased_test.go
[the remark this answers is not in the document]
vadasambar: Done

cluster-autoscaler/core/test/common.go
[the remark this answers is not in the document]
vadasambar: Updated the code to `NewTestPredicateCheckerWithCustomConfig`.

cluster-autoscaler/utils/scheduler/scheduler.go
```diff
@@ -106,3 +117,31 @@ func ResourceToResourceList(r *schedulerframework.Resource) apiv1.ResourceList {
 	}
 	return result
 }
+
+// GetSchedulerConfig loads scheduler config from a path.
+func GetSchedulerConfig(path string) (*scheduler_config.KubeSchedulerConfiguration, error) {
```
x13n, line 122: nit: `ConfigFromPath`? (So it becomes `scheduler.ConfigFromPath(path)` outside of this package.)

cluster-autoscaler/config/const.go
```diff
@@ -32,4 +36,53 @@ const (
 	DefaultScaleDownUnreadyTimeKey = "scaledownunreadytime"
 	// DefaultMaxNodeProvisionTimeKey identifies MaxNodeProvisionTime autoscaling option
 	DefaultMaxNodeProvisionTimeKey = "maxnodeprovisiontime"
+
+	// Custom scheduler configs for testing
```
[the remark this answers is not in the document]
vadasambar: I guess putting test consts in `const.go` feels a little out of place. But I don't have a strong opinion on where this should go. I will take another look at this (P.S.: open to suggestions).

cluster-autoscaler/utils/scheduler/scheduler.go
vadasambar: Thanks for the suggestion. I have changed the name to `ConfigFromPath`. `scheduler.ConfigFromPath` feels less vague than `scheduler.GetSchedulerConfig`.

cluster-autoscaler/config/const.go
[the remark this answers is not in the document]
vadasambar: I have moved the test config here to a separate package under at `cluster-autoscaler/config/test/config.go`
This test config is used at two different places: 
1. simulator/predicatechecker/schedulerbased_test.go
2. utils/scheduler/scheduler_test.go

Other place this can go into is the existing `cluster-autoscaler/utils/test/test_utils.go`. Since this is config for testing, putting it under a separate `test` package under `cluster-autoscaler/config` seems like a better place. 

Let me know what you think.

cluster-autoscaler/utils/scheduler/scheduler.go
vadasambar: PR to update scheduler code: https://github.com/kubernetes/kubernetes/pull/119057

vadasambar: Added TODO [here](https://github.com/kubernetes/autoscaler/pull/5708/files#diff-03046c2a86c17ed94a0d62f14449afb658d5a297113cbf894f3a1150858d0203R122-R123)

vadasambar:
@x13n I have addressed all the review comments. Waiting for your approval.

vadasambar:
> @x13n I have addressed all the review comments. Waiting for your approval.

@x13n waiting for your review.

x13n:
Apologies for the delay. The change LGTM, but there's a ton of commits here. Can you squash them before merging?

vadasambar:
> Apologies for the delay. The change LGTM, but there's a ton of commits here. Can you squash them before merging?

@x13n I have squashed the commits.

vadasambar:
> > Apologies for the delay. The change LGTM, but there's a ton of commits here. Can you squash them before merging?
> 
> @x13n I have squashed the commits.

@x13n waiting for your approval.

PR #5711, opened 2023-04-25
Reduce allowed termination time to the max of the pods on the node

domenicbozzuto:
#### What type of PR is this?

/kind feature

PodEvictionResults now track the grace period deadline set in the eviction request. `ctx.MaxGracefulTerminationSec` still defines the upper bound the CA will wait for pods to be evicted, but calculates the largest grace period deadline of the pods on the node and picks the minimum of that and the upper bound.

Previously, if a pod failed to be evicted within its specified termination grace period, the autoscaler would still wait `MaxGracefulTerminationSec` for the eviction to finish, which is problematic when this value is in the range of hours or even days. It's also especially problematic if parallel drain is not enabled, as the eviction attempt will block any subsequent drains from occurring until `MaxGracefulTerminationSec` has passed.

Now, with the addition of the eviction grace period added to `EvictionResult`, the eviction will only wait as long as the pod with the greatest eviction grace period on the drained node (+ eviction headroom). 

##### Sample logs
```
"2023-04-25T16:19:31.077Z","Waiting up to 10m30s for evictions from <node>"
"2023-04-25T16:12:14.674Z","Waiting up to 2m0s for evictions from <node>"
"2023-04-25T16:10:35.581Z","Waiting up to 1m5s for evictions from <node>"
"2023-04-25T16:09:14.105Z","Waiting up to 1m0s for evictions from <node>"
"2023-04-25T15:54:26.709Z","Waiting up to 2m0s for evictions from <node>"
"2023-04-25T15:54:26.682Z","Waiting up to 1m30s for evictions from <node>"
"2023-04-25T15:52:51.027Z","Waiting up to 1m30s for evictions from <node>"
"2023-04-25T15:52:49.647Z","Waiting up to 1m0s for evictions from <node>"
...
```

```release-note
NONE
```

### DIFF
--- a/cluster-autoscaler/core/scaledown/actuation/actuator_test.go
+++ b/cluster-autoscaler/core/scaledown/actuation/actuator_test.go
@@ -84,6 +84,8 @@ func getStartDeletionTestCases(testNg *testprovider.TestNodeGroup, ignoreDaemonS
 	atomic2mixed := sizedNodeGroup("atomic-2-mixed", 2, true)
 	atomic2drain := sizedNodeGroup("atomic-2-drain", 2, true)
 
+	maxGracefulTermination := int64(60)
+
 	testCases := map[string]startDeletionTestCase{
 		"nothing to delete": {
 			emptyNodes: nil,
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/core/scaledown/actuation/actuator_test.go
```diff
@@ -545,17 +545,17 @@ func TestStartDeletion(t *testing.T) {
...
 				"drain-node-1": {ResultType: status.NodeDeleteOk},
 				"drain-node-2": {
 					ResultType: status.NodeDeleteErrorFailedToEvictPods,
 					Err:        cmpopts.AnyError,
 					PodEvictionResults: map[string]status.PodEvictionResult{
-						"drain-node-2-pod-0": {Pod: removablePod("drain-node-2-pod-0", "drain-node-2")},
+						"drain-node-2-pod-0": {Pod: removablePod("drain-node-2-pod-0", "drain-node-2"), GracePeriodSeconds: apiv1.DefaultTerminationGracePeriodSeconds},
 						"drain-node-2-pod-1": {Pod: removablePod("drain-node-2-pod-1", "drain-node-2"), Err: cmpopts.AnyError, TimedOut: true},
-						"drain-node-2-pod-2": {Pod: removablePod("drain-node-2-pod-2", "drain-node-2")},
+						"drain-node-2-pod-2": {Pod: removablePod("drain-node-2-pod-2", "drain-node-2"), GracePeriodSeconds: apiv1.DefaultTerminationGracePeriodSeconds},
```
BigDarkClown, line 558: Could you add some test cases where this value is something different than the default?

BigDarkClown (commented):
I think this makes a lot of sense - the `maxTermination` is already passed to the Kubernetes API, so we can safely assume that after this time we can safely timeout.

My only worry are the pods without the `TerminationGracePeriodSeconds` set. For them the termination period is just the default 30 seconds. I feel like this can break some deployments which did not specify the `TerminationGracePeriodSeconds` but somehow depended on the higher `MaxGracefulTerminationSec`.

/assign @BigDarkClown

cluster-autoscaler/core/scaledown/actuation/actuator_test.go
domenicbozzuto: Added a test case to `drain_test.go` (since the original changes were in `drain.go`), and added a test to `actuator_test.go` that exercises the deletion in the context of `StartDeletion`.  Let me know if these are ok / if additional test cases are needed!

domenicbozzuto:
Rebased to reconcile the conflicts in `actuation_test.go`, no changes to the other files

gjtempleton:
@BigDarkClown are you able to give this another review now it's been rebased?

PR #5713, opened 2023-04-26
Binpacking Estimator pod priority

jayantjain93:
Refactor the Binpacking estimator by making the pod sort extendable by moving the logic into pod_priority_processor.go Cleanup sorting logic/structs from binpacking_estimator.go

#### What type of PR is this?
/kind feature

We would like to have to ability to customise pod order which is used while binpacking.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/estimator/binpacking_estimator.go
+++ b/cluster-autoscaler/estimator/binpacking_estimator.go
@@ -18,10 +18,8 @@ package estimator
 
 import (
 	"fmt"
-	"sort"
 
 	apiv1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator/clustersnapshot"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator/predicatechecker"
@@ -30,32 +28,32 @@ import (
 	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
 )
 
-// podInfo contains Pod and score that corresponds to how important it is to handle the pod first.
-type podInfo struct {
-	score float64
-	pod   *apiv1.Pod
-}
-
 // BinpackingNodeEstimator estimates the number of needed nodes to handle the given amount of pods.
 type BinpackingNodeEstimator struct {
 	predicateChecker predicatechecker.PredicateChecker
 	clusterSnapshot  clustersnapshot.ClusterSnapshot
 	limiter          EstimationLimiter
+	podOrderer       EstimationPodOrderer
 }
 
 // NewBinpackingNodeEstimator builds a new BinpackingNodeEstimator.
 func NewBinpackingNodeEstimator(
 	predicateChecker predicatechecker.PredicateChecker,
 	clusterSnapshot clustersnapshot.ClusterSnapshot,
-	limiter EstimationLimiter) *BinpackingNodeEstimator {
+	limiter EstimationLimiter,
+	podOrderer EstimationPodOrderer) *BinpackingNodeEstimator {
 	return &BinpackingNodeEstimator{
 		predicateChecker: predicateChecker,
 		clusterSnapshot:  clusterSnapshot,
 		limiter:          limiter,
+		podOrderer:       podOrderer,
 	}
 }
 
-// Estimate implements First Fit Decreasing bin-packing approximation algorithm.
+// Estimate implements First-Fit bin-packing approximation algorithm
+// The ordering of the pods depend on the EstimatePodOrderer, the default
+// order is DecreasingPodOrderer
+// First-Fit Decreasing bin-packing approximation algorithm.
 // See https://en.wikipedia.org/wiki/Bin_packing_problem for more details.
 // While it is a multi-dimensional bin packing (cpu, mem, ports) in most cases the main dimension
 // will be cpu thus the estimated overprovisioning of 11/9 * optimal + 6/9 should be
@@ -70,8 +68,7 @@ func (e *BinpackingNodeEstimator) Estimate(
 	e.limiter.StartEstimation(pods, nodeGroup)
 	defer e.limiter.EndEstimation()
 
-	podInfos := calculatePodScore(pods, nodeTemplate)
-	sort.Slice(podInfos, func(i, j int) bool { return podInfos[i].score > podInfos[j].score })
+	pods = e.podOrderer.Order(pods, nodeTemplate, nodeGroup)
 
 	newNodeNames := make(map[string]bool)
 	newNodesWithPods := make(map[string]bool)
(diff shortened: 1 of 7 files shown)


cluster-autoscaler/estimator/pod_priority_processor.go
```diff
@@ -0,0 +1,92 @@
...
+
+// PodScoreInfo contains Pod and score that corresponds to how important it is to handle the pod first.
+type PodScoreInfo struct {
+	score float64
+	pod   *apiv1.Pod
+}
+
+// PodPriorityProcessor is an interface used to determine the order of the pods
+// used while binpacking during scale up estimation
+type PodPriorityProcessor interface {
```
towca, line 35: IMO the name is both very generic, and seems misleadingly related to the k8s pod priority concept - which IIUC this interface isn't actually concerned with.

I'd go with one of these instead:
* If it's supposed to be a generic processor for pods at the beginning of binpacking for each node group, I'd omit the "priority" part, and add something about when it's called - e.g. "PreEstimationPodProcessor". But I'm not sure what use cases we'd have for generic processing of these pods at the beginning of binpacking. On the other hand IMO it'd have some conceptual issues - if it's a generic processing hook, the implementation could think it's a good idea to e.g. add pods to the list, which IIUC would not play well with BinpackingEstimator (it assumes all passed pods can be scheduled on the nodes from the node group). Probably not a big issue, more of a nit if you want to go this way.
* If it's supposed to be just for changing the order of processing the pods during binpacking, I'd tie the name to that - e.g. "EstimationPodOrderer". IMO this is a better way to go, as the interface clearly implies that implementations should only be concerned with determining the order of passed pods.

cluster-autoscaler/estimator/pod_priority_processor.go
```diff
@@ -0,0 +1,92 @@
...
+	Process(pods []*apiv1.Pod, nodeTemplate *framework.NodeInfo) []*apiv1.Pod
+}
+
+// DefaultPodPriorityProcessor is the default implementation of the pod priority processor
+// It refactors the existing implementation of sorting pods by pod score
+type DefaultPodPriorityProcessor struct {
```
towca, line 41: "Default" doesn't tell anything about the behavior of the type, could we name it according to the behavior? E.g. something like "DecreasingPodOrderer" (to match the intent of it implementing First-fit-decreasing binpacking: https://en.wikipedia.org/wiki/First-fit-decreasing_bin_packing).

cluster-autoscaler/estimator/pod_priority_processor.go
```diff
@@ -0,0 +1,92 @@
...
+		podList = append(podList, podInfo.pod)
+	}
+
+	return podList
+}
+
+// CalculatePodScore score for  pod and returns PodScoreInfo structure.
+// Score is defined as cpu_sum/node_capacity + mem_sum/node_capacity.
+// Pods that have bigger requirements should be processed first, thus have higher scores.
+func (d *DefaultPodPriorityProcessor) CalculatePodScore(pod *apiv1.Pod, nodeTemplate *framework.NodeInfo) *PodScoreInfo {
```
towca, line 67: Any reason for:
1. Exporting this method? Do we plan on using it outside of this package?
2. Having this as a method and not just a function? It doesn't use `d` in any way.

cluster-autoscaler/estimator/binpacking_estimator.go
```diff
@@ -30,28 +28,25 @@ import (
...
 	return &BinpackingNodeEstimator{
-		predicateChecker: predicateChecker,
-		clusterSnapshot:  clusterSnapshot,
-		limiter:          limiter,
+		predicateChecker:     predicateChecker,
+		clusterSnapshot:      clusterSnapshot,
+		limiter:              limiter,
+		podPriorityProcessor: podPriorityProcessor,
 	}
 }
```
towca, line 52: The Estimate comment got outdated, IMO it should be updated with:
* Estimate only implements First-Fit binpacking, and the order of processing pods depends on the pod orderer.
* The default orderer is "Decreasing", and then the rest about First-Fit-Decreasing can be left after that.

cluster-autoscaler/estimator/binpacking_estimator.go
```diff
@@ -70,8 +65,7 @@ func (e *BinpackingNodeEstimator) Estimate(
 	e.limiter.StartEstimation(pods, nodeGroup)
 	defer e.limiter.EndEstimation()
 
-	podInfos := calculatePodScore(pods, nodeTemplate)
-	sort.Slice(podInfos, func(i, j int) bool { return podInfos[i].score > podInfos[j].score })
+	pods = e.podPriorityProcessor.Process(pods, nodeTemplate)
```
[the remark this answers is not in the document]
jayantjain93: Sure. I think it would open up for future use-cases, although it might be used now. Done.

cluster-autoscaler/estimator/binpacking_estimator_test.go
```diff
@@ -174,7 +174,8 @@ func TestBinpackingEstimate(t *testing.T) {
 			predicateChecker, err := predicatechecker.NewTestPredicateChecker()
 			assert.NoError(t, err)
 			limiter := NewThresholdBasedEstimationLimiter(tc.maxNodes, time.Duration(0))
-			estimator := NewBinpackingNodeEstimator(predicateChecker, clusterSnapshot, limiter)
+			processor := NewDefaultPodPriorityProcessor()
```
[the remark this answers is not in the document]
jayantjain93: yeah, that's a good addition. done.

cluster-autoscaler/estimator/estimator.go
```diff
@@ -45,13 +45,13 @@ type Estimator interface {
...
 	switch name {
 	case BinpackingEstimatorName:
 		return func(
 			predicateChecker predicatechecker.PredicateChecker,
 			clusterSnapshot clustersnapshot.ClusterSnapshot) Estimator {
-			return NewBinpackingNodeEstimator(predicateChecker, clusterSnapshot, limiter)
+			return NewBinpackingNodeEstimator(predicateChecker, clusterSnapshot, limiter, processor)
 		}, nil
 	}
 	return nil, fmt.Errorf("unknown estimator: %s", name)
```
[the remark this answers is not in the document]
jayantjain93: Yeah, that makes it more consistent. Done.

cluster-autoscaler/estimator/pod_priority_processor.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"sort"
+
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/kubernetes/pkg/scheduler/framework"
+)
+
+// PodScoreInfo contains Pod and score that corresponds to how important it is to handle the pod first.
+type PodScoreInfo struct {
```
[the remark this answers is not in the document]
jayantjain93: > Why export this type? Isn't it an implementation detail of this specific implementation?
I think the idea here it for was extensions of this implementation to be able to use it. I think I don't intend to use it for now. I'll make it internal. Thanks.

cluster-autoscaler/estimator/pod_priority_processor.go
jayantjain93: Thanks for the detailed comment.
The definite concept behind the process is to just reorder the pods before binpacking instead of being a "generic"  pod processor before binpacking. I also agree having a generic processor before binpacking could be mis-represented and lead to unintended consequences.
I've renamed to `EstimationPodOrderer`.

cluster-autoscaler/estimator/pod_priority_processor.go
jayantjain93: > Having this as a method and not just a function? It doesn't use d in any way.
Yes. This one I specifically added as a method and not just a function. The idea behind it is, the method of calculation is "core functionality" of the object and rather not a utility for the processor, which is why I thought this should be attached to the object even though it doesn't use `d`. I'm okay to detach it.

> Exporting this method? Do we plan on using it outside of this package?
I think this was similar reason as `PodScoreInfo` being exported for possible extension of this implementation. However, I don't plan on using it at this moment. Will make the internal.

cluster-autoscaler/estimator/binpacking_estimator.go
jayantjain93: Done.

cluster-autoscaler/estimator/pod_priority_processor.go
jayantjain93: Yeah. I think better naming would be more appropriate. Done.

cluster-autoscaler/estimator/estimator.go
```diff
@@ -72,3 +73,9 @@ type EstimationLimiter interface {
 	// just not expected to add any more nodes.
 	PermissionToAddNode() bool
 }
+
+// EstimationPodOrderer is an interface used to determine the order of the pods
+// used while binpacking during scale up estimation
+type EstimationPodOrderer interface {
+	Process(pods []*apiv1.Pod, nodeTemplate *framework.NodeInfo, nodeGroup cloudprovider.NodeGroup) []*apiv1.Pod
```
towca, line 80: nit: Maybe "Order" would be a more accurate name? Process is pretty generic.

cluster-autoscaler/estimator/pod_priority_processor.go
towca: I don't think that's necessary, you're explanation makes sense, thanks.

towca:
In general LGTM, just one minor comment and one nit if you want to address, then we can merge.

cluster-autoscaler/estimator/pod_priority_processor.go
```diff
@@ -0,0 +1,87 @@
+/*
```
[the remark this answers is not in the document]
jayantjain93: Done.

cluster-autoscaler/estimator/estimator.go
jayantjain93: Done.

PR #5753, opened 2023-05-11
Refactor taints.go to support taint values

jayantjain93:
#### What type of PR is this?

/kind feature

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/utils/taints/taints.go
+++ b/cluster-autoscaler/utils/taints/taints.go
@@ -116,16 +116,26 @@ func getKeyShortName(key string) string {
 
 // MarkToBeDeleted sets a taint that makes the node unschedulable.
 func MarkToBeDeleted(node *apiv1.Node, client kube_client.Interface, cordonNode bool) error {
-	return AddTaint(node, client, ToBeDeletedTaint, apiv1.TaintEffectNoSchedule, cordonNode)
+	taint := apiv1.Taint{
+		Key:    ToBeDeletedTaint,
+		Value:  fmt.Sprint(time.Now().Unix()),
+		Effect: apiv1.TaintEffectNoSchedule,
+	}
+	return AddTaint(node, client, taint, cordonNode)
 }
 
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/utils/taints/taints.go
```diff
@@ -116,16 +116,16 @@ func getKeyShortName(key string) string {
...
 
 // MarkDeletionCandidate sets a soft taint that makes the node preferably unschedulable.
 func MarkDeletionCandidate(node *apiv1.Node, client kube_client.Interface) error {
-	return AddTaint(node, client, DeletionCandidateTaint, apiv1.TaintEffectPreferNoSchedule, false)
+	return AddTaint(node, client, DeletionCandidateTaint, fmt.Sprint(time.Now().Unix()), apiv1.TaintEffectPreferNoSchedule, false)
 }
 
 // AddTaint sets the specified taint on the node.
-func AddTaint(node *apiv1.Node, client kube_client.Interface, taintKey string, effect apiv1.TaintEffect, cordonNode bool) error {
+func AddTaint(node *apiv1.Node, client kube_client.Interface, taintKey string, taintValue string, effect apiv1.TaintEffect, cordonNode bool) error {
```
BigDarkClown, line 128: nit: Maybe add a wrapper function `AddTaintWithTimestamp` (not sure on the name) that would pass `fmt.Sprint(time.Now().Unix())` as the `taintValue`? It seems useful, at least two taints are following that pattern.

jayantjain93: I'm unsure if we should be creating a wrapper for two caller where one of the values is same, creating wrappers may not be sustainable. I would instead look to push the dependency back to the caller.
Made changes by sending `apiv1.Taint` obj instead of individual values since all the values of `apiv1.Taint` are expected from the caller now.

I would consider creating wrapper where a set of values (i.e whole taint obj) might be same.

PR #5754, opened 2023-05-11
Adjust Eviction Control AEP for in-place updates

jbartosik:
#### What type of PR is this?

/kind documentation
/kind feature
/kind api-change

Adjust Autoscaling Enhancement Proposal 4831 to work better with in place update support. See #5755 for details of in-place support

I divided the PR into a few changes to make the review easier (separating content main content change, formatting changes, file rename and minor edits).

@voelzmo @kgolab @wangchen615 please take a look

#4831


jbartosik:
@pbetkier please take a look

vertical-pod-autoscaler/enhancements/4831-control-actuation-behavior/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+The following analysis focuses on the `InPlaceOrRecreate` mode. In other modes having separate controls for in-place and
+eviction actuation doesn't offer any more control than single setting, because VPA won't actively actuate (`Off` and
+`Initial`) or because it will use only one mode of actuation (`Recreate` and `InPlaceOnly`). It also assumes that scale
+downs always succeed.
+
+VPA wil attempt to scale in place before attempting to evict. Since scale downs succeed VPA will never consider evicting
+pods to scale them down so configuring scale down in place separately from scale down by eviction is unnecessary.
+
+If user wants to scale up in place only, not evict they should use `InPlaceOnly` mode (since we assume scale downs in
```
kgolab, line 88: The only case I could think of to act differently here is if my app is not that good at reducing its resource (mostly memory) usage w/o a restart. IIUC this could be addressed within PodSpec, not via VPA settings, right?

kgolab:
> So, when in-place resource updates make it to the VPA, I don't think I would want any restrictions being applied to disruption-free resource updates and _still_ want to have restrictions for eviction-based resource updates – be it because VPA is falling back to eviction or because my workload is configured with an update mode which doesn't allow for in-place updates.

@voelzmo , I'm not sure I fully understood here: are you saying you'd always want in-place updates to happen and only care about evictions?

If yes: unlimited in-place scale down is likely to work but might end up in an eviction later as it might not be reversible (e.g. more Pods getting scheduled on a Node or others growing). Would that be a good reason do limit in-place scale down as well?

vertical-pod-autoscaler/enhancements/4831-control-actuation-behavior/README.md
```diff
@@ -0,0 +1,105 @@
...
+With this approach you either end up re-building half of the VPA (updater/webhook), or use a different mechanism to
+apply the recommendations, such as modifying the requests in the Pod owning Object – which has its own drawbacks.
+
+### Independent controls for eviction and in-place actuation
+We could introduce independent fields controlling scaling by eviction and scaling in place. This seems redundant.
+
+The following analysis focuses on the `InPlaceOrRecreate` mode. In other modes having separate controls for in-place and
+eviction actuation doesn't offer any more control than single setting, because VPA won't actively actuate (`Off` and
+`Initial`) or because it will use only one mode of actuation (`Recreate` and `InPlaceOnly`). It also assumes that scale
+downs always succeed.
```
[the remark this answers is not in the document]
jbartosik: I noticed I was making that assumption, it seemed like it would make sense so I wanted to make it explicit. I didn't check it very carefully.

I think without it it still makes sense to have one set of controls for actuation.

Without this assumption having separate controls makes some sense. One might want to scale down only in place (don't evict to save resources) but allow eviction to scale up (evict to allow continued functioning) but I'm not sure how useful that is

vertical-pod-autoscaler/enhancements/4831-control-actuation-behavior/README.md
jbartosik: I had some problems with updating requests of containers that require restart (restart didn't happen, resize got stuck in InProgress for >1h (I stopped looking)).

I hope this will be fixed before beta and we can rely on containers being restarted and properly configured

vertical-pod-autoscaler/enhancements/4831-control-actuation-behavior/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+VPA wil attempt to scale in place before attempting to evict. Since scale downs succeed VPA will never consider evicting
```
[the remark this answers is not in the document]
jbartosik: Done

> > So, when in-place resource updates make it to the VPA, I don't think I would want any restrictions being applied to disruption-free resource updates and _still_ want to have restrictions for eviction-based resource updates – be it because VPA is falling back to eviction or because my workload is configured with an update mode which doesn't allow for in-place updates.
> 
> Marco Voelz , I'm not sure I fully understood here: are you saying you'd always want in-place updates to happen and only care about evictions?
> 
> If yes: unlimited in-place scale down is likely to work but might end up in an eviction later as it might not be reversible (e.g. more Pods getting scheduled on a Node or others growing). Would that be a good reason do limit in-place scale down as well?

@voelzmo @kgolab 

I don't think eviction will go away completely in the foreseeable future. I don't think we'll have guarantee that in-place update will succeed so we'll want eviction as a fallback.

This is more about the case Karol described - user might want to avoid scale downs (because there is no guarantee that they'll be able to scale up without disruption) but allow scale-ups (preferring in-place but with eviction as fallback).

jbartosik:
So, when in-place resource updates make it to the VPA, I don't think I would want any restrictions being applied to disruption-free resource updates and still want to have restrictions for eviction-based resource updates – be it because VPA is falling back to eviction or because my workload is configured with an update mode which doesn't allow for in-place updates.

Marco Voelz , I'm not sure I fully understood here: are you saying you'd always want in-place updates to happen and only care about evictions?

If yes: unlimited in-place scale down is likely to work but might end up in an eviction later as it might not be reversible (e.g. more Pods getting scheduled on a Node or others growing). Would that be a good reason do limit in-place scale down as well?

[Marco Voelz](https://github.com/voelzmo) [Karol Gołąb](https://github.com/kgolab)

I don't think eviction will go away completely in the foreseeable future. I don't think we'll have guarantee that in-place update will succeed so we'll want eviction as a fallback.

This is more about the case Karol described - user might want to avoid scale downs (because there is no guarantee that they'll be able to scale up without disruption) but allow scale-ups (preferring in-place but with eviction as fallback).

liggitt:
Based on https://github.com/kubernetes/autoscaler/pull/5754#event-9230368884, it looks like this wasn't asking for API review

/remove-label api-review

vertical-pod-autoscaler/enhancements/4831-control-actuation-behavior/README.md
```diff
@@ -0,0 +1,106 @@
...
+* For a specific VPA object, allow turning off actuation for downscaling entirely while still allow for upscaling
+* Allow for resource-specific decisions: The desired policy may be different for CPU and Memory
+
+## Proposal
+Add a new field `ActuationRequirements` to
+[`PodUpdatePolicy`](https://github.com/kubernetes/autoscaler/blob/2f4385b72e304216cf745893747da45ef314898f/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L109)
+of type `[]*ActuationRequirement`. A single `ActuationRequirement` defines a condition which must be `true` to allow
+actuation for the corresponding `Container`. When multiple `ActuationRequirements` are specified for a `Pod`, all of
+them must evaluate to `true` to allow eviction. VPA may partially actuate its recommendation in place, obeying all the
+`ActuationRequirements` while not applying parts of the recommendation that would violate any `ActuationRequirement`.
```
kwiesmueller, line 50: If actuation still happens (using in-place) I think naming the field `EvictionRequirements` or `RestartRequirements` would be more transparent.

If we later want to add a similar mechanism to prevent even in-place changes in certain scenarios that would still be possible.

kwiesmueller: I see you mention that below in part, I agree with that.
For easier understanding I'd still suggest using a different name though.

vertical-pod-autoscaler/enhancements/4831-control-actuation-behavior/README.md
```diff
@@ -0,0 +1,106 @@
...
+uptime. It is possible to turn off eviction for downscaling entirely or have it managed by an external component to 
+allow downscaling only during certain time windows, days, or based on other criteria.
+
+Scaling containers in place might cause immediate disruption (when containers specify `RestartPolicy: RestartContainer`
+for one of the changed resources) which operators might want to avoid. Even if in place scale down doesn't cause
+immediate disruption there is no guarantee that container will be able to scale up later without eviction. 
+
+### Goals
+* For a specific VPA object, allow turning off actuation for downscaling entirely while still allow for upscaling
+* Allow for resource-specific decisions: The desired policy may be different for CPU and Memory
```
dbenque, line 42: This might be open for other resource than CPU and Memory int he future, so let's pay attention to have an API open to handle other type of resources.

vertical-pod-autoscaler/enhancements/4831-control-actuation-behavior/README.md
```diff
@@ -0,0 +1,106 @@
...
+of type `[]*ActuationRequirement`. A single `ActuationRequirement` defines a condition which must be `true` to allow
+actuation for the corresponding `Container`. When multiple `ActuationRequirements` are specified for a `Pod`, all of
+them must evaluate to `true` to allow eviction. VPA may partially actuate its recommendation in place, obeying all the
+`ActuationRequirements` while not applying parts of the recommendation that would violate any `ActuationRequirement`.
+
+A single `ActuationRequirement` specifies `Resources` and a `ChangeRequirement` comparing the new recommendation
+(`Target`) with the existing requests on a Pod (`Requests`). Possible values for `Resources` are `[CPU]` and `[Memory]`
+or both `[CPU,Memory]`. If `Resources: [CPU, Memory]`, the condition must be true for either of the two resources to
+allow for actuation. Possible values for `ChangeRequirement` are `TargetHigherThanRequests`,
+`TargetHigherThanOrEqualToRequests`, `TargetLowerThanRequests` and `TargetLowerThanOrEqualToRequests`.
```
dbenque, line 56: Why do we need the variations with `OrEqual` ?

voelzmo:
> > So, when in-place resource updates make it to the VPA, I don't think I would want any restrictions being applied to disruption-free resource updates and _still_ want to have restrictions for eviction-based resource updates – be it because VPA is falling back to eviction or because my workload is configured with an update mode which doesn't allow for in-place updates.
> 
> @voelzmo , I'm not sure I fully understood here: are you saying you'd always want in-place updates to happen and only care about evictions?

Sorry for dropping the ball on this one, I completely got sidetracked by other things.
You understood correctly: I'm voting for not extending the original proposal to also apply to in-place updates, but rather only deal with evictions, as it was intended.
This means, it should still apply whenever the Updater would fall back to eviction, but in-place resource updates (regardless if the containers specified they need a restart or not) would no be subject to the `EvictionRequirements`.

> 
> If yes: unlimited in-place scale down is likely to work but might end up in an eviction later as it might not be reversible (e.g. more Pods getting scheduled on a Node or others growing). Would that be a good reason do limit in-place scale down as well?

TBH, I'm currently not super concerned about that scenario. If you're running in a setting where Nodes are 99% allocated, even scale-ups without previous scale-downs would need to result in evictions. If that's the case, and you need to avoid disruptions at all cost, it seems like VPA is probably not what you want to use at all?

vertical-pod-autoscaler/enhancements/4831-control-actuation-behavior/README.md
voelzmo: Given that VPA has other mechanisms in place to ensure there is a large enough diff between the new recommendation and the current requests, I guess it should be fine to remove the `orEqual` variations entirely. WDYT, @jbartosik?

thockin:
I ACK the nomination to do API review.  I am a little out of touch with the VPA APIs, so I will have to come up to speed.  Maybe we can have a slack or Zoom discussion after code freeze?

voelzmo:
@thockin I think we tried to remove the request for api-review [once](https://github.com/kubernetes/autoscaler/pull/5754#event-9230368884) and [twice](https://github.com/kubernetes/autoscaler/pull/5754#issuecomment-1563173539), but this didn't work as expected. I think this should be fine without a review from your side. Sorry for the confusion ;(

We _are_ changing parts of the VPA API by adding an option for the user to control if a Pod should be evicted based on if the VPA is scaling up or down, but I think we're ok to discuss these changes inside SIG autoscaling? Don't want to waste your time here :)

@jbartosik: please correct me if I misunderstood the intent and you wanted @thockin to review this from the API side.

/remove-kind api-change

jbartosik:
I think it makes sense to have controls for in-place scaling. In place resize can still be a disruption (in place updates might cause container restart). So one might want to allow scale ups (to avoid even bigger disruption when containers start crashing).

vertical-pod-autoscaler/enhancements/4831-control-actuation-behavior/README.md
jbartosik: Yes, let's do that

vertical-pod-autoscaler/enhancements/4831-control-actuation-behavior/README.md
jbartosik: I think the proposal is easy to extend to other resources. Do you expect any problems?

jbartosik:
I think that we will need to allow control over direction of in-place resizes. From what @kwiesmueller and @voelzmo write like you'd prefer separate controls for in-place resizes and eviction. I preferred single control for both because having fewer config options makes things easier to use (by making it harder to have bad configs, less things to understand).

So I'd

1. Close this PR
2. Add a note to [4016-in-place-updates-support AEP](https://github.com/kubernetes/autoscaler/blob/fb1a4c28a3ab95f14ae064039733ae4a933030e1/vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md) that we should have separate controls for in-place resizes, with short explanation why we prefer that over having one set of controls for in-place and evict actuation (in-place is in alpha so I don't think we need to do full EAP yet)
3. Adjust [4831-control-eviction-behavior](https://github.com/kubernetes/autoscaler/blob/fb1a4c28a3ab95f14ae064039733ae4a933030e1/vertical-pod-autoscaler/enhancements/4831-control-eviction-behavior/README.md) to drop `TargetHigherThanOrEqualToRequests` and `TargetLowerThanOrEqualToRequests`

@kwiesmueller @voelzmo please:
- confirm that I understood you correctly
- confirm that actions I wrote down seem like a good plan to you
- let me know if you want to do any of the changes to existing AEPs I mention (or if I should do them)

kwiesmueller:
Sounds good to me, I don't think I'll get to making any changes this week :-/
If it has time I can work on a PR to update the in-place AEP next week.

voelzmo:
I'm happy with closing this PR and having `EvictionRequirements` apply only for evictions. I'm not sure that we have at this point a clear indication that we will need the same configuration options for in-place resource edits as well, but I think it is a good idea to move this into a separate PR/discussion.

I can take over point 3 
> Adjust [4831-control-eviction-behavior](https://github.com/kubernetes/autoscaler/blob/fb1a4c28a3ab95f14ae064039733ae4a933030e1/vertical-pod-autoscaler/enhancements/4831-control-eviction-behavior/README.md) to drop TargetHigherThanOrEqualToRequests and TargetLowerThanOrEqualToRequests

And remove the `OrEqual` options from the proposal and the API/implementation PRs.

Thanks for moving this forward, Joachim!

jbartosik:
Thank you @kwiesmueller and @voelzmo . I opened #5990 and #5991 and assigned them to you to make it easier for us to not forget about this.

PR #5755, opened 2023-05-11
AEP for support of in-place updates for VPA

jbartosik:
/kind documentation
/kind feature

Add Autoscaling Enhancement Proposal to add basic support for in place updates to VPA.

See also #5754 for changes I propose to eviction control enhancement proposal to work better with in place updates

#4016

@voelzmo @kgolab @pbetkier @wangchen615 please take a look

### DIFF
--- a/vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
+++ b/vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
@@ -1,0 +1,109 @@
+# AEP-4016: Support for in place updates in VPA
+
+<!-- toc -->
+- [Summary](#summary)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Context](#context)
+- [Design Details](#design-details)
+    - [Applying Updates During Pod Admission](#applying-updates-during-pod-admission)
+    - [Applying Disruption-free Updates](#Applying-disruption-free-updates)
+    - [Applying Disruptive Updates](#applying-disruptive-updates)
+    - [Test Plan](#test-plan)
+- [Implementation History](#implementation-history)
+<!-- /toc -->
+
+## Summary
+
+
+VPA applies its recommendations with a mutating webhook, during pod creation. It can also evict
+pods expecting that it will apply the recommendation when the pod is recreated. This is a
+disruptive process so VPA has some mechanism to avoid too frequent disruptive updates. 
+
+This proposal allows VPA to apply its recommendations more frequently, with less disruption by
+using in-place update feature which is an alpha feature available in Kubernetes 1.27.
+
+This proposal enables only core uses of  in place updates in VPA with intention of gathering more feedback. Any more
+advanced uses of in place updates in VPA (like applying different recommendations during pod initialization) will be
+introduced as separate enhancement proposals.
+
+### Goals
+
+* Allow VPA to actuate without disruption,
+* Allow VPA to actuate more frequently, when it can actuate without disruption,
+* Allow VPA to actuate in situations where actuation by eviction doesn't work.
+
+### Non-Goals
+
+* Improved handling of injected sidecars
+  * Separate AEP will improve VPAs handling of injected sidecars.
+
+## Proposal
+
+Add new supported values of
+[`UpdateMode`](https://github.com/kubernetes/autoscaler/blob/71b489f5aec3899157b37472cdf36a1de223d011/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L124):
+
+* `InPlaceOnly` and
+* `InPlaceOrRecreate`.
+
+## Context
+
+[In-place update of pod resources KEP](https://github.com/kubernetes/enhancements/issues/1287) is available in alpha in
+[Kubernetes 1.27](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.27.md#api-change-3). The
+feature allows changing container resources while the container is running. It also adds 
+[`ResizePolicy`](https://github.com/kubernetes/api/blob/8360d82aecbc72aa039281a394ebed2eaf0c0ccc/core/v1/types.go#L2448)
+field to Container. This field indicates that the container should be restarted when its resources
+
+## Design Details
+
+VPA Updater will execute in-place updates or eviction-updates in the newly added `InPlaceOnly` and `InPlaceOrRecreate`
+modes. 
+
+### Applying Updates During Pod Admission
+
+For VPAs in `InPlaceOnly` and `InPlaceOrRecreate` modes VPA Admission Controller will apply updates to starting pods,
+like it does for VPAs in `Initial`, `Auto`, and `Recreate` modes.
+
+### Applying Disruption-free Updates
+
+When update changes only resources for which container indicates that it doesn't require a restart then VPA can attempt
+to actuate the change without disrupting the pod. VPA Updater will:
+* attempt to actuate such updates in place,
+* attempt to actuate them if difference between recommendation and request is at least 10%
+  * even if pod has been running for less than 24h,
+* if necessary execute only partial updates (if some changes would require a restart).
+
+### Applying Disruptive Updates
+
+In both `InPlaceOnly` and `InPlaceOrRecreate` modes VPA updater will attempt to apply updates that require container
+restart in place. It will update them under conditions that would trigger update with `Recreate` mode. That is it will
+apply disruptive updates if:
+
+* Sum of pods requests is below sum of recommendations `LowerBound` or
+* Sum of pods requests is above sum of recommendations `UppelBound` or
+* Difference between sum of pods requests and sum of recommendation `Target`s is more than 10% and the pod has been
+  running undisrupted for at least 24h.
+  * Successful disruptive update counts as disruption here (and prevents further disruptive updates to the pod for 24h).
+
+In `InPlaceOrRecreate` mode (but not in `InPlaceOnly` mode) VPA updater will evict pod to actuate a recommendation if it
+attempted to apply the recommendation in place and failed.
+
+### Test Plan
+
+The following test scenarios will be added to e2e tests. Both `InPlaceOnly` and `InPlaceOrRecreate` modes will be tested
+and they should behave the same:
+
+* Admission controller applies recommendation to pod controlled by VPA. 
+* Disruption-free in-place update applied to all containers of a pod (request in recommendation bounds).
+* Partial disruption-free update applied to some containers of a pod, some disruptive changes skipped (request in
+  recommendation bounds).
+* Disruptive in-place update applied to all containers of a pod (request out ouf recommendation bounds).
+
+There will be also scenario testing difference between `InPlaceOnly` and `InPlaceOrRecreate` modes. Disruptive in-place
+update will fail. In `InPlaceOnly` pod should not be evicted, in the `InPlaceOrRecreate` pod should be evicted and a the
+recommendation applied.
+
+## Implementation History
+
+- 2023-05-10: initial version
--- a/vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
+++ b/vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
@@ -10,6 +10,7 @@
     - [Applying Updates During Pod Admission](#applying-updates-during-pod-admission)
     - [Applying Disruption-free Updates](#Applying-disruption-free-updates)
     - [Applying Disruptive Updates](#applying-disruptive-updates)
+    - [Comparison of `UpdateMode`s](#comparison-of-updatemodes)
     - [Test Plan](#test-plan)
 - [Implementation History](#implementation-history)
 <!-- /toc -->
@@ -22,11 +23,14 @@ pods expecting that it will apply the recommendation when the pod is recreated. 
 disruptive process so VPA has some mechanism to avoid too frequent disruptive updates. 
 
 This proposal allows VPA to apply its recommendations more frequently, with less disruption by
-using in-place update feature which is an alpha feature available in Kubernetes 1.27.
+using the
+[in-place update feature] which is an alpha feature [available in Kubernetes 1.27.] This proposal enables only core uses
+of in place updates in VPA with intention of gathering more feedback. Any more advanced uses of in place updates in VPA
+(like applying different recommendations during pod initialization) will be introduced as separate enhancement
+proposals.
 
-This proposal enables only core uses of  in place updates in VPA with intention of gathering more feedback. Any more
-advanced uses of in place updates in VPA (like applying different recommendations during pod initialization) will be
-introduced as separate enhancement proposals.
+[in-place update feature](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources)
+[available in Kubernetes 1.27.](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.27.md#api-change-3)
 
 ### Goals
 
@@ -41,24 +45,32 @@ introduced as separate enhancement proposals.
 
 ## Proposal
 
-Add new supported values of
-[`UpdateMode`](https://github.com/kubernetes/autoscaler/blob/71b489f5aec3899157b37472cdf36a1de223d011/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L124):
+Add new supported values of [`UpdateMode`]:
 
 * `InPlaceOnly` and
 * `InPlaceOrRecreate`.
 
+[`UpdateMode`]: https://github.com/kubernetes/autoscaler/blob/71b489f5aec3899157b37472cdf36a1de223d011/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L124
+
 ## Context
 
-[In-place update of pod resources KEP](https://github.com/kubernetes/enhancements/issues/1287) is available in alpha in
-[Kubernetes 1.27](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.27.md#api-change-3). The
-feature allows changing container resources while the container is running. It also adds 
-[`ResizePolicy`](https://github.com/kubernetes/api/blob/8360d82aecbc72aa039281a394ebed2eaf0c0ccc/core/v1/types.go#L2448)
-field to Container. This field indicates that the container should be restarted when its resources
+[In-place update of pod resources KEP] is available in alpha in [Kubernetes 1.27]. The feature allows changing container
+resources while the container is running. It also adds [`ResizePolicy`] field to Container. This field indicates for an
+individual resource if a container needs to be restarted by kubelet when the resource is changed. For example it may be
+the case that a Container automatically adapts to a change in CPU, but needs to be restarted for a change in Memory to
+take effect.
+
+[In-place update of pod resources KEP]: https://github.com/kubernetes/enhancements/issues/1287
+[Kubernetes 1.27]: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.27.md#api-change-3
+[`ResizePolicy`]: https://github.com/kubernetes/api/blob/8360d82aecbc72aa039281a394ebed2eaf0c0ccc/core/v1/types.go#L2448
 
 ## Design Details
 
-VPA Updater will execute in-place updates or eviction-updates in the newly added `InPlaceOnly` and `InPlaceOrRecreate`
-modes. 
+In the update modes existing before (`Initial` and `Recreate`) only admission controller was changing pod spec (updater
+was responsible for evicting pods so that admission controller could change them during admission).
+
+In the newly added `InPlaceOnly` and `InPlaceOrRecreate` modes VPA Updater will execute in-place updates (which change
+pod spec without involving admission controller).
 
 ### Applying Updates During Pod Admission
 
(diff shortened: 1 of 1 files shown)


vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+VPA Updater will execute in-place updates or eviction-updates in the newly added `InPlaceOnly` and `InPlaceOrRecreate`
+modes. 
+
+### Applying Updates During Pod Admission
+
+For VPAs in `InPlaceOnly` and `InPlaceOrRecreate` modes VPA Admission Controller will apply updates to starting pods,
+like it does for VPAs in `Initial`, `Auto`, and `Recreate` modes.
+
+### Applying Disruption-free Updates
```
voelzmo, line 68: Are we doing something about the case where an in-place update becomes `infeasible`, because the Pod doesn't fit on this Node anymore? Otherwise it could happen that using in-place updates creates more problems for people's workloads.

We could e.g. react to events on the Pod, [which should be generated by kubelet](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources#kubelet-and-api-server-interaction)
> kubelet will generate Events on the Pod whenever a resize is accepted or rejected

I wasn't sure where/how to find which events should be generated in this case, though?

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+* Admission controller applies recommendation to pod controlled by VPA. 
+* Disruption-free in-place update applied to all containers of a pod (request in recommendation bounds).
+* Partial disruption-free update applied to some containers of a pod, some disruptive changes skipped (request in
+  recommendation bounds).
+* Disruptive in-place update applied to all containers of a pod (request out ouf recommendation bounds).
+
+There will be also scenario testing difference between `InPlaceOnly` and `InPlaceOrRecreate` modes. Disruptive in-place
+update will fail. In `InPlaceOnly` pod should not be evicted, in the `InPlaceOrRecreate` pod should be evicted and a the
+recommendation applied.
```
voelzmo, line 105: I think [there is a test case preventing switching a Pod's QoS](https://github.com/vinaykul/kubernetes/blob/7db339dba22b2427059334aa48fb3b4f356e823a/test/e2e/node/pod_resize.go#L1406) (which could happen if we scale `RequestsOnly` and eventually scale `Requests` to equal `Limits`), so this seems also like a scenario we would need to avoid and test for.

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
kgolab: FWIW I think VPA should react to both Infeasible (this is rather easy) and also to long-lasting Deferred status: "it fits on this node" might require evicting all other Pods and it might take a long time to actually achieve this (IIUC there is no controller which will just do it for VPA).

I'd like to make sure that a long-lasting Deferred status doesn't block VPA from actuating the recommendation to other Pods in the workload.

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Applying Updates During Pod Admission
+
+For VPAs in `InPlaceOnly` and `InPlaceOrRecreate` modes VPA Admission Controller will apply updates to starting pods,
+like it does for VPAs in `Initial`, `Auto`, and `Recreate` modes.
+
+### Applying Disruption-free Updates
+
+When update changes only resources for which container indicates that it doesn't require a restart then VPA can attempt
+to actuate the change without disrupting the pod. VPA Updater will:
```
kgolab, line 71: Do you expect some form of hysteresis here? In some cases a recommendation might flicker between two values and I guess even w/o a full disruption it might not make sense to alter the resources too often (if only to limit the general load in the control plane & kubelet).

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+In both `InPlaceOnly` and `InPlaceOrRecreate` modes VPA updater will attempt to apply updates that require container
+restart in place. It will update them under conditions that would trigger update with `Recreate` mode. That is it will
+apply disruptive updates if:
+
+* Sum of pods requests is below sum of recommendations `LowerBound` or
+* Sum of pods requests is above sum of recommendations `UppelBound` or
+* Difference between sum of pods requests and sum of recommendation `Target`s is more than 10% and the pod has been
+  running undisrupted for at least 24h.
+  * Successful disruptive update counts as disruption here (and prevents further disruptive updates to the pod for 24h).
```
kgolab, line 87: Isn't "undisrupted time" simply the time Pod lives?

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+For VPAs in `InPlaceOnly` and `InPlaceOrRecreate` modes VPA Admission Controller will apply updates to starting pods,
+like it does for VPAs in `Initial`, `Auto`, and `Recreate` modes.
+
+### Applying Disruption-free Updates
+
+When update changes only resources for which container indicates that it doesn't require a restart then VPA can attempt
+to actuate the change without disrupting the pod. VPA Updater will:
+* attempt to actuate such updates in place,
+* attempt to actuate them if difference between recommendation and request is at least 10%
```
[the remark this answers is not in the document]
kgolab: +1 for making it easier to compare with "disruptive updates" section (a table, maybe?).

Maybe we also need a clarification as to what happens if in-place update ends up with an error; should VPA retry it (only on recommendation change or also after a timeout?) or should it be followed up with eviction (always or only assuming the eviction criteria are matched?)

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+* Sum of pods requests is below sum of recommendations `LowerBound` or
+* Sum of pods requests is above sum of recommendations `UppelBound` or
+* Difference between sum of pods requests and sum of recommendation `Target`s is more than 10% and the pod has been
+  running undisrupted for at least 24h.
+  * Successful disruptive update counts as disruption here (and prevents further disruptive updates to the pod for 24h).
+
+In `InPlaceOrRecreate` mode (but not in `InPlaceOnly` mode) VPA updater will evict pod to actuate a recommendation if it
+attempted to apply the recommendation in place and failed.
+
```
kgolab, line 91: I'd like to have the order of updater actions explicitly stated.
I expected it would attempt in-place updates first but a) it's only my assumption b) I'm not sure it always would be the case.
I can see at least the following possibilities:
* all resources can be updated in-place and the change matches eviction criteria: attempt in-place, evict on error?
* some resources can be updated in-place, some not and the change matches eviction criteria: would VPA evict already, w/o attempting in-place update (this seems contradictory to "if necessary execute only partial updates" earlier), does this depend on which resource matches eviction criteria, other?
* all or some resources can be updated in-place but the change doesn't match eviction criteria: in-place only?
* no possibility for in-place update: go to the existing updater flow.

We also have priority ordering in updater, maybe it's worth adding a sentence how it would be changed?

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
```diff
@@ -0,0 +1,109 @@
...
+* Admission controller applies recommendation to pod controlled by VPA. 
+* Disruption-free in-place update applied to all containers of a pod (request in recommendation bounds).
+* Partial disruption-free update applied to some containers of a pod, some disruptive changes skipped (request in
+  recommendation bounds).
+* Disruptive in-place update applied to all containers of a pod (request out ouf recommendation bounds).
+
+There will be also scenario testing difference between `InPlaceOnly` and `InPlaceOrRecreate` modes. Disruptive in-place
+update will fail. In `InPlaceOnly` pod should not be evicted, in the `InPlaceOrRecreate` pod should be evicted and a the
+recommendation applied.
+
```
kgolab, line 106: Can we please have something covering error handling?
I understand that such scenarios could be tricky or simply taking a long time to execute but maybe it's still worth testing what happens on error (at least when in-place failed and recommendation warrants eviction).

kgolab:
@jbartosik , thanks for starting this! I added some comments on top of @voelzmo reviews (thanks!) - nothing major but I think it's worth discussing some finer points to avoid misunderstanding (or getting wrong expectations).

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
jbartosik: I added this (see "VPA updater will consider that the update failed if").

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
jbartosik: I'm not sure I understand what you mean by "order of updater actions". Do you mean order of operations in updaters [`RunOnce`](https://github.com/kubernetes/autoscaler/blob/114a35961a85efdf3f36859350764e5e2c0c7013/vertical-pod-autoscaler/pkg/updater/logic/updater.go#L109)?

Or do you mean since now there are up to 3 different things we could attempt for any given pod, in which of them we actually do?

The things we could do:

1. Apply full update by eviction,
2. Apply full update in-place, trigger some container restarts,
3. Apply partial update in-place, don't restart any containers

We definitely need to try 2. before attempting 1. Attempt 1. only if 2. failed.

If both 2. and 3. are an option I think we should do 2. (there's some reason we didn't do the smaller disruption-free update earlier so let's move quickly).

I'm not updating AEP yet because I'm not sure what to put there.

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
jbartosik: Yes. VPA will not apply disruption-free update unless the change is by at least 10% of the request. It's possible for that to flicker (90% samples of 100m CPU, 10% samples of 1000 mCPU, percentile changes wildly between those 2 values) but it seems very unlikely.

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
```diff
@@ -0,0 +1,109 @@
...
+[In-place update of pod resources KEP](https://github.com/kubernetes/enhancements/issues/1287) is available in alpha in
+[Kubernetes 1.27](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.27.md#api-change-3). The
+feature allows changing container resources while the container is running. It also adds 
+[`ResizePolicy`](https://github.com/kubernetes/api/blob/8360d82aecbc72aa039281a394ebed2eaf0c0ccc/core/v1/types.go#L2448)
+field to Container. This field indicates that the container should be restarted when its resources
+
+## Design Details
+
+VPA Updater will execute in-place updates or eviction-updates in the newly added `InPlaceOnly` and `InPlaceOrRecreate`
```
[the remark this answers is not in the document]
jbartosik: I added a note about that.

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
jbartosik: I added note about the error and the test case.

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
```diff
@@ -0,0 +1,109 @@
...
+apply disruptive updates if:
+
+* Sum of pods requests is below sum of recommendations `LowerBound` or
+* Sum of pods requests is above sum of recommendations `UppelBound` or
+* Difference between sum of pods requests and sum of recommendation `Target`s is more than 10% and the pod has been
+  running undisrupted for at least 24h.
+  * Successful disruptive update counts as disruption here (and prevents further disruptive updates to the pod for 24h).
+
+In `InPlaceOrRecreate` mode (but not in `InPlaceOnly` mode) VPA updater will evict pod to actuate a recommendation if it
+attempted to apply the recommendation in place and failed.
```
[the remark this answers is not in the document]
jbartosik: I added a list of failures. Any of them would result in eviction. It seems to me that what happens when we run into one of those problems is that pod keeps running with resoures it had before the attempted update and it's just that the update doesn't progress (source my reading KEP and doing manual in place updates).

But we definitely should look for other more disruptive problems. If there's something serious then we might need to drop the `InPlaceOnly` mode. In noted that in the AEP

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
```diff
@@ -0,0 +1,109 @@
...
+  * even if pod has been running for less than 24h,
+* if necessary execute only partial updates (if some changes would require a restart).
+
+### Applying Disruptive Updates
+
+In both `InPlaceOnly` and `InPlaceOrRecreate` modes VPA updater will attempt to apply updates that require container
+restart in place. It will update them under conditions that would trigger update with `Recreate` mode. That is it will
+apply disruptive updates if:
+
+* Sum of pods requests is below sum of recommendations `LowerBound` or
```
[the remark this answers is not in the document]
jbartosik: Yes, you're right. For upper/lower bound it checks each container, for "significant" change it checks changes over the whole pod. Changed

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
```diff
@@ -0,0 +1,109 @@
...
+* Sum of pods requests is above sum of recommendations `UppelBound` or
```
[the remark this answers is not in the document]
jbartosik: Done

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
```diff
@@ -0,0 +1,109 @@
...
+modes. 
+
+### Applying Updates During Pod Admission
+
+For VPAs in `InPlaceOnly` and `InPlaceOrRecreate` modes VPA Admission Controller will apply updates to starting pods,
+like it does for VPAs in `Initial`, `Auto`, and `Recreate` modes.
+
+### Applying Disruption-free Updates
+
+When update changes only resources for which container indicates that it doesn't require a restart then VPA can attempt
```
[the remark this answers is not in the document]
jbartosik: Done

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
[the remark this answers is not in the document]
jbartosik: I don't think a table is feasible. There's just too many factors to consider:
- Does the pod meet CanEvict conditions?
- Does the pod meet Quick OOM conditions?
- Is the pod long-lived?
- Is there a significant change?
- Is the pod out of bounds?
- Is the update disruption-free?
- Is the update applied in-place or by eviction?
- Did VPA attempt to apply the update in-place before but failed?

Doing all combinations we get way too many (2^8 = 256) rows, I don't see a way to compress them nicely in a table.

I added a section about it (Comparison of `UpdateMode`s), please let me know if it makes this clearer.

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
```diff
@@ -0,0 +1,109 @@
...
+* `InPlaceOnly` and
+* `InPlaceOrRecreate`.
+
+## Context
+
+[In-place update of pod resources KEP](https://github.com/kubernetes/enhancements/issues/1287) is available in alpha in
+[Kubernetes 1.27](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.27.md#api-change-3). The
+feature allows changing container resources while the container is running. It also adds 
+[`ResizePolicy`](https://github.com/kubernetes/api/blob/8360d82aecbc72aa039281a394ebed2eaf0c0ccc/core/v1/types.go#L2448)
+field to Container. This field indicates that the container should be restarted when its resources
```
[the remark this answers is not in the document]
jbartosik: Added, I reworded slightly.

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Summary
+
+
+VPA applies its recommendations with a mutating webhook, during pod creation. It can also evict
+pods expecting that it will apply the recommendation when the pod is recreated. This is a
+disruptive process so VPA has some mechanism to avoid too frequent disruptive updates. 
+
+This proposal allows VPA to apply its recommendations more frequently, with less disruption by
+using in-place update feature which is an alpha feature available in Kubernetes 1.27.
```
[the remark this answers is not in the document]
jbartosik: I changed the 2nd link:
- added opening `[` bracket
- pointed the ling to release notes for K8s 1.27

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
jbartosik: Does the "VPA updater will consider that the update failed if" part I added answer your question? Or were you thinking about something else (scale down in place but then can't scale up in place to the old size)?

jbartosik:
@kgolab @voelzmo thank you for your comments. I replied to all and I think most are resolved. Please take a look

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
pbetkier: I think Karol meant the test plan and I agree we should test out the error handling, that's the interesting / difficult part. Some should be easily tested e2e (e.g. `in-place failed and recommendation warrants eviction`), some we can test with unit tests (stubbing out resize status).

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
pbetkier: 2. for `InPlaceOrRecreate` and 3. for `InPlaceOnly`?

katlol:
Hi @kgolab @voelzmo, could you please see if another round of reviews is appropriate?

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+[`CanEvict`]: https://github.com/kubernetes/autoscaler/blob/114a35961a85efdf3f36859350764e5e2c0c7013/vertical-pod-autoscaler/pkg/updater/eviction/pods_eviction_restriction.go#LL100C10-L100C37
+[by default less than 10 minutes]: https://github.com/kubernetes/autoscaler/blob/114a35961a85efdf3f36859350764e5e2c0c7013/vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go#L37
+[`UpdatePriorityCalculator.AddPod`]: https://github.com/kubernetes/autoscaler/blob/114a35961a85efdf3f36859350764e5e2c0c7013/vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go#L81
+[by default 12h]: https://github.com/kubernetes/autoscaler/blob/114a35961a85efdf3f36859350764e5e2c0c7013/vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go#L35
+[by default 10%]: https://github.com/kubernetes/autoscaler/blob/114a35961a85efdf3f36859350764e5e2c0c7013/vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go#L33
+[Outside recommendation range]: https://github.com/kubernetes/autoscaler/blob/114a35961a85efdf3f36859350764e5e2c0c7013/vertical-pod-autoscaler/pkg/updater/priority/priority_processor.go#L73
```
voelzmo, line 168:
```suggestion
[Outside recommended range]: https://github.com/kubernetes/autoscaler/blob/114a35961a85efdf3f36859350764e5e2c0c7013/vertical-pod-autoscaler/pkg/updater/priority/priority_processor.go#L73
```

voelzmo (approved):
/lgtm

Thanks for addressing all the comments! I've just added a few comments regarding link formatting for your consideration, otherwise this looks good to me!

katlol:
Thank you all for your work on this 🤗

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
jbartosik: @kgolab @voelzmo if you think I should change the doc please let me know and I'll add changes to #5877

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
jbartosik: 2. for both `InPlaceOrRecreate` and `InPlaceOnly`

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
```diff
@@ -0,0 +1,190 @@
...
+disruptive process so VPA has some mechanism to avoid too frequent disruptive updates. 
+
+This proposal allows VPA to apply its recommendations more frequently, with less disruption by
+using the
+[in-place update feature] which is an alpha feature [available in Kubernetes 1.27.] This proposal enables only core uses
+of in place updates in VPA with intention of gathering more feedback. Any more advanced uses of in place updates in VPA
+(like applying different recommendations during pod initialization) will be introduced as separate enhancement
+proposals.
+
+[in-place update feature](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources)
```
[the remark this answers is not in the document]
jbartosik: Done in #5877

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
```diff
@@ -0,0 +1,190 @@
...
+[available in Kubernetes 1.27.](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.27.md#api-change-3)
```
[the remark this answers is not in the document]
jbartosik: Done in #5877

vertical-pod-autoscaler/enhancements/4016-in-place-updates-support/README.md
jbartosik: Done in #5877

PR #5760, opened 2023-05-16
Fix similar node group calculation for non-existing node groups

BigDarkClown:
#### What type of PR is this?

/kind fix

This change fixes an issue introduced in #5750. For non-existing node groups we do not calculate `schedulablePods` for the main node group returned from the `CreateNodeGroup` call. This breaks the node group balancing if the selected scale-up included non-existing node group.

### DIFF
--- a/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
+++ b/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
@@ -193,14 +193,17 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 		mainCreatedNodeInfo, aErr := utils.GetNodeInfoFromTemplate(createNodeGroupResult.MainCreatedNodeGroup, daemonSets, o.taintConfig)
 		if aErr == nil {
 			nodeInfos[createNodeGroupResult.MainCreatedNodeGroup.Id()] = mainCreatedNodeInfo
+			schedulablePods[createNodeGroupResult.MainCreatedNodeGroup.Id()] = o.SchedulablePods(podEquivalenceGroups, createNodeGroupResult.MainCreatedNodeGroup, mainCreatedNodeInfo)
 		} else {
 			klog.Warningf("Cannot build node info for newly created main node group %v; balancing similar node groups may not work; err=%v", createNodeGroupResult.MainCreatedNodeGroup.Id(), aErr)
-			// Use node info based on expansion candidate but upadte Id which likely changed when node group was created.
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -192,15 +192,18 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 		// one should be more in line with nodes which will be created by node group.
 		mainCreatedNodeInfo, aErr := utils.GetNodeInfoFromTemplate(createNodeGroupResult.MainCreatedNodeGroup, daemonSets, o.taintConfig)
 		if aErr == nil {
-			nodeInfos[createNodeGroupResult.MainCreatedNodeGroup.Id()] = mainCreatedNodeInfo
+			nodeInfos[bestOption.NodeGroup.Id()] = mainCreatedNodeInfo
```
MaciekPytel, line 195: Why take NodeGroup from best option, rather than from createNodeGroupResult like before? The new version assumes the group creation mutates the non-existing NG rather than create a new one. Both approaches seem reasonable enough, so why add the assumption?

BigDarkClown: I just added it for the consistency with the `else` statement bellow. We are overriding the bestOption.NodeGroup either way few lines above, and this seems more concise. 

```
bestOption.NodeGroup = createNodeGroupResult.MainCreatedNodeGroup
```

BigDarkClown: I've changed all the `bestOption.NodeGroup` to `createNodeGroupResult.MainCreatedNodeGroup.` throughout the whole "create node group" block.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -192,15 +192,18 @@ func (o *ScaleUpOrchestrator) ScaleUp(
...
+			schedulablePods[bestOption.NodeGroup.Id()] = o.SchedulablePods(podEquivalenceGroups, bestOption.NodeGroup, mainCreatedNodeInfo)
```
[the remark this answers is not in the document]
BigDarkClown: I feel like recalculating is a better option. If we recreate the nodeInfo, then recalculating the schedulablePods is also reasonable. Also note that prior to the refactor this block included running the whole binpacking for all similar node groups, so we have some saved computation to spare.

PR #5762, opened 2023-05-17
helm: enable clusterapi namespace autodiscovery

jackfrancis:

This PR enables usage of the clusterapi provider's `clusterNamespace` node autodiscovery configuration in the helm chart.

I've reworked the combinatoric helm template foo a bit so hopefully it's a bit easier to wrangle going forward (I think there's no way to avoid it being sort of horrible).

Fixes #

```release-note
NONE
```

```docs

```


elmiko:
thanks @jackfrancis , just starting to take a look

jackfrancis:
Nevermind! I was using the wrong `clusterapi` key for `--node-group-auto-discovery`, should be `namespace=` not `clusterNamespace=`

cluster-autoscaler/cloudprovider/clusterapi/README.md
```diff
@@ -49,9 +49,7 @@ most likely need other command line flags. For more information you can invoke
 
 ## Configuring node group auto discovery
 
-If you do not configure node group auto discovery, cluster autoscaler will attempt
-to match nodes against any scalable resources found in any namespace and belonging
-to any Cluster.
+You must configure node group auto discovery to inform cluster autoscaler which cluster in which to find for scalable node groups.
```
jackfrancis, line 52: @elmiko FYI, I think this original statement is not correct. If you don't provide `autoDiscovery` configuration via the helm chart, you get an error.

elmiko: looking at this https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go#L419 it leads me to believe that if unconfigured then it should be watching everything. but perhaps the permissions for the rbac are not allowing it?

jackfrancis: I haven't attempted to get further beyond the helm chart validation, so I haven't actually tried to get it to work using the runtime.

But because we can only pass in one kubeconfig, and that kubeconfig is only going to work with a single cluster (thinking of `kubeconfig-incluster` context here, so CA is installed on the mgmt cluster and managing a remote workload cluster) why would you pass in a specific workload cluster kubeconfig secret but not declare the specific workload cluster that that secret pairs with?

Maybe this scenario makes sense for a self-managed CAPI cluster? (CAPI control plane installed onto the workload cluster)

elmiko: > why would you pass in a specific workload cluster kubeconfig secret but not declare the specific workload cluster that that secret pairs with?

+1, seems like it should be specified explicitly

charts/cluster-autoscaler/templates/NOTES.txt
```diff
@@ -1,4 +1,4 @@
-{{- if or .Values.autoDiscovery.clusterName .Values.autoscalingGroups -}}
+{{- if or ( or .Values.autoDiscovery.clusterName .Values.autoDiscovery.namespace .Values.autoDiscovery.labels ) .Values.autoscalingGroups }}
```
jackfrancis, line 1: This was a bug, AFAICT this condition needs to be the same condition that encapsulates the main template matter in the deployment template. (See L1 in `charts/cluster-autoscaler/templates/deployment.yaml` below)

charts/cluster-autoscaler/templates/_helpers.tpl
```diff
@@ -109,21 +109,52 @@ Return true if the priority expander is enabled
 {{- end -}}
 
 {{/*
-Return the autodiscoveryparameters for clusterapi.
+autoDiscovery.clusterName for clusterapi.
```
jackfrancis, line 112: Sorry for the gnarly diff in this file. Essentially with the introduction of `namespace` to the combinatorics below, it made sense to decompose this a bit. Not a non-horrible way to do this using go templates, at least I couldn't figure out one.

jackfrancis:
not sure why the helm docs CI is complaining (I did make changes there but I think that should be allowed)

gjtempleton:
Ah, that's something I've been meaning to update the docs on as it keeps on catching people out, the helm-docs action generates the README.md from the [README.md.gotmpl](https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/README.md.gotmpl), so changes to the text portions need to be made there.

jackfrancis:
@gjtempleton thanks for the pro tip, done!

jackfrancis:
@gjtempleton I wonder if this change in my PR that merged yesterday (which passed CI) is throwing a wrench in this PR?

- https://github.com/kubernetes/autoscaler/pull/5768/files#diff-d0c14bc33196b8f56c06b62d79b3f92cd75f94279dedecb4f3982493eafa9054

mwielgus:
Please rebase the PR.

jackfrancis:
@mwielgus rebased, thx for the nudge

@gjtempleton can't seem to figure out how to make `helm-docs` happy on this one...

gjtempleton:
@jackfrancis the fact the README.md hasn't been regenerated with new variable in the values file included in it is causing the failing test:

```
$ helm-docs charts/. && git diff --unified=0 | cat
INFO[2023-06-28T17:44:19+01:00] Found Chart directories [charts/cluster-autoscaler]
INFO[2023-06-28T17:44:19+01:00] Generating README Documentation for chart charts/cluster-autoscaler
diff --git a/charts/cluster-autoscaler/README.md b/charts/cluster-autoscaler/README.md
index cc18f7920..b054affee 100644
--- a/charts/cluster-autoscaler/README.md
+++ b/charts/cluster-autoscaler/README.md
@@ -347,0 +348 @@ vpa:
+| autoDiscovery.namespace | string | `nil` |  |
```

jackfrancis:
@gjtempleton `brew install norwoodj/tap/helm-docs && helm-docs` did the trick :)

elmiko:
@jackfrancis happy to help get this merged, is a rebase and review all that's needed?

Shubham82:
@jackfrancis, Please rebase the PR so that it will merge.

charts/cluster-autoscaler/templates/_helpers.tpl
```diff
@@ -40,14 +40,11 @@ app.kubernetes.io/name: {{ include "cluster-autoscaler.name" . | quote }}
...
 {{/*
-Return labels, including instance, name and version.
+Return labels, including instance and name.
 */}}
 {{- define "cluster-autoscaler.labels" -}}
 {{ include "cluster-autoscaler.instance-name" . }}
 app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
-{{- if .Chart.AppVersion }}
-app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
-{{- end }}
```
gjtempleton, line 50: Are we meaning to dump this or is this a rebase artefact?

charts/cluster-autoscaler/templates/_helpers.tpl
gjtempleton: Yeah, I don't think I've ever seen a nice way to do this, but this is at least an easier to grok end result.

gjtempleton (changes requested):
One nit, one query.

mwielgus (approved):
/lgtm
/approve

PR #5764, opened 2023-05-17
cluster-autoscaler: support Brightbox image pattern

NeilW:
cluster-autoscaler/cloudprovider/brightbox

Allow scaled workers to be built from an image name pattern as well as an image id. This deals with long running clusters where the official image is updated with security changes over time.

#### What type of PR is this?

/kind feature

```release-note
Add support for building new nodes using an image name as well as an ID. 
```


mwielgus (approved):
/lgtm
/approve

PR #5548, opened 2023-02-28
Azure: Fast nodegroup backoff on failed provisioning 

domenicbozzuto:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

Adds a case to the Azure VMSS instance status check to cover an instance that failed provisioning. If the instance fails to provision (generally due to Azure being unable to provide capacity for the instance type), the error percolate to the cluster state registry and put the nodegroup into backoff (faster than if it had to wait `maxNodeProvisionTime`). 

This is inspired by the other cloudProviders that implement similar behavior, like https://github.com/kubernetes/autoscaler/pull/4489

#### Testing Methodology

An nodegroup using an instance type that was actively experiencing Azure capacity issues was selected.  A scale up was triggered for that nodegroup which promptly fails.  The nodegroup is put into backoff in ~20s, and another compatible nodegroup can be upscaled instead, without having to wait `maxNodeProvisionTime` (15m):

image

```
2023-02-28T14:14:49.615Z: virtualMachineScaleSetsClient.WaitForCreateOrUpdateResult - updateVMSSCapacity for scale set ""<NODEGROUP>"" failed: Code=""ZonalAllocationFailed"" Message=""Allocation failed. We do not have sufficient capacity for the requested VM size in this zone. Read more about improving likelihood of allocation success at http://aka.ms/allocation-guidance"" Target=""53""
2023-02-28T14:14:49.615Z: Failed to update the capacity for <NODEGROUP> with error Code=""ZonalAllocationFailed"" Message=""Allocation failed. We do not have sufficient capacity for the requested VM size in this zone. Read more about improving likelihood of allocation success at http://aka.ms/allocation-guidance"" Target=""53"", invalidate the cache so as to get the real size from API
2023-02-28T14:14:49.619Z: Provisioning has failed for VM: azure:///subscriptions/<SUBSCRIPTION>/resourceGroups/<RESOURCE_GROUP>/providers/Microsoft.Compute/virtualMachineScaleSets/<NODEGROUP>/virtualMachines/53
2023-02-28T14:14:49.621Z: Nodegroup is nil for azure:///subscriptions/<SUBSCRIPTION>/resourceGroups/<RESOURCE_GROUP>/providers/Microsoft.Compute/virtualMachineScaleSets/<NODEGROUP>/virtualMachines/53
[...]

PR #5767, opened 2023-05-17
Azure: Bugfix: Check PowerState before setting OutOfResources on instance 

domenicbozzuto:
#### What type of PR is this?
/kind bug

In https://github.com/kubernetes/autoscaler/pull/5548, I had added support for faster backoff on a failed provisioning state. We recently found a bug where a many simultaneous updates to the network profiles of a running VMs in a VMSS could sometimes fail with a `ProvisioningState/failed/NetworkingInternalOperationError` error. This would be reflected in the `vm.ProvisioningState` (which represents the status of the _last_ provisioning operation on the VM), and would result in the VMSS size being decreased and the running VM instance being removed without first properly draining the running pods on the node.

This PR adds behavior to also consider the PowerState of the VM when determining if an OutOfResources error should be created -- it now only sets the OutOfResource error if the provisioning state is failed _AND_ the instance is not running. I extended the test case I added in #5548 with several more configurations to ensure the power state behavior should be correct.

##### Reproduction with Fix

I was able to reproduce the issue and captured a situation where one of the existing VMs in a VMSS was running and got updated to a ProvisioningFailed state at the same time as a ProvisioningFailure on a newly created VM. The existing VM is reported as `cloudprovider.InstanceRunning` but the new instance triggers the provisioning failure codepath that produces an `OutOfResources` error.

```
"2023-05-17T19:43:47.151Z","Disabling scale-up for node group <vmss> until 2023-05-17 19:48:46.903222688 +0000 UTC m=+6595.425006719; errorClass=OutOfResource; errorCode=provisioning-state-failed"
...
"2023-05-17T19:43:47.135Z","VM /subscriptions/<subscription>/resourceGroups/<resourceGroup>/providers/Microsoft.Compute/virtualMachineScaleSets/<vmss>/virtualMachines/1397 reports failed provisioning state with non-running power state: PowerState/stopped"
"2023-05-17T19:43:47.135Z","Getting vm instance provisioning state Failed for /subscriptions/<subscription>/resourceGroups/<resourceGroup>/providers/Microsoft.Compute/virtualMachineScaleSets/<vmss>/virtualMachines/1397"
"2023-05-17T19:43:47.135Z","Getting vm instance provisioning state Succeeded for /subscriptions/<subscription>/resourceGroups/<resourceGroup>/providers/Microsoft.Compute/virtualMachineScaleSets/<vmss>/virtualMachines/1390"
...
"2023-05-17T19:43:47.135Z","VM /subscriptions/<subscription>/resourceGroups/<resourceGroup>/providers/Microsoft.Compute/virtualMachineScaleSets/<vmss>/virtualMachines/1330 reports a failed provisioning state but is running (PowerState/running)"
"2023-05-17T19:43:47.135Z","Getting vm instance provisioning state Failed for /subscriptions/<subscription>/resourceGroups/<resourceGroup>/providers/Microsoft.Compute/virtualMachineScaleSets/<vmss>/virtualMachines/1330"
```

Fixes #

##### expand: "instanceView"

Setting the `expand` field of `.virtualMachineScaleSetVMsClient.List` to `"instanceView"` is required to retrieve the PowerState for the VMs. It looks like eventually using `instanceView` [was referenced a few years ago](https://github.com/kubernetes/autoscaler/pull/3141#discussion_r427830194) when the provisioning state logic was first introduced. I've been running a build of the CA with this fix (and `expand=instanceView`) in some of our larger Azure clusters (1k nodes) without issue for over a week, and the net number of API calls the CA makes has not increased.

If there's some hesitancy towards setting `expand=instanceView` by default, it's probably something that could be configured via an environment variable flag (and would probably act as a pseudo-feature-gate for the fast backoff, as `PowerState` is assumed to be `PowerState/running` if the `vm.InstanceView == nil`, so in the event of a provisioning failure we'd always still report `cloudprovider.InstanceRunning`, which was the behavior before #5548) -- I'm open to exploring this if it's preferred!

##### PowerState

The azure-sdk-for-go does not directly expose the enumerations for the PowerState, even though there are official values for these in other languages like [Java](https://learn.microsoft.com/en-us/java/api/com.microsoft.azure.management.compute.powerstate?view=azure-java-stable). I saw the legacy Azure cloud provider also had logic to parse the [power state from an instanceView.Statuses,](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/vendor/k8s.io/legacy-cloud-providers/azure/azure_instances.go#L39) so I largely based this fix on that.

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
@@ -35,6 +35,18 @@ import (
 	"github.com/Azure/go-autorest/autorest/azure"
 )
 
+// PowerStates reflect the operational state of a VM
+// From https://learn.microsoft.com/en-us/java/api/com.microsoft.azure.management.compute.powerstate?view=azure-java-stable
+const (
+	PowerStateStarting     = "PowerState/starting"
+	PowerStateRunning      = "PowerState/running"
+	PowerStateStopping     = "PowerState/stopping"
+	PowerStateStopped      = "PowerState/stopped"
+	PowerStateDeallocating = "PowerState/deallocating"
+	PowerStateDeallocated  = "PowerState/deallocated"
+	PowerStateUnknown      = "PowerState/unknown"
+)
+
 var (
 	defaultVmssInstancesRefreshPeriod = 5 * time.Minute
 	vmssContextTimeout                = 3 * time.Minute
@@ -295,7 +307,7 @@ func (scaleSet *ScaleSet) GetScaleSetVms() ([]compute.VirtualMachineScaleSetVM, 
 	defer cancel()
 
 	resourceGroup := scaleSet.manager.config.ResourceGroup
-	vmList, rerr := scaleSet.manager.azClient.virtualMachineScaleSetVMsClient.List(ctx, resourceGroup, scaleSet.Name, "")
+	vmList, rerr := scaleSet.manager.azClient.virtualMachineScaleSetVMsClient.List(ctx, resourceGroup, scaleSet.Name, "instanceView")
 	klog.V(4).Infof("GetScaleSetVms: scaleSet.Name: %s, vmList: %v", scaleSet.Name, vmList)
 	if rerr != nil {
 		klog.Errorf("VirtualMachineScaleSetVMsClient.List failed for %s: %v", scaleSet.Name, rerr)
@@ -612,18 +624,26 @@ func buildInstanceCache(vmList interface{}) []cloudprovider.Instance {
 	switch vms := vmList.(type) {
 	case []compute.VirtualMachineScaleSetVM:
 		for _, vm := range vms {
-			addInstanceToCache(&instances, vm.ID, vm.ProvisioningState)
+			powerState := PowerStateRunning
+			if vm.InstanceView != nil && vm.InstanceView.Statuses != nil {
+				powerState = powerStateFromStatuses(*vm.InstanceView.Statuses)
+			}
+			addInstanceToCache(&instances, vm.ID, vm.ProvisioningState, powerState)
 		}
 	case []compute.VirtualMachine:
 		for _, vm := range vms {
-			addInstanceToCache(&instances, vm.ID, vm.ProvisioningState)
+			powerState := PowerStateRunning
+			if vm.InstanceView != nil && vm.InstanceView.Statuses != nil {
+				powerState = powerStateFromStatuses(*vm.InstanceView.Statuses)
+			}
+			addInstanceToCache(&instances, vm.ID, vm.ProvisioningState, powerState)
 		}
 	}
 
 	return instances
 }
 
-func addInstanceToCache(instances *[]cloudprovider.Instance, id *string, provisioningState *string) {
+func addInstanceToCache(instances *[]cloudprovider.Instance, id *string, provisioningState *string, powerState string) {
 	// The resource ID is empty string, which indicates the instance may be in deleting state.
 	if len(*id) == 0 {
 		return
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -35,6 +35,18 @@ import (
 	"github.com/Azure/go-autorest/autorest/azure"
 )
 
+// PowerStates reflect the operational state of a VM
+// From https://learn.microsoft.com/en-us/java/api/com.microsoft.azure.management.compute.powerstate?view=azure-java-stable
+const (
```
comtalyst, line 40: Could we move this to `azure_util.go`? 
Those are where the potentially sharable constants are located.

Let's also keep the naming consistent and descriptive about power state being from VM, something like `vmPowerStateStarting`.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -680,11 +700,21 @@ func instanceStatusFromProvisioningState(provisioningState *string) *cloudprovid
...
-			ErrorMessage: "Azure failed to provision a node for this node group",
+		// Provisioning can fail both during instance creation or after the instance is running.
+		// Per https://learn.microsoft.com/en-us/azure/virtual-machines/states-billing#provisioning-states,
+		// ProvisioningState represents the most recent provisioning state, therefore only report
+		// InstanceCreating errors when the power state indicates the instance has not yet started running
+		if !isRunningPowerState(powerState) {
+			klog.V(4).Infof("VM %s reports failed provisioning state with non-running power state: %s", resourceId, powerState)
+			status.State = cloudprovider.InstanceCreating
+			status.ErrorInfo = &cloudprovider.InstanceErrorInfo{
+				ErrorClass:   cloudprovider.OutOfResourcesErrorClass,
```
comtalyst, line 711: I don't think we should assumed all errors to be `OutOfResourcesErrorClass`. The case where other errors like #5548 could happen again.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -693,6 +723,36 @@ func instanceStatusFromProvisioningState(provisioningState *string) *cloudprovid
...
+	for _, status := range statuses {
+		if status.Code == nil || !isKnownPowerState(*status.Code) {
+			continue
+		}
+		return *status.Code
+	}
+
+	// PowerState is not set if the VM is still creating (or has failed creation),
+	// so the absence of a PowerState is treated the same as a VM that is stopped
+	return PowerStateStopped
```
comtalyst, line 736: I think we should let `PowerStateUnknown` be the default in this case. The assumption in the comment could change anytime from the updates on their side. 
And since this case is not having a behavior difference between returning `PowerStateUnknown` and `PowerStateStopped` for now, the earlier one would be safer.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -693,6 +723,36 @@ func instanceStatusFromProvisioningState(provisioningState *string) *cloudprovid
...
+}
+
+func isRunningPowerState(powerState string) bool {
+	return powerState == PowerStateRunning || powerState == PowerStateStarting
+}
+
+func isKnownPowerState(powerState string) bool {
```
comtalyst, line 743: I think functions like this and `isRunningPowerState()` could be in `azure_util.go` too. In this way we could manage the definitions of each power state from the same place.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
domenicbozzuto: Sure thing, moved the constants and related helper functions to `azure_util.go`. I also adjusted the naming of the helper functions to better reference that it's a VM power state.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
domenicbozzuto: Agreed, I've updated to make `vmPowerStateUnknown` the default case -- thanks!

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
comtalyst: I see your point—let's keep it that way.

PR #5777, opened 2023-05-19
Snapshot actuation status before refreshing cache

krzysied:
#### What type of PR is this?

/kind bug

There is a race condition between node group size and deletion operation. Having  actuation status update done before cache refresh should change incorrect behavior from node group size going below min value to having adding loop for scale-down (in the worst case).

Follow-up to pull/5562

No

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -317,7 +317,8 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
 		klog.Errorf("Failed to get daemonset list: %v", err)
 		return caerrors.ToAutoscalerError(caerrors.ApiCallError, err)
 	}
-
+	// Snapshot scale-down actuation status before cache refresh.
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -317,7 +317,8 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
 		klog.Errorf("Failed to get daemonset list: %v", err)
 		return caerrors.ToAutoscalerError(caerrors.ApiCallError, err)
 	}
-
+	// Snapshot actuation status before cache refresh.
+	actuationStatus := a.scaleDownActuator.CheckStatus()
```
x13n, line 321: nit: Since the scope of the variable is now broader, I'd consider renaming it to `scaleDownActuationStatus`.

x13n (commented):
Thanks for the fix! Only one minor comment. Feel free to cancel the hold if you disagree.

/lgtm
/approve
/hold

PR #5800, opened 2023-05-23
Move the Mig instance refresh time to GceCache

BigDarkClown:
#### What type of PR is this?

/kind bug

This issue fixes a possible map write race condition in GCE cloud provider. The instance refresh map stored in `CachedMigInfoProvider` was not guarded behind mutex so the race condition with concurrent map writes was possible.


cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -19,13 +19,24 @@ package gce
...
 	"sync"
+	"time"
 
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
 
 	gce "google.golang.org/api/compute/v1"
 	klog "k8s.io/klog/v2"
 )
 
+type timeProvider interface {
```
jayantjain93, line 30: nit: `timer`?

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -184,6 +212,7 @@ func (gc *GceCache) SetMigInstances(migRef GceRef, instances []cloudprovider.Ins
 
 	gc.removeMigInstances(migRef)
 	gc.instances[migRef] = append([]cloudprovider.Instance{}, instances...)
+	gc.instancesRefreshTime[migRef] = gc.timeProvider.Now()
```
jayantjain93, line 215: since we're not using `gc.instancesRefreshTime` internally in GceCache, I don't think it makes sense to set the state in the package either. Especially since the refresh is happening externally and we're just setting the state of the cache.

I would look to create a meta func along the lines of `SetMigInstances(migRef GceRef, instances []cloudprovider.Instance, updateTimestamp time.Time)`, signifying when this information was generated.

Thoughts?

p.s. likewise we shouldn't have a need for `timeProvider` since it's the callers responsibility to provide that.

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -211,6 +240,7 @@ func (gc *GceCache) InvalidateAllMigInstances() {
 
 	klog.V(5).Infof("Mig instances cache invalidated")
 	gc.instances = make(map[GceRef][]cloudprovider.Instance)
+	gc.instancesRefreshTime = make(map[GceRef]time.Time)
```
jayantjain93, line 243: nit: following the above comment, we should probably rename the variable to like `gc.lastUpdatedTime` / `gc.instancesLastUpdatedTime`, since we're not "refreshing" the cache, but updating the cache.

cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
```diff
@@ -175,7 +163,7 @@ func (c *cachingMigInfoProvider) findMigWithMatchingBasename(instanceRef GceRef)
 }
 
 func (c *cachingMigInfoProvider) fillMigInstances(migRef GceRef) error {
-	if val, ok := c.migInstancesLastRefreshedInfo[migRef.String()]; ok {
+	if val, ok := c.cache.GetMigInstancesRefreshTime(migRef); ok {
```
jayantjain93, line 166: Since the scope of the PR is to fix concurrent access to the map `c.migInstancesLastRefreshedInfo`, this comment could be considered out of scope.

The current implementation of the `fillMigInstances` doesn't respect `c.migInstancesMinRefreshWaitTime` since concurrent calls can fetch the last refresh time and all calls can consider the data stale and make concurrent `c.gceClient.FetchMigInstances()` calls. This points a bug in the underlying implementation of RefreshWaitTime which is only enforced in linear calls this func but not concurrent.

jayantjain93 (commented):
I have reviewed the tests since there are couple of comments that could change them.

cluster-autoscaler/cloudprovider/gce/cache.go
BigDarkClown: Removed this one.

cluster-autoscaler/cloudprovider/gce/cache.go
BigDarkClown: That's a good point, updated the `SetMigInstances`

cluster-autoscaler/cloudprovider/gce/cache.go
BigDarkClown: Done

cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
BigDarkClown: I think this is a very good point, but a bit outside of the intended scope of this PR. We can improve on this mechanism further in a separate change.

cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
```diff
@@ -175,7 +173,7 @@ func (c *cachingMigInfoProvider) findMigWithMatchingBasename(instanceRef GceRef)
 }
 
 func (c *cachingMigInfoProvider) fillMigInstances(migRef GceRef) error {
```
jayantjain93, line 175: could we add some basic test for `fillMigInstances` testing different scenarios of when refresh should happen, and the values are being set.

jayantjain93 (commented):
Rest LGTM, just one testing comment.

cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
BigDarkClown: Added.

PR #5802, opened 2023-05-23
Recalculate similar node groups for the best scale-up option

BigDarkClown:
#### What type of PR is this?

/kind cleanup
/kind bug

This change ensures that the similar node groups for the best scale-up option picked by expander are always correctly set. This needs to be done as:
- Non-existing node groups can have different IDs after creation call.
- Expander does not guarantee that the SimilarNodeGroups field won't be overridden.


mwielgus (approved):
/lgtm
/approve

PR #5810, opened 2023-05-24
Addition of BinpackingLimiter interface

kushagra98:
Addition of BinpackingLimiter interface.
The idea behind this is to stop computing expansion options if we have already found good enough expansion option/s.

Talking about why we need it?
Currently we iterate over all the unschedulable pods and for each pod we try to check if it could fit in that node. However, this computation can be improved if somehow we could analyse the current expansion options and stop the binpacking early.

Eg.  Suppose a cluster has 3 nodegroups as follows:
1. NG1 with 100 unit CPU and 100 unit memory.
2. NG2 with 10 unit CPU and 10 unit memory.
3. NG3 with 1000 unit CPU and 1000 unit memory.

Let's say the extra pod that we want to schedule requires 80 unit CPU and 80 unit memory.

Now we can see that the pod can easily fit on NG1 and NG3, however, the expansion option of NG1 is definitely a great fit. So, if we can stop computing expansion option after processing one nodegroup (here NG1), we can save on time.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
+++ b/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
@@ -119,8 +119,12 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 		}
 	}
 
+	processedNodeGroups := make(map[string]bool)
+	for _, nodegroup := range nodeGroups {
+		processedNodeGroups[nodegroup.Id()] = false
+	}
+
 	useBinpackingLimiter := false
-	nodeGroupsWithExpansionOption := make(map[string]bool)
 	if o.processors.BinpackingLimiter != nil {
 		useBinpackingLimiter = true
 		o.processors.BinpackingLimiter.InitBinpacking(o.autoscalingContext, nodeGroups)
@@ -136,6 +140,11 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 	// Filter out invalid node groups
 	validNodeGroups, skippedNodeGroups := o.filterValidScaleUpNodeGroups(nodeGroups, nodeInfos, resourcesLeft, now)
 
+	// Mark skipped node groups as processed.
+	for nodegroupID, _ := range skippedNodeGroups {
+		processedNodeGroups[nodegroupID] = true
+	}
+
 	// Calculate expansion options
 	schedulablePods := map[string][]*apiv1.Pod{}
 	var options []expander.Option
@@ -146,15 +155,15 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 
 	for _, nodeGroup := range validNodeGroups {
 		option := o.ComputeExpansionOption(nodeGroup, schedulablePods, nodeInfos, upcomingNodes, now)
+		processedNodeGroups[nodeGroup.Id()] = true
+
 		if len(option.Pods) == 0 || option.NodeCount == 0 {
 			klog.V(4).Infof("No pod can fit to %s", nodeGroup.Id())
-			continue
+		} else {
+			options = append(options, option)
 		}
 
-		nodeGroupsWithExpansionOption[nodeGroup.Id()] = true
-		options = append(options, option)
-
-		if useBinpackingLimiter && o.processors.BinpackingLimiter.StopBinpacking(o.autoscalingContext, options, nodeGroupsWithExpansionOption) {
+		if useBinpackingLimiter && o.processors.BinpackingLimiter.StopBinpacking(o.autoscalingContext, options, processedNodeGroups) {
 			break
 		}
 	}
--- a/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
+++ b/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
@@ -141,7 +141,7 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 	validNodeGroups, skippedNodeGroups := o.filterValidScaleUpNodeGroups(nodeGroups, nodeInfos, resourcesLeft, now)
 
 	// Mark skipped node groups as processed.
-	for nodegroupID, _ := range skippedNodeGroups {
+	for nodegroupID := range skippedNodeGroups {
 		processedNodeGroups[nodegroupID] = true
 	}
 
--- a/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
+++ b/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
@@ -119,6 +119,13 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 		}
 	}
 
+	useBinpackingLimiter := false
+	nodeGroupsWithExpansionOption := make(map[string]bool)
+	if o.processors.BinpackingLimiter != nil {
+		useBinpackingLimiter = true
+		o.processors.BinpackingLimiter.InitBinpacking(o.autoscalingContext, nodeGroups)
+	}
+
 	resourcesLeft, aErr := o.resourceManager.ResourcesLeft(o.autoscalingContext, nodeInfos, nodes)
 	if aErr != nil {
 		return scaleUpError(&status.ScaleUpStatus{}, aErr.AddPrefix("could not compute total resources: "))
@@ -144,7 +151,12 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 			continue
 		}
 
+		nodeGroupsWithExpansionOption[nodeGroup.Id()] = true
 		options = append(options, option)
+
+		if useBinpackingLimiter && o.processors.BinpackingLimiter.StopBinpacking(o.autoscalingContext, options, nodeGroupsWithExpansionOption) {
+			break
+		}
 	}
 
 	if len(options) == 0 {
--- a/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
+++ b/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
@@ -119,16 +119,8 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 		}
 	}
 
-	processedNodeGroups := make(map[string]bool)
-	for _, nodegroup := range nodeGroups {
-		processedNodeGroups[nodegroup.Id()] = false
-	}
-
-	useBinpackingLimiter := false
-	if o.processors.BinpackingLimiter != nil {
-		useBinpackingLimiter = true
-		o.processors.BinpackingLimiter.InitBinpacking(o.autoscalingContext, nodeGroups)
-	}
+	// Initialise binpacking limiter.
+	o.processors.BinpackingLimiter.InitBinpacking(o.autoscalingContext, nodeGroups)
 
 	resourcesLeft, aErr := o.resourceManager.ResourcesLeft(o.autoscalingContext, nodeInfos, nodes)
 	if aErr != nil {
@@ -142,7 +134,7 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 
 	// Mark skipped node groups as processed.
 	for nodegroupID := range skippedNodeGroups {
-		processedNodeGroups[nodegroupID] = true
+		o.processors.BinpackingLimiter.MarkProcessed(o.autoscalingContext, nodegroupID)
 	}
 
 	// Calculate expansion options
@@ -155,7 +147,7 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 
 	for _, nodeGroup := range validNodeGroups {
 		option := o.ComputeExpansionOption(nodeGroup, schedulablePods, nodeInfos, upcomingNodes, now)
-		processedNodeGroups[nodeGroup.Id()] = true
+		o.processors.BinpackingLimiter.MarkProcessed(o.autoscalingContext, nodeGroup.Id())
 
 		if len(option.Pods) == 0 || option.NodeCount == 0 {
 			klog.V(4).Infof("No pod can fit to %s", nodeGroup.Id())
@@ -163,7 +155,7 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 			options = append(options, option)
 		}
 
-		if useBinpackingLimiter && o.processors.BinpackingLimiter.StopBinpacking(o.autoscalingContext, options, processedNodeGroups) {
+		if o.processors.BinpackingLimiter.StopBinpacking(o.autoscalingContext, options) {
 			break
 		}
 	}
--- a/cluster-autoscaler/core/scaleup/orchestrator/orchestrator_test.go
+++ b/cluster-autoscaler/core/scaleup/orchestrator/orchestrator_test.go
@@ -805,6 +805,69 @@ func TestScaleUpUnhealthy(t *testing.T) {
 	assert.False(t, scaleUpStatus.WasSuccessful())
 }
 
+func TestBinpackingLimiter(t *testing.T) {
+	n1 := BuildTestNode("n1", 1000, 1000)
+	n2 := BuildTestNode("n2", 100000, 100000)
+	now := time.Now()
+
+	SetNodeReadyState(n1, true, now.Add(-2*time.Minute))
+	SetNodeReadyState(n2, true, now.Add(-2*time.Minute))
+
+	nodes := []*apiv1.Node{n1, n2}
+
+	podLister := kube_util.NewTestPodLister([]*apiv1.Pod{})
+	listers := kube_util.NewListerRegistry(nil, nil, podLister, nil, nil, nil, nil, nil, nil, nil)
+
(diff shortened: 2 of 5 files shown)


cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -119,32 +119,54 @@ func (o *ScaleUpOrchestrator) ScaleUp(
...
 		}
 
-		options = append(options, option)
+		if len(options) == 0 {
+			// No expansion options found in this pool. Resetting map.
+			schedulablePods = map[string][]*apiv1.Pod{}
+		} else {
+			// Stop binpacking early if some pool has expansion options.
+			break
+		}
```
qianlei90, line 169: how about this:

```suggestion
		// Stop binpacking early if some pool has expansion options.
		if len(options) > 0 {
			break
		}
		// No expansion options found in this pool. Resetting map.
		schedulablePods = map[string][]*apiv1.Pod{}
```

cluster-autoscaler/processors/processors.go
```diff
@@ -38,6 +38,8 @@ type AutoscalingProcessors struct {
 	PodListProcessor pods.PodListProcessor
 	// NodeGroupListProcessor is used to process list of NodeGroups that can be used in scale-up.
 	NodeGroupListProcessor nodegroups.NodeGroupListProcessor
+	// NodeGroupPoolListProcessor processes lists of list of NodeGroups considered in scale-up.
+	NodeGroupPoolListProcessor nodegroups.NodeGroupPoolListProcessor
 	// NodeGroupSetProcessor is used to divide scale-up between similar NodeGroups.
 	NodeGroupSetProcessor nodegroupset.NodeGroupSetProcessor
 	// ScaleUpStatusProcessor is used to process the state of the cluster after a scale-up.
```
qianlei90, line 45: These names are confused

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -119,32 +119,53 @@ func (o *ScaleUpOrchestrator) ScaleUp(
...
+	nodeGroupPoolList = append(nodeGroupPoolList, nodeGroups)
+
+	if o.processors != nil && o.processors.NodeGroupPoolListProcessor != nil {
+		var err error
+		nodeGroupPoolList, nodeInfos, err = o.processors.NodeGroupPoolListProcessor.Process(o.autoscalingContext, nodeGroupPoolList, nodeInfos, unschedulablePods)
+		if err != nil {
+			return scaleUpError(&status.ScaleUpStatus{}, errors.ToAutoscalerError(errors.InternalError, err))
+		}
+	}
+
```
BigDarkClown, line 133: I feel like we should be using only either `NodeGroupListProcessor` or `NodeGroupPoolListProcessor` in a simple loop. It seems less ambiguous from the implementation perspective. Something like

```
if o.processors.NodeGroupPoolListProcessor != nil {
  nodeGroupPoolList = o.processors.NodeGroupPoolListProcessor.ProcessPool(//arguments)
} else if o.processors.NodeGroupPoolListProcessor != nil {
  nodeGroupPoolList = [][]NodeGroup{o.processors.NodeGroupPoolListProcessor.Process(//arguments)}
} else {
  nodeGroupPoolList = [][]NodeGroup{nodeGroups}
}
```

What do you think? It seems a bit less ambiguous to me, we can be always certain what path is the implementation taking after it is initialised.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -119,32 +119,53 @@ func (o *ScaleUpOrchestrator) ScaleUp(
...
+	var skippedNodeGroups map[string]status.Reasons
 	var options []expander.Option
+	schedulablePods := map[string][]*apiv1.Pod{}
 
-	for _, nodeGroup := range validNodeGroups {
-		schedulablePods[nodeGroup.Id()] = o.SchedulablePods(podEquivalenceGroups, nodeGroup, nodeInfos[nodeGroup.Id()])
-	}
+	for _, nodeGroups = range nodeGroupPoolList {
+		// Filter out invalid node groups
+		validNodeGroups, skippedNodeGroups = o.filterValidScaleUpNodeGroups(nodeGroups, nodeInfos, resourcesLeft, now)
```
BigDarkClown, line 148: We cannot override `skippedNodeGroups` every loop, this is used for Events about the scheduling. We should join the result with the outer map instead.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -119,32 +119,53 @@ func (o *ScaleUpOrchestrator) ScaleUp(
...
+			options = append(options, option)
 		}
 
-		options = append(options, option)
+		// Stop binpacking early if some pool has expansion options.
+		if len(options) > 0 {
+			break
+		}
+		// No expansion options found in this pool. Resetting map.
+		schedulablePods = map[string][]*apiv1.Pod{}
```
BigDarkClown, line 168: This is not strictly necessary, I think that leaving these cached schedulablePods should not break anything.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -119,32 +119,53 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 		}
 	}
 
+	var nodeGroupPoolList [][]cloudprovider.NodeGroup
```
BigDarkClown, line 122: Could you add some unit-tests to cover the usage of this processor?

cluster-autoscaler/processors/nodegroups/nodegroup_pool_list_processor.go
```diff
@@ -0,0 +1,52 @@
...
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
+	"k8s.io/autoscaler/cluster-autoscaler/context"
+	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
+)
+
+// NodeGroupPoolListProcessor processes list of pools of NodeGroup considered in scale-up.
+// This list of pools can serve as priority. Pool which comes first has higher priorty.
+type NodeGroupPoolListProcessor interface {
+	Process(context *context.AutoscalingContext, nodeGroupPoolList [][]cloudprovider.NodeGroup,
```
BigDarkClown: nit: I feel like taking `nodeGroupPools []cloudprovider.NodeGroup` instead of `nodeGroupPoolList [][]cloudprovider.NodeGroup` would be a bit cleaner (same arguments as `NodeGroupListProcessor`), but don't have a strong opinion on that.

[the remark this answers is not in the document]
kushagra98: No I don't think so.
In `NodeGroupListProcessor`, the `Process()`function returns `[]cloudprovider.NodeGroup`
In `NodeGroupPoolListProcessor` the `Process()`function returns `[][]cloudprovider.NodeGroup`

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
kushagra98: Makes sense. Done.

cluster-autoscaler/processors/nodegroups/nodegroup_pool_list_processor.go
kushagra98: Yes, I changed that in the previous commit.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
kushagra98: Yes, was just saving some space. Anyway changed.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
kushagra98: Done

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
kushagra98: Done!

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
kushagra98: Just added a UT similar to the one including `NodeGroupListProcessor`. This UT might get changed in future.

MaciekPytel:
I don't particularly like an idea of introducing a new processor here. It seems to overlap in responsibilities with NodeGroupListProcessor - both modify the list of NodeGroups to be processed in scale-up. I get it that one is meant for filtering and the other one for grouping, but honestly the split feels very artificial to me. Why can't I do the sorting in NodeGroupListProcessor? Or, alternatively, why not filter NodeGroups in NodeGroupPoolListProcessor?

I'd much prefer adding a new method to NodeGroupListProcessor or changing the signature of the existing one. 

Separately from that - as I mentioned offline, I think an ability to make decision to stop binpacking dynamically would be much more flexible than pre-grouping NodeGroups.
Some examples:
1. I want to enforce maximum loop time. If I already spent X time binpacking and I have at least one expander.Option I want to break the loop. The proposed interface doesn't let me do that, I only ever want to abort early if I hit some time threshold and I don't know when that will happen.
2. I may want to abort as soon as I find a "good enough" option. Once again, I don't know which group may yield such option, I want to look at options being created and make a continue/break decision based on the result (also - I don't want the check to be just the presence of expander.Option, I actually want to evaluate it).

For those reasons I'd consider a solution where we instead expand NodeGroupListProcessor with a function like:

StopFurtherBinpacking(ctx AutoscalingContext, alreadyCalculated []expander.Option) bool

I'd argue this would allow you to implement everything your processor lets you do - just group the NodeGroups in your processor implementation and move the if `len(options)>0` check inside the processor. However, it would be much more generic, letting you make the decision to abort at any time based on any criteria.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -144,7 +151,12 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 			continue
 		}
 
+		nodeGroupsWithExpansionOption[nodeGroup.Id()] = true
 		options = append(options, option)
+
+		if useBinpackingLimiter && o.processors.BinpackingLimiter.StopBinpacking(o.autoscalingContext, options, nodeGroupsWithExpansionOption) {
```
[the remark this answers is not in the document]
kushagra98: I don't get it.  The idea of `StopBinpacking` method is to stop the binpacking early by looking at the options (even if they are empty) and processNodeGroups. It's the method definition which would return bool to take a decision.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator_test.go
```diff
@@ -805,6 +805,69 @@ func TestScaleUpUnhealthy(t *testing.T) {
...
+
+	extraPod := BuildTestPod("p-new", 500, 0)
+
+	processors := NewTestProcessors(&context)
+
+	// We should stop binpacking after finding expansion option from first node.
+	// It is because we know that extra pod (requiring 500 cpu) is a good fit on
+	// node1. The pod can also be scheduled on node n2 but it's not a good fit.
+	// Hence we can use heuristics with BinpackingLimiter to stop binpacking early
+	// to save time on computation.
```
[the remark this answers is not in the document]
kushagra98: Done.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -119,6 +119,13 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 		}
 	}
 
+	useBinpackingLimiter := false
+	nodeGroupsWithExpansionOption := make(map[string]bool)
```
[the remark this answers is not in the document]
kushagra98: Done.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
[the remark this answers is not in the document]
kushagra98: I don't think it would help anyway. If we do not have an option and call `StopBinpacking`, we are basically checking at the same state as in the previous iteration. I mean the state of the method is not changing, `StopBinpacking` takes (autoscalingContext, options, processedNodeGroups) which is not changed if we have no options. Do you think we should have more info in `processedNodeGroups`?
Currently `processedNodeGroups[nodegroup_id] = true` means node is processed (either it has an expansion option or it is skipped) and `false` means either it is unprocessed or without an expansion option.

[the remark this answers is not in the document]
kushagra98: Or `processedNodeGroups` just means that we have processed node groups uptil here irrespective of the options they have or not?

cluster-autoscaler/core/test/common.go
```diff
@@ -329,6 +331,22 @@ func (p *MockAutoprovisioningNodeGroupListProcessor) Process(context *context.Au
 func (p *MockAutoprovisioningNodeGroupListProcessor) CleanUp() {
 }
 
+// MockBinpackingLimiter is a fake BinpackingLimiter to be used in tests.
+type MockBinpackingLimiter struct {
+	requiredExpansionOptions int
+	T                        *testing.T
```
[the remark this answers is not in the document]
kushagra98: Not required. Thanks!

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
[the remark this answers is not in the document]
BigDarkClown: How I think it should work is - we set `processedNodeGroups[nodeGroup.Id()] = true` when we either:
- skip node group
- calculate an empty expansion option
- calculate a non-empty expansion option

From the standpoint of someone who implements this interface later - I would like to know which node groups were already processed. Use cases which require this that come to mind:
- I want to stop binpacking if we considered all node groups satisfying `X`.
- I want to stop binpacking when we process `p` percent of all node groups.
- I want to stop binpacking if `nodegroup` is the last one satisfying `X`.

All these use cases require knowledge on which node groups have been already processed. Current implementation does not provide that - I only know which node groups were processed and created a non-empty expansion option.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -144,7 +151,12 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 			continue
 		}
 
+		nodeGroupsWithExpansionOption[nodeGroup.Id()] = true
```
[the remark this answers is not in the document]
kushagra98: Done

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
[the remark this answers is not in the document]
kushagra98: I see. Thanks! Done.

BigDarkClown (commented):
/lgtm
/approve

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -119,6 +119,17 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 		}
 	}
 
+	processedNodeGroups := make(map[string]bool)
```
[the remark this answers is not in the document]
kushagra98: Yes, agreed. Thanks! Done.

cluster-autoscaler/processors/binpacking/binpacking_limiter.go
```diff
@@ -25,11 +25,13 @@ import (
...
 type BinpackingLimiter interface {
 	InitBinpacking(context *context.AutoscalingContext, nodeGroups []cloudprovider.NodeGroup)
-	StopBinpacking(context *context.AutoscalingContext, evaluatedOptions []expander.Option, usedNodeGroups map[string]bool) bool
+	StopBinpacking(context *context.AutoscalingContext, evaluatedOptions []expander.Option) bool
+	MarkProcessed(context *context.AutoscalingContext, nodegroupId string)
 }
 
 // NoOpBinpackingLimiter returns true without processing expansion options.
 type NoOpBinpackingLimiter struct {
+	processedNodeGroups map[string]bool
```
x13n, line 34: Why do you need it here? It is only updated, never read. (Which makes sense if this is a no-op implementation.)

kushagra98: Makes sense. Removed it from here.

PR #5826, opened 2023-05-31
Add support for scaling up with ZeroToMaxNodesScaling option

hbostan:
#### What type of PR is this?

/kind feature

This PR adds a node group AtomicScaleUp option, that allows for all-or-nothing scale up of the node group.

N/A

```release-note
Add AtomicScaleUp option that allows all-or-nothing scale up of node groups.
In atomic scale-ups, the node group should be scaled up to max size (without partial scale-ups).
```

### DIFF
--- a/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
+++ b/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/core/scaleup/equivalence"
 	"k8s.io/autoscaler/cluster-autoscaler/core/scaleup/resource"
 	"k8s.io/autoscaler/cluster-autoscaler/core/utils"
+	"k8s.io/autoscaler/cluster-autoscaler/estimator"
 	"k8s.io/autoscaler/cluster-autoscaler/expander"
 	"k8s.io/autoscaler/cluster-autoscaler/metrics"
 	ca_processors "k8s.io/autoscaler/cluster-autoscaler/processors"
@@ -130,7 +131,7 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 	now := time.Now()
 
 	// Filter out invalid node groups
-	validNodeGroups, skippedNodeGroups := o.filterValidScaleUpNodeGroups(nodeGroups, nodeInfos, resourcesLeft, now)
+	validNodeGroups, skippedNodeGroups := o.filterValidScaleUpNodeGroups(nodeGroups, nodeInfos, resourcesLeft, len(nodes)+len(upcomingNodes), now)
 
 	// Mark skipped node groups as processed.
 	for nodegroupID := range skippedNodeGroups {
@@ -418,6 +419,7 @@ func (o *ScaleUpOrchestrator) filterValidScaleUpNodeGroups(
 	nodeGroups []cloudprovider.NodeGroup,
 	nodeInfos map[string]*schedulerframework.NodeInfo,
 	resourcesLeft resource.Limits,
+	currentNodeCount int,
 	now time.Time,
 ) ([]cloudprovider.NodeGroup, map[string]status.Reasons) {
 	var validNodeGroups []cloudprovider.NodeGroup
@@ -439,6 +441,17 @@ func (o *ScaleUpOrchestrator) filterValidScaleUpNodeGroups(
 			klog.V(4).Infof("Skipping node group %s - max size reached", nodeGroup.Id())
 			skippedNodeGroups[nodeGroup.Id()] = MaxLimitReachedReason
 			continue
+		}
+		autoscalingOptions, aErr := nodeGroup.GetOptions(o.autoscalingContext.NodeGroupDefaults)
+		if aErr != nil {
+			klog.Errorf("Couldn't get autoscaling options for ng: %v", nodeGroup.Id())
+		}
+		if autoscalingOptions != nil && autoscalingOptions.AtomicScaleUp {
+			if o.autoscalingContext.MaxNodesTotal != 0 && currentNodeCount+nodeGroup.MaxSize() > o.autoscalingContext.MaxNodesTotal {
+				klog.Errorf("Skipping node group %s - atomic scale-up exceetds cluster node count limit", nodeGroup.Id())
+				skippedNodeGroups[nodeGroup.Id()] = NewSkippedReasons("atomic scale-up exceeds cluster node count limit")
+				continue
+			}
 		}
 
 		nodeInfo, found := nodeInfos[nodeGroup.Id()]
@@ -473,6 +486,30 @@ func (o *ScaleUpOrchestrator) ComputeExpansionOption(
 		return option
 	}
 
+	autoscalingOptions, err := nodeGroup.GetOptions(o.autoscalingContext.NodeGroupDefaults)
+	if err != nil {
+		klog.Errorf("Failed to get autoscaling options for node group %s: %v", nodeGroup.Id(), err)
+	}
+	if autoscalingOptions != nil && autoscalingOptions.AtomicScaleUp {
+		// If atomic, there should be no similar node groups for balancing
+		// Also the number of pods should be capped according to the node count.
+		option.IsAtomic = true
+		// Limit the number of nodes we estimate to the max scale up possible.
+		currentTargetSize, err := nodeGroup.TargetSize()
+		if err != nil {
+			klog.Errorf("Failed to get node group size: %v", err)
+			return option
+		}
+		nodeEstimationCap := nodeGroup.MaxSize() - currentTargetSize
+		limiter := estimator.NewThresholdBasedEstimationLimiter(nodeEstimationCap, 0)
+		estimator := estimator.NewBinpackingNodeEstimator(o.autoscalingContext.PredicateChecker, o.autoscalingContext.ClusterSnapshot, limiter, estimator.NewDecreasingPodOrderer())
+		option.NodeCount, option.Pods = estimator.Estimate(pods, nodeInfo, nodeGroup)
+		if option.NodeCount != nodeGroup.MaxSize() {
+			klog.Warningf("Estimator NodeCount: %d setting to %d", option.NodeCount, nodeGroup.MaxSize())
+			option.NodeCount = nodeGroup.MaxSize()
+		}
+		return option
+	}
 	estimator := o.autoscalingContext.EstimatorBuilder(o.autoscalingContext.PredicateChecker, o.autoscalingContext.ClusterSnapshot)
 	option.NodeCount, option.Pods = estimator.Estimate(pods, nodeInfo, nodeGroup)
 	option.SimilarNodeGroups = o.ComputeSimilarNodeGroups(nodeGroup, nodeInfos, schedulablePods, now)
@@ -556,6 +593,15 @@ func (o *ScaleUpOrchestrator) IsNodeGroupResourceExceeded(resourcesLeft resource
 	if err != nil {
 		klog.Errorf("Skipping node group %s; error getting node group resources: %v", nodeGroup.Id(), err)
 		return NotReadyReason
+	}
+	autoscalingOptions, aErr := nodeGroup.GetOptions(o.autoscalingContext.NodeGroupDefaults)
+	if aErr != nil {
+		klog.Errorf("Couldn't get autoscaling options for ng: %v", nodeGroup.Id())
+	}
+	if autoscalingOptions != nil && autoscalingOptions.AtomicScaleUp {
+		for resource, delta := range resourcesDelta {
+			resourcesDelta[resource] = delta * int64(nodeGroup.MaxSize())
+		}
 	}
 
 	checkResult := resource.CheckDeltaWithinLimits(resourcesLeft, resourcesDelta)
--- a/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
+++ b/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
@@ -442,8 +442,8 @@ func (o *ScaleUpOrchestrator) filterValidScaleUpNodeGroups(
 			skippedNodeGroups[nodeGroup.Id()] = MaxLimitReachedReason
 			continue
 		}
-		autoscalingOptions, aErr := nodeGroup.GetOptions(o.autoscalingContext.NodeGroupDefaults)
-		if aErr != nil {
+		autoscalingOptions, err := nodeGroup.GetOptions(o.autoscalingContext.NodeGroupDefaults)
+		if err != nil {
 			klog.Errorf("Couldn't get autoscaling options for ng: %v", nodeGroup.Id())
 		}
 		if autoscalingOptions != nil && autoscalingOptions.AtomicScaleUp {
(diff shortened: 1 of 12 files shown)


Bryce-Soghigian:
So the purpose of this feature is we just want to scale up to the max when the flag is set to true? How should scale down be defined for this feature? What is the condition for atomic scale up? Just for a normal scale operation but we scale the entire nodepool up to max count?

hbostan:
> So the purpose of this feature is we just want to scale up to the max when the flag is set to true?

Correct, when the flag is set even if only a single node is sufficient for the pod we scale up to the max.

> How should scale down be defined for this feature?

Please take a look at #5695 

> What is the condition for atomic scale up? Just for a normal scale operation but we scale the entire nodepool up to max count?

I don't quite understand the question, it is  like a normal scale operation only difference is if the node group has the atomic option we scale up the whole group.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -449,6 +462,29 @@ func (o *ScaleUpOrchestrator) ComputeExpansionOption(
...
+		option.IsAtomic = true
+		// Limit the number of nodes we estimate to the max scale up possible.
+		currentTargetSize, err := nodeGroup.TargetSize()
+		if err != nil {
+			klog.Errorf("Failed to get node group size: %v", err)
+			return option
+		}
+		nodeEstimationCap := nodeGroup.MaxSize() - currentTargetSize
+		limiter := estimator.NewThresholdBasedEstimationLimiter(nodeEstimationCap, 0)
+		estimator := estimator.NewBinpackingNodeEstimator(o.autoscalingContext.PredicateChecker, o.autoscalingContext.ClusterSnapshot, limiter, estimator.NewDecreasingPodOrderer())
```
kawych, line 481: From looking at the code it's not very clear why you need a different estimator than the regular one, and I think in general it would be better to use the estimator builder from autoscaling context for consistency... AFAICT the only problem with that would be a cap for a node number (MaxNodesPerScaleUp)? Can you add a comment explaining that?

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
kawych, line 1: Apparently github doesn't allow to comment on lines that are not directly edited by you... so...

L 241
> "newNodes, aErr = o.resourceManager.ApplyLimits(o.autoscalingContext, newNodes, resourcesLeft, nodeInfo, bestOption.Node"
Is it possible for this line to alter the number of nodes and cause a non-atomic scale-up?

Same for line 175

hbostan: `filterValidScaleUpNodeGroups()` call in L131 filters out (marks as skipped) the atomic node groups that would be capped by the `o.resourceManager.ApplyLimits()` or `o.GetCappedNewNodeCount()` calls.

To be more specific: check at L426 covers the `GetCappedNewNodeCount` which is related to the cluster wide node count limit. And the change to `IsNodeGroupResourceExceeded()` L577, covers the `ApplyLimits()` call.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
hbostan: Added a comment explaining why we create a different estimator. Maybe, we can also change the definition of `EstimatorBuilder` to also take the limiter instance as an argument?

cluster-autoscaler/expander/expander.go
```diff
@@ -47,6 +47,7 @@ type Option struct {
 	NodeCount         int
 	Debug             string
 	Pods              []*apiv1.Pod
+	IsAtomic          bool
```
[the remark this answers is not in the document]
hbostan: Removed the field

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -416,6 +418,17 @@ func (o *ScaleUpOrchestrator) filterValidScaleUpNodeGroups(
 			skippedNodeGroups[nodeGroup.Id()] = MaxLimitReachedReason
 			continue
 		}
+		autoscalingOptions, aErr := nodeGroup.GetOptions(o.autoscalingContext.NodeGroupDefaults)
+		if aErr != nil {
```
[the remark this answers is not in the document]
hbostan: changed to err

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
olagacek: If we want to do it this way, please extend estimator builder. There is a possibility some other change will be introduced, changing this behaviour (moving limiters to processor), but not sure when. Once I know more I will comment here.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -449,6 +462,35 @@ func (o *ScaleUpOrchestrator) ComputeExpansionOption(
...
+		// that can be added to the node group should be capped according to the
+		// maximum size of the node group and not the 'MaxNodesPerScaleUp' option.
+		// Since we cannot change or update the limiter when we build the estimator
+		// with the 'autoscalingContext.EstimatorBuilder' we create a new estimator
+		// with a different limiter here.
+		nodeEstimationCap := nodeGroup.MaxSize() - currentTargetSize
+		limiter := estimator.NewThresholdBasedEstimationLimiter(nodeEstimationCap, 0)
+		estimator := estimator.NewBinpackingNodeEstimator(o.autoscalingContext.PredicateChecker, o.autoscalingContext.ClusterSnapshot, limiter, estimator.NewDecreasingPodOrderer())
+		option.NodeCount, option.Pods = estimator.Estimate(pods, nodeInfo, nodeGroup)
+		if option.NodeCount != nodeGroup.MaxSize() {
```
olagacek, line 489: Similar to one of the previous comments - I think this only works correctly as long as the node group can be only at 0 or MaxSize(). Is it guaranteed?

cluster-autoscaler/core/scaleup/orchestrator/orchestrator_test.go
```diff
@@ -124,6 +124,223 @@ func TestMixedScaleUp(t *testing.T) {
 	simpleScaleUpTest(t, config, expectedResults)
 }
 
+func TestAtomicScaleUpOK(t *testing.T) {
```
olagacek, line 127: can we merge tests introduced in this file into one, table driven test?

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -462,6 +475,35 @@ func (o *ScaleUpOrchestrator) ComputeExpansionOption(
...
+			return option
+		}
+		// When the node group has the atomic scale up option, the number of nodes
+		// that can be added to the node group should be capped according to the
+		// maximum size of the node group and not the 'MaxNodesPerScaleUp' option.
+		// Since we cannot change or update the limiter when we build the estimator
+		// with the 'autoscalingContext.EstimatorBuilder' we create a new estimator
+		// with a different limiter here.
+		nodeEstimationCap := nodeGroup.MaxSize() - currentTargetSize
+		limiter := estimator.NewThresholdBasedEstimationLimiter(nodeEstimationCap, 0)
```
BigDarkClown, line 499: I think recreating the limiter here is a really BAD idea for multitude of reasons:
- Limiter is an interface. The fact that we currently use the `ThresholdBasedEstimationLimiter` does not mean that we won't change it in the future and that this new implementation won't implement some *really* important functionality, which will be de-facto broken when used with AtomicScaleUp.
- Current `ThresholdBasedEstimationLimiter` was implemented to improve the scalability of the Cluster Autoscaler. Without it we had issues with large scale-ups which were causing the CA component to get stuck and fail healthchecks, resulting in crushlooping state. I'm *strongly* against avoiding these limits here, as this can result in crashlooping state.

But most importantly - why do you need this? I get that it would be "nice" to fit as many pods as you can, but after running the estimation you are still bumping the `NodeCount` to max possible. So it doesn't really make a difference what is the limit here. As I can see it with this change there are two posibilites:
- You are able to pack N nodes (N <= max nodes), and then change N to max nodes either way.
- You pack so many nodes that the CA becomes unhealthy and crashloops.

I don't see any reason to change the limiter here - you want to scale to max no matter what, so who cares when the limiter stops the binpacking? They only effect is that you need a hacky and hard to maintain piece of code, that *also* can result in the component crashlooping. Nothing other than that changes.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -462,6 +475,35 @@ func (o *ScaleUpOrchestrator) ComputeExpansionOption(
 		return option
 	}
 
+	autoscalingOptions, err := nodeGroup.GetOptions(o.autoscalingContext.NodeGroupDefaults)
+	if err != nil {
+		klog.Errorf("Failed to get autoscaling options for node group %s: %v", nodeGroup.Id(), err)
+	}
+	if autoscalingOptions != nil && autoscalingOptions.AtomicScaleUp {
```
BigDarkClown, line 482: How is this whole logic supposed to work with similar node groups? As I can see it you are currently taking only the main node group into account, totally ignoring the similar node groups. Current scale-up logic will distribute these new nodes between all the similar node groups, so I don't think you'll get the effect you're looking for.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -546,6 +588,15 @@ func (o *ScaleUpOrchestrator) IsNodeGroupResourceExceeded(resourcesLeft resource
 		klog.Errorf("Skipping node group %s; error getting node group resources: %v", nodeGroup.Id(), err)
 		return NotReadyReason
 	}
+	autoscalingOptions, aErr := nodeGroup.GetOptions(o.autoscalingContext.NodeGroupDefaults)
+	if aErr != nil {
+		klog.Errorf("Couldn't get autoscaling options for ng: %v", nodeGroup.Id())
+	}
+	if autoscalingOptions != nil && autoscalingOptions.AtomicScaleUp {
+		for resource, delta := range resourcesDelta {
+			resourcesDelta[resource] = delta * int64(nodeGroup.MaxSize())
```
BigDarkClown, line 597: This should be maxsize - targetsize to work correctly. But even then, I think it would be better to introduce either new function or extending `IsNodeGroupResourceExceeded` to take number of nodes as an argument - putting this piece of logic here is confusing.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -462,6 +475,35 @@ func (o *ScaleUpOrchestrator) ComputeExpansionOption(
...
+		// that can be added to the node group should be capped according to the
+		// maximum size of the node group and not the 'MaxNodesPerScaleUp' option.
+		// Since we cannot change or update the limiter when we build the estimator
+		// with the 'autoscalingContext.EstimatorBuilder' we create a new estimator
+		// with a different limiter here.
+		nodeEstimationCap := nodeGroup.MaxSize() - currentTargetSize
+		limiter := estimator.NewThresholdBasedEstimationLimiter(nodeEstimationCap, 0)
+		estimator := estimator.NewBinpackingNodeEstimator(o.autoscalingContext.PredicateChecker, o.autoscalingContext.ClusterSnapshot, limiter, estimator.NewDecreasingPodOrderer())
+		option.NodeCount, option.Pods = estimator.Estimate(pods, nodeInfo, nodeGroup)
+		if option.NodeCount != nodeGroup.MaxSize() {
```
BigDarkClown, line 502: What if `nodeCount = 0`, which means that no pods can schedule there? I think in this case we should not artificially inflate the nodeCount to max.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
hbostan: It turns out we don't actually need to create a new estimator, existing one works fine for this purpose

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -416,6 +418,17 @@ func (o *ScaleUpOrchestrator) filterValidScaleUpNodeGroups(
 			skippedNodeGroups[nodeGroup.Id()] = MaxLimitReachedReason
 			continue
 		}
+		autoscalingOptions, err := nodeGroup.GetOptions(o.autoscalingContext.NodeGroupDefaults)
+		if err != nil {
+			klog.Errorf("Couldn't get autoscaling options for ng: %v", nodeGroup.Id())
+		}
+		if autoscalingOptions != nil && autoscalingOptions.AtomicScaleUp {
+			if o.autoscalingContext.MaxNodesTotal != 0 && currentNodeCount+nodeGroup.MaxSize() > o.autoscalingContext.MaxNodesTotal {
+				klog.V(4).Infof("Skipping node group %s - atomic scale-up exceetds cluster node count limit", nodeGroup.Id())
```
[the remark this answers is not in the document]
hbostan: Fixed thanks :)

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
hbostan: In atomic scale up/down we should go from 0 -> max size -> 0. Node group should be either at 0 size or maximum size

cluster-autoscaler/core/scaleup/orchestrator/orchestrator_test.go
hbostan: Done

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
hbostan: You are right. I have changed the similar node group calculation for atomic node groups to not find similar node groups.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
hbostan: As we have discussed offline, it turns out we don't need to create a new estimator and using existing one works out for us. Removed this part, now we just set the node count after the estimation is performed for atomic scale up.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
hbostan: Extended the `IsNodeGroupResourceExceeded` to take in number of nodes as well, and now it calculates resource delta with the given number of nodes.

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
hbostan: Added a check to see if the node count is greater than 0.

MaciekPytel:
Can we rename this mechanism? My expectation for 'atomic scaling' would be along the lines of handling a set of pods (perhaps a job?) in a single scale-up operation on a single NodeGroup or something along those lines. 

This is very far from what I'd expect and I can easily imagine someone enabling this feature based on their intuition around the name of the flag and getting very surprised with a large bill after their nodegroup scaled to max. Maybe something like MaxOrNothingScaling?

BigDarkClown:
I like it, the solution is much more elegant now. Great work!
/lgtm

hbostan:
> Can we rename this mechanism? 

Renamed the mechanism to `ZeroOrMaxNodeScaling`. I think this is a bit more descriptive and clear than 'atomic scaling' about what the mechanism does.

MaciekPytel:
/approve
/hold
@hbostan Can you update release note to match the new name of the feature? Once you do please feel free to unhold

PR #5833, opened 2023-06-05
Filter out invalid scale-up infos before execution

BigDarkClown:
#### What type of PR is this?

/kind cleanup

This PR filters out invalid scale-up infos before executing them. Current implementation takes care of scale-up infos with new nodes count that's smaller or equal to the current nodes count. This will help with similar node group balancing, as it will no longer be necessary to filter out "empty" scale-up infos that would otherwise cause errors.


x13n:
Shouldn't we prevent creating such scale up infos in the first place?

x13n:
/close

Closing due to lack of activity. Please reopen if still relevant.

PR #5837, opened 2023-06-06
Update VPA scripts to use v1.

### LINKED ISSUE #5832: VPA deployment script generates bad CRD manifests in modes other than "create"

Component version: 0.11 but the behavior is the same on master

K8S version: 1.23.10

Expected behavior: The script `vpa-process-yaml.sh` in `print` mode should generate the same resources as in `create` mode, minus the secret.

What happened: The script `vpa-process-yaml.sh` in `print` generated the resources but also generated twice the CustomResourceDefinition manifests from two different files: `vpa-v1-crd-gen.yaml` and `vpa-beta2-crd.yaml`.
`vpa-v1-crd-gen.yaml` is used with the `create` mode so it is expected but `vpa-beta2-crd.yaml` is added in the mix just because the selected mode is `delete`, `diff` or `print`.

How to reproduce it: Run `./hack/vpa-process-yamls.sh print` in the `vertical-pod-autoscaler` directory

Shubham82:
#### What type of PR is this?

/kind bug

This PR update the [vpa-process-yamls.sh](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/hack/vpa-process-yamls.sh) script to print the VPA  V1 CRD data.

Fixes #5832

```release-note
None
```

```docs

```

### DIFF
--- a/vertical-pod-autoscaler/hack/vpa-process-yamls.sh
+++ b/vertical-pod-autoscaler/hack/vpa-process-yamls.sh
@@ -41,6 +41,9 @@ fi
 
 ACTION=$1
 COMPONENTS="vpa-v1-crd-gen vpa-rbac updater-deployment recommender-deployment admission-controller-deployment"
+case ${ACTION} in
+delete|diff) COMPONENTS+=" vpa-beta2-crd" ;;
+esac
 
 if [ $# -gt 1 ]; then
   COMPONENTS="$2-deployment"
--- a/vertical-pod-autoscaler/hack/vpa-process-yamls.sh
+++ b/vertical-pod-autoscaler/hack/vpa-process-yamls.sh
@@ -41,9 +41,6 @@ fi
 
 ACTION=$1
 COMPONENTS="vpa-v1-crd-gen vpa-rbac updater-deployment recommender-deployment admission-controller-deployment"
-case ${ACTION} in
-delete|diff|print) COMPONENTS+=" vpa-beta2-crd" ;;
-esac
 
 if [ $# -gt 1 ]; then
   COMPONENTS="$2-deployment"


Shubham82:
Hi @jbartosik @krzysied 
PTAL!

vertical-pod-autoscaler/hack/vpa-process-yamls.sh
```diff
@@ -41,9 +41,6 @@ fi
 
 ACTION=$1
 COMPONENTS="vpa-v1-crd-gen vpa-rbac updater-deployment recommender-deployment admission-controller-deployment"
-case ${ACTION} in
-delete|diff|print) COMPONENTS+=" vpa-beta2-crd" ;;
```
jbartosik, line 45: I think we still want to delete the old CRD

jbartosik (changes requested):

vertical-pod-autoscaler/hack/vpa-process-yamls.sh
Shubham82: > so you mean to mention the old CRDs until we delete them in future releases? As I mentioned in this [comment](https://github.com/kubernetes/autoscaler/issues/5832#issuecomment-1576567132). Or Did I miss something here?


@jbartosik, could you please take a look?

jbartosik: I mean if someone has beta CRD in cluster and runs the script to delete VPA we should delete beta CRD

Shubham82: > I mean if someone has beta CRD in cluster and runs the script to delete VPA we should delete beta CRD

yeah, you are right, it should delete the beta CRD if it would be in the cluster.

jbartosik (approved):
/lgtm
/approve

PR #4489, opened 2021-11-30
Early abort if AWS node group has no capacity

drmorr0:
## Summary

This change makes it so that Cluster Autoscaler will early-abort if AWS does not have any capacity available (e.g., it is a spot node group, and there are no spot instances available).  It works by polling the `DescribeScalingActivities` AWS EC2 endpoint, and setting the node group status to an error if the `DescribeScalingActivities` call returns "Failed".  This short-circuits the 15-minute waiting period for nodes to join the cluster before the node group is added to the backoff list, which allows Cluster Autoscaler to more rapidly try a different node group that is more likely to succeed.

## Notes for reviewer:

* Recommend reviewing the commits separately
* All existing tests pass
* New unit tests for the new behaviour
* Tested in a test cluster at Airbnb (see below for details)

## Testing details

I created 3 `c5ad.16xlarge` ASGs in `test-mc-a`, and tried to scale up a test deployment in the cluster that had a "required during scheduling" node affinity for `c5ad.16xlarge`.  The node group quickly ran out of capacity, and then deleted all of the "upcoming" nodes from the node group, as you can see in the follow series of graphs:

Screen Shot 2021-11-22 at 3 08 31 PM

Specifically, note that at multiple points in the time period, it tries to scale up, creates a bunch of un-registered nodes (yellow bars in the bottom graph), and then deletes them shortly thereafter.  We can see from the following graphs that the reason for the failed scale-up is `placeholder-cannot-be-fulfilled`:

Screen Shot 2021-11-22 at 3 09 51 PM

We can also look at the logs for this time period:

```
F I1122 17:02:51.876677       1 auto_scaling_groups.go:432] Instance group kubernetes-minion-test-mc-a-amd-asg-us-east-1a has only 0 instances created while requested count is 14. Creating placeholder instances.
...
F W1122 17:02:51.935464       1 auto_scaling_groups.go:440] Instance group kubernetes-minion-test-mc-a-amd-asg-us-east-1a cannot provision any more nodes!
...
F W1122 17:02:52.094981       1 clusterstate.go:281] Disabling scale-up for node group kubernetes-minion-test-mc-a-amd-asg-us-east-1b until 2021-11-22 17:07:50.399430339 +0000 UTC m=+48821.870874859; errorClass=OutOfResource; errorCode=placeholder-cannot-be-fullfilled
...
[...]

PR #5846, opened 2023-06-09
Don't scale down ASGs when placeholder nodes are discovered

### LINKED ISSUE #5829: [AWS] Unsafe decomissioning of nodes when ASGs are out of instances

**What version of the component are you using?**: cluster-autoscaler 1.25.1

Component version: chart cluster-autoscaler-9.28.0

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.0", GitCommit:"af46c47ce925f4c4ad5cc8d1fca46c7b77d13b38", GitTreeState:"clean", BuildDate:"2020-12-08T17:59:43Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"25+", GitVersion:"v1.25.9-eks-0a21954", GitCommit:"eb82cd845d007ae98d215744675dcf7ff024a5a3", GitTreeState:"clean", BuildDate:"2023-04-15T00:37:59Z", GoVersion:"go1.19.8", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What environment is this in?**: AWS EKS

**What did you expect to happen?**: We are running EKS on managed nodes using cluster-autoscaler to dynamically scale up and down the sizes of the node groups. Unfortunately, sometimes AWS is unable to fulfil requests for scaling up (Could not launch On-Demand Instances. InsufficientInstanceCapacity). When that happens, cluster-autoscaler gets confused about the size of the ASG and sends unnecessary `SetDesiredCapacity` requests, which result in unsafe scaling down of the ASG (details below).
The expectation is that CA handles these situations transparently and does not scale down the ASG.

**What happened instead?**: We are observing the following behavior, by looking at both the CA logs and the CloudTrail logs of the requests sent. I don't know enough about the internal workings of the CA, so some of this is an assumption.
* Pods are pending, waiting for capacity in a given node group.
* CA wants to scale up this node group and sends a SetDesiredCapacity request to the ASG to increase the size (say from 3 -> 5). This happens async, so the CA fires off the request and goes back to doing different things. However it updates its internal representation of the ASG to be at capacity 5.
* ASG can’t scale up and returns that info to the CA.
* CA comes back around to the request and realizes the ASG is blocked. It updates its internal representation (back to 3)  but crucially for some reason it also updates the ASG, sending a number of SetDesiredCapacity requests, to revert the capacity back down to 3.
* In the meantime, a new instance might have successfully started, so the actual capacity of the ASG is now 4.
* So as a result to receiving the SetDesiredCapacity requests from the CA, the ASG terminates one (or more) of its running instances resulting in the unsafe scale down.

image

See the attached image that shows the sequence of events as well.

**How to reproduce it (as minimally and precisely as possible)**: Difficult to reproduce as it only happens when the AWS capacity issues occur. We are regularly observing this behavior though, I am happy to assist with more information.

theintz:
#### What type of PR is this?

/kind bug

Fixes #5829 

```release-note
[AWS] ASGs are no longer scaled down when there are capacity issues within the ASG
```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
+++ b/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
@@ -277,10 +277,6 @@ func (m *asgCache) setAsgSizeNoLock(asg *asg, size int) error {
 	return nil
 }
 
-func (m *asgCache) decreaseAsgSizeByOneNoLock(asg *asg) error {
-	return m.setAsgSizeNoLock(asg, asg.curSize-1)
-}
-
 // DeleteInstances deletes the given instances. All instances must be controlled by the same ASG.
 func (m *asgCache) DeleteInstances(instances []*AwsInstanceRef) error {
 	m.mutex.Lock()
@@ -308,11 +304,9 @@ func (m *asgCache) DeleteInstances(instances []*AwsInstanceRef) error {
 
 	for _, instance := range instances {
 		// check if the instance is a placeholder - a requested instance that was never created by the node group
-		// if it is, just decrease the size of the node group, as there's no specific instance we can remove
+		// if it is, simply remove it from the cache
 		if m.isPlaceholderInstance(instance) {
-			klog.V(4).Infof("instance %s is detected as a placeholder, decreasing ASG requested size instead "+
-				"of deleting instance", instance.Name)
-			m.decreaseAsgSizeByOneNoLock(commonAsg)
+			klog.V(4).Infof("instance %s is detected as a placeholder", instance.Name)
 		} else {
 			// check if the instance is already terminating - if it is, don't bother terminating again
 			// as doing so causes unnecessary API calls and can cause the curSize cached value to decrement
@@ -341,10 +335,10 @@ func (m *asgCache) DeleteInstances(instances []*AwsInstanceRef) error {
 				return err
 			}
 			klog.V(4).Infof(*resp.Activity.Description)
-
-			// Proactively decrement the size so autoscaler makes better decisions
-			commonAsg.curSize--
 		}
+
+		// Proactively decrement the size so autoscaler makes better decisions
+		commonAsg.curSize--
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


theintz:
We've deployed this patch on our clusters a week ago (patched into v1.26.3) and it appears to have stopped the problems outlined in #5829. Looking forward to your feedback.

qianlei90:
I'm not familiar with AWS's ASG, but I think one use case should be consider(correct me if I'm wrong)：

If EC2 is out of stock and ASG can not create new instances, CA will delete these nodes after MaxNodeProvisionTime(default `15m`). If the desired instance number of ASG is not modified, new instances will still be created and registry to cluster when stock is restored, which is not the expected behavior.

jlamillan:
By removing the node(s), but not scaling down the corresponding ASG, the CA seems to be putting itself in a state where it is requesting more nodes than the cluster workload requires. 

I thought the AWS provider already [handled](https://github.com/kubernetes/autoscaler/pull/4489) this situation more or less. i.e. the scale up operations fails quickly short-circuiting the `--max-node-provision-time` if capacity errors are observed via `DescribeScalingActivities`.

cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
```diff
@@ -308,11 +304,9 @@ func (m *asgCache) DeleteInstances(instances []*AwsInstanceRef) error {
 
 	for _, instance := range instances {
 		// check if the instance is a placeholder - a requested instance that was never created by the node group
-		// if it is, just decrease the size of the node group, as there's no specific instance we can remove
+		// if it is, simply remove it from the cache
```
jlamillan, line 307: By removing the node(s) in the local cache, but not actually scaling down the corresponding ASG, the ongoing/pending request for additional capacity on the provider may impact what capacity is accessible subsequent requests. Basically the ongoing request is at best a capacity spoof and capacity dead-lock at worst.

Put another way, if you request a 100 instances in ASG_1, and get 25 before running into a `InsufficientInstanceCapacity`, it's probably not the right answer to leave that request on-going. It's not unreasonable to think that the ongoing request for an additional 75 nodes in ASG_1 could affect subsequent requests to provision instances of that type in the same or other ASGs.

Shubham82:
Closing this PR, as issue #5829 is fixed by PR #6911 

Please feel free to reopen if you have any concerns.

/close

Issue #5170, opened 2022-09-06
Cluster autoscaler improvements for AI workloads


asm582:
**Which component are you using?**:
Cluster autoscaler component which scales Kubernetes cluster

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:

We see that the current cluster autoscaler (CA) reacts to pending pods, this may not work well for AI/HPC workloads, we outline the scenarios below:

- CA reacts to pending pods, When a large number of pods are being co-scheduled it can wait tens of minutes with pending pods before triggering scale-up, stressing the control plane.
- When multiple AI/HPC workloads are submitted to the cluster, none of the jobs may progress even after scale-up by CA due to its inability to understand gang scheduling.
  - Example: user submitted 10 AI/HPC jobs (with 10 * N pods), based on N pending pods, lets's assume CA scaled to 10 nodes, and worst case only the head nodes of all the jobs got scheduled, such partial jobs will cause no user workloads to make progress.
- CA reactive scaling can hamper cluster scale-up when a user does not have sufficient quota to create pods in a namespace.
- CA performs termination based on traditional compute (i.e. CPU and memory ) utilization threshold which may not work for AI workload(s).
  - AI workload which uses GPU with large GPU memory may cause low CPU and memory utilization on the node which could lead to premature termination.
  - Setting the termination threshold to a very low value will cause CA to not evict any nodes.
- CA terminates nodes one at a time, scaling down a cluster that uses tens or hundreds of nodes will take time.
- Selection of Machineset(s) to scale cluster is random (or complex via expander) and periodically it causes over-provisioning.

**Describe the solution you'd like.**:

Looking for feedback on a better solution to have holistic autoscaling for large AI/HPC workloads

**Describe any alternative solutions you've considered.**:

We have not encountered any alternative solution yet, please share if you have a solution to above shortcomings.

**Additional context.**:

I had opened the same issue here: https://github.com/kubernetes/community/issues/6840 which will be closed.

x13n:
Thanks for creating this issue, I believe this is an important use case Cluster Autoscaler should support. Some of the pain points you're raising require fundamental changes to how CA operates, but some should already work properly with the right setup. Let me go through them one by one:

> * CA reacts to pending pods, When a large number of pods are being co-scheduled it can wait tens of minutes with pending pods before triggering scale-up, stressing the control plane.

[...]

PR #5848, opened 2023-06-13
Introduce AEP with Provisioning Request CRD

kisieland:
#### What type of PR is this?

/kind feature

This PR adds a new AEP to support a new CRD in cluster autoscaler. \
This novel API will allow users to specify a dependency between a group of pods.

```release-note
NONE
```

N/A

### DIFF
--- a/cluster-autoscaler/proposals/provisioning-request.md
+++ b/cluster-autoscaler/proposals/provisioning-request.md
@@ -1,0 +1,326 @@
+# Provisioning Request CRD
+
+author: kisieland
+
+## Background
+
+Currently CA does not provide any way to express that a group of pods would like
+to have a capacity available.  
+This is caused by the fact that each CA loop picks a group of unschedulable pods
+and works on provisioning capacity for them, meaning that the grouping is random
+(as it depends on the kube-scheduler and CA loop interactions).  
+This is especially problematic in couple of cases:
+
+  - Users would like to have all-or-nothing semantics for their workloads.
+    Currently CA will try to provision this capacity and if it is partially
+    successful it will leave it in cluster until user removes the workload.
+  - Users would like to lower e2e scale-up latency for huge scale-ups (100
+    nodes+). Due to CA nature and kube-scheduler throughput, CA will create
+    partial scale-ups, e.g. `0->200->400->600` rather than one `0->600`. This
+    significantly increases the e2e latency as there is non-negligible time tax
+    on each scale-up operation.
+
+## Proposal
+
+### High level
+
+Provisioning Request (abbr. ProvReq) is a new namespaced Custom Resource that
+aims to allow users to ask CA for capacity for groups of pods.  
+It allows users to express the fact that group of pods is connected and should
+be threated as one entity.  
+This AEP proposes an API that can have multiple provisioning classes and can be
+extended by cloud provider specific ones.  
+This object is meant as one-shot request to CA, so that if CA fails to provision
+the capacity it is up to users to retry (such retry functionality can be added
+later on).
+
+### ProvisioningRequest CRD
+
+The following code snippets assume [kubebuilder](https://book.kubebuilder.io/)
+is used to generate the CRD:
+
+```go
+// ProvisioningRequest is a way to express additional capacity
+// that we would like to provision in the cluster. Cluster Autoscaler
+// can use this information in its calculations and signal if the capacity
+// is available in the cluster or actively add capacity if needed.
+type ProvisioningRequest struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object metadata. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata
+	//
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Spec contains specification of the ProvisioningRequest object.
+	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#spec-and-status.
+	//
+	// +kubebuilder:validation:Required
+	Spec ProvisioningRequestSpec `json:"spec"`
+	// Status of the ProvisioningRequest. CA constantly reconciles this field.
+	//
+	// +optional
+	Status ProvisioningRequestStatus `json:"status,omitempty"`
+}
+
+// ProvisioningRequestList is a object for list of ProvisioningRequest.
+type ProvisioningRequestList struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard list metadata.
+	//
+	// +optional
+	metav1.ListMeta `json:"metadata"`
+	// Items, list of ProvisioningRequest returned from API.
+	//
+	// +optional
+	Items []ProvisioningRequest `json:"items"`
+}
+
+// ProvisioningRequestSpec is a specification of additional pods for which we
+// would like to provision additional resources in the cluster.
+type ProvisioningRequestSpec struct {
+	// PodSets lists groups of pods for which we would like to provision
+	// resources.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=32
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	PodSets []PodSet `json:"podSets"`
+
+	// ProvisioningClass describes the different modes of provisioning the resources.
+	// Supported values:
+	// * check-capacity.kubernetes.io - check if current cluster state can fullfil this request,
+	//   do not reserve the capacity.
+	// * atomic-scale-up.kubernetes.io - provision the resources in an atomic manner
+    // * ... - potential other classes that are specific to the cloud providers
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	ProvisioningClass string `json:"provisioningClass"`
+
+	// AdditionalParameters contains all other parameters custom classes may require.
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	AdditionalParameters map[string]string `json:"additionalParameters"`
+}
+
+type PodSet struct {
+	// PodTemplateRef is a reference to a PodTemplate object that is representing pods 
+	// that will consume this reservation (must be within the same namespace). 
+	// Users need to make sure that the  fields relevant to scheduler (e.g. node selector tolerations)
+	// are consistent between this template and actual pods consuming the Provisioning Request.
+	//
+	// +kubebuilder:validation:Required
+	PodTemplateRef Reference `json:"podTemplateRef"`
+	// Count contains the number of pods that will be created with a given
+	// template.
+	//
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=16384
+	Count int32 `json:"count"`
+}
+
+type Reference struct {
+	// Name of the referenced object.
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names#names
+	//
+	// +kubebuilder:validation:Required
+	Name string `json:"name,omitempty"`
+}
+
+// ProvisioningRequestStatus represents the status of the resource reservation.
+type ProvisioningRequestStatus struct {
+	// Conditions represent the observations of a Provisioning Request's
+	// current state. Those will contain information whether the capacity
+    // was found/created or if there were any issues. The condition types
+    // may differ between different provisioning classes.
+	//
+	// +listType=map
+    // +listMapKey=type
+    // +patchStrategy=merge
+    // +patchMergeKey=type
+    // +optional
+	Conditions []metav1.Condition `json:"conditions"`
+
+	// AdditionalStatus contains all other status values custom provisioning classes may require.
+	//
+	// +optional
+	// +kubebuilder:validation:MaxItems=64
+	AdditionalStatus map[string]string `json:"additionalStatus"`
+}
+```
+
+### Provisioning Classes
+
+#### check-capacity.kubernetes.io class
+
+The `check-capacity.kubernetes.io` is one-off check to verify that the in the cluster
+there is enough capacity to provision given set of pods.
+
+Note: If two of such objects are created around the same time, CA will consider
+them independently and place no guards for the capacity.
+Also the capacity is not reserved in any manner so it may be scaled-down.
+
+#### atomic-scale-up.kubernetes.io class
+
+The `atomic-scale-up.kubernetes.io` aims to provision the resources required for the
+specified pods in an atomic way. The proposed logic is to:
+1. Try to provision required VMs in one loop.
+2. If it failed, remove the partially provisioned VMs and back-off.
+3. Stop the back-off after a given duration (optional), which would be passed
+   via `AdditionalParameters` field, using `ValidUntilSeconds` key and would contain string
+   denoting duration for which we should retry (measured since creation fo the CR).
+
+Note: that the VMs created in this mode are subject to the scale-down logic.
+So the duration during which users need to create the Pods is equal to the
+value of `--scale-down-unneeded-time` flag.
+
+### Adding pods that consume given ProvisioningRequest
+
+To avoid generating double scale-ups and exclude pods that are meant to consume
+given capacity CA should be able to differentiate those from all other pods.  
+To do so users need to specify the following pod annotation (it is not required
+in ProvReq’s template, though it can be specified):
+
+```yaml
+annotations:
+    "cluster-autoscaler.kubernetes.io/consume-provisioning-request": "provreq-name"
+```
+
+If it is provided for the pods that consume the ProvReq with `check-capacity.kubernetes.io` class,
+the CA will not provision the capacity, even if it was needed (as some other pods might have been 
+scheduled on it) and will result in visibility events passed to the ProvReq and pods.
+If it is not passed the CA will behave normally and just provision the capacity if it needed.
+
+Note: CA will match all pods with this annotation to a corresponding ProvReq and
+ignore them when executing a scale-up loop (so that is up to users to make sure
+that the ProvReq count is matching the number of created pods).
+If the ProvReq is missing, all of the pods that consume it will be unschedulable indefinitely.
+
+### CRD lifecycle
+
+1.  A ProvReq will be created either by the end user or by a framework.
+    At this point needed PodTemplate objects should be also created.
+2.  CA will pick it up, choose a nodepool (or create a new one if NAP is
+    enabled), and try to create nodes.
+3.  If CA successfully creates capacity, ProvReq will receive information about
+    this fact in `Conditions` field.
+4.  At this moment, users can create pods in that will consume the ProvReq (in
+    the same namespace), those will be scheduled on the capacity that was
+    created by the CA.
+5.  Once all of the pods are scheduled users can delete the ProvReq object,
+    otherwise it will be garbage collected after some time.
+6.  When pods finish the work and nodes become unused the CA will scale them
+    down.
+
+Note: Users can create a ProvReq and pods consuming them at the same time (in a
+"fire and forget" manner), but this may result in the pods being unschedulable
+and triggering user configured alerts.
+
+### Canceling the requests
+
+To cancel a pending Provisioning Request with atomic class, all that the users need to do is
+to delete the Provisioning Request object.
+After that the CA will no longer guard the nodes from deletion and proceed with standard scale-down logic.
+
+### Conditions
+
+The following Condition states should encode the states of the ProvReq:
+
+  - Provisioned - VMs were created successfully (Atomic class)
+  - CapacityAvailable - cluster contains enough capacity to schedule pods (Check
+    class)
+    * `CapacityAvailable=true` will denote that cluster contains enough capacity to schedule pods
+	* `CapacityAvailable=false` will denote that cluster does not contain enough capacity to schedule pods
+  - Failed - failed to create or check capacity (both classes)
+
+The Reasons and Messages will contain more details about why the specific
+condition was triggered.
+
+Providers of the custom classes should reuse the conditions where available or create their own ones
+if items from the above list cannot be used to denote a specific situation.
+
+### CA implementation details
+
+The proposed implementation is to handle each ProvReq in a separate scale-up
+loop. This will require changes in multiple parts of CA:
+
+1.  Listing unschedulable pods where:
+      - pods that consume ProvReq need to filtered-out
+      - pods that are represented by the ProvReq need to be injected (we need to
+        ensure those are threated as one group by the sharding logic)
+2.  Scale-up logic, which as of now has no notion atomicity and grouping of
+    pods. This is simplified as the ScaleUp logic was recently put [behind an
+    interface](https://github.com/kubernetes/autoscaler/pull/5597).
+      - This is a place where the biggest part of the change will be made. Here
+        many parts of the logic are assuming best-effort semantics and the scale
+        up size is lowered in many situations:
+          - Estimation logic, which stops after some time-out or number of
+            pods/nodes.
+          - Size limiting, which caps the scale-up to match the size
+            restrictions (on node group or cluster level).
+3.  Node creation, which needs to support atomic resize. Either via native cloud
+    provider APIs or best effort with node removal if CA is unable to fulfill
+    the scale-up.
+      - This is also quite substantial change, we can provide a generic
+        best-effort implementation that will try to scale up and clean-up nodes
+        if it is unsuccessful, but it is up to cloud providers to integrate with
+        provider specific APIs.
+4.  Scale down path is not expected to change much. But users should follow
+    [best
+    practices](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node)
+    to avoid CA disturbing their workloads.
+
+## Testing
+
+The following e2e test scenarios will be created to check whether ProvReq
+handling works as expected:
+
+1.  A new ProvReq with `check-capacity.kubernetes.io` provisioning class is created, CA
+    checks if there is enough capacity in cluster to provision specified pods.
+2.  A new ProvReq with `atomic-scale-up.kubernetes.io` provisioning class is created, CA
+    picks an appropriate node group scales it up atomically.
+3.  A new atomic ProvReq is created for which a NAP needs to provision a new
+    node group. NAP creates it CA scales it atomically.
+      - Here we should cover some of the different reasons why NAP may be
+        required.
+4.  An atomic ProvReq fails due to node group size limits and NAP CPU and/or RAM
+    limits.
+5.  Scalability tests.
+      - Scenario in which many small ProvReqs are created (strain on the number
+        of scale-up loops).
+      - Scenario in which big ProvReq is created (strain on a single scale-up
+        loop).
+
+## Limitations
+
+The current Cluster Autoscaler implementation is not taking into account [Resource Quotas](https://kubernetes.io/docs/concepts/policy/resource-quotas/). \
+The current proposal is to not include handling of the Resource Quotas, but it could be added later on.
+
+## Future Expansions
+
+### ProvisioningClass CRD
+
+One of the expansion of this approach is to introduce the ProvisioningClass CRD,
+which follows the same approach as
+[StorageClass object](https://kubernetes.io/docs/concepts/storage/storage-classes/).
+Such approach would allow administrators of the cluster to introduce a list of allowed
+ProvisioningClasses. Such CRD can also contain a pre set configuration, i.e.
+administrators may set that `atomic-scale-up.kubernetes.io` would retry up to `2h`.
+
+Possible CRD definition:
+```go
+// ProvisioningClass is a way to express provisioning classes available in the cluster.
+type ProvisioningClass struct {
+	// Name denotes the name of the object, which is to be used in the ProvisioningClass 
+	// field in Provisioning Request CRD.
+	// 
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+
+	// AdditionalParameters contains all other parameters custom classes may require.
+	//
+	// +optional
+	AdditionalParameters map[string]string `json:"additionalParameters"`
+}
+```


qianlei90:
FYI：
- https://github.com/kubernetes/autoscaler/issues/5170#issuecomment-1272019435
- kueue want to use ResourceClaim API to interactive with CA：https://docs.google.com/document/d/1fYpdB_a3YYAOHfFkx6GnEBRuAZIVfN2TeILrwQWF8w4/edit#heading=h.33c0ab4ic73i

mwielgus:
@qianlei90 Yes, it should help with the issue.

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,232 @@
...
+
+	// OperationMode describes the different modes of provisioning the resources.
+	// Supported values:
+	// * GENERIC_CHECK_CAPACITY - check if current cluster state can fullfil this request
+	// * GENERIC_ATOMIC_SCALE_UP - provision the resources in an atomic manner
+    // * ... - potential other modes that are specific to the cloud providers
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	OperationMode string `json:"operationMode" protobuf:"bytes,2,name=operationMode"`
```
mwielgus, line 95: The standard way in K8S of doing this is with *Class pattern (see StorageClass for instance). Which is the very much the same string as here, but with additional object, lets say ProvisioningClass which can be used to list the allowed values for the field.

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,232 @@
...
+
+	// AdditionalParameters contains all other parameters custom modes may require.
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	AdditionalParameters map[string]string `json:"additionalParameters" protobuf:"bytes,3,name=additionalParameters"`
```
mwielgus, line 101: If we switched to ProvisioningClass, this would go to the ProvisioningClass object and would be set by the system administrator, not the caller, who may have less knowledge of what can possibly be set. On the other hand, some workload specific parameters may not fit well into ProvisioningClass (like expected workload duration, for instance). WDYT?

kisieland: I think addition of the ProvisioningClass CRD would result in some parameters being passed in ProvisioningRequest and others in ProvisioningClass. In my opinion this obscures the parameters users may be passing to CA.
I would restrain from adding yet another CRD as for now we do not have a clear use-case for parameters admins may want to pass through it.

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,232 @@
...
+2.  CA will pick it up, choose a nodepool (or create a new one if NAP is
+    enabled), and try to create nodes.
+3.  If CA successfully creates capacity, ProvReq will receive information about
+    this fact in `Conditions` field.
+4.  At this moment, users can create pods in that will consume the ProvReq (in
+    the same namespace), those will be scheduled on the capacity that was
+    created by the CA.
+5.  Once all of the pods are scheduled users can delete the ProvReq object,
+    otherwise it will be garbage collected after some time.
+6.  When pods finish the work and nodes become unused the CA will scale them
```
qelo, line 162: What are the rules for scale down when nodes are provisioned but pods didn't start (or didn't start yet)?

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,232 @@
...
+    interface](https://github.com/kubernetes/autoscaler/pull/5597).
+      - This is a place where the biggest part of the change will be made. Here
+        many parts of the logic are assuming best-effort semantics and the scale
+        up size is lowered in many situations:
+          - Estimation logic, which stops after some time-out or number of
+            pods/nodes.
+          - Size limiting, which caps the scale-up to match the size
+            restrictions (on node group or cluster level).
+3.  Node creation, which needs to support atomic resize. Either via native cloud
+    provider APIs or best effort with node removal if CA is unable to fulfill
```
qelo, line 202: Best effort atomic could be:
1. single attempt
2. retries -- scaling down to zero and retrying later
3. staircase -- holding what we got and trying further

Which one are we planning?

cluster-autoscaler/proposals/provisioning-request.md
qelo: +1 to @kisieland , additional CRD seems premature before we have some actual admin-level parameters. As for listing, we will only have a few classes (maybe 2-4 depending how many cloud providers add), so it's easy to list in the documentation.

mwielgus: We have 30+ cloud providers in cluster autoscaler. Each of them may introduce a vendor-specific modes/classes for provisioning the resources. In the worst/best case scenario (depending how you look at it), if each of them introduces 4 we will have 120 possible values, that I would rather don't want to list in the documentation/check in validating code. 

We don't need to have ProvisioningClass with all bells and whistles from day 1, but this direction should probably be mentioned in the doc, along with the intermediate steps (like sole string with no validation), and how we get from the intermediate solution to the final one.

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,232 @@
...
+	// PodSets lists groups of pods for which we would like to provision
+	// resources.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	PodSets []PodSet `json:"podSets" protobuf:"bytes,1,rep,name=podSets"`
+
+	// ProvisioningClass describes the different modes of provisioning the resources.
+	// Supported values:
+	// * GENERIC_CHECK_CAPACITY - check if current cluster state can fullfil this request
```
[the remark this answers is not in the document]
mwielgus: I think that there may be quite a few use cases. For example, trying to run low priority jobs only when there is capacity on the statically allocated nodes. Currently K8S doesn't have gang scheduling, and this may work as a best-effort replacement. 
The problem with invalidation exist in all cases where shared/non-exclusive resources are provisioned. This should be addressed in the doc. Possible options are:

* Invalidate request that is still there.
* Treat PR as a point-in-time query and don't bother about invalidation. 

Given that we already have 2 possible behaviors for a single capacity, the params/classes may be actually needed from day 1.

cluster-autoscaler/proposals/provisioning-request.md
mwielgus: It would be great if this was somehow configurable in the longer run.

cluster-autoscaler/proposals/provisioning-request.md
kisieland: Added a section with a possible future expansion and an example of how the ProvisioningClass CRD would look like.

cluster-autoscaler/proposals/provisioning-request.md
kisieland: The current proposal would be to use the same logic for scale-down so it would be configurable via `--scale-down-unneeded-time` flag.
For now there is no way for CA to know which VMs come from the atomic scale-up so we cannot differ.

cluster-autoscaler/proposals/provisioning-request.md
kisieland: 2. Retries

I've updated the doc to describe this in more details.

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,285 @@
...
+
+Note: CA will match all pods with this annotation to a corresponding ProvReq and
+ignore them when executing a scale-up loop (so that is up to users to make sure
+that the ProvReq count is matching the number of created pods).
+
+### CRD lifecycle
+
+1.  A ProvReq will be created either by the end user or by a framework.
+2.  CA will pick it up, choose a nodepool (or create a new one if NAP is
+    enabled), and try to create nodes.
```
qianlei90, line 179: Is it possible that CA chooses multiple nodepools？ If not, what is the behavior of CA when `--balance-similar-node-groups` is enabled？

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,285 @@
...
+and works on provisioning capacity for them, meaning that the grouping is random
+(as it depends on the kube-scheduler and CA loop interactions).  
+This is especially problematic in couple of cases:
+
+  - Users would like to have all-or-nothing semantics for their workloads.
+    Currently CA will try to provision this capacity and if it is partially
+    successful it will leave it in cluster until user removes the workload.
+  - Users would like to lower e2e scale-up latency for huge scale-ups (100
+    nodes+). Due to CA nature and kube-scheduler throughput, CA will create
+    partial scale-ups, e.g. `0->200->400->600` rather than one `0->600`. This
```
Bryce-Soghigian, line 19: All or nothing/Batched workloads are an interesting concept, woudl the same be provided for scale up or is it explicitly for scale down?

kisieland: THe current proposal is to just provide the atomicity at scale-up path and let the scale-down take nodes as those become unneeded.

severinson:
This capability is interesting for the Armada project, where we'd like to control resource provisioning separately from pod scheduling. Some context: with Armada we typically have long queues of pods that can't currently be scheduled, but we may not want to provision enough capacity to run those pods immediately.

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,285 @@
...
+This AEP proposes an API that can have multiple provisioning classes and can be
+extended by cloud provider specific ones.  
+This object is meant as one-shot request to CA, so that if CA fails to provision
+the capacity it is up to users to retry (such retry functionality can be added
+later on).
+
+### ProvisioningRequest CRD
+
+The following code snippets assume [kubebuilder](https://book.kubebuilder.io/)
+is ued to generate the CRD:
```
kannon92, line 40:
```suggestion
is used to generate the CRD:
```

asm582:
Can other external controllers instead of CA use provisioning CRD to allocate resources?

cluster-autoscaler/proposals/provisioning-request.md
kisieland: THx, addressed

asm582:
> Note: One potential way we can have different components reacting to CRDs in the same cluster is to split them via ProvisioningClass.

Thanks @kisieland, can you please explain the below statement a bit more:

```
Note: One potential way we can have different components reacting to CRDs in the same cluster is to split them via ProvisioningClass.
```

kisieland:
@asm582 

We can have ProvisioningClasses `GenericAtomicScaleUp` and `CloudProviderSomOtherProvisioningMode`.
CRs with first of which would be handled by the CA and the CRs with the second one would be handled by some other component.

alculquicondor (commented):
I forgot to send my review last week :(

cluster-autoscaler/proposals/provisioning-request.md
kisieland: We can make it inmutable, while allowing the annotations to change.
Also this will make it clear when a user is providing any additional data.
But if you prefer to pipe it through annotations we should be able to d o it as well.

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,285 @@
...
+Also the capacity is not reserved in any manner so it may be scaled-down.
+
+#### GenericAtomicScaleUp class
+
+The `GenericAtomicScaleUp` aims to provision the resources required for the
+specified pods in an atomic way. The proposed logic is to:
+1. Try to provision required VMs in one loop.
+2. If it failed, remove the partially provisioned VMs and back-off.
+3. Stop the back-off after a given duration (optional), which would be passed
+   via `AdditionalParameters` field with key `valid-until-seconds` and value
```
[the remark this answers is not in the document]
kisieland: Done.

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,285 @@
...
+and triggering user configured alerts.
+
+### Conditions
+
+The following Condition states should encode the states of the ProvReq:
+
+  - Provisioned - VMs were created successfully (Atomic class)
+  - CapacityAvailable - cluster contains enough capacity to schedule pods (Check
+    class)
+  - NotEnoughCapacity - not enough capacity in cluster (Check class)
```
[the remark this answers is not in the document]
kisieland: Updated.

kisieland:
Thanks @alculquicondor addressed your comments

cluster-autoscaler/proposals/provisioning-request.md
ahaysx: Would you consider a staircase implementation for day 2? This would be beneficial for gang-scheduling use cases that require difficult to acquire machine types (e.g. GPUs).

kisieland: The classes listed here are examples of what can be supported in CA, a class where we provide VMs in a staircase is definitely possible.
Though, I think the current CA behaviour would a very similar to it, i.e. if you have unschedulable pods CA will try to create a nodes in a back-off mode, yielding a staircase like behaviour.

mwielgus:
It seems that all comments have been addressed. If you have more comments, please give them now. Otherwise I will be merging the PR on Friday.

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,293 @@
...
+### Adding pods that consume given ProvisioningRequest
+
+To avoid generating double scale-ups and exclude pods that are meant to consume
+given capacity CA should be able to differentiate those from all other pods.  
+To do so users need to specify the following pod annotation (it is not required
+in ProvReq’s template, though it can be specified):
+
+```yaml
+annotations:
+    "cluster-autoscaler.kubernetes.io/consume-provisioning-request": "provreq-name"
```
liggitt, line 175: since we have to update pods to refer back to provisionrequest anyway, is there a way to simplify/trim down/deduplicate the info embedded in provisionrequest objects and pull the pod spec info either from pod or podtemplate objects that associate themselves with the provisionrequest?

For example,

provisioningrequest `request-1` just contains the info about the existence/count of pods in podsets:
```
podSets: [
{name:"foo",count:100},
{name:"bar",count:100},
{name:"baz",count:100},
]
```

and cluster autoscaler would wait until it sees pods (or podtemplates, if pre-provisioning was desired) identifying as members of `request-1` and has podspecs for all podsets:
```
"cluster-autoscaler.kubernetes.io/provisioning-request": "request-1"
"cluster-autoscaler.kubernetes.io/podset": "foo"
```

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,293 @@
...
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	AdditionalParameters map[string]string `json:"additionalParameters" protobuf:"bytes,4,name=additionalParameters"`
+}
+
+type PodSet struct {
+	// Template representing pods that will consume this reservation.
+	// Requirements for resources (CPU, RAM, GPUs, TPUs, storage) are
+	// necessary (must be non-zero). Users need to make sure that the
```
liggitt, line 114: what about requirements / tolerations / selectors that are not known ahead of pod creation time, or are modified at pod creation time by admission plugins? the approach below where the pod spec info is pulled from actual pod or podtemplate instances could resolve this issue

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,293 @@
...
+### Conditions
+
+The following Condition states should encode the states of the ProvReq:
+
+  - Provisioned - VMs were created successfully (Atomic class)
+  - CapacityAvailable - cluster contains enough capacity to schedule pods (Check
+    class)
+    * `CapacityAvailable=true` will denote that cluster contains enough capacity to schedule pods
+	* `CapacityAvailable=false` will denote that cluster does not contain enough capacity to schedule pods
+  - Failed - failed to create or check capacity (both classes)
```
liggitt, line 210: are these expected to be set by custom provisioning classes as well?

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,293 @@
...
+	// Name denotes the name of the object, which is to be used in the ProvisioningClass 
+	// field in Provisioning Request CRD.
+	// 
+	// +kubebuilder:validation:Required
+	Name string `json:"name" protobuf:"bytes,1,name=name"`
+
+	// AdditionalParameters contains all other parameters custom classes may require.
+	//
+	// +optional
+	AdditionalParameters map[string]string `json:"additionalParameters" protobuf:"bytes,2,name=additionalParameters"`
```
liggitt, line 291: how do these differ from the AdditionalParameters in the ProvisioningRequest?

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,293 @@
...
+	// PodSets lists groups of pods for which we would like to provision
+	// resources.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	PodSets []PodSet `json:"podSets" protobuf:"bytes,1,rep,name=podSets"`
+
+	// ProvisioningClass describes the different modes of provisioning the resources.
+	// Supported values:
+	// * GenericCheckCapacity - check if current cluster state can fullfil this request
```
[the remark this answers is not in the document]
kisieland: This mode is meant to be one-shot and the capacity is not reserved within CA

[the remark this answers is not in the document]
kisieland:
```
Typically ___Class points at the name of an object of that type (StorageClass, RuntimeClass, ...). This doesn't seem to have any object associated with it, so maybe this is less of a class and more just an identifier (like SchedulerName).
```
We list such CRD under `Future Expansions`

The domain approach makes sense, updated the AEP

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,293 @@
...
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	ValidUntilSeconds *int64 `json:"validUntilSeconds" protobuf:"bytes,3,name=validUntilSeconds"`
+
+	// AdditionalParameters contains all other parameters custom classes may require.
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	AdditionalParameters map[string]string `json:"additionalParameters" protobuf:"bytes,4,name=additionalParameters"`
```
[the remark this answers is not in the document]
kisieland: Users can provide additional parameters such as:
* machine types to try
* zones to try
* what is the expected runtime of the job 
* anything that is not clearly mappable to the pod spec

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,293 @@
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	ProvisioningClass string `json:"provisioningClass" protobuf:"bytes,2,name=provisioningClass"`
+
+	// ValidUntilSeconds contains information on how long we can wait for the ProvReq to provide VMs.
+	// Field specific to GenericAtomicScaleUp provisioning class.
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	ValidUntilSeconds *int64 `json:"validUntilSeconds" protobuf:"bytes,3,name=validUntilSeconds"`
```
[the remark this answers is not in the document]
kisieland: This field can definitely be reused by custom classes. Added comment. 
What I meant is that the `kubernetes.io/checkCapacity` is not supporting it.

cluster-autoscaler/proposals/provisioning-request.md
kisieland: Those would allow cluster admins to specify cluster wide defaults (that users either could override or would be enforced).

cluster-autoscaler/proposals/provisioning-request.md
kisieland: But this would also introduce other issues, what if each of the 1000 pods has a different set of tolerations and selectors, how should CA pick the representative?
This also does not solve the issue where the pods are pending and are unschedulable generating load for scheduler and noise in the alerts/logs.
And if we were to fetch the pod template from the workload definitions then CA would be required to understand each k8s object type that can result in creation of pod, which generates a lot of other problems.

cluster-autoscaler/proposals/provisioning-request.md
kisieland: Adding support for podTemplate is definitely something we would like to do, but referencing the pod (or workload definitions) themselves seems to generate more complexity that it allows us to hide by this CRD. 

The idea with the CRD is that users do not need to create the pods up front and create them once there is capacity in cluster.

cluster-autoscaler/proposals/provisioning-request.md
liggitt: > what if each of the 1000 pods has a different set of tolerations and selectors, how should CA pick the representative?

if that happens, this design won't work properly, right? the individual pods will instruct CA to ignore them, and CA will provision capacity based on the single podspec in the provisioningrequest, right?

cluster-autoscaler/proposals/provisioning-request.md
liggitt: Embedding more and more extensive schemas in CRDs doesn't seem like it will work well long-term (due to schema skew, validation limitations, and CRD / CR size). A PodTemplate object would also allow recording the podspec without creating pods up front

cluster-autoscaler/proposals/provisioning-request.md
kisieland: > if that happens, this design won't work properly, right? the individual pods will instruct CA to ignore them, and CA will provision capacity based on the single podspec in the provisioningrequest, right?

Yes

> CA reading from Pod objects and PodTemplate objects covers this case. A single PodTemplate would communicate the spec to CA without requiring spawning pods.

But in this case users would need to create one template pod (or PodTemplate), which further complicates the user interaction with this feature.
We can add both options to the CRD and remove the embedded pod template once we move the CRD to v1.
WDYT?

liggitt: > But in this case users would need to create one template pod (or PodTemplate), which further complicates the user interaction with this feature.

Two objects is not always more complicated than one... especially if existing tooling / validation / admission works better when creating podtemplate or pod or workload objects. Are we expecting users to construct provisioning request manifests by hand? If so, the duplication leading to inaccurate copies is also a potential user experience hazard with the current design.

I think embedding podspec into many different CRDs is a locally optimized decision that ends up being a _worse_ user experience overall.

> We can add both options to the CRD and remove the embedded pod template once we move the CRD to v1.
WDYT?

If the embedded spec makes it into the CRD, I don't believe it will ever get dropped.

kisieland: Do you know what native k8s objects allow users to reference the PodTemplates?

qelo: There is also an issue of snapshotting - capturing the exact spec that was used for provisioning. Some modes of provisioning won't allow for late changes, e.g. if queuing is involved. Inline spec is cleaner for that, as we can validate mutability and it's easy to check what was used for provisioning. If we refer to external objects such as PodTemplate or the workload, it's harder to track, but I guess we could store the snapshotted revision in ProvisioningRequest status.

> I was referring to the single PodTemplate API, not looking inside deployment/job/etc objects

PodTemplate doesn't really help with having a single source of truth if real deployment/job/pods don't use it, right? What's the benefit of the separate PodTemplate if it's only used by ProvisioningRequest?


For these reasons, I think we should choose between
1. Inline spec as currently proposed (better for snapshotting and CA clients/watches/caches, but with downsides of duplicating the spec)
2. Link to the pod template in the workload object such as Job -- gives a single source of truth, but more problematic in terms of snapshotting and CA clients/watches/caches.

qelo: Another consideration (speaking for inline spec) is ability to provision with incomplete spec if the information is not needed for provisioning. It should be fine to provision nodes before knowing the exact container image.

alculquicondor: Regarding the ability to provision for incomplete specs: What are the required fields in a PodTemplate that users might not know beforehand? I cannot think of anything other than container image. If so, we could ask users to inject anything or we could do it through a webhook. Long term, we could relax the validation in apiserver to allow empty container image in PodTemplate.

kisieland: Users should be able to inject anything that is not taken into account by the default scheduler, the biggest one being the container images and names

alculquicondor: That's my point, container image and name are the only mandatory fields which could easily be mocked by a webhook.
As a stable API, it's almost impossible to add mandatory fields to Pod, so this the set of mandatory fields is unlikely to change.
@liggitt wdyt?

liggitt: > this the set of mandatory fields is unlikely to change.

that's true

cluster-autoscaler/proposals/provisioning-request.md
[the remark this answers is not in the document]
kisieland: moved it to the additional parameters

cluster-autoscaler/proposals/provisioning-request.md
kisieland: we will proceed with referencing the PodTempalte

cluster-autoscaler/proposals/provisioning-request.md
kisieland: Moved to the PodTemplate, closing the thread for now

kisieland:
Hi, @liggitt I've updated the AEP to use the Reference to PodPemplate and removed the ValidUntilSeconds field. PTAL

mwielgus:
@liggitt Ping :)

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	AdditionalParameters map[string]string `json:"additionalParameters" protobuf:"bytes,3,name=additionalParameters"`
+}
+
+type PodSet struct {
+	// TemplateRef is a reference to a PodTemplate object that is representing pods 
+	// that will consume this reservation. Users need to make sure that the
+	// fields relevant to scheduler (e.g. node selector tolerations) are
+	// consistent between this template and actual pods consuming the Provisioning Request.
+	TemplateRef *metav1.ObjectReference `json:"templateRef" protobuf:"bytes,1,name=templateRef"`
```
liggitt, line 109: metav1.ObjectReference has many fields that will be unused, prefer defining a local reference type that only has the fields you need to make the reference (I think just `name`), with appropriate validation (can limit `name` to the regex / length that is allowed for an actual podtemplate

xref guidance on object reference fields in https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#object-references

> Object references on a namespaced type should usually refer only to objects in the same namespace

The name `podTemplateRef` is slightly better / more specific

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+// ProvisioningRequest is a way to express additional capacity
+// that we would like to provision in the cluster. Cluster Autoscaler
+// can use this information in its calculations and signal if the capacity
+// is available in the cluster or actively add capacity if needed.
+type ProvisioningRequest struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object metadata. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata
+	//
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
```
liggitt, line 52: nit: drop all the proto tags, since those don't do anything in CRDs

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+}
+
+// ProvisioningRequestSpec is a specification of additional pods for which we
+// would like to provision additional resources in the cluster.
+type ProvisioningRequestSpec struct {
+	// PodSets lists groups of pods for which we would like to provision
+	// resources.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
```
liggitt, line 84: since this field is immutable, I assume if you want to change your request you have to delete and recreate

it would be good to describe in the API docs how the autoscaler handles scale-down or cleanup of interrupted / cancelled / deleted ProvisioningRequest

internally, make sure you pay attention to the uid of the provisioning request is consistent, so you notice if a request gets deleted and recreated with the same name but a different uid

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+	Items []ProvisioningRequest `json:"items" protobuf:"bytes,2,rep,name=items"`
+}
+
+// ProvisioningRequestSpec is a specification of additional pods for which we
+// would like to provision additional resources in the cluster.
+type ProvisioningRequestSpec struct {
+	// PodSets lists groups of pods for which we would like to provision
+	// resources.
+	//
+	// +kubebuilder:validation:Required
```
liggitt, line 83: should this also be minItems > 0 in addition to being required?

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+
+type PodSet struct {
+	// TemplateRef is a reference to a PodTemplate object that is representing pods 
+	// that will consume this reservation. Users need to make sure that the
+	// fields relevant to scheduler (e.g. node selector tolerations) are
+	// consistent between this template and actual pods consuming the Provisioning Request.
+	TemplateRef *metav1.ObjectReference `json:"templateRef" protobuf:"bytes,1,name=templateRef"`
+	// Count contains the number of pods that will be created with a given
+	// template.
+	// +kubebuilder:validation:Minimum=1
```
liggitt, line 112: should there be a maximum here?

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	PodSets []PodSet `json:"podSets" protobuf:"bytes,1,rep,name=podSets"`
+
+	// ProvisioningClass describes the different modes of provisioning the resources.
+	// Supported values:
+	// * kubernetes.io/checkCapacity - check if current cluster state can fullfil this request, do not reserve the capacity.
+	// * kubernetes.io/atomicScaleUp - provision the resources in an atomic manner
+    // * ... - potential other classes that are specific to the cloud providers
```
liggitt, line 91: would the cluster-autoscaler implementation ignore things that didn't use the kubernetes.io names?

what would a user do if they wanted atomicScaleUp behavior AND some cloud-provider specific provisioner?

if we name this field provisioningClass, and have a ProvisioningClass API type in the future, I'd expect the value of  this field to be the name of the API object, but making it domain-qualified (e.g. `kubernetes.io/atomicScaleUp`) is an invalid API object name (can't contain a slash). How will the ProvisioningClass instances be matched to the values specified here?

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
```
liggitt, line 94: does this need validation to make sure this is a qualified name (`example.com/...`, etc) and has a max length, etc

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+	ProvisioningClass string `json:"provisioningClass" protobuf:"bytes,2,name=provisioningClass"`
+
+	// AdditionalParameters contains all other parameters custom classes may require.
```
liggitt, line 97: This API documentation should explicitly state that only *unprivileged* parameters (as defined by the ProvisioningClass provider) are allowed here, how parameters specified here interact with the same parameters specified in a ProvisioningClass object (which take precedence), and what happens if params unrecognized by the ProvisioningClass provider are specified.

This design should also describe how a ProvisioningClass provider would split parameters into privileged and unprivileged sets and document their use.

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
```
liggitt, line 100: should there be validation on the allowed keys (qualified names, length), allowed values (length, at least), and max number of items

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+
+// ProvisioningRequestStatus represents the status of the resource reservation.
+type ProvisioningRequestStatus struct {
+	// Conditions represent the observations of a Provisioning Request's
+	// current state. Those will contain information whether the capacity
+    // was found/created or if there were any issues. The condition types
+    // may differ between different provisioning classes.
+	//
+	// +optional
+	Conditions []metav1.Condition `json:"conditions" protobuf:"bytes,1,rep,name=conditions"`
```
liggitt, line 124: make sure you include +listMapKey=type and +listType=map directives to make merging of this work properly

liggitt: cc @jpbetz to fix the guidance at https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties to include SSA / kubebuilder directives so people can copy/paste and get what we want

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+### Provisioning Classes
+
+#### kubernetes.io/checkCapacity class
+
+The `kubernetes.io/checkCapacity` is one-off check to verify that the in the cluster
+there is enough capacity to provision given set of pods.
+
+Note: If two of such objects are created around the same time, CA will consider
+them independently and place no guards for the capacity.
+Also the capacity is not reserved in any manner so it may be scaled-down.
```
liggitt, line 142: Understanding the use case would be helpful here... does this mean this is only valid if there's a single actor creating pods and provisioning requests? If so, that seems really limiting / uncertain to use.

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+### Adding pods that consume given ProvisioningRequest
+
+To avoid generating double scale-ups and exclude pods that are meant to consume
+given capacity CA should be able to differentiate those from all other pods.  
+To do so users need to specify the following pod annotation (it is not required
+in ProvReq’s template, though it can be specified):
+
+```yaml
+annotations:
+    "cluster-autoscaler.kubernetes.io/consume-provisioning-request": "provreq-name"
```
liggitt, line 167: how does this interact with a `checkCapacity` provisioning request which didn't actually reserve anything?

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+```
+
+Note: CA will match all pods with this annotation to a corresponding ProvReq and
+ignore them when executing a scale-up loop (so that is up to users to make sure
+that the ProvReq count is matching the number of created pods).
```
liggitt, line 172: describe what happens if no provisioning request with the referenced name exists (ignoring the pods and waiting for the provisioning request seems reasonable, but be explicit)

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+The `kubernetes.io/checkCapacity` is one-off check to verify that the in the cluster
+there is enough capacity to provision given set of pods.
+
+Note: If two of such objects are created around the same time, CA will consider
+them independently and place no guards for the capacity.
+Also the capacity is not reserved in any manner so it may be scaled-down.
+
+#### kubernetes.io/atomicScaleUp class
+
+The `kubernetes.io/atomicScaleUp` aims to provision the resources required for the
```
liggitt, line 146: would custom provisioners also provide an all-or-nothing option? if so, would they use a parameter to let the user indicate that's what they wanted? would making `atomicScaleUp` a well-known parameter make more sense (alongside a well-known ValidUntilSeconds parameter)?

cluster-autoscaler/proposals/provisioning-request.md
jpbetz, line 142: What are the use cases this intends to solve? I'm concerned that this operations will lead to users expecting outcomes of future requests based on the result of the capacity check.

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+
+// ProvisioningRequestSpec is a specification of additional pods for which we
+// would like to provision additional resources in the cluster.
+type ProvisioningRequestSpec struct {
+	// PodSets lists groups of pods for which we would like to provision
+	// resources.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	PodSets []PodSet `json:"podSets" protobuf:"bytes,1,rep,name=podSets"`
```
liggitt, line 85: How would an administrator control the capacity a particular namespace could request? Creation of Pod objects is bounded by resourcequota, but this type/field could allow a namespace to trigger ~infinite capacity allocation. Does cluster autoscaler check ResourceQuota to see if the namespace is allowed to use the capacity it is requesting? Does cluster autoscaler consider other provisioning requests in the namespace to decide if the namespace is allowed to request even more capacity?

cluster-autoscaler/proposals/provisioning-request.md
jpbetz: https://github.com/kubernetes/community/pull/7477

cluster-autoscaler/proposals/provisioning-request.md
jonathan-innis, line 109: This requires the the pods exist in order to piggy-back off of their existing specs. If I'm trying to deploy both this resource and something like a Deployment or ReplicaSet object at the same time, I won't know what the pods names/namespaces are before the relevant controllers actually deploy them. Does it make more sense to have this reference higher-level objects that control pod deployments but have pod template fields or to have the podSpecs statically written into the pod set?

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+
+### Adding pods that consume given ProvisioningRequest
+
+To avoid generating double scale-ups and exclude pods that are meant to consume
+given capacity CA should be able to differentiate those from all other pods.  
+To do so users need to specify the following pod annotation (it is not required
+in ProvReq’s template, though it can be specified):
+
+```yaml
+annotations:
```
jonathan-innis, line 166: Is it possible for CAS to observe that there is other capacity that is available and managed by the the ProvisioningRequest and, therefore, not scale-up for a given set of pods that can fit inside of that requested capacity?

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+1.  A ProvReq will be created either by the end user or by a framework.
+    At this point needed PodTemplate objects should be also created.
+2.  CA will pick it up, choose a nodepool (or create a new one if NAP is
+    enabled), and try to create nodes.
+3.  If CA successfully creates capacity, ProvReq will receive information about
+    this fact in `Conditions` field.
+4.  At this moment, users can create pods in that will consume the ProvReq (in
+    the same namespace), those will be scheduled on the capacity that was
+    created by the CA.
+5.  Once all of the pods are scheduled users can delete the ProvReq object,
```
jonathan-innis, line 185: What if we never want this to be garbage collected? How does a mechanism like this work for always needed static capacity that conforms to a given pod shape?

jonathan-innis (commented):
From reading through this, this strikes me as a static capacity request for "ghost pods". i.e. I want negative preemption pods to represent my capacity without those pods actually existing and being interacted with by the kube-scheduler. Is there a similar hierarchical relationship that we can build here similar to how we have varying levels of abstraction for Pod/ReplicaSet/Deployment. Is there a way to represent a capacity request as a single pod and then build off of it with a replica count? Then CAS recognizes _either_ real pods _or_ these static capacity ghost pods. There's also a question in my head about the difference in use-cases here. One use-case is I always want overhead capacity for a warm pool at all times; the other is I eventually want that static capacity to be filled and I don't want to further scale up my capacity when that pre-provisioned static capacity is utilized.

cluster-autoscaler/proposals/provisioning-request.md
kisieland: Done

cluster-autoscaler/proposals/provisioning-request.md
kisieland: Done

cluster-autoscaler/proposals/provisioning-request.md
kisieland: I've added one: 16384
Do you expect it should be bigger?

cluster-autoscaler/proposals/provisioning-request.md
kisieland: Updated PTAL

cluster-autoscaler/proposals/provisioning-request.md
kisieland: This depends on the ProvisioningClass that is chosen for for the two that are proposed here, deletion of the Provisioning Request object should allow CA to scale down VMs using standard logic.
Added a section on it.

cluster-autoscaler/proposals/provisioning-request.md
kisieland: In such a scenario, should we drop the '/' and use a '-' or '.' instead?
will the following be ok?
```
kubernetes.io.atomic-scale-up
kubernetes.io.check-capacity 
```

cluster-autoscaler/proposals/provisioning-request.md
kisieland: Changed the format, so the '/' is not needed.

As for the max length I think we should mimic the validation of the name. 
Is there an example which does mimics the name validation.

cluster-autoscaler/proposals/provisioning-request.md
kisieland: Done

cluster-autoscaler/proposals/provisioning-request.md
kisieland: This annotation is meant to avoid double counting of the needed capacity, i.e. the actual pods and the ones inferred from Provisioning Request.
Which this CA knows that for those VMs there was a dedicated capacity managed by the Provisioning Request.

cluster-autoscaler/proposals/provisioning-request.md
kisieland: > Does it make more sense to have this reference higher-level objects that control pod deployments but have pod template fields or to have the podSpecs statically written into the pod set?

Fro the CUJ mentioned above it makes sense, but imposes a serious debt on the CA code-base as either it has to be able to access all K8s higher level objects (this will not play well with custom resources) or have a complex dynamic client to access a given field of an object (which may not necessary be a valid pod template).

Due to the above the approaches that make most sense are to embed the whole template or reference the pod template object. The later on being the proffered by the @liggitt

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,289 @@
...
+## Background
+
+Currently CA does not provide any way to express that a group of pods would like
+to have a capacity available.  
+This is caused by the fact that each CA loop picks a group of unschedulable pods
+and works on provisioning capacity for them, meaning that the grouping is random
+(as it depends on the kube-scheduler and CA loop interactions).  
+This is especially problematic in couple of cases:
+
+  - Users would like to have all-or-nothing semantics for their workloads.
```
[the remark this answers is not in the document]
kisieland: It cannot be guaranteed to succeed, but what we propose in the `kubernetes.io.atomic-scale-up` mode is to avoid paying for the partially provisioned VMs and clean them up, see  the `kubernetes.io.atomic-scale-up class` section

cluster-autoscaler/proposals/provisioning-request.md
kisieland: This refers to the 2 modes that are described in this proposal, if you would like to work on a new one (let's say `kubernetes.io.static-buffer`) then those could be excluded from garbage collection logic.

cluster-autoscaler/proposals/provisioning-request.md
kisieland: I would like to not include the details of how the privileged and unprivileged parameters work in this design as the ProvisioningClass is not _strictly_ proposed here.
We discuss it and make the API ready for it but those details can be covered in a dedicated AEP.

cluster-autoscaler/proposals/provisioning-request.md
kisieland: I am not sure about requiring the qualified names, but the length of the names, entries and number of entries seem reasonable.
Although I am not sure how to limit the length of the keys and values in kubebuilder.
Is there a macro that does that?

cluster-autoscaler/proposals/provisioning-request.md
kisieland: THe use-case we had in mind is the one of Kueue or similar, where there is only one actor that admits a workload to execute.

cluster-autoscaler/proposals/provisioning-request.md
kisieland: For the `checkCapacity` it may be optional.
If it is provided the CA will not provision the capacity, even if it was needed (some other pods used it) and result in additional events passed to the ProvReq and pods. 
If it is not passed the CA will behave normally and just provision the capacity if needed.

cluster-autoscaler/proposals/provisioning-request.md
kisieland: Done

cluster-autoscaler/proposals/provisioning-request.md
kisieland: I would imagine that most of the cloud providers would use this mode and use an appropriate atomic API in the CA.
If a specific provider wants to have a better all-or-nothing option they are free to do so, but I don't think a `ValidUntilSeconds` parameter is needed.

kisieland:
@liggitt friendly ping

cluster-autoscaler/proposals/provisioning-request.md
kisieland: If there is a single actor admitting jobs to the cluster then this can serve as a preliminary check to verify that there is enough capacity in the cluster.

cluster-autoscaler/proposals/provisioning-request.md
kisieland: > Does cluster autoscaler check ResourceQuota to see if the namespace is allowed to use the capacity it is requesting?

No, the cluster autoscaler is not currently taking ResourceQuota into account.
This would be validated only at the moment of creating actual pods.

kisieland: Added a Limitations section

mwielgus:
@liggitt @jpbetz Do you have any more comments? If no (or there is no response) i will be merging the PR soon.

kisieland:
> From reading through this, this strikes me as a static capacity request for "ghost pods". i.e. I want negative preemption pods to represent my capacity without those pods actually existing and being interacted with by the kube-scheduler. Is 
Re @jpbetz:
> ... Is there a way to represent a capacity request as a single pod and then build off of it with a replica count?

Provisioning Requests can represent one pod.  But building more of those objects and expecting CAS to group them beets the purpose of the ProvReq object in the first place, why use such ProvReqs, when one can just use pods?
The issue that ProvReq aims to represent groups of Pods is to avoid CA making decisions of partial state of the cluster. 
If we used labels to group pods CAS will not know when all of the pods were created and might have started work that is not needed.

>  One use-case is I always want overhead capacity for a warm pool at all times; the other is I eventually want that static capacity to be filled and I don't want to further scale up my capacity when that pre-provisioned static capacity is utilized.

Here we propose to implement the second mode, but there is nothing prevent you or anybody else from adding a constan size buffer mode, all the required fields are here, only CAS logic would need to be slightly modified.

mwielgus (approved):
/lgtm
/approve

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,326 @@
...
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	AdditionalParameters map[string]string `json:"additionalParameters"`
+}
+
+type PodSet struct {
+	// PodTemplateRef is a reference to a PodTemplate object that is representing pods 
+	// that will consume this reservation (must be within the same namespace). 
+	// Users need to make sure that the  fields relevant to scheduler (e.g. node selector tolerations)
```
kerthcet, line 110: How autoscaler reacts to the pod requirement, e.g.`node selector`, and scaling up satisfied nodes?

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,326 @@
...
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	ProvisioningClass string `json:"provisioningClass"`
+
+	// AdditionalParameters contains all other parameters custom classes may require.
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	AdditionalParameters map[string]string `json:"additionalParameters"`
```
kerthcet, line 104: This seems for save parameters in case of later use, can `annotations` replace this? Not against this idea, just thought that maybe we can populate this field based on feedbacks.

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,326 @@
...
+}
+
+type PodSet struct {
```
kerthcet, line 107: For scheduler specifically, the general scenario should be we have a group of pods pending for provisions, then it creates a `provReq` but what the podTemplateRef refer to? The already exist pods?

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,326 @@
...
+### Adding pods that consume given ProvisioningRequest
+
+To avoid generating double scale-ups and exclude pods that are meant to consume
+given capacity CA should be able to differentiate those from all other pods.  
+To do so users need to specify the following pod annotation (it is not required
+in ProvReq’s template, though it can be specified):
+
+```yaml
+annotations:
+    "cluster-autoscaler.kubernetes.io/consume-provisioning-request": "provreq-name"
```
kerthcet, line 187: For scheduler, we can't know the **provreq-name** before creating the **provReq**, am I misunderstanding here? Or do you mean I have to update the pods?(I guess no and shouldn't)

kerthcet (commented):
Hi, I'm not quite familiar with autoscaler source code, but do care about this feature, so I left some comments, maybe not mature, forgive me. 🥲

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,326 @@
...
+}
+
+type PodSet struct {
+	// PodTemplateRef is a reference to a PodTemplate object that is representing pods 
+	// that will consume this reservation (must be within the same namespace). 
+	// Users need to make sure that the  fields relevant to scheduler (e.g. node selector tolerations)
+	// are consistent between this template and actual pods consuming the Provisioning Request.
+	//
+	// +kubebuilder:validation:Required
+	PodTemplateRef Reference `json:"podTemplateRef"`
```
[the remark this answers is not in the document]
kisieland: ProvisioningRequest is namespaced, PodTemplateRef needs to reference an object within the same namespace

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,326 @@
...
+Note: If two of such objects are created around the same time, CA will consider
+them independently and place no guards for the capacity.
+Also the capacity is not reserved in any manner so it may be scaled-down.
+
+#### atomic-scale-up.kubernetes.io class
+
+The `atomic-scale-up.kubernetes.io` aims to provision the resources required for the
+specified pods in an atomic way. The proposed logic is to:
+1. Try to provision required VMs in one loop.
+2. If it failed, remove the partially provisioned VMs and back-off.
```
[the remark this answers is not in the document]
kisieland: The current CAS logic will work in step mode, i.e. if we need 100 and we provision 99, we will leave 99 and retry creation of the remaining one.

This has a disadvantage that if user needs all 100 VMs to make progress user pays for 99 until the remaining one is provisioned.
The `kubernetes.io.atomic-scale-up` mode is to be an alternative to the above mode

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,326 @@
...
+3. Stop the back-off after a given duration (optional), which would be passed
+   via `AdditionalParameters` field, using `ValidUntilSeconds` key and would contain string
```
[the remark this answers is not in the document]
kisieland: The field would be used only for one of the two modes that are proposed, so it would be skipped in the check-capacity mode.
Because of that it was moved to the map

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,326 @@
...
+	// +kubebuilder:validation:MaxItems=64
+	AdditionalStatus map[string]string `json:"additionalStatus"`
+}
+```
+
+### Provisioning Classes
+
+#### check-capacity.kubernetes.io class
+
+The `check-capacity.kubernetes.io` is one-off check to verify that the in the cluster
```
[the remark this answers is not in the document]
kisieland: This mode will be once checked by a CA to validate if there is enough capacity in the cluster.

The user journey this tries to address is of a cluster where only one component is working in cluster to admit jobs one-by-one.

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,326 @@
...
+1.  A ProvReq will be created either by the end user or by a framework.
+    At this point needed PodTemplate objects should be also created.
+2.  CA will pick it up, choose a nodepool (or create a new one if NAP is
+    enabled), and try to create nodes.
+3.  If CA successfully creates capacity, ProvReq will receive information about
+    this fact in `Conditions` field.
+4.  At this moment, users can create pods in that will consume the ProvReq (in
+    the same namespace), those will be scheduled on the capacity that was
+    created by the CA.
+5.  Once all of the pods are scheduled users can delete the ProvReq object,
```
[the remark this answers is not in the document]
kisieland: yup, but those can be left to be garbage collected

cluster-autoscaler/proposals/provisioning-request.md
```diff
@@ -0,0 +1,326 @@
...
+
+Note: Users can create a ProvReq and pods consuming them at the same time (in a
+"fire and forget" manner), but this may result in the pods being unschedulable
+and triggering user configured alerts.
+
+### Canceling the requests
+
+To cancel a pending Provisioning Request with atomic class, all that the users need to do is
+to delete the Provisioning Request object.
+After that the CA will no longer guard the nodes from deletion and proceed with standard scale-down logic.
```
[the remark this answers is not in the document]
kisieland: yes

cluster-autoscaler/proposals/provisioning-request.md
kisieland: By running a kube-scheduler ciode inside of the binary.

This is the same way CA does it for pods

cluster-autoscaler/proposals/provisioning-request.md
kisieland: PodTemplateRef should refer to the PodTemplate k8s object, which is just a template to create of Pod.

cluster-autoscaler/proposals/provisioning-request.md
kisieland: Users can create a ProvReq with arbitrary name, so it is only the matter of making sure that it is matching between the pods and the ProvReq

cluster-autoscaler/proposals/provisioning-request.md
kisieland: `annotations` could be used, but this way it is more explicit.

liggitt:
> @liggitt @jpbetz Do you have any more comments? If no (or there is no response) i will be merging the PR soon.

Sorry for the delay, we have a bi-weekly API review that will be reviewing the updates.

I see a CRD was merged already in https://github.com/kubernetes/autoscaler/pull/6104 linking to this PR as API approval. 
Lazy consensus does not constitute API approval for k8s.io APIs. Please move the CRD to the `x-k8s.io` domain if you want to proceed without API approval, or revert the PR until approval is given.

mwielgus:
The API was meant to go to x-k8s.io from the very beginning. Thanks for noticing it. It will be fixed soon.

kisieland:
Thanks @liggitt for noticing created https://github.com/kubernetes/autoscaler/pull/6108 to address this!

PR #5852, opened 2023-06-14
When comparing PodSpec semantically, drop projected volumes for init containers

azylinski:
#### What type of PR is this?

/kind bug

CA groups pods based on their scheduling requirements. They are grouped based on pod spec and we assume that pods belonging to the same controller should have similar requirements. However, there are certain fields that differ between the pods controlled by the same controller. If we know that they do not affect scheduling we should drop those for grouping purposes.

One of such fields is projected volume mount. We drop them currently, but only for containers, while they can exist for init containers as well.

```release-note
NONE
```


mwielgus (approved):
/lgtm
/approve

PR #5853, opened 2023-06-14
Fix zonal GCE outage breaking CA when only some of the zones are failed

damikag:
#### What type of PR is this?
/kind bug

GKE autopilot clusters ```InstanceGroupManagers.List``` and ```InstanceGroupManagers.Get``` methods consistently serving errors in a single zone caused CA to functionally grind to a halt in a regional cluster. This PR modifies ```fillMigInfoCache()``` method in ```cloudprovider/gce/mig_info_provider.go``` to do the following changes as a fix.
- Still log all errors encountered
- Return an error if all zones returned errors, not when one zone did
- For each broken zone, iterate over all MIGs from ```registeredMigRefs``` and mark them missing by calling ```c.migLister.HandleMigIssue(migRef, err)```

Fixes #

NONE

```release-note

```

```docs

```


x13n:
Discussed offline. The public functions behavior doesn't really change here (they still fall back to direct API call), so not much we can test.

/lgtm
/approve

PR #5864, opened 2023-06-15
Record all vpa api versions in recommender metrics

kwiesmueller:
#### What type of PR is this?

/kind cleanup

Change the tracking of APIVersion from a boolean
indicating if the VPA is v1beta1
to the version string
and make sure it gets exported
in metrics.

Add tests for the recommender metrics.

```release-note
* VPA API Versions reported in the `vpa_objects_count` metric now include `v1` VPAs
```

```docs

```

/cc @jbartosik

### DIFF
--- a/vertical-pod-autoscaler/pkg/utils/metrics/recommender/recommender.go
+++ b/vertical-pod-autoscaler/pkg/utils/metrics/recommender/recommender.go
@@ -33,7 +33,14 @@ const (
 )
 
 var (
-	modes = []string{string(vpa_types.UpdateModeOff), string(vpa_types.UpdateModeInitial), string(vpa_types.UpdateModeRecreate), string(vpa_types.UpdateModeAuto)}
+	// TODO: unify this list with the types defined in the VPA handler to avoid
+	// drift if one file is changed and the other one is missed.
+	modes = []string{
+		string(vpa_types.UpdateModeOff),
+		string(vpa_types.UpdateModeInitial),
+		string(vpa_types.UpdateModeRecreate),
+		string(vpa_types.UpdateModeAuto),
+	}
 )
 
 type apiVersion string
@@ -150,20 +157,15 @@ func NewObjectCounter() *ObjectCounter {
 
 // Add updates the helper state to include the given VPA object
 func (oc *ObjectCounter) Add(vpa *model.Vpa) {
-	mode := string(vpa_types.UpdateModeAuto)
+	mode := vpa_types.UpdateModeAuto
 	if vpa.UpdateMode != nil && string(*vpa.UpdateMode) != "" {
-		mode = string(*vpa.UpdateMode)
-	}
-	// TODO: Maybe report v1 version as well.
-	api := v1beta2
-	if vpa.IsV1Beta1API {
-		api = v1beta1
+		mode = *vpa.UpdateMode
 	}
 
 	key := objectCounterKey{
-		mode:              mode,
+		mode:              string(mode),
 		has:               vpa.HasRecommendation(),
-		apiVersion:        api,
+		apiVersion:        apiVersion(vpa.APIVersion),
 		matchesPods:       vpa.HasMatchedPods(),
 		unsupportedConfig: vpa.Conditions.ConditionActive(vpa_types.ConfigUnsupported),
 	}
--- a/vertical-pod-autoscaler/pkg/utils/metrics/recommender/recommender_test.go
+++ b/vertical-pod-autoscaler/pkg/utils/metrics/recommender/recommender_test.go
@@ -1,0 +1,322 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package recommender
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+	apiv1 "k8s.io/api/core/v1"
+	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1"
+	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/recommender/model"
+)
+
+func TestObjectCounter(t *testing.T) {
+	updateModeOff := vpa_types.UpdateModeOff
+	updateModeInitial := vpa_types.UpdateModeInitial
+	updateModeRecreate := vpa_types.UpdateModeRecreate
+	updateModeAuto := vpa_types.UpdateModeAuto
+	// We verify that other update modes are handled correctly as validation
+	// may not happen if there are issues with the admission controller.
+	updateModeUserDefined := vpa_types.UpdateMode("userDefined")
+
+	cases := []struct {
+		name        string
+		add         []*model.Vpa
+		wantMetrics map[string]float64
+	}{
+		{
+			name: "set empty api on metric if it is missing on the VPA",
+			add: []*model.Vpa{
+				{
+					APIVersion: "",
+				},
+			},
+			wantMetrics: map[string]float64{
+				"api=,has_recommendation=false,matches_pods=true,unsupported_config=false,update_mode=Auto,": 1,
+			},
+		},
+		{
+			name: "report api version v1beta1",
+			add: []*model.Vpa{
+				{
+					APIVersion: "v1beta1",
+				},
+			},
+			wantMetrics: map[string]float64{
+				"api=v1beta1,has_recommendation=false,matches_pods=true,unsupported_config=false,update_mode=Auto,": 1,
+			},
+		},
+		{
+			name: "report api version v1beta2",
+			add: []*model.Vpa{
+				{
+					APIVersion: "v1beta2",
+				},
+			},
+			wantMetrics: map[string]float64{
+				"api=v1beta2,has_recommendation=false,matches_pods=true,unsupported_config=false,update_mode=Auto,": 1,
+			},
+		},
+		{
+			name: "report api version v1",
+			add: []*model.Vpa{
+				{
+					APIVersion: "v1",
+				},
+			},
+			wantMetrics: map[string]float64{
+				"api=v1,has_recommendation=false,matches_pods=true,unsupported_config=false,update_mode=Auto,": 1,
+			},
+		},
+		{
+			name: "default update mode to auto",
+			add: []*model.Vpa{
+				{
+					APIVersion: "v1",
+					UpdateMode: nil,
+				},
+			},
+			wantMetrics: map[string]float64{
+				"api=v1,has_recommendation=false,matches_pods=true,unsupported_config=false,update_mode=Auto,": 1,
+			},
+		},
+		{
+			name: "report update mode auto",
+			add: []*model.Vpa{
+				{
+					APIVersion: "v1",
+					UpdateMode: &updateModeAuto,
+				},
+			},
+			wantMetrics: map[string]float64{
+				"api=v1,has_recommendation=false,matches_pods=true,unsupported_config=false,update_mode=Auto,": 1,
+			},
+		},
+		{
+			name: "report update mode initial",
+			add: []*model.Vpa{
+				{
+					APIVersion: "v1",
+					UpdateMode: &updateModeInitial,
+				},
+			},
+			wantMetrics: map[string]float64{
+				"api=v1,has_recommendation=false,matches_pods=true,unsupported_config=false,update_mode=Initial,": 1,
+			},
+		},
+		{
+			name: "report update mode recreate",
+			add: []*model.Vpa{
+				{
+					APIVersion: "v1",
+					UpdateMode: &updateModeRecreate,
+				},
+			},
+			wantMetrics: map[string]float64{
+				"api=v1,has_recommendation=false,matches_pods=true,unsupported_config=false,update_mode=Recreate,": 1,
+			},
+		},
+		{
+			name: "report update mode off",
+			add: []*model.Vpa{
+				{
+					APIVersion: "v1",
+					UpdateMode: &updateModeOff,
+				},
+			},
+			wantMetrics: map[string]float64{
+				"api=v1,has_recommendation=false,matches_pods=true,unsupported_config=false,update_mode=Off,": 1,
+			},
+		},
+		{
+			name: "report update mode user defined",
+			add: []*model.Vpa{
+				{
+					APIVersion: "v1",
+					UpdateMode: &updateModeUserDefined,
+				},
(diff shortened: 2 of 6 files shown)


jbartosik (changes requested):
Thank you for the PR. Mostly it looks good but please don't change go.mod nor vendor.

voelzmo:
I'm not sure this works how it is supposed to work. I saw that VPA objects which I uploaded as `v1` and which use properties which only exist in `v1` (`controlledResources`) were recorded as `v1beta2` objects.

Here's what I tried:
* Built and deployed a vpa-recommender with this change
* deployed `hamster` with `k apply -f examples/hamster.yaml`
* forwarded `vpa-recommender`'s metrics port and looked at the object count
```
$ k -n kube-system port-forward vpa-recommender-b99974c95-jqc4p 9002:8942

$ curl -s http://localhost:9002/metrics | grep 'vpa_recommender_vpa_objects_count{api="v1",has_recommendation="true",matches_pods="true",unsupported_config="false",update_mode="Auto"}'
vpa_recommender_vpa_objects_count{api="v1",has_recommendation="true",matches_pods="true",unsupported_config="false",update_mode="Auto"} 0

$ curl -s http://localhost:9002/metrics | grep 'vpa_recommender_vpa_objects_count{api="v1beta2",has_recommendation="true",matches_pods="true",unsupported_config="false",update_mode="Auto"}'
vpa_recommender_vpa_objects_count{api="v1beta2",has_recommendation="true",matches_pods="true",unsupported_config="false",update_mode="Auto"} 1
```

This most likely means that we're falling back to the default of `v1beta2` during [recording the object count metric](https://github.com/kubernetes/autoscaler/blob/c7ddd5a63d587b59c2d0d221de31e2168e7595a2/vertical-pod-autoscaler/pkg/recommender/routines/recommender.go#L123), because the VPA objects here are taken from [`ClusterState.ObservedVPAs`](https://github.com/kubernetes/autoscaler/blob/c7ddd5a63d587b59c2d0d221de31e2168e7595a2/vertical-pod-autoscaler/pkg/recommender/routines/recommender.go#L84). Unfortunately, this is populated in  [`cluster_feeder#LoadVPAs`](https://github.com/kubernetes/autoscaler/blob/c7ddd5a63d587b59c2d0d221de31e2168e7595a2/vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go#L386) and the VPA objects added here don't get modified by [`cluster.AddOrUpdateVPA`](https://github.com/kubernetes/autoscaler/blob/ab695bc59e90ee38f3186fdb6f15fd2c65db13d1/vertical-pod-autoscaler/pkg/recommender/model/cluster.go#L282), as this changes `cluster.VPAs`.

On a side note, I was wondering if it is even possible to do what you're trying to achieve here. Doesn't the API version that the `VPALister` uses determine the value of the `APIVersion` field in the object you're getting? Do we have any information about how the object was initially submitted or is stored? Clients can get objects in the API version they want, right?

voelzmo (changes requested):
I'm not sure this works how it is supposed to work. I saw that VPA objects which I uploaded as `v1` and which use properties which only exist in `v1` (`controlledResources`) were recorded as `v1beta2` objects.

Here's what I tried:
* Built and deployed a vpa-recommender with this change
* deployed `hamster` with `k apply -f examples/hamster.yaml`
* forwarded `vpa-recommender`'s metrics port and looked at the object count
```
$ k -n kube-system port-forward vpa-recommender-b99974c95-jqc4p 9002:8942

$ curl -s http://localhost:9002/metrics | grep 'vpa_recommender_vpa_objects_count{api="v1",has_recommendation="true",matches_pods="true",unsupported_config="false",update_mode="Auto"}'
vpa_recommender_vpa_objects_count{api="v1",has_recommendation="true",matches_pods="true",unsupported_config="false",update_mode="Auto"} 0

$ curl -s http://localhost:9002/metrics | grep 'vpa_recommender_vpa_objects_count{api="v1beta2",has_recommendation="true",matches_pods="true",unsupported_config="false",update_mode="Auto"}'
vpa_recommender_vpa_objects_count{api="v1beta2",has_recommendation="true",matches_pods="true",unsupported_config="false",update_mode="Auto"} 1
```

This most likely means that we're falling back to the default of `v1beta2` during [recording the object count metric](https://github.com/kubernetes/autoscaler/blob/c7ddd5a63d587b59c2d0d221de31e2168e7595a2/vertical-pod-autoscaler/pkg/recommender/routines/recommender.go#L123), because the VPA objects here are taken from [`ClusterState.ObservedVPAs`](https://github.com/kubernetes/autoscaler/blob/c7ddd5a63d587b59c2d0d221de31e2168e7595a2/vertical-pod-autoscaler/pkg/recommender/routines/recommender.go#L84). Unfortunately, this is populated in  [`cluster_feeder#LoadVPAs`](https://github.com/kubernetes/autoscaler/blob/c7ddd5a63d587b59c2d0d221de31e2168e7595a2/vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go#L386) and the VPA objects added here don't get modified by [`cluster.AddOrUpdateVPA`](https://github.com/kubernetes/autoscaler/blob/ab695bc59e90ee38f3186fdb6f15fd2c65db13d1/vertical-pod-autoscaler/pkg/recommender/model/cluster.go#L282), as this changes `cluster.VPAs`.

On a side note, I was wondering if it is even possible to do what you're trying to achieve here. Doesn't the API version that the `VPALister` uses determine the value of the `APIVersion` field in the object you're getting? Do we have any information about how the object was initially submitted or is stored? Clients can get objects in the API version they want, right?
I would expect (as soon as the the above inconsistency is fixed) to see all VPA objects reported as `v1`, even if they were submitted as `v1beta2` objects.

kwiesmueller:
Did some more testing and it looks like the client wipes the `TypeMeta` before passing it to the cache.
This seems to be on purpose: https://github.com/kubernetes/kubernetes/pull/59264#issuecomment-362579495

So I don't think we're actually able to track the apiVersion of resources we process.

@jbartosik what do you think, should we just remove this logic all together?
I'm still looking for a way to track the apiVersions we have, but it's unclear what the use-case for these labels is in the first place.

voelzmo:
> which makes me wonder if this apiversion gate ever worked?

I think the only way this worked was by looking at some selector that only the `v1beta1` version had and otherwise just defaulting to `v1beta2`. When we switched internally from `v1beta2` to `v1` I was also looking for a way to gather information about how many people were still using the `v1beta2` API, but I gave up because I ended up in a simlar place like you just did. 🙈 

I agree this feature would be very useful, I'm just not sure how to implement this.

kwiesmueller:
I don't think we can. If someone cares to know what api versions of a resource they use there should be better ways (etcd/apiserver metrics etc.). So I'd recommend removing this path if it doesn't work.

The question is if we can just remove a label from the metric or if we should hardcode a version constant instead (to `v1` now) that matches the api client we use?

jbartosik (commented):
I'd rather keep the label; I think it will be useful if we want to introduce a new API version

kwiesmueller:
How would we track the new version? The client doesn't give us the value so the only way we seem to have is hardcoding the api label based on the client we use.

jbartosik (commented):
> How would we track the new version? The client doesn't give us the value so the only way we seem to have is hardcoding the api label based on the client we use.

Similar to how we did it with `v1beta1` / `v1beta2` distinction. I'd guess that if we're introducing new API version then we're making some significant changes to the API. We can use those to detect API version.

(I'm ok with droping the label too)

kwiesmueller:
Then let's keep the label for now, but hardcode it to the client version.
I couldn't find the logic that did the v1beta1 detection before, so I guess we add something when needed on the fly.

jbartosik (commented):
Please
- don't change vendor
- ping me when this is ready for another round of review

kwiesmueller:
@jbartosik updated to hardcode to the client apiVersion for now.
Also added more comments for background.

I don't know why vendor changed so much, but the go.mod change seems necessary as it's moving dependencies from indirect to direct because I rely on `github.com/prometheus/client_model` to test the metrics work.

Is that a problem?

vertical-pod-autoscaler/go.mod
```diff
@@ -19,6 +19,11 @@ require (
 	k8s.io/utils v0.0.0-20221107191617-1a15be271d1d
 )
 
+require (
```
jbartosik, line 22: Please don't change this.

vertical-pod-autoscaler/vendor/golang.org/x/net/http2/flow.go
```diff
@@ -6,31 +6,99 @@
```
jbartosik, line 6: Please don't change vendor

vertical-pod-autoscaler/pkg/utils/metrics/recommender/recommender_test.go
```diff
@@ -0,0 +1,310 @@
...
+
+			t.Cleanup(func() {
+				// Reset the metric after the test to avoid collisions.
+				vpaObjectCount.Reset()
+			})
+			counter.Observe()
+
+			metrics := make(chan prometheus.Metric)
+
+			go func() {
```
jbartosik, line 259: Why  run this as goroutine?

vertical-pod-autoscaler/pkg/utils/metrics/recommender/recommender_test.go
```diff
@@ -0,0 +1,310 @@
...
+	updateModeInitial := vpa_types.UpdateModeInitial
+	updateModeRecreate := vpa_types.UpdateModeRecreate
+	updateModeAuto := vpa_types.UpdateModeAuto
+	updateModeUserDefined := vpa_types.UpdateMode("userDefined")
+
+	cases := []struct {
+		name                   string
+		add                    []*model.Vpa
+		wantMetrics            map[string]float64
+		ignoreOtherZeroMetrics bool
```
jbartosik, line 41: We don't set `ignoreOtherZeroMetrics` to true anywhere; I'd drop this

vertical-pod-autoscaler/pkg/utils/metrics/recommender/recommender_test.go
```diff
@@ -0,0 +1,310 @@
...
+			}()
+
+			gotMetrics := make(map[string]float64)
+			for metric := range metrics {
+				var metricProto dto.Metric
+				if err := metric.Write(&metricProto); err != nil {
+					t.Errorf("failed to write metric: %v", err)
+				}
+
+				key := labelsToKey(metricProto.GetLabel())
```
jbartosik, line 271: I think it's better to compare `LabelPair`s directly rather than convert them to custom string representations

jbartosik (changes requested):

voelzmo (changes requested):
Thanks for the PR! Seems like an investment in the future, but I do appreciate the additional unit tests for the metrics!

I've added a few small comments for your consideration.

vertical-pod-autoscaler/pkg/utils/metrics/recommender/recommender_test.go
```diff
@@ -0,0 +1,310 @@
...
+	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1"
+	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/recommender/model"
+)
+
+func TestObjectCounter(t *testing.T) {
+	updateModeOff := vpa_types.UpdateModeOff
+	updateModeInitial := vpa_types.UpdateModeInitial
+	updateModeRecreate := vpa_types.UpdateModeRecreate
+	updateModeAuto := vpa_types.UpdateModeAuto
+	updateModeUserDefined := vpa_types.UpdateMode("userDefined")
```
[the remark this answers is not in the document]
kwiesmueller: Just from the Go code it looks like you can use arbitrary update modes in your VPA objects.
If that is enforced differently at the API level, I'm happy with dropping it.

As the list of supported update modes is kind of duplicated in this file I thought I'd rather test that the code handles arbitrary values for update modes correctly (like if recommender would see `InPlace` but not know about it yet (or if someone forgot to update this code)

vertical-pod-autoscaler/pkg/utils/metrics/recommender/recommender.go
```diff
@@ -150,18 +155,18 @@ func NewObjectCounter() *ObjectCounter {
...
+		mode = *vpa.UpdateMode
 	}
-	// TODO: Maybe report v1 version as well.
-	api := v1beta2
-	if vpa.IsV1Beta1API {
-		api = v1beta1
+
+	api := apiVersion(vpa_types.SchemeGroupVersion.Version)
+	if vpaVersion := vpa.APIVersion; vpaVersion != "" {
+		api = apiVersion(vpa.APIVersion)
```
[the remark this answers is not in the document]
kwiesmueller: I used it as a safeguard against empty values, but am good with dropping the defaulting here.

vertical-pod-autoscaler/pkg/utils/metrics/recommender/recommender_test.go
kwiesmueller: I agree. Do you have a proposal how to record the metrics and define the cases in a still readable way?
Using cmp we could probably store the gotMetrics and wantMetrics as lists instead, but creating the test cases will be much harder to read.

jbartosik:
I created PR to clean up vendor: #5904

kwiesmueller:
Removed the vendor change.

vertical-pod-autoscaler/pkg/utils/metrics/recommender/recommender_test.go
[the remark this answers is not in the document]
jbartosik: We have a validation [here](https://github.com/kubernetes/autoscaler/blob/8f83f7ec434eda955f176fcb8ad7153d90000477/vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/handler.go#L119C18-L119C37).

jbartosik (commented):
Mostly looks good

kwiesmueller:
@jbartosik @voelzmo  as we validate the update mode there, do you mind if we keep around to make sure we handle desync between the two files correctly?
Otherwise I'd prefer to expose the supported update modes somewhere and reuse them instead of relying on humans to not forget about adding new modes in two places.

jbartosik (approved):
> Joachim Marco Voelz as we validate the update mode there, do you mind if we keep around to make sure we handle desync between the two files correctly? Otherwise I'd prefer to expose the supported update modes somewhere and reuse them instead of relying on humans to not forget about adding new modes in two places.

Sounds good to me. I think it would be nice to:

- Add a comment explaining that we're validating this but a bad config might slip through (for example because admission controller was down when )
- Add a TODO or open an issue to deduplicate list of valid modes (if you plan to do that)

Those are small improvements, this looks good to me; If nothing happens here I'll merge it on Monday

jbartosik:
> > Joachim Marco Voelz as we validate the update mode there, do you mind if we keep around to make sure we handle desync between the two files correctly? Otherwise I'd prefer to expose the supported update modes somewhere and reuse them instead of relying on humans to not forget about adding new modes in two places.
> 
> Sounds good to me. I think it would be nice to:
> 
> * Add a comment explaining that we're validating this but a bad config might slip through (for example because admission controller was down when )
> * Add a TODO or open an issue to deduplicate list of valid modes (if you plan to do that)
> 
> Those are small improvements, this looks good to me; If nothing happens here I'll merge it on Monday

We got a #5914 where it seems that admission webhook didn't apply default mode so it looks like it's a thing that happens

kwiesmueller:
Added comments and a TODO.

Issue #5851, opened 2023-06-14
VPA release 0.14.0


kgolab:
Following instructions in
[RELEASE.md](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/RELEASE.md).

I'm planning to do the release now, as asked by jbartosik@. It's been a while since K8s 1.27 was releases and IIUC we wanted to release a new minor VPA version around new K8s minor release.

I can also see that golang version has been bumped a few times to address vulnerabilities so I guess it's good to prepare a release for that reason as well.

I don't see any pending PRs which should block the release. In case I missed something, I will simply redo the process.

kgolab:
The instructions say:
```
Update VPA version const
[ ] Wait for all VPA changes that will be in the release to merge.
[ ] Wait for [the end to end tests](https://k8s-testgrid.appspot.com/sig-autoscaling-vpa) to run with all VPA changes included.
[ ] Make sure the end to end VPA tests are green.
```

I can see the tests being green here https://k8s-testgrid.appspot.com/sig-autoscaling-vpa#Summary (there's a single flake from actuation suite) but it's really hard to verify what exactly was built.

The only thing I'm sure is that we're using golang 1.20.5 which means that #5434 was already included. Luckily this is the last merged PR so it seems we're good to go further.

I'll try to see if we can improve (or simply document) matching a test run with the latest included commit.

kgolab:
@voelzmo , are there any special requirements to build the release after your #5746?

It seems that the previous release @jbartosik did was prior to this change.
I'm trying to build the next one but I ran into errors as shown in the previous comment. Hopefully you can help here :-)

voelzmo:
Ah I think the problem is that I didn't specify a `builder` explicitly in the Makefile and therefore it is using the one you currently have: `docker-container`, which needs an `output` specified. On my machine, I'm using the standard `docker` builder, which doesn't have the option to specify and `output`. Here's how the whole things looks for me (when using dockerhub as a registry)
```
 ALL_ARCHITECTURES="amd64 arm64" REGISTRY=voelzmo TAG=v0.99.1-dev make release
rm -f recommender-amd64
rm -f recommender-arm64
docker build -t vpa-autoscaling-builder ../../builder
[+] Building 2.4s (9/9) FINISHED
 => [internal] load .dockerignore                                                                                                                                                                                                                                                   0.0s
 => => transferring context: 2B                                                                                                                                                                                                                                                     0.0s
[...]

PR #5746, opened 2023-05-09
Switch to docker buildx for cross-platform builds

voelzmo:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:
Switch to `docker buildx --platform` for building cross-platform images, such that the `Architecture` metadata is correctly set in the image manifest.

#### Which issue(s) this PR fixes:

Fixes #5667

#### Special notes for your reviewer:
Tested this change by running
` ALL_ARCHITECTURES="amd64 arm64" REGISTRY=voelzmo TAG=v0.99.0-dev make release`

inspected the image with 
`docker image inspect voelzmo/vpa-admission-controller-amd64:v0.99.0-dev` and `docker image inspect voelzmo/vpa-admission-controller-arm64:v0.99.0-dev`

and saw that the `Architecture` metadata in the manifest was `amd64` and `arm64`, respectively.

#### Does this PR introduce a user-facing change?

```release-note
VPA images are now built using `docker buildx` to correctly set the `Architecture` metadata in the image manifest.
```

PR #5867, opened 2023-06-16
Explicitly create and remove buildx builders for VPA components

voelzmo:
#### What type of PR is this?

/kind bug

Explicitly create a `builder` to be used with `docker buildx` and configure it for use during the build. Previously, the build would take whichever builder was currently configured on the machine, which could lead to inconsistencies with the builder's `output` behavior. See [the discussion in the v0.14.0 release issue](https://github.com/kubernetes/autoscaler/issues/5851#issuecomment-1594502338) for more context.
The builder created here is of type `docker-container` and [needs and explicit `output` configured](https://docs.docker.com/engine/reference/commandline/buildx_create/#driver) – the flag [`--load` is a shorthand for `--output=type=docker`](https://docs.docker.com/engine/reference/commandline/buildx_build/#load).

Additional make target were created, such that we can re-use the builder across the different platforms. Creating and tearing down a builder takes time, which we want to avoid doing too often.

```release-note
NONE
```

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/Makefile
+++ b/vertical-pod-autoscaler/pkg/admission-controller/Makefile
@@ -42,7 +42,7 @@ ifndef TAG
 	ERR = $(error TAG is undefined)
 	$(ERR)
 endif
-	docker buildx build --pull --platform linux/$* -t ${REGISTRY}/${FULL_COMPONENT}-$*:${TAG} --build-arg ARCH=$* .
+	docker buildx build --pull --load --platform linux/$* -t ${REGISTRY}/${FULL_COMPONENT}-$*:${TAG} --build-arg ARCH=$* .
 
 .PHONY: docker-push
 docker-push: $(addprefix sub-push-,$(ALL_ARCHITECTURES)) push-multi-arch;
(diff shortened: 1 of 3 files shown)


vertical-pod-autoscaler/pkg/admission-controller/Makefile
```diff
@@ -83,9 +83,16 @@ build-in-docker-%: clean docker-builder
 	echo '=============== bulding from the above ==============='
 	docker run -v `pwd`/../..:/gopath/src/k8s.io/autoscaler/vertical-pod-autoscaler vpa-autoscaling-builder:latest bash -c 'cd /gopath/src/k8s.io/autoscaler/vertical-pod-autoscaler && make build-binary-with-vendor-$* -C pkg/admission-controller'
 
+.PHONY: create-buildx-builder
+create-buildx-builder:
+	BUILDER=$(shell docker buildx create --use)
```
kgolab, line 88: Is this guaranteed to create a docker-container or should this be specified explicitly?

jbartosik (commented):
Release has been delayed already. I'd consider rolling back #5746 and merging the fix after we've released 0.14 (to finish the release quicker)

@kgolab @voelzmo WDYT?

kgolab:
> Release has been delayed already. I'd consider rolling back #5746 and merging the fix after we've released 0.14 (to finish the release quicker)
> 
> @kgolab @voelzmo WDYT?

I'd rather merge this PR - we'll be able to verify the release process in a new (mine) environment then. If we hit trouble again, let's revert - but I can see a lot of value in independent verification by full release process in a previously-unused env.

vertical-pod-autoscaler/pkg/admission-controller/Makefile
voelzmo: Good catch, I've added a commit to set the driver explicitly to `docker-container`!

voelzmo: Done!

PR #5875, opened 2023-06-20
Explicitly create and remove buildx builders

kgolab:
#### What type of PR is this?

/kind bug

Cherry-picking fix from #5867 to release branch so it can be built from a clean state. We have a VPA release in progress (#5851) and I'd like to make sure that we follow the instructions precisely:
> Create a fresh clone of the repo and switch to the vpa-release-0.${minor} branch. This makes sure you have no local changes while building the images.

I verified already that with this c-p included everything builds fine and the images are pushed to GCR (my in this case).


jbartosik (approved):
/lgtm

PR #5880, opened 2023-06-22
Bump default VPA version to 0.14.0

kgolab:
#### What type of PR is this?

/kind cleanup
/kind documentation

Actually it's part of the release process.

The first step of the [finalize release section](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/RELEASE.md#finalize-release).

The release process itself is tracked under #5851.


vertical-pod-autoscaler/README.md
```diff
@@ -48,12 +48,13 @@ procedure described below.
...
 # Installation
 
-The current default version is Vertical Pod Autoscaler 0.13.0
+The current default version is Vertical Pod Autoscaler 0.14.0
 
 ### Compatibility
 
 | VPA version     | Kubernetes version |
 |-----------------|--------------------|
+| 0.14            | 1.25+              |
```
jbartosik, line 57: Did we test with 1.25?

jbartosik (commented):
Mostly looks good to me, I have 1 question

vertical-pod-autoscaler/README.md
kgolab: Yes, following the procedure directly meant that I used GKE cluster with version 1.25.8-gke.1000 (both KCP & nodes).

jbartosik (approved):
/lgtm
/approve

PR #5883, opened 2023-06-22
Add dynamic node count limits for binpacking estimator

ystryuchkov:
#### What type of PR is this?

/kind feature

Introduce new binpacking limit interface for binpacking estimator / threshold based limiter.

Instead of static node cap, we can have array of binpacking limits, that can be set both at limiter construction time (static cap) or at estimation start time (dynamic cap). It will give us the ability to tune the performance of the binpacking in future.

This PR implements threshold-based binpacking limit, replacing static node count value.

#### Special notes for reviewers:

No changes to existing logic. I will open additional PR that will introduce functional changes.


cluster-autoscaler/estimator/binpacking_limit.go
```diff
@@ -0,0 +1,5 @@
+package estimator
+
+type BinpackingLimit interface {
```
BigDarkClown, line 5: I think this interface would be better if it exposed both the Node limit by `GetNodeLimit() int` and the Duration limit by `GetDurationLimit() time.Duration`. That way we would be able to implement some duration based limiting easily in the future.

cluster-autoscaler/estimator/binpacking_limit.go
```diff
@@ -0,0 +1,5 @@
...
+	GetLimit() int
```
BigDarkClown, line 6: `GetLimit` is a bit ambiguous, I think `GetNodeLimit` would better show the intent.

cluster-autoscaler/estimator/estimator.go
```diff
@@ -39,7 +39,7 @@ var AvailableEstimators = []string{BinpackingEstimatorName}
 // It returns the number of new nodes needed as well as the list of pods it managed
 // to schedule on those nodes.
 type Estimator interface {
-	Estimate([]*apiv1.Pod, *schedulerframework.NodeInfo, cloudprovider.NodeGroup) (int, []*apiv1.Pod)
+	Estimate([]*apiv1.Pod, *schedulerframework.NodeInfo, cloudprovider.NodeGroup, []BinpackingLimit) (int, []*apiv1.Pod)
```
BigDarkClown, line 42: I'm not 100% sold on the runtime limits. I understand that this will make some implementation inside the orchestrator easier, but couldn't we have all the limits defined at the initialisation? We can give them a way to access some necessary variables either way.

cluster-autoscaler/estimator/threshold_based_limiter.go
```diff
@@ -56,9 +60,24 @@ func (tbel *thresholdBasedEstimationLimiter) PermissionToAddNode() bool {
...
-		maxNodes:    maxNodes,
-		maxDuration: maxDuration,
+		maxNodes:        maxNodes,
+		maxNodesDefault: maxNodes,
+		maxDuration:     maxDuration,
+	}
+}
+
+func computeLimit(binpackingLimits []BinpackingLimit, currentLimit int) int {
+	if binpackingLimits == nil || len(binpackingLimits) == 0 {
```
BigDarkClown, line 73: nit: This if is not necessary, if the slice is empty then we will just return currentLimit. Fortunately golang handles slices being `nil` gracefully when iterating over them.

cluster-autoscaler/estimator/threshold_based_limiter.go
```diff
@@ -56,9 +60,24 @@ func (tbel *thresholdBasedEstimationLimiter) PermissionToAddNode() bool {
 // after either a node count- of time-based threshold is reached. This is meant to prevent cases
 // where binpacking of hundreds or thousands of nodes takes extremely long time rendering CA
 // incredibly slow or even completely crashing it.
-func NewThresholdBasedEstimationLimiter(maxNodes int, maxDuration time.Duration) EstimationLimiter {
+func NewThresholdBasedEstimationLimiter(limits []BinpackingLimit, maxDuration time.Duration) EstimationLimiter {
+	maxNodes := computeLimit(limits, 0)
```
BigDarkClown, line 64: This does not work if any of the limits passed as the beginning is not constant, meaning that the limit it returns can change after some time. I think it would be better to store the limits here and calculate the maxNodes/maxDuration after each startEstimation.

cluster-autoscaler/estimator/threshold_based_limiter.go
```diff
@@ -25,18 +25,22 @@ import (
...
+	maxNodesDefault int
+	nodes           int
+	start           time.Time
 }
 
-func (tbel *thresholdBasedEstimationLimiter) StartEstimation([]*apiv1.Pod, cloudprovider.NodeGroup) {
+func (tbel *thresholdBasedEstimationLimiter) StartEstimation(_ []*apiv1.Pod, _ cloudprovider.NodeGroup, runtimeLimits []BinpackingLimit) {
 	tbel.start = time.Now()
 	tbel.nodes = 0
+	tbel.maxNodes = computeLimit(runtimeLimits, tbel.maxNodes)
```
BigDarkClown, line 38: This also relies on the limits passed during initialisation returning constant values - we should also include them here. Plus this way we won't need the second argument in the `computeLimit` method.

cluster-autoscaler/estimator/threshold_based_limiter.go
```diff
@@ -56,9 +60,24 @@ func (tbel *thresholdBasedEstimationLimiter) PermissionToAddNode() bool {
...
 	return &thresholdBasedEstimationLimiter{
-		maxNodes:    maxNodes,
-		maxDuration: maxDuration,
+		maxNodes:        maxNodes,
+		maxNodesDefault: maxNodes,
+		maxDuration:     maxDuration,
+	}
+}
+
+func computeLimit(binpackingLimits []BinpackingLimit, currentLimit int) int {
```
BigDarkClown, line 72: This is ambiguous, should be change to `getNodeLimit`.

cluster-autoscaler/estimator/binpacking_limit.go
ystryuchkov: I wonder what could be the logic of applying multiple time limits. Could it be same as for the node limits, i.e. take the smallest limit from all defined (>0)?

Currently duration-based limit is set once in the threshold-based limiter constructor method:

```
func NewThresholdBasedEstimationLimiter(
  limits []BinpackingLimit,
  maxDuration time.Duration) EstimationLimiter {...}
}
```

cluster-autoscaler/estimator/binpacking_limit.go
ystryuchkov: Makes sense - I'm renaming it

cluster-autoscaler/estimator/threshold_based_limiter.go
ystryuchkov: Thank you - I haven't accounted for limits that do computation on each invocation

ystryuchkov: Fixed in 0d0196dbc6b41c145c34cfe91c3e3b769aabd9e9

cluster-autoscaler/estimator/threshold_based_limiter.go
ystryuchkov: Good for me to know now :-) Thank you

cluster-autoscaler/estimator/threshold_based_limiter.go
ystryuchkov: Renamed in 7654f75852e2f3bbc63be0a199f53faa38bc3895

cluster-autoscaler/estimator/threshold_based_limiter.go
ystryuchkov: Removed in 7654f75852e2f3bbc63be0a199f53faa38bc3895

cluster-autoscaler/estimator/binpacking_limit.go
BigDarkClown: Yes, I think we should take the shortest duration limit returned by the limiters. So with pseudo code

```
nodeLimit = min(limit.NodeLimit() for limit in estimator.limits) // excluding zeros
durationLimit = min(limit.DurationLimit() for limit in estimator.limits) // excluding zeros
```

This gives us an advantage of being able to define many duration limits if needed.

cluster-autoscaler/estimator/estimator.go
BigDarkClown: Ok, I am more in favour now. Otherwise we would need to pass many different arguments to the StartEstimation if we wanted to take into account values calculated during scale-up. Runtime limits SGTM then.

cluster-autoscaler/estimator/threshold_based_limiter.go
```diff
@@ -27,16 +27,20 @@ import (
...
 	nodes       int
 	start       time.Time
 }
 
-func (tbel *thresholdBasedEstimationLimiter) StartEstimation([]*apiv1.Pod, cloudprovider.NodeGroup) {
+func (tbel *thresholdBasedEstimationLimiter) StartEstimation(_ []*apiv1.Pod, _ cloudprovider.NodeGroup, runtimeLimits []BinpackingLimit) {
 	tbel.start = time.Now()
 	tbel.nodes = 0
+	tbel.maxNodes = getNodeLimit(tbel.limits, 0)
+	tbel.maxNodes = getNodeLimit(runtimeLimits, tbel.maxNodes)
```
BigDarkClown, line 39: nit: `tbel.maxNodes = getNodeLimit(append(tbel.limits, runtimeLimits...), 0)` also works AND it would allow us to drop the second argument of getNodeLimit as it would always be 0.

BigDarkClown: We can potentially if the array underneath the `tbel.Limits` silce is big enough. But slice is a pair of (array, length), so even if we do we won't change anything in the slice itself, as all the relevant values in the array are the same and the length does not change.

cluster-autoscaler/estimator/binpacking_limit.go
ystryuchkov: Thank you - its now implemented in 2742946b9ca375b584f7a2ef2eedd69b8a736fea

cluster-autoscaler/estimator/estimator.go
ystryuchkov: I'm keeping them then :-)

cluster-autoscaler/estimator/threshold_based_limiter.go
ystryuchkov: initializing maxNodes & maxDuration was refactored - 2742946b9ca375b584f7a2ef2eedd69b8a736fea

cluster-autoscaler/estimator/threshold_based_limiter.go
ystryuchkov: I have refactored it in 2742946b9ca375b584f7a2ef2eedd69b8a736fea

BigDarkClown:
Looks great to me! Great job on this change, looks really clean now.
\lgtm

ystryuchkov:
Hi @x13n, could you please review this PR?

cluster-autoscaler/core/autoscaler.go
```diff
@@ -115,7 +115,12 @@ func initializeDefaultOptions(opts *AutoscalerOptions) error {
 		opts.ExpanderStrategy = expanderStrategy
 	}
 	if opts.EstimatorBuilder == nil {
-		estimatorBuilder, err := estimator.NewEstimatorBuilder(opts.EstimatorName, estimator.NewThresholdBasedEstimationLimiter(opts.MaxNodesPerScaleUp, opts.MaxNodeGroupBinpackingDuration), estimator.NewDecreasingPodOrderer())
+		estimatorLimits := []estimator.EstimationLimit{
+			estimator.NewThresholdEstimationLimit(opts.MaxNodesPerScaleUp, opts.MaxNodeGroupBinpackingDuration)}
```
x13n, line 119: nit: Can you move `}` to the next line for consistent formatting?

cluster-autoscaler/estimator/estimator.go
```diff
@@ -63,7 +63,7 @@ func NewEstimatorBuilder(name string, limiter EstimationLimiter, orderer Estimat
 // scale-up is limited by external factors.
 type EstimationLimiter interface {
 	// StartEstimation is called at the start of estimation.
-	StartEstimation([]*apiv1.Pod, cloudprovider.NodeGroup)
+	StartEstimation([]*apiv1.Pod, cloudprovider.NodeGroup, []EstimationLimit)
```
x13n, line 66: So we used to have a generic `EstimationLimiter` interface with threshold-based implementation, but now we're also introducing another `EstimationLimit` interface that returns thresholds? It seems the new interface is really only compatible with `thresholdBasedEstimationLimiter` - or am I missing something?

x13n (changes requested):

PR #5886, opened 2023-06-23
Do not calculate usage for missing resource

dbenque:
#### What type of PR is this?

/kind bug

In case of missing resource (mem or cpu) in the input, a default 0 value was returned for the missing resource(s).
Also this code is limiting itself to CPU and Mem while we can handle any kind of resource.

Not sure if the case can be triggered with the classic metric server as source of metrics. Maybe CPU and MEM are always provided... I am using another source of data (datadog) and CPU and MEM might not be returned in some cases (testing failure modes). It is important to not propagate source failure by pushing default values in the history.

NONE


vertical-pod-autoscaler/pkg/recommender/input/metrics/metrics_client.go
```diff
@@ -114,14 +114,13 @@ func newContainerMetricsSnapshot(containerMetrics v1beta1.ContainerMetrics, podM
...
-	cpuMillicores := cpuQuantity.MilliValue()
-
-	memoryQuantity := containerUsage[k8sapiv1.ResourceMemory]
-	memoryBytes := memoryQuantity.Value()
-
-	return model.Resources{
-		model.ResourceCPU:    model.ResourceAmount(cpuMillicores),
-		model.ResourceMemory: model.ResourceAmount(memoryBytes),
+	result := map[model.ResourceName]model.ResourceAmount{}
+	for resourceName, quantity := range containerUsage {
```
jbartosik, line 118: We're going to drop all samples that aren't CPU or memory later and log an error (for example [here](https://github.com/kubernetes/autoscaler/blob/d17b1aaaec7fdb2c2afad18ce8c076f01e0f6f4b/vertical-pod-autoscaler/pkg/recommender/model/types.go#L94)), VPA doesn't provide recommendation for other resources. Please return only CPU or memory samples.

jbartosik (changes requested):
Thank you. This is very important code for proper operation of VPA so please add some unit tests:
- `calculateUsage` handling samples
- other functions correctly handling incomplete (only CPU, only memory, neither)

vertical-pod-autoscaler/pkg/recommender/input/metrics/metrics_client.go
dbenque: In our kubernetes clusters are now running a custom recommender that could have other resources than CPU and Mem.
At least admission and updater should be completely generic in terms of resource Name and Count.
Should we try to do the same in the recommender? I mean handle as much as possible a list of resource instead of a pair CPU + Mem ?

jbartosik: If you want to change recommender code to make it easier to reuse in custom recommenders I think it's a good idea.

But we need to do that in a way that doesn't cause a bunch of spurious error logs in the default recommender

dbenque:
I have added unittest for `calculateUsage`

> But we need to do that in a way that doesn't cause a bunch of spurious error logs in the default recommender

I have browsed the code from `calculateUsage` --> `newContainerMetricsSnapshot` --> `createContainerMetricsSnapshots` --> `GetContainersMetrics` and I don't see any logs being produce in case we have only mem or cpu.

When we add the sample to the cluster state this is done using a switch, so it should be ok:
https://github.com/kubernetes/autoscaler/blob/fe2e4c0d9902cfd17456bbea05e4d123e54207f7/vertical-pod-autoscaler/pkg/recommender/model/container.go#L216

So I am not sure where I should add more changes/checks. Can you point me at some code lines so that I get a better idea?

sophieliu15:
> I have added unittest for `calculateUsage`
> 
> > But we need to do that in a way that doesn't cause a bunch of spurious error logs in the default recommender
> 
> I have browsed the code from `calculateUsage` --> `newContainerMetricsSnapshot` --> `createContainerMetricsSnapshots` --> `GetContainersMetrics` and I don't see any logs being produce in case we have only mem or cpu.
> 
> When we add the sample to the cluster state this is done using a switch, so it should be ok:
> 
> https://github.com/kubernetes/autoscaler/blob/fe2e4c0d9902cfd17456bbea05e4d123e54207f7/vertical-pod-autoscaler/pkg/recommender/model/container.go#L216
> 
> So I am not sure where I should add more changes/checks. Can you point me at some code lines so that I get a better idea?

One way is do a search of `klog` and evaluate each instance to confirm whether they are issuing logs when metrics are not cpu and memory. The other way is to manually run VPA including your change with a custom metrics and inspect the logs to see if there are any spurious error logs  we do not expect. Doing combination of both is preferred but I am not sure if the second approach is feasible for your dev environment.

vertical-pod-autoscaler/pkg/recommender/input/metrics/metrics_client.go
```diff
@@ -114,14 +114,13 @@ func newContainerMetricsSnapshot(containerMetrics v1beta1.ContainerMetrics, podM
 }
 
 func calculateUsage(containerUsage k8sapiv1.ResourceList) model.Resources {
-	cpuQuantity := containerUsage[k8sapiv1.ResourceCPU]
-	cpuMillicores := cpuQuantity.MilliValue()
-
-	memoryQuantity := containerUsage[k8sapiv1.ResourceMemory]
-	memoryBytes := memoryQuantity.Value()
-
-	return model.Resources{
```
sophieliu15, line 123: I am curious what is the current behaviour when CPU or memory resources are missing? It seems many places assume both resources exist? Will this change impact other parts of the recommender where it assumes both CPU and memory resources will be present?

E.g., https://sourcegraph.com/github.com/kubernetes/autoscaler@d17b1aa/-/blob/vertical-pod-autoscaler/pkg/recommender/input/spec/spec_client.go?L118 and https://sourcegraph.com/github.com/kubernetes/autoscaler@d17b1aa/-/blob/vertical-pod-autoscaler/pkg/recommender/logic/recommender.go?L66

vertical-pod-autoscaler/pkg/recommender/input/metrics/metrics_client.go
sophieliu15: It seems previously this function return both CPU and memory. After this change, it might only return one of the resource or none of them. Is there any downstream consumers of this function that explicitly expect both CPU and memory to be present and will panic or have unexpected behaviour when only one or none of the resource is returned?

vertical-pod-autoscaler/pkg/recommender/input/metrics/metrics_client_test.go
```diff
@@ -44,3 +48,52 @@ func TestGetContainersMetricsReturnsResults(t *testing.T) {
 		assert.Contains(t, tc.getAllSnaps(), snap, "One of returned ContainerMetricsSnapshot is different then expected ")
 	}
 }
+
+func Test_calculateUsage(t *testing.T) {
```
sophieliu15, line 52: Is it possible to add a test case when resource is not CPU or memory?

Issue #5618, opened 2023-03-24
Cluster Autoscaler resets unneeded since time to 0s


wreed4:
**Which component are you using?**:

cluster-autoscaler

**What version of the component are you using?**:

1.21.3
Component version:

**What k8s version are you using (`kubectl version`)?**:

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.17", GitCommit:"953be8927218ec8067e1af2641e540238ffd7576", GitTreeState:"clean", BuildDate:"2023-03-01T02:23:41Z", GoVersion:"go1.19.6", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"22+", GitVersion:"v1.22.17-eks-48e63af", GitCommit:"47b89ea2caa1f7958bc6539d6865820c86b4bf60", GitTreeState:"clean", BuildDate:"2023-01-24T09:34:06Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}

</pre></details>

**What environment is this in?**:

aws, EKS

**What did you expect to happen?**:

unneeded since would continue to increase until the node either is removed or becomes needed

**What happened instead?**:

unneeded since dropped to 0 for many nodes at once (but not all) even though they were never determined to be needed per the logs, causing the `--scale-down-unneeded-time` timer to be reset.

**How to reproduce it (as minimally and precisely as possible)**:

I'm unsure what causes this, but I know we have a fairly high churn rate on our cluster, around 300 nodes, mostly default settings with CA, and I did see a `Watch` on replicasets close in the loop that this happened in.. if that matters.

**Anything else we need to know?**:

I'm happy to answer more questions, but I'm unsure what else to put here.  The logs are far too verbose to copy in entirety, but I'll say, this is the piece of code I'm at that I think might possibly be lying. 
```
// Update stores nodes along with a time at which they were found to be
// unneeded. Previously existing timestamps are preserved.
```
https://github.com/kubernetes/autoscaler/blob/2f1c895b2a15a34729611e0a37e112357bb7c863/cluster-autoscaler/core/scaledown/unneeded/nodes.go#L77

vadasambar:
A snippet of logs around the problematic behavior might be helpful. 

Here's my understanding (feel free to correct me) 
[...]

PR #5890, opened 2023-06-26
fix: setting maxEmptyBulkDelete, and maxScaleDownParallelism to be the same value

### LINKED ISSUE #5870: max-empty-bulk-delete, max-scale-down-parallelism are not respected and instead the min(max-empty-bulk-delete, max-scale-down-parallelism) is taken
Cluster Autoscaler 
1.25.6 
1.25.6

$ kubectl version

</pre></details>

AKS
If I am using parallel Drain, we respect the Parallelism flags instead of max-empty-bulk-delete, and vice versa
**What happened instead?**:
The total nodes deleted is the minimum of the two flags, not respecting either one of them if the other is set. 

set max-empty-bulk-delete flag to 10, and max-scale-down-parallelism at 100 or vice versa, and you can observe that it will delete

min(max-empty-bulk-delete, max-scale-down-parallelism)

This is directly caused by both the planner and the actuator filtering out scale-down candidates. 

Nodes to delete

https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-release-1.25/cluster-autoscaler/core/scaledown/legacy/wrapper.go#L88
 will filter our nodes to honor max-empty-bulk delete, 
but then the actuator again inside of the actuators StartDeletion method when it considers max-scale-down-parallelism https://github.com/kubernetes/autoscaler/blob/c7ddd5a63d587b59c2d0d221de31e2168e7595a2/cluster-autoscaler/core/scaledown/actuation/actuator.go#L128

This results in filtering the nodes twice and neither flag being respected and instead the minimum or smaller amount is reset. 

Another piece, that I still need to investigate more closely, but figured I could mention in case someone has more experience here. 

We have the NodeDeletionTracker that assumes that the nodes in the planner were marked as deleted, but since they don't get deleted always in the actuator, the bulk delete may assume they were and remove them. 

Still investigating this but essentially it results in nodes being removed from the unneeded map and only being deleted in groups of 10(whatever the default value is of the  maxScaleDownParallelism) and getting reset and removed from the unneeded nodes map just to get added again.

Bryce-Soghigian:
#### What type of PR is this?
/kind bug

Cluster Autoscaler is in a migrating state where we make two decisions to filter down deletion candidates. We filter candidates in the Planner using the legacy maxEmptyBulkDelete flag, and we also filter down nodes in the cropNodesForDeletion function in the actuator that uses the maxScaleDownParallelism value. 

This leads to the following scenario
1) Neither MaxEmptyBulkDelete or MaxScaleDownParallelism is respected and the minimum of the two is taken, so if we aim to Delete 1000 nodes with MaxEmptyBulkDelete, but MaxScaleDownParallelism is set to 10(its default) then we only delete 10 Nodes, and the 1000 nodes are removed from the unneeded nodes map, causing them to have to wait the full unneeded time again before being canidates for deletion, and then we will only delete 10 of them. This cycle continues 100 more times before we delete all 1000 nodes. 

Fixes #5870, #5618 

```release-note
setting the value of max-scale-down-parallelism to the value of max-empty-bulk-delete if one of the flags is not set. And Vice Versa. 
The max-empty-bulk-delete flag will be deprecated as of 1.29
```

```docs

```

### DIFF
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -232,6 +232,16 @@ var (
 	forceDaemonSets                   = flag.Bool("force-ds", false, "Blocks scale-up of node groups too small for all suitable Daemon Sets pods.")
 )
 
+func isFlagPassed(name string) bool {
+	found := false
+	flag.Visit(func(f *flag.Flag) {
+		if f.Name == name {
+			found = true
+		}
+	})
+	return found
+}
+
 func createAutoscalingOptions() config.AutoscalingOptions {
 	minCoresTotal, maxCoresTotal, err := parseMinMaxFlag(*coresTotal)
 	if err != nil {
@@ -252,6 +262,17 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 	if *maxDrainParallelismFlag > 1 && !*parallelDrain {
 		klog.Fatalf("Invalid configuration, could not use --max-drain-parallelism > 1 if --parallel-drain is false")
 	}
+
+	// in order to avoid inconsistent deletion thresholds for the legacy planner and the new actuator, the max-empty-bulk-delete,
+	// and max-scale-down-parallelism flags must be set to the same value.
+	if isFlagPassed("max-empty-bulk-delete") && !isFlagPassed("max-scale-down-parallelism") {
+		*maxScaleDownParallelismFlag = *maxEmptyBulkDeleteFlag
+		klog.Warning("The max-empty-bulk-delete flag will be deprecated in k8s version 1.29. Please use max-scale-down-parallelism instead.")
+		klog.Infof("Setting max-scale-down-parallelism to %d, based on the max-empty-bulk-delete value %d", *maxScaleDownParallelismFlag, *maxEmptyBulkDeleteFlag)
+	} else if !isFlagPassed("max-empty-bulk-delete") && isFlagPassed("max-scale-down-parallelism") {
+		*maxEmptyBulkDeleteFlag = *maxScaleDownParallelismFlag
+	}
+
 	return config.AutoscalingOptions{
 		NodeGroupDefaults: config.NodeGroupAutoscalingOptions{
 			ScaleDownUtilizationThreshold:    *scaleDownUtilizationThreshold,


cluster-autoscaler/main.go
```diff
@@ -252,6 +253,12 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 	if *maxDrainParallelismFlag > 1 && !*parallelDrain {
 		klog.Fatalf("Invalid configuration, could not use --max-drain-parallelism > 1 if --parallel-drain is false")
 	}
+
+	// In order to avoid inconsistent deletion thresholds for the legacy planner and the new actuator,
+	// we set the deletion threshold to the maximum of the two flags that are used to filter down deletion candidates.
+	maxDeletionSize := int(math.Max(float64(*maxEmptyBulkDeleteFlag), float64(*maxScaleDownParallelismFlag)))
+	*maxEmptyBulkDeleteFlag = maxDeletionSize
+	*maxScaleDownParallelismFlag = maxDeletionSize
```
Bryce-Soghigian, line 261: This whole piece of code should be removed when we properly deprecate maxEmptyBulkDelete

x13n:
/assign

I think this will be problematic if someone explicitly sets one of the flags to be less than the default value. Instead of using Max, it seems safer to check which flag was provided (e.g. [like this](https://stackoverflow.com/questions/35809252/check-if-flag-was-provided-in-go)).

Bryce-Soghigian:
We have 4 cases here. 


1. Max-empty-bulk-delete is set but max-scale-down-parallelism is not
2. max-empty-bulk-delete is not set but max-scale-down-parallelism is 
3. both are set. 
4. neither are set and we just use the default value 10, currently thats fine. 


In cases 1 and 2, the solution is easy, we just set the value equal to the value of the one we set. 

For 3, its an interesting case, because if I set max-empty-bulk-delete to 3, then max-scale-down-parallelism to 11, which should we respect? 


```go
// cases we have 3, 1. Mebd was set, but msdp was not, 2 msdp was set, but mebd was not, 3. both were set
	if isFlagPassed("max-empty-bulk-delete") && !isFlagPassed("max-scale-down-parallelism") {
		// Set maxScaleDownParallelismFlag = maxEmptyBulkDeleteFlag to avoid inconsistent deletion thresholds for the legacy planner and the new actuator.
		*maxScaleDownParallelismFlag = *maxEmptyBulkDeleteFlag
	} else if !isFlagPassed("max-empty-bulk-delete") && isFlagPassed("max-scale-down-parallelism") {
		// Set maxEmptyBulkDeleteFlag = maxScaleDownParallelismFlag to avoid inconsistent deletion thresholds for the legacy planner and the new actuator.
		*maxEmptyBulkDeleteFlag = *maxScaleDownParallelismFlag
	} else {
		maxDeletionSize := int(math.Max(float64(*maxEmptyBulkDeleteFlag), float64(*maxScaleDownParallelismFlag)))
		*maxEmptyBulkDeleteFlag = maxDeletionSize
		*maxScaleDownParallelismFlag = maxDeletionSize

	}
	```
	
	Maybe something like this could work, where we only set maxDeletionSize, if we have both flags and have a conflict. Or since we are trying to deprecate the maxEmptyBulkDeleteFlag, we just respect maxScaleDownParallelismFlag.

**Comment by x13n** [comment]
If both are set ideally we'd just exit the process, but that's can break people already using such setup, especially if this is meant to be cherrypicked. I think max is ok in that case, but I'd log the problem to limit confusion.

**Comment by Bryce-Soghigian** [comment]
@x13n  I am thinking we should merge this in to 1.25, 1.26 and 1.27, and maybe for 1.28(IE Master), but maybe we should have another followup PR that panics if both flags are set for 1.28 or 1.29. This panic can take two paths. 

1. Panic whenever max-empty-bulk-delete is set. Maybe i should have a followup pr changing the planner to use max-scale-down-parallelism rather than max-empty-bulk-delete for cropping nodes down, having a single source of truth for the deletion canidates number. 

2. Panic whenever we see both flags set, and still support setting max-empty-bulk-delete and keep this logic currently in the autoscaler.

**Comment by Bryce-Soghigian** [comment]
https://github.com/kubernetes/autoscaler/pull/5926/files something like this could be useful as well, after this pr we are on is merged. 


This way we only set the total number of nodes we plan on removing/deleting based on one flag. With the pr we are on, we will have that value set to the intended value always. I am thinking maybe either in 1.28 or 1.29, we enforce max-empty-bulk-delete should not be set. 


What is the best way to announce flag deprecations to the community?

File: cluster-autoscaler/main.go [hunk 2]
```diff
@@ -252,6 +263,23 @@ func createAutoscalingOptions() config.AutoscalingOptions {
...
+		*maxScaleDownParallelismFlag = *maxEmptyBulkDeleteFlag
+	} else if !isFlagPassed("max-empty-bulk-delete") && isFlagPassed("max-scale-down-parallelism") {
+		// Set maxEmptyBulkDeleteFlag = maxScaleDownParallelismFlag to avoid inconsistent deletion thresholds for the legacy planner and the new actuator.
+		*maxEmptyBulkDeleteFlag = *maxScaleDownParallelismFlag
+	} else {
+		maxDeletionSize := int(math.Max(float64(*maxEmptyBulkDeleteFlag), float64(*maxScaleDownParallelismFlag)))
+		*maxEmptyBulkDeleteFlag = maxDeletionSize
+		*maxScaleDownParallelismFlag = maxDeletionSize
+
+		klog.Warning("setting maxEmptyBulkDeleteFlag and maxScaleDownParallelismFlag to the max of the two flags, in order to avoid inconsistent deletion thresholds for the legacy planner and the new actuator.")
```
x13n, line 280: Couple of nits:
- I'd use flag names, rather than variable names in the log.
- Maybe log the values of flags and the resulting value being used? That should be more readable.
- Let's start with capital letter.

cluster-autoscaler/main.go
```diff
@@ -252,6 +263,23 @@ func createAutoscalingOptions() config.AutoscalingOptions {
...
+
+	// in order to avoid inconsistent deletion thresholds for the legacy planner and the new actuator, the max-empty-bulk-delete,
+	// and max-scale-down-parallelism flags must be set to the same value.
+	if isFlagPassed("max-empty-bulk-delete") && !isFlagPassed("max-scale-down-parallelism") {
+		// Set maxScaleDownParallelismFlag = maxEmptyBulkDeleteFlag to avoid inconsistent deletion thresholds for the legacy planner and the new actuator.
+		*maxScaleDownParallelismFlag = *maxEmptyBulkDeleteFlag
+	} else if !isFlagPassed("max-empty-bulk-delete") && isFlagPassed("max-scale-down-parallelism") {
+		// Set maxEmptyBulkDeleteFlag = maxScaleDownParallelismFlag to avoid inconsistent deletion thresholds for the legacy planner and the new actuator.
+		*maxEmptyBulkDeleteFlag = *maxScaleDownParallelismFlag
+	} else {
```
x13n, line 275: If both flags are passed, we shouldn't emit the warning. No new code should be needed in such case.

x13n (changes requested):

cluster-autoscaler/main.go
Bryce-Soghigian: I agree, I was worried if both flags are set, and they are set to different values, we will still have inconsistent deletion behavior. 


We should just remove the reference to max-empty-bulk-delete https://github.com/kubernetes/autoscaler/pull/5926/files similar to this. I can do this in a separate PR, but this would solve the problem of a user can misconfigure the autoscaler. 


For now, removed the else block, and put a warning message inside of the if statement for max-empty-bulk-delete is set.

cluster-autoscaler/main.go
Bryce-Soghigian: Addressed, but rather than logging here, i am logging in the max-empty-bulk-delete block

cluster-autoscaler/main.go
```diff
@@ -252,6 +262,17 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 	if *maxDrainParallelismFlag > 1 && !*parallelDrain {
 		klog.Fatalf("Invalid configuration, could not use --max-drain-parallelism > 1 if --parallel-drain is false")
 	}
+
+	// in order to avoid inconsistent deletion thresholds for the legacy planner and the new actuator, the max-empty-bulk-delete,
+	// and max-scale-down-parallelism flags must be set to the same value.
+	if isFlagPassed("max-empty-bulk-delete") && !isFlagPassed("max-scale-down-parallelism") {
+		*maxScaleDownParallelismFlag = *maxEmptyBulkDeleteFlag
+		klog.Warning("The max-empty-bulk-delete flag will be deprecated in k8s version 1.29. Please use max-scale-down-parallelism instead.")
```
x13n, line 270: nit: "will be deprecated in k8s version 1.29" -> "is deprecated and will be removed in k8s version 1.29"?

x13n (approved):
LGTM, just one minor comment on wording. Putting on hold, but feel free to just cancel the hold if you disagree.

/lgtm
/approve
/hold

yaohuatj:
cause default value was set for max-scale-down-parallelism, maybe the function `!isFlagPassed("max-scale-down-parallelism")"` won't work ?

yaohuatj:
> cause default value was set for max-scale-down-parallelism, maybe the function `!isFlagPassed("max-scale-down-parallelism")"` won't work ?

I have figure out, flag.Pause() is missing, so
isFlagPassed("max-empty-bulk-delete") = false  and isFlagPassed("max-scale-down-parallelism") = false

PR #5893, opened 2023-06-27
Sort nodegroups in order of their ID

kushagra98:
#### What type of PR is this?
/kind bug
/kind failing-test
/kind flake

The test `TestBinpackingLimiter` from #5810 was flaky. The test cloud provider iterates over a map to create a list of nodegroups. In this test, I'm trying to limit the expansion options but it turns out that the list is not ordered and hence the BinpackingLimiter stops generating options after processing 1 nodegroup.

How to reproduce the issue (before these changes):
```
func TestBinpackingLimiter1000(t *testing.T) {
    loop := 100
    for i := 1; i < loop; i++ {
	TestBinpackingLimiter(t)
    }
}
```

Fixes #

```release-note
None
```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/test/test_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/test/test_cloud_provider.go
@@ -18,6 +18,7 @@ package test
 
 import (
 	"fmt"
+	"sort"
 	"sync"
 
 	apiv1 "k8s.io/api/core/v1"
@@ -136,6 +137,10 @@ func (tcp *TestCloudProvider) NodeGroups() []cloudprovider.NodeGroup {
 	for _, group := range tcp.groups {
 		result = append(result, group)
 	}
+	// Sort the nodegroups in order of Id.
+	sort.Slice(result, func(i, j int) bool {
+		return result[i].Id() < result[j].Id()
+	})
 	return result
 }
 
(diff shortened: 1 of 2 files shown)


x13n:
I don't think we should expect the limiter to get a sorted list of node groups from cloud provider. If it needs sorted node groups, it should do the sorting itself. Otherwise you're going to require each cloud provider to sort them.

kushagra98:
> I don't think we should expect the limiter to get a sorted list of node groups from cloud provider. If it needs sorted node groups, it should do the sorting itself. Otherwise you're going to require each cloud provider to sort them.

The functionality doesn't require nodegroups to be in sorted order. In the tests, the binpacking limiter with stop binpacking after 1 expansion option (just to showcase how binpacking limiter works - instead of calculating other expansion options). However, I was also asserting for the expansion option apart from just the length of expansion options. I think just asserting for the length should suffice if changing the logic in the test cloud provider is not an option.

x13n:
That's actually an interesting question - do we want to sort the list of node groups to consider or is random the intended ordering? It goes way out of "deflaking unit tests" scope, so I don't think we should block on it though.

/lgtm
/approve

jayantjain93:
nit: Ideally we should be amending the commit when addressing the comments.

PR #5894, opened 2023-06-27
Include short unregistered nodes in calculation of incorrect node group

BigDarkClown:
#### What type of PR is this?

/kind bug

There are 2 mechanisms used to fix the incorrect node group size:
- Unregistered nodes are removed after the default time of 15 minutes.
- Node group is resized to the number of registered nodes if this number is smaller than the target size after the default time of 15 minutes.

Both these mechanisms use the same time threshold, so there can occur race conditions between them. For example in the following situation:

Node group with:
- 1 registered node
- 1 unregistered node
- target size of 2

In this instance it is better to remove the unregistered node rather than resizing the node group. The issue with the second approach is that we don't have a guarantee which node will be removed - it can be either the unregistered one (desired) or the registered one (undesired).

To prevent this the second logic takes the nodes unregistered for > 15 minutes into account when deciding if it should resize the node group. It however does not take the nodes unready for < 15 minutes, which can result in the second mechanism being triggered.

This change takes all the unregistered nodes into account when deciding on the trigger condition for the second mechanism. This should prevent the race condition I've described here from happening. 

```
Because of this change some cloud providers may trigger the "fixNodeGroupSize" 
logic more frequently in their clusters than they did before. This can happen if the 
scale-ups take a long time to create nodes and can result in the CA treating the 
node groups as incorrectly sized and scaling them down. This can be prevented 
by increasing the max node provisioning time parameter of Cluster Autoscaler.

You can check if your cluster is experiencing this issue by looking for the following log line:
"Decreasing size of <node group name>, expected=X current=Y delta=Z"
```

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -648,8 +648,9 @@ func (csr *ClusterStateRegistry) updateIncorrectNodeGroupSizes(currentTime time.
 			}
 			continue
 		}
-		if len(readiness.Registered) > acceptableRange.MaxNodes ||
-			len(readiness.Registered) < acceptableRange.MinNodes {
+		unregisteredNodes := len(readiness.Unregistered) + len(readiness.LongUnregistered)
+		if len(readiness.Registered) > acceptableRange.CurrentTarget ||
+			len(readiness.Registered) < acceptableRange.CurrentTarget-unregisteredNodes {
 			incorrect := IncorrectNodeGroupSize{
 				CurrentSize:   len(readiness.Registered),
 				ExpectedSize:  acceptableRange.CurrentTarget,
--- a/cluster-autoscaler/clusterstate/clusterstate_test.go
+++ b/cluster-autoscaler/clusterstate/clusterstate_test.go
@@ -1064,3 +1064,319 @@ func newBackoff() backoff.Backoff {
 	return backoff.NewIdBasedExponentialBackoff(5*time.Minute, /*InitialNodeGroupBackoffDuration*/
 		30*time.Minute /*MaxNodeGroupBackoffDuration*/, 3*time.Hour /*NodeGroupBackoffResetTimeout*/)
 }
+
+func TestUpdateAcceptableRanges(t *testing.T) {
+	testCases := []struct {
+		name string
+
+		targetSizes      map[string]int
+		readiness        map[string]Readiness
+		scaleUpRequests  map[string]*ScaleUpRequest
+		scaledDownGroups []string
+
+		wantAcceptableRanges map[string]AcceptableRange
+	}{
+		{
+			name: "No scale-ups/scale-downs",
+			targetSizes: map[string]int{
+				"ng1": 10,
+				"ng2": 20,
+			},
+			readiness: map[string]Readiness{
+				"ng1": {Ready: make([]string, 10)},
+				"ng2": {Ready: make([]string, 20)},
+			},
+			wantAcceptableRanges: map[string]AcceptableRange{
+				"ng1": {MinNodes: 10, MaxNodes: 10, CurrentTarget: 10},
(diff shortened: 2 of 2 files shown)


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -512,7 +512,7 @@ func (csr *ClusterStateRegistry) updateAcceptableRanges(targetSize map[string]in
 		size := targetSize[nodeGroup.Id()]
 		readiness := csr.perNodeGroupReadiness[nodeGroup.Id()]
 		result[nodeGroup.Id()] = AcceptableRange{
-			MinNodes:      size - len(readiness.LongUnregistered),
+			MinNodes:      size - len(readiness.Unregistered) - len(readiness.LongUnregistered),
```
MaciekPytel, line 515: It makes sense to subtract Unregistered nodes here, given that we have a separate (and arguably better) mechanism for tracking them. I'm wondering though - doesn't that make us double-count scale-ups?
We subtract scaleUpRequest.Increase below and I'd imagine in most cases the unregistered nodes we're subtracting here come from scale-up requests.
Wouldn't it make sense to remove the scaleUpRequest part below? Sure, we will get out of acceptableRange for the window of time between resizing nodeGroup and instances being created, but having a timeout on instances being created seems like the purpose of this entire mechanism.

WDYT?

BigDarkClown: I guess it makes sense. The mechanism triggers only if the diff between the target size and current size is both outside of range and constant for 15 minutes. So we should not really be worried about scale-up requests, as the unregistered nodes should come and become registered.

The only worry I have is that it will trigger this mechanism more often than it already is. I'm not 100% sure if that is desirable for everyone, but I agree that it would be a more correct behaviour.

BigDarkClown: Ok, I did another 180 turn.

I've left the scale-up / scale-down logic in calculating acceptable range. This is used by the `IsNodeGroupHealthy` method, and I don't think we want to change the behaviour there. I've removed the long unregistered nodes from here as they IMO should impact the node pool healthiness. 

I however including the unregistered nodes when calculating the incorrect node group size. This is used only for correcting the node group size after 15 minutes, so we should be good here.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -648,8 +647,9 @@ func (csr *ClusterStateRegistry) updateIncorrectNodeGroupSizes(currentTime time.
 			}
 			continue
 		}
-		if len(readiness.Registered) > acceptableRange.MaxNodes ||
-			len(readiness.Registered) < acceptableRange.MinNodes {
+		unregisteredNodes := len(readiness.Unregistered) + len(readiness.LongUnregistered)
+		if len(readiness.Registered) > acceptableRange.CurrentTarget ||
```
MaciekPytel, line 651: Why current target and not MaxNodes? It seem like MaxNodes is an obvious fit here, there are no changes to how you calculate it and I don't see why we shouldn't offset for scale-downs that we're aware of (tbh I wouldn't implement the whole acceptableRanges logic if we were starting from scratch, but why not use it if it's already there?

BigDarkClown: The idea is that we don't want the incorrect node group size to depend on the ongoing scale-ups and scale-downs. We want to only guard on the expected size - that is the criteria that we won't to depend on. Otherwise we will effectively double our max node provisioning period for this mechanism.

Also in all fairness - the mechanism for fixing situations when `current size > expected size` is not implemented yet. It's possible that it will require some slightly different edge conditions, but I think it's better to leave figuring them to the person implementing this mechanism.

MaciekPytel: ack

PR #5896, opened 2023-06-27
Replace `BuildTestContainer` with use of builder

kwiesmueller:
#### What type of PR is this?

/kind cleanup

Resolves a code TODO by replacing uses of `BuildTestContainer` with use of the builder directly.

Fixes code TODO

I was resolving the TODO and generally think it might have helped clean things up, but would like to get some feedback if we all agree this change makes sense.
I can see it both ways that having this helper made sense. If so we should just remove the TODO instead.

```release-note
NONE
```

```docs

```

/assign @jbartosik

### DIFF
--- a/vertical-pod-autoscaler/pkg/updater/logic/updater_test.go
+++ b/vertical-pod-autoscaler/pkg/updater/logic/updater_test.go
@@ -27,6 +27,7 @@ import (
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
 	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1"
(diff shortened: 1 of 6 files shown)


jbartosik:
/retest

[Test failure](https://github.com/kubernetes/autoscaler/actions/runs/5394097024/jobs/9794784139?pr=5896) seems unrelated to the PR:

```
W0627 20:22:16.992770   40599 clusterstate.go:444] AcceptableRanges have not been populated yet. Skip checking
--- FAIL: TestBinpackingLimiter (0.00s)
    common.go:404: 
        	Error Trace:	/home/runner/work/autoscaler/autoscaler/go/src/k8s.io/autoscaler/cluster-autoscaler/core/test/common.go:404
        	            				/home/runner/work/autoscaler/autoscaler/go/src/k8s.io/autoscaler/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go:173
        	            				/home/runner/work/autoscaler/autoscaler/go/src/k8s.io/autoscaler/cluster-autoscaler/core/scaleup/orchestrator/orchestrator_test.go:855
        	Error:      	did not find expansionOptionToChoose %+v
        	Test:       	TestBinpackingLimiter
        	Messages:   	&{GroupName:ng1 SizeChange:1}
    orchestrator_test.go:858: 
        	Error Trace:	/home/runner/work/autoscaler/autoscaler/go/src/k8s.io/autoscaler/cluster-autoscaler/core/scaleup/orchestrator/orchestrator_test.go:858
        	Error:      	Should be true
        	Test:       	TestBinpackingLimiter
```

jbartosik (approved):
Looks good to me

One idea I have is that we're calling `resource.MustParse` each time we call `WithMemRequest` and similar functions (in existing code too). What do you think about changing the builder to take a string input instead?

(this is a strict improvement so if you think it's better to merge as-is let's do that)

kwiesmueller:
If we always use a string for the quantity it makes sense to simplify, agreed.
Would do that in a follow-up.

kwiesmueller:
Retest didn't trigger, trying again.

/retest
/test

jbartosik (approved):
/lgtm
/approve

jbartosik:
> If we always use a string for the quantity it makes sense to simplify, agreed.

It seems so to me.

PR #5778, opened 2023-05-19
Bump golang version & release tag

atwamahmoud:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:
Bumps Golang version to `1.20.4` & pump release tag to `1.8.19`
#### Which issue(s) this PR fixes:

Fixes 
 - [GO-2023-1751](https://pkg.go.dev/vuln/GO-2023-1751)
 - [GO-2023-1752](https://pkg.go.dev/vuln/GO-2023-1752)
 - [GO-2023-1753](https://pkg.go.dev/vuln/GO-2023-1753)

#### Does this PR introduce a user-facing change?
```release-note
 NONE
```

PR #5899, opened 2023-06-28
Bump golang to the latest version (1.20.5)

kgolab:
#### What type of PR is this?

/king bug

Need to fix GO-2023-1840

The govulncheck reports that:
```
Standard library
    Found in: runtime@go1.20.3
    Fixed in: runtime@go1.20.5
```

I'm not bumping the tag since the one from #5778 was actually never released.


mwielgus (approved):
/lgtm
/approve

PR #5900, opened 2023-06-28
protect admission against empty keys

dbenque:

vertical-pod-autoscaler/admission-controller

#### What type of PR is this?

/kind bug

Using external recommender that was coming with a bug, we ended up with a VPA object that was presenting a recommendation for a resource that has an empty name:

```
 resources:
          limits:
            cpu: "1"
            memory: 3Gi
          requests:
            "": "0"              <== empty key
            cpu: "1"
            ephemeral-storage: 10Gi
            memory: 3Gi
```

In this PR we are proposing to sanitise the recommendation by removing any resource key that would be empty so that the admission is not pushing "garbage" to the pods.

Possible extension: instead of purging empty key, we could have a list of allowed resourceName given as input parameter of the admission. This would also allow us to purge unknown resource or the one coming with typo.

If a recommendation comes with a resource key that is empty, this resource will be ignored in the admission-controller.


vertical-pod-autoscaler/pkg/utils/test/test_utils.go
```diff
@@ -89,6 +90,16 @@ func Resources(cpu, mem string) apiv1.ResourceList {
 	return result
 }
 
+// AddResource add a resource to the given resource list
+func AddResource(rl apiv1.ResourceList, resourceName apiv1.ResourceName, value string) apiv1.ResourceList {
```
jbartosik, line 94: There's only one place we're using this function. Please move it to vertical-pod-autoscaler/pkg/utils/test/test_recommendation.go and unexport it (rename to `addResource`)

vertical-pod-autoscaler/pkg/utils/test/test_recommendation.go
```diff
@@ -55,6 +57,15 @@ func (b *recommendationBuilder) WithTarget(cpu, memory string) RecommendationBui
 	return &c
 }
 
+func (b *recommendationBuilder) WithResource(resource apiv1.ResourceName, value string) RecommendationBuilder {
```
jbartosik, line 60:
```suggestion
func (b *recommendationBuilder) WithTargetResource(resource apiv1.ResourceName, value string) RecommendationBuilder {
```

vertical-pod-autoscaler/pkg/utils/test/test_vpa.go
```diff
@@ -131,6 +133,12 @@ func (b *verticalPodAutoscalerBuilder) WithTarget(cpu, memory string) VerticalPo
 	return &c
 }
 
+func (b *verticalPodAutoscalerBuilder) WithResourceInTarget(resource core.ResourceName, value string) VerticalPodAutoscalerBuilder {
```
jbartosik, line 136: Let's call it the same as the function in vertical-pod-autoscaler/pkg/utils/test/test_recommendation.go

Also please add a TODO, it looks like `verticalPodAutoscalerBuilder` (here) and `recommendationBuilder` (in test_recommendation.go) could use some deduplication

jbartosik:
@dbenque mentioning in case you're still interested in contributing this :)

vertical-pod-autoscaler/pkg/utils/test/test_vpa.go
dbenque: done.

vertical-pod-autoscaler/pkg/utils/test/test_utils.go
dbenque: done

dbenque:
Closing this PR due to a bad rebase that makes the history not clear.

Taking all the relevant changes to a new clean branch:
https://github.com/kubernetes/autoscaler/pull/6062

vertical-pod-autoscaler/pkg/admission-controller/resource/pod/recommendation/recommendation_provider_test.go
```diff
@@ -311,8 +311,7 @@ func TestUpdateResourceRequests(t *testing.T) {
 					return
 				}
 
-				_, foundEmpty := resources[0].Requests[""]
-				assert.Equal(t, foundEmpty, false, "empty resourceKey have not been purged")
+				assert.Contains(t, resources, "", "expected empty resource to be removed")
```
jbartosik, line 314: I know this is closed but isn't this check kind of opposite of the old one?

PR #5901, opened 2023-06-28
Remove unneeded node groups regardless of scale down being in cooldown.

olagacek:
#### What type of PR is this?

/kind feature

Currently unneeded node groups are deleted only if scale down isn't in cooldown, so in particular if there was a scale up in a given loop, no unneeded node group will get deleted. There are certain scenarios in which that might turn out problematic - eg if scale up fails in every loop, the node groups won't be deleted for a long time. At the same time, unneeded node groups might affect scale up failures (e.g. quota issues).

Due to that, we can decouple unneeded node group deletion from the rest of scale down logic. 

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -621,22 +621,26 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
 			a.processorCallbacks.disableScaleDownForLoop, scaleDownInCooldown)
 		metrics.UpdateScaleDownInCooldown(scaleDownInCooldown)
 
+		// We want to delete unneeded Node Groups only if here is no current delete
+		// in progress.
+		_, drained := scaleDownActuationStatus.DeletionsInProgress()
+		var removedNodeGroups []cloudprovider.NodeGroup
+		if len(drained) == 0 {
+			var err error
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -621,23 +621,27 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
...
+		// in progress.
+		_, drained := scaleDownActuationStatus.DeletionsInProgress()
+		var removedNodeGroups []cloudprovider.NodeGroup
+		if len(drained) == 0 {
+			var err error
+			removedNodeGroups, err = a.processors.NodeGroupManager.RemoveUnneededNodeGroups(autoscalingContext)
+			if err != nil {
+				klog.Errorf("Error while removing unneeded node groups: %v", err)
+			}
+			scaleDownStatus.RemovedNodeGroups = removedNodeGroups
```
x13n, line 634: I think this makes line 653 no-op now.

x13n (commented):
/assign

Looks good, just one comment.

cluster-autoscaler/core/static_autoscaler.go
olagacek: I don't think so, as in line 648 we are overwriting `scaleDownStatus`

x13n: Ah, good point. This function is way too long :disappointed:

PR #5917, opened 2023-06-30
Implement configurable thresholds for threshold based estimator

ystryuchkov:
#### What type of PR is this?

/kind feature

Introduce new binpacking limit (`Threshold`) interface for threshold based limiter.

Instead of static node cap, we can have an array of thresholds, that compute estimation limits based on parameters set both at construction and runtime. It will give us the ability to tune the performance of the binpacking in future.

This PR implements 3 thresholds:

1. Static maximum node & maximum duration
2. Static cluster-wide node number limit
3. Similar node groups capacity threshold

It is alternative implementation of #5883, but having orchestrator populating `EstimationContext` instead of constructing limits/thresholds, which gives us ability to inject thresholds outside of orchestrator if context has enough data for them.

@BigDarkClown @x13n

### DIFF
--- a/cluster-autoscaler/estimator/cluster_capacity_threshold.go
+++ b/cluster-autoscaler/estimator/cluster_capacity_threshold.go
@@ -25,22 +25,28 @@ import (
 type clusterCapacityThreshold struct {
 }
 
-// GetNodeLimit returns available capacity of the cluster. Return value of 0 means that no limit is set.
-func (l *clusterCapacityThreshold) GetNodeLimit(_ cloudprovider.NodeGroup, context *EstimationContext) int {
-	if context == nil || (context.clusterMaxNodeLimit <= 0) || (context.clusterMaxNodeLimit < context.currentNodeCount) {
+// NodeLimit returns maximum number of new nodes that can be added to the cluster
+// based on its capacity. Possible return values are:
+//   - -1 when cluster has no available capacity
+//   - 0 when context or cluster-wide node limit is not set. Return value of 0 means that there is no limit.
+//   - Any positive number representing maximum possible number of new nodes
+func (l *clusterCapacityThreshold) NodeLimit(_ cloudprovider.NodeGroup, context EstimationContext) int {
+	if context == nil || context.ClusterMaxNodeLimit() == 0 {
 		return 0
 	}
-	return context.clusterMaxNodeLimit - context.currentNodeCount
+	if (context.ClusterMaxNodeLimit() < 0) || (context.ClusterMaxNodeLimit() <= context.CurrentNodeCount()) {
+		return -1
+	}
+	return context.ClusterMaxNodeLimit() - context.CurrentNodeCount()
 }
 
-// GetDurationLimit always returns 0 for this threshold, i.e. no limit is set.
-func (l *clusterCapacityThreshold) GetDurationLimit() time.Duration {
+// DurationLimit always returns 0 for this threshold, meaning that no limit is set.
+func (l *clusterCapacityThreshold) DurationLimit(cloudprovider.NodeGroup, EstimationContext) time.Duration {
 	return 0
 }
 
-// NewClusterCapacityThreshold returns a variable Threshold to limit
-// maximum nodes in binpacking. Value of the limit depends on maximum number of nodes
-// allowed and total number of nodes currently running in the cluster
+// NewClusterCapacityThreshold returns a Threshold that can be used to limit binpacking
+// by available cluster capacity
 func NewClusterCapacityThreshold() Threshold {
 	return &clusterCapacityThreshold{}
 }
--- a/cluster-autoscaler/estimator/cluster_capacity_threshold.go
+++ b/cluster-autoscaler/estimator/cluster_capacity_threshold.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package estimator
+
+import (
+	"time"
+
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
+)
+
+type clusterCapacityThreshold struct {
+}
+
+// GetNodeLimit returns available capacity of the cluster. Return value of 0 means that no limit is set.
+func (l *clusterCapacityThreshold) GetNodeLimit(_ cloudprovider.NodeGroup, context *EstimationContext) int {
+	if context == nil || (context.clusterMaxNodeLimit <= 0) || (context.clusterMaxNodeLimit < context.currentNodeCount) {
+		return 0
+	}
+	return context.clusterMaxNodeLimit - context.currentNodeCount
+}
+
+// GetDurationLimit always returns 0 for this threshold, i.e. no limit is set.
+func (l *clusterCapacityThreshold) GetDurationLimit() time.Duration {
+	return 0
+}
+
+// NewClusterCapacityThreshold returns a variable Threshold to limit
+// maximum nodes in binpacking. Value of the limit depends on maximum number of nodes
+// allowed and total number of nodes currently running in the cluster
+func NewClusterCapacityThreshold() Threshold {
+	return &clusterCapacityThreshold{}
+}
--- a/cluster-autoscaler/estimator/cluster_capacity_threshold_test.go
+++ b/cluster-autoscaler/estimator/cluster_capacity_threshold_test.go
@@ -30,33 +30,39 @@ func TestNewClusterCapacityThreshold(t *testing.T) {
 		contextCurrentNodes int
 	}{
 		{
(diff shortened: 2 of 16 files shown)


cluster-autoscaler/estimator/threshold.go
```diff
@@ -0,0 +1,30 @@
...
+	"time"
+
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
+)
+
+// Threshold provides resources configuration for threshold based estimation limiter.
+// Return value of 0 means that no limit is set.
+type Threshold interface {
+	GetNodeLimit(cloudprovider.NodeGroup, *EstimationContext) int
+	GetDurationLimit() time.Duration
```
BigDarkClown, line 29: Why `GetDurationLimit` does not get the same `cloudprovider.NodeGroup, *EstimationContext` arguments? Currently there is no implementation that needs them, but it seems reasonable to include them as well

cluster-autoscaler/estimator/sng_capacity_threshold.go
```diff
@@ -0,0 +1,54 @@
...
+	"time"
+
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
+	"k8s.io/klog/v2"
+)
+
+type sngCapacityThreshold struct {
+}
+
+func (l *sngCapacityThreshold) GetNodeLimit(_ cloudprovider.NodeGroup, context *EstimationContext) int {
```
BigDarkClown, line 29: This limit should include the `targetSize - currentSize` of the "main" node group as well.

cluster-autoscaler/estimator/cluster_capacity_threshold_test.go
```diff
@@ -0,0 +1,62 @@
...
+			contextCurrentNodes: 5,
+		},
+		{
+			name:                "for not defined cluster max nodes returns 0",
+			wantThreshold:       0,
+			contextMaxNodes:     0,
+			contextCurrentNodes: 10,
+		},
+		{
+			name:                "for negative total capacity returns 0",
```
BigDarkClown, line 45: I think in this case the 0 should mean "You cannot add any nodes at all". This is a bit problematic with the current definition of 0. Maybe we can change the semantic so that "You can add however many nodes you want" would be represented by an arbitrary large number, eg. 10^9 for number of nodes and time.Hour for duration?

cluster-autoscaler/estimator/cluster_capacity_threshold.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
+)
+
+type clusterCapacityThreshold struct {
+}
+
+// GetNodeLimit returns available capacity of the cluster. Return value of 0 means that no limit is set.
+func (l *clusterCapacityThreshold) GetNodeLimit(_ cloudprovider.NodeGroup, context *EstimationContext) int {
+	if context == nil || (context.clusterMaxNodeLimit <= 0) || (context.clusterMaxNodeLimit < context.currentNodeCount) {
```
BigDarkClown, line 30: Please use getters instead of referencing the value.

cluster-autoscaler/estimator/cluster_capacity_threshold.go
```diff
@@ -0,0 +1,46 @@
...
+		return 0
+	}
+	return context.clusterMaxNodeLimit - context.currentNodeCount
+}
+
+// GetDurationLimit always returns 0 for this threshold, i.e. no limit is set.
+func (l *clusterCapacityThreshold) GetDurationLimit() time.Duration {
```
BigDarkClown, line 37: nit: getter

cluster-autoscaler/estimator/estimation_context.go
```diff
@@ -0,0 +1,68 @@
...
+// This patch is used to update existing context.
+func NewEstimationContextUpdate(similarNodeGroups []cloudprovider.NodeGroup, currentNodeCount int) *EstimationContext {
+	return &EstimationContext{
+		similarNodeGroups: similarNodeGroups,
+		currentNodeCount:  currentNodeCount,
+	}
+}
+
+// GetSimilarNodeGroups returns array of similar node groups
+func (c *EstimationContext) GetSimilarNodeGroups() []cloudprovider.NodeGroup {
```
BigDarkClown, line 45: nit: getter

cluster-autoscaler/estimator/cluster_capacity_threshold_test.go
ystryuchkov: Maybe we can have 3 states of thresholds and adjust thresholdBasedEstimationLimiter for them:
- negative: cannot add any nodes if at least one threshold is negative
- zero: can add as many nodes as you like, thresholds with return value 0 are ignored for current estimation
- positive: add no more than this number of nodes, it is the smallest possible value from all positive thresholds

What do you think?

BigDarkClown: I'm not a fan of encoding such things into numerical values. But also I see that the "0 means unlimited" concept is pretty well established here.

I think this should work as long as we make sure that nodegroups that are not able to scale-up even a single node are not passed to the estimator. This is a pretty strong assumption though, I would add a comment about this in the Estimator call site.

cluster-autoscaler/estimator/threshold.go
ystryuchkov: Amended in f3dfeeeb47a021ca4b854be5363e1a1663313b3b

cluster-autoscaler/estimator/sng_capacity_threshold.go
ystryuchkov: Threshold now includes capacity of current node group in addition to capacity of similar node groups in f3dfeeeb47a021ca4b854be5363e1a1663313b3b

cluster-autoscaler/estimator/cluster_capacity_threshold_test.go
ystryuchkov: Now thresholds may return negative value and its is handled by thresholdBasedEstimationLimiter - f3dfeeeb47a021ca4b854be5363e1a1663313b3b

cluster-autoscaler/estimator/cluster_capacity_threshold.go
ystryuchkov: Replaced fields with getters - f3dfeeeb47a021ca4b854be5363e1a1663313b3b

cluster-autoscaler/estimator/cluster_capacity_threshold.go
```diff
@@ -0,0 +1,46 @@
...
+	"time"
+
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
+)
+
+type clusterCapacityThreshold struct {
+}
+
+// GetNodeLimit returns available capacity of the cluster. Return value of 0 means that no limit is set.
+func (l *clusterCapacityThreshold) GetNodeLimit(_ cloudprovider.NodeGroup, context *EstimationContext) int {
```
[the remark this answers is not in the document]
ystryuchkov: Renamed in f3dfeeeb47a021ca4b854be5363e1a1663313b3b

cluster-autoscaler/estimator/cluster_capacity_threshold.go
ystryuchkov: Renamed in f3dfeeeb47a021ca4b854be5363e1a1663313b3b

cluster-autoscaler/estimator/estimation_context.go
ystryuchkov: Renamed in f3dfeeeb47a021ca4b854be5363e1a1663313b3b

cluster-autoscaler/estimator/estimation_context.go
```diff
@@ -0,0 +1,68 @@
...
+	return c.similarNodeGroups
+}
+
+// GetClusterMaxNodeLimit returns maximum node number allowed for the cluster
+func (c *EstimationContext) GetClusterMaxNodeLimit() int {
```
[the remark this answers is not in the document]
ystryuchkov: Renamed in f3dfeeeb47a021ca4b854be5363e1a1663313b3b

cluster-autoscaler/estimator/estimation_context.go
```diff
@@ -0,0 +1,68 @@
...
+	return c.clusterMaxNodeLimit
+}
+
+// GetCurrentNodeCount returns current number of nodes in the cluster
+func (c *EstimationContext) GetCurrentNodeCount() int {
```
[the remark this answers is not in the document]
ystryuchkov: Renamed in f3dfeeeb47a021ca4b854be5363e1a1663313b3b

cluster-autoscaler/estimator/threshold_based_limiter.go
```diff
@@ -29,17 +29,34 @@ type thresholdBasedEstimationLimiter struct {
...
+		return targetLimit
+	}
+	return baseLimit
 }
 
 func (*thresholdBasedEstimationLimiter) EndEstimation() {}
 
 func (tbel *thresholdBasedEstimationLimiter) PermissionToAddNode() bool {
-	if tbel.maxNodes > 0 && tbel.nodes >= tbel.maxNodes {
+	if tbel.maxNodes < 0 || (tbel.maxNodes > 0 && tbel.nodes >= tbel.maxNodes) {
```
BigDarkClown, line 59: We should add the same check for duration below, to include negative durations.

ystryuchkov: Corrected in b4213d82444eab039ba1fe0b6a0dbaf642deadc3

PR #5936, opened 2023-07-06
fix(cloudprovider/externalgrpc): `GetOptions` returns `MaxNodeProvisionTime` as 0 seconds

vadasambar:
Signed-off-by: vadasambar <surajrbanakar@gmail.com>

#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:
* Adds missing `MaxNodeProvisionTime` in externalgrpc proto
* Sets default `MaxNodeProvisionTime` for nodegroup's `GetOptions` function
* Check #5935 for more info
#### Which issue(s) this PR fixes:

Fixes #5935

#### Does this PR introduce a user-facing change?

```release-note
Fixed: cluster-autoscaler thinks newly scaled up nodegroup using externalgrpc provider has `MaxNodeProvisionTime` set as 0 seconds and expects the new node to be registered in 0-10 seconds instead of the default 15m. Check https://github.com/kubernetes/autoscaler/issues/5935 for more info.  
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #5937, opened 2023-07-08
fix(cloudprovider/exteranlgrpc): properly handle unimplemented methods

dbonfigli:
#### What type of PR is this?

/kind bug
/kind feature

Some `CloudProvider` and `nodeGroup` methods are optional: if a cloud provider does not implemented an optional method, it should return the `cloudprovider.ErrNotImplemented` error, this error is expected and the CA gracefully handles it.

Currently there is no way to pass this specific error from a grpc based cloud provider: The protobuf definition incorrectly states that an rpc is optional or a grpc error will result in using defaults, but this is not correct at the moment: in case of grpc errors, the client passes a generic error that is not correctly handled by the CA, so currently all rpc must be implemented.

This PR allows grpc cloud providers to not implement optional methods: a not implemented method should pass the grpc error code 12 `Unimplemented`, if so, the CA grpc client will transform this to the go `cloudprovider.ErrNotImplemented` error that is then correctly handled.

Fixes #

```release-note
externalgrpc cloud provider: grpc based cloud providers can now pass the grpc error code 12, Unimplemented, to signal they do not implement optional methods.
```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/externalgrpc/externalgrpc_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/externalgrpc/externalgrpc_cloud_provider.go
@@ -27,7 +27,9 @@ import (
 	"time"
 
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/status"
 	"gopkg.in/yaml.v2"
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -158,6 +160,10 @@ func (m *pricingModel) NodePrice(node *apiv1.Node, startTime time.Time, endTime 
 		EndTime:   &end,
 	})
 	if err != nil {
+		st, ok := status.FromError(err)
+		if ok && st.Code() == codes.Unimplemented {
+			return 0, cloudprovider.ErrNotImplemented
+		}
 		klog.V(1).Infof("Error on gRPC call PricingNodePrice: %v", err)
 		return 0, err
 	}
@@ -178,6 +184,10 @@ func (m *pricingModel) PodPrice(pod *apiv1.Pod, startTime time.Time, endTime tim
 		EndTime:   &end,
 	})
 	if err != nil {
+		st, ok := status.FromError(err)
+		if ok && st.Code() == codes.Unimplemented {
+			return 0, cloudprovider.ErrNotImplemented
+		}
 		klog.V(1).Infof("Error on gRPC call PricingPodPrice: %v", err)
 		return 0, err
 	}
--- a/cluster-autoscaler/cloudprovider/externalgrpc/protos/externalgrpc.proto
+++ b/cluster-autoscaler/cloudprovider/externalgrpc/protos/externalgrpc.proto
@@ -41,13 +41,13 @@ service CloudProvider {
 
   // PricingNodePrice returns a theoretical minimum price of running a node for
   // a given period of time on a perfectly matching machine.
-  // Implementation optional.
+  // Implementation optional: if unimplemented return error code 12 (for `Unimplemented`)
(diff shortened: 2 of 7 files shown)


cluster-autoscaler/cloudprovider/externalgrpc/externalgrpc_cloud_provider.go
```diff
@@ -178,6 +184,10 @@ func (m *pricingModel) PodPrice(pod *apiv1.Pod, startTime time.Time, endTime tim
 		EndTime:   &end,
 	})
 	if err != nil {
+		st, ok := status.FromError(err)
+		if ok && st.Code() == codes.Unimplemented {
+			return 0, cloudprovider.ErrNotImplemented
```
Bryce-Soghigian, line 189: why zero for the price? I mean we raise the error anyway sso it doesn't matter, but zero could potentially be a real price. Maybe -1 makes more sense? To further raise eyebrows this isn't successful data from the pricing model?

dbonfigli: > why zero for the price? 
no particular reason, it is a standard practice to return the default value for the type when the value does not matter
> I mean we raise the error anyway sso it doesn't matter
exactly
> but zero could potentially be a real price
also -1, and it is a even better price than 0 if it is used by mistake :wink: .
I think the point is that no matter the return value, errors should be handled, and the cluster autoscaler core does it, not doing that is a logic error by the caller.

dbonfigli:
keeping this on draft until https://github.com/kubernetes/autoscaler/pull/5936 is merged so to cleanly regenerate the code for the protobuf.

dbonfigli:
after merging https://github.com/kubernetes/autoscaler/pull/5936, this is now ready to be reviewed, tagging @vadasambar as you expressed interest on this part of the code and i am the only one listed (informally, commented out) on the OWNERS file for this cloud provider.

cluster-autoscaler/cloudprovider/externalgrpc/protos/externalgrpc.proto
```diff
@@ -41,13 +41,13 @@ service CloudProvider {
 
   // PricingNodePrice returns a theoretical minimum price of running a node for
   // a given period of time on a perfectly matching machine.
-  // Implementation optional.
+  // Implementation optional: if so should return error code 12 Unimplemented
```
vadasambar, line 44: nit: the wording feels a little ambiguous. Here's a suggestion:  
```suggestion
  // Implementation optional: if unimplemented return error code 12 (for `Unimplemented`) 
```

vadasambar:
Overall looks good. I will do another pass and approve it.

cluster-autoscaler/cloudprovider/externalgrpc/protos/externalgrpc.proto
dbonfigli: ok, suggestion accepted

vadasambar (approved):

dbonfigli:
after the lgtm label , I am assigning this to @towca for the approval.

/assign @towca

dbonfigli:
> @dbonfigli Did you think about becoming k8s org member so you can approve changes for externalgrpc?

If you are willing to sponsor me and we find another reviewer that wants to sponsor me I am willing to keep contributing to this project as an official k8s org member.

x13n:
Happy to do so! If you don't have another sponsor you could ask on the SIG meeting.

vadasambar:
> If you are willing to sponsor me and we find another reviewer that wants to sponsor me I am willing to keep contributing to this project as an official k8s org member.

Added it to the agenda for Monday: https://docs.google.com/document/d/1RvhQAEIrVLHbyNnuaT99-6u9ZUMp7BfkPupT2LAZK7w/edit#bookmark=id.xtfu11692zyn

elmiko:
i'm happy to help with sponsoring as well, i reviewed some of the original grpc work.

PR #5939, opened 2023-07-10
Fix: Do not inject fakeNode for instance which has errors on create

azylinski:
#### What type of PR is this?

/kind bug

CA doesn't handle well situation when instance is in error state (thus no pods can schedule on to it) and node pool is at min size, so it cannot scale down the faulty node.

In this situation the node is considered to be still upcoming, so CA still registers the fake "upcoming node" and CA doesn't trigger scale up immediately, only after 15 minutes (when the node passes the timeout for registration). In order to fix this, we should not inject the fake upcoming node, if we already detected it's in error state.

```release-note
NONE
```

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -988,7 +988,9 @@ func (csr *ClusterStateRegistry) getCloudProviderNodeInstances() (map[string][]c
 	return csr.cloudProviderNodeInstancesCache.GetCloudProviderNodeInstances()
 }
 
-// Calculates which of the existing cloud provider nodes are not registered in Kubernetes.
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -1008,6 +1010,10 @@ func getNotRegisteredNodes(allNodes []*apiv1.Node, cloudProviderNodeInstances ma
 	return notRegistered
 }
 
+func withoutErrors(instance cloudprovider.Instance) bool {
```
azylinski, line 1013: In essence, we want to verify, if we can expect the instance to be Ready or do we see an issue, e.g.
 - instance State is Creating, but ErrorInfo says STOCKOUT

Alternative names: `isUsable`, `canBeRegistered`

azylinski:
Suggestions applied, please @x13n re-review :)

PR #5944, opened 2023-07-12
Adapt limitations for Pods without controller

voelzmo:
#### What type of PR is this?

/kind documentation

Triggered by a [conversation on slack](https://kubernetes.slack.com/archives/C09R1LV8S/p1689102593778659): Adapt the limitations section to make clear that VPA doesn't scale individual Pods, but requires a controller. Previously it seemed that you would get initial recommendations, but just no evictions/updates. This may have been true in the times of `v1beta1`, when we still had a `selector` in the VPA CRD, but I don't think this is still possible.

```release-note
NONE
```


jbartosik:
I found some old notes explaining why VPA is allowed to point only to top-most scalable cot roller:

- With label selector it was easy to make mistakes. For example one could specify a selector that matched all pods in the cluster; That's unlikely to give good results
- With label selector (or VPA pointing to controllers that aren't top-level) one can have overlapping VPAs; It makes it hard for VPA to act reasonably (which of multiple recommendations to apply?)
- Specifying target controller is consistent with HPA, MPA, Balancer

Also I think its nice that we know all potential targets for VPA in the cluster.

jbartosik (approved):
/lgtm
/approve

PR #5945, opened 2023-07-12
Re-introduce asynchronous node deletion and clean node deletion logic.

kawych:
This PR fixes a bug where node deletions could sequentially even if parallelism is used. This changes AddNodes function from NodeDeletionBatcher to only add nodes synchronously, but continue with processing them in a goroutine.

Additionally the PR removes some dead code and changes map keys from NodeGroups to their IDs to make it less error prone for the future.

Shoutout to @krzysied who found this bug.

### DIFF
--- a/cluster-autoscaler/core/scaledown/actuation/delete_in_batch.go
+++ b/cluster-autoscaler/core/scaledown/actuation/delete_in_batch.go
@@ -67,19 +67,11 @@ func NewNodeDeletionBatcher(ctx *context.AutoscalingContext, csr *clusterstate.C
 	}
 }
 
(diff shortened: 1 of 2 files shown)


krzysied:
/lgtm

Although, I think it should be done in separate commits (can be the some PR). My rule of thumb is to keep clean-up separated from bugs fixes. The things done here:
- clean-up
- fix regarding concurrent ion deletion being blocked 
- fix regarding using NodeGroup object instead Node Group Id.

kawych:
Thanks, created separate commits as you recommended. Also ran a test for that:

/unhold

x13n:
Thanks for fixing this!

/lgtm
/approve

PR #5949, opened 2023-07-13
cleanup: removing the references to max-empty-bulk-delete

Bryce-Soghigian:
In https://github.com/kubernetes/autoscaler/pull/5890/files we effectively remove max-empty-bulk-delete, and in parallel drain, the new scale down mode, we use a different deletion flag. 

Only one flag, should control the deletion threshold/# of deletion candidates at a time. This PR switches over to the new deletion flag consistently in the application.

We will leave the flag piece, until 1.29, when i plan on completely removing the flag and panic() if it's passed.

Fixes #

```release-note
Removing max-empty-bulk-delete from the data model and legacy scale down. 
```

```docs
```

### DIFF
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -273,8 +273,6 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 		*maxScaleDownParallelismFlag = *maxEmptyBulkDeleteFlag
 		klog.Warning("The max-empty-bulk-delete flag will be deprecated in k8s version 1.29. Please use max-scale-down-parallelism instead.")
 		klog.Infof("Setting max-scale-down-parallelism to %d, based on the max-empty-bulk-delete value %d", *maxScaleDownParallelismFlag, *maxEmptyBulkDeleteFlag)
-	} else if !isFlagPassed("max-empty-bulk-delete") && isFlagPassed("max-scale-down-parallelism") {
-		*maxEmptyBulkDeleteFlag = *maxScaleDownParallelismFlag
 	}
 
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/main.go
```diff
@@ -270,8 +270,6 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 		*maxScaleDownParallelismFlag = *maxEmptyBulkDeleteFlag
 		klog.Warning("The max-empty-bulk-delete flag will be deprecated in k8s version 1.29. Please use max-scale-down-parallelism instead.")
```
x13n, line 271: This comment would probably better fit the previous PR, but I think it might make sense to update `max-empty-bulk-delete` flag description to add a deprecation notice too.

x13n:
/lgtm
/approve

Although I see some test failing, but it shouldn't be related.

cluster-autoscaler/main.go
Bryce-Soghigian: Let me add the flag, seems apt when removing the other references to the autoscaler. Now to figure out the failing tests

x13n:
Hm... retest doesn't seem to work, let's try closing and reopening instead.

Bryce-Soghigian:
if this fail i can spend some time debugging it friday

x13n:
Can you rebase to see if that helps with failing tests? @Bryce-Soghigian

x13n:
Also,

/lgtm

x13n:
I'm not sure how without digging too deeply into the code, but looks like the test failures are actually related to the change. Did you try running the tests locally?

x13n:
Closing for now, please reopen if you have time to follow up on this.

/close

PR #5951, opened 2023-07-13
clusterapi: refresh kubeconfig bearer tokens for management and workload kubeconfigs dynamically

cnmcavoy:
#### What type of PR is this?
/kind bug

Continuously update the bearer token provided by a kube config file for the clusterapi provider.

Fixes https://github.com/kubernetes/autoscaler/issues/4784

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/config/kubernetes.go
+++ b/cluster-autoscaler/config/kubernetes.go
@@ -19,9 +19,13 @@ package config
 import (
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"net/url"
+	"os"
 	"strconv"
+	"time"
 
+	utilnet "k8s.io/apimachinery/pkg/util/net"
 	kube_rest "k8s.io/client-go/rest"
 	kube_client_cmd "k8s.io/client-go/tools/clientcmd"
 )
@@ -113,3 +117,39 @@ func GetKubeClientConfig(uri *url.URL) (*kube_rest.Config, error) {
 
 	return kubeConfig, nil
 }
+
+// NewDynamicKubeConfigRoundTripper wraps the http.RoundTripper to provide the latest bearer token from the kube config on disk
+func NewDynamicKubeConfigRoundTripper(kubeconfigPath string, rt http.RoundTripper) http.RoundTripper {
+	return &bearerAuthRoundTripper{
+		kubeconfigPath: kubeconfigPath,
+		rt:             rt,
+	}
+}
+
+type bearerAuthRoundTripper struct {
+	kubeconfigPath string
+	rt             http.RoundTripper
+
+	lastModified time.Time
+	bearerToken  string
+}
+
+func (rt *bearerAuthRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	stat, err := os.Stat(rt.kubeconfigPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to stat kubeconfig path: %w", err)
+	}
+	if stat.ModTime().After(rt.lastModified) {
+		kubeConfig, err := kube_client_cmd.BuildConfigFromFlags("", rt.kubeconfigPath)
+		if err != nil {
+			return nil, fmt.Errorf("cannot build kube cluster config: %w", err)
+		}
+		rt.bearerToken = kubeConfig.BearerToken
+		rt.lastModified = stat.ModTime()
+	}
+	req = utilnet.CloneRequest(req)
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", rt.bearerToken))
+	return rt.rt.RoundTrip(req)
+}
+
+var _ http.RoundTripper = &bearerAuthRoundTripper{}
--- a/cluster-autoscaler/cloudprovider/clusterapi/README.md
+++ b/cluster-autoscaler/cloudprovider/clusterapi/README.md
@@ -165,6 +165,10 @@ cluster-autoscaler --cloud-provider=clusterapi \
                    --kubeconfig=/mnt/workload.kubeconfig
 ```
 
+### A note on kubeconfig credentials
+
+When the cluster autoscaler is configured to read kubeconfig(s) from a path, by passing a `--kubeconfig` or `--cloud-config` option, the autoscaler will re-read the contents of the kubeconfig file after it detects the file was modified by it's modtime. This was added to support EKS clusters with the Cluster API Provider for AWS, where CAPA would provide the cluster with a kubernetes secret resource populated with a short-lived EKS kubeconfig credential, but may be useful for other Cluster API providers as well. In the past, it was not possible to use the Cluster API's kubeconfig credential directly as a secret volume, because when the short-lived contents were refreshed by the CAPA controller, the cluster autoscaler would not re-read the secret mount. There were various work arounds, for example manually creating a longer-lived kubeconfig from a service account. This is no longer necessary, cluster autoscaler will automatically refresh the kubeconfig credentials as needed.
+
 ## Enabling Autoscaling
 
 To enable the automatic scaling of components in your cluster-api managed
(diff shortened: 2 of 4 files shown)


elmiko (commented):
thanks for the patch @cnmcavoy , i am a little out of my depth with some of this token wrapping stuff.  i wonder if we might bug @stlaz or @liggitt to give us a comment on this?

elmiko:
i'm still keen to see this get included, @cnmcavoy have you done some local testing of the refresh?

cnmcavoy:
> i'm still keen to see this get included, @cnmcavoy have you done some local testing of the refresh?

Indeed has been using a fork of the CA with this patch included for a month. Initially this PR only covered the kube config in the clusterapi provider package, but that led us to discover there is a third kubeconfig in use for leader election. I updated the PR on the 17th to address that issue.

elmiko:
thanks for the update @cnmcavoy , i'm inclined to merge this based on your comments but i will bring up the PR at the CAPI meeting today just to see if we can get an extra review or two. i will plan to merge this on friday if there is no further activity.

jackfrancis:
@cnmcavoy code seems sane to me, is there an appropriate slot in the Cluster API provider documentation that we can use to describe this new feature? I'd be happy to build a cluster-autoscaler image from this branch and run through that test flow and report back (which would have the benefit of testing both the functionality and the documentation 😀).

cnmcavoy:
@jackfrancis sure! I added a note after the various configuration modes between management and workload clusters.

enj:
Why is this not using something like [HTTPWrappersForConfig](https://github.com/kubernetes/kubernetes/blob/375e9ee7c4562d736bc2359744a2f3d63f860b89/staging/src/k8s.io/client-go/transport/round_trippers.go#L39) which invokes [NewBearerAuthWithRefreshRoundTripper](https://github.com/kubernetes/kubernetes/blob/375e9ee7c4562d736bc2359744a2f3d63f860b89/staging/src/k8s.io/client-go/transport/round_trippers.go#L289) to handle this flow?  You should not be writing any custom code to handle automatic token refresh when client-go already does it for you.

cluster-autoscaler/config/kubernetes.go
```diff
@@ -113,3 +117,39 @@ func GetKubeClientConfig(uri *url.URL) (*kube_rest.Config, error) {
...
+	stat, err := os.Stat(rt.kubeconfigPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to stat kubeconfig path: %w", err)
+	}
+	if stat.ModTime().After(rt.lastModified) {
+		kubeConfig, err := kube_client_cmd.BuildConfigFromFlags("", rt.kubeconfigPath)
+		if err != nil {
+			return nil, fmt.Errorf("cannot build kube cluster config: %w", err)
+		}
+		rt.bearerToken = kubeConfig.BearerToken
```
enj, line 147: Loading the entire kubeconfig but only honoring the bearer token changing doesn't really make sense.  The bearer token should be in a distinct file that is reloaded, the rest of the contents should remain static.

cluster-autoscaler/config/kubernetes.go
```diff
@@ -113,3 +117,39 @@ func GetKubeClientConfig(uri *url.URL) (*kube_rest.Config, error) {
...
+		rt.lastModified = stat.ModTime()
+	}
+	req = utilnet.CloneRequest(req)
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", rt.bearerToken))
+	return rt.rt.RoundTrip(req)
+}
+
+var _ http.RoundTripper = &bearerAuthRoundTripper{}
```
enj, line 155: You need to implement `utilnet.RoundTripperWrapper`

cluster-autoscaler/config/kubernetes.go
cnmcavoy: https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/clusterapi#autoscaler-running-anywhere-with-separate-kubeconfigs-for-management-and-workload-clusters

Cluster Autoscaler can be configured in a mode where it is running in Cluster "1", but the cluster and resources it needs to scale and manage are in cluster "3", and the cloud configuration and node groups to scale are in cluster "2". In this scenario, there is no service account bearer token, except for running in Cluster "1" due to these being separate clusters.

The contents of the file we are loading is created by the CAPA awscluster controller, only has a bearer token, and is shaped like this:
```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: LS0t...
    server: https://XXXXX.us-east-2.eks.amazonaws.com
  name: cluster-name
contexts:
- context:
    cluster: cluster-3
    user: cluster-3-capi-admin
  name: cluster-3-capi-admin@cluster-3
current-context: cluster-3-capi-admin@cluster-3
kind: Config
preferences: {}
users:
- name: cluster-3-capi-admin
  user:
    token: k8s-aws-v1...
```

Are you suggesting that CAPA should encode a second secret value with just the `token` value contents above, and have the Cluster Autoscaler reload that contents, to simulate the behavior of a mounted service account bearer token? Or something else?

enj: Right if you change the YAML to:

```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: LS0t...
    server: https://XXXXX.us-east-2.eks.amazonaws.com
  name: cluster-name
contexts:
- context:
    cluster: cluster-3
    user: cluster-3-capi-admin
  name: cluster-3-capi-admin@cluster-3
current-context: cluster-3-capi-admin@cluster-3
kind: Config
preferences: {}
users:
- name: cluster-3-capi-admin
  user:
    tokenFile: location_of_file
```

And have the contents of `location_of_file` be:

```
k8s-aws-v1...
```

Then client-go will handle the contents of `location_of_file` changing for you.

RichardoC:
Is this still needed given the [Kubernetes client-go library](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/go.mod#L43C22-L43C22) has been updated to a version that should [fix this](https://docs.aws.amazon.com/eks/latest/userguide/service-accounts.html#:~:text=This%20means%20that,required%20time%20frame%3A)?

elmiko:
@RichardoC that link to the aws docs is only about eks clusters though, is there a link to an issue in client-go that might talk about this?

the capi provider doesn't necessarily use eks, so we would still want a refresh here (as i understand it).

RichardoC:
This functionality is upstream k8s behaviour, just couldn't find the link earlier <https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#auto-generated-legacy-serviceaccount-token-clean-up>
My best guess is client-go was fixed with <https://github.com/kubernetes/client-go/commit/66e83da33c604751671a39f929644b8f471e5cd9#diff-a6baf0c4cd67228defd6ac03faa4c9922f5cf4dd2b27672509043b33e3d881f7R289> specifically the NewBearerAuthWithRefreshRoundTripper codepath then this shouldn't be needed?

cnmcavoy:
@RichardoC I think there is some confusion here, this change was about kubeconfig's with an embedded service account token expiring. This differs from the built-in client-go handling of service account tokens, where you are correct, client-go does automatically refresh. So this is still an active issue for CAPI users with EKS clusters.

However, I went to a SIG-APIMachinery meeting last year to discuss better approaches to handling this and they suggested changes downstream, in how the CAPI project manage their kubeconfigs, to avoid this type of issue entirely. So I'm pursuing that over in https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4648

If all goes well, there will likely need to be some small autoscaler changes from that, but only for it's helm chart / docs. I'll open a new PR for that work.

elmiko:
@cnmcavoy thanks for the update and continuing to follow this issue. happy to hear about the status, and thanks for closing this.

PR #5961, opened 2023-07-17
chore(deps): update vendored hcloud-go to 2.0.0

apricote:
#### What type of PR is this?

/kind feature

Update the vendored copy of `github.com/hetznercloud/hcloud-go` to 2.0.0. This includes a breaking change where we changed the type of all `id` fields from `int` to `int64`. Read more about this in our [docs](https://docs.hetzner.cloud/#52-bit-ids). 

As updating the vendored code is annoying, I wrote a script that automatically updates the vendor. This has removed the License headers (Apache) which were added when the code was originally vendored. Instead I kept around the "LICENSE" (MIT) file from the hcloud-go repo.

```release-note
NONE
```

### DIFF
--- a/cluster-autoscaler/cloudprovider/hetzner/hack/update-vendor.sh
+++ b/cluster-autoscaler/cloudprovider/hetzner/hack/update-vendor.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+set -o pipefail
+set -o nounset
+set -o errexit
+
+UPSTREAM_REPO=${UPSTREAM_REPO:-https://github.com/hetznercloud/hcloud-go.git}
+UPSTREAM_REF=${UPSTREAM_REF:-main}
+
+vendor_path=hcloud-go
+
+original_module_path="github.com/hetznercloud/hcloud-go/v2/"
+vendor_module_path=k8s.io/autoscaler/cluster-autoscaler/cloudprovider/hetzner/hcloud-go/
+
+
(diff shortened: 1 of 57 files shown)


cluster-autoscaler/cloudprovider/hetzner/hack/update-vendor.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+set -o pipefail
+set -o nounset
+set -o errexit
+
+UPSTREAM_REPO=${UPSTREAM_REPO:-git@github.com:hetznercloud/hcloud-go.git}
```
jooola, line 6: To confirm, but this won't work if the user of the script doesn't have an ssh key configured to clone from Github.

Maybe you should use the https URL to clone the repo.

apricote: Updated :+1:

LKaemmerling (approved):

apricote:
Can I lgtm my own PR?

Lukas is in the Owners file but not part of the Kubernetes org: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/hetzner/OWNERS

/lgtm

apricote:
@feiskyer @BigDarkClown Could one of you /lgtm this PR?

mwielgus (approved):
/lgtm
/approve

PR #5974, opened 2023-07-20
Add PostPackingFunc to be run at the end of the estimation logic

azylinski:
#### What type of PR is this?

/kind refactor

Add PostPackingFunc, so the BinPackingEstimator can be extended easily.

```release-note
NONE
```

### DIFF
--- a/cluster-autoscaler/estimator/estimator.go
+++ b/cluster-autoscaler/estimator/estimator.go
@@ -45,15 +45,18 @@ type Estimator interface {
 // EstimatorBuilder creates a new estimator object.
 type EstimatorBuilder func(predicatechecker.PredicateChecker, clustersnapshot.ClusterSnapshot, EstimationContext) Estimator
 
+// EstimationAnalyserFunc to be run at the end of the estimation logic.
+type EstimationAnalyserFunc func(clustersnapshot.ClusterSnapshot, cloudprovider.NodeGroup, map[string]bool)
+
 // NewEstimatorBuilder creates a new estimator object from flag.
-func NewEstimatorBuilder(name string, limiter EstimationLimiter, orderer EstimationPodOrderer) (EstimatorBuilder, error) {
+func NewEstimatorBuilder(name string, limiter EstimationLimiter, orderer EstimationPodOrderer, estimationAnalyserFunc EstimationAnalyserFunc) (EstimatorBuilder, error) {
 	switch name {
 	case BinpackingEstimatorName:
 		return func(
 			predicateChecker predicatechecker.PredicateChecker,
 			clusterSnapshot clustersnapshot.ClusterSnapshot,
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/estimator/estimator.go
```diff
@@ -45,15 +45,18 @@ type Estimator interface {
 // EstimatorBuilder creates a new estimator object.
 type EstimatorBuilder func(predicatechecker.PredicateChecker, clustersnapshot.ClusterSnapshot, EstimationContext) Estimator
 
+// PostPackingFunc to be run at the end of the estimation logic.
+type PostPackingFunc func(clustersnapshot.ClusterSnapshot, cloudprovider.NodeGroup, map[string]bool)
```
jayantjain93, line 49: nit: rename to `PostEstimationProcessor` ?

jayantjain93 (commented):
Just a nit, rest LGTM.

/lgtm

cluster-autoscaler/estimator/estimator.go
azylinski: I'm not a huge fan of `PostEstimation` naming, as it would indicate something is happening after `Estimation`, which is the function name itself. And that is not true, as we're running it after fitting / packing / distribution / ...  but before snapshot revert, which is part of the Estimate function. You could argue it's happening `PostEstimation` core logic, but it's still not so obvious, especially at first glance.

The same goes with `Processor` suffix - in every other usage, the Processors are happening one after another (as we chain them together) - but that's not the case here.

Ideally, I would love the name, that would indicate the function is happening either: after bin-packing OR on modified cluster snapshot. 

"Naming is the hardest problem in computer science" :)

jayantjain93: Since this was a nit, feel free to resolve it. I don't have strong opinions on renaming it.

BigDarkClown: Really the only comment I have here is that I agree with Jayant. I understand where you're coming from with `PostEstimation` not being great, but adding `postPackagingFunc` here is not great either. The packaging is an implementation detail from the `BinpackingEstimator`, it should not be an interface level concept. Also it would be good for the name to imply that it is not changing the estimation (number of nodes etc.).

I think maybe something like `ObserveEstimation`, `RegisterEstimation`, `RecordEstimation`, `AnalyzeEstimation`? This is exactly what we are doing - analysing the estimation result.

BigDarkClown: Maybe even `XYZEstimationResult`? This way there is no ambiguity with the `Estimate` method ending.

azylinski: changed to: `EstimationAnalyserFunc` - please re-review :)

PR #5981, opened 2023-07-25
Remove ChangeRequirements with `OrEqual` from EvictionControl enhancement proposal

voelzmo:
#### What type of PR is this?
/kind cleanup

Adapt the proposal for EvictionControl: remove the `ChangeRequirement`s which include an equality check. [As discussed](https://github.com/kubernetes/autoscaler/pull/5754#pullrequestreview-1511181490): The VPA has other means to avoid evicting/updating the resource requests if no changes happened, there's no need to make the API for EvictionControl also expose this.

Mentioned as a taks in https://github.com/kubernetes/autoscaler/pull/5754#issuecomment-1647577540


Shubham82:
Thanks @voelzmo
/lgtm

jbartosik (approved):
/lgtm
/approve

fixes #5990

PR #5989, opened 2023-07-27
Clarify Eviction Control for Pods with multiple Containers

voelzmo:
#### What type of PR is this?
/kind documentation

Adapt the enhancement proposal for Eviction Control to Clarify what happens with Pods that have more than one Container. As discussed in https://github.com/kubernetes/autoscaler/pull/5599#discussion_r1197703104 this wasn't very obvious when reading the proposal.

```release-note
NONE
```


jbartosik (approved):
/lgtm
/approve

PR #5998, opened 2023-07-29
fix MaxNodeProvisionTime and ZeroOrMaxNodeScaling node-groups handlin…

kruftik:
#### What type of PR is this?

/kind bug

current code has a bug: externalgrpc service proto specification has `GetOptions(...)`  *optional* method but the response type does not have MaxNodeProvisionTime ZeroOrMaxNodeScaling fields. GetOption(..) wrapper does not respect such a situration it results in 0 (zero) MaxNodeProvisionTime duration and false value for ZeroOrMaxNodeScaling options since they are default type values in case if the specific externalgrpc service implements the `GetOptions()` method.

No


kruftik:
@BigDarkClown , @feiskyer ,
could you take a look at the PR please ?

towca:
Cloud provider PRs should be reviewed by the relevant OWNERS - in this case it looks like @dbonfigli. I took a look and the overall approach looks good, but I can't verify the proto/grpc parts.

/assign @dbonfigli
/label area/provider/externalgrpc

dbonfigli:
I am not an official kubernetes member so i cannot give a review that can be used to approve this PR even if i am listed in the ownes file for this cloud provider , anyway, it looks good but the same changes and a bit more are covered on PRs https://github.com/kubernetes/autoscaler/pull/5936 and https://github.com/kubernetes/autoscaler/pull/5937 that have been created before this, @towca can we give priority to those PRs, especially https://github.com/kubernetes/autoscaler/pull/5936 that in turn is needed for the final version of https://github.com/kubernetes/autoscaler/pull/5937 that otherwise would conflict on the proto files?

x13n:
Per the comment above it looks like this PR should be closed in favor of #5936 and #5937.

/close

PR #6001, opened 2023-07-31
Allow setting a separate timeout for instance registration.

kawych:
MaxNodeProvisionTimeProvider may be now passed via AutoscalerOptions, by default the same timeout is applied for end-to-end node provisioning and node registration after the instance is created. If needed, MaxNodeProvisionTimeProvider might be extended with other timeouts with custom logic (e.g. instance provisioning excluding registration as a node).

This PR allows setting max node provisioning time on higher granularity.

Directly this PR doesn't have any effect, but it adds support to override MaxNodeProvisionTimeProvider.

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -50,9 +50,17 @@ var (
 	errMaxNodeProvisionTimeProviderNotSet = errors.New("MaxNodeProvisionTimeProvider was not set in cluster state")
 )
 
+// maxNodeProvisionTimeProvider provides maximum times allowed for different stages of node provisioning. After
+// these times are reached, Cluster Autoscaler will clean up the nodes and try again.
+// "provisioning" refers to the duration from the point when a VM creation was requested until the node is registered in K8s.
+// "creation" refers to the portion of that time from VM creation request to a VM running.
+// "registration" refers to the duration starting when the VM was created until a node is registered in K8s.
+// "startup" refers to the duration starting when the node is registered in K8s to the time the node is ready.
 type maxNodeProvisionTimeProvider interface {
-	// GetMaxNodeProvisionTime returns MaxNodeProvisionTime value that should be used for the given NodeGroup.
+	// GetMaxNodeProvisionTime is a time a node has to register since its creation started
 	GetMaxNodeProvisionTime(nodeGroup cloudprovider.NodeGroup) (time.Duration, error)
+	// GetMaxNodeRegisterTime is a time a node has to register, starting from when its creation finished
+	GetMaxNodeRegisterTime(nodeGroup cloudprovider.NodeGroup) (time.Duration, error)
 }
 
 // ScaleUpRequest contains information about the requested node group scale up.
@@ -61,8 +69,11 @@ type ScaleUpRequest struct {
 	NodeGroup cloudprovider.NodeGroup
 	// Time is the time when the request was submitted.
 	Time time.Time
-	// ExpectedAddTime is the time at which the request should be fulfilled.
-	ExpectedAddTime time.Time
+	// ExpectedProvisionTime is the time at which the request should be fulfilled.
+	ExpectedProvisionTime time.Time
+	// ExpectedRegisterTime is the time at which the nodes that are already created are expected to register.
+	// Zero (time.Time{}) indicates unknown registration time, equivalent to not enforcing a timeout for now.
+	ExpectedRegisterTime time.Time
 	// How much the node group is increased.
 	Increase int
 }
@@ -105,8 +116,10 @@ type IncorrectNodeGroupSize struct {
 type UnregisteredNode struct {
 	// Node is a dummy node that contains only the name of the node.
 	Node *apiv1.Node
-	// UnregisteredSince is the time when the node was first spotted.
-	UnregisteredSince time.Time
+	// CreationStartTime is the time when the node was first spotted.
+	CreationStartTime time.Time
+	// RegistrationStartTime is the time when the node was provisioned and should start registration.
+	RegistrationStartTime time.Time
 }
 
 // ScaleUpFailure contains information about a failure of a scale-up.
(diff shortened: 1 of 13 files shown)


kawych:
/hold
I need to add unit tests specific to this functionality

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -271,6 +279,13 @@ func (csr *ClusterStateRegistry) updateScaleRequests(currentTime time.Time) {
 				nodeGroupName, currentTime.Sub(scaleUpRequest.Time))
 			continue
 		}
+		_, registrationTime, err := csr.GetProvisioningTimeouts(scaleUpRequest.NodeGroup)
+		if err != nil {
+			klog.Warningf("failed to retrieve max node registering time for nodeGroup %s", nodeGroupName)
+			continue
+		} else if csr.allInstancesAreCreated(nodeGroupName) && currentTime.Add(registrationTime).Before(scaleUpRequest.ExpectedAddTime) {
+			scaleUpRequest.ExpectedAddTime = currentTime.Add(registrationTime)
+		}
```
BigDarkClown, line 288: This feels extremely hacky. IIUC we want to update the `ExpectedAddTime` once all the nodes are created to handle the separate registration time limit.

Can we just add a different variable to `scaleUpRequest` that would be initially 0 and then set once the nodes are created? IMO we should rename the `ExpectedAddTime` to `ExpectedProvisionTime` and add a new `ExpectedRegistrationTime` variable that would be dedicated to registration limits. Otherwise this becomes really unreadable.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -300,6 +315,15 @@ func (csr *ClusterStateRegistry) updateScaleRequests(currentTime time.Time) {
 	csr.scaleDownRequests = newScaleDownRequests
 }
 
+func (csr *ClusterStateRegistry) allInstancesAreCreated(nodeGroupId string) bool {
```
BigDarkClown, line 318: Firstly, this should be called `areAllNodeCreated`. Secondly, this does not use passed `nodeGroupId`, which makes me think that it is not working as intended. Maybe with a single ongoing scale-up it works as expected, but I totally don't understand how this will behave if there are multiple ongoing scale0ups at once, especially if those would be in the same node group.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -1023,6 +1050,13 @@ func expectedToRegister(instance cloudprovider.Instance) bool {
 	return instance.Status != nil && instance.Status.State != cloudprovider.InstanceDeleting && instance.Status.ErrorInfo == nil
 }
 
+func provisionedTime(instance cloudprovider.Instance, currentTime time.Time) time.Time {
```
BigDarkClown, line 1053: This method is used in a single place and is totally non descriptive, please inline it instead.

cluster-autoscaler/clusterstate/providers/max_node_provision_time_provider.go
```diff
@@ -25,16 +26,34 @@ import (
...
 }
 
 type defultMaxNodeProvisionTimeProvider struct {
+	initialized              bool
 	context                  *context.AutoscalingContext
 	nodeGroupConfigProcessor nodegroupconfig.NodeGroupConfigProcessor
 }
 
-// GetMaxNodeProvisionTime returns MaxNodeProvisionTime value that should be used for the given NodeGroup.
+func (p *defultMaxNodeProvisionTimeProvider) Initialize(context *context.AutoscalingContext, nodeGroupConfigProcessor nodegroupconfig.NodeGroupConfigProcessor) {
```
BigDarkClown, line 39: Please add comment

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -1246,3 +1280,16 @@ func (csr *ClusterStateRegistry) GetScaleUpFailures() map[string][]ScaleUpFailur
 	}
 	return result
 }
+
+// ShouldRegisterByNow returns true i.f.f. the given node should be registered by now, according to the provided time.
+func (csr *ClusterStateRegistry) ShouldRegisterByNow(nodeGroup cloudprovider.NodeGroup, un UnregisteredNode, currentTime time.Time) (bool, error) {
```
BigDarkClown, line 1285: nit: `ShouldBeProvisionedByNow`? I like this method much more than the existing approach of exposing the time limits and adding them in the call site. Could we wrap other cases in this and not expose the limits externally?

BigDarkClown (commented):
Overall high level comments on this PR:

- Time limit naming is inconsistent across these changes. You are using multiple terms for nodes (nodes, intances), provisioning time (provisioning, end-to-end) or registration time (registration, instalation). Please unify the naming convention, this code is confusing to read right now.

- I understand that the maxNodeProvisionTimeProvider was added to AutoscalerOptions to enable switching it in the forked repositories. I however don't think we need to duplicate the interface and mock implementation in the `core` package - we should use the one defined in `clusterstate`. If the cyclic dependency is an issue we can move it to `clusterstate.providers`, this should resolve the issues.

- I think there are some bugs in the implementation, mostly in the `allInstancesAreCreated` method. I am not certain on how this code will behave if there are multiple ongoing scale-ups, some of which have registration issues. Has this case been tested? 

I feel like these are the most urgent comments, I'll do another round of review once those are addressed.

kawych:
Thanks for the review. Addressing comments, in a meantime a quick answer to your bigger comments, as it could spark some discussion
1. I can fix the consistency. Some of that is based on the existing conventions, CA already has a notion of Instance in addition to Node. I understand "node" as a K8s object, and instance as an underlying VM. Do you prefer to just consistently stick to terms "node", "provisioning" and "registration"? (regarding "provisioning", I think it's useful to explain at least once that it refers to "creation requested -> node registered" time, as my intuition would be to use this term for just VM provisioning).
2. Yeah, this is caused by circular dependencies. I'm not sure if it can be easily avoided. Now, clusterstate has to get maxNodeProvisionTimeProvider passed, this one gets AutoscalingContext in the "initialize" function and the AutoscalingContext stores ClusterState. I figured one version to untangle that would be to create a separate maxNodeProvisionTimeProvider interface for ClusterState's own use, that doesn't have Initialize function.
3. Thanks, great catch! I'll make sure to add test cases to cover that

BigDarkClown:
1. Yes, I think sticking to "nodes", "provisioning", "creation" and "registration" would be best. I totally agree that those need to be explained at some point, they are not that straightforward. But doing that once and then sticking to it will make the code more readable IMO.
2.  Yes, this is a bit messy at the moment. I have an idea on how we could simplify it, I'll give it a try.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -51,8 +51,10 @@ var (
 )
 
 type maxNodeProvisionTimeProvider interface {
```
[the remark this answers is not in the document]
kawych: Added a larger comment to explain this differentiation

cluster-autoscaler/clusterstate/clusterstate.go
kawych: Yeah, thanks for spotting this. Added filtering by nodeGroupId and extended TestScaleUpNodeRegistrationTimeout test so that reversing this change will fail an unit test.

cluster-autoscaler/clusterstate/clusterstate.go
kawych: Agreed, but I prefer to leave it out of this PR. Currently static_autoscaler uses MaxNodeProvisionTime in fixNodeGroupSize(), I'd prefer not to add too much scope to that. Added a TODO on MaxNodeProvisionTime and made maxNodeRegisterTime private.

PR #6011, opened 2023-08-03
Clean up the NodeGroupConfigProcessor interface and remove the MaxNodeProvisioningTimeProvider interface.

BigDarkClown:
#### What type of PR is this?

/kind cleanup

This PR cleans up the NodeGroupConfigProcessor and MaxNodeProvisioningTimeProvider interfaces.


cluster-autoscaler/processors/nodegroupconfig/node_group_config_processor.go
```diff
@@ -20,23 +20,23 @@ import (
...
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
-	"k8s.io/autoscaler/cluster-autoscaler/context"
+	"k8s.io/autoscaler/cluster-autoscaler/config"
 )
 
 // NodeGroupConfigProcessor provides config values for a particular NodeGroup.
 type NodeGroupConfigProcessor interface {
 	// GetScaleDownUnneededTime returns ScaleDownUnneededTime value that should be used for a given NodeGroup.
-	GetScaleDownUnneededTime(context *context.AutoscalingContext, nodeGroup cloudprovider.NodeGroup) (time.Duration, error)
+	GetScaleDownUnneededTime(options config.AutoscalingOptions, nodeGroup cloudprovider.NodeGroup) (time.Duration, error)
```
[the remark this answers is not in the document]
BigDarkClown: SGTM, changed to remove the options entirely.

cluster-autoscaler/clusterstate/providers/max_node_provision_time_provider.go
```diff
@@ -20,21 +20,21 @@ import (
...
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
-	"k8s.io/autoscaler/cluster-autoscaler/context"
+	"k8s.io/autoscaler/cluster-autoscaler/config"
 	"k8s.io/autoscaler/cluster-autoscaler/processors/nodegroupconfig"
 )
 
 // NewDefaultMaxNodeProvisionTimeProvider returns the default maxNodeProvisionTimeProvider which uses the NodeGroupConfigProcessor.
-func NewDefaultMaxNodeProvisionTimeProvider(context *context.AutoscalingContext, nodeGroupConfigProcessor nodegroupconfig.NodeGroupConfigProcessor) *defultMaxNodeProvisionTimeProvider {
-	return &defultMaxNodeProvisionTimeProvider{context: context, nodeGroupConfigProcessor: nodeGroupConfigProcessor}
+func NewDefaultMaxNodeProvisionTimeProvider(options config.AutoscalingOptions, nodeGroupConfigProcessor nodegroupconfig.NodeGroupConfigProcessor) *defaultMaxNodeProvisionTimeProvider {
```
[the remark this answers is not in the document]
BigDarkClown: You're right, it becomes obsolete. I've removed it entirely.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -197,17 +194,9 @@ func (csr *ClusterStateRegistry) RegisterOrUpdateScaleUp(nodeGroup cloudprovider
 	csr.registerOrUpdateScaleUpNoLock(nodeGroup, delta, currentTime)
 }
 
-// RegisterProviders registers providers in the cluster state registry.
-func (csr *ClusterStateRegistry) RegisterProviders(maxNodeProvisionTimeProvider maxNodeProvisionTimeProvider) {
-	csr.maxNodeProvisionTimeProvider = maxNodeProvisionTimeProvider
-}
-
 // MaxNodeProvisionTime returns MaxNodeProvisionTime value that should be used for the given NodeGroup.
 func (csr *ClusterStateRegistry) MaxNodeProvisionTime(nodeGroup cloudprovider.NodeGroup) (time.Duration, error) {
```
towca, line 198: I don't think we need this method either, it's body is literally just 1 line. Previously it was used to encapsulate the nil check, which is not relevant anymore IIUC?

cluster-autoscaler/clusterstate/testutils.go
```diff
@@ -20,18 +20,47 @@ import (
 	"time"
 
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
+	k8scloudprovider "k8s.io/cloud-provider"
 )
 
-// NewMockMaxNodeProvisionTimeProvider returns static maxNodeProvisionTimeProvider which returns constant MaxNodeProvisionTime for every NodeGroup.
-func NewMockMaxNodeProvisionTimeProvider(maxNodeProvisionTime time.Duration) *staticMockMaxNodeProvisionTimeProvider {
-	return &staticMockMaxNodeProvisionTimeProvider{maxNodeProvisionTime}
+// NewMockMaxNodeProvisionTimeProcessor returns static NodeGroupConfigProcessor which returns constant MaxNodeProvisionTime for every NodeGroup.
```
towca, line 26: Wouldn't this sort of functionality be better placed in the same package as the processor itself? It's not inherently connected to CSR.

I'm also not convinced we need it at all in most places - any test that is not testing overriding features would get the same behavior with the default implementations and setting the value in NodeGroupDefaults, right? But that's a smaller issue - it's repeated in a lot of places, so maybe the syntactic sugar is worth it.

cluster-autoscaler/clusterstate/clusterstate.go
BigDarkClown: It's also called outside of this package, so this is needed if we want to keep the processor as a private value. I think it makes sense to leave it as is.

cluster-autoscaler/processors/nodegroupconfig/node_group_config_processor_test.go
```diff
@@ -22,18 +22,16 @@ import (
...
-
-	"github.com/stretchr/testify/assert"
 )
 
 // This test covers all Get* methods implemented by
-// DelegatingNodeGroupConfigProcessor. The implementation and expectations are
+// NodeGroupConfigProcessor. The implementation and expectations are
 // identical hence a single test for the whole bunch.
-func TestDelegatingNodeGroupConfigProcessor(t *testing.T) {
+func TestNodeGroupConfigProcessor(t *testing.T) {
```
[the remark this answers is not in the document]
BigDarkClown: Fully agree, I got too trigger happy with Ctrl+R.

cluster-autoscaler/clusterstate/clusterstate.go
towca: Ok, I'd never expect that (because there's a dedicated processor with this functionality). The other callsites should definitely be using the processor for that and not CSR, this is a very roundabout way. Let's not bloat this PR further though, maybe add a TODO if you're up for it.

cluster-autoscaler/clusterstate/testutils.go
BigDarkClown: I've removed the static mock entirely, that was a good suggestion.

cluster-autoscaler/clusterstate/clusterstate.go
BigDarkClown: Fully agree, added to do to sort this out in a follow up.

PR #6019, opened 2023-08-07
Implement dynamically adjustment of NodeDeleteDelayAfterTaint based on round trip time between CA and api-server

damikag:
Implement dynamically adjustment of NodeDeleteDelayAfterTaint based on round trip time between CA and api-server

#### What type of PR is this?
/kind bug

Since CA and Scheduler are not synced, there could be race condition because scheduler might put pods on tainted nodes for deletion on scale down. Currently the probability of a race condition is reduced by waiting ```NodeDeleteDelayAfterTaint``` amount of time before deleting a node after tainting. The value of ```NodeDeleteDelayAfterTaint``` affects scale down throughput so this PR makes it dynamic based on the round trip time between CA and api-server.

Fixes #

NONE
```release-note

```

```docs

```


cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -270,6 +270,8 @@ type AutoscalingOptions struct {
 	MinReplicaCount int
 	// NodeDeleteDelayAfterTaint is the duration to wait before deleting a node after tainting it
 	NodeDeleteDelayAfterTaint time.Duration
+	// NodeDeleteDelayAfterTaintUpdateTimestamp it the last update timestamp of the NodeDeleteDelayAfterTaint
+	NodeDeleteDelayAfterTaintUpdateTimestamp time.Time
```
x13n, line 274: Why is this in `AutoscalingOptions`? It isn't something user would configure. What they might want to configure instead though is whether the dynamic logic is enabled at all, instead of using a fixed interval.

cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -170,6 +175,33 @@ func (a *Actuator) taintNodesSync(NodeGroupViews []*budgets.NodeGroupView) error
...
+
+			if ind == 0 && time.Now().After(a.ctx.NodeDeleteDelayAfterTaintUpdateTimestamp.Add(nodeDeleteDelayAfterTaintUpdateDuration)) {
+				// Get the newest version of the node.
+				newestNode, err := a.ctx.ClientSet.CoreV1().Nodes().Get(ctx.TODO(), node.Name, metav1.GetOptions{})
+				if err != nil || newestNode == nil {
+					klog.Errorf("Error getting newest version of the node: %v", err)
+					continue
+				}
+				for !taints.HasToBeDeletedTaint(newestNode) {
+					time.Sleep(20 * time.Millisecond)
```
x13n, line 187: Polling like this is going to generate too much traffic to the apiserver. CA should wait for an update on watch instead - listers are already getting this information.

cluster-autoscaler/core/scaledown/actuation/actuator.go
x13n, line 1: I don't think this logic should be inlined in the actuator. The logic for calculating NodeDeleteDelayAfterTaint dynamically can be encapsulated in a dedicated object that is unit-testable.

x13n (changes requested):

cluster-autoscaler/core/scaledown/actuation/actuator.go
damikag: Done

cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -159,18 +160,33 @@ func (a *Actuator) deleteAsyncEmpty(NodeGroupViews []*budgets.NodeGroupView) (re
 func (a *Actuator) taintNodesSync(NodeGroupViews []*budgets.NodeGroupView) errors.AutoscalerError {
 	var taintedNodes []*apiv1.Node
 	for _, bucket := range NodeGroupViews {
+		startTimestamp := map[string]time.Time{}
```
x13n, line 163: I think a lot of the code here could be moved to `UpdateLatencyTracker`. Specifically:
- `startTimestamp` is already a member of `UpdateLatencyTracker`, it actually breaks encapsulation to update it here. What's worse, you're writing it from one goroutine and reading from another without any synchronization.
- `resultChan` isn't actually necessary. You can have a synchronous interface:

    ```
    updateLatencyTracker := NewUpdateLatencyTracker(a.ctx.AutoscalingKubeClients.ListerRegistry.AllNodeLister())
    updateLatencyTracker.Start()
    (...)
    updateLatencyTracker.Add(node.Name, time.Now())
    (...)
    if err != nil { updateLatencyTracker.Stop() }
    (...)
    latency := updateLatencyTracker.Await()
    ```

- or async:

    ```
    updateLatencyTracker := NewUpdateLatencyTracker(a.ctx.AutoscalingKubeClients.ListerRegistry.AllNodeLister())
    go updateLatencyTracker.Start()
    (...)
    updateLatencyTracker.StartTimes <- startTime{node.Name, time.Now()}
    (...)
    if err != nil { close(updateLatencyTracker.StopChan) }
    (...)
    latency := <-updateLatencyTracker.ResultChan
    ```

Now it's mixed, which might cause bugs like when trying to update the map directly.

cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -159,18 +160,33 @@ func (a *Actuator) deleteAsyncEmpty(NodeGroupViews []*budgets.NodeGroupView) (re
...
 					_, _ = taints.CleanToBeDeleted(taintedNode, a.ctx.ClientSet, a.ctx.CordonNodeBeforeTerminate)
 				}
+				close(stopChan)
 				return errors.NewAutoscalerError(errors.ApiCallError, "couldn't taint node %q with ToBeDeleted", node)
 			}
 			taintedNodes = append(taintedNodes, node)
 		}
+		close(stopChan)
+		latency := <-resultChan
+		latencyDuration, err := time.ParseDuration(fmt.Sprintf("%vms", latency))
```
x13n, line 184: Woot, why convert int to a string, then parse it back to int (i.e. Duration) again?

cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -159,18 +160,33 @@ func (a *Actuator) deleteAsyncEmpty(NodeGroupViews []*budgets.NodeGroupView) (re
...
+		if err != nil {
+			klog.Errorf("Error parsing latency: ", err)
+			continue
+		}
+		a.ctx.NodeDeleteDelayAfterTaint = 2 * latencyDuration
```
x13n, line 189: If you're updating it for every bucket, you will just end up with latency from the last update, this should be updated after collecting results for all nodes. Also, I wouldn't modify context, but rather make this a field in actuator, to avoid side effects. That field should then be set either to `a.ctx.NodeDeleteDelayAfterTaint` or to the dynamic latency calculated here based on a flag.

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
```diff
@@ -0,0 +1,79 @@
...
+func (u *UpdateLatencyTracker) Start(stopChan <-chan struct{}, resultChan chan int64) {
+	for {
+		select {
+		case <-stopChan:
+			var remainingNodes []string
+			for node, _ := range u.startTimestamp {
+				remainingNodes = append(remainingNodes, node)
+			}
+			for remainingNodes != nil {
+				time.Sleep(100 * time.Millisecond)
```
x13n, line 34: Why this sleeps for 100ms, while the other sleep is for 50ms? By the way, hardcoding them like this will make testing harder.

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (u *UpdateLatencyTracker) Start(stopChan <-chan struct{}, resultChan chan int64) {
+	for {
+		select {
+		case <-stopChan:
+			var remainingNodes []string
+			for node, _ := range u.startTimestamp {
+				remainingNodes = append(remainingNodes, node)
+			}
+			for remainingNodes != nil {
```
x13n, line 33: This is a potentially infinite loop if something goes wrong. There has to be a timeout. I'd actually consider passing a context instead of stopChan here. When this is taking too long, the context can time out. When there's an error, the context can be cancelled.

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
```diff
@@ -0,0 +1,79 @@
...
+			return
+		default:
+		}
+		u.updateFinishTime()
+		time.Sleep(50 * time.Millisecond)
+	}
+}
+
+func (u *UpdateLatencyTracker) updateFinishTime() {
+	nodes, err := u.nodeLister.List()
```
x13n, line 49: Actually, you don't need to list all nodes and iterate through them - you can iterate through nodes with start timestamp and call `Get` on the lister.

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
```diff
@@ -0,0 +1,79 @@
...
+	var remainingNodes []string
+	for _, node := range nodes {
+		if _, ok := u.finishTimestamp[node]; !ok {
+			remainingNodes = append(remainingNodes, node)
+		}
+	}
+	return remainingNodes
+}
+
+func (u *UpdateLatencyTracker) calculateLatency() int64 {
```
x13n, line 70: I'd take max or some high percentile of durations here to be on the safe side. I'd also expect this to return a duration.

x13n (changes requested):

cluster-autoscaler/core/scaledown/actuation/actuator.go
damikag: Done

cluster-autoscaler/core/scaledown/actuation/actuator.go
damikag: Done

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
damikag: Extracted to a constant

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
damikag: Added a timeout

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
damikag: Done

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
damikag: Done

cluster-autoscaler/core/scaledown/actuation/actuator.go
damikag: Done

cluster-autoscaler/core/scaledown/actuation/actuator.go
damikag: Removed that part

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
```diff
@@ -0,0 +1,100 @@
...
+import (
+	"time"
+
+	"k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/taints"
+	"k8s.io/klog/v2"
+)
+
+const sleepDurationWhenPolling = 50 * time.Millisecond
+const watiForTaintingTimeoutDuration = 5 * time.Minute
```
x13n, line 12: typo: wati -> wait

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
```diff
@@ -0,0 +1,100 @@
...
+		stopChan:           make(chan struct{}),
+		resultChan:         make(chan time.Duration),
+	}
+}
+
+func (u *UpdateLatencyTracker) Start() {
+	for {
+		select {
+		case <-u.stopChan:
+			waitingForTaintingStartTime := time.Now()
```
x13n, line 45: That's a bit too much nesting. Can you extract the logic after `stopChan` is closed to a separate function?

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
```diff
@@ -0,0 +1,100 @@
...
+
+const sleepDurationWhenPolling = 50 * time.Millisecond
+const watiForTaintingTimeoutDuration = 5 * time.Minute
+
+type NodeTaintStartTime struct {
+	nodeName  string
+	startTime time.Time
+}
+
+type UpdateLatencyTracker struct {
```
x13n, line 19: Please capitalize all members that are part of the public interface of `UpdateLatencyTracker`. Right now it is hard to tell from this code which fields are meant to be directly modified and which ones are internal state of the object.

cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -158,20 +160,37 @@ func (a *Actuator) deleteAsyncEmpty(NodeGroupViews []*budgets.NodeGroupView) (re
...
+			}
 			err := a.taintNode(node)
 			if err != nil {
 				a.ctx.Recorder.Eventf(node, apiv1.EventTypeWarning, "ScaleDownFailed", "failed to mark the node as toBeDeleted/unschedulable: %v", err)
 				// Clean up already applied taints in case of issues.
 				for _, taintedNode := range taintedNodes {
 					_, _ = taints.CleanToBeDeleted(taintedNode, a.ctx.ClientSet, a.ctx.CordonNodeBeforeTerminate)
 				}
+				if a.ctx.AutoscalingOptions.DynamicNodeDeleteDelayAfterTaintEnabled {
+					close(updateLatencyTracker.stopChan)
```
x13n, line 180: This will cause the tracker to calculate latency even though we don't need it. With the sync api I suggested in another thread I distinguished between `Stop` and `Await` - I think something like this should be done here.

x13n (changes requested):

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
damikag: Done

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
damikag: Done

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
damikag: Done

cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -158,20 +160,37 @@ func (a *Actuator) deleteAsyncEmpty(NodeGroupViews []*budgets.NodeGroupView) (re
...
 				return errors.NewAutoscalerError(errors.ApiCallError, "couldn't taint node %q with ToBeDeleted", node)
 			}
 			taintedNodes = append(taintedNodes, node)
 		}
 	}
+	if a.ctx.AutoscalingOptions.DynamicNodeDeleteDelayAfterTaintEnabled {
+		updateLatencyTracker.AwaitOrStopChan <- true
+		latency, ok := <-updateLatencyTracker.ResultChan
+		if ok {
+			a.nodeDeleteDelayAfterTaint = 2 * latency
```
x13n, line 191: Please add a comment explaining where the magic constant (2) is coming from.

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
```diff
@@ -0,0 +1,107 @@
...
+import (
+	"time"
+
+	"k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/taints"
+	"k8s.io/klog/v2"
+)
+
+const sleepDurationWhenPolling = 50 * time.Millisecond
+const waitForTaintingTimeoutDuration = 2 * time.Minute
```
x13n, line 12: 2m seem a bit too generous. Let's reduce this to, say, 30s?

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
```diff
@@ -0,0 +1,107 @@
...
+
+type NodeTaintStartTime struct {
+	nodeName  string
+	startTime time.Time
+}
+
+type UpdateLatencyTracker struct {
```
x13n, line 19: Please add a comment explaining the purpose of this struct.

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
```diff
@@ -0,0 +1,107 @@
...
+	startTimestamp     map[string]time.Time
+	finishTimestamp    map[string]time.Time
+	remainingNodeCount int
+	nodeLister         kubernetes.NodeLister
+	StartTimeChan      chan NodeTaintStartTime
+	// Send bool for await close the chanell to stop
```
x13n, line 25: Can you clarify this a bit? Also, I think punctuation would really help here.

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
```diff
@@ -0,0 +1,107 @@
...
+func (u *UpdateLatencyTracker) await() {
+	waitingForTaintingStartTime := time.Now()
+	for {
+		switch {
+		case u.remainingNodeCount == 0:
+			latency := u.calculateLatency()
+			u.ResultChan <- latency
+			return
+		case time.Now().After(waitingForTaintingStartTime.Add(waitForTaintingTimeoutDuration)):
+			klog.Errorf("Timeout before tainting all nodes")
```
x13n, line 99: WDYT about mentioning the result of this error, rather than just the error:

"Timeout before tainting all nodes, latency measurement will be stale"

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
x13n, line 1: Thanks, this looks much better now. Please add unit tests.

x13n (changes requested):
Looks like gofmting is required:

```
Please run hack/update-gofmt.sh to fix the following files:
./cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
```

cluster-autoscaler/core/scaledown/actuation/actuator.go
damikag: Done

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
damikag: Done

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
damikag: Updated

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
damikag: Done

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
damikag: Done

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker_test.go
```diff
@@ -0,0 +1,112 @@
...
+	n2.Spec.Taints = append(n2.Spec.Taints, toBeDeletedTaint)
+	n3 := test.BuildTestNode("n3", 100, 100)
+
+	testCases := []struct {
+		description             string
+		startTime               time.Time
+		nodes                   []*apiv1.Node
+		durations               []time.Duration
+		wantLatency             time.Duration
+		wantResultChanNotClosed bool
```
x13n, line 66: Why the negation? Can't we have `wantResultChanClosed` instead? This makes the test cases unnecessarily harder to read.

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
```diff
@@ -0,0 +1,116 @@
...
+
+import (
+	"time"
+
+	"k8s.io/autoscaler/cluster-autoscaler/utils/kubernetes"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/taints"
+	"k8s.io/klog/v2"
+)
+
+const sleepDurationWhenPolling = 50 * time.Millisecond
```
x13n, line 11: Can it be shortened for testing?

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker_test.go
```diff
@@ -0,0 +1,112 @@
...
+	return timeToSend
+}
+
+func (m *mockClock) getIndex() int {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+	return m.index
+}
+
+func TestUpdateLatencyCalculation(t *testing.T) {
```
x13n, line 52: Can you add a test case in which a node is updated rather than tainted first time it is checked?

x13n (changes requested):

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker_test.go
damikag: Changed

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker_test.go
damikag: Done

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
damikag: Done

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
```diff
@@ -0,0 +1,117 @@
...
+	// waitForTaintingTimeoutDuration after passing a bool, latency calculation will be
+	// aborted and the ResultChan will be closed without returning a value) Closing the
+	// AwaitOrStopChan without passing any bool will abort the latency calculation.
+	AwaitOrStopChan chan bool
+	ResultChan      chan time.Duration
+	// now is used only to make the testing easier
+	now func() time.Time
+}
+
+func NewUpdateLatencyTracker(nodeLister kubernetes.NodeLister, sleepDurationWhenPolling time.Duration, now func() time.Time) *UpdateLatencyTracker {
```
x13n, line 38: nit: You don't need to take all these parameters here, an alternative would be to just use default and use a separate function like `NewUpdateLatencyTrackerForTesting` that uses shorter polling period and accepts a mock for `time.Now`.

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker_test.go
```diff
@@ -0,0 +1,176 @@
...
+		nodeTaintAfterNthGetCall: nodeTaintAfterNthGetCall,
+	}
+}
+
+func TestUpdateLatencyCalculation(t *testing.T) {
+
+	testCases := []struct {
+		description string
+		startTime   time.Time
+		// N < 1 result in node will not get tainted before timeout
```
x13n, line 106: nit: Maybe just don't put anything in the map to indicate the node never gets tainted?

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker_test.go
```diff
@@ -0,0 +1,176 @@
...
+		nodes = append(nodes, node)
+	}
+	return nodes, nil
+}
+
+// Get returns node from test lister. Add ToBeDeletedTaint to the node
+// during the N th call specified in the nodeTaintAfterNthGetCall
+func (l *TestCustomNodeLister) Get(name string) (*apiv1.Node, error) {
+	for _, node := range l.nodes {
+		l.getCallCount[node.Name] += 1
```
x13n, line 76: Why are you bumping call count for the node before checking whether it is the one requested?

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker.go
damikag: Done

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker_test.go
damikag: Changed

cluster-autoscaler/core/scaledown/actuation/update_latency_tracker_test.go
damikag: Done

PR #6027, opened 2023-08-08
feat: deprecating max-scale-down-parallelism in place of existing flag

Bryce-Soghigian:
#### What type of PR is this?

/kind bug 
/kind deprecation 
There are two flags that control the deletion thresholds. FOr now to mitigate, we just set them to the same value, but long term, we should only have one flag that controls the number of deletion candidates. 

It might be easier to deprecate the parallelism flag rather than maxEmptyBulk delete since it has been around for ages. So this pr is the implementation of that approach. I have an alternative PR that If the discussion dictates, we can use instead https://github.com/kubernetes/autoscaler/pull/5949

Fixes #

```release-note
Deprecation Notice added for max-scale-down-parallelism
```

```docs

```


x13n:
Note that semantics of each flag is slightly different. Max parallelism controls how many nodes are drained simultaneously, while max empty bulk delete controlled how many new drains can at most start during each CA loop. I think the former is simpler to reason about for CA users, since the latter behaves differently depending on how often the internal loop is running.

PR #6034, opened 2023-08-10
Set batch size to target size for atomically scaled groups

aleksandra-malinowska:
For atomically scaled groups with both empty and drainable nodes, we need to wait for both sets of nodes before proceeding with deletion. Before this change, batch size was set separately for empty and drainable node sets.

1st commit: test case exposing error with batch size
2nd commit: fix for the error with batch size
3rd commit: some extra validation and fix for tests (test cases leak state)
4th commit: test case exposing error with atomic groups interfering with each other's scale down (fails with `-race`)
5th commit: fix for the error with atomic groups interfering with each other's scale down

Sent for review together because changes depend on the previous ones, happy to sqush/split based on review comments.

/kind bug

```release-note
Fixes issue where scale down of atomically scaled node group may fail post-drain.
```

### DIFF
--- a/cluster-autoscaler/core/scaledown/actuation/actuator.go
+++ b/cluster-autoscaler/core/scaledown/actuation/actuator.go
@@ -148,7 +148,7 @@ func (a *Actuator) deleteAsyncEmpty(NodeGroupViews []*budgets.NodeGroupView) (re
 	}
 
 	for _, bucket := range NodeGroupViews {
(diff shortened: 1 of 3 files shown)


MaciekPytel (commented):
/lgtm
/approve

I'm ok merging this as it's fixing an existing bug, that I think should be addressed in a follow-up.

PR #6041, opened 2023-08-16
test(core): add test case for filterOutSchedulableByPacking

qianlei90:
#### What type of PR is this?

/kind flake

This PR only introduces additional unit test cases for the `filterOutSchedulableByPacking` function. 

I notice that the order of unschedulable pods affects the result of filterOutSchedulableByPacking, which may lead to the following unexpected scenario:

Assuming there are 2 pending pods using 12c cpu and 2 pending pods using 6c cpu. The node group has 16c allocatable cpu.

loop 1: CA finds the 4 pending pod and estimates 3 nodes: node1 (6c pod + 6c pod), node2 (12c pod), and node3 (12c pod). Node group is triggered to scale up for 3 nodes.

loop 2: CA finds the same pending pod, and 3 new fake nodes. The expected result of `filterOutSchedulableByPacking` is that all 4 pending pods can be scheduled to these 3 fake nodes: node1 (6c, 6c), node2 (12c), and node3 (12c). However, it is possible that one of the 12c pods fails to be scheduled, the nodes maybe like this: node1 (6c), node2 (6c), node3 (12c), and then this unschedulable pod triggers the node group to scale up for 1 more node.

The root cause of this issue is that `filterOutSchedulableByPacking` and `estimator` have a different schedule order. `filterOutSchedulableByPacking` schedule pods by spec.priority first, if pods have same priority, it depends on the list order from `lister`. `estimator` schedules pods by score, pods with higher utilization will be scheduled with higher priority.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/podlistprocessor/filter_out_schedulable_test.go
+++ b/cluster-autoscaler/core/podlistprocessor/filter_out_schedulable_test.go
@@ -151,6 +151,64 @@ func TestFilterOutSchedulable(t *testing.T) {
 			},
 			nodeMatches: matchesAllNodes,
 		},
+		"different schedule order 1: 12c 12c 6c 6c": {
+			nodesWithPods: map[*apiv1.Node][]*apiv1.Pod{
+				BuildTestNode("node16c_1", 15600, 200000): {},
+				BuildTestNode("node16c_2", 15600, 200000): {},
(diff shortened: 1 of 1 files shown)


qianlei90:
/hold 

The behavior of the scheduler appears to be inconsistent, the test is passed in my local development environment and failed in github checkes.

Issue #6050, opened 2023-08-18
Timing issue when loading history from AWS AMP


Evedel:
**Which component are you using?**:
Vertical Pod Autoscaler (Recommender only)

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:
In my case, VPA runs with [AWS AMP](https://aws.amazon.com/prometheus/) as a history provider. Pods also use iam role-based permissions.
That means that the VPA recommender deployment consist of two containers:
1. aws-sigv4-proxy ([repo](https://github.com/awslabs/aws-sigv4-proxy), [aws docs](https://docs.aws.amazon.com/prometheus/latest/userguide/AMP-onboard-ingest-metrics-existing-Prometheus.html)) with:
      ```
      - --host
      - aps-workspaces.${REGION}.amazonaws.com
      - --port
      - :8005
      ```
1. VPA recommender with
      ```
       - --storage=prometheus
       - --prometheus-address=http://localhost:8005/workspaces/${WORKSPACE_ID}
      ```
It works when proxy starts faster then recommender, no error in logs at all on any verbosity level and memory consumption is ~2Gi.

However, there is some timing issue. Approximately half of the time the `recommender` starts faster than the proxy, try to load history, fails on the first query with a connection error, and exists from loading history at all. Memory consumption is then ~50Mi. Recommendations are still given within approximately the same ranges.

I can see the error in logs, saying that:
```
Cannot get cluster history: cannot get usage history: cannot get timeseries for cpu: Post "http://localhost:8005/workspaces/${WORKSPACE_ID}/api/v1/query_range": dial tcp localhost:80: connect: connection refused
```
The error is from this "stacktrace":
https://github.com/kubernetes/autoscaler/blob/e1b03fac9958791790bfc18eeba9fab5cac0ccc1/vertical-pod-autoscaler/pkg/recommender/main.go#L188

https://github.com/kubernetes/autoscaler/blob/e1b03fac9958791790bfc18eeba9fab5cac0ccc1/vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go#L199

https://github.com/kubernetes/autoscaler/blob/e1b03fac9958791790bfc18eeba9fab5cac0ccc1/vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go#L216

**Describe the solution you'd like.**:
As the `recommender` works fine without historical data, I would ask if it is possible to add an argument to skip the history initialisation explicitly.

[...]

Issue #5031, opened 2022-07-15
`Cannot get cluster history...got result type model.Vector` with Promethus as storage


gsweene2:
**Which component are you using?**:

vertical-pod-autoscaler

**What version of the component are you using?**:

`v0.11.0` (note below I am not seeing this with `v0.10.0`)

Component version:

**What k8s version are you using (`kubectl version`)?**:

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.1", GitCommit:"632ed300f2c34f6d6d15ca4cef3d3c7073412212", GitTreeState:"clean", BuildDate:"2021-08-19T15:45:37Z", GoVersion:"go1.16.7", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.0", GitCommit:"ab69524f795c42094a6630298ff53f3c3ebab7f4", GitTreeState:"clean", BuildDate:"2021-12-07T18:09:57Z", GoVersion:"go1.17.3", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What environment is this in?**:

Minikube (also replicated on EKS)

**What did you expect to happen?**:

Expected to get the correct type back when Initializing VPA from history provider

**What happened instead?**:

Logged error
`Cannot get cluster history: cannot read last labels: expected query to return a matrix; got result type model.Vector`

**How to reproduce it (as minimally and precisely as possible)**:

Below is a bash script that will
* Restart minikube with necessary args
* Clone and install kube-prometheus
* Clone autscaler repo
* Checkout v0.11.0, patch to use prometheus as storage provider, hack-up, and save the logs to file (error seen)
* Checkout v0.10.0, patch to use prometheus as storage provider, hack-up, and save the logs to file (no error seen)

Expected `v0.10.0.logs` file
```
I0715 19:50:58.222415       1 cluster_feeder.go:229] Initializing VPA from history provider
```

Expected `v0.11.0.logs` file
```
I0715 19:50:13.957450       1 cluster_feeder.go:230] Initializing VPA from history provider
E0715 19:50:14.348896       1 cluster_feeder.go:233] Cannot get cluster history: cannot read last labels: expected query to return a matrix; got result type model.Vector
```

Script to report issue:
```bash
#!/bin/bash

set -e

# Once the scipt has been run once these can be set to false
[...]

PR #6054, opened 2023-08-22
Require prometheus storage to init history and add option to explicitly not use history storage

Evedel:
#### What type of PR is this?
/kind bug

It will allow to run VPA explicitly without history initialisation, but also will make it compulsory to initialise history if prometheus is a provider.

Fixes https://github.com/kubernetes/autoscaler/issues/6050

Please see the bug/issue description detailed in https://github.com/kubernetes/autoscaler/issues/6050

No

No


vertical-pod-autoscaler/pkg/recommender/main.go
```diff
@@ -55,7 +55,7 @@ var (
 	kubeApiQps             = flag.Float64("kube-api-qps", 5.0, `QPS limit when making requests to Kubernetes apiserver`)
 	kubeApiBurst           = flag.Float64("kube-api-burst", 10.0, `QPS burst limit when making requests to Kubernetes apiserver`)
 
-	storage = flag.String("storage", "", `Specifies storage mode. Supported values: prometheus, checkpoint (default)`)
+	storage = flag.String("storage", "", `Specifies storage mode. Supported values: prometheus, none, checkpoint (default)`)
```
mwielgus, line 58: Why default value is "" and not "checkpoint" ?

vertical-pod-autoscaler/pkg/recommender/main.go
```diff
@@ -118,7 +118,9 @@ func main() {
 	metrics_recommender.Register()
 	metrics_quality.Register()
 
-	useCheckpoints := *storage != "prometheus"
+	usePrometheus := *storage == "prometheus"
+	useNoStorage := *storage == "none"
+	useCheckpoints := (!usePrometheus) && (!useNoStorage)
```
mwielgus, line 123: The code should quit with an error if the flag value is incorrect, here it would just ignore the incorrect value and proceed with default.

Evedel: done

image

vertical-pod-autoscaler/pkg/recommender/main.go
Evedel: changed

image

vertical-pod-autoscaler/pkg/recommender/main.go
```diff
@@ -181,11 +193,18 @@ func main() {
...
 		if err != nil {
 			klog.Fatalf("Could not initialize history provider: %v", err)
 		}
-		recommender.GetClusterStateFeeder().InitFromHistoryProvider(provider)
+
+		historyInitErr := recommender.GetClusterStateFeeder().InitFromHistoryProvider(provider)
+		if historyInitErr != nil {
+			klog.Fatalf("Failed to load prometheus history")
+		}
+	} else if useNoStorage {
```
sophieliu15, line 206: nit: Could be just `else` since it is the last case.

sophieliu15 (commented):
LGTM.

vertical-pod-autoscaler/pkg/recommender/main.go
Evedel: in both cases was following existing behaviour, but agree that the fail fast is a better way of handling it

vertical-pod-autoscaler/pkg/recommender/main.go
Evedel: Personally, I'd prefer to list all options explicitly for readability (as there are many of them, and `if` bodies can grow big). Looking at this again, however, I think there might be some weird cases in the future world without a final `else`, so I added that as a "no storage selected" option instead.

vertical-pod-autoscaler/pkg/recommender/main.go
jbartosik: The default value is still "". The second argument of `flag.String()` is the default value of the flag the call defines: https://pkg.go.dev/flag#String

vertical-pod-autoscaler/pkg/recommender/main.go
```diff
@@ -118,7 +118,17 @@ func main() {
...
+	switch *storage {
+	case "checkpoint":
+		useCheckpoints = true
+	case "prometheus":
+		usePrometheus = true
+	case "none":
+		useNoStorage = true
+	default:
+		klog.Fatalf("Storage option '%s' is not supported. Supported values: prometheus, none, checkpoint", *storage)
+	}
```
jbartosik, line 131: Using bools here seems wrong - we want exactly one way to init. I'd use enum. Also I'd move as much as feasible to helper functions to keep main() small. Something like:

```
type InitSource  // this type doesn't have to be in main so let's keep it some somewhere else
const (
initFromCheckpoints initSource = iota
initFromPrometheus
noinitFromStorage
)

func getInitStorageSource(storage string) (initSource, error) { // this function doesn't have to be in main  so let's keep it some somewhere else
switch storage {
// ...
}

func InitStorage(storage string) error {  // this function doesn't have to be in main  so let's keep it some somewhere else
  initStorage, err := getInitStorageSource()
  if err != nil {
    return err
  } 
  switch initStorage {
    case initFromPrometheus:
       // ....
  }
}
```

then in main

just call `InitStorage(*storage)` and log fatal if it returns an error

jbartosik (changes requested):

vertical-pod-autoscaler/pkg/recommender/main.go
Evedel: that has to be github taking you to the outdated commits via permalink, please see this one
https://github.com/Evedel/autoscaler/blob/add-strict-mode-to-force-retries-during-prometheus-history-loading/vertical-pod-autoscaler/pkg/recommender/main.go#L60

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -192,13 +202,61 @@ type clusterStateFeeder struct {
...
+func (feeder *clusterStateFeeder) initFromHistoryProvider(historyProvider history.HistoryProvider) (historyInitError error) {
+	historyInitError = nil
 	klog.V(3).Info("Initializing VPA from history provider")
 	clusterHistory, err := historyProvider.GetClusterHistory()
 	if err != nil {
-		klog.Errorf("Cannot get cluster history: %v", err)
+		historyInitError = err
+	}
+	if len(clusterHistory) == 0 {
+		klog.Warningf("history provider returned no pods")
```
Evedel, line 259: This potentially could also be an error/restart, but not sure if any valid cases are possible, so leaving with a warning for now.

As for the error cases, at the moment `--metric-for-pod-labels=up{job=\"kubernetes-pods\"}`, if prom is not configured to have a given job label, the query will not return any results, so history will not be used (and users won't be notified about the misconfiguration)

vertical-pod-autoscaler/pkg/recommender/main.go
Evedel: Thanks for the suggestion @jbartosik , I think that is a good way to reorganize code, true. I moved the init logic from `main` to `cluster_feeder`, please let me know what you think about this.

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
sophieliu15: This case does not seem to be treated differently (e.g., issue warning logs) before this PR. Is anything changed in this PR that make this case special? Or you just hope to add a log for this case?

sophieliu15:
@kwiesmueller Could you do a final round of review and approve if it looks good to you?
cc: @laoj2

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -192,13 +202,60 @@ type clusterStateFeeder struct {
...
+		return fmt.Errorf("history provider option is not set. Supported values: prometheus, none, checkpoint")
+	}
+	return nil
 }
 
-func (feeder *clusterStateFeeder) InitFromHistoryProvider(historyProvider history.HistoryProvider) {
+func GetHistorySourceFromArg(historySource string) (HistorySource, error) {
+	switch historySource {
+	case "checkpoint":
+		return Checkpoints, nil
```
kwiesmueller, line 240: I think I'd prefer those "enums" to be named more explicitly like `HistorySourceCheckpoints` or `CheckpointsHistorySource`. It's a little duplicated with the types, but when used inside the package it can be unclear that this is really just an enum value.
But more of a nit.

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -192,13 +202,60 @@ type clusterStateFeeder struct {
...
+		return Prometheus, nil
+	case "none":
+		return None, nil
+	default:
+		return Undefined, fmt.Errorf("storage option '%s' is not supported. Supported values: prometheus, none, checkpoint", historySource)
+	}
+}
+
+func (feeder *clusterStateFeeder) initFromHistoryProvider(historyProvider history.HistoryProvider) (historyInitError error) {
+	historyInitError = nil
```
kwiesmueller, line 251: I think this is nil by default. Would you mind making the return unnamed and do `var historyInitError error` here instead?

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -192,13 +202,60 @@ type clusterStateFeeder struct {
...
 	klog.V(3).Info("Initializing VPA from history provider")
 	clusterHistory, err := historyProvider.GetClusterHistory()
 	if err != nil {
-		klog.Errorf("Cannot get cluster history: %v", err)
+		historyInitError = err
```
kwiesmueller, line 255: Can you clarify why returning the error here right away doesn't work. I'm not sure if the historyInitError variable is really needed.
I think when a function returns a value and an error, as soon as the error is non-nil the caller should usually make no assumptions about the value. If you get an error this function should probably handle that error and return.

kwiesmueller (changes requested):
Just a few nits.

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
Evedel: Yes, this is behaviour preexisting before this PR. (also it can be a valid behaviour, say when a user specified the metric filter to target a subset of pods, but no such pods existed in the required history window).

Adding warning here just in case when this behaviour is not intentional, so it is easier to uncover a misconfiguration, for example see https://github.com/kubernetes/autoscaler/issues/6221 and https://github.com/kubernetes/autoscaler/issues/5031#issuecomment-1450583325.

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
Evedel: That was intentional, as I personally, prefer to avoid "smurf" naming, but please let me know if this is a requested change, will be happy to make it

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
Evedel: Very valid point, I was thinking that more error flavours are possible with generalised `historyProvider`, but that is clearly overthinking, I'll push requested change when will have a free moment

ben-bertrands-hs:
Hi, Can this be merged in?

Evedel:
Any chance someone can have a look please? 

@jbartosik
@kgolab
@mwielgus
@kwiesmueller
@sophieliu15

PR #6056, opened 2023-08-23
Add support to filter out pods being deleted.

jayantjain93:
#### What type of PR is this?

/kind bug

Pods in process of being deleted should be filtered out from being considered for scale up. This should be done before filter out schedulable, as the pod can "virtually" consume capacity of the cluster.

This being in line with schedule code (https://github.com/kubernetes/kubernetes/blob/cb56c4c627fc627b2eb1ba69a435cc3f205f7363/pkg/scheduler/schedule_one.go#L350-L354), which will no schedule pods that have a deletionTimestamp set.

This has further implications on scale down also, the pod can stay alive for their gracePeriod, however, they will stay pending and consume "virtual" capacity on the cluster, having impact on scale down decisions.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/utils/kubernetes/listers.go
+++ b/cluster-autoscaler/utils/kubernetes/listers.go
@@ -201,7 +201,9 @@ func (lister *scheduledAndUnschedulablePodLister) List() (scheduledPods []*apiv1
 		}
 		_, condition := podv1.GetPodCondition(&pod.Status, apiv1.PodScheduled)
 		if condition != nil && condition.Status == apiv1.ConditionFalse && condition.Reason == apiv1.PodReasonUnschedulable {
-			unschedulablePods = append(unschedulablePods, pod)
+			if pod.GetDeletionTimestamp() == nil {
+				unschedulablePods = append(unschedulablePods, pod)
+			}
 		}
 	}
 	return scheduledPods, unschedulablePods, nil


jayantjain93:
@k8s-ci-robot retest

olagacek:
As an alternative: have you considered filtering out in lister by adding additional selector term (https://github.com/kubernetes/autoscaler/blob/8c749fc655b807da8e321d634c35e202088da3f1/cluster-autoscaler/utils/kubernetes/listers.go#L210-L222 ) ? This way we won't need additional processor and code would be more straightforward. If there are some issues with this approach I'm open to discussion

olagacek:
> As an alternative: have you considered filtering out in lister by adding additional selector term (
> 
> https://github.com/kubernetes/autoscaler/blob/8c749fc655b807da8e321d634c35e202088da3f1/cluster-autoscaler/utils/kubernetes/listers.go#L210-L222
> 
> ) ? This way we won't need additional processor and code would be more straightforward. If there are some issues with this approach I'm open to discussion

Or we could leave selector as is, but remove pods with deletionTimestamp from unschedulablePods slice by adding a new filter here: https://github.com/kubernetes/autoscaler/blob/8c749fc655b807da8e321d634c35e202088da3f1/cluster-autoscaler/utils/kubernetes/listers.go#L204

jayantjain93:
> > As an alternative: have you considered filtering out in lister by adding additional selector term (
> > https://github.com/kubernetes/autoscaler/blob/8c749fc655b807da8e321d634c35e202088da3f1/cluster-autoscaler/utils/kubernetes/listers.go#L210-L222
> > 
> > ) ? This way we won't need additional processor and code would be more straightforward. If there are some issues with this approach I'm open to discussion
> 
> Or we could leave selector as is, but remove pods with deletionTimestamp from unschedulablePods slice by adding a new filter here:
> 
> https://github.com/kubernetes/autoscaler/blob/8c749fc655b807da8e321d634c35e202088da3f1/cluster-autoscaler/utils/kubernetes/listers.go#L204

I would prefer to avoid overloading a lister with filtering logic. However, the bigger underlying issue I see is to filter these pods as early in the process as possible.
I've changed the implementation to account for this. And we now filter the pods right after listing.

olagacek:
> > > As an alternative: have you considered filtering out in lister by adding additional selector term (
> > > https://github.com/kubernetes/autoscaler/blob/8c749fc655b807da8e321d634c35e202088da3f1/cluster-autoscaler/utils/kubernetes/listers.go#L210-L222
> > > 
> > > ) ? This way we won't need additional processor and code would be more straightforward. If there are some issues with this approach I'm open to discussion
> > 
> > 
> > Or we could leave selector as is, but remove pods with deletionTimestamp from unschedulablePods slice by adding a new filter here:
> > https://github.com/kubernetes/autoscaler/blob/8c749fc655b807da8e321d634c35e202088da3f1/cluster-autoscaler/utils/kubernetes/listers.go#L204
> 
> I would prefer to avoid overloading a lister with filtering logic. However, the bigger underlying issue I see is to filter these pods as early in the process as possible. I've changed the implementation to account for this. And we now filter the pods right after listing.

After thinking a little about it, I don't think it's overloading lister with filtering logic. IMHO it should be listers job to return only the pods that CA should care about and not delegate it to some different place. This pattern might become problematic if we start to use lister in different place and forget about the filtering being done in a different place. 

Also, we already do some filtering inside a lister - we skip Failed and Succeeded pods. It's done in a different place - using selector - but nevertheless it's filtering.

jayantjain93:
> As

That makes sense. Updated the code to reflect this.

PR #6061, opened 2023-08-25
use vpaObject instead of vpaModel to reduce deps of post-processors

dbenque:
#### What type of PR is this?

/kind cleanup

The post-processor interface is taking a model.Vpa when a simple vpa object is enough.
Using a vpa object reduces dependencies of the post-processing layer on the rest of the code base. This is better for user that creates external recommender and that want to include the post-processor.

Not done in this PR to reduce the size of the diff: move the post-processors in a dedicated package. This move should be done in a follow-up PR. This will isolate the post-processors from the rest of the recommender routine and make it easy to import/maintain/understand.

This can be considered as a code cleanup. Thanks to this reusing the post-processing layer in external-recommender will be easier since it will come with less dependencies on the rest of the codebase.

NONE

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/routines/capping_post_processor.go
+++ b/vertical-pod-autoscaler/pkg/recommender/routines/capping_post_processor.go
@@ -17,10 +17,10 @@ limitations under the License.
 package routines
 
 import (
-	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1"
-	"k8s.io/autoscaler/vertical-pod-autoscaler/pkg/recommender/model"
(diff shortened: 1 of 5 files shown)


vertical-pod-autoscaler/pkg/recommender/routines/capping_post_processor.go
```diff
@@ -17,10 +17,10 @@ limitations under the License.
 package routines
 
 import (
+	"k8s.io/klog/v2"
```
dbenque, line 20: this is not an add, just lastest version of linter making different organisation of imports.
Same as https://github.com/kubernetes/autoscaler/pull/5676/files/143402350b219a8ee36ea451ec2f5535a097ddcf#r1173813414

PR #6062, opened 2023-08-28
David.benque/prevent empty resource propagation upstream 2

dbenque:

vertical-pod-autoscaler/admission-controller

#### What type of PR is this?

/kind bug

Using external recommender that was coming with a bug, we ended up with a VPA object that was presenting a recommendation for a resource that has an empty name:

```
 resources:
          limits:
            cpu: "1"
            memory: 3Gi
          requests:
            "": "0"              <== empty key
            cpu: "1"
            ephemeral-storage: 10Gi
            memory: 3Gi
```

In this PR we are proposing to sanitise the recommendation by removing any resource key that would be empty so that the admission is not pushing "garbage" to the pods.

Possible extension: instead of purging empty key, we could have a list of allowed resourceName given as input parameter of the admission. This would also allow us to purge unknown resource or the one coming with typo.

If a recommendation comes with a resource key that is empty, this resource will be ignored in the admission-controller.

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/resource/pod/recommendation/recommendation_provider.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/resource/pod/recommendation/recommendation_provider.go
@@ -20,10 +20,11 @@ import (
 	"fmt"
 
 	core "k8s.io/api/core/v1"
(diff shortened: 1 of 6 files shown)


mwielgus (approved):
/lgtm
/approve

PR #6066, opened 2023-08-28
The autoscaler does not scale node groups on non-amd64 clusters when pods explicitly require non-amd64 nodes in node affinity

### LINKED ISSUE #6064: The autoscaler does not scale node groups on non-amd64 clusters when pods explicitly require non-amd64 nodes in node affinity
The autoscaler does not scale node groups on non-amd64 clusters when pods explicitly require non-amd64 nodes in node affinity

**What version of the component are you using?**: master branch

Component version: any

**What k8s version are you using (`kubectl version`)?**: any

**What environment is this in?**: cluster API and most cloud providers

**What did you expect to happen?**: when the autoscaler provider build the generic labels to perform autoscale from zero decisions, most cloud providers consider `amd64` as the default architecture to use, leading a pod reporting a node selector term to be scheduled on non-amd64 nodes, unable to trigger the autoscaler although non-amd64 nodes are available in the cluster.

What behaviour did you expect to see?

Users should be allowed to override the default architecture in the buildGenericLabels method. This will not take into account cases like clusters heterogeneous in architecture (CP on arm64, DP on amd64, or CP on amd64 and DP arm64 + amd64 nodes). However, this case should be considered user-specific or cloud-provider-specific.

**What happened instead?**:

When a pod reports a node selector term like `kubernetes.io/arch in {arm64}` in an arm64 cluster, the pod does not trigger the scale out of node groups.

In any kubernetes arm64 deployment with the autoscaler enabled, try to define a deployment with node selector term `kubernetes.io/arch in {arm64}` , resource requests, and set the replicas such that a pod keeps pending scheduling due to missing resources.

Example:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: 'my-deployment'
spec:
  selector:
    matchLabels:
      app: name
  replicas: 3
  template:
    metadata:
      labels:
        app: name
    spec:
      affinity:
        nodeAffinity:
            nodeSelectorTerms:
              - matchExpressions:
                - key: kubernetes.io/arch
                  operator: In
                  values:
                    - "arm64"
      containers:
        - name: container
          image: >-
            quay.io/openshifttest/hello-openshift:1.2.0
          resources:
              requests:
                cpu: "2"
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
  paused: false
```

aleskandro:
#### What type of PR is this?

/kind bug

The architecture label in the build generic labels method of the Cluster API provider is now populated using the `cpu.GetDefaultScaleFromZeroArchitecture().Name()` method.

The method allows users deploying the cluster-autoscaler to define the default architecture to be used by the cluster-autoscaler for scale up from zero via the `--scale-up-from-zero-default-arch` flag. `Amd64` is kept as a fallback for historical reasons and to maintain compatibility with current providers.

The `SystemArchitecture` type and the functions for parsing architecture names are copied over from the GCE implementation, to be generic across all the defined cloud providers.

However, the GCE implementation and the other cloud providers are currently left untouched.

The introduced changes will not take into account the case of nodes heterogeneous in architecture. The labels generation to infer properties like the cpu architecture from the node groups' features should be considered cloud-provider/user specific.

Cloud providers willing to opt into this implementation are expected to change their manager code by replacing
the `cloudprovider.DefaultArch` constant with the `GetDefaultScaleFromZeroArchitecture` function exposed by this package.
Usually, this should be done in the function responsible for generating the generic labels of the given
cloud provider's manager code (for example, `buildGenericLabels(.)`) as follows:

```go
func (...) buildGenericLabels(...) (map[string]string) {
	result := make(map[string]string)
	result[apiv1.LabelArchStable] = cpu.GetDefaultScaleFromZeroArchitecture().Name()
	// ...
}
```

Fixes #6064 

```release-note
Fixes #6064: Users of Cluster API can override the default architecture to consider in the templates for autoscaling from zero so that pods requesting non-amd64 nodes in their node selector terms can trigger the scale-up in non-amd64 single-arch clusters. 
```

```docs

```

cc @elmiko @jeffdyoung @Prashanth684

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/README.md
+++ b/cluster-autoscaler/cloudprovider/clusterapi/README.md
@@ -20,6 +20,7 @@ cluster.
   * [Scale from zero support](#scale-from-zero-support)
     * [RBAC changes for scaling from zero](#rbac-changes-for-scaling-from-zero)
     * [Pre-defined labels and taints on nodes scaled from zero](#pre-defined-labels-and-taints-on-nodes-scaled-from-zero)
+    * [CPU Architecture awareness for single-arch clusters](#cpu-architecture-awareness-for-single-arch-clusters)
 * [Specifying a Custom Resource Group](#specifying-a-custom-resource-group)
 * [Specifying a Custom Resource Version](#specifying-a-custom-resource-version)
 * [Sample manifest](#sample-manifest)
@@ -275,6 +276,16 @@ metadata:
     capacity.cluster-autoscaler.kubernetes.io/labels: "key1=value1,key2=value2"
     capacity.cluster-autoscaler.kubernetes.io/taints: "key1=value1:NoSchedule,key2=value2:NoExecute"
 ```
+
+#### CPU Architecture awareness for single-arch clusters 
+
+Users of single-arch non-amd64 clusters who are using scale from zero 
+support should also set the `CAPI_SCALE_ZERO_DEFAULT_ARCH` environment variable
+to set the architecture of the nodes they want to default the node group templates to.
+The autoscaler will default to `amd64` if it is not set, and the node 
+group templates may not match the nodes' architecture, specifically when 
+the workload triggering the scale-up uses a node affinity predicate checking 
+for the node's architecture.
 
 ## Specifying a Custom Resource Group
 
--- a/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_utils.go
+++ b/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_utils.go
@@ -18,8 +18,11 @@ package clusterapi
 
 import (
 	"fmt"
+	"k8s.io/klog/v2"
+	"os"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -36,6 +39,20 @@ const (
 	maxPodsKey      = "capacity.cluster-autoscaler.kubernetes.io/maxPods"
 	taintsKey       = "capacity.cluster-autoscaler.kubernetes.io/taints"
 	labelsKey       = "capacity.cluster-autoscaler.kubernetes.io/labels"
+	// UnknownArch is used if the Architecture is Unknown
+	UnknownArch SystemArchitecture = ""
+	// Amd64 is used if the Architecture is x86_64
+	Amd64 SystemArchitecture = "amd64"
+	// Arm64 is used if the Architecture is ARM64
+	Arm64 SystemArchitecture = "arm64"
+	// Ppc64le is used if the Architecture is ppc64le
+	Ppc64le SystemArchitecture = "ppc64le"
+	// S390x is used if the Architecture is s390x
+	S390x SystemArchitecture = "s390x"
+	// DefaultArch should be used as a fallback if not passed by the environment via the --scale-up-from-zero-default-arch
+	DefaultArch = Amd64
+	// scaleUpFromZeroDefaultEnvVar is the name of the env var for the default architecture
+	scaleUpFromZeroDefaultArchEnvVar = "CAPI_SCALE_ZERO_DEFAULT_ARCH"
 )
 
 var (
@@ -79,9 +96,24 @@ var (
 	nodeGroupMinSizeAnnotationKey = getNodeGroupMinSizeAnnotationKey()
 	nodeGroupMaxSizeAnnotationKey = getNodeGroupMaxSizeAnnotationKey()
 	zeroQuantity                  = resource.MustParse("0")
+
+	systemArchitecture *SystemArchitecture
+	once               sync.Once
 )
 
 type normalizedProviderID string
+
+// SystemArchitecture represents a CPU architecture (e.g., amd64, arm64, ppc64le, s390x).
+// It is used to determine the default architecture to use when building the nodes templates for scaling up from zero
+// by some cloud providers. This code is the same as the GCE implementation at
+// https://github.com/kubernetes/autoscaler/blob/3852f352d96b8763292a9122163c1152dfedec55/cluster-autoscaler/cloudprovider/gce/templates.go#L611-L657
+// which is kept to allow for a smooth transition to this package, once the GCE team is ready to use it.
+type SystemArchitecture string
+
+// Name returns the string value for SystemArchitecture
+func (s SystemArchitecture) Name() string {
+	return string(s)
+}
 
 // minSize returns the minimum value encoded in the annotations keyed
 // by nodeGroupMinSizeAnnotationKey. Returns errMissingMinAnnotation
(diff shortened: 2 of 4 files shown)


elmiko (commented):
thanks for this update @aleskandro , in general this looks good to me and i'm happy to approve the cluster-api changes.

i am concerned about getting consensus on this pr since it touches so many other providers.

/lgtm

elmiko:
we talked about this at the sig meeting today, the general consensus is that this should be limited to just cluster-api for now. we can add the flag to help determine the default, and provide some code comments for how other providers can use it. but to start, only cluster-api will respect the flag.

/lgtm cancel

aleskandro:
Hi @elmiko , thanks for the review and thanks all for the discussion during the sig meeting today.

I made the requested changes:
- No other cloud providers than Cluster API are now affected by changes in this PR
- The SystemArchietcture type and its methods are now copied over from the GCE implementation (it is left untouched, too)
- Updated usage description and godoc comments with information about how to allow other cloud providers to opt in the facility given by the new command line argument
- Updated the CAPI readme to inform users about Arch-awareness in the scale from zero for single-arch clusters.

cluster-autoscaler/cloudprovider/clusterapi/README.md
```diff
@@ -275,6 +276,16 @@ metadata:
     capacity.cluster-autoscaler.kubernetes.io/taints: "key1=value1:NoSchedule,key2=value2:NoExecute"
 ```
 
+#### CPU Architecture awareness for single-arch clusters 
+
+Users of non-amd64, single-arch clusters willing to enable scale from zero
```
elmiko, line 281: i think i would reword this as:

```suggestion
Users of single-arch non-amd64 clusters who are using scale from zero 
```

cluster-autoscaler/cloudprovider/clusterapi/README.md
```diff
@@ -275,6 +276,16 @@ metadata:
     capacity.cluster-autoscaler.kubernetes.io/taints: "key1=value1:NoSchedule,key2=value2:NoExecute"
 ```
 
+#### CPU Architecture awareness for single-arch clusters 
+
+Users of non-amd64, single-arch clusters willing to enable scale from zero 
+support should also set the `--scale-up-from-zero-default-arch` flag to the 
+architecture of the nodes they want to default the node group templates to.
+The autoscaler will default to amd64 if the flag is not set, and the node
```
elmiko, line 284: i would add some code blocks here for style

```suggestion
The autoscaler will default to `amd64` if the flag is not set, and the node 
```

cluster-autoscaler/cloudprovider/clusterapi/README.md
```diff
@@ -275,6 +276,16 @@ metadata:
     capacity.cluster-autoscaler.kubernetes.io/taints: "key1=value1:NoSchedule,key2=value2:NoExecute"
 ```
 
+#### CPU Architecture awareness for single-arch clusters 
+
+Users of non-amd64, single-arch clusters willing to enable scale from zero 
+support should also set the `--scale-up-from-zero-default-arch` flag to the 
+architecture of the nodes they want to default the node group templates to.
+The autoscaler will default to amd64 if the flag is not set, and the node 
+group templates will not match the nodes' architecture, particularly when
```
elmiko, line 285: minor change

```suggestion
group templates will not match the nodes' architecture, specifically when 
```

elmiko (commented):
i think this is looking really nice @aleskandro , much easier to review. i also like that you updated the PR description, hopefully we could get more providers to adopt this logic.

i just have some suggestions for the readme, but otherwise i'm happy to approve this. we will also need to get approval for the flag changes and cpu module addition.

cluster-autoscaler/utils/cpu/architecture.go
```diff
@@ -0,0 +1,108 @@
...
+// Cloud providers willing to opt into this implementation are expected to change their manager code by replacing
+// the cloudprovider.DefaultArch constant with the GetDefaultScaleFromZeroArchitecture function exposed by this package.
+// Usually, this should be done in the function responsible for generating the generic labels of the given
+// cloud provider's manager code (for example, buildGenericLabels(.)) as follows:
+//
+//	func (...) buildGenericLabels(...) (map[string]string) {
+//		result := make(map[string]string)
+//		result[apiv1.LabelArchStable] = cpu.GetDefaultScaleFromZeroArchitecture().Name()
+//		// ...
+//	}
```
elmiko, line 64: :heart:

aleskandro:
> i think this is looking really nice @aleskandro , much easier to review. i also like that you updated the PR description, hopefully we could get more providers to adopt this logic.
> 
> i just have some suggestions for the readme, but otherwise i'm happy to approve this. we will also need to get approval for the flag changes and cpu module addition.

Hi @elmiko , thanks again. PTAL when you get the bandwidth for it.

elmiko (commented):
/lgtm

i like the comments in the new cpu util package, makes it clear how others can participate. if the sig is ok with adding the new flag i think this is great.

elmiko (commented):
thanks @aleskandro 

/lgtm

elmiko (commented):
oops, forgot to add approval from my side

/approve

we will need a core reviewer to approve the util stuff.

BigDarkClown:
Really my only comment is that this feature seems *very* cloud provider specific. I'm not sure how useful it will be to other cloud providers (different architectures etc.), so I'm a bit iffy on placing this architecture defaulting logic in the shared space.

Additionally, the CA is following the practise of specifying all the component-wide flags in the `main.go`. I'm very far from arguing that this is in any way superior to the method you used (I actually think it's worse), but for the sake of consistency it would be nice to add the flag in already established space, just so it does not get lost in the future.

elmiko:
> Really my only comment is that this feature seems very cloud provider specific. I'm not sure how useful it will be to other cloud providers (different architectures etc.), so I'm a bit iffy on placing this architecture defaulting logic in the shared space.

i think that's fair, we had initially thought that other providers might want to use the same defaulting, but if this should just be a clusterapi change then we could certainly use an environment variable (as we have done with other clusterapi specific config) instead of the flag. that would allow us to move all the code into the clusterapi directory.

good point about the flag.

aleskandro:
> Additionally, the CA is following the practise of specifying all the component-wide flags in the main.go. I'm very far from arguing that this is in any way superior to the method you used (I actually think it's worse), but for the sake of consistency it would be nice to add the flag in already established space, just so it does not get lost in the future.

Yeah, I imagined this but we wanted to add it in the review discussion as adding the flag to the main and exposing the variable as public to all the packages is far from the ideal too IMHO.

> > Really my only comment is that this feature seems very cloud provider specific. I'm not sure how useful it will be to other cloud providers (different architectures etc.), so I'm a bit iffy on placing this architecture defaulting logic in the shared space.
> 
> i think that's fair, we had initially thought that other providers might want to use the same defaulting, but if this should just be a clusterapi change then we could certainly use an environment variable (as we have done with other clusterapi specific config) instead of the flag. that would allow us to move all the code into the clusterapi directory.
> 
> good point about the flag.

This issue should affect all the cloud providers but GCE currently. However yes, other providers could decide for a different way of fixing it. I hoped to provide a more general solution.

Changed the implementation to be cluster api specific and using an env var.

elmiko:
thanks @aleskandro , that will make it easier to merge.

i'm fine to merge this PR as only cluster-api, but if you wanted to try and build a more general solution perhaps it would be appropriate to make a new [proposal](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/proposals) ?

then we could debate the implementation on the proposal, and we would have a central document about how to implement the change (even though i loved your code comments).

aleskandro:
Ok @elmiko, would we merge this first and run the proposal as a follow up, keeping the issue open?

elmiko:
good question @aleskandro , i think we should definitely merge this and then create the proposal. given that the issue is wider in scope than just clusterapi, i'd say yes we probably want to keep it open.

cluster-autoscaler/cloudprovider/clusterapi/README.md
```diff
@@ -276,6 +277,16 @@ metadata:
     capacity.cluster-autoscaler.kubernetes.io/taints: "key1=value1:NoSchedule,key2=value2:NoExecute"
 ```
 
+#### CPU Architecture awareness for single-arch clusters 
+
+Users of single-arch non-amd64 clusters who are using scale from zero 
+support should also set the `CAPI_SCALE_ZERO_DEFAULT_ARCH` environment variable
+to the architecture of the nodes they want to default the node group templates to.
```
elmiko, line 284: minor nit, missing a word

```suggestion
to set the architecture of the nodes they want to default the node group templates to.
```

cluster-autoscaler/cloudprovider/clusterapi/clusterapi_utils.go
```diff
@@ -279,3 +311,37 @@ func getClusterNameLabel() string {
...
+	}
+}
+
+// GetDefaultScaleFromZeroArchitecture returns the SystemArchitecture from the environment variable
+// CAPI_SCALE_ZERO_DEFAULT_ARCH or DefaultArch if the variable is set to an invalid value.
+func GetDefaultScaleFromZeroArchitecture() SystemArchitecture {
+	once.Do(func() {
+		archStr := os.Getenv(scaleUpFromZeroDefaultArchEnvVar)
+		arch := SystemArchitectureFromString(archStr)
+		klog.V(5).Infof("the  value is set to %s (%s)", scaleUpFromZeroDefaultArchEnvVar, archStr, arch.Name())
```
elmiko, line 338: i think we should add little extra detail to this log

```suggestion
		klog.V(5).Infof("the default scale from zero architecture value is set to %s (%s)", scaleUpFromZeroDefaultArchEnvVar, archStr, arch.Name())
```

elmiko (commented):
2 minor nits them i'm ready to merge @aleskandro

elmiko (commented):
thanks @aleskandro 

/approve
/lgtm

PR #6077, opened 2023-08-29
Process annotation marking pod as not safe to evict first

### LINKED ISSUE #5434: "Pod is blocking scale down because it’s not backed by a controller" but the pod is a jenkins agent running a job

Cluster autoscaler

This is my use case, jenkins agents created to run jobs cause the autoscaler rise a warning that a pod is blocking the scale down because is not backed by a controller, those agents eventually will finish the job and the autoscaler will be able to shutdown the node. Is not a big deal but is annoying to see that warning time to time and have to check if it is real or not

An annotation like "cluster-autoscaler.kubernetes.io/safe-to-evict": "true" to said to the autoscaler to wait or ignore the pod without controller and not shutdown the node

lauraseidler:
#### What type of PR is this?

/kind feature

By processing this annotation first, pods that would match also other, subsequent rules (e.g. pods without a controller), will generate the `NotSafeToEvictAnnotation` instead of e.g. `NotReplicated`.

This allows marking pods as not safe to evict to change the warning to one that can be more effectively ignored, e.g. for ephemeral build pods that are known to not be backed by a controller, and should not be evicted.

Fixes #5434

Unsure if this counts, please advise :) 

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/utils/drain/drain.go
+++ b/cluster-autoscaler/utils/drain/drain.go
@@ -127,6 +127,9 @@ func GetPodsForDeletionOnNodeDrain(
 		}
 
 		if !safeToEvict && !terminal {
+			if hasNotSafeToEvictAnnotation(pod) {
+				return []*apiv1.Pod{}, []*apiv1.Pod{}, &BlockingPod{Pod: pod, Reason: NotSafeToEvictAnnotation}, fmt.Errorf("pod annotated as not safe to evict present: %s", pod.Name)
+			}
 			if !replicated {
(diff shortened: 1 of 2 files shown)


towca:
The change is slightly user-facing, but the warnings are not generally expected to be a stable interface anyway. If you want, you can edit the description to include something like 'Pods with the "cluster-autoscaler.kubernetes.io/safe-to-evict": "true" annotation will now always report an annotation-related warning message if they block scale-down (where previously they might've reported e.g. a message about not being replicated).'

/lgtm
/approve

PR #6092, opened 2023-09-06
Use same Store for ScheduledPodLister and ScheduledAndUnschedulablePodLister

linxiulei:
#### What type of PR is this?

/kind cleanup

Fixes #

Using same Store will reduce memory usage since two pod listers overlap in pod list

```release-note
NONE
```


x13n:
I think we should just drop `ScheduledPodLister`, otherwise we're going to see race conditions when code tries to separately list scheduled and unschedulable pods.

/close

PR #6093, opened 2023-09-07
Add additional error codes for invalid reservations to GCE client

hbostan:
#### What type of PR is this?

/kind feature

Adds more error codes related to the reservations increasing the visibility of GCE Errors.

```release-note
NONE
```

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
+++ b/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
@@ -72,6 +72,22 @@ const (
 	ErrorCodeOther = "OTHER"
 )
 
+var (
+	regexReservationErrors = []*regexp.Regexp{
+		regexp.MustCompile("Incompatible AggregateReservation VMFamily"),
+		regexp.MustCompile("Could not find the given reservation with the following name"),
+		regexp.MustCompile("must use ReservationAffinity of"),
+		regexp.MustCompile("The reservation must exist in the same project as the instance"),
+		regexp.MustCompile("only compatible with Aggregate Reservations"),
+		regexp.MustCompile("Please target a reservation with workload_type ="),
+		regexp.MustCompile("AggregateReservation VMFamily: should be a (.*) VM Family for instance with (.*) machine type"),
+		regexp.MustCompile("VM Family: (.*) is not supported for aggregate reservations. It must be one of"),
+		regexp.MustCompile("Reservation (.*) is incorrect for the requested resources"),
+		regexp.MustCompile("Zone does not currently have sufficient capacity for the requested resources"),
+		regexp.MustCompile("Reservation (.*) does not have sufficient capacity for the requested resources."),
+	}
+)
+
 // AutoscalingGceClient is used for communicating with GCE API.
 type AutoscalingGceClient interface {
 	// reading resources
@@ -453,15 +469,8 @@ func isReservationNotReady(errorCode, errorMessage string) bool {
 }
 
 func isInvalidReservationError(errorCode, errorMessage string) bool {
-	reservationErrors := []string{
-		"Incompatible AggregateReservation VMFamily",
-		"Could not find the given reservation with the following name",
-		"must use ReservationAffinity of",
-		"The reservation must exist in the same project as the instance",
-		"only compatible with Aggregate Reservations",
-	}
-	for _, rErr := range reservationErrors {
-		if strings.Contains(errorMessage, rErr) {
+	for _, re := range regexReservationErrors {
+		if re.MatchString(errorMessage) {
 			return true
 		}
 	}


cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -453,15 +453,18 @@ func isReservationNotReady(errorCode, errorMessage string) bool {
...
-		"only compatible with Aggregate Reservations",
+	regexReservationErrors := []*regexp.Regexp{
+		regexp.MustCompile("Incompatible AggregateReservation VMFamily"),
+		regexp.MustCompile("Could not find the given reservation with the following name"),
+		regexp.MustCompile("must use ReservationAffinity of"),
+		regexp.MustCompile("The reservation must exist in the same project as the instance"),
+		regexp.MustCompile("only compatible with Aggregate Reservations"),
+		regexp.MustCompile("instance configuration is optimized for ([a-z0-9*-]+) workloads"),
+		regexp.MustCompile("should be a ([a-zA-Z0-9*-]+) VM Family for instance with ([a-zA-Z0-9*\\ -]*) machine type"),
+		regexp.MustCompile("is not supported for aggregate reservations. It must be one of"),
```
kawych, line 464: nit: can you check this with a prefix "VM Family: %s "? Makes it a little bit clearer

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -453,15 +453,18 @@ func isReservationNotReady(errorCode, errorMessage string) bool {
...
-		"must use ReservationAffinity of",
-		"The reservation must exist in the same project as the instance",
-		"only compatible with Aggregate Reservations",
+	regexReservationErrors := []*regexp.Regexp{
+		regexp.MustCompile("Incompatible AggregateReservation VMFamily"),
+		regexp.MustCompile("Could not find the given reservation with the following name"),
+		regexp.MustCompile("must use ReservationAffinity of"),
+		regexp.MustCompile("The reservation must exist in the same project as the instance"),
+		regexp.MustCompile("only compatible with Aggregate Reservations"),
+		regexp.MustCompile("instance configuration is optimized for ([a-z0-9*-]+) workloads"),
```
kawych, line 462: Please use the other part of the error message: "Please target a reservation with workload_type" (makes it clear that the error is about reservations and what's it about).

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -453,15 +453,18 @@ func isReservationNotReady(errorCode, errorMessage string) bool {
...
+		regexp.MustCompile("should be a ([a-zA-Z0-9*-]+) VM Family for instance with ([a-zA-Z0-9*\\ -]*) machine type"),
```
kawych, line 463: nit: can you check this with a prefix "AggregateReservation VMFamily: %s?

kawych (changes requested):

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
hbostan: Done

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
hbostan: Done

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
hbostan: Done

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -453,15 +453,21 @@ func isReservationNotReady(errorCode, errorMessage string) bool {
 }
 
 func isInvalidReservationError(errorCode, errorMessage string) bool {
-	reservationErrors := []string{
-		"Incompatible AggregateReservation VMFamily",
-		"Could not find the given reservation with the following name",
-		"must use ReservationAffinity of",
-		"The reservation must exist in the same project as the instance",
-		"only compatible with Aggregate Reservations",
+	regexReservationErrors := []*regexp.Regexp{
```
x13n, line 456: This list is static. Can we compile the regexps once, rather than on every call to this function?

hbostan: Moved the regexp slice to a `var` block.

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -453,15 +469,8 @@ func isReservationNotReady(errorCode, errorMessage string) bool {
...
-		"Incompatible AggregateReservation VMFamily",
-		"Could not find the given reservation with the following name",
-		"must use ReservationAffinity of",
-		"The reservation must exist in the same project as the instance",
-		"only compatible with Aggregate Reservations",
-	}
-	for _, rErr := range reservationErrors {
-		if strings.Contains(errorMessage, rErr) {
+	for _, re := range regexReservationErrors {
+		if re.MatchString(errorMessage) {
```
jayantjain93, line 473: I'm concerned that this change will make the matching more restrictive.
Old checkers uses a contains logic, wheres now we're doing a full-match (I think).
This could potentially break existing logic where we would have stacktrace to the errorMessage?

Can we explore if we can do a regex contains?

PR #6103, opened 2023-09-11
Refactor ScaleDownSet processor into a composite processor

kawych:
Split ScaleDownSet processor into two processors doing their individual tasks. This allows more flexibility when overriding these processors, such as adding another targeted processor to the pipeline, or removing one of the processors if it's not needed.

CC @towca 

```release-note
NONE
```

### DIFF
--- a/cluster-autoscaler/core/test/common.go
+++ b/cluster-autoscaler/core/test/common.go
@@ -178,7 +178,10 @@ func NewTestProcessors(context *context.AutoscalingContext) *processors.Autoscal
 		NodeGroupListProcessor: &nodegroups.NoOpNodeGroupListProcessor{},
 		BinpackingLimiter:      binpacking.NewDefaultBinpackingLimiter(),
 		NodeGroupSetProcessor:  nodegroupset.NewDefaultNodeGroupSetProcessor([]string{}, config.NodeGroupDifferenceRatios{}),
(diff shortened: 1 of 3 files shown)


towca:
/lgtm
/approve

Thanks for making the change!

cluster-autoscaler/core/test/common.go
```diff
@@ -178,7 +178,10 @@ func NewTestProcessors(context *context.AutoscalingContext) *processors.Autoscal
 		NodeGroupListProcessor: &nodegroups.NoOpNodeGroupListProcessor{},
 		BinpackingLimiter:      binpacking.NewDefaultBinpackingLimiter(),
 		NodeGroupSetProcessor:  nodegroupset.NewDefaultNodeGroupSetProcessor([]string{}, config.NodeGroupDifferenceRatios{}),
-		ScaleDownSetProcessor:  nodes.NewPostFilteringScaleDownNodeProcessor(),
+		ScaleDownSetProcessor: nodes.NewCompositeScaleDownSetProcessor([]nodes.ScaleDownSetProcessor{
```
towca, line 181: Oops, missed this: why is the order different for test processors? Was this intended? I think we always want to call the max nodes one first so that we avoid breaking the atomic property (could be worth capturing that in a comment somewhere btw)?

PR #6104, opened 2023-09-13
Prov req client

kisieland:
#### What type of PR is this?
/kind feature

This PR adds a definition of Provisioning Request that was proposed in https://github.com/kubernetes/autoscaler/pull/5848.
PR also adds mechanisms to generate a CRD yaml file and client to interact with the Provisioning Request objects.

N/A

The PR consists of 3 commints:
1. Adds the type definition (main addition of the PR)
2. Adds tooling to generate CRD manifest and clients
3. Adds generated manifest and clients

```release-note
NONE
```

```docs

```


mwielgus (approved):
/lgtm
/approve

cluster-autoscaler/config/crd/provisioning.k8s.io_provisioningrequests.yaml
```diff
@@ -0,0 +1,203 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    api-approved.kubernetes.io: https://github.com/kubernetes/autoscaler/pull/5848
```
liggitt, line 6: commented at https://github.com/kubernetes/autoscaler/pull/5848#issuecomment-1719515802 but wanted to signal boost here... that PR didn't get API approval yet – we're reviewing the updates today.

If you want to merge ahead of API approval, you can work in the `autoscaling.x-k8s.io` API group with only SIG oversight, which explicitly marks the API as experimental and doesn't require API approval.

cluster-autoscaler/config/crd/provisioning.k8s.io_provisioningrequests.yaml
```diff
@@ -0,0 +1,203 @@
...
+    controller-gen.kubebuilder.io/version: v0.13.0
+  name: provisioningrequests.provisioning.k8s.io
```
liggitt, line 8: The `provisioning.k8s.io` API group looks novel. What else would go in this group, would it only contain autoscaling APIs?

cluster-autoscaler/config/crd/provisioning.k8s.io_provisioningrequests.yaml
```diff
@@ -0,0 +1,203 @@
...
+    - provreq
+    - provreqs
+    singular: provisioningrequest
+  scope: Namespaced
+  versions:
+  - additionalPrinterColumns:
+    - jsonPath: .metadata.creationTimestamp
+      name: Age
+      type: date
+    name: v1beta1
```
liggitt, line 25: This should not start as beta, that carries forward compatibility guarantees for k8s APIs.

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
```diff
@@ -0,0 +1,173 @@
...
+	// +kubebuilder:validation:Maximum=16384
+	Count int32 `json:"count"`
+}
+
+// Reference represents reference to an object within the same namespace.
+type Reference struct {
+	// Name of the referenced object.
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names#names
+	//
+	// +kubebuilder:validation:Required
```
liggitt, line 125: this could add validation to match what is allowed for pod template names (length, dns1123 format, etc) to keep a typo here from failing the lookup

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
```diff
@@ -0,0 +1,173 @@
...
+// ProvisioningRequestSpec is a specification of additional pods for which we
+// would like to provision additional resources in the cluster.
+type ProvisioningRequestSpec struct {
+	// PodSets lists groups of pods for which we would like to provision
+	// resources.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=32
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
```
liggitt, line 82: from https://github.com/kubernetes/autoscaler/pull/5848#discussion_r1297456283

if these fields are immutable, changing requires deleting / recreating? would we expect someone to delete and recreate rapidly with the same name, or issue a delete of one request and create a different one with a different name?

if delete / recreate rapidly with the same name, does the autoscaler notice and handle the uid and shape of immutable fields changing on an update?

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
```diff
@@ -0,0 +1,173 @@
...
+	// Users need to make sure that the  fields relevant to scheduler (e.g. node selector tolerations)
+	// are consistent between this template and actual pods consuming the Provisioning Request.
+	//
+	// +kubebuilder:validation:Required
+	PodTemplateRef Reference `json:"podTemplateRef"`
+	// Count contains the number of pods that will be created with a given
+	// template.
+	//
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=16384
```
liggitt, line 116: a max is good, 2^14 seems random... keeping this within int32 is necessary for correctness. is there a more principled limit we could place here (max nodes in ecosystem * max pods per node)?

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
```diff
@@ -0,0 +1,173 @@
...
+	// +kubebuilder:validation:MaxItems=32
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	PodSets []PodSet `json:"podSets"`
+
+	// ProvisioningClass describes the different modes of provisioning the resources.
+	// Supported values:
+	// * check-capacity.kubernetes.io - check if current cluster state can fullfil this request,
+	//   do not reserve the capacity.
+	// * atomic-scale-up.kubernetes.io - provision the resources in an atomic manner
+	// * ... - potential other classes that are specific to the cloud providers
```
liggitt, line 90: everywhere else we have a *Class field that points at an object, we use `*ClassName`

making this *sometimes* point at an API object and sometimes not will be tricky to understand if the intent is to use this for ProvisioningClass instances later... is there a way to make that clearer here in docs?

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
```diff
@@ -0,0 +1,173 @@
...
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
```
liggitt, line 93: add validation to ensure this stays a valid custom resource object name (IsDNS1123Subdomain) and has multiple `.`-delimited segments to ensure people use qualified names from the start

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
```diff
@@ -0,0 +1,173 @@
...
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=32
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	PodSets []PodSet `json:"podSets"`
+
+	// ProvisioningClass describes the different modes of provisioning the resources.
+	// Supported values:
+	// * check-capacity.kubernetes.io - check if current cluster state can fullfil this request,
+	//   do not reserve the capacity.
+	// * atomic-scale-up.kubernetes.io - provision the resources in an atomic manner
```
liggitt, line 89: probably worth repeating in documentation here that `.kubernetes.io` is reserved for Kubernetes-project-defined modes

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
```diff
@@ -0,0 +1,173 @@
...
+	// * ... - potential other classes that are specific to the cloud providers
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	ProvisioningClass string `json:"provisioningClass"`
+
+	// AdditionalParameters contains all other parameters custom classes may require.
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
```
liggitt, line 99: TODO to limit number of entries (e.g. 100), length of keys and values (e.g. 255) to keep the size bounded

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
```diff
@@ -0,0 +1,173 @@
...
+	// +listMapKey=type
+	// +patchStrategy=merge
+	// +patchMergeKey=type
+	// +optional
+	Conditions []metav1.Condition `json:"conditions"`
+
+	// AdditionalStatus contains all other status values custom provisioning classes may require.
+	//
+	// +optional
+	// +kubebuilder:validation:MinProperties=64
```
liggitt, line 146: should this be MaxProperties or are we requiring all setters to populate 64 values here :)

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
```diff
@@ -0,0 +1,173 @@
...
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	ProvisioningClass string `json:"provisioningClass"`
+
+	// AdditionalParameters contains all other parameters custom classes may require.
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	AdditionalParameters map[string]string `json:"additionalParameters"`
```
liggitt, line 100: I'd suggest calling this `parameters` rather than `additionalParameters`, since we didn't put any first-class parameter fields in the spec

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
```diff
@@ -0,0 +1,173 @@
...
+	// +patchStrategy=merge
+	// +patchMergeKey=type
+	// +optional
+	Conditions []metav1.Condition `json:"conditions"`
+
+	// AdditionalStatus contains all other status values custom provisioning classes may require.
+	//
+	// +optional
+	// +kubebuilder:validation:MinProperties=64
+	AdditionalStatus map[string]string `json:"additionalStatus"`
```
liggitt, line 147: do you have examples of what would go in this field that couldn't fit into `status.conditions`? we don't have any other examples of free-form status map[string]string extension points I can think of

couldn't everything you would put in this map be modeled as a condition where the key is the type and the value is the reason?

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
liggitt, line 89: the documentation of these behaviors (including timeout, failure modes, expected conditions, supported parameter name/values) should be included here... users expect `kubectl explain --recursive provisioningrequest` to tell them how to use the API, not to need to reference the design doc

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
kisieland: Added, seems there is no kube validation to do it out of the box, so I've added it manually.

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
kisieland: Added to docs

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
kisieland: Added, though I did not see a way for us to limit the length of the keys, only entries could be capped.

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
kisieland: Done

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
kisieland: Example of information that could be passed here for `atomic-scale-up.kubernetes.io` is a number of retries that were executed so far, this is something that wouldn't make sense to export as an condition.

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
kisieland: Updated.

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
kisieland: Updated to ProvisioningClassName, but clarified that `ProvisioningClass` object are not yet supported. PTAL

cluster-autoscaler/provisioningrequests/apis/provisioning.k8s.io/v1beta1/types.go
kisieland: Updated

PR #6108, opened 2023-09-14
Fix the api group to x-k8s.io

kisieland:
#### What type of PR is this?
/kind bug

Changes the API from `provisioning.k8s.io` to `provisioning.x-k8s.io` which was the initial aim

Fixes #

```release-note

```

```docs

```


mwielgus (approved):
/lgtm
/approve

cluster-autoscaler/config/crd/provisioning.x-k8s.io_provisioningrequests.yaml
```diff
@@ -3,11 +3,10 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    api-approved.kubernetes.io: https://github.com/kubernetes/autoscaler/pull/5848
     controller-gen.kubebuilder.io/version: v0.13.0
-  name: provisioningrequests.provisioning.k8s.io
+  name: provisioningrequests.provisioning.x-k8s.io
 spec:
-  group: provisioning.k8s.io
+  group: provisioning.x-k8s.io
```
liggitt, line 9: provisioning is pretty generic... if this will only contain autoscaling-owned APIs, consider making a more focused group name

cluster-autoscaler/provisioningrequests/apis/provisioning.x-k8s.io/v1beta1/register.go
```diff
@@ -25,7 +25,7 @@ import (
 
 const (
 	// GroupName represents the group name for ProvisioningRequest resources.
-	GroupName = "provisioning.k8s.io"
+	GroupName = "provisioning.x-k8s.io"
 	// GroupVersion represents the group name for ProvisioningRequest resources.
 	GroupVersion = "v1beta1"
```
liggitt, line 30: I'd still recommend you start with an alpha to avoid committing yourself to forward compatibility without any feedback

mwielgus: We want to start integration with Kueue soon, so the commitment to the API is happening right now. We also ran the API on GKE as a part of early access program to some features and the customers were OK with it.

cluster-autoscaler/config/crd/provisioning.x-k8s.io_provisioningrequests.yaml
mwielgus: Do you have any naming suggestions (assuming that nothing more gets into the group in the near future)?

liggitt: single-resource API groups are sort of a smell, as is repeating the concept in the type name (ProvisioningRequest) and the group name (provisioning...).

if this is owned by sig-autoscaling, `autoscaling.x-k8s.io` doesn't seem bad

kisieland: Renamed to `autoscaling.x-k8s.io`

mwielgus (approved):
Comments addressed, removing hold.

mwielgus (approved):
/lgtm

PR #6114, opened 2023-09-15
Allow setting content-type in command

linxiulei:
#### What type of PR is this?

/kind feature

Allow setting content-type in command line, which might be preferable for some users to use protobuf than json

#### Does this PR introduce a user-facing change

For more information on release notes see: https://git.k8s.io/community/contributors/guide/release-notes.md
-->
```release-note
New command line option `--kube-api-content-type`  is added to specify content type to communicate with apiserver
```


cluster-autoscaler/main.go
```diff
@@ -101,6 +101,7 @@ var (
 	address                 = flag.String("address", ":8085", "The address to expose prometheus metrics.")
 	kubernetes              = flag.String("kubernetes", "", "Kubernetes master location. Leave blank for default")
 	kubeConfigFile          = flag.String("kubeconfig", "", "Path to kubeconfig file with authorization and master location information.")
+	kubeAPIContentType      = flag.String("kube-api-content-type", "", "Content type of requests sent to apiserver.")
```
linxiulei, line 104: Do we want to make a change in default behavior here to make it protobuf?

linxiulei:
> I think it should be possible already to set content type through kubeconfig: https://github.com/kubernetes/client-go/blob/master/rest/config.go#L290 Isn't this sufficient?

unfortunnately, kubeconfig is parsed with this Config type: https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/api/types.go#L31, which has no content type field

cluster-autoscaler/main.go
```diff
@@ -408,6 +409,8 @@ func getKubeConfig() *rest.Config {
 		klog.Fatalf("Failed to build Kubernetes client configuration: %v", err)
 	}
 
+	kubeConfig.ContentType = *kubeAPIContentType
```
x13n, line 412: I understand `ContentType` is empty now, but it could change in the future, so maybe let's only use the flag value if it was actually provided?

linxiulei: I do like to keep the consistency for CA with other k8s components in using this flag. 

`--kube-api-content-type string     Default: application/vnd.kubernetes.protobuf`

(eg. https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/)

If it ever changes, maybe change this part in the future? alternatively we can specify a default value in `kubeAPIContentType`.

cluster-autoscaler/main.go
linxiulei: Changed default behavior to use protobuf

cluster-autoscaler/main.go
linxiulei: Done, thanks!

PR #6116, opened 2023-09-18
add selectorLabels

nalbam:
#### What type of PR is this?

/kind feature

/kind helm-chart

Once deployed, if you change the `podLabels`, the next deployment will not match `selector.matchLabels` and will throw an error. 
So I added `selectorLabels` to create them separately.

```release-note
NONE
```

```docs

```

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -11,4 +11,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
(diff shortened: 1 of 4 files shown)


vadasambar:
> Once deployed, if you change the podLabels, the next deployment will not match selector.matchLabels and will throw an error.
> So I added selectorLabels to create them separately.

We want `podLabels` to be the same as `selector.matchLabels` because we want the cluster-autoscaler Deployment to always own the cluster-autoscaler pods (pod labels and `selector.matchLabels` need to match for this). 

What error are we talking about? Is there a particular scenario which causes a problem? I am not sure I understand the problem here.

PR #6122, opened 2023-09-20
fix incomplete startup of informers + broken scale down

pohly:
#### What type of PR is this?

/kind bug

Previously, SharedInformerFactory.Start was called before core.NewAutoscaler. That had the effect that any new informer created as part of core.NewAutoscaler, in particular in kubernetes.NewListerRegistryWithDefaultListers, never got started.

One of them was the DaemonSet informer. This had the effect that the DaemonSet lister had an empty cache and scale down failed with:

    I0920 11:06:36.046889   31805 cluster.go:164] node gke-cluster-pohly-default-pool-c9f60a43-5rvz cannot be removed: daemonset for kube-system/pdcsi-node-7hnmc is not present, err: daemonset.apps "pdcsi-node" not found

This was on a GKE cluster with cluster-autoscaler running outside of the cluster on a development machine. I have no idea which "real" deployments of cluster-autoscaler might be affected by this.

```release-note
Scale down may have failed with "daemonset.apps not found".
```

### DIFF
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -498,16 +498,23 @@ func buildAutoscaler(debuggingSnapshotter debuggingsnapshot.DebuggingSnapshotter
 		Comparator: nodeInfoComparator,
 	}
 
-	stop := make(chan struct{})
-	informerFactory.Start(stop)
-
 	// These metrics should be published only once.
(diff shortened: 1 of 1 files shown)


pohly:
Probably broken since 0f02235e9826e91800067523930f00107460cad2, i.e. fairly recently (five days ago).

/cc @linxiulei

x13n:
Thanks for the fix!

/lgtm
/approve

Issue #6113, opened 2023-09-15
VPA release 1.0.0


raywainman:
Following instructions in [RELEASE.md](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/RELEASE.md).

Starting this release as recommended by @jbartosik. This corresponds to the k8s 1.28 release.

Please provide any information that is related to the release:

- When we plan to do the release?
   - Now
- Are there any issues / PRs blocking the release?
   - TBD

aadyyb:
is there a plan to publish an official helm chart? There is issue open (https://github.com/kubernetes/autoscaler/issues/5369) with reference to a few of community options, but we are challenged by our security to use 'unofficial' helm chart.

jbartosik:
> is there a plan to publish an official helm chart? There is issue open (#5369) with reference to a few of community options, but we are challenged by our security to use 'unofficial' helm chart.

Like I mentioned in [a comment on the issue](https://github.com/kubernetes/autoscaler/issues/5369#issuecomment-1408733510) I think we can accept contributions but don't have capacity to maintain them.

As far as I know none contributed helm charts for VPA specifically (one can [use CA helm chart to deploy VPA](https://github.com/kubernetes/autoscaler/tree/1a6b1e9a65521e5f3ec3d62a0406efaab879b5f3/charts/cluster-autoscaler#verticalpodautoscaler)).

jbartosik:
I created new branch: https://github.com/kubernetes/autoscaler/tree/vpa-release-1.0

BTW we should update release documentation to refer to 1.n instead of 0.n from now on

Shubham82:
I think we have to backport #6151 into vpa-release-1.0
@raywainman @jbartosik WDYT?

If seems good then I will open a PR for it.

Shubham82:
> @Shubham82 seems totally reasonable, sgtm

Thanks @raywainman, will raise PR soon.

Shubham82:
> > @Shubham82 seems totally reasonable, sgtm
> 
> Thanks @raywainman, will raise PR soon.

Fixed it here: #6160

Ubun1:
hi all!
Can someone compile changelog with user facing features for 1.0.0 release ?

jbartosik:
Yes, we're going to. It's taking a bit longer than I expected, sorry about that.

kwiesmueller:
[...]

PR #6149, opened 2023-09-26
Fix duplicate -addext when generating certificates with admission-controller/gencerts.sh

raywainman:
#### What type of PR is this?

/kind bug

The autoscaler/vertical-pod-autoscaler/pkg/admission-controller/gencerts.sh script fails with the following error:

```
Generating certs for the VPA Admission Controller in /tmp/vpa-certs.
req: Skipping unknown subject name attribute "/CN"
Error adding extensions defined via -addext
40C7A2C93F7F0000:error:0580008C:x509 certificate routines:X509at_add1_attr:duplicate attribute:../crypto/x509/x509_att.c:86:
```

This is because https://github.com/kubernetes/autoscaler/blob/cf8c507d2421e7ed8b8d6f9ae44e9a624e221c43/vertical-pod-autoscaler/pkg/admission-controller/gencerts.sh#L54 adds a `-addext` argument that duplicates what is already added in https://github.com/kubernetes/autoscaler/blob/cf8c507d2421e7ed8b8d6f9ae44e9a624e221c43/vertical-pod-autoscaler/pkg/admission-controller/gencerts.sh#L38.

Perhaps this used to work because openSSL ignored duplicate entries? However, in the latest version of OpenSSL I'm using (3.0.10) this causes an error and prevents the script from finishing.

Part of #6113, this fix was needed to unblock release testing.

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/gencerts.sh
+++ b/vertical-pod-autoscaler/pkg/admission-controller/gencerts.sh
@@ -51,7 +51,7 @@ set -o errexit
 # Create a server certificate
 openssl genrsa -out ${TMP_DIR}/serverKey.pem 2048
 # Note the CN is the DNS name of the service of the webhook.
-openssl req -new -key ${TMP_DIR}/serverKey.pem -out ${TMP_DIR}/server.csr -subj "/CN=vpa-webhook.kube-system.svc" -config ${TMP_DIR}/server.conf -addext "subjectAltName = DNS:vpa-webhook.kube-system.svc"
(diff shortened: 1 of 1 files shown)


jbartosik (approved):
/lgtm
/approve

Should we cherry pick this to the 1.0 branch?

We'll have to release 1.0.1 for this to be useful and 1.1 will be triggered on 2023-12-05 by K8s 1.29 release so I'm not sure if we'll have capacity to do the release but there's little harm in cherrypicking.

PR #6166, opened 2023-10-02
aws: check for all possible Terminated states

tomwans:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it:**
It's possible that an EC2 instance is in the Terminated state for multiple autoscaler loops, so to avoid trying to terminate it more than once and causing the bug seen for this initial fix, we include this check as well.

Also, make the log more verbose and write out what state we found the instance in.

This builds upon this change: https://github.com/kubernetes/autoscaler/pull/5411

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider_test.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider_test.go
@@ -504,6 +504,54 @@ func TestDeleteNodesTerminatingInstances(t *testing.T) {
 	assert.Equal(t, 2, newSize)
 }
 
+func TestDeleteNodesTerminatedInstances(t *testing.T) {
+	a := &autoScalingMock{}
+	provider := testProvider(t, newTestAwsManagerWithAsgs(t, a, nil, []string{"1:5:test-asg"}))
+	asgs := provider.NodeGroups()
+
+	a.On("TerminateInstanceInAutoScalingGroup", &autoscaling.TerminateInstanceInAutoScalingGroupInput{
+		InstanceId:                     aws.String("test-instance-id"),
+		ShouldDecrementDesiredCapacity: aws.Bool(true),
+	}).Return(&autoscaling.TerminateInstanceInAutoScalingGroupOutput{
+		Activity: &autoscaling.Activity{Description: aws.String("Deleted instance")},
+	})
+
+	// Look up the current number of instances...
+	var expectedInstancesCount int64 = 2
+	a.On("DescribeAutoScalingGroupsPages",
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/cloudprovider/aws/aws_cloud_provider_test.go
```diff
@@ -504,6 +504,54 @@ func TestDeleteNodesTerminatingInstances(t *testing.T) {
...
+		},
+	}
+	err = asgs[0].DeleteNodes([]*apiv1.Node{node})
+	assert.NoError(t, err)
+	a.AssertNumberOfCalls(t, "TerminateInstanceInAutoScalingGroup", 0) // instances which are terminating don't need to be terminated again
+	a.AssertNumberOfCalls(t, "DescribeAutoScalingGroupsPages", 1)
+
+	newSize, err := asgs[0].TargetSize()
+	assert.NoError(t, err)
+	assert.Equal(t, 2, newSize)
```
drmorr0, line 552: Why isn't this `expectedInstancesCount`?

tomwans: Going to double check, but I just copy pasted the tests from the last PR for this and added the new Terminated state; will double check both.

tomwans:
@drmorr0 thanks for the review. I cleaned up the test a bit, please let me know how it looks now. It just seems like the TerminatingWait test was copy pasted from something else, with light adjustments. The expectedInstanceCount variable is not really relevant, it was just a distraction. I reworded some of that from the Terminated test I added.

If you'd like, I can copy paste this and do the same for the TerminatingWait test, so they look the same. Or, I can just bring them all in to one test and do a table test. I have no preference. What do you think?

drmorr0:
> @drmorr0 thanks for the review. I cleaned up the test a bit, please let me know how it looks now. It just seems like the TerminatingWait test was copy pasted from something else, with light adjustments. The expectedInstanceCount variable is not really relevant, it was just a distraction. I reworded some of that from the Terminated test I added.
> 
> If you'd like, I can copy paste this and do the same for the TerminatingWait test, so they look the same. Or, I can just bring them all in to one test and do a table test, which might be fine since we expect the same result out of all the states anyway. I have no preference. What do you think?

@tomwans If you have some time and it's not terribly difficult, putting into a table test would be nice, I guess?  But I don't have particularly strong feelings on the matter, so if you don't have time/want to just ship it, I'll stamp it.

tomwans:
Let's just stamp it out. Theres no other case of table tests in this test file, so it wouldn't be matching conventions anyway.

PR #6169, opened 2023-10-03
Add HasNodeGroupStartedScaleUp to cluster state registry.

hbostan:
#### What type of PR is this?

/kind feature

Adds a new function to `ClusterStateRegistry` called `HasNodeGroupStartedScaleUp()` which returns whether the node group has a scale up request without checking if there are any upcoming nodes in the node group.

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -501,6 +501,17 @@ func (csr *ClusterStateRegistry) IsNodeGroupScalingUp(nodeGroupName string) bool
 	return found
 }
 
+// HasNodeGroupStartedScaleUp returns true if the node group has started scale up regardless
+// of whether there are any upcoming nodes. This is useful in the case when the node group's
+// size reverts back to its previous size before the next UpdatesCall and we want to know
+// if a scale up for node group has started.
+func (csr *ClusterStateRegistry) HasNodeGroupStartedScaleUp(nodeGroupName string) bool {
+	csr.Lock()
+	defer csr.Unlock()
+	_, found := csr.scaleUpRequests[nodeGroupName]
+	return found
+}
+
 // AcceptableRange contains information about acceptable size of a node group.
 type AcceptableRange struct {
 	// MinNodes is the minimum number of nodes in the group.
--- a/cluster-autoscaler/clusterstate/clusterstate_test.go
+++ b/cluster-autoscaler/clusterstate/clusterstate_test.go
@@ -122,6 +122,7 @@ func TestEmptyOK(t *testing.T) {
 	assert.Empty(t, clusterstate.GetScaleUpFailures())
 	assert.True(t, clusterstate.IsNodeGroupHealthy("ng1"))
 	assert.False(t, clusterstate.IsNodeGroupScalingUp("ng1"))
+	assert.False(t, clusterstate.HasNodeGroupStartedScaleUp("ng1"))
 
 	provider.AddNodeGroup("ng1", 0, 10, 3)
 	clusterstate.RegisterOrUpdateScaleUp(provider.GetNodeGroup("ng1"), 3, now.Add(-3*time.Second))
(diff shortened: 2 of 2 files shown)


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -501,6 +501,14 @@ func (csr *ClusterStateRegistry) IsNodeGroupScalingUp(nodeGroupName string) bool
 	return found
 }
 
+// HasNodeGroupStartedScaleUp returns true if the node group has started scale up.
+// Difference between IsNodeGroupScalingUp() is, this function doesn't check whether
+// there are any upcoming nodes for the node group.
+func (csr *ClusterStateRegistry) HasNodeGroupStartedScaleUp(nodeGroupName string) bool {
+	_, found := csr.scaleUpRequests[nodeGroupName]
```
x13n, line 508: Node group requests are cleared in https://github.com/kubernetes/autoscaler/blob/763c0e85492f57a5b51936767a95557c36fcc40e/cluster-autoscaler/clusterstate/clusterstate.go#L258, what semantics exactly are you looking for here?

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -501,6 +501,14 @@ func (csr *ClusterStateRegistry) IsNodeGroupScalingUp(nodeGroupName string) bool
 	return found
 }
 
+// HasNodeGroupStartedScaleUp returns true if the node group has started scale up.
+// Difference between IsNodeGroupScalingUp() is, this function doesn't check whether
```
x13n, line 505: This comment would be more useful if it explained why this is useful rather than describing how it is implemented.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -501,6 +501,14 @@ func (csr *ClusterStateRegistry) IsNodeGroupScalingUp(nodeGroupName string) bool
...
+// there are any upcoming nodes for the node group.
+func (csr *ClusterStateRegistry) HasNodeGroupStartedScaleUp(nodeGroupName string) bool {
```
x13n, line 507: Why is this not mutex protected? FWIW, looks like `IsNodeGroupScalingUp` should really be `isNodeGroupScalingUpNoLock` since it is used internally from mutex-protected public functions.

cluster-autoscaler/clusterstate/clusterstate_test.go
x13n, line 1: Please add a test case that would illustrate the difference betwen `IsNodeGroupScalingUp` and `HasNodeGroupStartedScaleUp`. As far as unit testing goes, they seem to be identical.

x13n (changes requested):

cluster-autoscaler/clusterstate/clusterstate_test.go
hbostan: Added a new test demonstrating the case where two methods differ.

cluster-autoscaler/clusterstate/clusterstate.go
hbostan: Added mutex protection, however left renaming the `IsNodeGroupScalingUp` to another pr.

cluster-autoscaler/clusterstate/clusterstate.go
hbostan: Reworded the comment

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -501,6 +501,17 @@ func (csr *ClusterStateRegistry) IsNodeGroupScalingUp(nodeGroupName string) bool
 	return found
 }
 
+// HasNodeGroupStartedScaleUp returns true if the node group has started scale up regarless
```
x13n, line 504: nit, typo: regardless

x13n (commented):
Looks good, just a minor nit.

/approve

PR #5820, opened 2023-05-30
feat: implement `kwok` cloud provider

vadasambar:
Signed-off-by: vadasambar <surajrbanakar@gmail.com>

#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:
This PR implements [`kwok`](https://kwok.sigs.k8s.io/) provider (`kwok` is a more performant alternative to kubemark; it doesn't require creation of a separate master/cluster for testing) for scale testing CA to identify and fix issues like https://github.com/kubernetes/autoscaler/issues/5769 

More more info, check the [README](https://github.com/vadasambar/autoscaler/blob/kwok-poc/cluster-autoscaler/cloudprovider/kwok/README.md)

#### Which issue(s) this PR fixes:

There is no issue for this PR. I can create one if needed. 

#### Does this PR introduce a user-facing change?

```release-note
Added: new kwok cloud provider (check https://github.com/kubernetes/autoscaler/blob/kwok-poc/cluster-autoscaler/cloudprovider/kwok/README.md) for more info
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #6180, opened 2023-10-08
refactor(*): move getKubeClient to utils/kubernetes

qianlei90:
#### What type of PR is this?

/kind cleanup

move getKubeClient to `utils/kubernetes` and make this function public.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -21,7 +21,6 @@ import (
 	"flag"
 	"fmt"
 	"net/http"
-	"net/url"
 	"os"
 	"os/signal"
 	"strconv"
@@ -63,7 +62,6 @@ import (
 	"k8s.io/client-go/informers"
 	kube_client "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/leaderelection"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	kube_flag "k8s.io/component-base/cli/flag"
@@ -353,7 +351,9 @@ func createAutoscalingOptions() config.AutoscalingOptions {
 		StatusTaints:                     *statusTaintsFlag,
 		BalancingExtraIgnoredLabels:      *balancingIgnoreLabelsFlag,
 		BalancingLabels:                  *balancingLabelsFlag,
+		Kubernetes:                       *kubernetes,
 		KubeConfigPath:                   *kubeConfigFile,
+		KubeAPIContentType:               *kubeAPIContentType,
 		KubeClientBurst:                  *kubeClientBurst,
 		KubeClientQPS:                    *kubeClientQPS,
 		NodeDeletionDelayTimeout:         *nodeDeletionDelayTimeout,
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/utils/kubernetes/client.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"k8s.io/autoscaler/cluster-autoscaler/config"
+
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
+)
+
+// GetKubeConfig returns the rest config from AutoscalingOptions.
+func GetKubeConfig(opts config.AutoscalingOptions) *rest.Config {
```
x13n, line 30: Passing all `AutoscalingOptions` here seems like overkill for this: it only requires a few values. WDYT about moving all kube-client related fields to a dedicated struct and only passing that struct here? If it is necessary in `AutoscalingOptions`, it can still be there.

cluster-autoscaler/main.go
```diff
@@ -455,7 +430,7 @@ func buildAutoscaler(debuggingSnapshotter debuggingsnapshot.DebuggingSnapshotter
 	}
 	informerFactory := informers.NewSharedInformerFactoryWithOptions(kubeClient, 0, informers.WithTransform(trim))
 
-	eventsKubeClient := createKubeClient(getKubeConfig())
+	eventsKubeClient := createKubeClient(kube_util.GetKubeConfig(autoscalingOptions))
```
x13n, line 433: Is `KubeConfig` abstraction needed at all in main? Can the entire client creation be moved to a dedicated package?

x13n (commented):
/assign

Thanks for the change! Left some comments, let me know what you think.

cluster-autoscaler/utils/kubernetes/client.go
qianlei90:
```go
type AutoscalingOptions struct {
    	...

    	// Kubernetes master location.
	Kubernetes string
	// Path to kube configuration if available
	KubeConfigPath string
	// Content type of requests sent to APIServer.
	KubeAPIContentType string
	// Burst setting for kubernetes client
	KubeClientBurst int
	// QPS setting for kubernetes client
	KubeClientQPS float64

    	...
}
```

Yeah I think it's very reasonable. How about moving these fields to a dedicated struct `KubeClientOptions` in *utils/kubernetes/client.go*？

I want to create a kube client in cloudprovider(see  https://github.com/kubernetes/autoscaler/pull/5820#discussion_r1349697340) but `buildCloudProvider` only accept `AutoscalingOptions`:

https://github.com/kubernetes/autoscaler/blob/e7bf3ec8405a2e87cc81ed8779cc8b447ada10cd/cluster-autoscaler/cloudprovider/builder/builder_all.go#L90-L90

If we move kube client releated fields and functions to *utils/kubernetes/client.go*, I'm wondering how to create kube client in cloudprovider.

cluster-autoscaler/main.go
qianlei90: Yes. releated fields and `getKubeConfig`、`createKubeClient` now are only used in `main.go`. I do this refactor because I want to reuse these functions in cloudprovider.

cluster-autoscaler/utils/kubernetes/client.go
x13n: I think the struct should still live next to `AutoscalingOptions` - this package has almost no dependencies and I'd like to keep it that way. Otherwise it would depend on the new file. (And transitively depend on anything else the new file imports.)

What I'm proposing is:

```go
type AutoscalingOptions struct {
	...
	KubeClient KubeClientOptions
	...
}

type KubeClientOptions struct {
 	// Kubernetes master location.
	Kubernetes string
 	// Path to kube configuration if available
 	KubeConfigPath string
 	// Content type of requests sent to APIServer.
 	ContentType string
 	// Burst setting for kubernetes client
 	BurstLimit int
 	// QPS setting for kubernetes client
 	QPSLimit float64
}
```

And then in the new package:

```go
package client

...

func New(opts config.KubeClientOptions) kube_client.Interface {
	...
}
```

With that, given `AutoscalingOptions`, you can still use the new utils to create kube client. Does that make sense?

vadasambar:
@qianlei90 any updates on this? I wanted to include changes in this PR in https://github.com/kubernetes/autoscaler/pull/5820. Since this is related to the kwok provider PR, I can help with the changes if you want. Other options I am thinking of:
1. Make code changes in https://github.com/kubernetes/autoscaler/pull/5820 to match this PR
2. Pull this PR into https://github.com/kubernetes/autoscaler/pull/5820 after it is approved and merged

Let me know what you think.

vadasambar:
@qianlei90 wanted to remind you about https://github.com/kubernetes/autoscaler/pull/6180#issuecomment-1786429682

vadasambar:
Seems like @qianlei90 might not be available to work on this. 
@x13n do you mind if I create a separate PR based off this. We can close this one and the review can be continued in the new PR. `@qianlei90` generously created this PR in response to the PR I am working on: https://github.com/kubernetes/autoscaler/pull/5820#discussion_r1349697340 
I would like to close the loop.

x13n:
Yup, makes sense. If you are going to follow up separately, I'll just close this one. Please assign me to the follow-up PR once it exists.

/close

PR #6184, opened 2023-10-10
feat: each node pool can now have different init configs

Silvest89:
#### What type of PR is this?
/kind feature

Currently the Hetzner provider is very lacking and the autoscaler does not know beforehand which custom taints/labels the provided node pools have.
This PR adds support for a each nodepool to have their own initconfig (currently breaking change, but will see if I have the time to make it backwards compatible). It also supports multi arch node pools. I have been running this a month or so~ build my own image

```release-note
Currently this PR introduces a breaking change. However if required I can refactor it so it is backwards compatible
```

### DIFF
--- a/cluster-autoscaler/cloudprovider/hetzner/hetzner_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/hetzner/hetzner_cloud_provider.go
@@ -191,9 +191,12 @@ func BuildHetzner(_ config.AutoscalingOptions, do cloudprovider.NodeGroupDiscove
 		klog.Fatalf("Failed to create Hetzner cloud provider: %v", err)
 	}
 
+	if manager.clusterConfig.IsUsingNewFormat && len(manager.clusterConfig.NodeConfigs) == 0 {
+		klog.Fatalf("No cluster config present provider: %v", err)
+	}
+
 	validNodePoolName := regexp.MustCompile(`^[a-z0-9A-Z]+[a-z0-9A-Z\-\.\_]*[a-z0-9A-Z]+$|^[a-z0-9A-Z]{1}$`)
 	clusterUpdateLock := sync.Mutex{}
-
 	for _, nodegroupSpec := range do.NodeGroupSpecs {
 		spec, err := createNodePoolSpec(nodegroupSpec)
 		if err != nil {
@@ -204,6 +207,13 @@ func BuildHetzner(_ config.AutoscalingOptions, do cloudprovider.NodeGroupDiscove
 		servers, err := manager.allServers(spec.name)
 		if err != nil {
 			klog.Fatalf("Failed to get servers for for node pool %s error: %v", nodegroupSpec, err)
+		}
+
+		if manager.clusterConfig.IsUsingNewFormat {
+			_, ok := manager.clusterConfig.NodeConfigs[spec.name]
+			if !ok {
+				klog.Fatalf("No node config present for node group id `%s` error: %v", spec.name, err)
+			}
 		}
 
 		manager.nodeGroups[spec.name] = &hetznerNodeGroup{
--- a/cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go
+++ b/cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go
@@ -19,6 +19,7 @@ package hetzner
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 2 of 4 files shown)


Shubham82:
Hi @Silvest89 
Please sign the CLA before the PR can be reviewed.

Shubham82:
To check EasyCLA

/easycla

Silvest89:
Alright, just signed it :)

apricote (changes requested):
Looks very nice :100: 

Two things I would like to see before approving this:

**Backwards compatibility**: We do not want to break users of the older options, so please add `HCLOUD_CLOUD_INIT` & `HCLOUD_IMAGE` back in. If one of the old options as well as a the `HCLOUD_CLUSTER_CONFIG` is defined, I think we should use the newer options.

**Docs**: It would be great if you can provide an example of how this can be used and how they relate to the older options.

Silvest89:
> Looks very nice 💯
> 
> Two things I would like to see before approving this:
> 
> **Backwards compatibility**: We do not want to break users of the older options, so please add `HCLOUD_CLOUD_INIT` & `HCLOUD_IMAGE` back in. If one of the old options as well as a the `HCLOUD_CLUSTER_CONFIG` is defined, I think we should use the newer options.
> 
> **Docs**: It would be great if you can provide an example of how this can be used and how they relate to the older options.

Alright I will be on it. Will update the PR later this week. With updated docs as well :)

Silvest89:
@apricote 
Backwards compatibility and updated docs has been implemented

cluster-autoscaler/cloudprovider/hetzner/hetzner_cloud_provider.go
```diff
@@ -206,6 +209,13 @@ func BuildHetzner(_ config.AutoscalingOptions, do cloudprovider.NodeGroupDiscove
 			klog.Fatalf("Failed to get servers for for node pool %s error: %v", nodegroupSpec, err)
 		}
 
+		if manager.clusterConfig.IsUsingNewFormat {
+			_, ok := manager.clusterConfig.NodeConfigs[spec.name]
+			if !ok {
+				klog.Fatalf("No node config present for node id `%s` error: %v", spec.name, err)
```
apricote, line 215:
```suggestion
				klog.Fatalf("No node config present for node group id `%s` error: %v", spec.name, err)
```

cluster-autoscaler/cloudprovider/hetzner/hetzner_manager.go
```diff
@@ -71,19 +94,41 @@ func newManager() (*hetznerManager, error) {
...
+	if clusterConfigBase64 != "" {
+		clusterConfig.IsUsingNewFormat = true
+	}
+
+	if clusterConfig.IsUsingNewFormat {
+		clusterConfigEnv, err := base64.StdEncoding.DecodeString(clusterConfigBase64)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse cluster config error: %s", err)
+		}
+		json.Unmarshal([]byte(string(clusterConfigEnv)), &clusterConfig)
```
apricote, line 116: `clusterConfigEnv` already is a `[]byte`, so no need to convert to string and back to `[]byte`. https://pkg.go.dev/encoding/base64#Encoding.DecodeString

```suggestion
		json.Unmarshal(clusterConfigEnv, &clusterConfig)
```

---

Also please add error handling for failed `json.Unmarshal()`.

cluster-autoscaler/cloudprovider/hetzner/hetzner_node_group.go
```diff
@@ -392,10 +412,16 @@ func createServer(n *hetznerNodeGroup) error {
...
+
+	if n.manager.clusterConfig.IsUsingNewFormat {
+		cloudInit = n.manager.clusterConfig.NodeConfigs[n.id].CloudInit
+	}
+
 	StartAfterCreate := true
 	opts := hcloud.ServerCreateOpts{
 		Name:             newNodeName(n),
-		UserData:         n.manager.cloudInit,
+		UserData:         string(cloudInit),
```
apricote, line 424: `cloudInit` is already a string

```suggestion
		UserData:         cloudInit,
```

apricote (changes requested):
Looking good besides one missed error handling and the linting errors from CI :) 

https://github.com/kubernetes/autoscaler/actions/runs/6500103013/job/17654801323?pr=6184#step:6:152

Silvest89:
@apricote 
I've pushed your suggestions! :) Should be alright now

bgervan:
Am I seeing correctly that it is not included in the latest version yet?

Shubham82:
yes, @bgervan these changes weren't include in the latest version yet i.e. CA v1.28. This feature will be available in CA v1.29.

PR #6187, opened 2023-10-11
Rule ordering

artemvmin:
#### What type of PR is this?

/kind cleanup

Adds priority to drainability rules. This allows custom rules that return a lower priority outcome (e.g. DrainOk) to not get blocked by higher priority outcomes (e.g. SkipDrain).

Parent rebase: https://github.com/kubernetes/autoscaler/pull/6164

The one complication is that the legacy replica count check is now run after DrainOk rules (i.e. daemonset, safetoevict, and terminal). See drainability/rules.go and the now-combined replicated rule. If this is not okay, I have some alternatives:

1. How far out are we from completely deprecating the legacy logic? Can we remove this logic?
2. Temporarily add duplicate "legacy checks" to the DrainOk rules, so that they are skipped. This is the least ideal, but should be a reasonable last resort.

```release-note
NONE
```


x13n:
I'm not a fan of this approach. It implies rule ordering doesn't matter - only outcome priority does, which is not true (see https://github.com/kubernetes/autoscaler/pull/6077 for an example why).

Another issue is that it will be impossible to create custom rules blocking drain if there's an existing rule resulting in `DrainOk`/`SkipDrain`.

artemvmin:
> I'm not a fan of this approach. It implies rule ordering doesn't matter - only outcome priority does, which is not true (see #6077 for an example why).
> 
> Another issue is that it will be impossible to create custom rules blocking drain if there's an existing rule resulting in `DrainOk`/`SkipDrain`.

Agreed. Consider this alternative implementation, which makes the framework more flexible, not less: https://github.com/kubernetes/autoscaler/pull/6196

x13n:
Thanks. Reviewed the other PR and closing this one.

/close

PR #6189, opened 2023-10-11
Template the autoDiscovery.clusterName variable in the Helm chart

mtougeron:
#### What type of PR is this?

/kind feature

This allows templating of the `autoDiscovery.clusterName` parameter in the Helm chart so that you can do more dynamic processing of the cluster's name.

Fixes # n/a

```release-note
Template the autoDiscovery.clusterName variable in the Helm chart
```

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -11,4 +11,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.30.1
+version: 9.31.0
--- a/charts/cluster-autoscaler/README.md
+++ b/charts/cluster-autoscaler/README.md
@@ -74,6 +74,10 @@ To create a valid configuration, follow instructions for your cloud provider:
 - [OpenStack Magnum](#openstack-magnum)
 - [Cluster API](#cluster-api)
 
+### Templating the autoDiscovery.clusterName
+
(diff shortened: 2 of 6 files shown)


mwielgus (approved):
/lgtm
/approve

gjtempleton (commented):
Do we want to also support it in the templating for CAPI? https://github.com/kubernetes/autoscaler/blob/6a7e6a23f45b4fd0d1a73e44456497d7763794e2/charts/cluster-autoscaler/templates/_helpers.tpl#L119

Would also be good to add something to the readme explaining that this can be used I think?

charts/cluster-autoscaler/Chart.yaml
```diff
@@ -11,4 +11,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.30.0
+version: 9.30.1
```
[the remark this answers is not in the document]
mtougeron: No worries. :)

mtougeron:
> Do we want to also support it in the templating for CAPI?

Yes I should update it there. I must have just missed that one. Thanks!

charts/cluster-autoscaler/Chart.yaml
```diff
@@ -11,4 +11,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.30.1
+version: 9.30.2
```
gjtempleton, line 14: One last request actually, can we make this a minor version bump rather than a patch as we're adding functionality.

mtougeron: done

gjtempleton:
Thanks!
/lgtm
/approve

PR #6193, opened 2023-10-13
Update README.md: Update link to "Charts".

guettli:
I think most people want to read the autogenerate README containing the docs for the helm chart values.

#### What type of PR is this?
/kind documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ amount of CPU and memory requested by pods running in the Kubernetes Cluster. Cu
 [Addon Resizer](https://github.com/kubernetes/autoscaler/tree/master/addon-resizer) - a simplified version of vertical pod autoscaler that modifies
 resource requests of a deployment based on the number of nodes in the Kubernetes Cluster. Current state - beta.
 
-[Charts](https://github.com/kubernetes/autoscaler/tree/master/charts) - Supported Helm charts for components above.
+[Charts](https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/README.md) - Supported Helm charts for components above.
 
 ## Contact Info
 


README.md
```diff
@@ -15,7 +15,7 @@ amount of CPU and memory requested by pods running in the Kubernetes Cluster. Cu
 [Addon Resizer](https://github.com/kubernetes/autoscaler/tree/master/addon-resizer) - a simplified version of vertical pod autoscaler that modifies
 resource requests of a deployment based on the number of nodes in the Kubernetes Cluster. Current state - beta.
 
-[Charts](https://github.com/kubernetes/autoscaler/tree/master/charts) - Supported Helm charts for components above.
+[Charts](https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/README.md) - Supported Helm charts for components above.
```
Shubham82, line 18: @guettli I think the current link is correct, it redirects to the charts directory, which contains the components for which helm charts are supported.

guettli: The old link: https://github.com/kubernetes/autoscaler/tree/master/charts

The new link: https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/README.md

I think most uses want to read the new link. It explains the which values are supported by the helm chart.

The old link explains how the README gets build. I think only a fraction of the readers are about this detail.

I would like to add the new link. 

Maybe we should have both links. The new and the old link.

What do you think?

Shubham82: If we want to change the link to see the values, so instead of the `https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/README.md` link which redirects specifically to README, we can use the following link: `https://github.com/kubernetes/autoscaler/tree/master/charts/cluster-autoscaler`. Which also shows the supported helm charts for the component (CA) and README.

Shubham82: Open to suggestions.
cc @MaciekPytel @mwielgus

gjtempleton: The intent of this link originally was that it's chart neutral, if we chose to add additional supported helm charts to this repo.

Shubham82: I agree with @gjtempleton, The same thing I mentioned above comment. IMO so we don't need to change this link.

PR #6196, opened 2023-10-13
Add mechanism to override drainability status

artemvmin:
#### What type of PR is this?

/kind feature

Adds a mechanism for one rule to selectively override other rules. This allows for custom rules that override some outcomes (e.g. BlockDrain -> DrainOk) while not overriding others (e.g. SkipDrain).

```release-note
NONE
```

### DIFF
--- a/cluster-autoscaler/simulator/drainability/rules/rules.go
+++ b/cluster-autoscaler/simulator/drainability/rules/rules.go
@@ -86,10 +86,23 @@ func (rs Rules) Drainable(drainCtx *drainability.DrainContext, pod *apiv1.Pod) d
 		drainCtx.RemainingPdbTracker = pdb.NewBasicRemainingPdbTracker()
 	}
 
+	var candidates []drainability.Status
+
 	for _, r := range rs {
-		d := r.Drainable(drainCtx, pod)
-		if d.Outcome != drainability.UndefinedOutcome {
-			return d
+		status := r.Drainable(drainCtx, pod)
+		if len(status.Overrides) > 0 {
+			candidates = append(candidates, status)
+			continue
+		}
+		for _, candidate := range candidates {
+			for _, override := range candidate.Overrides {
+				if status.Outcome == override {
+					return candidate
+				}
+			}
+		}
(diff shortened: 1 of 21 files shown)


cluster-autoscaler/simulator/drainability/rules/rules_test.go
```diff
@@ -0,0 +1,128 @@
...
+				fakeRule{drainability.NewBlockedStatus(drain.NotEnoughPdb, nil)},
+			},
+			want: drainability.Status{
+				Outcome:   drainability.DrainOk,
+				Overrides: []drainability.OutcomeType{drainability.BlockDrain},
+			},
+		},
+		"override no match": {
+			rules: Rules{
+				fakeRule{drainability.Status{
```
x13n, line 59: nit: How about `drainability.NewDrainableStatus(drainability.OverridingOnly(drainability.SkipDrain))`? Or is it not really easier to read?

cluster-autoscaler/simulator/drainability/rules/rules.go
```diff
@@ -86,10 +86,23 @@ func (rs Rules) Drainable(drainCtx *drainability.DrainContext, pod *apiv1.Pod) d
...
-		if d.Outcome != drainability.UndefinedOutcome {
-			return d
+		status := r.Drainable(drainCtx, pod)
+		if len(status.Overrides) > 0 {
+			candidates = append(candidates, status)
+			continue
+		}
+		for _, candidate := range candidates {
+			for _, override := range candidate.Overrides {
+				if status.Outcome == override {
```
x13n, line 99: Maybe worth logging the fact that override happened? At least at `V(5)`? This logic becomes non-trivial with this change, so some observability wouldn't hurt.

x13n (changes requested):
Just minor comments, this looks like a reasonable mechanism.

cluster-autoscaler/simulator/drainability/rules/rules.go
artemvmin: Great idea. Added a Name() to the Rule interface to make this log more useful.

PR #6201, opened 2023-10-17
Adds and implements node taints count metric.

piotrwrotniak:
#### What type of PR is this?

/kind feature

Adds and implements node taints count metric.

```release-note
Adds and implements node taints count metric.
```

### DIFF
--- a/cluster-autoscaler/utils/taints/taints.go
+++ b/cluster-autoscaler/utils/taints/taints.go
@@ -19,6 +19,7 @@ package taints
 import (
 	"context"
 	"fmt"
+	"maps"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/utils/taints/taints.go
```diff
@@ -416,3 +442,50 @@ func FilterOutNodesWithStartupTaints(taintConfig TaintConfig, allNodes, readyNod
...
 	return newAllNodes, newReadyNodes
 }
+
+// CountNodeTaints counts used node taints.
+func CountNodeTaints(nodes []*apiv1.Node, taintConfig TaintConfig) map[string]int {
+	foundTaintsCount := make(map[string]int)
+	for _, node := range nodes {
+		for _, taint := range node.Spec.Taints {
+			key := getTaintTypeToReport(taint.Key, taintConfig)
+			if _, ok := foundTaintsCount[key]; ok {
```
x13n, line 452: Missing map entries will default to 0, so there's no need to handle "not found" case separately - simple increment will suffice.

cluster-autoscaler/utils/taints/taints.go
```diff
@@ -108,6 +117,23 @@ var (
 	// Mutable only in unit tests
 	maxRetryDeadline      time.Duration = 5 * time.Second
 	conflictRetryInterval time.Duration = 750 * time.Millisecond
+
+	explicitlyReportedNodeTaints = TaintKeySet{
```
x13n, line 121: This is duplicating the list from `NodeConditionTaints`, can we initialize this with `NodeConditionTaints` with a few extra taints? I worry they may get out of sync.

cluster-autoscaler/proposals/metrics.md
```diff
@@ -21,17 +21,18 @@ All the metrics are prefixed with `cluster_autoscaler_`.
...
+| max_nodes_count | Gauge |                                            | Maximum number of nodes in all node groups.                                        |
+| cluster_cpu_current_cores | Gauge |                                            |                                                                                    | Current number of cores in the cluster, minus deleting nodes. |
+| cpu_limits_cores | Gauge | `direction`=&lt;`minimum` or `maximum`&gt; | Minimum and maximum number of cores in the cluster.                                |
+| cluster_memory_current_bytes | Gauge |                                            | Current number of bytes of memory in the cluster, minus deleting nodes.            |
+| memory_limits_bytes | Gauge | `direction`=&lt;`minimum` or `maximum`&gt; | Minimum and maximum number of bytes of memory in cluster.                          |
+ | node_taints_count | Gauge | `type`=&lt;taint-type&gt;                  | Number of taints per type in cluster.                                              |
```
x13n, line 35: This is a pretty old proposal, not sure if there's much value in adding the metric here. There's already a lot of metrics that were added in the code but not to this doc.

cluster-autoscaler/utils/taints/taints.go
```diff
@@ -416,3 +442,50 @@ func FilterOutNodesWithStartupTaints(taintConfig TaintConfig, allNodes, readyNod
...
+	// Track deprecated taints.
+	if strings.HasPrefix(key, IgnoreTaintPrefix) {
+		return IgnoreTaintPrefix
+	}
+
+	if _, ok := explicitlyReportedNodeTaints[key]; ok {
+		return key
+	}
+
+	if _, ok := taintConfig.StartupTaints[key]; ok {
```
x13n, line 472: optional: There's already 3-4 places in this file that separately check prefix and exact values in `taintConfig`. WDYT about slightly refactoring this to make fields in `TaintConfig` private and expose `IsStartupTaint`/`IsStatusTaint` functions instead?

x13n (changes requested):
Thanks for the change!

cluster-autoscaler/utils/taints/taints.go
piotrwrotniak: Done

cluster-autoscaler/utils/taints/taints.go
piotrwrotniak: Changed, PTAL

cluster-autoscaler/utils/taints/taints.go
piotrwrotniak: Added

cluster-autoscaler/proposals/metrics.md
piotrwrotniak: Removed

PR #6202, opened 2023-10-17
feat: node hibernate

### LINKED ISSUE #5923: Support deallocate mode for Azure vmss when shutting down

Cluster autoscaler for Azure

Azure does not support customizing the node image, but it does allow to stop (Deallocate) nodes in a nodepool. This would drastically speed up the startup of our nodes using sysbox (which is now installed by a deamonset at startup).  https://github.com/nestybox/sysbox/blob/master/docs/user-guide/install-k8s.md

- We should be able to configure a given vmss with a tag/label to indicate it should Deallocate a node instead of Delete it (delete should be the default).
- In the scale_down code we should use this property to either call DeleteInstances or DeallocateInstances for a vmss.
- in the scale_up: first start deallocated nodes before scaling up the vmss

- First looked at customizing the node image (this was our preferred option), but this is not allowad by Azure, only the capi project  or the legacy aks-engine project support it.

**Additional context.**:
In the Azure docs about the shutdownMode, they mention it as their solution to speed up the initial node creation. https://learn.microsoft.com/en-us/azure/aks/scale-down-mode

jackfrancis:
#### What type of PR is this?

/kind feature

This PR introduces a new Azure-specific `ScaleDownPolicy` feature that allows configurable scale down, either "Delete" (default, which is the current behavior) or "Deallocate", a new behavior which deallocates but doesn't delete the underlying VM, making it more responsively available for re-use on a future scale up.

Ref:

- https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/deallocate?tabs=HTTP

Fixes #5923 

```release-note
azure: ScaleDownPolicy feature, including Deallocate
```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/azure/azure_config.go
+++ b/cluster-autoscaler/cloudprovider/azure/azure_config.go
@@ -135,6 +135,9 @@ type Config struct {
 
 	// EnableVmssFlex defines whether to enable Vmss Flex support or not
 	EnableVmssFlex bool `json:"enableVmssFlex,omitempty" yaml:"enableVmssFlex,omitempty"`
+
+	// ScaleDownPolicy is the VMSS scale down policy, either "Delete" or "Hibernate"
+	ScaleDownPolicy string `json:"scaleDownPolicy" yaml:"scaleDownPolicy"`
 }
 
 // BuildAzureConfig returns a Config object for the Azure clients
@@ -169,6 +172,7 @@ func BuildAzureConfig(configReader io.Reader) (*Config, error) {
 		cfg.AADClientCertPath = os.Getenv("ARM_CLIENT_CERT_PATH")
 		cfg.AADClientCertPassword = os.Getenv("ARM_CLIENT_CERT_PASSWORD")
 		cfg.Deployment = os.Getenv("ARM_DEPLOYMENT")
+		cfg.ScaleDownPolicy = os.Getenv("AZURE_SCALE_DOWN_POLICY")
 
 		subscriptionID, err := getSubscriptionIdFromInstanceMetadata()
 		if err != nil {
--- a/cluster-autoscaler/cloudprovider/cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/cloud_provider.go
@@ -236,6 +236,23 @@ type NodeGroup interface {
 	GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error)
 }
 
+// PolicyNodeGroup is a wrapper for a nodegroup that can have scaling policies
+type PolicyNodeGroup interface {
+	// ScaleDownPolicy returns whether this nodegroup instances will be deleted or hibernated on scale down
+	ScaleDownPolicy() ScaleDownPolicy
+}
+
+// ScaleDownPolicy is a string representation of the ScaleDownPolicy
+type ScaleDownPolicy string
+
+const (
+	// Delete means that on scale down, nodes will be deleted
+	Delete ScaleDownPolicy = "Delete"
+	// Hibernate means that on scale down, node infrastructure will be stopped
+	// and on scale-up they will be prioritized and restarted
+	Hibernate ScaleDownPolicy = "Hibernate"
+)
+
 // Instance represents a cloud-provider node. The node does not necessarily map to k8s node
 // i.e it does not have to be registered in k8s cluster despite being returned by NodeGroup.Nodes()
 // method. Also it is sane to have Instance object for nodes which are being created or deleted.
@@ -265,6 +282,10 @@ const (
 	InstanceCreating InstanceState = 2
 	// InstanceDeleting means instance is being deleted
 	InstanceDeleting InstanceState = 3
+	// InstanceHibernated means that the instance is hibernated for future scale up
+	InstanceHibernated InstanceState = 4
+	// InstanceHibernating means the instance is currently being hibernated
+	InstanceHibernating InstanceState = 5
 )
 
 // InstanceErrorInfo provides information about error condition on instance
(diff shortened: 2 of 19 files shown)


cluster-autoscaler/cloudprovider/azure/azure_config.go
```diff
@@ -135,6 +135,9 @@ type Config struct {
 
 	// EnableVmssFlex defines whether to enable Vmss Flex support or not
 	EnableVmssFlex bool `json:"enableVmssFlex,omitempty" yaml:"enableVmssFlex,omitempty"`
+
+	// ScaleDownPolicy is the VMSS scale down policy, either "Delete" or "Deallocate"
```
Jont828, line 139:
```suggestion
	// ScaleDownPolicy is the VMSS scale down policy, either "Delete" or "Deallocate"
	// +kubebuilder:validation:Enum=Delete;Deallocate
```

cluster-autoscaler/cloudprovider/azure/azure_config.go
```diff
@@ -135,6 +135,9 @@ type Config struct {
...
+	ScaleDownPolicy string `json:"scaleDownPolicy" yaml:"scaleDownPolicy"`
```
jackfrancis, line 140: @Jont828 FYI this is where a new feature config lives for the azure provider

MaciekPytel:
Just taking a very quick look my feeling is that we should entirely keep in cloudprovider or make it fully a first class citizen (preferably the latter). What I mean by that, if we change clusterstate to be able to account for Deallocated nodes, I think we should also explicitly add functionality to Deallocate to cloudprovider API. Ideally other providers should be able to add the functionality just by implementing this new method / methods.

x13n:
+1 to the above, it should either be cloudprovider implementation detail (i.e. cloud provider can decide whether how to handle `DeleteNodes` - really delete or just deallocate) or a first class citizen that requires extending cloudprovider interface to differentiate between deletion and hibernation/deallocation. I will be looking to build a similar proof of concept for GKE in the near future. @jackfrancis - would you like to cooperate on this?

jackfrancis:
@MaciekPytel @x13n thanks for your notes. I'm 100% aligned on making this cloud provider-reusable and implementing this in partnership w/ at least one other provider. I'll reserve some more cycles this week to incorporate notes and we can start figuring out what the best approache is for core + cloud provider surface area.

thx, hang tight for just a few more days!

cluster-autoscaler/cloudprovider/cloud_provider.go
```diff
@@ -236,6 +236,23 @@ type NodeGroup interface {
 	GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error)
 }
 
+// PolicyNodeGroup is a wrapper for a nodegroup that can have scaling policies
+type PolicyNodeGroup interface {
```
jackfrancis, line 240: @MaciekPytel @x13n WDYT about this kind of optional interface definition for cloud providers?

This would standardize what this looks like for all cloud providers but doesn't require any provider to use it at the cost of this type of runtime check:

`if policyNg, ok := ng.(PolicyNodeGroup); ok && policyNg.ScaleDownPolicy() == Hibernate {`

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -987,6 +1005,12 @@ func (csr *ClusterStateRegistry) GetUpcomingNodes() (upcomingCounts map[string]i
 		ar := csr.acceptableRanges[id]
 		// newNodes is the number of nodes that
 		newNodes := ar.CurrentTarget - (len(readiness.Ready) + len(readiness.Unready) + len(readiness.LongUnregistered))
+		if cloudprovider.IsNodeGroupHibernateEnabled(nodeGroup) {
+			newNodes -= len(readiness.Hibernating)
```
x13n, line 1009: How exactly are the hibernated nodes modeled? Here you are subtracting them from `CurrentTarget` here, but in Azure cloudprovider code you write:

> // If the policy for this ScaleSet is Hibernate, the TargetSize is the capacity reported by VMSS minus the nodes
	// in deallocated and deallocating states

Which suggests hibernated nodes are subtracted twice.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -605,14 +620,17 @@ func (csr *ClusterStateRegistry) updateReadinessStats(currentTime time.Time) {
...
-	update := func(current Readiness, node *apiv1.Node, nr kube_util.NodeReadiness) Readiness {
+	update := func(current Readiness, node *apiv1.Node, nr kube_util.NodeReadiness, nodeGroup cloudprovider.NodeGroup) Readiness {
 		current.Registered = append(current.Registered, node.Name)
 		if _, isDeleted := csr.deletedNodes[node.Name]; isDeleted {
 			current.Deleted = append(current.Deleted, node.Name)
 		} else if nr.Ready {
 			current.Ready = append(current.Ready, node.Name)
 		} else if node.CreationTimestamp.Time.Add(MaxNodeStartupTime).After(currentTime) {
 			current.NotStarted = append(current.NotStarted, node.Name)
+		} else if cloudprovider.IsNodeGroupHibernateEnabled(nodeGroup) && taints.HasShutdownTaint(node) || taints.HasUnreachableTaint(node) {
```
x13n, line 631: Relying on generic taints here seems quite fragile. What if the node is unreachable for other reasons than hibernation? Misconfigured firewall rules, perhaps? Since you introduced `InstanceHibernating` and `InstanceHibernated` states for instances, why not rely on that instead?

x13n (commented):
Thanks for the change. It will take me some time to review this properly, but for now adding some comments after skimming through the code.

PR #6204, opened 2023-10-18
Update PodListProcessor & Use non processed pods by scheduler as unschedulable pods 

atwamahmoud:
#### What type of PR is this?

/kind cleanup
/kind feature

- Moves `FilterOutAndSplitExpendables` calls to a new PodListProcessor
- Moves `clearTpuRequests` calls to a new PodListProcessor
- Makes both `UnschedulablePodTimeBuffer` & `UnschedulablePodWithGpuTimeBuffer` configurable with flags
- Use pods which are not marked by scheduler yet (unknown) as Unschedulable pods since schedulable pods will be filtered by the current PodListProcessor effectively making CA not wait for scheduler
```release-note
Adds 2 new flags "unschedulable-pod-time-buffer" & "unschedulable-pod-with-gpu-time-buffer" with default values of 2sec & 30sec respectively.
The values were constants but now can be configured according to users' needs.
Furthermore, CA will not wait for scheduler to mark pods as unschedulable and will evaluate non processed pods.
```

### DIFF
--- a/cluster-autoscaler/core/podlistprocessor/filter_out_expendable.go
+++ b/cluster-autoscaler/core/podlistprocessor/filter_out_expendable.go
@@ -1,0 +1,66 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package podlistprocessor
+
+import (
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/autoscaler/cluster-autoscaler/context"
+	core_utils "k8s.io/autoscaler/cluster-autoscaler/core/utils"
+	caerrors "k8s.io/autoscaler/cluster-autoscaler/utils/errors"
+	klog "k8s.io/klog/v2"
+)
+
+type filterOutExpandable struct {
+}
+
+// NewFilterOutExpandablePodListProcessor creates a PodListProcessor filtering out expendable pods
+func NewFilterOutExpandablePodListProcessor() *filterOutExpandable {
+	return &filterOutExpandable{}
+}
+
+// Process filters out pods which are expendable and adds pods which is waiting for lower priority pods preemption to the cluster snapshot
+func (p *filterOutExpandable) Process(context *context.AutoscalingContext, pods []*apiv1.Pod) ([]*apiv1.Pod, error) {
+	klog.V(4).Infof("Filtering out expandable pods")
+	nodes, err := context.AllNodeLister().List()
+	if err != nil {
+		return nil, err
+	}
+	expendablePodsPriorityCutoff := context.AutoscalingOptions.ExpendablePodsPriorityCutoff
+
+	unschedulablePods, waitingForLowerPriorityPreemption := core_utils.FilterOutExpendableAndSplit(pods, nodes, expendablePodsPriorityCutoff)
+	if err = p.addPreemptiblePodsToSnapshot(waitingForLowerPriorityPreemption, context); err != nil {
+		return nil, err
+	}
+
+	return unschedulablePods, nil
+}
+
+// addPreemptiblePodsToSnapshot modifies the snapshot simulating scheduling of pods waiting for preemption.
+// this is not strictly correct as we are not simulating preemption itself but it matches
+// CA logic from before migration to scheduler framework. So let's keep it for now
+func (p *filterOutExpandable) addPreemptiblePodsToSnapshot(pods []*apiv1.Pod, ctx *context.AutoscalingContext) error {
+	for _, p := range pods {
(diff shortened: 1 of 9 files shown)


cluster-autoscaler/metrics/metrics.go
```diff
@@ -135,12 +135,13 @@ var (
...
 
-	unschedulablePodsCount = k8smetrics.NewGauge(
+	// Unschedulable pod count can be from scheduler-marked-unschedulable pods or not-yet-processed pods (unknown)
+	unschedulablePodsCount = k8smetrics.NewGaugeVec(
 		&k8smetrics.GaugeOpts{
 			Namespace: caNamespace,
 			Name:      "unschedulable_pods_count",
 			Help:      "Number of unschedulable pods in the cluster.",
-		},
+		}, []string{"count_type"},
```
x13n, line 144: nit: Isn't `count` a bit redundant here? The metric already has `_count` suffix, so this could be simply `type`?

cluster-autoscaler/metrics/metrics.go
```diff
@@ -462,8 +463,9 @@ func UpdateNodeGroupsCount(autoscaled, autoprovisioned int) {
 }
 
 // UpdateUnschedulablePodsCount records number of currently unschedulable pods
-func UpdateUnschedulablePodsCount(podsCount int) {
-	unschedulablePodsCount.Set(float64(podsCount))
+func UpdateUnschedulablePodsCount(uschedulablePodsCount, unknownPodsCount int) {
+	unschedulablePodsCount.WithLabelValues("unschedulable").Set(float64(uschedulablePodsCount))
+	unschedulablePodsCount.WithLabelValues("unknown").Set(float64(unknownPodsCount))
```
x13n, line 468: ditto on the `unknown`. Can we find a better name?

cluster-autoscaler/main.go
```diff
@@ -243,6 +243,8 @@ var (
 	maxAllocatableDifferenceRatio           = flag.Float64("max-allocatable-difference-ratio", config.DefaultMaxAllocatableDifferenceRatio, "Maximum difference in allocatable resources between two similar node groups to be considered for balancing. Value is a ratio of the smaller node group's allocatable resource.")
 	forceDaemonSets                         = flag.Bool("force-ds", false, "Blocks scale-up of node groups too small for all suitable Daemon Sets pods.")
 	dynamicNodeDeleteDelayAfterTaintEnabled = flag.Bool("dynamic-node-delete-delay-after-taint-enabled", false, "Enables dynamic adjustment of NodeDeleteDelayAfterTaint based of the latency between CA and api-server")
+	unschedulablePodTimeBuffer              = flag.Duration("unschedulable-pod-time-buffer", 2*time.Second, "How old the oldest unschedulable pod should be before starting scale up.")
```
x13n, line 246: I'm not convinced we need these flags and I'm pretty sure we need one to control whether CA waits for scheduler to mark pods as uschedulable.

cluster-autoscaler/core/podlistprocessor/filter_out_expendable_test.go
```diff
@@ -0,0 +1,179 @@
...
+					podsInSnapshot = append(podsInSnapshot, podInfo.Pod)
+				}
+			}
+
+			assert.ElementsMatch(t, tc.wantPodsInSnapshot, podsInSnapshot)
+		})
+	}
+}
+
+func getPrioritySetter(priority int32) func(*apiv1.Pod) {
```
x13n, line 144: nit: `getXSetter` is a bit too verbose, I'd just rename to `x`: `getPrioritySetter`->`priority`, `getNominatedNodeNameSetter`->`nominatedNodeName`. It'll be easier to read in the test cases.

cluster-autoscaler/core/podlistprocessor/filter_out_expendable.go
```diff
@@ -0,0 +1,66 @@
...
+
+	unschedulablePods, waitingForLowerPriorityPreemption := core_utils.FilterOutExpendableAndSplit(pods, nodes, expendablePodsPriorityCutoff)
+	if err = p.addPreemptiblePodsToSnapshot(waitingForLowerPriorityPreemption, context); err != nil {
+		return nil, err
+	}
+
+	return unschedulablePods, nil
+}
+
+// addPreemptiblePodsToSnapshot modifies the snapshot simulating scheduling of pods waiting for preemption.
```
x13n, line 52: nit: `Preempting`? `Preemptible` suggests they can be preempted, while these are the pods that are themselves preempting other pods.

cluster-autoscaler/core/podlistprocessor/filter_out_expendable.go
```diff
@@ -0,0 +1,66 @@
...
+// Process filters out pods which are expendable and adds pods which is waiting for lower priority pods preemption to the cluster snapshot
+func (p *filterOutExpandable) Process(context *context.AutoscalingContext, pods []*apiv1.Pod) ([]*apiv1.Pod, error) {
+	klog.V(4).Infof("Filtering out expandable pods")
+	nodes, err := context.AllNodeLister().List()
+	if err != nil {
+		return nil, err
+	}
+	expendablePodsPriorityCutoff := context.AutoscalingOptions.ExpendablePodsPriorityCutoff
+
+	unschedulablePods, waitingForLowerPriorityPreemption := core_utils.FilterOutExpendableAndSplit(pods, nodes, expendablePodsPriorityCutoff)
```
x13n, line 44: `FilterOutExpendableAndSplit` is only used here now, maybe it belongs in this file?

x13n (changes requested):
/assign

PR #6208, opened 2023-10-19
Update FAQ.md

axelgMS:
Looks like there's also this scenario that can prevent CAS from scaling down. It's been especially observed with Calico-typha.

1.	Calico-typha will use a hostPort configuration to listen a port on host by design. This means only max 1 pod of calico-typha can exist on the same node.
2.	Calico-typha is a deployment, which has a “auto-scale” configuration from the tigera operator to set its replica count according to node count.
3.	When scaling down, cluster-autoscaler will try to find another node that can schedule the pod, however it will be blocked due to (1), as every other node also have typha scheduled.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/FAQ.md
+++ b/cluster-autoscaler/FAQ.md
@@ -89,6 +89,7 @@ Cluster Autoscaler decreases the size of the cluster when some nodes are consist
   * are not run on the node by default, *
   * don't have a [pod disruption budget](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#how-disruption-budgets-work) set or their PDB is too restrictive (since CA 0.6).
 * Pods that are not backed by a controller object (so not created by deployment, replica set, job, stateful set etc). *
+* Pods with a hostPort configuration - causing the scheduler to fit only one of those pods per node. Scale-down might not work as the existing nodes will already have a pod on them and can't schedule the new one.
 * Pods with local storage **. *
     - unless the pod has the following annotation set:
       ```


axelgMS:
Given I'm not sure about the EasyCLA, I'm opening an issue so that someone can update the doc - if relevant ?

https://github.com/kubernetes/autoscaler/issues/6209

Shubham82:
@axelgMS 
you have to sign the CLA before the PR can be reviewed.

Shubham82:
To check EasyCLA

/easycla

mwielgus:
CLA not signed, we are unable to take the contribution. Please reopen once signed.

lualvare:
Hi @towca, 

I have experienced the same problem as reported on this issue #6209 , could you help to the CAS FAQ updated with this scenario which prevents CAS to scale down?

Thank you.

towca:
This is already captured (perhaps not clearly enough) in this point of the FAQ:
```
* Pods that cannot be moved elsewhere due to various constraints (lack of resources, non-matching node selectors or affinity,
matching anti-affinity, etc)
```

If a scheduled pod can't be scheduled on any other node in the cluster, Cluster Autoscaler won't evict it - otherwise it would just stay pending, or trigger a scale-up. The described scenario with hostPorts is just one example of this. We should probably make the wording of that point clearer. We can also list the hostPort scenario as an example, but it shouldn't be a separate point.

@lualvare I'll try to update the FAQ some time next week, or you can always send out a PR of course.

lualvare:
Hi @towca,

Thank you so much for the reply, I guess this would be good enough to make the wording clearer and add that example for hostPort scenarios, this will definitely help when cases like this are met.

Have a good day.

Thanks.

lualvare:
Hello @towca,  

I hope you are doing well, just wanted to follow up on this PR to see if this is would be updated as a feasible update to the [CAS FAQ](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node)?

towca:
@lualvare Thanks for the reminder, sent out #6567!

PR #5993, opened 2023-07-27
Feat scale based on container count

AhmedGrati:
#### What type of PR is this?

/kind feature
#### What this PR does / why we need it:
Mainly related to #5547. Adds scaling based on container count.

#### Which issue(s) this PR fixes:

Fixes #5546.

#### Special notes for your reviewer:
It's my first contribution to the project, so please bear with me. Also this is a docker image with this feature:
`ahmedgrati/addon-resizer-amd64:1.8.19`. You can use it for manual testing.

#### Does this PR introduce a user-facing change?

```release-note
Add scaling mode option with the flag --scaling-mode which accepts two options: 'node-proportional' or 'container-proportional'.
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

- KEP: https://github.com/kubernetes/autoscaler/blob/master/addon-resizer/enhancements/5546-scaling-based-on-container-count/README.md

PR #6211, opened 2023-10-19
Fix scalingMode

rsgowman:
https://github.com/kubernetes/autoscaler/pull/5993 added a 'scalingMode' but reversed it's sense (so if you specified 'containers' it would scale on nodes, and vice versa).

/kind bug

```release-note
NONE
```

### DIFF
--- a/addon-resizer/nanny/nanny_lib_test.go
+++ b/addon-resizer/nanny/nanny_lib_test.go
@@ -350,9 +350,19 @@ func TestNodeVsContainerProportional(t *testing.T) {
 			t.Errorf("updateResources got %d, want %d for test case %d.", gotScaling, tc.wantScaling, i)
 		}
 		if gotScaling == overwrite {
-			wantCPU := "400m"  // 300m + 10 * 10m
-			if gotCPU := k8s.newResources.Requests.Cpu().String(); gotCPU != wantCPU {
-				t.Errorf("updateResources got %q, want %q for test case %d.", gotCPU, wantCPU, i)
+			wantCPU := cpuBase
+			n := 0
+			if tc.scalingMode == ContainerProportional {
+				n = int(tc.containers)
+			} else {
+				n = int(tc.nodes)
+			}
+			for i := 0; i < n; i++ {
+				wantCPU.Add(cpuExtra)
+			}
+
+			if gotCPU := k8s.newResources.Requests.Cpu().String(); gotCPU != wantCPU.String() {
+				t.Errorf("updateResources got %q, want %q for test case %d.", gotCPU, wantCPU.String(), i)
 			}
 		}
 	}
--- a/addon-resizer/nanny/nanny_lib_test.go
+++ b/addon-resizer/nanny/nanny_lib_test.go
@@ -286,6 +286,78 @@ func TestUpdateResources(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


addon-resizer/nanny/nanny_lib_test.go
```diff
@@ -286,6 +286,88 @@ func TestUpdateResources(t *testing.T) {
 	}
 }
 
+func TestNodeVsContainerProportional(t *testing.T) {
```
sophieliu15, line 289: I am curious why the issue is not caught in the TestUpdateResources() above? Is it because we are using static node and container number? Is it possible to increase the coverage of the unit test above if it is not too much work?

rsgowman: It's a bit that, plus using fakes instead of real objects. (That test effectively ignores the actual scaling, and instead just returns whatever we tell the fake to return.)

I initially tried to adapt that test, rather than adding a new one, but wasn't able to do so without tearing the whole test apart and rebuilding it from scratch. (Which might be worth while... but not something I'd want to do in a bugfix PR.) Once I realized what I was doing, I abandoned that approach and just added a new test instead. <shrug> Maybe we should followup with a refactoring?

rsgowman: Or maybe rephrasing the above a bit into code: I can make this patch:


```
$ git diff -U10
diff --git a/addon-resizer/nanny/nanny_lib.go b/addon-resizer/nanny/nanny_lib.go
index 4fcc7e2a5..847ab645d 100644
--- a/addon-resizer/nanny/nanny_lib.go
+++ b/addon-resizer/nanny/nanny_lib.go
@@ -128,20 +128,22 @@ func PollAPIServer(k8s KubernetesClient, est ResourceEstimator, hc *healthcheck.
 // expected ResourceRequirements are in line with the actual ResourceRequirements.
 func updateResources(k8s KubernetesClient, est ResourceEstimator, now, lastChange time.Time, scaleDownDelay, scaleUpDelay time.Duration, threshold uint64, prevResult updateResult, scalingMode string) updateResult {
        // Query the apiserver for the cluster size.
        var num uint64
        var err error
        if scalingMode == ContainerProportional {
                num, err = k8s.CountContainers()
        } else {
                num, err = k8s.CountNodes()
        }
+  num = 999999
+  err = nil
 
        if err != nil {
                log.Error(err)
                return noChange
        }
        log.V(4).Infof("The cluster size is %d", num)
 
        // Query the apiserver for this pod's information.
        resources, err := k8s.ContainerResources()
        if err != nil {
```

And all the (previously existing) tests still pass. (Newly added test fails, of course.)

sophieliu15: Did you intend to assign default value to `num` and `err` before `if scalingMode...`? Otherwise, it looks like you are overwriting the actual value with default value after calling `k8s.CountContainers()` or `k8s.CountNotes()`.

Anyway, I know there is an urgency in submitting the fix to unblock the release. I will let you decide if you want to do the refactoring in this PR or following up with a refractory after this is submitted.

You mentioned that you plan to manually test the behaviour yesterday. How is the test going? Is the behaviour after this fix expected?

jbartosik (commented):
/lgtm
/approve

PR #6212, opened 2023-10-20
Update image builder to use Go 1.21.3

towca:
Some of Cluster Autoscaler code is now using features only available in Go 1.21.


MaciekPytel:
It's also the same version that k8s is now using -> https://github.com/kubernetes/kubernetes/blob/master/.go-version
/lgtm
/approve

PR #6217, opened 2023-10-23
Rules type must implement Rule interface

jbartosik:
The interface changed in #6196 but the Rules class didn't

/kind bug
/kind cleanup

### DIFF
--- a/cluster-autoscaler/simulator/drainability/rules/rules.go
+++ b/cluster-autoscaler/simulator/drainability/rules/rules.go
@@ -79,6 +79,11 @@ func Default(deleteOptions options.NodeDeleteOptions) Rules {
 // Rules defines operations on a collections of rules.
 type Rules []Rule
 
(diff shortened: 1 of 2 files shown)


x13n:
I don't think `Rules` has to implement the `Rule` interface at all - it represents a collection, so adding a name seems quite artificial.

jbartosik:
> I don't think `Rules` has to implement the `Rule` interface at all - it represents a collection, so adding a name seems quite artificial.

It already implements [`Drainable`](https://github.com/kubernetes/autoscaler/blob/557b0f27e5169a11e7636480c6f605bddc0c9295/cluster-autoscaler/simulator/drainability/rules/rules.go#L84) which looks to me like it's more of a composite rule than a collection of rules

x13n:
The fact the it did implement the interface before doesn't imply we want it to continue doing so.

/close

PR #6220, opened 2023-10-24
Bump CA golang to 1.21

azylinski:
#### What type of PR is this?

/kind cleanup

Bump cluster-autoscaler golang version to 1.21.
The Docker build was already using `golang:1.21.3` - it's only the CI and go.mod that was lacking behind.
K/K is already on 1.21.3 as well: https://github.com/kubernetes/kubernetes/blob/master/.go-version

With upgrade, I've removed min/max functions, which are now build-in: https://tip.golang.org/ref/spec#Min_and_max

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,7 +17,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: '>=1.20.0'
+          go-version: '1.21.3'
 
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: '>=1.20.0'
+          go-version: '>=1.21.0'
```
x13n, line 20: I'd make this 1.21 exactly instead of >=1.21, so that when 1.22 comes out, we are still consistent between build scripts and CI.

azylinski: Done

x13n:
Right, this is outside of CA dir. @gjtempleton can you approve this?

Issue #6222, opened 2023-10-24
[UX] Improve the "memory utilization too big" message


shapirus:
**Which component are you using?**:

cluster-autoscaler

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:

On seeing the following message in the log:

```
Node <node_name> is not suitable for removal - memory utilization too big (0.664205)
```

one can naturally think that it happens because it's the actual memory usage that is too high and become confused on checking it with `kubectl top node` and seeing very low reported usage, e.g.:

```
NAME                  CPU(cores)   CPU%   MEMORY(bytes)   MEMORY%   
<node_name>           47m          1%     3794Mi          24%       
```

...whereas the real reason is that it's the total memory _requested_, rather than used, by pods, that is too high for considering the node for removal.

A proof that it's a UX issue is that this confusion lead to opening issues in the past, e.g., https://github.com/kubernetes/autoscaler/issues/2303 (and we don't know how many people just ignored it as a weird quirk and didn't care to open an issue).

**Describe the solution you'd like.**:

Change the message to get rid of any chance of it being misinterpreted and make it easy to understand how this aspect of CA works without even looking in the docs, e.g.:

```
Node %s is not suitable for removal - memory requests allocation too high (%.2f)
```

There is quite possibly a similar situation with the CPU-related message. If that's the case, it would benefit from such a change too.

shapirus:
Created a PR with a suggested change: https://github.com/kubernetes/autoscaler/pull/6223

I used the terminology along the lines of the `kubectl describe node` output:

```
Allocated resources:
  (Total limits may be over 100 percent, i.e., overcommitted.)
  Resource           Requests       Limits
  --------           --------       ------
  cpu                125m (3%)      0 (0%)
  memory             10373Mi (66%)  66660Mi (426%)
```

did not replace `%f` with `%.2f` in the PR, but I think it'd be a good idea as well. To be determined in further discussion, if there will be any.

PR #6223, opened 2023-10-24
Disambiguate the resource usage node removal eligibility messages

shapirus:
#### What type of PR is this?
/kind feature

Improves user experience as described in https://github.com/kubernetes/autoscaler/issues/6222

https://github.com/kubernetes/autoscaler/issues/6222

I would also change the `%f` to `%.2f` in the messages to remove the extra significant digits that provide no added value from the output. Please let me know in the comments if it's a good idea, I'll update my commit if needed.

```release-note
Disambiguated wording in the log messages related to node removal ineligibility caused by high resource allocation
```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/scaledown/eligibility/eligibility.go
+++ b/cluster-autoscaler/core/scaledown/eligibility/eligibility.go
@@ -159,11 +159,11 @@ func (c *Checker) unremovableReasonAndNodeUtilization(context *context.Autoscali
 		return simulator.UnexpectedError, nil
 	}
 	if !underutilized {
-		klog.V(4).Infof("Node %s is not suitable for removal - %s utilization too big (%f)", node.Name, utilInfo.ResourceName, utilInfo.Utilization)
+		klog.V(4).Infof("Node %s unremovable: %s requested (%.6g%% of allocatable) is above the scale-down utilization threshold", node.Name, utilInfo.ResourceName, utilInfo.Utilization*100)
 		return simulator.NotUnderutilized, &utilInfo
 	}
 
-	klogx.V(4).UpTo(utilLogsQuota).Infof("Node %s - %s utilization %f", node.Name, utilInfo.ResourceName, utilInfo.Utilization)
+	klogx.V(4).UpTo(utilLogsQuota).Infof("Node %s - %s requested is %.6g%% of allocatable", node.Name, utilInfo.ResourceName, utilInfo.Utilization*100)
 
 	return simulator.NoReason, &utilInfo
 }


jayantjain93:
/lgtm

I see the use different messaging for UX, but I don't see a need for decimal to be more restrictive.
I think the values specifically is a ratio and for larger values this may be a "noticeable" value. This believe these finer margins of the value are used by developers to analyse issues with resource predication/bad scheduling, and losing visibility for that is not something I would like to do.

jayantjain93:
@x13n you could review approval for this change.

shapirus:
> I believe these finer margins of the value are used by developers to analyse issues with resource predication/bad scheduling, and losing visibility for that is not something I would like to do.

I agree, this justifies having them.

x13n:
I can see how `utilization` may be confusing in this context, but then `resource allocation` is not a well defined concept I think. Maybe make it a bit more verbose? For example:

```cpu request to node allocatable ratio is too high (99.9%)```

If going the verbose way, might be useful to also extend it like this:

```cpu request to node allocatable ratio (excluding DaemonSets) is too high (99.9%)```

(in case when `skipDaemonSetPods` is `true`). WDYT?

**Comment by shapirus** [comment]
If we are ok to change it a little more, what about this?

```go
klog.V(4).Infof("Node %s unremovable: %s request allocation %.2f%% is above the threshold", node.Name, utilInfo.ResourceName, utilInfo.Utilization * 100)
```

- reword to avoid bloating it too much, however it's still 7 bytes longer than the original (I tried hard to make it shorter; happy to accept suggestions for reducing the length of the message without losing its readability, meaning and the useful information that it provides)
- percents instead of the as-is float for better readability
- %.2f for the value to avoid adding extra digits while retaining the original precision, because the number is multiplied by 100 to get percents
- make it clear that there is a threshold that we compare against, and not just randomly decide that it's too high when we feel like doing it

This makes it a small self-contained change aiming to solve one particular problem.

On to your second suggestion, regarding the DaemonSet logic. I think it should be a matter of another PR, because it's going to introduce new logic where this log message is produced (checking if daemonsets are skipped or not), thus it should not block the initially proposed change.

If we speak about further improvements, it would be beneficial for the user's experience to include the actual value of the menitoned threshold in the message.
It's not quite trivial, because we don't have its value in the context where this log message is produced. It's calculated (to make things worse, conditionally, depending on the presence of a GPU) in `isNodeBelowUtilizationThreshold()`. To have it printed in the log, the calculation code would need to be moved into a separate function that could then be called by both `isNodeBelowUtilizationThreshold()` and `unremovableReasonAndNodeUtilization()`, I guess.
I guess it warrants having a separate PR for this as well (which I'll be happy to submit).

x13n:
I'm in favor of improving readability of logs, but I don't think introducing new concepts is the right approach. In Kubernetes, there's a number of well defined quantities: [pod request, pod limit, node allocatable, etc.](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/) My main point here is that if the current message is not using these concepts (i.e. "request allocation" as a percentage). I don't have a strong opinion on percentage vs fraction or how many digits the number has.

shapirus:
What about the following:

```go
klog.V(4).Infof("Node %s unremovable: requested %s (%.2f%% of allocatable) is above the scale-down utilization threshold", node.Name, utilInfo.ResourceName, utilInfo.Utilization * 100)
```

?

this will produce e.g.:

```
Node i-dfa1aa285b1240e88 unremovable: requested memory (70.33% of allocatable) is above the scale-down utilization threshold
```

jayantjain93:
> I'm in favor of improving readability of logs, but I don't think introducing new concepts is the right approach. In Kubernetes, there's a number of well defined quantities: [pod request, pod limit, node allocatable, etc.](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/) My main point here is that if the current message is not using these concepts (i.e. "request allocation" as a percentage). I don't have a strong opinion on percentage vs fraction or how many digits the number has.

@x13n, I do have a preference on how many digits. I don't think we should lower the accuracy. However, no preference on percentage vs fraction. :)

shapirus:
(meanwhile, I force-pushed the commit to the lasted version I ended up with)

shapirus:
...one more update: `%.6g` instead of `.2f` (that I used for percents in the previous version). This will output 6 significant digits: 0.123456%, 1.23456%, 12.3456% etc. This is no less precise than the original version.

I personally think that such precision is meaningless. There is no way of making any practical use of it. Such precision is way above the practical measurement accuracy.

4 digits (12.34%, 1.234%) should be more than enough for any practical use case and it will be more readable.

@x13n what do you think of having 4 significant digits for it?

x13n:
> What about the following:
> 
> ```go
> klog.V(4).Infof("Node %s unremovable: requested %s (%.2f%% of allocatable) is above the scale-down utilization threshold", node.Name, utilInfo.ResourceName, utilInfo.Utilization * 100)
> ```
> 
> ?
> 
> this will produce e.g.:
> 
> ```
> Node i-dfa1aa285b1240e88 unremovable: requested memory (70.33% of allocatable) is above the scale-down utilization threshold
> ```

Yup. Looks good to me. With one caveat that it is actually not true when `skipDaemonSetPods` is `true` :slightly_smiling_face: DS requests are then excluded from the denominator instead of being added to the numerator.




> ...one more update: `%.6g` instead of `.2f` (that I used for percents in the previous version). This will output 6 significant digits: 0.123456%, 1.23456%, 12.3456% etc. This is no less precise than the original version.
> 
> However I personally think that such precision makes at least two least-significant digits meaningless. There is no way of making any practical use of it. Such precision is way above the practical measurement accuracy.
> 
> 4 digits (12.34%, 1.234%) should be more than enough for any practical use case and it will be more readable.
> 
> @x13n what do you think of having 4 significant digits for it?

Yeah, I struggle to find a use case for keeping 6 significant digits. @jayantjain93  - do you have a specific use case for when high precision would actually help? Does 4 sound reasonable to you?

shapirus:
> With one caveat that it is actually not true when `skipDaemonSetPods` is `true` 🙂 DS requests are then excluded from the denominator instead of being added to the numerator.

So we're talking about this code here:

https://github.com/kubernetes/autoscaler/blob/db5e83bc1afeaef6bfd3d69befef9c12ad469f94/cluster-autoscaler/simulator/utilization/info.go#L94-L127

As far as I understand, they are excluded from _both_ the numerator and denominator. Which essentially means that the message will still be correct, if it is understood that what it says implies something like "...where _requests_ and _allocatable_ are what they mean to CA in this context instead of raw values from k8s", but we obviously can't add these words as is.

We could use something like "Node i-dfa1aa285b1240e88 unremovable: effective requested memory (70.33% of effective allocatable) is above the scale-down utilization threshold", but I think it will look ugly.

To be strictly correct there, we could say `...[, DaemonSet [and mirror] pods excluded]`, but will it really warrant the added verbosity of the message and complication of the code? I'm not sure.

That being said, I agree, it would be nice to indicate that the way that CA calculates utilization depends on certain settings, and then it'll be up to the user to understand what configuration options they set. If only we could make it simple and concise.

Any ideas on possible wording to use without adding extra logic (that'll be partially duplicating what's already done in `CalculateUtilizationOfResource()`)? How do we say: "..., where the meaning of _requests_ and _allocatable_ depend on what pods CA considers for calculating the usage percentage" in one or two words? :)

x13n:
Well, all of that was neatly hidden behind the word "utilization" in the existing code. I understand it is easy to confuse with other notions of utilization, so maybe it would suffice to alter the term a bit? "Allocatable utilization" is almost correct, modulo the caveats discussed. How about a completely different term like "node efficiency"? It'd probably make sense to extend the FAQ to explain what that means, but it would definitely resolve the current confusion problem.

Naming is hard...

shapirus:
The sole purpose of the proposed change was to make the message clear enough for the user to understand at a glance what's going on, without opening a web browser to search for documentation or explanation (after looking at `kubectl top node` and seeing much lower usage than that reported by CA).

I believe it serves that purpose, while staying _sufficiently correct_ for (arguably) most use cases. Even if the actual number reported may be more or less incorrect when it's affected by the daemonset-related condition, the message will still serve its main purpose and not confuse the user:

- ok, why are my nodes not scaling down? let me see the logs!
- oh, there's too much `requests` for this resource placed on this node, I need to check my pods and adjust resource allocation settings.

The difference caused by the different states of the skipDaemonSetPods attribute will come into play, if ever, only at much lower level of debugging, at which point the user will very likely be reading the documentation anyway.

Question is what is our goal about this: provide the best user experience with some technical caveats (until someone comes up with a good wording to cover both), or be strictly correct in the terminology, but make the logs significantly more noisy and potentially more confusing?

x13n:
Whoops, I actually don't think the hold is needed. `gofmt` seems to be needed though.

/hold cancel

shapirus:
> `gofmt` seems to be needed though.

Can you please elaborate on this? Do I need to change something?
(I'm rather new to the go universe, so I can miss something obvious)

x13n:
Yes - if you follow the `Details` link next to test-and-verify check, you'll see logs explaining that the linter thinks the code requires formatting. It should be sufficient to run `gofmt -s -w ./cluster-autoscaler/core/scaledown/eligibility/eligibility.go` and force push the changes. There are also some unrelated errors from another component (balancer) - if they persist, a rebase might be required.

shapirus:
@x13n that's fixed now.

x13n:
Thanks!

/lgtm

PR #6225, opened 2023-10-25
Fix: Include restartable init containers in Pod utilization calc

azylinski:
#### What type of PR is this?

/kind bug

Fix pod resources utilization calculation - Include restartable init containers

Similar logic is already in k/k: https://github.com/kubernetes/kubernetes/blob/8453eb0c24ea61794157b451e4442ea570abce45/pkg/controller/podautoscaler/replica_calculator.go#L424

```release-note
NONE
```


x13n:
Can you clarify what problem this is solving?

towca:
> Can you clarify what problem this is solving?
It's bringing CA util calculations in line with the scheduler. Without it, in some weird cases CA could be underestimating how much of a node's resources a pod is taking. E.g. a pod with containers requesting only 1 CPU, but with an initContainer requesting 7 CPU scheduled on an 8-core Node. CA could scale the node down because it'd calculate the utilization as 1/8, when actually the pod is taking 7/8 of the Node.

cluster-autoscaler/simulator/utilization/info.go
```diff
@@ -118,7 +118,14 @@ func CalculateUtilizationOfResource(nodeInfo *schedulerframework.NodeInfo, resou
 		if drain.IsPodLongTerminating(podInfo.Pod, currentTime) {
 			continue
 		}
-		for _, container := range podInfo.Pod.Spec.Containers {
+		// calculate all regular containers and restartable init containers requests
```
towca, line 121: The new calculation takes restartable initContainers (which is a new thing available in k8s 1.28+) into account, but not regular initContainers. We should be calculating the same way as kube-scheduler, the logic is here: https://github.com/kubernetes/kubernetes/blob/8509ab82b96caa2365552efa08c8ba8baf11c5ec/pkg/api/v1/resource/helpers.go#L50.

Roughly:
* The previous code just looks at a sum of container requests.
* The proposed code looks at the sum of container and restartable initContainer requests.
* The correct logic should take a max of the proposed code (sum of container and restartable initContainer requests), and the biggest initContainer requests. Now computing the effective initContainer request unfortunately requires iterating over them, because you have to also add all previous restartable initContainer requests.

The correct logic is explained in this KEP: https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/753-sidecar-containers#exposing-pod-resource-requirements. The KEP also mentions that the computed value should already be available in the `TotalResourcesRequested` field on the pod, but I don't think that's available yet.

towca (changes requested):

cluster-autoscaler/simulator/utilization/info_test.go
```diff
@@ -33,6 +34,19 @@ func TestCalculate(t *testing.T) {
 	testTime := time.Date(2020, time.December, 18, 17, 0, 0, 0, time.UTC)
 	pod := BuildTestPod("p1", 100, 200000)
 	pod2 := BuildTestPod("p2", -1, -1)
+	podWithInitContainers := BuildTestPod("p-init", 100, 200000)
```
towca, line 37: I'd add a test case for non-restartable initContainers, as they're treated completely differently from restartable ones, and we're changing the calculations to include both kinds.

cluster-autoscaler/simulator/utilization/info.go
towca: Great idea, somehow skipped my mind :sweat_smile:

LGTM, I'd just like to see the non-restartable initContainer test case.

cluster-autoscaler/simulator/utilization/info_test.go
azylinski: Sure, I've extended the tests.

towca (approved):

PR #6226, opened 2023-10-25
Add support for passing custom headers to VPA requests

asafadar:
#### What type of PR is this?
/kind Feature

Add functionality for injecting custom auth headers to prometheus client requests.

https://github.com/kubernetes/kubernetes/issues/121507
Fixes #

This is just a minimal suggestion, generally it seems like the change should come from prometheus client itself, this is a minimal suggestion that might be applied in the meantime.
```release-note
Adding request headers injection as part of VPA round tripper.

Adding 2 CLI flags:
1. authHeaderName - supply the auth header name e.g Token
2. authHeaderValueName - supply env variable name containing the header value
```

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
@@ -37,11 +37,21 @@ import (
 type PrometheusBasicAuthTransport struct {
 	Username string
 	Password string
+	Headers  *map[string]string
 }
 
 // RoundTrip function injects the username and password in the request's basic auth header
 func (t *PrometheusBasicAuthTransport) RoundTrip(req *http.Request) (*http.Response, error) {
-	req.SetBasicAuth(t.Username, t.Password)
+	if t.Username != "" && t.Password != "" {
+		req.SetBasicAuth(t.Username, t.Password)
+	}
+
+	if t.Headers != nil {
+		for k, v := range *t.Headers {
+			req.Header.Set(k, v)
+		}
+	}
+
 	return http.DefaultTransport.RoundTrip(req)
 }
 
@@ -57,6 +67,7 @@ type PrometheusHistoryProviderConfig struct {
 	CadvisorMetricsJobName                           string
 	Namespace                                        string
 	PrometheusBasicAuthTransport
+	Headers map[string]string
 }
 
 // PodHistory represents history of usage and labels for a given pod.
@@ -94,12 +105,13 @@ func NewPrometheusHistoryProvider(config PrometheusHistoryProviderConfig) (Histo
 		Address: config.Address,
 	}
 
-	if config.Username != "" && config.Password != "" {
-		transport := &PrometheusBasicAuthTransport{
+	if (config.Username != "" && config.Password != "") || config.Headers != nil {
+		transport := PrometheusBasicAuthTransport{
 			Username: config.Username,
 			Password: config.Password,
+			Headers:  &config.Headers,
 		}
-		promConfig.RoundTripper = transport
+		promConfig.RoundTripper = &transport
 	}
 
 	promClient, err := promapi.NewClient(promConfig)
--- a/vertical-pod-autoscaler/pkg/recommender/main.go
+++ b/vertical-pod-autoscaler/pkg/recommender/main.go
@@ -19,8 +19,10 @@ package main
 import (
 	"context"
 	"flag"
-	resourceclient "k8s.io/metrics/pkg/client/clientset/versioned/typed/metrics/v1beta1"
+	"os"
 	"time"
+
+	resourceclient "k8s.io/metrics/pkg/client/clientset/versioned/typed/metrics/v1beta1"
 
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/informers"
@@ -72,6 +74,10 @@ var (
 	username            = flag.String("username", "", "The username used in the prometheus server basic auth")
 	password            = flag.String("password", "", "The password used in the prometheus server basic auth")
 	memorySaver         = flag.Bool("memory-saver", false, `If true, only track pods which have an associated VPA`)
+	authHeader          = flag.Bool("auth-header", false, `If True will use header authentication, If true aslo authHeaderName and authHeaderValueName must be supplied`)
+	authHeaderName      = flag.String("auth-header-name", "", `Header name to use for header based Auth`)
+	authHeaderValueName = flag.String("auth-header-value-name", "", `Env variable name holding the auth header value e.g the value of the bearer token etc..`)
+
 	// external metrics provider config
 	useExternalMetrics   = flag.Bool("use-external-metrics", false, "ALPHA.  Use an external metrics provider instead of metrics_server.")
 	externalCpuMetric    = flag.String("external-metrics-cpu-metric", "", "ALPHA.  Metric to use with external metrics provider for CPU usage.")
@@ -202,6 +208,12 @@ func main() {
 				Username: *username,
 				Password: *password,
 			},
+		}
+		if *authHeader {
+			authValue := os.Getenv(*authHeaderValueName)
+			config.Headers = map[string]string{
+				*authHeaderName: authValue,
+			}
 		}
 		provider, err := history.NewPrometheusHistoryProvider(config)
 		if err != nil {


asafadar:
@krzysied 
@kwiesmueller 
Hey there :) Struggling a bit with the CI tests - they are failing on something that doesn't seem related to my changes at all. Would love a bit of guidance if possible. 

Thank you

sophieliu15:
> Krzysztof Siedlecki Kevin Wiesmüller Hey there :) Struggling a bit with the CI tests - they are failing on something that doesn't seem related to my changes at all. Would love a bit of guidance if possible.
> 
> Thank you

You are right I don't think the [test failure](https://github.com/kubernetes/autoscaler/actions/runs/6641866558/job/18045331036?pr=6226) is related to your change. There seem to be some importing issues for balancer/ which were introduced in https://github.com/kubernetes/autoscaler/commit/fb425d81f1befc3106d4c1067803106871709cd1

@mwielgus Do you have any insight on the testing failure related to balancer/

vertical-pod-autoscaler/pkg/recommender/main.go
```diff
@@ -72,6 +74,10 @@ var (
 	username            = flag.String("username", "", "The username used in the prometheus server basic auth")
 	password            = flag.String("password", "", "The password used in the prometheus server basic auth")
 	memorySaver         = flag.Bool("memory-saver", false, `If true, only track pods which have an associated VPA`)
+	authHeader          = flag.Bool("auth-header", false, `If True will use header authentication, If true aslo authHeaderName and authHeaderValueName must be supplied`)
+	authHeaderName      = flag.String("auth-header-name", "", `What is the header name to use for header based Auth`)
```
sophieliu15, line 78: nit: "The header name for header based Auth". We usually prefer a statement instead of question for flag description

vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
```diff
@@ -57,6 +66,7 @@ type PrometheusHistoryProviderConfig struct {
 	CadvisorMetricsJobName                           string
 	Namespace                                        string
 	PrometheusBasicAuthTransport
+	Headers map[string]string
```
sophieliu15, line 69: Could you add unit tests to test the path when Headers are provided in history_provider_test.go?

vertical-pod-autoscaler/pkg/recommender/main.go
asafadar: done :)

vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
asafadar: Added please review :)

sophieliu15:
@kwiesmueller Could you do a final round of review and approve if it looks good to you?

cc: @laoj2

vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
```diff
@@ -40,8 +40,17 @@ type PrometheusBasicAuthTransport struct {
 }
 
 // RoundTrip function injects the username and password in the request's basic auth header
-func (t *PrometheusBasicAuthTransport) RoundTrip(req *http.Request) (*http.Response, error) {
-	req.SetBasicAuth(t.Username, t.Password)
+func (c *PrometheusHistoryProviderConfig) RoundTrip(req *http.Request) (*http.Response, error) {
+	if c.PrometheusBasicAuthTransport.Username != "" && c.PrometheusBasicAuthTransport.Password != "" {
```
kwiesmueller, line 44: Is this a behavior change? Are there use-cases where one wants to only set a username and no password?

vertical-pod-autoscaler/pkg/recommender/input/history/history_provider_test.go
```diff
@@ -257,6 +262,25 @@ func TestGetMemorySamples(t *testing.T) {
...
 
+func TestAuthHeader(t *testing.T) {
+	historyProvider, err := NewPrometheusHistoryProvider(getDefaultPrometheusHistoryProviderConfigForTest())
+	if err != nil {
+		t.Fatal(err)
+	}
+	authHeaderName := "Token"
+	testToken := "aaaaaaaaaaaa"
+	tokenEnvVariableName := "TEST_TOKEN"
+	os.Setenv(tokenEnvVariableName, testToken)
```
kwiesmueller, line 273: You could use https://pkg.go.dev/testing#T.Setenv here to avoid the unset later. The defer down there is kind of out of place I think.

vertical-pod-autoscaler/pkg/recommender/input/history/history_provider_test.go
```diff
@@ -257,6 +262,25 @@ func TestGetMemorySamples(t *testing.T) {
...
+
+	req, _ := http.NewRequest("GET", "http://example.com", nil)
+
+	_, err = historyProvider.GetClusterHistory()
+
+	assert.Nil(t, err)
+	assert.Equal(t, testToken, req.Header.Get(authHeaderName))
+	defer os.Unsetenv(tokenEnvVariableName)
```
kwiesmueller, line 281: I don't think this correctly unsets the env if the asserts above fail. Maybe consider using https://pkg.go.dev/testing#T.Setenv instead or moving the defer to where the env is set.

vertical-pod-autoscaler/pkg/recommender/main.go
```diff
@@ -72,6 +74,10 @@ var (
 	username            = flag.String("username", "", "The username used in the prometheus server basic auth")
 	password            = flag.String("password", "", "The password used in the prometheus server basic auth")
 	memorySaver         = flag.Bool("memory-saver", false, `If true, only track pods which have an associated VPA`)
+	authHeader          = flag.Bool("auth-header", false, `If True will use header authentication, If true aslo authHeaderName and authHeaderValueName must be supplied`)
```
kwiesmueller, line 77:
```suggestion
	authHeader          = flag.Bool("auth-header", false, `If True will use header authentication, If true authHeaderName and authHeaderValueName must also be supplied`)
```

vertical-pod-autoscaler/pkg/recommender/main.go
```diff
@@ -72,6 +74,10 @@ var (
...
+	authHeaderName      = flag.String("auth-header-name", "", `Header name to use for header based Auth`)
+	authHeaderValueName = flag.String("auth-header-value-name", "", `Env variable name holding the auth header value e.g the value of the bearer token etc..`)
```
kwiesmueller, line 79: I think it's not clear from the flag name that an env variable key is expected here.
Could we either clarify in the flag name that this needs to be the key of an env var, or alternatively you could when reading the flag to an envsubstr and allow users to set the value to something like `${MY_ENV_VAR}`.
Either way is fine with me, what does @sophieliu15 think?

kwiesmueller (changes requested):

vertical-pod-autoscaler/pkg/recommender/input/history/history_provider.go
asafadar: Hey @kwiesmueller 
Kinda - Prior to this change round tripper was enabled only in case both username and password were supplied. Now round tripper is enabled any how and the logic was moved into the round tripper func. So in terms of username password no behavior changes - Will be used only if both supplied. But behavior did changed in case non will be provided but still round tripper will be called, Hoping I was able to answer your q

vertical-pod-autoscaler/pkg/recommender/input/history/history_provider_test.go
asafadar: done :)

vertical-pod-autoscaler/pkg/recommender/input/history/history_provider_test.go
asafadar: done @kwiesmueller

vertical-pod-autoscaler/pkg/recommender/main.go
asafadar: Fixed in code

vertical-pod-autoscaler/pkg/recommender/main.go
asafadar: @kwiesmueller tnx for the clarification. Updated the flag name to be more meaningful

vertical-pod-autoscaler/pkg/recommender/input/history/history_provider_test.go
asafadar: @kwiesmueller I removed the Unit test for now since it seems not to work. I also saw that there isn't any unit testing for round tripper at all and couldn't find any way on implementing it without making changes that will allow exposing the request object. Should I implement it using a different client just for sake of testing?

voelzmo:
@kwiesmueller @asafadar What's the current state of this PR? Do you want to rebase it and we want to get this into the next release or can we close this?

PR #6229, opened 2023-10-27
Add possbility to override the max size fo groups

kisieland:
This allowes for more control over for the CA scale-up simulation.

Allows users to control the size of the pods per group in the simulation. 
This will help in cases where multiple similar pods are being scheduled.
The original hard-coded value of 10 was limiting in cases where there are 1000 of similar pods being created at once.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/scaleup/equivalence/groups.go
+++ b/cluster-autoscaler/core/scaleup/equivalence/groups.go
@@ -36,9 +36,9 @@ type PodGroup struct {
 }
 
 // BuildPodGroups prepares pod groups with equivalent scheduling properties.
-func BuildPodGroups(pods []*apiv1.Pod) []*PodGroup {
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/core/scaleup/equivalence/groups.go
```diff
@@ -76,7 +74,7 @@ func groupPodsBySchedulingProperties(pods []*apiv1.Pod) map[equivalenceGroupId][
 			podEquivalenceGroups[*gid] = append(podEquivalenceGroups[*gid], pod)
 			continue
 		}
-		if len(egs) < maxEquivalenceGroupsByController {
+		if len(egs) < maxPodsPerGroup {
```
x13n, line 77: Wait, this used to be # of equivalence groups per controller and now you're suddenly using max pods per group here. What problem are you trying to solve?

kisieland:
Thanks @x13n, the pods were missing the Controller flag to be grouped.

PR #6235, opened 2023-10-31
Ignore scheduler processing

atwamahmoud:
#### What type of PR is this?

/kind cleanup
/kind feature

- Moves `FilterOutAndSplitExpendables` calls to a new PodListProcessor
- Moves `clearTpuRequests` calls to a new PodListProcessor
- Use pods which are not marked by scheduler yet (schedulerUnprocessed) as Unschedulable pods since schedulable pods will be filtered by the current PodListProcessor effectively making CA not wait for scheduler
```release-note
Adds new flag "pod-time-buffer-enabled"  with default value of `true` to maintain original behaviour.
Furthermore, CA will not wait for scheduler to mark pods as unschedulable and will evaluate non processed pods.
```


cluster-autoscaler/main.go
```diff
@@ -243,6 +243,7 @@ var (
 	maxAllocatableDifferenceRatio           = flag.Float64("max-allocatable-difference-ratio", config.DefaultMaxAllocatableDifferenceRatio, "Maximum difference in allocatable resources between two similar node groups to be considered for balancing. Value is a ratio of the smaller node group's allocatable resource.")
 	forceDaemonSets                         = flag.Bool("force-ds", false, "Blocks scale-up of node groups too small for all suitable Daemon Sets pods.")
 	dynamicNodeDeleteDelayAfterTaintEnabled = flag.Bool("dynamic-node-delete-delay-after-taint-enabled", false, "Enables dynamic adjustment of NodeDeleteDelayAfterTaint based of the latency between CA and api-server")
+	enablePodTimeBuffer                     = flag.Bool("pod-time-buffer-enabled", true, "Enables waiting for a time buffer so that the oldest unschedulable pod should be before starting scale up.")
```
x13n, line 246: We should control the entire logic with a flag, not just the "wait for more pods part". What I'd expect to see here is a single flag with two settings:
- Keep the time buffers, don't inject scheduler unprocessed pods in CA
- Drop the time buffest, start processing all unscheduled pods

cluster-autoscaler/core/podlistprocessor/filter_out_expendable.go
```diff
@@ -0,0 +1,66 @@
...
+	return unschedulablePods, nil
+}
+
+// addPreemptingPodsToSnapshot modifies the snapshot simulating scheduling of pods waiting for preemption.
+// this is not strictly correct as we are not simulating preemption itself but it matches
+// CA logic from before migration to scheduler framework. So let's keep it for now
+func (p *filterOutExpendable) addPreemptingPodsToSnapshot(pods []*apiv1.Pod, ctx *context.AutoscalingContext) error {
+	for _, p := range pods {
+		if err := ctx.ClusterSnapshot.AddPod(p, p.Status.NominatedNodeName); err != nil {
+			klog.Errorf("Failed to update snapshot with pod %s waiting for preemption", err)
```
x13n, line 58: Log message is missing %v for the error. Also, it would be useful to include pod namespace/name?

cluster-autoscaler/core/podlistprocessor/filter_out_expendable.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// NewFilterOutExpendablePodListProcessor creates a PodListProcessor filtering out expendable pods
+func NewFilterOutExpendablePodListProcessor() *filterOutExpendable {
+	return &filterOutExpendable{}
+}
+
+// Process filters out pods which are expendable and adds pods which is waiting for lower priority pods preemption to the cluster snapshot
+func (p *filterOutExpendable) Process(context *context.AutoscalingContext, pods []*apiv1.Pod) ([]*apiv1.Pod, error) {
+	klog.V(4).Infof("Filtering out expendable pods")
```
x13n, line 37: This will be logged every loop, what is the value of having it here?

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -451,24 +450,9 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
 
 	metrics.UpdateLastTime(metrics.Autoscaling, time.Now())
 
-	metrics.UpdateUnschedulablePodsCount(len(unschedulablePods))
-
-	unschedulablePods = tpu.ClearTPURequests(unschedulablePods)
```
x13n, line 456: Shouldn't this also be moved to a separate processor?

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -144,18 +144,63 @@ type PodLister interface {
 	List() ([]*apiv1.Pod, error)
 }
 
+func isScheduled(pod *apiv1.Pod) bool {
+	if pod == nil {
```
x13n, line 148: Why do we need to nil check? Shouldn't `pod` always be non-nil in these functions?

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -144,18 +144,63 @@ type PodLister interface {
...
+	}
+	return pod.Spec.NodeName != ""
+}
+func isDeleted(pod *apiv1.Pod) bool {
+	if pod == nil {
+		return false
+	}
+	return pod.GetDeletionTimestamp() != nil
+}
+func isUnschedulable(pod *apiv1.Pod) bool {
```
x13n, line 159: Shouldn't this be used by `UnschedulablePods`?

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -144,18 +144,63 @@ type PodLister interface {
...
+	for _, pod := range allPods {
+		// Make sure it's not scheduled or deleted
+		if isScheduled(pod) || isDeleted(pod) || isUnschedulable(pod) {
+			continue
+		}
+		// Make sure that if it's not scheduled it's either
+		// Not processed (condition is nil)
+		// Or Reason is empty (not schedulerError, terminated, ...etc)
+		_, condition := podv1.GetPodCondition(&pod.Status, apiv1.PodScheduled)
+		if condition == nil || (condition.Status == apiv1.ConditionFalse && condition.Reason == "") {
```
x13n, line 197: Do we need this extra check if the pod is neither scheduled, deleted nor unschedulable? It would be useful to have tests for this, especially if the answer is "yes".

x13n (changes requested):

cluster-autoscaler/main.go
atwamahmoud: Updated flag name to `ignore-scheduler-processing` & used it in unschedulable pod processing in static autoscaler

cluster-autoscaler/core/podlistprocessor/filter_out_expendable.go
atwamahmoud: Not much really, just wanted to be in sync with [filter_out_daemon_sets.go](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/podlistprocessor/filter_out_daemon_sets.go#L40)

- Removed

cluster-autoscaler/utils/kubernetes/listers.go
atwamahmoud: Yes, but it can be used in other functions as well, so it'd be safer to check it inside of the function.

cluster-autoscaler/utils/kubernetes/listers.go
atwamahmoud: For Unschedulable, we need the pod scheduling reason to be `PodReasonUnschedulable=Unschedulable`
However, we can have a scheduler error reason `PodReasonSchedulerError=SchedulerError`
In that case, the pod would be neither scheduled, deleted nor unschedulable (according to our definition), but  we don't want to include it as scheduler unprocessed because it's not.
In general our definition of "scheduler unprocessed" means the scheduler has not tried to schedule this pod yet. Which leaves out any non-empty pod scheduling reason

cluster-autoscaler/utils/kubernetes/listers.go
x13n: Yes, but it is private so these potential future functions would be in the same module, which limits the risk.

cluster-autoscaler/core/podlistprocessor/filter_out_expendable.go
```diff
@@ -0,0 +1,65 @@
...
+// NewFilterOutExpendablePodListProcessor creates a PodListProcessor filtering out expendable pods
+func NewFilterOutExpendablePodListProcessor() *filterOutExpendable {
+	return &filterOutExpendable{}
+}
+
+// Process filters out pods which are expendable and adds pods which is waiting for lower priority pods preemption to the cluster snapshot
+func (p *filterOutExpendable) Process(context *context.AutoscalingContext, pods []*apiv1.Pod) ([]*apiv1.Pod, error) {
+	nodes, err := context.AllNodeLister().List()
+	if err != nil {
+		return nil, err
```
x13n, line 39: Looks like static_autoscaler ignores the error here, let's make sure it is handled properly.

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -991,6 +991,118 @@ func TestStaticAutoscalerRunOnceWithFilteringOnUpcomingNodesEnabledNoScaleUp(t *
 		podDisruptionBudgetListerMock, daemonSetListerMock, onScaleUpMock, onScaleDownMock)
 }
 
+func TestStaticAutoscalerRunOnceWithSchedulerProcessingIgnored(t *testing.T) {
```
x13n, line 994: Can you make this test table-based? Possibly refactoring some of the existing tests.

cluster-autoscaler/main.go
```diff
@@ -243,6 +244,8 @@ var (
...
 	forceDaemonSets                         = flag.Bool("force-ds", false, "Blocks scale-up of node groups too small for all suitable Daemon Sets pods.")
 	dynamicNodeDeleteDelayAfterTaintEnabled = flag.Bool("dynamic-node-delete-delay-after-taint-enabled", false, "Enables dynamic adjustment of NodeDeleteDelayAfterTaint based of the latency between CA and api-server")
+	ignoreSchedulerProcessing               = flag.Bool("ignore-scheduler-processing", false, "If true, cluster autoscaler will not wait for scheduler to mark pods as unschedulable and will process both marked & non-marked pods (Schedulable pods will be filtered before scaling-up) it will also disable waiting for pod time buffers before triggering a scale-up.")
+	ignoredSchedulers                       = pflag.StringSlice("ignore-schedulers", []string{apiv1.DefaultSchedulerName}, fmt.Sprintf("Names of schedulers to be ignored if '--ignore-scheduler-processing' is set to true. default value '%s' is used", apiv1.DefaultSchedulerName))
```
x13n, line 248: Ditto to another comment about options: If we're providing a list of schedulers, is there value in adding yet another flag `ignore-scheduler-processing`?

Also, nit & optional: `ignore-schedulers` sounds scary :) `bypassed-scheduler-names`?

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -144,30 +144,92 @@ type PodLister interface {
...
+	}
+	_, condition := podv1.GetPodCondition(&pod.Status, apiv1.PodScheduled)
+	if condition == nil || condition.Status != apiv1.ConditionFalse || condition.Reason != apiv1.PodReasonUnschedulable {
+		return false
+	}
+	return true
+}
+
+// getIsDefaultSchedulerIgnored checks if the default scheduler should be ignored or not
+func getIsDefaultSchedulerIgnored(ignoredSchedulers map[string]bool) bool {
```
x13n, line 179: Instead of checking this for every pod, maybe just extend the map once at startup? `if ignoredSchedulers[apiv1.DefaultSchedulerName] { ignoredSchedulers[""] = true }`.

x13n (changes requested):

cluster-autoscaler/core/static_autoscaler_test.go
atwamahmoud: Updated tests & addsed a new function `testAutoscaler` to be used in other tests as well, also added a todo to refactor old tests to use the new `testAutoscaler` function

cluster-autoscaler/utils/kubernetes/listers.go
atwamahmoud: Updated `GetIgnoredSchedulersMap` to set that field

cluster-autoscaler/main.go
atwamahmoud: Yes, updated flags & also changed "ignore schedulers" to "bypass schedulers"

cluster-autoscaler/core/podlistprocessor/filter_out_expendable.go
atwamahmoud: Added a warning here, if nil is returned to static_autoscaler unschedulable pods would be empty and no scale up would occur so it's implicitly handled. (Same is also being done at `filterOutSchedulablePodListProcessor` without the new warning)

cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -277,4 +277,11 @@ type AutoscalingOptions struct {
 	// dynamicNodeDeleteDelayAfterTaintEnabled is used to enable/disable dynamic adjustment of NodeDeleteDelayAfterTaint
 	// based on the latency between the CA and the api-server
 	DynamicNodeDeleteDelayAfterTaintEnabled bool
+	//IgnoreSchedulerProcessing is used to signal whether CA will/won't wait
+	//for scheduler to mark pods as unschedulable and will process both marked & non-marked pods
+	//it will also signal whether we enable/disable waiting for pod time buffers before triggering a scale-up.
+	IgnoreSchedulerProcessing bool
```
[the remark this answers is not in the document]
atwamahmoud: True, updated

cluster-autoscaler/utils/kubernetes/listers.go
atwamahmoud: Updated & added a note comment for callers, still not sure about that decision though :')

cluster-autoscaler/config/autoscaling_options.go
```diff
@@ -277,11 +277,6 @@ type AutoscalingOptions struct {
...
 	// based on the latency between the CA and the api-server
 	DynamicNodeDeleteDelayAfterTaintEnabled bool
-	//IgnoreSchedulerProcessing is used to signal whether CA will/won't wait
-	//for scheduler to mark pods as unschedulable and will process both marked & non-marked pods
-	//it will also signal whether we enable/disable waiting for pod time buffers before triggering a scale-up.
-	IgnoreSchedulerProcessing bool
-	//IgnoredSchedulers are used to specify which schedulers to ignore their processing
-	//if IgnoreSchedulerProcessing is set to true
-	IgnoredSchedulers map[string]bool
+	//BypassedSchedulers are used to specify which schedulers to bypass their processing
```
x13n, line 280: nit: Add a space after `//` to keep comments consistent.

cluster-autoscaler/utils/kubernetes/listers.go
```diff
@@ -193,18 +181,12 @@ func ScheduledPods(allPods []*apiv1.Pod) []*apiv1.Pod {
...
 
-	isDefaultSchedulerIgnored := getIsDefaultSchedulerIgnored(ignoredSchedulers)
-
 	for _, pod := range allPods {
-		// Don't add a pod with a scheduler that isn't specified by the user
-		if !isDefaultSchedulerIgnored && pod.Spec.SchedulerName == "" {
-			continue
-		}
-		if isIgnored, found := ignoredSchedulers[pod.Spec.SchedulerName]; !found || !isIgnored {
+		if isIgnored, found := bypassedSchedulers[pod.Spec.SchedulerName]; !found || !isIgnored {
```
x13n, line 189: Checking for presence of key in map is redundant, `isIgnored` always implies `found`. Besides, we're only storing `true` values there.

Also, nit: rename `isIgnored` to `canBypass`?

cluster-autoscaler/core/podlistprocessor/filter_out_expendable.go
x13n: I don't think the warning and returning the error at the same time is useful. Errors should either be handled or passed back to the caller, logging & returning tries to do both. Often error handling is responsibility of the caller when function can return `error` type, but I see a problem here - the return value is ignored: https://github.com/kubernetes/autoscaler/blob/dcaba4483d512f4a93b1e60c7272e1de8fe181a9/cluster-autoscaler/core/static_autoscaler.go#L518

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -992,115 +1108,123 @@ func TestStaticAutoscalerRunOnceWithFilteringOnUpcomingNodesEnabledNoScaleUp(t *
...
-		MaxCoresTotal:             10,
-		MaxMemoryTotal:            100000,
-		IgnoreSchedulerProcessing: true,
-		IgnoredSchedulers: map[string]bool{
+		EstimatorName:    estimator.BinpackingEstimatorName,
+		ScaleDownEnabled: true,
+		MaxNodesTotal:    10,
+		MaxCoresTotal:    10,
+		MaxMemoryTotal:   100000,
+		BypassedSchedulers: map[string]bool{
```
x13n, line 1125: Can you use `GetBypassedSchedulersMap` here? Right now this list contains default scheduler, but not empty scheduler name, which is a config we will never see in practice.

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -992,115 +1108,123 @@ func TestStaticAutoscalerRunOnceWithFilteringOnUpcomingNodesEnabledNoScaleUp(t *
...
+			nodeGroups: []*nodeGroup{{
+				name:  "ng1",
+				min:   1,
+				max:   10,
+				nodes: []*apiv1.Node{n1},
+			}},
+			daemonSetListerCallTimes: 1,
+			pdbListerCallTimes:       1,
+			expectedScaleUps:         []*scaleCall{},
+			now:                      time.Now(),
```
x13n, line 1154: Each `time.Now()` call returns a slightly different timestamp, which leads me to think we don't really need to include all these values in each test case?

cluster-autoscaler/core/static_autoscaler_test.go
x13n: I'm not sure if one huge testing function is the way to go. The actual assertions/checks belong in the test function, not some humongous utility. If you want to abstract some of it, I think the test setup could be abstracted, but I'd make it clear what the testing function is actually testing (i.e. keep the checks there).

x13n (changes requested):

cluster-autoscaler/core/podlistprocessor/filter_out_expendable.go
atwamahmoud: Added warnings @ static_autoscaler & removed the one in processor. The processor now returns a formatted error

cluster-autoscaler/utils/kubernetes/listers.go
atwamahmoud: Yes, updated now

cluster-autoscaler/config/autoscaling_options.go
atwamahmoud: Done

cluster-autoscaler/core/static_autoscaler_test.go
atwamahmoud: Done

cluster-autoscaler/core/static_autoscaler_test.go
atwamahmoud: Yes it was made so that I'm consistent with other tests so the original "testAutoscaler" wouldn't need much updates to be used for other tests, updated now

x13n:
Thanks! Can you squash the commits?

/approve

x13n:
Thanks!

/lgtm

PR #6245, opened 2023-11-03
AWS: cache instance requirements

alexanderConstantinescu:
#### What type of PR is this?

We can discuss this, but I am going to go ahead and label this:

/kind bug

since it forces users of the CA to re-configure the CA with a work-around that is sub-optimal, see below for more information. 

The cluster-autoscaler on AWS currently builds a representation of the Node object during each scale up/down computation. This computation is performed depending on the `scan-interval` defined. Parts of this computation requires setting the Node capacity. The capacity is built by fetching instance requirements from the AWS API. For clusters which use mixed instance policy and which specify a launch template, the API call made is `DescribeLaunchTemplateVersions`. This information does not need to be fetched during each scale up/down computation, but can instead be cached along with all the other information required by the cluster-autoscaler running on AWS. By doing this the cluster-autoscaler can reduce the amount of API calls performed. If the `scan-interval` is 10s; the amount of API calls to this endpoint will be: 6 calls per minute, or 8640 calls per day. This obviously incurs unnecessary costs for people operating AWS clusters at scale and moreover: only leaves them with the possibility of increasing the `scan-interal` as to reduce the amount of API calls. This is not great since that has secondary impacts on how quickly the CA can react to changes on the cluster. The current `refreshInterval` is set to one minute and defines the interval at which the CA refreshes its cache, so by caching the instance requirements we can make sure that this API endpoint is only called once per minute, regardless of the `scan-interval` chosen.  

This PR therefore caches the instance requirements as done with all the other `MixedInstancesPolicy` data, and uses the cached data when executing `buildNodeFromTemplate`. 

Fixes #

```release-note
None
```

```docs

```


cluster-autoscaler/cloudprovider/aws/aws_manager_test.go
```diff
@@ -521,6 +521,22 @@ func TestBuildNodeFromTemplate(t *testing.T) {
...
+				Min: aws.Int64(4),
+				Max: aws.Int64(8),
+			},
+			AcceleratorTypes:         []*string{aws.String(autoscaling.AcceleratorTypeGpu)},
+			AcceleratorManufacturers: []*string{aws.String(autoscaling.AcceleratorManufacturerNvidia)},
+			AcceleratorCount: &ec2.AcceleratorCount{
+				Min: aws.Int64(4),
+				Max: aws.Int64(8),
+			},
+		},
```
gjtempleton, line 539: Is this actually testing anything at the moment given the overrides override every element of the requirements?

alexanderConstantinescu: Good point, the test didn't make much sense with my changes. Previously this test specified `instanceRequirementsOverrides` as to make sure  the code ended up calling https://github.com/kubernetes/autoscaler/pull/6245/files#diff-22984a3a02b16ff49b2a94a43b49f3aa61c856483c3612b06b69dd51e347746fL392-L396 

This will now be performed on the cache side, so what I added was just confusing for the readability of the test. Instead I have now changed the test to use `instanceRequirements` which will be populated using `instanceRequirementsOverrides` on the cache side. The test surface should thus remain the same.

PR #6248, opened 2023-11-06
liveness probe for vpa-admission

jklaw90:
#### What type of PR is this?

/kind bug

* Adding a liveness probe 
* Handle sigint/sigterm and fixed shutdown to be graceful 
* update timer to ticker. for more consistent timing on updates.

Fixes #

```release-note
None
```

```docs
None
```

### DIFF
--- a/vertical-pod-autoscaler/deploy/admission-controller-deployment.yaml
+++ b/vertical-pod-autoscaler/deploy/admission-controller-deployment.yaml
@@ -42,6 +42,14 @@ spec:
             - containerPort: 8000
             - name: prometheus
               containerPort: 8944
+          livenessProbe:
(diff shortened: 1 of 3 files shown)


vertical-pod-autoscaler/deploy/admission-controller-deployment.yaml
```diff
@@ -42,6 +42,14 @@ spec:
             - containerPort: 8000
             - name: prometheus
               containerPort: 8944
+          livenessProbe:
+            failureThreshold: 3
+            httpGet:
+              path: /health-check
+              port: prometheus
```
jklaw90, line 49: not sure if we want to move this to the regular http port but this is where it is currently.

PR #6251, opened 2023-11-06
Fix '.spec.updatePolicy.evictionRequirements.resources' to be plural in yaml

voelzmo:
#### What type of PR is this?

/kind bug

Fix '.spec.updatePolicy.evictionRequirements.resources' to be plural in yaml. Seems like this slipped through the cracks in #5599. We didn't release the CRD yet, so changing it shouldn't be an issue?

### DIFF
--- a/vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml
+++ b/vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml
@@ -283,9 +283,11 @@ spec:
               resourcePolicy:
                 description: Controls how the autoscaler computes recommended resources.
                   The resource policy may be used to set constraints on the recommendations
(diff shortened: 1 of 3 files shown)


mwielgus:
Just to be on the safe side, this part of api has not been actually released yet, right?

voelzmo:
> Just to be on the safe side, this part of api has not been actually released yet, right?

Correct, that's what I tried to say in the first post of the PR
> We didn't release the CRD yet, so changing it shouldn't be an issue?

voelzmo:
Before we do a new release, can we please merge this to avoid modifying a released CRD?
I'm asking because I stumbled upon https://github.com/kubernetes/autoscaler/pull/6276
/cc @sophieliu15

sophieliu15:
@kwiesmueller Kevin, could you review and merge this change?

PR #6261, opened 2023-11-08
Fix capacityType label in AWS ManagedNodeGroup

sibucan:
#### What type of PR is this?
/kind bug

When using a ManagedNodeGroup (MNG), CAS is supposed to use the EKS's DescribeNodegroup API call to [get the information about the nodes in the NodeGroup](https://github.com/kubernetes/autoscaler/blob/e9a698f33ddba270ff13b780450dd4e77dfd78cd/cluster-autoscaler/cloudprovider/aws/aws_manager.go#L306C3-L306C3) to know where a Pod is supposed to go if no nodes currently exist for the MNG. This is because there's currently no native way of setting tags on a MNG's underlying ASG using the MNG API, and CAS traditionally depended on the ASG's tags to be able to tell when it should scale up from 0 when a Pod could potentially fit there.

However, the `capacityType` label that CAS gets from the DescribeNodegroup API call is missing a crucial prefix. A Spot instance node will have the `eks.amazonaws.com/capacityType` label placed on it, but when CAS gets the node label list from the API, it'll look like this:

```
I1005 17:40:24.332839       1 aws_manager.go:310] node.Labels : map[amiType:BOTTLEROCKET_x86_64 capacityType:SPOT cluster-name:eks01 eks.amazonaws.com/nodegroup:bottlerocket_x86_spot-20231003194631949400000003 k8sVersion:1.28 kubernetes.io/arch:amd64 kubernetes.io hostname:eks-bottlerocket_x86_spot-20231003194631949400000003-bcc57b89-f4ef-9c63-7c70-b4ab5325eb78-asg-6921437539965377742 kubernetes.io/os:linux node.kubernetes.io/instance-type:m5.xlarge nodegroup-name:bottlerocket_x86_spot-20231003194631949400000003 topology.kubernetes.io/region:us-east-1 topology.kubernetes.io/zone:us-east-1a]
```

Note that the label CAS determined from the API call was:
>`capacityType:SPOT`

In a real node, the labels look like this (taken from `kubectl describe node`):
```
Labels:             beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/instance-type=m6i.xlarge
                    beta.kubernetes.io/os=linux
                    eks.amazonaws.com/capacityType=SPOT
                    eks.amazonaws.com/nodegroup=bottlerocket_x86_spot-20231003194631949400000003
                    eks.amazonaws.com/nodegroup-image=ami-09c9ab15237fd5960
                    eks.amazonaws.com/sourceLaunchTemplateId=lt-0d333d9b550f53aad
                    eks.amazonaws.com/sourceLaunchTemplateVersion=3
                    failure-domain.beta.kubernetes.io/region=us-east-1
                    failure-domain.beta.kubernetes.io/zone=us-east-1a
                    kubernetes.io/arch=amd64
                    kubernetes.io/hostname=ip-10-250-242-179.ec2.internal
                    kubernetes.io/os=linux
                    node.kubernetes.io/instance-type=m6i.xlarge
                    topology.ebs.csi.aws.com/zone=us-east-1a
                    topology.kubernetes.io/region=us-east-1
                    topology.kubernetes.io/zone=us-east-1a
```

So if I tried to schedule a Pod using the following nodeaffinity:
```
      affinity:
        nodeAffinity:
            nodeSelectorTerms:
              - matchExpressions:
                - key: "eks.amazonaws.com/capacityType"
                  operator: In
                  values: ["SPOT"]
```
The CAS will see that my Spot MNG (with discovered label capacityType: SPOT) will "never" be able fit my Pod that requests `eks.amazonaws.com/capacityType: SPOT`, and says "we can't fit your Pod in the cluster":

```
I1005 17:56:50.997360       1 orchestrator.go:546] Pod autoscaler-test/test-59f845b4f8-qvtll can't be scheduled on eks-bottlerocket_x86_spot-20231003194631949400000003-bcc57b89-f4ef-9c63-7c70-b4ab5325eb78, predicate checking error: node(s) didn't match Pod's node affinity/selector; predicateName=NodeAffinity; reasons: node(s) didn't match Pod's node affinity/selector; debugInfo=
```

However, If I used the "wrong" label `capacityType: SPOT`, CAS will autoscale the correct MNG, but my Pod then stays Pending because the label is missing the `eks.amazonaws.com` prefix and so it can't actually go into that node. 😢 

The fix in here makes it so the autodiscovered label is the same one that's placed on the nodes once the MNG scales up for the first time.

**Note:** This does not happen when there's at least one node in the MNG, as CAS will then cache the correct set of labels, since it will circumvent the need for EKS's DescribeNodegroup API.

Fixes #

```release-note
Fixed an issue where the capacityType label inferred from an empty AWS ManagedNodeGroup does not match the same label on the nodes after it scales from 0 -> 1.
```


cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
```diff
@@ -85,7 +85,7 @@ func (m *awsWrapper) getManagedNodegroupInfo(nodegroupName string, clusterName s
 	}
 
 	if r.Nodegroup.CapacityType != nil && len(*r.Nodegroup.CapacityType) > 0 {
-		labels["capacityType"] = *r.Nodegroup.CapacityType
+		labels["eks.amazonaws.com/capacityType"] = *r.Nodegroup.CapacityType
```
vadasambar, line 88: Should we abstract this label to a const variable and use it everywhere so that it is easy to change if the describe API starts using the correct label in the future?

cluster-autoscaler/cloudprovider/aws/aws_wrapper_test.go
```diff
@@ -174,7 +174,7 @@ func TestGetManagedNodegroup(t *testing.T) {
 	assert.Equal(t, labelMap[labelKey2], labelValue2)
 	assert.Equal(t, labelMap["diskSize"], strconv.FormatInt(diskSize, 10))
 	assert.Equal(t, labelMap["amiType"], amiType)
-	assert.Equal(t, labelMap["capacityType"], capacityType)
+	assert.Equal(t, labelMap["eks.amazonaws.com/capacityType"], capacityType)
```
vadasambar, line 177: Same comment as https://github.com/kubernetes/autoscaler/pull/6261/files#r1394608346

vadasambar (changes requested):

vadasambar:
A nit, LGTM otherwise. Description is easy to understand (thank you)

vadasambar (approved):

vadasambar:
@drmorr, @gjtempleton need your approval.

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
sibucan: Hm, would you mind rephrasing the question? The NodeGroupDescribe API for capacity type only returns a value (in this case, either `SPOT` or `ON_DEMAND`), and this change here only adds the prefix that's used on the node, so that the label map cache that gets pre-populated at startup actually matches the labels added on the node. The API call isn't really using an "incorrect" label, it was just being indexed wrong on the map 😅

sibucan: That being said, using a variable to save the "label key" does make sense! But I'd preferably like to do that as part of a separate PR so that it encompasses all the labels used by the "GetNodeGroupInfo()" function, as a larger refactor.

vadasambar: > The NodeGroupDescribe API for capacity type only returns a value (in this case, either SPOT or ON_DEMAND)

You are right. Please ignore my comment.

vadasambar: > But I'd preferably like to do that as part of a separate PR so that it encompasses all the labels used by the getManagedNodegroupInfo() function, as a refactor of sorts. 😄

Sounds good to me :+1:

gjtempleton:
Thanks for this!
/lgtm
/approve

Issue #6096, opened 2023-09-07
"Failed to check cloud provider" flooding logs


grosser:
**Which component are you using?**:

cluster-autoscaler

**What version of the component are you using?**:

1.27.3 with patch https://github.com/kubernetes/autoscaler/pull/5887

**What k8s version are you using (`kubectl version`)?**:

v1.27.2

**What environment is this in?**:

aws

**What did you expect to happen?**:

no log spam

**What happened instead?**:

log spam

```
{"level":"WARN","message":"Failed to check cloud provider has instance for ip-xxx.us-west-2.compute.internal: node is not present in aws: could not find instance {aws:///us-west-2b/i-xyz i-xyz}"}
```

seems like 1 line for every instance in use

**How to reproduce it (as minimally and precisely as possible)**:

unclear

**Anything else we need to know?**:

seems very similar to https://github.com/kubernetes/autoscaler/issues/5842 but the patch for it did not solve our issue

mro-amboss:
I have CA 1.27.2 and having the same error.
Kubernetes version: 1.24.
Env: AWS EKS

brydoncheyney:
Observing the same issue. The CA is logging each non-ASG node every control loop. We're provisioning "platform" nodes on Managed Node Group instances, and use karpenter to provision "worker" nodes, which should be excluded from the check.

CA: 1.27.2
Kubernetes version: 1.28
Env: AWS EKS

shapirus:
Same issue here with CA 1.28.0 / k8s 1.28.2.

The ASGs are managed by kops.

```
I1017 10:21:10.920315       1 cluster.go:175] node i-06de46c41ffb0aaaa is not suitable for removal: can reschedule only 0 out of 1 pods
[at this point all is working fine]
...
I1017 10:25:01.973574       1 auto_scaling_groups.go:393] Regenerating instance to ASG map for ASG names: [ list-of-all-ASGs ]
I1017 10:25:02.099392       1 auto_scaling_groups.go:400] Regenerating instance to ASG map for ASG tags: map[]
I1017 10:25:02.100362       1 auto_scaling_groups.go:142] Updating ASG <ASG-that-contains-the-node-in-question>
...
W1017 10:27:32.735121       1 clusterstate.go:1033] Failed to check cloud provider has instance for i-06de46c41ffb0aaaa: node is not present in aws: could not find instance {aws:///eu-central-1c/i-06de46c41ffb0aaaa i-06de46c41ffb0aaaa}
[...]

PR #6265, opened 2023-11-09
allow users to avoid aws instance not found spam

grosser:
#### What type of PR is this?

/kind bug

allows users to mark nodes for "do not spam me about missing aws instance"

fixes https://github.com/kubernetes/autoscaler/issues/6096

not ready to be merged, just for discussion

```release-note
TODO
```

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
@@ -139,6 +139,14 @@ func (aws *awsCloudProvider) HasInstance(node *apiv1.Node) (bool, error) {
 		return true, cloudprovider.ErrNotImplemented
 	}
 
+	// avoid log spam for not autoscaled asgs:
+	//   Nodes that belong to an asg that is not autoscaled will not be found in the asgCache below,
+	//   so do not trigger warning spam by returning an error from being unable to find them.
+	//   Annotation is not automated, but users that see the warning can add the annotation to avoid it.
+	if node.Annotations != nil && node.Annotations["k8s.io/cluster-autoscaler/enabled"] == "false" {
+		return false, nil
+	}
+
 	awsRef, err := AwsRefFromProviderId(node.Spec.ProviderID)
 	if err != nil {
 		return false, err
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider_test.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider_test.go
@@ -723,4 +723,19 @@ func TestHasInstance(t *testing.T) {
 	assert.ErrorContains(t, err, nodeNotPresentErr)
 	assert.False(t, present)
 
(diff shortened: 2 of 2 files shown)


cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -139,6 +139,13 @@ func (aws *awsCloudProvider) HasInstance(node *apiv1.Node) (bool, error) {
 		return true, cloudprovider.ErrNotImplemented
 	}
 
+	// nodes that belong to an asg that is not autoscaled will not be found in the asgCache below
+	// do not spam warnings about them being unable to be found
+	// users that see this error can add an annotation to their nodes to avoid it
+	if node.Annotations != nil && node.Annotations["k8s.io/cluster-autoscaler/enabled"] == "false" {
```
drmorr0, line 145: Can we add a test confirming this works correctly?

drmorr0 (changes requested):

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
grosser: yes, I can add tests if anyone with merge permission can confirm this is a good approach

drmorr0: This seems fine to me, yes.

Rather: imo the _ideal_ approach would be to just have that error be logged at a lower verbosity, but I think that's going to be non-trivial to accomplish given how the code is currently organized, so this seems fine as an alternate solution.

grosser:
improved comment, added test

brydoncheyney:
`k8s.io/cluster-autoscaler/enabled` is not a valid annotation key - https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set - which expects only a _single_ DNS prefix.

Attempting to manually add this annotation to an existing node will fail -

```
# nodes "ip-XXX-XXX-XXX-XXX.ec2.internal" was not valid:
# * metadata.annotations: Invalid value: "k8s.io/cluster-autoscaler/enabled": a qualified name must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]') with an optional DNS subdomain prefix and '/' (e.g. 'example.com/MyName')
```

Note that the use of the `k8s.io/cluster-autoscaler/enabled` _tag_ is valid within AWS.

This could be fixed by using an alternative key format, such as `k8s.io/cluster-autoscaler-enabled`

I'll pull a PR together for this.

grosser:
thx for checking/fixing! :)

scottblack1:
Curious why we have used node.annotation here instead of a label for example?

As far as I can see there is no easy way to automatically annotate nodes in EKS as this is not an available argument in the eks_bootstrap.sh.

PR #6273, opened 2023-11-13
Stop (un)tainting nodes from unselected node groups.

fische:
#### What type of PR is this?
/kind bug

This PR fixes a bug where at the start of `RunOnce` and during scale down, the code cleans DeletionCandidate taints from pretty much all nodes, even from those that are not part of the selected node groups. By selected node groups, I mean the ones passed through the `--nodes` flag. This is obviously undesired behaviour but I'd like to give a bit more context around how we came across this issue:
We are currently running our cluster on GKE which comes with its own cluster autoscaler. It lacks loads of options, so we have deployed our own CA and disabled the managed one on all node pools, as there is no way to completely remove it. This means we have 2 CAs running at the same time, even though one is not supposed to do anything. Given the bug I described above, they both are conflicting because while one (ours) tries to scale down some nodes, the other (GKE's) will remove the DeletionCandidate taint from those as from its perspective those nodes should not be scaled down even though their node groups have not been selected. This makes scale down **very** slow.

To give you a better idea, with the test case I've added, it fails with the following on master:
```
                Error Trace:    /home/fische/local/autoscaler/cluster-autoscaler/core/static_autoscaler_test.go:372
                Error:          Not equal:
                                expected: []v1.Taint{v1.Taint{Key:"DeletionCandidateOfClusterAutoscaler", Value:"1699914797", Effect:"PreferNoSchedule", TimeAdded:<nil>}}
                                actual  : []v1.Taint{}

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,8 +1,2 @@
                                -([]v1.Taint) (len=1) {
                                - (v1.Taint) {
                                -  Key: (string) (len=36) "DeletionCandidateOfClusterAutoscaler",
                                -  Value: (string) (len=10) "1699914797",
                                -  Effect: (v1.TaintEffect) (len=16) "PreferNoSchedule",
                                -  TimeAdded: (*v1.Time)(<nil>)
                                - }
                                +([]v1.Taint) {
                                 }
                Test:           TestStaticAutoscalerRunOnce
```

I haven't raised any issue for this. Shall I?

I'm just not sure what we should do if there's an error filtering the nodes during scale down. Shall we continue? I've added a comment in the code in case you don't see what I'm talking about.

This does somewhat introduce a user-facing change, yes, so I've added a release note, even though it is behaviour that users should not rely on.

```release-note
Fixed undesired behaviour at the beginning of autoscaler main logic and during scale down, where it would clean taints from nodes that are not part of selected node groups (--nodes flag).
```

```docs

```


Shubham82:
@fische, you have to sign the CLA before the PR can be reviewed.

Shubham82:
To check EasyCLA

/easycla

fische:
> @fische, you have to sign the CLA before the PR can be reviewed. See the following document to sign the CLA: [Signing Contributor License Agreements(CLA)](https://github.com/kubernetes/autoscaler/blob/master/CONTRIBUTING.md#signing-contributor-license-agreementscla)

Yup, sorry, I forgot about that.

fische:
@Shubham82 Would you mind taking a look now that the CLA has been signed please? :pray:

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -967,6 +974,19 @@ func (a *StaticAutoscaler) obtainNodeLists(cp cloudprovider.CloudProvider) ([]*a
 	return allNodes, readyNodes, nil
 }
 
+func filterNodeGroups(cp cloudprovider.CloudProvider, nodes ...*apiv1.Node) ([]*apiv1.Node, caerrors.AutoscalerError) {
```
BigDarkClown, line 977: nit: This method does not filter node groups, it filters nodes that have an associated node group. Please rename it so it better reflects this.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -665,6 +667,11 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
 				scaleDownStatus.Result == scaledownstatus.ScaleDownNoUnneeded) &&
 				a.AutoscalingContext.AutoscalingOptions.MaxBulkSoftTaintCount != 0 {
 				taintableNodes := a.scaleDownPlanner.UnneededNodes()
+				allNodes, err := filterNodeGroups(a.CloudProvider, allNodes...)
+				if err != nil {
+					klog.Warningf("Failed filtering nodes based on selected node groups: %v", err)
+				}
+				// REVIEW: Shall we do the below calls if there's an error? This is essentially gonna mean that untaintableNodes is empty since allNodes is empty.
```
BigDarkClown, line 674: Two main comments:
- I don't think the failure mode for err should be "don't clean the taints at all", I think it should be "clean all the nodes". In other words - if the filter returns an error, then I think we should actually use the previous allNodes value. This applies to the previous case as well.
- After these changes taintableNodes used here may "theoretically" be something than is not a strict subset of filtered nodes? Just for sanity check I would limit the taintableNodes to the ones present in filtered nodes as well.

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -172,6 +173,13 @@ func TestStaticAutoscalerRunOnce(t *testing.T) {
 	SetNodeReadyState(n2, true, time.Now())
 	n3 := BuildTestNode("n3", 1000, 1000)
 	n4 := BuildTestNode("n4", 1000, 1000)
+	n5 := BuildTestNode("n5", 1000, 1000)
```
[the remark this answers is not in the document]
fische: Yeah, you're right on the separate test. I was initially not sure this was best given multiple cases are tested per test.

cluster-autoscaler/core/static_autoscaler.go
fische: I felt like based on the whole function signature it made sense, but sure I can make this even more explicit.

cluster-autoscaler/core/static_autoscaler.go
fische: > I don't think the failure mode for err should be "don't clean the taints at all", I think it should be "clean all the nodes". In other words - if the filter returns an error, then I think we should actually use the previous allNodes value. This applies to the previous case as well.

Sounds good, going to make the change :+1: 

> After these changes taintableNodes used here may "theoretically" be something than is not a strict subset of filtered nodes? Just for sanity check I would limit the taintableNodes to the ones present in filtered nodes as well.

Yeah "theoretically" the planner should already take into account the selected node groups, but I agree it's a good idea to make sure this is the case. I'll add that.

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -1120,6 +1121,129 @@ func TestStaticAutoscalerRunOnceWithFilteringOnUpcomingNodesEnabledNoScaleUp(t *
...
+	}
+
+	// Node from selected node groups can get their deletion candidate taints removed.
+	readyNodeLister.SetNodes([]*apiv1.Node{n1})
+	allNodeLister.SetNodes([]*apiv1.Node{n1})
+	allPodListerMock.On("List").Return([]*apiv1.Pod{p1}, nil).Twice()
+	daemonSetListerMock.On("List", labels.Everything()).Return([]*appsv1.DaemonSet{}, nil)
+	podDisruptionBudgetListerMock.On("List").Return([]*policyv1.PodDisruptionBudget{}, nil)
+
+	err = autoscaler.RunOnce(time.Now().Add(5 * time.Hour))
```
x13n, line 1227: Can you make this test table-based? Also - optional - there's a ton of dependencies here, which suggests maybe it would be cleaner to migrate to a new file that doesn't require most of the setup above and can be tested more easily.

x13n (changes requested):

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -964,6 +983,19 @@ func (a *StaticAutoscaler) obtainNodeLists(cp cloudprovider.CloudProvider) ([]*a
 	return allNodes, readyNodes, nil
 }
 
+func filterNodesFromSelectedGroups(cp cloudprovider.CloudProvider, nodes ...*apiv1.Node) ([]*apiv1.Node, caerrors.AutoscalerError) {
+	filtered := make([]*apiv1.Node, 0, len(nodes))
+	for _, n := range nodes {
+		if ng, err := cp.NodeGroupForNode(n); err != nil {
+			klog.Errorf("Failed to get a node group node node: %v", err)
+			return nil, caerrors.ToAutoscalerError(caerrors.CloudProviderError, err)
```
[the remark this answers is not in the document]
fische: I've removed the return :+1:

cluster-autoscaler/core/static_autoscaler_test.go
fische: Yeah I already tried to remove a bunch of those dependencies but given I don't have much knowledge about all the interactions, I wasn't sure about exactly what needed to be removed.
Regarding the table test part, I can do that but that will mean there are 2 different testing styles in the codebase, so I'm not sure if you're OK with that? Maybe you're already planning on moving to table tests?

x13n: Yeah, moving everything to table based is a lot of effort, so for now we're just encouraging this approach in new or modified tests.

x13n: And by getting rid of dependencies I meant factoring out this code to a dedicated package/object. You don't really need to understand all of the deps of static_autoscaler - simply depend only on what your new code needs to run.

fische: Right, I've made the test table-based and tried removing as many deps as possible. Unfortunately I can't use the setupAutoscaler as it does not allow the caller to set a specific ClientSet, which I need in this test.

fische:
@x13n Could you please have another look?

x13n:
Apologies for slow review. LGTM now.

/lgtm
/approve

VelorumS:
@fische I think that this MR is incomplete. Should also add a fix for the `MaxBulkSoftTaintCount == 0` : https://github.com/kubernetes/autoscaler/pull/6273/files#diff-5d3331e59ebce666578d33c0c52f7686b92ff18bcab405eb8eab637881f78080L235

PR #6275, opened 2023-11-14
Remove gce-expander-ephemeral-storage-support flag

azylinski:
#### What type of PR is this?

/kind cleanup

Always enable the ephemeral-storage feature

```release-note
Flag --gce-expander-ephemeral-storage-support is now Deprecated. The ephemeral-storage support is always enabled and the flag itself would be ignored.
```

/assign @x13n

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
@@ -374,7 +374,7 @@ func BuildGCE(opts config.AutoscalingOptions, do cloudprovider.NodeGroupDiscover
 		klog.Fatalf("Failed to create GCE Manager: %v", err)
 	}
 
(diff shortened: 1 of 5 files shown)


azylinski:
> > Does this PR introduce a user-facing change?
> 
> Yes, it will crash CA if anyone is setting this flag explicitly.

Fair enough. Let me leave the flag, but:
* ignore it's value passed
* mark as deprecated / to be deleted

x13n:
O, I mainly meant changes like this should be mentioned in release notes  - instead of the current `NONE` release note. Please update it to mention that this flag became deprecated.

azylinski:
> O, I mainly meant changes like this should be mentioned in release notes - instead of the current `NONE` release note. Please update it to mention that this flag became deprecated.

Right. I've updated the release notes.

x13n:
Thanks!

Issue #6281, opened 2023-11-15
Unexpected scaling behaviour when using cluster-autoscaler as a library


mads-hartmann:
**Which component are you using?**:

We're using the cluster-autoscaler in our custom autoscaler-expander implementation

**What version of the component are you using?**:

Component version: k8s.io/autoscaler/cluster-autoscaler v0.0.0-20230811122327-e1b03fac9958

**What k8s version are you using (`kubectl version`)?**:

I don't think the k8s version is relevant. See rest of issue for more context.

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version

</pre></details>

**What environment is this in?**:

AWS

**What did you expect to happen?**:

We're invoking `NodeGroupIncreaseSize` with a delta of `1`. To the best of our knowledge the current desired capacity is `7` at the time. I would expect it to to increase the desired capacity to `8`.

**What happened instead?**:

It increased it to `71`. A bit more context

We're experiencing an odd scaling behaviour when using cluster-autoscaler as a library from a custom autoscaler expander. Our custom autoscaler-expander is invoking `NodeGroupIncreaseSize` with a delta of `1` but somehow that makes it want to scale the node group from `7` to `71`. See screenshots:

![image](https://github.com/kubernetes/autoscaler/assets/83561/631193d2-a17e-47f2-bf05-d0b9cd25c340)
![image-2](https://github.com/kubernetes/autoscaler/assets/83561/5e963e92-ea58-468c-83c1-d708069ff31b)

While my gut reaction was that this feels like a string concatenation bug ("7" + 1 = "71") I can't finde any evidence of that in the code.

So the only explanation I can see is that AWS somehow returned `70` as the current taget size. However, we don't see any evidence in CloudTrail. We see two SetDesiredCapacity events, first one with 7 and then 71.

**How to reproduce it (as minimally and precisely as possible)**:

This can't be easily reproduced outside of our setup I'm afraid.

**Anything else we need to know?**:

PR #6282, opened 2023-11-15
Add min/max/current asg size to log

mads-hartmann:
#### What type of PR is this?

/kind bug

We're experiencing an odd scaling behaviour when using cluster-autoscaler as a library from a custom autoscaler expander. Our custom autoscaler-expander is invoking `NodeGroupIncreaseSize` with a delta of `1` but somehow that makes it want to scale the node group from 7 to 71. See https://github.com/kubernetes/autoscaler/issues/6281 for details.

The only explanation I can see is that AWS somehow returned `70` as the current taget size. We don't see any evidence in CloudTrail this though. In CloudTrail we only have SetDesiredCapacity of 7 and then 71.

This PR logs out the minimum, maximum, and current size of the auto scaling group when reading the values from AWS. This information would have helped us understand where the jump from 7 to 70+1 is coming from.

Part of https://github.com/kubernetes/autoscaler/issues/6281

```release-note
Adds additional log output when updating the ASG information form AWS
```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
+++ b/cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
@@ -139,8 +139,6 @@ func (m *asgCache) register(asg *asg) *asg {
 			return existing
 		}
 
-		klog.V(4).Infof("Updating ASG %s", asg.AwsRef.Name)
-
 		// Explicit registered groups should always use the manually provided min/max
 		// values and the not the ones returned by the API
 		if !m.explicitlyConfigured[asg.AwsRef] {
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/cloudprovider/aws/auto_scaling_groups.go
```diff
@@ -158,6 +156,8 @@ func (m *asgCache) register(asg *asg) *asg {
 		existing.MixedInstancesPolicy = asg.MixedInstancesPolicy
 		existing.Tags = asg.Tags
 
+		klog.V(4).Infof("Updated ASG %s. min/max/current is %d/%d/%d", asg.AwsRef.Name, existing.minSize, existing.maxSize, existing.curSize)
```
mwielgus, line 159: This log suggests that ASG was updated on the cloud provider side.

mads-hartmann: Ah right. How about "Updated ASG cache for %s" instead?

mwielgus (approved):
/lgtm
/approve

PR #6284, opened 2023-11-15
Allow overriding domain suffix in GCE cloud provider.

olagacek:
#### What type of PR is this?

/kind feature

This PR allows overriding Domain Suffix in calls made to GCE cloud provider. If it's not specified the default domain suffix is used.

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/gce_url.go
+++ b/cluster-autoscaler/cloudprovider/gce/gce_url.go
@@ -18,16 +18,13 @@ package gce
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 )
 
 const (
-	gceUrlSchema    = "https"
-	gceDomainSuffix = "googleapis.com/compute/v1/projects/"
-	// Cluster Autoscaler previously used "content" instead of "www" here, for reasons unknown.
-	gcePrefix           = gceUrlSchema + "://www." + gceDomainSuffix
-	instanceUrlTemplate = gcePrefix + "%s/zones/%s/instances/%s"
-	migUrlTemplate      = gcePrefix + "%s/zones/%s/instanceGroups/%s"
+	projectsSubstring = "/projects/"
+	defaultDomainUrl  = "https://www.googleapis.com/compute/v1"
 )
 
 // ParseMigUrl expects url in format:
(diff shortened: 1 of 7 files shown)


cluster-autoscaler/cloudprovider/gce/gce_url.go
```diff
@@ -64,24 +61,36 @@ func ParseInstanceUrlRef(url string) (GceRef, error) {
...
-func GenerateMigUrl(ref GceRef) string {
+func GenerateMigUrl(domainSuffix string, ref GceRef) string {
+	if domainSuffix == "" {
+		domainSuffix = defaultGceDomainSuffix
+	}
+	migUrlTemplate := prefix(domainSuffix) + "%s/zones/%s/instanceGroups/%s"
 	return fmt.Sprintf(migUrlTemplate, ref.Project, ref.Zone, ref.Name)
 }
 
+func prefix(domainSuffix string) string {
```
kawych, line 81: Can we allow to override the whole API endpoint? It allows slightly more flexibility while shortening the code slightly.

Also, I believe "API endpoint" wording is more obvious than "domain suffix" when reading the code :)

cluster-autoscaler/cloudprovider/gce/gce_url.go
```diff
@@ -22,12 +22,9 @@ import (
...
 const (
-	gceUrlSchema    = "https"
-	gceDomainSuffix = "googleapis.com/compute/v1/projects/"
-	// Cluster Autoscaler previously used "content" instead of "www" here, for reasons unknown.
-	gcePrefix           = gceUrlSchema + "://www." + gceDomainSuffix
-	instanceUrlTemplate = gcePrefix + "%s/zones/%s/instances/%s"
-	migUrlTemplate      = gcePrefix + "%s/zones/%s/instanceGroups/%s"
+	gceUrlSchema           = "https"
+	projectsSubstring      = "/projects/"
+	defaultGceDomainSuffix = "googleapis.com/compute/v1/projects/"
```
kawych, line 27: Is it intended that "/projects/" is already a part of domain suffix?

kawych (changes requested):

cluster-autoscaler/cloudprovider/gce/gce_url.go
olagacek: changed it to domain url

cluster-autoscaler/cloudprovider/gce/gce_url.go
```diff
@@ -22,12 +22,9 @@ import (
 )
 
 const (
-	gceUrlSchema    = "https"
-	gceDomainSuffix = "googleapis.com/compute/v1/projects/"
-	// Cluster Autoscaler previously used "content" instead of "www" here, for reasons unknown.
-	gcePrefix           = gceUrlSchema + "://www." + gceDomainSuffix
-	instanceUrlTemplate = gcePrefix + "%s/zones/%s/instances/%s"
-	migUrlTemplate      = gcePrefix + "%s/zones/%s/instanceGroups/%s"
+	gceUrlSchema      = "https"
```
BigDarkClown, line 25: Can we remain this to something like `urlPrefix` or `httpsPrefix`? The current does not accurately describe the content.

cluster-autoscaler/cloudprovider/gce/gce_url.go
```diff
@@ -64,24 +61,32 @@ func ParseInstanceUrlRef(url string) (GceRef, error) {
...
-func GenerateMigUrl(ref GceRef) string {
+func GenerateMigUrl(domainUrl string, ref GceRef) string {
+	if domainUrl == "" {
+		domainUrl = defaultDomainUrl
+	}
+	migUrlTemplate := domainUrl + projectsSubstring + "%s/zones/%s/instanceGroups/%s"
 	return fmt.Sprintf(migUrlTemplate, ref.Project, ref.Zone, ref.Name)
 }
 
 func parseGceUrl(url, expectedResource string) (project string, zone string, name string, err error) {
```
BigDarkClown, line 81: nit: I feel like this whole function could be just a single regexp check. What do you think about rewriting this?

BigDarkClown (commented):
LGTM, but I think we can clean this a bit more without much additional work, just by using regexp.

cluster-autoscaler/cloudprovider/gce/gce_url.go
olagacek: I replaced the checks with regexp, however I still left the split logic. Let me know if you think we should change it further

cluster-autoscaler/cloudprovider/gce/gce_url.go
olagacek: removed it alltogether

cluster-autoscaler/cloudprovider/gce/gce_url.go
BigDarkClown: I think this is good enough for now - ideal would be replacing the split with regexp as well, but this is already much better.

BigDarkClown (commented):
/lgtm
/approve

PR #6294, opened 2023-11-16
refactor(*): move getKubeClient to utils/kubernetes

vadasambar:
#### What type of PR is this?

/kind cleanup

This is a follow-up to the closed PR: https://github.com/kubernetes/autoscaler/pull/6180 (<- check this PR for more info on the current PR)

No issue exists for this. Can create one if needed. 

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/utils/kubernetes/client.go
+++ b/cluster-autoscaler/utils/kubernetes/client.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kubernetes
+
+import (
+	"net/url"
+
+	"k8s.io/autoscaler/cluster-autoscaler/config"
+
+	kube_client "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
+)
+
+const (
+	failedToBuildConfigErr       = "Failed to build config"
+	failedToParseK8sUrlErr       = "Failed to parse Kubernetes url"
+	failedToBuildClientConfigErr = "Failed to build Kubernetes client configuration"
+)
+
+// CreateKubeClient creates kube client based on AutoscalingOptions.KubeClientOptions
+func CreateKubeClient(opts config.KubeClientOptions) kube_client.Interface {
+	return kube_client.NewForConfigOrDie(getKubeConfig(opts))
+}
+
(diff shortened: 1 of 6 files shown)


vadasambar:
This PR is currently WIP. I will assign x13n here once the PR is ready for review.

vadasambar:
@x13n I think the PR is ready for review :pray:

elmiko:
thanks @vadasambar , i should have some time to review this on monday

cluster-autoscaler/utils/kubernetes/client.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+const (
+	failedToBuildConfigErr       = "Failed to build config"
+	failedToParseK8sUrlErr       = "Failed to parse Kubernetes url"
+	failedToBuildClientConfigErr = "Failed to build Kubernetes client configuration"
+)
+
+// CreateKubeClient creates kube client based on AutoscalingOptions.KubeClientOptions
+func CreateKubeClient(opts *config.KubeClientOptions) kube_client.Interface {
```
x13n, line 37: nit: Do we need to pass this via pointer? The struct is small and this won't be called frequently, so performance is not an issue and when passed by value it is clear at the call site that the data under pointer is not getting modified. You also can stop taking the address of `KubeClientOps` then.

x13n (commented):
Thanks for following up on this. Just a single minor comment, feel free to ignore. Putting on hold for you to decide whether to change anything (or wait for @elmiko to take a look as well).

/lgtm
/approve
/hold

cluster-autoscaler/utils/kubernetes/client.go
vadasambar: > this won't be called frequently

Agreed. 

>  when passed by value it is clear at the call site that the data under pointer is not getting modified

I see what you mean. 

Honestly, I don't have a strong opinion here. The argument for pointer is, the number of fields in `KubeClientOptions` may increase in the future ([there are many rest config options](https://github.com/kubernetes/autoscaler/blob/57f2814267399e0900a70d567a7813089a3cfa5b/cluster-autoscaler/vendor/k8s.io/client-go/rest/config.go#L50-L146) which aren't supported through `KubeClientOptions` as of now). Changing this to a pointer later would mean a change across core and cloudproviders (i.e., places which use this function) and getting review from cloudprovider maintainers.

vadasambar: Let's see if @elmiko also has any comment on this as a cloudprovider maintainer.

vadasambar: I will wait until Monday and make the change to pass by value if there are no strong opinions on this.

elmiko: i think what @x13n is proposing sounds reasonable, passing by pointer does make me wonder if some mutation is happening.

elmiko (commented):
lgtm, happy to approve for capi depending on what @vadasambar wants to do regarding the pointer reference.

vadasambar:
> @vadasambar: you cannot LGTM your own PR.

:laughing: thought so. @x13n , @elmiko I have made the change to pass by value. Need lgtm from you folks :pray:

elmiko (commented):
/lgtm

x13n:
Thanks! Un-holding as well so it actually gets merged.

/hold cancel

elmiko:
oops, my bad, i thought there was an open question on the hold.

PR #6296, opened 2023-11-19
fix: alicloud the function NodeGroupForNode return nil

guopeng0:
#### What type of PR is this?
/kind bug

I noticed an issue in the implementation of the NodeGroupForNode function in the alicloud provider of the cluster-autoscaler project. This issue results in the function sometimes returning nil for nodes in a specific nodeGroup.

Currently, the implementation regenerates the cache data and calls Alibaba Cloud to query all instances when a specified instance is not found. However, the DescribeScalingInstances function has default values of pageSize=10 and pageNumber=1, which only returns a subset of the data.

To address this issue, I have made the following modifications:
Modify the code logic to perform multiple queries based on the TotalCount of response information, ensuring the complete set of instances in the nodeGroup is returned.
I am a frequent user of cluster-autoscaler and have thoroughly tested and validated these modifications on Alibaba Cloud. 

The error in this function can lead to inaccurate nodeGroup healthy status, resulting in scaling issues.

Fixes #

NONE

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/alicloud/alicloud_auto_scaling.go
+++ b/cluster-autoscaler/cloudprovider/alicloud/alicloud_auto_scaling.go
@@ -28,6 +28,7 @@ const (
 	refreshClientInterval   = 60 * time.Minute
 	acsAutogenIncreaseRules = "acs-autogen-increase-rules"
 	defaultAdjustmentType   = "TotalCapacity"
+	defaultRequestPageSize  = 10
 )
 
 // autoScaling define the interface usage in alibaba-cloud-sdk-go.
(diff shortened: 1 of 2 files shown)


guopeng0:
@x13n @BigDarkClown

vadasambar:
Looks like I can't add the `area/provider/alicloud` label. @ringtail you might want to take a look at this.

vadasambar:
Just in case, I will do a first pass tomorrow. Thank you for the PR @guopeng0! The description you wrote is easy to understand. 

/assign vadasambar

cluster-autoscaler/cloudprovider/alicloud/alicloud_auto_scaling.go
```diff
@@ -166,14 +167,29 @@ func (m autoScalingWrapper) getScalingGroupByName(groupName string) (*ess.Scalin
...
-		return nil, err
+	instances := make([]ess.ScalingInstance, 0)
+	pageNumber := 1
+
+	for {
+		params := ess.CreateDescribeScalingInstancesRequest()
+		params.ScalingGroupId = asgId
+		params.PageNumber = requests.NewInteger(pageNumber)
+		params.PageSize = requests.NewInteger(defaultRequestPageSize)
+		resp, err := m.DescribeScalingInstances(params)
```
vadasambar, line 178: Should we add some unit tests around this? Since this is an unexported function, another option could be to test it by adding a test for `FindForInstance`. For starters, we can have a test related to the bug which fails for the old code.

vadasambar (changes requested):

cluster-autoscaler/cloudprovider/alicloud/alicloud_auto_scaling.go
guopeng0: Thank you for your suggestion, I fully agree and will add unit tests as soon as possible.

guopeng0: The unit tests have been added.

vadasambar:
/lgtm 
/approve

@guopeng0 can you please squash the commits into a single commit? 

@x13n , @BigDarkClown can we get an approve on this PR :pray:

vadasambar (approved):

x13n:
/approve

(Based on @ringtail's lgtm above.)

PR #6302, opened 2023-11-20
Fixes scale down logic.

piotrwrotniak:
#### What type of PR is this?

/kind bug
/kind regression

Fixes scale down logic removing redundant check.


piotrwrotniak:
Verified this fixes the bug.

Issue #6293, opened 2023-11-16
Cluster Autoscaler patch releases Nov 2023


x13n:
I'm planning to cut new Cluster Autoscaler patch releases for 1.26-28 next week. If there are any bugfix cherry-picks you want included, please prepare the PRs and let me know in the comments.

Tagging cloud provider owners for visibility:
@4ND3R50N
@7fELF
@alphajc
@andrewsykim
@Andrius521
@apricote
@ArturasRa
@arunmk
@avorima
@BigDarkClown
@cprivitere
@ctrox
@davidjumani
@dbonfigli
@deitch
@detiber
@displague
@d-mo
@dougsong
@drmorr0
@ellistarn
@elmiko
@enxebre
@ericrrath
@feiskyer
@gajicdev
@gandhipr
@gjtempleton
@happytreees
@hardikdr
@hello2mao
@jackfrancis
@jayantjain93
@jaypipes
@Jeffwan
@jlamillan
@johnl
@jtherin
@kevin-wangzefeng
@LKaemmerling
@louisportay
@maciekpytel
@marwanad
@MorrisLaw
@mrajashree
@NeilW
@nilo19
@optik-aper
@OriHoch
@pawelkuc

x13n:
@PhilippeChepy
@piepmatz
@pierre-emmanuelJ
@RainbowMango
@remyleone
@ringtail
@schegi
@Sh4d1
@shysank
@sohan-oracle
@tallaxes
@tghartland
@thirdeyenick
@towca
@trungng92
@vishalanarase
@v-pap
@zalmarge

RainbowMango:
LGTM thank you for the notice

ricardov1:
@x13n these have been merged but just want to make sure they're included in the releases:
1.26: https://github.com/kubernetes/autoscaler/pull/6287
1.27: https://github.com/kubernetes/autoscaler/pull/6285
1.28: https://github.com/kubernetes/autoscaler/pull/6283

x13n:
@ricardov1 @ctrox Thanks, all the PRs you listed will be included.

x13n:
Unfortunately, 1.25 reached end of life a month ago, I'm only patching the k/k versions that are still supported: https://kubernetes.io/releases/

ricardov1:
[...]

Issue #6259, opened 2023-11-08
Go vulnerabilities in image autoscaling/cluster-autoscaler-amd64:v1.28.0, 1.27.2, 1.26.2.


zhouke1991:
**Which component are you using?**:
cluster-autoscaler

**What version of the component are you using?**:
1.26.2, 1.27.2, 1.28.0

**How to reproduce it (as minimally and precisely as possible)**:

The security vulnerability [CVE-2023-3955](https://nvd.nist.gov/vuln/detail/CVE-2023-3955) and [CVE-2023-3676](https://nvd.nist.gov/vuln/detail/CVE-2023-3676)  that is pointed out by Trivy scan for cluster-autoscaler 1.28.0
```
─────────────────┼──────────────────────────────────────────────────────────────┤
│ k8s.io/kubernetes                                            │ CVE-2023-3676       │ HIGH     │ v1.28.0           │ 1.28.1, 1.27.5, 1.26.8, 1.25.13, 1.24.17 │ Insufficient input sanitization on Windows nodes leads to    │
│                                                              │                     │          │                   │                                          │ privilege escalation                                         │
│                                                              │                     │          │                   │                                          │ https://avd.aquasec.com/nvd/cve-2023-3676                    │
│                                                              ├─────────────────────┤          │                   │                                          ├──────────────────────────────────────────────────────────────┤
│                                                              │ CVE-2023-3955       │          │                   │                                          │ Insufficient input sanitization on Windows nodes leads to    │
│                                                              │                     │          │                   │                                          │ privilege escalation                                         │
│                                                              │                     │          │                   │                                          │ https://avd.aquasec.com/nvd/cve-2023-3955                    │
└──────────────────────────────────────────────────────────────┴─────────────────────┴──────────┴───────────────────┴──────────────────────────────────────────┴──────────────────────────────────────────────────────────────┘
```

**Anything else we need to know?**:

We need to bump the [k8s.io/kubernetes](http://k8s.io/kubernetes) package to a fixed version. Fixed versions include 1.28.1, 1.27.5, 1.26.8. 

Can we fix these CVEs and release new versions for the above cluster-autoscaler image?

PR #6311, opened 2023-11-22
Update kubernetes vendor to 1.27.8

x13n:


x13n:
xref https://github.com/kubernetes/autoscaler/issues/6259 https://github.com/kubernetes/autoscaler/issues/6293

cluster-autoscaler/version/version.go
```diff
@@ -17,4 +17,4 @@ limitations under the License.
 package version
 
 // ClusterAutoscalerVersion contains version of CA.
-const ClusterAutoscalerVersion = "1.27.3"
+const ClusterAutoscalerVersion = "1.27.8"
```
BigDarkClown, line 20: This is a CA version not a K8s version. We should bump the patch number once to `1.27.4`

x13n: This was autogenerated. Overwritten now.

BigDarkClown (commented):
/lgtm

PR #6315, opened 2023-11-24
ClusterAutoscaler: Put APIs in a separate go module

### LINKED ISSUE #6307: CA: Put APIs in a separate go module

ClusterAutoscaler

If the external project imports cluster-autoscaler, the external project can't run `go list -modfile=./go.mod -m all`
because the cluster-autoscaler depends on k/k repository.

By these limitations, Goland (JetBrains) wouldn't work fine.
Please see the details: https://github.com/kubernetes-sigs/kueue/issues/1345

As we discussed in the above issue with @x13n and @alculquicondor, the best solution is to put APIs in a separate go module.

**Additional context.**:

tenzen-y:
#### What type of PR is this?

/kind cleanup

I put APIs in a separate go module.

Fixes #6307

```release-note

```

```docs

```


tenzen-y:
Also, it might be worth it to create a separate go module in https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/provisioningrequest/client as well.

@x13n WDYT?

alculquicondor:
I don't think it's worth maintaining 3 packages. Maybe the client and APIs can be in the same package?

cluster-autoscaler/provisioningrequest/apis/go.mod
```diff
@@ -0,0 +1,47 @@
+module sigs.k8s.io/autoscaler/cluster-autoscaler/provisioningrequest/apis
```
alculquicondor, line 1: I think the module should be
`sigs.k8s.io/autoscaler/cluster-autoscaler/apis`

where one of the folders is `provisioningrequest`.

alculquicondor: The reason being that we might have an generic (independent of the cloud provider) implementation of `provisioningrequest` and it might have dependencies that are not worth leaking to users of the API.

tenzen-y:
> I don't think it's worth maintaining 3 packages. Maybe the client and APIs can be in the same package?

It makes sense.

cluster-autoscaler/provisioningrequest/apis/go.mod
tenzen-y: I see. Maybe, we can put a client folder on `sigs.k8s.io/autoscaler/cluster-autoscaler/apis/provisioningrequest/client` as well.

x13n: Doesn't `sigs.k8s.io` resolve to kubernetes-sigs org? This repo is in main kubernetes org, so shouldn't it be under `k8s.io/autoscaler/cluster-autoscaler`?

tenzen-y: Oops. Sorry, you're right.

cluster-autoscaler/boilerplate/boilerplate.generatego.txt
tenzen-y, line 1: I added this file since there are no templates.

tenzen-y:
Through some investigation, I found that `k8s.io/autoscaler/cluster-autoscaler/apis` pkg can not have `vendor` separate from `k8s.io/autoscaler/cluster-autoscaler`'s one due to the following error:

@x13n Do we need to have `vendor` for the `k8s.io/autoscaler/cluster-autoscaler/apis`?

> # k8s.io/autoscaler/cluster-autoscaler/provisioningrequest/provreqwrapper
> Error: provisioningrequest/provreqwrapper/wrapper.go:64:9: cannot use pr.v1Beta1PR.CreationTimestamp (variable of type "k8s.io/autoscaler/cluster-autoscaler/apis/vendor/k8s.io/apimachinery/pkg/apis/meta/v1".Time) as "k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/apimachinery/pkg/apis/meta/v1".Time value in return statement
> Error: provisioningrequest/provreqwrapper/wrapper.go:69:9: cannot use pr.v1Beta1PR (variable of type *v1beta1.ProvisioningRequest) as "k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/apimachinery/pkg/runtime".Object value in return statement: *v1beta1.ProvisioningRequest does not implement "k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/apimachinery/pkg/runtime".Object (wrong type for method DeepCopyObject)
> 		have DeepCopyObject() "k8s.io/autoscaler/cluster-autoscaler/apis/vendor/k8s.io/apimachinery/pkg/runtime".Object
> 		want DeepCopyObject() "k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/apimachinery/pkg/runtime".Object
> Error: provisioningrequest/provreqwrapper/wrapper.go:85:9: cannot use pr.v1Beta1PR.UID (variable of type "k8s.io/autoscaler/cluster-autoscaler/apis/vendor/k8s.io/apimachinery/pkg/types".UID) as "k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/apimachinery/pkg/types".UID value in return statement
> Error: provisioningrequest/provreqwrapper/wrapper.go:90:9: cannot use pr.v1Beta1PR.Status.Conditions (variable of type []"k8s.io/autoscaler/cluster-autoscaler/apis/vendor/k8s.io/apimachinery/pkg/apis/meta/v1".Condition) as []"k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/apimachinery/pkg/apis/meta/v1".Condition value in return statement
> Error: provisioningrequest/provreqwrapper/wrapper.go:95:35: cannot use conditions (variable of type []"k8s.io/autoscaler/cluster-autoscaler/vendor/k8s.io/apimachinery/pkg/apis/meta/v1".Condition) as []"k8s.io/autoscaler/cluster-autoscaler/apis/vendor/k8s.io/apimachinery/pkg/apis/meta/v1".Condition value in assignment

https://github.com/kubernetes/autoscaler/actions/runs/7267376813/job/19801050209?pr=6315

Once removing vendor for the `k8s.io/autoscaler/cluster-autoscaler/apis` pkg, CI should work fine: https://github.com/kubernetes/autoscaler/actions/runs/7267869755/job/19802584596?pr=6315

tenzen-y:
@x13n friendly ping :)

tenzen-y:
I resolved conflicts.

x13n:
Apologies for slow response.

/lgtm
/approve

@kisieland FYI

tenzen-y:
/assign @mwielgus 
for `hack/for-go-proj.sh`

cluster-autoscaler/apis/go.mod
```diff
@@ -0,0 +1,64 @@
...
+	k8s.io/utils v0.0.0-20230726121419-3b25d923346b // indirect
+	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
+	sigs.k8s.io/yaml v1.3.0 // indirect
+)
+
+replace k8s.io/apimachinery => k8s.io/apimachinery v0.29.0-alpha.3
+
+replace k8s.io/client-go => k8s.io/client-go v0.29.0-alpha.3
+
+replace k8s.io/code-generator => k8s.io/code-generator v0.29.0-alpha.3
```
alculquicondor, line 64: Do we need these replace?

tenzen-y: I followed the go module in the following:

https://github.com/kubernetes/autoscaler/blob/d29ffd03b961b0cdd7b2f547035db9548c4ba022/cluster-autoscaler/go.mod#L211-L216

If this replacement is not needed, we can remove those.
@x13n WDYT?

alculquicondor: I think these `replace` could cause anyone using this package to have issues.

Also, maybe we can use released v0.29 versions instead.

tenzen-y: > I think these replace could cause anyone using this package to also have to replace.

That makes sense.

> Also, maybe we can use released v0.29 versions instead.

Indeed, the apis package could use the released version since that package doesn't depend on any k/k codes.

tenzen-y:
I squashed old commits into one, and then I replaced alpha k8s libraries with the released ones.

@x13n @alculquicondor Please take another look.

alculquicondor:
go.mod LGTM, in the sense that it will be usable from Kueue. I'll left the rest to the owners :)

x13n:
Hm... This makes me think about a couple of things:
- Versioning should be consistent between main CA module and the new API module. Specifically, the k8s dependency should use the same tag.
- There's no binary that will be released for the new module, but we should keep the deps in sync going forwards. Do we need any changes to https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/hack/update-vendor.sh for this?
- Since 1.29 was already released, I think we should use `v1.30.0-alpha.0` on the master branch.

tenzen-y:
> There's no binary that will be released for the new module, but we should keep the deps in sync going forwards. Do we need any changes to https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/hack/update-vendor.sh for this?

Yes, we need an additional script or modified `hack/update-vendor.sh`. I think another script would be better since the `cluster-autoscaler/hack/apis` doesn't have vendor.

x13n:
> > Versioning should be consistent between main CA module and the new API module. Specifically, the k8s dependency should use the same tag.
> > Since 1.29 was already released, I think we should use v1.30.0-alpha.0 on the master branch.
> 
> These two points are in conflict. Users of the CA APIs shouldn't have to depend on alpha k8s modules.
> 
> I think using 1.29.x in the APIs and 1.30.0-alpha.y would work in most scenarios. CA code would compile with 1.30, as that satisfies both dependencies.

That makes sense, but will be inconsistent with tags unless we start releasing the API module independently from CA? Unless we follow something like this:
- master branch - API module uses latest k/k release deps, CA uses latest k/k alpha
- release branch 1.XX - both API module and CA use latest k/k patch: 1.XX.YY

Does that make sense to you? This should be documented in our release docs. CC @gjtempleton @MaciekPytel for thoughts.

> Yes, we need an additional script or modified `hack/update-vendor.sh`. I think another script would be better since the `cluster-autoscaler/hack/apis` doesn't have vendor.

Honestly I'm not sure if we really need CA to have vendor either, we could rename the script to `update-deps.sh` so it doesn't imply vendoring.

tenzen-y:
> That makes sense, but will be inconsistent with tags unless we start releasing the API module independently from CA? Unless we follow something like this:
> 
> master branch - API module uses latest k/k release deps, CA uses latest k/k alpha
> release branch 1.XX - both API module and CA use latest k/k patch: 1.XX.YY
> Does that make sense to you? This should be documented in our release docs. CC @gjtempleton @MaciekPytel for thoughts.

@x13n I imagined that this (autoscaler) repository would cut releases separately from `cluster-autoscaler-1.x.y` like `cluster-autoscaler/apis`. If this repo wouldn't cut dedicated releases, we can not import `cluster-autoscaler/apis` to external projects using the tagging version due to Golang module restriction.

cc: @alculquicondor

tenzen-y:
> > Yes, we need an additional script or modified hack/update-vendor.sh. I think another script would be better since the cluster-autoscaler/hack/apis doesn't have vendor.
> 
> Honestly I'm not sure if we really need CA to have vendor either, we could rename the script to update-deps.sh so it doesn't imply vendoring.

I'm ok with either way. I'll try to modify the existing script.

x13n:
If we need separate process for cutting releases of the new module, we need to document it as well. I was thinking it could be piggybacking on the existing CA release tags though. Why is this an issue? We have 1:1 mappings between CA minor, k8s minor and go version used - does that render the api module unusable? I'd think if it depends on a specific k/k version, it is ok to rely on Go version used by the k/k as well.

elmiko:
i haven't studied this PR deeply yet, but i agree with the thinking that it would be nice if these API pieces could be included from other projects that might be implementing common autoscaling components. thanks for the work here!

x13n:
Yes, that makes sense to me, thanks. While reading about this a bit more, I came across https://go.dev/wiki/Modules#faqs--multi-module-repositories (see `Is it possible to add a module to a multi-module repository?` section). I think this is exactly the case we have here? Won't this PR lead to “ambiguous import” error when trying to use the new apis module?

x13n:
The more I read about this, the more I wonder if it wouldn't be a better idea to put this module in a separate repo under kubernetes-sigs.

alculquicondor:
That has the disadvantage that PRs that need API updates will have to be split among 2 repositories. The best model is what we have for k/k with the staging folder, but that sounds like overkill for just one submodule.

We might be better off doing a submodule in the repo. It's just 1 or 2 extra steps when branching a new version.

x13n:
Yeah, I agree trying to follow in k/k footsteps here would be an overkill. Submodule is ok, just need to follow these extra steps. I see right now CA repo depends on `k8s.io/autoscaler/cluster-autoscaler/apis v0.0.0-00010101000000-000000000000`. What are the next steps after this PR is merged?

tenzen-y:
> Yeah, I agree trying to follow in k/k footsteps here would be an overkill. Submodule is ok, just need to follow these extra steps. I see right now CA repo depends on `k8s.io/autoscaler/cluster-autoscaler/apis v0.0.0-00010101000000-000000000000`. What are the next steps after this PR is merged?

@x13n We need to work on https://github.com/kubernetes/autoscaler/pull/6315#issuecomment-1884726098 in this PR.
Please give me some time.

I think that we don't have any steps after this PR is merged. Any concerns?

tenzen-y:
I have rebased. I'm still working on scripts.

tenzen-y:
@x13n I modified the `update-deps.sh`. PTAL, thanks.

tenzen-y:
@x13n friendly ping :)

cluster-autoscaler/hack/update-deps.sh
```diff
@@ -48,22 +48,32 @@ MODS=($(
...
-            exit 1
-        fi
-        echo "${JSON}" | sed -n 's|.*"Version": "\(.*\)".*|\1|p'
-    )
-    go mod edit "-replace=${MOD}=${MOD}@${V}"
+PKGS=("." "./apis")
+for pkg in "${PKGS[@]}"; do
+  pushd "${pkg}"
+
+  for MOD in "${MODS[@]}"; do
```
x13n, line 55: I think `MODS` depend on `pkg`, shouldn't `MOD` definition be moved inside the inner loop here? Also, perhaps move to a function to avoid too much nesting.

x13n (commented):
Thanks, just one minor comment, otherwise should be good to go.

Approving the PR, anyone can then lgtm & unhold later in case I won't (I'm about to go on vacation for a week).

/approve
/hold

cluster-autoscaler/hack/update-deps.sh
tenzen-y: > I think MODS depend on pkg

@x13n I think it's not correct since `MODS` is set based on the `staging` directory in the k/k repository.
So, we can not move the `MOD` definition inside the inner loop.

tenzen-y:
I have rebased this PR to resolve conflicts.

tenzen-y:
Hi @mwielgus! Could you take a look this PR since we need to get approve from you due to changing of hack/for-go-proj.sh.
Thanks.

cluster-autoscaler/hack/update-deps.sh
x13n: `MODS` is a result of catting `go.mod`, which is different for each `PKG`. For instance, `./go.mod` uses `k8s.io/component-base`, but `./apis/go.mod` doesn't.

tenzen-y: Oh, it makes sense.
I'll try to fix it!

tenzen-y: @x13n Indeed, the `MODS` is the result of catting `go.mod`, but we cat the `go.mod` after we perform "pushed ${REPO}". It means that we cat the go.mod in k/k.

So, I think that the current script is correct. WDYT?

```
pushd ${REPO}
git fetch --depth 1 origin v${VERSION}
git checkout FETCH_HEAD
MODS=($(
    cat go.mod | sed -n 's|.*k8s.io/\(.*\) => ./staging/src/k8s.io/.*|k8s.io/\1|p'
))
```

cluster-autoscaler/hack/update-deps.sh
tenzen-y, line 1: @x13n As you proposed in another thread, I cleaned up this script. 
We can verify this script with `./hack/update-deps.sh ${K/K_ALPHA_VERSION} ${STABLE_APIS_VERSION}` like `./hack/update-deps.sh v1.30.0-alpha.3 v1.29.2`.

cluster-autoscaler/hack/update-deps.sh
tenzen-y: Anyway, I cleaned up this script: https://github.com/kubernetes/autoscaler/pull/6315#discussion_r1518672607

x13n:
Thanks! LGTM from my perspective.

/lgtm
/approve

tenzen-y:
@mwielgus @maciekpytel @gjtempleton Could you get a chance to review this PR?

mwielgus (approved):
/lgtm
/approve

Issue #6318, opened 2023-11-27
Add backoff data to cluster-autoscaler-status and make the status easier for parsing


walidghallab:
**Which component are you using?**:
cluster-autoscaler

**The problem this feature should solve**:
cluster-autoscaler-status config map shows which nodegroups are in backoff but it doesn't show the reasons behind backoff of nodegroups which makes debugging harder especially for large clusters.
Also the status of cluster-autoscaler doesn't confine to a pattern that is easily parsable.

**Describe the solution you'd like**:
To add backoff reason (errorCode & errorMessage) for each nodegroup being in backoff in the status field of cluster-autoscaler-status config map.

Also change the current format of status field from the current format to yaml which is both human-readable and easier in parsing.

**Example for the status field:**

Current state (Human readable string without backoff data):

```
status: |
Cluster-autoscaler status at 2023-11-24 04:28:19.546750398 +0000 UTC:
Cluster-wide:
  Health:      Healthy (ready=7 unready=0 (resourceUnready=0) notStarted=0 longNotStarted=0 registered=7 longUnregistered=0)
               LastProbeTime:      2023-11-24 04:28:19.408724799 +0000 UTC m=+48988.192673876
               LastTransitionTime: 2023-11-23 14:52:02.009391854 +0000 UTC m=+10.793340938
  ScaleUp:     NoActivity (ready=7 registered=7)
               LastProbeTime:      2023-11-24 04:28:19.408724799 +0000 UTC m=+48988.192673876
               LastTransitionTime: 2023-11-24 04:15:34.972108466 +0000 UTC m=+48223.756057543
  ScaleDown:   NoCandidates (candidates=0)
               LastProbeTime:      2023-11-24 04:28:19.408724799 +0000 UTC m=+48988.192673876
               LastTransitionTime: 0001-01-01 00:00:00 +0000 UTC

NodeGroups:
  Name:        https://www.googleapis.com/compute/v1/projects/sample-project/zones/us-central1-c/instanceGroups/gke-sample-cluster-default-pool-40ce0341-grp
  Health:      Healthy (ready=7 unready=0 (resourceUnready=0) notStarted=0 longNotStarted=0 registered=7 longUnregistered=0 cloudProviderTarget=7 (minSize=4, maxSize=10))
               LastProbeTime:      2023-11-24 04:28:19.408724799 +0000 UTC m=+48988.192673876
[...]

PR #6327, opened 2023-11-30
Add error details to autoscaling backoff.

walidghallab:
#### What type of PR is this?

/kind feature

This PR adds more detailed error information to backoff interface & implementation. 
This is needed as it will allow getting this info easily to show them as a reason for backoff.

#6318 

NONE

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -143,6 +143,13 @@ type ClusterStateRegistry struct {
 	scaleUpFailures map[string][]ScaleUpFailure
 }
 
+// NodeGroupScalingSafety contains information about the safety of the node group to scale up/down.
+type NodeGroupScalingSafety struct {
+	SafeToScale   bool
+	Healthy       bool
+	BackoffStatus backoff.Status
+}
+
 // NewClusterStateRegistry creates new ClusterStateRegistry.
 func NewClusterStateRegistry(cloudProvider cloudprovider.CloudProvider, config ClusterStateRegistryConfig, logRecorder *utils.LogEventRecorder, backoff backoff.Backoff, nodeGroupConfigProcessor nodegroupconfig.NodeGroupConfigProcessor) *ClusterStateRegistry {
 	emptyStatus := &api.ClusterAutoscalerStatus{
@@ -276,7 +283,11 @@ func (csr *ClusterStateRegistry) updateScaleRequests(currentTime time.Time) {
 			} else {
 				gpuResource, gpuType = gpu.GetGpuInfoForMetrics(csr.cloudProvider.GetNodeGpuConfig(nodeInfo.Node()), availableGPUTypes, nodeInfo.Node(), scaleUpRequest.NodeGroup)
 			}
-			csr.registerFailedScaleUpNoLock(scaleUpRequest.NodeGroup, metrics.Timeout, cloudprovider.OtherErrorClass, "timeout", gpuResource, gpuType, currentTime)
+			csr.registerFailedScaleUpNoLock(scaleUpRequest.NodeGroup, metrics.Timeout, cloudprovider.InstanceErrorInfo{
+				ErrorClass:   cloudprovider.OtherErrorClass,
+				ErrorCode:    "timeout",
+				ErrorMessage: fmt.Sprintf("Scale-up timed out for node group %v after %v", nodeGroupName, currentTime.Sub(scaleUpRequest.Time)),
+			}, gpuResource, gpuType, currentTime)
 			delete(csr.scaleUpRequests, nodeGroupName)
 		}
 	}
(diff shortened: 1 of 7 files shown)


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -441,12 +482,14 @@ func (csr *ClusterStateRegistry) updateNodeGroupMetrics() {
...
 
-// IsNodeGroupSafeToScaleUp returns true if node group can be scaled up now.
-func (csr *ClusterStateRegistry) IsNodeGroupSafeToScaleUp(nodeGroup cloudprovider.NodeGroup, now time.Time) bool {
-	if !csr.IsNodeGroupHealthy(nodeGroup.Id()) {
-		return false
+// IsNodeGroupSafeToScaleUp returns information about node group safety to be scaled up now.
+func (csr *ClusterStateRegistry) IsNodeGroupSafeToScaleUp(nodeGroup cloudprovider.NodeGroup, now time.Time) NodeGroupScalingSafety {
+	nodeGroupStatus := csr.GetNodeGroupHealth(nodeGroup.Id())
+	if !nodeGroupStatus.IsHealthy {
+		return NodeGroupScalingSafety{false, nodeGroupStatus.ErrorInfo}
```
towca, line 489: nit: Using named struct parameters is preferred (no need to modify struct literals whenever a new field is added).

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -756,7 +799,7 @@ func (csr *ClusterStateRegistry) GetStatus(now time.Time) *api.ClusterAutoscaler
 
 		// Health.
 		nodeGroupStatus.Conditions = append(nodeGroupStatus.Conditions, buildHealthStatusNodeGroup(
-			csr.IsNodeGroupHealthy(nodeGroup.Id()), readiness, acceptable, nodeGroup.MinSize(), nodeGroup.MaxSize()))
+			csr.GetNodeGroupHealth(nodeGroup.Id()).IsHealthy, readiness, acceptable, nodeGroup.MinSize(), nodeGroup.MaxSize()))
```
towca, line 802: nit: Why not pass the whole Health object here like we do for Safety and `buildScaleUpStatusNodeGroup`?

cluster-autoscaler/utils/backoff/exponential_backoff.go
```diff
@@ -87,14 +88,21 @@ func (b *exponentialBackoff) Backoff(nodeGroup cloudprovider.NodeGroup, nodeInfo
...
 	return backoffUntil
 }
 
-// IsBackedOff returns true if execution is backed off for the given node group.
-func (b *exponentialBackoff) IsBackedOff(nodeGroup cloudprovider.NodeGroup, nodeInfo *schedulerframework.NodeInfo, currentTime time.Time) bool {
+// BackoffStatus returns whether the execution is backed off for the given node group and error info when the node group is backed off.
+func (b *exponentialBackoff) BackoffStatus(nodeGroup cloudprovider.NodeGroup, nodeInfo *schedulerframework.NodeInfo, currentTime time.Time) Status {
 	backoffInfo, found := b.backoffInfo[b.nodeGroupKey(nodeGroup)]
-	return found && backoffInfo.backoffUntil.After(currentTime)
+	if !found || !backoffInfo.backoffUntil.After(currentTime) {
```
towca, line 99: nit: `backoffInfo.backoffUntil.Before(currentTime)` seems more readable.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -143,6 +143,18 @@ type ClusterStateRegistry struct {
 	scaleUpFailures map[string][]ScaleUpFailure
 }
 
+// NodeGroupHealth contains information about the health of the node group.
+type NodeGroupHealth struct {
```
towca, line 147: Hmm, I still don't understand why we need these two types at all. Either I'm really not seeing something here, or the `ErrorInfo` parameter seems completely unused other than passing it through from `GetNodeGroupHealth` to `IsNodeGroupSafeToScaleUp` (after which it's not used anymore).

Wouldn't the PR work exactly the same if we didn't change `IsNodeGroupSafeToScaleUp` and `IsNodeGroupHealthy` at all, just with less complexity? From a high level, the "safety" determination is based on backoff info, rather than being an input to backoff - so we shouldn't need to keep this error around to feed it to the new backoff interface, right?

towca (commented):
Everything apart from the 2 structs I mention in one comment LGTM, just some minor comments. Still need to understand the structs

cluster-autoscaler/clusterstate/clusterstate.go
walidghallab: Sorry about that. It was needed but the needed part didn't get in the this commit after splitting.
I've added a line where it is used.

cluster-autoscaler/clusterstate/clusterstate.go
walidghallab: Done, thank you!

cluster-autoscaler/utils/backoff/exponential_backoff.go
walidghallab: The reason I didn't do it was to make it the same as before, as using Before will be different in the case backoffUntil == currentTIme.
But I have changed it to the suggested code since the chances to happen is almost zero (except for the tests which I have made some changes), and also it is similar to the code in Backoff()

cluster-autoscaler/clusterstate/clusterstate.go
walidghallab: It is not currently needed, but I've added it in case we need to add more error info to Health status. Thank you!

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -441,12 +455,16 @@ func (csr *ClusterStateRegistry) updateNodeGroupMetrics() {
...
-// IsNodeGroupSafeToScaleUp returns true if node group can be scaled up now.
-func (csr *ClusterStateRegistry) IsNodeGroupSafeToScaleUp(nodeGroup cloudprovider.NodeGroup, now time.Time) bool {
+// IsNodeGroupSafeToScaleUp returns information about node group safety to be scaled up now.
+func (csr *ClusterStateRegistry) IsNodeGroupSafeToScaleUp(nodeGroup cloudprovider.NodeGroup, now time.Time) NodeGroupScalingSafety {
 	if !csr.IsNodeGroupHealthy(nodeGroup.Id()) {
-		return false
+		return NodeGroupScalingSafety{SafeToScale: false, Healthy: false, BackoffStatus: nil}
 	}
-	return !csr.backoff.IsBackedOff(nodeGroup, csr.nodeInfosForGroups[nodeGroup.Id()], now)
+	backoffStatus := csr.backoff.BackoffStatus(nodeGroup, csr.nodeInfosForGroups[nodeGroup.Id()], now)
```
towca, line 463: Hmm, wouldn't it be simpler to have BackoffStatus as a struct and not a pointer, and just always set it?  I'm slightly worried about somebody assuming it's always set and hitting a panic.

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -821,8 +839,11 @@ func buildScaleUpStatusNodeGroup(isScaleUpInProgress bool, isSafeToScaleUp bool,
 	}
 	if isScaleUpInProgress {
 		condition.Status = api.ClusterAutoscalerInProgress
-	} else if !isSafeToScaleUp {
+	} else if !scaleUpSafety.Healthy {
+		condition.Status = api.ClusterAutoscalerUnhealthy
+	} else if !scaleUpSafety.SafeToScale {
 		condition.Status = api.ClusterAutoscalerBackoff
+		condition.Message = scaleUpSafety.BackoffStatus.ErrorInfo.ErrorMessage
```
towca, line 846: I wouldn't do this in this PR yet, can we hold off on actual status changes for the follow-up PR and keep this as a pure refactor? This would change the current status format, but AFAIU not in the intended way we're aiming for in #6318

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -569,9 +569,11 @@ func (o *ScaleUpOrchestrator) UpcomingNodes(nodeInfos map[string]*schedulerframe
...
 func (o *ScaleUpOrchestrator) IsNodeGroupReadyToScaleUp(nodeGroup cloudprovider.NodeGroup, now time.Time) *SkippedReasons {
 	// Non-existing node groups are created later so skip check for them.
-	if nodeGroup.Exist() && !o.clusterStateRegistry.IsNodeGroupSafeToScaleUp(nodeGroup, now) {
-		// Hack that depends on internals of IsNodeGroupSafeToScaleUp.
-		if !o.clusterStateRegistry.IsNodeGroupHealthy(nodeGroup.Id()) {
+	if !nodeGroup.Exist() {
+		return nil
+	}
+	if scaleUpSafety := o.clusterStateRegistry.IsNodeGroupSafeToScaleUp(nodeGroup, now); !scaleUpSafety.SafeToScale {
+		if !scaleUpSafety.SafeToScale {
```
towca, line 576: This should be `if !scaleUpSafety.Healthy`, right? Also, maybe we can add more details to the backoff warning below now that we have them in `scaleUpSafety`?

towca:
Thanks for addressing the comments!
/lgtm
/approve

PR #6328, opened 2023-11-30
Validate APIVersion when checking for DaemonSet pods

### LINKED ISSUE #5977: Autoscaler not scaling down pods when`kind` is of the resource is same between controllers
- **Which component are you using?**: `cluster-autoscaler`
- **What version of the component are you using?**: `1.25.0`
- **What k8s version are you using?**: `v1.25.10`
- **What environment is this in?**: `kops` on AWS

We are using OpenKruise and [Advanced DaemonSet](https://openkruise.io/docs/user-manuals/advanceddaemonset/).
Autoscaler seems like detects it as regular `DaemonSet` and trying to find the corresponding `DaemonSet` for the workload, and fails. This prevents from scaling down to occur.

I'm not sure whether it's the root cause or not, but I suspect that Autoscaler doesn't respect the API group. The pods created by Advanced DaemonSet have the following `ownerReferenses`:

```yaml
  ownerReferences:
  - apiVersion: apps.kruise.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: DaemonSet
    name: example-advanced-daemonset
    uid: 6f0f4fa8-2694-44df-9e68-8f00630f19c1
```

The `kind` is `DaemonSet`, but the `apiVersion` is `apps.kruise.io/v1alpha1`. So might be that the `apiVersion` is being ignored, and autoscaler just looks for regular `DaemonSet` named `example-advanced-daemonset`, which obviously doesn't exist.

Here are relevant logs from autoscaler:

```
aws-cluster-autoscaler I0724 08:09:58.813598       1 cluster.go:170] node i-0bb670250f677c0c0 cannot be removed: daemonset for devops/example-advanced-daemonset-x44c5 is not present, err: daemonset.apps "example-advanced-daemonset" not found
```

Can someone advice if this expected that the `apiVersion` is ignored or not being respected, am I missing something?

shamil:
#### What type of PR is this?

/kind bug

The issue is described in #5977

Fixes #5977

In the change I added a validation for `APIVersion`, and adjusted all relevant tests to support that. If `APIVersion` shouldn't be required when validating the DaemonSet's owner referense, I can ensure that the `APIVersion` validation happens only if it's not empty. Let me know...

NONE

### DIFF
--- a/cluster-autoscaler/core/scaledown/actuation/drain_test.go
+++ b/cluster-autoscaler/core/scaledown/actuation/drain_test.go
@@ -124,7 +124,7 @@ func TestDaemonSetEvictionForEmptyNodes(t *testing.T) {
 			for i, dsName := range scenario.dsPods {
 				ds := BuildTestPod(dsName, 100, 0)
 				ds.Spec.NodeName = "n1"
-				ds.OwnerReferences = GenerateOwnerReferences("", "DaemonSet", "", "")
+				ds.OwnerReferences = GenerateOwnerReferences("", "DaemonSet", "apps/v1", "")
 				if v, ok := scenario.extraAnnotationValue[dsName]; ok {
 					ds.Annotations[daemonset.EnableDsEvictionKey] = v
 				}
--- a/cluster-autoscaler/simulator/nodes_test.go
+++ b/cluster-autoscaler/simulator/nodes_test.go
@@ -227,7 +227,7 @@ func buildDSPod(ds *appsv1.DaemonSet, nodeName string) *apiv1.Pod {
 	pod.Name = ds.Name
 	ptrVal := true
 	pod.ObjectMeta.OwnerReferences = []metav1.OwnerReference{
-		{Kind: "DaemonSet", UID: ds.UID, Controller: &ptrVal},
+		{APIVersion: "apps/v1", Kind: "DaemonSet", UID: ds.UID, Controller: &ptrVal},
 	}
 	return pod
 }
--- a/cluster-autoscaler/utils/pod/pod.go
+++ b/cluster-autoscaler/utils/pod/pod.go
@@ -31,7 +31,7 @@ const (
 // IsDaemonSetPod returns true if the Pod should be considered as Pod managed by a DaemonSet
 func IsDaemonSetPod(pod *apiv1.Pod) bool {
 	controllerRef := metav1.GetControllerOf(pod)
-	if controllerRef != nil && controllerRef.Kind == "DaemonSet" {
+	if controllerRef != nil && controllerRef.APIVersion == "apps/v1" && controllerRef.Kind == "DaemonSet" {
 		return true
 	}
 
--- a/cluster-autoscaler/utils/pod/pod.go
+++ b/cluster-autoscaler/utils/pod/pod.go
@@ -21,6 +21,7 @@ import (
 
 	apiv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 const (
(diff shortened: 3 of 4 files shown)


vadasambar:
I can do a first pass tomorrow if you don't get any reviews from the assigned reviewers. /assign vadasambar

vadasambar:
Thank you for the PR @shamil. 
/lgtm 
/approve

@x13n @BigDarkClown can you approve the PR :pray:

shamil:
@x13n @BigDarkClown can you please take a look on this change?

x13n:
Could we verify just the API group rather than the whole API version? If a new version of DS is created, this will break.

shamil:
> Could we verify just the API group rather than the whole API version? If a new version of DS is created, this will break.

@x13n you mean something like?

```go
strings.HasPrefix(controllerRef.APIVersion, "apps/")
```

x13n:
Ideally I'd check if there's some apimachinery util for this, but if there is none, I guess prefix should be ok.

shamil:
@x13n

I'm now retrieving the API Group version from `k8s.io/api/apps/v1`, this is still versioned but at least once we migrate from `v1` to newer API we won't need to adjust the code to support it. Let me know if that good enough.

x13n:
This doesn't solve the problem I mentioned - if a new version is added, let's say we start to see `v2beta1` DaemonSets, this code will no longer treat them like DaemonSets. `apps/v1` will still work as before. I was expecting something like https://github.com/kubernetes/apimachinery/blob/e2f405af78de67e27b5dd5ceb1a3eca76331e4d5/pkg/runtime/schema/group_version.go#L296-L305 be used to avoid direct string manipulation.

shamil:
@x13n 

Sorry for the confusion, I've recreated my last commit to use `FromAPIVersionAndKind` from `apimachinery`. 
Please let me know if any other change needed.

x13n:
Thanks, the code itself looks good now, but I think it will cause other problems: my guess is that anything called `DaemonSet` is a pod per node thing, is that true for the one you're using? The issue described in #5977 is caused by CA trying to get the owner ref from apiserver and failing. We actually want to treat some pods as DaemonSets in autoscaler, because they may exhibit DS-like behavior of popping up on every node (which CA simulations have to account for). So, I'd suggest keeping the existing DS checking code as-is, but fixing the check here instead: https://github.com/kubernetes/autoscaler/blob/5bf33b23f2bcf5f9c8ccaf99d445e25366ee7f40/cluster-autoscaler/simulator/drainability/rules/replicacount/rule.go#L73-L86

If there are DaemonSets that should not be treated as DaemonSets, perhaps we should extend support for `cluster-autoscaler.kubernetes.io/daemonset-pod` annotation to allow specifying `false` as a value.

Apologies for not spotting this earlier.

shamil:
@x13n if the code should decide if it daemonset just by its kind, then I agree, the code in `IsDaemonSetPod` method should be kept as is. But then it means we won't be able drain he node, because we don't have the required APIs to query for this custom DaemonSet. Right?

Just to understand what you're asking here, to fix this check I need to validate that the `DaemonSet` in question is not part of `apps` API group in this code? Or you also want to validate that the Daemonset in question running with custom controller?

x13n:
I'm suggesting a change to the code linked in my previous comment so that it only performs a lookup of DS object if the kind and api group are correct. You don't need to modify the criteria for what is considered to be a DS by autoscaler - only add a check before attempting to use DS lister. That way, pods owner by "regular" DS will have the extra check for owner existence, while all the "custom" DS pods will simply assume it is safe to drain. (Returning `NewUndefinedStatus` from this rule means: as far as this rule is concerned, there is no problem draining the pod, evaluation of other rules should decide.)

shamil:
@x13n Gotcha
I've created a separate PR (#6412) with the suggested changes. I think this one can be closed.

x13n:
Thanks, I'll take a look at the other PR, for now closing this one.

/close

PR #6346, opened 2023-12-05
Regenerate patch release schedule

x13n:
Schedule based on https://kubernetes.io/releases/patch-releases/#upcoming-monthly-releases. Included more rows to have one entry per maintainer.

#### What type of PR is this?

/kind documentation

@gjtempleton @BigDarkClown @towca PTAL

### DIFF
--- a/cluster-autoscaler/README.md
+++ b/cluster-autoscaler/README.md
@@ -87,12 +87,11 @@ target ETA and the actual releases.
 
 | Date       | Maintainer Preparing Release | Backup Maintainer |
 |------------|------------------------------|-------------------|
-| 2023-03-15 | MaciekPytel                  | gjtempleton       |
-| 2023-05-17 | gjtempleton                  | BigDarkClown      |
-| 2023-07-19 | BigDarkClown                 | towca             |
-| 2023-09-13 | towca                        | x13n              |
-| 2023-11-15 | x13n                         | MaciekPytel       |
 | 2024-01-17 | MaciekPytel                  | gjtempleton       |
+| 2024-03-13 | gjtempleton                  | BigDarkClown      |
+| 2024-05-?? | BigDarkClown                 | towca             |
+| 2024-07-?? | towca                        | x13n              |
+| 2024-09-?? | x13n                         | MaciekPytel       |
 
 Additional patch releases may happen outside of the schedule in case of critical
 bugs or vulnerabilities.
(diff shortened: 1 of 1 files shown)


towca:
/lgtm
/hold

The schedule LGTM, are the question marks intended? Maybe we can just leave the month? Hard to predict the exact day so far in the future.

gjtempleton:
Same question on the question marks as @towca though, giving an idea of at least the week we intend to cut the release makes it easier for us to give cut-offs for PRs people want to make the releases in advance.

x13n:
Ok, added some tentative dates (second-to-last Wed each month). Should be good enough for now and we can adjust later if needed.

x13n:
@gjtempleton @BigDarkClown @towca PTAL

x13n:
Also @MaciekPytel - since you're added as backup in the 2024-09 release.

gjtempleton:
Looks good to me.

One other thought I've had given the release of 1.29 over the weekend is whether we want to also set up a schedule for cutting the .0 releases for new versions of kubernetes?

MaciekPytel:
/lgtm

I see no reason for not adding a schedule for minor releases (.0 versions).
The only limitation is that I think k8s only publishes the timelines for the next release I think (?), so we may not be able to put specific dates. But a table mapping version to maintainer (e.g. 1.29.0 -> MaciekPytel) should be sufficient I think.

cluster-autoscaler/README.md
```diff
@@ -87,12 +87,11 @@ target ETA and the actual releases.
...
 | Date       | Maintainer Preparing Release | Backup Maintainer |
 |------------|------------------------------|-------------------|
-| 2023-03-15 | MaciekPytel                  | gjtempleton       |
-| 2023-05-17 | gjtempleton                  | BigDarkClown      |
-| 2023-07-19 | BigDarkClown                 | towca             |
-| 2023-09-13 | towca                        | x13n              |
-| 2023-11-15 | x13n                         | MaciekPytel       |
 | 2024-01-17 | MaciekPytel                  | gjtempleton       |
+| 2024-03-13 | gjtempleton                  | BigDarkClown      |
+| 2024-05-22 | BigDarkClown                 | towca             |
```
MaciekPytel, line 92: One minor comment - would you mind swapping backup shifts @towca? I would take this one and you'd take the one in September. 
My 10th marriage anniversary is in September and I estimate a very high chance that I will take some vacation to travel at this time :)

x13n:
We don't have specific dates for patches either, so all of them are approximate. Included minor versions in the schedule as well, PTAL.

cluster-autoscaler/README.md
x13n: After including minor releases this is no longer an issue, your anniversary travel plans have been de-risked now :)

gjtempleton:
Thanks.
/lgtm

x13n:
Thanks! This has been hanging for a while now, so let's merge and address any concerns in follow up PRs if needed.

/label tide/merge-method-squash
/unhold

PR #6361, opened 2023-12-08
feat: add metrics to show target size of every node group

guopeng0:
#### What type of PR is this?

/kind feature

This PR aims to add a new metric that displays the target size of all node groups.This metric is essential for obtaining a clear understanding of the target size of each node group while using the cluster autoscaler.

for example, we can on comparing this metric with the number of Kubernetes nodes to determine if the environment behaves as expected.This is crucial in detecting issues promptly, especially in the production environment.

The periodic update of this value has already been implemented in the UpdateNodes function of ClusterStateRegistry.We just need to add the reporting metrics functionality.This method does not increase the number of calls to the cloud provider nor does it affect other logic.

We kindly request your review of this PR and your thoughts on the proposed implementation. 

Thank you for your time and consideration.

Fixes #

```release-note

```
NONE

```docs

```

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -319,6 +319,7 @@ func (csr *ClusterStateRegistry) UpdateNodes(nodes []*apiv1.Node, nodeInfosForGr
 	if err != nil {
 		return err
 	}
+	metrics.UpdateNodeGroupTargetSize(targetSizes)
 
 	cloudProviderNodeInstances, err := csr.getCloudProviderNodeInstances()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


x13n:
Thanks for the change! It would be a bit more consistent with other metrics if it was exported with them:

https://github.com/kubernetes/autoscaler/blob/b4586f9b4468f595ac81632d55a1eda4a415dc99/cluster-autoscaler/core/static_autoscaler.go#L339-L345

Though I guess it could have performance implications for cloud provideres that don't cache target size, so let's merge as is.

/lgtm
/approve

guopeng0:
@x13n  Thank you so much for merging my Pull Request

PR #6364, opened 2023-12-11
fix(hetzner): insufficient nodes when boot fails

apricote:
#### What type of PR is this?

/kind bug

The Hetzner Cloud API returns "Actions" for anything asynchronous that happens inside the backend. When creating a new server multiple actions are returned: `create_server`, `start_server`, `attach_to_network` (if set).

Our current code waits for the `create_server` and if it fails, it makes sure to delete the server so cluster-autoscaler can create a new one immediately to provide the required capacity. If one of the "follow up" actions fails though, we do not handle this. This causes issues when the server for whatever reason did not start properly on the first try, as then the customer has a shutdown server, is paying for it, but does not receive the additional capacity for their Kubernetes cluster.

This commit fixes the bug, by awaiting all actions returned by the create server API call, and deleting the server if any of them fail.

https://docs.hetzner.cloud/#servers-create-a-server

-

I intend to back-port this fix to all current releases (1.26+)

```release-note
Fixed a bug where failed servers are kept for longer than necessary
```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/hetzner/hetzner_node_group.go
+++ b/cluster-autoscaler/cloudprovider/hetzner/hetzner_node_group.go
@@ -451,12 +451,18 @@ func createServer(n *hetznerNodeGroup) error {
 		return fmt.Errorf("could not create server type %s in region %s: %v", n.instanceType, n.region, err)
 	}
 
-	action := serverCreateResult.Action
 	server := serverCreateResult.Server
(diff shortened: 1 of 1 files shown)


jooola (approved):

Shubham82:
Thanks @apricote 
/lgtm

apricote:
Backports:

- #6367
- #6366
- #6365

PR #6371, opened 2023-12-12
fix: updated readme for hetzner cloud provider

pmdroid:
#### What type of PR is this?

/kind documentation

This provides a better way of letting someone know that it's also possible to use the id of a network, firewall or ssh key instead of only the name as before mentioned on the readme.


apricote:
Thanks!

/lgtm
/approve

PR #6378, opened 2023-12-15
Taint utils taking multiple taints

BigDarkClown:
#### What type of PR is this?

/kind cleanup

Update the taint methods to take multiple taints as arguments instead of one. This can be used to optimize the number of API calls needed to add/delete more than one taint at the time.

```release-note
NA
```

### DIFF
--- a/cluster-autoscaler/utils/taints/taints.go
+++ b/cluster-autoscaler/utils/taints/taints.go
@@ -151,16 +151,12 @@ func (tc TaintConfig) isExplicitlyReportedTaint(taint string) bool {
 	return ok
 }
 
-// getKeyShortName converts taint key to short name for logging
-func getKeyShortName(key string) string {
-	switch key {
-	case ToBeDeletedTaint:
-		return "ToBeDeletedTaint"
-	case DeletionCandidateTaint:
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/utils/taints/taints.go
```diff
@@ -215,22 +211,34 @@ func AddTaint(node *apiv1.Node, client kube_client.Interface, taint apiv1.Taint,
...
 
-func addTaintToSpec(node *apiv1.Node, taint apiv1.Taint, cordonNode bool) bool {
-	for _, t := range node.Spec.Taints {
-		if t.Key == taint.Key {
-			klog.V(2).Infof("%v already present on node %v, t: %v", taint.Key, node.Name, t)
-			return false
+func addTaintsToSpec(node *apiv1.Node, taints []apiv1.Taint, cordonNode bool) bool {
+	taintsAdded := false
+	for _, taint := range taints {
+		taintPresent := false
```
x13n, line 225: Instead of a nested loop, it would be a bit more readable to use `HasTaint` function which already exists in this file.

x13n (changes requested):
Just one minor comment, otherwise LGTM.

PR #6186, opened 2023-10-10
Azure: Remove AKS vmType

jackfrancis:
#### What type of PR is this?

/kind deprecation
/kind documentation
/kind cleanup

#### What this PR does / why we need it:

This PR removes the obsolete, non-working implementation for manual AKS cluster integration as expressed via the `"aks"` `vmType` configuration.

Node autoscaling has long been a first class GA feature of AKS. See:

- https://docs.microsoft.com/azure/aks/autoscaler

Running the Azure cluster-autoscaler provider manually on your cluster is not a supported feature. To aid future maintainability of the project, we're removing these non-working code paths and updating documentation.

#### Which issue(s) this PR fixes:

Fixes #6190

#### Does this PR introduce a user-facing change?

```release-note
Azure: Remove AKS vmType
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #6384, opened 2023-12-15
azure: fix chart bugs after AKS vmType deprecation

jackfrancis:
#### What type of PR is this?

/kind bug

This PR fixes some chart errata introduced in https://github.com/kubernetes/autoscaler/pull/6186 (mea culpa). Specifically:

- PR https://github.com/kubernetes/autoscaler/pull/6186 removed the "AKS" vmType configuration, and so we don't to default the `azureVMType` chart value to be "AKS"
- PR https://github.com/kubernetes/autoscaler/pull/6186 removed the Azure provider "clusterName" configuration, so we don't want a reference to that as a required value in the chart deployment spec
- PR https://github.com/kubernetes/autoscaler/pull/6186 removed the Azure provider "nodeResourceGroup" configuration, so we don't want a reference to that as a required value in the chart deployment spec

Here is a representative example in PR https://github.com/kubernetes/autoscaler/pull/6186 of the removals mentioned above:

- https://github.com/kubernetes/autoscaler/pull/6186/files#diff-e350455eeabc5662e1fc8162e570b2f45f95699270a30ce51443ba61afa49da6

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/charts/cluster-autoscaler/Chart.yaml
+++ b/charts/cluster-autoscaler/Chart.yaml
@@ -11,4 +11,4 @@ name: cluster-autoscaler
 sources:
   - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
 type: application
-version: 9.34.0
+version: 9.34.1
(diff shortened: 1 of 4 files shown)


mwielgus (approved):
/lgtm
/approve

PR #6391, opened 2023-12-20
Fix VPA e2e test failures

### LINKED ISSUE #6389: VPA E2E tests are broken
This is blocking the VPA 1.1.0 release

vertical-pod-autoscaler 

gcr.io/k8s-infra-e2e-boskos-079/vpa-updater-amd64:latest

gcr.io/k8s-infra-e2e-boskos-079/vpa-admission-controller-amd64:latest

Testgrid: https://testgrid.k8s.io/sig-autoscaling-vpa#Summary

Logs of the latest failed run: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-autoscaling-vpa-actuation/1737090055995920384

voelzmo:
#### What type of PR is this?

/kind bug

Fixes failing e2e tests for VPA. After the refactorings done in #5599, the tests started failing – but nobody realized this until now: #6389

This PR fixes the tests by doing two things:
* The way the new builder pattern for VPA test objects was introduced in the e2e tests had a problem with how recommendations are created. For the first container, a recommendation is created implicitly when you call `Build()` on the VPA object. This interferes with the pattern of `AppendRecommendation()`, which lead to a second entry in the `containerRecommendations` section, where one of those entries was empty.
* During the refactorings mentioned above, the test for the named recommender were broken and we didn't realize it. This PR brings back the recommender name for the VPA object created in the test, such that it can succeed again.

Fixes #6389

```release-note
NONE
```

### DIFF
--- a/vertical-pod-autoscaler/pkg/utils/test/test_vpa.go
+++ b/vertical-pod-autoscaler/pkg/utils/test/test_vpa.go
@@ -248,7 +248,12 @@ func (b *verticalPodAutoscalerBuilder) Get() *vpa_types.VerticalPodAutoscaler {
 			Mode:             b.scalingMode[containerName],
 		})
 	}
-	recommendation = b.recommendation.WithContainer(b.containerNames[0]).Get()
+	// VPAs with a single container may still use the old/implicit way of adding recommendations
+	r := b.recommendation.WithContainer(b.containerNames[0]).Get()
+	if r.ContainerRecommendations[0].Target != nil {
+		recommendation = r
+	}
+
 	recommendation.ContainerRecommendations = append(recommendation.ContainerRecommendations, b.appendedRecommendations...)
 
 	return &vpa_types.VerticalPodAutoscaler{
--- a/vertical-pod-autoscaler/e2e/v1/full_vpa.go
+++ b/vertical-pod-autoscaler/e2e/v1/full_vpa.go
@@ -238,6 +238,7 @@ var _ = FullVpaE2eDescribe("Pods under VPA with non-recognized recommender expli
 		containerName := GetHamsterContainerNameByIndex(0)
 		vpaCRD := test.VerticalPodAutoscaler().
 			WithName("hamster-vpa").
+			WithRecommender("non-recognized").
 			WithNamespace(f.Namespace.Name).
 			WithTargetRef(targetRef).
 			WithContainer(containerName).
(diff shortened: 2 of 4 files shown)


voelzmo:
ping @laoj2 this fixes the tests and should enable us to cut a new release soon

vertical-pod-autoscaler/pkg/utils/test/test_vpa.go
```diff
@@ -248,7 +248,12 @@ func (b *verticalPodAutoscalerBuilder) Get() *vpa_types.VerticalPodAutoscaler {
 			Mode:             b.scalingMode[containerName],
 		})
 	}
-	recommendation = b.recommendation.WithContainer(b.containerNames[0]).Get()
+	// VPAs with a single container may still use the old/implicit way of adding recommendations
+	r := b.recommendation.WithContainer(b.containerNames[0]).Get()
```
laoj2, line 252: Why do we hardcode the container name index here and append to `recommendation.ContainerRecommendations` instead of doing this logic for all containers in the loop above?

vertical-pod-autoscaler/pkg/utils/test/test_vpa.go
```diff
@@ -248,7 +248,12 @@ func (b *verticalPodAutoscalerBuilder) Get() *vpa_types.VerticalPodAutoscaler {
 			Mode:             b.scalingMode[containerName],
 		})
 	}
-	recommendation = b.recommendation.WithContainer(b.containerNames[0]).Get()
+	// VPAs with a single container may still use the old/implicit way of adding recommendations
```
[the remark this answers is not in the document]
voelzmo: I think this weird split was introduced a long time ago when the option was added to provide more than one recommendation to the test VPA object: https://github.com/kubernetes/autoscaler/pull/2866/files#diff-294b6f5855b82faa084f2866e8d3542c0c129776ecbbfd44f8d921f210706490

You're right, ideally this should be simplified to work with a single way to set the recommendations. We probably should do this in a follow-up PR to get the tests green again first?

vertical-pod-autoscaler/pkg/utils/test/test_vpa.go
voelzmo: See comment above, probably the same issue/root cause?

voelzmo:
Hey @laoj2 thanks for the review! I agree that this part is a mess in the test code and we should clean it up. I'd rather do this in a follow-up PR and get the tests green as soon as possible, does that sound good?

voelzmo:
ping @laoj2 can we merge this and get the tests back to green again, so we  can cut a new release, or do you want me to do any additional changes on this PR?

laoj2:
Yes, sounds good @voelzmo. Thanks for the fix!
 
/lgtm 

cc @kwiesmueller

PR #6396, opened 2023-12-21
feat:add node group health and back off metrics

guopeng0:
#### What type of PR is this?
/kind feature

This pull request aims to introduce new metrics to display the health status and back off situation of node groups in the cluster. Currently, there are cluster health metrics, but this PR intends to add additional metrics at the node group level.

The two new metrics being added are:
node_group_healthiness: This metric will indicate the health status of each node group, providing clear visibility into whether the node group is currently healthy or not. If a node group remains unhealthy for a prolonged period, it can trigger alerts for further investigation.
node_group_backoff_status: This metric displays the back off status of each node group, along with the specific reasons causing it.

Implementation Details:
The proposed changes involve adding an updateMetrics function to the ClusterStateRegistry class. This function will be called at the end of each RunOnce execution, ensuring that the new metrics are regularly updated. While these metrics are primarily used in scale-up scenarios, it is important to keep them updated continuously, not just during scale-up.

We kindly request your review of this PR and your thoughts on the proposed implementation.
Thank you for your time and consideration.

Fixes #

NONE

```docs

```

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -456,6 +456,18 @@ func (csr *ClusterStateRegistry) updateNodeGroupMetrics() {
 	metrics.UpdateNodeGroupsCount(autoscaled, autoprovisioned)
 }
 
+// UpdateSafeScaleUpMetricsForNodeGroup queries the health status and backoff situation of the node group and updates metrics
+func (csr *ClusterStateRegistry) UpdateSafeScaleUpMetricsForNodeGroup(now time.Time) {
+	for _, nodeGroup := range csr.cloudProvider.NodeGroups() {
+		if !nodeGroup.Exist() {
+			continue
+		}
+		metrics.UpdateNodeGroupHealthStatus(nodeGroup.Id(), csr.IsNodeGroupHealthy(nodeGroup.Id()))
+		backoffStatus := csr.backoff.BackoffStatus(nodeGroup, csr.nodeInfosForGroups[nodeGroup.Id()], now)
+		metrics.UpdateNodeGroupBackOffStatus(nodeGroup.Id(), backoffStatus.IsBackedOff, backoffStatus.ErrorInfo.ErrorCode)
+	}
+}
+
 // IsNodeGroupSafeToScaleUp returns information about node group safety to be scaled up now.
 func (csr *ClusterStateRegistry) IsNodeGroupSafeToScaleUp(nodeGroup cloudprovider.NodeGroup, now time.Time) NodeGroupScalingSafety {
 	isHealthy := csr.IsNodeGroupHealthy(nodeGroup.Id())
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -405,6 +405,7 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
 		if err != nil {
 			klog.Errorf("AutoscalingStatusProcessor error: %v.", err)
 		}
+		a.clusterStateRegistry.UpdateSafeScaleUpMetricsForNodeGroup(currentTime)
 	}()
 
 	// Check if there are any nodes that failed to register in Kubernetes
--- a/cluster-autoscaler/metrics/metrics.go
+++ b/cluster-autoscaler/metrics/metrics.go
@@ -208,6 +208,22 @@ var (
 		}, []string{"node_group"},
 	)
 
+	nodesGroupHealthiness = k8smetrics.NewGaugeVec(
+		&k8smetrics.GaugeOpts{
+			Namespace: caNamespace,
+			Name:      "node_group_healthiness",
+			Help:      "Whether or not node group is healthy enough for autoscaling. 1 if it is, 0 otherwise.",
+		}, []string{"node_group"},
+	)
+
+	nodeGroupBackOffStatus = k8smetrics.NewGaugeVec(
+		&k8smetrics.GaugeOpts{
+			Namespace: caNamespace,
+			Name:      "node_group_backoff_status",
+			Help:      "Whether or not node group is backoff for not autoscaling. 1 if it is, 0 otherwise.",
+		}, []string{"node_group", "reason"},
+	)
+
 	/**** Metrics related to autoscaler execution ****/
 	lastActivity = k8smetrics.NewGaugeVec(
 		&k8smetrics.GaugeOpts{
@@ -431,6 +447,8 @@ func RegisterAll(emitPerNodeGroupMetrics bool) {
 		legacyregistry.MustRegister(nodesGroupMinNodes)
 		legacyregistry.MustRegister(nodesGroupMaxNodes)
 		legacyregistry.MustRegister(nodesGroupTargetSize)
+		legacyregistry.MustRegister(nodesGroupHealthiness)
+		legacyregistry.MustRegister(nodeGroupBackOffStatus)
 	}
 }
 
@@ -533,6 +551,24 @@ func UpdateNodeGroupMax(nodeGroup string, maxNodes int) {
 func UpdateNodeGroupTargetSize(targetSizes map[string]int) {
 	for nodeGroup, targetSize := range targetSizes {
 		nodesGroupTargetSize.WithLabelValues(nodeGroup).Set(float64(targetSize))
+	}
+}
+
+// UpdateNodeGroupHealthStatus records if node group is healthy to autoscaling
+func UpdateNodeGroupHealthStatus(nodeGroup string, healthy bool) {
+	if healthy {
+		nodesGroupHealthiness.WithLabelValues(nodeGroup).Set(1)
+	} else {
+		nodesGroupHealthiness.WithLabelValues(nodeGroup).Set(0)
+	}
+}
+
+// UpdateNodeGroupBackOffStatus records if node group is backoff for not autoscaling
+func UpdateNodeGroupBackOffStatus(nodeGroup string, backOff bool, reason string) {
+	if backOff {
+		nodeGroupBackOffStatus.WithLabelValues(nodeGroup, reason).Set(1)
+	} else {
+		nodeGroupBackOffStatus.WithLabelValues(nodeGroup, reason).Set(0)
 	}
 }
 


cluster-autoscaler/metrics/metrics.go
```diff
@@ -536,6 +554,24 @@ func UpdateNodeGroupTargetSize(targetSizes map[string]int) {
...
+		nodesGroupHealthiness.WithLabelValues(nodeGroup).Set(1)
+	} else {
+		nodesGroupHealthiness.WithLabelValues(nodeGroup).Set(0)
+	}
+}
+
+// UpdateNodeGroupBackOffStatus records if node group is backoff for not autoscaling
+func UpdateNodeGroupBackOffStatus(nodeGroup string, backOff bool, reason string) {
+	if backOff {
+		nodeGroupBackOffStatus.WithLabelValues(nodeGroup, reason).Set(1)
```
x13n, line 569: This will not work the way you want: once the backoff is over, the value of 1 will keep on being emitted for the old reason. Consider:

1. `UpdateNodeGroupBackOffStatus("foo", true, "reason")`

   Under /metrics endpoint you'll get something like:
   
   ```
   node_group_backoff_status{node_group=foo,reason=reason} 1
   ```

2. `UpdateNodeGroupBackOffStatus("foo", false, "")`

   Under /metrics endpoint you'll get something like:
   
   ```
   node_group_backoff_status{node_group=foo,reason=reason} 1
   node_group_backoff_status{node_group=foo,reason=""} 0
   ```

What you may want to do instead is to use a string metric with value equal to the backoff reason.

x13n (changes requested):

cluster-autoscaler/metrics/metrics.go
guopeng0: I agree with the suggestion and will make the necessary modifications as soon as possible.

guopeng0: Modifications done and PR submitted.

x13n: Apologies for not being clear - I think the current version of the code will have exactly the same issue: I expect `reason` to be an empty string every time `backoff` is `false`. If you want to track backoff reasons like this, you need to keep track of the last `reason` value and explicitly clear it when `backoff` becomes `false`. Alternatively, you could implement a custom collector that will only emit metrics with up-to-date reason. This would be the cleanest solution. With the existing code, in the scenario from my previous comment, after backoff is over, you get:

```
node_group_backoff_status{node_group=foo,reason=reason} 1
node_group_backoff_status{node_group=foo,reason=""} 0
```

If you instead track nodegroup -> reason mapping and clear the last reason you will instead get:

```
node_group_backoff_status{node_group=foo,reason=reason} 0
```

If you decide to write a custom collector, once backoff is over, you could simply get empty output:

```
```

guopeng0: Thank you for raising this issue, and you’re right, I misunderstood it. It is indeed not expected behavior that the previous nodeGroup backoff metric continues to be reported as 1 after the backoff is over. I will take some time to reconsider how to address this and submit the code changes as soon as possible.

guopeng0: Modifications done and PR submitted.

guopeng0: @x13n  Please review my changes

cluster-autoscaler/metrics/metrics.go
```diff
@@ -543,6 +562,24 @@ func UpdateNodeGroupTargetSize(targetSizes map[string]int) {
...
+		nodesGroupHealthiness.WithLabelValues(nodeGroup).Set(1)
+	} else {
+		nodesGroupHealthiness.WithLabelValues(nodeGroup).Set(0)
+	}
+}
+
+// UpdateNodeGroupBackOffStatus records if node group is backoff for not autoscaling
+func UpdateNodeGroupBackOffStatus(nodeGroup string, backOff bool, errorClass cloudprovider.InstanceErrorClass) {
+	if backOff {
+		nodeGroupBackOffStatus.WithLabelValues(nodeGroup).Set(float64(errorClass))
```
x13n, line 577: That is quite hacky and also meaningless when aggregated. Adding error codes together doesn't make sense and yet it is exactly what will happen when querying for multiple node groups. The previous approach with label for error code makes sense, you just need to make sure it stops reporting 1 when the node group is no longer in backoff.

guopeng0: Thank you for your suggestion. I just wanted to confirm the expected metric behavior:
There are two ways to display this metrics since it involves different handling scenarios based on the reason. 
I am not sure which approach is more appropriate, and I would appreciate any advice you can provide.

x13n: I think both are fine, though I slightly prefer the former. The reason is that whatever is collecting these metrics, will be able to clearly say that a given metric stream is "done" when backoff reason changes or goes away. With the second approach I think there can be artifacts on the graph, since a dashboard won't correlate different streams without some careful query crafting.

The way I would implement this is by keeping a map with {node_group, reason} pair as a key. It would then be updated in bulk:
1. Iterate through the map and set all values to 0
2. Iterate through the node groups and set corresponding {node_group, reason} value to 1 (or `true`). Optionally, inject {node_group, ""} keys set to 0 for all node groups that are not in backoff. Not sure if that second part is actually useful for anything, perhaps not.
3. Use a custom collector that will read values from the map when prometheus endpoint is scraped.

guopeng0: Understood, thank you! and submit the code changes as soon as possible.

guopeng0: Modifications done and PR submitted. Because the errorCode is provided by the nodeProvider and is open, this collector is dynamic and will search all historical collections of reported reasons.

guopeng0: @x13n

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -116,6 +116,9 @@ type ScaleUpFailure struct {
 	Time      time.Time
 }
 
+// BackoffReasonStatus contains information about backoff status and reason
+type BackoffReasonStatus map[string]int
```
x13n, line 120: Since the values here will always be either 0 or 1, maybe `map[string]bool` would make more sense?

cluster-autoscaler/cloudprovider/cloud_provider.go
```diff
@@ -267,6 +267,11 @@ const (
 	InstanceDeleting InstanceState = 3
 )
 
+const (
+	// UnknownErrorCode means that the cloud provider has not provided an error code.
+	UnknownErrorCode = "unknown"
```
x13n, line 272: Why is this in `cloudprovider` package?

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -462,6 +467,38 @@ func (csr *ClusterStateRegistry) updateNodeGroupMetrics() {
 	metrics.UpdateNodeGroupsCount(autoscaled, autoprovisioned)
 }
 
+// UpdateSafeScaleUpMetricsForNodeGroup queries the health status and backoff situation of the node group and updates metrics
+func (csr *ClusterStateRegistry) UpdateSafeScaleUpMetricsForNodeGroup(now time.Time) {
```
x13n, line 471: Both the `ForNodeGroup` suffix and the comment above suggest this is about a specific node group, while in fact it iterates over all of them.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -406,6 +406,7 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
 		if err != nil {
 			klog.Errorf("AutoscalingStatusProcessor error: %v.", err)
 		}
+		a.clusterStateRegistry.UpdateSafeScaleUpMetricsForNodeGroup(currentTime)
```
x13n, line 409: This would better fit as an `AutoscalingStatusProcessor` implementation - all the relevant params are already passed in the `Process` call above. Let's update this function:

https://github.com/kubernetes/autoscaler/blob/a3a29cf5ddfbfffd6518af9a89b644fbb31c895f/cluster-autoscaler/processors/status/autoscaling_status_processor.go#L33-L35

to return a new processor that does the metric update. The new processor would also be a better place to keep the `backoffReasonStatus` map - `clusterStateRegistry` should be responsible for tracking the cluster state, not for metrics handling.

x13n (commented):
Apologies for late response, added some comments.

cluster-autoscaler/core/static_autoscaler.go
guopeng0: Modifications done and PR submitted

cluster-autoscaler/clusterstate/clusterstate.go
guopeng0: Modifications done and PR submitted

cluster-autoscaler/cloudprovider/cloud_provider.go
guopeng0: Modifications done and PR submitted

cluster-autoscaler/clusterstate/clusterstate.go
guopeng0: Modifications done and PR submitted

cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -470,8 +475,28 @@ func (csr *ClusterStateRegistry) UpdateSafeScaleUpMetricsForNodeGroup(now time.T
...
 		metrics.UpdateNodeGroupHealthStatus(nodeGroup.Id(), csr.IsNodeGroupHealthy(nodeGroup.Id()))
 		backoffStatus := csr.backoff.BackoffStatus(nodeGroup, csr.nodeInfosForGroups[nodeGroup.Id()], now)
-		metrics.UpdateNodeGroupBackOffStatus(nodeGroup.Id(), backoffStatus.IsBackedOff, backoffStatus.ErrorInfo.ErrorClass)
+		csr.updateNodeGroupBackoffStatusMetrics(nodeGroup.Id(), backoffStatus)
+	}
+}
+
+// updateNodeGroupBackoffStatusMetrics returns information about backoff situation and reason of the node group
+func (csr *ClusterStateRegistry) updateNodeGroupBackoffStatusMetrics(nodeGroup string, backoffStatus backoff.Status) {
+	backoffReasonStatus := make(BackoffReasonStatus)
```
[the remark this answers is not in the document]
guopeng0: Modifications done and PR submitted

x13n:
Thanks for the changes!

/lgtm
/approve

lukasmrtvy:
How should this work for AWS when there are no spots available? For example `Could not launch Spot Instances. UnfulfillableCapacity - Unable to fulfill capacity due to your request configuration. Please adjust your request and try again. Launching EC2 instance failed.` 

Thanks

PR #6274, opened 2023-11-14
Cleanup: Remove separate client for k8s events

azylinski:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

There are two parts to this change:
 1) Remove kube-client RateLimiting - replay on APF for apiserver protection. Details: https://github.com/kubernetes/kubernetes/issues/111880
 2) Remove dedicated client for events, for the same reasons

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

/assign @MaciekPytel
/assign @x13n

PR #6405, opened 2023-12-28
Restore flags for setting QPS limit in CA

jbartosik:
Partially undo #6274. I noticed that with this change CA get rate limited and slows down significantly (especially during large scale downs).

/kind bug

Fix a regression - I noticed that CA is significantly slower (especially during large scale downs) because of rate limiting.

### DIFF
--- a/cluster-autoscaler/main.go
+++ b/cluster-autoscaler/main.go
@@ -103,8 +103,8 @@ var (
 	kubernetes              = flag.String("kubernetes", "", "Kubernetes master location. Leave blank for default")
 	kubeConfigFile          = flag.String("kubeconfig", "", "Path to kubeconfig file with authorization and master location information.")
 	kubeAPIContentType      = flag.String("kube-api-content-type", "application/vnd.kubernetes.protobuf", "Content type of requests sent to apiserver.")
-	_                       = flag.Int("kube-client-burst", rest.DefaultBurst, "Burst value for kubernetes client. (Deprecated, relay on APF for rate limiting)")
-	_                       = flag.Float64("kube-client-qps", float64(rest.DefaultQPS), "QPS value for kubernetes client. (Deprecated, relay on APF for rate limiting)")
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/main.go
```diff
@@ -103,8 +103,8 @@ var (
 	kubernetes              = flag.String("kubernetes", "", "Kubernetes master location. Leave blank for default")
 	kubeConfigFile          = flag.String("kubeconfig", "", "Path to kubeconfig file with authorization and master location information.")
 	kubeAPIContentType      = flag.String("kube-api-content-type", "application/vnd.kubernetes.protobuf", "Content type of requests sent to apiserver.")
-	_                       = flag.Int("kube-client-burst", rest.DefaultBurst, "Burst value for kubernetes client. (Deprecated, relay on APF for rate limiting)")
-	_                       = flag.Float64("kube-client-qps", float64(rest.DefaultQPS), "QPS value for kubernetes client. (Deprecated, relay on APF for rate limiting)")
+	kubeClientBurst         = flag.Int("kube-client-burst", rest.DefaultBurst, "Burst value for kubernetes client. (Deprecated, relay on APF for rate limiting)")
```
MaciekPytel, line 106: Can we remove the deprecated part? I don't think those should be deprecated, at least not with the current semantic of defaulting to low limit.

cluster-autoscaler/main.go
```diff
@@ -103,8 +103,8 @@ var (
...
+	kubeClientQPS           = flag.Float64("kube-client-qps", float64(rest.DefaultQPS), "QPS value for kubernetes client. (Deprecated, relay on APF for rate limiting)")
```
MaciekPytel, line 107: Same as above

cluster-autoscaler/main.go
jbartosik: Done

MaciekPytel:
/lgtm
/approve
Would you mind also opening a backport to 1.29? I'll lgtm and include it in 1.29.1 (which I'll be building in January as per the schedule)

PR #6412, opened 2023-12-29
Allow draining when DaemonSet kind has custom API Group

### LINKED ISSUE #5977: Autoscaler not scaling down pods when`kind` is of the resource is same between controllers

shamil:
#### What type of PR is this?

/kind bug

The issue is described in #5977

Fixes #5977
Suppresses #6328

This PR is a continuation from #6328. Made the changes as requested by @x13n 

NONE

### DIFF
--- a/cluster-autoscaler/simulator/drainability/rules/replicacount/rule.go
+++ b/cluster-autoscaler/simulator/drainability/rules/replicacount/rule.go
@@ -21,6 +21,7 @@ import (
 
 	apiv1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator/drainability"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/drain"
 	pod_util "k8s.io/autoscaler/cluster-autoscaler/utils/pod"
@@ -57,7 +58,10 @@ func (r *Rule) Drainable(drainCtx *drainability.DrainContext, pod *apiv1.Pod) dr
 	if controllerRef == nil {
 		return drainability.NewUndefinedStatus()
 	}
-	refKind := controllerRef.Kind
+
+	groupVersionKind := schema.FromAPIVersionAndKind(controllerRef.APIVersion, controllerRef.Kind)
+	refGroup := groupVersionKind.Group
+	refKind := groupVersionKind.Kind
 
 	if refKind == "ReplicationController" {
 		rc, err := drainCtx.Listers.ReplicationControllerLister().ReplicationControllers(controllerNamespace).Get(controllerRef.Name)
@@ -71,8 +75,8 @@ func (r *Rule) Drainable(drainCtx *drainability.DrainContext, pod *apiv1.Pod) dr
 			return drainability.NewBlockedStatus(drain.MinReplicasReached, fmt.Errorf("replication controller for %s/%s has too few replicas spec: %d min: %d", pod.Namespace, pod.Name, rc.Spec.Replicas, r.minReplicaCount))
 		}
 	} else if pod_util.IsDaemonSetPod(pod) {
-		if refKind != "DaemonSet" {
-			// We don't have a listener for the other DaemonSet kind.
+		if refGroup != "apps" || refKind != "DaemonSet" {
+			// We don't have a listener for the other DaemonSet group or kind.
 			// TODO: Use a generic client for checking the reference.
 			return drainability.NewUndefinedStatus()
 		}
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/simulator/drainability/rules/replicacount/rule.go
```diff
@@ -57,7 +58,10 @@ func (r *Rule) Drainable(drainCtx *drainability.DrainContext, pod *apiv1.Pod) dr
 	if controllerRef == nil {
 		return drainability.NewUndefinedStatus()
 	}
-	refKind := controllerRef.Kind
+
+	groupVersionKind := schema.FromAPIVersionAndKind(controllerRef.APIVersion, controllerRef.Kind)
```
vadasambar, line 62: Docs for `FromAPIVersionAndKind` says:
```go
// FromAPIVersionAndKind returns a GVK representing the provided fields for types that
// do not use TypeMeta. This method exists to support test types and legacy serializations
// that have a distinct group and kind.
// TODO: further reduce usage of this method.
```
https://github.com/kubernetes/apimachinery/blob/e2f405af78de67e27b5dd5ceb1a3eca76331e4d5/pkg/runtime/schema/group_version.go#L296-L305

`DaemonSet` uses `TypeMeta`: https://pkg.go.dev/k8s.io/api/apps/v1#DaemonSet

Should we be using `TypeMeta`'s [`GroupVersionKind()`](https://github.com/kubernetes/apimachinery/blob/v0.29.0/pkg/apis/meta/v1/meta.go#L125) function here instead? `GroupVersionKind()` uses [`FromAPIVersionAndKInd()` internally](https://github.com/kubernetes/apimachinery/blob/3c8c1f22dc332a8a6b51033f6a7c0566b885d290/pkg/apis/meta/v1/meta.go#L125-L127) but there is less chance of `GroupVersionKind()` getting removed in the future (while `FromAPIVersionKind()` might)  

Something like (at the cost of more verbosity):
```go
v1.TypeMeta{Kind: controllerRef.Kind, APIVersion: controllerRef.APIVersion}.GroupVersionKind()
```

vadasambar: cc: @x13n

x13n: How about:

```
refGroup := schema.ParseGroupVersion(controllerRef.APIVersion)
refKind := controllerRef.Kind
```

`ParseGroupVersion` isn't annotated as something that is planned for removal.

vadasambar: Sounds good to me :+1:

shamil: Hi @vladimirvivien , @x13n 

`schema.ParseGroupVersion` returns error in addition to `schema.GroupVersion`, will the following code be sufficient?

```go
refGroup, err := schema.ParseGroupVersion(controllerRef.APIVersion)
if err != nil {
	return drainability.NewUndefinedStatus()
}

refKind := controllerRef.Kind
```

Or I should return `drainability.NewBlockedStatus` instead?

x13n: Undefined is ok in this case. If it didn't parse, it means it wasn't a correctly formatted `APIVersion` string.

shamil: Allright, changed

Issue #6333, opened 2023-12-01
Excessive cient-side throttling in the clusterapi provider for cluster-autoscaler


kdw174:
**Which component are you using?**:
cluster-autoscaler

**What version of the component are you using?**:
1.27.1 and 1.26.2

**What environment is this in?**:
clusterapi

**What did you expect to happen?**:
cluster-autoscaler to meet the SLO described [here](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-service-level-objectives-for-cluster-autoscaler)

**What happened instead?**:
We are seeing excessive, 200-300 times a minute, client-side throttling in our clusters that have a higher number of machinedeployments and nodes. ~50 machinedeployments and 100-200 nodes. 
```
Waited for 194.425534ms due to client-side throttling, not priority and fairness, request: GET:https://{apiserver}:443/apis/cluster.x-k8s.io/v1beta1/namespaces/{namesapce}/machinedeployments/{machinedeplyoment}/scale
```

This is causing the main loop to take close to 2 minutes in some cases.
```
2023-12-01 12:55:21.692 | I1201 12:55:21.692565       1 metrics.go:409] Function main took 1m46.646684648s to complete
```

When we need to scale up multiple machinedeployments at once, this can severely delay a pod getting scheduled. We have seen pods pending for over an hour. Let's say we have 10 machinedeployments that all need to scale up at once because of a rolling deployment on dedicated nodes or scaling event. It will take a minimum of 10*1m46s to scale up replicas in each machinedeployment from the initial list. If additional machinedeployments need scaled during that scaling period, it will make it take even longer.

**How to reproduce it (as minimally and precisely as possible)**:
Use cluster-autoscaler with the clusterapi provider. Setup a cluster with a higher number of nodes (100-200) and machinedeployments (~50). You should see heavy client-side throttling in the logs and the main function taking over a minute to complete.

If you scale up pod replicas so multiple (10) machinedeployments scale up at once, it will take many minutes for the last pod to get scheduled on a node. 

**Anything else we need to know?**:
The clusterapi provider creates additional [kube client configs](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_provider.go#L157-L162) that use the default qps and burst configuration in the rest client. Currently, there is no way to override those values to either increase the rate limit or remove it all together and rely on  the API Priority and Fairness (APF). 

[...]

PR #6416, opened 2024-01-02
Pass Burst and QPS client params to clusterapi k8s clients

kdw174:
#### What type of PR is this?
/kind bug

Passes the kubeClientBurst and kubeClientQPS flags to the clusterapi provider k8s clients to allow tuning of client and avoid excessive client-side throttling. 

Fixes https://github.com/kubernetes/autoscaler/issues/6333

```release-note
kubeClientBurst and kubeClientQPS flags are now passed to the clusterapi provider k8s clients
```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_provider.go
+++ b/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_provider.go
@@ -158,6 +158,8 @@ func BuildClusterAPI(opts config.AutoscalingOptions, do cloudprovider.NodeGroupD
 	if err != nil {
 		klog.Fatalf("cannot build management cluster config: %v", err)
 	}
+	managementConfig.QPS = float32(opts.KubeClientQPS)
+	managementConfig.Burst = opts.KubeClientBurst
 
 	workloadKubeconfig := opts.KubeClientOpts.KubeConfigPath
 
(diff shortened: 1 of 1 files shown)


elmiko:
thanks for the PR @kdw174 !

i think this looks good, i wonder if we should add something to the clusterapi/README.md to let users know about this?

also, do you think we would ever need to vary these values between the management kubeconfig and the workload kubeconfig?

elmiko:
my second question is more speculative, i think this PR is probably good as it is for now.

kdw174:
I don't think we should add that as part of the clusterapi readme since this parameter isn't clusterapi provider specific. The clients are going to have the same default QPS and Burst config they had before. 

I could add it to the cluster-autoscaler FAQ here https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca if you think that makes sense

elmiko:
thanks. i feel like we should definitely have these options in the main FAQ, but it might be better to open that as another PR since it will require approvers from the core team to approve.

i'm ready to approve this, but i just thought of one other question. is it possible that a user could set the QPS and burst in the kubeconfig itself? (i'm not sure this is even possible but i want to make sure we don't overwrite a config value from the kubeconfig if it's already present)

kdw174:
I don't believe it's possible to set the qps or burst from the kubeconfig as this is the pattern the default client follows in cluster-autoscaler. Digging through the [BuildConfigFromFlags](https://github.com/kubernetes/client-go/blob/v0.29.0/tools/clientcmd/client_config.go#L616) function that's called, I wasn't able to find anywhere that the qps could be set from the kubeconfig or an env var.

elmiko:
ack, thanks for investigating @kdw174 

/approve
/lgtm

PR #6418, opened 2024-01-04
Allow ignoring volumes in drainable check

### LINKED ISSUE #6417: Local PV will prevent node scaling down

cluster-autoscaler

Component version: v1.26.3

$ kubectl version
Client Version: v1.28.1
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.26.11-eks-8cb36c9
</pre></details>

EKS

Under-utilized nodes with local PV should be scaled down

**What happened instead?**:

Under-utilized nodes with local PV cannot be scaled down

If pod has local PV (provisioned by [local-volume-provisioner](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/docs/provisioner.md) in our use case) mounted
CA refuses to evict the pod because the local pv has NodeAffinity for specific node
There is no way we can bypass this restriction to scale down those under-utilized nodes

Specifying `safe-to-evict-local-volumes` and `safe-to-evict` does no help,
because pod will eventually enter bounded PV check where it will fail

I think we should provide an option to remove some volumes before bounded PV check

jewelzqiu:
#### What type of PR is this?

/kind feature

Fixes #6417

```release-note
Allow ignoring volumes in drainable check
```

```docs

```

### DIFF
--- a/cluster-autoscaler/simulator/drain.go
+++ b/cluster-autoscaler/simulator/drain.go
@@ -52,6 +52,7 @@ func GetPodsToMove(nodeInfo *schedulerframework.NodeInfo, deleteOptions options.
 	}
 	for _, podInfo := range nodeInfo.Pods {
 		pod := podInfo.Pod
+		drain.RemoveIgnoredVolumes(pod)
 		status := drainabilityRules.Drainable(drainCtx, pod)
 		switch status.Outcome {
 		case drainability.UndefinedOutcome, drainability.DrainOk:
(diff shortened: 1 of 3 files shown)


x13n:
/assign

Isn't it sufficient to specify `--skip-nodes-with-local-storage=false`? This should have the same effect.

wackxu:
> /assign
> 
> Isn't it sufficient to specify `--skip-nodes-with-local-storage=false`? This should have the same effect.


I've tried it but it doesn't work, can you give me some advice?

x13n:
What reason for not scaling down do you see? There should be a log line matching `node .* cannot be removed:`

PR #6419, opened 2024-01-04
list vpacheckpoint from local cache

yunwang0911:
Use vpacheckpoint lister instead of vpacheckpoint client when vpa recommender lists vpacheckpoint.

As we all know, query objects from kube-apiservice may increase the stress to kube-apiservice. What's more, vpa recommender might be limited when it lists vpacheckpoint by client. This pr changes the vpa recommender to use lister instead of client.

No

No. 
```release-note
changes the vpa recommender to use lister instead of client to list vpacheckpoint.
```

### DIFF
--- a/vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
+++ b/vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
@@ -80,6 +80,7 @@ type ClusterStateFeederFactory struct {
 	KubeClient          kube_client.Interface
 	MetricsClient       metrics.MetricsClient
 	VpaCheckpointClient vpa_api.VerticalPodAutoscalerCheckpointsGetter
+	VpaCheckpointLister vpa_lister.VerticalPodAutoscalerCheckpointLister
 	VpaLister           vpa_lister.VerticalPodAutoscalerLister
 	PodLister           v1lister.PodLister
 	OOMObserver         oom.Observer
@@ -96,6 +97,7 @@ func (m ClusterStateFeederFactory) Make() *clusterStateFeeder {
 		metricsClient:       m.MetricsClient,
 		oomChan:             m.OOMObserver.GetObservedOomsChannel(),
 		vpaCheckpointClient: m.VpaCheckpointClient,
+		vpaCheckpointLister: m.VpaCheckpointLister,
 		vpaLister:           m.VpaLister,
 		clusterState:        m.ClusterState,
 		specClient:          spec.NewSpecClient(m.PodLister),
@@ -186,6 +188,7 @@ type clusterStateFeeder struct {
 	metricsClient       metrics.MetricsClient
 	oomChan             <-chan oom.OomInfo
 	vpaCheckpointClient vpa_api.VerticalPodAutoscalerCheckpointsGetter
+	vpaCheckpointLister vpa_lister.VerticalPodAutoscalerCheckpointLister
 	vpaLister           vpa_lister.VerticalPodAutoscalerLister
 	clusterState        *model.ClusterState
 	selectorFetcher     target.VpaTargetSelectorFetcher
@@ -252,18 +255,16 @@ func (feeder *clusterStateFeeder) InitFromCheckpoints() {
 
 	for namespace := range namespaces {
 		klog.V(3).Infof("Fetching checkpoints from namespace %s", namespace)
-		checkpointList, err := feeder.vpaCheckpointClient.VerticalPodAutoscalerCheckpoints(namespace).List(context.TODO(), metav1.ListOptions{})
+		checkpointList, err := feeder.vpaCheckpointLister.VerticalPodAutoscalerCheckpoints(namespace).List(labels.Everything())
 		if err != nil {
 			klog.Errorf("Cannot list VPA checkpoints from namespace %v. Reason: %+v", namespace, err)
 		}
-		for _, checkpoint := range checkpointList.Items {
-
+		for _, checkpoint := range checkpointList {
 			klog.V(3).Infof("Loading VPA %s/%s checkpoint for %s", checkpoint.ObjectMeta.Namespace, checkpoint.Spec.VPAObjectName, checkpoint.Spec.ContainerName)
-			err = feeder.setVpaCheckpoint(&checkpoint)
+			err = feeder.setVpaCheckpoint(checkpoint)
 			if err != nil {
 				klog.Errorf("Error while loading checkpoint. Reason: %+v", err)
 			}
-
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


vertical-pod-autoscaler/pkg/utils/vpa/api.go
```diff
@@ -94,6 +94,26 @@ func NewVpasLister(vpaClient *vpa_clientset.Clientset, stopChannel <-chan struct
 	return vpaLister
 }
 
+// NewVpaCheckpointLister returns VerticalPodAutoscalerCheckpointLister configured to fetch all VPACheckpoint objects from namespace,
+// set namespace to k8sapiv1.NamespaceAll to select all namespaces.
+// The method blocks until vpaCheckpointLister is initially populated.
+func NewVpaCheckpointLister(vpaClient *vpa_clientset.Clientset, stopChannel <-chan struct{}, namespace string) vpa_lister.VerticalPodAutoscalerCheckpointLister {
```
alvaroaleman, line 100: Doesn't this require changes to the manifests to allow watching these?

yunwang0911: Do you mean the clusterrole/role should be improved?

alvaroaleman: Yes

yunwang0911: No need to update rbac. The service account already was bound the permission to list and watch vpacheckpoint.
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: system:vpa-checkpoint-actor
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: system:vpa-checkpoint-actor
subjects:
  - kind: ServiceAccount
    name: vpa-recommender
    namespace: kube-system
```
https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/deploy/vpa-rbac.yaml

adrianmoisey:
@yunwang0911 are you willing to update this with master? I have capacity to review it

yunwang0911:
@adrianmoisey Rebased

vertical-pod-autoscaler/pkg/utils/vpa/api.go
```diff
@@ -96,6 +96,26 @@ func NewVpasLister(vpaClient *vpa_clientset.Clientset, stopChannel <-chan struct
...
+	vpaListWatch := cache.NewListWatchFromClient(vpaClient.AutoscalingV1().RESTClient(), "verticalpodautoscalercheckpoints", namespace, fields.Everything())
+	indexer, controller := cache.NewIndexerInformer(vpaListWatch,
+		&vpa_types.VerticalPodAutoscalerCheckpoint{},
+		1*time.Hour,
+		&cache.ResourceEventHandlerFuncs{},
+		cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+	vpaCheckpointLister := vpa_lister.NewVerticalPodAutoscalerCheckpointLister(indexer)
+	go controller.Run(stopChannel)
+	if !cache.WaitForCacheSync(make(chan struct{}), controller.HasSynced) {
+		klog.Fatalf("Failed to sync VPA checkpoint cache during initialization")
```
adrianmoisey, line 112: I see that this is `Fatalf`, while the `NewVpasLister` uses `ErrorS`. At the time you made this PR, `NewVpasLister` used `Fatalf`, but it was changed in https://github.com/kubernetes/autoscaler/pull/7294

I just want to conform with @omerap12 if this change was intended. However, I assume we want `Fatalf` here.

omerap12: Thanks for pinging me. I’ve added a comment.

omerap12, line 112: Please use `ErrorS` + `os.Exit(255)` instead of `Fatalf`

vertical-pod-autoscaler/pkg/utils/vpa/api.go
```diff
@@ -96,6 +96,26 @@ func NewVpasLister(vpaClient *vpa_clientset.Clientset, stopChannel <-chan struct
...
+	} else {
+		klog.Info("Initial VPA checkpoint synced successfully")
```
omerap12, line 114: Please use `InfoS`

omerap12 (changes requested):

vertical-pod-autoscaler/pkg/utils/vpa/api.go
yunwang0911: updated

yunwang0911: updated

yunwang0911:
@voelzmo @adrianmoisey @kwiesmueller Could you please approve this PR?

vertical-pod-autoscaler/pkg/utils/vpa/api.go
```diff
@@ -30,14 +31,13 @@ import (
...
-	"k8s.io/client-go/tools/cache"
-	"k8s.io/klog/v2"
-
 	vpa_types "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1"
 	vpa_clientset "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/client/clientset/versioned"
 	vpa_api "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/client/clientset/versioned/typed/autoscaling.k8s.io/v1"
 	vpa_lister "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/client/listers/autoscaling.k8s.io/v1"
 	controllerfetcher "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/target/controller_fetcher"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
```
voelzmo, line 40: I don't think we want to re-order these imports to the lower section, please leave them as-is

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
```diff
@@ -284,11 +285,11 @@ func (feeder *clusterStateFeeder) GarbageCollectCheckpoints() {
 
 	for _, namespaceItem := range namespaceList.Items {
 		namespace := namespaceItem.Name
-		checkpointList, err := feeder.vpaCheckpointClient.VerticalPodAutoscalerCheckpoints(namespace).List(context.TODO(), metav1.ListOptions{})
+		checkpointList, err := feeder.vpaCheckpointLister.VerticalPodAutoscalerCheckpoints(namespace).List(labels.Everything())
```
voelzmo, line 288: I think now that we have a `vpaCheckpointLister` inside the `cluster_feeder`, we no longer need the list of namespaces here.
Instead, we can just list _all checkpoints_ (already filtered by `vpaObjectNamespace`, if that is set) above like this
```
checkpointList, err := feeder. vpaCheckpointLister.List(labels.Everything())
```

And then iterate over those checkpoints and filter them, if they are in a namespace that is in `vpaIgnoredNamespaces`. Alternatively, we could do the filtering once on the entire list, similar to what is done in `filterVPAs`

voelzmo (changes requested):
Thanks for the PR. I guess we'll have to do some rebasing between this and https://github.com/kubernetes/autoscaler/pull/7716 though.

vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go
omerap12, line 288: I agree. You can refer to this: [cluster_feeder.go#L330](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/recommender/input/cluster_feeder.go#L330) for an example of how to filter the checkpoints based on the desired namespaces.

omerap12:
Hey @yunwang0911 , do you still work on that?

adrianmoisey:
@yunwang0911 Just a friendly ping. Do you want to keep working on this or can someone take it over?
It seems like a change we should get landed.

adrianmoisey:
@raywainman could you add a 1.5.0 milestone? I'm happy to pick this work up and get it done in 1.5.0

adrianmoisey:
Last ping, @yuanwang04 you still keen to push this through?

Issue #6232, opened 2023-10-30
Exclude some namespaces from VPA webhook


adrianmoisey:
**Which component are you using?**:

vertical-pod-autoscaler - admission-controller

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:
The source of this feature request is https://cloud.google.com/kubernetes-engine/docs/how-to/optimize-webhooks#unsafe-webhooks

GKE displays a warning for webhooks that intercept resources in the kube-system namespace. 

This warning seems sane to me.

**Describe the solution you'd like.**:

Ability to ignore some namespaces for the vertical-pod-autoscaler admission-controller webhook.

**Describe any alternative solutions you've considered.**:

N/A

**Additional context.**:

N/A

voelzmo:
will be fixed by: #6428

PR #6428, opened 2024-01-08
[GCE] fix unsafe webhook vpa-webhook-config

britdm:
#### What type of PR is this?

/kind feature

This PR cleans up the unsafe webhook `vpa-webhook-config` deployed on GCE. The namespaces `kube-system` and `kube-node-lease` are considered system-managed. See [unsafe webhooks](https://cloud.google.com/kubernetes-engine/docs/how-to/optimize-webhooks?&_ga=2.18119354.-2077099904.1664205505#unsafe-webhooks).

Fixes https://github.com/cowboysysop/charts/issues/578

To keep the namespaceSelector optionally empty, the variable definition can be wrapped in an if statement that verifies that the provider is `gce`.


sophieliu15:
@kwiesmueller Could you take a look to decide whether to approve or not?

vertical-pod-autoscaler/pkg/admission-controller/config.go
```diff
@@ -65,6 +65,15 @@ func selfRegistration(clientset *kubernetes.Clientset, caCert []byte, namespace,
 	sideEffects := admissionregistration.SideEffectClassNone
 	failurePolicy := admissionregistration.Ignore
 	RegisterClientConfig.CABundle = caCert
+	namespaceSelector := metav1.LabelSelector{
```
alvaroaleman, line 68: Does this really matter, considering it runs with `failurePolicy: ignore`?

voelzmo: Even with `failurePolicy: ignore`, GKE will still create an error for you and display it in your cluster, so I guess it makes sense to ignore those namespaces. This is a breaking change, though, if for some reason people would auto-scale stuff in `kube-system`, it would stop working for them. Therefore, I think we should only do this conditionally and hide it behind some configuration parameter.

alvaroaleman: Yeah but if GKE raises error when it shouldn't that is a problem with GKE, not with the components that happen to trigger that behavior. I think we should have better reasoning than "Cloudprovider UI displays an error".

kwiesmueller: Agreed with both of you.
It's breaking and should be flag gated.

Not sure if it's about the UI, but agree that if it's really a cloud provider specific UI thing that shouldn't force us to fix it. I thought webhooks against system namespaces are considered "dangerous" either way.
Honestly I'm surprised to see this selfRegistration thing, was not aware that it's something people do. @alvaroaleman do you know if that's common?

I think a flag that lets you specify namespaces that should be ignored could be one option.
I'd like to make the default safer, prevent people from accidentally locking their clusters, but as the failurePolicy is ignore. that should be fine, right?

kwiesmueller (changes requested):

vertical-pod-autoscaler/pkg/admission-controller/config.go
alvaroaleman: > Honestly I'm surprised to see this selfRegistration thing, was not aware that it's something people do. @alvaroaleman do you know if that's common?

Its uncommon for projects to provide this IME. If it is provided, it is probably at least semi-common to use, because webhooks require you to generate certs and not everyone has some other mechanism in their cluster available to do or knows how to do it through helm and friends.

voelzmo: @alvaroaleman 
> I think we should have better reasoning than "Cloudprovider UI displays an error".

Absolutely agreed! I guess in this specific case, it is hard to say "the UI is just wrong, so if you want to get rid of this error, go talk to your cloud provider". As @kwiesmueller pointed out, even with `failurePolicy: ignore`, user deployed webhooks in system namespaces should be avoided. In our internal k8s platform we do a very similar thing and flag these things for users.

@kwiesmueller 
> I think a flag that lets you specify namespaces that should be ignored could be one option.

This is a double-edged feature. It was proposed with #6232 and would allow users to fix this issue themselves. However, at the same time it makes it easier/possible to configure something that leads to an endless eviction loop: If you only configure the webhook to ignore a certain namespace but happen to have objects under VPA control in those namespaces, Updater will permanently evict your workload.
Ideally, we wouldn't make it easy or even possible for people to create these situations and shoot themselves in the foot.

I imagine the best compromise here is probable a namespace denylist, similar to how the VPA components currently have the concept of an _allowlist_ for namespaces they work in (parameter `vpa-object-namespace`).  The configuration parameter is the the same on all 3 components and  you could argue that a user setting this parameter on one component should take care of setting it on the other two components with the same values. Denylist and allowlist should be made exclusive to each other, and hopefully we found a good solution for everyone. WDYT?

alvaroaleman: > failurePolicy: ignore, user deployed webhooks in system namespaces should be avoided

Where did they point that out? And could you elaborate on the rationale behind that?

voelzmo: This is a quote from the [GKE docs](https://cloud.google.com/kubernetes-engine/docs/how-to/optimize-webhooks?&_ga=2.18119354.-2077099904.1664205505#unsafe-webhooks)

> If a webhook is intercepting any resources in system-managed namespaces, or [certain types of resources](https://cloud.google.com/kubernetes-engine/docs/how-to/optimize-webhooks?&_ga=2.18119354.-2077099904.1664205505#how-gke-detect), GKE considers this unsafe and recommends that you update the webhooks to avoid intercepting these resources.

The same is mentioned in the [official k8s docs](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#avoiding-operating-on-the-kube-system-namespace), but maybe with a bit softer wording

> If your admission webhooks don't intend to modify the behavior of the Kubernetes control plane, exclude the kube-system namespace from being intercepted using a [namespaceSelector](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-namespaceselector).

I guess that `failurePolicy: ignore` will help in most cases, [but there are still scenarios where a failing webhook _can_ stop k8s from working properly](https://kubernetes.slack.com/archives/C0EG7JC6T/p1650375336872579), so that's probably why even with this `failurePolicy` webhooks are still flagged as problematic. However, I see that you[ were involved in the discussion around this](https://github.com/kubernetes-sigs/controller-runtime/issues/2662), so you probably have a better understanding of this than I do.

alvaroaleman: So the webhook returning a valid http response that is not parseable or such would be an issue. The other issue I am aware of as it caused an outage for us is if you validate anything related to leader election, because the client-side timeouts there are way below the 30 seconds, so all of these requests can fail before the webhook times out, which can brick a cluster.

However, we are only validating create pod requests, i think that should be fine.

alvaroaleman (approved):

voelzmo:
Hey @britdm thanks for the new change! I think, however, we need to have a paramater on _all three VPA components_, otherwise it can easily happen that users end up in an endless eviction loop: If you have VPA objects in `kube-system`, recommender will provide a recommendation, updater will evict and the admission-controller ignores this namespace and won't adapt the requests.

I tried to sketch this out in an [earlier, longer comment](url)

> I imagine the best compromise here is probable a namespace denylist, similar to how the VPA components currently have the concept of an allowlist for namespaces they work in (parameter vpa-object-namespace). The configuration parameter is the the same on all 3 components and you could argue that a user setting this parameter on one component should take care of setting it on the other two components with the same values. Denylist and allowlist should be made exclusive to each other.

So I think the missing changes here are

- rename parameter to something like `ignored-vpa-object-namespaces`, such that it doesn't reference the webhook and has some resemblence to the already existing `vpa-object-namespace`
- hrow an error, if both parameters are provided. It should only be possible to provide one of `ignored-vpa-object-namespaces` and `vpa-object-namespace`
- move documentation for this parameter to the [top README](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/README.md#examples)
- Bonus: add a test that makes sure this flag works, similar to what we already have [for recommenders with different names](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/e2e/v1/full_vpa.go#L203)

adrianmoisey:
Hey @britdm, are you planning to continue this PR?
If not, do you mind if I try work on it?

britdm:
> Hey @britdm, are you planning to continue this PR? If not, do you mind if I try work on it?

Sure, I have not yet had time to clean up my local copy with changes for this PR, and would appreciate the help to move it forward. One caveat I found with updating the other VPA components is that the conditional statement used to check if both variables are used in the admission controller would need slight modifications to work for the other components.

voelzmo:
/close
In favor of #6788

PR #6460, opened 2024-01-19
[VPA] check OwnerRef against TargetRef to confirm VPA/Pod association

dbenque:
#### What type of PR is this?

/kind bug

The function `GetControllingVPAForPod` tries to match pods with VPA. The problem is that it does it based on label only.

If some pods have label that match the labelSelector that was compute for the VPA the program will consider that the pod is controlled by the VPA. This is not always tru. 

For example an orphan pod (no controller) can match the labelSelector but it should not be associated with any VPA! We had 
that particular case in our infra. Another theoric case is that 2 deployments having same set of labels (or overlapping set) could result in bad assignment pod<=>vpa.

To fix these kind of issue, in this PR, we propose to validate that the function `GetControllingVPAForPod` validates that the pod matches in terms of labels but also in terms of ownerRef<=>targetRef.

The first commit is just code move to expose the ControllerFetecher to other packages.
The second commit adds the missing bits to have the ownerRef<=>targetRef validation.

NONE


dbenque:
@mwielgus 
As discussed during the SIG Meeting here is a reference to some kubernetes code showing that an orphan pod (not having ownerRef) is not considered as part of the deployment even if the labelSelector matches:
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/deployment_controller.go#L553-L581

This function also shows that if the pod matches in terms of label, it is not considered as part as the deployment if the ownerReference chain (Pod=>ReplicaSet=>Deployment) is not correct.

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/matcher_test.go
```diff
@@ -37,7 +40,22 @@ func parseLabelSelector(selector string) labels.Selector {
 }
 
 func TestGetMatchingVpa(t *testing.T) {
```
sophieliu15, line 42: Is there a testing case where VPA does not select a pod when its selector matches but the target ref does not match?

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/matcher_test.go
```diff
@@ -52,7 +70,7 @@ func TestGetMatchingVpa(t *testing.T) {
 			name: "matching selector",
 			pod:  podBuilder.Get(),
 			vpas: []*vpa_types.VerticalPodAutoscaler{
-				vpaBuilder.WithUpdateMode(vpa_types.UpdateModeAuto).WithName("auto-vpa").Get(),
+				vpaBuilder.WithUpdateMode(vpa_types.UpdateModeAuto).WithName("auto-vpa").WithTargetRef(targetRef).Get(),
```
sophieliu15, line 73: Please correct me if I am wrong. I thought the target ref of vpa is usually "Deployment" kind. E.g., https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler#example-vpa-configuration

Are we assigning `ReplicationController` kind here?

dbenque: The test is using a `controllerfetcher.FakeControllerFetcher` to retrieve the owner reference. The implementation of this mock returns the same reference that is passed. So for the test it is important that the reference set in the pod is the same that the one set in the VPA. We are left with 2 options, none reflecting really the reality, and I would say it is not important because here we are not testing the code of the embedded `controllerfetcher`:
Option 1: vpa.targetRef=> Deployment    and pod.ownerRef=> Deployment (which is not possible in reality)
Option 2: pod.ownerRef=> ReplicaSet    and pod.ownerRef=> Replicaset (which is not happening in reality)

In fact there is a 3rd option that we can implement and the would be more correct compare to reality: we use Statefulset. I can amend the PR for that.

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/matcher_test.go
dbenque: The test was missing, I have added a test where selector match but not the ownerRef, here it is:
https://github.com/kubernetes/autoscaler/pull/6460/commits/a5cae3b59f454c6b5a801b6cc76abe65d57ee348

sophieliu15:
/lgtm

cc: @kwiesmueller for final code review and approval.

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/matcher_test.go
kwiesmueller: Just to clarify, the functional code does traverse ownerRefs (pod -> replica-set -> deployment)?
If this test is really not meant to test the owner chain resolution (and if we really have another test making sure that still happens correctly) I'm okay with a surreal test case given it's explained in a comment.

vertical-pod-autoscaler/pkg/utils/vpa/api.go
```diff
@@ -125,11 +127,45 @@ func stronger(a, b *vpa_types.VerticalPodAutoscaler) bool {
...
+
+	var ownerRefrence *meta.OwnerReference
+	for i := range pod.OwnerReferences {
+		r := pod.OwnerReferences[i]
+		if r.Controller != nil && *r.Controller {
+			ownerRefrence = &r
+		}
+	}
+	if ownerRefrence == nil {
+		// If the pod has no ownerReference, it cannot be under a VPA.
```
kwiesmueller, line 140: Is that a behavior change?
If so I think we need API review and make sure we version the change accordingly.

With the prospect of in-place pod updates, @dbenque what do you think about this rule. I'm wondering if in the future it might make sense to have VPAs for single pods to still allow growing them?
Or would we expect people to still have them behind a Deployment with replica=1?

kwiesmueller (commented):
Overall this looks good to me.
I'd prefer that we clarify with API review if this change is okay. I think we'll need to cut this as a behavior change.

kwiesmueller:
@sophieliu15 can you look into the API question? Feel free to loop me in if help is needed.

dbenque:
@kwiesmueller , following [your comment](https://github.com/kubernetes/autoscaler/pull/6460#discussion_r1470285369), I am adding a comment into the test:
[4f9f840](https://github.com/kubernetes/autoscaler/pull/6460/commits/4f9f840018686e4444b9812e9871fb79b6f15142)

vertical-pod-autoscaler/pkg/utils/vpa/api.go
dbenque: > Is that a behavior change?

It changes the behaviour like any bug fix would do. Today a VPA can adopt orphan pods or even pods of another deployment if ever the labelSelectors overlap. This PR will prevent such "unexpected" behaviours.

> I'm wondering if in the future ...

When it comes to VPA (in-place update or not), a pod (or a set of pods) is expected to be behind a Deployment or any other controller (Statefulset, DaemonSet or exposing Scale subresource). This is a direct consequence of or VPA definition: [it uses targetRef](https://github.com/kubernetes/autoscaler/blob/7e95c7e63a5f0773e51bc484b1f7dbe802fe668a/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1/types.go#L76-L87)

A single pod has to be behind a Deployment with replica=1 or any other custom controller implementing scale if we want to be able to apply VPA recommendation to it.

vertical-pod-autoscaler/pkg/admission-controller/resource/vpa/matcher_test.go
dbenque: https://github.com/kubernetes/autoscaler/pull/6460/commits/4f9f840018686e4444b9812e9871fb79b6f15142

vertical-pod-autoscaler/pkg/utils/vpa/api.go
sophieliu15: Re API review.

Not aware of any API review process for this sig. TBH, I believe many PRs change behaviours of VPA. @voelzmo and @mwielgus is there any API review process for this sig? Do you think a change like this PR requires a review?

sophieliu15: Actually this PR was already discussed in the sig meeting. @dbenque is there any remaining concerns from the sig meeting that needs to be addressed? If not it is good to me.

kwiesmueller (approved):
/lgtm
/approve

dbenque:
Complete with fix:
https://github.com/kubernetes/autoscaler/pull/6724

PR #6462, opened 2024-01-22
Small logging enchancement for easier debugging

kawych:
Better logging

```release-note
NONE
```

### DIFF
--- a/cluster-autoscaler/core/scaledown/actuation/actuator.go
+++ b/cluster-autoscaler/core/scaledown/actuation/actuator.go
@@ -98,7 +98,7 @@ func (a *Actuator) ClearResultsNotNewerThan(t time.Time) {
 
 // StartDeletion triggers a new deletion process.
 func (a *Actuator) StartDeletion(empty, drain []*apiv1.Node) (*status.ScaleDownStatus, errors.AutoscalerError) {
-	a.nodeDeletionScheduler.ReportMetrics()
+	a.nodeDeletionScheduler.ResetAndReportMetrics()
 	deletionStartTime := time.Now()
 	defer func() { metrics.UpdateDuration(metrics.ScaleDownNodeDeletion, time.Since(deletionStartTime)) }()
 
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -160,8 +160,9 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 	// Finalize binpacking limiter.
 	o.processors.BinpackingLimiter.FinalizeBinpacking(o.autoscalingContext, options)
 
+	
+	klog.V(1).Infof("%v expansion options", len(options))
```
vadasambar, line 164: How important is being able to see this? V(1) means we would have a log emitted every iteration of the scaling loop (every 10 seconds)

kawych: Anything wrong with that? It's just one line per iteration, we already have some lines like that: "best option to resize", "Estimated %d nodes needed", "Final scale-up plan". Although I think I would be OK with logging it at V(2) or adding this number to some of the other lines, e.g. "Best option to resize out of %v".

The exact case I have in mind is debugging a situation where Cluster Autoscaler made a different decision than I expected. There are actually no logs to show whether the expected node group was even considered for a scale-up.

vadasambar: If debugging is the only reason, maybe having 2 logs resolve the problem. Something like this:
```go
// default level is 2, so the level can be 3 or 4 
if len(options) != 0 {
  klog.V(3).Infof("%v expansion options", len(options))
}
if len(options==0) {
  // retain the old log as is
  klog.V(1).Info("No expansion options")
}
````

x13n:
@kawych Are you working on this one? Closing as it seems abandoned (and content doesn't match description anymore). Please reopen if still relevant.

/close

PR #6472, opened 2024-01-26
chore: adding telemetry to monitor run once loop

Bryce-Soghigian:
#### What type of PR is this?

/kind feature 

We would like to be able to alert and monitor if a given cluster is exceeding its run once loop scan interval. 

```release-note
Adding Metric cluster_autoscaler_runonce_duration_seconds to monitor the length of the run once loop
```

```docs

```


x13n:
How is this different from existing `function_duration_seconds` with label `main`?

https://github.com/kubernetes/autoscaler/blob/00fbbe19a7045536a2d967902eb132c37f5a9f8e/cluster-autoscaler/main.go#L596

x13n:
There's an existing metric for exactly the same thing, let's avoid duplication.

/close

PR #6473, opened 2024-01-28
Introduce GceInstance that extends cloudprovider.Instance with NumericId

scotch-bonnet:
#### What type of PR is this?
/kind feature

Allows access to numeric instance id in GCE Cloud Provider implementation.

NONE

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
+++ b/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
@@ -88,6 +88,12 @@ var (
 	}
 )
 
+// GceInstance extends cloudprovider.Instance with GCE specific numeric id.
+type GceInstance struct {
+	cloudprovider.Instance
+	NumericId uint64
+}
+
 // AutoscalingGceClient is used for communicating with GCE API.
 type AutoscalingGceClient interface {
 	// reading resources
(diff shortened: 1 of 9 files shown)


cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -88,6 +88,13 @@ var (
 	}
 )
 
+// GceInstance extends cloudprovider.Instance with GCE specific numeric id.
+type GceInstance struct {
+	Id        string
+	Status    *cloudprovider.InstanceStatus
+	NumericId uint64
+}
```
BigDarkClown, line 96:
```suggestion
type GceInstance struct {
        cloudprovider.Instance
	NumericId uint64
}
```

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client_test.go
```diff
@@ -390,7 +398,7 @@ func TestFetchMigInstancesInstanceUrlHandling(t *testing.T) {
 					},
 				},
 			},
-			wantInstances: []cloudprovider.Instance{
+			wantInstances: []GceInstance{
```
BigDarkClown, line 401: nit: Please add NumericId to all test cases.

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
BigDarkClown: This will both guard us from diverging from `cloudprovider.Instance` and simplify conversion from `GceInstance` to `cloudprovider.Instance`.

BigDarkClown (commented):
/lgtm
/approve

PR #6477, opened 2024-01-29
Consider atomic nodes

kushagra98:
#### What type of PR is this?

/kind bug
/kind regression

`unneededNodesLimit()` stops considering scale down nodes when this limit is reached.  However, if a nodepool which has nodes > this limit and only allowed for atomic scaledown will always starve.
Eg. Consider a nodepool with 1000 nodes in a cluster and `unneededNodesLimit()` = 500. Let's consider that this nodepool is only allowed to scale down atomically. The nodepool will always starve for scale down as the limit is 500 but we cannot scaledown 500 nodes, we can only scale down 1000 nodes.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/scaledown/planner/planner.go
+++ b/cluster-autoscaler/core/scaledown/planner/planner.go
@@ -23,6 +23,7 @@ import (
 
 	apiv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
 	"k8s.io/autoscaler/cluster-autoscaler/context"
 	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown"
 	"k8s.io/autoscaler/cluster-autoscaler/core/scaledown/eligibility"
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/core/scaledown/planner/planner.go
```diff
@@ -274,7 +276,7 @@ func (p *Planner) categorizeNodes(podDestinations map[string]bool, scaleDownCand
 			klog.Warningf("%d out of %d nodes skipped in scale down simulation due to timeout.", len(currentlyUnneededNodeNames)-i, len(currentlyUnneededNodeNames))
 			break
 		}
-		if len(removableList) >= p.unneededNodesLimit() {
+		if len(removableList)-atomicScaleDownNodesCount >= p.unneededNodesLimit() {
 			klog.V(4).Infof("%d out of %d nodes skipped in scale down simulation: there are already %d unneeded nodes so no point in looking for more.", len(currentlyUnneededNodeNames)-i, len(currentlyUnneededNodeNames), len(removableList))
```
x13n, line 280: Let's include `atomicScaleDownNodesCount` in the log.

x13n (commented):
Looks good, just one minor comment - let's address before merging.

/approve

cluster-autoscaler/core/scaledown/planner/planner.go
x13n: I meant in this line: just report `atomicScaleDownNodesCount` value. Doing it for every considered node is going to result in a lot of log spam.

kushagra98: Makes sense. Done.

PR #6482, opened 2024-01-30
Fixed bug with missing taint translation layer between AWS EKS format and Kubernetes format

### LINKED ISSUE #6481: Taints on EKS Managed Node Groups scaled to zero not read correctly, causing scale-up of unsuitable nodes

cluster-autoscaler

1.27.5, but reproduced out-of-cluster off `main` branch (as of 28. January 2024)

Component version:

$ kubectl version

Client Version: v1.29.0
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.28.4-eks-8cb36c9
</pre></details>

AWS EKS using multiple Managed Node Groups. Some of these are used for batch jobs, and will scale to zero during idle periods.

It's also quite acceptable to us for the pods to be pending until previous pods have completed, if/when we reach the node group's maximum size.

When EKS Managed Node Groups are scaled to zero, so there's no Kubernetes node to read taints from, taints should be read correctly from the Managed Node Group.

This avoids scaling up nodes, pending pods cannot be scheduled on anyway.

This should be supported as of this commit: [[AWS EKS - Scale-to-0] Add Managed Nodegroup Cache](https://github.com/kubernetes/autoscaler/commit/b4cadfb4e25b6660c41dbe2b73e66e9a2f3a2cc9).

In Cluster Autoscaler log, we should see "predicate checking error: node(s) had untolerated taint" for each node group with untolerated taints.

**What happened instead?**:

In Cluster Autoscaler log, we see no "predicate checking error: node(s) had untolerated taint" (for node groups with untolerated taints), but we do see "predicate checking error: node(s) didn't match Pod's node affinity/selector", so node labels are read correctly.

When evalutating node groups to scale-up, and there are no running nodes to query, taints are read from the [Auto-Sscaling Group](https://github.com/kubernetes/autoscaler/blob/ed6ebbe8baacab2fe1befd468ffaa049c66b0d73/cluster-autoscaler/cloudprovider/aws/aws_manager.go#L300), and [Managed Node Group](https://github.com/kubernetes/autoscaler/blob/ed6ebbe8baacab2fe1befd468ffaa049c66b0d73/cluster-autoscaler/cloudprovider/aws/aws_manager.go#L314-L320) (if applicable).

Kubernetes operate with the taint effects "NoSchedule", "PreferNoSchedule" and "NoExecute" (as per [Kubernetes' NodeSpec documentation](https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/node-v1/#NodeSpec)).

However, the [`DescribeNodegroup` call](https://github.com/kubernetes/autoscaler/blob/ed6ebbe8baacab2fe1befd468ffaa049c66b0d73/cluster-autoscaler/cloudprovider/aws/aws_wrapper.go#L66) returns taint effects as "NO_SCHEDULE", "NO_EXECUTE" and "PREFER_NO_SCHEDULE" (as per [AWS' EKS API documentation](https://docs.aws.amazon.com/eks/latest/APIReference/API_Taint.html))

This is also evident when describing MNG's with the AWS CLI, for example:

```bash
aws eks describe-nodegroup --cluster-name $cluster --nodegroup-name $ng --query "nodegroup.taints" --output json
```

```json
[
    {
        "key": "veo.co/nodegroup-instance-spot",
        "value": "true",
        "effect": "NO_SCHEDULE"
    }
]
```

When we propagate the taints as tags on *Auto-Scaling Groups*, Cluster Autoscaler correctly reads the taints, and does not scale-up nodegroups with untolerated taints.

But in the debug log we then see this gem - two distinct taints, with the same key, but effect in different format (one from MNG *taint*, one from ASG *tag*), only one of which is correct and works as intended (the one from the ASG):

```text
debugInfo=taints on node:
v1.Taint{Key:"veo.co/nodegroup-instance-spot", Value:"true", Effect:"NO_SCHEDULE", TimeAdded:<nil>}, <-- from MNG taint
v1.Taint{Key:"veo.co/nodegroup-instance-spot", Value:"true", Effect:"NoSchedule", TimeAdded:<nil>},  <-- from ASG tag
```

This is particularly problematic for us when we reach the maximum size of the node groups our pods *should* (eventually) be scheduled on. In this case we saw Cluster Autoscaler scale-up seemingly random node groups that *matches* node selector, but with *untolerated* taints. As the pods kept being pending, this happens repeatedly.

Preparation:

1. Create two new EKS Managed Node Groups (see script below)
   1. Set a node label we'll target later (e.g. `veo.co/test=true`)
   2. Set maximum size of both to 1
   3. Add a taint to one node group that we'll *not* tolerate (e.g. `veo.co/active=false`)
   4. Instance type `t3.medium` should do
2. Set Cluster Autoscaler logging level to 5 (`-v==5`)
3. Add a dummy `deployment` (see manifest below)
   1. Target the node label from above (`veo.co/test=true`)
   2. Do *not* tolerate any taints
   3. Set memory request so we don't have to scale to many replicas (e.g. `1Gi`)

This assumes there's already a suitable IAM role for the EKS node group. If not create one as described in [Creating the Amazon EKS node IAM role](https://docs.aws.amazon.com/eks/latest/userguide/create-node-role.html#create-worker-node-role).

Observe *undesirable* behaviour:

1. Scale the deployment to 2 replicas which need to span two nodes
2. Wait 30 seconds, grep Cluster Autoscaler logs and observe:
   1. There are no "predicate checking error: node(s) had untolerated taint" entries (at least before the group has scaled up once, and its actual taints are cached)
   2. Node group `test2` will scale up, though its taint is not tolerated
   3. Cluster Autoscaler thinks the pod can be moved the upcoming node

```text
$ kubectl logs -n kube-system -l app=cluster-autoscaler --since=5m --tail=-1 | grep "test2"

orchestrator.go:310] Final scale-up plan: [{eks-test2-SOME_GUID 0->1 (max: 1)}]
...
klogx.go:87] Pod default/http-https-echo-6d9c4b769f-rmfb4 can be moved to template-node-for-eks-test2-SOME_GUID-upcoming-0
```

Observe *desirable* behaviour:

1. Scale deployment and test node groups to zero
2. Propagate node group taint to ASG - which shouldn't be necessary (see script below)
3. Restart Cluster Autoscaler
4. Scale the deployment to 2 replicas which need to span two nodes
5. Wait 30 seconds, grep Cluster Autoscaler logs and observe:
    1. The `test2` node group is recognised as having untolerated taints
    2. `debugInfo` reveals taints are read from both MNG and ASG, albeit only has the correct format of "effect"
    3. The correct node group, `test1`, is scaled up
    4. As `test1` reaches its max size, options are exhausted and Cluster Autoscaler does not spin up some other node group, with untolerated taints

```text
$ kubectl logs -n kube-system -l app=cluster-autoscaler --since=5m --tail=-1 | grep "test"

orchestrator.go:466] Pod http-https-echo-6d9c4b769f-rz65v can't be scheduled on eks-test2-SOME_GUID, predicate checking error: node(s) had untolerated taint {veo.co/active: false}; predicateName=TaintToleration; reasons: node(s) had untolerated taint {veo.co/active: false}; debugInfo=taints on node: []v1.Taint{v1.Taint{Key:"veo.co/active", Value:"false", Effect:"NoSchedule", TimeAdded:<nil>}, v1.Taint{Key:"veo.co/active", Value:"false", Effect:"NO_SCHEDULE", TimeAdded:<nil>}}
...
orchestrator.go:310] Final scale-up plan: [{eks-test1-SOME_GUID 0->1 (max: 1)}]
...
klogx.go:87] Pod default/http-https-echo-6d9c4b769f-6bbb8 can be moved to template-node-for-eks-test1-SOME_GUID-upcoming-0
...
orchestrator.go:142] Skipping node group eks-test1-SOME_GUID - max size reached
```

Create EKS Managed Node Groups:

```bash
cluster=<CLUSTER_NAME>
role=<ROLE_ARN>
subnets=subnet-xxxxxxxxxxxxxxxxx

aws eks create-nodegroup --cluster-name $cluster --nodegroup-name test1 --node-role $role --subnets $subnets \
    --scaling-config minSize=0,desiredSize=0,maxSize=1 \
    --labels "veo.co/test=true"

aws eks create-nodegroup --cluster-name $cluster --nodegroup-name test2 --node-role $role --subnets $subnets \
    --scaling-config minSize=0,desiredSize=0,maxSize=1 \
    --labels "veo.co/test=true" --taints "key=veo.co/active,value=false,effect=NO_SCHEDULE"
```

Dummy `deployment`:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: http-https-echo
  name: http-https-echo
  namespace: default
spec:
  replicas: 0
  selector:
    matchLabels:
      app: http-https-echo
  template:
    metadata:
      labels:
        app: http-https-echo
    spec:
      nodeSelector:
        veo.co/test: "true"
      containers:
        - image: mendhak/http-https-echo
          name: http-https-echo
          resources:
            requests:
              memory: 2Gi
```

Set taints as Auto-Scaling Group tags:

```bash
asg=$(aws autoscaling describe-auto-scaling-groups --filters "Name=tag:eks:cluster-name,Values=$cluster" --filters "Name=tag:eks:nodegroup-name,Values=test2" --query 'AutoScalingGroups[].AutoScalingGroupName' --output text)
aws autoscaling create-or-update-tags \
    --tags ResourceId=$asg,ResourceType=auto-scaling-group,Key=k8s.io/cluster-autoscaler/node-template/taint/veo.co/active,Value=false:NoSchedule,PropagateAtLaunch=false
```

Similar issues:

- [Autoscaler not respecting the taint tag in AWS (#2434)](https://github.com/kubernetes/autoscaler/issues/2434)
- [cluster autoscaler doesn't apply eks-managed-ng's taint (#5902)](https://github.com/kubernetes/autoscaler/issues/5902)

Huge shout-out to @wcarlsen for invaluable help in finding the culprit! He's also opened PR #6482 to translate EKS taint effects to Kubernetes taint effects, which in our tests resolves this issue.

wcarlsen:
#### What type of PR is this?

/kind bug

This PR fixes a bug between AWS EKS API understanding of Kubernetes taints and normal Kubernetes understanding of taints by introducing a translation method. Since the AWS API response when calling DescribeNodegroup have taints in the format of NO_SCHEDULE instead of NoSchedule as an example.

Fixes #6481

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
@@ -118,10 +118,14 @@ func (m *awsWrapper) getManagedNodegroupInfo(nodegroupName string, clusterName s
 		taintList := r.Nodegroup.Taints
 		for _, taint := range taintList {
 			if taint != nil && taint.Effect != nil && taint.Key != nil && taint.Value != nil {
+				formattedEffect, err := taintEksTranslator(taint)
+				if err != nil {
+					return nil, labels, tags, err
+				}
 				taints = append(taints, apiv1.Taint{
 					Key:    *taint.Key,
 					Value:  *taint.Value,
-					Effect: apiv1.TaintEffect(*taint.Effect),
+					Effect: apiv1.TaintEffect(formattedEffect),
 				})
 			}
 		}
@@ -778,5 +782,19 @@ func buildLaunchTemplateFromSpec(ltSpec *autoscaling.LaunchTemplateSpecification
 	return &launchTemplate{
 		name:    aws.StringValue(ltSpec.LaunchTemplateName),
 		version: version,
+	}
+}
+
+func taintEksTranslator(t *eks.Taint) (apiv1.TaintEffect, error) {
+	// Translation between AWS EKS and Kubernetes taints
+	switch effect := *t.Effect; effect {
+	case eks.TaintEffectNoSchedule:
+		return apiv1.TaintEffectNoSchedule, nil
+	case eks.TaintEffectNoExecute:
+		return apiv1.TaintEffectNoExecute, nil
+	case eks.TaintEffectPreferNoSchedule:
+		return apiv1.TaintEffectPreferNoSchedule, nil
+	default:
+		return "", fmt.Errorf("couldn't translate EKS DescribeNodegroup response taint %s into Kubernetes format", effect)
 	}
 }
--- a/cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
@@ -787,6 +787,10 @@ func buildLaunchTemplateFromSpec(ltSpec *autoscaling.LaunchTemplateSpecification
 
 func taintEksTranslator(t *eks.Taint) (apiv1.TaintEffect, error) {
 	// Translation between AWS EKS and Kubernetes taints
+	//
+	// See:
+	//
+	// https://docs.aws.amazon.com/eks/latest/APIReference/API_Taint.html
 	switch effect := *t.Effect; effect {
 	case eks.TaintEffectNoSchedule:
 		return apiv1.TaintEffectNoSchedule, nil
--- a/cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
@@ -120,7 +120,7 @@ func (m *awsWrapper) getManagedNodegroupInfo(nodegroupName string, clusterName s
 			if taint != nil && taint.Effect != nil && taint.Key != nil && taint.Value != nil {
 				formattedEffect, err := taintEksTranslator(taint)
 				if err != nil {
-					return nil, labels, tags, err
+					return nil, nil, nil, err
 				}
 				taints = append(taints, apiv1.Taint{
 					Key:    *taint.Key,
--- a/cluster-autoscaler/cloudprovider/aws/aws_wrapper_test.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_wrapper_test.go
@@ -114,7 +114,8 @@ func TestGetManagedNodegroup(t *testing.T) {
 	nodegroupName := "testNodegroup"
 	clusterName := "testCluster"
 
-	taintEffect1 := "effect 1"
+	taintEffect1 := eks.TaintEffectNoSchedule
+	taintEffectTranslated1 := apiv1.TaintEffectNoSchedule
 	taintKey1 := "key 1"
 	taintValue1 := "value 1"
 	taint1 := eks.Taint{
@@ -123,7 +124,8 @@ func TestGetManagedNodegroup(t *testing.T) {
 		Value:  &taintValue1,
 	}
 
-	taintEffect2 := "effect 2"
+	taintEffect2 := eks.TaintEffectNoExecute
+	taintEffectTranslated2 := apiv1.TaintEffectNoExecute
 	taintKey2 := "key 2"
 	taintValue2 := "value 2"
 	taint2 := eks.Taint{
@@ -163,10 +165,10 @@ func TestGetManagedNodegroup(t *testing.T) {
 	taintList, labelMap, tagMap, err := awsWrapper.getManagedNodegroupInfo(nodegroupName, clusterName)
 	assert.Nil(t, err)
 	assert.Equal(t, len(taintList), 2)
-	assert.Equal(t, taintList[0].Effect, apiv1.TaintEffect(taintEffect1))
+	assert.Equal(t, taintList[0].Effect, taintEffectTranslated1)
 	assert.Equal(t, taintList[0].Key, taintKey1)
 	assert.Equal(t, taintList[0].Value, taintValue1)
-	assert.Equal(t, taintList[1].Effect, apiv1.TaintEffect(taintEffect2))
+	assert.Equal(t, taintList[1].Effect, taintEffectTranslated2)
 	assert.Equal(t, taintList[1].Key, taintKey2)
 	assert.Equal(t, taintList[1].Value, taintValue2)
 	assert.Equal(t, len(labelMap), 7)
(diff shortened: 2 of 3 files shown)


x13n:
/assign @gjtempleton @drmorr0 

Can (one of) you take a look?

wcarlsen:
So anything we can do to get the ball rolling on this one?

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
```diff
@@ -780,3 +784,17 @@ func buildLaunchTemplateFromSpec(ltSpec *autoscaling.LaunchTemplateSpecification
 		version: version,
 	}
 }
+
+func taintEksTranslator(t *eks.Taint) (apiv1.TaintEffect, error) {
+	// Translation between AWS EKS and Kubernetes taints
```
drmorr0, line 789: Can you add a link to the AWS docs here?

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
```diff
@@ -118,10 +118,14 @@ func (m *awsWrapper) getManagedNodegroupInfo(nodegroupName string, clusterName s
 		taintList := r.Nodegroup.Taints
 		for _, taint := range taintList {
 			if taint != nil && taint.Effect != nil && taint.Key != nil && taint.Value != nil {
+				formattedEffect, err := taintEksTranslator(taint)
+				if err != nil {
+					return nil, labels, tags, err
```
drmorr0, line 123: If there's an error, I generally feel like everything else should be nil -- prevents partial data bugs in the future.

drmorr0 (changes requested):
Just a couple minor comments, will approve when those are fixed.

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
abstrask: Here https://docs.aws.amazon.com/eks/latest/APIReference/API_Taint.html, @wcarlsen

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
wcarlsen: A valid and good point. Thanks

cluster-autoscaler/cloudprovider/aws/aws_wrapper.go
wcarlsen: Very good idea. Thanks

abstrask:
> /approved

Should this be `/approve`?

> Approvers can indicate their approval by writing /approve in a comment

drmorr0:
d'oh.  thank you.

abstrask:
The fix is included in [release 1.31.0](https://github.com/kubernetes/autoscaler/releases/tag/cluster-autoscaler-1.31.0) 🥳

PR #6488, opened 2024-01-31
Add ProvisioningRequestProcessor

yaroslava-serdiuk:
#### What type of PR is this?
/kind feature

ProvisioningRequestProcessor reconciles ProvisioningRequest state/conditions. It checks whether BookingExpired condition should be added or ProvisioningRequest was pending too long.
Also removed CapacityFound condition and use uniform Provisioned condition instead.

Part of implementation https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/proposals/provisioning-request.md


cluster-autoscaler/provisioningrequest/checkcapacity/processor.go
```diff
@@ -0,0 +1,79 @@
...
+
+// Process iterates over ProvisioningRequests and apply:
+// -BookingExpired condition for Provisioned ProvisioningRequest if capacity reservation time is expired.
+// -Failed condition for ProvisioningRequest that were not provisioned during defaultExpirationTime.
+// TODO(yaroslava): fetch reservation and expiration time from ProvisioningRequest
+func (p *checkCapacityProcessor) Process() error {
+	provReqs, err := p.client.ProvisioningRequests()
+	if err != nil {
+		return fmt.Errorf("Couldn't fetch ProvisioningRequests in the cluster: %v", err)
+	}
```
kisieland, line 57: Shouldn't we cap the number of ProvREqs we handle in one loop to not spend a lot of time in this refresh?
Can we start with 10-20 ProvReqs?

cluster-autoscaler/processors/processors.go
```diff
@@ -73,7 +74,8 @@ type AutoscalingProcessors struct {
 	// * scale-downs per nodegroup
 	// * scale-up failures per nodegroup
 	// * scale-down failures per nodegroup
-	ScaleStateNotifier *nodegroupchange.NodeGroupChangeObserversList
+	ScaleStateNotifier           *nodegroupchange.NodeGroupChangeObserversList
+	ProvisioningRequestProcessor provreq.ProvisoningRequestProcessor
```
kisieland, line 78: QQ: how will we add the processor for the atomic for provreq?

cluster-autoscaler/provisioningrequest/checkcapacity/processor.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &checkCapacityProcessor{client: client, now: time.Now}, nil
+}
+
+// Process iterates over ProvisioningRequests and apply:
+// -BookingExpired condition for Provisioned ProvisioningRequest if capacity reservation time is expired.
+// -Failed condition for ProvisioningRequest that were not provisioned during defaultExpirationTime.
+// TODO(yaroslava): fetch reservation and expiration time from ProvisioningRequest
```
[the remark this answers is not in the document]
yaroslava-serdiuk: We would like to have configurable time for booking capacity and probably for how long CA should try to fit the ProvReq. So when we have that this function should be updated to read the values from ProvReq and fallback to default one if no values are specified.

cluster-autoscaler/processors/processors.go
yaroslava-serdiuk: When we have >1 ProvisioningClass we will need to update processor to contain a map of processors and pick the appropriate processor for each ProvReq or have a list of processors and iterate over each processor and each ProvReq.

cluster-autoscaler/provisioningrequest/checkcapacity/processor.go
yaroslava-serdiuk: I don't think we should limit the number of ProvReqs, we need to make sure we have updated conditions for all ProvReq

cluster-autoscaler/processors/processors.go
kisieland: I would prefer to have a list of processors from the start to avoid changing the logic of the autoscaler multiple times.

cluster-autoscaler/provisioningrequest/checkcapacity/processor.go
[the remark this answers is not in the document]
kisieland: makes sense.

cluster-autoscaler/provisioningrequest/checkcapacity/processor.go
kisieland: Ack, I would still prefer to have some cap to avoid overloading CA in this part and starving the actual scale up logic.

yaroslava-serdiuk: We might spend some time here, but this won't happen each CA loop, because we setCondition once for each ProvReq: or BookingExpired or Failed at some point. 
I'm worried that with ProvReqs cut, we might have a situation when some ProvReq won't be processed for sometime, so we overbook capacity for it or tried to scaleUp that will result in blocking other ProvReqs from scaleUp.

yaroslava-serdiuk: make sense, updated

cluster-autoscaler/processors/processors.go
yaroslava-serdiuk: done

cluster-autoscaler/provisioningrequest/checkcapacity/processor.go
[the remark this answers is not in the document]
kisieland: I would cap both kinds of operations, i.e. the expired and failed.

We should prefer the expiring, but if there are too many provreqs to expire we still could spend a significant time in this processor.

Also we should add tests.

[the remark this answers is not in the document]
yaroslava-serdiuk: Updated

cluster-autoscaler/provisioningrequest/checkcapacity/processor.go
```diff
@@ -0,0 +1,91 @@
...
+func (p *checkCapacityProcessor) Process(provReqs []*provreqwrapper.ProvisioningRequest) error {
+	expiredProvReq := []*provreqwrapper.ProvisioningRequest{}
+	failedProvReq := []*provreqwrapper.ProvisioningRequest{}
+	for _, provReq := range provReqs {
```
kisieland, line 52: one think I just remembered, we should only do this for check capacity provreqs.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -507,6 +507,10 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
 		a.AutoscalingContext.DebuggingSnapshotter.SetClusterNodes(l)
 	}
 
+	if err := a.processors.ProvisioningRequestProcessor.Process(nil); err != nil {
```
kisieland, line 510: I would make the `ProvisioningRequestProcessor.Process(nil)` take no arguments.
Only the sub-processors could receive the provreqs, which would be fetched by the combined processor.

yaroslava-serdiuk: Done

cluster-autoscaler/provisioningrequest/checkcapacity/processor.go
yaroslava-serdiuk: Done

x13n (changes requested):
Thanks, mostly LGTM. I don't particularly like the two-level processor, but I think with a bit of renaming and shuffling the code around, it would be slightly cleaner. Specifically, I suggest:
-  Renaming `ProvisioningRequestProcessor provreq.ProvisioningRequestProcessor` to `LoopStartNotifier *loopstart.ObserversList` (and perhaps its `Process` method to `Refresh` akin to cloud provider `Refresh` also happening on loop start). This would be a new package & object (not interface) similar to the one in `scale_down_candidates_observer.go`.
-  Renaming `ProvisioningClassProcessor` interface to `ProvisioningRequestProcessor`

That should allow us to later move more logic out of `RunOnce` in the future. Does that make sense?

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -507,6 +507,10 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
 		a.AutoscalingContext.DebuggingSnapshotter.SetClusterNodes(l)
 	}
 
+	if err := a.processors.ProvisioningRequestProcessor.Process(); err != nil {
```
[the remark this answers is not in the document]
yaroslava-serdiuk: ProvisioningRequestProcessor is aimed to process ProvisioningRequest conditions and check if they are expired, or the booking capacity mechanism is expired and update ProvisioningRequest conditions if needed.
This Processor should be called before ProvisioningRequestInjector (I haven't introduced it yet, it's will be in next PR), and ProvisioningRequestInjector will be a part of PodListProcessor (or it will be called just after PodListProcessor) and will be called after ProvisioningRequestFilter (which filter out real unscheduled pods that consume ProvisioningRequest from unscheduled pods to not trigger another scaleUp).

Currently we don't list ProvisioningRequests in the beginning of the loop (as we do for regular pods). I added a client to the processor and list ProvReqs inside the processor. That's why the input is empty.

Currently I'm adding support for check-capacity ProvisioningClass, but then I'll add support for atomicScaleUp ProvisioningClass. Each class will process ProvisioningRequest differently, so that's why ProvisioningRequestProcessor has a list of ProvisioningClassProcessor where each processor receives a list of ProvisioningRequests. 
Initially I thought to have one interface for ProvisioningRequestProcessor and ProvisioningClassProcessor and wanted to pass nil list to ProvisioningRequestProcessor in the RunOnce() function, but @kisieland asked to have an empty input instead, so I introduced two processors.

cluster-autoscaler/processors/provreq/provsioning_request_processor.go
```diff
@@ -0,0 +1,90 @@
...
+	"k8s.io/client-go/rest"
+)
+
+// ProvisioningRequestProcessor process ProvisionignRequests in the cluster.
+type ProvisioningRequestProcessor interface {
+	Process() error
+	CleanUp()
+}
+
+// ProvisioningClassProcessor process ProvisionignRequests in the cluster.
```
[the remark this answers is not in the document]
yaroslava-serdiuk: I wrote a comment below with full context, will copy the part of it that related to your question: 
"Currently I'm adding support for check-capacity ProvisioningClass, but then I'll add support for atomicScaleUp ProvisioningClass. Each class will process ProvisioningRequest differently, so that's why ProvisioningRequestProcessor has a list of ProvisioningClassProcessor where each processor receives a list of ProvisioningRequests.
Initially I thought to have one interface for ProvisioningRequestProcessor and ProvisioningClassProcessor and wanted to pass nil list to ProvisioningRequestProcessor in the RunOnce() function, but @kisieland asked to have an empty input instead, so I introduced two processors."

cluster-autoscaler/processors/provreq/provsioning_request_processor.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"k8s.io/autoscaler/cluster-autoscaler/provisioningrequest/provreqclient"
+	"k8s.io/autoscaler/cluster-autoscaler/provisioningrequest/provreqwrapper"
+	"k8s.io/client-go/rest"
+)
+
+// ProvisioningRequestProcessor process ProvisionignRequests in the cluster.
+type ProvisioningRequestProcessor interface {
+	Process() error
```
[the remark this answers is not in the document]
yaroslava-serdiuk: Yes, currently ProvisioningRequest are hidden, so we don't have a list of ProvisioningRequests to pass it here and we don't need to return anything since nothing accept the list of ProvisioningRequest. 
This was initial design decision from @kisieland . But I'm open to discussion if you think that we need to change this approach.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -336,6 +340,15 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
 		klog.Errorf("Failed to refresh cloud provider config: %v", err)
 		return caerrors.ToAutoscalerError(caerrors.CloudProviderError, err)
 	}
+	if a.loopStartNotifier != nil {
```
x13n, line 343: I'd just assume non-nil. If you want to handle `nil` separately, it doesn't make sense to do the defaulting in `autoscaler.go`.

cluster-autoscaler/observers/loopstart/loopstart.go
```diff
@@ -0,0 +1,51 @@
...
+*/
+
+package loopstart
+
+import "errors"
+
+// Observer interface is used to store object that needed to be refreshed in each CA loop.
+// It returns error and a bool value whether the loop should be skipped.
+type Observer interface {
+	Refresh() (bool, error)
```
x13n, line 24: Do we need a return type here? Observers should ideally just observe, not break CA loop. Specifically in ProvReq case - we may prevent regular pod-based autoscaling from working only because some there was an error in handling ProvisioningRequests.

cluster-autoscaler/processors/provreq/provsioning_request_processor.go
```diff
@@ -0,0 +1,67 @@
...
+package provreq
+
+import (
+	"k8s.io/autoscaler/cluster-autoscaler/observers/loopstart"
+	"k8s.io/autoscaler/cluster-autoscaler/provisioningrequest/provreqclient"
+	"k8s.io/autoscaler/cluster-autoscaler/provisioningrequest/provreqwrapper"
+	"k8s.io/client-go/rest"
+)
+
+// ProvisioningRequestProcessor process ProvisionignRequests in the cluster.
```
x13n, line 26: nit: Another `ProvisionignRequests` -> `ProvisioningRequests`. Also I noticed there's a typo in the filename as well, can you rename it?

x13n (changes requested):

cluster-autoscaler/processors/provreq/provsioning_request_processor.go
```diff
@@ -0,0 +1,67 @@
...
+		return nil, err
+	}
+	return &CombinedProvReqProcessor{client: client, processors: processors}, nil
+}
+
+// Refresh iterates over ProvisioningRequests and updates its conditions/state.
+func (cp *CombinedProvReqProcessor) Refresh() (bool, error) {
+	provReqs, err := cp.client.ProvisioningRequests()
+	if err != nil {
+		return true, err
```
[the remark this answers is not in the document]
yaroslava-serdiuk: We actually should have error in main (when initialising orchestrator), so should be safe enough to just log the error.

cluster-autoscaler/observers/loopstart/loopstart.go
yaroslava-serdiuk: You're right. Actually we never return error for ProvisioningRequestProcessor.

x13n (changes requested):
Just two minor comments, otherwise LGTM.

cluster-autoscaler/go.mod
```diff
@@ -1,6 +1,6 @@
 module k8s.io/autoscaler/cluster-autoscaler
 
-go 1.21
+go 1.21.3
```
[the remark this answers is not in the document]
yaroslava-serdiuk: Sorry, that's for internal testing

yaroslava-serdiuk:
#6814

PR #6497, opened 2024-02-01
Reduce time waited for pod evictions

domenicbozzuto:
#### What type of PR is this?
/kind feature

The time waited for all pods in a group to be drained from a node is guaranteed to be no more than the grace period configured for the PriorityClass group + the pod eviction headroom. However, if all the pod's termination grace periods are less than the PriorityClass-level grace period, we can reduce the time waited to the largest termination grace period of the pods in the group.  This prevents waiting an unnecessarily long time when a group is configured with a very high grace period but all the pods were safely evicted sooner.

Note: I originally opened https://github.com/kubernetes/autoscaler/pull/5711 for this a while ago, but the implementation of the `scaledown/actuation` package has changed drastically since then; I decided opening a new PR probably made more sense.

Fixes #

RE: testing; I opted to add tests for the `curtailTerminationGracePeriod` function directly; I'm generally not a fan of unit tests for non-exported functions, but this felt cleaner/less invasive than trying to add timing/sleep logic into the tests.

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/scaledown/actuation/drain.go
+++ b/cluster-autoscaler/core/scaledown/actuation/drain.go
@@ -122,8 +122,9 @@ func (e Evictor) drainNodeWithPodsBasedOnPodPriority(ctx *acontext.AutoscalingCo
 			return evictionResults, err
 		}
 
-		// Evictions created successfully, wait ShutdownGracePeriodSeconds + podEvictionHeadroom to see if fullEviction pods really disappeared.
-		evictionResults, err = e.waitPodsToDisappear(ctx, node, group.FullEvictionPods, evictionResults, group.ShutdownGracePeriodSeconds)
+		// Evictions created successfully, wait up to ShutdownGracePeriodSeconds + podEvictionHeadroom to see if fullEviction pods really disappeared.
+		maxTerminationGracePeriodSeconds := curtailTerminationGracePeriod(evictionResults, group.ShutdownGracePeriodSeconds)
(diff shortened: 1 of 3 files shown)


x13n:
The change itself generally LGTM, I'm wondering though what you're trying to achieve with it. The only outcome is going to be that scale down failures will become more likely (and sooner) with less time to evict pods. Note that if all the pods get evicted earlier, actual VM deletion will also happen sooner, since CA is just polling for that in a loop:

https://github.com/kubernetes/autoscaler/blob/00fbbe19a7045536a2d967902eb132c37f5a9f8e/cluster-autoscaler/core/scaledown/actuation/drain.go#L138-L156

Is the goal here to fail faster? Because this PR won't improve anything on the happy path.

domenicbozzuto:
> The change itself generally LGTM, I'm wondering though what you're trying to achieve with it. The only outcome is going to be that scale down failures will become more likely (and sooner) with less time to evict pods. Note that if all the pods get evicted earlier, actual VM deletion will also happen sooner, since CA is just polling for that in a loop

Yeah after reviewing some more I totally agree, I think I must've been considering an error case that's not really possible; `waitForDeletion` won't be called if `initiateEvictions` has any errors, and `initiateEvictions` will only succeed if evictions are createable for _every_ pod (and like you said, the polling loop in `waitForDeletion` should therefore already end as soon as the longest assigned grace period takes).

Going to close as unnecessary, thank you much for the commentary and apologies for the noise 🙇

PR #6525, opened 2024-02-12
Refactor: Move AutoscalingContext to Estimate func

azylinski:
#### What type of PR is this?

/kind cleanup

The Estimator Builder should use (wide) AutoscalingContext instead of (specific sub-) params.
That would be consistent with other interfaces, e.g.: all processors, scaleup/scaledown.

```release-note
"NONE"
```

/assign @x13n

### DIFF
--- a/cluster-autoscaler/estimator/binpacking_estimator.go
+++ b/cluster-autoscaler/estimator/binpacking_estimator.go
@@ -21,8 +21,8 @@ import (
 
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/autoscaler/cluster-autoscaler/cloudprovider"
+	cactx "k8s.io/autoscaler/cluster-autoscaler/context"
 	"k8s.io/autoscaler/cluster-autoscaler/simulator/clustersnapshot"
-	"k8s.io/autoscaler/cluster-autoscaler/simulator/predicatechecker"
 	"k8s.io/autoscaler/cluster-autoscaler/utils/scheduler"
 	klog "k8s.io/klog/v2"
 	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
@@ -30,29 +30,23 @@ import (
 
 // BinpackingNodeEstimator estimates the number of needed nodes to handle the given amount of pods.
 type BinpackingNodeEstimator struct {
-	predicateChecker       predicatechecker.PredicateChecker
-	clusterSnapshot        clustersnapshot.ClusterSnapshot
 	limiter                EstimationLimiter
 	podOrderer             EstimationPodOrderer
-	context                EstimationContext
+	estimationContext      EstimationContext
 	estimationAnalyserFunc EstimationAnalyserFunc // optional
 }
 
 // NewBinpackingNodeEstimator builds a new BinpackingNodeEstimator.
 func NewBinpackingNodeEstimator(
-	predicateChecker predicatechecker.PredicateChecker,
-	clusterSnapshot clustersnapshot.ClusterSnapshot,
 	limiter EstimationLimiter,
 	podOrderer EstimationPodOrderer,
-	context EstimationContext,
+	estimationContext EstimationContext,
 	estimationAnalyserFunc EstimationAnalyserFunc,
 ) *BinpackingNodeEstimator {
 	return &BinpackingNodeEstimator{
-		predicateChecker:       predicateChecker,
-		clusterSnapshot:        clusterSnapshot,
 		limiter:                limiter,
 		podOrderer:             podOrderer,
-		context:                context,
+		estimationContext:      estimationContext,
 		estimationAnalyserFunc: estimationAnalyserFunc,
 	}
 }
@@ -68,11 +62,12 @@ func NewBinpackingNodeEstimator(
 // It is assumed that all pods from the given list can fit to nodeTemplate.
 // Returns the number of nodes needed to accommodate all pods from the list.
 func (e *BinpackingNodeEstimator) Estimate(
+	ctx *cactx.AutoscalingContext,
 	pods []*apiv1.Pod,
 	nodeTemplate *schedulerframework.NodeInfo,
 	nodeGroup cloudprovider.NodeGroup) (int, []*apiv1.Pod) {
 
-	e.limiter.StartEstimation(pods, nodeGroup, e.context)
+	e.limiter.StartEstimation(pods, nodeGroup, e.estimationContext)
 	defer e.limiter.EndEstimation()
 
 	pods = e.podOrderer.Order(pods, nodeTemplate, nodeGroup)
@@ -80,9 +75,9 @@ func (e *BinpackingNodeEstimator) Estimate(
 	newNodeNames := make(map[string]bool)
 	newNodesWithPods := make(map[string]bool)
 
-	e.clusterSnapshot.Fork()
+	ctx.ClusterSnapshot.Fork()
 	defer func() {
-		e.clusterSnapshot.Revert()
+		ctx.ClusterSnapshot.Revert()
 	}()
 
 	newNodeNameIndex := 0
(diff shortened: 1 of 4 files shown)


vadasambar:
> The Estimator Builder should use (wide) AutoscalingContext instead of (specific sub-) params.
> That would be consistent with other interfaces, e.g.: all processors, scaleup/scaledown.

Is there any other reason apart from consistency? I don't see why estimator should match processors or scaleup/scaledown when it comes to passing parameters. They are quite different constructs. 

I think passing context to the estimator makes the function definition less prone to changes in the future because you can just use anything from the context (right now you have to pluck the value from the context and pass it to the estimator) but that also means estimator now knows a whole lot more than it should. Do we really need this?

azylinski:
> Is there any other reason apart from consistency? I don't see why estimator should match processors or scaleup/scaledown when it comes to passing parameters. They are quite different constructs.

I'd say that estimator (especially: `estimationAnalyserFunc` and `binpackingLimiter`) are pretty much processors.


> estimator now knows a whole lot more than it should

I agree 100%. IMHO `AutoscalerContext` by itself is more than it should.
I see it as a collection of: options, clients and "global" state, that we pass around. It's not ideal (again, just my opinion), but it's a design decision that was made long time ago.
And again, this is currently the case for all the processors.


> definition less prone to changes in the future because you can just use anything from the context

In essence, that's the direction I'm aiming for.
Current PR is 2nd part of refactor - and there's one more to come, when I'd like to move nodeInfos (or better: TemplateNodesForGroups) to context as well. That's not only a huge code simplification, but also pairs well wth `ClusterStateRegistry` "level".

MaciekPytel:
> > estimator now knows a whole lot more than it should
>
> I agree 100%. IMHO AutoscalerContext by itself is more than it should.

I think this is a decision we already took with introduction of processors and the definition of AutoscalingContext. The idea was that we want to allow cloud providers / advanced users to have broad power to customize CA to fit the specifics of their platform, including arbitrary changes to internal CA logic. 

CA cloudprovider interface expects very little from provider, which makes it relatively easy to add integrations (as demonstrated by the number of integrations we have). The downside is that our interface and our generic logic is essentially a lowest common denominator which makes it very hard to use more advanced cloud provider features (e.g. how long it took for CA integration with MixedInstancesPolicy? How limiting that integration is to this day?). 

The idea of introducing processors was to address this limitation, by giving providers tools to essentially completely change arbitrary parts of CA logic. To this end we've added programming hooks in most critical parts of CA and gave them full access to all the internal state (i.e. AutoscalingContext). The expectation was users forking CA and writing custom processors that can significantly change / override core logic in order to support their specific use-cases that cannot be satisfied purely via implementation of cloudprovider interface. 

To sum up: we designed AutoscalingContext the way it is, specifically to allow people to inject arbitrary and unexpected logic in CA extension points (processors). It can be argued if this is the right approach, but I think pushing AutoscalingContext to any interface that we reasonably expect to be customized by users (e.g. EstimationAnalyserFunc or BinpackingLimiter) is at least consistent with what we've been doing elsewhere in CA codebase.

cluster-autoscaler/estimator/binpacking_estimator.go
```diff
@@ -21,38 +21,34 @@ import (
...
 	"k8s.io/autoscaler/cluster-autoscaler/utils/scheduler"
 	klog "k8s.io/klog/v2"
 	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
 )
 
 // BinpackingNodeEstimator estimates the number of needed nodes to handle the given amount of pods.
 type BinpackingNodeEstimator struct {
-	predicateChecker       predicatechecker.PredicateChecker
-	clusterSnapshot        clustersnapshot.ClusterSnapshot
+	ctx                    *cactx.AutoscalingContext
```
MaciekPytel, line 32: Feel free to ignore this comment if it's very hard to implement, but it would be much more consistent with the rest of the codebase to have ctx as a first parameter to Estimate() rather than keep it here. 
Also it seems counterintuitive to persist context. Our implementation happens to re-use same context object rather than recreating it between loops, but I'd rather not have that as an assumption if we can avoid it.

azylinski: Moved to `Estimate()` - your are absolutely right, context should not be stored.

azylinski: And I couldn't agree more about "the love for EstimationContext". If you look closer: 
 1st: this more of a Limiter / Threshold Context 
 2nd: this is the only reason why we have this builder pattern, so the threshold limits could be updated inline

This should be simplified, but as you pointed out, that's out of PR scope.

vadasambar:
> > > estimator now knows a whole lot more than it should
> > 
> > 
> > I agree 100%. IMHO AutoscalerContext by itself is more than it should.
> 
> I think this is a decision we already took with introduction of processors and the definition of AutoscalingContext. The idea was that we want to allow cloud providers / advanced users to have broad power to customize CA to fit the specifics of their platform, including arbitrary changes to internal CA logic.
> 
> CA cloudprovider interface expects very little from provider, which makes it relatively easy to add integrations (as demonstrated by the number of integrations we have). The downside is that our interface and our generic logic is essentially a lowest common denominator which makes it very hard to use more advanced cloud provider features (e.g. how long it took for CA integration with MixedInstancesPolicy? How limiting that integration is to this day?).
> 
> The idea of introducing processors was to address this limitation, by giving providers tools to essentially completely change arbitrary parts of CA logic. To this end we've added programming hooks (processors) in most critical parts of CA and gave them full access to all the internal state (i.e. AutoscalingContext). The expectation was users forking CA and writing custom processors that can significantly change / override core logic in order to support their specific use-cases that cannot be satisfied purely via implementation of cloudprovider interface.
> 
> To sum up: we designed AutoscalingContext the way it is, specifically to allow people to inject arbitrary and unexpected logic in CA extension points (processors). It can be argued if this is the right approach, but I think pushing AutoscalingContext to any interface that we reasonably expect to be customized by users (e.g. EstimationAnalyserFunc or BinpackingLimiter) is at least consistent with what we've been doing elsewhere in CA codebase.

Appreciate the context @MaciekPytel . I will review the PR from the 

> `binpackingLimiter`

On a second look I see why you say estimator and limiter are processors :)

vadasambar:
> > Is there any other reason apart from consistency? I don't see why estimator should match processors or scaleup/scaledown when it comes to passing parameters. They are quite different constructs.
> 
> I'd say that estimator (especially: `estimationAnalyserFunc` and `binpackingLimiter`) are pretty much processors.
> 
> > estimator now knows a whole lot more than it should
> 
> I agree 100%. IMHO `AutoscalerContext` by itself is more than it should. I see it as a collection of: options, clients and "global" state, that we pass around. It's not ideal (again, just my opinion), but it's a design decision that was made long time ago. And again, this is currently the case for all the processors.
> 
> > definition less prone to changes in the future because you can just use anything from the context
> 
> In essence, that's the direction I'm aiming for. Current PR is 2nd part of refactor - and there's one more to come, when I'd like to move nodeInfos (or better: TemplateNodesForGroups) to context as well. That's not only a huge code simplification, but also pairs well wth `ClusterStateRegistry` "level".

On a second look, I see why you say estimator and limiter are processors :)

vadasambar:
> > > estimator now knows a whole lot more than it should
> > 
> > 
> > I agree 100%. IMHO AutoscalerContext by itself is more than it should.
> 
> I think this is a decision we already took with introduction of processors and the definition of AutoscalingContext. The idea was that we want to allow cloud providers / advanced users to have broad power to customize CA to fit the specifics of their platform, including arbitrary changes to internal CA logic.
> 
> CA cloudprovider interface expects very little from provider, which makes it relatively easy to add integrations (as demonstrated by the number of integrations we have). The downside is that our interface and our generic logic is essentially a lowest common denominator which makes it very hard to use more advanced cloud provider features (e.g. how long it took for CA integration with MixedInstancesPolicy? How limiting that integration is to this day?).
> 
> The idea of introducing processors was to address this limitation, by giving providers tools to essentially completely change arbitrary parts of CA logic. To this end we've added programming hooks (processors) in most critical parts of CA and gave them full access to all the internal state (i.e. AutoscalingContext). The expectation was users forking CA and writing custom processors that can significantly change / override core logic in order to support their specific use-cases that cannot be satisfied purely via implementation of cloudprovider interface.
> 
> To sum up: we designed AutoscalingContext the way it is, specifically to allow people to inject arbitrary and unexpected logic in CA extension points (processors). It can be argued if this is the right approach, but I think pushing AutoscalingContext to any interface that we reasonably expect to be customized by users (e.g. EstimationAnalyserFunc or BinpackingLimiter) is at least consistent with what we've been doing elsewhere in CA codebase.

Appreciate the context. Might be a good idea to document this somewhere (feels like tribal knowledge).

x13n:
I just realized I never replied here. Approving, based on the reasoning above. @azylinski can you rebase?

/approve

PR #6529, opened 2024-02-15
Add ProvisioningRequest injector

yaroslava-serdiuk:
#### What type of PR is this?
/kind feature

Inject pods from ProvisioningRequest to unschedulable pods, so CA will try to run ScaleUp.

Part of implementation https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/proposals/provisioning-request.md

```release-note
Support for Check-capacity ProvisioningClass is added. To enable the feature set --enable-provisioning-requests=true. If there is capacity in the cluster for ProvisioningRequest of Check-capacity ProvisioniningClass, the Provisioned=True condition will be set and the capacity will be reserved from scheduling other ProvisioningRequest for 10 minutes. 
Note: ClusterAutoscaler doesn't block other pods to be scheduled on reserved capacity; ClusterAutoscaler can scale down reserved capacity.
```

### DIFF
--- a/cluster-autoscaler/processors/provreq/provisioning_request_injector.go
+++ b/cluster-autoscaler/processors/provreq/provisioning_request_injector.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/autoscaler/cluster-autoscaler/provisioningrequest/provreqclient"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/clock"
 )
 
 const (
@@ -42,7 +43,7 @@ var SupportedProvisioningClasses = []string{v1beta1.ProvisioningClassCheckCapaci
 // ProvisioningRequestPodsInjector creates in-memory pods from ProvisioningRequest and inject them to unscheduled pods list.
 type ProvisioningRequestPodsInjector struct {
 	client provisioningRequestClient
-	now    func() time.Time
+	clock  clock.PassiveClock
 }
 
 // Process pick one ProvisioningRequest, update Accepted condition and inject pods to unscheduled pods list.
@@ -64,16 +65,23 @@ func (p *ProvisioningRequestPodsInjector) Process(
 
 		//TODO(yaroslava): support exponential backoff
 		// Inject pods if ProvReq wasn't scaled up before or it has Provisioned == False condition more than defaultRetryTime
-		if provisioned == nil || (provisioned != nil && provisioned.Status == metav1.ConditionFalse &&
-			provisioned.LastTransitionTime.Add(defaultRetryTime).Before(p.now())) {
+		inject := true
+		if provisioned != nil {
+			if provisioned.Status == metav1.ConditionFalse && provisioned.LastTransitionTime.Add(defaultRetryTime).Before(p.clock.Now()) {
+				inject = true
+			} else {
+				inject = false
+			}
+		}
+		if inject {
 			provreqpods, err := provreqpods.PodsForProvisioningRequest(pr)
 			if err != nil {
 				klog.Errorf("Failed to get pods for ProvisioningRequest %v", pr.Name())
-				provreqconditions.AddOrUpdateCondition(pr, v1beta1.Failed, metav1.ConditionTrue, provreqconditions.FailedToCreatePodsReason, err.Error(), metav1.NewTime(p.now()))
+				provreqconditions.AddOrUpdateCondition(pr, v1beta1.Failed, metav1.ConditionTrue, provreqconditions.FailedToCreatePodsReason, err.Error(), metav1.NewTime(p.clock.Now()))
 				continue
 			}
 			unschedulablePods := append(unschedulablePods, provreqpods...)
-			provreqconditions.AddOrUpdateCondition(pr, v1beta1.Accepted, metav1.ConditionTrue, provreqconditions.AcceptedReason, provreqconditions.AcceptedMsg, metav1.NewTime(p.now()))
+			provreqconditions.AddOrUpdateCondition(pr, v1beta1.Accepted, metav1.ConditionTrue, provreqconditions.AcceptedReason, provreqconditions.AcceptedMsg, metav1.NewTime(p.clock.Now()))
 			return unschedulablePods, nil
 		}
 	}
@@ -89,5 +97,5 @@ func NewProvisioningRequestPodsInjector(kubeConfig *rest.Config) (pods.PodListPr
 	if err != nil {
 		return nil, err
 	}
-	return &ProvisioningRequestPodsInjector{client: client, now: time.Now}, nil
+	return &ProvisioningRequestPodsInjector{client: client, clock: clock.RealClock{}}, nil
 }
--- a/cluster-autoscaler/processors/provreq/provisioning_request_injector.go
+++ b/cluster-autoscaler/processors/provreq/provisioning_request_injector.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package provreq
+
+import (
+	"time"
+
+	apiv1 "k8s.io/api/core/v1"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/autoscaler/cluster-autoscaler/context"
+	"k8s.io/autoscaler/cluster-autoscaler/processors/pods"
+	"k8s.io/autoscaler/cluster-autoscaler/provisioningrequest/apis/autoscaling.x-k8s.io/v1beta1"
+	provreqconditions "k8s.io/autoscaler/cluster-autoscaler/provisioningrequest/conditions"
+	provreqpods "k8s.io/autoscaler/cluster-autoscaler/provisioningrequest/pods"
+	"k8s.io/autoscaler/cluster-autoscaler/provisioningrequest/provreqclient"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
+)
+
+const (
+	defaultRetryTime = 10 * time.Minute
+)
+
+// ProvisioningRequestPodsInjector creates in-memory pods from ProvisioningRequest and inject them to unscheduled pods list.
+type ProvisioningRequestPodsInjector struct {
+	client provisioningRequestClient
+	now    func() time.Time
+}
+
+// Process pick one ProvisioningRequest, update Accepted condition and inject pods to unscheduled pods list.
+func (p *ProvisioningRequestPodsInjector) Process(
+	_ *context.AutoscalingContext,
+	unschedulablePods []*apiv1.Pod,
+) ([]*apiv1.Pod, error) {
+	provReqs, err := p.client.ProvisioningRequests()
+	if err != nil {
+		return nil, err
+	}
+	for _, pr := range provReqs {
+		conditions := pr.Conditions()
+		if apimeta.IsStatusConditionTrue(conditions, v1beta1.Failed) || apimeta.IsStatusConditionTrue(conditions, v1beta1.Provisioned) {
+			continue
+		}
+
+		provisioned := apimeta.FindStatusCondition(conditions, v1beta1.Provisioned)
+
+		//TODO(yaroslava): support exponential backoff
+		// Inject pods if ProvReq wasn't scaled up before or it has Provisioned == False condition more than defaultRetryTime
+		if provisioned == nil || (provisioned != nil && provisioned.Status == metav1.ConditionFalse &&
+			provisioned.LastTransitionTime.Add(defaultRetryTime).Before(p.now())) {
+			provreqpods, err := provreqpods.PodsForProvisioningRequest(pr)
+			if err != nil {
+				klog.Errorf("Failed to get pods for ProvisioningRequest %v", pr.Name())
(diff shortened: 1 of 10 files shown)


cluster-autoscaler/processors/provreq/provisioning_request_injector.go
```diff
@@ -0,0 +1,91 @@
...
+			continue
+		}
+
+		provisioned := apimeta.FindStatusCondition(conditions, v1beta1.Provisioned)
+		accepted := apimeta.FindStatusCondition(conditions, v1beta1.Provisioned)
+
+		//TODO(yaroslava): support exponential backoff
+		// Inject pods if ProvReq is new or it has Provisioned == False condition more than defaultRetryTime
+		if accepted == nil || provisioned != nil &&
+			provisioned.LastTransitionTime.Add(defaultRetryTime).Before(p.now()) {
```
kisieland, line 66: shouldn't we check if the condition is actually `Provisioned=False` ?
It could be also `Provisioned=Unknown`

kisieland: We should add a test case for it as well

cluster-autoscaler/provisioningrequest/apis/autoscaling.x-k8s.io/v1beta1/types.go
```diff
@@ -175,6 +175,9 @@ type Detail string
 // The following constants list all currently available Conditions Type values.
 // See: https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition
 const (
+	// Accepted indicates that the ProvisioningRequest was accepted by ClusterAutoscaler,
+	// so ClusterAutoscaler will attempt to provision the nodes for it.
+	Accepted string = "Accepted"
```
kisieland, line 180: Can we also add the `CapacityRevoked` condition?

yaroslava-serdiuk: Added

cluster-autoscaler/processors/provreq/provisioning_request_injector.go
yaroslava-serdiuk: Right, updated

cluster-autoscaler/processors/provreq/provisioning_request_injector.go
```diff
@@ -0,0 +1,91 @@
...
+		return nil, err
+	}
+	for _, pr := range provReqs {
+		conditions := pr.Conditions()
+		if apimeta.IsStatusConditionTrue(conditions, v1beta1.Failed) || apimeta.IsStatusConditionTrue(conditions, v1beta1.Provisioned) {
+			continue
+		}
+
+		provisioned := apimeta.FindStatusCondition(conditions, v1beta1.Provisioned)
+		accepted := apimeta.FindStatusCondition(conditions, v1beta1.Provisioned)
```
jayantjain93, line 61: I'm don't quite understand why the same call method with same variables, be used to two create two variables.
Maybe a bug?

cluster-autoscaler/processors/provreq/provisioning_request_injector.go
```diff
@@ -0,0 +1,91 @@
...
+		if accepted == nil || provisioned != nil && provisioned.Status == metav1.ConditionFalse &&
+			provisioned.LastTransitionTime.Add(defaultRetryTime).Before(p.now()) {
+			provreqpods, err := provreqpods.PodsForProvisioningRequest(pr)
+			if err != nil {
+				klog.Errorf("Failed to get pods for ProvisioningRequest %v", pr.Name())
+				provreqconditions.AddOrUpdateCondition(pr, v1beta1.Failed, metav1.ConditionTrue, provreqconditions.FailedToCreatePodsReason, err.Error(), metav1.NewTime(p.now()))
+				continue
+			}
+			unschedulablePods := append(unschedulablePods, provreqpods...)
+			provreqconditions.AddOrUpdateCondition(pr, v1beta1.Accepted, metav1.ConditionTrue, provreqconditions.AcceptedReason, provreqconditions.AcceptedMsg, metav1.NewTime(p.now()))
```
jayantjain93, line 74: Maybe I'm missing some context, but why would we have Status be updated to "Accepted" when the check is done as `accepted == nil`.
Is it a transitive which you're setting here, that will change the flag to `accepted != nil`

cluster-autoscaler/provisioningrequest/apis/autoscaling.x-k8s.io/v1beta1/types.go
```diff
@@ -175,9 +175,14 @@ type Detail string
...
 	// BookingExpired indicates that the ProvisioningRequest had Provisioned condition before
 	// and capacity reservation time is expired.
 	BookingExpired string = "BookingExpired"
+	// CapacityRevoked indicates that requested resources are not longer valid.
+	CapacityRevoked string = "CapacityRevoked"
```
jayantjain93, line 185: we added a new type, but I don't see any usage of it. Maybe I'm missing something?

If not, we should introduce new type with the usage imo.

cluster-autoscaler/provisioningrequest/checkcapacity/orchestrator.go
```diff
@@ -151,10 +152,10 @@ func (o *provReqOrchestrator) scaleUp(unschedulablePods []*apiv1.Pod) (bool, err
 	}
 	st, _, err := o.injector.TrySchedulePods(o.context.ClusterSnapshot, unschedulablePods, scheduling.ScheduleAnywhere, true)
 	if len(st) < len(unschedulablePods) || err != nil {
-		setCondition(provReq, v1beta1.Provisioned, metav1.ConditionFalse, CapacityIsFoundReason, "Capacity is not found, CA will try to find it later.", metav1.Now())
+		conditions.AddOrUpdateCondition(provReq, v1beta1.Provisioned, metav1.ConditionFalse, conditions.CapacityIsFoundReason, "Capacity is not found, CA will try to find it later.", metav1.Now())
 		return false, err
 	}
-	setCondition(provReq, v1beta1.Provisioned, metav1.ConditionTrue, CapacityIsFoundReason, "Capacity is found in the cluster.", metav1.Now())
+	conditions.AddOrUpdateCondition(provReq, v1beta1.Provisioned, metav1.ConditionTrue, conditions.CapacityIsFoundReason, "Capacity is found in the cluster.", metav1.Now())
```
jayantjain93, line 158: nit: maybe we should also move this to `CapacityIsFoundMesg`?

cluster-autoscaler/provisioningrequest/checkcapacity/orchestrator.go
```diff
@@ -151,10 +152,10 @@ func (o *provReqOrchestrator) scaleUp(unschedulablePods []*apiv1.Pod) (bool, err
 	}
 	st, _, err := o.injector.TrySchedulePods(o.context.ClusterSnapshot, unschedulablePods, scheduling.ScheduleAnywhere, true)
 	if len(st) < len(unschedulablePods) || err != nil {
-		setCondition(provReq, v1beta1.Provisioned, metav1.ConditionFalse, CapacityIsFoundReason, "Capacity is not found, CA will try to find it later.", metav1.Now())
+		conditions.AddOrUpdateCondition(provReq, v1beta1.Provisioned, metav1.ConditionFalse, conditions.CapacityIsFoundReason, "Capacity is not found, CA will try to find it later.", metav1.Now())
```
jayantjain93, line 155: Maybe I'm missing something, byt the Condition Reason and Mesg don't seem to match.
`CapacityIsFoundReason` --> `Capacity is not found, CA will try to find it later. `
Is this expected?

cluster-autoscaler/provisioningrequest/conditions/conditions.go
```diff
@@ -40,27 +47,24 @@ const (
...
-func shouldCapacityBeBooked(pr *provreqwrapper.ProvisioningRequest) bool {
+// ShouldCapacityBeBooked returns whether capacity should be booked.
+func ShouldCapacityBeBooked(pr *provreqwrapper.ProvisioningRequest) bool {
 	if pr.V1Beta1().Spec.ProvisioningClassName != v1beta1.ProvisioningClassCheckCapacity {
 		return false
 	}
-	if pr.Conditions() == nil || len(pr.Conditions()) == 0 {
+	conditions := pr.Conditions()
+	if apimeta.IsStatusConditionTrue(conditions, v1beta1.Failed) || apimeta.IsStatusConditionTrue(conditions, v1beta1.BookingExpired) {
 		return false
```
jayantjain93, line 57: maybe we can simplify this code to:

```
conditions := pr.Conditions()
 if apimeta.IsStatusConditionTrue(conditions, v1beta1.Provisioned) {
		return true
}
return false
```

cluster-autoscaler/provisioningrequest/conditions/condition_test.go
```diff
@@ -242,7 +242,7 @@ func TestSetCondition(t *testing.T) {
 						Conditions: test.oldConditions,
 					},
 				}, nil)
-			setCondition(pr, test.newType, test.newStatus, "", "", v1.Now())
+			AddOrUpdateCondition(pr, test.newType, test.newStatus, "", "", v1.Now())
```
jayantjain93, line 245: Since we'd introduced new hardcoded condition of `Accepted` state, do you think we should add a test case where only the three selected conditions can make modification and any new case is added should need to update the test to reflect this?

cluster-autoscaler/processors/provreq/provisioning_request_injector.go
yaroslava-serdiuk: `Accepted indicates that the ProvisioningRequest was accepted by ClusterAutoscaler, so ClusterAutoscaler will attempt to provision the nodes for it.`
If Accepted == nil, means CA see ProvReq first time and should set this condition, since the pods from ProvReq are injected.

cluster-autoscaler/processors/provreq/provisioning_request_injector.go
yaroslava-serdiuk: Good catch, thanks!

cluster-autoscaler/provisioningrequest/conditions/conditions.go
yaroslava-serdiuk: No, we should check BookingExpired condition. If it's present, the capacity should not be booked

cluster-autoscaler/provisioningrequest/conditions/condition_test.go
yaroslava-serdiuk: What test do you have in mind? 
I have a test case that checks that Unknown condition is not updated.

cluster-autoscaler/provisioningrequest/apis/autoscaling.x-k8s.io/v1beta1/types.go
jayantjain93: Discussed offline on why we should keep it.

cluster-autoscaler/provisioningrequest/conditions/conditions.go
jayantjain93: The states are no exclusive.
https://maelvls.dev/kubernetes-conditions/#conditions-vs-state-machine

cluster-autoscaler/provisioningrequest/conditions/condition_test.go
jayantjain93: Ideal case would be run all the condition that get add [by default], which would try to run the update func, and there would be an expectation in the func to accept positive results to a sub-set of them.

This would ensure we'd have to modify the test if we add/remove conditions there, which would avoid un-intended changes.

yaroslava-serdiuk: Extended test with the case of Accepted condition. The current test will catch if any of "default" conditions is removed, but I'm not sure how to catch if the list is extended. That's a developer responsibility to cover implementation with tests.

jayantjain93: sounds like a good interaction.

cluster-autoscaler/processors/provreq/provisioning_request_injector.go
```diff
@@ -0,0 +1,93 @@
...
+	defaultRetryTime = 10 * time.Minute
+)
+
+// SupportedProvisioningClasses is a list of supported ProvisioningClasses in ClusterAutoscaler.
+var SupportedProvisioningClasses = []string{v1beta1.ProvisioningClassCheckCapacity}
+
+// ProvisioningRequestPodsInjector creates in-memory pods from ProvisioningRequest and inject them to unscheduled pods list.
+type ProvisioningRequestPodsInjector struct {
+	client provisioningRequestClient
+	now    func() time.Time
```
BigDarkClown, line 45: Please use `clock.PassiveClock` instead.

cluster-autoscaler/processors/provreq/provisioning_request_injector.go
```diff
@@ -0,0 +1,93 @@
...
+		conditions := pr.Conditions()
+		if apimeta.IsStatusConditionTrue(conditions, v1beta1.Failed) || apimeta.IsStatusConditionTrue(conditions, v1beta1.Provisioned) {
+			continue
+		}
+
+		provisioned := apimeta.FindStatusCondition(conditions, v1beta1.Provisioned)
+
+		//TODO(yaroslava): support exponential backoff
+		// Inject pods if ProvReq wasn't scaled up before or it has Provisioned == False condition more than defaultRetryTime
+		if provisioned == nil || (provisioned != nil && provisioned.Status == metav1.ConditionFalse &&
```
BigDarkClown, line 67: Please put this if statement into variable, it's hard too read. Something like

```
inject := true
if provisioned != nil && provisioned.Status == metav1.CondistionFalse {
  inject = provisioned.LastTransitionTime.Add(defaultRetryTime).Before(p.now()))
}

if inject {
```

BigDarkClown (commented):
Minor comments, otherwise lgtm

cluster-autoscaler/processors/provreq/provisioning_request_injector.go
yaroslava-serdiuk: Thanks!

cluster-autoscaler/processors/provreq/provisioning_request_injector.go
yaroslava-serdiuk: updated

BigDarkClown (commented):
/lgtm

yaroslava-serdiuk:
#6814

PR #6530, opened 2024-02-15
Introduce LocalSSDSizeProvider interface for GCE

atwamahmoud:
#### What type of PR is this?

/kind feature

Introduces `LocalSSDSizeProvider` interface to GCE cloud provider to be used to calculate local ssd disk size instead of using a plain constant.
Also introduces `SimpleLocalSSDProvider` struct implementing `LocalSSDSizeProvider` which returns a constant value `375GiB` to maintain original behaviour

```release-note
NONE
```

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/localssdsize/local_ssd_size_provider.go
+++ b/cluster-autoscaler/cloudprovider/gce/localssdsize/local_ssd_size_provider.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


cluster-autoscaler/cloudprovider/gce/localssdsize/local_ssd_size_provider.go
```diff
@@ -0,0 +1,46 @@
...
+
+// LocalSSDSizeProvider contains methods to calculate local ssd disk size for GCE based on some input parameters (e.g. machine type name)
+type LocalSSDSizeProvider interface {
+	// Computes local ssd disk size in GiB based on machine type name
+	SSDSizeInGiB(string) uint64
+}
+
+// LocalSSDDiskSizeInGiB is the size of each local SSD in GiB
+// (cf. https://cloud.google.com/compute/docs/disks/local-ssd)
+const LocalSSDDiskSizeInGiB = uint64(375)
```
BigDarkClown, line 27: micro nit: I would make this variable local as a sanity check, but I won't block this change on this. LGTM

BigDarkClown (commented):
/lgtm
/approve

PR #6531, opened 2024-02-15
GCE: Implement kube-env caching

BigDarkClown:
#### What type of PR is this?

/kind cleanup

Caching of kube-env for instance templates. The change greatly improves the CPU usage for idle clusters, as ~80% of the main loop time is spend on repetitive unmarshalling of kube-env.

The improvement is drastic for idle clusters even with only 3 MIGs. In my testing the loop time decreased from 1.25s to 0.25s.

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/cache.go
+++ b/cluster-autoscaler/cloudprovider/gce/cache.go
@@ -68,6 +68,7 @@ type GceCache struct {
 	migBaseNameCache          map[GceRef]string
 	instanceTemplateNameCache map[GceRef]string
 	instanceTemplatesCache    map[GceRef]*gce.InstanceTemplate
+	kubeEnvCache              map[GceRef]KubeEnv
 }
 
 // NewGceCache creates empty GceCache.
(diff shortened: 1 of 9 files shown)


atwamahmoud (approved):
In General, LGTM
However I might be messing some stuff so it'd be safer if someone else took a quick look

atwamahmoud:
/lgtm

It would be better to omit the `Get` from function names but it's out of scope for this PR since we'll have to rename not just added functions but older ones to remain consistent

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -409,6 +411,46 @@ func (gc *GceCache) InvalidateAllMigInstanceTemplates() {
...
+	defer gc.cacheMutex.Unlock()
+
+	if _, found := gc.kubeEnvCache[ref]; found {
+		klog.V(5).Infof("Kube-env cache invalidated for %s", ref)
+		delete(gc.kubeEnvCache, ref)
+	}
+}
+
+// InvalidateAllMigKubeEnvs clears the kube-env cache
+func (gc *GceCache) InvalidateAllMigKubeEnvs() {
```
x13n, line 446: Will we ever need either `Invalidate` ? Since instance templates are immutable, we probably won't ever call this. What we might want instead though is some expiration mechanism - to avoid memleaks in clusters with high MIG churn. This is a nice to have though, I don't expect single CA instance to observe enough instance templates over entire process lifetime to visibly inflate memory usage.

x13n:
I left a comment, but that is something to follow up on later, not in this PR, so let's merge as is.

/lgtm

Btw, @atwamahmoud - you don't need to be an OWNER to /lgtm, you just need to be a Kubernetes org member.

PR #6535, opened 2024-02-15
CA: GCE: add pricing for new Z3 machines

AnishShah:


cluster-autoscaler/cloudprovider/gce/gce_price_info.go
```diff
@@ -513,6 +518,8 @@ var (
 		"t2d-standard-32":  0.3271,
 		"t2d-standard-48":  0.4907,
 		"t2d-standard-60":  0.6134,
+		"z3-highmem-88":    6.75,
+		"z3-highmem-176":   10.37,
```
jayantjain93, line 522: prices we use are for us-cental1.
Could you change that this to reflect that.

https://cloud.google.com/compute/vm-instance-pricing

cluster-autoscaler/cloudprovider/gce/gce_price_info.go
```diff
@@ -336,6 +339,8 @@ var (
 		"t2d-standard-32":  1.3519,
 		"t2d-standard-48":  2.0278,
 		"t2d-standard-60":  2.5348,
+		"z3-highmem-88":    13.0,
+		"z3-highmem-176":   22.06,
```
jayantjain93, line 343: same as below

cluster-autoscaler/cloudprovider/gce/gce_price_info.go
AnishShah: Done.

cluster-autoscaler/cloudprovider/gce/gce_price_info.go
AnishShah: Done.

Issue #6544, opened 2024-02-20
Cluster Autoscaler should support multiple profiles


daimaxiaxie:
**Which component are you using?**:

Cluster Autoscaler

**Is your feature request designed to solve a problem? If so describe the problem this feature should solve.**:

The cluster scheduler can use multiple profiles to perform different scheduling for different pods. However, the current Cluster Autoscaler only supports one profile, which will lead to inaccurate scale up and down. So Cluster Autoscaler should support multiple profiles.

**Describe the solution you'd like.**:

Use the corresponding framework according to pod scheduler name in `SchedulerBasedPredicateChecker`.

**Describe any alternative solutions you've considered.**:

Added new predicate checker `MultiSchedulerPredicateChecker`, use flag enabled.

**Additional context.**:

NONE

PR #6545, opened 2024-02-20
feature: Support multiple profiles

daimaxiaxie:
#### What type of PR is this?

/kind feature

The cluster scheduler can use multiple profiles to perform different scheduling for different pods. However, the current Cluster Autoscaler only supports one profile, which will lead to inaccurate scale up and down.

Feature issue  https://github.com/kubernetes/autoscaler/issues/6544

```release-note
flag scheduler-config-file support the same configuration file as scheduler.
```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler_test.go
+++ b/cluster-autoscaler/core/static_autoscaler_test.go
@@ -21,7 +21,9 @@ import (
 	stdcontext "context"
 	"flag"
 	"fmt"
+	testconfig "k8s.io/autoscaler/cluster-autoscaler/config/test"
 	"os"
+	"path/filepath"
 	"reflect"
 	"strings"
 	"testing"
@@ -1469,6 +1471,22 @@ func TestStaticAutoscalerRunOnceWithUnselectedNodeGroups(t *testing.T) {
 func TestStaticAutoscalerRunOnceWithBypassedSchedulers(t *testing.T) {
 	bypassedScheduler := "bypassed-scheduler"
 	nonBypassedScheduler := "non-bypassed-scheduler"
+
+	tmpDir, err := os.MkdirTemp("", "scheduler-configs")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmpDir)
+	customConfigFile := filepath.Join(tmpDir, "custom_config.yaml")
+	if err := os.WriteFile(customConfigFile,
+		generateSchedulerConfig(apiv1.DefaultSchedulerName, bypassedScheduler, nonBypassedScheduler),
+		os.FileMode(0600)); err != nil {
+		t.Fatal(err)
+	}
+
+	customConfig, err := scheduler.ConfigFromPath(customConfigFile)
+	assert.NoError(t, err)
+
 	options := config.AutoscalingOptions{
 		NodeGroupDefaults: config.NodeGroupAutoscalingOptions{
 			ScaleDownUnneededTime:         time.Minute,
@@ -1485,6 +1503,7 @@ func TestStaticAutoscalerRunOnceWithBypassedSchedulers(t *testing.T) {
 			apiv1.DefaultSchedulerName,
 			bypassedScheduler,
 		}),
+		SchedulerConfig: customConfig,
 	}
 	now := time.Now()
 
@@ -2617,4 +2636,12 @@ func newEstimatorBuilder() estimator.EstimatorBuilder {
 	)
 
 	return estimatorBuilder
+}
+
+func generateSchedulerConfig(schedulerNames ...string) []byte {
+	schedulerConfig := testconfig.SchedulerConfigMinimalCorrect + "\nprofiles:\n"
+	for _, name := range schedulerNames {
+		schedulerConfig += "- schedulerName: " + name + "\n"
+	}
+	return []byte(schedulerConfig)
 }
--- a/cluster-autoscaler/core/static_autoscaler_test.go
+++ b/cluster-autoscaler/core/static_autoscaler_test.go
@@ -21,9 +21,7 @@ import (
 	stdcontext "context"
 	"flag"
 	"fmt"
-	testconfig "k8s.io/autoscaler/cluster-autoscaler/config/test"
 	"os"
-	"path/filepath"
 	"reflect"
 	"strings"
 	"testing"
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -1452,6 +1454,22 @@ func TestStaticAutoscalerRunOnceWithUnselectedNodeGroups(t *testing.T) {
 func TestStaticAutoscalerRunOnceWithBypassedSchedulers(t *testing.T) {
 	bypassedScheduler := "bypassed-scheduler"
 	nonBypassedScheduler := "non-bypassed-scheduler"
+
+	tmpDir, err := os.MkdirTemp("", "scheduler-configs")
```
MaciekPytel, line 1458: This seems to copy-paste logic used in schedulerbased_test.go, can we move it to a single utility function (for example in config/test/test_utils.go)?

cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
```diff
@@ -53,15 +56,13 @@ func NewSchedulerBasedPredicateChecker(informerFactory informers.SharedInformerF
...
 		return nil, fmt.Errorf("unexpected scheduler config: expected one scheduler profile only (found %d profiles)", len(schedConfig.Profiles))
 	}
 	sharedLister := NewDelegatingSchedulerSharedLister()
 
-	framework, err := schedulerframeworkruntime.NewFramework(
-		context.TODO(),
-		scheduler_plugins.NewInTreeRegistry(),
-		&schedConfig.Profiles[0],
+	recorderFactory := func(string) events.EventRecorder { return nil }
+	profiles, err := scheduler_profile.NewMap(context.TODO(), schedConfig.Profiles, scheduler_plugins.NewInTreeRegistry(), recorderFactory,
```
MaciekPytel, line 65: nit: can we make line breaking more consistent?

cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
```diff
@@ -71,7 +72,8 @@ func NewSchedulerBasedPredicateChecker(informerFactory informers.SharedInformerF
 	}
 
 	checker := &SchedulerBasedPredicateChecker{
-		framework:              framework,
+		profiles:               profiles,
+		defaultSchedulerName:   schedConfig.Profiles[0].SchedulerName,
```
MaciekPytel, line 76: Is it a requirement for the first config in file to be the default one? If not this seems very brittle.

cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
```diff
@@ -104,8 +106,16 @@ func (p *SchedulerBasedPredicateChecker) FitsAnyNodeMatching(clusterSnapshot clu
 	p.delegatingSharedLister.UpdateDelegate(clusterSnapshot)
 	defer p.delegatingSharedLister.ResetDelegate()
 
+	fwk, err := p.frameworkForPod(pod)
+	if err != nil {
+		// This shouldn't happen, because we only accept for scheduling the pods
```
MaciekPytel, line 111: What does that mean? Is there an admission controller built-into kubernetes that prevents users from creating pods with arbitrary names. What if someone has a custom scheduler that is not a profile of kubernetes scheduler, but a completely custom binary? Today we would still scale-up for them - potentially using different criteria, but maybe they've already accounted for that. Isn't this a breaking change?

Wouldn't it be more backward compatible to return default scheduler if scheduler name is not found?

MaciekPytel:
Overall this looks good to me, modulo the backward compatibility aspect of failing on unknown scheduler names. @alculquicondor I wonder if you would mind doing a quick sanity check on changes in schedulerbased.go to see if there are any potential problems with initializing framework that way from CA? I don't see anything, but you have much more expertise here.

cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
```diff
@@ -184,6 +200,21 @@ func (p *SchedulerBasedPredicateChecker) CheckPredicates(clusterSnapshot cluster
...
+	if !ok {
+		return nil, fmt.Errorf("profile not found for scheduler name %q", pod.Spec.SchedulerName)
+	}
+	return fwk, nil
+}
+
+func (p *SchedulerBasedPredicateChecker) schedulerNameForPod(pod *apiv1.Pod) string {
+	if len(pod.Spec.SchedulerName) == 0 {
+		return p.defaultSchedulerName
+	}
```
alculquicondor, line 214: This is not useful. The apiserver always adds a default.

cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
alculquicondor: > Is there an admission controller built-into kubernetes that prevents users from creating pods with arbitrary name

No.

> What if someone has a custom scheduler that is not a profile of kubernetes scheduler, but a completely custom binary? Today we would still scale-up for them

Also cases where people have multiple kube-scheduler profiles (for example, just to have different scoring strategies). These users would suddenly be broken unless they create all the same profiles for their cluster autoscaler instance.

> Wouldn't it be more backward compatible to return default scheduler if scheduler name is not found?

I agree. Maybe just the first profile from the list.

cluster-autoscaler/core/static_autoscaler_test.go
daimaxiaxie: I restored this unit test.

cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
daimaxiaxie: ok

cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
daimaxiaxie: use profiles[0] as default, check len in line 59.

cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
daimaxiaxie: ok, I made the modifications as requested.

cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
daimaxiaxie: ok

cluster-autoscaler/simulator/predicatechecker/schedulerbased.go
```diff
@@ -184,6 +198,16 @@ func (p *SchedulerBasedPredicateChecker) CheckPredicates(clusterSnapshot cluster
 	return nil
 }
 
+func (p *SchedulerBasedPredicateChecker) frameworkForPod(pod *apiv1.Pod) schedulerframework.Framework {
+	fwk, ok := p.profiles[pod.Spec.SchedulerName]
+	if !ok {
+		klog.Warningf("profile not found for scheduler name %q", pod.Spec.SchedulerName)
```
alculquicondor, line 204: Avoid the use of `Warning`

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#change-log-functions

daimaxiaxie: use `InfoS`

x13n:
/assign @MaciekPytel 

Looks like you were reviewing this already, can you follow up? Is it ok to merge now?

PR #6547, opened 2024-02-20
Migrate GCE client to server-side wait

mtrqq:
#### What type of PR is this?

/kind feature

[GCE CloudProvider](cluster-autoscaler/cloudprovider/gce) is currently using polling as a primary method for awaiting async operations. GCE now provides [method for server side wait](https://cloud.google.com/compute/docs/reference/rest/v1/zoneOperations/wait), which will decrease RPS compared to polling approach. 

1. WaitForOperation method becomes part of public interface to allow further extension of interface without reimplementing await logic once again;
2. In case we are sure that timeouts more than 2 minutes are required we may combine wait method with polling as the last resort.

```release-note
NONE
```

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
+++ b/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
@@ -37,9 +37,8 @@ import (
 )
 
 const (
-	defaultOperationWaitTimeout          = 20 * time.Second
-	defaultOperationPollInterval         = 100 * time.Millisecond
-	defaultOperationDeletionPollInterval = 1 * time.Second
+	defaultOperationWaitTimeout  = 20 * time.Second
+	defaultOperationPollInterval = 100 * time.Millisecond
 	// ErrorCodeQuotaExceeded is an error code used in InstanceErrorInfo if quota exceeded error occurs.
 	ErrorCodeQuotaExceeded = "QUOTA_EXCEEDED"
 
@@ -115,6 +114,11 @@ type AutoscalingGceClient interface {
 	ResizeMig(GceRef, int64) error
 	DeleteInstances(migRef GceRef, instances []GceRef) error
 	CreateInstances(GceRef, string, int64, []string) error
+
+	// WaitForOperation can be used to poll GCE operations until completion/timeout using WAIT calls.
+	// Calling this is normally not needed when interacting with the client, other methods should call it internally.
+	// Can be used to extend the interface with more methods outside of this package.
+	WaitForOperation(operationName, operationType, project, zone string) error
 }
 
 type autoscalingGceClientV1 struct {
@@ -123,39 +127,36 @@ type autoscalingGceClientV1 struct {
 	projectId string
 	domainUrl string
 
-	// These can be overridden, e.g. for testing.
-	operationWaitTimeout          time.Duration
-	operationPollInterval         time.Duration
-	operationDeletionPollInterval time.Duration
+	// Can be overridden, e.g. for testing.
+	operationWaitTimeout  time.Duration
+	operationPollInterval time.Duration
 }
 
 // NewAutoscalingGceClientV1WithTimeout creates a new client with custom timeouts
 // for communicating with GCE v1 API
-func NewAutoscalingGceClientV1WithTimeout(client *http.Client, projectId string, userAgent string,
-	waitTimeout, pollInterval, deletionPollInterval time.Duration) (*autoscalingGceClientV1, error) {
+func NewAutoscalingGceClientV1WithTimeout(client *http.Client, projectId string, userAgent string, waitTimeout time.Duration, pollInterval time.Duration) (*autoscalingGceClientV1, error) {
 	gceService, err := gce.New(client)
 	if err != nil {
 		return nil, err
 	}
 	gceService.UserAgent = userAgent
+
 	return &autoscalingGceClientV1{
-		projectId:                     projectId,
-		gceService:                    gceService,
-		operationWaitTimeout:          waitTimeout,
-		operationPollInterval:         pollInterval,
-		operationDeletionPollInterval: deletionPollInterval,
+		projectId:             projectId,
+		gceService:            gceService,
+		operationWaitTimeout:  waitTimeout,
+		operationPollInterval: pollInterval,
 	}, nil
 }
 
 // NewAutoscalingGceClientV1 creates a new client for communicating with GCE v1 API.
 func NewAutoscalingGceClientV1(client *http.Client, projectId string, userAgent string) (*autoscalingGceClientV1, error) {
-	return NewAutoscalingGceClientV1WithTimeout(client, projectId, userAgent, defaultOperationWaitTimeout, defaultOperationPollInterval, defaultOperationDeletionPollInterval)
+	return NewAutoscalingGceClientV1WithTimeout(client, projectId, userAgent, defaultOperationWaitTimeout, defaultOperationPollInterval)
 }
 
 // NewCustomAutoscalingGceClientV1 creates a new client using custom server url and timeouts
 // for communicating with GCE v1 API.
-func NewCustomAutoscalingGceClientV1(client *http.Client, projectId, serverUrl, userAgent, domainUrl string,
-	waitTimeout, pollInterval time.Duration, deletionPollInterval time.Duration) (*autoscalingGceClientV1, error) {
+func NewCustomAutoscalingGceClientV1(client *http.Client, projectId, serverUrl, userAgent, domainUrl string, waitTimeout time.Duration, pollInterval time.Duration) (*autoscalingGceClientV1, error) {
 	gceService, err := gce.New(client)
 	if err != nil {
 		return nil, err
@@ -164,12 +165,11 @@ func NewCustomAutoscalingGceClientV1(client *http.Client, projectId, serverUrl, 
 	gceService.UserAgent = userAgent
 
 	return &autoscalingGceClientV1{
-		projectId:                     projectId,
-		gceService:                    gceService,
-		domainUrl:                     domainUrl,
-		operationWaitTimeout:          waitTimeout,
-		operationPollInterval:         pollInterval,
-		operationDeletionPollInterval: deletionPollInterval,
+		projectId:             projectId,
+		gceService:            gceService,
+		domainUrl:             domainUrl,
+		operationWaitTimeout:  waitTimeout,
+		operationPollInterval: pollInterval,
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -115,6 +115,9 @@ type AutoscalingGceClient interface {
 	ResizeMig(GceRef, int64) error
 	DeleteInstances(migRef GceRef, instances []GceRef) error
 	CreateInstances(GceRef, string, int64, []string) error
+
+	// extension point for new operations
```
towca, line 119: Not sure I understand this comment

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -274,32 +270,31 @@ func instanceIdsToNamesMap(instanceProviderIds []string) map[string]bool {
...
-			if op.Status == "DONE" {
-				if op.Error != nil {
-					errBytes, err := op.Error.MarshalJSON()
-					if err != nil {
-						errBytes = []byte(fmt.Sprintf("operation failed, but error couldn't be recovered: %v", err))
-					}
-					return fmt.Errorf("error while getting operation %s on %s: %s", operation.Name, operation.TargetLink, errBytes)
-				}
+func (client *autoscalingGceClientV1) WaitForOperation(operationName, operationType, project, zone string) error {
+	ctx, cancel := context.WithTimeout(context.TODO(), client.operationWaitTimeout)
```
towca, line 274: I don't think the cancel works - it's only called in the defer when exiting the function, but we'd never reach the exit of the function if we e.g. got stuck in the for loop below. The cancel should be issued after `client.operationWaitTimeout` passes since entering the method, you can use `time.After(client.operationWaitTimeout)` for that.

The cleanest solution that comes to mind would be to extract the body of the for loop into a separate function that takes a channel, and returns the status and error through that channel. Start that function in a goroutine here, and then do a `select` on either the result channel (in which case either return if DONE/err!=nil, or trigger another call of the extracted function otherwise), or the channel returned from `time.After` (in which case call cancel and return). WDYT?

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -274,32 +270,31 @@ func instanceIdsToNamesMap(instanceProviderIds []string) map[string]bool {
...
+	defer cancel()
+
+	for {
```
towca, line 277: Reading the documentation, it seems that if the server is overloaded, the Wait call can return immediately. I'd still keep some interval between retries for this scenario, but IMO it doesn't make sense to differentiate the timeouts between different operations anymore.

towca (changes requested):

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
mtrqq: The cancel function is here only to release the resources, all the timeouts are being handled inside `context.WithTimeout` as they are wrapping parent context into a new one with timer which executes cancel function by itself after some period of time.

I've also checked the internals of `Do()` call and they are also accounting for cancelled channel, so I think we are safe from this perspective. We of course may add select statement with `time.After` and `ctx.Done` and it shouldn't affect performance anyhow, but it's redundant from my perspective

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
mtrqq: Fully agree with that, thanks for mentioning! Yeah, I don't think that it makes sense to make it configurable even, but the tests may struggle from that potentially. 

The only thing that bothers me is that there's no mention in docs how to identify that server is overloaded based on response (so that sleep would be called only in this scenario)

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
mtrqq: Yeah, it's very ambiguous to me as well, that basically means that this piece is only exported in order not to reimplement this method in case of interface extension. Do you have any ideas how to replace it?

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
mtrqq: As a result - added `operationPollInterval`, but made it not configurable from outside of the package, so that only tests may access that for modification

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
towca: You're right, I misunderstood the `WithTimeout` method. This should work, could you just change `context.TODO()` to `context.Background()`? In this case we want a fresh, empty context, `TODO()` is more for cases where some context should actually be propagated to this function, but isn't yet.

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
mtrqq: Awesome, the comment seems good from my side, thanks for help! I think that it makes more sense to make it part of the GCE client itself because  semantically it's still a method of a client, even if it has fully utility functionality. 

Also the downsides of this approach that we either need to pass client itself as a parameter (which is relatively same as making it a receiver) or pass http client, poll interval and wait timeout and the parameters we are already passing, which to me seems a bit messy. I personally lean towards leaving it as a receiver function, is that okay from your side?

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
mtrqq: Totally makes sense, changing it from TODO to Background. I personally would expect all the client methods to accept context as the first parameter, but it's probably the topic for the other PR

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -274,32 +272,34 @@ func instanceIdsToNamesMap(instanceProviderIds []string) map[string]bool {
...
+		klog.V(4).Infof("Waiting for operation %s (%s/%s/%s)", operationName, project, zone, operationType)
```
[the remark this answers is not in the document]
mtrqq: Definitely makes sense, done

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client_test.go
```diff
@@ -110,11 +110,10 @@ func TestWaitForOpError(t *testing.T) {
...
 	g := newTestAutoscalingGceClient(t, "project1", server.URL, "")
 
-	server.On("handle", "/projects/project1/zones/us-central1-b/operations/operation-1505728466148-d16f5197").Return(operationDoneResponseError).Once()
-
+	server.On("handle", "/projects/project1/zones/us-central1-b/operations/operation-1505728466148-d16f5197/wait").Return(operationDoneResponseError).Once()
 	operation := &gce_api.Operation{Name: "operation-1505728466148-d16f5197"}
 
-	err := g.waitForOp(operation, projectId, zoneB, false)
+	err := g.WaitForOperation(operation.Name, "TestWaitForOpError", projectId, zoneB)
 	assert.Error(t, err)
```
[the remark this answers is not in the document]
mtrqq: sure, added this call to this test and some others

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client_test.go
```diff
@@ -123,21 +122,28 @@ func TestWaitForOpTimeout(t *testing.T) {
...
 
-	// The values here are higher than in other tests since we're aiming for timeout.
-	// Lower values make this fragile and flakey.
-	g.operationPollInterval = 10 * time.Millisecond
-	g.operationWaitTimeout = 49 * time.Millisecond
-
-	// Sometimes, only 3 calls are made, but it doesn't really matter,
-	// so let's not assert expectations for this mock, just check for timeout error.
-	server.On("handle", "/projects/project1/zones/us-central1-b/operations/operation-1505728466148-d16f5197").Return(operationRunningResponse).Times(5)
+	server.On("handle", "/projects/project1/zones/us-central1-b/operations/operation-1505728466148-d16f5197/wait").Return(operationRunningResponse).Once()
```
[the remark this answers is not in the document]
mtrqq: Fully agree, decreased intervals down to 1 ms, mock is not complaining about that, also removed Once() spec just not to bring more confuse

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client_test.go
```diff
@@ -110,32 +109,41 @@ func TestWaitForOpError(t *testing.T) {
...
-	// The values here are higher than in other tests since we're aiming for timeout.
-	// Lower values make this fragile and flakey.
-	g.operationPollInterval = 10 * time.Millisecond
-	g.operationWaitTimeout = 49 * time.Millisecond
-
-	// Sometimes, only 3 calls are made, but it doesn't really matter,
-	// so let's not assert expectations for this mock, just check for timeout error.
-	server.On("handle", "/projects/project1/zones/us-central1-b/operations/operation-1505728466148-d16f5197").Return(operationRunningResponse).Times(5)
+	// default polling interval and wait time are too big for the test
+	g.operationWaitTimeout = 1 * time.Millisecond
```
towca, line 125: With the timeout being this low, the test looks even more identical to `TestWaitForOpContextTimeout` - I think in both we're testing if timing out the API call via context works correctly (since 1ms likely isn't enough to obtain the result, so the call will just fail). Or am I not seeing some difference here?

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client_test.go
```diff
@@ -110,32 +109,41 @@ func TestWaitForOpError(t *testing.T) {
...
 	assert.Error(t, err)
+	mock.AssertExpectationsForObjects(t, server)
+}
+
+func TestWaitForOpContextTimeout(t *testing.T) {
+	server := test_util.NewHttpServerMock()
+	defer server.Close()
+	g := newTestAutoscalingGceClient(t, "project1", server.URL, "")
+
+	g.operationWaitTimeout = 1 * time.Millisecond
```
towca, line 140: I'd make this a bit higher, 1ms probably means that the context is cancelled immediately, before even reaching the API call (and increase the value in `After` below proportionally).

mtrqq: Agree, increased timeout to 10 ms and After to 1 minute in order to better reflect that we are testing context timeout during API call

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client_test.go
mtrqq: I think you are right that such low timeout values actually make them look the same, I've adjusted those slightly to be more easily differentiated. The main difference between these tests is that `TestWaitForOpContextTimeout` should never achieve http client receiving response (to better reflect that - I've added a minute delay and a comment), TestWaitForOpContextTimeout should receive the response and respond with an error that operation is still running

PR #6548, opened 2024-02-20
extract create group to a function

dsafdsa1:
#### What type of PR is this?
/kind feature

It extracts node group creation into a separate function, enabling its reuse in a custom cloud provider orchestrator.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
+++ b/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
@@ -194,48 +194,11 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 
 	createNodeGroupResults := make([]nodegroups.CreateNodeGroupResult, 0)
 	if !bestOption.NodeGroup.Exist() {
-		oldId := bestOption.NodeGroup.Id()
-		createNodeGroupResult, aErr := o.processors.NodeGroupManager.CreateNodeGroup(o.autoscalingContext, bestOption.NodeGroup)
+		var scaleUpStatus *status.ScaleUpStatus
+		createNodeGroupResults, scaleUpStatus, aErr = o.CreateNodeGroup(bestOption, nodeInfos, schedulablePods, podEquivalenceGroups, daemonSets)
 		if aErr != nil {
-			return status.UpdateScaleUpError(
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -194,48 +194,11 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 
 	createNodeGroupResults := make([]nodegroups.CreateNodeGroupResult, 0)
 	if !bestOption.NodeGroup.Exist() {
-		oldId := bestOption.NodeGroup.Id()
-		createNodeGroupResult, aErr := o.processors.NodeGroupManager.CreateNodeGroup(o.autoscalingContext, bestOption.NodeGroup)
+		var scaleUpStatus *status.ScaleUpStatus
+		createNodeGroupResults, scaleUpStatus, aErr = o.CreateNodeGroup(bestOption, nodeInfos, schedulablePods, podEquivalenceGroups, daemonSets)
```
jayantjain93, line 198: nit: It would be a nice to have if we can separate some unit test which are core to the new refactored logic to directly call `CreateNodeGroup`, and leave the rest as is.

this will decouple the testing also internal bits of the new func.

dsafdsa1: The unit tests that check this path are [TestScaleUpAutoprovisionedNodeGroup](https://github.com/kubernetes/autoscaler/blob/a3c8978e119f56a90a6f8406e76685c21cfecd51/cluster-autoscaler/core/scaleup/orchestrator/orchestrator_test.go#L1421) and  [TestScaleUpBalanceAutoprovisionedNodeGroups](https://github.com/kubernetes/autoscaler/blob/a3c8978e119f56a90a6f8406e76685c21cfecd51/cluster-autoscaler/core/scaleup/orchestrator/orchestrator_test.go#L1473). I would prefer to leave them as they are currently, because they test both the creation and the scale up.

jayantjain93: ack.

PR #6552, opened 2024-02-21
Add BuildTestNodeWithAllocatable test utility method.

walidghallab:
#### What type of PR is this?

/kind cleanup

Adding a way to create test node with allocatable. Currently the existing method only supports specifying capacity.

```release-note
None
```

```docs

```

### DIFF
--- a/cluster-autoscaler/utils/test/test_utils.go
+++ b/cluster-autoscaler/utils/test/test_utils.go
@@ -212,7 +212,7 @@ func TolerateGpuForPod(pod *apiv1.Pod) {
 }
 
 // BuildTestNode creates a node with specified capacity.
-func BuildTestNode(name string, millicpu int64, mem int64) *apiv1.Node {
+func BuildTestNode(name string, millicpuCapacity int64, memCapacity int64) *apiv1.Node {
 	node := &apiv1.Node{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:     name,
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/utils/test/test_utils.go
```diff
@@ -213,6 +213,12 @@ func TolerateGpuForPod(pod *apiv1.Pod) {
 
 // BuildTestNode creates a node with specified capacity.
 func BuildTestNode(name string, millicpu int64, mem int64) *apiv1.Node {
+	return BuildTestNodeWithAllocatable(name, millicpu, mem, millicpu, mem)
+}
+
+// BuildTestNodeWithAllocatable creates a node with specified capacity and allocatable.
+// When allocatable is negative number, the allocatable will be the same as capacity.
```
BigDarkClown, line 220: I don't like this defaulting - test util functions should be as simple as possible, and this adds another layer. If I have to understand that negative values mean that capacity == allocatable, then I can just pass this value explicitely in test, making it more readable.

cluster-autoscaler/utils/test/test_utils.go
```diff
@@ -213,6 +213,12 @@ func TolerateGpuForPod(pod *apiv1.Pod) {
...
+func BuildTestNodeWithAllocatable(name string, millicpuCapacity, memCapacity, millicpuAllocatable, memAllocatable int64) *apiv1.Node {
```
BigDarkClown, line 221: With so many arguments, all of the same type, it is easy to mess up by accidentally reordering them.

IMHO opinion the best way to implement this to be as readable as possible would be to use a builder pattern. Something like `BuildTestNode("n", 100, 1000).WithAllocatable(50, 500).Build()` is much easier to understand than `BuildTestNodeWithAllocatable("n", 100, 1000, 50, 500)`. This would be my preferred approach - you could go even further by creating more `WithXYZ(x)` functions for better granularity.

Other option would be to create a wrapper `WithAllocatable(node *apiv1.Node, millicpu int64, mem int64) *apiv1.Node`. Gives better readability and does not compromise on the expressiveness.

Overall, I am not going to block on this, as this change is very small. But if you have 15 minutes free - I think it would greatly improve the readability and prevent refactors in the future :)

PR #6567, opened 2024-02-26
CA FAQ: clarify the point about scheduling constraints blocking scale-down

towca:
Prompted by #6208

### DIFF
--- a/cluster-autoscaler/FAQ.md
+++ b/cluster-autoscaler/FAQ.md
@@ -96,8 +96,13 @@ Cluster Autoscaler decreases the size of the cluster when some nodes are consist
       "cluster-autoscaler.kubernetes.io/safe-to-evict-local-volumes": "volume-1,volume-2,.."
       ```
       and all of the pod's local volumes are listed in the annotation value.
-* Pods that cannot be moved elsewhere due to various constraints (lack of resources, non-matching node selectors or affinity,
-matching anti-affinity, etc)
+* Pods that cannot be moved elsewhere due to scheduling constraints. CA simulates kube-scheduler behavior, and if there's no other node where a given pod can schedule, the pod's node won't be scaled down.
+  * This can be particularly visible if a given workloads' pods are configured to only fit one pod per node on some subset of nodes. Such pods will always block CA from scaling down their nodes, because all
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/FAQ.md
```diff
@@ -96,8 +96,13 @@ Cluster Autoscaler decreases the size of the cluster when some nodes are consist
       "cluster-autoscaler.kubernetes.io/safe-to-evict-local-volumes": "volume-1,volume-2,.."
       ```
       and all of the pod's local volumes are listed in the annotation value.
-* Pods that cannot be moved elsewhere due to various constraints (lack of resources, non-matching node selectors or affinity,
-matching anti-affinity, etc)
+* Pods that cannot be moved elsewhere due to scheduling constraints. CA simulates kube-scheduler behavior, and if there's no other node where a given pod can schedule, the pod's node won't be scaled down.
```
Shubham82, line 99:
```suggestion
* Pods that cannot be moved elsewhere due to scheduling constraints. CA simulates
kube-scheduler behavior, and if there's no other node where a given pod can schedule, the
pod's node won't be scaled down.
```

cluster-autoscaler/FAQ.md
```diff
@@ -96,8 +96,13 @@ Cluster Autoscaler decreases the size of the cluster when some nodes are consist
...
+  * This can be particularly visible if a given workloads' pods are configured to only fit one pod per node on some subset of nodes. Such pods will always block CA from scaling down their nodes, because all
```
Shubham82, line 100:
```suggestion
  * This can be particularly visible if a given workloads' pods are configured to only fit
  one pod per node on some subset of nodes. Such pods will always block CA from scaling
  down their nodes, because all
```

Shubham82 (changes requested):

cluster-autoscaler/FAQ.md
Shubham82: Just small nit: Wrap the line

cluster-autoscaler/FAQ.md
Shubham82: nit: Wrap the line

PR #6570, opened 2024-02-27
Regional instance

Edwinhr716:
#### What type of PR is this?

/kind feature
Need to support Regional Instance Templates. TWithout this change, whenever FetchMIGTemplate gets called, it will fail to find the instance template that is attached to the MIG, as it will call the global GCE API instead of regional one. 

Fixes #

NONE
```release-note
```

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
+++ b/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
@@ -595,7 +595,7 @@ func (client *autoscalingGceClientV1) FetchMigTemplateName(migRef GceRef) (Insta
 	if err != nil {
 		return InstanceTemplateName{}, err
 	}
-	regional, err := regexp.MatchString("(/projects/.*[A-Za-z0-9]+.*/regions/)", templateUrl.String())
+	regional, err := IsInstanceTemplateRegional(templateUrl.String())
 	if err != nil {
 		return InstanceTemplateName{}, err
 	}
--- a/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
+++ b/cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
@@ -103,8 +103,8 @@ type AutoscalingGceClient interface {
 	FetchMigTargetSize(GceRef) (int64, error)
 	FetchMigBasename(GceRef) (string, error)
 	FetchMigInstances(GceRef) ([]GceInstance, error)
-	FetchMigTemplateName(migRef GceRef) (string, error)
-	FetchMigTemplate(migRef GceRef, templateName string) (*gce.InstanceTemplate, error)
+	FetchMigTemplateName(migRef GceRef) (InstanceTemplateNameType, error)
+	FetchMigTemplate(migRef GceRef, templateName string, regional bool) (*gce.InstanceTemplate, error)
 	FetchMigsWithName(zone string, filter *regexp.Regexp) ([]string, error)
 	FetchZones(region string) ([]string, error)
 	FetchAvailableCpuPlatforms() (map[string][]string, error)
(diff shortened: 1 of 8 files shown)


cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -33,6 +33,13 @@ type MachineTypeKey struct {
 	MachineTypeName string
 }
 
+// InstanceTemplateNameType is used to store the name, and
+// whether or not the instance template is regional
+type InstanceTemplateNameType struct {
```
BigDarkClown, line 38: Please remove the `Type` suffix, it is not necessary here.

cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
```diff
@@ -239,44 +240,44 @@ func (c *cachingMigInfoProvider) GetMigBasename(migRef GceRef) (string, error) {
...
 func (c *cachingMigInfoProvider) GetMigInstanceTemplate(migRef GceRef) (*gce.InstanceTemplate, error) {
-	templateName, err := c.GetMigInstanceTemplateName(migRef)
+	instanceTemplateNameType, err := c.GetMigInstanceTemplateName(migRef)
 	if err != nil {
 		return nil, err
 	}
 
 	template, found := c.cache.GetMigInstanceTemplate(migRef)
-	if found && template.Name == templateName {
+	if found && template.Name == instanceTemplateNameType.Name {
```
BigDarkClown, line 275: Is it enough now that the 2 templates have the same name to determine that they are the same? IIUC the change the logic should now be:
- If template is global, then check if the names of the templates are the same
- if template is regional, then check if both the names and the regions of the templates are the same

cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
```diff
@@ -43,7 +44,7 @@ type MigInfoProvider interface {
 	// GetMigBasename returns basename for given MIG ref
 	GetMigBasename(migRef GceRef) (string, error)
 	// GetMigInstanceTemplateName returns instance template name for given MIG ref
-	GetMigInstanceTemplateName(migRef GceRef) (string, error)
+	GetMigInstanceTemplateName(migRef GceRef) (InstanceTemplateNameType, error)
 	// GetMigInstanceTemplate returns instance template for given MIG ref
 	GetMigInstanceTemplate(migRef GceRef) (*gce.InstanceTemplate, error)
```
BigDarkClown, line 49: Please rebase your changes on the current master. We've recently added new logic around caching the kube-env for templates (https://github.com/kubernetes/autoscaler/pull/6531) which also relies on storing the template naming to work properly.

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
```diff
@@ -565,26 +565,33 @@ func (client *autoscalingGceClientV1) FetchAvailableCpuPlatforms() (map[string][
...
 			if err.Code == http.StatusNotFound {
-				return "", errors.NewAutoscalerError(errors.NodeGroupDoesNotExistError, "%s", err.Error())
+				return InstanceTemplateNameType{}, errors.NewAutoscalerError(errors.NodeGroupDoesNotExistError, "%s", err.Error())
 			}
 		}
-		return "", err
+		return InstanceTemplateNameType{}, err
 	}
 	templateUrl, err := url.Parse(igm.InstanceTemplate)
+	regional, _ := regexp.MatchString("(/projects/.*[A-Za-z0-9]+.*/regions/)", templateUrl.String())
```
BigDarkClown, line 580: Missing error handling, please add. Also add an unit test covering this.

cluster-autoscaler/cloudprovider/gce/cache.go
```diff
@@ -33,6 +33,13 @@ type MachineTypeKey struct {
...
+	Name     string
+	Regional bool
```
BigDarkClown, line 40: I think you will need to store the region here as well, otherwise it is not possible to discern if 2 template names are the same. Please also add a dedicated `Equal` function that compares 2 template names.

BigDarkClown (commented):
Initial review done. Please rebase on master and fix the issues, then ping me to take another look.

cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
Edwinhr716: Done

cluster-autoscaler/cloudprovider/gce/cache.go
Edwinhr716: Will add the Equal function. No need to store the region, as the instance template name is unique. So two instance templates of the same nodepool but of different regions will have different names.

cluster-autoscaler/cloudprovider/gce/autoscaling_gce_client.go
Edwinhr716: IS there an existing test for FetchMigTemplate or should I create one from scratch? I couldn't find one in autoscaling_gce_client_test.go

cluster-autoscaler/cloudprovider/gce/mig_info_provider.go
```diff
@@ -360,7 +361,8 @@ func (c *cachingMigInfoProvider) fillMigInfoCache() error {
 				templateUrl, err := url.Parse(zoneMig.InstanceTemplate)
 				if err == nil {
 					_, templateName := path.Split(templateUrl.EscapedPath())
-					c.cache.SetMigInstanceTemplateName(zoneMigRef, templateName)
+					regional, _ := regexp.MatchString("(/projects/.*[A-Za-z0-9]+.*/regions/)", templateUrl.String())
```
BigDarkClown, line 364: Please put this regexp matching into a helper function, it is used in more than one place.

Edwinhr716: Done

cluster-autoscaler/cloudprovider/gce/mig_info_provider_test.go
```diff
@@ -52,8 +52,8 @@ type mockAutoscalingGceClient struct {
 	fetchMigTargetSize   func(GceRef) (int64, error)
 	fetchMigBasename     func(GceRef) (string, error)
 	fetchMigInstances    func(GceRef) ([]GceInstance, error)
-	fetchMigTemplateName func(GceRef) (string, error)
-	fetchMigTemplate     func(GceRef, string) (*gce.InstanceTemplate, error)
+	fetchMigTemplateName func(GceRef) (InstanceTemplateName, error)
+	fetchMigTemplate     func(GceRef, string, bool) (*gce.InstanceTemplate, error)
 	fetchMachineType     func(string, string) (*gce.MachineType, error)
 }
```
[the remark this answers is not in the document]
Edwinhr716: done

PR #6575, opened 2024-02-27
Refactor StartDeletion usage patterns and enforce periodic scaledown

mtrqq:
#### What type of PR is this?

/cleanup

This PR does 2 things:
* Removes responsibility from Actuator.StartDeletion to create full ScaleDownStatus object. It allows us to remove some semantically incorrect usage patterns like [this](https://github.com/kubernetes/autoscaler/compare/master...mtrqq:feat/periodic-scaledown-status-calls?expand=1#diff-a7aace3350322c394f2546d1c4f098372716d343f4883ded0e3aaed75c820cfcL1283) and [this](https://github.com/kubernetes/autoscaler/compare/master...mtrqq:feat/periodic-scaledown-status-calls?expand=1#diff-b1ecfd03ab78ce8627f09152ca317b1feaf5003297f00fbac5ff77710af6be66L99) and instead query for status object explicitly via Actuator.ScaleDownStatus call
* Enforces periodic checks of ScaleDownStatusProcessor, because as of right now - it can be called twice in case of cooldown period due to [scaleDownStatusProcessorAlreadyCalled](https://github.com/kubernetes/autoscaler/compare/master...mtrqq:feat/periodic-scaledown-status-calls?expand=1#diff-5d3331e59ebce666578d33c0c52f7686b92ff18bcab405eb8eab637881f78080L391) not being set in the case of scale down cooldown. Generally, the logic of explicit status processor calls around the loop itself was removed in favor of the periodic check in a [defer statement](https://github.com/kubernetes/autoscaler/compare/master...mtrqq:feat/periodic-scaledown-status-calls?expand=1#diff-5d3331e59ebce666578d33c0c52f7686b92ff18bcab405eb8eab637881f78080R410)

* Please give a special attention to legacy wrapper around Actuator because we are *changing interface quite significantly*
* Some tests were not completely changed from using partial ScaleDownStatus objects as it brings as bit more readability (arguable)

```release-note
NONE
```

### DIFF
--- a/cluster-autoscaler/core/scaledown/actuation/actuator.go
+++ b/cluster-autoscaler/core/scaledown/actuation/actuator.go
@@ -96,47 +96,47 @@ func (a *Actuator) ClearResultsNotNewerThan(t time.Time) {
 	a.nodeDeletionTracker.ClearResultsNotNewerThan(t)
 }
 
+// DeletionResults returns deletion results since the last ClearResultsNotNewerThan call
+// in a map form, along with the timestamp of last result.
+func (a *Actuator) DeletionResults() (map[string]status.NodeDeleteResult, time.Time) {
+	return a.nodeDeletionTracker.DeletionResults()
+}
+
 // StartDeletion triggers a new deletion process.
-func (a *Actuator) StartDeletion(empty, drain []*apiv1.Node) (*status.ScaleDownStatus, errors.AutoscalerError) {
+func (a *Actuator) StartDeletion(empty, drain []*apiv1.Node) (status.ScaleDownResult, []*status.ScaleDownNode, errors.AutoscalerError) {
 	a.nodeDeletionScheduler.ResetAndReportMetrics()
 	deletionStartTime := time.Now()
 	defer func() { metrics.UpdateDuration(metrics.ScaleDownNodeDeletion, time.Since(deletionStartTime)) }()
 
-	results, ts := a.nodeDeletionTracker.DeletionResults()
-	scaleDownStatus := &status.ScaleDownStatus{NodeDeleteResults: results, NodeDeleteResultsAsOf: ts}
-
+	scaledDownNodes := make([]*status.ScaleDownNode, 0)
 	emptyToDelete, drainToDelete := a.budgetProcessor.CropNodes(a.nodeDeletionTracker, empty, drain)
 	if len(emptyToDelete) == 0 && len(drainToDelete) == 0 {
-		scaleDownStatus.Result = status.ScaleDownNoNodeDeleted
-		return scaleDownStatus, nil
+		return status.ScaleDownNoNodeDeleted, nil, nil
 	}
 
 	if len(emptyToDelete) > 0 {
 		// Taint all empty nodes synchronously
 		if err := a.taintNodesSync(emptyToDelete); err != nil {
-			scaleDownStatus.Result = status.ScaleDownError
-			return scaleDownStatus, err
+			return status.ScaleDownError, scaledDownNodes, err
 		}
 
 		emptyScaledDown := a.deleteAsyncEmpty(emptyToDelete)
-		scaleDownStatus.ScaledDownNodes = append(scaleDownStatus.ScaledDownNodes, emptyScaledDown...)
+		scaledDownNodes = append(scaledDownNodes, emptyScaledDown...)
 	}
 
 	if len(drainToDelete) > 0 {
 		// Taint all nodes that need drain synchronously, but don't start any drain/deletion yet. Otherwise, pods evicted from one to-be-deleted node
 		// could get recreated on another.
 		if err := a.taintNodesSync(drainToDelete); err != nil {
-			scaleDownStatus.Result = status.ScaleDownError
-			return scaleDownStatus, err
+			return status.ScaleDownError, scaledDownNodes, err
 		}
 
 		// All nodes involved in the scale-down should be tainted now - start draining and deleting nodes asynchronously.
 		drainScaledDown := a.deleteAsyncDrain(drainToDelete)
-		scaleDownStatus.ScaledDownNodes = append(scaleDownStatus.ScaledDownNodes, drainScaledDown...)
+		scaledDownNodes = append(scaledDownNodes, drainScaledDown...)
 	}
 
-	scaleDownStatus.Result = status.ScaleDownNodeDeleteStarted
-	return scaleDownStatus, nil
+	return status.ScaleDownNodeDeleteStarted, scaledDownNodes, nil
 }
 
 // deleteAsyncEmpty immediately starts deletions asynchronously.
--- a/cluster-autoscaler/core/scaledown/scaledown.go
+++ b/cluster-autoscaler/core/scaledown/scaledown.go
@@ -56,12 +56,15 @@ type Actuator interface {
 	// function are not guaranteed to be deleted, it is possible for the
 	// Actuator to ignore some of them e.g. if max configured level of
 	// parallelism is reached.
-	StartDeletion(empty, needDrain []*apiv1.Node) (*status.ScaleDownStatus, errors.AutoscalerError)
+	StartDeletion(empty, needDrain []*apiv1.Node) (status.ScaleDownResult, []*status.ScaleDownNode, errors.AutoscalerError)
 	// CheckStatus returns an immutable snapshot of ongoing deletions.
 	CheckStatus() ActuationStatus
 	// ClearResultsNotNewerThan removes information about deletions finished
 	// before or exactly at the provided timestamp.
 	ClearResultsNotNewerThan(time.Time)
+	// DeletionResults returns deletion results since the last ClearResultsNotNewerThan call
+	// in a map form, along with the timestamp of last result.
+	DeletionResults() (map[string]status.NodeDeleteResult, time.Time)
 }
 
 // ActuationStatus is used for feeding Actuator status back into Planner
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -388,7 +388,6 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
 	scaleUpStatus := &status.ScaleUpStatus{Result: status.ScaleUpNotTried}
 	scaleUpStatusProcessorAlreadyCalled := false
 	scaleDownStatus := &scaledownstatus.ScaleDownStatus{Result: scaledownstatus.ScaleDownNotTried}
-	scaleDownStatusProcessorAlreadyCalled := false
 
 	defer func() {
 		// Update status information when the loop is done (regardless of reason)
@@ -403,14 +402,22 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
 		if !scaleUpStatusProcessorAlreadyCalled && a.processors != nil && a.processors.ScaleUpStatusProcessor != nil {
 			a.processors.ScaleUpStatusProcessor.Process(a.AutoscalingContext, scaleUpStatus)
 		}
-		if !scaleDownStatusProcessorAlreadyCalled && a.processors != nil && a.processors.ScaleDownStatusProcessor != nil {
+		if a.processors != nil && a.processors.ScaleDownStatusProcessor != nil {
+			// Gather status before scaledown status processor invocation
+			nodeDeletionResults, nodeDeletionResultsAsOf := a.scaleDownActuator.DeletionResults()
+			scaleDownStatus.NodeDeleteResults = nodeDeletionResults
+			scaleDownStatus.NodeDeleteResultsAsOf = nodeDeletionResultsAsOf
+			a.scaleDownActuator.ClearResultsNotNewerThan(scaleDownStatus.NodeDeleteResultsAsOf)
 			scaleDownStatus.SetUnremovableNodesInfo(a.scaleDownPlanner.UnremovableNodes(), a.scaleDownPlanner.NodeUtilizationMap(), a.CloudProvider)
+
 			a.processors.ScaleDownStatusProcessor.Process(a.AutoscalingContext, scaleDownStatus)
 		}
 
-		err := a.processors.AutoscalingStatusProcessor.Process(a.AutoscalingContext, a.clusterStateRegistry, currentTime)
-		if err != nil {
-			klog.Errorf("AutoscalingStatusProcessor error: %v.", err)
+		if a.processors != nil && a.processors.AutoscalingStatusProcessor != nil {
+			err := a.processors.AutoscalingStatusProcessor.Process(a.AutoscalingContext, a.clusterStateRegistry, currentTime)
+			if err != nil {
+				klog.Errorf("AutoscalingStatusProcessor error: %v.", err)
+			}
 		}
 	}()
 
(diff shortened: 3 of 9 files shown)


cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -96,47 +96,52 @@ func (a *Actuator) ClearResultsNotNewerThan(t time.Time) {
 	a.nodeDeletionTracker.ClearResultsNotNewerThan(t)
 }
 
+// ScaleDownStatus returns current scale down status.
+func (a *Actuator) ScaleDownStatus() *status.ScaleDownStatus {
```
towca, line 100: The name and description for the method suggest that a complete status is returned, where in reality it only sets 2 specific fields - IMO it's misleading.

It could make sense as something like `BaseScaleDownStatus`, but then it'd still be inconsistent with how other fields are set. `UnremovableNodesInfo` is set in `RunOnce` using data obtained from `Planner`, `ScaledDownNodes` are set in `RunOnce` using data obtained from `Actuator`, `Result` is set in `RunOnce` based on various things. I see no reason why `NodeDeleteResults` and `NodeDeleteResultsAsOf` would be set in a different way (and if they are, I start wondering why it's different for them).

I'd definitely set these 2 fields in `RunOnce` like the others. The necessary data is in `NodeDeletionTracker`, which `RunOnce` currently doesn't have access to. We can either expose the data from `Actuator` by delegating to the tracker, or plumb `NodeDeletionTracker` to `StaticAutoscaler` and call it directly. The former approach probably makes more sense, as `Actuator` already exposes a method to clear the results.

cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -96,47 +96,52 @@ func (a *Actuator) ClearResultsNotNewerThan(t time.Time) {
...
 
-	results, ts := a.nodeDeletionTracker.DeletionResults()
-	scaleDownStatus := &status.ScaleDownStatus{NodeDeleteResults: results, NodeDeleteResultsAsOf: ts}
-
 	emptyToDelete, drainToDelete := a.budgetProcessor.CropNodes(a.nodeDeletionTracker, empty, drain)
 	if len(emptyToDelete) == 0 && len(drainToDelete) == 0 {
-		scaleDownStatus.Result = status.ScaleDownNoNodeDeleted
-		return scaleDownStatus, nil
+		result = status.ScaleDownNoNodeDeleted
+		return
```
towca, line 116: Could you used named return parameters instead of the naked return? IMO the function is long/complex enough where it decreases the readability.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -635,19 +636,19 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
 			scaleDownStatus.RemovedNodeGroups = removedNodeGroups
 		}
 
+		scaleDownStatus = a.scaleDownActuator.ScaleDownStatus()
```
towca, line 639: Why not do this where `scaleDownStatus` is defined?

towca (changes requested):

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -403,14 +402,16 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
...
-		if !scaleDownStatusProcessorAlreadyCalled && a.processors != nil && a.processors.ScaleDownStatusProcessor != nil {
+		if a.processors != nil && a.processors.ScaleDownStatusProcessor != nil {
 			scaleDownStatus.SetUnremovableNodesInfo(a.scaleDownPlanner.UnremovableNodes(), a.scaleDownPlanner.NodeUtilizationMap(), a.CloudProvider)
 			a.processors.ScaleDownStatusProcessor.Process(a.AutoscalingContext, scaleDownStatus)
 		}
 
-		err := a.processors.AutoscalingStatusProcessor.Process(a.AutoscalingContext, a.clusterStateRegistry, currentTime)
-		if err != nil {
-			klog.Errorf("AutoscalingStatusProcessor error: %v.", err)
+		if a.processors != nil {
```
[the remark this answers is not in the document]
mtrqq: good point, I'll add additional check for `AutoscalingStatusProcessor` not being nil

cluster-autoscaler/core/scaledown/actuation/actuator.go
mtrqq: Makes sense, will move setting part back to `RunOnce` and will add `DeletionResults` as a delegate method of actuator

cluster-autoscaler/core/static_autoscaler.go
mtrqq: Good question, currently I'm trying to keep it as close to original place of capturing in order to get 100% same behavior and state of data as was intended before my changes. If you are saying that it's safe to move it to status initialization - I may do that

cluster-autoscaler/core/scaledown/actuation/actuator.go
```diff
@@ -96,47 +96,46 @@ func (a *Actuator) ClearResultsNotNewerThan(t time.Time) {
...
 
-	results, ts := a.nodeDeletionTracker.DeletionResults()
-	scaleDownStatus := &status.ScaleDownStatus{NodeDeleteResults: results, NodeDeleteResultsAsOf: ts}
-
+	scaledDownNodes := make([]*status.ScaleDownNode, 0)
 	emptyToDelete, drainToDelete := a.budgetProcessor.CropNodes(a.nodeDeletionTracker, empty, drain)
 	if len(emptyToDelete) == 0 && len(drainToDelete) == 0 {
-		scaleDownStatus.Result = status.ScaleDownNoNodeDeleted
-		return scaleDownStatus, nil
+		return status.ScaleDownNoNodeDeleted, scaledDownNodes, nil
```
towca, line 113: nit: `nil` instead  of `scaledDownNodes` would be slightly better to read IMO.

cluster-autoscaler/core/scaledown/scaledown.go
```diff
@@ -56,12 +56,14 @@ type Actuator interface {
...
 	// Actuator to ignore some of them e.g. if max configured level of
 	// parallelism is reached.
-	StartDeletion(empty, needDrain []*apiv1.Node) (*status.ScaleDownStatus, errors.AutoscalerError)
+	StartDeletion(empty, needDrain []*apiv1.Node) (status.ScaleDownResult, []*status.ScaleDownNode, errors.AutoscalerError)
 	// CheckStatus returns an immutable snapshot of ongoing deletions.
 	CheckStatus() ActuationStatus
 	// ClearResultsNotNewerThan removes information about deletions finished
 	// before or exactly at the provided timestamp.
 	ClearResultsNotNewerThan(time.Time)
+	// DeletionResults returns deletion results in a map form, along with the timestamp of last result.
```
towca, line 65: Could you add a mention that it returns all deletions since the last `ClearResultsNotNewerThan` call?

cluster-autoscaler/core/static_autoscaler.go
towca: IMO the best place to set the results is actually in the defer block, right before the processor is called. It's consistent with how `UnremovableNodesInfo` is set, and it makes the most sense to me semantically - "before calling the processor, finalize the status by filling all asynchronously-obtained information".

There shouldn't be any change in behavior - we care about reflecting the results in the status as soon as the tracker reports them, and only reflecting each result once. IIUC before and after the changes, each result should only be reflected in the status once, since we always clear the results immediately after reflecting. After the changes, if anything the results will be reported earlier than before the changes (since the defer call will always happen regardless of the rest of the loop).

towca (commented):
Just some last comments, overall LGTM. Have you tested this manually? Do you think it'd be viable to add a unit test that checks if the results are reported correctly (something that would break if we were to repeat the bug in the future)?

cluster-autoscaler/core/scaledown/scaledown.go
mtrqq: Done

cluster-autoscaler/core/scaledown/actuation/actuator.go
mtrqq: agree, done

cluster-autoscaler/core/static_autoscaler.go
mtrqq: Thanks for explanations! Moved node deletion results collection to defer statement

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -2396,6 +2412,108 @@ func TestFilterOutYoungPods(t *testing.T) {
...
+				RemovedNodeGroups:     nil,
+				NodeDeleteResults:     map[string]status.NodeDeleteResult{},
+				NodeDeleteResultsAsOf: time.Time{},
+			},
+		},
+	}
+
+	for testName, test := range testCases {
+		// prevent issues with scoping, we should be able to get rid of that with Go 1.22
+		test := test
```
towca, line 2461: Since we have so few test cases and these are just unit tests, I'd remove the `t.Parallel` below, and then we don't need this part.

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -2396,6 +2412,108 @@ func TestFilterOutYoungPods(t *testing.T) {
 	}
 }
 
+func TestStaticAutoscalerRunOnceInvokesScaleDownStatusProcessorNoCandidates(t *testing.T) {
```
towca, line 2415: Could we remove the "NoCandidates" part from the name, and add test cases for both when there are and aren't candidates?

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -2396,6 +2412,108 @@ func TestFilterOutYoungPods(t *testing.T) {
...
+				ScaledDownNodes: []*status.ScaleDownNode{},
+				UnremovableNodes: []*status.UnremovableNode{
+					{
+						Node:        n1,
+						BlockingPod: nil,
+						Reason:      simulator.NotUnderutilized,
+					},
+				},
+				RemovedNodeGroups:     nil,
+				NodeDeleteResults:     map[string]status.NodeDeleteResult{},
```
towca, line 2453: We have to have something inside the results here, otherwise we're not verifying that they're passed correctly at all. I think your code has a bug right now (see the other comment), but the test isn't failing because this is empty.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -403,14 +402,22 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) caerrors.AutoscalerErr
...
+		if a.processors != nil && a.processors.ScaleDownStatusProcessor != nil {
 			scaleDownStatus.SetUnremovableNodesInfo(a.scaleDownPlanner.UnremovableNodes(), a.scaleDownPlanner.NodeUtilizationMap(), a.CloudProvider)
 			a.processors.ScaleDownStatusProcessor.Process(a.AutoscalingContext, scaleDownStatus)
 		}
 
-		err := a.processors.AutoscalingStatusProcessor.Process(a.AutoscalingContext, a.clusterStateRegistry, currentTime)
-		if err != nil {
-			klog.Errorf("AutoscalingStatusProcessor error: %v.", err)
+		// Gather node deletion result before scaledown status processor invocation
+		nodeDeletionResults, nodeDeletionResultsAsOf := a.scaleDownActuator.DeletionResults()
```
towca, line 411: Isn't this before the wrong processor? It should be inside the if block above, right before `scaleDownStatus.SetUnremovableNodesInfo`. Right now it's setting the fields after the scale-down processor call, right before the loop exits and the status is lost.

towca (changes requested):

cluster-autoscaler/core/static_autoscaler_test.go
mtrqq: t.Parallel() would make this test to be able to run with other parallel tests (not only this small group) which greatly increases speed of a run. I don't agree that it's a big sacrifice for a speed up

cluster-autoscaler/core/static_autoscaler_test.go
mtrqq: Totally, makes sense, added 3 more test cases to cover this behavior

cluster-autoscaler/core/static_autoscaler_test.go
mtrqq: You're right, it was gathered in the wrong place, I'll conduct another pass of manual testing

mtrqq: Made a run on a patched commit, status processor reports periodically with or without scaledown candidates

cluster-autoscaler/core/static_autoscaler_test.go
towca: Fair enough! FYI `t.Parallel()` only works within a single package, and there's just one other `t.Parallel()` call in this package, so this might be achieving less than you think though.

cluster-autoscaler/core/static_autoscaler_test.go
towca: It looks like still none of the test cases check if any actual results are reported? Or am I missing something here? If you run this test on the previous version which retrieved the results in the wrong place - does it start failing?

mtrqq: If you are talking about NodeDeletionResults object itself it would be quite hard to get this part tested in a reliable way because those are being populated when async drain/deletion completed. Talking about whether it fails if I move that - due to the fact that it's empty in all tests - it's not failing

towca: Yeah, I specifically meant for the test to check if the results are reported, otherwise it wouldn't have caught the previous bug. The async part is why I though it might not be viable to implement such a test. Looking at the test code, it might not be that hard though. Wouldn't it be enough to make `nodeDeletionTracker` configurable through `autoscalerSetupConfig`, and then just fake-report the deletions directly through the tracker from the test code?

mtrqq: Got your point, I've made a kind of mocked node deletion tracker to simulate the behavior

cluster-autoscaler/core/static_autoscaler_test.go
```diff
@@ -2396,6 +2413,240 @@ func TestFilterOutYoungPods(t *testing.T) {
...
+	underUtilizedPod := BuildTestPod("p1", 20, 20, WithNodeName("n1"))
+	utilizedPod := BuildTestPod("p1", 800, 800, WithNodeName("n1"))
+
+	testCases := map[string]struct {
+		pods                         []*apiv1.Pod
+		nodes                        []*apiv1.Node
+		fakeDeletionResults          map[string]status.NodeDeleteResult
+		fakeDeletionResultsNodeGroup string
+		expectedStatus               *status.ScaleDownStatus
+	}{
```
towca, line 2445: IMO the test cases test the internals of dependencies more than the RunOnce logic they're supposed to. 

Ideally, for this test we'd like to verify that:
1. The processor always gets called during the loop, regardless of other logic.
2. The processor always gets all relevant data when it's called.

Obviously we can't cover every scenario, but for 1. I could see e.g. 3 cases: "scale-down happened", "only scale-up happened, scale-down didn't because no candidates", "nothing happened, the loop short-circuited early". For 2., we need to verify that in all 3 cases, all relevant pieces of the status are filled correctly. I don't see much value in testing the internals of the status (e.g. different unremovable reasons, or deletion result types), that should be covered by the unit tests of the components where we get the data from. All we need to verify here is that the data was passed from the components correctly (and as much data as it makes sense for every test case).

The test is still good enough IMO, this is a nit/something to consider in the future.

towca:
/lgtm
/approve
/hold

Thanks for the change! Left one nit, feel free to unhold the PR if you prefer not to fix.

towca (approved):

cluster-autoscaler/core/static_autoscaler_test.go
mtrqq: I think you are right, it's quite hard to keep concentration on isolation while implementing test which indirectly touches so many parts of CA, thanks for the feedback! 

As a result I've reduced tests count to 3, now it's testing following scenarious: scaledown, no scaledown, no scaledown with deleted node

towca: Oh yeah definitely, this line is frequently blurred, and CA is frequently not easy to test because of old tech debt. The test cases are very clear now, thank you!

PR #6577, opened 2024-02-28
Consider preemption policy for expandable pods

### LINKED ISSUE #6227: Computation of expendable pods does not consider preemption policy
cluster-autoscaler

$ kubectl version
Server Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.7", GitCommit:"84e1fc493a47446df2e155e70fca768d2653a398", GitTreeState:"clean", BuildDate:"2023-07-19T12:16:45Z", GoVersion:"go1.20.6", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

If there is an unschedulable pod, then CA attempts to determine if this pod should be considered as unschedulable (as indicated by the kube-scheduler) or it can be scheduled onto an existing node. This check is done in every run of the `static_autoscaler.go` -  this is done by the [podlistprocessor](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/podlistprocessor/filter_out_schedulable.go#L46).  To do this it also runs a scheduler simulation and one of the filter plugins that will be called is `scheduler/framework/plugins/noderesources/fit.go`. This is where the requests from the Pod are matched against the `NodeInfo` which is constructed by CA. 

> I have changed the name of the pod and node for easy reference elsewhere.

In our case kube-scheduler failed to schedule the pod with the following captured as part of the Pod status:
```
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2023-10-25T10:25:21Z"
    message: '0/13 nodes are available: 2 Insufficient memory, 2 node(s) had untolerated
      taint {taint-name: arm64}, 9 node(s) didn''t match Pod''s node affinity/selector.
      preemption: not eligible due to preemptionPolicy=Never..'
    reason: Unschedulable
    status: "False"
    type: PodScheduled
  phase: Pending
```

Where as CA suggested that it is possible to schedule this pod onto an existing node.

```
I1026 03:55:42.730901       1 filter_out_schedulable.go:63] Filtering out schedulables
I1026 03:55:42.730914       1 hinting_simulator.go:63] Looking for place for default/test-pod
I1026 03:55:42.731025       1 hinting_simulator.go:77] Pod default/test-pod can be moved to test-node
I1026 03:55:42.731043       1 filter_out_schedulable.go:120] 1 pods marked as unschedulable can be scheduled.
I1026 03:55:42.731057       1 filter_out_schedulable.go:75] Schedulable pods present
I1026 03:55:42.731079       1 static_autoscaler.go:512] No unschedulable pods
```
The node `test-node` in question already had 18 pods deployed onto it and it had the following allocatable resources:
```
Allocated resources:
  (Total limits may be over 100 percent, i.e., overcommitted.)
  Resource           Requests            Limits
  --------           --------            ------
  cpu                3566m (45%)         17520m (221%)
  memory             129289570250 (96%)  130410Mi (102%)
```
The `test-pod` was requesting the following resources:
```
  requests:
     cpu: "2"
     memory: 18000Mi
```
This would never fit the `test-node` and kube-scheduler was correct in determining that. But still CA was suggesting that this pod should be scheduled onto this `test-node`

When we looked closer at the pods that were populated in the `ClusterSnapshot` that is constructed by CA we noticed that it only listed 17/18 pods and thus its computation of total requested resources on the Node was not accurate.

We found the deployed pod that was not listed/considered by CA and for the computation of total requests on the node. This pod (lets call it `exp-pod-1`) had the following in its Spec:
```
 preemptionPolicy: Never
 priority: -15
```
The cluster-autoscaler was started (among other flags) with:
```
--expendable-pods-priority-cutoff=-10
```
When we looked [expendable.go](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/utils/expendable.go#L66-L68) we noticed that CA filters out pods when computing total requested resources on a node which satisfy:
```
pod.Spec.Priority != nil && int(*pod.Spec.Priority) < expendablePodsPriorityCutoff
```

This is incomplete as it should consider the preemption policy of the unschedulable pods as well as that is considered by the kube-scheduler. Without that there is always going to be a difference between the decision taken by CA and kube-scheduler and it will result in no scale-up when it should have been done in this case.

**What happened instead?**:
We expected a scale up to happen as `test-pod` was marked as unschedulable by the kube-scheduler and the node that CA suggested for this pod did not sufficient memory.

1. start cluster-autoscaler with `--expendable-pods-priority-cutoff=-10`
2. Create a pod with:
```
 preemptionPolicy: Never
 priority: -15
```
Make sure that this pod gets deployed on to a node.
3. Create another pod with requests higher than what is available on this node. This pod should have `preemptionPolicy: Never`

BigDarkClown:
#### What type of PR is this?

/kind bug

Consider preemption policy when determining the expandable pods. The pods with preemption policy set to never should never be considered as expandable pods.

Fixes #6227

### DIFF
--- a/cluster-autoscaler/core/utils/expendable.go
+++ b/cluster-autoscaler/core/utils/expendable.go
@@ -34,7 +34,7 @@ func FilterOutExpendableAndSplit(unschedulableCandidates []*apiv1.Pod, nodes []*
 	}
 
 	for _, pod := range unschedulableCandidates {
-		if pod.Spec.Priority != nil && int(*pod.Spec.Priority) < expendablePodsPriorityCutoff {
+		if IsExpendablePod(pod, expendablePodsPriorityCutoff) {
 			klog.V(4).Infof("Pod %s has priority below %d (%d) and will scheduled when enough resources is free. Ignoring in scale up.", pod.Name, expendablePodsPriorityCutoff, *pod.Spec.Priority)
 		} else if nominatedNodeName := pod.Status.NominatedNodeName; nominatedNodeName != "" {
 			if nodeNames[nominatedNodeName] {
@@ -64,5 +64,7 @@ func FilterOutExpendablePods(pods []*apiv1.Pod, expendablePodsPriorityCutoff int
 
 // IsExpendablePod tests if pod is expendable for give priority cutoff
 func IsExpendablePod(pod *apiv1.Pod, expendablePodsPriorityCutoff int) bool {
-	return pod.Spec.Priority != nil && int(*pod.Spec.Priority) < expendablePodsPriorityCutoff
+	preemptLowerPriority := pod.Spec.PreemptionPolicy == nil || *pod.Spec.PreemptionPolicy == apiv1.PreemptLowerPriority
+	lowPriority := pod.Spec.Priority != nil && int(*pod.Spec.Priority) < expendablePodsPriorityCutoff
+	return preemptLowerPriority && lowPriority
 }
--- a/cluster-autoscaler/core/utils/expendable_test.go
+++ b/cluster-autoscaler/core/utils/expendable_test.go
@@ -88,20 +88,92 @@ func TestFilterOutExpendablePods(t *testing.T) {
 	assert.Equal(t, podWaitingForPreemption2, res[2])
 }
 
-func TestIsExpendablePod(t *testing.T) {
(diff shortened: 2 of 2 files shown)


Shubham82:
Thanks @BigDarkClown 
/lgtm

PR #6584, opened 2024-03-04
Delay force refresh by DefaultInterval when OCI GetNodePool call retu…

vbhargav875:
What type of PR is this?

/kind feature

What this PR does / why we need it:

In the OCI Nodepools Implementation of ClusterAutoscaler, we do not set any delay to the calls that ClusterAutoscaler makes to the OKE GetNodePool API when the API returns a 404 error (which is possible when the Nodepool is deleted or an incorrect Nodepool ocid is passed) and hence we continuously trigger forceRefresh (when CA triggers Refresh) and therefore CA will indefinitely make GetNodePool calls to CP.
Introducing a max retry count and setting a delay to the next force refresh when the retries is exhausted will help prevent this.

Does this PR introduce a user-facing change?

No

Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

None

### DIFF
--- a/cluster-autoscaler/cloudprovider/oci/nodepools/cache.go
+++ b/cluster-autoscaler/cloudprovider/oci/nodepools/cache.go
@@ -40,24 +40,33 @@ func (c *nodePoolCache) nodePools() map[string]*oke.NodePool {
 	return result
 }
 
-func (c *nodePoolCache) rebuild(staticNodePools map[string]NodePool) error {
+func (c *nodePoolCache) rebuild(staticNodePools map[string]NodePool, maxGetNodepoolRetries int) (httpStatusCode int, err error) {
 	klog.Infof("rebuilding cache")
+	var resp oke.GetNodePoolResponse
+	var statusCode int
 	for id := range staticNodePools {
-		// prevent us from getting a node pool at the same time that we're performing delete actions on the node pool.
-		c.mu.Lock()
-		resp, err := c.okeClient.GetNodePool(context.Background(), oke.GetNodePoolRequest{
-			NodePoolId: common.String(id),
-		})
-		c.mu.Unlock()
-
-		if err != nil {
-			return err
+		for i := 1; i <= maxGetNodepoolRetries; i++ {
+			// prevent us from getting a node pool at the same time that we're performing delete actions on the node pool.
+			c.mu.Lock()
+			resp, err = c.okeClient.GetNodePool(context.Background(), oke.GetNodePoolRequest{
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/cloudprovider/oci/nodepools/oci_manager.go
```diff
@@ -249,11 +250,15 @@ func (m *ociManagerImpl) TaintToPreventFurtherSchedulingOnRestart(nodes []*apiv1
 }
 
 func (m *ociManagerImpl) forceRefresh() error {
-	err := m.nodePoolCache.rebuild(m.staticNodePools)
+	httpStatusCode, err := m.nodePoolCache.rebuild(m.staticNodePools, maxGetNodepoolRetries)
 	if err != nil {
+		if httpStatusCode == 404 {
```
trungng92, line 255: If the OCI SDK returns a 404, is that considered an error?

cluster-autoscaler/cloudprovider/oci/nodepools/cache.go
```diff
@@ -40,24 +40,33 @@ func (c *nodePoolCache) nodePools() map[string]*oke.NodePool {
 	return result
 }
 
-func (c *nodePoolCache) rebuild(staticNodePools map[string]NodePool) error {
+func (c *nodePoolCache) rebuild(staticNodePools map[string]NodePool, maxGetNodepoolRetries int) (httpStatusCode int, err error) {
 	klog.Infof("rebuilding cache")
+	var resp oke.GetNodePoolResponse
+	var statusCode int
 	for id := range staticNodePools {
```
trungng92, line 47: One consideration. We're looping over all of the node pools, but we exit/error out on the first one that errors out, which means you could have 20 node pools, and if the first one is deleted/errors out, then we won't reconcile any of them.

This is how the previous behavior worked as well, so we don't need to address it now, but we should note it as a bug (perhaps as a code comment and/or an internal ticket).

vbhargav875: Agreed

cluster-autoscaler/cloudprovider/oci/nodepools/oci_manager.go
vbhargav875: Yes, in case of 404, err is not nil.

trungng92:
Changes look good to me

jlamillan (approved):
/lgtm
/approve

PR #6589, opened 2024-03-06
Add an option to trigger iterations more frequently

kawych:
Trigger new autoscaling iterations based on two additional criteria:
1. There are new unschedulable pods - reduces autoscaling latency slightly
2. Last iteration was productive (there was a scale-up or scale-down) - increases autoscaling throughput in cases where there are multiple heterogeneous workloads, each requiring a separate iteration.

This also avoids burning CPU unnecessarily, which would happen if we just reduced scanInterval to some value close to zero.

This functionality is flag-guarded, disabled by default.

```release-note
NONE
```

### DIFF
--- a/cluster-autoscaler/loop/trigger.go
+++ b/cluster-autoscaler/loop/trigger.go
@@ -1,0 +1,144 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package loop
+
+import (
+	"context"
+	"time"
+
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/autoscaler/cluster-autoscaler/metrics"
+	kube_client "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+	podv1 "k8s.io/kubernetes/pkg/api/v1/pod"
+)
+
+const maxPodChangeAge = 10 * time.Second
+
+var (
+	podsResource             = "pods"
+	unschedulablePodSelector = fields.ParseSelectorOrDie("spec.nodeName==" + "" + ",status.phase!=" +
+		string(apiv1.PodSucceeded) + ",status.phase!=" + string(apiv1.PodFailed))
+)
+
+// scalingTimesGetter exposes recent autoscaler activity
+type scalingTimesGetter interface {
+	LastScaleUpTime() time.Time
+	LastScaleDownDeleteTime() time.Time
+}
+
+// LoopTrigger object implements criteria used to start new autoscaling iteration
+type LoopTrigger struct {
+	podObserver        *UnschedulablePodObserver
+	scanInterval       time.Duration
+	scalingTimesGetter scalingTimesGetter
+}
+
+// NewLoopTrigger creates a LoopTrigger object
+func NewLoopTrigger(podObserver *UnschedulablePodObserver, scalingTimesGetter scalingTimesGetter, scanInterval time.Duration) *LoopTrigger {
+	return &LoopTrigger{
+		podObserver:        podObserver,
+		scanInterval:       scanInterval,
+		scalingTimesGetter: scalingTimesGetter,
+	}
+}
+
+// Wait waits for the next autoscaling iteration
+func (t *LoopTrigger) Wait(lastRun time.Time) {
+	sleepStart := time.Now()
+	defer metrics.UpdateDurationFromStart(metrics.LoopWait, sleepStart)
+
+	// To improve scale-up throughput, Cluster Autoscaler starts new iteration
+	// immediately if the previous one was productive.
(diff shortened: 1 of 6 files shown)


kawych:
CC @x13n

gjtempleton:
I take it we're planning on following this up with a separate PR to make use of this functionality?

x13n:
I a bit hesitant exposing internal fields in static autoscaler like this when there's plenty of existing hooks in static autoscaler already. Could this instead be implemented as scale up/down status processor?

kawych:
@gjtempleton I actually planned to do this in a forked Cluster Autoscaler implementation, assuming that exposing these fields won't hurt. But I can follow up with enabling the same in this repo if the repo owners are OK with it.

@x13n can you elaborate more on what you are worried about? I can agree that exposing all of the internal state would not be a good idea, but for these specific fields it's fairly natural to show them to the external world (readonly).  Additionally, the "run" function where these values would be actually used [1] can easily access outputs from cluster autoscaler, while for processors don't have any natural way of passing the results there.

[1] https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/main.go#L574

x13n:
This PR only exposes two fields, so it is a bit hard to discuss how exactly it will be used. Regading the `run` function - we should limit the amount of logic in `main`, rather than keep adding to it.

But to Guy's point above: Let's maybe start with a purpose of the change and then decide how to best fulfill that purpose instead of starting by exposing internal fields of StaticAutoscaler without extra context? Maybe exposing these fields is the way to go, but it is hard to decide if that's the entire change.

kawych:
We discussed this change offline with @x13n and decided to implement a more comprehensive (flag-guarded) feature of improving autoscaling throughput and latency in the same PR.

cluster-autoscaler/main.go
```diff
@@ -591,23 +593,34 @@ func run(healthCheck *metrics.HealthCheck, debuggingSnapshotter debuggingsnapsho
...
-				err := autoscaler.RunOnce(loopStart)
-				if err != nil && err.Type() != errors.TransientError {
-					metrics.RegisterError(err)
-				} else {
-					healthCheck.UpdateLastSuccessfulRun(time.Now())
-				}
-
-				metrics.UpdateDurationFromStart(metrics.Main, loopStart)
-			}
+	runAutoscalerOnce := func(loopStart time.Time) {
```
x13n, line 596: Any reason not to make it a proper function, perhaps in the new `loop` module?

cluster-autoscaler/loop/trigger.go
```diff
@@ -0,0 +1,144 @@
...
+const maxPodChangeAge = 10 * time.Second
+
+var (
+	podsResource             = "pods"
+	unschedulablePodSelector = fields.ParseSelectorOrDie("spec.nodeName==" + "" + ",status.phase!=" +
+		string(apiv1.PodSucceeded) + ",status.phase!=" + string(apiv1.PodFailed))
+)
+
+// StaticAutoscaler exposes recent autoscaler activity
+type StaticAutoscaler interface {
```
x13n, line 41: It doesn't need to be public, this interface is meant to be consumed locally in this module. Also, `StaticAutoscaler` is the current implementation of the interface, but it doesn't make a lot of sense as the name - with the current set of exposed functions something along the lines of `scalingTimesGetter` would be more appropriate.

cluster-autoscaler/loop/trigger.go
```diff
@@ -0,0 +1,144 @@
...
+		podObserver:  podObserver,
+		scanInterval: scanInterval,
+		autoscaler:   autoscaler,
+	}
+}
+
+// Wait waits for the next autoscaling iteration
+func (t *LoopTrigger) Wait(lastRun time.Time) {
+	sleepStart := time.Now()
+	defer metrics.UpdateDurationFromStart("loopWait", sleepStart)
```
x13n, line 65: `"loopWait"` should be a constant in `metrics.go`, same as other function labels.

cluster-autoscaler/loop/trigger.go
```diff
@@ -0,0 +1,144 @@
...
+
+	// To improve scale-up throughput, Cluster Autoscaler starts new iteration
+	// immediately if the previous one was productive.
+	if !t.autoscaler.LastScaleUpTime().Before(lastRun) ||
+		!t.autoscaler.LastScaleDownDeleteTime().Before(lastRun) {
+		select {
+		case <-t.podObserver.unschedulablePodChan:
```
x13n, line 72: So you're clearing the signal about unschedulable pod appearing here, but not about scan interval passing. Why are you treating some triggers differently than others? This will lead to wasting a loop that isn't needed every now and then, which is not terrible, but definitely something we could avoid.

x13n (changes requested):

cluster-autoscaler/loop/trigger.go
kawych: Done

cluster-autoscaler/loop/trigger.go
kawych: Done

cluster-autoscaler/loop/trigger.go
kawych: We're clearing just this one channel because it persists between loops, while the time.After() is re-created with each iteration. I don't think it will have the effect you're suggesting, but I'll actually test that, for now publishing the other responses.

cluster-autoscaler/main.go
kawych: Done

kawych (commented):
Thanks, these are great suggestions.

Putting it on hold for now since I'll want to re-test it when approved.

/hold

cluster-autoscaler/loop/trigger.go
x13n: Ah, you're right, we don't make a call to `After()` at all in this branch, nvm my comment then :)

kawych:
I re-tested it, submitting.

/unhold

PR #6595, opened 2024-03-06
Update expander options for the AWS cloud provider README

drmorr0:
#### What type of PR is this?

/kind documentation

Cluster autoscaler supports more expansion options than it used to


Shubham82:
Thanks @drmorr0 for this PR.
/lgtm

PR #6596, opened 2024-03-07
Remove shadow err variable in deleteCreatedNodesWithErros func

damikag:
#### What type of PR is this?
/kind bug

This PR removes shadowed ```err``` variable in ```deleteCreatedNodesWithErrors()``` in ```static_autoscaler.go```. Previously the returned error from ```nodeGroup.DeleteNodes()``` was assigned to a shadowed ```err``` variable. Because of that the original variable remains ```nil``` even when node deletion failed and that original variable is used to determine whether there was a successful node deletion. If there was a successful node deletion the current CA loop will skip. Because of this shadowed variable, CA can get stuck in an infinite loop when there are created nodes with errors and node deletion fails for those nodes.

NONE


x13n:
Good catch!

/lgtm
/approve

PR #6604, opened 2024-03-08
Track auto-provisioned queued provisioning node groups.

dsafdsa1:
#### What type of PR is this?
/kind feature

It allows us to differentiate between standard and queued-provisioning node groups that are getting auto-provisioned.

```release-note
NONE
```

```docs

```

### DIFF
--- a/cluster-autoscaler/metrics/metrics.go
+++ b/cluster-autoscaler/metrics/metrics.go
@@ -390,20 +390,22 @@ var (
 		},
 	)
 
-	nodeGroupCreationCount = k8smetrics.NewCounter(
+	nodeGroupCreationCount = k8smetrics.NewCounterVec(
 		&k8smetrics.CounterOpts{
 			Namespace: caNamespace,
 			Name:      "created_node_groups_total",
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/metrics/metrics.go
```diff
@@ -390,20 +391,22 @@ var (
...
 	)
 
-	nodeGroupCreationCount = k8smetrics.NewCounter(
+	nodeGroupCreationCount = k8smetrics.NewCounterVec(
 		&k8smetrics.CounterOpts{
 			Namespace: caNamespace,
 			Name:      "created_node_groups_total",
 			Help:      "Number of node groups created by Node Autoprovisioning.",
 		},
+		[]string{"queued_provisioning"},
```
towca, line 400: IIRC `queued_provisioning` is a GCE/GKE-specific concept, right? These metrics are supposed to be cloud-provider-agnostic. WDYT about a type/kind (or other appropriate more generic term) label instead? And then cloud-provider-specific code can set the value to `queued_provisioning`?

dsafdsa1: Changed to group_type.

PR #6607, opened 2024-03-11
Support scale up of nodepool with zero nodes with ephemeral storage r…

vbhargav875:
#### What type of PR is this?
/kind feature

With OCI Nodepools current implementation, we do not support scaling up of a nodepool with zero nodes when the pod spec was ephemeral-storage requests since does not know what ephemeral-storage size should be set to the node. With this PR we allow users to set the ephemeral-storage of a nodepool using OCI freeform tags.

### DIFF
--- a/cluster-autoscaler/cloudprovider/oci/common/oci_tag.go
+++ b/cluster-autoscaler/cloudprovider/oci/common/oci_tag.go
@@ -1,0 +1,27 @@
+/*
+Copyright 2021-2024 Oracle and/or its affiliates.
+*/
+
+package common
+
+import (
+	oke "k8s.io/autoscaler/cluster-autoscaler/cloudprovider/oci/vendor-internal/github.com/oracle/oci-go-sdk/v65/containerengine"
+)
+
+// TagsGetter returns the oci tags for the pool.
+type TagsGetter interface {
+	GetNodePoolFreeformTags(*oke.NodePool) (map[string]string, error)
+}
+
+// TagsGetterImpl is the implementation to fetch the oci tags for the pool.
+type TagsGetterImpl struct{}
+
+// CreateTagsGetter creates a new oci tags getter.
+func CreateTagsGetter() TagsGetter {
+	return &TagsGetterImpl{}
+}
+
+// GetNodePoolFreeformTags returns the FreeformTags for the nodepool
+func (tgi *TagsGetterImpl) GetNodePoolFreeformTags(np *oke.NodePool) (map[string]string, error) {
+	return np.FreeformTags, nil
+}
--- a/cluster-autoscaler/cloudprovider/oci/nodepools/consts/annotations.go
+++ b/cluster-autoscaler/cloudprovider/oci/nodepools/consts/annotations.go
@@ -25,4 +25,7 @@ const (
 	// ToBeDeletedByClusterAutoscaler is the taint used to ensure that after a node has been called to be deleted
 	// no more pods will schedule onto it
 	ToBeDeletedByClusterAutoscaler = "ignore-taint.cluster-autoscaler.kubernetes.io/oke-impending-node-termination"
+
+	// EphemeralStorageSize is the freeform tag key that would be used to determine the ephemeral-storage size of the node
+	EphemeralStorageSize = "cluster-autoscaler/node-ephemeral-storage"
 )
--- a/cluster-autoscaler/cloudprovider/oci/nodepools/oci_manager.go
+++ b/cluster-autoscaler/cloudprovider/oci/nodepools/oci_manager.go
@@ -136,6 +136,7 @@ func CreateNodePoolManager(cloudConfigPath string, discoveryOpts cloudprovider.N
 
 	//ociShapeGetter := ocicommon.CreateShapeGetter(computeClient)
 	ociShapeGetter := ocicommon.CreateShapeGetter(ocicommon.ShapeClientImpl{ComputeMgmtClient: computeMgmtClient, ComputeClient: computeClient})
+	ociTagsGetter := ocicommon.CreateTagsGetter()
 
 	registeredTaintsGetter := CreateRegisteredTaintsGetter()
 
@@ -145,6 +146,7 @@ func CreateNodePoolManager(cloudConfigPath string, discoveryOpts cloudprovider.N
 		computeClient:          &computeClient,
 		staticNodePools:        map[string]NodePool{},
 		ociShapeGetter:         ociShapeGetter,
+		ociTagsGetter:          ociTagsGetter,
 		registeredTaintsGetter: registeredTaintsGetter,
 		nodePoolCache:          newNodePoolCache(&okeClient),
 	}
@@ -210,6 +212,7 @@ type ociManagerImpl struct {
 	okeClient              okeClient
 	computeClient          *core.ComputeClient
 	ociShapeGetter         ocicommon.ShapeGetter
+	ociTagsGetter          ocicommon.TagsGetter
 	registeredTaintsGetter RegisteredTaintsGetter
 	staticNodePools        map[string]NodePool
 
@@ -521,7 +524,15 @@ func (m *ociManagerImpl) buildNodeFromTemplate(nodePool *oke.NodePool) (*apiv1.N
 		Capacity: apiv1.ResourceList{},
 	}
 
-	shape, err := m.ociShapeGetter.GetNodePoolShape(nodePool)
+	freeformTags, err := m.ociTagsGetter.GetNodePoolFreeformTags(nodePool)
+	if err != nil {
+		return nil, err
+	}
+	ephemeralStorage, err := getEphemeralResourceRequestsInBytes(freeformTags)
+	if err != nil {
+		klog.Error(err)
+	}
+	shape, err := m.ociShapeGetter.GetNodePoolShape(nodePool, ephemeralStorage)
 	if err != nil {
 		return nil, err
 	}
@@ -542,11 +553,16 @@ func (m *ociManagerImpl) buildNodeFromTemplate(nodePool *oke.NodePool) (*apiv1.N
 		})
 	}
 
+	if err != nil {
+		return nil, err
+	}
 	node.Status.Capacity[apiv1.ResourcePods] = *resource.NewQuantity(110, resource.DecimalSI)
-
 	node.Status.Capacity[apiv1.ResourceCPU] = *resource.NewQuantity(int64(shape.CPU), resource.DecimalSI)
 	node.Status.Capacity[apiv1.ResourceMemory] = *resource.NewQuantity(int64(shape.MemoryInBytes), resource.DecimalSI)
 	node.Status.Capacity[ipconsts.ResourceGPU] = *resource.NewQuantity(int64(shape.GPU), resource.DecimalSI)
+	if ephemeralStorage != -1 {
+		node.Status.Capacity[apiv1.ResourceEphemeralStorage] = *resource.NewQuantity(ephemeralStorage, resource.DecimalSI)
+	}
 
 	node.Status.Allocatable = node.Status.Capacity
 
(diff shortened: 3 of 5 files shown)


cluster-autoscaler/cloudprovider/oci/nodepools/oci_manager.go
```diff
@@ -542,12 +545,22 @@ func (m *ociManagerImpl) buildNodeFromTemplate(nodePool *oke.NodePool) (*apiv1.N
...
+	}
 	node.Status.Capacity[apiv1.ResourcePods] = *resource.NewQuantity(110, resource.DecimalSI)
-
 	node.Status.Capacity[apiv1.ResourceCPU] = *resource.NewQuantity(int64(shape.CPU), resource.DecimalSI)
 	node.Status.Capacity[apiv1.ResourceMemory] = *resource.NewQuantity(int64(shape.MemoryInBytes), resource.DecimalSI)
 	node.Status.Capacity[ipconsts.ResourceGPU] = *resource.NewQuantity(int64(shape.GPU), resource.DecimalSI)
 
+	ephemeralStorage, err := getEphemeralResourceRequestsInBytes(freeformTags)
+	if err != nil {
+		klog.V(4).Info(err)
```
trungng92, line 559: Hm, curious, why do we log the error, but not return the error?

cluster-autoscaler/cloudprovider/oci/nodepools/oci_manager.go
```diff
@@ -634,6 +647,34 @@ func addTaintToSpec(node *apiv1.Node, taintKey string, effect apiv1.TaintEffect)
...
+		if key == npconsts.EphemeralStorageSize {
+			klog.V(0).Infof("ephemeral-storage size set with value : %v", value)
+			value = strings.ReplaceAll(value, " ", "")
+			valueLength := len(value)
+			if valueLength < 3 {
+				return 0, fmt.Errorf("invalid value set for ephemeral-storage requests")
+			}
+			unit := value[valueLength-3:]
+			if unit != "GiB" && unit != "MiB" {
+				return 0, fmt.Errorf("invalid value set for ephemeral-storage requests")
```
trungng92, line 661: If possible, we should make the error measurement easier for the user to know what to do.
"I put a wrong value, but the error doesn't indicate what the value should be"

This comment might be irrelevant if we use the kubernetes resource library.

cluster-autoscaler/cloudprovider/oci/nodepools/consts/annotations.go
```diff
@@ -25,4 +25,7 @@ const (
 	// ToBeDeletedByClusterAutoscaler is the taint used to ensure that after a node has been called to be deleted
 	// no more pods will schedule onto it
 	ToBeDeletedByClusterAutoscaler = "ignore-taint.cluster-autoscaler.kubernetes.io/oke-impending-node-termination"
+
+	// EphemeralStorageSize is the freeform tag key that would be used to determine the ephemeral-storage size of the node
+	EphemeralStorageSize = "cluster-autoscaler/node-ephemeral-storage"
```
trungng92, line 30: Given that this is an OKE specific implementation, I'm on the fence whether we should make the name indicate that it's custom/OKE specific. 🤔

cluster-autoscaler/cloudprovider/oci/nodepools/oci_manager.go
vbhargav875: My thinking was that, since it is not mandatory for users to add the required freeform tag, so we could just leave it as an info log with higher verbosity level so it does not spam the logs. This would only be needed if customers want CA to spin up nodes with a specific ephemeral-storage size (especially when they want to scale up the pool from 0).

Let me know if you would still like me to make modifications here.

cluster-autoscaler/cloudprovider/oci/nodepools/oci_manager.go
```diff
@@ -634,6 +647,34 @@ func addTaintToSpec(node *apiv1.Node, taintKey string, effect apiv1.TaintEffect)
...
+			}
+			size, err := strconv.Atoi(value[0 : valueLength-3])
+			if err != nil {
+				return 0, err
+			}
+			if unit == "GiB" {
```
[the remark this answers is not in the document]
vbhargav875: Thank you, updated this.

cluster-autoscaler/cloudprovider/oci/nodepools/oci_manager.go
vbhargav875: Thanks, yes, I have made the changes to use the k8s library.

cluster-autoscaler/cloudprovider/oci/nodepools/oci_manager.go
trungng92: I think if a user is specifying a value, but it is not getting parsed correctly, I personally think we should error out so that the user can quickly know that something is wrong and correct it.

Otherwise, they will think everything is successful, even when it's not.

However, I could see arguments where it would be okay to warn the user. For example, the autoscaler is working, then the user mistypes a free form tag which will cause a failure, but doesn't see a failure until days later when a scale operation actually occurs. Or the fact that specifying one wrong value will cause scaling to not happen at all.

Overall, I'd be okay with erroring out or warning (but slightly prefer erroring).

cluster-autoscaler/cloudprovider/oci/nodepools/oci_manager.go
```diff
@@ -634,6 +647,26 @@ func addTaintToSpec(node *apiv1.Node, taintKey string, effect apiv1.TaintEffect)
...
+			klog.V(0).Infof("ephemeral-storage size set with value : %v", value)
+			value = strings.ReplaceAll(value, " ", "")
+
+			defer func() {
+				if err := recover(); err != nil {
+					klog.Errorf("Exception: %v", err)
+				}
+			}()
+
+			resourceSize := resource.MustParse(value)
```
trungng92, line 662: https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#ParseQuantity
We should use ParseQuantity instead of MustParse.

vbhargav875: Thanks, made the changes.

cluster-autoscaler/cloudprovider/oci/nodepools/oci_manager.go
vbhargav875: Agreed, thanks. I've changed the logic here.

trungng92:
Changes look good to me.

cluster-autoscaler/cloudprovider/oci/nodepools/consts/annotations.go
jlamillan, line 30: Why are we using free-form tags to provide local disk information? Is there something special about the scale from `0`  scenario that prevents the OKE and InstancePool implementations from using the `LocalDisksTotalSizeInGBs` and `LocalDisks` fields of  [core.shape](https://github.com/oracle/oci-go-sdk/blob/master/core/shape.go#L27) when populating the node template in `buildNodeFromTemplate`?

jlamillan: Assuming ephemeral storage, like other resources, can be determined from the shape, it still useful to optionally support using free-form tags to _override_ shape specific values like cpu, memory, ephemeral-storage, gpu, etc..

However, IMO, the values from [core.shape](https://github.com/oracle/oci-go-sdk/blob/master/core/shape.go#L27) should remain the _default_ source resource information for the node when available.

jlamillan: OK... after looking into this more, it looks like [most shapes](https://docs.oracle.com/en-us/iaas/Content/Compute/References/computeshapes.htm) use "block storage only". This means that _most of the time_ `LocalDisksTotalSizeInGBs` and `LocalDisks` will not be set. 

I still think it's a good idea for both the node-pool and instance-pool implementations to check for and use `shape.LocalDisksTotalSizeInGBs` from [core.shape](https://github.com/oracle/oci-go-sdk/blob/master/core/shape.go#L27) by default. However, given that it's going to be unset most of the time, it seems like a good idea to add the ability to receive these values via free-form tags.

cluster-autoscaler/cloudprovider/oci/nodepools/oci_manager.go
```diff
@@ -542,12 +545,22 @@ func (m *ociManagerImpl) buildNodeFromTemplate(nodePool *oke.NodePool) (*apiv1.N
 		})
 	}
 
+	freeformTags, err := m.ociTagsGetter.GetFreeformTags(nodePool)
```
jlamillan, line 548: Given the discussion in the [other thread](https://github.com/kubernetes/autoscaler/pull/6607#discussion_r1534710107), I would expect [our shape struct](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/oci/common/oci_shape.go#L53) to have a new field, say, `EphemeralStorageInGB`, which may or may not be set after the call to `m.ociShapeGetter.GetNodePoolShape(nodePool)` based on [disk info](https://docs.oracle.com/en-us/iaas/Content/Compute/References/computeshapes.htm) of the node shape.

It is at this point that `buildNodeFromTemplate()` would set the value of `node.Status.Capacity[apiv1.ResourceEphemeralStorage]` to either `shape.EphemeralStorageInGB` or the value from the free-form tag.

Does that make sense?

vbhargav875: We get the LocalDiskSize of a shape from the Core Service/Compute API - ([LINK](https://github.com/oracle/oci-go-sdk/blob/master/core/shape.go#L70)).
But this is not part of the [NodeShapeConfig](https://github.com/oracle/oci-go-sdk/blob/master/containerengine/node_shape_config.go).
We would need to work on adding the changes to oci-go-sdk so that the NodeShapeConfig has the LocalDiskSize, but this would be a whole api level change and would involve a lot of time and effort.

jlamillan: OK. Since there are _two_ implementations of this provider (and because the core API may evolve to offer ephemeral storage), I think it still makes sense to expose this attribute to both implementation using our [common.Shape struct](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/oci/common/oci_shape.go#L53) alongside the other capacity attributes (CPU, memory, GPU). 

So, for now, you still fetch the value from the free-form tag using a  `common.getEphemeralResourceRequestsInBytes(freeformTags)` function, but instead of setting a local / private variable, you'd set [shape.EphemeralStorageGB](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/oci/common/oci_shape.go#L53).


cluster-autoscaler/cloudprovider/oci/common/oci_shape.go

```
type Shape struct {
	Name             string
	CPU              float32
	GPU              int
	MemoryInBytes    float32
++	EphemeralStorageGB float32
}
```


As a followup MR, we'd need to ensure the _instancepool_ implementation is also using the `common.getEphemeralResourceRequestsInBytes(freeformTags)` function to check for ans set `shape.EphemeralStorageGB`

vbhargav875: Agreed. I have updated the PR with the above suggested changes.
Please take a look.

cluster-autoscaler/cloudprovider/oci/common/oci_tag.go
```diff
@@ -0,0 +1,27 @@
...
+
+package common
+
+import (
+	oke "k8s.io/autoscaler/cluster-autoscaler/cloudprovider/oci/vendor-internal/github.com/oracle/oci-go-sdk/v65/containerengine"
+)
+
+// TagsGetter returns the oci tags for the pool.
+type TagsGetter interface {
+	GetFreeformTags(*oke.NodePool) (map[string]string, error)
```
jlamillan, line 13: Since there will be two separate implementations of a `TagsGetter`, please rename this function to `GetNodePoolFreeformTags(*oke.NodePool) (map[string]string, error)`. 

Later, we can add a _second_ function to the interface called `GetInstancePoolFreeformTags(*core.InstancePool) (map[string]string, error)` and the associated implementation.

jlamillan (changes requested):
Other than the requested rename of `GetFreeformTags` to `GetNodePoolFreeformTags`, this change is looking good.

cluster-autoscaler/cloudprovider/oci/common/oci_tag.go
vbhargav875: Agreed, done.

jlamillan (approved):
Thanks Vijay.

/lgtm
/approve

PR #6613, opened 2024-03-11
Deprecate the Linode Cluster Autoscaler provider

### LINKED ISSUE #6495: Proposal: Deprecate Linode autoscaler implementation
Hey,
I'm from Linode/Akamai and wanted to propose we deprecate "our" [in-tree provider](https://github.com/kubernetes/autoscaler/tree/7e95c7e63a5f0773e51bc484b1f7dbe802fe668a/cluster-autoscaler/cloudprovider/linode). We have the following reasons for this:

1. This in-tree implementation duplicates [our official implementation](https://www.linode.com/blog/kubernetes/horizontal-cluster-autoscaling-on-linode-kubernetes-engine/) withing LKE (Linode Kubernetes Engine), which serves the very same purpose. (This implementation pre-dates our offering.) And it's not like this in-tree autoscaler can be used for bootstrapped clusters on bare VMs, this implementation is specific to our LKE offering. So a user is really deciding between enabling autoscaling in their existing nodepools and running their own autoscaler that does the very same thing.
2. There are [no OWNERS](https://github.com/kubernetes/autoscaler/blob/7e95c7e63a5f0773e51bc484b1f7dbe802fe668a/cluster-autoscaler/cloudprovider/linode/OWNERS) set for this codebase, which seems to be in violation of [the current policy](https://github.com/kubernetes/autoscaler/blob/7e95c7e63a5f0773e51bc484b1f7dbe802fe668a/cluster-autoscaler/cloudprovider/POLICY.md#in-tree-provider).
3. It's [largely unmaintained](https://github.com/kubernetes/autoscaler/commits/master/cluster-autoscaler/cloudprovider/linode).

The gist of it is that we're sending the wrong signal to users - this implementation doesn't allow for anything they don't already have within LKE. And since it only interacts with LKE, it doesn't provide any value to Linode Kubernetes users that don't leverage LKE.

Let me know what you think and if you agree, I'd love to know what our next steps are.

kokes:
#### What type of PR is this?

/kind deprecation

The Linode provider for CA has been redundant for years now and should be deprecated (and eventually removed).

Fixes #6495

```release-note
The Linode Cluster Autoscaler provider is removed.
```

None

### DIFF
--- a/cluster-autoscaler/cloudprovider/linode/README.md
+++ b/cluster-autoscaler/cloudprovider/linode/README.md
@@ -1,5 +1,9 @@
 # Cluster Autoscaler for Linode
 
+**⚠️ The Linode Cluster Autoscaler has been deprecated and will be removed some time in 2025. This implementation has only ever scaled the Linode Kubernetes Engine (LKE)
+deployments and [this product has had native autoscaling for years](https://www.linode.com/blog/kubernetes/horizontal-cluster-autoscaling-on-linode-kubernetes-engine/).
+This implementation is thus redundant.**
+
 The cluster autoscaler for Linode scales nodes in a LKE cluster.
 
 ## Linode Kubernetes Engine


elmiko:
been talking about this with @kokes  on slack, i think this is a solid first step towards removing this provider.

/lgtm

gjtempleton:
I think on past deprecations (though I now can't find which ones... @MaciekPytel can you remember which deprecation we followed this for?) we chose to match the [CLI deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-a-flag-or-cli) of marking as deprecated for 2 releases, so we could choose to be more specific and say that it'll only continue to function as far as the CA v1.32.x release?

mwielgus (approved):
/lgtm
/approve

PR #6616, opened 2024-03-12
Clarify VPA and HPA limitations when scaling on resource metrics

### LINKED ISSUE #6247: Can the VPA and HPA be used together on separate resource metrics?
I have found similar questions and answers in [other issues](https://github.com/kubernetes/autoscaler/issues/6060) and in the [multidimensional autoscaler AEP](https://github.com/kubernetes/autoscaler/blob/db5e83bc1afeaef6bfd3d69befef9c12ad469f94/multidimensional-pod-autoscaler/AEP.md#goals) however I am still unsure on the answer.

The [VPA limitations](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler#known-limitations) state the following:

> Vertical Pod Autoscaler should not be used with the [Horizontal Pod Autoscaler](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-resource-metrics) (HPA) on CPU or memory at this moment. However, you can use VPA with [HPA on custom and external metrics](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#scaling-on-custom-metrics).

This sounds to me like you cannot use VPA and HPA together on separate resource metrics, however I do not see why you couldn't use VPA to vertically scale only on memory and use HPA to horizontally scale only on CPU (as done by Google's [MultidimPodAutoscaler](https://github.com/kubernetes/autoscaler/blob/db5e83bc1afeaef6bfd3d69befef9c12ad469f94/multidimensional-pod-autoscaler/AEP.md#google-gkes-approach-of-mpa)). The multidimensional autoscaler AEP does state that for some applications this can [lead to memory being scaled down as the number of replicas are scaled up](https://github.com/kubernetes/autoscaler/blob/db5e83bc1afeaef6bfd3d69befef9c12ad469f94/multidimensional-pod-autoscaler/AEP.md#different-scaling-actions-for-different-types-of-resources), however this sounds like expected (and desired) behaviour and not something that would prevent VPA and HPA being used together.

From my limited testing, this seems to work great for Nginx Ingress Controller:

```yaml
kind: VerticalPodAutoscaler
metadata:
  name: ingress-nginx-controller
  namespace: ingress-nginx
spec:
  targetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: ingress-nginx-controller
  updatePolicy:
    updateMode: Auto
  resourcePolicy:
    containerPolicies:
    - containerName: "*"
      controlledResources:
      - memory
---
apiVersion: autoscaling/v2
kind: HorizontalPodAutoscaler
metadata:
  name: ingress-nginx-controller
  namespace: ingress-nginx
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: ingress-nginx-controller
  minReplicas: 2
  maxReplicas: 10
  metrics:
  - type: Resource
    resource:
      name: cpu
      target:
        type: Utilization
        averageUtilization: 60
```

The multidimensional autoscaler AEP describes why you cannot use VPA and HPA on the same resource metric and I can also see why the multidimensional autoscaler would provide more value than combining VPA and HPA (since it might be useful to vertically scale on CPU too), however until that is ready combining VPA and HPA on separate resource metrics is still very powerful.

If this is possible and there are not any known edge cases when doing this, I would be happy to create a PR to clarify this in the VPA limitations.

dippynark:
#### What type of PR is this?

/kind documentation

Clarifies whether it's possible to use VPA together with HPA on separate resource metrics

Fixes #6247

```release-note
NONE
```

### DIFF
--- a/vertical-pod-autoscaler/README.md
+++ b/vertical-pod-autoscaler/README.md
@@ -384,8 +384,12 @@ vpa-post-processor.kubernetes.io/{containerName}_integerCPU=true
   recreated. This can be partly
   addressed by using VPA together with [Cluster Autoscaler](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#basics).
 * VPA does not update resources of pods which are not run under a controller.
-* Vertical Pod Autoscaler **should not be used with the [Horizontal Pod Autoscaler](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-resource-metrics) (HPA) on CPU or memory** at this moment.
-  However, you can use VPA with [HPA on custom and external metrics](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#scaling-on-custom-metrics).
+* Vertical Pod Autoscaler **should not be used with the [Horizontal Pod
+  Autoscaler](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-resource-metrics)
+  (HPA) on the same resource metric (CPU or memory)** at this moment. However, you can use [VPA with
+  HPA on separate resource metrics](https://github.com/kubernetes/autoscaler/issues/6247) (e.g. VPA
+  on memory and HPA on CPU) as well as with [HPA on custom and external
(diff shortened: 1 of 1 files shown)


mwielgus (approved):
/lgtm
/approve

PR #6625, opened 2024-03-12
Updated VPA admission-controller to have adjustable minimum TLS version and TLS ciphers

### LINKED ISSUE #6624: For VPA, make the admission-controller expose the minimum TLS version and/or the TLS ciphers

Vertical Pod Autoscaler

At my company we have stricter TLS rules than normal.  Currently VPA's admission controller is accepting TLS ciphers for `ECDHE-RSA-DES-CBC3-SHA` and `DES-CBC3-SHA` which are considered to be weak ciphers.  We are being pinged for vulnerabilities when we use VPA.

With this feature, it would solve our issue as we could just have an allow-list of ciphers we accept or just bump the minimum TLS version to 1.3.

I would like to see two new flags exposed `--min-tls-version` and `--tls-ciphers` on the VPA admission-controller as it would allow for customers to adjust the server to their needs.

N/A
**Additional context.**:

allenmun197:
#### What type of PR is this?

/kind feature
This PR allows the VPA admission-controller to have an adjustable minimum TLS version and ciphersuites.  This will help people using VPA to be more secure as you can now set a minimum TLS version of 1.3 or only allow strong ciphers.

Fixes #6624

I don't know the best way to create a mapping of ciphers/tls_versions to convert the string from the flag to the accepted uint16 accepted by the server params.  If there is a better/cleaner solution I would be very much open to changing the code.

```release-note
NONE
```

```docs

```

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/README.md
+++ b/vertical-pod-autoscaler/pkg/admission-controller/README.md
@@ -32,6 +32,8 @@ up the changes: ```sudo systemctl restart kubelet.service```
    for pods on their creation & updates.
 1. You can specify a path for it to register as a part of the installation process
    by setting `--register-by-url=true` and passing `--webhook-address` and `--webhook-port`.
+1. You can specify a minimum TLS version with `--min-tls-version` with acceptable values being `tls1_2` (default), or `tls1_3`.
+1. You can also specify a comma or colon separated list of ciphers for the server to use with `--tls-ciphers` if `--min-tls-version` is set to `tls1_2`.
 
 ## Implementation
 
--- a/vertical-pod-autoscaler/pkg/admission-controller/config.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/config.go
@@ -19,6 +19,8 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"fmt"
+	"strings"
 	"time"
 
 	admissionregistration "k8s.io/api/admissionregistration/v1"
@@ -31,14 +33,43 @@ const (
 	webhookConfigName = "vpa-webhook-config"
 )
 
-func configTLS(serverCert, serverKey []byte) *tls.Config {
+func configTLS(serverCert, serverKey []byte, minTlsVersion, ciphers string) *tls.Config {
+	var tlsVersion uint16
+	var ciphersuites []uint16
+	reverseCipherMap := make(map[string]uint16)
 	sCert, err := tls.X509KeyPair(serverCert, serverKey)
 	if err != nil {
 		klog.Fatal(err)
 	}
+
+	for _, c := range tls.CipherSuites() {
+		reverseCipherMap[c.Name] = c.ID
+	}
+	for _, c := range strings.Split(strings.ReplaceAll(ciphers, ",", ":"), ":") {
+		cipher, ok := reverseCipherMap[c]
+		if ok {
+			ciphersuites = append(ciphersuites, cipher)
+		}
+	}
+	if len(ciphersuites) == 0 {
+		ciphersuites = nil
+	}
+
+	switch minTlsVersion {
+	case "":
+		fallthrough
+	case "tls1_2":
+		tlsVersion = tls.VersionTLS12
+	case "tls1_3":
+		tlsVersion = tls.VersionTLS13
+	default:
+		klog.Fatal(fmt.Errorf("Unable to determine value for --min-tls-version (%s), must be either tls1_2 or tls1_3", minTlsVersion))
+	}
+
 	return &tls.Config{
-		MinVersion:   tls.VersionTLS12,
+		MinVersion:   tlsVersion,
 		Certificates: []tls.Certificate{sCert},
+		CipherSuites: ciphersuites,
 	}
 }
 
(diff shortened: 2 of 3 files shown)


mwielgus (approved):
/lgtm
/approve

allenmun197:
@mwielgus fixed the formatting issue, should be good now

vertical-pod-autoscaler/pkg/admission-controller/README.md
```diff
@@ -32,6 +32,8 @@ up the changes: ```sudo systemctl restart kubelet.service```
    for pods on their creation & updates.
 1. You can specify a path for it to register as a part of the installation process
    by setting `--register-by-url=true` and passing `--webhook-address` and `--webhook-port`.
+1. You can specify a minimum TLS version with `--min-tls-version` with acceptable values being `tls1_0`, `tls1_1`, `tls1_2` (default), or `tls_13`.
```
voelzmo, line 35:
```suggestion
1. You can specify a minimum TLS version with `--min-tls-version` with acceptable values being `tls1_0`, `tls1_1`, `tls1_2` (default), or `tls1_3`.
```

vertical-pod-autoscaler/pkg/admission-controller/config.go
```diff
@@ -31,14 +32,43 @@ const (
...
+
+	switch minTlsVersion {
+	case "tls1_0":
+		tlsVersion = tls.VersionTLS10
+	case "tls1_1":
+		tlsVersion = tls.VersionTLS11
+	case "tls1_3":
+		tlsVersion = tls.VersionTLS13
+	default:
+		tlsVersion = tls.VersionTLS12
```
voelzmo, line 65: This is somewhat dangerous, IMHO. This code allows for arbitrary strings to be set for `--min-tls-version` and any typo will not result in an error, but in TLS1.2 being set. Users will never realize they're not getting what they intended. For example, if I'm setting `--min-tls-version=1.3`, admission-controller will silently configure TLS1.2. I think we should make it explicit to the user that they entered a value that is unknown and fail to start.

vertical-pod-autoscaler/pkg/admission-controller/config.go
```diff
@@ -31,14 +32,43 @@ const (
...
+		if ok {
+			ciphersuites = append(ciphersuites, cipher)
+		}
+	}
+	if len(ciphersuites) == 0 {
+		ciphersuites = nil
+	}
+
+	switch minTlsVersion {
+	case "tls1_0":
```
voelzmo, line 58: Do we have a good reason to "go back" and allow for less secure TLS configurations than we do now? If not, let's not include TLS1.0 and TLS1.1 here.

vertical-pod-autoscaler/pkg/admission-controller/main.go
```diff
@@ -59,6 +59,8 @@ var (
 		tlsCertFile:   flag.String("tls-cert-file", "/etc/tls-certs/serverCert.pem", "Path to server certificate PEM file."),
 		tlsPrivateKey: flag.String("tls-private-key", "/etc/tls-certs/serverKey.pem", "Path to server certificate key PEM file."),
 	}
+	ciphers       = flag.String("tls-ciphers", "", "A comma-separated or colon-separated list of ciphers to accept.  Only works when min-tls-version is set to tls1_2 or below.")
+	minTlsVersion = flag.String("min-tls-version", "tls1_2", "The minimum TLS version to accept.  Defaults to tls1_2.")
```
voelzmo, line 63: We should list all allowed values here, to make it clear which values are valid without taking a look at the implementation.

voelzmo (changes requested):
Thanks for immediately providing a PR for your issue! I have a few requests regarding the ux, where I think we can be a bit nicer to users.

vertical-pod-autoscaler/pkg/admission-controller/README.md
allenmun197: updated typo

vertical-pod-autoscaler/pkg/admission-controller/config.go
allenmun197: updated code to default to `tls1_2` and only accept `tls1_2` and `tls1_3`.  If a different value is given, we now throw a fatal exception.  Error now looks like
```
F0318 15:31:04.356258       1 config.go:66] Unable to determine value for --min-tls-version (tls1_1), must be either tls1_2 or tls1_3
```

vertical-pod-autoscaler/pkg/admission-controller/config.go
allenmun197: removed TLS 1.0 and TLS 1.1

vertical-pod-autoscaler/pkg/admission-controller/main.go
allenmun197: updated description

allenmun197:
@voelzmo Updated the PR, made the changes. LMK if there is anything else that needs to be updated.

voelzmo:
Looks great, thanks!
/lgtm

voelzmo (approved):

robbiezhang:
@voelzmo , is it fine to cherry-pick this change to vpa 1.1 release?

voelzmo:
@robbiezhang I don't think we cherry-pick new features to releases. We probably would rather release a 1.2.0
@kwiesmueller @raywainman WDYT?

robbiezhang:
@voelzmo thanks for the reply. That makes sense. Wondering what's the release cadence and 1.2 release timeline?

kwiesmueller:
Agreed. A new version would be necessary for features.

PR #6627, opened 2024-03-13
Add provreqOrchestrator that handle ProvReq classes

yaroslava-serdiuk:
#### What type of PR is this?
/kind feature

Part of implemenation of https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/proposals/provisioning-request.md

```release-note
NONE
```


cluster-autoscaler/provisioningrequest/checkcapacity/scaleup.go
```diff
@@ -28,76 +28,57 @@ import (
...
-	"k8s.io/klog/v2"
 
 	ca_processors "k8s.io/autoscaler/cluster-autoscaler/processors"
 	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
 )
 
-type provisioningRequestClient interface {
-	ProvisioningRequests() ([]*provreqwrapper.ProvisioningRequest, error)
-	ProvisioningRequest(namespace, name string) (*provreqwrapper.ProvisioningRequest, error)
+type scaleUpMode struct {
```
kisieland, line 41: `scaleUpMode` seems rather generic.
Maybe `checkCapacityOrchestratorMode` ?

cluster-autoscaler/provisioningrequest/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,149 @@
...
+	ca_errors "k8s.io/autoscaler/cluster-autoscaler/utils/errors"
+	"k8s.io/autoscaler/cluster-autoscaler/utils/taints"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
+
+	ca_processors "k8s.io/autoscaler/cluster-autoscaler/processors"
+	schedulerframework "k8s.io/kubernetes/pkg/scheduler/framework"
+)
+
+type scaleUpMode interface {
```
kisieland, line 45: ditto on the interface name.
Maybe `provisioningReqeustOrchestratorMode`

cluster-autoscaler/provisioningrequest/provreqclient/client.go
```diff
@@ -157,3 +162,26 @@ func newPodTemplatesLister(client *kubernetes.Clientset, stopChannel <-chan stru
 	klog.V(2).Info("Successful initial Pod Template sync")
 	return podTemplLister, nil
 }
+
+// VerifyProvisioningRequestClass check that all pods belong to one ProvisioningRequest that belongs to check-capacity ProvisioningRequst class.
+func VerifyProvisioningRequestClass(client ProvisioningRequestClient, unschedulablePods []*apiv1.Pod, className string) (*provreqwrapper.ProvisioningRequest, error) {
```
kisieland, line 167: Please add tests for this code

cluster-autoscaler/provisioningrequest/orchestrator/orchestrator.go
yaroslava-serdiuk: I think orchestrator should be in general one, however we have pods and ProvReq, so we have two different orchestrators and a wrapper one (3 in total), so different ProvClasses I introduced as different scaleUp modes. I'm ok name it provisioningRequestScaleUpMode, but IMO it's too big and scaleUpMode is actually is not limited to provreq.

kisieland (approved):
/lgtm

cluster-autoscaler/utils/errors/errors.go
```diff
@@ -81,6 +81,9 @@ func ToAutoscalerError(defaultType AutoscalerErrorType, err error) AutoscalerErr
 	if e, ok := err.(AutoscalerError); ok {
 		return e
 	}
+	if err == nil {
```
towca, line 84: Why?

cluster-autoscaler/provisioningrequest/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,149 @@
...
+	o.injector = scheduling.NewHintingSimulator(autoscalingContext.PredicateChecker)
+	for _, mode := range o.scaleUpModes {
+		mode.Initialize(autoscalingContext, processors, clusterStateRegistry, estimatorBuilder, taintConfig, o.injector)
+	}
+}
+
+// ScaleUp run ScaleUp for each Provisionining Class. As of now, CA pick one ProvisioningRequest,
+// so only one ProvisioningClass return non empty scaleUp result.
+// In case we implement multiple ProvisioningRequest ScaleUp, the function should return combined status
+func (o *provReqOrchestrator) ScaleUp(
```
towca, line 90: Please make this (and other similar places) one line: https://google.github.io/styleguide/go/guide.html#line-length.

cluster-autoscaler/provisioningrequest/checkcapacity/scaleup.go
```diff
@@ -0,0 +1,105 @@
...
+		return status.UpdateScaleUpError(&status.ScaleUpStatus{}, errors.NewAutoscalerError(errors.InternalError, "error during ScaleUp: %s", err.Error()))
+	}
+	if scaleUpIsSuccessful {
+		return &status.ScaleUpStatus{Result: status.ScaleUpSuccessful}, nil
+	}
+	return &status.ScaleUpStatus{Result: status.ScaleUpNoOptionsAvailable}, nil
+}
+
+// Assuming that all unschedulable pods comes from one ProvisioningRequest.
+func (o *checkCapacityScaleUpMode) scaleUp(unschedulablePods []*apiv1.Pod) (bool, error) {
```
towca, line 93: Why is this called `scaleUp`? It doesn't scale anything up

cluster-autoscaler/provisioningrequest/checkcapacity/scaleup.go
```diff
@@ -0,0 +1,105 @@
...
+	provReq, err := o.client.ProvisioningRequest(unschedulablePods[0].Namespace, unschedulablePods[0].OwnerReferences[0].Name)
```
towca, line 94: Why do we retrieve the ProvReq from the client again? Looks like we already have it when calling the method.

cluster-autoscaler/provisioningrequest/provreqclient/client.go
```diff
@@ -41,15 +41,21 @@ const (
 	provisioningRequestClientCallTimeout = 4 * time.Second
 )
 
-// ProvisioningRequestClient represents client for v1beta1 ProvReq CRD.
-type ProvisioningRequestClient struct {
+// ProvisioningRequestClient is an interface with methods for v1beta1 ProvReq CRD client.
+type ProvisioningRequestClient interface {
```
towca, line 45: Why make this an interface? Usually interfaces should be defined in consumer packages (like they were before this PR), not the implementer package. See e.g. https://medium.com/@mbinjamil/using-interfaces-in-go-the-right-way-99384bc69d39 for an explanation why

cluster-autoscaler/provisioningrequest/orchestrator/orchestrator.go
towca: I have a couple questions about the general delegation setup here:

1. Why is `WrapperOrchestrator` defined under `core` instead of `provisioningrequest`? It seems very ProvReq-specific.
2. Why do we want to have 2 levels of delegation here? Wouldn't it be easier to understand if we integrated delegating between ProvReq classes into `WrapperOrchestrator` (together with moving it under `provisioningrequest`)? The proposed setup is a ProvReq-specific Orchestrator delegating between the regular scale-up Orchestrator and another ProvReq-specific Orchestrator which then itself delegates between ProvReq-class-specific Orchestrators (which are called Modes for added confusion).
3. Is the only reason for introducing `scaleUpMode` not having to implement methods we don't care about (i.e. `ScaleUpToNodeGroupMinSize`)? If so, IMO `provReqOrchestrator` would be the best name. It's not a mode, it's a component performing something. Plus then the tie to the full `Orchestrator` interface is clear.
4. Have you thought about turning `WrapperOrchestrator` into a generic `DelegatingOrchestrator`? E.g. it would take a set of orchestrators and a set of functions that assign the pods to the orchestrators (e.g. by grouping by the `ProvisioningRequestPodAnnotationKey` value). And then we'd just have one Orchestrator per-ProvReq-class, and one level of delegation, handled under `core`. This would also make it easier to introduce new classes in downstream CA implementations.

towca (changes requested):

cluster-autoscaler/provisioningrequest/orchestrator/orchestrator.go
yaroslava-serdiuk: 1) Agree, make sense to move wrapper to provisioningrequest package.
2) My goal was to make code easier to understand. I would like to have each object doing a small part of logic. I.e wrapper orchestrator is doing split between regular pods and ProvReq pods and choose the appropriate orchestrator. The provreq orchestrator may have a different implementation for other cloud providers (for example if someone implement their own ProvReq class). The current implementation of ProvReq orchestrator doesn't care about specific ProvReq classes. 
3) Yes and no :) Sure, provReq orchestrator doesn't implement `ScaleUpToNodeGroupMinSize` method. Another reason to introduce scaleUpMode is because in my mental model orchestrator is smth major and we shouldn't have many orchestrators. I'm fine to change the name to provReqOrchestrator but I think there will be confusion between interface name and provreqorchestrator struct. 
4) Mostly addressed in 2). I wouldn't mix main provReqOrchestator and wrapperOrchestrator. 
>This would also make it easier to introduce new classes in downstream CA implementations
I don't see how it will make easier to introduce new classes, it should be the same as it is now, no?

towca: Re 2.:

I get aiming for more objects with narrower responsibilities, but hopping between abstraction levels also has a readability cost. And the responsibilities for `wrapperOrchestrator` and `provReqOrchestator` actually seem quite similar to me - delegate pods to dependent orchestrators based on some pod properties. So while reading I'm actually wondering why they're separate, plus the hop has its own cost.

There are 2 meaningful differences I see between them:
* `wrapperOrchestrator` splits the pods and delegates to one of them only, while `provReqOrchestrator` leaves 
the splitting to the dependent orchestrators, and calls all of them.
Do we actually the splitting/delegating semantics to be different between the orchestrators? Wouldn't `provReqOrchestrator` work if it called only one of the dependent orchestrators (e.g. in an LRU fashion), and then `wrapperOrchestrator` could handle both cases?
* `provReqOrchestrator` books the capacity for all dependent orchestrators. Wouldn't it be better to block that capacity also from the regular scale-up? Then again, `wrapperOrchestrator` could handle both cases.

Re 3.:

Are there differences in responsibilities/expected semantics between `scaleUpMode.ScaleUp()` and `Orchestrator.ScaleUp()`? If there aren't, I'd want to highlight that `scaleUpMode` is a subset of `Orchestrator`, with the same expectations etc. - in which case I'd try to stay as close to the original `Orchestrator` name as possible. If there are meaningful differences, I'd change the interface/methods/arguments/comment to highlight these differences, or at least make it not seem like it's a subset of `Orchestrator`.

In any case I'd remove the "Mode" part of the name. It suggests that this component changes (parts of) the logic of some other component (and not that some other component will delegate the logic to this component). There's nothing changing in `provReqOrchestrator` logic if you provide it with different modes. The overall behavior might change, but not the `provReqOrchestrator` logic itself, so the "Mode" part seems misleading to me.

Re 4.:

Right now `provReqOrchestrator` doesn't allow downstream implementations to change the set of modes at all, and if it did it would also make sense to export this interface. You also said that other cloud providers might need to implement their own `provReqOrchestrator` if they want to implement their own class. This is basically what I'm trying to avoid - I'd like to have a delegating orchestrator generic and robust enough for cloud providers to use it to implement their own classes by just implementing a specific orchestrator for their pods, and providing a function to determine which pods are theirs. With what's proposed, they'd have to re-implement their own `provReqOrchestrator`, and then plumb it into `wrapperOrchestrator` carefully (splitting the provreq pods between different prov-req specific orchestrators), right?

We can make `provReqOrchestrator` configurable enough to achieve the same effect, but then we have to define clear semantics for this interface that the cloud-provider-specific implementations can rely on.

cluster-autoscaler/provisioningrequest/orchestrator/orchestrator.go
yaroslava-serdiuk: The line will be too long. The style guide suggest "If a line feels too long, prefer refactoring instead of splitting it.", but refactoring would complicate this PR.
If we aim readability first, then the current format is more readable than the long line.

cluster-autoscaler/provisioningrequest/provreqclient/client.go
```diff
@@ -157,3 +163,32 @@ func newPodTemplatesLister(client *kubernetes.Clientset, stopChannel <-chan stru
 	klog.V(2).Info("Successful initial Pod Template sync")
 	return podTemplLister, nil
 }
+
+// VerifyProvisioningRequestClass check that all pods belong to one ProvisioningRequest that belongs to check-capacity ProvisioningRequst class.
+func VerifyProvisioningRequestClass(client ProvisioningRequestClient, unschedulablePods []*apiv1.Pod, className string) (*provreqwrapper.ProvisioningRequest, error) {
+	if len(unschedulablePods) == 0 {
```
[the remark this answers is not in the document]
yaroslava-serdiuk: Make sense. Added a check for empty list to ProvClass

cluster-autoscaler/provisioningrequest/orchestrator/orchestrator.go
yaroslava-serdiuk: Moved wrapper to provisioningrequest orchestrator package

towca (approved):

cluster-autoscaler/provisioningrequest/orchestrator/wrapper_orchestrator.go
```diff
@@ -41,20 +38,16 @@ import (
...
 	provReqOrchestrator scaleup.Orchestrator
 }
 
 // NewWrapperOrchestrator return WrapperOrchestrator
-func NewWrapperOrchestrator(kubeConfig *rest.Config) (scaleup.Orchestrator, error) {
-	provReqOrchestrator, err := checkcapacity.New(kubeConfig)
-	if err != nil {
-		return nil, fmt.Errorf("failed create ScaleUp orchestrator for ProvisioningRequests, error: %v", err)
-	}
+func NewWrapperOrchestrator(provReqOrchestrator scaleup.Orchestrator) *WrapperOrchestrator {
```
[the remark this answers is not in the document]
yaroslava-serdiuk: WrapperOrchestrator consist of podsOrchestrator and provReqOrchestrator, since there is only one option for podsOrchestrator (at least for now), it doesn't make sense to have a parameter for podsOrchestrator. However provReqOrchestrator could be different depends on which provReqClasses are supported, so I added a parameter for it.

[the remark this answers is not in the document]
towca: Fair enough, this makes sense. Thanks for the explanation!

towca:
Thanks for the change and addressing my comments!
/lgtm
/approve

yaroslava-serdiuk:
#6814

PR #6628, opened 2024-03-13
[clusterapi] Do not skip nodegroups with minSize=maxSize

### LINKED ISSUE #6549: cluster-api-autoscaler and enforce-node-group-min-size flag

Component version: 1.27.4

As was discussed in https://github.com/kubernetes/autoscaler/issues/5267#issuecomment-1861440996 user can specify an `enforce-node-group-min-size` for cluster autoscaler to enforce min nodes for the nodegroup.

But this solution won't work for clusterapi, if user specify minNodes = maxNodes, because of this check:
https://github.com/kubernetes/autoscaler/blob/a3c8978e119f56a90a6f8406e76685c21cfecd51/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_nodegroup.go#L357-L360

so if, for example, user adds the following annotations to his machineDeployment:
```
cluster.x-k8s.io/cluster-api-autoscaler-node-group-max-size: "2"
cluster.x-k8s.io/cluster-api-autoscaler-node-group-min-size: "2"
```
min size won't be enforced

@elmiko, pinging you in this issue because seems to be related to our previous discussion.

MaxFedotov:
#### What type of PR is this?

/kind feature

Fixes #6549 

```release-note
None
```

### DIFF
--- a/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller_test.go
+++ b/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller_test.go
@@ -1043,9 +1043,8 @@ func TestControllerNodeGroupForNodeWithPositiveScalingBounds(t *testing.T) {
 		if err != nil {
 			t.Fatalf("unexpected error: %v", err)
 		}
-		// We don't scale from 0 so nodes must belong to a
-		// nodegroup that has a scale size of at least 1.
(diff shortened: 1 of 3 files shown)


elmiko (commented):
i think this is what we want, and thanks for the tests too!

/lgtm
/approve

PR #6634, opened 2024-03-14
azure labels to skip in nodegroupset

gandhipr:
#### What type of PR is this?
feature - skip a few azure specific labels in nodegroupset

This PR add a few azure specific labels that needs to be skipped when checking for nodegroup similarity.
This change doesn't include moving azure specific lables under /azure directory - that chore will be covered in another change

Fixes #

```release-note
Skips Azure-specific node labels that might mistakenly categorize nodegroups as different when, in reality, they are similar.

```

```docs

```

### DIFF
--- a/cluster-autoscaler/processors/nodegroupset/azure_nodegroups.go
+++ b/cluster-autoscaler/processors/nodegroupset/azure_nodegroups.go
@@ -30,6 +30,16 @@ const AzureNodepoolLabel = "kubernetes.azure.com/agentpool"
 // AzureDiskTopologyKey is the topology key of Azure Disk CSI driver
 const AzureDiskTopologyKey = "topology.disk.csi.azure.com/zone"
 
+// Those labels are added on the VMSS and shouldn't affect nodepool similarity
+const aksEngineVersionLabel = "aksEngineVersion"
+const creationSource = "creationSource"
+const poolName = "poolName"
+const resourceNameSuffix = "resourceNameSuffix"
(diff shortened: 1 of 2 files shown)


x13n:
Change itself LGTM, though WDYT about moving this code to `cluster-autoscaler/cloudprovider/azure/processors`? You wouldn't require global CA approval that way.

/assign

rakechill (approved):

tallaxes (changes requested):
The changes themselves look good (may be a good idea to normalize capitalization of constants), but the code currently does not build, likely as a result of the package move; additional post-move work needed.

tallaxes (approved):
If we are keeping these outside of provider-specific space temporarily (for change sequencing reasons), and plan to move it the near future - let's add a comment to this effect in the description.

jackfrancis:
/lgtm

@MaciekPytel @gjtempleton @x13n would it make sense to schedule a follow-up item to move all of the cloud-provider-specific stuff in `cluster-autoscaler/processors/nodegroupset/` into the provider-specific code surface area so that this maintenance can be performed without having to loop in core cluster-autoscaler approvers?

x13n:
Yeah, that makes a lot of sense to me. There's no point having cloudprovider-specific processors in core CA dirs.

jackfrancis:
> Yeah, that makes a lot of sense to me. There's no point having cloudprovider-specific processors in core CA dirs.

Cool, once this gets and approval and merge I'll do that.

Issue #6601, opened 2024-03-08
cluster-autoscaler gets stuck with "Some nodes that failed to create were removed"


daimaxiaxie:
**Which component are you using?**:

cluster-autoscaler

**What version of the component are you using?**:

Component version: v1.28.0

**What k8s version are you using (`kubectl version`)?**:

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.16", GitCommit:"e37e4ab4cc8dcda84f1344dda47a97bb1927d074", GitTreeState:"clean", BuildDate:"2021-10-27T16:25:59Z", GoVersion:"go1.15.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19+", GitVersion:"v1.19.3-5", GitCommit:"dcc97265743078854c5328e30727147bdc5d1c37", GitTreeState:"clean", BuildDate:"2020-12-04T03:52:29Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What environment is this in?**:

aws via the aws cloud provider.

**What did you expect to happen?**:

I expect cluster-autoscaler to be able to scale ASGs up/down without error.

**What happened instead?**:

cluster-autoscaler is stuck main loop with the following error:
```
I0304 08:20:40.253243       1 static_autoscaler.go:287] Starting main loop
I0304 08:20:40.539180       1 mixed_nodeinfos_processor.go:205] GetNodeInfosForGroups: setup node infos for group size: 41
I0304 08:20:40.549177       1 clusterstate.go:1081] Found 5 instances with errorCode OutOfResource.placeholder-cannot-be-fulfilled in nodeGroup spot-32c128g-b
I0304 08:20:40.549203       1 clusterstate.go:1099] Failed adding 5 nodes (0 unseen previously) to group spot-32c128g-b due to OutOfResource.placeholder-cannot-be-fulfilled; errorMessages=[]string{"AWS cannot provision any more instances for this node group"}
I0304 08:20:40.549223       1 clusterstate.go:1081] Found 3 instances with errorCode OutOfResource.placeholder-cannot-be-fulfilled in nodeGroup spot-32c128g-c
[...]

PR #6650, opened 2024-03-22
Fix: Check MinSize in fix nodegroup size

daimaxiaxie:
#### What type of PR is this?

/kind bug

When an ASG out of stock and the instance is less than MinSize(especially spot), the entire Cluster Autoscaler will be stuck due to `deleteCreatedNodesWithErrors and fixNodeGroupSize ` return error.

Cluster Autoscaler cannot handle this situation well.

Fixes  https://github.com/kubernetes/autoscaler/issues/6601

Respect MinSize.

```release-note
NONE
```

```docs

```


MaciekPytel:
This PR fixes problem for ASGs, but at the same time breaks it for other cloudproviders. We generally don't assume min size is enforced at provider side (i.e. the expectation is that if CA decides to delete a node it can do so) and the current behavior of going below is the desired one. The reason is that there is no point keeping a VM that cannot register; it's not really providing capacity to kubernetes cluster. Min size is there to guarantee a minimum capacity of the cluster, but a node that cannot register shouldn't really be counted.

We need a more nuanced solution to the problem, one that allows CA to delete unregistered nodes below min size, while giving a chance to providers to handle it. IIRC we discussed extending cloudprovider to support this with @Bryce-Soghigian - do I remember correctly that you have a fix for this that you wanted to upstream?

MaciekPytel:
Actually a good question is why do we have such a check in https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/static_autoscaler.go#L785. That feels really inconsistent, why are we ok going below min in deleteCreatedNodesWithErrors, but not in removeOldUnregisteredNodes? 

Sure, in the case of deleteCreatedNodesWithErrors we're probably deleting some placeholders and not actual VMs, but what exactly those placeholders are would differ by provider and at the end of the day we're changing target size to be less than min size. If we say that's ok - and I think we should, as per my comment above the meaningful interpretation of min size is "minimum number of nodes" and a VM that cannot register is not really a node - we should do it in all paths where VMs fail to register as nodes. Thoughts @x13n @towca @BigDarkClown ?

Bryce-Soghigian:
Yep @MaciekPytel  https://github.com/kubernetes/autoscaler/pull/6658


The idea is that when the number of unregistered nodes is below min count, we won't ever remove them. This Draft PR will add a respectMinCount to the delete nodes interface that will allow us to make min count violating deletions for unregistered nodes.


I need to sit down and finish it. I wanted to upstream the change so that all cloud providers could benefit from it as well but that involved touching a lot more code thus taking longer than I thought it would :)

Bryce-Soghigian:
> Actually a good question is why do we have such a check in https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/static_autoscaler.go#L785. That feels really inconsistent, why are we ok going below min in deleteCreatedNodesWithErrors, but not in removeOldUnregisteredNodes?

Another thing to noted is not all cloudproviders enforce min count on DeleteNodes() level as well. Some of them rely on these higher level checks.(They shouldn't?)


>  the meaningful interpretation of min size is "minimum number of nodes" and a VM that cannot register is not really a node - we should do it in all paths where VMs fail to register as nodes.

For things like IsClusterHealthy, or IsNodegroupHealthy, the opposite is true. We only represent real nodes and not these partial vms (vms that never join apiserver) in the health signal. IMO if the bootstrap is failing its more indicative of a health problem than not ready nodes.

daimaxiaxie:
> Actually a good question is why do we have such a check in https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/static_autoscaler.go#L785. That feels really inconsistent, why are we ok going below min in deleteCreatedNodesWithErrors, but not in removeOldUnregisteredNodes?

I also have this question, this code comes from @kawych . I think `removeOldUnregisteredNodes` and `deleteCreatedNodesWithErrors` should be consistent because they face the same environment.

towca:
@daimaxiaxie I think the logic was the same even before @kawych changes FWIW.

@MaciekPytel I have no idea why the behavior is inconsistent between the 2 places, I agree that CA should be able to delete the VMs in both scenarios even if it means going below the MinSize.

@Bryce-Soghigian brought up a good point - if we expect the limit enforcement to happen outside fo CP, it should also mean that CPs shouldn't enforce the limits within `DeleteNodes`, right. @MaciekPytel for example the GCE cloud provider does so for example, should we change it?

Bryce-Soghigian:
One could also argue that having that cloud provider enforce min count in case some upper level logic is helpful. It means that no matter what we can't violate min count and scale beyond max count. 

That being said there are cases(like unregistered nodes) where we want to be able to force a deletion.

x13n:
I agree we should be able to go below min in case of long unregistered nodes and creation errors. There's no point in keeping such nodes around.

However, I'm not convinced complicating CA logic to pass the "ignore min size" flag down to cloud providers is the right thing to do. When CA is managing a node group, it should be the source of truth for determining group size. When CA tries to delete a node, cloud provider code should actuate this decision (or return an error). If we're allowing force delete in certain cases, the additional safeguard on cloud provider side isn't really going to provide a lot of additional protection I think? What is the use case for sometimes honoring min size internally in cloud provider code? Is it meant as extra layer of protection in case CA goes amok or is there more to it?

Bryce-Soghigian:
Yeah exactly. It serves as a safeguard if someone adds a new deletion call. Really, if we didn't have that cloudprovider code I could just delete the unregistered nodes outright. But with the cloudprovider safeguards, i cannot just delete the unregistered nodes. 


Since min count is a fundamental principle and people expect it to never be violated it makes sense to have that extra safety in the cloudprovider. But also I would like to be able to remove nodes that are bad. Either modifying that interface for DeleteNodes() or have a new method called UnsafeDeleteNodes()  might make sense.

Bryce-Soghigian:
Force makes a lot of sense lets do that instead.

x13n:
Thanks, so I think we can close this one then.

/close

daimaxiaxie:
Hi @x13n , I updated the code and now only respects MinSize in `fixNodeGroupSize`. Hope it can be reopened.

Respecting MinSize is necessary in `fixNodeGroupSize`, even with `ForceDeleteNodes`, because `fixNodeGroupSize` does not remove nodes. Thanks.

cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -728,6 +728,17 @@ func fixNodeGroupSize(context *context.AutoscalingContext, clusterStateRegistry
 					incorrectSize.ExpectedSize,
 					incorrectSize.CurrentSize,
 					delta)
+
+				possibleDelta := nodeGroup.MinSize() - incorrectSize.ExpectedSize
```
bskiba, line 732: Do we really want this check? MinSize is there to protect Cluster Autoscaler against scaling down below MinSize. If we want to fix the target size to sth below minSize, it means that the nodegroup started below minSize, we tried to scale up, but nodes are not registering correctly. I would expect that adjusting the target size of the group even to a value below minSize is something that we want to be able to do, we're not violating the minSize constraint, just reacting to a failed scale up situation.

daimaxiaxie: From the user's perspective, minsize is the minimum value of nodegroup target, and it cannot be less than minsize at any time.

On the other hand, many cloud vendors do not allow you to set a value lower than minsize.

x13n: I don't think this makes sense in CA core. For cloud providers that prevent target size setting below min, adjusting the size can happen on cloud provider side. Otherwise it is going to prevent cloud providers that do allow going below min from working correctly. Note that incorrect size in CSR may stay incorrect if target size is adjusted up to match min size, so the subsequent loops will run this code again.

better0332:
I also encountered this bug

PR #6659, opened 2024-03-26
fix alicloud target size

guopeng0:
#### What type of PR is this?
/kind bug

The targetSize function in alicloud is implemented by calling the Alibaba Cloud ESS API DescribeScalingGroups. In the function func (m *AliCloudManager) GetAsgSize(asgConfig *Asg) (int64, error), targetSize is calculated as sg.ActiveCapacity + sg.PendingCapacity. However, this approach poses certain risks and issues during actual execution.
﻿
ActiveCapacity represents the number of ECS instances that have successfully joined the scaling group and are running normally. PendingCapacity represents the number of ECS instances that are in the process of joining the scaling group but have not completed the relevant configurations. However, since ESS operates asynchronously, there is a risk of the sum of these two values not changing even after modifying the expected value, especially when desired is large.
﻿
The official documentation mentions that DesiredCapacity represents the desired number of ECS instances within the scaling group, which seems more reasonable.
﻿
During execution, I encountered this issue where despite modifying desired, the targetSize remained 0 at the start of the second onceOnce. However, a new value was retrieved during increase, resulting in an incorrect number of instances being scaled up, exceeding the expected value.

image
Added logs at key locations for easier localization.
NONE

```release-note

```

https://help.aliyun.com/zh/auto-scaling/developer-reference/api-describescalinggroups?spm=a2c4g.11186623.0.i3#doc-api-Ess-DescribeScalingGroups

### DIFF
--- a/cluster-autoscaler/cloudprovider/alicloud/alibaba-cloud-sdk-go/services/ess/describe_scaling_groups.go
+++ b/cluster-autoscaler/cloudprovider/alicloud/alibaba-cloud-sdk-go/services/ess/describe_scaling_groups.go
@@ -167,6 +167,7 @@ type ScalingGroup struct {
 	RemovingCapacity             int             `json:"RemovingCapacity" xml:"RemovingCapacity"`
 	ScalingGroupName             string          `json:"ScalingGroupName" xml:"ScalingGroupName"`
 	ActiveCapacity               int             `json:"ActiveCapacity" xml:"ActiveCapacity"`
+	DesiredCapacity              int             `json:"DesiredCapacity" xml:"DesiredCapacity"`
 	StandbyCapacity              int             `json:"StandbyCapacity" xml:"StandbyCapacity"`
 	ProtectedCapacity            int             `json:"ProtectedCapacity" xml:"ProtectedCapacity"`
 	ActiveScalingConfigurationId string          `json:"ActiveScalingConfigurationId" xml:"ActiveScalingConfigurationId"`
--- a/cluster-autoscaler/cloudprovider/alicloud/alicloud_manager.go
+++ b/cluster-autoscaler/cloudprovider/alicloud/alicloud_manager.go
@@ -107,7 +107,7 @@ func (m *AliCloudManager) GetAsgSize(asgConfig *Asg) (int64, error) {
 	if err != nil {
 		return -1, fmt.Errorf("failed to describe ASG %s,Because of %s", asgConfig.id, err.Error())
 	}
-	return int64(sg.ActiveCapacity + sg.PendingCapacity), nil
+	return int64(sg.DesiredCapacity), nil
 }
 
 // SetAsgSize sets ASG size.
--- a/cluster-autoscaler/cloudprovider/alicloud/alicloud_manager.go
+++ b/cluster-autoscaler/cloudprovider/alicloud/alicloud_manager.go
@@ -107,7 +107,7 @@ func (m *AliCloudManager) GetAsgSize(asgConfig *Asg) (int64, error) {
 	if err != nil {
 		return -1, fmt.Errorf("failed to describe ASG %s,Because of %s", asgConfig.id, err.Error())
 	}
-	return int64(sg.DesiredCapacity), nil
+	return int64(max(sg.DesiredCapacity, sg.ActiveCapacity + sg.PendingCapacity)), nil
 }
 
 // SetAsgSize sets ASG size.
--- a/cluster-autoscaler/cloudprovider/alicloud/alicloud_manager.go
+++ b/cluster-autoscaler/cloudprovider/alicloud/alicloud_manager.go
@@ -107,7 +107,7 @@ func (m *AliCloudManager) GetAsgSize(asgConfig *Asg) (int64, error) {
 	if err != nil {
 		return -1, fmt.Errorf("failed to describe ASG %s,Because of %s", asgConfig.id, err.Error())
 	}
-	return int64(max(sg.DesiredCapacity, sg.ActiveCapacity + sg.PendingCapacity)), nil
+	return int64(sg.DesiredCapacity), nil
 }
 
 // SetAsgSize sets ASG size.


guopeng0:
Hi @feiskyer
Please review this pull request. your valuable feedback is appreciated, and I will make changes as soon as possible.

daimaxiaxie:
I seem to have encountered a similar problem, repeated the scale up. The difference is that the second scale up reduced the TargetSize (maybe the pending pod has changed slightly). But I think the root cause is the same.

feiskyer:
@ringtail would you like to have a look?

Shubham82:
@ringtail @feiskyer `/lgtm` label above didn't apply, Could you please review it and approve it so it will merge?

Thanks!

x13n:
/assign @feiskyer 

Looks like @ringtail LGTMed, do you have any concerns for approving?

daimaxiaxie:
I encountered some situations in the process of applying this pr to the cluster. Users must enable `Enable Expected Number of Instances`(default disabled) for each asg in aliyun, otherwise `DesiredCapacity` = 0. Therefore, users must be reminded!!!!

guopeng0:
> I encountered some situations in the process of applying this pr to the cluster. Users must enable `Enable Expected Number of Instances`(default disabled) for each asg in ali, otherwise `DesiredCapacity` = 0. This is very dangerous!! Therefore, users must be reminded!!!!

Hi @daimaxiaxie 
Yes, indeed. The “Expected Number of Instances” feature needs to be enabled. Adding or removing nodes in the cluster autoscaler requires this feature. The current code on the master branch also relies on it, and this PR does not modify this logic.

daimaxiaxie:
But after disabled the "Expected number of instances", the current logic can still work normally. In other words, the current Cluster Autoscaler does not require 'Expected number of instances' feature. 

Therefore, the user is likely to have an asg with disabled feature. Once only the `DesiredCapacity` is taken, all instances will be destroyed. @guopeng0

cluster-autoscaler/cloudprovider/alicloud/alicloud_manager.go
```diff
@@ -107,7 +107,7 @@ func (m *AliCloudManager) GetAsgSize(asgConfig *Asg) (int64, error) {
 	if err != nil {
 		return -1, fmt.Errorf("failed to describe ASG %s,Because of %s", asgConfig.id, err.Error())
 	}
-	return int64(sg.ActiveCapacity + sg.PendingCapacity), nil
+	return int64(sg.DesiredCapacity), nil
```
daimaxiaxie, line 110: I think it should be compatible with both enabled 'Expected number of instances' and disabeld situations.

Maybe this is an option:
```suggestion
	return int64(max(sg.DesiredCapacity, sg.ActiveCapacity + sg.PendingCapacity)), nil
```

guopeng0: This is incorrect. Firstly, there are syntax errors. Additionally, configuring desiredSize is necessary. It should prompt rather than be compatible. Moreover, compatibility with downsizing could pose risks, so I don’t recommend this change.

cluster-autoscaler/cloudprovider/alicloud/alibaba-cloud-sdk-go/services/ess/describe_scaling_groups.go
```diff
@@ -167,6 +167,7 @@ type ScalingGroup struct {
 	RemovingCapacity             int             `json:"RemovingCapacity" xml:"RemovingCapacity"`
 	ScalingGroupName             string          `json:"ScalingGroupName" xml:"ScalingGroupName"`
 	ActiveCapacity               int             `json:"ActiveCapacity" xml:"ActiveCapacity"`
+	DesiredCapacity              int             `json:"DesiredCapacity" xml:"DesiredCapacity"`
```
daimaxiaxie, line 170:
```suggestion
	DesiredCapacity              int             `json:"DesiredCapacity" xml:"DesiredCapacity"`
	EnableDesiredCapacity        bool            `json:"EnableDesiredCapacity,omitempty" xml:"EnableDesiredCapacity,omitempty"`
```

`EnableEnableDesiredCapacity` required. 
https://next.api.aliyun.com/document/Ess/2014-08-28/DescribeScalingGroups

cluster-autoscaler/cloudprovider/alicloud/alicloud_manager.go
```diff
@@ -107,7 +107,7 @@ func (m *AliCloudManager) GetAsgSize(asgConfig *Asg) (int64, error) {
 	if err != nil {
 		return -1, fmt.Errorf("failed to describe ASG %s,Because of %s", asgConfig.id, err.Error())
 	}
```
daimaxiaxie, line 109:
```suggestion
	}
	if !sg.EnableDesiredCapacity {
	    klog.Warningf("The Expected Number of Instances of ASG %s is disabled", asgConfig.id)
	    return int64(sg.ActiveCapacity + sg.PendingCapacity), nil
	}
```

This is basically consistent with our code. 
But we have done more compatibility(*max*). If you don't want to, I think this is enough.

daimaxiaxie: Feel free to share your thoughts.

daimaxiaxie (changes requested):

IrisIris:
After reading the above information,  I believe this issue is discussing how to make cluster-autoscaler compatible with ACK desired capacity, right? There were good discussions, but this problem can't be solved by simply modifying the way to get autoscaling group size（ i.e. the implement of GetAsgSize). First, I'd like to expalin the reasons why it‘ll  not work:
1)  Desired capacity is not a required configuration for the AliCloud autoscaling group. When an autoscaling group's desired capacity is not enabled, the desired capacity would be zero even many nodes has been created in the group.  So it's not ok, directly treated the desired capacity as the nodes number.
2) Actually the reason why autoscaling group size is different is mostly due to the delay of executing scaling when desired size enbaled. AliCloud Ess has suppoted a more efficient and comaptiable way to solve this delay.
I'd like to give a new PR to fix this problem this week.

daimaxiaxie:
Yes, solving the problem depends on user enabled DesiredCapacity.

PR #6665, opened 2024-03-28
Add ability to the VPA admission-controller to reload it's certificate v2

### LINKED ISSUE #6272: VPA admission controller certificate is not reloaded

**What version of the component are you using?**: 0.13.0

Component version: 0.13.0

**What k8s version are you using (`kubectl version`)?**: 1.24

$ kubectl version

</pre></details>

**What environment is this in?**: AWS EKS

**What did you expect to happen?**: When using cert-manager to generate the secret containing TLS certs for the VPA admission controller, updates to the secret should trigger a reload so that the VPA admission controller is still available.

**What happened instead?**: When using cert-manager to generate the secret containing TLS certs for the VPA admission controller, updates to the secret result in the VPA admission controller reporting a bad certificate.

**How to reproduce it (as minimally and precisely as possible)**: See above

Re-filing this as a +1 to https://github.com/kubernetes/autoscaler/issues/3458.
I see that https://github.com/kubernetes/autoscaler/pull/3462 was opened but eventually closed.

Nuckal777:
#### What type of PR is this?

/kind feature

Adds the ability for the admission-controller to reload it's certificate on changes to the certificate file. This is useful when provisioning certificates with cert-manager.

Fixes #6272.

```release-note
Added `reload-cert` flag to the VPA admission-controller, which will reload the certificates it use, when they change.
```

None.

### DIFF
--- a/vertical-pod-autoscaler/pkg/admission-controller/certs.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/certs.go
@@ -17,17 +17,18 @@ limitations under the License.
 package main
 
 import (
+	"crypto/tls"
 	"os"
+	"path"
+	"sync"
 
+	"github.com/fsnotify/fsnotify"
 	"k8s.io/klog/v2"
 )
 
-type certsContainer struct {
-	caCert, serverKey, serverCert []byte
-}
-
 type certsConfig struct {
 	clientCaFile, tlsCertFile, tlsPrivateKey *string
+	reload                                   *bool
 }
 
 func readFile(filePath string) []byte {
@@ -41,10 +42,59 @@ func readFile(filePath string) []byte {
 	return res
 }
 
-func initCerts(config certsConfig) certsContainer {
-	res := certsContainer{}
-	res.caCert = readFile(*config.clientCaFile)
-	res.serverCert = readFile(*config.tlsCertFile)
-	res.serverKey = readFile(*config.tlsPrivateKey)
-	return res
+type CertReloader struct {
+	tlsCertPath string
+	tlsKeyPath  string
+	cert        *tls.Certificate
+	mu          sync.RWMutex
+}
+
+func (cr *CertReloader) Start(stop <-chan struct{}) error {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return err
+	}
+
+	if err = watcher.Add(path.Dir(cr.tlsCertPath)); err != nil {
+		return err
+	}
+	if err = watcher.Add(path.Dir(cr.tlsKeyPath)); err != nil {
+		return err
+	}
+	go func() {
+		defer watcher.Close()
+		for {
+			select {
+			case event := <-watcher.Events:
+				if event.Has(fsnotify.Create) || event.Has(fsnotify.Write) {
+					klog.V(2).Info("New certificate found, reloading")
+					if err := cr.Load(); err != nil {
+						klog.Errorf("Failed to reload certificate: %s", err)
+					}
+				}
+			case err := <-watcher.Errors:
+				klog.Warningf("Error watching certificate files: %s", err)
+			case <-stop:
+				return
+			}
+		}
+	}()
+	return nil
+}
+
+func (cr *CertReloader) Load() error {
+	cert, err := tls.LoadX509KeyPair(cr.tlsCertPath, cr.tlsKeyPath)
+	if err != nil {
+		return err
+	}
+	cr.mu.Lock()
+	defer cr.mu.Unlock()
+	cr.cert = &cert
+	return nil
+}
+
+func (cr *CertReloader) GetCertificate(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
+	cr.mu.RLock()
+	defer cr.mu.RUnlock()
+	return cr.cert, nil
 }
--- a/vertical-pod-autoscaler/pkg/admission-controller/certs_test.go
+++ b/vertical-pod-autoscaler/pkg/admission-controller/certs_test.go
@@ -1,0 +1,154 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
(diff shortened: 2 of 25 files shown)


Nuckal777:
/assign krzysied kwiesmueller

Would appreciate a review.

vertical-pod-autoscaler/pkg/admission-controller/certs.go
```diff
@@ -41,10 +42,59 @@ func readFile(filePath string) []byte {
...
+	if err = watcher.Add(path.Dir(cr.tlsCertPath)); err != nil {
+		return err
+	}
+	if err = watcher.Add(path.Dir(cr.tlsKeyPath)); err != nil {
+		return err
+	}
+	go func() {
+		defer watcher.Close()
+		for {
+			select {
```
sophieliu15, line 67: Do we need to add a default case here?

vertical-pod-autoscaler/pkg/admission-controller/certs_test.go
```diff
@@ -0,0 +1,154 @@
...
+	keyPath := path.Join(tempDir, "cert.key")
+	if err = os.WriteFile(keyPath, privateKey, 0666); err != nil {
+		t.Error(err)
+	}
+
+	reloader := certReloader{
+		tlsCertPath: certPath,
+		tlsKeyPath:  keyPath,
+	}
+	if err != nil {
```
sophieliu15, line 129: It seems you already handled the error in line 121. Should this `err` variable be assigned to something else before you checking if it is `nil` in this line?

Nuckal777: The check is superfluous.

vertical-pod-autoscaler/pkg/admission-controller/certs.go
Nuckal777: I think that no default case is required. A goroutine is forked off, which runs the blocking select. The channels are filled by the fsnotify implementation. Do you have a use case for a default case?

vertical-pod-autoscaler/pkg/admission-controller/certs_test.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	err = pem.Encode(caFile, &pem.Block{
+		Type:  "CERTIFICATE",
+		Bytes: caBytes,
+	})
+	if err != nil {
+		t.Error(err)
+	}
+
+	generateCerts := func(org string) ([]byte, []byte) {
```
kwiesmueller, line 69: I would prefer this to be a named function as inline functions tend to make the code more complex and harder to read.

vertical-pod-autoscaler/pkg/admission-controller/certs.go
```diff
@@ -41,10 +42,59 @@ func readFile(filePath string) []byte {
...
+	if err != nil {
+		return err
+	}
+	cr.mu.Lock()
+	defer cr.mu.Unlock()
+	cr.cert = &cert
+	return nil
+}
+
+func (cr *certReloader) getCertificate(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
```
kwiesmueller, line 96: Why is the ClientHelloInfo needed?

vertical-pod-autoscaler/pkg/admission-controller/certs_test.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	pub, privateKey = generateCerts("second")
+	if err = os.WriteFile(certPath, pub, 0666); err != nil {
+		t.Error(err)
+	}
+	if err = os.WriteFile(keyPath, privateKey, 0666); err != nil {
+		t.Error(err)
+	}
+	time.Sleep(40 * time.Millisecond)
```
kwiesmueller, line 137: Sleeps in tests can cause issues with flakiness.
Is there a way to test this without needing a sleep?

What do you think about calling getCertificate in a loop until it passes (or the test times out?), could also set a custom timeout/max number of tries there.

Nuckal777: Fixed.

Nuckal777:
I hope, I addressed everything. Did I miss something?

kwiesmueller (changes requested):
Generally lgtm.
Sorry for not bringing this up earlier. Do you think you can add an e2e/integration test for this feature?

Nuckal777:
Would generating a second pair of certificates in `gencerts.sh` and swapping the certificates in a test be fine?

kwiesmueller:
That sounds good to me, yes.
As long as we do it within the integration cluster so we verify the configmap reloading within the pod happens correctly when the configmap changes.

Nuckal777:
Should we run `selfRegistration` to update the webhook on reload as well, if enabled?

Nuckal777:
I had some thought about regarding the self registration again. I'd like to use the VPA together with cert-manager. This means the `MutatingAdmissionWebhook` needs to have the [annotation for cert-managers ca-injection](https://cert-manager.io/docs/concepts/ca-injector/). Likely, this implies the webhook needs to deployed by something else than the admission-controller. So, the self registration needs to be disabled in that setup. cert-manager would take of the CA in the webhook and the certificates in the secret. The admission-controller would just need to reload the leaf certificates, which is implemented in the PR.

Given the self registration would not be helpful in that setup, I think we don't need to touch it here. Any opinions?

voelzmo:
I'm fine with not touching the self-registration here. I also think it is unlikely that the webhook can be deployed with self-registgration if you have a mechanism which automatically renews the certificate. So this is good to go from my side.
/lgtm

@kwiesmueller do you have any additional changes you want to see in this PR?

kwiesmueller (approved):
/lgtm
/approve

kwiesmueller:
Thanks for adding the test!

Issue #4337, opened 2021-09-16
Daemonset Eviction during Scale down


rangarb885:
Hello : 
I want to understand how to handle the below scenario with CA. 

1. When EKS CA decides to scale down a node (which is a part of managed node-group) which has a daemonset like fluent-bit (shipping logs from apps) & SignalFx (tracing & metrics), what configuration i need to have on CA to make sure that daemonset are not evicted as app may be using this during this scaling down time (under their graceful timeout window).?

2. Is there a config on CA setup to skip the daemonset eviction and allow them to run until the node is terminated. 
I am good even if these daemonset are not gracefully stopped as app using them are gracefully stopping (with their own graceful shutdown timeouts) 

My current CA configuration (EKS1.21)

Image : `k8s.gcr.io/autoscaling/cluster-autoscaler:v1.21.0`

```
            "./cluster-autoscaler",
            "--v=2",
            "--stderrthreshold=2",
            "--cloud-provider=aws",
            "--scan-interval=10s",
            "--skip-nodes-with-local-storage=false",
            "--aws-use-static-instance-list=true",
            "--expander=least-waste",
            "--node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,k8s.io/cluster-autoscaler/${var.cluster_name}"
```

Thank you
Balaji

jim-barber-he:
Hi.

If I'm right then I believe setting the new `--daemonset-eviction-for-occupied-nodes=false` parameter that was introduced in version `1.22` of the cluster-autoscaler might handle the scenario where the daemonsets aren't stopped at all and will not be gracefully stopped but killed when the node terminates.

I'm adding to this case because it is similar enough to the request I was about to make.
I would like a way to have the daemonsets stop gracefully, but only after all the "normal" pods have completed.
As far as I can tell there is no way to do that?

We have application deployments with a preStop lifecycle to make sure they complete their work before they shutdown.
We also have supporting daemonsets such as the node-local-dns, kiam, and fluent-bit running to provide DNS; AWS IAM access; and logging services for the application pods.
However when cluster autoscaler chooses a node to scale in, the daemonsets are terminated before the application pods have finished running, resulting in various errors (such as not being able to resolve hostnames for example).

In order to prove that the daemonsets were being terminated too early I set up a test.
I created a dedicated instance group, and then the following:
- A daemonset called `testing-daemonset`
- A deployment called `testing-balloon` configured like so:
    - 2 replicas
[...]

PR #6666, opened 2024-03-28
fix daemonset eviction

jfcoz:
evict daemonset pods only on empty nodes by default

#### What type of PR is this?

/kind bug

/kind regression

Change the default for daemonset pods eviction : 
- do not evict daemonset pods on non-empty nodes
- evict daemonset pods on empty nodes

Fixes https://github.com/kubernetes/autoscaler/issues/5240
Fixes https://github.com/kubernetes-sigs/aws-fsx-csi-driver/issues/253
Fixes https://github.com/kubernetes/autoscaler/issues/4337

```release-note
NONE
```

```docs

```


x13n:
This can be a breaking change to anyone depending on current defaulting. At the same time, if anyone indeed prefers different behavior, they can just flip the flag themselves. In fact, this can now be solved using `--drain-priority-config` flag to evict lower priority pods first (assuming DS are higher priority which generally is a reasonable setup).

/close

PR #6667, opened 2024-03-28
Refactor estimation

kisieland:
#### What type of PR is this?
/kind feature

Introduces optimization to the binpacking estimator, which changes the time complexity from O(pods * nodes) to O(pods) in cases of homogeneous pods.

The PR also introduces a benchmark that shows 5x improvement:
The previous version took ~25s:
```
goos: linux
goarch: amd64
pkg: k8s.io/autoscaler/cluster-autoscaler/estimator
cpu: AMD EPYC 7B12
BenchmarkBinpackingEstimate-48    	       1	25418848308 ns/op	15018690016 B/op	144793817 allocs/op
PASS
ok  	k8s.io/autoscaler/cluster-autoscaler/estimator	25.488s
```
The new version takes less than 5s:
```
goos: linux
goarch: amd64
pkg: k8s.io/autoscaler/cluster-autoscaler/estimator
cpu: AMD EPYC 7B12
BenchmarkBinpackingEstimate-48    	       1	4805105872 ns/op	2023739088 B/op	 3574976 allocs/op
PASS
ok  	k8s.io/autoscaler/cluster-autoscaler/estimator	4.869s
```

Split the PR into two commits:
First one introduces a required pod-equivalence abstraction
Second one uses this new fact in the estimation

For now I kept the logic in one function in the estimation to make clear what changes were made, but I am happy to split it into further functions.

No

NA

### DIFF
--- a/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
+++ b/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
@@ -696,10 +696,10 @@ func (o *ScaleUpOrchestrator) ComputeSimilarNodeGroups(
 func matchingSchedulablePodGroups(podGroups []estimator.PodEquivalenceGroup, similarPodGroups []estimator.PodEquivalenceGroup) bool {
 	schedulableSamplePods := make(map[*apiv1.Pod]bool)
 	for _, podGroup := range similarPodGroups {
-		schedulableSamplePods[podGroup.Pods[0]] = true
+		schedulableSamplePods[podGroup.Exemplar()] = true
 	}
 	for _, podGroup := range podGroups {
-		if _, found := schedulableSamplePods[podGroup.Pods[0]]; !found {
+		if _, found := schedulableSamplePods[podGroup.Exemplar()]; !found {
 			return false
 		}
 	}
--- a/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
+++ b/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
@@ -138,15 +138,15 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 	}
 
 	// Calculate expansion options
-	schedulablePods := map[string][]*apiv1.Pod{}
+	schedulablePodGroups := map[string][]estimator.PodEquivalenceGroup{}
 	var options []expander.Option
 
 	for _, nodeGroup := range validNodeGroups {
-		schedulablePods[nodeGroup.Id()] = o.SchedulablePods(podEquivalenceGroups, nodeGroup, nodeInfos[nodeGroup.Id()])
+		schedulablePodGroups[nodeGroup.Id()] = o.SchedulablePodGroups(podEquivalenceGroups, nodeGroup, nodeInfos[nodeGroup.Id()])
 	}
 
 	for _, nodeGroup := range validNodeGroups {
-		option := o.ComputeExpansionOption(nodeGroup, schedulablePods, nodeInfos, len(nodes)+len(upcomingNodes), now)
+		option := o.ComputeExpansionOption(nodeGroup, schedulablePodGroups, nodeInfos, len(nodes)+len(upcomingNodes), now)
 		o.processors.BinpackingLimiter.MarkProcessed(o.autoscalingContext, nodeGroup.Id())
 
 		if len(option.Pods) == 0 || option.NodeCount == 0 {
@@ -195,14 +195,14 @@ func (o *ScaleUpOrchestrator) ScaleUp(
 	createNodeGroupResults := make([]nodegroups.CreateNodeGroupResult, 0)
 	if !bestOption.NodeGroup.Exist() {
 		var scaleUpStatus *status.ScaleUpStatus
-		createNodeGroupResults, scaleUpStatus, aErr = o.CreateNodeGroup(bestOption, nodeInfos, schedulablePods, podEquivalenceGroups, daemonSets)
+		createNodeGroupResults, scaleUpStatus, aErr = o.CreateNodeGroup(bestOption, nodeInfos, schedulablePodGroups, podEquivalenceGroups, daemonSets)
 		if aErr != nil {
 			return scaleUpStatus, aErr
 		}
 	}
 
 	// Recompute similar node groups in case they need to be updated
-	bestOption.SimilarNodeGroups = o.ComputeSimilarNodeGroups(bestOption.NodeGroup, nodeInfos, schedulablePods, now)
+	bestOption.SimilarNodeGroups = o.ComputeSimilarNodeGroups(bestOption.NodeGroup, nodeInfos, schedulablePodGroups, now)
 	if bestOption.SimilarNodeGroups != nil {
 		// if similar node groups are found, log about them
 		similarNodeGroupIds := make([]string, 0)
@@ -440,20 +440,20 @@ func (o *ScaleUpOrchestrator) filterValidScaleUpNodeGroups(
 // ComputeExpansionOption computes expansion option based on pending pods and cluster state.
 func (o *ScaleUpOrchestrator) ComputeExpansionOption(
 	nodeGroup cloudprovider.NodeGroup,
-	schedulablePods map[string][]*apiv1.Pod,
+	schedulablePodGroups map[string][]estimator.PodEquivalenceGroup,
 	nodeInfos map[string]*schedulerframework.NodeInfo,
 	currentNodeCount int,
 	now time.Time,
 ) expander.Option {
 	option := expander.Option{NodeGroup: nodeGroup}
-	pods := schedulablePods[nodeGroup.Id()]
+	podGroups := schedulablePodGroups[nodeGroup.Id()]
 	nodeInfo := nodeInfos[nodeGroup.Id()]
 
-	if len(pods) == 0 {
+	if len(podGroups) == 0 {
 		return option
 	}
 
-	option.SimilarNodeGroups = o.ComputeSimilarNodeGroups(nodeGroup, nodeInfos, schedulablePods, now)
+	option.SimilarNodeGroups = o.ComputeSimilarNodeGroups(nodeGroup, nodeInfos, schedulablePodGroups, now)
 
 	estimateStart := time.Now()
 	expansionEstimator := o.estimatorBuilder(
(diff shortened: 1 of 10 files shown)


cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
```diff
@@ -683,22 +685,26 @@ func (o *ScaleUpOrchestrator) ComputeSimilarNodeGroups(
...
 
 	return validSimilarNodeGroups
 }
 
-func matchingSchedulablePods(groupSchedulablePods []*apiv1.Pod, similarSchedulablePods []*apiv1.Pod) bool {
+func matchingSchedulablePods(podGroups []estimator.PodEquivalenceGroup, similarPodGroups []estimator.PodEquivalenceGroup) bool {
 	schedulablePods := make(map[*apiv1.Pod]bool)
-	for _, pod := range similarSchedulablePods {
-		schedulablePods[pod] = true
+	for _, podGroup := range similarPodGroups {
```
x13n, line 698: Can we just compare groups not the pods in them? The groups should be immutable, so iterating over all pods seems wasteful.

cluster-autoscaler/estimator/binpacking_estimator.go
```diff
@@ -89,14 +89,19 @@ func (e *BinpackingNodeEstimator) Estimate(
 	scheduledPods := []*apiv1.Pod{}
 	lastNodeName := ""
 
-	for _, pod := range pods {
-		found := false
+	for _, podsEquivalenceGroup := range podsEquivalenceGroups {
```
x13n, line 92: Can you split this into smaller functions? It's pretty long right now, long nested loops are hard to read and there's a bit of code duplication that perhaps could be avoided.

cluster-autoscaler/estimator/binpacking_estimator.go
```diff
@@ -105,47 +110,66 @@ func (e *BinpackingNodeEstimator) Estimate(
...
-			// on a new node either. There is no point adding more nodes to snapshot in such case, especially because of
-			// performance cost each extra node adds to future FitsAnyNodeMatching calls.
-			if lastNodeName != "" && !newNodesWithPods[lastNodeName] {
-				continue
+		for ; index < len(podsEquivalenceGroup.Pods); index++ {
+			pod := podsEquivalenceGroup.Pods[index]
+			found := false
+
+			// Check schedulability on only newly created node
+			nodeName, err := e.predicateChecker.FitsAnyNodeMatching(e.clusterSnapshot, pod, func(nodeInfo *schedulerframework.NodeInfo) bool {
```
x13n, line 118: Since we only want to check one node here, perhaps it would suffice to call `CheckPredicates` instead? That'll avoid listing all nodes from cluster snapshot just to pick one of them later on. Also, no point in running this check when `lastNodeName` is empty I think?

cluster-autoscaler/estimator/binpacking_estimator.go
```diff
@@ -105,47 +110,66 @@ func (e *BinpackingNodeEstimator) Estimate(
...
+				newNodeNameIndex++
+				newNodeNames[newNodeName] = true
+				lastNodeName = newNodeName
+
+				// And try to schedule pod to it.
+				// Note that this may still fail (ex. if topology spreading with zonal topologyKey is used);
+				// in this case we can't help the pending pod. We keep the node in clusterSnapshot to avoid
+				// adding and removing node to snapshot for each such pod.
+				if err := e.predicateChecker.CheckPredicates(e.clusterSnapshot, pod, newNodeName); err != nil {
+					continue
```
x13n, line 164: Shouldn't we skip the entire group in this case? If one pod failed, what's the point of trying another identical pod?

cluster-autoscaler/estimator/binpacking_estimator_test.go
```diff
@@ -79,10 +79,12 @@ func makePods(cpuPerPod int64, memoryPerPod int64, hostport int32, maxSkew int32
...
 		pods = append(pods, pod)
 	}
-	return pods
+	return PodEquivalenceGroup{
+		Pods: pods,
+	}
 }
 
-func makeNode(cpu int64, mem int64, name string, zone string) *apiv1.Node {
+func makeNode(cpu, mem, podCount int64, name string, zone string) *apiv1.Node {
```
x13n, line 87: Similar to the above. Could this use `BuildTestNode` instead? https://github.com/kubernetes/autoscaler/blob/7184d23e62367c57b2f0d66203a5d7c7a7eb661e/cluster-autoscaler/utils/test/test_utils.go#L215

Missing params can be added via option pattern. Right now it is hard to tell what `makeNode(1, 2, 3, "4", "5")` is doing without referring back here.

cluster-autoscaler/estimator/binpacking_estimator_test.go
```diff
@@ -32,7 +32,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func makePods(cpuPerPod int64, memoryPerPod int64, hostport int32, maxSkew int32, topologySpreadingKey string, podCount int) []*apiv1.Pod {
+func makePodEquivalenceGroup(cpuPerPod int64, memoryPerPod int64, hostport int32, maxSkew int32, topologySpreadingKey string, podCount int) PodEquivalenceGroup {
```
x13n, line 35: I know you're just modifying an existing function, but could this just accept a pod and count to be composed with `BuildTestPod`? https://github.com/kubernetes/autoscaler/blob/7184d23e62367c57b2f0d66203a5d7c7a7eb661e/cluster-autoscaler/utils/test/test_utils.go#L36

For instance: `makePodEquivalenceGroup(BuildTestPod(...), 10)`? Missing params can be added via option pattern.

x13n (changes requested):

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
kisieland: Made the function take a look only on the sample pod (i.e. the first pod in the slice)

cluster-autoscaler/estimator/binpacking_estimator_test.go
kisieland: kept the `BuildTestPod` hidden as a call inside in `makePodEquivalenceGroup`
But this simplifies the code quite.

cluster-autoscaler/estimator/binpacking_estimator.go
kisieland: Make sense, rewrote.

cluster-autoscaler/estimator/binpacking_estimator.go
kisieland: Done.

cluster-autoscaler/estimator/binpacking_estimator_test.go
kisieland: Done

cluster-autoscaler/estimator/binpacking_estimator.go
```diff
@@ -68,49 +78,98 @@ func NewBinpackingNodeEstimator(
...
-	newNodeNameIndex := 0
 	scheduledPods := []*apiv1.Pod{}
-	lastNodeName := ""
+	for _, podsEquivalenceGroup := range podsEquivalenceGroups {
+		var err error
+		var remainingPods []*apiv1.Pod
+
+		scheduledPods, remainingPods, err = e.tryToScheduleOnExistingNodes(scheduledPods, podsEquivalenceGroup.Pods)
+		if err != nil {
+			klog.Errorf("%v", err)
```
x13n, line 104: nit: This is equivalent to `klog.Errorf(err.Error())`, just with unnecessary formatting. Same below.

cluster-autoscaler/estimator/binpacking_estimator.go
```diff
@@ -68,49 +78,98 @@ func NewBinpackingNodeEstimator(
...
-		if err == nil {
-			found = true
-			if err := e.clusterSnapshot.AddPod(pod, nodeName); err != nil {
-				klog.Errorf("Error adding pod %v.%v to node %v in ClusterSnapshot; %v", pod.Namespace, pod.Name, nodeName, err)
-				return 0, nil
+		if err != nil {
+			break
+		}
+
+		if err := e.clusterSnapshot.AddPod(pod, nodeName); err != nil {
```
x13n, line 137: This pattern repeats in 3 places: add pod to cluster snapshot, append to `scheduledPods`, update `newNodesWithPods`. Perhaps worth extracting to a function?

cluster-autoscaler/estimator/binpacking_estimator.go
```diff
@@ -124,49 +183,45 @@ func (e *BinpackingNodeEstimator) Estimate(
...
-			if err != nil {
-				klog.Errorf("Error while adding new node for template to ClusterSnapshot; %v", err)
-				return 0, nil
+			if err := e.addNewNodeToSnapshot(nodeTemplate); err != nil {
+				return nil, fmt.Errorf("Error while adding new node for template to ClusterSnapshot; %w", err)
 			}
-			newNodeNameIndex++
-			newNodeNames[newNodeName] = true
-			lastNodeName = newNodeName
+			e.newNodeNames[e.lastNodeName] = true
```
x13n, line 189: nit: make it a part of `addNewNodeToSnapshot`?

cluster-autoscaler/estimator/binpacking_estimator_test.go
x13n: I think with 6 unnamed variables, mostly of the same type, it is hard to read test cases without referring back here. Hence my proposal to explicitly call `BuildTestPod` there - the options are named, is it is clear what they do.

cluster-autoscaler/estimator/binpacking_estimator_test.go
x13n: Is it? I still see `makeNode` not `BuildTestNode` here.

x13n (changes requested):

cluster-autoscaler/estimator/binpacking_estimator.go
```diff
@@ -36,6 +36,12 @@ type BinpackingNodeEstimator struct {
 	podOrderer             EstimationPodOrderer
 	context                EstimationContext
 	estimationAnalyserFunc EstimationAnalyserFunc // optional
+
+	// Helper variables to avoid coping them in each function
+	newNodeNameIndex int
```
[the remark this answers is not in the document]
kisieland: SG, done

cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
kisieland: Added, though there were only few calls added to this function.

cluster-autoscaler/estimator/binpacking_estimator_test.go
kisieland: Updated

cluster-autoscaler/estimator/binpacking_estimator_test.go
kisieland: Updated to use `BuildTestNode` directly.

PR #6670, opened 2024-03-29
Cleanup ProvReq wrapper

yaroslava-serdiuk:
#### What type of PR is this?

/kind cleanup

```release-note
NONE
```


towca:
/lgtm
/approve
/hold

The change itself LGTM, but I'm wondering about the motivation behind it. It seems like the wrapper was designed as a translation layer between different ProvReq API versions. Do you think this won't be needed after all? What happens when we introduce v1 ProvReq?

yaroslava-serdiuk:
/unhold

API is unlikely to change much, but there are a lot of code duplication in current version of wrapper.

yaroslava-serdiuk:
#6814

PR #6672, opened 2024-04-01
[Azure] Don't crash when vmss not present or has no nodes

rakechill:
#### What type of PR is this?

Add one of the following kinds:
/kind bug

#### Testing
* One unit test in `azure_manager_test.go` that acts as an acceptance test.
* One e2e test (in-progress)

We currently crash CAS if there are no nodes or if VMSS (Azure `NodeGroup` implementation) is not present. Ideally, we'd like to avoid crashing in most situations and instead log.

**Main problem was in BuildAzure (call path shown below)**:
* [cloudprovider.BuildAzure](https://github.com/kubernetes/autoscaler/blob/9b41b9dc5bf431a4ad657c044e809b15cc28a7e2/cluster-autoscaler/cloudprovider/azure/azure_cloud_provider.go#L206) --> `klog.Fatalf("Failed to create Azure Manager: %v", err`
* [CreateAzureManager](https://github.com/kubernetes/autoscaler/blob/9b41b9dc5bf431a4ad657c044e809b15cc28a7e2/cluster-autoscaler/cloudprovider/azure/azure_manager.go#L138) --> returns forceRefresh err
* [azureManager.forceRefresh](https://github.com/kubernetes/autoscaler/blob/9b41b9dc5bf431a4ad657c044e809b15cc28a7e2/cluster-autoscaler/cloudprovider/azure/azure_manager.go#L208) --> `klog.Errorf("Failed to regenerate Azure cache: %v", err), return err`
* [azureCache.regenerate](https://github.com/kubernetes/autoscaler/blob/9b41b9dc5bf431a4ad657c044e809b15cc28a7e2/cluster-autoscaler/cloudprovider/azure/azure_cache.go#L169-L171) --> return ng.Nodes() error
* [scaleSet.Nodes](https://github.com/kubernetes/autoscaler/blob/9b41b9dc5bf431a4ad657c044e809b15cc28a7e2/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go#L651-L655) --> `klog.Errorf("Failed to get current size for vmss %q: %v", scaleSet.Name, err), return err`
* [scaleSet.getCurSize](https://github.com/kubernetes/autoscaler/blob/9b41b9dc5bf431a4ad657c044e809b15cc28a7e2/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go#L189-L194) --> `klog.Errorf("failed to get information for VMSS: %s, error: %v", scaleSet.Name, err), return err`
* [getVMSSFromCache](https://github.com/kubernetes/autoscaler/blob/9b41b9dc5bf431a4ad657c044e809b15cc28a7e2/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go#L179-L181) --> `return compute.VirtualMachineScaleSet{}, fmt.Errorf("could not find vmss: %s", scaleSet.Name)`

**Main updates include:**
* updating `Nodes()` to not error if there are zero nodes, but rather log and return an empty list of instances
* updated `GetOptions()` to not error
* updated `regenerate` to handle `instances == nil`

**Key exceptions found**:
* `TemplateNodeInfo()` errors are handled by the core code. Typically, if the node group is not found, core logs a warning and skips the node group. The only exception is for `GetNodeFromTemplate()`. If `TemplateNodeInfo()` returns an error, this case is not handled since it's required to spin up a new node.
* `GetOptions()` should not error, because if options are nil, we just use default options. GCE + AWS both return nil error here.


Shubham82:
@rakechill, you have to sign the CLA before the PR can be reviewed.

Shubham82:
To check EasyCLA

/easycla

rakechill:
converted back to a draft since I need to address the azure scale set tests
cc @gandhipr @jackfrancis who were added as reviewers. will address the tests tonight and re-publish the PR.

cluster-autoscaler/cloudprovider/azure/azure_cache.go
```diff
@@ -120,6 +120,10 @@ func (m *azureCache) regenerate() error {
 		if err != nil {
 			return err
 		}
+		if instances == nil {
+			klog.V(4).Infof("regenerate: no nodes found for node group %s", ng.Id())
```
Bryce-Soghigian, line 124: Is default for the chart set to v(3)? If so are we ok with not seeing this for most clusters?

x13n:
/ok-to-test
/assign @gandhipr @jackfrancis 

Can you take a look?

jackfrancis:
@rakechill is this ready for final review?

cluster-autoscaler/cloudprovider/azure/azure_cache.go
jackfrancis: This seems appropriate as a debug-level log event (and matches its complement at L118)

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -186,7 +189,13 @@ func (scaleSet *ScaleSet) getCurSize() (int64, error) {
 
 // GetScaleSetSize gets Scale Set size.
 func (scaleSet *ScaleSet) GetScaleSetSize() (int64, error) {
-	return scaleSet.getCurSize()
+	// First, get the size of the ScaleSet reported by API
```
jackfrancis, line 192: Why are we ever in the flow of "get me the size of a scaleset" before the cache has a record of the scaleset? Would it be possible to enforce more serial sanity here, so that we populate the cache for newly observed scalesets (either literally new scalelets, or e.g., the cluster-autoscaler runtime has restarted and so has to repopulate its cache) as part of the flow of retrieving details about it (e.g., its size).

I think we definitely want to continue having all GET flows derive data from the cache, just wondering if we can improve the behavior of "cache miss" flows so that we don't have these weird edge cases that may slow down eventual consistency.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -163,7 +163,8 @@ func (scaleSet *ScaleSet) Autoprovisioned() bool {
 func (scaleSet *ScaleSet) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
 	template, err := scaleSet.getVMSSFromCache()
 	if err != nil {
-		return nil, err
+		klog.Errorf("failed to get information for VMSS: %s", scaleSet.Name)
```
rakechill, line 166: `GetOptions` shouldn't error. Failure to get info for VMSS should result in defaulting instead. Erroring here results in extra logs from core that we don't need.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -227,7 +228,7 @@ func (scaleSet *ScaleSet) getCurSize() (int64, error) {
 		set, rerr = scaleSet.manager.azClient.virtualMachineScaleSetsClient.Get(ctx, scaleSet.manager.config.ResourceGroup, scaleSet.Name)
 		if rerr != nil {
 			klog.Errorf("failed to get information for VMSS: %s, error: %v", scaleSet.Name, rerr)
-			return -1, err
+			return -1, rerr.Error()
```
rakechill, line 231: Open to discussion: I updated this to `rerr.Error()`, because I feel like we should be returning the error from the compute RP rather than the `err` from the `getVMSSFromCache` call.
cc @comtalyst who made this `rerr` update during cloud-provider-azure refactor.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -254,9 +255,8 @@ func (scaleSet *ScaleSet) getScaleSetSize() (int64, error) {
 	size, err := scaleSet.getCurSize()
 	if size == -1 || err != nil {
 		klog.V(3).Infof("getScaleSetSize: either size is -1 (actual: %d) or error exists (actual err:%v)", size, err)
-		return size, err
 	}
-	return size, nil
+	return size, err
```
rakechill, line 259: This is just the same. Just removing the extra line of code, since we want to return `err` always (even if it's nil)

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -651,7 +651,7 @@ func (scaleSet *ScaleSet) Nodes() ([]cloudprovider.Instance, error) {
 	curSize, err := scaleSet.getCurSize()
 	if err != nil {
 		klog.Errorf("Failed to get current size for vmss %q: %v", scaleSet.Name, err)
-		return nil, err
+		return nil, nil // Don't return error if VMSS not found
```
rakechill, line 654: This prevents us from failing to create AzureManager if one vmss is not found/not yet created.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
tallaxes: Is this true of `GetOptions` in general? In which chase we can remove the error from the return. Otherwise, let's add a comment in code to explain why the error is not returned.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
tallaxes: Could getCurSize return other kinds of errors, beside not found?

tallaxes (commented):
Overall looks good, left a couple of questions. Has this been tested in a real no-VMSS scenario?

rakechill:
> Overall looks good, left a couple of questions. Has this been tested in a real no-VMSS scenario?

Running in self-hosted, we see:
>
image

Expected errors (e.g. not able to get info for node group if vmss not found) are present, but CAS doesn't crash and is able to initialize AzureManager.

We were seeing errors like this previously:
```
// Entire cache regeneration would fail due to even on vmss not found
azure_manager.go:221] Failed to regenerate Azure cache: could not find vmss

// Entire manager would fail creation due to even one vmss not found
// This line leads to a log.Fatalf()
azure_cloud_provider.go:202] Failed to create Azure Manager: could not find vmss
```

These failures seem to be retried every 5-6 min and continuously loop through this until a manual intervention (e.g. manual scale up on aks nodepool).

What is still unknown: will this ever recover? How does AKS know to recreate another nodepool if the creation of the first one failed.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
rakechill: That is actually a fair call. 

There is another error scenario where we fail to fetch latest info from VMSS client for spot instances:
https://github.com/kubernetes/autoscaler/blob/9b41b9dc5bf431a4ad657c044e809b15cc28a7e2/cluster-autoscaler/cloudprovider/azure/azure_scale_set.go#L221-L233

I need to think more on how to differentiate between the two. An ugly solution would be to return -2 for this one...lol

A better solution may be to have different error types to switch on:
1. not found via VMSS cache
2. not found via VMSS client

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
rakechill: I linked some more details about this in the PR description, but we have GetOptions() interface-level function definition: https://github.com/kubernetes/autoscaler/blob/9b41b9dc5bf431a4ad657c044e809b15cc28a7e2/cluster-autoscaler/cloudprovider/cloud_provider.go#L241-L244

**If options aren't found (e.g. == nil), then core will use default options.**

---
### AWS + GCE implementations
```
// AWS
func (ng *AwsNodeGroup) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
	if ng.asg == nil || ng.asg.Tags == nil || len(ng.asg.Tags) == 0 {
		return &defaults, nil
	}
	return ng.awsManager.GetAsgOptions(*ng.asg, defaults), nil
}

// GCE
func (mig *gceMig) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
	return mig.gceManager.GetMigOptions(mig, defaults), nil
}
```
---
### Without this change 
If instead we error due to vmss not found, then we will get more `vmss not found` logs than we need to. For the sake of moving this bug fix forward, I'm happy to revert this part and return to it later.

**Callers of GetOptions**
Refs from deletion/scale down:
  * static autoscaler: uses default if err + opts == nil; no panic, **errors** 
  * group deletion scheduler: **fails if err != nil**, sets empty if opts == nil, **err not used**
  * budgets: **err logged**
  * scaledown planner: **err logged**
  * scaledown set processor: **errors**

Refs from scaleup:
  * orchestrator scaleup:  **err not used**
  * node_group_config_processor: **errors**

tallaxes: No need to revert, let's just add a comment in the code (rather than PR comment) to explain why the error is not returned.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
tallaxes: I am not sure we need to treat these two cases differently, do we?

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -227,7 +232,8 @@ func (scaleSet *ScaleSet) getCurSize() (int64, error) {
 		set, rerr = scaleSet.manager.azClient.virtualMachineScaleSetsClient.Get(ctx, scaleSet.manager.config.ResourceGroup, scaleSet.Name)
 		if rerr != nil {
 			klog.Errorf("failed to get information for VMSS: %s, error: %v", scaleSet.Name, rerr)
-			return -1, err
+			notFound := rerr.IsNotFound()
+			return -1, newGetVMSSFailedError(rerr.Error(), notFound)
```
tallaxes, line 236: Can compress to

```suggestion
			return -1, newGetVMSSFailedError(rerr.Error(), rerr.IsNotFound())
```

Bryce-Soghigian:
Ack, will review by EOD

tallaxes (approved):
LGTM. It feels like there may be a cleaner way to represent not found errors (maybe with error wrapping), but your implementation should work as well.

cluster-autoscaler/cloudprovider/azure/azure_cache.go
```diff
@@ -170,6 +170,10 @@ func (m *azureCache) regenerate() error {
...
+			return nil
```
comtalyst, line 175: Should we return here?
It looks like this loop is going to each group. Not sure if we want to halt the entire loop/method when one group is empty. Maybe `continue` is more appropriate?

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -163,7 +163,12 @@ func (scaleSet *ScaleSet) Autoprovisioned() bool {
 func (scaleSet *ScaleSet) GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error) {
 	template, err := scaleSet.getVMSSFromCache()
 	if err != nil {
-		return nil, err
+		klog.Errorf("failed to get information for VMSS: %s", scaleSet.Name)
+		// Note: We don't return an error here and instead accept defaults.
+		// Every invocation of GetOptions() returns an error if this condition is met:
+		// `if err != nil && err != cloudprovider.ErrNotImplemented`
+		// The error return value is intended to only capture unimplemented.
+		return nil, nil
```
comtalyst, line 171: What do you think about returning `cloudprovider.ErrNotImplemented` instead?
It is true that we don't implement this, and the handling will be the same as just returning `nil, nil`. Just with clearer visual.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
```diff
@@ -648,10 +652,13 @@ func (scaleSet *ScaleSet) TemplateNodeInfo() (*schedulerframework.NodeInfo, erro
...
 func (scaleSet *ScaleSet) Nodes() ([]cloudprovider.Instance, error) {
-	curSize, err := scaleSet.getCurSize()
-	if err != nil {
-		klog.Errorf("Failed to get current size for vmss %q: %v", scaleSet.Name, err)
-		return nil, err
+	curSize, getVMSSError := scaleSet.getCurSize()
+	if getVMSSError != nil {
+		klog.Errorf("Failed to get current size for vmss %q: %v", scaleSet.Name, getVMSSError.error)
+		if getVMSSError.notFound {
+			return nil, nil // Don't return error if VMSS not found
```
comtalyst, line 659: Should we return empty array instead of `nil`?

cluster-autoscaler/cloudprovider/azure/azure_cache.go
rakechill: yeah agreed that makes more sense

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
rakechill: Hm... I don't think that's exactly accurate. `ErrNotImplemented` signifies that we don't implement GetOptions() at all which isn't the case. We just were unable to get options, because we were unable to get the scaleset. 

The function comment here explains what nil will do:
>  Returning a nil will result in using default options.

cluster-autoscaler/cloudprovider/azure/azure_scale_set.go
rakechill: hm I'm not 100% sure. We don't always follow the same convention here in our codebase wrt empty vs. nil slice return values. 
fwiw this random blog I found suggests nil: https://blog.boot.dev/golang/empty-vs-nil-slices-golang/ and the code previously used nil.

tallaxes: What do other providers do?

rakechill: * [azure vms pool inits a slice (and thus would return empty)](https://github.com/kubernetes/autoscaler/blob/8c71ebfe2e16262c27adef61c6285cd9c9361dc6/cluster-autoscaler/cloudprovider/azure/azure_vms_pool.go#L156)
* [aws inits a slice (and thus would return empty)](https://github.com/kubernetes/autoscaler/blob/8c71ebfe2e16262c27adef61c6285cd9c9361dc6/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go#L369C2-L369C60)
* [gce inits a slice (and thus would return empty)](https://github.com/kubernetes/autoscaler/blob/8c71ebfe2e16262c27adef61c6285cd9c9361dc6/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_nodegroup.go#L235)

think robin's idea is better then, will refactor to do something similar

cluster-autoscaler/cloudprovider/azure/azure_cache.go
comtalyst: Yea, I think we must not return here. It could effectively disable autoscaling on other node groups. Let's be extra careful on the decision to return early.
Although, I feel like unit tests should have caught it. Do you think it is useful to add one in this case?

cluster-autoscaler/cloudprovider/azure/azure_cache.go
```diff
@@ -170,7 +170,7 @@ func (m *azureCache) regenerate() error {
 		if err != nil {
 			return err
 		}
-		klog.V(4).Infof("regenerate: found nodes for node group %s: %+v", ng.Id(), instances)
+		klog.V(4).Infof("regenerate: found %d nodes for node group %s: %+v", ng.Id(), len(instances), instances)
```
rakechill, line 173: Updated this to 1) not return early when there are no nodes and 2) updated the log message to include # of nodes found.

comtalyst:
/lgtm
/approve
/hold

Unhold whenever you are ready. And I think this is something we should cherry-pick to all minor versions. Do you agree?

cluster-autoscaler/cloudprovider/azure/azure_cache.go
comtalyst, line 173: The order of variables here is a bit incorrect...

comtalyst:
```suggestion
		klog.V(4).Infof("regenerate: found %d nodes for node group %s: %+v", len(instances), ng.Id(), instances)
```

comtalyst (changes requested):
(per previous comment)

comtalyst:
Will be using https://github.com/kubernetes/autoscaler/pull/7708 instead.

rakechill:
Closing since @comtalyst copied this to another PR (no permissions to edit this PR since it's from my personal fork)

PR #6677, opened 2024-04-02
fix: scale up broken for providers not implementing NodeGroup.GetOptions()

### LINKED ISSUE #6676: Scale-up broken for Cloud Providers not implementing NodeGroup.GetOptions()
**Which component are you using?**: `cluster-autoscaler`

Component version: 1.28.0+

**What k8s version are you using (`kubectl version`)?**: 1.28.0+

**What environment is this in?**: Hetzner Cloud, kubespray

When pods are pending, a scale up should happen.

**What happened instead?**:

On scale up, the user sees the following error message and no nodes are created:

```
orchestrator.go:446] Couldn't get autoscaling options for ng: pool1
```

Unclear

- Reported on Slack: https://kubernetes.slack.com/archives/C01FCM1943Y/p1708699690894189

This error message was added in (#5826) `1.28.0`: https://github.com/kubernetes/autoscaler/pull/5826/files#diff-95d98f54ac57fc3c60c23c1daca659d1fb752fc1822d32e42e12614b96331183R444-R447

The PR calls `NodeGroup.GetOptions`. Providers are allowed to return `errNotImplemented` from this, and the calling code should handle them. This is not being done in the current code and it fails instead. I have fixed a similar issue (on scale down) in #6037 & #6038.

apricote:
#### What type of PR is this?

/kind bug
/kind documentation

Properly handle calls to `NodeGroup.GetOptions()` that return `cloudprovider.ErrNotImplemented` in the scale up path.

Fixes #6676

All call sites check for `options != nil` right after the error check.

```release-note
Fixed scaling up node groups for Hetzner Provider
```

I have only confirmed that this was actually affecting the Hetzner Provider, there are other providers that also return `ErrNotImplemented`, but I did not verify if there was user visible breakage for them.

```docs

```

### DIFF
--- a/cluster-autoscaler/cloudprovider/cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/cloud_provider.go
@@ -243,7 +243,7 @@ type NodeGroup interface {
 
 	// GetOptions returns NodeGroupAutoscalingOptions that should be used for this particular
 	// NodeGroup. Returning a nil will result in using default options.
-	// Implementation optional.
+	// Implementation optional. Callers MUST handle `cloudprovider.ErrNotImplemented`.
 	GetOptions(defaults config.NodeGroupAutoscalingOptions) (*config.NodeGroupAutoscalingOptions, error)
 }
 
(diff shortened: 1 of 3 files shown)


towca:
Thanks for reporting this!

I'm wondering if we could just change all of the cloud providers that return `nil, cloudprovider.ErrNotImplemented` to return `nil, nil`? That seems safer going forward, as the callers wouldn't have to remember to handle `cloudprovider.ErrNotImplemented`, and would preserve the semantics since returning `nil` options is supposed to fall back to the default options anyway.

@gjtempleton @MaciekPytel WDYT?

towca:
@MaciekPytel volunteered for the long-term fix, merging this before that happens.

/lgtm
/approve

